PR #2796, opened 2018-11-30
Properly bypass connection limits for cluster peers

nbougalis:
Tag: @mDuo13 for potential documentation update

### DIFF
--- a/src/ripple/peerfinder/impl/Logic.h
+++ b/src/ripple/peerfinder/impl/Logic.h
@@ -392,6 +392,11 @@ public:
         if (keys_.find (key) != keys_.end())
             return Result::duplicate;
 
+        // If the peer belongs to a cluster, update the slot to reflect that.
+        counts_.remove (*slot);
+        slot->cluster (cluster);
+        counts_.add (*slot);
+
         // See if we have an open space for this slot
         if (! counts_.can_activate (*slot))
         {
@@ -401,18 +406,15 @@ public:
             return Result::full;
         }
 
-        // Set key and cluster right before adding to the map
-        // otherwise we could assert later when erasing the key.
-        counts_.remove (*slot);
+        // Set the key right before adding to the map, otherwise we might
+        // assert later when erasing the key.
         slot->public_key (key);
-        slot->cluster (cluster);
-        counts_.add (*slot);
-
-        // Add the public key to the active set
-        auto const result = keys_.insert (key);
-        // Public key must not already exist
-        assert (result.second);
-        (void) result.second;
+        {
+            auto const result = keys_.insert(key);
+            // Public key must not already exist
+            assert (result.second);
+            (void)result.second;
+        }
 
         // Change state and update counts
         counts_.remove (*slot);


src/ripple/peerfinder/impl/Logic.h
```diff
@@ -401,18 +406,15 @@ class Logic
...
-        slot->cluster (cluster);
-        counts_.add (*slot);
-
-        // Add the public key to the active set
-        auto const result = keys_.insert (key);
-        // Public key must not already exist
-        assert (result.second);
-        (void) result.second;
+        {
+            auto const result = keys_.insert(key);
```
thejohnfreeman: Do we enforce "one space between function/macro name and parenthesized argument list"?

src/ripple/peerfinder/impl/Logic.h
```diff
@@ -401,18 +406,15 @@ class Logic
...
+            // Public key must not already exist
+            assert (result.second);
+            (void)result.second;
```
thejohnfreeman: What about space after C-style cast?

src/ripple/peerfinder/impl/Logic.h
nbougalis: We aren’t super-strict about it. Our “space after” policy is a bit unclear and some files contain a mix of several iterations of the “preferred” style.

src/ripple/peerfinder/impl/Logic.h
nbougalis: Ideally, we’d avoid C-style casts altogether. Sad that we don’t have `std::assert` which requires the compiler to treat the thing being asserted as used to avoid issuing warnings.

thejohnfreeman: Why can't we use `std::assert`?

src/ripple/peerfinder/impl/Logic.h
```diff
@@ -392,6 +392,11 @@ class Logic
         if (keys_.find (key) != keys_.end())
             return Result::duplicate;
 
+        // If the peer belongs to a cluster, update the slot to reflect that.
+        counts_.remove (*slot);
+        slot->cluster (cluster);
+        counts_.add (*slot);
```
miguelportilla: If the slot cannot activate we've upped `counts_` and return, which differs from the prior behavior.

Albeit ugly, we could preserve the prior behavior with...
```
        // If the peer belongs to a cluster, update the slot to reflect that.
        auto const saved {slot->cluster()};
        slot->cluster(cluster);

        // See if we have an open space for this slot
        if (! counts_.can_activate (*slot))
        {
            if (! slot->inbound())
                bootcache_.on_success (
                    slot->remote_endpoint());
            return Result::full;
        }

        slot->cluster(saved);
        counts_.remove(*slot);
        slot->cluster(cluster);
        counts_.add(*slot);
```

thejohnfreeman: Would you need to reset `slot->cluster(saved)` before the `return Result::full` too?

miguelportilla: @thejohnfreeman Most likely, good catch.

nbougalis: The `counts_` should be properly decremented when the slot is released because the connection will be dropped. That should invoke [`Counts::remove`](https://github.com/ripple/rippled/blob/381a1b948b06d9526cc73f14cfc69635fabf8605/src/ripple/peerfinder/impl/Counts.h#L65) which will then call [`Counts::adjust`](https://github.com/ripple/rippled/blob/381a1b948b06d9526cc73f14cfc69635fabf8605/src/ripple/peerfinder/impl/Counts.h#L254).

nbougalis:
Merged as 61f443e3bbf3bf1f6e13f3ef25bb5fc60fe85078

PR #2805, opened 2018-12-13
Improve crawl shard resource usage

miguelportilla:


src/ripple/overlay/impl/PeerImp.cpp
```diff
@@ -1119,36 +1119,77 @@ PeerImp::onMessage(std::shared_ptr <protocol::TMShardInfo> const& m)
     // Parse the shard indexes received in the shard info
```
thejohnfreeman: Maybe amend this comment to explain that we're (1) parsing indices from a comma-separated string of indices (e.g. 123) and index ranges (e.g. 456-789), and (2) raising an error if any of them are out of some bounds that we assume.

src/ripple/nodestore/DatabaseShard.h
```diff
@@ -211,6 +211,14 @@ class DatabaseShard : public Database
...
 };
 
+constexpr
+std::uint32_t
+seqToShardIndex(std::uint32_t seq,
+    std::uint32_t ledgersPerShard = DatabaseShard::ledgersPerShardDefault)
+{
+    return (seq - 1) / ledgersPerShard;
+}
+
```
HowardHinnant: `PeerImp::hasLedger` in PeerImp.cpp:380 might be an additional place to use this new function.

src/ripple/overlay/impl/PeerImp.cpp
```diff
@@ -1119,36 +1119,77 @@ PeerImp::onMessage(std::shared_ptr <protocol::TMShardInfo> const& m)
...
 
-            if (seqs.size() == 1)
-                shardIndexes.insert(first);
-            else
+        std::vector<std::string> tokens;
+        boost::split(tokens, m->shardindexes(),
+            boost::algorithm::is_any_of(","));
+        for (auto const& t : tokens)
+        {
+            std::vector<std::string> indexes;
```
HowardHinnant: This can be made a bit more efficient by moving `indexes` outside of the loop and calling `indexes.clear()` at this spot.  It won't make any difference on the first iteration, but subsequent iterations can benefit from `indexes`'s `capacity()` to avoid allocations.

src/ripple/nodestore/DatabaseShard.h
miguelportilla: Indeed but I didn't want to muddy this PR with it. I will be pushing a separate PR of related miscellaneous fixes that include your suggestion. Thanks!

nbougalis:
Merged as 1be8094ee265167c7b40b4a33f6468ce5b43a34d.

PR #2807, opened 2018-12-13
Add zaphod.alloy.ee to default hub configuration

alloyxrp:
Add pool of servers from alloy.ee to default hub bootstrap. This pool will be operated for a minimum of 1 year. https://www.alloy.ee is a service provider from Estonia, with hardened application hosting as the core competence.

### DIFF
--- a/src/ripple/overlay/impl/OverlayImpl.cpp
+++ b/src/ripple/overlay/impl/OverlayImpl.cpp
@@ -484,13 +484,22 @@ OverlayImpl::onPrepare()
     m_peerFinder->setConfig (config);
 
     // Populate our boot cache: if there are no entries in [ips] then we use
-    // the entries in [ips_fixed]. If both are empty, we resort to a round-robin
-    // pool.
+    // the entries in [ips_fixed].
     auto bootstrapIps = app_.config().IPS.empty()
         ? app_.config().IPS_FIXED
         : app_.config().IPS;
+
+
+    // If nothing is specified, default to several well-known high-capacity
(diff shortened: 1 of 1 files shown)


nbougalis:
Merged as f443439f1fc380602d2c09b5ae70551cb9effb06. Thanks for the contribution @alloyxrp and your support of the XRP Ledger!

wilsonianb:
Should the pool also be added to `[ips]` in the default config?
As is, it would only be used if rippled runs without `[ips]` or `[ips_fixed]`.

alloyxrp:
The other option could be to comment out the `[ips]` section in the default rippled.cfg example file, and have no entry there at all. Any entries under the sections `[ips]` and `[ips_fixed]` would only be added by the user. @nbougalis ?

garay-daniel:
I think we should have the default hubs explicitly listed in [ips_fixed] as the default. Users can remove/add to this list as they want. I think this is more clear. I do not like that the default only works if [ips_fixed] and [ips] is unused.

MarkusTeufelberger:
@garay-daniel - that would seriously hinder any semi-private/forked network which doesn't want to be connected with the mainnet.

alloyxrp:
@MarkusTeufelberger : This is currently the case since r.ripple.com is under `[ips]` in the default config. This means that the condition of bootstrapping with zaphod.alloy.ee will never be reached. So, either @wilsonianb suggestion, or removing all entries will need to be implemented in the config.

intelliot:
(#2844 resolves the issue by removing the [ips] section)

Issue #1751, opened 2016-06-22
Add "share_UNL" to config, exposing the currently used validators in /crawl


MarkusTeufelberger:
While it would be a rather informal way to share your UNL (and could of course be faked etc.), it still would give at least some hints about validator topology which would be needed for decentralizing.

Additionally, a signature with your validator's key might be added, to give some more concrete proof that if the node fakes this statement, it is done knowingly and deliberately.

MarkusTeufelberger:
UNLs can be exposed via /crawl now (and by default are...). They don't carry signatures though, so they are fairly easy to plausibly deny.

Since the current implementation fulfills the ticket's title, I'll close this.

@garay-daniel, feel free to comment or reopen.

PR #2823, opened 2019-01-10
Enchance /crawl API endpoint with local server information (RIPD-1644):

jwbusch:
The /crawl API endpoint allows developers to examine the structure of the XRP Ledger's overlay network.
This commit adds additional information about the local server to the /crawl endpoint, making it
possible for a developer to create rich network-wide status dashboards.

Related:
 - https://developers.ripple.com/peer-protocol.html
 - https://github.com/ripple/rippled-network-crawler

### DIFF
--- a/cfg/rippled-example.cfg
+++ b/cfg/rippled-example.cfg
@@ -1057,7 +1057,7 @@
 #
 #   server = <flag>
 #
-#       Report information about the local server, similar to the "server_info"
+#       Report information about the local server, similar to the "server_state"
 #       RPC API. Default is '1'.
 #
 #   counts = <flag>
--- a/cfg/rippled-example.cfg
+++ b/cfg/rippled-example.cfg
@@ -20,7 +20,9 @@
 #
 #   7. Voting
 #
-#   8. Example Settings
+#   8. Misc Settings
+#
+#   9. Example Settings
 #
 #-------------------------------------------------------------------------------
 #
@@ -1038,6 +1040,46 @@
 #
 #   [signing_support]
 #   true
+#
+# [crawl]
+#
+#   List of options to control what data is reported through the /crawl endpoint
+#   See https://developers.ripple.com/peer-protocol.html#peer-crawler
+#
+#   off
+#
+#       Do not respond to /crawl requests
+#
+#   overlay = <flag>
+#
+#       Report information about peers this server is connected to, similar
+#       to the "peers" RPC API. Default is '1'.
+#
+#   server = <flag>
+#
+#       Report information about the local server, similar to the "server_info"
+#       RPC API. Default is '1'.
+#
+#   counts = <flag>
+#
+#       Report information about the local server health counters, similar to
+#       the "get_counts" RPC API. Default is '0'.
+#
+#   unl = <flag>
+#
+#       Report information about the local server's validator lists, similar to
+#       the "validators" and "validator_list_sites" RPC APIs. Default is '1'.
+#
+#   Example:
+#
+#   [crawl]
+#   off           # disable /crawl responses
+#
+#   [crawl]
+#   overlay = 1   # report peer overlay info
+#   server = 1    # report local server info
+#   counts = 0    # do not report server counts
+#   unl = 1       # report server validator lists
 #
 #-------------------------------------------------------------------------------
 #
--- a/cfg/rippled-example.cfg
+++ b/cfg/rippled-example.cfg
@@ -1046,9 +1046,9 @@
 #   List of options to control what data is reported through the /crawl endpoint
 #   See https://developers.ripple.com/peer-protocol.html#peer-crawler
 #
-#   off
+#   <flag>
 #
-#       Do not respond to /crawl requests
+#       Enable or disable access to /crawl requests. Default is '1'
 #
 #   overlay = <flag>
 #
@@ -1073,7 +1073,7 @@
 #   Example:
 #
 #   [crawl]
-#   off           # disable /crawl responses
+#   0
 #
 #   [crawl]
 #   overlay = 1   # report peer overlay info
--- a/src/ripple/app/misc/impl/ValidatorList.cpp
+++ b/src/ripple/app/misc/impl/ValidatorList.cpp
@@ -192,11 +192,28 @@ ValidatorList::load (
 
 
 ListDisposition
-ValidatorList::applyList (
+ValidatorList::applyList(
     std::string const& manifest,
     std::string const& blob,
     std::string const& signature,
     std::uint32_t version)
+{
+    // This API is only used by tests
+    return applyList(
+        manifest,
+        blob,
+        signature,
+        version,
+        std::string("unknown.test"));
+}
+
+ListDisposition
+ValidatorList::applyList (
+    std::string const& manifest,
+    std::string const& blob,
+    std::string const& signature,
+    std::uint32_t version,
+    std::string siteUri)
 {
     if (version != requiredListVersion)
         return ListDisposition::unsupported_version;
@@ -215,6 +232,7 @@ ValidatorList::applyList (
     publisherLists_[pubKey].sequence = list["sequence"].asUInt ();
     publisherLists_[pubKey].expiration = TimeKeeper::time_point{
         TimeKeeper::duration{list["expiration"].asUInt()}};
+    publisherLists_[pubKey].siteUri = std::move(siteUri);
     std::vector<PublicKey>& publisherList = publisherLists_[pubKey].list;
 
     std::vector<PublicKey> oldList = publisherList;
@@ -544,6 +562,7 @@ ValidatorList::getJson() const
         Json::Value& curr = jPublisherLists.append(Json::objectValue);
         curr[jss::pubkey_publisher] = strHex(p.first);
         curr[jss::available] = p.second.available;
+        curr[jss::uri] = p.second.siteUri;
         if(p.second.expiration != TimeKeeper::time_point{})
         {
             curr[jss::seq] = static_cast<Json::UInt>(p.second.sequence);
--- a/src/ripple/app/misc/impl/ValidatorList.cpp
+++ b/src/ripple/app/misc/impl/ValidatorList.cpp
@@ -190,23 +190,6 @@ ValidatorList::load (
     return true;
 }
 
-
-ListDisposition
-ValidatorList::applyList(
-    std::string const& manifest,
-    std::string const& blob,
-    std::string const& signature,
-    std::uint32_t version)
-{
-    // This API is only used by tests
-    return applyList(
-        manifest,
-        blob,
-        signature,
-        version,
-        std::string("unknown.test"));
-}
-
 ListDisposition
 ValidatorList::applyList (
     std::string const& manifest,
--- a/src/ripple/overlay/Overlay.h
+++ b/src/ripple/overlay/Overlay.h
@@ -70,15 +70,6 @@ public:
         bool expire = false;
         beast::IP::Address public_ip;
         int ipLimit = 0;
-
-        enum CrawlOptions {
-            Disabled = 0x00,
-            Overlay = 0x01,
-            ServerInfo = 0x02,
-            ServerCounts = 0x04,
-            Unl = 0x08
-        };
-
         std::uint32_t crawlOptions = 0;
     };
 
--- a/src/ripple/overlay/Overlay.h
+++ b/src/ripple/overlay/Overlay.h
@@ -104,11 +104,6 @@ public:
     std::size_t
     size () = 0;
 
-    /** Returns information reported to the crawl cgi command. */
-    virtual
-    Json::Value
-    crawl() = 0;
-
     /** Return diagnostics on the status of all peers.
         @deprecated This is superceded by PropertyStream
     */
--- a/src/ripple/overlay/Overlay.h
+++ b/src/ripple/overlay/Overlay.h
@@ -70,6 +70,16 @@ public:
         bool expire = false;
         beast::IP::Address public_ip;
         int ipLimit = 0;
+
+        enum CrawlOptions {
+            Disabled = 0x00,
+            Overlay = 0x01,
+            ServerInfo = 0x02,
+            ServerCounts = 0x04,
+            Unl = 0x08
+        };
+
+        std::uint32_t crawlOptions = 0;
     };
 
     using PeerSequence = std::vector <std::shared_ptr<Peer>>;
--- a/src/ripple/overlay/impl/OverlayImpl.cpp
+++ b/src/ripple/overlay/impl/OverlayImpl.cpp
@@ -1244,10 +1244,22 @@ setup_Overlay (BasicConfig const& config)
     }
     {
         auto const& section = config.section("crawl");
+        auto const& values = section.values();
+
+        if (values.size() > 1) {
+            Throw<std::runtime_error>("Configured [crawl] section is invalid");
+        }
+
         bool crawlEnabled = true;
 
-        if (section.lines().size() == 1 && section.legacy() == "off") {
-            crawlEnabled = false;
+        // Only allow "off" as a value
+        if (values.size() == 1) {
+            if (values.front() == "off") {
+                crawlEnabled = false;
+            } else {
+                Throw<std::runtime_error>(
+                    "Configured [crawl] section has invalid value: " + values.front());
+            }
         }
 
         if (crawlEnabled) {
--- a/src/ripple/overlay/impl/OverlayImpl.cpp
+++ b/src/ripple/overlay/impl/OverlayImpl.cpp
@@ -1246,8 +1246,11 @@ setup_Overlay (BasicConfig const& config)
         auto const& section = config.section("crawl");
         auto const& values = section.values();
 
-        if (values.size() > 1) {
-            Throw<std::runtime_error>("Configured [crawl] section is invalid");
+        if (values.size() > 1 ||
+            values.size() == 1 && section.lines().size() > 1)
+        {
+            Throw<std::runtime_error>(
+                "Configured [crawl] section is invalid, too many values");
         }
 
         bool crawlEnabled = true;
--- a/src/ripple/overlay/impl/OverlayImpl.cpp
+++ b/src/ripple/overlay/impl/OverlayImpl.cpp
@@ -21,8 +21,10 @@
 #include <ripple/app/misc/HashRouter.h>
 #include <ripple/app/misc/NetworkOPs.h>
 #include <ripple/app/misc/ValidatorList.h>
+#include <ripple/app/misc/ValidatorSite.h>
 #include <ripple/basics/base64.h>
 #include <ripple/basics/make_SSLContext.h>
+#include <ripple/basics/PerfLog.h>
 #include <ripple/beast/core/LexicalCast.h>
 #include <ripple/core/DatabaseCon.h>
 #include <ripple/nodestore/DatabaseShard.h>
@@ -32,6 +34,7 @@
 #include <ripple/overlay/impl/PeerImp.h>
 #include <ripple/peerfinder/make_Manager.h>
 #include <ripple/rpc/json_body.h>
+#include <ripple/rpc/handlers/GetCounts.h>
 #include <ripple/server/SimpleWriter.h>
 
 #include <boost/utility/in_place_factory.hpp>
@@ -912,6 +915,61 @@ OverlayImpl::crawl()
     return jv;
 }
 
+Json::Value
+OverlayImpl::crawl_server()
+{
+    Json::Value jv;
+
+    bool const humanReadable = false;
+    bool const admin = false;
+    bool const counters = false;
+
+    Json::Value server_info = app_.getOPs().getServerInfo(humanReadable, admin, counters);
+
+    // Filter out some information
+    server_info.removeMember(jss::hostid);
+    server_info.removeMember(jss::load_factor_fee_escalation);
+    server_info.removeMember(jss::load_factor_fee_queue);
+
+    if (server_info.isMember(jss::validated_ledger))
+    {
+        Json::Value& validated_ledger = server_info[jss::validated_ledger];
+
+        validated_ledger.removeMember(jss::base_fee);
+        validated_ledger.removeMember(jss::reserve_base_xrp);
+        validated_ledger.removeMember(jss::reserve_inc_xrp);
+    }
+
+    jv[jss::info] = std::move(server_info);
+
+    jv[jss::counters] = getCountsJson(app_, 10);
+
+    Json::Value validators = app_.validators().getJson();
+
+    if (validators.isMember(jss::publisher_lists))
+    {
+        Json::Value& publisher_lists = validators[jss::publisher_lists];
+
+        for (auto& publisher : publisher_lists)
+        {
+            publisher.removeMember(jss::list);
+        }
+    }
+
+    validators.removeMember(jss::signing_keys);
+
+    Json::Value validatorSites = app_.validatorSites().getJson();
+
+    if (validatorSites.isMember(jss::validator_sites))
+    {
+        validators[jss::validator_sites] = std::move(validatorSites[jss::validator_sites]);
+    }
+
+    jv[jss::validators] = std::move(validators);
+
+    return jv;
+}
+
 // Returns information on verified peers.
 Json::Value
 OverlayImpl::json ()
@@ -933,6 +991,7 @@ OverlayImpl::processRequest (http_request_type const& req,
     msg.insert("Content-Type", "application/json");
     msg.insert("Connection", "close");
     msg.body()["overlay"] = crawl();
+    msg.body()["server"] = crawl_server();
     msg.prepare_payload();
     handoff.response = std::make_shared<SimpleWriter>(msg);
     return true;
--- a/src/ripple/overlay/impl/OverlayImpl.cpp
+++ b/src/ripple/overlay/impl/OverlayImpl.cpp
@@ -24,7 +24,6 @@
 #include <ripple/app/misc/ValidatorSite.h>
 #include <ripple/basics/base64.h>
 #include <ripple/basics/make_SSLContext.h>
-#include <ripple/basics/PerfLog.h>
 #include <ripple/beast/core/LexicalCast.h>
 #include <ripple/core/DatabaseCon.h>
 #include <ripple/nodestore/DatabaseShard.h>
@@ -863,7 +862,7 @@ OverlayImpl::limit()
 }
 
 Json::Value
-OverlayImpl::crawl()
+OverlayImpl::getOverlayInfo()
 {
     using namespace std::chrono;
     Json::Value jv;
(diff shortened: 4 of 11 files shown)


MarkusTeufelberger:
What additional information exactly is added and why?

nbougalis:
It includes the (non-privileged) output of the `server_info` RPC call, as well as the list of validators on the server's UNL. This will be useful to allow anyone interested to build network explorers.

nbougalis:
Here's a sample output:

```
{
   "overlay":{
      "active":[
         {
            "complete_ledgers":"44290046-44290499",
            "ip":"46.166.162.76",
            "port":"51235",
            "public_key":"A56oGKbcr8lbcNx9K75+8U/5/w8cm2hYX6dFz8oRQNQ7",
            "type":"out",
            "uptime":28,
            "version":"rippled-1.1.2"
         },
         {
            "complete_ledgers":"44290243-44290499",
            "ip":"34.213.185.56",
            "port":"51235",
            "public_key":"Atv/wrJ7jkcXjB7gmvj5X27MOmRoa+CfZYvyCxCXNtC0",
            "type":"out",
            "uptime":28,
            "version":"rippled-1.1.2"
         },
         {
            "complete_ledgers":"44290243-44290499",
            "ip":"54.218.232.1",
            "port":"51235",
            "public_key":"ArFMArh4JCU5at5mhJVweAuAiNc4FyxpgeP+ZwsEruIU",
            "type":"out",
            "uptime":28,
            "version":"rippled-1.1.2"
         },
         {
            "complete_ledgers":"44290243-44290499",
            "ip":"54.213.19.217",
            "port":"51235",
            "public_key":"AmPlGnCXbn2AaoDkKpdb1WODEMKZ/gZRapk/HkuTpO5x",
            "type":"out",
            "uptime":28,
            "version":"rippled-1.1.2"
         },
         {
            "complete_ledgers":"44290243-44290499",
            "ip":"184.173.45.38",
            "port":"51235",
            "public_key":"AjSNOnXBSWuMOQBuyaPQ2q5x32doQFv4pMTO5+N+Q1lr",
            "type":"out",
            "uptime":28,
            "version":"rippled-1.1.2"
         },
         {
            "complete_ledgers":"44289475-44290499",
            "ip":"54.84.21.230",
            "port":"51235",
            "public_key":"AmQgZGrN0mII1lnAFCfFHUe/A9gn7ApTNEgEt406omfu",
            "type":"out",
            "uptime":28,
            "version":"rippled-1.1.2"
         },
         {
            "complete_ledgers":"44280499-44290499",
            "ip":"162.243.114.118",
            "port":"51235",
            "public_key":"AufDkW4E1DOxjzRPj46Eu+AyJdsakUeJTz3xklv1kCfp",
            "type":"out",
            "uptime":28,
            "version":"rippled-1.1.2"
         },
         {
            "complete_ledgers":"44289475-44290499",
            "ip":"54.86.175.122",
            "port":"51235",
            "public_key":"A8fY0NYbO2astDjP7HuVL1k+hwLhESFEcb+zVg9C3d1a",
            "type":"out",
            "uptime":28,
            "version":"rippled-1.1.2"
         },
         {
            "complete_ledgers":"44290243-44290499",
            "ip":"198.11.206.6",
            "port":"51235",
            "public_key":"A2ceMJkfKeyTq6Lp4bOYk78TpWbHMxw0JVAyKgjArsG1",
            "type":"out",
            "uptime":28,
            "version":"rippled-1.1.2"
         },
         {
            "complete_ledgers":"32570-44290498",
            "ip":"198.11.206.10",
            "port":"51235",
            "public_key":"Aw7J0CVhFKt0h6PDEpqu6t4LbPY0PsX8jCFbvSQFDOkW",
            "type":"out",
            "uptime":28,
            "version":"rippled-1.1.2"
         }
      ]
   },
   "server":{
      "counters":{
         "AL_hit_rate":0,
         "HashRouterEntry":1844,
         "SLE_hit_rate":0,
         "STObject":208,
         "STValidation":208,
         "dbKBLedger":178,
         "dbKBTotal":955,
         "dbKBTransaction":9,
         "fullbelow_size":0,
         "historical_perminute":0,
         "ledger_hit_rate":33.33333206176758,
         "node_hit_rate":0,
         "node_read_bytes":0,
         "node_reads_hit":0,
         "node_reads_total":0,
         "node_writes":2,
         "node_written_bytes":639,
         "treenode_cache_size":2,
         "treenode_track_size":2,
         "uptime":"30 seconds",
         "write_load":0
      },
      "info":{
         "build_version":"1.2.0-b10+DEBUG",
         "closed_ledger":{
            "base_fee":10,
            "close_time":600377640,
            "hash":"EC16632394537D4950FA4C8B964EB51E227C71B4A6C1860ACFFF3BFC4C699D34",
            "reserve_base":200000000,
            "reserve_inc":50000000,
            "seq":2
         },
         "complete_ledgers":"empty",
         "io_latency_ms":1,
         "jq_trans_overflow":"0",
         "last_close":{
            "converge_time":0,
            "proposers":0
         },
         "load_base":256,
         "load_factor":256,
         "load_factor_fee_reference":256,
         "load_factor_server":256,
         "peer_disconnects":"0",
         "peer_disconnects_resources":"0",
         "peers":10,
         "pubkey_node":"n9LkoGdQjcfE1y7Qjw2RkfdGT2GkZi8YVuT3hWQANd7eEyM4bntg",
         "published_ledger":"none",
         "server_state":"connected",
         "server_state_duration_us":"28966463",
         "state_accounting":{
            "connected":{
               "duration_us":"28966463",
               "transitions":1
            },
            "disconnected":{
               "duration_us":"1384071",
               "transitions":1
            },
            "full":{
               "duration_us":"0",
               "transitions":0
            },
            "syncing":{
               "duration_us":"0",
               "transitions":0
            },
            "tracking":{
               "duration_us":"0",
               "transitions":0
            }
         },
         "time":"2019-Jan-09 19:34:30.844047",
         "uptime":30,
         "validation_quorum":4294967295
      },
      "validators":{
         "local_static_keys":[
            "nHUDK4Tff855pM1cGseNEyzedi4XiHk9XLfbDJYnzbvB2heGVDnZ",
            "nHBZfAVZ3vQYZ4pWym6UXPLEGycvF1LDzzcaEC1xUSSZ6hqYraWF"
         ],
         "publisher_lists":[
            {
               "available":true,
               "expiration":"2019-Jan-15 00:00:00.000000000",
               "pubkey_publisher":"ED2677ABFFD1B33AC6FBC3062B71F1E8397C1505E1C42C64D11AD1B28FF73F4734",
               "seq":41,
               "uri":"https://vl.ripple.com",
               "version":1
            }
         ],
         "trusted_validator_keys":[
            "nHULGMbQHyXe92wtnxA1X9TEp26qPiRfwfTv5MskD1yMro7bQ2df",
         ],
         "validation_quorum":4294967295,
         "validator_list":{
            "count":2,
            "expiration":"2019-Jan-15 00:00:00.000000000",
            "status":"active"
         },
         "validator_sites":[
            {
               "last_refresh_status":"accepted",
               "last_refresh_time":"2019-Jan-09 19:34:00.992644674",
               "next_refresh_time":"2019-Jan-09 19:39:00.720214781",
               "refresh_interval_min":5,
               "uri":"https://vl.ripple.com"
            }
         ]
      }
   }
}
```

MarkusTeufelberger:
If someone wanted to share this information, they also could just publish their public RPC-port in the `/crawl`-endpoint and let users poll it.

Anyways, if you want to go ahead with this, please make sure that private peers are not counted in the number of peers reported (otherwise their existence would leak).
Also please change the output to the output of the `server_state` RPC call, this endpoint is not for human consumption anyways.
Lastly the `counters` section is not part of the `server_info` output on my node, is that really publically available information on every node? It seems quite specific and might give hints on the underlying hardware and configuration (and its limits...).

nbougalis:
Good point about `server_state` and maybe we need to reconsider the output of `get_counts` (although I don't think it includes anything particularly sensitive). Perhaps we can add a config option that allows operators to opt-in/opt-out of including this additional information.

To your other point: sure, people could expose this information by allowing access to the RPC port, but that exposes other APIs and places an administrative burden on operators which I think is unnecessary.

The goal is to allow everyone to get a better idea of the state of the network and to create new and exciting tools that improve the community's insight in the network. I believe that you were one of the people that asked for such information (specifically for information about which validators a server was using).

If you have a better idea for how to go about making this data, I'd love to hear about it!

src/ripple/overlay/impl/OverlayImpl.cpp
```diff
@@ -912,6 +915,61 @@ OverlayImpl::crawl()
...
+Json::Value
+OverlayImpl::crawl_server()
+{
+    Json::Value jv;
+
+    bool const humanReadable = false;
+    bool const admin = false;
+    bool const counters = false;
+
+    Json::Value server_info = app_.getOPs().getServerInfo(humanReadable, admin, counters);
```
mellery451: I'm guessing you were trying to counteract the trouble of comprehending params a call like `getServerInfo(false, false, false)`? I'm not particularly sold on this approach here, but it's fine either way to me.

src/ripple/overlay/impl/OverlayImpl.cpp
```diff
@@ -912,6 +915,61 @@ OverlayImpl::crawl()
     return jv;
 }
 
+Json::Value
+OverlayImpl::crawl_server()
```
mellery451: nitpicking the name - this and the existing `crawl` seem poorly named. Probably should be `getCrawlInfo` or something like that. No action needed - just throwing it around.

jwbusch: I'm down to change it `crawl -> getOverlayInfo` and `crawl_server -> getLocalServerInfo`

src/ripple/overlay/impl/OverlayImpl.cpp
jwbusch: There's not really any overhead, and if it changes later you'll also get a more readable diff: 
`bool const counters = `~~false~~` true;` vs. `(false, false, `~~false~~`true);`

src/ripple/overlay/impl/OverlayImpl.cpp
```diff
@@ -21,8 +21,10 @@
 #include <ripple/app/misc/HashRouter.h>
 #include <ripple/app/misc/NetworkOPs.h>
 #include <ripple/app/misc/ValidatorList.h>
+#include <ripple/app/misc/ValidatorSite.h>
 #include <ripple/basics/base64.h>
 #include <ripple/basics/make_SSLContext.h>
+#include <ripple/basics/PerfLog.h>
```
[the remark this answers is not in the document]
jwbusch: You're right, not sure how that got in either...

src/ripple/app/misc/impl/ValidatorList.cpp
```diff
@@ -192,11 +192,28 @@ ValidatorList::load (
...
 
 ListDisposition
-ValidatorList::applyList (
+ValidatorList::applyList(
     std::string const& manifest,
     std::string const& blob,
     std::string const& signature,
     std::uint32_t version)
+{
+    // This API is only used by tests
```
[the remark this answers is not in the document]
jwbusch: I did that previously, but I have a habit of doing things like this to avoid unnecessary code churn... I can update the tests if that's preferred.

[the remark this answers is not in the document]
mellery451: I think we can afford the churn, especially if it's unit test code 👍

src/ripple/overlay/impl/OverlayImpl.cpp
jwbusch: Also was able to remove it from the `Overlay` interface since it's unused there

MarkusTeufelberger:
> I believe that you were one of the people that asked for such information (specifically for information about which validators a server was using).

Yes, see #1751 - I'd like this data to be signed with the server's key pair (so I can receive this data federated), I'd like that kind of data sharing to be opt-in (UNLs are private for a reason) and if get_counts contains only benign data, I'd like it to be made a public API on the RPC interface. I still think it leaks far too much meta data to be benign, but then again you are paying bounties for bugs, not security advice.

I'd like to see public RPC/WS(S) ports available on the server listed in `/crawl` too btw., even if you keep the output of `server_state` in there, since it is currently hard to discover public endpoints without doing a port scan (which is a bit impolite). And I'd like each section (public endpoints, peers, UNL settings, server_state, counters) to be individually enables/disabled. I don't care that much about what the defaults are (I'd say peers, public API endpoints and server_state should be on, UNL and counters off by default), as long as it is configurable.

MarkusTeufelberger:
Oh, and the "humanized" output in `counters` for uptime really isn't very useful imho, maybe that field should be removed alltogether - it is available elsewhere too and if you return humanized values, I'd expect them in `server_info` (which ironically reports uptime in seconds only).

nbougalis:
You make some good points @MarkusTeufelberger, thanks for the feedback. We should definitely add the switch to allow the admin to opt-in to the enhanced crawl.

Also, signing the data could prove useful, but that can be slightly tricky to implement, so maybe we can punt on this for round one?

src/ripple/overlay/impl/OverlayImpl.cpp
```diff
@@ -912,6 +914,61 @@ OverlayImpl::crawl()
...
+    if (server_info.isMember(jss::validated_ledger))
+    {
+        Json::Value& validated_ledger = server_info[jss::validated_ledger];
+
+        validated_ledger.removeMember(jss::base_fee);
+        validated_ledger.removeMember(jss::reserve_base_xrp);
+        validated_ledger.removeMember(jss::reserve_inc_xrp);
+    }
+
+    jv[jss::info] = std::move(server_info);
```
seelabs: We move from both `server_info` and `validator_sites` in this function, and they shouldn't be used after they are used from (they aren't, but I'd like to make that clear from the code for easy inspection and to make futures changes less likely to use them). I'd either:
1) Just put them into their own scopes, starting at the declaration of the varialbe and ending when they are moved.
or
2) Use a lambda to set the values. I.e.: `jv[jss::info] = [&]{...}()`. RVO should make this as efficient as moving.
Either way is fine with me.

src/ripple/rpc/handlers/GetCounts.h
```diff
@@ -0,0 +1,32 @@
+//------------------------------------------------------------------------------
+/*
+    This file is part of rippled: https://github.com/ripple/rippled
+    Copyright (c) 2012, 2013 Ripple Labs Inc.
```
seelabs: Nit: Old copyright

src/ripple/rpc/handlers/GetCounts.cpp
```diff
@@ -58,58 +58,50 @@ textTime(std::string& text, UptimeClock::time_point& seconds,
...
-    dbKB = getKBUsedDB (context.app.getTxnDB ().getSession ());
+    dbKB = getKBUsedDB (app.getTxnDB ().getSession ());
 
     if (dbKB > 0)
         ret[jss::dbKBTransaction] = dbKB;
 
     {
-        std::size_t c = context.app.getOPs().getLocalTxCount ();
+        std::size_t c = app.getOPs().getLocalTxCount ();
         if (c > 0)
```
seelabs: Not in your changeset, but there's some strangeness in this code. For exmaple, `c` is an unsigned type, why are we checking for `>0` instead of just `0`. Also, why is the declaration of `c` outside the if. I'd expect this to be written something like:
```
    if (Json::UInt c = app.getOPs().getLocalTxCount())
        ret[jss::local_txs] = c;
```
Ditto for all the `dbKB` uses above. I would have expected that to be an unsigned type and also to have it's declaration in an if statement and check for zero, not `>0`.
I wouldn't make this change in this changeset, just making a note to myself.

src/ripple/rpc/handlers/GetCounts.cpp
```diff
@@ -121,21 +113,21 @@ Json::Value doGetCounts (RPC::Context& context)
     textTime (uptime, s, "second", 1s);
     ret[jss::uptime] = uptime;
```
seelabs: Also not in your changeset, but I'd put this into it's own block. Something like:
```
    {
        std::string uptime;
        uptime.reserve(16);
        auto s = UptimeClock::now();
        textTime(uptime, s, "year", 365 * 24h);
        textTime(uptime, s, "day", 24h);
        textTime(uptime, s, "hour", 1h);
        textTime(uptime, s, "minute", 1min);
        textTime(uptime, s, "second", 1s);
        ret[jss::uptime] = uptime;
    }
```

src/ripple/rpc/handlers/GetCounts.cpp
```diff
@@ -58,58 +58,50 @@ textTime(std::string& text, UptimeClock::time_point& seconds,
...
+        app.getInboundLedgers().fetchRate());
+    ret[jss::SLE_hit_rate] = app.cachedSLEs().rate();
+    ret[jss::node_hit_rate] = app.getNodeStore ().getCacheHitRate ();
+    ret[jss::ledger_hit_rate] = app.getLedgerMaster ().getCacheHitRate ();
+    ret[jss::AL_hit_rate] = app.getAcceptedLedgerCache ().getHitRate ();
 
-    ret[jss::fullbelow_size] = static_cast<int>(context.app.family().fullbelow().size());
-    ret[jss::treenode_cache_size] = context.app.family().treecache().getCacheSize();
-    ret[jss::treenode_track_size] = context.app.family().treecache().getTrackSize();
+    ret[jss::fullbelow_size] = static_cast<int>(app.family().fullbelow().size());
```
seelabs: I don't understand why we are casting to `int` instead of `Json::UInt`. I know the rest of the function does something similar, so I get why you did it. I'm not understanding why it was done origionally.

src/ripple/overlay/impl/OverlayImpl.cpp
seelabs: This is a hard problem. In my own (non-rippled) code, I either use enum classes for the parameters or call the function like `getServerInfo(/*humanreadable*/ false, /*admin*/ false, /*counters*/ false)`. Of course, you're right that any function with three bool switches should be looked at as ripe for refactoring.

src/ripple/rpc/handlers/GetCounts.cpp
nbougalis: By the way, @HowardHinnant: this kind of thing (format duration to human-readable form) would probably make a nice addition to your time library.

MarkusTeufelberger: Just FYI - a "year" is not a static time span of 365 days. Uptime in that regions is unhealthy anyways, so I'd recommend leaving it out.

nbougalis: Howard is the master of all things `<chrono>` so I'm sure he can sort it out ;)

HowardHinnant: I agree with Markus.  The year unit is both never used and somewhat ambiguous if it ever were to be used.  Simply removing that one line of code leaves a correct and accurate uptime.

Whether it should be done as part of this PR, or in a separate PR is a another question.

src/ripple/overlay/impl/OverlayImpl.cpp
jwbusch: This is now pulled into separate functions (so they can be controlled through the config file), so no moving necessary.

src/ripple/rpc/handlers/GetCounts.cpp
jwbusch: I didn't actually touch any logic in this code, just pulled the core of it from `doGetCounts` to `getCountsJson` so that it could be called from `OverlayImpl`. I tried to keep the formatting identical so the diff shows that just `context.` was the part that was removed.

src/ripple/rpc/handlers/GetCounts.cpp
jwbusch: It can be removed if that's the consensus. I didn't change the core of this code, just refactored it so I could re-use it in the `OverlayImpl` class.

src/ripple/rpc/handlers/GetCounts.h
jwbusch: 2012-2019? Most of the copyrights are out of date :)

jwbusch:
@MarkusTeufelberger I added some options to the `rippled.cfg` that would allow an admin to turn `/crawl` off entirely, or limit some of the information being reported through that endpoint. Let me know if that addresses your concern.

MarkusTeufelberger:
Great, thanks - I'd just like them to be documented in the example config file. :-)

cfg/rippled-example.cfg
```diff
@@ -1039,6 +1041,46 @@
...
+#       Do not respond to /crawl requests
+#
+#   overlay = <flag>
+#
+#       Report information about peers this server is connected to, similar
+#       to the "peers" RPC API. Default is '1'.
+#
+#   server = <flag>
+#
+#       Report information about the local server, similar to the "server_info"
```
MarkusTeufelberger: Still `server_info`, not `server_state`?

MarkusTeufelberger:
Thanks a lot, both for the feature and adapting it. :-)

src/ripple/overlay/impl/OverlayImpl.cpp
```diff
@@ -1203,23 +1223,49 @@ Overlay::Setup
...
+    }
     {
-        boost::system::error_code ec;
-        setup.public_ip = beast::IP::Address::from_string (ip, ec);
-        if (ec || beast::IP::is_private (setup.public_ip))
-            Throw<std::runtime_error> ("Configured public IP is invalid");
+        auto const& section = config.section("crawl");
+        bool crawlEnabled = true;
+
+        if (section.lines().size() == 1 && section.legacy() == "off") {
```
seelabs: A couple comments here:
1) If there's an error in the config file, it's silently ignored here. It also enables crawl on error, which isn't a good default. I'd probably throw here and not allow rippled to run with an invalid config file. But at a minimum I'd disable crawl on error and write to the log.
2) I wouldn't call the value "off". Since it's a boolean, I'd use 0/1 instead.
3) I wouldn't use the `legacy` function in new code. I'd use `values` instead.

jwbusch: 1.) the old default was to enable crawl, so I tried to keep it that way (i.e. always try to enable it unless explicitly disabled), but I've reworked it a bit.  
2.) I prefer "off" in this case because it's not one of the key/value pairs, it's just a single value, and seems clearer with the text. The example config is updated with listing "off" as the only option, it's not listed as a `<flag>`.
```
[crawl]
off
``` 
vs
```
[crawl]
0
```
3.) That works when adding more error handling. It wasn't clear upon first pass what happens if I throw an exception here, but the code above it does Throw for errors in the config file. I just learned that you don't see those exceptions in the debug-log or stdout by default though.

seelabs: For 1) I'm OK with default on, but not on error.

For 2) I don't want to introduce the new vocabulary words of `off` and `on` unless all the sections support it. 

Other configuration sections use `0`. For example:
```
[ssl_verify]
0
```
It's not consitant that we can say `[crawl] off` and not `[ssl_verify] off`. Also, if you really do want to keep `off`, I'd explictly check for `on`/`off`/`0`/`1`/`true`/`false` and signal an error if it's not in that set. I.e. `[crawl] o` should be flagged as an error. But to be clear, I vote for `0/1` unless we make this change for all sections.

For 3) I did not know that, I thought we did log errors. At any rate, I think that's an issue for a different PR. I do think we should throw on an error in the config file. We shouldn't let a misconfigured rippled run.

src/ripple/overlay/Overlay.h
```diff
@@ -70,6 +70,16 @@ class Overlay
         bool expire = false;
         beast::IP::Address public_ip;
         int ipLimit = 0;
+
+        enum CrawlOptions {
```
nbougalis: Just noting a micro-nit: these values are never exposed via the `Overlay` API, so they could just be migrated into `OverlayImpl`. _No change is necessary._

src/ripple/overlay/impl/OverlayImpl.cpp
```diff
@@ -912,6 +914,65 @@ OverlayImpl::crawl()
...
+    // Filter out some information
+    server_info.removeMember(jss::hostid);
+    server_info.removeMember(jss::load_factor_fee_escalation);
+    server_info.removeMember(jss::load_factor_fee_queue);
+
+    if (server_info.isMember(jss::validated_ledger))
+    {
+        Json::Value& validated_ledger = server_info[jss::validated_ledger];
+
+        validated_ledger.removeMember(jss::base_fee);
```
nbougalis: Why do we remove these three members?

src/ripple/overlay/impl/OverlayImpl.cpp
jwbusch: That's fair, updated.

src/ripple/overlay/impl/OverlayImpl.cpp
jwbusch: That was by request of dgaray, he mentioned they weren't very useful/just noise.

src/ripple/overlay/Overlay.h
jwbusch: That's true, I was going to use the pattern of bit flags in an enum, but realized that's a Microsoft specific quirk: [msdn.com](https://blogs.msdn.microsoft.com/vcblog/2008/03/10/visual-studio-2008-enum-bit-flags-visualization/).

src/ripple/overlay/impl/OverlayImpl.cpp
```diff
@@ -1244,10 +1244,22 @@ setup_Overlay (BasicConfig const& config)
     }
     {
         auto const& section = config.section("crawl");
+        auto const& values = section.values();
+
+        if (values.size() > 1) {
+            Throw<std::runtime_error>("Configured [crawl] section is invalid");
+        }
+
```
seelabs: Nit: We use Allman style for braces where the braces go on a new line. I.e.
```
        if (values.size() > 1) 
        {
            Throw<std::runtime_error>("Configured [crawl] section is invalid");
        }
```

nbougalis:
@MarkusTeufelberger: if your concerns have been addressed, can you sign-off on this? I'd like to get it into the next 1.2.0 beta.

MarkusTeufelberger:
:+1:

src/ripple/overlay/impl/OverlayImpl.cpp
```diff
@@ -932,7 +1005,25 @@ OverlayImpl::processRequest (http_request_type const& req,
     msg.insert("Server", BuildInfo::getFullVersionString());
     msg.insert("Content-Type", "application/json");
     msg.insert("Connection", "close");
-    msg.body()["overlay"] = crawl();
+    msg.body()["version"] = Json::Value(1u);
```
MarkusTeufelberger: I just stumbled upon this `"version": 1` field in the `/crawl` json data. What does it signify and under which circumstances would this get changed/incremented?

jwbusch: It would be used for breaking changes, i.e. a field is removed or if the representation of the data changed significantly.

PR #2619, opened 2018-07-03
Add a 'domain' field to validator lists:

cryptobrad:
This patch adds a new optional field to validator lists allowing
a validation list publisher to include the domain associated with
a validator it is listing.

I also rewrote an awkward loop using C++ constructs.

PR #2825, opened 2019-01-14
Simplify and document validator list parsing

cryptobrad:
This is a slimdown version of the more ambitious #2619 which is withdranw.
It changes one loop from a single-pass to a three pass but the code is not performance critical & is much now easier to read and verify that it is correct.

### DIFF
--- a/src/ripple/app/misc/impl/ValidatorList.cpp
+++ b/src/ripple/app/misc/impl/ValidatorList.cpp
@@ -205,80 +205,78 @@ ValidatorList::applyList (
 
     Json::Value list;
     PublicKey pubKey;
-    auto const result = verify (list, pubKey, manifest, blob, signature);
+    auto const result = verify(list, pubKey, manifest, blob, signature);
     if (result != ListDisposition::accepted)
         return result;
 
     // Update publisher's list
     Json::Value const& newList = list["validators"];
     publisherLists_[pubKey].available = true;
-    publisherLists_[pubKey].sequence = list["sequence"].asUInt ();
+    publisherLists_[pubKey].sequence = list["sequence"].asUInt();
     publisherLists_[pubKey].expiration = TimeKeeper::time_point{
         TimeKeeper::duration{list["expiration"].asUInt()}};
-    std::vector<PublicKey>& publisherList = publisherLists_[pubKey].list;
+    auto& publisherList = publisherLists_[pubKey].list;
 
-    std::vector<PublicKey> oldList = publisherList;
-    publisherList.clear ();
-    publisherList.reserve (newList.size ());
+    auto oldList = publisherList;
+    publisherList.clear();
+    publisherList.reserve(newList.size());
     std::vector<std::string> manifests;
     for (auto const& val : newList)
     {
-        if (val.isObject() &&
-            val.isMember ("validation_public_key") &&
-            val["validation_public_key"].isString ())
+        if (val.isObject())
         {
-            std::pair<Blob, bool> ret (strUnHex (
-                val["validation_public_key"].asString ()));
-
-            if (! ret.second || ! publicKeyType(makeSlice(ret.first)))
+            if (val.isMember("validation_public_key") &&
+                val["validation_public_key"].isString())
             {
-                JLOG (j_.error()) <<
-                    "Invalid node identity: " <<
-                    val["validation_public_key"].asString ();
-            }
-            else
-            {
-                publisherList.push_back (
-                    PublicKey(Slice{ ret.first.data (), ret.first.size() }));
+                std::pair<Blob, bool> ret(
+                    strUnHex(val["validation_public_key"].asString()));
+
+                if (!ret.second || !publicKeyType(makeSlice(ret.first)))
+                {
+                    JLOG (j_.error()) << "Invalid node identity: "
+                        << val["validation_public_key"].asString();
+                    continue;
+                }
+
+                publisherList.emplace_back(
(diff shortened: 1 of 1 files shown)


scottschurr:
FWIW, here is the build error reported by the gcc.DEBUG -Dunity=OFF build:
```
../../src/ripple/app/misc/impl/ValidatorList.cpp: In member function â€˜ripple::ListDisposition ripple::ValidatorList::applyList(const string&, const string&, const string&, uint32_t)â€™:
../../src/ripple/app/misc/impl/ValidatorList.cpp:259:46: error: â€˜comparatorâ€™ was not declared in this scope
                 std::back_inserter(removed), comparator);
                                              ^~~~~~~~~~
../../src/ripple/app/misc/impl/ValidatorList.cpp:259:46: note: suggested alternative: â€˜operator<â€™
                 std::back_inserter(removed), comparator);
                                              ^~~~~~~~~~
                                              operator<
```

cryptobrad:
please run the tesst

scottschurr:
Looks like all tests have passed.

HowardHinnant:
I'm getting a lot of messages that look like:
```
2019-Jan-18 01:43:47.254816 Flow:ERR Strand re-execute check failed. Step: 3
2019-Jan-18 01:43:47.287482 Flow:ERR Strand re-execute check failed. ExpectedIn: 0.0864060699458496 CachedIn: 0.08640606049562684 ExpectedOut: 263602 CachedOut: 263601
```

This is the first time I'm noticing this, but I'm not positive if these messages are new with this change or not.  Testing develop...

scottschurr:
FWIW, I've been seeing those `Flow:ERR Strand re-execute check failed` messages for over a year now.  They come from the Flow code.  You should only see them in a debug build, which may be why you haven't seen them a lot before.

HowardHinnant:
develop is not behaving with the same messages.  I'll investigate this further tomorrow.  In the mean time I don't recommend this PR until this is better understood.  The volume of messages I note above was enough to run my disk space low.

cryptobrad:
:question: This code does not make these errors at all or use payment but only work to oarse of the validator list
i shall look to make sure it is not wrong. thank you!

src/ripple/app/misc/impl/ValidatorList.cpp
```diff
@@ -205,80 +205,78 @@ ValidatorList::applyList (
...
+    publisherLists_[pubKey].sequence = list["sequence"].asUInt();
     publisherLists_[pubKey].expiration = TimeKeeper::time_point{
         TimeKeeper::duration{list["expiration"].asUInt()}};
-    std::vector<PublicKey>& publisherList = publisherLists_[pubKey].list;
+    auto& publisherList = publisherLists_[pubKey].list;
 
-    std::vector<PublicKey> oldList = publisherList;
-    publisherList.clear ();
-    publisherList.reserve (newList.size ());
+    auto oldList = publisherList;
```
jwbusch: What about 
```
auto oldList = std::move(publisherList);
publisherList.reserve(newList.size());
```
Old code was making a copy + reallocating, this would save us from the copy.

src/ripple/app/misc/impl/ValidatorList.cpp
```diff
@@ -205,80 +205,78 @@ ValidatorList::applyList (
...
-                publisherList.push_back (
-                    PublicKey(Slice{ ret.first.data (), ret.first.size() }));
+                std::pair<Blob, bool> ret(
+                    strUnHex(val["validation_public_key"].asString()));
+
+                if (!ret.second || !publicKeyType(makeSlice(ret.first)))
+                {
+                    JLOG (j_.error()) << "Invalid node identity: "
+                        << val["validation_public_key"].asString();
+                    continue;
```
jwbusch: this `continue;` changes the behavior, although it might be an improvement.

src/ripple/app/misc/impl/ValidatorList.cpp
```diff
@@ -529,9 +527,9 @@ ValidatorList::getJson() const
     auto it = publisherLists_.find(local);
     if (it != publisherLists_.end())
     {
-        for (auto const& key : it->second.list)
+        for (auto const& e : it->second.list)
```
jwbusch: I'd leave this alone, `key` is easier to read than `e`. (same with 552, 553)

ximinez:
Also, you probably should rebase onto the current tip of develop and resolve the conflict.

src/ripple/app/misc/impl/ValidatorList.cpp
```diff
@@ -205,80 +205,78 @@ ValidatorList::applyList (
...
-            *iNew < *iOld))
+        // Update keyListings_ for added and removed keys
+        std::sort(publisherList.begin(), publisherList.end());
+
         {
-            // Increment list count for added keys
-            ++keyListings_[*iNew];
-            ++iNew;
+            // Removed elements appear in the old list but not in the new
+            std::vector<PublicKey> removed;
```
jwbusch: This new code creates 2 `std::vectors` and copies a bunch of elements into each (with `set_difference`), set_difference also iterates over the lists twice. The old code has better performance, albeit uglier/harder to read. Would there be a way to keep the existing algorithm but add comments to explain it better?

nbougalis:
While I like this PR and the changes that it makes, no progress has been made. I'm going to be closing; @cryptobrad, feel free to re-open it if you'd like to have it considered again.

PR #2826, opened 2019-01-17
Include more information LedgerTrie json

mellery451:


MarkusTeufelberger:
Uhm, what is this change supposed to do?

mellery451:
@MarkusTeufelberger this change just adds a few more fields to some debug log information (dumped as json). No operational impact.

PR #2829, opened 2019-01-21
Slow consensus to allow lagging proposers to catch up.

mtrippled:


src/ripple/consensus/Consensus.h
```diff
@@ -700,6 +718,15 @@ Consensus<Adaptor>::peerProposalInternal(
 
     if (newPeerProp.prevLedger() != prevLedgerID_)
     {
+        uint32_t ourSeq = previousLedger_.seq();
+        if (newPeerProp.proposeSeq() < ++ourSeq)
```
bachase: This is mixing the sequence number of a `Ledger` and a `Proposal`, which are two different things and are not comparable. 

Proposals are part of the deliberation phase of consensus and indicate which transactions a node believes belong in the next ledger. A sequence of such proposals (each identified by the increasing `proposeSeq`) eventually converge on a final set of transactions, which are used to generate the next ledger.

The ledger sequence number identifies the ledger index, which is one more than the parent ledger number.

src/test/csf/Peer.h
```diff
@@ -298,6 +298,15 @@ struct Peer
...
     }
 
+    /**
+     * Stubs
+     */
+    std::size_t
+    allowedDisagree()
+    {
+        return 0;
+    }
```
bachase: I think its important to have the consensus simulation framework testing this change in the same way the production code operates.

src/ripple/consensus/Consensus.h
bachase: Aside: Note that the compiler errors are picking up on this difference, because the simulation framework uses tagged integer types to distinguish different kinds of sequence numbers.

src/ripple/app/misc/ValidatorList.h
```diff
@@ -241,6 +241,18 @@ class ValidatorList
...
+    /** Get maximum quantity of trusted validators allowed to disagree
+     * while still allowing consensus (converse of quorum).
+     *
+     * @return quantity allowed to disagree.
+     */
+    std::size_t
+    allowedDisagree () const
+    {
+        std::lock_guard<std::shared_timed_mutex> lock{mutex_};
+        return trustedKeys_.size() - quorum_;
```
jwbusch: Is there a guarantee that `trustedKeys_.size() >= quorum_`? Otherwise this could return a massive number.

src/ripple/consensus/Consensus.h
```diff
@@ -88,6 +91,8 @@ checkConsensus(
     std::chrono::milliseconds currentAgreeTime,
     ConsensusParms const & parms,
     bool proposing,
+    std::size_t const laggards,
```
jwbusch: nit: having two params of the same type is easy to mix up at the call-site. 

Best solution is probably to adopt more tagged integers.

src/ripple/consensus/Consensus.h
cryptobrad: so why do we not always use tag integer for such integers?

src/ripple/app/misc/ValidatorList.h
cryptobrad: i beieve that quorum_ can be 0xFFFFFFFF when lists are not available from calculateQuroum

src/ripple/consensus/ConsensusParms.h
```diff
@@ -82,6 +82,9 @@ struct ConsensusParms
     std::chrono::milliseconds ledgerMIN_CONSENSUS =
         std::chrono::milliseconds {1950};
 
+    //! Maximum time to delay consensus to let others catch up
+    std::chrono::seconds ledgerMAX_CONSENSUS {20};
```
cryptobrad: Maybe this should be a times of the MINCLOSE time? Like 10 x MINCLOSE

src/ripple/consensus/Consensus.h
bachase: The types used in the production code were in place before the consensus code refactor and are used in lots of other parts of the code. I think using tagged integers for them would be great, but would be a decent sized task by itself.

JoelKatz:
I'm not sure I understand what the intent of this change is. You already cannot declare a consensus unless 80% of validators on your UNL that have proposed have proposed the exact same transaction set you have. Was the intent to require 80% of your total UNL?

So, for example, say there are 12 servers on my UNL. I need to see 80% of 12 validate a ledger to call it fully validated. But if I only see 5 servers proposing in a particular consensus round, if they all agree with me, I'll declare a consensus. I won't wait for the other servers to join this round.

If the intent was to wait for the other servers to join this round, that's just not sane. Say the network is split into two. Neither side will ever see 80% of anyone's UNL agreeing and so both sides will wait for the other.

Our current design requires constant forward progress to repair from splits. Slowing down in a split of this type is not sensible. You have to keep going to produce validations, which is what you need to heal a split.

mtrippled:
> I'm not sure I understand what the intent of this change is. You already cannot declare a consensus unless 80% of validators on your UNL that have proposed have proposed the exact same transaction set you have. Was the intent to require 80% of your total UNL?
> 
> So, for example, say there are 12 servers on my UNL. I need to see 80% of 12 validate a ledger to call it fully validated. But if I only see 5 servers proposing in a particular consensus round, if they all agree with me, I'll declare a consensus. I won't wait for the other servers to join this round.
> 
> If the intent was to wait for the other servers to join this round, that's just not sane. Say the network is split into two different consensus rounds. (There's no connectivity issue, they just don't agree on the last closed ledger.) Neither side will ever see 80% of anyone's UNL agreeing and so both sides will wait for the other.
> 
> Our current design requires constant forward progress to repair from splits. Slowing down in a split of this type is not sensible. You have to keep going to produce validations, which is what you need to heal a split. You can't wait for the split to fix itself -- it won't.
> 
> That said, issuing a partial validation in the case where fewer than 50% of the servers in your total UNL proposed might make sense.

@bachase pointed out to me on this PR that I'm conflating the proposal round sequence with ledger sequence. The intention is to not build a ledger until it's possible for it to achieve network consensus. If I know that >20% of proposers are on a different ledger sequence, then I know it's impossible. So wait.

I'm closing this PR for now: whatever the best solution, it will basically require re-writing this PR. (@JoelKatz, @bachase )

PR #2830, opened 2019-01-22
Trivial: remove assert that accesses object post-dtor (RIPD-1704)

scottschurr:


jwbusch:
Took me a minute to understand the code and why it would be a case of  `use-after-free`... 

It's because `Stoppable` has a `RootStoppable` member, even though `RootStoppable` inherits from `Stoppable`... 

This bug only happens when the actual `RootStoppable` destructor runs (destroying the actual instance that is referenced by the member `m_root`) and then the `Stoppable` destructor runs, which accesses `m_root` after it was destroyed. 

Fix looks good, although it's a sign something could be redesigned. Minor nit: no need to mark `~RootStoppable` virtual, but also doesn't hurt anything.

nbougalis:
We should slipstream this into the next *1.2.0* beta or RC.

PR #2840, opened 2019-01-26
 Disallow both single- and multi-signing in RPC (RIPD-1713)

scottschurr:
This pull request is composed of two related commits.

The first commit adds some checking in the signing RPC commands (like submit and sign_for) to return an error when a transaction being multisigned is already single-signed and vice versa.  The transaction layer already protected against this problem.  So the commit just improves the error message for the user and stops processing earlier on this kind of error.  It also adds a couple of new error codes to ErrorCodes.h.  More on that in a bit.

In the process of making this change a bug was uncovered in the RPCCall unit test.  The first commit fixes that bug as well.

Also during the work on the first commit it was discovered that there was at least one code in ErrorCode.h that is currently unused anywhere in the current code base.  That led to an audit.  The audit in turn led to the second commit.

The second commit removes all of the codes in ErrorCodes.h that are unused.  The corresponding support code is also removed from ErrorCodes.cpp.

This leads me to my last question.  The source code in ErrorCodes.h notes that the numeric values below a certain point are unstable: https://github.com/ripple/rippled/blob/develop/src/ripple/protocol/ErrorCodes.h#L40-L41  And they certainly will be unstable after these two commits.

However the online documentation does not seem to spell out that the numeric codes are not stable.  Here's where I would expect to see that indication: https://developers.ripple.com/error-formatting.html  It doesn't assert that the numeric codes are _stable_.  But it also doesn't say they can't be relied on.  There _is_ documentation to say that `TER` codes are unstable: https://developers.ripple.com/transaction-results.html#immediate-response  Still, `TER` codes are a different matter from RPC error codes.

One additional data point is that the numeric values in ErrorCodes.h _have_ been changing over time.  So it seems like users must not be relying on the stability of these values; we probably would have heard.  The last change was in version 1.1.0.  Before that there were changes in versions 0.80.0 and 0.33.0.  And there were lots of changes before 0.33.0.

So my feeling is that we can probably continue to get away with changing these codes.  But if we do we should probably add a warning regarding their stability in https://developers.ripple.com/error-formatting.html

Alternatively, if we decide that the codes should be treated as stable, then I should rework the first commit and throw out the second commit.  I'd be good with that too.

Thanks for your thoughts.

### DIFF
--- a/src/ripple/protocol/impl/ErrorCodes.cpp
+++ b/src/ripple/protocol/impl/ErrorCodes.cpp
@@ -17,137 +17,149 @@
 */
 //==============================================================================
 
-#include <ripple/basics/contract.h>
-#include <ripple/basics/safe_cast.h>
 #include <ripple/protocol/ErrorCodes.h>
 #include <cassert>
-#include <unordered_map>
-#include <utility>
 
-namespace std {
-
-template <>
-struct hash <ripple::error_code_i>
-{
-    explicit hash() = default;
-
-    std::size_t operator() (ripple::error_code_i value) const
-    {
-        return value;
-    }
-};
-
-}
 namespace ripple {
 namespace RPC {
 
 namespace detail {
 
-class ErrorCategory
+// Unordered array of ErrorInfos, so we don't have to maintain the list
+// ordering by hand.
+//
+// This array will be omitted from the object file; only the sorted version
+// will remain in the object file.  But the string literals will remain.
+constexpr static ErrorInfo unorderedErrorInfos[]
 {
-public:
-    using Map = std::unordered_map <error_code_i, ErrorInfo>;
-
-    ErrorCategory ()
-        : m_unknown (rpcUNKNOWN, "unknown", "An unknown error code.")
-    {
-        add (rpcACT_BITCOIN,           "actBitcoin",          "Account is bitcoin address.");
-        add (rpcACT_MALFORMED,         "actMalformed",        "Account malformed.");
-        add (rpcACT_NOT_FOUND,         "actNotFound",         "Account not found.");
-        add (rpcALREADY_MULTISIG,      "alreadyMultisig",     "Already multisigned.");
-        add (rpcALREADY_SINGLE_SIG,    "alreadySingleSig",    "Already single-signed.");
-        add (rpcAMENDMENT_BLOCKED,     "amendmentBlocked",    "Amendment blocked, need upgrade.");
-        add (rpcATX_DEPRECATED,        "deprecated",          "Use the new API or specify a ledger range.");
-        add (rpcBAD_FEATURE,           "badFeature",          "Feature unknown or invalid.");
-        add (rpcBAD_ISSUER,            "badIssuer",           "Issuer account malformed.");
-        add (rpcBAD_MARKET,            "badMarket",           "No such market.");
-        add (rpcBAD_SECRET,            "badSecret",           "Secret does not match account.");
-        add (rpcBAD_SEED,              "badSeed",             "Disallowed seed.");
-        add (rpcBAD_SYNTAX,            "badSyntax",           "Syntax error.");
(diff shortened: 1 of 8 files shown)


src/ripple/protocol/impl/ErrorCodes.cpp
```diff
@@ -50,83 +50,71 @@ class ErrorCategory
...
+        add (rpcSTREAM_MALFORMED,      "malformedStream",     "Stream malformed.");
+        add (rpcTOO_BUSY,              "tooBusy",             "The server is too busy to help you now.");
+        add (rpcTXN_NOT_FOUND,         "txnNotFound",         "Transaction not found.");
+        add (rpcUNKNOWN_COMMAND,       "unknownCmd",          "Unknown method.");
+        add (rpcSENDMAX_MALFORMED,     "sendMaxMalformed",    "SendMax amount malformed.");
+
+        // Verify that the number of entries in m_map equals the number of
+        // enums that have descriptions.  That skips rpcUNKNOWN and rpcSUCCESS,
+        // which are -1 and 0 respectively.
+        assert (static_cast<int>(rpcNOT_SUPPORTED) == m_map.size());
```
thejohnfreeman: Can we add an `rpcNUM_CODES` value (I'm not picky on the name) with a comment at the end of the enum and use that here instead (we'll have to subtract 1, but I'm fine with that)?

src/ripple/rpc/impl/TransactionSign.cpp
```diff
@@ -447,9 +447,20 @@ transactionPreProcessImpl (
...
+    {
+        if (tx_json.isMember (jss::TxnSignature))
+            return rpcError (rpcALREADY_SINGLE_SIG);
+
+        // If multisigning then we need to return the public key.
         signingArgs.setPublicKey (keypair.first);
+    }
+    else
+    {
+        if (tx_json.isMember (sfSigners.jsonName))
```
thejohnfreeman: Can we start on the path to consistency? Add `jss:Signers` to `JsonFields.h` and use it here.

How about adding a `signingArgs.isSingleSigning` method, using it for this branch, and adding a new else-branch that diagnoses an error ("unknown signing method").

scottschurr: I started looking at the consistency problem.  I think it deserves its own pull request that really does the whole job.  Otherwise all we accomplish is stirring mud.  And my personal preference would be that the `SField` objects would not be represented in the `jss` namespace at all.  All `SField` names should be retrieved from their primary source.

I'll put together a separate pull request with that change.

I'll add an `isSingleSigning()` method.  That does help with documentation a bit.  I can't bring myself to justify a third branch to the if-else.  Adding a third kind of signing will be a big deal.  The person adding a third kind of signing will need to add a lot more than a third condition in the if-else.

scottschurr:
At prompting from @thejohnfreeman I investigated what it would take to do error checking of the `ErrorCode` lookup table at compile time.  It turned out not to be that hard.  But it does take a recent enough compiler to swallow the `constexpr` code.

So I've added two commits to this pull request.  One, from @mellery451 (thanks very much to him), pushes our Travis CI ahead to a more recent gcc version.  The gcc version that Travis was previously at could not handle the `for` loop in `constexpr` evaluation.  The last commit takes an unordered list of `ErrorInfo` which is sorted and validated at compile time.  The sorted list is used in place of what was previously a hash map.

These most recent commits are optional.  They add nothing to the functionality of the original pull request.  But I think that validating the `ErrorInfo` to this extent at compile time is the sort of thing we should be doing when we can.

I encourage people to make a local branch and mess around by re-ordering, adding, or commenting out various entries in the `unorderedErrorInfos` toward the top of ErrorCodes.cpp and then re-building.  That will give you a sense for the kinds of compile-time errors we can see from `constexpr` error checking.

scottschurr:
Squashed and rebased.

PR #2849, opened 2019-02-09
Resource limit proxied clients whose IP addresses are known but

mtrippled:
have no corresponding user. Unit tests for unlimited vs limited ports.
Import boost::beast::websocket insert custom headers for unit tests.

### DIFF
--- a/src/ripple/net/impl/RPCCall.cpp
+++ b/src/ripple/net/impl/RPCCall.cpp
@@ -44,6 +44,7 @@
 #include <array>
 #include <iostream>
 #include <type_traits>
+#include <unordered_map>
 
 namespace ripple {
 
@@ -60,7 +61,7 @@ std::string createHTTPPost (
     std::string const& strHost,
     std::string const& strPath,
     std::string const& strMsg,
-    std::map<std::string, std::string> const& mapRequestHeaders)
+    std::unordered_map<std::string, std::string> const& mapRequestHeaders)
 {
     std::ostringstream s;
 
@@ -1264,9 +1265,14 @@ struct RPCCallImp
     }
 
     // Build the request.
-    static void onRequest (std::string const& strMethod, Json::Value const& jvParams,
-        const std::map<std::string, std::string>& mHeaders, std::string const& strPath,
-            boost::asio::streambuf& sb, std::string const& strHost, beast::Journal j)
+    static void onRequest (
+        std::string const& strMethod,
+        Json::Value const& jvParams,
+        std::unordered_map<std::string, std::string> const& headers,
+        std::string const& strPath,
+        boost::asio::streambuf& sb,
+        std::string const& strHost,
+        beast::Journal j)
     {
         JLOG (j.debug()) << "requestRPC: strPath='" << strPath << "'";
 
@@ -1276,7 +1282,7 @@ struct RPCCallImp
                       strHost,
                       strPath,
                       JSONRPCRequest (strMethod, jvParams, Json::Value (1)),
-                      mHeaders);
+                      headers);
     }
 };
 
@@ -1338,7 +1344,8 @@ cmdLineToJSONRPC (std::vector<std::string> const& args, beast::Journal j)
 
 std::pair<int, Json::Value>
 rpcClient(std::vector<std::string> const& args,
-    Config const& config, Logs& logs)
+    Config const& config, Logs& logs,
+    std::unordered_map<std::string, std::string> const& headers)
 {
     static_assert(rpcBAD_SYNTAX == 1 && rpcSUCCESS == 0,
         "Expect specific rpc enum values.");
@@ -1415,7 +1422,8 @@ rpcClient(std::vector<std::string> const& args,
                     config.quiet(),
                     logs,
                     std::bind (RPCCallImp::callRPCHandler, &jvOutput,
-                               std::placeholders::_1));
+                               std::placeholders::_1),
+                    headers);
                 isService.run(); // This blocks until there is no more outstanding async calls.
             }
             if (jvOutput.isMember ("result"))
@@ -1499,7 +1507,8 @@ void fromNetwork (
     std::string const& strPath, std::string const& strMethod,
     Json::Value const& jvParams, const bool bSSL, const bool quiet,
     Logs& logs,
-    std::function<void (Json::Value const& jvInput)> callbackFuncP)
+    std::function<void (Json::Value const& jvInput)> callbackFuncP,
+    std::unordered_map<std::string, std::string> headers)
 {
     auto j = logs.journal ("HTTPClient");
 
@@ -1511,11 +1520,8 @@ void fromNetwork (
     }
 
     // HTTP basic authentication
-    auto const auth = base64_encode(strUsername + ":" + strPassword);
-
-    std::map<std::string, std::string> mapRequestHeaders;
-
-    mapRequestHeaders["Authorization"] = std::string ("Basic ") + auth;
+    headers["Authorization"] = std::string("Basic ") + base64_encode(
+        strUsername + ":" + strPassword);
 
     // Send request
 
@@ -1535,7 +1541,7 @@ void fromNetwork (
             &RPCCallImp::onRequest,
             strMethod,
             jvParams,
-            mapRequestHeaders,
+            headers,
             strPath, std::placeholders::_1, std::placeholders::_2, j),
         RPC_REPLY_MAX_BYTES,
         RPC_NOTIFY,
--- a/src/ripple/resource/impl/ResourceManager.cpp
+++ b/src/ripple/resource/impl/ResourceManager.cpp
@@ -22,7 +22,10 @@
 #include <ripple/basics/chrono.h>
 #include <ripple/basics/Log.h>
 #include <ripple/beast/core/CurrentThreadName.h>
+#include <ripple/beast/net/IPAddressConversion.h>
+#include <boost/asio/ip/address_v4.hpp>
 #include <boost/core/ignore_unused.hpp>
+#include <boost/system/error_code.hpp>
 #include <condition_variable>
 #include <memory>
 #include <mutex>
@@ -46,7 +49,6 @@ public:
         : journal_ (journal)
         , logic_ (collector, stopwatch(), journal)
     {
-        boost::ignore_unused (journal_); // Keep unused journal_ just in case.
         thread_ = std::thread {&ManagerImp::run, this};
     }
 
@@ -69,14 +71,37 @@ public:
         return logic_.newInboundEndpoint (address);
     }
 
+    Consumer newInboundEndpoint (beast::IP::Endpoint const& address,
+        bool const proxy, boost::string_view const& forwardedFor) override
+    {
+        if (! proxy)
+            return newInboundEndpoint(address);
+
+        boost::system::error_code ec;
+        auto const proxiedIp = boost::asio::ip::make_address(
+            forwardedFor.to_string(), ec);
+        if (ec)
+        {
+            journal_.warn() << "forwarded for ("
+                << forwardedFor
+                << ") from proxy "
+                << address.to_string()
+                << " doesn't convert to IP endpoint: "
+                << ec.message();
+            return newInboundEndpoint(address);
+        }
+        return newInboundEndpoint(
+            beast::IPAddressConversion::from_asio(proxiedIp));
+    }
+
     Consumer newOutboundEndpoint (beast::IP::Endpoint const& address) override
     {
         return logic_.newOutboundEndpoint (address);
     }
 
-    Consumer newUnlimitedEndpoint (std::string const& name) override
+    Consumer newUnlimitedEndpoint (beast::IP::Endpoint const& address) override
     {
-        return logic_.newUnlimitedEndpoint (name);
+        return logic_.newUnlimitedEndpoint (address);
     }
 
     Gossip exportConsumers () override
--- a/src/ripple/rpc/Role.h
+++ b/src/ripple/rpc/Role.h
@@ -20,10 +20,13 @@
 #ifndef RIPPLE_SERVER_ROLE_H_INCLUDED
 #define RIPPLE_SERVER_ROLE_H_INCLUDED
 
-#include <ripple/server/Port.h>
+#include <ripple/beast/net/IPEndpoint.h>
 #include <ripple/json/json_value.h>
 #include <ripple/resource/ResourceManager.h>
-#include <ripple/beast/net/IPEndpoint.h>
+#include <ripple/server/Handoff.h>
+#include <ripple/server/Port.h>
+#include <boost/utility/string_view.hpp>
+#include <string>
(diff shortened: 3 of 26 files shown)


mtrippled:
Here's diffs in the boost::beast::websocket files compared with boost 1.67:

diff -r websocket/impl/handshake.ipp /home/mtravis/proj/boost_1_67_0/boost/beast/websocket/impl/handshake.ipp
13c13
---
28d27
58,59c57
<             Decorator const& decorator,
<             std::unordered_map<std::string, std::string> const& headers = {})
---
>             Decorator const& decorator)
63c61
<                 host, target, decorator, headers))
---
>                 host, target, decorator))
258,259c256
<     string_view target,
<     std::unordered_map<std::string, std::string> const& headers)
---
>     string_view target)
265c262
<         host, target, ec, headers);
---
>         host, target, ec);
328,329c325
<     string_view target, error_code& ec,
<     std::unordered_map<std::string, std::string> const& headers)
---
>     string_view target, error_code& ec)
334c330
<         host, target, &default_decorate_req, ec, headers);
---
>         host, target, &default_decorate_req, ec);
399,400c395
<     error_code& ec,
<     std::unordered_map<std::string, std::string> const& headers)
---
>     error_code& ec)
407c402
<             key, host, target, decorator, headers);
---
>             key, host, target, decorator);

diff -r websocket/impl/stream.ipp /home/mtravis/proj/boost_1_67_0/boost/beast/websocket/impl/stream.ipp
13,16c13,16
---
35,36d34
637,638c635
<         Decorator const& decorator,
<         std::unordered_map<std::string, std::string> const& headers)
---
>         Decorator const& decorator)
650,651d646
<     for (auto const& h : headers)
< 	req.set(h.first.c_str(), h.second.c_str());

diff -r websocket/stream.hpp /home/mtravis/proj/boost_1_67_0/boost/beast/websocket/stream.hpp
14,25c14,25
---
40d39
760,761c759
<         string_view target,
<         std::unordered_map<std::string, std::string> const& headers = {});
---
>         string_view target);
976,977c974
<         error_code& ec,
<         std::unordered_map<std::string, std::string> const& headers = {});
---
>         error_code& ec);
3459,3460c3456
<                 Decorator const& decorator,
<                 std::unordered_map<std::string, std::string> const& headers = {});
---
>                 Decorator const& decorator);
3557,3558c3553
<                 error_code& ec,
<                     std::unordered_map<std::string, std::string> const& headers = {});
---
>                 error_code& ec);

miguelportilla:
It is not necessary to add the beast websocket class to rippled, which will present a problem with upstream changes. You may need to rework the wsclient files in test to get it working.

mtrippled:
How do I insert custom headers for websocket client connections?

mtrippled:
And I'm OK with not including that particular test--the only use for it in
our code is for this particular test, which actually isn't part of the
initial change I was trying to make. But I figured that since there was no
test coverage for the related code either, I should make tests for that as
well.

mtrippled:
Basically, here's where beast needs to change:
https://github.com/ripple/rippled/pull/2849/files#diff-66c131c3f746149a3a6c3da857c6ed52L140
The handshake() function is where the initial HTTP request is made. It upgrades the connection to websocket. There's no way to insert custom headers into this currently.

vinniefalco:
> The handshake() function is where the initial HTTP request is made. It upgrades the connection to websocket. There's no way to insert custom headers into this currently.

Covered here:
https://www.boost.org/doc/libs/1_69_0/libs/beast/doc/html/beast/using_websocket/handshaking_clients.html#beast.using_websocket.handshaking_clients.decorators

mtrippled:
> > The handshake() function is where the initial HTTP request is made. It upgrades the connection to websocket. There's no way to insert custom headers into this currently.
> 
> Covered here:
> https://www.boost.org/doc/libs/1_69_0/libs/beast/doc/html/beast/using_websocket/handshaking_clients.html#beast.using_websocket.handshaking_clients.decorators

Thanks, @vinniefalco . @mellery451 pointed me to an overload of handshake_ex() that lets me pass in a lambda to set custom headers. I'll try that instead.

mtrippled:
> It is not necessary to add the beast websocket class to rippled, which will present a problem with upstream changes. You may need to rework the wsclient files in test to get it working.

@miguelportilla fixed

mtrippled:
> I agree with @miguelportilla ... we can't be duplicating beast source code here like this. What's more, I guess I don't understand why you can't use the existing handshake method that let's you pass a lambda that can modify the `req` object: https://github.com/boostorg/beast/blob/develop/include/boost/beast/websocket/stream.hpp#L975-L980

@mellery451 fixed

src/ripple/net/impl/RPCCall.cpp
```diff
@@ -1264,9 +1265,12 @@ struct RPCCallImp
     }
 
     // Build the request.
-    static void onRequest (std::string const& strMethod, Json::Value const& jvParams,
-        const std::map<std::string, std::string>& mHeaders, std::string const& strPath,
-            boost::asio::streambuf& sb, std::string const& strHost, beast::Journal j)
+    static void onRequest (std::string const& strMethod,
+        Json::Value const& jvParams,
+        const std::unordered_map<std::string,
+        std::string>& mHeaders, std::string const& strPath,
```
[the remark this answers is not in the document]
mtrippled: fixed

src/ripple/resource/impl/ResourceManager.cpp
```diff
@@ -69,14 +71,37 @@ class ManagerImp : public Manager
...
     }
 
+    Consumer newInboundEndpoint (beast::IP::Endpoint const& address,
+        bool const proxy, std::string const& forwardedFor) override
+    {
+        if (! proxy)
+            return newInboundEndpoint(address);
+
+        boost::system::error_code ec;
+        auto const proxiedIp = boost::asio::ip::make_address_v4(
```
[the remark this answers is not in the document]
mtrippled: fixed so either IP address type is converted. Fixed support for comma-delimited X-Forwarded-For and now also Forwarded header with its syntax. New unit tests for those header fields.

src/ripple/net/impl/RPCCall.cpp
```diff
@@ -1267,9 +1267,10 @@ struct RPCCallImp
     // Build the request.
     static void onRequest (std::string const& strMethod,
         Json::Value const& jvParams,
-        const std::unordered_map<std::string,
-        std::string>& mHeaders, std::string const& strPath,
-        boost::asio::streambuf& sb, std::string const& strHost,
+        std::unordered_map<std::string, std::string> const& mHeaders,
+        std::string const& strPath,
```
vinniefalco: Consider `std::string_view` instead of `std::string const&`

seelabs: `string_view` is very nice. Unfortunately `std::string_view` is C++-17 only and we're still on 14.

vinniefalco: Oh? I'm surprised...C++14 is a minor update, most people either stay on C++11 or go straight to C++17. What is the obstacle?

seelabs: @vinniefalco Mostly it's just concervatism and making sure we have a good solution for all our supported platforms (centos is usually the most behind). I expect we'll upgrade to 17 soon; there are good things in there that would make rippled nicer.

vinniefalco: You can always write
```
namespace ripple {
using string_view = boost::string_view;
} // ripple
```

And then change that to `std::string_view` later

mtrippled: Is there some mechanism to ensure that the original ```char *``` for the string_view stays in scope? I understand that avoiding a string copy is desirable, but is going back to consciously managing the lifetime of strings worth it?

vinniefalco: Howard could give you a concise answer but I can put it a different way - are you okay making unnecessary copies of the string depending on the type of argument at the call site?

src/ripple/rpc/Role.h
```diff
@@ -52,13 +55,12 @@ enum class Role
 */
 Role
 requestRole (Role const& required, Port const& port,
```
mellery451: the implementation of this function still has the old signature (thus the no-unity build failures)

mtrippled: fixed

src/ripple/rpc/impl/Role.cpp
```diff
@@ -74,40 +79,61 @@ requestRole (Role const& required, Port const& port,
...
+        std::string forwardedFor = it->value().to_string();
+        std::size_t const pos = forwardedFor.find(',');
+        if (pos != forwardedFor.npos)
+            forwardedFor.resize(pos);
+        return forwardedFor;
+    }
+    it = request.find("Forwarded");
+    if (it != request.end())
+    {
+        std::stringstream ss(it->value().to_string());
```
mellery451: since we don't really care about all the parts of `Forwarded`, we could just do something like:
```
        static std::string const forStr{"for="};
        auto forStart = it->value().find(forStr);
        if (forStart != boost::beast::string_view::npos)
        {
            forStart += forStr.size();
            const auto forEnd = it->value().find(";", forStart);
            return it->value().substr(forStart, forEnd-forStart).to_string();
        }
```
..which I think avoids a couple of temp strings.

vinniefalco: is the field a list of semicolon-delimited values? There might be something in here that could help:
https://github.com/boostorg/beast/blob/develop/include/boost/beast/http/rfc7230.hpp#L34

src/ripple/rpc/impl/Role.cpp
```diff
@@ -74,40 +79,61 @@ requestRole (Role const& required, Port const& port,
...
+        std::string directive;
+        while (std::getline(ss, directive, ';'))
+        {
+            std::size_t const found = directive.find(forStr);
```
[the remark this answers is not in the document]
vinniefalco: Note that `std::toupper` uses the current locale, but you probably want to limit it to low-ascii:
https://en.cppreference.com/w/cpp/locale/toupper

You can use the beast ascii character operations:
https://github.com/boostorg/beast/blob/develop/include/boost/beast/core/string.hpp#L49

src/ripple/rpc/Role.h
```diff
@@ -52,13 +53,12 @@ enum class Role
...
-    Json::Value const& jsonRPC, beast::IP::Endpoint const& remoteIp,
-    std::string const& user);
+    Json::Value const& jsonRPC, beast::IP::Endpoint const& remoteIp);
```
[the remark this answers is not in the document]
mtrippled: This was fixed in 4eb0217f8c7d838fe3de0441a8fb550b82d24a92

src/ripple/resource/impl/ResourceManager.cpp
```diff
@@ -69,14 +71,37 @@ class ManagerImp : public Manager
         return logic_.newInboundEndpoint (address);
     }
 
+    Consumer newInboundEndpoint (beast::IP::Endpoint const& address,
+        bool const proxy, std::string const& forwardedFor) override
```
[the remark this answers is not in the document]
mtrippled: More complex to put that logic in the callers imo.

src/ripple/rpc/impl/Role.cpp
mtrippled: @vinniefalco I'd like to use beast::param_list, but the format of the "Forwarded" value uses ";" as a delimiter between field-value pairs, and not as a start-of-record as in rfc7230.

mtrippled: @vinniefalco However, ext_list is suitable for simplyfing how I extract from X-Forwarded-For.

src/ripple/net/impl/RPCCall.cpp
mtrippled: > Consider `std::string_view` instead of `std::string const&`

Based on your suggestion, I put string_view in for the server-side part of handling headers, but not the part that you suggested because the client side of this is just for testing, and string_view lacks a hash comparitor for it to live in an unordered_map. Not worth the hassle just for a couple of unit tests. But for the server side handling of headers, I think it should be beneficial.

src/ripple/net/impl/RPCCall.cpp
```diff
@@ -1264,9 +1265,12 @@ struct RPCCallImp
...
     // Build the request.
-    static void onRequest (std::string const& strMethod, Json::Value const& jvParams,
-        const std::map<std::string, std::string>& mHeaders, std::string const& strPath,
-            boost::asio::streambuf& sb, std::string const& strHost, beast::Journal j)
+    static void onRequest (std::string const& strMethod,
+        Json::Value const& jvParams,
+        const std::unordered_map<std::string,
+        std::string>& mHeaders, std::string const& strPath,
+        boost::asio::streambuf& sb, std::string const& strHost,
+        beast::Journal j)
```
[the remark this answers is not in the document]
mtrippled: fixed

mtrippled:
Based on your suggestions, I refactored the forwardedFor() function to more efficiently extract those headers, including case insensitivity, locale-awareness, and eliminating string copies.

src/ripple/net/impl/RPCCall.cpp
vinniefalco: Entirely reasonable

src/ripple/rpc/impl/Role.cpp
```diff
@@ -104,36 +105,43 @@ requestInboundEndpoint (Resource::Manager& manager,
...
     {
-        std::stringstream ss(it->value().to_string());
-        std::string directive;
-        while (std::getline(ss, directive, ';'))
+        static std::string const forStr{"for="};
+        auto found = std::search(it->value().begin(), it->value().end(),
+            forStr.begin(), forStr.end(),
+            [](char c1, char c2)
+            {
+                return boost::beast::detail::ascii_tolower(c1) ==
```
vinniefalco: See https://github.com/boostorg/beast/issues/1512

src/ripple/rpc/impl/Role.cpp
```diff
@@ -104,36 +105,43 @@ requestInboundEndpoint (Resource::Manager& manager,
...
+        for (n = 0; n < std::distance(&val, it->value().end()); ++n)
         {
-            std::size_t const found = directive.find(forStr);
-            if (found == directive.npos)
-                continue;
-            return directive.substr(found + forStrLen);
+            if (*(&val + n) == ';')
+                break;
         }
+        return boost::string_view{&val, n};
```
miguelportilla: Consider this alternative:
```
        found += forStr.size();
        auto pos{it->value().find(';', forStr.size())};
        if (pos != boost::string_view::npos)
            return {found, pos + 1};
        return {found, static_cast<size_t>(std::distance(found, it->value().end()))};
```

mtrippled: Thanks, @miguelportilla -- this is preferable.

PR #2854, opened 2019-02-14
 Remove runtime inference of unrecognized SFields

scottschurr:
For years the `SField` implementation has been able to infer certain classes of `SField`s from serialized data.  This capability gives the `SField` handling a certain degree of "forward compatibility".  It allows a server to handle `SField`s that another server might know about, but the local server does not.

However, inferring `SField`s comes with a disadvantage too.  Given the current system with amendments, all servers that can successfully participate in the protocol should have hard definitions for every `SField` used in communications.  So there should never be any need to infer `SField`s; the definitions should in all cases be known.  If an unknown `SField` is encountered it should be flagged as an error and not inferred as a new and unknown `SField`.

This branch removes the ability to infer `SField`s.

As a sanity check, I ran this branch against the network for 12 hours.  During that run the server never lost sync with the network and I shutdown the server normally.  That suggests to me that there is probably not some unexpected reliance on `SField` inference.

With this change, malformed serialized data is detected sooner.  That affected a few of our unit tests which contained malformed serialized data.  The affected unit tests now fail with an "Unknown field" rather than their intended error condition.  In one case I was able to tweak the serialized data so it no longer contained inferred `SField`s.  In other cases I added unit tests that use well known `SField`s to recreate the desired error condition.

Only the top-most commit needs to be reviewed.  That commit sits atop a different pull request by @ximinez who is also making improvements to `SField`s.  I'm assuming that the underlying pull request will be merged and the top-most commit of this branch can be merged afterwards (assuming it passes code review).

### DIFF
--- a/src/ripple/protocol/impl/SOTemplate.cpp
+++ b/src/ripple/protocol/impl/SOTemplate.cpp
@@ -28,18 +28,20 @@ void SOTemplate::push_back (SOElement const& r)
     //
     if (mIndex.empty ())
     {
-        // Unmapped indices will be set to -1
+        // Unmapped indices are initialized to -1
         //
         mIndex.resize (SField::getNumFields () + 1, -1);
     }
 
     // Make sure the field's index is in range
     //
-    assert (r.e_field.getNum () < mIndex.size ());
+    if (r.e_field.getNum() <= 0 || r.e_field.getNum() >= mIndex.size())
+        Throw<std::runtime_error> ("Invalid field index for SOTemplate.");
 
     // Make sure that this field hasn't already been assigned
     //
-    assert (getIndex (r.e_field) == -1);
+    if (getIndex (r.e_field) != -1)
+        Throw<std::runtime_error> ("Duplicate field index for SOTemplate.");
 
     // Add the field to the index mapping table
     //
--- a/src/test/protocol/STTx_test.cpp
+++ b/src/test/protocol/STTx_test.cpp
@@ -28,6 +28,8 @@
 #include <ripple/basics/Slice.h>
 #include <ripple/protocol/messages.h>
 
+#include <memory>
+
 namespace ripple {
 
 class STTx_test : public beast::unit_test::suite
@@ -1218,7 +1220,7 @@ public:
             0x12, 0x12, 0x12, 0xff
         };
 
-        constexpr unsigned char dupField[] =
+        constexpr unsigned char payload3[] =
         {
             0x12, 0x00, 0x65, 0x24, 0x00, 0x00, 0x00, 0x00, 0x20, 0x1e, 0x00, 0x4f,
             0x00, 0x00, 0x20, 0x1f, 0x03, 0xf6, 0x00, 0x00, 0x20, 0x20, 0x00, 0x00,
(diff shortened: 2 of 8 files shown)


src/ripple/protocol/impl/SOTemplate.cpp
```diff
@@ -28,18 +28,20 @@ void SOTemplate::push_back (SOElement const& r)
...
+        // Unmapped indices are initialized to -1
         //
         mIndex.resize (SField::getNumFields () + 1, -1);
     }
 
     // Make sure the field's index is in range
     //
-    assert (r.e_field.getNum () < mIndex.size ());
+    if (r.e_field.getNum() <= 0 || r.e_field.getNum() >= mIndex.size())
+        Throw<std::runtime_error> ("Invalid field index for SOTemplate.");
```
mellery451: I'm curious about your take on exceptions here. As near as I can tell, this code will be accessed by a couple of static instance constructors, but the items we pass are fixed/known - so won't ever throw? I realize we'd probably like a violation of these invariants to be caught at compile time, but that's tricky here. Is there a scenario outside of unit tests that would trigger an exception and will it be handled as we expect?

src/test/protocol/STTx_test.cpp
```diff
@@ -1243,6 +1245,173 @@ class STTx_test : public beast::unit_test::suite
...
+                fail("An exception should have been thrown");
+            }
+            catch (std::exception const& ex)
+            {
+                BEAST_EXPECT(
+                    strcmp(ex.what(), "Duplicate field detected") == 0);
+            }
+        }
+
+        // Exercise the three raw transactions.
```
mellery451: nice test!

scottschurr:
Thanks for the review @mellery451.  I'm not ignoring your comments, but I'm tied up elsewhere for the rest of the week.  I'll be able to respond to your comments early next week...

src/ripple/protocol/impl/SOTemplate.cpp
scottschurr: That's a fair question.  Let me see if I can recreate what I was thinking when I made this change.

I believe my concern was that `assert`s only occur in a debug build.  If a bad `SField` is passed here then previously the bounds checking only occurred in a debug build.  In a release build we would index out of range, leading to unknown bad things.

Also, `SOTemplate` construction is relatively rare so, in terms of execution time, performing the validation in a release build adds very little aggregate runtime cost.  I think we can afford to do this validation in a release build.

It just increased my confidence a bit if we do the `SOTemplate` validation checking all the time, not just in a debug build.  And that's not something I get from an `assert`.  But I also didn't want to blindly shutdown a release-build of a server, which is why I did not use `LogicError`.

So I can't say I have clear evidence that this is a necessary change.  I can't point to a way that foreign code could cause an invalid `SOTemplate` to be constructed in a release build.  I simply preferred to put in a check that both can't be ignored and doesn't shutdown the server.

The consequence of the `throw` would be that the `SOTemplate` would still exist, but it would not contain all of the intended fields.  That's because fields are added by `SOTemplate::push_back()` not in the `SOTemplate` constructor.

That could be an excellent argument to add a constructor to the `SOTemplate` that takes a `std::initializer_list<SOElement>`.  I think if I did that I could remove `SOTemplate::push_back`.  That would lead to the consequence that if an invalid field were part of the `SOTemplate` the `SOTemplate` constructor would throw, so the entire `SOTemplate` with the invalid field would not even exist.

Let me know if you think adding the `std::initializer_list<SOElement>` constructor would be a worthwhile change.  It feels like a good idea to me at the moment.  There are few enough `SOTemplates` that I believe I could make such a change part of this pull request.  Or I could make a separate pull request...

mellery451: I'm fine with either approach - thanks for clarifying. If you think `initializer_list` makes the error state slightly saner, that sounds fine to me.

nbougalis: Consider the change I made in #2863, allowing a function to be passed into the `SOTemplate` constructor:

https://github.com/ripple/rippled/blob/9a4583042d0166c44daca829efe9cc80de19a845/src/ripple/protocol/SOTemplate.h#L80-L92

See it in use here: https://github.com/ripple/rippled/blob/9a4583042d0166c44daca829efe9cc80de19a845/src/ripple/app/misc/impl/Manifest.cpp#L42

I'd _prefer_ that we move to this sort of construction for _all_ `SOTemplate` objects going forward and eschew the piece-meal approach we have now.

src/test/protocol/STTx_test.cpp
nbougalis: Yeah, agreed. Dang!

src/ripple/protocol/impl/SOTemplate.cpp
scottschurr: I like your suggestion, but I also have a different approach which calls the `SOTemplate` constructor with a `std::initializer_list`.  I just about have the `initializer_list` approach coded up and tested.  I'm going to submit it as a separate pull request because the changes are too extensive to just slide on top of this pull request.  With luck I'll have that pull request ready today or tomorrow.

Once you see the `initializer_list` approach we can talk about whether we prefer that or the callback.  My feeling is we should only use one approach or the other, and I personally don't have a strong preference at the moment.

scottschurr:
@ximinez, thanks for the :+1:.

Since this pull request is currently designed to sit on top of your pull request https://github.com/ripple/rippled/pull/2770, I'll wait until that pull request is marked as passed and is squashed.  Then I'll rebase this pull request on top of yours and resolve the conflicts.  At that point I'l mark this pull request passed.

scottschurr:
Incorporated into https://github.com/ripple/rippled/pull/2869.  Closing this pull request so the code is not accidentally merged to develop twice.

PR #2855, opened 2019-02-15
Properly transition state to disconnected:

nbougalis:
If the number of peers a server has is below the configured minimum peer limit, this commit will properly transition the server's state to "disconnected".

The default limit for the minimum number of peers required was 0 meaning that a server that was connected but lost all its peers would never transition to disconnected, since it could never drop below zero peers.

This commit redefines the default minimum number of peers to 1 and produces a warning if the server is configured in a way that will prevent it from ever achieving sufficient connectivity.

### DIFF
--- a/src/ripple/app/misc/NetworkOPs.cpp
+++ b/src/ripple/app/misc/NetworkOPs.cpp
@@ -196,7 +196,7 @@ public:
     // VFALCO TODO Make LedgerMaster a SharedPtr or a reference.
     //
     NetworkOPsImp (Application& app, NetworkOPs::clock_type& clock,
-        bool standalone, std::size_t network_quorum, bool start_valid,
+        bool standalone, std::size_t minPeerCount, bool start_valid,
         JobQueue& job_queue, LedgerMaster& ledgerMaster, Stoppable& parent,
         ValidatorKeys const & validatorKeys, boost::asio::io_service& io_svc,
         beast::Journal journal)
@@ -220,7 +220,7 @@ public:
         , m_ledgerMaster (ledgerMaster)
         , m_job_queue (job_queue)
         , m_standalone (standalone)
-        , m_network_quorum (start_valid ? 0 : network_quorum)
+        , minPeerCount_ (start_valid ? 0 : minPeerCount)
     {
     }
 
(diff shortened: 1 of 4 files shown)


HowardHinnant:
It looks like the unit tests need to be tweaked for this change.

src/ripple/app/misc/NetworkOPs.cpp
```diff
@@ -598,7 +598,7 @@ class NetworkOPsImp final
     bool const m_standalone;
 
     // The number of nodes that we need to consider ourselves connected.
-    std::size_t const m_network_quorum;
+    std::size_t const minPeerCount_;
```
thejohnfreeman: Another mix of naming conventions for members: trailing underscore and leading `m`.

src/ripple/app/misc/NetworkOPs.cpp
```diff
@@ -196,7 +196,7 @@ class NetworkOPsImp final
     // VFALCO TODO Make LedgerMaster a SharedPtr or a reference.
     //
     NetworkOPsImp (Application& app, NetworkOPs::clock_type& clock,
-        bool standalone, std::size_t network_quorum, bool start_valid,
+        bool standalone, std::size_t minPeerCount, bool start_valid,
```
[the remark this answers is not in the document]
nbougalis: We do want to be consistent. Several files contain an arcane mix of styles. Really old, old old, new old, old new, new new... it's weird.

I'm not a huge fan of going through and renaming _everything_ at once, especially as part of some other fix, and I'd normally leave this well enough alone, except the name was bad to begin with. Since I was fixing it, I decided to also move it to the current preferred style.

I don't feel that strongly about it though and am happy to just rename it `m_min_peer_count`, even if that's way too many underscores for my overworked SHIFT finger.

src/ripple/app/misc/NetworkOPs.cpp
nbougalis: Agreed; responded in the other comment.

nbougalis:
Merged as 2529edd2b6d107256170ddcc1045f69a29a0d954

PR #2858, opened 2019-02-18
Refine parseUrl regular expression

thejohnfreeman:
We need narrower character sets to handle pathological cases. In case of
an exception, just bail.

Fixes RIPD-1708

### DIFF
--- a/src/ripple/basics/impl/StringUtilities.cpp
+++ b/src/ripple/basics/impl/StringUtilities.cpp
@@ -110,36 +110,34 @@ bool parseUrl (parsedURL& pUrl, std::string const& strUrl)
         "\\s*?\\'");
     boost::smatch smMatch;
 
-    bool bMatch = false;
+    // Bail if there is no match.
     try {
-      bMatch = boost::regex_match (strUrl, smMatch, reUrl); // Match status code.
+        if (! boost::regex_match (strUrl, smMatch, reUrl))
+            return false;
     } catch (...) {
-      return false;
+        return false;
     }
 
-    if (bMatch)
+    pUrl.scheme = smMatch[1];
+    boost::algorithm::to_lower (pUrl.scheme);
+    pUrl.username = smMatch[2];
+    pUrl.password = smMatch[3];
+    const std::string domain = smMatch[4];
+    // We need to use Endpoint to parse the domain to
+    // strip surrounding brackets from IPv6 addresses,
+    // e.g. [::1] => ::1.
+    const auto result {beast::IP::Endpoint::from_string_checked (domain)};
+    pUrl.domain = result.second
+        ? result.first.address().to_string()
+        : domain;
+    const std::string port = smMatch[5];
+    if (!port.empty())
     {
-        pUrl.scheme = smMatch[1];
-        boost::algorithm::to_lower (pUrl.scheme);
-        pUrl.username = smMatch[2];
-        pUrl.password = smMatch[3];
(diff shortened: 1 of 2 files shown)


thejohnfreeman:
The problem is this URI: `http://::1:1234/validators`. `::1:1234` is an IPv6 address.

- [RFC3986 requires that it be enclosed in square brackets.](https://tools.ietf.org/html/rfc3986#appendix-A).
- RFC2396 excludes IPv6 addresses. See [3.2.2](https://tools.ietf.org/html/rfc2396#section-3.2.2): "Literal IPv6 addresses are not supported." The [recommended regex](https://tools.ietf.org/html/rfc2396#appendix-B) (`^(([^:/?#]+):)?(//([^/?#]*))?([^?#]*)(\?([^#]*))?(#(.*))?`) happily consumes it as the authority, but implementations, e.g. [java.net.URI](https://repl.it/repls/StudiousCourteousPixel), leave it to the developer to parse the authority into user info, host, and port. This two-phase parsing is what we wanted to avoid in the first place by consolidating to one regex.

How can we tell that `:1234` is not a port without baking a large IPv6 parser into our regex? I'm going to change the regex so that if you use an unenclosed IPv6 address, you don't get to use a port.

thejohnfreeman:
Separately, maybe we can get our code to behave by enclosing its IPv6 addresses.

src/test/basics/StringUtilities_test.cpp
```diff
@@ -259,6 +259,15 @@ class StringUtilities_test : public beast::unit_test::suite
             BEAST_EXPECT(! parseUrl (pUrl, "://"));
             BEAST_EXPECT(! parseUrl (pUrl, ":///"));
         }
+
+        {
+          std::string strUrl("s://");
+          for (size_t i = 0; i < 8192; i++) {
+            strUrl += ":";
+          }
```
miguelportilla: Alternative:
`std::string strUrl("s://" + std::string(8192, ':'));`

Nit: use four space indent

src/ripple/basics/impl/StringUtilities.cpp
```diff
@@ -100,17 +100,22 @@ bool parseUrl (parsedURL& pUrl, std::string const& strUrl)
...
         "\\s*?\\'");
     boost::smatch smMatch;
 
-    bool bMatch = boost::regex_match (strUrl, smMatch, reUrl); // Match status code.
+    bool bMatch = false;
+    try {
+      bMatch = boost::regex_match (strUrl, smMatch, reUrl); // Match status code.
+    } catch (...) {
+      return false;
+    }
```
miguelportilla: Consider omitting `bMatch`
```
    try {
        if (! boost::regex_match (strUrl, smMatch, reUrl)) // Match status code.
            return false;
    }
    catch (...) {
        return false;
    }

    pUrl.scheme = smMatch[1];
    ...

    return true;
```

thejohnfreeman:
@mellery451 let me know that we should ignore the Jenkins build in the comments and just pay attention to the Jenkins build (continuous-integration/jenkins/pr-head) in the PR status block that appears at the end.

mellery451:
@thejohnfreeman yes, please ignore that one failure. If that's your only hold-up, you should mark as passed

nbougalis:
This has been merged as https://github.com/ripple/rippled/commit/0c20e2eb8b33e45b35628ff0a721c609624c6719

PR #2862, opened 2019-02-21
Add `delivered_amount` to ledger RPC and websocket commands

seelabs:
This PR adds a synthetic `delivered_amount` to payment and check_cash transactions for the ledger RPC and websocket commands. It only adds these synthetic fields for non-binary requests only.

Note, there are some minor bugs in earlier commits that are fixed in later commits (such as using `meta` instead of `metaData` for the RPC command). Normally I would squash these down so the bug wouldn't be present while reviewing. However, the commit that moves the function out of the `impl` directory makes this difficult to do. I also opted to keep the commits as-is rather than squash them into a single commit, as I believe it's easier to review with some of the refactoring commits kept separate.

### DIFF
--- a/src/ripple/app/ledger/impl/LedgerToJson.cpp
+++ b/src/ripple/app/ledger/impl/LedgerToJson.cpp
@@ -21,6 +21,8 @@
 #include <ripple/app/misc/TxQ.h>
 #include <ripple/basics/base_uint.h>
 #include <ripple/basics/date.h>
+#include <ripple/rpc/Context.h>
+#include <ripple/rpc/DeliveredAmount.h>
 
 namespace ripple {
 
@@ -97,49 +99,63 @@ void fillJsonBinary(Object& json, bool closed, LedgerInfo const& info)
     }
 }
 
-Json::Value fillJsonTx (LedgerFill const& fill,
-    bool bBinary, bool bExpanded,
-        std::pair<std::shared_ptr<STTx const>,
-            std::shared_ptr<STObject const>> const i)
+Json::Value
+fillJsonTx(
+    LedgerFill const& fill,
+    bool bBinary,
+    bool bExpanded,
+    std::shared_ptr<STTx const> const& txn,
+    std::shared_ptr<STObject const> const& stMeta)
 {
-    if (! bExpanded)
+    if (!bExpanded)
+        return to_string(txn->getTransactionID());
+
+    Json::Value txJson{Json::objectValue};
+    auto const txnType = txn->getTxnType();
+    if (bBinary)
     {
-        return to_string(i.first->getTransactionID());
+        txJson[jss::tx_blob] = serializeHex(*txn);
+        if (stMeta)
+            txJson[jss::meta] = serializeHex(*stMeta);
     }
     else
     {
-        Json::Value txJson{ Json::objectValue };
-        if (bBinary)
+        copyFrom(txJson, txn->getJson(0));
+        if (stMeta)
         {
-            txJson[jss::tx_blob] = serializeHex(*i.first);
-            if (i.second)
-                txJson[jss::meta] = serializeHex(*i.second);
-        }
-        else
-        {
-            copyFrom(txJson, i.first->getJson(0));
-            if (i.second)
-                txJson[jss::metaData] = i.second->getJson(0);
-        }
-
-        if ((fill.options & LedgerFill::ownerFunds) &&
-            i.first->getTxnType() == ttOFFER_CREATE)
-        {
-            auto const account = i.first->getAccountID(sfAccount);
-            auto const amount = i.first->getFieldAmount(sfTakerGets);
-
-            // If the offer create is not self funded then add the
-            // owner balance
-            if (account != amount.getIssuer())
+            txJson[jss::metaData] = stMeta->getJson(0);
+            if (txnType == ttPAYMENT || txnType == ttCHECK_CASH)
             {
-                auto const ownerFunds = accountFunds(fill.ledger,
-                    account, amount, fhIGNORE_FREEZE,
-                    beast::Journal {beast::Journal::getNullSink()});
-                txJson[jss::owner_funds] = ownerFunds.getText ();
+                // Insert delivered amount
+                auto txMeta = std::make_shared<TxMeta>(
+                    txn->getTransactionID(), fill.ledger.seq(), *stMeta);
+                RPC::insertDeliveredAmount(
+                    txJson[jss::metaData], fill.ledger, txn, *txMeta);
             }
         }
-        return txJson;
     }
+
+    if ((fill.options & LedgerFill::ownerFunds) &&
+        txn->getTxnType() == ttOFFER_CREATE)
+    {
+        auto const account = txn->getAccountID(sfAccount);
+        auto const amount = txn->getFieldAmount(sfTakerGets);
+
+        // If the offer create is not self funded then add the
+        // owner balance
+        if (account != amount.getIssuer())
+        {
+            auto const ownerFunds = accountFunds(
+                fill.ledger,
+                account,
+                amount,
+                fhIGNORE_FREEZE,
+                beast::Journal{beast::Journal::getNullSink()});
+            txJson[jss::owner_funds] = ownerFunds.getText();
+        }
+    }
+
+    return txJson;
 }
 
 template <class Object>
@@ -153,7 +169,7 @@ void fillJsonTx (Object& json, LedgerFill const& fill)
     {
         for (auto& i: fill.ledger.txs)
         {
-            txns.append(fillJsonTx(fill, bBinary, bExpanded, i));
+            txns.append(fillJsonTx(fill, bBinary, bExpanded, i.first, i.second));
         }
     }
     catch (std::exception const&)
@@ -219,8 +235,7 @@ void fillJsonQueue(Object& json, LedgerFill const& fill)
         if (tx.lastResult)
             txJson["last_result"] = transToken(*tx.lastResult);
 
-        txJson[jss::tx] = fillJsonTx(fill, bBinary, bExpanded,
-            std::make_pair(tx.txn, nullptr));
+        txJson[jss::tx] = fillJsonTx(fill, bBinary, bExpanded, tx.txn, nullptr);
     }
 }
 
--- a/src/ripple/rpc/DeliveredAmount.h
+++ b/src/ripple/rpc/DeliveredAmount.h
@@ -1,0 +1,68 @@
+//------------------------------------------------------------------------------
+/*
+    This file is part of rippled: https://github.com/ripple/rippled
+    Copyright (c) 2019 Ripple Labs Inc.
+
+    Permission to use, copy, modify, and/or distribute this software for any
+    purpose  with  or without fee is hereby granted, provided that the above
+    copyright notice and this permission notice appear in all copies.
+
+    THE  SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+    WITH  REGARD  TO  THIS  SOFTWARE  INCLUDING  ALL  IMPLIED  WARRANTIES  OF
+    MERCHANTABILITY  AND  FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+    ANY  SPECIAL ,  DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+    WHATSOEVER  RESULTING  FROM  LOSS  OF USE, DATA OR PROFITS, WHETHER IN AN
+    ACTION  OF  CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+    OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+//==============================================================================
+
+#ifndef RIPPLE_RPC_DELIVEREDAMOUNT_H_INCLUDED
+#define RIPPLE_RPC_DELIVEREDAMOUNT_H_INCLUDED
+
+#include <memory>
+
+namespace Json {
+class Value;
+}
+
+namespace ripple {
+
+class ReadView;
+class Transaction;
+class TxMeta;
+class STTx;
+
+namespace RPC {
+
+struct Context;
+
+/**
+   Add a `delivered_amount` field to the `meta` input/output parameter.
+   The field is only added to successful payment and check cash transactions.
+   If a delivered amount field is available in the TxMeta parameter, that value
+   is used. Otherwise, the transaction's `Amount` field is used. If neither is available,
+   then the delivered amount is set to "unavailable".
+
+   @{
+ */
+void
+insertDeliveredAmount(
+    Json::Value& meta,
+    ReadView const&,
+    std::shared_ptr<STTx const> serializedTx,
+    TxMeta const&);
+
+void
+insertDeliveredAmount(
+    Json::Value& meta,
+    Context&,
+    std::shared_ptr<Transaction>,
+    TxMeta const&);
+
+/** @} */
+
+} // RPC
+} // ripple
+
+#endif
--- a/src/ripple/rpc/handlers/Tx.cpp
+++ b/src/ripple/rpc/handlers/Tx.cpp
@@ -26,6 +26,7 @@
 #include <ripple/protocol/ErrorCodes.h>
 #include <ripple/protocol/JsonFields.h>
 #include <ripple/rpc/Context.h>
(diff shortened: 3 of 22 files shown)


src/ripple/rpc/handlers/Tx.cpp
```diff
@@ -130,7 +130,7 @@ Json::Value doTx (RPC::Context& context)
                     lgr->seq (), *rawMeta, context.app.journal ("TxMeta"));
                 okay = true;
                 auto meta = txMeta->getJson (0);
-                addPaymentDeliveredAmount (meta, context, txn, txMeta);
+                addPaymentDeliveredAmount (meta, context, txn, *txMeta);
                 ret[jss::meta] = meta;
```
miguelportilla: move meta?

src/ripple/rpc/impl/DeliveredAmount.cpp
```diff
@@ -71,12 +81,12 @@ insertDeliveredAmount(
...
         // was present when a partial payment was made and its absence indicates
         // that the amount delivered is listed in the Amount field.
-        bool const ledgerIndexCheck = ledgerIndex >= 4594095;
+        bool const ledgerIndexCheck = getLedgerIndex() >= 4594095;
         // If the ledger closed long after the DeliveredAmount code was deployed
         // then its absence indicates that the amount delivered is listed in the
         // Amount field. DeliveredAmount went live January 24, 2014.
         // 446000000 is in Feb 2014, well after DeliveredAmount went live
-        bool const closeTimeCheck = closeTime > NetClock::time_point{446000000s};
+        bool const closeTimeCheck = getCloseTime() > NetClock::time_point{446000000s};
```
miguelportilla: `getCloseTime` can be expensive as it hits the node store and possibly SQLite. I recommend evaluating `closeTimeCheck`  only if `ledgerIndexCheck` is none.

seelabs: @miguelportilla Agreed. The whole point of passing in the lambda was to evaluate it only when needed, and then I evaluated here unnesisarily anyway. Good catch. Fixed in: `575212aa1 * delivered_amt [FOLD] Do not evaluate `getCloseTime` unless needed`

src/ripple/rpc/handlers/Tx.cpp
seelabs: Fixed in: `3464b7117 * delivered_amt [fold] Resolve feedback from Mickey 1`

src/ripple/app/ledger/impl/LedgerToJson.cpp
```diff
@@ -99,46 +99,42 @@ void fillJsonBinary(Object& json, bool closed, LedgerInfo const& info)
...
-                txJson[jss::metaData] = i.second->getJson(0);
+                txJson[jss::metaData] = stMeta->getJson(0);
                 if (txnType == ttPAYMENT || txnType == ttCHECK_CASH)
                 {
-                    auto const& txn = i.first;
+                    // Insert delivered amount
                     auto const& txnID = txn->getTransactionID();
-                    auto rawMeta = fill.ledger.txRead(txnID).second;
-                    if (rawMeta)
+                    if (stMeta)
```
miguelportilla: This check is unnecessary, `stMeta` must be valid here.

seelabs: Agreed. Fixed in `Mickey fixes 2` (to be pushed)

src/ripple/rpc/impl/DeliveredAmount.cpp
```diff
@@ -0,0 +1,161 @@
+//------------------------------------------------------------------------------
+/*
+    This file is part of rippled: https://github.com/ripple/rippled
+    Copyright (c) 2012-2014 Ripple Labs Inc.
```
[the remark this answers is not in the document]
seelabs: Fixed in `Resolve RP feedback 2` (to be pushed)

src/ripple/rpc/impl/DeliveredAmount.cpp
```diff
@@ -0,0 +1,161 @@
...
+#include <ripple/protocol/Feature.h>
+#include <ripple/rpc/Context.h>
+#include <ripple/rpc/impl/RPCHelpers.h>
+#include <boost/algorithm/string/case_conv.hpp>
+
+namespace ripple {
+namespace RPC {
+
+/*
+  GetLedgerIndex and GetCloseTime and lambdas that allow the close time and ledger index to be lazily calculated.
```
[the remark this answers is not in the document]
seelabs: Fixed in `Resolve RP feedback 2` (to be pushed)

src/test/rpc/Subscribe_test.cpp
```diff
@@ -651,6 +651,109 @@ class Subscribe_test : public beast::unit_test::suite
...
+        Json::Value stream;
+
+        {
+            // RPC subscribe to ledger stream
+            stream[jss::streams] = Json::arrayValue;
+            stream[jss::streams].append("ledger");
+            stream[jss::accounts] = Json::arrayValue;
+            stream[jss::accounts].append(toBase58(alice.id()));
+            stream[jss::accounts].append(toBase58(bob.id()));
+            auto jv = wsc->invoke("subscribe", stream);
```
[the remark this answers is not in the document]
seelabs: Fixed in `Resolve PR feedback 2` (to be pushed)

src/ripple/rpc/DeliveredAmount.h
```diff
@@ -0,0 +1,68 @@
+//------------------------------------------------------------------------------
+/*
+    This file is part of rippled: https://github.com/ripple/rippled
+    Copyright (c) 2012=2014 Ripple Labs Inc.
```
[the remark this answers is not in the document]
seelabs: Fixed in `Resolve PR feedback 2` (to be pushed)

src/ripple/app/ledger/impl/LedgerToJson.cpp
```diff
@@ -97,49 +99,67 @@ void fillJsonBinary(Object& json, bool closed, LedgerInfo const& info)
...
+                // Insert delivered amount
+                auto const& txnID = txn->getTransactionID();
+                if (stMeta)
+                {
+                    auto txMeta = std::make_shared<TxMeta>(
+                        txnID, fill.ledger.seq(), *stMeta);
+                    RPC::insertDeliveredAmount(
+                        txJson[jss::metaData], fill.ledger, txn, *txMeta);
+                }
+            }
```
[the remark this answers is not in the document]
seelabs: Fixed in `Resolve PR feedback 2` (to be pushed) (tho you can't write it as you suggest here. You still need to execute line 126 even if not a payment or check_cash). I just removed the inner `stMeta` check

src/ripple/rpc/DeliveredAmount.h
```diff
@@ -0,0 +1,68 @@
...
+class Transaction;
+class TxMeta;
+class STTx;
+
+namespace RPC {
+
+struct Context;
+
+/**
+   Add a deviveredAmount field to the `meta` input/output parameter.
```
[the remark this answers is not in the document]
seelabs: Fixed in `Resolve PR feedback 2` (to be pushed)

seelabs:
@ximinez I agree with your comment that a larger refactoring should be looked at. However, I do think this PR improves the situation in the meantime. I'd like to get this in first, and later on we can tackle a larger refactoring (which I fully support).

src/test/rpc/Subscribe_test.cpp
```diff
@@ -651,6 +651,110 @@ class Subscribe_test : public beast::unit_test::suite
...
+        using namespace std::chrono_literals;
+
+        Account const alice("alice");
+        Account const bob("bob");
+        Account const carol("carol");
+        auto const gw = Account("gateway");
+        auto const USD = gw["USD"];
+
+        // number of expected 'delivered_amount' available
+        int numExpectedAvailable = 0;
```
ximinez: I hate to come back to this again, but it occurred to me that it would be good to add a `numExpectedNotAvailable` and a corresponding check here and in LedgerRPC_test.cpp

src/test/rpc/Subscribe_test.cpp
```diff
@@ -651,6 +651,109 @@ class Subscribe_test : public beast::unit_test::suite
...
+            env.require(balance(bob, USD(1000)));
+
+            // failed payment
+            env(pay(gw, carol, USD(9999999)), ter(tecPATH_PARTIAL));
+            // do not incExpected - the payment failed
+            env.require(balance(carol, USD(0)));
+        }
+
+        auto wsc = makeWSClient(env.app().config());
+        Json::Value stream;
```
miguelportilla: `stream` scope can be narrowed.

src/test/rpc/Subscribe_test.cpp
```diff
@@ -651,6 +651,109 @@ class Subscribe_test : public beast::unit_test::suite
...
+                    m.isMember("delivered_amount") &&
+                    m["delivered_amount"] != "unavailable";
+                BEAST_EXPECT(expectDeliveredAmount == hasDeliveredAmount);
+                if (hasDeliveredAmount)
+                {
+                    ++numAvailable;
+                    if (m.isMember("DeliveredAmount"))
+                    {
+                        BEAST_EXPECT(
+                            m["delivered_amount"] == m["DeliveredAmount"]);
```
miguelportilla: Consider using `sfDeliveredAmount.jsonName`

src/test/rpc/LedgerRPC_test.cpp
```diff
@@ -1501,6 +1501,81 @@ class LedgerRPC_test : public beast::unit_test::suite
...
+                if (t["meta"].isMember("DeliveredAmount"))
+                {
+                    BEAST_EXPECT(
+                        t["meta"]["delivered_amount"] ==
+                        t["meta"]["DeliveredAmount"]);
+                }
+                else
+                {
+                    BEAST_EXPECT(t["Amount"] == t["meta"]["delivered_amount"]);
+                }
```
miguelportilla: Would there be any value in comparing the `tx` RPC result here? eg.
```Json::Value const meta = env.rpc("tx", t[jss::hash].asString())[jss::result][jss::meta];```

File: src/test/rpc/Subscribe_test.cpp (hunk 10 above)
Comment (seelabs) [reply src/test/rpc/Subscribe_test.cpp]: Fixed in `Refactor tests and resolve PR feedback 3:`

File: src/ripple/app/ledger/impl/LedgerToJson.cpp [hunk 14]
```diff
@@ -99,65 +99,67 @@ void fillJsonBinary(Object& json, bool closed, LedgerInfo const& info)
...
-            if (account != amount.getIssuer())
+            txJson[jss::metaData] = stMeta->getJson(0);
+            if (txnType == ttPAYMENT || txnType == ttCHECK_CASH)
             {
-                auto const ownerFunds = accountFunds(fill.ledger,
-                    account, amount, fhIGNORE_FREEZE,
-                    beast::Journal {beast::Journal::getNullSink()});
-                txJson[jss::owner_funds] = ownerFunds.getText ();
+                // Insert delivered amount
+                auto const& txnID = txn->getTransactionID();
```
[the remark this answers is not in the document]
seelabs: Fixed in `Refactor tests and resolve PR feedback 3:`

src/ripple/rpc/impl/DeliveredAmount.cpp
```diff
@@ -102,9 +112,9 @@ insertDeliveredAmount(
     auto const info = ledger.info();
     insertDeliveredAmount(
         meta,
-        ledger.rules(),
-        info.seq,
-        info.closeTime,
+        [&ledger]{return ledger.rules().enabled(fix1623);},
+        [&info]{return info.seq;},
+        [&info]{return info.closeTime;},
```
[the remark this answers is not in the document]
seelabs: Fixed in `Refactor tests and resolve PR feedback 3:` I made individual variables for each lambda; I didn't like adding the space when called inline.

src/test/rpc/LedgerRPC_test.cpp
```diff
@@ -1555,21 +1555,22 @@ class LedgerRPC_test : public beast::unit_test::suite
...
             {
                 ++numAvailable;
-                if (t["meta"].isMember("DeliveredAmount"))
+                if (t["metaData"].isMember("DeliveredAmount"))
                 {
                     BEAST_EXPECT(
-                        t["meta"]["delivered_amount"] ==
-                        t["meta"]["DeliveredAmount"]);
+                        t["metaData"]["delivered_amount"] ==
+                        t["metaData"]["DeliveredAmount"]);
```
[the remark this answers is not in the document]
seelabs: Fixed in `Refactor tests and resolve PR feedback 3:`

src/ripple/rpc/impl/DeliveredAmount.cpp
```diff
@@ -34,7 +34,7 @@ namespace ripple {
 namespace RPC {
 
 /*
-  GetLedgerIndex and GetCloseTime and lambdas that allow the close time and ledger index to be lazily calculated.
+  GetLedgerIndex and GetCloseTime are lambdas that allow the close time and ledger index to be lazily calculated.
```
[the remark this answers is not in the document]
seelabs: Fixed in `Refactor tests and resolve PR feedback 3:`

src/test/rpc/Subscribe_test.cpp
seelabs: Er, I don't know about this one. For a test I think the string "DeliveredAmount" is fine `sfDeliveredAmount.jsonName` seems cumbersome. I left it as-is for now, but will change it if you feel strongly.

src/test/rpc/Subscribe_test.cpp
seelabs: Fixed in `Refactor tests and resolve PR feedback 3:`

src/test/rpc/LedgerRPC_test.cpp
seelabs: Take a look at the new refactored tests in `Fixed in `Refactor tests and resolve PR feedback 3:`; I'm not sure if this resolves this issue or not (I do consider the tx result, but I'm not sure if that's what you meant or not). I'll leave this issue unresolved until I talk to you on Monday.

src/test/rpc/Subscribe_test.cpp
ximinez: The biggest advantage of `jsonName` is that the reference will be easier to find if we ever rename or restructure the field variable names.

src/test/rpc/DeliveredAmount_test.cpp
```diff
@@ -0,0 +1,342 @@
...
+    int numExpectedAvailable_ = 0;
+    // Number of payments with field with `delivered_amount` set to the
+    // string "unavailable"
+    int numExpectedSetUnavailable_ = 0;
+    // Number of payments with no `delivered_amount` field
+    int numExpectedNotSet_ = 0;
+
+    // increment the expected numbers based on xxx, partial payment, and
+    // ter success is true is the ter is tesSUCCESS partial is true if a
+    // partial payment;
```
ximinez: This comment needs a rewrite.

src/test/rpc/DeliveredAmount_test.cpp
```diff
@@ -0,0 +1,342 @@
...
+                else
+                {
+                    if (metaData[jss::delivered_amount] != "unavailable")
+                        return false;
+                }
+            }
+        }
+        else
+        {
+        }
```
ximinez: Empty block

src/test/rpc/Subscribe_test.cpp
seelabs: Fixed with: `Replace raw strings with sf<SomeName>.jsonName` commit

src/test/rpc/DeliveredAmount_test.cpp
seelabs: Fixed in `Remove empty block and reword comment` commit

src/test/rpc/DeliveredAmount_test.cpp
seelabs: Fixed in `Remove empty block and reword comment` comment

src/test/rpc/LedgerRPC_test.cpp
miguelportilla: I meant with every check iteration, cross-reference the relevant portion of an RPC "tx" (or "account_tx") result using the transaction hash from this "ledger" result. But since they both call insertDeliveredAmount, I don't believe there would be much benefit other than making sure the RPCs are reporting similar results. The refactor looks good as is, no change necessary.

seelabs:
Squashed and marked as passed

nbougalis:
Merged as c5d215d901d25b1ea18bf0e96529799b1b5158cc

PR #2864, opened 2019-02-22
Allow build to support XCode 10.2

mellery451:
tested with current 10.1 and 10.2 beta 3

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -366,7 +366,6 @@ add_library (opts INTERFACE)
 add_library (Ripple::opts ALIAS opts)
 target_compile_definitions (opts
   INTERFACE
-    BOOST_ASIO_HAS_STD_ARRAY
     BOOST_ASIO_DISABLE_HANDLER_TYPE_REQUIREMENTS
     $<$<BOOL:${boost_show_deprecated}>:
       BOOST_ASIO_NO_DEPRECATED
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -567,6 +566,15 @@ target_link_libraries (ripple_boost
     Boost::system
     Boost::thread)
 
+# workaround for xcode 10.2 and boost < 1.69
+# once we require Boost 1.69 or higher, this can be removed
+# see:  https://github.com/boostorg/asio/commit/43874d5
+if (CMAKE_CXX_COMPILER_ID STREQUAL "AppleClang" AND
+    CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 10.0.1.10010043 AND
+    Boost_VERSION LESS 106900)
+  target_compile_definitions (opts INTERFACE BOOST_ASIO_HAS_STD_STRING_VIEW)
```
scottschurr: After nosing around, it looks like this problem is because in boost 1.69.0 ASIO thinks it can use `std::string_view` because we have the header, even though we're still running C++14 (so `std:string_view` is not really available).  Did I get that right?

mellery451: @scottschurr I think this pull request mostly describes the issue and links to the fix:  https://github.com/boostorg/asio/pull/91  -- so 1.69 fixes the issue and we can remove this check/macro once we *require* boost 1.69 or higher. For now, I wanted to be able to support 10.2 + boost 1.67 was well as 1.69

scottschurr: :+1: thanks for the clarification.

PR #2867, opened 2019-02-26
rippled-example.cfg - Remove trailing comments from [crawl] example

intelliot:
Trailing comments are not permitted; they cause the lines containing
them to be ignored.


mDuo13:
Thank you for doing this. The revised version looks good to me.

PR #2869, opened 2019-02-27
Reduce likelihood of malformed SOTemplate

scottschurr:
During the code review of https://github.com/ripple/rippled/pull/2854, @mellery451 asked about the consequences of some added `throw`s.  It was noted that those `throw`s could result in a malformed `SOTemplate` being left behind.  This pull request sits atop https://github.com/ripple/rippled/pull/2854 and does its best to address that issue.

Formerly an `SOTemplate` was default constructed and, later, its elements added using `push_back()`.  This left open the possibility of a malformed `SOTemplate` if adding one of the elements caused a throw.

With this commit the `SOTemplate` requires an `initializer_list` of its elements at construction.  Elements may not be added after construction.  With this approach either...

- The `SOTemplate` is fully constructed with all of its elements or
- The constructor throws, which prevents an invalid `SOTemplate` from even existing.

This change requires all callers of the `SOTemplate` constructor to be adjusted at the call site.  Those changes are also in this commit.

The `SOE_Flags` `enum` is also renamed to `SOEStyle`, which harmonizes the name with other uses of the `enum` in the code base.  `SOEStyle` enumerators are renamed (slightly) to have an `soe` prefix rather than `SOE_`.  This heads toward reserving identifiers with all upper case for macros.  The new style also aligns with other prominent `enum`s in the code base like the collection of `TER` identifiers.

`SOElement` is adjusted so it can be stored directly in an STL container (by using `std::reference_wrapper`).  Previously `SOElement` needed to be stored in a `std::unique_ptr` which could be stored in an STL container.  As a result of this change `unique_ptr` usage is removed from both `SOTemplate` and `KnownFormats`.

Only the top-most commit needs to be reviewed. That commit sits atop two other unmerged pull requests:

- https://github.com/ripple/rippled/pull/2854 (its immediate predecessor) and
- https://github.com/ripple/rippled/pull/2770 (which underlies https://github.com/ripple/rippled/pull/2854 ).

I'm assuming that both of the underlying pull requests will be merged and the top-most commit of this branch can be merged afterwards (assuming it passes code review).

### DIFF
--- a/src/ripple/protocol/KnownFormats.h
+++ b/src/ripple/protocol/KnownFormats.h
@@ -22,7 +22,9 @@
 
 #include <ripple/basics/contract.h>
 #include <ripple/protocol/SOTemplate.h>
-#include <memory>
+#include <boost/container/flat_map.hpp>
+#include <algorithm>
+#include <forward_list>
 
 namespace ripple {
 
@@ -42,50 +44,49 @@ public:
     class Item
     {
     public:
-        Item (char const* name, KeyType type)
-            : m_name (name)
-            , m_type (type)
+        Item (char const* name,
+            KeyType type,
+            std::initializer_list<SOElement> uniqueFields,
+            std::initializer_list<SOElement> commonFields)
+            : soTemplate_ (uniqueFields, commonFields)
+            , name_ (name)
+            , type_ (type)
         {
-        }
-
-        Item& operator<< (SOElement const& el)
-        {
-            elements.push_back (el);
-
-            return *this;
+            // Verify that KeyType is appropriate.
+            static_assert (
+                std::is_enum<KeyType>::value ||
+                std::is_integral<KeyType>::value,
+                "KnownFormats KeyType must be integral or enum.");
         }
 
         /** Retrieve the name of the format.
         */
-        std::string const& getName () const noexcept
+        std::string const& getName () const
         {
-            return m_name;
+            return name_;
         }
 
         /** Retrieve the transaction type this format represents.
         */
-        KeyType getType () const noexcept
+        KeyType getType () const
         {
-            return m_type;
+            return type_;
         }
 
-    public:
-        // VFALCO TODO make an accessor for this
-        SOTemplate elements;
+        SOTemplate const& getSOTemplate() const
+        {
+            return soTemplate_;
+        }
 
     private:
-        std::string const m_name;
-        KeyType const m_type;
+        SOTemplate soTemplate_;
+        std::string const name_;
+        KeyType const type_;
     };
 
-private:
-    using NameMap = std::map <std::string, Item*>;
-    using TypeMap = std::map <KeyType, Item*>;
-
-public:
     /** Create the known formats object.
 
-        Derived classes will load the object will all the known formats.
+        Derived classes will load the object with all the known formats.
     */
     KnownFormats () = default;
 
@@ -104,7 +105,7 @@ public:
         @param  name The name of the type.
         @return      The type.
     */
-    KeyType findTypeByName (std::string const name) const
+    KeyType findTypeByName (std::string const& name) const
     {
         Item const* const result = findByName (name);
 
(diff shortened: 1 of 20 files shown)


src/ripple/protocol/KnownFormats.h
```diff
@@ -42,47 +44,46 @@ class KnownFormats
...
         KeyType getType () const noexcept
         {
-            return m_type;
+            return type_;
         }
 
-    public:
-        // VFALCO TODO make an accessor for this
-        SOTemplate elements;
+        SOTemplate const& getSOTemplate() const noexcept
```
seelabs: I don't think there's any benefit to `noexept` on functions like this, and I don't think there's ever likely to be one.  I'm fine if you want to keep it, I don't see the harm either. 

I do understand the benefit of annotating some constructors or assignments with `noexcept` for the benefit of generic code. But I don't see the benefit here. In general, if there's no benefit I'd rather leave things like this out.

src/ripple/protocol/SOTemplate.h
```diff
@@ -61,50 +75,44 @@ class SOElement
 class SOTemplate
 {
 public:
-    using list_type = std::vector <std::unique_ptr <SOElement const>>;
-    using iterator_range = boost::iterator_range<list_type::const_iterator>;
+    SOTemplate(SOTemplate&& other) = default;
```
seelabs: I'd either explictly delete the copy constructor/assignment operator or add a comment here explaining the reason for the defaulted move constructor (assuming this is the reason).

BTW, I don't see an obvious reason why this is a move-only type, but I'm not advocating for that change now.

src/ripple/protocol/SOTemplate.h
```diff
@@ -61,50 +75,44 @@ class SOElement
...
     */
-    SOTemplate () = default;
+    SOTemplate (std::initializer_list<SOElement> uniqueFields,
+        std::initializer_list<SOElement> commonFields = {});
 
-    SOTemplate(SOTemplate&& other)
-        : mTypes(std::move(other.mTypes))
-        , mIndex(std::move(other.mIndex))
+    /* Provide for the enumeration of fields */
+    std::vector<SOElement>::const_iterator begin() const
```
seelabs: I use `begin/end` myself, but lately we've been adding `cbegin/cend` to classes with `begin/end` as well. May as well do it here to be consistent.

src/ripple/protocol/impl/SOTemplate.cpp
```diff
@@ -21,42 +21,46 @@
 
 namespace ripple {
 
-void SOTemplate::push_back (SOElement const& r)
+SOTemplate::SOTemplate (std::initializer_list<SOElement> uniqueFields,
+    std::initializer_list<SOElement> commonFields)
+: elements_ ()
```
seelabs: Why the explicit default constructor call? I'd lose it. Also formatting seems off.

src/ripple/protocol/impl/SOTemplate.cpp
```diff
@@ -21,42 +21,46 @@
...
-    {
-        // Unmapped indices will be set to -1
-        //
-        mIndex.resize (SField::getNumFields () + 1, -1);
-    }
+    // Add all SOElements.
+    elements_.reserve (uniqueFields.size() + commonFields.size());
+    elements_.assign (uniqueFields);
+    for (SOElement const& elem : commonFields)
+        elements_.push_back (elem);
```
seelabs: I'd `insert` instead of the loop. I.e.: `elements_.insert(elements_.end(), commonFields)`

src/ripple/protocol/impl/SOTemplate.cpp
scottschurr: I've removed the explicit constructor for `elements_`.  I'm curious about what seems off with the formatting.  If you can be specific I'll fix it.  Thanks.

seelabs: The member initializations (the `:` part) should be indented four spaces. So something like:
```

SOTemplate::SOTemplate(
    std::initializer_list<SOElement> uniqueFields,
    std::initializer_list<SOElement> commonFields)
    : indices_(SField::getNumFields() + 1, -1)  // Unmapped indices == -1
```

src/ripple/protocol/SOTemplate.h
scottschurr: Thanks for poking at this.  It was not a corner I thought hard about.

I have a habit, when I _am_ thinking about it, of only giving classes capabilities that they need at the moment.  Since we have non-library code adding that capability later is not really a hardship.  And, if the class doesn't have that capability initially, the person who wants that feature can think about why they need it rather an accidentally introducing, say, a copy they weren't thinking about.

That's why, for instance, I was on the side of the Standard Committee that wanted operator
spaceship to be op-in.  I didn't want yet one more thing that I need to remember to explicitly remove from most of the classes I write.

In this case it turns out that the usage model in this pull request doesn't require `SOTemplate` to have _any_ constructor or assignment operator beyond the constructor that takes an `initializer_list`.  So the obvious thing (to me) is to simply remove the capabilities that I don't need: all those constructors and assignment operators.

However my approach does make it more inconvenient for folks who want to do new stuff in the future.  And my approach would never work for library development (where we have no idea how a potential user might want to use a class in the library).

So at this point I would be inclined to explicitly delete the constructors and assignment operators that are not needed by the code base at the moment.  I'd also add a comment saying that they are just not needed and can be enabled if the need arises.  Does that work for you?

Thanks.  And thanks for noting things that look off to you.

seelabs: By default, I'd like our types to regular types, and then remove functionality we want to forbid for good reasons - and document those reasons. A move-only type should be a signal about how that type is meant to be used (like `unique_pointer` say), not a signal that "no one needed that functionality yet".

In this particular case, the type was already move-only, and I'm not proposing changing that in this PR. However, given your reasoning above it sounds like you want to do this for new classes as well, and I do disagree with that. I want our rippled classes to be movable/copyable/assignable/equal comparable (and orderable and hashable when applicable). Even when there is no code using that functionality.

scottschurr: I can live with that.  I'm coding for a shared code base, and I recognize that my approach is not the only one.  Thanks.

src/ripple/protocol/SOTemplate.h
```diff
@@ -61,50 +75,44 @@ class SOElement
...
-    /** Create an empty template.
-        After creating the template, call @ref push_back with the
-        desired fields.
-        @see push_back
+    /** Create a template populated with all fields.
+        After creating the template fields cannot be
+        added, modified, or removed.
     */
-    SOTemplate () = default;
+    SOTemplate (std::initializer_list<SOElement> uniqueFields,
```
nbougalis: I like your idea of using `std::initializer_list` much better than my approach (passing a lambda). I also like that we have the "common fields" separate from the "unique fields" because it helps avoid mistakes, but I'm not sure I'm sold on passing _two_ initializer lists.

Rationale: it pollutes the interface of `SOTemplate` with a particular implementation detail (namely that some fields are "common" to multiple templates and some fields are not).

It would sit better with me, I think, if the callers handled the "detail" of inserting the common fields alongside the unique fields and passing one thing to this constructor.

scottschurr:
Even though this pull request has three passing reviews, I won't mark it as passed until the two underlying pull requests (https://github.com/ripple/rippled/pull/2770 and https://github.com/ripple/rippled/pull/2854) are passed and/or merged.

scottschurr:
Rebased to the most recent develop and also built on top of the most recent (and passed) versions of https://github.com/ripple/rippled/pull/2770 and https://github.com/ripple/rippled/pull/2854.  Those two pull requests will shortly be marked closed not due to any problems with them, but to prevent accidentally merging them into develop twice.

PR #2872, opened 2019-03-05
Allow regular key even when it matches the disabled master key

intelliot:
If the account has a regular key, first check whether it is being used for this transaction, before checking if it matches the master key. It is currently possible for accounts to set the regular key to be the master key (and then disable the master key). Although this serves no purpose, it should not prevent the account from being used.

This may need to be on an amendment: some transactions that were previously invalid will become valid.

- [ ] Unit test
- [ ] Test/review for performance implications


thejohnfreeman:
Merged into #2873.

PR #2873, opened 2019-03-05
[Amendment] Disallow setting an account's regular key to match its master key

thejohnfreeman:
(You'll want to ignore white space in the diff to read the new tests and test changes.)

What should we name the amendment? I defaulted to `fix1721` based on a pattern among some of the other names, naming it after our internal JIRA issue, but I would prefer a name that (1) tries to describe the change and (2) is not tied to an issue number (given that our issue platform may change).

Fixes JIRA-1721.


MarkusTeufelberger:
Maybe something like "fixUniqueMasterKey"? It is more or less an edge case and likely unintended, so I'd keep the "fix" part in front (compared to pure feature amendments like SHAmap_v2).

Alternatively maybe "fixRegularNeqMasterKey".

src/test/app/SetRegularKey_test.cpp
```diff
@@ -27,76 +27,106 @@ class SetRegularKey_test : public beast::unit_test::suite
...
+            testcase("Universal mask");
+            Env env(*this);
+            Account const alice("alice");
+            Account const bob("bob");
+            env.fund(XRP(10000), alice, bob);
+
+            auto jv = regkey(alice, bob);
+            jv[sfFlags.fieldName] = tfUniversalMask;
+            env(jv, ter(temINVALID_FLAG));
+        }
```
ximinez: Why did you roll the separate test functions into `run`? Having the functions divided up helps with readability and maintainability.

src/ripple/protocol/TER.h
```diff
@@ -96,6 +96,7 @@ enum TEMcodes : TERUnderlyingType
     temBAD_SIGNATURE,
     temBAD_SRC_ACCOUNT,
     temBAD_TRANSFER_RATE,
+    temCANT_USE_MASTER_KEY,
```
ximinez: I'm not super thrilled with this name. The existing `temBAD_SIGNER` may be perfectly appropriate, and if not I suggest something like `temBAD_REGKEY`.

src/test/app/SetRegularKey_test.cpp
thejohnfreeman: I disagree. I came into this class thinking adding a method would add a test. No, you must add a call to that method. Adding a method does not give the test a name either, though you have to give the method a name. Methods are not run independently, with separate setup and teardown. This class tries to look like other test frameworks, but it falls to live up to the expectations set by those frameworks. Instead, I'd rather it look exactly like what it's doing, so that adding to it is straightforward and intuitive.

scottschurr: I have side with Ed on this one.  I hear that you have expectations based on other test frameworks that you've used.  And, believe me, I sympathize that the framework we're using could be better.  But there are a ton of tests already in this code base that follow the model that you just unrolled.  I believe, at least, the tests within this code base should be self consistent.

If you want to argue that _all_ of the tests in the code base should be unrolled you're welcome to make that argument to the team and provide a plan for making all the tests consistent with your vision.  If the team agrees with your vision then charge ahead.  But flattening just one test only adds to confusion.

Just my opinion.

thejohnfreeman: The tests are already inconsistent. Here are some that have a fully unrolled `run()` matching the pattern I use here: [`Buffer_test`](https://github.com/ripple/rippled/blob/develop/src/test/basics/Buffer_test.cpp#L38), [`TaggedCache_test`](https://github.com/ripple/rippled/blob/develop/src/test/basics/TaggedCache_test.cpp#L41), [`Slice_test`](https://github.com/ripple/rippled/blob/develop/src/test/basics/Slice_test.cpp#L30), [`Scheduler_test`](https://github.com/ripple/rippled/blob/develop/src/test/csf/Scheduler_test.cpp#L32), [`LedgerReplay_test`](https://github.com/ripple/rippled/blob/develop/src/test/app/LedgerReplay_test.cpp#L30). One surprising case is [`Manifest_test`](https://github.com/ripple/rippled/blob/develop/src/test/app/Manifest_test.cpp#L653), which has a mix of the two patterns. Given that both patterns are in common use, I'd like to tip the scales in favor of the unroll. I'll hold a quick poll in chat to see which way the winds blow, and if I'm in the minority, I'll revert the unroll, but I don't think a plan beyond "unroll them as I touch them" should be required given the existing inconsistency.

thejohnfreeman: Well, I've got quite the uphill battle ahead of me. I've reverted the unroll for now... soon.jpg

nbougalis: We are inconsistent: some tests are in methods called from `run` and some are in `run` directly. I don't think we should settle on a single "hard" rule because such rules force us to do the wrong thing.

Consider that we have several tests that are several hundred lines long. Put all those in a single function and you'd end up with a function that's several thousand lines and that's just a recipe for disaster. Grouping a single set of "tests" inside a function just makes for easier to read code.

But then again, we have tests that are less than 10 lines long. Do we want a separate function for those? Probably not because it seems like an unnecessary level of indirection.

I think we need to exercise our judgement, and decide on a case-by-case basis. Yes, that means some inconsistency between files; Things like `Manifest_test` should probably be cleaned up to be one way or the other.

One thought might be to make `testcase` take both a name and a lambda. That way, we could develop a common style:

```
void run()
{
    testcase("Try one thing", []()
    {
    });

    testcase("Try a second thing", []()
    {
    });

    for (int i = 0; i != 8; ++i)
    {
        testcase ("Try various things: " + std::to_string(i), []()
        {
        }(i));
    }
}
```

At least until  we migrate to a different testing framework and away from `test_suite`; then we can see if one particular style is better suited.

JoelKatz:
Should we also change the behavior for any existing accounts whose master key matches their regular key? Perhaps in the code where we reject a transaction because it's signed with the master key and the master key was disabled, check if the regular key and master keys are identical and, if so, allow the transaction to execute.

I realize this changes the behavior of existing accounts, but I doubt anyone intentionally rendered an account unusable by setting the regular key to the master key and disabling the master key. We have sane ways to render an account unusable (set regular key to provably unusable value and disable master key).

MarkusTeufelberger:
How many accounts are currently affected by this anyways?

thejohnfreeman:
We've discussed reversing the checks for "does the signing key match the master key and the master key is not disabled" and "does the signing key match the regular key", which would unlock currently locked accounts (there are only around 20). That could go in this change (I think it requires an amendment anyway).

src/ripple/app/tx/impl/Transactor.cpp
```diff
@@ -340,44 +340,53 @@ Transactor::checkSign (PreclaimContext const& ctx)
...
+    // Look up the account.
+    auto const idSigner = calcAccountID(PublicKey(makeSlice(pkSigner)));
+    auto const idAccount = ctx.tx.getAccountID(sfAccount);
+    auto const sleAccount = ctx.view.read(keylet::account(idAccount));
+    auto const hasRegularKey = sleAccount->isFieldPresent(sfRegularKey);
+
+    auto const amended = ctx.view.rules().enabled(fixDisabledRegularKey);
 
-    if (pkAccount == id)
+    // Check that the signing key is authorized.
```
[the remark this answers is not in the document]
nbougalis: Coincidentally, the first `if` can be simplified further with only a bit of extra work:

Add `#include <ripple/protocol/STAccount.h>` at the top, then change the `if` to:

if ((*sleAccount)[~sfRegularKey] == idSigner)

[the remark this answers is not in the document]
thejohnfreeman: Are we sure we want to change the return codes? It involves both (1) changing the implementation in `Transactor` _and_ (2) copying any test that asserts `tefBAD_AUTH_MASTER` or `tefMASTER_DISABLED` to run once with the amendment disabled and once with it enabled.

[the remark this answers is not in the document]
thejohnfreeman: The other downside is less granular errors, leaving users to wonder: which of the N possible ways was this a bad authorization?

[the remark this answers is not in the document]
thejohnfreeman: Alright, I restored `tefMASTER_DISABLED` and changed tests expecting `tefBAD_AUTH_MASTER` to expect `tefBAD_AUTH`.

hexage09:
As far as I know, at least one person in Japan has created an account for this problem. This problem is serious given the frozen assets of that purse. About 1.4 million XRP.

src/ripple/app/tx/impl/SetRegularKey.cpp
```diff
@@ -55,6 +56,13 @@ SetRegularKey::preflight (PreflightContext const& ctx)
     if (!isTesSuccess (ret))
         return ret;
 
+    if (ctx.rules.enabled(fixDisabledRegularKey)
+        && ctx.tx.isFieldPresent(sfRegularKey)
+        && (ctx.tx.getAccountID(sfRegularKey) == ctx.tx.getAccountID(sfAccount)))
```
ximinez: Nit: Long line. We try to keep lines under 80 characters so that folks using terminals or other small screens don't have to deal with weird wrapping. I suggest a line break after the `==`.

src/test/app/SetRegularKey_test.cpp
```diff
@@ -30,33 +30,120 @@ class SetRegularKey_test : public beast::unit_test::suite
...
+        // No way to sign...
+        // log << "implicitly sign transaction with master key" << std::endl;
+        env(noop(alice), ter(tefMASTER_DISABLED));
+        env(noop(alice), sig(alice), ter(tefMASTER_DISABLED));
+
+        // ... until now.
+        // log << "pass amendment: fixDisabledRegularKey" << std::endl;
+        env.enableFeature(fixDisabledRegularKey);
+        env(noop(alice));
+        env(noop(alice), sig(alice));
```
ximinez: It might be useful to add a couple more transactions to clean up the "bad" regular key. eg.
```
env(regkey(alice, disabled), ter(tecNO_ALTERNATIVE_KEY));
env(fclear(alice, asfDisableMaster));
env(regkey(alice, disabled));
env(fset(alice, asfDisableMaster), ter(tecNO_ALTERNATIVE_KEY));
```
Those may not be the exact correct `ter` results, but that's the idea.

src/test/app/SetRegularKey_test.cpp
```diff
@@ -30,33 +30,120 @@ class SetRegularKey_test : public beast::unit_test::suite
...
+        // log << "disable master key" << std::endl;
+        env(fset(alice, asfDisableMaster), sig(alice));
+
+        // No way to sign...
+        // log << "implicitly sign transaction with master key" << std::endl;
+        env(noop(alice), ter(tefMASTER_DISABLED));
+        env(noop(alice), sig(alice), ter(tefMASTER_DISABLED));
+
+        // ... until now.
+        // log << "pass amendment: fixDisabledRegularKey" << std::endl;
```
ximinez: Go ahead and remove these commented out `log` lines. Chances are they won't be needed again.

src/test/jtx/Env_test.cpp
```diff
@@ -250,7 +250,7 @@ class Env_test : public beast::unit_test::suite
     {
         using namespace jtx;
 
-        Env env(*this);
+        Env env{*this, supported_amendments() | fixDisabledRegularKey};
```
ximinez: Because `fixDisabledRegularKey` is in `supported_amendments()`, is it necessary to explicitly add it in here?

src/ripple/app/tx/impl/SetRegularKey.cpp
thejohnfreeman: Agreed. I'm one of those people. I can't explain why Vim didn't already wrap it for me.

src/test/jtx/Env_test.cpp
thejohnfreeman: I want to be robust against changes in `supported_amendments`. To be honest, I'm surprised that the tests aren't already like this, because it meant I had to add `- fixDisabledRegularKey` to some existing tests. I prefer that new code not have to change existing tests.

ximinez: Amendments aren't removed from `supported_amendments` very often, so I hadn't really considered the need. I agree about new code not changing existing tests in general, but when you're changing functionality, sometimes you gotta do what you gotta do. :smile:

src/ripple/app/tx/impl/Transactor.cpp
```diff
@@ -340,44 +341,57 @@ Transactor::checkSign (PreclaimContext const& ctx)
...
         return tefBAD_AUTH; // FIXME: should be better error!
     }
 
-    auto const pkAccount = calcAccountID (
-        PublicKey (makeSlice (spk)));
+    // Look up the account.
+    auto const idSigner = calcAccountID(PublicKey(makeSlice(pkSigner)));
+    auto const idAccount = ctx.tx.getAccountID(sfAccount);
+    auto const sleAccount = ctx.view.read(keylet::account(idAccount));
+    auto const isMasterDisabled = sleAccount->isFlag(lsfDisableMaster);
```
jwbusch: nit: I'd rather see `bool` over `auto` if we know that it's a `bool` already.

src/ripple/app/tx/impl/Transactor.cpp
```diff
@@ -340,44 +341,57 @@ Transactor::checkSign (PreclaimContext const& ctx)
...
+
+    if (ctx.view.rules().enabled(fixDisabledRegularKey)) {
+
```
jwbusch: nit: spacing/indentation around here looks a little weird. (I think `{` should be on a newline).

ximinez: Oh, yeah, I missed that. Our style guide requires `{` to always be on a newline except for namespaces.

hexage09:
I will write something that everyone already knows, by the way.

This bug will kill the master key first, or even if you set a regular key from Bob or Charlie to Alice's address, the first thing I've identified is the reverse procedure of Alice setting to Alice's public address also success, this is of course not an error code in the process of disable_master flag after this.

So I think the nature of the problem is this: return a ERR_duplicate_key, or at the time of error recovery code such as RPE_duplicated_key is appropriate.

Recently I am good report here to the owner of the wallet.

hexage09:
The problem is as trivial as a mosquito bite, and it won't even be an exploit.

Still, if you care about this boring and tedious tweak, we should know this word.

The really beautiful code is simple and powerful. the same is true for the name.

src/ripple/app/tx/impl/Transactor.cpp
```diff
@@ -340,44 +341,58 @@ Transactor::checkSign (PreclaimContext const& ctx)
...
+    auto const idSigner = calcAccountID(PublicKey(makeSlice(pkSigner)));
+    auto const idAccount = ctx.tx.getAccountID(sfAccount);
+    auto const sleAccount = ctx.view.read(keylet::account(idAccount));
+    bool const isMasterDisabled = sleAccount->isFlag(lsfDisableMaster);
+
+    if (ctx.view.rules().enabled(fixDisabledRegularKey))
+    {
+
+        if ((*sleAccount)[~sfRegularKey] == idSigner
+                || (!isMasterDisabled && idAccount == idSigner))
```
nbougalis: Keeping in mind that this is the new flow, I believe that we are making the logic a little complicated to follow. I would prefer to see this block take the following shape:

```
if (ctx.view.rules().enabled(fixDisabledRegularKey))
{
    // Signature made using the regular key: accepted
    if ((*sleAccount)[~sfRegularKey] == idSigner)
        return tesSUCCESS;

    // Signature made using the master key: accepted if not disabled
    if (!sleAccount->isFlag(lsfDisableMaster) && idAccount == idSigner)
        return tesSUCCESS;

    // Signature made with a key that's not authorized to sign
    return tefBAD_AUTH;
}
```

I understand that this "removes" a theoretically useful error message but I highly doubt that the distinct error message is useful _in practice_. Whether the failure was because the master key is disabled or because the wrong key was used doesn't seem particularly actionable. I _can_ imagine a scenario (a client that knows both the master and the regular key tries to sign with the master key first and, if disabled, transparently resigns with the regular key) but honestly, I find that flow broken.

The cost of having the extra result is, to be sure, minor but it also complicates (if only slightly) one of the most critical parts of the code: checking whether a given public key is authorized to sign for a particular account. In that context, every single instruction matters. Every single additional `if` is an opportunity to get it wrong or mess something up.

scottschurr: I'll say that I do find @nbougalis's reformulation significantly easier to read and reason about, especially with the comments above each condition.  I wonder is we can keep both this readability and the `tefMASTER_DISABLED`?  Consider something like...
```
if (ctx.view.rules().enabled(fixDisabledRegularKey))
{
    // Signature made using the regular key: accepted
    if ((*sleAccount)[~sfRegularKey] == idSigner)
        return tesSUCCESS;

    // Signature made using the master key: accepted if not disabled
    if (!isMasterDisabled && idAccount == idSigner)
        return tesSUCCESS;

    // Signature made using disabled master key
    if (isMasterDisabled && idAccount == idSigner)
        return tefMASTER_DISABLED;

    // Signature made with a key that's not authorized to sign
    return tefBAD_AUTH;
}
```

ximinez: I like the readability that comes with splitting up the success checks (especially because they can be commented separately), but I don't necessarily think we need to remove the other error message. For example,
```
if (ctx.view.rules().enabled(fixDisabledRegularKey))
{
    // Signature made using the regular key: accepted
    if ((*sleAccount)[~sfRegularKey] == idSigner)
        return tesSUCCESS;

    // Signature made using the master key: accepted if not disabled
    if (!isMasterDisabled && idAccount == idSigner)
        return tesSUCCESS;

    // Signature made using the master key: reject if disabled
    if (isMasterDisabled && idAccount == idSigner)
        return tefMASTER_DISABLED;

    // Signature made with a key that's not authorized to sign
    return tefBAD_AUTH;
}
```

nbougalis: Thanks, @scottschurr, @ximinez, @thejohnfreeman.

nbougalis:
:+1: - will need squashing together, once we're ready to merge this.

hexage09:
It is surprising that this problem has been neglected for two months.
I feel a lack of technology in logic circuits.

nbougalis:
“Neglected”? The PR has been merged and will go out with the **1.3.0** release.

MarkusTeufelberger:
It'll still need to be activated, but that's up to the validators, not the people that write the code here. In code, the issue is solved already.

PR #2874, opened 2019-03-11
Make LedgerTrie remove work for truncated history

mellery451:


src/ripple/consensus/LedgerTrie.h
```diff
@@ -513,58 +529,51 @@ class LedgerTrie
     bool
     remove(Ledger const& ledger, std::uint32_t count = 1)
```
nbougalis: Is calling this function with `count == 0` allowed?

nbougalis: Is calling this function with `count == 0` allowed?

src/ripple/consensus/LedgerTrie.h
```diff
@@ -575,14 +584,8 @@ class LedgerTrie
...
-        Node const* loc;
-        Seq diffSeq;
-        std::tie(loc, diffSeq) = find(ledger);
-
-        // Exact match
-        if (loc && diffSeq == loc->span.end() && diffSeq > ledger.seq())
-            return loc->tipSupport;
-        return 0;
+        Node const* loc = findByLedgerID(ledger);
+        return loc ? loc->tipSupport : 0;
```
nbougalis: This is more of a stylistic nit than anything, but what about:

    if (auto const* loc = findLedgerID(ledger))
        return loc->tipSupport;
    return 0;

src/ripple/consensus/LedgerTrie.h
```diff
@@ -395,6 +395,22 @@ class LedgerTrie
         return std::make_pair(curr, pos);
     }
 
+    Node*
```
bachase: Nit: Consider adding documentation for this function. Its internal, but might be helpful nonetheless.

src/ripple/consensus/LedgerTrie.h
bachase: It should work but not change anything in the trie.

PR #2876, opened 2019-03-15
Update RELEASENOTES.md

intelliot:
Fix broken image and update links


MarkusTeufelberger:
Isn't the protocol called "RTXP"?

PR #2880, opened 2019-03-21
Cxx 17

seelabs:
This PR is build on 1.3.0-b2 and proposes that we enable C++-17 support. I'm assigning @HowardHinnant and @nbougalis as reviewers: Howard for the code changes and Nik to make the broader decision if we want to approve turning on c++ 17.

Just the commits after `Set version to 1.3.0-b2` need to be reviewed.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -422,7 +422,8 @@ add_library (Ripple::common ALIAS common)
 link_libraries (Ripple::common)
 set_target_properties (common
   PROPERTIES INTERFACE_POSITION_INDEPENDENT_CODE ON)
-target_compile_features (common INTERFACE cxx_std_14)
+set(CMAKE_CXX_EXTENSIONS OFF)
+target_compile_features (common INTERFACE cxx_std_17)
 target_compile_definitions (common
   INTERFACE
     $<$<CONFIG:Debug>:DEBUG _DEBUG>
--- a/src/ripple/basics/random.h
+++ b/src/ripple/basics/random.h
@@ -51,7 +51,7 @@ namespace detail {
 // Determines if a type can be called like an Engine
 template <class Engine, class Result = typename Engine::result_type>
 using is_engine =
-    std::is_invocable<Engine, Result()>;
+    std::is_invocable_r<Result, Engine>;
 }
 
 /** Return the default random engine.
--- a/src/ripple/beast/cxx17/type_traits.h
+++ b/src/ripple/beast/cxx17/type_traits.h
@@ -24,82 +24,18 @@
 #include <type_traits>
 #include <utility>
 
+#if defined(_LIBCPP_VERSION) && _LIBCPP_VERSION < 6000
+
 namespace std {
 
-#ifndef _MSC_VER
+template <class _Ret, class _Fn, class... _Args>
+struct _LIBCPP_TEMPLATE_VIS is_invocable_r
+    : integral_constant<bool, __invokable_r<_Ret, _Fn, _Args...>::value>
+{
+};
 
-    #if ! __cpp_lib_void_t
-
-        template<class...>
-        using void_t = void;
-
-    #endif // ! __cpp_lib_void_t
-
-    #if ! __cpp_lib_bool_constant
-
-        template<bool B>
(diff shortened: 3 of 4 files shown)


scottschurr:
FWIW, regarding the clang build failures, I tried a local clang build on my Mac OS machine and got this diagnostic:
```
In file included from /Users/scottschurr/Development/rippled/src/ripple/core/impl/SNTPClock.cpp:22:
/Users/scottschurr/Development/rippled/src/ripple/basics/random.h:54:10: error: 
      no template named 'is_invocable_r' in namespace 'std'; did you mean
      '__invokable_r'?
    std::is_invocable_r<Result, Engine>;
    ~~~~~^~~~~~~~~~~~~~
         __invokable_r
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/include/c++/v1/type_traits:4336:8: note: 
      '__invokable_r' declared here
struct __invokable_r
       ^
```

seelabs:
@scottschurr The problem is the version of xcode you're using ships with the equivalent to clang 5; Support for `is_invocable_r` didn't happen until clang 6. Unforuntately, this requires our mac users to be on the xcode 10, which only came out last September.

The options I see are:

1) Require xcode 10;
2) Code up a workaround for `is_invocable_r` for xcode 9
3) Don't upgrade to c++-17 yet.

Right now, I'm leaning to (3). But I'll leave this PR open for a little while longer.

HowardHinnant:
I can confirm that things compile on Xcode 10.

My opinion is that neither requiring Xcode 10, nor patching `is_invocable_r` for Xcode 9 are big enough deals to impact the decision to move to C++17.  The decision should be made on other factors such as the cost to those clients who are compiling with gcc (our only officially supported platform), and what benefits C++17 will bring us (CTAD seems like a very useful language feature to me).

HowardHinnant:
This passes all tests on macOS, using Xcode 10.  The code changes look good to me.

scottschurr:
FWIW, if all that is required is a relatively simple work around for Xcode 9 I'd be in favor of pushing ahead.  There are some specific capabilities if C++17 that I'd like to use in the code base.  I can think of `inline` variables and `if constexpr` specifically.

HowardHinnant:
`if constexpr` isn't supported by gcc-5.2.  We need at least gcc-7 for that.  I haven't checked `inline` variables.  I agree these would both be good to have in our toolbox.

HowardHinnant:
Here is what src/ripple/beast/cxx17/type_traits.h could look like to patch Xcode 9:

```
#if defined(_LIBCPP_VERSION) && _LIBCPP_VERSION < 6000

namespace std
{

template <class _Ret, class _Fn, class ..._Args>
struct _LIBCPP_TEMPLATE_VIS is_invocable_r
    : integral_constant<bool, __invokable_r<_Ret, _Fn, _Args...>::value> {};

}  // namespace std

#endif
```

seelabs:
Forced push with Howard's patch for xcode 9

scottschurr:
With @HowardHinnant's patch everything builds, unit tests pass, and server syncs and shuts down cleanly on my Mac using Xcode 9!  Nice work!

seelabs:
The problem with the clang jenkins builds is clang is picking up the standard library from gcc 6 (even though gcc 7 is installed) and gcc 6 does not have `string_view`.

@mellery451  and I brainstormed about this and see two options:

1) We could update the jenkins clang to clang 7. This _does_ pick up the correct stdlib. The downside to that is we no longer have a CI test for clang 5 at all (xcode 9 still uses clang 5, so I don't want to lose that test).

2) Upgrade the jenkins to clang 7, and add a mac tests to travis. Mike says travis has good support for mac, so as much as I dislike travis, this may be a good solution, at least until we stop supporting xcode 9.

I think (2) is the best option.

JoeLoser:
Is there any consensus on how to move forward with this? I'd love to see this land.

sublimator:
Winter is coming

>

seelabs:
@JoeLoser The plan is to land this early into the 1.4 release. I don't want to put it into the 1.3 release, which is too far along to bump the C++ version.

PR #2881, opened 2019-03-23
Arguments passed to jtx Env::operator() must be invocable:

seelabs:
Before this patch, jtx allowed non-invocable functions to be passed to
operator(). However, these arguments are ignored. This caused erronious code
code such as:

```
env (offer (account_to_test, BTC (250), XRP (1000)),
         offers (account_to_test, 1));
```

While it looks like the number of offers are checked, they are not. The `offers`
funclet is never run. While we could modify jtx to make the above code correct,
a cleaner solution is to run post conditions in a `require` statement after a
transasction runs.

### DIFF
--- a/src/test/app/Offer_test.cpp
+++ b/src/test/app/Offer_test.cpp
@@ -1126,15 +1126,15 @@ public:
         // PART 1:
         // we will make two offers that can be used to bridge BTC to USD
         // through XRP
-        env (offer (account_to_test, BTC (250), XRP (1000)),
-                 offers (account_to_test, 1));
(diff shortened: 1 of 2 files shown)


seelabs:
Note: as soon as we switch to c++-17 I'll replace all the "for each argument" hacks we have in place now with cleaner fold expressions.

seelabs:
The CI visual studio tests are failing. I'm closing this PR until I can debug them.

seelabs:
Reopening this. The windows issues appears to be a compiler issue that is fixed in newer versions of visual studio. We are upgrading the visual studio we are using in CI.

PR #2888, opened 2019-04-01
Rename JsonFields.h to jss.h:

scottschurr:
At this point all of the jss::* names are defined in the same file.  That file has been named JsonFields.h.  That file name has little to do with either JsonStaticStrings (which is what jss is short for) or with the text "jss".  The file is renamed to jss.h so the file name better reflects what the file contains.

All includes of that file are fixed.  A few include order issues are tidied up along the way.

Include order rules:

1. If the including file is a cpp file and it has a corresponding .h file, the corresponding .h file should be included first.  This assures that the corresponding .h file stands alone.

2. In general, more specialized includes should come before less specialized includes.  For example, local includes come before boost includes which come before standard library includes.

3. Within groupings of includes, files should be listed alphabetically.  This makes it easier to find duplicate include files.

I didn't look for (or fix) all include order issues.  But if I spotted something in the area I was touching I tried to fix it up.

This would be considered a trivial pull request except for the large number of files touched.  No (intentional) risky changes are included.

Travis is failing consistently with timeouts.  I don't believe those failures are a result of these changes.


MarkusTeufelberger:
The travis issue is because you're trying to build both boost and rippled on a weak build machine in under 50 minutes. Once a build succeeds, it would cache the boost build and then rippled (barely) should manage to build in 40-50 minutes on Travis-CI too.

PR #2893, opened 2019-04-03
Improve windows build README

mellery451:


Builds/VisualStudio2017/README.md
```diff
@@ -242,19 +219,35 @@ cmake ..\.. -G"Visual Studio 15 2017 Win64" -DBOOST_ROOT="C:\lib\boost_1_68_0" -
...
+ 
+Edit your `CmakeSettings.json` (described above) by adding `-Dunity=OFF` to the `cmakeCommandArgs` entry 
+for each build configuration.
+
+## Standalone CMake Builds
+
+When running cmake to generate the Visual Studio project files, add `-Dunity=OFF` to the 
+command line options passed to cmake.
+
+**Note:** you will need to re-run the cmake configuration step anytime you want to switch between unity/no-unity builds.
```
jwbusch: Do you need to clear the CMake cache too, or is it okay?

ximinez: FWIW, I usually build my unity and non-unity CMake projects in separate directories, so it's not an issue. But I did the experiment generating with `-unity=ON`, building, `-unity=OFF`, and building again. Both builds worked.

PR #2894, opened 2019-04-03
[TRIVIAL] Reserve memory before inserting into a flat set

seelabs:


src/ripple/app/paths/impl/BookStep.cpp
```diff
@@ -729,8 +729,14 @@ BookStep<TIn, TOut, TDerived>::revImp (
         auto const r = forEachOffer (sb, afView, prevStepRedeems, eachOffer);
         boost::container::flat_set<uint256> toRm = std::move(std::get<0>(r));
         std::uint32_t const offersConsumed = std::get<1>(r);
-        ofrsToRm.insert (boost::container::ordered_unique_range_t{},
-            toRm.begin (), toRm.end ());
+        if (!toRm.empty())
```
jwbusch: Seems like this pattern is repeated a few times, could it be pulled out into a function and re-used? 

Something like... 
```
template<class FlatSet>
void merge_flat_set(FlatSet& lhs, FlatSet const& rhs) {
    if (!rhs.empty()) {
        lhs.reserve(lhs.size() + rhs.size());
        lhs.insert(boost::container::ordered_unique_range_t{}, rhs.begin(), rhs.end());
    }
}
```

seelabs: Fixed in 99e6a3dcee6eb18b1ef6283380f1917585b1031b

PR #2897, opened 2019-04-04
Add GitLab CI configuration

thejohnfreeman:
The pieces of this change include:

- The job matrix in `.gitlab-ci.yml`. This should not trigger jobs (_pipelines_ in GitLab parlance) on our internal GitLab because of the `only` clause that restricts them to projects hosted at `gitlab.com`, which include [`ripple/rippled`](https://gitlab.com/ripple/rippled) and any forks there.
- Separate build and test scripts for *nix systems. These borrow some parts from the `build_and_test.sh` script that sits near them in the [`bin/ci/ubuntu` directory](https://github.com/ripple/rippled/blob/fa57859477441b60914e6239382c6fba286a0c26/bin/ci/ubuntu/build-and-test.sh#L1). That script did not seem easy to use in a new context (nor did it have enough comments for me to easily understand it), so I started from scratch and borrowed what I could, documenting everything on the way, both in the scripts and in a `README` next to them.

The job matrix in `.gitlab-ci.yml` implements much of the [18-job matrix from the Jenkinsfile](https://github.com/ripple/rippled/blob/fa57859477441b60914e6239382c6fba286a0c26/Jenkinsfile#L113):

- The 4 MSVC jobs are blocked because [Docker Machine cannot provision Windows machines](https://github.com/thejohnfreeman/ami-gitlab-runner/tree/master/windows). Docker Machine is one of two options for autoscaling GitLab Runners, the service that handles GitLab CI jobs. Kubernetes is the other option; I have not investigated it (and don't plan to yet). It could be possible to keep some Windows machines running constantly, but the size I chose for builds (`c5.2xlarge`, 8 vCPU, 16 GB) costs over $500 / mo for each. Autoscaling lets us pay for the minutes we use on spot instances that cost about 70% less per hour.
- The docs and coverage jobs (1 each) are out because I haven't worked on (a) adding those dependencies to the build environment and (2) reverse engineering the commands from the Jenkinsfile and `build_and_test.sh`.
- That leaves 12 jobs that are implemented, covering various combinations of
  - GCC vs Clang
  - Make vs Ninja (I changed the default to Ninja because it is faster, even for cold builds)
  - Debug vs Release configuration
  - unity vs non-unity
  - static vs dynamic linking

Not seen in this pull request is work I did for operating the build fleet:

- [`gitlab-runner-aws`](https://github.com/thejohnfreeman/gitlab-runner-aws): Documentation and tools for setting up the GitLab Runner bastion.
- [`ami-gitlab-runner`](https://github.com/thejohnfreeman/ami-gitlab-runner): Ready-to-go AMIs for GitLab Runners using Docker.
- [`rippled-docker`](https://github.com/thejohnfreeman/rippled-docker): The build image I'm using and publish to Docker Hub. Happy to use a different one, but I would like to see us manage the Docker images separately from the `rippled` project (1) to keep projects highly focused and as small as possible (i.e. separation of concerns) and (2) to potentially build different configurations of Docker images, each supporting different versions of `rippled`, as needed, from the same checkout.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -21,9 +21,10 @@
     - bin/ci/build.sh
     - bin/ci/test.sh
   cache:
-    # Generate a new UUID whenever you want to bust the cache, e.g. when the
+    # Use a different key for each unique combination of (generator, compiler,
+    # build type). Caches are stored as `.zip` files; they are not merged.
+    # Generate a new key whenever you want to bust the cache, e.g. when the
     # dependency versions have been bumped.
-    key: 910f11a6-8869-4ee4-9681-905f21b14b0c
     # By default, jobs pull the cache. Only a few specially chosen jobs update
     # the cache (with policy `pull-push`); one for each unique combination of
     # (generator, compiler, build type).
@@ -38,6 +39,7 @@
     COMPILER: gcc
     BUILD_TYPE: Debug
   cache:
+    key: ae412fb0-5fc8-4fd5-8e48-e3a60e91ed46
     policy: pull-push
 
 'build+test Ninja GCC Debug':
@@ -47,6 +49,7 @@
     COMPILER: gcc
     BUILD_TYPE: Debug
   cache:
+    key: 3a314872-4faa-4712-b8ab-a8d54ac83342
     policy: pull-push
 
 'build+test Ninja GCC Debug -Dstatic=OFF':
@@ -56,6 +59,8 @@
     COMPILER: gcc
     BUILD_TYPE: Debug
     CMAKE_ARGS: '-Dstatic=OFF'
+  cache:
+    key: 3a314872-4faa-4712-b8ab-a8d54ac83342
 
 'build+test Ninja GCC Debug -Dstatic=OFF -DBUILD_SHARED_LIBS=ON':
   extends: .job_linux_build_test
@@ -64,6 +69,8 @@
     COMPILER: gcc
     BUILD_TYPE: Debug
     CMAKE_ARGS: '-Dstatic=OFF -DBUILD_SHARED_LIBS=ON'
+  cache:
+    key: 3a314872-4faa-4712-b8ab-a8d54ac83342
 
 'build+test Ninja GCC Debug -Dunity=OFF':
   extends: .job_linux_build_test
@@ -72,6 +79,8 @@
     COMPILER: gcc
     BUILD_TYPE: Debug
     CMAKE_ARGS: '-Dunity=OFF'
+  cache:
+    key: 3a314872-4faa-4712-b8ab-a8d54ac83342
 
 'build+test Ninja GCC Release -Dassert=ON':
   extends: .job_linux_build_test
@@ -81,6 +90,7 @@
     BUILD_TYPE: Release
     CMAKE_ARGS: '-Dassert=ON'
   cache:
+    key: ac5e7a8a-bfcb-4480-a3cc-07d0336f4408
     policy: pull-push
 
 'build+test(manual) Ninja GCC Release -Dassert=ON':
@@ -91,6 +101,8 @@
     BUILD_TYPE: Release
     CMAKE_ARGS: '-Dassert=ON'
     MANUAL_TEST: 'true'
+  cache:
+    key: ac5e7a8a-bfcb-4480-a3cc-07d0336f4408
 
 'build+test Make clang Debug':
   extends: .job_linux_build_test
@@ -99,6 +111,7 @@
     COMPILER: clang
     BUILD_TYPE: Debug
   cache:
+    key: 09cec2ce-83f5-4ce9-8a3b-23737edefd4f
     policy: pull-push
 
 'build+test Ninja clang Debug':
@@ -108,6 +121,7 @@
     COMPILER: clang
     BUILD_TYPE: Debug
   cache:
+    key: c6e29541-e539-4d57-86c5-57d923521f35
     policy: pull-push
 
 'build+test Ninja clang Debug -Dunity=OFF':
@@ -117,6 +131,8 @@
     COMPILER: clang
     BUILD_TYPE: Debug
     CMAKE_ARGS: '-Dunity=OFF'
+  cache:
+    key: c6e29541-e539-4d57-86c5-57d923521f35
 
 'build+test Ninja clang Debug -Dunity=OFF -Dsan=address':
   extends: .job_linux_build_test
@@ -126,6 +142,8 @@
     BUILD_TYPE: Debug
     CMAKE_ARGS: '-Dunity=OFF -Dsan=address'
     CONCURRENT_TESTS: 1
+  cache:
+    key: c6e29541-e539-4d57-86c5-57d923521f35
 
 'build+test Ninja clang Debug -Dunity=OFF -Dsan=undefined':
   extends: .job_linux_build_test
@@ -134,6 +152,8 @@
     COMPILER: clang
     BUILD_TYPE: Debug
     CMAKE_ARGS: '-Dunity=OFF -Dsan=undefined'
+  cache:
+    key: c6e29541-e539-4d57-86c5-57d923521f35
 
 'build+test Ninja clang Release -Dassert=ON':
   extends: .job_linux_build_test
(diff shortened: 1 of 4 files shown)


ximinez:
For reference, this is what the build "pipeline" output looks like on gitlab.com https://gitlab.com/ripple/rippled/pipelines/55229386/builds

.gitlab-ci.yml
```diff
@@ -0,0 +1,169 @@
...
+
+.job_linux_build_test:
+  only:
+    variables:
+      - $CI_PROJECT_URL =~ /^https?:\/\/gitlab.com\//
+  stage: build
+  tags:
+    - linux
+    - c5.2xlarge
+  image: thejohnfreeman/rippled-build-ubuntu:develop
```
mellery451: how is this going to work when I want to test out builds with some other version of boost? or openssl ? I might have master that needs to build against a certain set and develop that needs to build against entirely different deps.

.gitlab-ci.yml
```diff
@@ -0,0 +1,169 @@
+# I don't know what the minimum size is, but we cannot build on t3.micro.
+
+# TODO: Factor common builds between different tests.
+
+# The parameters for our job matrix:
+#
+# 1. Generator (Make, Ninja, MSBuild)
+# 2. Compiler (GCC, Clang, MSVC)
```
mellery451: what versions of the compilers? Can we have builds against a few different versions?

bin/ci/build.sh
```diff
@@ -0,0 +1,26 @@
...
+set -o errexit
+
+# The build system. Either 'Unix Makefiles' or 'Ninja'.
+GENERATOR=${GENERATOR:-Unix Makefiles}
+# The compiler. Either 'gcc' or 'clang'.
+COMPILER=${COMPILER:-gcc}
+# The build type. Either 'Debug' or 'Release'.
+BUILD_TYPE=${BUILD_TYPE:-Debug}
+# Additional arguments to CMake.
+CMAKE_ARGS=${CMAKE_ARGS:-}
```
mellery451: for CI builds I think we want `-Dwerr=ON` always...I guess it could be added here or somewhere in the yaml config

bin/ci/test.sh
```diff
@@ -0,0 +1,42 @@
...
+CONCURRENT_TESTS=${CONCURRENT_TESTS:-$(nproc)}
+# The path to rippled.
+RIPPLED=${RIPPLED:-build/rippled}
+# Additional arguments to rippled.
+RIPPLED_ARGS=${RIPPLED_ARGS:-}
+
+function join_by { local IFS="$1"; shift; echo "$*"; }
+
+declare -a manual_tests=(
+  'beast.chrono.abstract_clock'
```
mellery451: we now have this list in two places - that's unfortunate.

thejohnfreeman:
@ximinez:

1. My general practice is to use multiple commits in the PR until it is ready to merge. At that point, I expect the maintainer (whether it is me or someone else) to choose and follow the merge strategy, whether it is squash and rebase (you and I share the same preference for this) or otherwise.
2. There is a paid GitLab integration for GitHub: https://docs.gitlab.com/ee/ci/ci_cd_for_external_repos/github_integration.html
3. I would rather not drop Jenkins until we have fully transitioned every job to GitLab.

.gitlab-ci.yml
thejohnfreeman: I'm assuming that an individual Git tree is expecting to build against exactly one version of Boost or OpenSSL. The `.gitlab-ci.yml` in that tree should choose an image that has those versions. That means, in your example, that the `.gitlab-ci.yml` on `master` has one image and the `.gitlab-ci.yml` on `develop` has a different image. The developer is expected to update the `.gitlab-ci.yml` exactly when they change any other code to expect a new version. Does that make sense?

The way we tag the images on Docker Hub is up-in-the-air. We want immutable tags so that old builds can be repeated deterministically. We could just tag an image with a unique prefix of its ID. How does that sound?

.gitlab-ci.yml
thejohnfreeman: The versions are determined by the image. It is definitely possible to set up jobs that use different images to test against different versions of compilers.

The only editable description of an image on Docker Hub is its tag, but it is not possible to encode, in a human-readable way, all the metadata about an image, including versions of dependencies installed, in the tag. We will have to maintain a separate catalog of images until Docker Hub expands its capabilities. If we build images out of [rippled-docker](https://docs.gitlab.com/ee/ci/ci_cd_for_external_repos/github_integration.html), we can put that catalog in the README.

bin/ci/build.sh
thejohnfreeman: I'll set that as the default with `${CMAKE_ARGS-'-Dwerr=ON'}` substitution. That way callers can remove it by setting an empty `$CMAKE_ARGS`.

bin/ci/test.sh
thejohnfreeman: I agree. I'd like to have a well-documented pair of scripts, `build.sh` and `test.sh`, and replace `build-and-test.sh` with calls to them, but I need some help understanding `build-and-test.sh`. I don't have experience with everything it is using, and the parts I'm unfamiliar with are undocumented or underdocumented.

bin/ci/build.sh
```diff
@@ -0,0 +1,26 @@
...
+
+set -o xtrace
+set -o errexit
+
+# The build system. Either 'Unix Makefiles' or 'Ninja'.
+GENERATOR=${GENERATOR:-Unix Makefiles}
+# The compiler. Either 'gcc' or 'clang'.
+COMPILER=${COMPILER:-gcc}
+# The build type. Either 'Debug' or 'Release'.
+BUILD_TYPE=${BUILD_TYPE:-Debug}
```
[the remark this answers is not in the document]
thejohnfreeman: I prefer to leave them open as string-typed values and let the underlying tools (CMake in this case) complain about unknown values.

thejohnfreeman:
Windows builds have roadblocks with GitLab CI.

- With the Docker Machine executor (I used this one, but we're not locked into it), we need Docker Machine to provision Windows containers. They have [shown no interest](https://github.com/docker/machine/issues/2907) in building it. They have an [open PR](https://github.com/docker/machine/pull/3329) adding the functionality (for Azure only) that has been neglected for nearly 3 years.
- GitLab is progressing slowly on [launching Windows containers with the Kubernetes executor](https://gitlab.com/gitlab-org/gitlab-runner/issues/4014).
- GitLab is very close to finishing [support for Windows with the Docker executor](https://gitlab.com/groups/gitlab-org/-/epics/535), but that will require us to keep a set of Windows machines up, even when there are no Windows jobs.

thejohnfreeman:
The GitHub integration on `gitlab.com` is [free until September 22nd, 2019](https://docs.gitlab.com/ee/ci/ci_cd_for_external_repos/). I've got it set up on [our project over there](https://gitlab.com/ripple/rippled/), and you can see in [this PR](https://github.com/ripple/rippled/pull/2923) that it triggered a pipeline and updated the commit status.  We're almost there!

Now we just have to approve this PR and merge it into `develop` so that everyone can base their future work off of it and get the `.gitlab-ci.yml` in their feature branch.

The one caveat is that if you want to take advantage of this integration, then __you must push your work to a feature branch in *this repo*__, not in your personal fork. GitLab will not run pipelines on code from forks for security reasons. I don't think it's too much of a lift, but we can discuss as a team.

ximinez:
> 1. My general practice is to use multiple commits in the PR until it is ready to merge. At that point, I expect the maintainer (whether it is me or someone else) to choose and follow the merge strategy, whether it is squash and rebase (you and I share the same preference for this) or otherwise.

Ok, that's what we usually do. We do have sort of a convention to start the commit message with `[FOLD]` if your intention is for that commit to be merged into the one before. 

> 2. There is a paid GitLab integration for GitHub: https://docs.gitlab.com/ee/ci/ci_cd_for_external_repos/github_integration.html

:+1: 

> 3. I would rather not drop Jenkins until we have fully transitioned every job to GitLab.

:+1:

thejohnfreeman:
Cool, I will start using `[FOLD]` then.

PR #2899, opened 2019-04-05
Update to NuDB 2.0.1

miguelportilla:
This PR implements new functionality in NuDB introduced here: [NuDB 2.0](https://github.com/CPPAlliance/NuDB/pull/68)

The new context object in NuDB 2.0, allows one or more database instances to share a pool of threads. This feature is especially helpful to Rippled in reducing processing overhead on servers that have many shards. This version of NuDB requires Boost 1.69 which is unfortunately broken on MacOS. I expect the PR approval to be delayed until that matter is resolved but can be tested on other platforms in the meantime.

It is important to note that with these changes, NuDB will be the only permanent storage backend supported for shards. RocksDB will no longer be supported. The `type` field in the `shard_db` is no longer required.

Because of MacOS, this PR depends on [PR2905](https://github.com/ripple/rippled/pull/2905).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -714,7 +714,7 @@ if (static AND NOT APPLE)
 else ()
   set (Boost_USE_STATIC_RUNTIME OFF)
 endif ()
-find_package (Boost 1.67 REQUIRED
+find_package (Boost 1.70 REQUIRED
   COMPONENTS
     chrono
     context
@@ -748,15 +748,6 @@ target_link_libraries (ripple_boost
     Boost::serialization
     Boost::system
     Boost::thread)
-
-# workaround for xcode 10.2 and boost < 1.69
-# once we require Boost 1.69 or higher, this can be removed
-# see:  https://github.com/boostorg/asio/commit/43874d5
-if (CMAKE_CXX_COMPILER_ID STREQUAL "AppleClang" AND
-    CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 10.0.1.10010043 AND
-    Boost_VERSION LESS 106900)
-  target_compile_definitions (opts INTERFACE BOOST_ASIO_HAS_STD_STRING_VIEW)
-endif ()
 
 #[===================================================================[
    NIH dep: openssl
@@ -1419,8 +1410,8 @@ if (is_root_project) # NuDB not needed in the case of xrpl_core inclusion build
   if (CMAKE_VERSION VERSION_GREATER_EQUAL 3.11)
     FetchContent_Declare(
       nudb_src
-      GIT_REPOSITORY https://github.com/vinniefalco/NuDB.git
-      GIT_TAG        1.0.0
+      GIT_REPOSITORY https://github.com/CPPAlliance/NuDB.git
+      GIT_TAG        2.0.1
     )
     FetchContent_GetProperties(nudb_src)
     if(NOT nudb_src_POPULATED)
@@ -1430,8 +1421,8 @@ if (is_root_project) # NuDB not needed in the case of xrpl_core inclusion build
   else ()
     ExternalProject_Add (nudb_src
       PREFIX ${nih_cache_path}
-      GIT_REPOSITORY https://github.com/vinniefalco/NuDB.git
-      GIT_TAG 1.0.0
+      GIT_REPOSITORY https://github.com/CPPAlliance/NuDB.git
+      GIT_TAG 2.0.1
       CONFIGURE_COMMAND ""
       BUILD_COMMAND ""
       TEST_COMMAND ""
--- a/cfg/rippled-example.cfg
+++ b/cfg/rippled-example.cfg
@@ -868,18 +868,10 @@
 #       ...
 #
 #   Example:
-#       type=nudb
 #       path=db/shards/nudb
 #
-#   The "type" field must be present and controls the choice of backend:
-#
-#   type = NuDB
-#       NuDB is recommended for shards.
-#
-#   type = RocksDB
-#
 #   Required keys:
-#       path                Location to store the database (all types)
+#       path                Location to store the database
 #
 #       max_size_gb         Maximum disk space the database will utilize (in gigabytes)
 #
@@ -1193,7 +1185,6 @@ advisory_delete=0
 # NuDB requires SSD storage. Helpful information can be found here
 # https://ripple.com/build/history-sharding
 #[shard_db]
-#type=NuDB
 #path=/var/lib/rippled/db/shards/nudb
 #max_size_gb=500
 
--- a/src/ripple/nodestore/backend/NuDBFactory.cpp
+++ b/src/ripple/nodestore/backend/NuDBFactory.cpp
@@ -40,9 +40,6 @@ class NuDBBackend
     : public Backend
 {
 public:
-    // This needs to be tuned for the
-    // distribution of data sizes.
-    static constexpr std::size_t arena_alloc_size = megabytes(16);
     static constexpr std::size_t currentType = 1;
 
     beast::Journal j_;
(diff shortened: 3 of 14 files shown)


src/ripple/nodestore/impl/DatabaseShardImp.cpp
```diff
@@ -34,61 +34,97 @@ namespace NodeStore {
...
     {
-        fdLimit_ = Manager::instance().make_Backend(
-            config_, scheduler_, j_)->fdlimit();
+        JLOG(j_.error()) <<
+            "ledgers_per_shard must be a multiple of 256";
+        return false;
     }
-    catch (std::exception const&)
+
+    // NuDB is the default and only supported permanent storage backend
```
jwbusch: should this say `"nudb"` instead? (just to be consistent with how `"memory"` and `"none"` are written)

src/ripple/nodestore/impl/Shard.cpp
```diff
@@ -49,79 +49,68 @@ Shard::Shard(DatabaseShard const& db, std::uint32_t index,
...
     try
     {
-        backend_ = Manager::instance().make_Backend(
-            config, scheduler, j_);
-        backend_->open(!dirPreexist || dirEmpty);
+        backend_->open(!preexist);
 
+        // If the file descriptor limit is zero, then
+        // the backend does not use permanent storage
         if (backend_->fdlimit() == 0)
```
jwbusch: This seems like it should be a member function inside of the `backend_` interface...
`bool isPermanentStorage() const`... which internally might be `return fdlimit() == 0;` but then you don't need a comment explaining it here...

src/ripple/nodestore/impl/Shard.cpp
```diff
@@ -49,79 +49,68 @@ Shard::Shard(DatabaseShard const& db, std::uint32_t index,
...
+
     auto fail = [&](std::string msg)
     {
-        JLOG(j_.error()) <<
-            "shard " << std::to_string(index_) << " error: " << msg;
-
-        if (!dirPreexist)
-            removeAll(dir_, j_);
-        else if (dirEmpty)
+        if (!msg.empty())
```
jwbusch: Don't you miss out on logging some errors this way? Line 106 will be empty string, which means nothing in the journal.

src/ripple/nodestore/impl/Shard.cpp
```diff
@@ -49,79 +49,68 @@ Shard::Shard(DatabaseShard const& db, std::uint32_t index,
...
-            "shard " + std::to_string(index_) + " exception: " + e.what();
+            "shard " << index_ << ": " << ec.message();
         return false;
     }
 
+    config.set("path", dir_.string());
+    backend_ = factory->createInstance(
+        NodeObject::keyBytes, config, scheduler, ctx, j_);
+
     auto fail = [&](std::string msg)
```
jwbusch: nit: could be `std::string_view`

src/ripple/nodestore/impl/DatabaseShardImp.cpp
```diff
@@ -34,61 +34,97 @@ namespace NodeStore {
...
+                "Failed to create shard store type " << backendName_;
+            return false;
+        }
+        fdLimit_ = factory->createInstance(NodeObject::keyBytes,
+            config_, scheduler_, *ctx_, j_)->fdlimit();
+    }
+
+    // If the file descriptor limit is zero, then
+    // the backend does not use permanent storage
+    if (fdLimit_ == 0)
```
jwbusch: Seeing the same comment in 2 places definitely makes me think there should be a function.

src/ripple/nodestore/impl/DatabaseShardImp.cpp
```diff
@@ -34,61 +34,97 @@ namespace NodeStore {
...
 DatabaseShardImp::~DatabaseShardImp()
 {
     // Stop threads before data members are destroyed
     stopThreads();
+
+    std::lock_guard<std::mutex> lock(m_);
+    complete_.clear();
+    if (incomplete_)
+        incomplete_.reset();
+    try
```
jwbusch: any reason `ctx_` is special and needs exceptions caught and but not the rest of it? i.e.  could wrap the entire destructor in a try-catch? It's a little concerning that we need to catch anything during destruction, was there an instance where you saw it throwing exceptions under normal cleanup? 

```
DatabaseShardImp::~DatabaseShardImp() try {
  // cool code goes here
} catch(std::exception const& e) {
    JLOG(j_.error()) << 
        e.what(); 
}
```

HowardHinnant:
In boost 1.70 b1, the deprecated member function `get_io_service` has been removed from `basic_stream_socket`:
```
rippled/src/ripple/server/impl/PlainWSPeer.h:68:36: error: no member named 'get_io_service' in
      'boost::asio::basic_stream_socket<boost::asio::ip::tcp, boost::asio::executor>'
        std::move(request), socket.get_io_service(), journal)
                            ~~~~~~ ^
```

MarkusTeufelberger:
In what context are shards queried for data by rippled by the way? Do they need ultra-fast random access by key or would even a sorted csv file be enough to share history (and import shard contents into the node store on demand) in theory?

cfg/rippled-example.cfg
```diff
@@ -864,19 +864,8 @@
...
-#       NuDB is recommended for shards.
-#
-#   type = RocksDB
-#
-#       The RocksDB backend also provides these optional parameters:
-#
-#       compression         0 for none, 1 for Snappy compression
 #
 #   Required keys:
 #       path                Location to store the database (all types)
```
MarkusTeufelberger:
```suggestion
#       path                Location to store the database
```

CMakeLists.txt
```diff
@@ -698,7 +698,7 @@ if (static AND NOT APPLE)
 else ()
   set (Boost_USE_STATIC_RUNTIME OFF)
 endif ()
-find_package (Boost 1.67 REQUIRED
+find_package (Boost 1.69 REQUIRED
```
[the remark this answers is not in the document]
miguelportilla: Unfortunately, because 1.69 is broken on MacOS, I may need to bump the requirement to 1.70.

cfg/rippled-example.cfg
miguelportilla: @MarkusTeufelberger nice catch, thank you.

src/ripple/nodestore/impl/DatabaseShardImp.cpp
miguelportilla: The `context` dtor will throw if any database using the object is open. It is the caller's responsibility to close databases before destroying the `context`. This try/catch is not needed since we are closing the databases beforehand.

src/ripple/nodestore/impl/DatabaseShardImp.cpp
miguelportilla: No, since this is referring to the name. But the string literals should be "NuDB",  "Memory" and "none", thanks.

src/ripple/nodestore/impl/Shard.cpp
miguelportilla: No, the fail in `saveControl` already logged the error.

src/ripple/nodestore/backend/NuDBFactory.cpp
```diff
@@ -65,6 +65,24 @@ class NuDBBackend
                 "nodestore: Missing path in NuDB backend");
     }
 
+    NuDBBackend (
+        int keyBytes,
```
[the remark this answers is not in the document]
miguelportilla: Will fix, thanks.

src/ripple/nodestore/impl/Shard.cpp
miguelportilla: Agreed, will implement. 👍

miguelportilla: 👍

miguelportilla:
@HowardHinnant, @seelabs is working on a PR that addresses the necessary rippled changes for boost 1.70.

miguelportilla:
@MarkusTeufelberger The current version of rippled queries shards when node objects aren't found in the node store, primarily for history. The future plan is to have shards replace the need entirely for the node store. One of the planned features is to allow more than a single storage path. That would remove or alleviate the requirement of large SSD drives and RAID hardware. It would also allow a mixed-use of drives such as SSD for recent shards and economical mechanical drives for older history.

HowardHinnant:
After cherry-picking https://github.com/seelabs/rippled/commit/7070d01bb420f50bc85baa63c332475cc54154a0 on to this branch, this passes on macOS using boost 1.70.0b1.  However, as noted at https://github.com/seelabs/rippled/commit/7070d01bb420f50bc85baa63c332475cc54154a0

On macOS using boost 1.70b1, I get about 100 link warnings that look like:

ld: warning: direct access in function 'boost::exception_ptr boost::exception_detail::get_static_exception_objectboost::exception_detail::bad_alloc_()' from file '/Users/howardhinnant/Development/boost_1_70_0_B1/stage/lib/libboost_coroutine.a(stack_traits.o)' to global weak symbol 'guard variable for boost::exception_ptr boost::exception_detail::get_static_exception_objectboost::exception_detail::bad_alloc_()::ep' from file 'CMakeFiles/rippled.dir/src/ripple/app/consensus/RCLConsensus.cpp.o' means the weak symbol cannot be overridden at runtime. This was likely caused by different translation units being compiled with different visibility settings.

I currently do not know how to address these warnings, or why they have appeared by simply changing boost versions. boost 1.70b1 was built using the macOS build instructions at Builds/macos/README.md (as was boost 1.67).

This commit does not build on macOS using boost 1.67, as designed.

nbougalis:
I believe this has been merged as a988b3224fa898523e94769b4b518e808aa1eb90

PR #2902, opened 2019-04-10
Add request timeout for ValidatorSites:

mellery451:
Fixes: RIPD-1737

Enforce a 20s timeout for making validator list requests.

### DIFF
--- a/src/ripple/app/misc/impl/ValidatorSite.cpp
+++ b/src/ripple/app/misc/impl/ValidatorSite.cpp
@@ -177,7 +177,15 @@ ValidatorSite::stop()
         sp->cancel();
     cv_.wait(lock, [&]{ return ! fetching_; });
 
-    timer_.cancel();
+    // docs indicate cancel() can throw, but this should be
+    // reconsidered if it changes to noexcept
+    try
+    {
+        timer_.cancel();
+    }
+    catch (boost::system::system_error const&)
+    {
+    }
     stopping_ = false;
     pending_ = false;
     cv_.notify_all();
@@ -220,7 +228,15 @@ ValidatorSite::makeRequest (
         [this] ()
         {
             std::lock_guard <std::mutex> lock_state{state_mutex_};
-            timer_.cancel_one();
+            // docs indicate cancel_one() can throw, but this
+            // should be reconsidered if it changes to noexcept
+            try
+            {
+                timer_.cancel_one();
+            }
+            catch (boost::system::system_error const&)
+            {
+            }
         };
     auto onFetch =
         [this, siteIdx, timeoutCancel] (
--- a/src/ripple/app/misc/impl/ValidatorSite.cpp
+++ b/src/ripple/app/misc/impl/ValidatorSite.cpp
@@ -26,15 +26,15 @@
 #include <ripple/basics/Slice.h>
 #include <ripple/json/json_reader.h>
 #include <ripple/protocol/JsonFields.h>
+#include <boost/algorithm/clamp.hpp>
 #include <boost/regex.hpp>
 #include <algorithm>
 
 namespace ripple {
 
-// default site query frequency - 5 minutes
-auto constexpr DEFAULT_REFRESH_INTERVAL = std::chrono::minutes{5};
-auto constexpr ERROR_RETRY_INTERVAL = std::chrono::seconds{30};
-unsigned short constexpr MAX_REDIRECTS = 3;
+auto           constexpr DEFAULT_REFRESH_INTERVAL = std::chrono::minutes{5};
+auto           constexpr ERROR_RETRY_INTERVAL     = std::chrono::seconds{30};
+unsigned short constexpr MAX_REDIRECTS            = 3;
 
 ValidatorSite::Site::Resource::Resource (std::string uri_)
     : uri {std::move(uri_)}
@@ -90,7 +90,8 @@ ValidatorSite::Site::Site (std::string uri)
 ValidatorSite::ValidatorSite (
     boost::asio::io_service& ios,
     ValidatorList& validators,
-    beast::Journal j)
+    beast::Journal j,
+    std::chrono::seconds timeout)
     : ios_ (ios)
     , validators_ (validators)
     , j_ (j)
@@ -98,6 +99,7 @@ ValidatorSite::ValidatorSite (
     , fetching_ (false)
     , pending_ (false)
     , stopping_ (false)
+    , requestTimeout_ (timeout)
 {
 }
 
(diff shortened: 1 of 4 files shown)


src/test/jtx/TrustedPublisherServer.h
```diff
@@ -82,14 +83,16 @@ class TrustedPublisherServer
...
         std::string blob = base64_encode(data);
-
-        list_ = "{\"blob\":\"" + blob + "\"";
-
         auto const sig = sign(keys.first, keys.second, makeSlice(data));
-
-        list_ += ",\"signature\":\"" + strHex(sig) + "\"";
-        list_ += ",\"manifest\":\"" + manifest + "\"";
-        list_ += ",\"version\":" + std::to_string(version) + '}';
+        getList_ = [blob, sig, manifest, version](int interval) {
```
scottschurr: :+1: Nice solution to passing in the interval.

src/test/jtx/TrustedPublisherServer.h
```diff
@@ -181,7 +184,19 @@ class TrustedPublisherServer
...
+                    int refresh = 5;
+                    if (boost::starts_with(path, "/validators/refresh"))
+                        refresh =
+                            boost::lexical_cast<unsigned int>(path.substr(20));
+                    res.body() = getList_(refresh);
+                }
+            }
+            else if (boost::starts_with(path, "/sleep/"))
+            {
+                auto sleep_sec =
```
scottschurr: `auto const`?

src/ripple/app/misc/impl/ValidatorSite.cpp
```diff
@@ -210,17 +214,30 @@ ValidatorSite::makeRequest (
     fetching_ = true;
     sites_[siteIdx].activeResource = resource;
     std::shared_ptr<detail::Work> sp;
+    auto timeoutCancel =
+        [this] ()
+        {
+            std::unique_lock <std::mutex> lock_state{state_mutex_};
```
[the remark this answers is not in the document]
mellery451: fixed - thanks for pointing that out.

src/ripple/app/misc/impl/ValidatorSite.cpp
```diff
@@ -370,10 +413,14 @@ ValidatorSite::parseJsonResponse (
...
-        sites_[siteIdx].refreshInterval =
-            std::chrono::minutes{body["refresh_interval"].asUInt ()};
+        auto refresh =
+            boost::algorithm::clamp(
+                body["refresh_interval"].asUInt (),
+                1u,
+                60u*24);
+        sites_[siteIdx].refreshInterval = std::chrono::minutes{refresh};
+        sites_[siteIdx].nextRefresh =
+            clock_type::now() + sites_[siteIdx].refreshInterval;
```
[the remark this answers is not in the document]
mellery451: 👍  - fixed

src/ripple/app/misc/impl/ValidatorSite.cpp
```diff
@@ -252,6 +269,32 @@ ValidatorSite::makeRequest (
 
     work_ = sp;
     sp->run ();
+    // start a timer for the request, which shouldn't take more
+    // than requestTimeout_ to complete
+    std::unique_lock <std::mutex> lock_state{state_mutex_};
+    timer_.expires_at ( clock_type::now() + requestTimeout_);
```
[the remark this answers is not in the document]
mellery451: yep, `expires_after` is better here - thanks.

src/ripple/app/misc/impl/ValidatorSite.cpp
```diff
@@ -210,17 +214,30 @@ ValidatorSite::makeRequest (
     fetching_ = true;
     sites_[siteIdx].activeResource = resource;
     std::shared_ptr<detail::Work> sp;
+    auto timeoutCancel =
+        [this] ()
+        {
+            std::unique_lock <std::mutex> lock_state{state_mutex_};
+            error_code ec;
+            timer_.cancel_one(ec);
+        };
```
miguelportilla: This overload of `cancel_one` is deprecated. You can use the non-error_code overload. eg.  `timer_.cancel_one();`

src/ripple/app/misc/impl/ValidatorSite.cpp
```diff
@@ -252,6 +269,32 @@ ValidatorSite::makeRequest (
...
+void
+ValidatorSite::onRequestTimeout (
+    std::size_t siteIdx,
+    error_code const& ec)
+{
+    if (ec)
+        return;
+
+    {
+        std::unique_lock <std::mutex> lock_site{sites_mutex_};
```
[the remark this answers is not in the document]
miguelportilla: While you are here... it seems the caught exception at line 327, `onSiteFetch` is called while already locked on `sites_mutex`. I believe that is UB.

It would also be nice to add a documenting param to `setTimer`  to make it clear that it is always called with a lock on `state_mutex_`

src/ripple/app/misc/impl/ValidatorSite.cpp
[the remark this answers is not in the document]
miguelportilla: suggesion... `1h * 24` -> `24h`

src/test/app/ValidatorSite_test.cpp
```diff
@@ -185,12 +187,24 @@ class ValidatorSite_test : public beast::unit_test::suite
...
     };
 
+    struct fetchListConfig
+    {
+        std::string path_;
+        std::string msg_;
+        bool failFetch_ = false;
+        bool failApply_ = false;
+        int serverVersion_ = 1;
+        std::chrono::seconds expiresFromNow_ = std::chrono::seconds{3600};
```
miguelportilla: nit: Our format typically does not use a trailing underscore denoting private on public struct members.

src/test/jtx/TrustedPublisherServer.h
```diff
@@ -181,7 +184,19 @@ class TrustedPublisherServer
...
+                    boost::lexical_cast<unsigned int>(path.substr(7));
```
miguelportilla: This can throw.

src/test/jtx/TrustedPublisherServer.h
```diff
@@ -181,7 +184,19 @@ class TrustedPublisherServer
                 else if (path == "/validators/missing")
                     res.body() = "{\"version\": 1}";
                 else
-                    res.body() = list_;
+                {
+                    int refresh = 5;
+                    if (boost::starts_with(path, "/validators/refresh"))
+                        refresh =
+                            boost::lexical_cast<unsigned int>(path.substr(20));
```
miguelportilla: This can throw.

src/ripple/app/misc/impl/ValidatorSite.cpp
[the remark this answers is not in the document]
mellery451: yep - good point. I'll move the lock inside the try and that should fix it. I'll also remove that assert since we don't have a state lock to protect that access (strictly..perhaps this could have just been an atomic). I'll add the lock param to setTimer too.

src/ripple/app/misc/impl/ValidatorSite.cpp
mellery451: fixed

seelabs: This doesn't appear to be fixed yet.

PR #2905, opened 2019-04-12
Support Boost 1.70:

seelabs:
This patch removes calls to several deprecated asio functions.

* `io_service::post` becomes `post` (free function)
* `io_service::work` becomes `executor_work_guard`
* `io_service::wrap` becomes `bind_executor`
* `get_io_context`   becomes `get_executor` or `get_executor().context()`

This patch was tested with boost 1.69 and 1.70. The functions
`ripple::get_lowest_layer` and `beast::create_waitable_timer` are required to
handle a breaking difference between these versions. When rippled no longer
needs to support pre 1.70 boost versions, both of these functions may be
removed, and the waitable timer injections may also be removed.

Note: this was tested against boost 1.69 and boost 1.70 rc2. We should not merge this until we test against the actual released boost 1.70.

@HowardHinnant @miguelportilla


seelabs:
Pushed with updated readme with @HowardHinnant changes. Note: reverted the change saying the min version was 1.70. I believe it's still 1.67 with this commit.

HowardHinnant:
I'll double check that.  I tried with boost 1.67 and it failed.  But I now think that during yesterday's boost 1.70 linker warning hunt I left my 1.67 in a compromised state.  Rebuilding 1.67 and will retest this against it, and will comment back here.

seelabs:
Re-added @HowardHinnant docs about 1.70 being the min boost version on mac. There are build issues with older versions.

HowardHinnant:
More info:  I'm unable to build boost 1.67 with the latest Apple developer tools.  However boost binaries are available (https://sourceforge.net/projects/boost/files/boost-binaries/).  And I've confirmed that this commit works with boost 1.67, assuming 1.67 is already built.

seelabs:
Thanks @HowardHinnant Given this is a boost build issue and not a rippled issue, I changed the min version back to 1.67

seelabs:
The the gcc.debug jenkins test are failing with `ripple.http.Server failed to complete. The child process may have crashed.` I'll look into this.

miguelportilla:
Tested on Windows using Boost 1.67, 1.69 and 1.70.

src/ripple/server/impl/LowestLayer.h
```diff
@@ -0,0 +1,46 @@
...
+
+#if BOOST_VERSION >= 107000
+#include <boost/beast/core/stream_traits.hpp>
+#else
+#include <boost/beast/core/type_traits.hpp>
+#endif
+
+namespace ripple {
+
+// Before boost 1.70, get_lowest_layer required an explicit templat parameter
```
miguelportilla: misspell `templat`

seelabs:
Rebased and pushed FileBoost change from @mellery451

Issue #2675, opened 2018-08-30
"Connection reset by peer" - which one?


MarkusTeufelberger:
The "Connection reset by peer" log message shows up from time to time. To detect and debug potential problems there, it would be nice if the IP and the public key of the peer that reset the connection would also get logged, so it is easier to find out if it is just one single peer that is having troubles or if it is a more general issue.

PR #2907, opened 2019-04-18
Add the peer IP to log messages.

empatogen:
This partially fixes Issue #2675 (the IP should be enough)


MarkusTeufelberger:
The IP is not enough in all cases, since there can be more than one peer behind the same public IP.

MarkusTeufelberger:
Great, thanks! :-)

empatogen:
Yeah, I thought it didn't matter really, as the IP routing the messages is still the peer that disconnects. Though, it could probably fail in different ways and more detailed messages is never a bad thing. :-)

PR #2909, opened 2019-04-18
Packaging - add distros to smoketest

mellery451:
minor change to add/clarify distro versions that we use for pkg smoketests


mellery451:
closed and rolled into https://github.com/ripple/rippled/pull/2913

PR #2910, opened 2019-04-23
 Add units to all fee calculations:

### LINKED ISSUE #2451: [FeeVote] Voting for fee change and its implications are unclear
Hello!
On my rippled fork i'm trying to figure out how are all Fee-related things working. I think it might even be a bug in how certain fee-related variables are used in code.
There are several places from which we can obtain fee stats including:

1. server_state (raw values (in drops???))
2. server_info (human readable server_state output - with potential bug?)
3. fee
4. subscription to closed ledgers where we can see 'fee_ref' value (which is actually not votable 'reference_fee_units' == 10, not the votable 'reference_fee').
----
at first i set up recommended values on my standalone server. and then advance 256 ledgers to first flag ledger and query server_state, server_info, fee

rozek@***:~$ tail /etc/rippled/rippled.cfg
[features]
MultiSign

[voting]
**reference_fee = 10**
account_reserve = 20000000
owner_reserve = 5000000

rozek@***:~$ /usr/bin/rippled --conf=/etc/rippled/rippled.cfg server_info
Loading: "/etc/rippled/rippled.cfg"
2018-Mar-23 12:38:54 HTTPClient:NFO Connecting to 127.0.0.1:5005

{
   "result" : {
      "info" : {
         "build_version" : "1.0.0-b2",
         "complete_ledgers" : "2-261",
         "hostid" : "***",
         "io_latency_ms" : 1,
         "jq_trans_overflow" : "0",
         "last_close" : {
            "converge_time_s" : 0.1,
            "proposers" : 0
         },
         --- editors cut --- 
         "validated_ledger" : {
            **"base_fee_xrp" : 1e-05,**
            "hash" : "94B62A82705D7B241AA8B2BA4C0EBF772905E9A3F8CF76B4F971C1BCDA82D233",
            **"reserve_base_xrp" : 20,
            "reserve_inc_xrp" : 5,**
            "seq" : 261
         },
         "validation_quorum" : 0,
         "validator_list_expires" : "unknown"
      },
      "status" : "success"
   }
}
rozek@***:~$ /usr/bin/rippled --conf=/etc/rippled/rippled.cfg server_state
Loading: "/etc/rippled/rippled.cfg"
2018-Mar-23 12:38:57 HTTPClient:NFO Connecting to 127.0.0.1:5005

{
   "result" : {
      "state" : {
         "build_version" : "1.0.0-b2",
         "complete_ledgers" : "2-261",
         "io_latency_ms" : 1,
         "jq_trans_overflow" : "0",
         "last_close" : {
            "converge_time" : 100,
            "proposers" : 0
         },
         "load" : {
            "job_types" : [
               {
                  "in_progress" : 1,
                  "job_type" : "clientCommand"
               }
            ],
            "threads" : 1
         },
         "load_base" : 256,
         "load_factor" : 256,
         "load_factor_fee_escalation" : 256,
         "load_factor_fee_queue" : 256,
         "load_factor_fee_reference" : 256,
         "load_factor_server" : 256,
         --- editors cut --- 
         "validated_ledger" : {
            **"base_fee" : 10,**
            "close_time" : 575124139,
            "hash" : "94B62A82705D7B241AA8B2BA4C0EBF772905E9A3F8CF76B4F971C1BCDA82D233",
            **"reserve_base" : 20000000,
            "reserve_inc" : 5000000,**
            "seq" : 261
         },
         "validation_quorum" : 0,
         "validator_list_expires" : 0
      },
      "status" : "success"
   }
}
```
rozek@***:~$ /usr/bin/rippled --conf=/etc/rippled/rippled.cfg fee
Loading: "/etc/rippled/rippled.cfg"
2018-Mar-23 12:39:02 HTTPClient:NFO Connecting to 127.0.0.1:5005

{
   "result" : {
      "current_ledger_size" : "1",
      "current_queue_size" : "0",
      "drops" : {
         "base_fee" : "10",
         "median_fee" : "5000",
         "minimum_fee" : "10",
         "open_ledger_fee" : "10"
      },
      "expected_ledger_size" : "1000",
      "ledger_current_index" : 262,
      "levels" : {
         "median_level" : "128000",
         "minimum_level" : "256",
         "open_ledger_level" : "256",
         "reference_level" : "256"
      },
      "max_queue_size" : "20000",
      "status" : "success"
   }
}
```

so far so good, everything is allright.
now modify rippled.cfg to reference_fee = 100  and hit next flag ledger
rozek@***:~$ tail /etc/rippled/rippled.cfg
[features]
MultiSign
[voting]
**reference_fee = 100**
account_reserve = 20000000
owner_reserve = 5000000

in code reference_fee from conf file is mapped to FeeVote::Setup::reference_fee which is loaded to Fees::base (Ledger class member). there is also an Fees::units member which is not voteable (it is added to SetFee tx though) and hardcoded to 10.
Now what seems not correct to me is calculation of human-readable response to server_info command in NetworkOPs.cpp:2286
```
l[jss::reserve_base_xrp]   = static_cast<double> (Json::UInt (
                    lpClosed->fees().accountReserve(0).drops() * baseFee / baseRef))
/ SYSTEM_CURRENCY_PARTS;
```
baseFee / baseRef gives fine result when both are equal to 10, but if baseFee (which is votable) changes, this also has impact on how values are reported in server_info.
**similar construction * baseFee / baseRef is used in a few more places in code and i'm wondering if that has any logical impact on applied fees if fee is changed via voting?**
proof malformed server_info below:

rozek@***:~$ /usr/bin/rippled --conf=/etc/rippled/rippled.cfg server_state
Loading: "/etc/rippled/rippled.cfg"
2018-Mar-23 12:44:57 HTTPClient:NFO Connecting to 127.0.0.1:5005

{
   "result" : {
      "state" : {
         "build_version" : "1.0.0-b2",
         "complete_ledgers" : "2-572",
         "io_latency_ms" : 1,
         "jq_trans_overflow" : "0",
         "last_close" : {
            "converge_time" : 100,
            "proposers" : 0
         },
         "load" : {
            "job_types" : [
               {
                  "avg_time" : 4,
                  "in_progress" : 1,
                  "job_type" : "clientCommand",
                  "peak_time" : 29
               },
               {
                  "job_type" : "WriteNode",
                  "per_second" : 1
               }
            ],
            "threads" : 1
         },
         "load_base" : 256,
         "load_factor" : 256,
         "load_factor_fee_escalation" : 256,
         "load_factor_fee_queue" : 256,
         "load_factor_fee_reference" : 256,
         "load_factor_server" : 256,
          --- editors cut ---
         "validated_ledger" : {
            **"base_fee" : 100,**
            "close_time" : 575124750,
            "hash" : "E4773984777DCDBFEE7CF84077FB88FFE0AC920A58F9096E03D599D287F4B3B1",
            **"reserve_base" : 20000000,
            "reserve_inc" : 5000000,**
            "seq" : 572
         },
         "validation_quorum" : 0,
         "validator_list_expires" : 0
      },
      "status" : "success"
   }
}
rozek@***:~$ /usr/bin/rippled --conf=/etc/rippled/rippled.cfg server_info
Loading: "/etc/rippled/rippled.cfg"
2018-Mar-23 12:45:13 HTTPClient:NFO Connecting to 127.0.0.1:5005

{
   "result" : {
      "info" : {
         "build_version" : "1.0.0-b2",
         "complete_ledgers" : "2-572",
         "hostid" : "***",
         "io_latency_ms" : 1,
         "jq_trans_overflow" : "0",
         --- editors cut ---
         "validated_ledger" : {
            **"base_fee_xrp" : 0.0001,**
            "hash" : "E4773984777DCDBFEE7CF84077FB88FFE0AC920A58F9096E03D599D287F4B3B1",
            **"reserve_base_xrp" : 200,
            "reserve_inc_xrp" : 50,**
            "seq" : 572
         },
         "validation_quorum" : 0,
         "validator_list_expires" : "unknown"
      },
      "status" : "success"
   }
}
```
rozek@***:~$ /usr/bin/rippled --conf=/etc/rippled/rippled.cfg fee
Loading: "/etc/rippled/rippled.cfg"
2018-Mar-23 12:45:26 HTTPClient:NFO Connecting to 127.0.0.1:5005

{
   "result" : {
      "current_ledger_size" : "1",
      "current_queue_size" : "0",
      "drops" : {
         "base_fee" : "100",
         "median_fee" : "50000",
         "minimum_fee" : "100",
         "open_ledger_fee" : "100"
      },
      "expected_ledger_size" : "1000",
      "ledger_current_index" : 573,
      "levels" : {
         "median_level" : "128000",
         "minimum_level" : "256",
         "open_ledger_level" : "256",
         "reference_level" : "256"
      },
      "max_queue_size" : "20000",
      "status" : "success"
   }
}
```

Now my questions...
is it a bug in displaying human-readable values? ( i guess so?)
is it working as expected? if so, why fee increase results in account_reserve increase?
and most important: what does 'fee in fee units' or 'fee units' or FeeVote::Setup::reference_fee_units mean? could anyone enlighten me please?
(it is for instance referenced in TransactionSign.cpp in checkFee function)

Thank you very much in advance!

btw. Great work!

ximinez:
* Uses the boost::units library with units for drops,
  fee units (LoadFeeTrack), and fee levels (TxQ).
* Resolves #2451

This is a rewrite of the closed PR #2884.

While investigating #2451, I decided to do an audit of all fee calculations. I found a couple where units were not used correctly (eg. treating a drops value as fee units, and applying the conversion again), including those mentioned in the original problem. I figured that rather than fixing problems one at a time only to risk them being reintroduced, I'd convert all usages of fee values into a typed variable rather than a generic `uint32_t` or `uint64_t`.

Implementation notes:
1. I created a new `unsafe_cast` alongside `safe_cast`, because it helped me better visualize some of what was going on.

Here are links to the locations in 1.2.4 (so the links will stay consistent) that fell out of this work as using the wrong units.
* https://github.com/ripple/rippled/blob/1.2.4/src/ripple/app/misc/NetworkOPs.cpp#L2327-L2329
  * Treats drops as fee units and tries to convert back to drops.
* https://github.com/ripple/rippled/blob/1.2.4/src/ripple/app/misc/NetworkOPs.cpp#L2331-L2333
  * Treats drops as fee units and tries to convert back to drops.
* https://github.com/ripple/rippled/blob/1.2.4/src/ripple/app/misc/impl/TxQ.cpp#L93-L95
  * `calculateBaseFee` returns fee units, but the result is treated as drops.
* https://github.com/ripple/rippled/blob/1.2.4/src/ripple/app/misc/impl/TxQ.cpp#L671-L673
  * `calculateBaseFee` returns fee units, but the result is treated as drops.
* https://github.com/ripple/rippled/blob/1.2.4/src/test/ledger/Invariants_test.cpp#L97
  * `ApplyContext` wants fee units, not drops.
* https://github.com/ripple/rippled/blob/1.2.4/src/test/rpc/NoRippleCheck_test.cpp#L293
  * Tries to use a fee level as the transaction fee, instead of drops.
* https://github.com/ripple/rippled/blob/1.2.4/src/test/rpc/NoRippleCheck_test.cpp#L297
  * Tries to use a fee level as the transaction fee, instead of drops.
* https://github.com/ripple/rippled/blob/1.2.4/src/test/rpc/NoRippleCheck_test.cpp#L300
  * Tries to use a fee level as the transaction fee, instead of drops.


ximinez:
@HowardHinnant Could you make sure this one builds on MacOS and that I didn't reintroduce any problems?

src/ripple/rpc/impl/TransactionSign.cpp
```diff
@@ -711,22 +711,23 @@ Json::Value checkFee (
     }
 
     // Default fee in fee units.
-    std::uint64_t const feeDefault = config.TRANSACTION_FEE_BASE;
+    FeeUnit32 const feeDefault = config.TRANSACTION_FEE_BASE;
```
ximinez: I intentionally changed this from 64 to 32 to match the ledger, and to make downstream conversions easier

HowardHinnant:
> @HowardHinnant Could you make sure this one builds on MacOS and that I didn't reintroduce any problems?

Still :thumbsup: here.

ximinez:
> > @HowardHinnant Could you make sure this one builds on MacOS and that I didn't reintroduce any problems?
> 
> Still  here.

Woo hoo! Thanks!

ximinez:
Rebased on top of 1.3.0-b3. Everything went swimmingly.

src/ripple/protocol/STObject.h
```diff
@@ -734,6 +761,22 @@ STObject::ValueProxy<T>::operator= (U&& u)
     return *this;
 }
 
+template <class T>
+template <class Unit, class U >
+std::enable_if_t<
+    std::is_assignable_v<T, U> &&
+    (
+    std::is_same_v<Unit, units::drop_unit> ||
```
mellery451: do we actually need to restrict `Unit` to these three types? I'll admit I don't fully understand this code, but I was hoping the assign from argument type would be sufficient.

ximinez: I consider this an important safety mechanism because we're stripping out unit information to make the assignment. Mechanically, the main intention is to keep values with "intermediate" or "composed" units from incorrectly being assigned into objects and losing that unit information. Consider the following pseudocode:
```
STObject o = [whatever];
auto x = 150 * units::drops;
auto y = 10 * units::feeunits;

o[sfField1] = x; // This is good. x is in drops, so it makes sense to allow this assignment. sfField1 is 150.
o[sfField2] = y; // This is also good. y is fee units, so it makes sense to allow this assignment. sfField2 is 10.
o[sfField3] = x * y; // This is problematic. The unit of x*y is basically "fee_unit drops". sfField3 is 1500, but 1500 what?
o[sfField4] = x / y; // Wat. The unit of x/y is basically "drops/fee_units". sfField4 has a value of 15, but again 15 what?
```
So the intention of this restriction is to prevent assignments like sfField3 & 4 from happening by accident. If one ever needs to be done intentionally, then we can add that case later. For now, I'd rather it be restrictive.

src/ripple/app/misc/impl/LoadFeeTrack.cpp
```diff
@@ -120,17 +119,90 @@ void lowestTerms(T1& a,  T2& b)
...
+
+template <class T1, class T2,
+    class = std::enable_if_t <
+        std::is_unsigned<T1>::value &&
+        sizeof(T1) <= sizeof(std::uint64_t) >,
+    class = std::enable_if_t <
+        std::is_unsigned<T2>::value &&
+        sizeof(T2) <= sizeof(std::uint64_t) >
+>
+void lowestTerms(T1& a,  T2& b)
```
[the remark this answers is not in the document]
ximinez: Looks like "no" and "yes".

src/ripple/app/misc/impl/LoadFeeTrack.cpp
```diff
@@ -120,17 +119,90 @@ void lowestTerms(T1& a,  T2& b)
...
+        sizeof(T1) <= sizeof(std::uint64_t) >,
+    class = std::enable_if_t <
+        std::is_unsigned<T2>::value &&
+        sizeof(T2) <= sizeof(std::uint64_t)>,
+    class = std::enable_if_t <
+        std::is_same_v<Unit1, units::feeunit_unit> ||
+        std::is_same_v<Unit1, units::drop_unit>
+    >
+>
+void lowestTerms(boost::units::quantity<Unit1, T1>& a,
```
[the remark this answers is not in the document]
ximinez: I was able to make it work with lambdas local to `scaleFeeLoad`.

src/ripple/basics/feeunits.h
```diff
@@ -0,0 +1,431 @@
...
+std::basic_ostream<Char, Traits>&
+operator<<(std::basic_ostream<Char, Traits>& os,
+    const quantity<Unit, T>& q)
+{
+    os << q.value();
+
+    return(os);
+}
+
+// get string representation of a quantity
```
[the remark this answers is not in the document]
ximinez: Even better, I can just do `return to_string(t.value());`. I originally used `operator<<` to future-proof it in case I only wanted to change the output format in the future. I doubt that's going to happen, though, honestly.

src/ripple/basics/feeunits.h
```diff
@@ -0,0 +1,431 @@
...
+    return unsafe_cast<typename Dest::value_type>(s.value()) * uus;
+}
+
+namespace detail
+{
+
+// Need to cap to uint64 to uint32 due to JSON limitations
+template<class Unit>
+boost::units::quantity<Unit, std::uint32_t>
+clamp(boost::units::quantity<Unit, std::uint64_t> v)
```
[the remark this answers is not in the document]
ximinez: Renamed to `clamp32`. Easy enough since it's in the `detail` namespace.

ximinez:
Rebased on to 1.3.0-b4, and resolved a couple of conflicts

src/ripple/app/misc/impl/LoadFeeTrack.cpp
```diff
@@ -120,17 +104,49 @@ void lowestTerms(T1& a,  T2& b)
...
 // Scale using load as well as base rate
-std::uint64_t
-scaleFeeLoad(std::uint64_t fee, LoadFeeTrack const& feeTrack,
+Drops64
+scaleFeeLoad(FeeUnit64 fee, LoadFeeTrack const& feeTrack,
     Fees const& fees, bool bUnlimited)
 {
-    if (fee == 0)
-        return fee;
+    auto lowestTerms1 = [](auto& a, auto& b)
```
jwbusch: nit: would prefer a better name than `lowestTerms1` and `lowestTerms2`, or find a way to combine the two functions with `if constexpr` something like... 
``` 
auto lowestTerms = [](auto& a, auto& b) {
    auto get_value = [](auto v) {
        if constexpr (std::is_same_v<decltype(v), TYPE_WITH_VALUE_MEMBER_FUNC>) {
            return v.value();
        } else {
            return v;
        }
    };

    if (auto const g = gcd(get_value(a), get_value(b))) 
    { 
        a /= g;
        b /= g;
    }
};
```
but no big deal...

seelabs:
@ximinez I didn't do a deep review of this PR, but I took a quick look and have two concerns.

The first is the use of the boost units library. This is usually used for dimensional analysis, and it doesn't appear like that's how it's used here. For example, the PR defines three base dimensions: "drop", "fee", and "fee level". However, both "drop" and "fee" are measured in XRP, so they shouldn't be distinct dimensions - they are both currencies. In fact, "LoadFeeTrack.cpp" has a function "maybe_swap" what swaps a variable of type "fee" with one of type "drops", showing they have the same dimension. Also "fee levels" appears to be dimensionless: multiplying a "fee" by a "fee level" gives a fee.

I don't object to making types for "fees" and "fee levels" and making the interface more type safe by doing so. However, using a dimensional analysis library for this seems like the wrong tool for the job.

For this PR, there are very few combinations of drops, fees, and fee units that make sense. I'd just define those functions and lose the boost units library.

My second concern is adding yet another type to represent XRP amounts. We can if we must, but is there something you need that `XRPAmount` can already do (or be modified to do)?

ximinez:
I've basically done a "back to the drawing board" rewrite, which primarily does not introduce yet another type to represent drops / XRP. Instead, it very slightly and safely expands `XRPAmount`, and uses it.

Rather than closing the PR and opening a new one, I've "rerequested" reviews from the current reviewers.

src/ripple/app/misc/NetworkOPs.cpp
```diff
@@ -2310,33 +2306,35 @@ Json::Value NetworkOPsImp::getServerInfo (bool human, bool admin, bool counters)
...
+            l[jss::close_time] = Json::Value::UInt (
+                lpClosed->info().closeTime.time_since_epoch().count());
         }
         else
         {
-            l[jss::base_fee_xrp] = static_cast<double> (baseFee) /
-                    SYSTEM_CURRENCY_PARTS;
+            // Make a local specialization of the TaggedFee class, using
+            // XRPAmount as the "unit" to make some of the math here easier.
+            using DropsDouble = units::TaggedFee<XRPAmount, double>;
```
seelabs: It looks like this type is `XRPDouble`, not `DropsDouble`. I also wouldn't use `XRPAmount` as a tag, I'd make another empty struct like the others. I don't think this type is buying us much anyway, and it does seem like we're using this `TaggedFee` type differently than the others. I'd consider just using `double` for this.

src/ripple/app/misc/impl/LoadFeeTrack.cpp
```diff
@@ -81,56 +82,110 @@ LoadFeeTrack::lowerLocalFee ()
...
+template <class T>
+std::enable_if_t<std::is_unsigned_v<T>, T>
+abs(T t)
+{
+    return t;
+}
+
+template <class T>
+std::enable_if_t<std::is_signed_v<T>, T>
+abs(T t)
```
seelabs: Given how `abs` is used, I might have both functions return `std::uint64_t`. For the signed version, I'd cast to `int64` before taking the abs (E.g. as written we'll get the wrong answer for `numeric_limits<std::int8_t>::min()`).

src/ripple/basics/XRPAmount.h
```diff
@@ -0,0 +1,298 @@
...
+    XRPAmount&
+    operator= (Other drops)
+    {
+        drops_ = static_cast<value_type> (drops);
+        return *this;
+    }
+
+    constexpr
+    XRPAmount
+    operator*(value_type const& rhs) const
```
seelabs: This adds `XRPAmount * scalar`, we should add `scalar * XRPAmount` as well. Also, if we can multiply we should probably be able to divide as well (I'm usually against adding functions we don't use, but it does seem incomplete without it).

src/ripple/basics/XRPAmount.h
```diff
@@ -0,0 +1,298 @@
...
+    if (m % den)
+    {
+        if (!neg && roundUp)
+            r += 1;
+        if (neg && !roundUp)
+            r -= 1;
+    }
+    if (r > std::numeric_limits<XRPAmount::value_type>::max ())
+        Throw<std::overflow_error> ("XRP mulRatio overflow");
+    if (r < std::numeric_limits<XRPAmount::value_type>::min ())
```
seelabs: Strictly speaking, this is transaction breaking. However, I don't think it comes up in practice, and I do think the check should be here. I'd rather not have to put a switch here. Let me think about if we can get away without one (I think we can).

src/ripple/app/misc/impl/LoadFeeTrack.cpp
```diff
@@ -81,56 +82,110 @@ LoadFeeTrack::lowerLocalFee ()
...
+    return x;
 }
 
+namespace detail
+{
+
+struct xrp_unit_product_tag;
+
+using xrp_unit_product =
+    units::TaggedFee<detail::xrp_unit_product_tag, std::uint64_t>;
```
seelabs: I don't think I'd define this type. I see why it's done, tho. What do you think about defining a function `mulDiv(FeeUnit64, XRPAmount, FeeUnit64) -> XRPAmount` and don't use a tagged type inside that function?

src/ripple/ledger/ReadView.h
```diff
@@ -45,10 +46,10 @@ namespace ripple {
 */
 struct Fees
 {
-    std::uint64_t base = 0;         // Reference tx cost (drops)
-    std::uint32_t units = 0;        // Reference fee units
-    std::uint32_t reserve = 0;      // Reserve base (drops)
-    std::uint32_t increment = 0;    // Reserve increment (drops)
+    XRPAmount base{ 0 };         // Reference tx cost (drops)
+    FeeUnit32 units{ 0 };        // Reference fee units
```
seelabs: I know you didn't change this, but does a reference fee of zero make semantic sense?

src/ripple/app/misc/impl/TxQ.cpp
```diff
@@ -66,31 +66,33 @@ getLastLedgerSequence(STTx const& tx)
...
     auto const txBegin = view.txs.begin();
     auto const txEnd = view.txs.end();
     auto const size = std::distance(txBegin, txEnd);
     feeLevels.reserve(size);
     std::for_each(txBegin, txEnd,
         [&](auto const& tx)
         {
-            auto const baseFee = calculateBaseFee(view, *tx.first);
+            auto const baseFee = view.fees().toDrops(
+                calculateBaseFee(view, *tx.first)).second;
```
seelabs: This looks like a change in behavior, is that intentional?

src/ripple/app/misc/impl/TxQ.cpp
```diff
@@ -1492,27 +1497,19 @@ TxQ::doRPC(Application& app) const
     levels[jss::median_level] = to_string(metrics.medFeeLevel);
     levels[jss::open_ledger_level] = to_string(metrics.openLedgerFeeLevel);
 
-    auto const baseFee = view->fees().base;
+    auto const& baseFee = view->fees().base;
```
seelabs: Nit: XRPAmount is small, taking it by value is fine (no change needed).

src/ripple/basics/feeunits.h
```diff
@@ -0,0 +1,540 @@
...
+    fee () const
+    {
+        return fee_;
+    }
+
+    Json::Value
+    json () const
+    {
+        static_assert(is_usable_unit_v<TaggedFee>, "Invalid unit type");
+        using jsontype = std::conditional_t<std::is_signed_v<value_type>,
```
seelabs: This won't work for floating point types. Should we restrict this class to integer types? (I do propose getting rid of the one use of `double` for this class in another comment).

src/ripple/basics/feeunits.h
```diff
@@ -0,0 +1,540 @@
...
+#include <ios>
+#include <sstream>
+
+namespace ripple {
+
+namespace units {
+
+struct feeunit_tag;
+struct feelevel_tag;
+struct unitless_tag;
```
seelabs: I'd add a quick comment explaining for what each tag represents

src/ripple/protocol/impl/STObject.cpp
```diff
@@ -603,11 +603,24 @@ void STObject::setFieldU32 (SField const& field, std::uint32_t v)
...
+}
+
 void STObject::setFieldU64 (SField const& field, std::uint64_t v)
 {
     setFieldUsingSetValue <STUInt64> (field, v);
 }
 
+void STObject::setFieldU64 (SField const& field, XRPAmount v)
+{
+    assert(v > beast::zero);
```
seelabs: I'm glad there's an assert here, but I'm thinking this might be a really bad error to let slip. Do we want to crash here instead?

src/ripple/app/misc/impl/LoadFeeTrack.cpp
```diff
@@ -150,43 +205,43 @@ scaleFeeLoad(std::uint64_t fee, LoadFeeTrack const& feeTrack,
...
         Throw<std::overflow_error> ("scaleFeeLoad");
     baseFee *= feeFactor;
     // Reorder fee and baseFee
-    if (fee < baseFee)
-        std::swap(fee, baseFee);
-    // If fee * baseFee / den might overflow...
-    if (fee > max / baseFee)
+    maybe_swap(fee, baseFee);
+    // If fee * baseFee overflows, do the division first
+    if (fee > FeeUnit64{ max / baseFee })
```
jwbusch: Do you need this `FeeUnit64` ?

src/ripple/basics/feeunits.h
```diff
@@ -0,0 +1,540 @@
...
+    operator%= (value_type const& rhs)
+    {
+        fee_ %= rhs;
+        return *this;
+    }
+
+    TaggedFee
+    operator- () const
+    {
+        static_assert( std::is_unsigned_v<T>,
```
jwbusch: Should this be `is_signed_v<value_type>` -> assert that it is signed?

src/ripple/beast/cxx17/type_traits.h
```diff
@@ -35,13 +35,76 @@ namespace std {
 
     #endif // ! __cpp_lib_void_t
 
+    #if ! __cpp_lib_common_type_t
```
HowardHinnant: This breaks the macOS build as `__cpp_lib_common_type_t` doesn't exist in libc++.  So we end up getting two definitions of `common_type_t`.  Furthermore no feature-test macros are mandated in C++14 or C++17.  And `__cpp_lib_common_type_t` is not mandated in C++20.

Will gcc compile without this addition?

src/ripple/basics/feeunits.h
```diff
@@ -0,0 +1,540 @@
...
+    ANY  SPECIAL ,  DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+    WHATSOEVER  RESULTING  FROM  LOSS  OF USE, DATA OR PROFITS, WHETHER IN AN
+    ACTION  OF  CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+    OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+
+#ifndef BASICS_FEES_H_INCLUDED
+#define BASICS_FEES_H_INCLUDED
+
+#include <ripple/basics/BasicConfig.h>
```
mellery451: you have added this header file to the `xrpl_core` files, which means all of its includes are now part of the core lib interface. Pretty sure we *don't* install `BasicConfig.h` currently in the lib, but do we actually need it here anyhow? If we do, then we need to add it to the xrpl_core installed headers and probably need to move it's TU into the lib as well.

src/ripple/app/misc/FeeVoteImpl.cpp
```diff
@@ -28,14 +28,13 @@ namespace ripple {
 
 namespace detail {
 
-template <typename Integer>
 class VotableInteger
 {
 private:
-    using map_type = std::map <Integer, int>;
+    using Integer = XRPAmount;
```
mellery451: Is there a particular reason you stuck with the `Integer` type alias? Now that the class is no longer generic, it seems like it's specifically a class about voting on `XRPAmounts`.

src/ripple/beast/cxx17/type_traits.h
mellery451: can confirm - getting compile errors related to `common_type_t` on macos.

src/ripple/basics/XRPAmount.h
```diff
@@ -0,0 +1,298 @@
+//------------------------------------------------------------------------------
```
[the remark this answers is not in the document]
ximinez: Done.

src/ripple/app/misc/FeeVote.h
```diff
@@ -40,16 +40,16 @@ class FeeVote
     struct Setup
     {
         /** The cost of a reference transaction in drops. */
-        std::uint64_t reference_fee = 10;
+        XRPAmount reference_fee{ 10 };
```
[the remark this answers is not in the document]
ximinez: I'm pretty sure that's how VS auto-formats it. I almost want to change it back to `= 10`, but I'll leave it as-is.

src/ripple/app/misc/NetworkOPs.cpp
```diff
@@ -2288,14 +2283,15 @@ Json::Value NetworkOPsImp::getServerInfo (bool human, bool admin, bool counters)
         }
         if (loadFactorFeeEscalation !=
                 escalationMetrics.referenceFeeLevel &&
-                    (admin || loadFactorFeeEscalation != loadFactor))
+                (admin ||
+                loadFactorFeeEscalation != FeeLevel64{ loadFactor }))
```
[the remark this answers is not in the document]
ximinez: In the process of writing this comment, I realized that the straightforward cast was just plain _wrong_! It works because the base for server load and the base for fee levels happen to be the same, but (a) that's not guaranteed to be true forever, and (b) this is exactly the kind of problem that this whole PR is supposed to be solving. 

So thanks for pointing this out!

src/ripple/basics/feeunits.h
```diff
@@ -0,0 +1,540 @@
...
+#include <cassert>
+#include <cmath>
+#include <string>
+#include <iosfwd>
+#include <ios>
+#include <sstream>
+
+namespace ripple {
+
+namespace units {
```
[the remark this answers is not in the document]
ximinez: Renamed to `feeunit`

src/ripple/app/misc/NetworkOPs.cpp
ximinez: The reason I'm using `XRPAmount` as a tag is so that the type will be compatible with `XRPAmount` itself using as far as `TaggedFee` is concerned. The reason that works is that `XRPAmount` uses itself as its `unit_type`. https://github.com/ripple/rippled/pull/2910/files/3b93c0eb76d2559a2e33ad2bd3b5e672fad361e5#diff-715593421d046dd34694688c71cfb223R42

What do you think? Overly clever?

src/ripple/app/misc/impl/LoadFeeTrack.cpp
ximinez: My intention was to have the same interface as `std::abs` without writing a bunch of overloads. But I've fixed it now.

src/ripple/basics/XRPAmount.h
ximinez: The `scalar * XRPAmount` operator was defined further down, but I moved it to be a friend inside the class, so that's clearer. Division is provided by `boost::dividable`. (I would have used `boost::multiplicative`, but that doesn't give me `constexpr` multiplication.)

src/ripple/app/misc/NetworkOPs.cpp
seelabs: It's definitely clever (and I mean that as a compliment - I love "cute" solutions like that). But in this case, I don't think we can justify the clever code. Here's my thinking:

1) It's used to convert "drops" into "xrp" (millions of drops). There's no other calculation here, and not likely to be one. I'd probably just use a lambda for this. 

2) A new type that represents xrp (a fundamentally integral type) as a double at least raises my eyebrows a little.

3) We're using the `TaggedFee` type in a different way than we use it in other places (including the way the tag is used). There is joy is reusing a structure in a new way, but it's also easier to understand something if it's always used in roughly the same way.

Your call, of course. But I'd probably lose the `XRPDouble` class.

src/ripple/basics/XRPAmount.h
ximinez: I did a little test, and discovered that the `convert_to` function will convert an underflow to `min()`, so if I undo this change, the result will at least be a really really big negative number, as opposed to some random mapping of the underlying bits. It's also worth noting that all of these calculations deal with `XRPAmount`s, and any amount that does overflow or underflow will be dealing with more XRP than can ever exist, so it'll fail somewhere else in the process.

I can go either way.

src/ripple/app/misc/impl/LoadFeeTrack.cpp
ximinez: Duh. The templated unit-handling version of that function in feeunits.h already does this. So that's an easy update, and then I can remove `xrp_unit_product_tag` and functions.

src/ripple/ledger/ReadView.h
ximinez: I suspect the intention of initializing to 0 is just to prevent it from being uninitialized, and so that if it is used, there will be a very attention getting divide-by-0, instead of meaningless values. The values are set in `Ledger::setup` before it's ever used.

ximinez: I suspect the intention of initializing to 0 is just to prevent it from being uninitialized, and so that if it is used, there will be a very attention getting divide-by-0, instead of meaningless values. The values are set in `Ledger::setup` before it's ever used.

src/ripple/app/misc/impl/TxQ.cpp
ximinez: Yes, absolutely. This is one of the places that I identified as treating the value returned as a different unit than it was intended to be. In practice, right now, the numeric value is going to be equivalent, so there will be no behavior change, but if the base fee ever changes, that will be significant.

src/ripple/app/misc/impl/TxQ.cpp
ximinez: Fixed anyway

src/ripple/ledger/ReadView.h
seelabs: If something doesn't have a reasonable default state, I usually delete the default constructor. I can see how the two phase setup in Ledger complicates this. No change needed (it's beyond the scope of this PR to fix this, and I don't have an easy fix anyway). Marking this as resolved.

src/ripple/basics/XRPAmount.h
seelabs: I did miss the one lower down. Thanks for moving it up. Nit: Is there a reason that one is defined as a member function and one is defined as a friend? I'd define them both as friends (tho no change needed, it's a minor nit).

src/ripple/basics/XRPAmount.h
seelabs: I agree your change is an improvement. I do want to keep it. I'm just trying to decide if we need an amendment or not.

src/ripple/basics/XRPAmount.h
ximinez: There are two reasons one is a member function, and one is a friend:
1. That's how it was originally written, and I'm lazy.
2. Because I can't make them both member functions. Seriously, if I could, I would. I have a significant preference for member functions, but it's just a preference.

ximinez: I wrote a different reply first, but realized that I remembered incorrectly how I got there. The main reason I did one as a member and one as a free function, later a friend, is just that I have a preference for using member functions instead of free functions. I'm not going to change it, but I don't have any objection either.

src/ripple/basics/XRPAmount.h
ximinez: In my latest commit ("Review feedback #3"), I commented the change out, with a todo to determine if we need an amendment, and, more significantly, updated the unit tests to verify and demonstrate the behavior. The relevant unit tests had several holes and incorrectly documented test cases.

src/ripple/app/misc/NetworkOPs.cpp
seelabs: I like your `decimalXRP()` solution for this. Thanks for doing that.

src/ripple/basics/XRPAmount.h
seelabs: I'm fine with it. I'm marking this resolved.

src/ripple/basics/feeunits.h
ximinez: Yes, I think it makes sense to restrict this function to integer types. But I figured out how to overload depending on the `value_type`, so it'll work for both the integer and floating types. I'm never 100% confident with template metaprogramming, so I'm all ears if there's a better way to implement it.

seelabs: When 1.4 b1lands we'll get `if constexpr`. I'd probably use that over `enable_if` here. I'm also not sure why we need the `is_usable_unit` check.

ximinez: Oh yeah! That'll be much nicer!

> I'm also not sure why we need the `is_usable_unit` check.

Future-proofing. I should document that. Right now, all defined unit tags are valid, but if this class is used in a more general sense later (and I can imagine it being useful for anything that needs a compile-time unit), then I don't want new units to be convertible to JSON by default. There was an example with that `xrp_unit_product_tag` tag before I got rid of it at your very helpful suggestion. They can be added to that value later if it's appropriate, but I want it to be intentional.

src/ripple/basics/feeunits.h
ximinez: Done

src/ripple/protocol/impl/STObject.cpp
ximinez: After some discussion, rather than range checking, we determined that it would likely be a lot safer to implement type safety. To that end, I templatized `XRPAmount`, leaving the "default" `XRPAmount` as `XRPAmountBase<std::int64_t>`, then rewrote the `setFieldU32` and `U64` to take `XRPAmountU32` and `U64`, respectively.

`XRPAmountBase` has an `explicit` ctor for other `XRPAmountBase`s with different template types, so they can be converted when necessary, but not accidentally.

I think this could work, yo.

src/ripple/app/misc/NetworkOPs.cpp
```diff
@@ -2310,33 +2306,35 @@ Json::Value NetworkOPsImp::getServerInfo (bool human, bool admin, bool counters)
...
         else
         {
-            l[jss::base_fee_xrp] = static_cast<double> (baseFee) /
-                    SYSTEM_CURRENCY_PARTS;
+            // Make a local specialization of the TaggedFee class, using
+            // XRPAmount as the "unit" to make some of the math here easier.
+            using DropsDouble = units::TaggedFee<XRPAmount, double>;
+
+            l[jss::base_fee_xrp] =
+                static_cast<DropsDouble>(baseFee) /
```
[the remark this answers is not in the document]
ximinez: I rewrote this as a member function: `XRPAmount::decimalXRP()`.

src/ripple/app/misc/impl/LoadFeeTrack.cpp
ximinez: This part was rewritten, so technically, no.

src/ripple/basics/feeunits.h
ximinez: WAT!  Nice catch! Fixed, and I'll add unit tests.

src/ripple/beast/cxx17/type_traits.h
ximinez: @HowardHinnant @mellery451 I've rebased the branch onto 1.4.0-b1, which gives us C++17, and thus got rid of this definition plus the use of it in LoadFeeTrack.cpp. Let me know if this is still a problem, please.

src/ripple/basics/feeunits.h
ximinez: I was able to remove this dependency with some other changes.

src/ripple/app/misc/FeeVoteImpl.cpp
ximinez: Mainly laziness. But it's back to being more generic again, so that's awesome.

src/ripple/protocol/impl/STObject.cpp
ximinez: Update: No more converting ctors for these classes. Instead, there's an `as<Dest>()` function that does range checking if appropriate and returns the requested type.

ximinez:
@mellery451 @seelabs @HowardHinnant @jwbusch  CI problems aside, I think I've addressed all of everybody's concerns. Please check out the changes and let me know if I've missed anything, or if you spot any new problems.

seelabs:
This is a long comment, but the TL;DR is I'd rather not template `XRPAmount`. I know how we got here, but when I look at the code needed to support the template versus the benefit, I think it's not worth it.

First the benefit. Templating `XRPAmount` allows the code in `STValidations` to set fields without worring about overflow or underflow:

```c++
    if (fees.baseFee)
        setFieldU64(sfBaseFee, *fees.baseFee);

    if (fees.reserveBase)
        setFieldU32(sfReserveBase, *fees.reserveBase);

    if (fees.reserveIncrement)
        setFieldU32(sfReserveIncrement, *fees.reserveIncrement);
```

The `baseFee`, `reserveBase`, and `reserveIncrement` are strongly typed so they will always be able to set their fields without error. Without a doubt, this is a win.

The cost of doing this is templating the `XRPAmount` class. This requires handling interactions between the different concrete `XRPAmountBase` classes: handing type-safe conversions and handling equality/comparison tests. It also requires templating `VotableInteger` and the `DropsPerXRP` constant.

None of that is killer, and this PR does a good job of supporting all of that. However, when I look at what the code this is supporting - the six lines of code in `STValidations` - and look at what that code would look like without the templated XRPAmount class - I don't think the template is worth it. I also don't think new code is likely to use these templated classes. The alternative I'd propose is add a templated `dropsAs` member function to XRPAmount that returns an seated optional if the conversion can succeed. Then remove the templates and the `setFieldXXX(XRPAmountXXX)` overloads in `STObject`. The Code in STValidations isn't as nice (it would look like this now), but the rest of the code is simpler.

```c++
    if (fees.baseFee)
    {
        if (auto const v = fees.baseFee->dropsAs<std::uint64_t>())
        {
            setFieldU64(sfBaseFee, *v);
        }
        else
        {
            Throw<std::runtime_error>("XRPAmount conversion out of range");
        }
    }

    if (fees.reserveBase)
    {
        if (auto const v = fees.reserveBase->dropsAs<std::uint32_t>())
        {
            setFieldU32(sfReserveBase, *v);
        }
        else
        {
            Throw<std::runtime_error>("XRPAmount conversion out of range");
        }
    }

    if (fees.reserveIncrement)
    {
        if (auto const v = fees.reserveIncrement->dropsAs<std::uint32_t>())
        {
            setFieldU32(sfReserveIncrement, *v);
        }
        else
        {
            Throw<std::runtime_error>("XRPAmount conversion out of range");
        }
    }
```

Here's a patch to look at to compare the two approaches: https://github.com/seelabs/rippled/commit/e418075d5f570eaaa35567f5a15ddf564f3032f2

I also recognize that this is a trade-off (and one we've discussed in DM). While I vote for removing the template, I'll also approve this if you decide to keep it.

src/ripple/basics/XRPAmount.h
```diff
@@ -0,0 +1,388 @@
...
+            return { safe_cast<desttype>(drops_) };
+        }
+        else
+        {
+            if ((drops_ > std::numeric_limits<desttype>::max()) ||
+                (!std::numeric_limits<desttype>::is_signed && drops_ < 0) ||
+                (std::numeric_limits<desttype>::is_signed &&
+                    drops_ < std::numeric_limits<desttype>::lowest()))
+            {
+                Throw<std::runtime_error>("XRPAmount conversion out of range");
```
seelabs: I'd return an optional rather than throwing.

src/ripple/beast/cxx17/type_traits.h
HowardHinnant: Looks good on macOS. 👍

src/ripple/basics/XRPAmount.h
ximinez: @seelabs You're rewrite is a great simplification. I pushed your changes to the branch, and updated the unit test you removed.

src/ripple/app/misc/FeeVote.h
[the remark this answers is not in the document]
nbougalis: Honestly, `= 10` is way better, IMO. No change needed however; we need to get with the times, I guess.

src/ripple/protocol/SystemParameters.h
```diff
@@ -37,23 +38,27 @@ systemName ()
...
+{
+    return amount <= INITIAL_XRP;
+}
 
-/** Number of drops in the genesis account. */
-static
-std::uint64_t const
-SYSTEM_CURRENCY_START = SYSTEM_CURRENCY_GIFT * SYSTEM_CURRENCY_USERS * SYSTEM_CURRENCY_PARTS;
+inline
+bool isLegalAmount (std::uint64_t amount)
```
nbougalis: I don't like this. It's a _dangerous_ API to use: it accepts a value in _drops_, making it really easy for someone to do the wrong thing. And it only gives us a marginal benefit in a couple of cases, in fee vote processing.

I'd prefer that, wherever this overload is called from, we just construct an instance of `XRPAmount` and pass that into the `isLegalAmount` variant above. That way it's very clear exactly what's happening and you can't accidentally pass in XRP where drops are expected, or vice-versa.

src/ripple/basics/XRPAmount.h
```diff
@@ -0,0 +1,342 @@
...
+
+} // feeunit
+
+class XRPAmount
+    : private boost::totally_ordered <XRPAmount>
+    , private boost::additive <XRPAmount>
+    , private boost::equality_comparable <XRPAmount, std::int64_t>
+    , private boost::additive <XRPAmount, std::int64_t>
+    , private boost::dividable <XRPAmount, std::int64_t>
+    , private boost::modable <XRPAmount, std::int64_t>
```
nbougalis: I have some concerns about implementing multiplication and _significant_ concerns about implementing division in the general purpose class that represents XRP.

src/ripple/app/misc/FeeVoteImpl.cpp
```diff
@@ -217,10 +229,22 @@ FeeVoteImpl::doVoting(
...
-                obj[sfReserveBase] = baseReserve;
-                obj[sfReserveIncrement] = incReserve;
-                obj[sfReferenceFeeUnits] = feeUnits;
+                // These throws are likely to terminate the app if any fire.
+                // They should not be possible to hit, though, because the voting
+                // process (FeeVoteImpl) ignores any out of range values.
+                if(auto const v = baseFee.dropsAs<std::uint64_t>())
+                    obj[sfBaseFee] = *v;
+                else
+                    Throw<std::runtime_error>("XRPAmount conversion out of range");
```
nbougalis: This is a new exception and I'm not sure that we ever catch this. Can you point to the catch handler @ximinez?

src/ripple/protocol/SystemParameters.h
ximinez: AFAIK, outside of unit test code, every place that works with XRP uses a value expressed in drops, so I don't know how risky this really is. The bigger problem is that in the few places this overload is used, it's *also* dangerous to construct an `XRPAmount` directly because an absurdly large value will overflow `int64_t`, and thus end up as a negative value, which would also be really bad.

I'm really happy that I was able to convert existing fees to use `XRPAmount`, but there are a couple of cases where the signed/unsigned conversions make it just a little awkward.

There are a couple of potential solutions:
1. Get rid of the overload, check _only_ for `int64_t` overflow in these couple of places, then construct an `XRPAmount` as you suggested.
2. Define an unsigned `XRPAmount`-like instance of the `TaggedFee` class. and use that as the overload parameter for these couple of cases.
3. Decide that using `XRPAmount` wasn't the best idea and use a `TaggedFee` instance everywhere for fees.

I suspect @seelabs will have opinions about some of these options.

src/ripple/app/misc/FeeVoteImpl.cpp
ximinez: There is no catch handler. The original intent was that if the values here are not valid, then something has gone horribly wrong, so the intention was to allow the throw to crash the program.

However, I've given your feedback some thought, and came up with two alternate approaches that I think address all of your concerns
1. https://github.com/ripple/rippled/commit/7ba8fbbf5a9ae22f9ef42ce54fd481f1b84b832a (Comparison to develop: https://github.com/ripple/rippled/compare/develop...ximinez:strong-xrpfee). This change makes a new class, `XRPFee`, which is *derived* from `XRPAmount`, and adds the functionality that isn't appropriate for `XRPAmount` including: division, `dropsAs`. `decimalXRP`, `jsonClipped`, and the `>>` operator.  This allows us to continue to use `XRPAmount` as _the_ class to represent XRP / drops, while still having functionality to represent fees when we need to. It also gets rid of these conversion throws by allowing `dropsAs` to return a default value if the stored value is out of range, and uses the fee values from the LCL as those defaults.
2. https://github.com/ripple/rippled/commit/ce568a5dfc405baeddb552508d49424d0fa25805 (Comparison to develop: https://github.com/ripple/rippled/compare/develop...ximinez:strong-template). This change basically puts `XRPAmount` back in protocol (with a few improvements), and creates `XRPFee` as an instance of `TaggedFee`, and then converts values as necessary. This approach has already been rejected once thanks to @seelabs, but it does also get rid of the throws because the fees are always the right type throughout the voting process. I also don't like this approach, because there are so many conversions between the two types that I worry about one of them being done incorrectly.

So, @nbougalis, @seelabs, thoughts?

seelabs: @ximinez What do you think about removing the division, mod, increment, and decrement functions from XRPAmount? Here's a patch that does so: https://github.com/seelabs/rippled/commit/6cf2c3802ed3bef6f75648750212bf3db2a0812e

The only place that loses some of the type safety you added was the lambda in `scaleFeeLoad`, but I think that's OK, and on balance I think that's worth it to remove those operators (which I agree with @nbougalis, I'd rather they weren't part of the XRPAmount class).

I did a patch that got rid of the multiplication operator, but I think that operator is worth keeping. We only allow multiplication by a scalar, and it's useful for calculating reserves (multiply by owner count), scaling fees, as converting "whole xrps" to drops (tho that could easily be a function if we wanted).

See what you think of this patch, and if @nbougalis is OK with keeping multiplication, I'd rather do something like this than derive a subclass from `XRPAmount`.

Note: this does not solve the issues with throwing, but does solve the operators problem.

src/ripple/basics/XRPAmount.h
```diff
@@ -0,0 +1,318 @@
...
+    decimalXRP () const;
+
+    template <class Dest>
+    boost::optional<Dest>
+    dropsAs() const
+    {
+        if ((drops_ > std::numeric_limits<Dest>::max()) ||
+            (!std::numeric_limits<Dest>::is_signed && drops_ < 0) ||
+            (std::numeric_limits<Dest>::is_signed &&
+                drops_ < std::numeric_limits<Dest>::lowest()))
```
seelabs: Fine as-is. However, we've been moving toward indenting as the `.clang_format` in the project indents. That current does not add the extra space.

ximinez:
@mellery451 @jwbusch @nbougalis @seelabs
Ok, after going over this one more time, I've made one more commit to tweak a few things that I caught. I think this is finally done. Can I get a final sign-off? (Especially from @nbougalis)

src/ripple/app/misc/impl/LoadFeeTrack.cpp
```diff
@@ -96,21 +96,17 @@ scaleFeeLoad(FeeUnit64 fee, LoadFeeTrack const& feeTrack,
     // compatible. This function is an exception.
     auto lowestTerms = [](auto& a, auto& b)
     {
-        auto value = [](auto val)
+        auto const value = [](auto val)
```
seelabs: I'd remove the `const` from the lambdas. Declaring a lambda `const` doesn't add any restrictions (the lambda's `operator()` is already const`).

src/ripple/app/misc/impl/LoadFeeTrack.cpp
```diff
@@ -96,21 +96,17 @@ scaleFeeLoad(FeeUnit64 fee, LoadFeeTrack const& feeTrack,
...
         {
             if constexpr(std::is_arithmetic_v<decltype(val)>)
                 return val;
             else
                 return val.value();
         };
 
-        auto divide = [](auto& val, auto gcd)
+        auto const divide = [&value](auto& val, auto gcd)
```
seelabs: I'd remove this lambda. At this point the lambda is so simple that it's better to just replace the call to the lambda with this code directly.

src/ripple/app/misc/FeeVoteImpl.cpp
```diff
@@ -171,7 +171,7 @@ FeeVoteImpl::doVoting(
             if (val->isFieldPresent (sfBaseFee))
             {
                 auto const vote = val->getFieldU64 (sfBaseFee);
-                if (isLegalAmount(vote))
+                if (vote <= std::numeric_limits<XRPAmount::value_type>::max())
```
seelabs: I _think_ I prefer the lower amount restriction in `isLegalAmount` to the larger amount in `value_type`. Is there a motivation for changing this that I'm missing?

src/ripple/app/misc/impl/LoadFeeTrack.cpp
ximinez: That's fair. I was thinking that `const` didn't hurt, particularly when I captured it in the second lambda, but it makes sense to remove it, especially if I remove the second lambda.

src/ripple/app/misc/impl/LoadFeeTrack.cpp
ximinez: Done.

src/ripple/app/misc/FeeVoteImpl.cpp
ximinez: Yes. `vote` is a uint64. I got rid of the `isLegalAmount` override that explicitly takes a `uint64`, so the only remaining definition takes an `XRPAmount`. `XRPAmount` has an underlying int64 value is implicitly constructed from any integral value. If I call `isLegalAmount(vote)`, and vote is larger than the int64 limit, it will map to a negative value, and thus `isLegalAmount` will return true, which is a logic error.

I'll change it to check both.

jwbusch:
Looked again, latest looks good 👍

ximinez:
Squashed and rebased on to 1.4.0-rc1.

ximinez:
Squashed and rebased on to 1.5.0-b1 to resolve conflicts.

PR #2922, opened 2019-04-30
Add locking to two PeerImp member variables

scottschurr:
It appears that two of `PeerImp`'s member variables (`closedLedgerHash_` and `last_status_`) are not synchronized while being written in one thread (`addMessage()`) and being read in a different thread (`json()`).

This pull request leverages a pre-existing mutex (`recentLock_`) to add the appropriate locking.

### DIFF
--- a/src/ripple/overlay/impl/PeerImp.cpp
+++ b/src/ripple/overlay/impl/PeerImp.cpp
@@ -88,14 +88,25 @@ PeerImp::PeerImp (Application& app, id_t id, endpoint_type remote_endpoint,
 
 PeerImp::~PeerImp ()
 {
-    if (cluster())
-    {
-        JLOG(journal_.warn()) << name_ << " left cluster";
-    }
+    const bool inCluster {cluster()};
+
     if (state_ == State::active)
         overlay_.onPeerDeactivate(id_);
     overlay_.peerFinder().on_closed (slot_);
     overlay_.remove (slot_);
+
+    if (inCluster)
+    {
+        JLOG(journal_.warn()) << getName() << " left cluster";
+    }
+}
+
+// Helper function to check for valid uint256 values in protobuf buffers
+static
+bool
+stringIsUint256Sized (std::string const& pBuffStr)
+{
+    return pBuffStr.size() == uint256::size();
 }
 
 void
@@ -113,17 +124,21 @@ PeerImp::run()
         assert (state_ == State::active);
         // XXX Set timer: connection is in grace period to be useful.
         // XXX Set timer: connection idle (idle may vary depending on connection type.)
-        if ((hello_.has_ledgerclosed ()) && (
-            hello_.ledgerclosed ().size () == (256 / 8)))
+        if (hello_.has_ledgerclosed() && 
+            stringIsUint256Sized (hello_.ledgerclosed()))
         {
-            memcpy (closedLedgerHash_.begin (),
-                hello_.ledgerclosed ().data (), 256 / 8);
-            if ((hello_.has_ledgerprevious ()) &&
-                (hello_.ledgerprevious ().size () == (256 / 8)))
+            // Operations on closedLedgerHash_ and previousLedgerHash_ must be
+            // guarded by recentLock_.
+            std::lock_guard<std::mutex> sl(recentLock_);
+
+            closedLedgerHash_ = hello_.ledgerclosed();
+
+            if (hello_.has_ledgerprevious() &&
+                stringIsUint256Sized (hello_.ledgerprevious()))
             {
-                memcpy (previousLedgerHash_.begin (),
-                    hello_.ledgerprevious ().data (), 256 / 8);
-                addLedger (previousLedgerHash_);
+                previousLedgerHash_ = hello_.ledgerprevious();
+
+                addLedger (previousLedgerHash_, sl);
             }
             else
             {
@@ -190,10 +205,12 @@ PeerImp::send (Message::pointer const& m)
         // a small senq periodically
         large_sendq_ = 0;
     }
-    else if ((sendq_size % Tuning::sendQueueLogFreq) == 0)
+    else if (journal_.active (beast::severities::kDebug) &&
+        (sendq_size % Tuning::sendQueueLogFreq) == 0)
     {
+        std::string const name {getName()};
         JLOG (journal_.debug()) <<
-            (name_.empty() ? remote_address_.to_string() : name_) <<
+            (name.empty() ? remote_address_.to_string() : name) <<
                 " sendq: " << sendq_size;
     }
 
@@ -262,8 +279,9 @@ PeerImp::json()
     {
         ret[jss::cluster] = true;
 
-        if (!name_.empty ())
-            ret[jss::name] = name_;
+        std::string name {getName()};
+        if (!name.empty ())
+            ret[jss::name] = std::move(name);
     }
 
     ret[jss::load] = usage_.balance ();
@@ -295,9 +313,6 @@ PeerImp::json()
         ret[jss::complete_ledgers] = std::to_string(minSeq) +
             " - " + std::to_string(maxSeq);
 
-    if (closedLedgerHash_ != beast::zero)
-        ret[jss::ledger] = to_string (closedLedgerHash_);
-
     switch (sanity_.load ())
     {
         case Sanity::insane:
@@ -313,9 +328,20 @@ PeerImp::json()
             break;
     }
 
-    if (last_status_.has_newstatus ())
+    uint256 closedLedgerHash;
+    protocol::TMStatusChange last_status;
     {
-        switch (last_status_.newstatus ())
+        std::lock_guard<std::mutex> sl(recentLock_);
+        closedLedgerHash = closedLedgerHash_;
+        last_status = last_status_;
+    }
+
+    if (closedLedgerHash != beast::zero)
+        ret[jss::ledger] = to_string (closedLedgerHash);
+
+    if (last_status.has_newstatus ())
+    {
+        switch (last_status.newstatus ())
         {
         case protocol::nsCONNECTING:
             ret[jss::status] = "connecting";
@@ -338,9 +364,8 @@ PeerImp::json()
             break;
 
         default:
-            // FIXME: do we really want this?
             JLOG(p_journal_.warn()) <<
-                "Unknown status: " << last_status_.newstatus ();
+                "Unknown status: " << last_status.newstatus ();
         }
     }
 
@@ -397,6 +422,9 @@ PeerImp::hasTxSet (uint256 const& hash) const
 void
 PeerImp::cycleStatus ()
 {
+    // Operations on closedLedgerHash_ and previousLedgerHash_ must be
+    // guarded by recentLock_.
+    std::lock_guard<std::mutex> sl(recentLock_);
     previousLedgerHash_ = closedLedgerHash_;
     closedLedgerHash_.zero ();
 }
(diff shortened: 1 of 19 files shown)


src/ripple/overlay/impl/PeerImp.cpp
```diff
@@ -116,8 +116,11 @@ PeerImp::run()
         if ((hello_.has_ledgerclosed ()) && (
             hello_.ledgerclosed ().size () == (256 / 8)))
         {
-            memcpy (closedLedgerHash_.begin (),
-                hello_.ledgerclosed ().data (), 256 / 8);
+            {
+                std::lock_guard<std::mutex> sl(recentLock_);
+                memcpy (closedLedgerHash_.begin (),
+                    hello_.ledgerclosed ().data (), 256 / 8);
```
JoeLoser: I know this isn't anything new you added, but while in this neighborhood of code, do you mind giving a good identifier to the expression `256/8`? It is now used four times within this method and seventeen times within this file.

miguelportilla:
The changes look good. You may want to also consider similarly protecting `previousLedgerHash_`.

`lastPingSeq_`, `lastPingTime_`, and `name_` also seem susceptible.

scottschurr:
Thanks @miguelportilla, I'll give those member variables a good hard look and submit a [FOLD] commit.

src/ripple/overlay/impl/PeerImp.cpp
nbougalis: `closedLedgerHash_` is a `uint256` and `hello_.ledgerclosed()` is a `std::string` that contains the raw bytes for the `uint256`. The `256 / 8` is the size, in bytes, of a 256-bit variable.

Silly, I know. It's poor form and bad style in code that hasn't been rewritten cleanly.
 
Personally, I'd prefer the `if` on line 118 to use:

`m->ledgerhash ().size () == closedLedgerHash_.size()`

We could add a second overload of `copyFrom` that would do the `memcpy` (and the size check) allowing us to change this code to:

`closedLedgerHash_.copyFrom(m->ledgerhash().begin(), m->ledgerhash().end());`

Alternatively, we could just avoid adding a function and using code that's simply more readable:

`std::memcpy(closedLedgerHash_.begin(), hello_.ledgerclosed().data(), closedLedgerHash_.size());`

or even:

`std::copy(m->ledgerhash().begin(), m->ledgerhash().end(), closedLedgerHash_.begin());`

scottschurr: @nbougalis, I like your suggestion of replacing the magic numbers with `closedLedgerHash_.size()`.

I'm inclined to take a slightly different tack and use `sizeof (closedLedgerHash_)`.  To me that has the tiny advantage of clearly (to the reader) being a compile-time constant.  `base_uint::size()` is `constexpr`, so it is probably handled as a compile-time constant, at least by Clang.  But, as a reader, you have to go the the declaration of `base_uint::size()` to find that out.

Regardless, I think the outcome is comparable.  Replace a magic number with a value that is clearly associated with the correct value/type.

scottschurr:
Sorry for the delay, reviewers.  I looked at the member variables you pointed out, @miguelportilla, and I agree; they should be protected too.  So I've added several commits to this pull request.  I find thread locking can be hard to think about, so I tried to keep each commit focused on a single task.  This helped me to isolate each problem.  It may make it easier to review by going through individual commits as well.  But that's your call.

Thanks for your eyes and comments.

src/ripple/overlay/impl/PeerImp.cpp
nbougalis: @scottschurr, I like that idea but it's dangerous. We need to be _very_ careful with `sizeof(variable)` in this context because that hinges on assumptions that, while true today, may not be true tomorrow. In this case, it assumes that the size of a `base_uint<N>` is the same as `N / 8`.; that's not guaranteed.

Of course, this is something we rely on anyways, so it's not a stretch, but you can see how it could lead to trouble if someone refactors the code. It may be a good idea to add a `static_assert(std::is_pod(base_uint<...>), "Ack, base_uint needs to be POD")` or `std::is_trivial` or what-have-you, so that if things change out from under us, at least, we'll know right away instead of continuing oblivious to the change.

scottschurr: @nbougalis, good point.  Yes, I'll add that `static_assert`.

scottschurr:
`static_assert`s added and a few more magic numbers removed...

seelabs:
@scottschurr @miguelportilla 

I looked at `name_`, needing and mutex, and I don't believe it needs one. I did
not look at the others where a new lock was added. 

`name_` is initialized once, in `PeerImp::doAccept`. Afterward it is read in:

1) PeerImp::send
2) PeerImp::json
3) PeerImp::fail
4) PeerImp::onMessage
5) PeerImp::destructor

doAccept is called once per PeerImp, after the constructor, from `PeerImp::run`.
In effect, `PeerImp` is doing mutli-phase initialization, with `PeerImp::run`
being the second phase after the constructor. It is always run on
`PeerImp::strand_`. `doAccept` adds the peer to to the overlays active peers.

If we can guarantee that `doAccept` is run before any of the other function read
name, we don't need a mutex.

1) PeerImp::send runs on the same strand as `doAccept`. So this will be run after `doAccept`.
2) PeerImp::json is run from OverlayImpl::json, which iterates through the
   active peers. A peer isn't an active peer until it is added to the
   `OverlayImpl::ids_` collection in `PeerImpl::doAccept`. So this cannot run
   until after `PeerImpl::doAccept`.
3) PeerImp::fail runs on the `PeerImp::strand_`, This will run after `doAccept`.
4) PeerImp::onMessage is called from (eventually) `PeerImp::onReadMessage`,
   which is always called from a function that uses `PeerImp::strand_`
5) The destructor. It's possible this one is problematic as-is. However, if we
   move the log message to the end of the destructor I think we're safe.

Given that, I don't think `name_` needs a new mutex. Did I miss something (it's very possible, I took a relatively quick look)?

miguelportilla:
@seelabs Playing the Devil's advocate, my thoughts are that a weak argument could be made about a future use where name_ is read via something like ::json before its initialization. Although that would that hardly cause a real problem in the case of name_ and seems like improper use, the code, unfortunately, lacks documentation making the user aware. I generally prefer a safer implementation, documentation or at the very least an assertion. A similar situation applies to previousLedgerHash_, where future use might improperly call something like cycleStatus while the peer is receiving a status change message. Publicly exposing unprotected members (not deviantly!) during an undocumented and complex multi-phase initialization can be concerning.

scottschurr:
Just to keep everyone up to date on where this pull request is at...

1. I am personally aligned with @miguelportilla's concerns about future uses of a type that does multiphase initialization.  Since we have the locking figured out I'm inclined to leave it in place even though some of it is not strictly required at the moment.  It is a form of future proofing PeerImp.  If there are objections to this plan let me know.

2. @HowardHinnant just submitted a pull request (https://github.com/ripple/rippled/pull/2951) that significantly improves `base_uint` and conflicts with some of the changes here.  The PeerImp changes in this pull request can be significantly improved by being rebased on top of @HowardHinnant's work.  So my intent is to wait until https://github.com/ripple/rippled/pull/2951 is marked passed.  Then I'll rebase this pull request on top of that.

Once that rebase is done I'll ping my reviewers to request an additional once over.  I won't mark the pull request passed until the final reviews are complete.  Thanks for your patience.

seelabs:
@scottschurr 

How about something like this:

1) Keep the unneeded locks. You audited the code and decided the new locks were needed. That's a **HUGE** warning that the existing protection scheme is a tangled mess.
2) In this PR, document which locks are needed and which are being added because the existing locking scheme was too hard to audit. 
3) In a separate PR (and not necessarily my you - I can look at it too), explore using separate types for each phase to get rid of the multi-phase initialization. If this look OK, get rid of the locks we are now adding for multi-phase initialization.
4) In a separate PR, try to make it clearer _in code_ when data is protected by strands rather than mutexes. I'm not sure how yet. Remove unneeded locks after this.

It's very possible once I look at (2) and (3) in depth it'll be worse than adding the unneeded locks in the first place, but I'd at least like to try to remove them.

scottschurr:
@seelabs, I think that sounds like a great plan.  And I couldn't agree more that using locks in the multiphase initialization is a mess.  I'm delighted to have someone make a run at untangling it.  I'll add a **[FOLD]** to this pull request that lists which of the members actually need protection now and which ones are being protected preemptively.

seelabs:
@scottschurr I took a look at the new comment regarding locking, and I like to make two changes:

1) I would not include what functions that read or modify variables in that table. It already doesn't include all the functions that modify the variables (for example, `run` modifies `closedLedgerHash_`). And even if we updated the table so it's accurate now, it's very unlikely to stay in sync with the code. And even if it somehow manages to stay in sync with the code, no one would ever trust it, and would audit the code anyway.

2) The table only says `name_` does not require locking. I believe none of `last_status_`, `lastPingSeq_`, `lastPingTime_`, `no_ping_` require locks. `name_` will not require a lock once the destructor is modified (and I'd like to modify the destructor in this PR so we access name after it is removed from the `overlay_`).

Setting aside `name_`, the reason the other variables don't require a lock is because they always run on the `PeerImp` strand. Take `no_ping_` as an example. It is modified in `onTimer` and `onMessage`. It's easy to see that `onTimer` is always run on the strand (see PeerImp::setTimer). `onMessage` is a little more involved. `onMessage` is:
```
* called from ProtocolMessage.h detail::invoke
** called from invokeProtcolMessage
*** called from PeerImp::onReadMessage
**** called from `self` in the PeerImp's strand
**** called from doProtocolStart
***** called from run, which always runs on the peer imps strand
***** called from onWriteResponse
****** called from self, on the peer imp strand
****** called from doAccept
******* called from run, which always runs on peer's strand
```

We can do a similar analysis for the others to show they don't need a lock (of course, you and the other reviewers should double check this).

seelabs:
I'm not an official reviewer on this, but :+1: from me after 685380229fff93f899dc8c07bfb2ebd62354be04, and thanks!

scottschurr:
@seelabs, anyone who offers an opinion is an official reviewer :smile:.  Thanks for stepping in.

scottschurr:
This branch is now rebased atop https://github.com/ripple/rippled/pull/2951 which is passed and includes significant improvements to `base_uint`.  The rebase was pretty straight forward.

The (new) top-most commit on this branch applies the changes in `base_uint` to the `PeerImp` code.  In particular, it replaced all explicit calls to `memcpy` with assignment or constructor calls.  It also allowed removal of many `static_assert`s.  So the top-most commit deserves significant scrutiny.  It would be great if I could enlist @HowardHinnant to help with that review.

I think we may be close to done with this pull request.  Thanks for the help everyone.

src/ripple/overlay/impl/PeerImp.cpp
```diff
@@ -101,6 +101,14 @@ PeerImp::~PeerImp ()
     }
 }
 
+// Helper function to check for valid uint256 values in protobuf buffers
+static
+bool
+stringIsUint256Sized (std::string const& pBuffStr)
+{
+    return pBuffStr.size() == sizeof (uint256);
```
HowardHinnant: Tiniest nitpick:  Consider `return pBuffStr.size() == uint256::size();`

This is guaranteed to be the same functionality, but seems like a little bit higher level abstraction to me.  But nothing I'll loose sleep over.

scottschurr:
@miguelportilla, at your convenience it would be great if we could get your eyes on the last two commits of this pull request.  If you're okay with those changes then I'll squash the branch and mark it passed.  Thanks.

scottschurr:
Sorry.  My error.  @miguelportilla is on PTO for the rest of the week.  Since the last two commits were reviewed and okayed by @thejohnfreeman and @HowardHinnant, I'm inclined to mark this pull request as passed.  But I'll wait the rest of the day in case someone wants to add a review or object.  If I don't hear anything additional by Tuesday Jun 11th I'll squash and mark the pull request as passed.

PR #2927, opened 2019-05-06
Get names of transactions and ledger types from jss

scottschurr:
The code base, on the whole, is pretty good about using symbolic names, rather than local string literals, for strings that are visible through the API.  Sharing one symbol for a given name has the advantage that misspelling that symbol will cause a compile error, and all users of the symbol are guaranteed to use identical spellings.

One place where we haven't been following the symbolic name pattern is has been the names of transactions and the names of ledger objects.  This pull request corrects that oversight.

The pull request has many small commits.  The goal was to make the code a little easier for reviewers to audit.  Only a few names are transitioned from string literal to symbolic name in each commit.  By doing the review one commit at a time a reviewer will have fewer changing names to track.  But feel free to do the review any way you like.

Once the review is complete all of the commits will be squashed into a single commit.

### DIFF
--- a/src/test/rpc/LedgerData_test.cpp
+++ b/src/test/rpc/LedgerData_test.cpp
@@ -299,7 +299,7 @@ public:
 
         {
             Json::Value jv;
-            jv[jss::TransactionType] = "EscrowCreate";
+            jv[jss::TransactionType] = jss::EscrowCreate;
             jv[jss::Flags] = tfUniversal;
             jv[jss::Account] = Account{"bob5"}.human();
             jv[jss::Destination] = Account{"bob6"}.human();
@@ -312,7 +312,7 @@ public:
 
         {
             Json::Value jv;
-            jv[jss::TransactionType] = "PaymentChannelCreate";
+            jv[jss::TransactionType] = jss::PaymentChannelCreate;
             jv[jss::Flags] = tfUniversal;
             jv[jss::Account] = Account{"bob6"}.human ();
             jv[jss::Destination] = Account{"bob7"}.human ();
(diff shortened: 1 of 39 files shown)


src/test/rpc/LedgerData_test.cpp
```diff
@@ -344,84 +354,91 @@ class LedgerData_test : public beast::unit_test::suite
         auto const jrr = makeRequest(jss::account);
         BEAST_EXPECT( checkArraySize(jrr[jss::state], 12) );
         for (auto const& j : jrr[jss::state])
-            BEAST_EXPECT( j["LedgerEntryType"] == "AccountRoot" );
+            BEAST_EXPECT( j["LedgerEntryType"] == jss::AccountRoot );
```
jwbusch: should `"LedgerEntryType"` also be in `jss` ?

src/test/rpc/TransactionHistory_test.cpp
```diff
@@ -125,10 +125,10 @@ class TransactionHistory_test : public beast::unit_test::suite
             }
         }
         BEAST_EXPECT(total == 117);
-        BEAST_EXPECT(typeCounts["AccountSet"] == 20);
-        BEAST_EXPECT(typeCounts["TrustSet"] == 19);
-        BEAST_EXPECT(typeCounts["Payment"] == 58);
-        BEAST_EXPECT(typeCounts["OfferCreate"] == 20);
+        BEAST_EXPECT(typeCounts[jss::AccountSet.c_str()] == 20);
```
jwbusch: Seems a little strange indexing by c_str, Is it actually doing a string comparison or is it comparing addresses? i.e. a pointer to that jss literal?

If they're all in jss then pointer lookup should be fine, but wanted to double check.

src/test/rpc/LedgerData_test.cpp
scottschurr: `LedgerEntryType` is one of the serialized fields (defined in SField.cpp).  We should probably getting the string from `sfLedgerEntryType.fieldName`.

My eventual hope is that we get any strings that are in an `SField` from there.  All other strings should come from `jss`.  But I'm a long way from achieving that.  For now I'm limiting myself to transaction and ledger types.  This pull request is just a step along the path.

src/test/rpc/TransactionHistory_test.cpp
scottschurr: It's doing a string comparison.  The declaration of `typeCounts` is
```
std::unordered_map<std::string, unsigned> typeCounts;
```
So we're comparing the string hash to the hash of the null-terminated c-string.  Final comparison is the built-in comparison operator between a `std::string` and a `char const*`.  Actual comparison of the pointers would probably go badly.  And, yeah, it's probably silly that an `std::unordered_map` is being used there...

PR #2930, opened 2019-05-09
Payment paths with a zero output step are dry (RIPD-1749):

seelabs:
A tiny input amount to a payment step can cause this step to output zero. For
example, if a previous steps outputs a dust amount of 10^-80, and this step is a
IOU -> XRP offer, the offer may output zero drops. In this case, call the strand
dry. Before this patch, an error would be logged, the strand would be called
dry; in debug mode an assert triggered.

Note, this patch is not transaction breaking, as the caller did not user the ter
code. The caller only checked for success or failuer.

This patch addresses github issue issue reported here:
https://github.com/ripple/rippled/issues/2929

@HowardHinnant @jwbusch

### DIFF
--- a/src/ripple/app/paths/impl/StrandFlow.h
+++ b/src/ripple/app/paths/impl/StrandFlow.h
@@ -144,14 +144,21 @@ flow (
                         *sb, *afView, ofrsToRm, EitherAmount (*maxIn));
                     limitStepOut = r.second;
 
-                    if (strand[i]->isZero (r.second) ||
-                        get<TInAmt> (r.first) != *maxIn)
+                    if (strand[i]->isZero(r.second))
+                    {
+                        JLOG(j.trace()) << "First step found dry";
+                        return {tecPATH_DRY, std::move(ofrsToRm)};
+                    }
+                    if (get<TInAmt> (r.first) != *maxIn)
                     {
                         // Something is very wrong
                         // throwing out the sandbox can only increase liquidity
                         // yet the limiting is still limiting
-                        JLOG (j.fatal()) << "Re-executed limiting step failed";
-                        assert (0);
+                        JLOG(j.fatal())
+                            << "Re-executed limiting step failed. r.first: "
+                            << to_string(get<TInAmt>(r.first))
+                            << " maxIn: " << to_string(*maxIn);
+                        assert(0);
                         return {telFAILED_PROCESSING, std::move (ofrsToRm)};
                     }
                 }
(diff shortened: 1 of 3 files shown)


src/ripple/app/paths/impl/StrandFlow.h
```diff
@@ -144,14 +144,21 @@ flow (
...
                         // Something is very wrong
                         // throwing out the sandbox can only increase liquidity
                         // yet the limiting is still limiting
-                        JLOG (j.fatal()) << "Re-executed limiting step failed";
-                        assert (0);
+                        JLOG(j.fatal())
+                            << "Re-executed limiting step failed. r.first: "
+                            << to_string(get<TInAmt>(r.first))
+                            << " maxIn: " << to_string(*maxIn);
+                        assert(0);
```
HowardHinnant: I would like to see us develop a way to bring a run-time issue to our attention without terminating the process unless we believe that the process is corrupted to the point that it would be unsafe for it to continue.  I'm not sure what such a process would look like, but improved access to the logs of validators we control would be a step in the right direction.

seelabs: @HowardHinnant You are correct that the purpose of the assert is to bring the issue to our attention, and the process is safe to continue (if it were not I'd throw a logic error). 

I'm coming from the assumption that `asserts` are not used in production (validators that use our package are built with `NDEBUG` so the asserts won't fire). So long as that is true, the `asserts` are doing what I want. 

Are you worried that someone is going to turn off `NDEBUG` on the production machines, or do you object to bringing down the process even on dogfood machines? I'll try to think of a different way to do this, but I do think it's safe to merge with the `assert` as-is if we don't think of another way.

HowardHinnant: I'm concerned that the program with asserts enabled behaves differently than without.  It would be better if we could depend on the asserts that might be downstream of returning `telFAILED_PROCESSING`.

seelabs: One solution would be to remove the asserts and set a breakpoint there on dogfood machines. The downside to that is I'd want to make sure that everyone on the team ran their dogfood machines that way. I still want to keep the asserts, but I'd be OK with this solution too.

HowardHinnant: I'm not loosing sleep over this.  I've tested this patch and everything looks ok to me.  But I did want to start this conversation.  E.g. what if we want to debug the case that `telFAILED_PROCESSING` get returned and something unexpected happens afterward.  Or maybe we just want to assure that it is safe to keep the process up afterwards.

seelabs: I DM'd @HowardHinnant and I do see where he's coming from. However, we agreed that it's OK to keep the asserts for now (and we'll look for a better solution for this class of problems - maybe an external tool).

src/ripple/app/paths/impl/StrandFlow.h
```diff
@@ -144,14 +144,21 @@ flow (
...
+                    }
+                    if (get<TInAmt> (r.first) != *maxIn)
                     {
                         // Something is very wrong
                         // throwing out the sandbox can only increase liquidity
                         // yet the limiting is still limiting
-                        JLOG (j.fatal()) << "Re-executed limiting step failed";
-                        assert (0);
+                        JLOG(j.fatal())
+                            << "Re-executed limiting step failed. r.first: "
```
jwbusch: nit: my only comment is around these debug messages. `r.first` and `r.second` aren't particularly useful names in the log messages.
 
If I was debugging given this message I'd have to copy/paste a sub-string of it and search the codebase for where it's logged to get more context vs. getting more context from the log message itself.

Issue #2861, opened 2019-02-21
Append default port (51235) in [ips_fixed] when no port specified


alloyxrp:
If a port is not specified against an IP in `[ips_fixed]`, `rippled` fails to start. The daemon should try with the default port if it is not specified.

nbougalis:
Yep that’s a bug. Thanks for reporting it!

alloyxrp:
This fix should also apply to `[ips]`.

MarkusTeufelberger:
The example config currently says "A port may must be specified after adding a space to the address."

https://github.com/ripple/rippled/commit/b39c1eb92c5884c248baaaf2af62e88cd8681b83 sounds like this was intentionally set to "must", so this is a feature, not a bug.

nbougalis:
While the original intent may have been to require a port number, it doesn't mean this isn't a bug, albeit not a code bug. The decision to mandate a port when a port isn't mandatory in any other instance is, at best, curious. Our config is complex enough as is; having arcane differences here and there is not good.

I believe that this should be corrected.

movitto:
Figured this would be a good easy one to get my feet wet with rippled contributions! Just submitted a first stab, let me know if specific changes are needed/preferred.

PR #2931, opened 2019-05-11
Assign default port to [ips_fixed] and [ips] if unspecified

movitto:
Replaces previous behaviour where error was thrown.

Fixes issue #2861

### DIFF
--- a/src/ripple/overlay/impl/OverlayImpl.cpp
+++ b/src/ripple/overlay/impl/OverlayImpl.cpp
@@ -534,11 +534,12 @@ OverlayImpl::onPrepare()
             {
                 if (addr.port () == 0)
                 {
-                    Throw<std::runtime_error> ("Port not specified for "
-                        "address:" + addr.to_string ());
+                    ips.push_back (to_string (addr.at_port(51235)));
                 }
-
-                ips.push_back (to_string (addr));
+                else
+                {
+                    ips.push_back (to_string (addr));
+                }
             }
 
             std::string const base ("config: ");
--- a/src/ripple/overlay/impl/OverlayImpl.cpp
+++ b/src/ripple/overlay/impl/OverlayImpl.cpp
@@ -534,7 +534,7 @@ OverlayImpl::onPrepare()
             {
                 if (addr.port () == 0)
                 {
-                    ips.push_back (to_string (addr.at_port(51235)));
+                    ips.push_back (to_string (addr.at_port(2459)));
                 }
                 else
                 {
--- a/src/ripple/peerfinder/impl/Logic.h
+++ b/src/ripple/peerfinder/impl/Logic.h
@@ -188,12 +188,11 @@ public:
             return;
         }
 
-        for (auto const& remote_address : addresses)
+        for (auto remote_address : addresses)
         {
             if (remote_address.port () == 0)
             {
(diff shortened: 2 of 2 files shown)


src/ripple/overlay/impl/OverlayImpl.cpp
```diff
@@ -534,11 +534,12 @@ OverlayImpl::onPrepare()
             {
                 if (addr.port () == 0)
                 {
-                    Throw<std::runtime_error> ("Port not specified for "
-                        "address:" + addr.to_string ());
+                    ips.push_back (to_string (addr.at_port(51235)));
```
MarkusTeufelberger: Please don't make a port in the dynamic/private ports range the default.

movitto: 51235 is already the default port (see existing instances a few lines above). This patch is just to automatically set it on ips & fixed ips when not specified in the config.

MarkusTeufelberger: There currently is no default port in `rippled`, this is just the one these instances use.

You are suggesting to make a port from a port range that is supposed to be dynamic/private to be the default port of `rippled`.

movitto: OK I see what you're saying, though I would argue that 51235 is currently the default port in all but name, afterall it is the default port set in the config & used by most instances (that I know of).

What do you suggest as far as default port assignment, or are you saying that #2861  is a non-issue?

MarkusTeufelberger: I'd recommend just grabbing an unassigned port from the proper range and maybe in parallel also asking officially for an assigned port - it is actually not that difficult or impossible to get that: https://tools.ietf.org/html/rfc6335#page-16

nbougalis: I am supportive of registering a port. Give me a day, I have a plan :smiling_imp:

nbougalis: Update on the plan I hinted at above:

There already is a port ([2459](https://www.iana.org/assignments/service-names-port-numbers/service-names-port-numbers.xhtml?search=Schwartz#David_Schwartz)) that @JoelKatz has registered; we could repurpose it for this, and make it the new default for `rippled`. We'll follow-up with IANA about this.

@movitto, it may be a good idea to hold off on this PR so that we don't add an "implicit default" that people end up relying on only to then change it. Let's wait and see whether we'll go with 2459 or register something else.

movitto: > @movitto, it may be a good idea to hold off on this PR so that we don't add an "implicit default" that people end up relying on only to then change it. Let's wait and see whether we'll go with 2459 or register something else.

Sounds good @nbougalis, let us know how the IANA follow-up goes and I can update this PR from there.

movitto:
@nbougalis @MarkusTeufelberger @alloyxrp @mDuo13 updated the default port to reflect #3037. This patch just affects the port assigned to ips and ips_fixed entries if not specified, it does _not_ change the default ports listened on in rippled.cfg. After this any configurations w/ unspecified ports may result in connection errors if peers are still listening on the legacy port 51235

nbougalis:
This has been superseded by #3235.

PR #2932, opened 2019-05-11
Update some old comments

movitto:
Ledger index was removed from persistent nodestore a while back:
  https://github.com/ripple/rippled/commit/f946d7b447452d2b28a086696889034e2d894f2a#diff-06e0962214bab48948f689d95d5e6dc7

But these docs were never updated

### DIFF
--- a/src/ripple/nodestore/README.md
+++ b/src/ripple/nodestore/README.md
@@ -3,7 +3,7 @@
 ## Introduction
 
 A `NodeObject` is a simple object that the Ledger uses to store entries. It is 
-comprised of a type, a hash, a ledger index and a blob. It can be uniquely 
+comprised of a type, a hash and a blob. It can be uniquely
 identified by the hash, which is a 256 bit hash of the blob. The blob is a 
 variable length block of serialized data. The type identifies what the blob 
 contains. The fields are as follows: 
@@ -33,19 +33,14 @@ contains. The fields are as follows:
 
  A 256-bit hash of the blob.
 
-* `mLedgerIndex`
-      
- An unsigned integer that uniquely identifies the ledger in which this 
-   NodeObject appears.
-
 * `mData`
       
  A blob containing the payload. Stored in the following format.
  
 |Byte   |                     |                          |
 |:------|:--------------------|:-------------------------|
-|0...3  |ledger index         |32-bit big endian integer |
-|4...7  |reserved ledger index|32-bit big endian integer |
+|0...3  |unused               |32-bit big endian integer |
+|4...7  |unused               |32-bit big endian integer |
 |8      |type                 |NodeObjectType enumeration|
 |9...end|data                 |body of the object data   |
 ---    
--- a/src/ripple/nodestore/README.md
+++ b/src/ripple/nodestore/README.md
@@ -39,8 +39,7 @@ contains. The fields are as follows:
  
 |Byte   |                     |                          |
 |:------|:--------------------|:-------------------------|
-|0...3  |unused               |32-bit big endian integer |
-|4...7  |unused               |32-bit big endian integer |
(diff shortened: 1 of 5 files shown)


MarkusTeufelberger:
These are currently set to all zeros, but is that enforced/checked? Also having 2 32 bit "unused" integers in there is weird - why not a 64 bit "unused" integer or 8 "unused" null-bytes?

Maybe these should be removed alltogether anways... could save a bit of space, even with compression.

movitto:
> These are currently set to all zeros, but is that enforced/checked? Also having 2 32 bit "unused" integers in there is weird - why not a 64 bit "unused" integer or 8 "unused" null-bytes?
> 
> Maybe these should be removed alltogether anways... could save a bit of space, even with compression.

Fair point, though my intent with the preliminary patches is to be the least disruptive as possible, hence just changing the one field name from 'ledger_index' to 'unusued' as opposed to changing the types of / merging two fields.

movitto:
Splitting up #2952. Merging patch tidying comments with existing PR in similar domain

src/ripple/nodestore/README.md
```diff
@@ -33,19 +33,14 @@ contains. The fields are as follows:
...
-
 * `mData`
       
  A blob containing the payload. Stored in the following format.
  
 |Byte   |                     |                          |
 |:------|:--------------------|:-------------------------|
-|0...3  |ledger index         |32-bit big endian integer |
-|4...7  |reserved ledger index|32-bit big endian integer |
+|0...3  |unused               |32-bit big endian integer |
```
scottschurr: @nbougalis or @JoelKatz, do you have an opinion about whether marking these fields as unused is adequate?  It is possible that they are currently unused but, for example, must never be reused for any other purpose (with an appropriate explanation).  It's outside my area of expertise, but I thought I should ask the question.

src/ripple/nodestore/impl/DecodedBlob.cpp
```diff
@@ -31,8 +31,8 @@ DecodedBlob::DecodedBlob (void const* key, void const* value, int valueBytes)
 
         Bytes
 
-        0...3       LedgerIndex     32-bit big endian integer
-        4...7       Unused?         An unused copy of the LedgerIndex
+        0...3       Unused
```
scottschurr: Same question as noted in the nodestore README.md.

mtrippled:
Why not get rid of this line? https://github.com/ripple/rippled/pull/2932/files#diff-a15e5dba8ff474dc1749c03d1420ce1cR36
`//hotTRANSACTION = 2` is at least 4 years old

mtrippled:
The enum can be replaced:
https://github.com/ripple/rippled/blob/966f4bc7eb7061ddb0048e0db6dec13ef2e8a06e/src/ripple/nodestore/NodeObject.h#L55-L63
with 
```static constexpr std::size_t keyBytes = 32;```

File: src/ripple/nodestore/README.md [hunk 3]
```diff
@@ -33,19 +33,14 @@ contains. The fields are as follows:
...
+|4...7  |unused               |32-bit big endian integer |
```
mtrippled: 0...7 | unused |

Type can be left blank.

src/ripple/nodestore/README.md
mtrippled: I'm OK with documenting as is "unused" and leaving eternity to unfold in a possible different direction.

src/ripple/nodestore/impl/DecodedBlob.cpp
mtrippled: 0...7 Unused

movitto:
@mtrippled @scottschurr incorporated all feedback into the last patch and rebased / pushed.

PR #2934, opened 2019-05-14
Force snappy compression for RocksDB (remove option):

mellery451:
FIXES: https://github.com/ripple/rippled/issues/2860

 * Also remove RocksDBQuick backend which is non-functional.


cfg/rippled-example.cfg
```diff
@@ -818,13 +818,11 @@
 #       RocksDB is an alternative backend for systems that don't use solid-state
 #       drives.  Because RocksDB's performance degrades as it stores more data,
 #       keeping full history is not advised, and using online delete is
-#       recommended.  RocksDB is not available on Windows.
+#       recommended.
 #
 #       The RocksDB backend also provides these optional parameters:
```
MarkusTeufelberger:
```suggestion
#       The RocksDB backend provides no optional parameters
```

cfg/rippled-example.cfg
```diff
@@ -818,13 +818,11 @@
...
 #
-#       compression         0 for none, 1 for Snappy compression
-#
-#
+#             TBD
```
miguelportilla: This comment isn't needed.

cfg/rippled-example.cfg
mellery451: ...except that the rocks backend *does* have a number of optional configs but we have chosen (so far) to not document them because they are considered "expert only" 😬 .

mellery451: @MarkusTeufelberger sorry - I initially misunderstood your comment. This is confusing and I will remove the sentence entirely (I'll say nothing about optional config params for now). thanks.

MarkusTeufelberger: Yeah, that's probably even better, RocksDB was/is the odd one out regarding optional parameters anyways.

However, at the end of the file quite a lot of existing undocumented optional parameters for RocksDB are set and used...

mellery451: I agree..it's a little off-putting that we use some options here that we don't document. We might need to address that in the future (for now, all rocks specific options are considered "expert only")

mtrippled:
:+1:

PR #2935, opened 2019-05-15
Peer reservations

thejohnfreeman:
I want to get this PR started now, while the work is in progress, to get more eyes and some direction. I'm going to link some internal resources for now; once they are polished, I'd like to find some way to publish them publicly.

For context, see #2938 (and [JIRA 1712](https://ripplelabs.atlassian.net/browse/RIPD-1712) if you have access).

I've written what I've deduced to be the steps for adding a command to rippled on [our internal wiki](https://ripplelabs.atlassian.net/wiki/spaces/~189280334/pages/543031357/How+to+add+a+method+to+rippled). I've been following those steps below. Please comment there or here with any corrections or blind spots (especially blind spots; I don't know what I don't know).

I've left some questions in comments in the changes below. Please take a look and see if you can answer any. If you just want to skim for them, Ctrl-f for `?`.

I'm having trouble figuring out how I should think about and handle "public key for a node" as a string. If there are any docs on this, please point me to them.

Similarly, our RPC "framework" seems to be immune to function abstraction. I've taken a stab at it below, but I'm not happy with it. Here's the pattern I see in the rest of the code:

```cpp
Json::Value result;

auto tmpVar = tryToParse(someField);
if (notGood(tmpVar))
    return rpcError(rpcERROR_CONSTANT);

result[jss::some_field] = tmpVar;
return result;
```

The problem is the temporary variable combined with the early return for an error. If I try to put this in a function, I can't return both, so I end up taking an out parameter for the value. But I don't think I want to test for an empty `Json::Value` in the event of success, so I take that as an out parameter too and return a `bool`. It's all very gross to me. This kind of control flow is better suited by exceptions, in my opinion, or by an Error monad (do we have one available?). Thoughts?


MarkusTeufelberger:
> For context, see JIRA 1712.

:-(

src/ripple/app/main/Application.cpp
```diff
@@ -52,6 +52,7 @@
 #include <ripple/json/json_reader.h>
 #include <ripple/nodestore/DummyScheduler.h>
 #include <ripple/overlay/Cluster.h>
+#include <ripple/overlay/PeerReservationTable.h>
```
HowardHinnant: Nitpick:  We like to order our "like includes" alphabetically.

src/ripple/overlay/PeerReservationTable.h
```diff
@@ -0,0 +1,48 @@
...
+#include <string>
+#include <unordered_map>
+#include <ripple/beast/hash/uhash.h>
+#include <ripple/protocol/PublicKey.h>
+
+namespace ripple {
+// The rest of this directory is just in the `ripple` namespace.
+// Do we want to stick with that pattern, or do we want the namespace to
+// match the directory structure (to avoid name collisions and abide by
+// the principle of least surprise)?
```
HowardHinnant: My vote would be to keep this in `ripple` namespace.  But it is not something I'll loose any sleep over.  If we do go with `ripple::overlay` we should start migrating other types to it for consistency.

src/ripple/rpc/handlers/Reservations.cpp
```diff
@@ -0,0 +1,69 @@
...
+        // jaReservation["public_key"] = reservation.identity_;
+        if (reservation.name_) {
+            jaReservation["name"] = *reservation.name_;
+        }
+        jaReservations.append(jaReservation);
+    }
+    return result;
+}
+
+}
```
HowardHinnant: Needs newline

thejohnfreeman: VS Code... :(

nbougalis:
@MarkusTeufelberger, obviously having the conversation on our internal wiki isn't idea. Here's the relevant content in the form of an issue: https://github.com/ripple/rippled/issues/2938

MarkusTeufelberger:
Thanks! :-)

Much clearer what this is supposed to do (without having to guess at the intentions).

miguelportilla:
@thejohnfreeman For unity builds, `rcpcx1.cpp` should include `Reservations.cpp`

src/ripple/basics/Result.h
```diff
@@ -0,0 +1,146 @@
...
+};
+
+// A specialization for when the two types are the same.
+// Do we need enable_if protection to compare decayed types? Why or why not?
+template <typename TE>
+class Result<TE, TE> {
+private:
+    enum Which {
+        VALUE,
+        ERROR
```
miguelportilla: MS compiler complains about `ERROR` being a constant (error C2059: syntax error: 'constant' ). I assume because it was previously defined in `wingdi.h`.

You might also consider making this an enum class.

src/ripple/rpc/handlers/Reservations.cpp
```diff
@@ -18,22 +18,90 @@
...
 doReservationsAdd(RPC::Context& context)
 {
-    // TODO:
-    // 1. Get the parameters of the method.
-    // 2. Check that the parameters have the right types.
-    // 3. Check if the node already has a reservation.
-    // 4. Add the reservation to context.app.peerReservations().
+    auto const& params = context.params;
+
+    if (!params.isMember("public_key"))
```
miguelportilla: Consider using `jss::public_key` vs `public_key`

src/ripple/rpc/handlers/Reservations.cpp
```diff
@@ -18,22 +18,90 @@
...
+    {
+        return RPC::missing_field_error("public_key");
```
miguelportilla: Consider using `RPC::missing_field_error(jss::public_key);`

src/ripple/rpc/handlers/Reservations.cpp
```diff
@@ -0,0 +1,126 @@
...
+    }
+
+    Json::Value result;
+
+    RPCResult<PublicKey> resPk = parsePublicKey(
+        // TODO: Can asString() fail?
```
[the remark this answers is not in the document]
miguelportilla: I suggest checking the type after checking `isMember`.
```
    if (!params[jss::public_key].isString())
    {
        return RPC::make_error(
            rpcINVALID_PARAMS,
            RPC::expected_field_message(jss::public_key, "a string"));
    }
```

src/ripple/rpc/handlers/Reservations.cpp
```diff
@@ -18,22 +18,90 @@
...
+            result,
+            identity,
+            // TODO: Can asString() fail?
+            params["public_key"].asString(),
+            TokenType::NodePublic))
+        return result;
+
+    const boost::optional<std::string> name = params.isMember("name")
+        ? boost::make_optional(params["name"].asString())
+        : boost::none;
```
miguelportilla: Consider using `jss::name` and checking the param type.

src/ripple/rpc/handlers/Reservations.cpp
```diff
@@ -18,22 +18,90 @@
...
+
+    // I think most people just want to overwrite existing reservations.
+    // TODO: Make a formal team decision that we do not want to raise an
+    // error upon a collision.
+    auto emplaced = context.app.peerReservations().emplace(
+        std::make_pair(identity, overlay::PeerReservation{identity, name}));
```
miguelportilla: I realize its a WIP but you might get away with simply storing an optional string versus a `PeerReservation`.

src/ripple/overlay/PeerReservationTable.h
```diff
@@ -0,0 +1,48 @@
...
+    PublicKey const identity_;
+    std::string name_;
+};
+
+using PeerReservationTable = std::unordered_map<PublicKey, PeerReservation, beast::uhash<>>;
+
+}
+}
+
+#endif
```
miguelportilla: Missing carriage return.

miguelportilla:
You might want to consider having add and remove accept an array of objects allowing the caller to specify one or more pubkey/name pairs in a single call.

src/ripple/overlay/PeerReservationTable.h
thejohnfreeman: `ripple` alone it is, then.

src/ripple/basics/Result.h
thejohnfreeman: I'm just removing this class. Boost 1.70 is coming soon, and it brings `boost::result`.

scottschurr:
Sadly, this branch is is not currently building for me on macOS using Boost 1.70.0.  It looks like 1.70.0 lost the `get_io_service()` method in a couple of places.  All of the build errors I'm seeing involve the invocation of `get_io_service()` member functions.  The lines that I'm seeing build errors on are:
```
src/ripple/server/impl/PlainWSPeer.h:68
src/ripple/server/impl/PlainHTTPPeer.h:71
src/ripple/server/impl/SSLWSPeer.h:76
src/ripple/server/impl/SSLHTTPPeer.h:81
src/ripple/server/impl/Door.h:180
src/ripple/server/impl/Door.h:182
src/ripple/server/impl/Door.h:376
```
It's possible that a rebase atop 1.3.0-b6 will fix the issues.  It looks like the branch is currently sitting atop 1.3.0-b3

src/ripple/peerfinder/impl/Logic.h
```diff
@@ -1088,8 +1088,8 @@ class Logic
                 item ["inbound"]    = "yes";
             if (slot.fixed())
                 item ["fixed"]      = "yes";
-            if (slot.cluster())
-                item ["cluster"]    = "yes";
+            if (slot.reserved())
+                item ["reserved"]   = "yes";
```
scottschurr: Similar to Counts.h, I don't _think_ this is written to a public API, so I'm guessing the rename from `cluster` to `reserved` won't hurt anybody. But that would be a good thing to research and confirm.  And maybe leave a comment once you've done the research.  Thanks.

src/ripple/app/main/DBInit.cpp
```diff
@@ -131,6 +131,13 @@ const char* WalletDBInit[] =
         PrivateKey      CHARACTER(52)               \
     );",
 
+    // Peer reservations.
+    // REVIEWER: How do we handle table migrations if we need to add a column?
```
scottschurr: Great question.  I don't know the answer.  But when you do find out the answer I think you should write up that answer as a comment.  Thanks.

src/ripple/overlay/PeerReservationTable.h
```diff
@@ -0,0 +1,142 @@
...
+
+    auto
+    toJson() const -> Json::Value;
+};
+
+class PeerReservationTable
+{
+public:
+    using table_type =
+        std::unordered_map<PublicKey, PeerReservation, beast::uhash<>>;
```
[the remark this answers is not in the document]
thejohnfreeman: Every time I've tried to remove the key from a record type that is kept in a map, I've come to regret it. At some point, I want the full record, and then my only choice is a tuple of `(Key, Record)` (speaking abstractly).

Duplicating the key costs 41 bytes on a 64-bit machine, and we don't expect anywhere near 100 peer reservations. I'd rather not slice this record type until it is known that its size is a problem.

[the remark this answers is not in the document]
scottschurr: Your call.  However it's not just about space savings.  It's also about duplication of information, which can lead to bugs.  Certainly your current approach works.  But it also allows for the situation where the `PublicKey` in the key doesn't match the `PublicKey` in the `PeerReservation` value.  Yes, that would be a bug.  But bugs happen.

At any rate, I won't lose sleep over either approach.  Use your best judgement.

src/ripple/overlay/impl/PeerReservationTable.cpp
```diff
@@ -0,0 +1,142 @@
...
+    PeerReservation const rvn{nodeId, desc};
+    boost::optional<PeerReservation> previous;
+
+    auto emplaced = table_.emplace(nodeId, rvn);
+    if (!emplaced.second)
+    {
+        // The node already has a reservation.
+        // I think most people just want to overwrite existing reservations.
+        // TODO: Make a formal team decision that we do not want to raise an
+        // error upon a collision.
```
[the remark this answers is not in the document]
thejohnfreeman: That's exactly how I feel. Maybe I didn't express it well enough.

src/ripple/overlay/PeerReservationTable.h
```diff
@@ -0,0 +1,142 @@
...
+#include <boost/optional.hpp>
+#include <soci/soci.h>
+
+#include <string>
+#include <unordered_map>
+
+namespace ripple {
+
+// Value type for reservations.
+// REVIEWER: What is the preferred naming convention for member variables?
```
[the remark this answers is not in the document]
thejohnfreeman: Because I want this to be a value type, ideally one that can be constructed as an aggregate, I'm trying to omit any constructors, which lets me use the nice underscore-less convention without any complications.

src/ripple/overlay/impl/OverlayImpl.cpp
```diff
@@ -278,22 +278,24 @@ OverlayImpl::onHandoff (std::unique_ptr <beast::asio::ssl_bundle>&& ssl_bundle,
...
-        JLOG(journal.debug()) <<
-            "Peer " << remote_endpoint << " redirected, slots full";
-        handoff.moved = false;
-        handoff.response = makeRedirectResponse(slot, request,
-            remote_endpoint.address());
-        handoff.keep_alive = beast::rfc2616::is_keep_alive(request);
-        return handoff;
+        // The node gets a reserved slot if it is in our cluster
+        // or if it has a reservation.
+        bool const reserved = (
```
scottschurr: I'm a little surprised by the indentation here, but I won't lose sleep over it.

If you use brace initialization without the `=` then your compiler will stop you if you are making any narrowing conversions here.  It would look like this:
```
bool const reserved {
    static_cast<bool>(app_.cluster().member(*publicKey))
    || app_.peerReservations().contains(*publicKey)};
```
In C++14 don't use that technique for `auto`, only for explicit types.  But you're initializing a `bool` here, so you're okay.  When we get to C++17 it will work with `auto ` as well.  Personally, I think it's a good habit to develop.

src/ripple/overlay/impl/OverlayImpl.cpp
```diff
@@ -278,22 +278,24 @@ OverlayImpl::onHandoff (std::unique_ptr <beast::asio::ssl_bundle>&& ssl_bundle,
...
+                static_cast<bool>(app_.cluster().member(*publicKey))
+                || app_.peerReservations().contains(*publicKey)
+        );
+        auto const result = m_peerFinder->activate(slot, *publicKey, reserved);
+        if (result != PeerFinder::Result::success) {
```
scottschurr: FWIW, in this code base the opening brace of an `if` is typically on the next line rather than at the end of the `if`.  That may take some getting used to for some folks...

src/ripple/overlay/PeerReservationTable.h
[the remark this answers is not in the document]
scottschurr: I realized overnight that, since you want to keep the key in the record type, what you really want is not a `map`.  You want a `set`.  Yes, since you're using a hash table you have to write a special hash function that doesn't include the `description` in the hash.  But, in my opinion, a `set` really matches your design intent.

Again, I won't lose any sleep over the choice you make here.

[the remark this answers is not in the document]
thejohnfreeman: You're exactly right. I'll make that change. Hoping I can understand @HowardHinnant's `hash` enough to write my own overload...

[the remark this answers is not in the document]
HowardHinnant: If you have any questions, I'm happy to help.  And happy to review.

src/ripple/overlay/PeerReservationTable.h
```diff
@@ -0,0 +1,139 @@
...
+            PeerReservation const& lhs, PeerReservation const& rhs) const
+    {
+        return lhs.nodeId == rhs.nodeId;
+    }
+};
+
+class PeerReservationTable
+{
+public:
+    using table_type = std::unordered_set<
```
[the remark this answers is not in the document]
thejohnfreeman: This way doesn't require any special gymnastics if we ever decide to add more fields. I believe this is the best abstraction for the data held in an SQL table: a set of records indexed on a subset of the record's fields.

src/ripple/app/main/Main.cpp
```diff
@@ -149,6 +149,9 @@ void printHelp (const po::options_description& desc)
            "     peers\n"
            "     ping\n"
            "     random\n"
+           "     reservations_add <public_key> [<description>]\n"
+           "     reservations_del <public_key>\n"
+           "     reservations_list\n"
```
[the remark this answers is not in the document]
thejohnfreeman: I would like that syntax as well, but I don't think it's supported by the current framework, so I don't want to shoehorn that functionality into this change.

src/ripple/overlay/PeerReservationTable.h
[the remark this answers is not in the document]
nbougalis: I would have gone with the `std::unordered_map` because it's unlikely we'll add more data to this and the map simplifies (imo) the interface we use and removes the need for an additional type.

However, I understand your argument and, ultimately, it's your call. Thanks for explaining your rationale.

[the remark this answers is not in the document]
thejohnfreeman: You might also think of it as better naming. The `map` stores `pair`s, with `first` and `second` members. The `set` stores `PeerReservation`s, with `nodeId` and `name` members.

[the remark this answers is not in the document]
seelabs: I understand why the `set` is attractive. It is a nice solution for a database row indexed by a subset of its columns. For this table, we are indexing by one row - the `PublicKey` - and we need to get an `optional<string>`. I looked at what a solution using `unordered_map<PublicKey, optional<string>` would look like (see: https://github.com/seelabs/rippled/commit/9c68c5a677e88cc9eeb03b29dcd6670a342f4c83). (Aside: the previous `map` solution stored the `PeerReservation`. I think I prefer storing the description as a value instead).

Here's how I think it compares with using `set`

Points in favor of `set`:

1) If we add more columns to the db table, or we index on multiple columns, the `set` solution naturally handles them.

2) Iterating through the set is more natural. The `set` iterator returns a `PeerReservation` with named field values. Iterating through the `map` solution is less natural, it returns a `pair<PublicKey, optional<string>>`

Points in favor of the `map` solution:

1) It's easy to understand what the `map` solution is doing. We're using `map` in ways it's been used many times in the past. The `set` solution is a rarer use of `set`. It's not immediately obvious that we're just indexing on `PublicKey`. (I'm not claiming it's super difficult, I just find the intention of the `map` solution clearer).

2) The `map` solution requires less boilerplate. I.e `hash_append` and `KeyEqual` go away.

3) the `mutable` public data member naturally goes away. I don't think we want the mutable public member even in the `set` solution, as it breaks const correctness in a bad way (i.e. a caller can accidentally modify a description). I think we can make it go away in the `set` solution as well (we probably need to `const_cast` in upsert, tho I admit I haven't looked closely).

I probably vote for the `map` solution myself, tho there are points in favor of both solutions. And of course, it's your PR, whatever you decide is fine with me, just thought I'd give my view.

src/ripple/app/main/Main.cpp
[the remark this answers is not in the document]
nbougalis: I hear you and agree with you 100%.

src/ripple/overlay/impl/PeerReservationTable.cpp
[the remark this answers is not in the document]
nbougalis: :bulb: We should use `ripple::rand_bool` to decide which name to keep...

In all seriousness, I don't think this merits a warning; we can quietly replace one string with another, effectively formalizing the rule that "what we called you last is what we'll call you next" which seems reasonable enough to me.

src/ripple/overlay/PeerReservationTable.h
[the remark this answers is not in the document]
nbougalis: > You might also think of it as better naming. The `map` stores `pair`s, with `first` and `second` members. The `set` stores `PeerReservation`s, with `nodeId` and `description` members.

That's a fair point: the `first` / `second` nomenclature is a bit of a pain in almost every instance. I said my piece and I'm fine with your decision/approach.

However, I'll point something else out (and @seelabs did it too, in his response above): `boost::optional<std::string>`.

Please ditch the `boost::optional<string>`. There's no semantic difference (than I can think of) between an empty `string` and an unseated `boost::optional<string>` so the latter only adds complexity.

src/ripple/overlay/impl/PeerReservationTable.cpp
```diff
@@ -0,0 +1,168 @@
...
+        previous = *hint;
+        // We should pick an adjacent location for the insertion hint.
+        // Decrementing may be illegal if the found reservation is at the
+        // beginning. Incrementing is always legal; at worst we'll point to
+        // the end.
+        auto const deleteme = hint;
+        ++hint;
+        table_.erase(deleteme);
+    }
+    auto existed = table_.insert(hint, reservation);
```
seelabs: Clang warns that `existed` is an unused variable.

src/ripple/overlay/PeerReservationTable.h
```diff
@@ -0,0 +1,127 @@
...
+        return lhs.nodeId == rhs.nodeId;
+    }
+};
+
+class PeerReservationTable final
+{
+public:
+    using table_type = std::unordered_set<
+        PeerReservation, beast::uhash<>, KeyEqual>;
+    using const_iterator = table_type::const_iterator;
```
seelabs: Nit: I'd remove `table_type` and `const_iterator` type aliases. `table_type` is only used once, in the declaration of `table_`. `const_iterator` isn't used at all.

scottschurr: I'll second @seelabs recommendation.

src/ripple/overlay/impl/PeerReservationTable.cpp
```diff
@@ -0,0 +1,168 @@
...
+        result[jss::description] = description;
+    }
+    return result;
+}
+
+auto
+PeerReservationTable::list() const -> std::vector<PeerReservation>
+{
+    std::vector<PeerReservation> list;
+    list.reserve(table_.size());
```
scottschurr: I understand why you want to do this, but it's not safe to call `table_.size()` outside of the lock.  So you should move this line of code so it is inside the `lock_guard`.

src/ripple/overlay/impl/PeerReservationTable.cpp
```diff
@@ -0,0 +1,168 @@
...
+
+    auto hint = table_.find(reservation);
+    if (hint != table_.end()) {
+        // The node already has a reservation. Remove it.
+        // `std::unordered_set` does not have an `insert_or_assign` method,
+        // and sadly makes it impossible for us to implement one efficiently:
+        // https://stackoverflow.com/q/49651835/618906
+        // Regardless, we don't expect this function to be called often, or
+        // for the table to be very large, so this less-than-ideal
+        // remove-then-insert is acceptable in order to present a better API.
```
scottschurr: FWIW, I suspect the awkwardness of this method suggests that the right choice for `table_`'s container would have been a `map` or `unordered_map`.  Then you'd be able to replace the `value` without needing to remove the entire node.  But I also know this has been discussed plenty, and I'm not asking for any changes.

src/ripple/overlay/impl/PeerReservationTable.cpp
thejohnfreeman: Can you teach me why it's not safe?

src/ripple/overlay/impl/PeerReservationTable.cpp
thejohnfreeman: I disagree. I think it's a problem with `unordered_set`. If it had an `insert_or_assign` method the same as `unordered_map`, then there would be no awkwardness. There is no technical reason it cannot have such a method. I don't think a deficiency in one standard container means the right choice is to pick a differently deficient container.

src/ripple/overlay/impl/PeerReservationTable.cpp
scottschurr: From a standees perspective, the `size()` of `table_` is not synchronized.  Any unsynchronized access invokes undefined behavior.

From a practical perspective, you have two concerns: since the return value of `size()` is bigger than a byte, the value might be torn in which case who knows what value `size()` will return.  Another concern is what the optimizer might do with this stuff.  Who know where `size()` might actually be invoked if you don't place it under the lock.  Of the optimizer might recognize that you have undefined behavior here and remove chunks of the code.

But given what optimizers do today I take any threat of undefined behavior very seriously.

src/ripple/overlay/impl/PeerReservationTable.cpp
scottschurr: From the perspective of the standard designers there is no deficiency.  The `key` to both a `set` and a `map` _must_ be immutable.  If you were allowed to mutate a `key` in place that would corrupt the data structure.  So, sensibly enough, the interface doesn't allow you to mutate the `key` (without working hard).

A `set` _only_ has a `key`, and there are no options for a partially non-`const` `key`.  If you need a partially non-`const` key, well, that's what `map` is for.

src/ripple/overlay/impl/PeerReservationTable.cpp
thejohnfreeman: An unsynchronized call of a const method is UB? Where can I read more about this. First I've heard of it.

src/ripple/overlay/impl/PeerReservationTable.cpp
thejohnfreeman: Can you define "mutate the key"? I don't see it as a mutation of the key. I see it as an atomic replacement. It is not like asking the container to return an iterator to a non-const key where I can mutate the key outside the purview of the container. I am asking the container itself to atomically remove one key and insert another. It has full control of its invariants before, during, and after that process. No corruption would be possible. I don't need the `key_type` or `value_type` of `unordered_set` to be mutable at all. I just need an amotic `replace`/`upsert`/`insert_or_assign`/whatever-you-want-to-call-it method.

src/ripple/overlay/impl/PeerReservationTable.cpp
thejohnfreeman: TIL any unsynchronized access (even single-writer) of a "memory location" (even a scalar type like `std::size_t`) from multiple threads is UB.

> A memory location is
>
> - an object of scalar type (arithmetic type, pointer type, enumeration type, or `std::nullptr_t`)
> - or the largest contiguous sequence of bit fields of non-zero length

> When an evaluation of an expression writes to a memory location and another evaluation reads or modifies the same memory location, the expressions are said to conflict. A program that has two conflicting evaluations has a data race unless
>
> - both evaluations execute on the same thread or in the same signal handler, or
> - both conflicting evaluations are atomic operations (see std::atomic), or
> - one of the conflicting evaluations happens-before another
>
> If a data race occurs, the behavior of the program is undefined.

https://en.cppreference.com/w/cpp/language/memory_model

scottschurr: In http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/n4820.pdf (the most current working draft) `Section 6.8.2.1 Data races` paragraph 21.2

> The execution of a program contains a data race if it contains two potentially concurrent conflicting actions, at least one of which is not atomic, and neither happens before the other, except for the special case for signal handlers described below. Any such data race results in undefined behavior. [Note: It can be shown that programs that correctly use mutexes and memory_order::seq_cst operations to prevent all data races and use no other synchronization operations behave as if the operations executed by their constituent threads were simply interleaved, with each value computation of an object being taken from the last side effect on that object in that interleaving. This is normally referred to as “sequential consistency”. However, this applies only to data-race-free programs, and data-race-free programs cannot observe most program transformations that do not change single-threaded program semantics. In fact, most single-threaded program transformations continue to be allowed, since any program that behaves differently as a result must perform an undefined operation.

And this isn't new.  I'm fairly confident it has been true since C++11 when the standard became thread aware.

src/ripple/overlay/impl/PeerReservationTable.cpp
scottschurr: The reason I was referring to "mutate the key" is because the part of the key that determines the location in the `set` does not change during the "atomic replacement".  The only thing that changes is the `description_`.  So you could leave the key in place and "mutate the key" by changing (only) `description_`.

I'm not going to argue with you about whether `insert_or_assign` _should_ be a part of the standard `set` interface.

src/ripple/overlay/impl/PeerReservationTable.cpp
thejohnfreeman: I'm not saying it is new, it's just new to me. I stepped away from C++ soon after C++11 was passed, and I was never exposed to the concurrency model. I'm still catching up.

PR #2937, opened 2019-05-15
[TRIVIAL] Add some missing items to help command list:

ximinez:
* validators
* validator_list_sites
* Put "version" in the right place

Since these commands ignore parameters, there's not much to it.

### DIFF
--- a/src/ripple/app/main/Main.cpp
+++ b/src/ripple/app/main/Main.cpp
@@ -151,7 +151,6 @@ void printHelp (const po::options_description& desc)
            "     random\n"
            "     ripple ...\n"
            "     ripple_path_find <json> [<ledger>]\n"
-           "     version\n"
            "     server_info [counters]\n"
            "     server_state [counters]\n"
            "     sign <private_key> <tx_json> [offline]\n"
@@ -161,6 +160,9 @@ void printHelp (const po::options_description& desc)
            "     submit_multisigned <tx_json>\n"
            "     tx <id>\n"
            "     validation_create [<seed>|<pass_phrase>|<key>]\n"
+           "     validators\n"
+           "     validator_list_sites\n"
+           "     version\n"
            "     wallet_propose [<passphrase>]\n";
 }
 


src/ripple/app/main/Main.cpp
```diff
@@ -161,6 +160,9 @@ void printHelp (const po::options_description& desc)
            "     submit_multisigned <tx_json>\n"
            "     tx <id>\n"
            "     validation_create [<seed>|<pass_phrase>|<key>]\n"
+           "     validators"
+           "     validator_list_sites"
+           "     version\n"
            "     wallet_propose [<passphrase>]\n";
 }
```
mellery451: this looks good, but seems like it still doesn't completely match the list at https://github.com/ripple/rippled/blob/5214b3c1b09749420ed0682a2e49fbbd759741e5/src/ripple/rpc/impl/Handler.cpp#L58-L116  -- should it or do we deliberately omit some ?

nbougalis: We deliberately omit some of these.

src/ripple/app/main/Main.cpp
```diff
@@ -161,6 +160,9 @@ void printHelp (const po::options_description& desc)
            "     submit_multisigned <tx_json>\n"
            "     tx <id>\n"
            "     validation_create [<seed>|<pass_phrase>|<key>]\n"
+           "     validators"
```
MarkusTeufelberger:
```suggestion
           "     validators\n"
```

src/ripple/app/main/Main.cpp
```diff
@@ -161,6 +160,9 @@ void printHelp (const po::options_description& desc)
...
+           "     validator_list_sites"
```
MarkusTeufelberger:
```suggestion
           "     validator_list_sites\n"
```

src/ripple/app/main/Main.cpp
ximinez: Which ones do we deliberately omit? Should we make a notation in `handlerArray` as such?

MarkusTeufelberger:
Oh damn it, my company mail address seems to be used by default in these "suggested changes"... O_o

GitHub still manages to put in slightly annoying things for people working on open source both privately and professionally... I'm sure my employer is ok with me committing 2 newlines though, so no biggie.

ximinez:
@MarkusTeufelberger I squashed the commits.

src/ripple/app/main/Main.cpp
nbougalis: Generally, the ones that require multiple parameters or involve complicated parsing.

nbougalis:
@MarkusTeufelberger: you can choose which e-mail address to use when making changes via the GitHub web-based UI (I'm assuming you use it for your commit)

Issue #2883, opened 2019-03-26
Websocket closes and dies


tuloski:
I have a rippled with the following modif on the rippled.cfg:

[server]
port_ws_public

[port_ws_public]
port = 8888
ip = 0.0.0.0
protocol = ws

I have multiple peers (locally) that connect to that ws.

If one host requests lots of data via the ws (for example lots of books), the ws closes with code 1000, but:

* It's not technically a code 1000.
* I'm not able to reconnect to the ws until I restart rippled.
* All other peers that were using that ws are disconnected﻿﻿ and can't reconnect until restart of rippled.

The following code (with RippleAPI) can be used to trigger the error:

```
const RippleAPI = require('ripple-lib').RippleAPI;

var util = require('util');
var fs = require('fs');

var CURRENCIES = [
	{ 
		currency : 'USD',
		  issuer : 'rvYAfWj5gh67oV6fW32ZzP3Aw4Eubs59B'
	},
	{ 
		currency : 'BTC',
		  issuer : 'rvYAfWj5gh67oV6fW32ZzP3Aw4Eubs59B'
	},
	{ 
		currency : 'USD',
		  issuer : 'rhub8VRN55s94qWKDv6jmDy1pUykJzF3wq'
	},
	{ 
		currency : 'EUR',
		  issuer : 'rhub8VRN55s94qWKDv6jmDy1pUykJzF3wq'
	},
	{ 
		currency : 'BTC',
		  issuer : 'rchGBxcD1A1C2tdxF6papQYZ8kjRKMYcL'
	},
	{ 
		currency : 'BCH',
		  issuer : 'rchGBxcD1A1C2tdxF6papQYZ8kjRKMYcL'
	},
	{ 
		currency : 'ETH',
		  issuer : 'rchGBxcD1A1C2tdxF6papQYZ8kjRKMYcL'
	},
	{ 
		currency : 'ETC',
		  issuer : 'rchGBxcD1A1C2tdxF6papQYZ8kjRKMYcL'
	},
	{ 
		currency : 'REP',
		  issuer : 'rchGBxcD1A1C2tdxF6papQYZ8kjRKMYcL'
	},
];

api = new RippleAPI({
	//server: 'wss://s-west.ripple.com:443'
	//server: 'wss://s1.ripple.com:443'
	server: 'ws://127.0.0.1:8888',
	maxFeeXRP: '0.1'
});

//----- RIPPLE API EVENTS ----------------
[...]

Issue #2887, opened 2019-03-29
Report to WebSocket clients connection refusal reason


miguelportilla:
Report connection refusal reason to WebSocket clients when the rate limit threshold has been surpassed.

[issue 2883](https://github.com/ripple/rippled/issues/2883)

movitto:
Ran into this locally as well so decided to implement a fix. Let me know if the style needs updating, etc.

miguelportilla:
@movitto Thank you!

mDuo13:
The fix has been released in 1.3.1. I plan to add documentation sometime soon.

PR #2948, opened 2019-05-20
Return WS error code & reason on closure due to balance threshold exceeded

movitto:
Clients can now inspect error code & msg to ascertain reason for connection being closed.

Fixes issues #2883 and #2887

### DIFF
--- a/src/ripple/rpc/impl/ServerHandlerImp.cpp
+++ b/src/ripple/rpc/impl/ServerHandlerImp.cpp
@@ -359,7 +359,7 @@ ServerHandlerImp::onWSMessage(
     if (postResult == nullptr)
     {
         // The coroutine was rejected, probably because we're shutting down.
-        session->close();
+        session->close({boost::beast::websocket::going_away, "Shutting Down"});
     }
 }
 
@@ -388,7 +388,7 @@ ServerHandlerImp::processSession(
     auto is = std::static_pointer_cast<WSInfoSub> (session->appDefined);
     if (is->getConsumer().disconnect())
     {
-        session->close();
+        session->close({boost::beast::websocket::policy_error, "threshold exceeded"});
         // FIX: This rpcError is not delivered since the session
         // was just closed.
         return rpcError(rpcSLOW_DOWN);
--- a/src/ripple/server/WSSession.h
+++ b/src/ripple/server/WSSession.h
@@ -136,9 +136,12 @@ struct WSSession
     void
     send(std::shared_ptr<WSMsg> w) = 0;
 
+    virtual void
+    close() = 0;
+
     virtual
     void
-    close() = 0;
+    close(boost::beast::websocket::close_reason const& reason) = 0;
 
     /** Indicate that the response is complete.
         The handler should call this when it has completed writing
--- a/src/ripple/server/impl/BaseWSPeer.h
+++ b/src/ripple/server/impl/BaseWSPeer.h
@@ -105,6 +105,9 @@ public:
     close() override;
 
     void
+    close(boost::beast::websocket::close_reason const& reason) override;
+
+    void
     complete() override;
 
 protected:
@@ -224,13 +227,12 @@ send(std::shared_ptr<WSMsg> w)
         return;
     if(wq_.size() > port().ws_queue_limit)
     {
-        JLOG(this->j_.info()) <<
-            "closing slow client";
         cr_.code = safe_cast<decltype(cr_.code)>
                       (boost::beast::websocket::close_code::policy_error);
         cr_.reason = "Policy error: client is too slow.";
+        JLOG(this->j_.info()) << cr_.reason;
         wq_.erase(std::next(wq_.begin()), wq_.end());
-        close();
+        close(cr_);
         return;
     }
     wq_.emplace_back(std::move(w));
@@ -238,24 +240,38 @@ send(std::shared_ptr<WSMsg> w)
         on_write({});
 }
 
-template<class Handler, class Impl>
+template <class Handler, class Impl>
 void
-BaseWSPeer<Handler, Impl>::
-close()
+BaseWSPeer<Handler, Impl>::close()
 {
-    if(! strand_.running_in_this_thread())
-        return post(
-            strand_, std::bind(&BaseWSPeer::close, impl().shared_from_this()));
+    close(boost::beast::websocket::close_reason{});
+}
+
+template <class Handler, class Impl>
+void
+BaseWSPeer<Handler, Impl>::close(
+    boost::beast::websocket::close_reason const& reason)
+{
+    if (!strand_.running_in_this_thread())
+        return post(strand_, [self = impl().shared_from_this(), reason] {
+            self->close(reason);
+        });
     do_close_ = true;
-    if(wq_.empty())
+    if (wq_.empty())
+    {
         impl().ws_.async_close(
-            {},
+            reason,
             bind_executor(
                 strand_,
-                std::bind(
-                    &BaseWSPeer::on_close,
-                    impl().shared_from_this(),
-                    std::placeholders::_1)));
+                [self = impl().shared_from_this()](
+                    boost::beast::error_code const& ec) {
+                    self->on_close(ec);
+                }));
+    }
+    else
+    {
+        cr_ = reason;
+    }
 }
 
 template<class Handler, class Impl>


src/ripple/server/impl/BaseWSPeer.h
```diff
@@ -234,14 +237,36 @@ close()
 {
     if(! strand_.running_in_this_thread())
         return strand_.post(std::bind(
-            &BaseWSPeer::close, impl().shared_from_this()));
+            static_cast<void(BaseWSPeer<Handler, Impl>::*)(void)>(&BaseWSPeer::close),
```
HowardHinnant: I suggest a clarifying comment for the need for this case, and a slight simplification in its syntax:
```
            // disambiguate to this close overload
            static_cast<void(BaseWSPeer::*)()>(&BaseWSPeer::close),
```

movitto:
@HowardHinnant updated

scottschurr:
@movitto, congratulations.  It looks like both @intelliot and @HowardHinnant are good with your changes (I'm guessing based on a :+1:).  I think the next step would be to rebase your work on top of the current develop (which is 1.3.0-b4).  That should resolve whatever conflicts there are.  Thanks.

movitto:
@scottschurr sure thing will do. That being said, I have a few approved PR's that are outstanding now & I'm a bit concerned about bit rot w/ them, eg the longer these things wait & are in the queue the more likely they'll need to be rebased upon changes pushed to develop (not a big deal for the smaller ones but for #2959 every rebase will be a big undertaking). Perhaps we can establish a merge window in the near future to get these in?

scottschurr:
@movitto bit rot on pull requests is a legit concern.  I took a quick glance at your current pull requests:

- https://github.com/ripple/rippled/pull/2959: **Passed**.  I would expect this to be merged to 1.3.0-b4.
- https://github.com/ripple/rippled/pull/2949: **Passed**.  I would expect this to be merged to 1.3.0-b4.
- https://github.com/ripple/rippled/pull/2948: Not yet passed.  Can't be merged with conflicts.
- https://github.com/ripple/rippled/pull/2932: Awaiting review from @nbougalis and @mtrippled.
- https://github.com/ripple/rippled/pull/2931: Blocked indefinitely waiting for IANA.

Your biggest concern for bit rot, https://github.com/ripple/rippled/pull/2959, is very likely to go into 1.3.0-b4.  This pull request, https://github.com/ripple/rippled/pull/2948, looks like it has a good chance of going into 1.3.0-b4 once the conflicts are resolved.  I can't really speak to your remaining non-passed pull requests.

movitto:
Cool, thx for addressing my concern @scottschurr. Just rebased & pushed.

nbougalis:
@manojsdoshi, can you please include this PR in the **rc1** candidate you're putting together? Thanks.

scottschurr:
Actually, @manojsdoshi, @seelabs is currently reviewing the asio portion of this pull request and may request some changes.  So please don't include this pull request in rc1.  @seelabs has some obligations today so he's unable to finish the review immediately.  However I expect to see his feedback tomorrow.

@movitto, sorry for the delay.  But I think the more eyes the better.

src/ripple/rpc/impl/ServerHandlerImp.cpp
```diff
@@ -388,7 +388,7 @@ ServerHandlerImp::processSession(
     auto is = std::static_pointer_cast<WSInfoSub> (session->appDefined);
     if (is->getConsumer().disconnect())
     {
-        session->close();
+        session->close(1008, "threshold exceeded");
```
seelabs: Instead of the magic number `1008` here, could we use `boost::beast::websocket::policy_error`?

src/ripple/server/WSSession.h
```diff
@@ -140,6 +140,10 @@ struct WSSession
     void
     close() = 0;
 
+    virtual
+    void
+    close(const std::uint16_t& code, const std::string& reason) = 0;
```
seelabs: Instead of passing the `code` and `reason`, could be pass `boost::beast::websocket::close_reason` instead?

(also, Nit: We prefer to take integer types like `std::uint16_t` by value instead of `const&`)

src/ripple/server/impl/BaseWSPeer.h
```diff
@@ -258,6 +264,33 @@ close()
...
+void
+BaseWSPeer<Handler, Impl>::
+close(const std::uint16_t& code, const std::string& reason)
+{
+    if(! strand_.running_in_this_thread())
+        return post(strand_, std::bind(
+            // disambiguate to this close overload (also above)
+            static_cast<void(BaseWSPeer::*)(const std::uint16_t&, const std::string&)>(&BaseWSPeer::close),
+            impl().shared_from_this(),
+            code, reason));
```
seelabs: Instead of `std::bind` and a `static_cast` to disambiguate the overload, what do you think about using a lambda instead? Something like:
```
    if (!strand_.running_in_this_thread())
        return post(strand_, [self = impl().shared_from_this(), reason] {
            self->close(reason);
        });
```

src/ripple/server/impl/BaseWSPeer.h
```diff
@@ -258,6 +264,33 @@ close()
...
+          _reason,
+          bind_executor(
+              strand_,
+              std::bind(
+                  &BaseWSPeer::on_close,
+                  impl().shared_from_this(),
+                  std::placeholders::_1)));
+    }
+}
+
```
seelabs: This function and the one above are almost identical. The only difference is one passes an empty reason while the other passes the caller given reason. How about having the `close()` method create an empty reason and call the `close(boost::beast::websocket::close_reason const&)` function? (Note: I suggest replacing the `close(std::uint16_t const&, std::string const&)` function with one taking a single `reason` parameter elsewhere in this review).

src/ripple/server/impl/BaseWSPeer.h
```diff
@@ -258,6 +264,33 @@ close()
...
+{
+    if(! strand_.running_in_this_thread())
+        return post(strand_, std::bind(
+            // disambiguate to this close overload (also above)
+            static_cast<void(BaseWSPeer::*)(const std::uint16_t&, const std::string&)>(&BaseWSPeer::close),
+            impl().shared_from_this(),
+            code, reason));
+    do_close_ = true;
+    if(wq_.empty()){
+      boost::beast::websocket::close_reason _reason;
```
seelabs: Nit: we usually don't name variables with a leading underscore.

scottschurr: Additional nit: standard indentation in this code base is four spaces.  The bulk of this pull request adheres to that quite well.  However this particular area, starting what the `if(wq_.empty())` has some indentation issues.

movitto:
Hey I was able to give @seelabs' followup patch a quick lookover this morning (am travelling & out of town for a few days) and I agree with the proposed changes, the code looks much cleaner. I also gave it a whirl and it looks to cleanly build locally though I didn't get as far as testing/verify the actual functionality (I had previously done this using the example provided in #2883). But outside of that, no objections from me, merge away!

movitto:
Speaking of merging is the idea for me to fold @seelabs' patch into mine? I can do this though would be a bit concerned that we'd lose the history of both myself and @seelabs contributing to this one. It's probably NBD for this one, but it's often nice to see all the contributors to patchset in the commit log (incase of questions, issues, etc later on). Just let me know!

And thanks again for the review! (I'll get to the other PR's with outstanding comments early next week)

seelabs:
@movitto Great, glad the changes look OK to you. I'd like to get a thumbs up from @miguelportilla before we mark this passed. 

Yes, the idea is for you to merge my changes into yours. Usually the PR owner would cherry pick my commit and add it to this PR (so reviewers can leave comments). When this eventually passes, squash all the commits down to one commit with you as the author.

miguelportilla:
@seelabs I've reviewed your patch and left some comments.

With @seelabs patch, this PR looks fine to me. Thank you for the contribution @movitto!

seelabs:
@movitto I just pushed another patch to address @miguelportilla's feedback. Can you cherry pick this one as well: https://github.com/seelabs/rippled/commit/2741d6750336ea63719e5e98867423ef3a19c04e (https://github.com/seelabs/rippled/commit/891889c35f49ba937bdcdd10725ab9f8313f8af5 should be cherry picked first).

Of course, this is your PR, so please review the code in the new patch, and if you object to anything let me know. 

@HowardHinnant Can you also review this new patch?

movitto:
@seelabs just left some comments on @miguelportilla comments (comment'ception?! :smile: ), just let me know when it's good to squash (back in town, should be fairly quick)

movitto:
Ahh looks like our comment collided in mid air! (we didn't even reach cruising altitude!) Will try out the changes and if all works locally squash / repush

movitto:
@seelabs @miguelportilla @HowardHinnant @scottschurr OK, updated:

- cherry picked @seelabs' follow-on patches and reviewed locally. Everything seems inorder tho TBH I'm not 100% up to speed w/ the semantics of the message queue (wr_) and the relevant state (empty, non-empty, etc). That being said I did try out the example provided in #2883 and re-verified the fix.
- squashed & rebased onto develop
- force pushed

seelabs:
@movitto Thanks!

One quick nit: we prefer not to squash the commits down until the PR is officially passed (it makes it easier to review if the reviews can see in the commits what's new). No big deal that it was squashed now, just wanted to bring it up for next time.

HowardHinnant:
Still 👍 from me.

miguelportilla:
👍

PR #2949, opened 2019-05-20
Fix a few issues encountered during RPM/Docker buildprocess

movitto:
When running "**cmake --build . --target rpm**" on a recent
Fedora Release (28) against Docker 1.13.1 (both stock Fedora/Docker
installations with updates) I encountered two issues prevening RPM
building from completing.

First a Docker-specific error about "FROM" macro usage:
```
  Step 1/25 : ARG GIT_COMMIT=unknown
  Please provide a source image with `from` prior to commit
  gmake[3]: *** [CMakeFiles/rpm_container.dir/build.make:57: CMakeFiles/rpm_container] Error 1
  gmake[2]: *** [CMakeFiles/Makefile2:202: CMakeFiles/rpm_container.dir/all] Error 2
  gmake[1]: *** [CMakeFiles/Makefile2:361: CMakeFiles/rpm.dir/rule] Error 2
  gmake: *** [Makefile:268: rpm] Error 2
```

And after that is fixed an error about rpmbuild check-buildroot failing:
```
  + /usr/lib/rpm/check-buildroot
  Binary file /opt/rippled_bld/pkg/rpmbuild/BUILDROOT/rippled-1.3.0-0.1.b3.el7.x86_64/opt/ripple/bin/validator-keys matches
  Binary file /opt/rippled_bld/pkg/rpmbuild/BUILDROOT/rippled-1.3.0-0.1.b3.el7.x86_64/usr/lib/debug/opt/ripple/bin/validator-keys.debug matches
  Found '/opt/rippled_bld/pkg/rpmbuild/BUILDROOT/rippled-1.3.0-0.1.b3.el7.x86_64' in installed files; aborting
  error: Bad exit status from /var/tmp/rpm-tmp.YPR0mV (%install)

  RPM build errors:
      Bad exit status from /var/tmp/rpm-tmp.YPR0mV (%install)
```

This patch fixes both of those issues such that the RPM can sucessfully be built locally.

### DIFF
--- a/Builds/containers/centos-builder/Dockerfile
+++ b/Builds/containers/centos-builder/Dockerfile
@@ -1,6 +1,6 @@
+FROM centos:7
 ARG GIT_COMMIT=unknown
 
-FROM centos:7
 LABEL git-commit=$GIT_COMMIT
 
 COPY centos-builder/centos_setup.sh /tmp/


movitto:
With this PR I'd also like to open up a discussion pertaining to RPM build process, I found little documentation in the developer docs pertaining to this subject.

As far as my understanding the process above(cmake --build . --target rpm) is the supported mechanism for building the RPM for the CentOS / RH platforms correct? And then 'alien' is used to convert the build rpms to ubuntu packages (according to dev portal). Has there been any discussion as far as building the package for downstream linux distributions?

As a Fedora packager I know the process pretty well, and the biggest blocker would be the vendored dependencies in the project (the various libraries that are pulled in via git in CMakeLists). I'd suspect those will need to be extracted and packaged as dependencies before rippled could be Fedora-acceptable. The plus side to getting it in Fedora though would be that it could be added to the 'EPEL' repo, facilitating including in CentOS & RHEL by simply enabling that repo and installing from there.

Thoughts?

MarkusTeufelberger:
I wrote a few tests to explore building `rippled` (not even packages, just compiling) on various distributions: https://github.com/MarkusTeufelberger/rippled-distrotest

Usually one of the biggest hurdles is rippled's dependency on relatively recent versions of boost (you won't be able to build it on Fedora 28 for example using the Fedora supplied version of boost, you'll need Fedora 30 to get boost 1.69).

At least most external dependencies are not directly vendored any more. It still could be a bit easier to provide external dependencies as binaries instead of having them fetched and compiled (sometimes statically) by cmake.

movitto:
> Usually one of the biggest hurdles is rippled's dependency on relatively recent versions of boost (you won't be able to build it on Fedora 28 for example using the Fedora supplied version of boost, you'll need Fedora 30 to get boost 1.69).

Agreed. Luckily though we're not the only one's with this problem and recent versions of Fedora / RPM / DNF (eg yum) support software collections which would allow us to ship a custom version of boost (eg boost-rippled)

https://www.softwarecollections.org/en/docs/guide/#sect-What_Are_Software_Collections

For example one could enable this one to enable boost 1.69 on older versions of Fedora (pre-30)

https://www.softwarecollections.org/en/scls/denisarnaud/boost169/

> 
> At least most external dependencies are not directly vendored any more. It still could be a bit easier to provide external dependencies as binaries instead of having them fetched and compiled (sometimes statically) by cmake.

Not a fan of this one, eg introducing binary dependencies outside of standard build / platform / distribution channels.

mellery451:
@movitto sorry for the delay in responding. I'm going to study your changes here and get back to you soon.

I made a few changes to the package build in `b4` (see 73c6e47e8a) that would likely need to be merged with your changes, although I need to investigate this a bit more. I'm curious though if `b4` without your changes still has the original issues you encountered. I'm surprised the Dockerfile change hasn't been required by our systems, but it could be a version difference.

MarkusTeufelberger:
> Not a fan of this one, eg introducing binary dependencies outside of standard build / platform / distribution channels.

I meant "external dependencies as binaries" from the PoV of rippled/cmake, so e.g. providing a `librocksdb.so` (or `librocksdb.a`) from an RPM or DEB package instead of always pulling a repo and compiling RocksDB when compiling `rippled`.

MarkusTeufelberger:
>  I'm surprised the Dockerfile change hasn't been required by our systems, but it could be a version difference.

Docker 1.13.1 is about 2 years old, Fedora 28 is even already no longer supported by Fedora (https://fedoraproject.org/wiki/Releases#Current_Supported_Releases), so probably back then using build args in `Dockerfile`s was not that common or well supported.

mellery451:
@movitto I looked a little more closely at these changes. I'm fine with the `Dockerfile` change, but I think the spec file changes overlap considerably with 73c6e47e8a. I'd appreciate it if you can give https://github.com/ripple/rippled/releases/tag/1.2.4 a try and see if it resolves that particular `check-buildroot` issue. If not, I can work with you make necessary changes to the current spec file.

As for the notion of getting the rpm into EPEL or similar, if you have a specific set of requirements that would need to be met in order to do so, I'm happy to discuss it. As @MarkusTeufelberger points out, we prefer to depend on very specific (usually recent) versions of libraries and that makes distro packages generally untenable. If we link statically, though, does it really matter that we pull specific versions during the build? I understand, though, that this is not a great fit for distros that prohibit/discourage static linking.

Oh, @MarkusTeufelberger you'll the thrilled to know the as of the latest beta, we now require boost 1.70 (up from 1.67) - sorry for any hassle that might cause.

MarkusTeufelberger:
You require boost 1.70 or you can build with boost 1.70?

Because this reads like you oly require boost 1.67+:

https://github.com/ripple/rippled/blob/e83d367f49cceb4b5f28830fb1965a5978744450/CMakeLists.txt#L717

mellery451:
ah - you are correct. We had another change that would make it a hard requirement, but that's not merged yet (thus I was confused). Thanks for clarifying that.

MarkusTeufelberger:
> If we link statically, though, does it really matter that we pull specific versions during the build? I understand, though, that this is not a great fit for distros that prohibit/discourage static linking.

...which are basically all major Linux distros out there. I'm not even sure if there's any major distro that actually would encourage static linking or allow static binaries into their repositories, maybe Gentoo has some settings for this? Anyways, pulling in potentially already packaged dependencies automatically during a build process is usually a big no-no if you ever want to have people install `rippled` directly from a distro. You'll need to find someone like Alpine who then have to add patches on top of every one of your releases (currently ironically to make 1.2.4 build with a more modern boost version, I'm assuming because you are vendoring FindBoost.cmake): https://github.com/alpinelinux/aports/tree/master/community/rippled

Alpine seems to have a more relaxed approach regarding dependencies, I doubt it would be possible to sneak `rippled` into Debian or Fedora with that many external dependencies not exposed or statically linked (RocksDB, soci, lz4, snappy, sqlite, libarchive...) but it might be worth a try.

movitto:
> > I'm surprised the Dockerfile change hasn't been required by our systems, but it could be a version difference.
> 
> Docker 1.13.1 is about 2 years old, Fedora 28 is even already no longer supported by Fedora (https://fedoraproject.org/wiki/Releases#Current_Supported_Releases), so probably back then using build args in `Dockerfile`s was not that common or well supported.

Yes my system is a little overdue for an update, problem is Fedora releases are pretty frequent, so one has to update often! (then again I suppose that's a good thing!)

> @movitto I looked a little more closely at these changes. I'm fine with the `Dockerfile` change, but I think the spec file changes overlap considerably with [73c6e47](https://github.com/ripple/rippled/commit/73c6e47e8a56f643e7e2c5cb1b9fe6efd8d1294c). I'd appreciate it if you can give https://github.com/ripple/rippled/releases/tag/1.2.4 a try and see if it resolves that particular `check-buildroot` issue. If not, I can work with you make necessary changes to the current spec file.
> 

Cool will do, I'll also update my distro to Fedora 30 and give it a while from there / update this with the status when I know more.


> As for the notion of getting the rpm into EPEL or similar, if you have a specific set of requirements that would need to be met in order to do so, I'm happy to discuss it. As @MarkusTeufelberger points out, we prefer to depend on very specific (usually recent) versions of libraries and that makes distro packages generally untenable. If we link statically, though, does it really matter that we pull specific versions during the build? I understand, though, that this is not a great fit for distros that prohibit/discourage static linking.

There are quite a few guidelines to make a Fedora/EPEL compliant package, the overall guide can be found [here](https://docs.fedoraproject.org/en-US/packaging-guidelines/).

As you mentioned specific library versions & static deps would impose challenges so perhaps a logical first step would be to organize the CMakeLists.txt file a bit to allow configurable dependency resolution, eg pull if/rely on system packages if we can, else fall back to the vendored/static defaults as they stand now. Any thoughts on the merits of this?

movitto:
> > If we link statically, though, does it really matter that we pull specific versions during the build? I understand, though, that this is not a great fit for distros that prohibit/discourage static linking.
> 
> ...which are basically all major Linux distros out there. 

Correct, here is the Fedora guideline pertaining [linking to static libraries](https://docs.fedoraproject.org/en-US/packaging-guidelines/#_statically_linking_executables)

> Alpine seems to have a more relaxed approach regarding dependencies, I doubt it would be possible to sneak `rippled` into Debian or Fedora with that many external dependencies not exposed or statically linked (RocksDB, soci, lz4, snappy, sqlite, libarchive...) but it might be worth a try.

I doubt you'd get it through the review process and feel it'd be in bad taste to try. I'm sure if the communities are interested, we could find a way to get it in.

mellery451:
@movitto ok, that's an interesting idea and probably worthwhile. The main hassle is that not all of those libraries provide proper cmake configs, so we'll probably have to resort to local find modules for some of these. I've opened https://github.com/ripple/rippled/issues/2955 to track this. thanks.

movitto:
@MarkusTeufelberger @mellery451 Updated to Fedora 30 and gave the rpm build a try again from the latest develop head. Docker issue still present (F30 still ships dock 1.13.1) but the rpmbuild issue seemed to be resolved. Removed that bit from the patch, rebased, and pushed.

mellery451:
looks good  - thanks for this change.

PR #2921, opened 2019-04-29
[OPS-508] Adding recursive remove and clean to apt

jfryman:
This PR updates the apt steps to recursively remove any outstanding lists and run a `clean` directive for good measure.

PR #2950, opened 2019-05-21
Proposed 1.3.0-b4

manojsdoshi:
If merged this PR will: 

- close #2897 
- close #2937 
- close #2927 
- close #2930 
- close #2905 
- close #2914 
- close #2921 
- close #2933 
- close #2936  
- close #2865 
- close #2866


MarkusTeufelberger:
I don't think #2935 is part of this (and I don't think it should be).

PR #2952, opened 2019-05-27
Resolve Various FIXMEs & TODOs

movitto:
Tackling what seems to be some low hanging fruit in the codebase as I continue get acquainted with things. These few patches were the result for grepping for "FIXME" and "TODO" in the codebase and addressing the issues to my best understanding. If I misinterpreted something and / or there are style changes needed please let me know and I can make the updates!

### DIFF
--- a/src/ripple/app/consensus/RCLConsensus.cpp
+++ b/src/ripple/app/consensus/RCLConsensus.cpp
@@ -633,29 +633,31 @@ RCLConsensus::Adaptor::doAccept(
         JLOG(j_.info()) << "We closed at "
                         << closeTime.time_since_epoch().count();
         using usec64_t = std::chrono::duration<std::uint64_t>;
-        usec64_t closeTotal =
-            std::chrono::duration_cast<usec64_t>(closeTime.time_since_epoch());
+        std::vector<usec64_t> closeTimes;
         int closeCount = 1;
 
         for (auto const& p : rawCloseTimes.peers)
         {
-            // FIXME: Use median, not average
             JLOG(j_.info())
                 << std::to_string(p.second) << " time votes for "
                 << std::to_string(p.first.time_since_epoch().count());
+
             closeCount += p.second;
-            closeTotal += std::chrono::duration_cast<usec64_t>(
-                              p.first.time_since_epoch()) *
-                p.second;
+
+            auto time = std::chrono::duration_cast<usec64_t>(p.first.time_since_epoch());
+            for(int pi = 0; pi < p.second; ++pi)
+              closeTimes.push_back(time);
         }
 
-        closeTotal += usec64_t(closeCount / 2);  // for round to nearest
-        closeTotal /= closeCount;
+        auto our_time = std::chrono::duration_cast<usec64_t>(closeTime.time_since_epoch());
+        closeTimes.insert(std::upper_bound(closeTimes.begin(), closeTimes.end(), our_time), our_time);
 
         // Use signed times since we are subtracting
         using duration = std::chrono::duration<std::int32_t>;
         using time_point = std::chrono::time_point<NetClock, duration>;
-        auto offset = time_point{closeTotal} -
+
+        // Use median time
+        auto offset = time_point{closeTimes[closeTimes.size()/2]} -
             std::chrono::time_point_cast<duration>(closeTime);
         JLOG(j_.info()) << "Our close offset is estimated at " << offset.count()
                         << " (" << closeCount << ")";
--- a/src/ripple/protocol/json_options.h
+++ b/src/ripple/protocol/json_options.h
@@ -1,0 +1,11 @@
+#ifndef RIPPLE_RPC_JSON_OPTIONS_H
+#define RIPPLE_RPC_JSON_OPTIONS_H
+
+namespace ripple {
+  enum JsonOption {
+    none         = 0,
+    include_date = 1
+  };
+} // ripple
+
+#endif
--- a/src/ripple/app/ledger/AcceptedLedgerTx.cpp
+++ b/src/ripple/app/ledger/AcceptedLedgerTx.cpp
@@ -77,11 +77,11 @@ std::string AcceptedLedgerTx::getEscMeta () const
 void AcceptedLedgerTx::buildJson ()
 {
     mJson = Json::objectValue;
-    mJson[jss::transaction] = mTxn->getJson (0);
+    mJson[jss::transaction] = mTxn->getJson (JsonOption::none);
 
     if (mMeta)
     {
-        mJson[jss::meta] = mMeta->getJson (0);
+        mJson[jss::meta] = mMeta->getJson (JsonOption::none);
         mJson[jss::raw_meta] = strHex (mRawMeta);
     }
 
--- a/src/ripple/app/ledger/Ledger.cpp
+++ b/src/ripple/app/ledger/Ledger.cpp
@@ -928,7 +928,7 @@ static bool saveValidatedLedger (
                     << "Transaction in ledger " << seq
                     << " affects no accounts";
                 JLOG (j.warn())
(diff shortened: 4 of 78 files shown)


movitto:
Even though this has conflicts, if possible I'd like to get some feedback before addressing those since the last patch is rather large and touches alot of files (eg keeping it in sync with the latest develop HEAD will take some work). Thanks!

thejohnfreeman:
I forgot to say: thank you for contributing! I got a little swallowed up in the review, but every new contributor we get is very exciting. Welcome :)

scottschurr:
I'll second @thejohnfreeman; thanks for the contribution @movitto, and welcome!

I agree with @thejohnfreeman about separating the commits into separate pull requests.  My inclination would be to divide these three commits into three different pull requests.  As far as I can tell they are independent of one another.  Making them separate pull requests allows us to argue the points (What? Us? Argue?) independently.  And that way if one gets stalled for some reason (as @thejohnfreeman noted, good consensus is fragile) the others can proceed on their way into the code base.

I am _really_ excited to see the commit that makes the argument to `STBase::getJson()` an enum!  That's annoyed me for the longest time, and I never got around to addressing it.  Now, out of the blue, here it is!  Done!  Thanks a ton.

Since that's the commit which is close to my heart, I'll offer a few minor suggestions:

1. The `getJson(int)` interface you're fixing predates my presence in the code base.  Having said that, I don't believe that the original interface was intended to be used as a bit mask.  I think they were hackish integers.  The `enum` approach you've taken looks appropriate to me given the history I've seen.  If we made them bit masks we'd run the risk of people actually introducing bit masks which at this point I believe is complexity we don't need.

In summary, I like your `enum`.  Let's keep it.

2. There are a few (unrelated) `getJson()` methods in the code base that take an integer as a level (yet a different mystery).  To reduce the likelihood of folks getting confused about that, I suggest moving the definition of your `enum` from its stand alone file into the top of STBase.h.  STBase.h is a good place for your `enum` definition, because your `enum` is really a part of the `STBase` interface.  At that point we'll remove the newly introduced stand alone file, since it no longer contains anything useful.

3. I suggest you make your `enum` an `enum class`.  Two reasons:

- The term `JsonOption::none` is really expressive in the source code.  The reader is not left wondering about what that thing means.  By making the `enum` an `enum class` you _require_ that future maintainers keep the text `JsonOption` in those calls.

- The term `none` is a really useful term in a lot of circumstances.  If you put `none` in a standard `enum`, then the name `none` leaks out into the entire `ripple` namespace.  That may make it harder for other people to use the term in their own locality.  If you make your `enum` an `enum class`, then the `none` stays trapped inside the `enum`.  Folks outside your `enum` are less likely to run into problems if they also try to use the term `none`.

4. On the very picky front, we typically indent four spaces in this code base.  I suggest you use the same convention in the declaration of your `enum`.

5. On a similar, very picky, note, we usually keep line lengths under 80 characters (with some notable exceptions for data tables).  Since `JsonOption::none` has more characters than `0`, just a few line lengths we're pushed past 80 characters by your change.  It would be great to wrap those lines back within 80 characters.

There are those who argue that line lengths are no longer relevant.  But I find that keeping line lengths to 80 characters helps a lot when I'm doing side-by-side code comparisons.  So I think the 80 character guidance still has relevance even in today's world.

Thanks again for the contribution!

src/ripple/app/consensus/RCLConsensus.cpp
```diff
@@ -633,29 +633,31 @@ RCLConsensus::Adaptor::doAccept(
...
                 << std::to_string(p.first.time_since_epoch().count());
+
             closeCount += p.second;
-            closeTotal += std::chrono::duration_cast<usec64_t>(
-                              p.first.time_since_epoch()) *
-                p.second;
+
+            auto time = std::chrono::duration_cast<usec64_t>(p.first.time_since_epoch());
+            for(int pi = 0; pi < p.second; ++pi)
+              closeTimes.push_back(time);
```
nbougalis: Potential for improving readability/comprehension by replacing this loop with:

    closeTimes.insert(closeTimes.end(), p.second, p.first.time_since_epoch());

src/ripple/app/consensus/RCLConsensus.cpp
```diff
@@ -633,29 +633,31 @@ RCLConsensus::Adaptor::doAccept(
...
+        auto our_time = std::chrono::duration_cast<usec64_t>(closeTime.time_since_epoch());
+        closeTimes.insert(std::upper_bound(closeTimes.begin(), closeTimes.end(), our_time), our_time);
 
         // Use signed times since we are subtracting
         using duration = std::chrono::duration<std::int32_t>;
         using time_point = std::chrono::time_point<NetClock, duration>;
-        auto offset = time_point{closeTotal} -
+
+        // Use median time
+        auto offset = time_point{closeTimes[closeTimes.size()/2]} -
```
nbougalis: This change will, unfortunately, make the code more flakey in a very subtle and hard to detect way: it relies on an implementation detail (namely that `rawCloseTimes.peers` is an ordered container) without documenting this.

I remain unconvinced that the median (especially the weighted median) is the right metric here. Why not use a simpler and more easily understood metric of central tendency? For example by picking the most popular close time and resolving conflicts by the earliest time. While in theory such a metric could be less robust than the median, I think in practice it would be fine, especially if we improve the dampening/hysteresis in `adjustCloseTime`.

Pinging Messrs. @HowardHinnant and @JoelKatz.

src/ripple/protocol/json_options.h
```diff
@@ -0,0 +1,11 @@
+#ifndef RIPPLE_RPC_JSON_OPTIONS_H
+#define RIPPLE_RPC_JSON_OPTIONS_H
+
+namespace ripple {
+  enum JsonOption {
```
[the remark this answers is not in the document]
nbougalis: In the past, we mostly assumed the existence of "normal" and "verbose" markers, but I don't think we should be constrained by the past.

In fact, I think that this change, coupled with your suggestion for more "descriptive" flags makes a lot of sense.

movitto:
Awesome, thanks for the feedback @thejohnfreeman and @scottschurr 

All the above comments above make sense, I will incorporate the changes into separate PRs and close this one.

src/ripple/app/consensus/RCLConsensus.cpp
movitto: Hey @nbougalis, thanks for the feedback as well, I hit reply before I saw your comment. I'm curious as to more details to the "most popular" close time idea, would this entail selecting the time with >= 51% of the votes, or simply the time w/ the highest percentage of votes (and would there be a minimum threshold enforced, eg what if the most popular option only got 25% of the votes).

bachase: @movitto Thanks for looking at this. I agree with Nik that median may not be the best for this purpose and is probably better to do holistically by considering how we adjust the network clock based on this distribution of peer close times. I suspect that in practice, median and mean and almost anything else aren't terribly different, but it would be useful to see what the distribution does look like.

If that is out of scope for this PR, I'd be vote removing the FIXME in the code and opening an issue instead to track this.

movitto: @nbougalis @bachase Hey Nik, Brad sorry this past week has been hectic, haven't been able to get to it yet, but I am planning on following up on this ASAP!

movitto:
OK, created an issue (#2958) to discuss concensus close time and moved the 2nd patch to #2932 targed towards cleaning up some old comments.

Will get to rebasing & updating the last patch (JSONOptions enum) and sending a seperate PR soon (today if possible, else early this week). I will close this PR after that.

Thanks again to everyone for the great feedback!

movitto:
OK split the last patch into a separate PR & verified it works, closing this one.

PR #2967, opened 2019-06-07
Add a reset code for the regular keys

cryptobrad:
The wallets provider (like Gatehub) stores the wallet seed or master private key in database, and encrypt  using the account password so wallet provider cannot access the wallet without the user loging in. This is good but also it is the problem as Gatehub hack shows this can mean that wallet provider is not able to act to protect users until users login.

If providers used a regular private key to sign better uses security could be possible by having a reset code that is in the cold storage. That way if the database is taken the provider can use the reset code form the cold storage to remove the regular key without having the wait for the user. That way the provider loses the access but they users still has the regular key and can use the account without  problem.

I add a reset reguarl key code using the crypto-condition PreimageSha256 which can be used to remove a regular key from the wallet.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2075,6 +2075,7 @@ else ()
     src/ripple/app/tx/impl/OfferStream.cpp
     src/ripple/app/tx/impl/PayChan.cpp
     src/ripple/app/tx/impl/Payment.cpp
+    src/ripple/app/tx/impl/ResetRegularKey.cpp
     src/ripple/app/tx/impl/SetAccount.cpp
     src/ripple/app/tx/impl/SetRegularKey.cpp
     src/ripple/app/tx/impl/SetSignerList.cpp
--- a/src/ripple/app/tx/impl/ResetRegularKey.cpp
+++ b/src/ripple/app/tx/impl/ResetRegularKey.cpp
@@ -1,0 +1,80 @@
+//------------------------------------------------------------------------------
+/*
+    This file is part of rippled: https://github.com/ripple/rippled
+    Copyright (c) 2019, Ripple Labs Inc.
+
+    Permission to use, copy, modify, and/or distribute this software for any
+    purpose  with  or without fee is hereby granted, provided that the above
+    copyright notice and this permission notice appear in all copies.
+
+    THE  SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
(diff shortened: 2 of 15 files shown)


MarkusTeufelberger:
Would/Could/Should a "reset token" be linked to a specific regular key? So If I have a master key, a regular key "1" and then switch to regular key "2", could the wallet provider kick out the second one at any time?

Also imagine this in practice: Gatehub probably has at least hundreds, if not thousands of customers. As far as I understand the current implementation (might be nicer to have some design discussion beforehand...), it would require them to send one transaction for each customer. It might be better to have a "reset my RegularKey if the following secret gets revealed" field instead on AccountRoot objects that can only be controlled through the MasterKey?

I'm also not sure how/if this should work with a disabled MasterKey or MultiSig.

In general giving someone the ability to revoke secrets might be useful, though I'm a bit sceptical of this approach. The design for Gatehub's secret storage is by the way not new, it existed with the blobstore for the better part of a decade by now and was originally designed by Ripple.

nbougalis:
This is certainly an interesting PR, however I believe there is a better way to accomplish this _and more_ right now by using multisigning capabilities and without any changes to the ledger:

To begin, a wallet provider creates and funds a new wallet which becomes their signing address **A**. They then [configure a new regular key for **A**](https://developers.ripple.com/setregularkey.html) and [disable the master key](https://developers.ripple.com/accountset.html).

For every wallet they want to be able to sign for using that address, they simply need to [configure a 1-of-1 multisign list](https://developers.ripple.com/set-up-multi-signing.html), containing the address **A**. They can then sign (technically multi-sign) transactions as **A**.

Here's the neat part: since the master key for **A** is disabled, they only way they can sign as **A** is using the regular key associated with **A**—which is, of course, changeable.

So if the provider comes to suspect that the regular key for **A** is compromised, they can change it and further signatures with that regular key will be rejected. The provider could even rotate the key on a regular basis.

The regular key need not be a wallet that's funded on the ledger. It can just be a keypair.

scottschurr:
FWIW, all of the non-unity builds are failing.  It looks like SetAccount.cpp is missing an include:
```
#include <ripple/conditions/Condition.h>
```
I'm unsure whether it's worth fixing the include since @nbougalis has proposed a different approach (which requires no code changes) to solving the problem you're identifying.

nbougalis:
I think we should close this PR: it has been inactive for over a month and, based on the comments here, there are alternative mechanisms to achieve substantially the same objectives without requiring code changes.

If you think closing this is premature or that it requires further consideration, please ping this issue.

PR #2977, opened 2019-06-18
Enhance AccountTx unit test

scottschurr:
While working on tickets I found that the account_tx RPC test did not examine all possible returned transaction types.  So I'm adding that test.

Additionally, it seems that there are enough unit tests exercising Checks that it is worth moving jtx interfaces for check handling into a central location.

### DIFF
--- a/src/test/rpc/AccountTx_test.cpp
+++ b/src/test/rpc/AccountTx_test.cpp
@@ -26,7 +26,7 @@ namespace ripple {
 
 namespace test {
 
-class AccountTX_test : public beast::unit_test::suite
+class AccountTx_test : public beast::unit_test::suite
 {
     void
     testParameters()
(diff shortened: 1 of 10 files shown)


src/test/rpc/AccountTx_test.cpp
```diff
@@ -164,15 +164,299 @@ class AccountTX_test : public beast::unit_test::suite
...
+            { 15, jss::TrustSet,             {jss::RippleState, jss::DirectoryNode, jss::DirectoryNode}, {},                               {jss::AccountRoot, jss::AccountRoot}},
+            { 16, jss::SetRegularKey,        {},                                                         {},                               {jss::AccountRoot}},
+            { 17, jss::Payment,              {},                                                         {},                               {jss::AccountRoot, jss::AccountRoot}},
+            { 18, jss::AccountSet,           {},                                                         {},                               {jss::AccountRoot}},
+            { 19, jss::AccountSet,           {},                                                         {},                               {jss::AccountRoot}},
+            { 20, jss::Payment,              {jss::AccountRoot},                                         {},                               {jss::AccountRoot}},
+        };
+
+        BEAST_EXPECT (std::extent<decltype (sanity)>::value ==
```
mellery451: L340 sort of overlaps with this check...I think I like this one better since it verifies against the sanity data

scottschurr: Good point.  Would you prefer that I remove the check on line 340?  Or is it okay to leave it?

mellery451: I'm weakly in favor or removing L340, but I'm fine either way.

scottschurr:
Rebased to 1.3.0-b6.

scottschurr:
Squashed and rebased to 1.3.0.

PR #2978, opened 2019-06-18
Remove SQLite Validations table

miguelportilla:
This PR is a small step in preparation for self-contained portable shards. The `ledger.Validations` table and related indexes are removed entirely. The Validations table stores observed validations but the data is not used by rippled. The same information can be gathered using a WebSocket subscription.


scottschurr:
@miguelportilla, regarding verifying "... an existing ledger database is properly modified and the new validation.Validations table is populated", what methods would you suggest for doing this verification on macOS?

Also, just from a curiosity stand point, what's the backwards compatibility story for this change?  What will happen if I...

1. Run a new server (with this change) that migrates the database,
2. Shut that server down, and then
3. Use an older server version without wiping the database.

I suspect it's not pretty, and that may be inevitable.  But we should have some idea what to expect.  Thanks.

miguelportilla:
@scottschurr I recommend inspecting the tables before and after on the ledger and validation databases. The number of table records should be the same after migration. The ledger database should be absent of the indexes and the Validations table etc... You can do that using a tool like sqlitebrowser.org (Mac OS is supported), which is easy to use.

Although I haven't tested it, running an older rippled version after this branch edition should work fine. The older rippled version will ignore the new validation database and create a new Validations table and indexes inside the ledger database.

@scottschurr @thejohnfreeman @mtrippled  Please halt review of this PR until further notice. After some discussion with David and Nik, I will be pushing a commit that removes the Validations table entirely from the project.

scottschurr:
Looks like the build failure, at least in part, is due to this line: https://github.com/ripple/rippled/blob/828b531764e92bc0591ba3e67600e5f3a6b0d031/src/test/consensus/Validations_test.cpp#L740
Clang thinks that's an unused variable.  Remove that line and it builds for me.

Beyond that, I see one unit test failure in `SHAMapStore clearPrior`
```
#5 failed: unhandled exception: sqlite3_statement_backend::prepare: no such table: Validations while preparing "SELECT count(*) AS rows FROM Validations;"
```
Hope that helps.

mellery451:
can `src/test/csf/Peer.h|148 col 9| onStale(Validation&& v)` be removed too?

src/ripple/app/main/Application.cpp
```diff
@@ -824,24 +824,145 @@ class ApplicationImp
...
+
+            if (!setup.standAlone ||
+                setup.startUp == Config::LOAD ||
+                setup.startUp == Config::LOAD_FILE ||
+                setup.startUp == Config::REPLAY)
+            {
+                // perform any needed table updates
+                updateTxnDB();
+
+                // Check if AccountTransactions has primary key
```
mtrippled: Why does it matter if AccountTransactions doesn't have a primary key? Why specifically check for this schema characteristic?

src/ripple/app/main/Application.cpp
```diff
@@ -824,24 +824,145 @@ class ApplicationImp
     beast::Journal journal (std::string const& name) override;
 
     //--------------------------------------------------------------------------
-    bool initSqliteDbs ()
+
+    bool
+    initDBs()
```
mtrippled: Conceptually, I see SQLite and nodestore databases as distinct. My preference is to have separate functions called within Application:setup() that each pertain to initializing/tuning these things accordingly. I agree that it's a good idea to move these activities from the body of setup(), and I think that it would make it even easier to reason about if there were 2 functions, such as initSqliteDbs() for SQL, and initNodestoreDbs() for nodestore stuff.

src/ripple/app/main/DBInit.cpp
```diff
@@ -92,30 +92,18 @@ const char* LedgerDBInit[] =
...
-    "CREATE INDEX IF NOT EXISTS ValidationsByInitialSeq ON              \
-        Validations(InitialSeq, LedgerSeq);",
-    "CREATE INDEX IF NOT EXISTS ValidationsByTime ON              \
-        Validations(SignTime);",
+    // Old table and indexes no longer needed
+    "DROP TABLE IF EXISTS Validations;",
+    "DROP INDEX IF EXISTS ValidationsByHash;",
+    "DROP INDEX IF EXISTS ValidationsBySeq;",
+    "DROP INDEX IF EXISTS ValidationsByInitialSeq;",
+    "DROP INDEX IF EXISTS ValidationsByTime;",
```
mtrippled: DROP INDEXes are redundant here. Table objects such as indices go away once the table goes away: https://www.sqlite.org/lang_droptable.html

src/ripple/app/main/Application.cpp
miguelportilla: I don't know the reason, its what rippled currently does.

MarkusTeufelberger: It would be probably a bit expensive to list all transactions by the Bitstamp issuing account without an index I guess.

miguelportilla:
@scottschurr @mtrippled @thejohnfreeman The PR is ready for review, thanks.

scottschurr:
FWIW, I just remembered to try the experiment of running this branch and then running the current develop.

There is an interesting artifact to be aware of.

1. After running a rippled server from this branch I gracefully shutdown that server.
2. Without changing any of the database I started a rippled server using 1.3.0-b6.  It started fine.
3. I let the server achieve sync.
4. After syncing I requested a graceful shutdown of 1.3.0-b6 and got this message:
```
libc++abi.dylib: terminating with uncaught exception of type boost::system::system_error: No such file or directory
Abort trap: 6
```
I only saw this symptom once -- the first time I tried it.  I attempted unsuccessfully to reproduce the symptom three additional times.  So I'm unsure anything needs to be done.  But you should be aware.

src/ripple/app/main/Application.cpp
```diff
@@ -824,24 +824,149 @@ class ApplicationImp
...
-        DatabaseCon::Setup setup = setup_DatabaseCon (*config_);
-        mTxnDB = std::make_unique <DatabaseCon> (setup, TxnDBName,
-                TxnDBInit, TxnDBCount);
-        mLedgerDB = std::make_unique <DatabaseCon> (setup, "ledger.db",
-                LedgerDBInit, LedgerDBCount);
-        mWalletDB = std::make_unique <DatabaseCon> (setup, "wallet.db",
-                WalletDBInit, WalletDBCount);
+        try
+        {
+            auto setup = setup_DatabaseCon(*config_);
```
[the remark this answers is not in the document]
miguelportilla: It can and will be. 👍

src/ripple/app/main/Application.cpp
```diff
@@ -824,24 +824,149 @@ class ApplicationImp
...
+                setup,
+                TxnDBName,
+                TxnDBInit,
+                TxnDBCount);
+            mTxnDB->getSession() <<
+                boost::str(boost::format("PRAGMA cache_size=-%d;") %
+                (config_->getSize(siTxnDBCache) * kilobytes(1)));
+            mTxnDB->setupCheckpointing(m_jobQueue.get(), logs());
+
+            if (!setup.standAlone ||
```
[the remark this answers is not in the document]
miguelportilla: The condition determines if temporary files won't be used. The checks and updates are applied only to an existing database as a new one meets the criteria. Opting out earlier is a tiny optimization.

src/ripple/app/main/Application.cpp
```diff
@@ -1978,49 +2061,36 @@ ApplicationImp::journal (std::string const& name)
     return logs_->journal (name);
 }
 
-//VFALCO TODO clean this up since it is just a file holding a single member function definition
-
 static
```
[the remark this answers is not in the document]
miguelportilla: I like narrowing the scope, will do, thanks. 👍

miguelportilla:
@mtrippled I think your suggested log changes to `RCLValidations.cpp` are outside this scope and may be better suited for another PR.

scottschurr:
I like the lambda.  Still :+1: from me.

mtrippled:
Oh, the commit message should be "Validations table" not "validation database".

ximinez:
@miguelportilla Can you rebase this on to the latest develop? I'm having trouble building it, and I think it's more because of the older dependencies.

miguelportilla:
Thanks for the feedback and review! @scottschurr @mtrippled @thejohnfreeman @ximinez

nbougalis:
Incorporated in #2995.

PR #2981, opened 2019-06-21
Make protocol message counters more granular:

nbougalis:
A running instance of the server tracks the number of protocol messages and the number of bytes it sends and receives.

This commit makes the counters more granular, allowing server operators to better track and understand bandwidth usage.

### DIFF
--- a/src/ripple/overlay/impl/TrafficCount.cpp
+++ b/src/ripple/overlay/impl/TrafficCount.cpp
@@ -21,55 +21,6 @@
 
 namespace ripple {
 
-const char* TrafficCount::getName (category c)
-{
-    switch (c)
-    {
-    case category::base:              return "overhead";
-    case category::cluster:           return "cluster management";
-    case category::overlay:           return "overlay management";
-    case category::manifests:         return "manifest management";
-    case category::transaction:       return "transactions";
-    case category::proposal:          return "proposals management";
-    case category::validation:        return "validations";
-    case category::shards:            return "shards";
-    case category::get_set:           return "set management (get)";
-    case category::share_set:         return "set management(share)";
-    case category::ld_tsc_get:        return "ledger data: Transaction Set candidate (get)";
-    case category::ld_tsc_share:      return "ledger data: Transaction Set candidate (share)";
-    case category::ld_txn_get:        return "ledger data: Transaction Node (get)";
-    case category::ld_txn_share:      return "ledger data: Transaction Node (share)";
-    case category::ld_asn_get:        return "ledger data: Account State Node (get)";
-    case category::ld_asn_share:      return "ledger data: Account State Node (share)";
-    case category::ld_get:            return "ledger data (get)";
-    case category::ld_share:          return "ledger data (share)";
-    case category::gl_tsc_share:      return "ledger: Transaction Set candidate (share)";
-    case category::gl_tsc_get:        return "ledger: Transaction Set candidate (get)";
-    case category::gl_txn_share:      return "ledger: Transaction node (share)";
-    case category::gl_txn_get:        return "ledger: Transaction node (get)";
-    case category::gl_asn_share:      return "ledger: Account State node (share)";
-    case category::gl_asn_get:        return "ledger: Account State node (get)";
-    case category::gl_share:          return "ledger (share)";
-    case category::gl_get:            return "ledger (get)";
-    case category::share_hash_ledger: return "getobject: Ledger (share)";
-    case category::get_hash_ledger:   return "getobject: Ledger (get)";
-    case category::share_hash_tx:     return "getobject: Transaction (share)";
-    case category::get_hash_tx:       return "getobject: Transaction (get)";
-    case category::share_hash_txnode: return "getobject: Transaction node (share)";
-    case category::get_hash_txnode:   return "getobject: Transaction node (get)";
-    case category::share_hash_asnode: return "getobject: Account State node (share)";
-    case category::get_hash_asnode:   return "getobject: Account State node (get)";
-    case category::share_cas_object:  return "getobject: CAS (share)";
-    case category::get_cas_object:    return "getobject: CAS (get)";
-    case category::share_fetch_pack:  return "getobject: Fetch Pack (share)";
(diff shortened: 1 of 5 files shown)


src/ripple/overlay/impl/TrafficCount.cpp
```diff
@@ -25,30 +25,48 @@ const char* TrafficCount::getName (category c)
...
+    case category::base:              return "overhead";
+    case category::cluster:           return "cluster management";
+    case category::overlay:           return "overlay management";
+    case category::manifests:         return "manifest management";
+    case category::transaction:       return "transactions";
+    case category::proposal:          return "proposals management";
+    case category::validation:        return "validations";
+    case category::shards:            return "shards";
+    case category::get_set:           return "set management (get)";
+    case category::share_set:         return "set management(share)";
```
jwbusch: nit: `set management (share)` (add space after `management`)

src/ripple/overlay/impl/TrafficCount.cpp
```diff
@@ -25,30 +25,48 @@ const char* TrafficCount::getName (category c)
...
-        case category::CT_get_ledger:
-            return "ledger_get";
-        case category::CT_share_ledger:
-            return "ledger_share";
-        case category::CT_get_trans:
-            return "transaction_set_get";
-        case category::CT_share_trans:
-            return "transaction_set_share";
-        case category::CT_unknown:
-            assert (false);
```
thejohnfreeman: Removed assertions?

src/ripple/overlay/impl/TrafficCount.cpp
```diff
@@ -58,81 +76,122 @@ TrafficCount::category TrafficCount::categorize (
...
+    if (auto msg = dynamic_cast<protocol::TMGetObjectByHash const*>(&message))
     {
-        auto msg =
-            dynamic_cast <protocol::TMGetObjectByHash const*>
-                (&message);
-        if (msg)
-        {
-            // inbound queries and outbound responses are sharing
-            // outbound queries and inbound responses are getting
+        if (msg->type() == protocol::TMGetObjectByHash::otLEDGER)
```
jwbusch: Another way to go about this is pull these into a loop... makes it clear that the logic is identical for every case.

You could also cache the `bool const isInbound = msg->query() == inbound` outside the loop to avoid indirecting through the `msg` pointer for every check.

```
struct {
    message_type_enum message_type;
    TrafficCount::category share;
    TrafficCount::category get;
} checks[] = {
    { protocol::TMGetObjectByHash::otLEDGER,      TrafficCount::category::share_hash_ledger, TrafficCount::category::get_hash_ledger },
    { protocol::TMGetObjectByHash::otTRANSACTION, TrafficCount::category::share_hash_tx,     TrafficCount::category::get_hash_tx     },
    // etc. 
};

for (auto& check : checks) {
    if (msg->query() == check.message_type) {
        return (msg->query() == inbound) ? check.share : check.get;
    }
}
```

src/ripple/overlay/impl/TrafficCount.h
```diff
@@ -122,9 +183,8 @@ class TrafficCount
         return ret;
     }
 
-    protected:
-
-    std::map <category, TrafficStats> counts_;
+protected:
+    std::map<category, TrafficStats> counts_;
```
jwbusch: Given that `category` is an enum with default values, could we use a `std::vector/std::array` instead and use the category as the index? Avoids a tree traversal/(potentially) fragmented memory. 

i.e. add a `count` member at the end of the enum class (after `unknown`), and use that to determine the size of the array/vector.

src/ripple/overlay/impl/TrafficCount.h
```diff
@@ -122,9 +183,8 @@ class TrafficCount
         return ret;
```
thejohnfreeman: I can't leave a comment on lines 179-180, but I'm pretty sure `std::piecewise_construct` and `std::forward_as_tuple` are not required here because we're only passing one argument and there are no ambiguous constructors. I'm guessing someone just copied the pattern on cppreference.com without thinking too much.

Further, now that we have `std::atomic`s in `TrafficStats`, do we want to construct copies of those for each category, or are raw `std::uint64_t`s good enough? I don't know much about the cost of `std::atomic`.

I'm guessing using `std::atomic` implies this class wants to be thread-safe, but there is a race condition here: interleaving the copies of related values (e.g. bytes in + messages in) on line 180 with updates to those values in lines 150-156. So there is a small chance that a small discrepancy can arise, but since these are just diagnostics, I'm guessing that potential error can be tolerated.

src/ripple/overlay/impl/TrafficCount.cpp
thejohnfreeman: Depending on the compiler, it might be able to consolidate those common expressions, but I would just do it myself as a matter of style. Gives an opportunity to give a meaningful name to the value, and makes it easier to repeat (via autocompletion) compared to the full expression.

src/ripple/overlay/impl/TrafficCount.h
nbougalis: I like that, but I'm not sure I want to do that for this PR. Let me see how it would work.

src/ripple/overlay/impl/TrafficCount.h
jwbusch: They were `std::atomic` before (behind a typedef), so the behavior should be unchanged, just changed from `unsigned long` to `uint64_t`

thejohnfreeman: Sure, but that doesn't change the point. Anyone requesting a copy probably doesn't need atomics. `TrafficStats` could be a template instantiated for `uint64_t` and `std::atomic<uint64_t>`. `std::atomic<T>` has an `operator T` so you can even write a polymorphic copy constructor.

nbougalis: I think this is a fair point, @thejohnfreeman. I'd rather not make this change at this time, but it's something we should look at for the next round of improvements (and there are a few others too) we do.

PR #2994, opened 2019-07-01
Stabilize RPC error code values:

scottschurr:
The original intent was that RPC error codes were not stable.  But those codes were made available through the API, so some users came to depend on the code values.  This change adapts to the current state of affairs.

### DIFF
--- a/src/ripple/protocol/ErrorCodes.h
+++ b/src/ripple/protocol/ErrorCodes.h
@@ -27,82 +27,110 @@ namespace ripple {
 
 // VFALCO NOTE These are outside the RPC namespace
 
+// NOTE: Although the precise numeric values of these codes were never
+// intended to be stable, several API endpoints include the numeric values.
+// Some users came to rely on the values, meaning that renumbering would be
+// a breaking change for those users.
+//
+// We therefore treat the range of values as stable although they are not
+// and are subject to change.
+//
+// Please only append to this table. Do not "fill-in" gaps and do not re-use
+// or repurpose error code values.
 enum error_code_i
 {
-    rpcUNKNOWN = -1,     // Represents codes not listed in this enumeration
+    // -1 represents codes not listed in this enumeration
+    rpcUNKNOWN               = -1,
 
-    rpcSUCCESS = 0,
+    rpcSUCCESS               = 0,
 
-    rpcBAD_SYNTAX,  // Must be 1 to print usage to command line.
-    rpcJSON_RPC,
-    rpcFORBIDDEN,
-
-    // Error numbers beyond this line are not stable between versions.
-    // Programs should use error tokens.
+    rpcBAD_SYNTAX            = 1,
+    rpcJSON_RPC              = 2,
+    rpcFORBIDDEN             = 3,
 
     // Misc failure
-    rpcNO_PERMISSION,
-    rpcNO_EVENTS,
-    rpcTOO_BUSY,
-    rpcSLOW_DOWN,
-    rpcHIGH_FEE,
-    rpcNOT_ENABLED,
-    rpcNOT_READY,
-    rpcAMENDMENT_BLOCKED,
+    // unused                  4,
+    // unused                  5,
+    rpcNO_PERMISSION         = 6,
+    rpcNO_EVENTS             = 7,
+    // unused                  8,
+    rpcTOO_BUSY              = 9,
+    rpcSLOW_DOWN             = 10,
+    rpcHIGH_FEE              = 11,
+    rpcNOT_ENABLED           = 12,
+    rpcNOT_READY             = 13,
+    rpcAMENDMENT_BLOCKED     = 14,
 
     // Networking
-    rpcNO_CLOSED,
-    rpcNO_CURRENT,
-    rpcNO_NETWORK,
+    rpcNO_CLOSED             = 15,
+    rpcNO_CURRENT            = 16,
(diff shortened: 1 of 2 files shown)


src/ripple/protocol/ErrorCodes.h
```diff
@@ -27,82 +27,110 @@ namespace ripple {
...
-    rpcNO_EVENTS,
-    rpcTOO_BUSY,
-    rpcSLOW_DOWN,
-    rpcHIGH_FEE,
-    rpcNOT_ENABLED,
-    rpcNOT_READY,
-    rpcAMENDMENT_BLOCKED,
+    // unused                  4,
+    // unused                  5,
+    rpcNO_PERMISSION         = 6,
```
jwbusch: So users were expecting `rpcNO_PERMISSION` to be 6, but we were returning 4? Just want to clarify that this change is to fix an issue that was caused by adding/removing error codes.

src/ripple/protocol/impl/ErrorCodes.cpp
```diff
@@ -136,19 +136,31 @@ sortErrorInfos (ErrorInfo const (&unordered)[N]) -> ErrorInfoArray<N>
 
     // Verify that all entries are filled in starting with 1 and proceeding
     // to rpcLAST.
-    int expect = 0;
+    //
+    // It's okay for there to be missing entries; they will contain the code
+    // rpcUNKNOWN.  But other than that all entries should match their index.
+    int codeCount {0};
+    int expect {rpcBAD_SYNTAX - 1};
```
jwbusch: You could add an `rpcFIRST = rpcBAD_SYNTAX` and use that to initialize `expect`, similar to how rpcLAST is used to check bounds above, it could be added there too instead to become `if (info.code < rpcFIRST || info.code > rpcLAST)`

It still has the oddity of the - 1, but I see why it had to be that way.

src/ripple/protocol/ErrorCodes.h
scottschurr: That's basically correct.  The specific complaint was that the error code for `actNotFound` changed from 19 to 15.  But the machinery is identical.

The reason that change occurred is that In a previous commit I had removed a bunch of RPC error messages that were no longer being used.  That's what caused the issue.  To recover I pulled up an old version of rippled and inserted `// unused` comments where the old errors and their codes had been.  The enum values now match the values found in this commit: https://github.com/ripple/rippled/commit/36d6758945954fab757b024b2afc8c3a5ad76919

src/ripple/protocol/impl/ErrorCodes.cpp
scottschurr: I could.  `rpcBAD_SYNTAX` was _supposed_ to be another magic numeric code (like `rpcUNKNOWN`), so it's value was supposed to be known not to change.  But, as @ximinez recently pointed out to me, that magic is not working at the moment.

Still, I'm not very worried about the value of `rpcBAD_SYNTAX` changing in the future.  I do expect `rpcLAST` to be a moving target, so I think it's important to have that alias to the actual last numeric code.

If you feel strongly about adding `rpcFIRST` I can do that, but my inclination is not to do so.  In my experience alias proliferation can lead to confusion, although I don't think that specific alias would be particularly confusing.

jwbusch: Nah, no strong feelings, I just like it when things are symmetrical.

src/ripple/protocol/ErrorCodes.h
```diff
@@ -27,82 +27,110 @@ namespace ripple {
...
-    rpcSRC_ACT_NOT_FOUND,
-    rpcSRC_CUR_MALFORMED,
-    rpcSRC_ISR_MALFORMED,
-    rpcSTREAM_MALFORMED,
-    rpcATX_DEPRECATED,
+    rpcACT_BITCOIN           = 34,
+    rpcACT_MALFORMED         = 35,
+    rpcALREADY_MULTISIG      = 36,
+    rpcALREADY_SINGLE_SIG    = 37,
+    rpcBAD_FEATURE           = 38,
```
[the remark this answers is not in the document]
jwbusch: Basically, should be -> `rpcBAD_FEATURE = 40,` 

I double checked again after Ed found this, and the rest seem correct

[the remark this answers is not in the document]
scottschurr: Thanks gents!  Good spotting @ximinez!.  I have pushed a `[FOLD]` commit which I believe addresses the problem.  The earlier spot checking I'd done was not sufficient to catch a problem of this nature.  My error.

src/ripple/protocol/ErrorCodes.h
```diff
@@ -91,9 +91,9 @@ enum error_code_i
     rpcACT_MALFORMED         = 35,
     rpcALREADY_MULTISIG      = 36,
     rpcALREADY_SINGLE_SIG    = 37,
-    rpcBAD_FEATURE           = 38,
-    // unused                  39,
-    // unused                  40,
+    // unused                  38,
+    // unused                  49,
```
jwbusch: should be 39

scottschurr: <Slaps Forehead> Doh!  Thanks.  Another `[FOLD]` coming soon...

scottschurr:
Squashed.  Thanks everybody!

PR #2995, opened 2019-07-03
Move shard store init to Application

miguelportilla:
This PR is a small step in preparation for self-contained portable shards. The initialization of the shard store is moved to the Application object. Once the shard store relieves the node store of duty, there will be no need for the SHAMapStore class and will be eventually removed.


src/ripple/app/main/Application.cpp
```diff
@@ -824,24 +835,149 @@ class ApplicationImp
...
         assert (mWalletDB.get () == nullptr);
 
-        DatabaseCon::Setup setup = setup_DatabaseCon (*config_);
-        mTxnDB = std::make_unique <DatabaseCon> (setup, TxnDBName,
-                TxnDBInit, TxnDBCount);
-        mLedgerDB = std::make_unique <DatabaseCon> (setup, "ledger.db",
-                LedgerDBInit, LedgerDBCount);
-        mWalletDB = std::make_unique <DatabaseCon> (setup, "wallet.db",
-                WalletDBInit, WalletDBCount);
+        try
```
jwbusch: nit: could make this a function scope `try-catch`, little cleaner

```
bool
initSQLiteDBs() try 
{
//...
} 
catch(std::exception const& ex)
{}
```

src/ripple/app/main/Application.cpp
```diff
@@ -1978,49 +2079,31 @@ ApplicationImp::journal (std::string const& name)
...
-        JLOG (j.fatal()) << "Schema for " << dbName << " has too few lines";
-        Throw<std::runtime_error> ("bad schema");
-    }
+        return false;
+    };
 
-    return schema[line].find (content) != std::string::npos;
-}
+    assert(schemaHas("TransID"));
+    assert(!schemaHas("foobar"));
```
jwbusch: Did you want to keep the `foobar` assert?

src/ripple/nodestore/impl/DatabaseShardImp.cpp
```diff
@@ -34,64 +35,165 @@ namespace NodeStore {
...
     , app_(app)
-    , config_(config)
-    , dir_(get<std::string>(config, "path"))
+    , ctx_(std::make_unique<nudb::context>())
+    , config_(app.config().section(ConfigSection::shardDatabase()))
+    , dir_(get<std::string>(config_, "path"))
     , backendName_(Manager::instance().find(
-        get<std::string>(config_, "type"))->getName())
-    , maxDiskSpace_(get<std::uint64_t>(config, "max_size_gb") << 30)
+        get<std::string>(config_, "type", "nudb"))->getName())
```
jwbusch: Could `Manager::instance().find` return a nullptr (and crash on accessing `getName`)? We could just insert the string directly from the config and let `init` handle the error.

src/ripple/app/main/Application.cpp
miguelportilla: I don't know the reason it was previously added, so I am opting to keep it.

src/ripple/nodestore/impl/DatabaseShardImp.cpp
miguelportilla: Nice find!

src/ripple/app/main/Application.cpp
```diff
@@ -1978,49 +2079,31 @@ ApplicationImp::journal (std::string const& name)
     return logs_->journal (name);
 }
 
-//VFALCO TODO clean this up since it is just a file holding a single member function definition
-
-static
-std::vector<std::string>
-getSchema (DatabaseCon& dbc, std::string const& dbName)
+void
+ApplicationImp::updateTxnDB()
```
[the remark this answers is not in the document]
miguelportilla: Agreed, removed. 👍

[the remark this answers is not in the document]
miguelportilla: The ```AccountTransactions has a primary key check``` can probably also be removed but I'll leave it in unless someone feels strongly about removing it.

miguelportilla:
@mellery451 @jwbusch  I added a new commit (to be folded) that has additional changes belonging to this PR. Apologies for adding the additional work, it was going to be a separate PR but the changes really belong here. Thanks!

src/ripple/app/misc/SHAMapStoreImp.cpp
```diff
@@ -220,17 +241,27 @@ SHAMapStoreImp::makeDatabase (std::string const& name,
...
+            scheduler_,
+            readThreads,
+            app_.getJobQueue(), // Uses the JobQueue as a Stoppable parent
+            std::move(writableBackend),
+            std::move(archiveBackend),
+            app_.config().section(ConfigSection::nodeDatabase()),
+            app_.logs().journal("NodeStore"));
         fdlimit_ += dbr->fdlimit();
         dbRotating_ = dbr.get();
         db.reset(dynamic_cast<NodeStore::Database*>(dbr.release()));
```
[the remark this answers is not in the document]
miguelportilla: Possibly but I am going to leave it untouched. Keep in mind that SHAMapStore will go away entirely with coming changes.

src/ripple/app/misc/SHAMapStoreImp.cpp
```diff
@@ -220,17 +241,27 @@ SHAMapStoreImp::makeDatabase (std::string const& name,
...
     {
-        db = NodeStore::Manager::instance().make_Database (name, scheduler_,
-            readThreads, parent, setup_.nodeDatabase, nodeStoreJournal_);
+        db = NodeStore::Manager::instance().make_Database(
+            name,
+            scheduler_,
+            readThreads,
+            app_.getJobQueue(), // Uses the JobQueue as a Stoppable parent
+            app_.config().section(ConfigSection::nodeDatabase()),
+            app_.logs().journal("NodeStore"));
```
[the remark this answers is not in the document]
miguelportilla: "NodeStore" is a better description.

src/ripple/app/misc/SHAMapStoreImp.cpp
```diff
@@ -159,54 +160,74 @@ SHAMapStoreImp::SavedStateDB::setLastRotated (LedgerIndex seq)
...
-        beast::Journal journal,
-        beast::Journal nodeStoreJournal,
-        TransactionMaster& transactionMaster,
-        BasicConfig const& config)
-    : SHAMapStore (parent)
+SHAMapStoreImp::SHAMapStoreImp(
+    Application& app,
+    NodeStore::Scheduler& scheduler,
+    beast::Journal journal)
+    : SHAMapStore (dynamic_cast<Stoppable&>(app))
```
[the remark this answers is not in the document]
miguelportilla: Good point.

src/ripple/nodestore/impl/DatabaseShardImp.cpp
```diff
@@ -1301,17 +1264,28 @@ DatabaseShardImp::available() const
...
+    auto shardStore = std::make_unique<DatabaseShardImp>(
         app,
-        name,
+        parent,
+        "ShardStore",
         scheduler,
         readThreads,
         j);
+    if (!shardStore->init())
+        Throw<std::runtime_error>("shard store failed to initialize.");
```
jwbusch: Only concern was that this function didn't throw before, but now it could. We could keep the behavior the same by changing it to:
```
if (!shardStore->init()) 
    shardStore.reset();
```

miguelportilla: Good catch. Init already logged.

src/ripple/app/misc/SHAMapStoreImp.cpp
```diff
@@ -21,6 +21,7 @@
 #include <ripple/app/ledger/TransactionMaster.h>
 #include <ripple/app/misc/NetworkOPs.h>
 #include <ripple/app/misc/SHAMapStoreImp.h>
+#include <ripple/app/misc/SHAMapStoreImp.h>
```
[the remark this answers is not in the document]
miguelportilla: Bizarre, thanks. 👍

src/ripple/nodestore/impl/DatabaseShardImp.cpp
```diff
@@ -85,36 +80,86 @@ DatabaseShardImp::init()
...
+
+    if (boost::filesystem::exists(dir_))
+    {
+        if (!boost::filesystem::is_directory(dir_))
+            return fail("'path' must be a directory");
+    }
+    else
+        boost::filesystem::create_directories(dir_);
+
+    if (!get_if_exists<std::uint64_t>(section, "max_size_gb", maxDiskSpace_))
```
mellery451: I'm pretty sure the existing comparisons with this value are in terms of bytes, but this value is gb. The old code converted to bytes before passing to this class, so we should probably do it here after reading and checking overflow.

miguelportilla: Thanks! 👍

src/ripple/nodestore/impl/DatabaseNodeImp.h
```diff
@@ -49,6 +49,7 @@ class DatabaseNodeImp : public Database
         , backend_(std::move(backend))
     {
         assert(backend_);
+        addToParent(parent);
```
jwbusch: Doesn't this mean it's attached to two parents? One in the `Database::Stoppable()` (using `parent.getRoot()`) and one here using `parent`

src/ripple/nodestore/impl/DatabaseShardImp.cpp
```diff
@@ -85,36 +80,89 @@ DatabaseShardImp::init()
...
+        return fail("'max_size_gb' missing");
+
+    if (maxDiskSpace_ < minDiskSpace_)
```
jwbusch: is `minDiskSpace` also in GB? does it need to be converted to bytes too?

src/ripple/nodestore/impl/DatabaseNodeImp.h
jwbusch: You could add a bool to verify that `addToParent` is only called once, as well as verify that the parent's root matches `m_root`

src/ripple/core/Stoppable.h
```diff
@@ -318,6 +328,7 @@ class Stoppable
     std::condition_variable m_cv;
     std::mutex              m_mut;
     bool                    m_is_stopping = false;
+    bool hasParent_ {false};
```
mellery451: nit : stick with the same formatting/init style as the line above, or change both ?

src/ripple/core/Stoppable.h
```diff
@@ -209,6 +209,16 @@ class Stoppable
     /** Destroy the Stoppable. */
     virtual ~Stoppable ();
 
+    RootStoppable& getRoot() {return m_root;}
```
mellery451: should we just go ahead and make `m_root` private?

src/ripple/core/impl/Stoppable.cpp
```diff
@@ -34,16 +36,26 @@ Stoppable::Stoppable (std::string name, Stoppable& parent)
...
 {
 }
 
+void Stoppable::setParent (Stoppable& parent)
+{
+    if (hasParent_ ||
+        parent.isStopping() ||
+        std::addressof(m_root) != std::addressof(parent.m_root))
+    {
+        LogicError("Invalid parent");
```
mellery451: can you elaborate on the switch from `assert` to `LogicError`? I feel like these would be the result of programming errors and should still be `asserts` but I might be thinking about it wrong.

jwbusch: I suggested it, as it's not something that could get accidentally `#ifdef'd` out. As soon as the programmer tried to run their program w/ the wrong code they'd hit the crash and have to fix it. But if they happened to be running w/ asserts off, they might miss it.

mellery451:
still 👍 with latest commit

src/ripple/core/Stoppable.h
miguelportilla: Not sure what you mean, `m_root` is private.

mellery451: you're right - please ignore.

src/ripple/core/Stoppable.h
miguelportilla: I'll fix the formatting. I am okay having mixed naming and initialization. That has been acceptable in the past, not sure what the consensus is today. In a former life, I would have vigorously updated the file to conform to the latest standard. One benefit I like about mixed is that it documents evolutionary changes. If you have strong feelings about the uniformity I can oblige.

src/ripple/nodestore/impl/DatabaseShardImp.cpp
miguelportilla: Yes! thanks. :+1:

jwbusch: Might be helpful to use labels for units -> `minDiskSpace_` means it's in bytes, `minDiskSpaceGB_` means it's in gigabytes... since we don't have any types for these units afaik. Maybe we need some std::chrono-like system :)

miguelportilla: Bah, actually `minDiskSpace ` _is_ always in GB and before the conversion (at this point), so is `maxDiskSpace_`. So the code is correct as is. I am not a fan of the label idea but using a temp here and assigning a converted value to `maxDiskSpace_` will help avoid confusion.  I'll also narrow `minDiskSpace`'s scope to be local since its only used in one place.

src/ripple/core/impl/Stoppable.cpp
miguelportilla: @jwbusch That applies to all asserts. Nice thing is they get optimized away when turned off. I am leaning to asserting.

miguelportilla:
Thanks for the excellent feedback and review! @mellery451 @jwbusch

nbougalis:
This PR includes & closes #2978.

PR #2999, opened 2019-07-11
Fix GitLab CI

thejohnfreeman:
- Update Docker image to Boost 1.70
- Bust dependency cache
- Pass `Boost_NO_BOOST_CMAKE` to CMake


bin/ci/build.sh
```diff
@@ -14,6 +14,9 @@ BUILD_TYPE=${BUILD_TYPE:-Debug}
 # the default by setting `$CMAKE_ARGS` to the empty string.
 CMAKE_ARGS=${CMAKE_ARGS-'-Dwerr=ON'}
 
+# https://gitlab.kitware.com/cmake/cmake/issues/18865
```
mellery451: However we don't rely on `Boost_LIBRARIES`. That said, perhaps in future boost releases you'll want to consider re-enabling Config mode.

PR #3000, opened 2019-07-15
Update appveyor dependencies for boost 1.70

ximinez:
Rippled 1.3.* increased the minimum required version of boost to 1.70, but did not include an update to the appveyor config to keep up to date. This resolves that, using the update set of archives in `Downloads`.

Note that Gitlab CI will fail on this PR because of the boost version. That is fixed in #2999

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,12 +1,15 @@
 # Set environment variables.
 environment:
 
-  # We bundle up protoc.exe and only the parts of boost and openssl we need so
+  # We bundle up only the parts of boost and openssl we need so
   # that it's a small download. We also use appveyor's free cache, avoiding fees
   # downloading from S3 each time.
   # TODO: script to create this package.
-  RIPPLED_DEPS_PATH: rippled_deps17.04
-  RIPPLED_DEPS_URL: https://ripple.github.io/Downloads/appveyor/%RIPPLED_DEPS_PATH%.zip
+  RIPPLED_DEPS_PATH: rippled_deps17.05
+  RIPPLED_DEPS_BASE_URL: https://ripple.github.io/Downloads/appveyor
+  RIPPLED_OPENSSL: rippled_deps.openssl.1.0.2j.zip
+  RIPPLED_BOOST: rippled_deps.boost.1.70.zip
+  RIPPLED_BOOST_STAGE: rippled_deps.boost.stage.1.70.zip
 
   # CMake honors these environment variables, setting the include/lib paths.
   BOOST_ROOT: C:/%RIPPLED_DEPS_PATH%/boost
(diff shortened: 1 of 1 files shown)


appveyor.yml
```diff
@@ -36,24 +39,35 @@ cache:
...
-
   # Download dependencies if appveyor didn't restore them from the cache.
   # Use 7zip to unzip.
   - ps: |
-        if (-not(Test-Path 'C:/$env:RIPPLED_DEPS_PATH')) {
-            echo "Download from $env:RIPPLED_DEPS_URL"
-            Start-FileDownload "$env:RIPPLED_DEPS_URL"
-            7z x "$($env:RIPPLED_DEPS_PATH).zip" -oC:\ -y > $null
-            if ($LastExitCode -ne 0) { throw "7z failed" }
+        if (-not(Test-Path "C:/$env:RIPPLED_DEPS_PATH/openssl")) {
```
mellery451: I don't feel strongly about this, but checking for just one of the expected dirs seems weird. How about using that version file referenced below? Or possibly create an all_done file after the file extractions and look for that? If none of these suggestions make sense, I'm fine with it as is.

appveyor.yml
```diff
@@ -36,24 +39,35 @@ cache:
...
+            "Dependencies are in cache"
+            ls "C:/$env:RIPPLED_DEPS_PATH"
         }
 
   # Newer DEPS include a versions file.
   # Dump it so we can verify correct behavior.
   - ps: |
-        if (Test-Path "C:/$env:RIPPLED_DEPS_PATH/versions.txt") {
-          cat "C:/$env:RIPPLED_DEPS_PATH/versions.txt"
+        if (Test-Path "C:/$env:RIPPLED_DEPS_PATH/version*.txt") {
```
mellery451: since we know the version of deps file we're using, can't we safely assume this file exists? If it doesn't exist, would that be a sign is something wrong?

appveyor.yml
ximinez: Yeah, that makes sense. I used `openssl` because it is the last item on the list to be downloaded and extracted, so I was in a sense treating it as an `all_done` file. The version file, specifically, doesn't really work because each archive has one, so I'd have to check for the last item there, too.

Because Appveyor doesn't cache on failures, it might actually be sufficient to only check for `RIPPLED_DEPS_PATH`, and make sure we update that path any time the dependencies change (which is the intention anyway). Thoughts?

appveyor.yml
ximinez: There's more than one file - one per package. I think I agree that not having any files would be a sign of a problem.

appveyor.yml
mellery451: unless it pre-creates the cache path, checking just for that seems fine. I'm fine with whatever you decide.

ximinez:
I pushed a small commit to integrate @mellery451 's changes.

appveyor.yml
```diff
@@ -36,25 +39,34 @@ cache:
...
+                $url = "$env:RIPPLED_DEPS_BASE_URL/$file"
+                echo "Download $file from $url"
+                Start-FileDownload "$url"
+                7z x "$file" -o"C:\$env:RIPPLED_DEPS_PATH" -y > $null
+                if ($LastExitCode -ne 0) { throw "7z failed" }
+            }
+        }
+        else
+        {
+            "Dependencies are in cache"
```
jwbusch: is that supposed to be an `echo "Dependencies are in cache"` ?

ximinez: No. One of the few tricks I know in Powershell is that a plain string is just written to stdout.

jwbusch: Ah, line 54 had an `echo`, wonder if there is some nuance to it. (could you hook calls to `echo`? but not stdout?)

ximinez: It's probably even simpler than that. Different authors at different times, for example. If you think it's confusing, I can make it consistent.

jwbusch: I was just curious, I had to test it out myself to see if it worked or not, I wasn't sure if it was supposed to be a comment or if you intended to print to the console.

PR #3003, opened 2019-07-17
Log process name and version upon startup. Ensure that worker

mtrippled:
thread count gets logged to a file.

### DIFF
--- a/src/ripple/app/main/Application.cpp
+++ b/src/ripple/app/main/Application.cpp
@@ -53,6 +53,7 @@
 #include <ripple/nodestore/DummyScheduler.h>
 #include <ripple/overlay/Cluster.h>
 #include <ripple/overlay/make_Overlay.h>
+#include <ripple/protocol/BuildInfo.h>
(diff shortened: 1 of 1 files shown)


src/ripple/app/main/Application.cpp
```diff
@@ -1182,9 +1180,15 @@ bool ApplicationImp::setup()
         if (logs_->threshold() > kDebug)
             logs_->threshold (kDebug);
     }
+    JLOG(m_journal.info()) << "process starting: "
+        << BuildInfo::getFullVersionString();
 
+    // Optionally turn off logging to console.
     logs_->silent (config_->silent());
 
+    // VFALCO NOTE: 0 means use heuristics to determine the thread count.
```
nbougalis: Since you're here, this comment should just be removed from here.

mtrippled: fixed

nbougalis:
This needs to go into the next release we make - whether that's **1.3.1** or **1.4.0-b1**.

PR #3004, opened 2019-07-19
Add a LogicError when a deadlock is detected

jwbusch:
In the rare scenario of a deadlock being detected, it's better for the system to restart instead of continuing in the degenerate state. One way to trigger a restart is via crashing, rippled already has a mechanism to detect a crash and restart the process. See Sustain.cpp -> DoSustain.

Tested by adding a `sleep(100)` to `PeerImpl.cpp` to simulate a deadlock in the system, verified it restarts correctly.

### DIFF
--- a/src/ripple/app/main/LoadManager.cpp
+++ b/src/ripple/app/main/LoadManager.cpp
@@ -139,7 +139,7 @@ void LoadManager::run ()
                 // as undefined behavior.
                 //
                 constexpr auto deadlockTimeLimit = 90s;
-                assert (timeSpentDeadlocked < deadlockTimeLimits);
+                assert (timeSpentDeadlocked < deadlockTimeLimit);
 
                 if (timeSpentDeadlocked > deadlockTimeLimit)
                     LogicError("Deadlock detected");
--- a/src/ripple/app/main/LoadManager.cpp
+++ b/src/ripple/app/main/LoadManager.cpp
@@ -134,11 +134,15 @@ void LoadManager::run ()
                         << timeSpentDeadlocked.count() << " seconds.";
                 }
 
-                // If we go over 500 seconds spent deadlocked, it means that
+                // If we go over 90 seconds spent deadlocked, it means that
                 // the deadlock resolution code has failed, which qualifies
                 // as undefined behavior.
                 //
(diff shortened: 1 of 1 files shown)


src/ripple/app/main/LoadManager.cpp
```diff
@@ -134,11 +134,15 @@ void LoadManager::run ()
...
                 }
 
-                // If we go over 500 seconds spent deadlocked, it means that
+                // If we go over 90 seconds spent deadlocked, it means that
                 // the deadlock resolution code has failed, which qualifies
                 // as undefined behavior.
                 //
-                assert (timeSpentDeadlocked < 500s);
+                constexpr auto deadlockTimeLimit = 90s;
+                assert (timeSpentDeadlocked < deadlockTimeLimits);
```
mellery451: typo on this line (trailing `s`)

nbougalis: How did this compile on MSVC (see CI table, above)? 😕

jwbusch: On MSVC the entire assert expression is removed in release builds 🙃, Which is coincidentally what I was compiling before... but I had to test on Linux and I didn't want to wait the full 90s so I actually changed this to 15s before compiling. I probably fixed the typo to get it to build while testing and forgot about it (since MSVC was building fine locally).

src/ripple/app/main/LoadManager.cpp
```diff
@@ -139,7 +139,7 @@ void LoadManager::run ()
                 // as undefined behavior.
                 //
                 constexpr auto deadlockTimeLimit = 90s;
-                assert (timeSpentDeadlocked < deadlockTimeLimits);
+                assert (timeSpentDeadlocked < deadlockTimeLimit);
```
HowardHinnant: Nitpick:  To be consistent with the comment, and with `timeSpentDeadlocked > deadlockTimeLimit`, this should be `timeSpentDeadlocked <= deadlockTimeLimit`.

jwbusch: I did the other condition, but same effect. I do like symmetry :)

Issue #2847, opened 2019-02-05
fail_hard is not working as described


ximinez:
The `fail_hard` parameter to submit (https://developers.ripple.com/submit.html) is described as
> If true, and the transaction fails locally, do not retry or relay the transaction to other servers

This is not the current rippled behavior. Instead, only `ter` result codes (except for `terQUEUED`) are not retried or relayed. All other failure codes do retry. This is unlikely to be a problem because those failure codes will not succeed on most retry events, BUT some of them could potentially succeed if the protocol rules change while it's retrying, eg. an amendment goes live.

ximinez:
To keep implementation simple for now, continue to treat `terQUEUED` specially. ie. those transactions will still be relayed, and retried locally if necessary.

intelliot:
As a client of rippled, I care about which category the result code falls under:
1) The transaction has been forwarded (or will be forwarded) to other server(s).
2) The transaction was NOT forwarded to other servers, and never will be.

In case of (1), I'll watch fully validated ledgers to see if they included my transaction, and if one does, I'll look to see what effect my transaction had on the ledger.

In case of (2), I'll know that my transaction "failed" with finality, so I can feel free to create/submit a new transaction re-using that tx's sequence number.

Currently, I have to assume everything is (1) because even `tem` txs get held for a few tries. With `fail_hard: true`, I would like to have an additional field in the `submit` response that indicates which of the above two categories applies to the transaction. It could be a boolean called `hard_failed`.

mDuo13:
My 2¢:

- Yeah, `terQUEUED` should be treated as not a hard failure. All other `ter`, `tem`, `tef`, and `tel` codes should be hard failures. 
- Maybe another "stronger" option could instruct the server to skip the queue, so that `terQUEUED` would also be a hard failure.
- Possibly even `tec` codes could be treated as hard failures and not relayed (but I guess removing them from the open ledger could be messy?). That way, if you're using your own server, you don't have to destroy XRP for every time you screw up and got `tecPATH_DRY` because you forgot to include proper `Paths` or something. You might want to limit this to admins if it's significantly more work than rejecting `tel` / `tef` / etc. Or just assume that `tec` transactions are forwarded and will probably make it into a ledger.
[...]

PR #3007, opened 2019-07-23
Charge fees for use of memo

cryptobrad:


WietseWind:
👍 

Alternative (credits to @codetsunami for the idea) is a simple PoW like solution (find a hash, amount of prefix zeroes based on the memo size. Some free bytes and after that: [find the zeroes](https://runkit.com/wietsewind/xrpl-memo-pow-tryout))

mDuo13:
I like the XRP fee rather than a proof-of-work solution. It's more effective (and more in-keeping with existing XRPL features) to charge real value in the form of XRP than pseudo-value in the form of "work" (whose actual value depends on your electricity costs and computing power).

I also like how this implementation keeps transactions memos < 32 bytes free—equivalent to adding an InvoiceID or similar Hash256-type field..

src/ripple/app/tx/impl/Transactor.cpp
```diff
@@ -139,7 +139,20 @@ std::uint64_t Transactor::calculateBaseFee (
...
+
+    if (tx.isFieldPresent(sfMemos) && view.rules().enabled(featureMemoFee))
+    {
+        auto const& memos = tx.getFieldArray (sfMemos);
+
+        Serializer s (2048);
+        memos.add (s);
+
+        if (s.size() > 32)
+            memoCount = 2 * static_cast<std::uint32_t>(s.size());
```
nbougalis: Seems reasonable. I'd probably consider making the multiplier a function of length too. For example:
`memoCount = std::max(1, (s.size() / 128)) * static_cast<std::uint32_t>(s.size());`

This gives it a nice "exponential-like" curve as you can see on this [WolframAlpha graph](https://www.wolframalpha.com/input/?i=plot+f(x)+%3D+max(1,+(x+%2F+128))+*+x+from+0+to+1024)

thejohnfreeman: Maybe just `s.size() * s.size() / 128` for simplicity. With a minimum `s.size()` of 32, this expression starts at 8 and scales to 8192.

src/ripple/protocol/impl/Feature.cpp
```diff
@@ -118,6 +118,7 @@ detail::supportedAmendments ()
         "MultiSignReserve",
         "fixTakerDryOfferRemoval",
         "fixMasterKeyAsRegularKey",
+        "MemoFee"
```
ximinez: Need to also add "MemoFee" to `featureNames` in Feature.h.

thejohnfreeman: Please add a trailing comma to this line as well.

src/ripple/app/tx/impl/Transactor.cpp
thejohnfreeman: @miguelportilla makes a good point that super-linear fee scaling will just encourage splitting into multiple transactions at the point where the marginal cost of a memo byte exceeds the cost of a new transaction. That means we should just use `s.size()` for `memoCount`.

nbougalis:
The more I think about this, the more I think it’s the wrong approach. There’s nothing inherently wrong with memos, and charging more for them less sense the more I think about it.

OTNSam:
Wietse - Charging for this won't stop the child porn concern you expressed. I believe it already happened on BTC years ago. I tend to agree with you on people attempting to use XRPL for file storage. 

An alternative approach might be to write code fingerprinting these apps pulling the txn's from your servers. Then you could block/slow/timeout the results - rendering them useless.

Nik - I'm not entirely opposed to a fee, but the idea of starting it at 32 bytes? That's a 1/4 tweet! I think it's killing innovation before we even know the possibilities of what could be developed. 

I also think the scaled approach is going to be a nightmare down the road. If anything I would suggest 4 tiers 128, 256, 512, and 768. 

Apps developed using the memo field would at least have a framework and cost structure that's a bit more stable.

WietseWind:
@OTNSam I agree, however: allowing to store a relatively large chunk of data, almost for free, makes it pretty easy to _abuse_ the XRPL for something it's not designed for. 

Interesting you mention the fingerprinting and slowing/blocking certain responses. I started working on a rippled webservice proxy just now, allowing me to do just that.

OTNSam:
And that's the thing Wietse - Nobody is required to keep full history, so it's not technically a free lunch. 
We LOVE that you and others do, and I agree it's taxing your and other validators' resources (full history or not)

The bigger question:
- If somebody wanted to run an app like that, should they be able to run a full history validator and host the service on XRPL? 

- Are we closing the door on this kind of XRPL innovation? (And milder forms that may have a wonderful use case)

I can see the case for discouraging, but not hamstringing. Perhaps Free,1x,4x,8x fees? (Sorry Nik, just noticed yours has tiers - long time since I wrote any C++)

yxxyun:
How you can block a valid tx? Make no sense.

WietseWind:
@yxxyun I think you are replying at me? I mean blocking out the response message when one of those transactions is requested (fetching account transactions or individual transactions). As in: they won't be retrieving their files from my full history node.

whaus7:
Why do we need images in memos at all? I just can't think of a single automated use case which requires passing image data. It's a silly niche thing for human eyes and only helps bloat the size of XRPL

Still a little fuzzy on the approach for slowing down/blocking requests. Doesn't seem like this prevents the biggest issue which is inappropriate images stored on the chain for eternity?

JamesGDiaz:
Is there a way to track down if a memo points to another one (e.g. if a memo contains another tx id) and then increase the fee exponentially with every chained transaction?
That, coupled with a smaller max length on the memo would discourage this behaviour.

WietseWind:
@JamesGDiaz That's easy to bypass by encoding TX ID's some other way, and that wouldn't prevent the "Just dump lots of random data in it" [attack](https://pgqakmz.dlvr.cloud/pasted_2.png).

JamesGDiaz:
@WietseWind Yes, I saw that discord. 
Maybe flag addresses that continually send full length memos, kind of like when an API limits requests to n amount per second. If the network flags an address as malicious (maybe not to block all transactions, just the ability to send memos), the attacker would need new addresses every time, adding 20 zerps to the cost for every new address.
Of course this doesnt prevent anyone from hosting illegal content, but it deters the DDoS part of it.

I'm just brainstorming here :)

jnr101:
Taking some other approach and considering privacy and e.g. GDPR an issue, wouldn't it be an idea to enforce by the rippled software a hashed sha256 storage of the memo in question ? In that case an exchange of the content of the memo would always have to be off-chain, but size is guaranteed small and current fee schedule would suffice

MarkusTeufelberger:
My idea would also have been along the lines of increasing fees according to the total size of transactions in the last x units of time. People manage to spam the ledger with other things than memos too (high frequency OrderCreate/OrderCancel comes to mind). Looking not only at the current transaction but also the recent history might make sense (and would be more costly, resource-wise).

This PR however seems to hit the wrong spots, the more I think about it. Memos really hurt servers with full history, validators shouldn't even store ANY history imho and bootstrap from a verified ledger state instead, so they are completely unaffected by excessive memo spam. It is not really in their interest to do anything against memos, as these only hurt the ecosystem as a whole, not their operations. Unfortunately there's only mostly fee and reserve settings as  a knob to turn. There used to be a PoW implementation by the way on the PeerProtocol side of things, but it got removed after a while, as far as I understood it was mostly used for allowing peers to connect to you while making sure you're not being overloaded.

In general the base fees on XRPL are too low, validators should just increase them from 10 drops to 1000 or 10000 drops and this whole attack is a non-starter while still costing fractions of an US-cent at current XRP prices per transaction.
Also full history servers should have software in front such as the one @WietseWind suggested that charge for recovering historic data or block access (either rate limit, fully block, require a captcha or micropayment via ILP...). So far it was enough to view them as public good, maybe this now means it is time to start charging for a service that's being provided.

I personally switched mine off for now, I can't compile `rippled` on my distro anyways and this spam situation seems unresolved, yet a known issue/attack vector for years.

Generally,I tend to agree more with Wietse: Having MemoType, MemoData, MemoFormat and then MemoPoW (or MemoPoWNonce) in the spec would make memos a bit larger in general, but could keep them smaller in total by requiring a larger PoW for each byte added. This would be a similar approach to BitMessage, who run a "sliding window" blockchain of ~2 weeks and require a relatively hefty PoW to be attached to each of their (encrypted) message transactions. Still an incomplete solution though (can still be spammed by someone with a botnet and doesn't help anyone with a full history server or the network in general to keep history/shard size down).

Another quick-fix non-solution would be a config switch to NOT return memos from transactions earlier than a certain amount of ledgers (similar to `fetch_depth`) that node operators can enable. There's only a handful of us out there anyways who even have this data. This still doesn't help against spam, but helps against people that want to use memos for anything other than ephemeral messages/data (their intended use case).

The fastest solution would be to limit memos to a rather small size instead of 1 kB and allow larger memos only after implementing a strict(er) verifier (e.g. you can add IPFS hashes as MemoType/MemoFormat, but stuff in MemoData has to be at least syntactically valid, not just random bytes or base64 encoded). 64 bytes of Memo should be enough for everyone after all! ;-)

yxxyun:
Stellar‘s implement：

> Memo
optional The memo contains optional extra information. It is the responsibility of the client to interpret this value. Memos can be one of the following types:

MEMO_TEXT : A string encoded using either ASCII or UTF-8, up to 28-bytes long.
MEMO_ID : A 64 bit unsigned integer.
MEMO_HASH : A 32 byte hash.
MEMO_RETURN : A 32 byte hash intended to be interpreted as the hash of the transaction the sender is refunding.

nikitasius:
Ripple have **9Tb**+ and requires 32Gb RAM while exists since 2012 regarding wiki.
There are bit late when you decided "ohh hell, lets trim memos"!

* BTC have 303Gb with zindex
* ETH have 293Gb
* XMR have 73Gb

MarkusTeufelberger:
ETH with full state has ~300 GiB? Keep in mind that these ~9 TiB contain both transactions and state, the transactions themselves are a bit less than a TiB last time I checked. BTC for example only calculates, but doesn't store state (if you want to know the UTXO set at block 1000000, you need to calculate it by replaying transactions since genesis).

Anyways, the "memo" feature in Bitcoin is `OP_RETURN`, limited to 80 bytes payload and 3 bytes of type identifier afaik. (some graphs about usage: https://opreturn.org/)

yxxyun:
ETH full node is over 2TB
https://etherscan.io/chartsync/chainarchive

nikitasius:
> ETH with full state has ~300 GiB? Keep in mind that these ~9 TiB

@MarkusTeufelberger yes, as @yxxyun  mentioned it's not the 300Gb, it's about ~2.6Tb . Meainwhile BTC and XMR are much smaller and the question is what **only now** devs wanna trim the memos.

Same time [capacity-planning](https://xrpl.org/capacity-planning.html) and [configure-history-sharding](https://xrpl.org/configure-history-sharding.html) tells us what a 2 weeks (300k ledgers regarding this article) will take upto 76Gb of disk space as is (upto 256b per ledger if i calc well).

I personally have no negative to this pull request and every dev team must improve the codebase, but it looks **very late**, just when someoone wanted to inject 1Tb of trash into a blockchain.

So i want to ask: _what you did before? Why only now?_
And i don't want to even think about the impact on the actual business flow, if someone already build on actual xrp model his own stable business model and now trimming memos will affect this.

P.S. personally i did not build anything yet on XRP, but i have a nice roadmap to do it.

1maginarium:
I think this is a good idea, whether or not the Dev's have been alerted about it; there are ongoing "raids" to flood and spam the memo network on chain. You can see discussions in this discord, as well as these threads here.
They want to load another terabyte of bad meme's and try to fud XRP.

https://boards.4channel.org/biz/thread/14945724#p14945850
https://boards.4channel.org/biz/thread/14936827
https://boards.4channel.org/biz/thread/14951155
https://ndm-inf.github.io/ndm/viewFile/26580FF11BFD4DA4EA482E8062D160FBD4BFCDEC2690C366BAED8763FEC57170-48863076
unknown

codetsunami:
As some people have already discussed above instead of directly including the memo fields in the transaction, just leave a memo_hash = hash(accountID + seq_no + memo) in the transaction and then supply the memo as a separate entity after the transaction. Memos would not be inherently required for "full" history then, and some servers can collect and store memos if they wish to. I think this is the most flexible solution. It also allows memos containing illegal content to be removed from servers.

nikitasius:
> It also allows memos containing illegal content to be removed from servers.

@codetsunami you mean blockchain under moderation/cenzorship enabled?

Lol, thats better than just to store 1Tb :smirk:

codetsunami:
@nikitasius it wouldn’t be censorship. You’d still be able to submit whatever memo you want, and it’d still be relayed and processed by the network, it just makes it easier for history nodes to omit non essential blockchain data if they choose to

nikitasius:
@codetsunami we put to retieve it back. When we move the information out of the blockchain it's not a part of blockchain anymore.

In such case if everyone (all full nodes) forget something, it will be lost forever. This is a moderation. Idk who need to use a blockchain with a moderation.

codetsunami:
> @codetsunami we put to retieve it back. When we move the information out of the blockchain it's not a part of blockchain anymore.
> 
> In such case if everyone (all full nodes) forget something, it will be lost forever. This is a moderation. Idk who need to use a blockchain with a moderation.

Ripple is designed to operate with a current set of accounts. History is optional. Full history is even more optional. There's no reason not to introduce tiers. Some nodes have partial history, some have full core history (excluding memos) and some have full history (including memos). Everyone gets to keep data to the degree they actually want to instead of being forced to host other people's terabytes of worthless spam.

thejohnfreeman:
Here is a list of non-problems that I see people focusing on in the discussion:

- Images in memos. Who cares what data people store in memos? Memos are meant to be free-form, to let the payer decide what information to include and how to encode it, because we cannot anticipate every legitimate use case for memos.
- Payments that reference other payments.
- Long memos.
- Large numbers of payments. In fact, we want this!

Here are the problems:

- **The cost of transmitting a large memo is externalized.** Ethereum has a great model we can learn from in that every cost for a transaction processor, every operation and byte of data, must be paid for with a fee (gas). We need to make agents submitting large memos pay for the cost of transmitting that memo throughout the network.
- **The cost of storing a large memo is externalized.** The network does not require full history storage. This is a cost that full history servers take on voluntarily. It might make sense for the network to charge a fee (more commonly known as "rent") for storage of objects in the ledger, and we have a (poor) approximation of that with the reserve requirement, but it doesn't make sense for the network to charge a fee for storage of objects off the ledger. Instead, full history _services_ should choose how they want to handle this problem, from charging fees for processing queries to deleting unwanted data.

Any fee for a memo byte that scales greater than 1x the base fee for a transaction is easily defeated by submitting a new transaction. In other words, if the 256th byte costs a 2x fee, then agents will just cut their memo short at 255 bytes and roll the data over to a new transaction, where the 256th byte only costs the base fee for a payment transaction plus the cost of the 1st memo byte. An entire second transaction imposes transmission (and processing!) costs on the network far greater than one memo byte.

Any fee for memo bytes that is zero will let agents bypass memo fees entirely by chunking within that free tier. Because the transaction ID is 256 bits, a transaction that references another transaction needs at least 256 bits (32 bytes) to make a chain, so rolling over 1 byte of data in a chain requires 33 bytes in the next transaction.

The [optional InvoiceID field of a payment](https://xrpl.org/payment.html#payment-fields) already provides an optional no-fee free-form field for including 32 bytes of information. There may be others I'm forgetting. The point is:

- We need to charge a fee for all of the free-form optional bytes taken together, not just Memo.
- No more than 32 free-form bytes should be free.
- The fee for the last byte should not scale higher than the fee for the first byte, which means every byte needs to have the same fee.
- The fee for any given byte should not exceed the base fee for a transaction.

codetsunami:
> * We need to charge a fee for all of the free-form optional bytes taken together, not just Memo.

Good points and I agree with this particularly

MarkusTeufelberger:
Solutions proposed so far in order of proposal in this PR:

1) Increase fees for larger transactions (various schemes - free tiers, ).

2) Attach a Proof of Work scheme for larger transactions.

3) Don't/slowly serve certain transactions from history to disincentivize the use case of "split one file in chunks and have it easily stored/retrieved from XRPL full history servers".

4) Only block certain identifiable data encoding schemes.

5) Store a pointer to an arbitrary blob (Sha256 hash, IPFS hash...) and provide this data out-of-band - only the pointer stays on-ledger.

6) Increase the base fee, since transactions are relatively large and relatively cheap to spam at 10 drops a piece anyways (adding 1 TB of tx spam to the ledger doesn't require memos...).

7) Decrease the max. allowed memo size to levels similar as in other cryptocurrencies.

8) ...do nothing and let them come, I guess? (It's not like this would be undetectable or extremely cheap, hopefully in a while we'll have deterministic history shards and can host the result of this stuff on HDD...)

I hope this kinda summarizes it - some proposals (e.g. 3 and 4) are probably too easy to get around or non-issues for malicious uses, others (1, 5, 7) could have quite a limiting effect on currently unknowable use cases for adding 1 kB of data to transactions instead of only double digit numbers as in other blockchain systems.

codetsunami:
> I hope this kinda summarizes it

My second proposal (after client side PoW) was to store a pointer (hash) to the memo [point 5] but the memo is not out of band, on the contrary it's still part of the transaction and sent with/in the transaction. The difference is the transaction is only signed with the memo hash included. This way the memo data may be optionally discarded by history servers without compromising the ability to prove the correctness of the chain.

Thinking about it more I increasingly like this option.

MarkusTeufelberger:
Why should any server forward this content with the transaction to its peers? It would be in every node operator's interest to strip this information asap since the transaction is valid without it, it might be illegal to transmit anyways and it takes up precious bandwidth. Worse: Checking that the memo is the data that's being hashed costs resources, but there's nobody paying for it, so nobody will/should do this. This means that it would be easy to write a server that sends fake data instead of correct memos.

Requiring an IPFS hash (or BitTorrent infohash or ed2k hash or ...) might make it a bit more clear that the contents are not part of the transaction and that one can't expect `rippled` servers to transmit them for free reliably.

nbougalis:
As I said earlier, this PR seemed reasonable at first, as did my suggestion to change it to increase the fee non-linearly with the size of the memo. But it's become clear to me that this is *not* the right approach.

Certainly adopting a segwit-style approach, where you embed the hash of the memo block in the transaction and the memo itself is kept out is one option going forward and it could be used to allow for arbritrarily large memos, but being able to actually embed memos within the transaction is also useful so I can't quite see that going away.

Lastly a word of caution: as other have highlighted, storing history is **_optional_** on XRP Ledger, as the system does not require history to continue operating, only the last validated ledger, since each ledger is a fully self-contained state of the network as of the time the ledger closed. What does this mean? Any service that relies on those who are operating full history servers to provide storage and allow for the retrieval of historical data will operate only at the pleasure of those who operate full history servers.

codetsunami:
> Why should any server forward this content with the transaction to its peers? It would be in every node operator's interest to strip this information asap since the transaction is valid without it, it might be illegal to transmit anyways and it takes up precious bandwidth. Worse: Checking that the memo is the data that's being hashed costs resources, but there's nobody paying for it

@MarkusTeufelberger fair points. I would make it a consensus rule that the entire tx including the memo has to be present and the memo hash needs to be checked before the tx is considered valid by a node. That would prevent the behaviour you're talking about. Yes it is an additional hash and yes that costs resources. However that can easily be addressed by a small fee increase.

The only outcome of my proposed change is to allow nodes storing the history to discard or block certain memos if they want to without destroying their ability to compute the chain.

thejohnfreeman:
@MarkusTeufelberger Charging a fee for extra bytes (which I will start calling the **Size Fee**) has nothing to do with the amount of the base fee. I'm guessing you asked that rhetorically. We can consider the issue of a low base fee separately, but it should have no impact on the Size Fee proposal. Perhaps you can fork an issue proposing a higher base fee?

The Size Fee has nothing to do with the cost of storing memos, and I had tried to say as much:

>  it doesn't make sense for the network to charge a fee for storage of objects off the ledger

The Size Fee helps to internalize the cost of transmitting large transactions by levying a fee against accounts submitting large transactions. There is no better way, not even proof-of-work, which costs different payers different values based on their cost of computation (which may be zero if they're using other people's machines).

There will never be a way to stop spammers. Consider that even legitimately high traffic (e.g. Google searches) can _look_ like spam from the right angle. We can't be thinking about "stopping" spam. We can only try to deter spam. We want a filter that catches spam buts lets through legitimate traffic. A filter needs a way to differentiate the two. The biggest difference between spam and legitimately high traffic is the economic value behind them: very little for spam, very much for real users. Real users are willing to pay fees much higher than spammers to facilitate their traffic. The XRP ledger **already uses fees as a spam deterrent**. You seem to already know all this, based on the fact that you like the base fee so much you want to see it raised.

I'm arguing that the only problem with the fees is that **they do not go far enough**. They leave some loopholes for spam to squeeze through. The base fee is uncorrelated with the size of the transaction. We already [scale fees based on transaction processing costs](https://xrpl.org/transaction-cost.html#special-transaction-costs), e.g. for multi-signed transactions. I'm arguing that we need to scale fees based on transmission costs too.

Presently, we have a very simple fee schedule. If I had to guess, it would be because when it was first suggested, everyone agreed it would be better to test the waters by starting simple. I think the simplicity of this pull request shows we can easily extend the schedule, and I think the time has come. I don't want to stop with the Size Fee. It's not the only way to fight spam. But we can add new fees one at a time for the same reason we started with only a few special fees: baby steps are easier to make.

If we go with Memo Hashes, which seems to be gaining momentum, I want to make sure that the existing legitimate users of memos are not inconvenienced. That means that the correct memo must be preserved alongside its transaction at least until it is published in a validation.

MarkusTeufelberger:
Looking at babies, they also don't seem very good at making steps and be having quite a hard time walking...

Charging a **Size Fee** based on the size of all transactions included into the ledger in the past X seconds instead of only charging on-the-fly for optional fields of certain transaction types could also help and would be a more generic solution. There are transaction types or configurations that are much larger than others after all.

> the existing legitimate users of memos

Actually, from the top of my head I don't know of any existing legitimate users of memos. Whom do you have in mind?

> That means that the correct memo must be preserved alongside its transaction at least until it is published in a validation.

Ensuring data availability is the job of the transaction submitter, not the Ledger. Otherwise I could submit a LOT of large transactions that are kept in-flight (e.g. with `terPRE_SEQ`) and validators/nodes are required (how could this possibly be enforced btw?) to preserve their contents.

---

By the way:

All this talk about fees etc. glances over the issue that fees are implementation details. There just isn't a second implementation for a validating `rippled` software out there yet, so whatever fee scheme `rippled` implements will eventually be the fee scheme for XRPL. Still, tuning the fee calculation parameters of `rippled` doesn't really solve this on a protocol level.

thejohnfreeman:
Not sure what you mean by "more generic solution", but the cost of computing a fee with hysteresis is much, much higher: walking the past N transaction sets for every transaction, or building and maintaining an optimized data structure that can quickly answer that query.

If you want to suggest a fee on the raw transaction size, I like that. It's a more faithful internalization of the transmission cost.

From what I've heard, the primary use case for memos is to pass information required by the [FinCEN Travel Rule](https://www.sec.gov/about/offices/ocie/aml2007/fincen-advissu7.pdf).

I'm not talking about preserving all submitted transactions. I'm saying transactions that end up in a proposal or a validation must be accompanied by their memos, as they are right now. Nodes will need to take whatever means necessary to ensure that, which is no more than what they are doing right now.

MarkusTeufelberger:
By "more generic solution" I mean that it would also automatically apply for cases of someone spamming a lot of small, but cheap transactions as well as potential future transaction types that have unlimited size for whatever reason.

Charging for transaction size in general might be a good start, but could turn out to be too cheap. In my opinion there also needs to be a factor similar to "Load" that increases fees for accounts that put a lot of transactions onto the ledger in short bursts.

> From what I've heard, the primary use case for memos is to pass information required by the FinCEN Travel Rule.

Interesting, do you have a few txids of transactions with such memos? This data must be encrypted after all and would still likely violate the "right to be forgotten" part of GDPR. Only storing a pointer (and never actually releasing this data to anyone other than the recipient!) might be even a requirement for compliance.

> I'm saying transactions that end up in a proposal or a validation must be accompanied by their memos, as they are right now.

Even with Memo Hashes/Pointers instead of Memo Data? What's the incentive for doing that?

> Nodes will need to take whatever means necessary to ensure that, [...]

How would you find out if they didn't and what would be the consequences? After all with your proposal, immediately after validation memo data could/should be discarded. The data wouldn't be necessary to calculate any state hashes etc. after all.

nbougalis:
@MarkusTeufelberger, you wrote: "Actually, from the top of my head I don't know of any existing _legitimate_ users of memos." (emphasis mine)

I don't know that there any illegimate uses of memos. They are, essentially, free-form message fields allowing information to be attached to a transaction. Who's to say what is legitimate anyways? Is a quote from Airplane! ok? What about Satoshi Nakamoto's message in the genesis block: _"The Times 03/Jan/2009 Chancellor on brink of second bailout for banks"_?

> > From what I've heard, the primary use case for memos is to pass information required by the FinCEN Travel Rule.
> 
> Interesting, do you have a few txids of transactions with such memos? This data must be encrypted after all and would still likely violate the "right to be forgotten" part of GDPR. Only storing a pointer (and never actually releasing this data to anyone other than the recipient!) might be even a requirement for compliance.

I don't know that there's a standard for communicating such information, but here's an example of such a transaction: [07FD156038CA08E44C1A303FDC0F1A1CC743026E18965F313B8E9E3FE03ADC86](http://livenet.xrpl.org/transactions/07FD156038CA08E44C1A303FDC0F1A1CC743026E18965F313B8E9E3FE03ADC86/raw)

thejohnfreeman:
Here is a [transaction](https://livenet.xrpl.org/transactions/07FD156038CA08E44C1A303FDC0F1A1CC743026E18965F313B8E9E3FE03ADC86/raw) where the MemoType is `travelrule` (which you can [check](https://www.rapidtables.com/convert/number/hex-to-ascii.html)).

Yes, even with Memo Hashes. The reason is so that senders using memos as intended now do not have to spin up and operate a separate "memo server" so that their recipients can get the memo. We have an existing, legitimate use case for memos, today, right now. We do not (yet?) have a spam problem. We should not kill actual legitimate use cases because of a hypothetical problem. A spam counter-measure should not inconvenience real users more than the spam would.

I'm not yet on the Memo Hash bandwagon. I'm leaving the question of "how" to the proposers.

MarkusTeufelberger:
@nbougalis - John and I wrote about users, not uses. Apparently Ripple uses these travelrule formats - the (imho useless/spammy) `Client: rt1.4.3-13-g582a3a5` memo is also a good hint for that. Maybe give your colleagues at XRP II a call if you want to decode this data, they also send such memos: https://bithomp.com/explorer/r9B3osCvQ28Chgv1qKDqwcGkQaDZJf4iqB.

Ok, Ripple apparently uses Memos to append some long-ish base64 encoded (ugh! You have raw bytes available!) data to transactions that probably relates to FinCen regulations. That is one user so far and this is your employer too... but fine, let's say there are some more users that rely on this field being not suddenly an IPFS hash. They also might rely on their transactions not suddenly costing more (or worse: their libraries creating transactions with too low fees, because they didn't update their software).

@thejohnfreeman - What's the incentive and what's your additional enforcement mechanism? Currently Memos are part of a transaction and you can't verify its signature without them.

The issue is anyways that senders don't spend any resources towards aiding operation of the network, but that's a different discussion.

---

I'm personally turning towards the "Raise tx-fees in general, build deterministic shards and store history on HDD" camp, the more I think about it. A couple TB of extra history are easy to handle in that case, if something really bad ends up in transactions, one can always manually filter requests as an operator (jurisdictions are different after all) and with larger actual use transaction traffic would dwarf a few TB of meme pictures anyways. Maybe implement a blanket Size Fee that charges 1 drop per byte of serialized transaction or something along these lines so the current script kiddies (who seem to tell people to short while telling others to buy XRP?!) don't cause too many issues until the features are in place.

codetsunami:
> Yes, even with Memo Hashes. The reason is so that senders using memos as intended now do not have to spin up and operate a separate "memo server" so that their recipients can get the memo. We have an existing, legitimate use case for memos, today, right now. We do not (yet?) have a spam problem. We should not kill actual legitimate use cases because of a hypothetical problem. A spam counter-measure should not inconvenience real users more than the spam would.
> 
> I'm not yet on the Memo Hash bandwagon. I'm leaving the question of "how" to the proposers.

For regular use a change to a seg-wit style memo doesn't create any need to spin up new 'memo' nodes. As you know a transaction actually consists of the transaction + the signature. The signature can't sign itself so it is effectively 'sent with' the transaction. The memo block would be simply another concatenated entity. Now three things are sent together and all must be present and match before a node will accept the transaction as valid. There would be no tangible change for existing legitimate or future memo users, literally all that changes is the ability of full history nodes to discard large or illegal memos if they wish without compromising their ability to correctly compute ledgers from the remaining data.

MarkusTeufelberger:
@codetsunami - should these memos then also end up in the nodestore and/or shards or should they be discarded before storing them to disk?

codetsunami:
> @codetsunami - should these memos then also end up in the nodestore and/or shards or should they be discarded before storing them to disk?

@MarkusTeufelberger I'd say by default they should be sharded and all data kept. Later if it is posing a problem an amendment to prune can be put through. Doesn't hurt anyone and gives full history node runners an easy way to remove any illegal content if they get a takedown notice.

MarkusTeufelberger:
Then there could be lots of different versions of a shard (every transaction with and without memo).
Why would pruning need an amendment?

codetsunami:
@MarkusTeufelberger agree shards would also need a revision to place any segwit content in a secondary shard (a segwit for shards if you will) which may or may not be available from a given history node. Does seem reasonably complicated, however network spam might force hands here

yxxyun:
https://github.com/seelabs/rippled/commit/5b34dd08f52e0fd7dd6ca087e9bf15db56ec9d8a
So ripple planing charge fee based on tx size?

nbougalis:
@yxxyun it’s just an experimental branch... a “how would this look?” kind of thing. Ripple hasn’t decided whether to submit any proposed changes in response to this, and even if we do, it’s up to the network to decide whether to adopt those changes.

yxxyun:
Well，it's more fair but a bit complicated, for （muiti）path payment or future muiti-operation tx need determine the base fee before sign it.

JoelKatz:
I will always strongly oppose the use of PoW as any part of a transaction fee. It is heavily biased in favor of the attacker and against the legitimate user. The attacker can seek out the most cost-effective hardware and the cheapest power. Legitimate users have whatever they have. It's a terrible solution, forcing value that the ledger creates to go to equipment manufacturers and electric companies rather than staying in the ecosystem.

I would strongly prefer proposals that are the least likely to existing code. There is a lot of software, much of it unfortunately unmaintained, that interacts with the XRP Ledger and any changes that make anything that's widely used that currently works break strikes me as undesirable unless there's really no better choice.

For example, I'd prefer approaches that de-prioritize transactions based on size when recent ledger space consumption is especially elevated over approaches that raise the required minimum fees. Another possible option is to track recent ledger space consumption by account and require higher fees from accounts that have sent a lot of transactions, or a lot of large transactions, recently. Though this might just encourage determined attackers to use large numbers of accounts, I guess.

MarkusTeufelberger:
Something of concern is that if the first N transaction of an account in short local (or global) size bursts are cheaper than the next ones, this incentivizes using up actual state (by funding new accounts) instead of just spamming transaction history. Yes, this has a certain cost (currently only 20 XRP!) in the form of reserves, but is hurting the system a lot more than just using up disk space of full history nodes or badly configured partial history ones.

JoelKatz:
I guess it comes down to exactly what sort of problem we think we have. If we think that average-sized transactions are sufficiently handled with the existing fee escalation and the only real issue is that large and small transactions cost the same, then we never have to burden the most common transactions that people make today such as transfers of XRP with no (or minimal) memo.

I suppose the cleanest change would be to start by taking size into account when computing the base transaction fee with things adjusted so that the vast majority of transactions made today still have a 10 drop base fee. For example (and I'm totally making numbers up here, I haven't measured) if 95% of transactions are less than 1KB, we can make transactions under 1KB pay the base fee and transactions over 1KB pay one base fee per kilobyte.

If we think the problem is that a bad actor can just spam the ledger with 10 drop transactions and not deny any service (except to people who submit 10 drop transactions) but cause an unacceptable ledger growth rate, then I think we need to raise the base fee from 10 drops to a higher amount or, alternatively, trigger fee escalation whenever the transaction rate is higher than some amount.

One the bright side, I don't believe there's any serious short-term attack. I believe the threat is that an attacker can maliciously gradually increase the cost of running a server and keeping history over a long period of time.

socratesfin:
https://bithomp.com/explorer/r3WRgUARJi8G6SDDjE8w3s6u17d7SrpbWZ

What is the problem that currently needs to be solved?

Find a way to disincentivize actors from spamming with long useless memos by making it financially expensive to do so? 

_or_

Prevent the upload of illegal data via memos that both contain a chunk and reference the previous one needed to rebuild a file?

TexasHodlem:
I'd place the issues into two categories:
1) costs imposed on the network (bandwidth; full-history storage costs, etc.)
2) illicit content imposed on history providers

Proposed solutions seem to fit into these boxes:
A) fee-based deterrents; these solve for 1, but not for 2
B) additional processing burdens for data storage (e.g. PoW); same as A, but with a greater computational burden
C) ignoring 2 because no one is _forced_ to maintain history; I believe _someone_ should have the history, and they shouldn't have to host it in a jurisdiction that has no limits on content (e.g. child pornography)
D) storing memo hashes; solves 2 and some parts of 1, but loses functionality since off-ledger storage is required

TL;DR - I don't consider B/C to be viable. I think a combination of A (fee changes) and D (memo hashing) are required to _really_ resolve both resource burdens and censorship of illicit/spammy content.

D is the only one of these that attempts to attack issue 2. Costs are easier to resolve than legal issues. As such, I find some form of D to be imperative if there are parties engaging in illicit-content attacks. So long as any form of arbitrary data exists with a transaction, this is a possibility. Limiting the memo to a single, arbitrary byte would still allow storage of an obnoxiously chunked 1kb file, just at a significantly higher cost. Allowing the memo type "YourFavoriteHash" still leaves the possibility of repeating the string "XRPScam" (or other arbitrary data) up to the byte-length expected of the hash, _unless the hash is verified by validators_.

As much as I'd hate to see another processing burden on validators, the extremely solid solution I see is:
- sign transaction data including a memo (+sequence# to force submitter recalculation) hash, but not the memo itself so that transactions can be verified without the raw memo
- require tx submitter to send both the signed transaction and the memo
- require validators to verify the memo hash (note the arbitrary/false hash "XRPScam" example above)
- rippled configuration bool StoreMemos (or similar limit)
- rippled configuration MemoStorageBlacklist: {'accounts':[...], 'txIds':[...]} so that even a full-history node operator who's willing/able to also store all memos can do so while blocking spam/bloat memos and complying with law-enforcement

Drawbacks:
- someone needs to code this
- n validators will all have to do a checksum on the submitted memo
- full-history+memo nodes will have to store the memo AND the memo hash; this is a negligible bit of storage, but I'll admit it crossed my mind
- memos aren't guaranteed to be stored; arguably, the point in XRP is not storage, but a trustless public record, so any service relying on memos can store their own (not unreasonable for a business case), and the on-ledger memo hash will provide a way to keep them honest
- without _also_ implementing solutions from category A (fee changes, such as those based on memo size), spammers could submit 1k memos repeatedly, putting a bandwidth and processing burden on validators

Note: I see that while I was busy with this novel, @socratesfin had the same question about which problem(s) were being addressed here.

Silkjaer:
Most chunks are submitted to the same ledger, and have the same sender and recipient. Maybe it’s time to consider an anti-spam measure to prevent accounts from submitting too many transactions too fast?

It could be limited to check for e.g. transactions of same type or same sender and recipient. And enforced by having a max number of transactions per ledger or using a velocity algorithm to calculate velocity over multiple ledgers and simply fail transactions when the limit is hit.

It would prevent this kind of spam, but also mixer services sending many minor payments to obfuscate size and recipients. A backtest could be done to check who would’ve been caught in the past to ensure the right balance is found.

TexasHodlem:
@Silkjaer - Regarding a velocity limit:
- Legitimate senders may be impacted. As an example, think of two xRapid partner exchanges doing a  high volume of low value transactions. David/Nik may or may not be able to comment on whether those transactions ever use payment channels or batching. Even if not for xRapid, it's feasible that two parties transact _very_ frequently.
- The criteria for identifying all possible attacks seems unnecessarily complex. One disadvantage to open source software is that spammers will just transact to SPAM_LIMIT - 1 transactions per ledger based on a publicly available SPAM_LIMIT.
- Velocity limits don't create a volume limit. The ledger (hopefully) will be around for our lifetimes, leaving bad actors plenty of time for continuous, low-rate spam.
- Relying on sender/recipient for filtering may mean lead bad actors to use numerous accounts, only imposing the additional reserve fee on them. Given a SPAM_LIMIT, they're only charge n*20 XRP to hit the ledger at a rate of n*SPAM_LIMIT.
- Illicit content and large content are not addressed.

JoelKatz:
As a general principle, my philosophy is that the cost to perform a transaction on a public ledger should never be significantly less than the actual cost the transaction imposes on the other users of the network. There can be a certain base level of subsidized service when demand is low (as 10 drop transactions currently are) so long as that can't be abused to discourage people from operating full nodes.

codetsunami:
The cost to store a chunk of data “forever” actually depends on the contents of the data, because the primary business model for paying for data storage is to charge people who want to access it. The greater the number of people who want to access the data the more viable it is to store.

That’s why splitting the data into “core” and “auxiliary” with a segwit style division makes sense. Relatively speaking of people want access to all the core blockchain data but some guys 1Gib podcast encoded into 100,000 memos, not so much. 

In a sense the current system forces people who want access to historical core data to subsidise storage for non core data. If abuse continues indefinitely I foresee the historical core data becoming simply unavailable as a result.

downspot:
from discord https://ndm-inf.github.io/ndm/fileIndex

sublimator:
lol, I actually embedded some images using memos back in the day ...
lost the code for piecing them back together though ...
can't recall what ledger/txn etc

>

Silkjaer:
@TexasHodlem It could also use payment amount as a factor and all in an effort to limit DDOS-type attacks. 

As long as history is public, fees are low and speed is fast, it will be an open invitation to use the system unintentionally. Like banks are also targets for graffiti artists because it has walls. 

If a velocity factor is used, transparency of when limit is hit is opaque, and any high-frequency service would have to implement error handling. It would complicate misuse, not avoid it, hence not targeting any illicit/unintentional use directly. 

The tricky part would be to find the right factors to use and avoid trapping real use. I suppose high-frequency, low amount payments would be better served using payment channels anyway :)

Silkjaer:
Loosing access to full history (or long term history) would/could complicate AML measures such as checks to see if a VASP is receiving “dirty” XRP. 

If history is paywalled every VASP would have to rely on established parties to buy access, host a full history node themselves or use AML service providers who does. It would give established parties in the space an added benefit – as it is now everyone is equal only limited by own technical competences, allowing small players to participate in or even accelerate innovation.

jargoman:
Is this 32 bytes per memo or 32 bytes of memo(s) per tx? A tx can have multiple memo's.

I tend to use two memo's per tx. The first memo is the wallet version used to sign the tx. The second memo I use to mark my orders with an identifying word.

thejohnfreeman:
@jargoman 32 bytes across all Memos, MemoTypes, and MemoFormats together.

TrashAccount3424:
1) Please please do not suggest PoW. It's broken. It's cheaper the more someone wants to produce. The only positive aspect of PoW is that it can be "paid" without holding or buying a specific token. In a system like the XRPL this is irrelevant no one can actively participate without XRPs anyway.

2) Using the memo to store/transmit data is the purpose of the memo. It's a feature not the problem. So we should focus on fixing the problem(s) and not the feature.

3) What data is added to the memo should be irrelevant. I understand that some people don't want to run a node which useless data on it but there is absolutely not way to detect whats useful and what not. Even the meme images could hold hidden data valuable to someone but undetectable to others or the whole file could be the key to decrypt something important. There is no point in painting bits.
Although I personally do not get in the concept of illegal bit combinations, its how the world works. So for reported illegal content there should be an easy way to remove access, and remove the data from the history. This however is a completely separate problem which requires absolutely no changes in the way Tx are made or how fees are calculated etc. Personally I consider this a waste of time. Some easy way for each node owner to restrict access is probably a good idea trough. But overall this is a game you can't win. In a world where 09 F9 11 02 9D 74 E3 5B D8 41 56 C5 63 56 88 C0 was once considered illegal data, what stops the law maker to declare any bit combination illegal at any time. Also what stops people form encoding illegal data in literally anything. Destination tag for example can hold arbitrary data as-well. An if that's somehow killed, how about encoding data in the Tx value and in the fee burned.

mDuo13:
Regarding the "illicit data" problem, I'd like to reiterate (as several others have mentioned) that the "memo hash" proposition **does not ultimately fix the problem**. The hash field itself doesn't have to be a hash; it could just be used as an arbitrary data field of whatever size hash it is, and servers themselves probably couldn't tell the difference. There are other fields that can be used to store arbitrary data, like destination tags, which have important and legitimate use cases. At best, memo hashing makes it more effort to encode illicit data into the indispensable fields of a transaction and might reduce how many bytes any single transaction could store. 

This is an intrinsic problem of all blockchains. As long as the idea that "having" a specific string of bytes is illegal, operators are forced to choose between storing illegal data or not being able to reconstruct all history. The XRP Ledger has it less bad than Bitcoin, because on the XRP Ledger you don't need to replay all history to know the latest state.

Memo hashing also impacts legitimate uses of memos today so we should only take action if there's a reasonable belief that the pros and cons of memo hashing are weighted appropriately. 

- Pros: 
    - makes encoding illicit data more complicated and maybe more expensive
- Cons: 
    - makes storing, legitimate memos more complicated, by potentially changing the workflow around sending and looking up memos.
    - Might invalidate existing (frequently unmaintained) software that signs & submits transactions to the XRP Ledger with memos
    - Could make server management more complex. ("Do you want to store memos? Configure your memo storage backend." etc.)
    - Requires implementation work.

----

The other problem is "spammy data". This problem can be summed up as:

- If using XRPL as a file-storage service catches on, a significant portion of full-history services' costs could be taken up by serving a use case that's a bit tangential to the intended purpose of the XRP Ledger. So some may choose to stop offering full history as it becomes too expensive.
- A higher volume of large transactions would increase the costs of running even a server with minimal history (disk, bandwidth, etc.), although it remains to be seen if the change is substantial enough to have a major effect on server operators' costs. If it does, that would strongly suggest that current transaction fees are too cheap for their intended purpose, and fees should increase for large transactions, or maybe all transactions across the board.

As some commenters above have pointed out, it's already pretty expensive to serve full history, even without that many large-size transactions. It would improve the overall utility and accessibility of the network if we could do things to make it easier to store history in general. History sharding is a good step in that regard. Charging a size fee for transactions might help, but my intuition is that it wouldn't have a big impact on the current ledger. 

Another idea (credit @ximinez for the idea) would be to make an effort to reduce the number of failed transactions that get stored in the ledger. If you submit a cross-currency Payment that can't succeed because it requires paths that don't exist, for example, you'll probably get a `tec` code, which means every server in the network has to try applying your transaction, realize it doesn't work, burn your XRP fee, relay it to their peers, and validate that result via consensus. What if the transaction failed with a `ter` code instead? You wouldn't have a "final" result as quickly, but only the server(s) you submitted the transaction to would have to process it, and it wouldn't cost you any XRP, either. So one server in the network does some transaction processing that doesn't result in an XRP cost, but no other member of the network has to do any work for your failed transaction, and it doesn't have to be encoded into the immutable full history. Given that there are plenty of other ways that can make an individual server do work without burning XRP (anything that results in a tef code, for example, or just hammering their APIs in general) and the fact that this is localized to individual servers and not relayed through the network, it sounds like it could be a worthwhile change. In other words, maybe instead of just making `fail_hard` work as expected ( #2847 ), we should make it on by default.

codetsunami:
@TrashAccount3424 
Obviously a throwaway account to make your point but I'll bite.
> 3\. What data is added to the memo should be irrelevant.
I don't necessarily disagree with you that all combinations of bits should be legally equal as far as the machinery of the Internet is concerned however this is effectively activism and not really appropriate for this discussion. The reality is full history nodes need to comply with the laws in the jurisdictions where they are hosted. If this becomes impossible for most jurisdictions there simply won't be any FH nodes.

nbougalis:
>I'm personally turning towards the "Raise tx-fees in general, build deterministic shards and store history on HDD" camp, the more I think about it.

We are certainly moving in the _"build deterministic shards and store history on HDD"_ direction and moving rapidly: the idea is that each server will maintain a minimum of two shards: the "current" (open) shard, and the the "previous" shard. Beyond that, a server can opt to keep as many shards as it wants, with each shard being fully self-contained.

th3penguinwhisperer:
I have mostly read this thread but might have missed something.

To me a solution where the cost of the transaction is increased with 1 (or a multiplier of it) drop for every byte of a memo (or other free-form data fields) looks best:
- KISS, so not compute intensive and less likely to break something (as David mentions is very important)
- transparent fee
- no need for tiers
- no need for a (scaling) window to define the fee
- legit high number of transactions are not disadvantaged/discouraged
- in legit cases the extra drops won't make a "real" difference
- the "damage" of using XRPL as a file store might be countered by burning more XRP and so having a reduced supply.

I don't think I've read such a proposal in this PR. Perhaps I missed something that makes this proposal nonviable.

There's no way to prevent illicit content from being uploaded. Nor is there certainty that you would be able to trace it back to the person who did it (e.g. hacked account).

nbougalis:
Thanks for contributing code, but the consensus seems to be that this is premature and that we shoulnd't rush to get this adopted, so I am going to be closing this PR.

If you would like to continue discussing the topic of fees and ensuring that fees charged remain effective at preventing ledger and tx spam, please feel free to open a separate issue and reference this PR for context.

PR #3011, opened 2019-07-29
Check privileges of user

alloyxrp:
Check if the script is being run as root or sudo, else exit with message.

### DIFF
--- a/Builds/containers/shared/update-rippled.sh
+++ b/Builds/containers/shared/update-rippled.sh
@@ -2,6 +2,13 @@
 
 # auto-update script for rippled daemon
 
+# Check for sudo/root permissions
+if [ `id -u` -ne 0 ]; then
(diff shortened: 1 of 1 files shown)


Builds/containers/shared/update-rippled.sh
```diff
@@ -2,6 +2,13 @@
 
 # auto-update script for rippled daemon
 
+# Check for sudo/root permissions
+if [ `id -u` -ne 0 ]; then
```
mellery451: it's a nit-pick, but other places here use `$(cmd)` for subshells, so I guess I would prefer that strictly for consistency (I don't think there is any practical difference in this case, but let me know if I'm wrong...)

alloyxrp: No issues. I can change it for consistency.

PR #3014, opened 2019-07-30
Fix shard directory detection

miguelportilla:
Addresses problem introduced in PR 2899 where a shard can fail to detect if a directory exists.

Thank you @mDuo13,  for discovering this problem!

### DIFF
--- a/src/ripple/nodestore/impl/Shard.cpp
+++ b/src/ripple/nodestore/impl/Shard.cpp
@@ -65,20 +65,12 @@ Shard::open(Section config, Scheduler& scheduler, nudb::context& ctx)
         return false;
     }
 
-    boost::system::error_code ec;
-    auto const preexist {exists(dir_, ec)};
-    if (ec)
-    {
-        JLOG(j_.error()) <<
-            "shard " << index_ << ": " << ec.message();
-        return false;
-    }
-
     config.set("path", dir_.string());
     backend_ = factory->createInstance(
         NodeObject::keyBytes, config, scheduler, ctx, j_);
 
-    auto fail = [&](std::string msg)
+    bool const preexist {exists(dir_)};
+    auto fail = [this, preexist](std::string const& msg)
     {
         if (!msg.empty())
         {


miguelportilla:
@HowardHinnant Thanks. `exists` does not throw if the path to the database files does not exist. It simply returns `false`. I believe an exception will only be thrown when the OS or other underlying API results in an error that prevents the function from meeting its specifications.

HowardHinnant:
I've edited the suggested commit message.

src/ripple/nodestore/impl/Shard.cpp
```diff
@@ -65,19 +65,11 @@ Shard::open(Section config, Scheduler& scheduler, nudb::context& ctx)
...
-            "shard " << index_ << ": " << ec.message();
-        return false;
-    }
-
     config.set("path", dir_.string());
     backend_ = factory->createInstance(
         NodeObject::keyBytes, config, scheduler, ctx, j_);
 
+    auto const preexist {exists(dir_)};
     auto fail = [&](std::string msg)
```
seelabs: Nit: msg should be `const&`

miguelportilla: 👍

src/ripple/nodestore/impl/Shard.cpp
```diff
@@ -65,20 +65,12 @@ Shard::open(Section config, Scheduler& scheduler, nudb::context& ctx)
...
-            "shard " << index_ << ": " << ec.message();
-        return false;
-    }
-
     config.set("path", dir_.string());
     backend_ = factory->createInstance(
         NodeObject::keyBytes, config, scheduler, ctx, j_);
 
-    auto fail = [&](std::string msg)
+    auto const preexist {exists(dir_)};
```
scottschurr: Until we are officially on C++17, combining list initialization with `auto` is dangerous.  For me (using Clang) `preexist` is being deduced as `const bool`, I'm not sure why.  But I think this declaration is supposed to yield `preexist` as `const std::initializer_list<bool>` in C++14.  See, e.g., https://mariusbancila.ro/blog/2017/04/13/cpp17-new-rules-for-auto-deduction-from-braced-init-list/.

So my habit has been to always specify the type when I use list initialization.  And that gives me an excuse to not use `auto` :wink:.  My preferred form for this line would be:
```
bool const preexist {exists(dir_)};
```
1. You know that you get a `const bool` and can read it in the source file.
2. Because you're using list initialization without an equals you know there is no narrowing.

src/ripple/nodestore/impl/Shard.cpp
```diff
@@ -65,20 +65,12 @@ Shard::open(Section config, Scheduler& scheduler, nudb::context& ctx)
...
+    auto fail = [this, preexist](std::string const& msg)
```
scottschurr: Thanks for the explicit capture.  That's my preference as well.

mDuo13:
Just dropping in to confirm that dfedbc2 fixed my problem of being unable to create new shards.

src/ripple/nodestore/impl/Shard.cpp
miguelportilla: @scottschurr Thanks.

PR #3015, opened 2019-07-31
Allow user to make backups

yusufsahinhamza:
When `[node_db]` path changed and there is a blockchain already: rippled deletes `[node_db]` path content so user can't make blockchain backups. The deleted codes running because there is a something that has no `dbPrefix_` in the **changed** `[node_db]` path.

Also if deleted codes are necessary that problem can be fixed with using multiple conditions in comparing `dbPrefix_`.

### DIFF
--- a/src/ripple/app/misc/SHAMapStoreImp.cpp
+++ b/src/ripple/app/misc/SHAMapStoreImp.cpp
@@ -527,8 +527,6 @@ SHAMapStoreImp::dbPaths()
             writableDbExists = true;
         else if (! state.archiveDb.compare (it->path().string()))
             archiveDbExists = true;
-        else if (! dbPrefix_.compare (it->path().stem().string()))
-            boost::filesystem::remove_all (it->path());
     }
 
     if ((!writableDbExists && state.writableDb.size()) ||


nbougalis:
I'm not sure I follow what exactly is happening here. Can you explain the problem this is solving, or provide a list of steps that will "replicate" it so we can understand what this is solving?

yusufsahinhamza:
@nbougalis

I installed the rippled to my desktop completely and run it. After rippled synced, i wanted to change my rippled path so i moved rippled to another place then old path and i get that message:

```
$ ./rippled
Loading: "/Users/myuser/Desktop/rippled/build/rippled.cfg"
2019-Aug-01 04:39:40.033744 SHAMapStore:ERR state db error: 
  writableDbExists false archiveDbExists false
  writableDb '/users/myuser/desktop/rippled/build/blockchain/testnet/db/rocksdb/rippledb.f59f' archiveDb '/users/myuser/desktop/rippled/build/blockchain/testnet/db/rocksdb/rippledb.03f1'

To resume operation, make backups of and remove the files matching /users/myuser/desktop/rippled/build/data/testnet/db/state* and contents of the directory /users/myuser/desktop/rippled/build/data/testnet/db/rocksdb

libc++abi.dylib: terminating with uncaught exception of type std::runtime_error: state db error
Abort trap: 6
```
rippled didn't find the database files and gave me that message. As you can see rippled said: 

> To resume operation, make backups of and remove the files...

But after i got that message my blockchain folders (writableDb and archiveDb) in the `rocksdb` (my db type) folder already was gone because of deleted codes in the commit was worked so i actually can't make backups. After this i build my rippled like that commit and tried same scenario then that problem was fixed, my folders not deleted.

Also these deleted codes are working when there is a folder that has no `dbPrefix_` in the `rocksdb` (my db type) folder. So if i want to create a folder at `rocksdb` (my db type) folder that has no `dbPrefix_`; writableDb and archiveDb folder contents gone again. If this is for security or something that i dont know, that problem can be fixed with using multiple conditions in comparing `dbPrefix_` but i think that these codes are working at that wrong time.

src/ripple/app/misc/SHAMapStoreImp.cpp
```diff
@@ -527,8 +527,6 @@ SHAMapStoreImp::dbPaths()
             writableDbExists = true;
         else if (! state.archiveDb.compare (it->path().string()))
             archiveDbExists = true;
-        else if (! dbPrefix_.compare (it->path().stem().string()))
-            boost::filesystem::remove_all (it->path());
```
mtrippled: This change can lead to the data directory being cluttered if a directories that are not active as either the writable or archive db are not cleaned up when the process starts. I'm not sure how you managed to get an extra directory in the data directory--under normal circumstances, there should only be the 2 directories with a small window of time in which a 3rd directory is brought into existence and the archive db is deleted. If the process crashes during that window of time, it is possible for there to be 3 directories. That is why the code you are trying to remove exists.

I don't know what you intend by backing up one of these directories. However, if you want to do something like back up the archive database while rippled is running, then you can activate online_delete manually by enabling the advisory_delete option in the config, and then only executing can_delete when you are ready to allow data to be deleted. The behavior of online_delete with these options is explained in detail here: https://xrpl.org/online-deletion.html

I believe that the change that you suggest will cause unintended consequences, and also what you are trying to do can be better accomplished by using the mechanisms described in that document. Thank you very much for your submission. I hope this explanation helps you to accomplish your objectives and also helps you to learn about how we implemented this feature.

yusufsahinhamza: @mtrippled Instead of remove the codes as i said; there could be multiple conditions in comparing  `dbPrefix_` for do not delete the files if database path changed? Its actually working as expected but i think it shouldn't work when database path changed because we need to re-download everything and sync again and i dont know how users can make backups after rippled message if these folders are gone already.

mtrippled:
@yusufsahinhamza Please be careful with changes to rippled's database directories. The initial error you saw was caused because the data rippled thought it had was missing and it didn't know how to proceed. If you need to move the data from one hard drive partition to another, then stop rippled and move the contents of the entire data directory, and configure rippled so that it knows to find it's data in the new location when it starts again. Any other kind of change to the data will cause problems. If you feel the need to move the data around, it is a good idea to stop rippled and then to fully back up the data before you move it. That was if there is a problem, you can roll back to your original situation and try again. In no circumstances should you expect rippled or any other application to work well if its data is modified by some other process.

yusufsahinhamza:
@mtrippled

> If you need to move the data from one hard drive partition to another, then stop rippled and move the contents of the entire data directory, and configure rippled so that it knows to find it's data in the new location when it starts again.

That's what i am talking about, rippled deletes the database files instead of find it's data in the new location and gaving a error that says make backups when these data's already was gone.

mtrippled:
> @mtrippled
> 
> > If you need to move the data from one hard drive partition to another, then stop rippled and move the contents of the entire data directory, and configure rippled so that it knows to find it's data in the new location when it starts again.
> 
> That's what i am talking about, rippled deletes the database files instead of find it's data in the new location and gaving a error that says make backups when these data are already was gone because rippled think that data are still in the old location.

I have created a pull request to clarify the error message when this happens: https://github.com/ripple/rippled/pull/3016

Also, @mDuo13, is it reasonable to add to the operations or trouble-shooting guide the specific steps to migrate data  from one filesystem location to another? (https://xrpl.org/server-wont-start.html)

mtrippled:
Also, @yusufsahinhamza, if you wish to implement a feature for rippled to automatically know how to move data when the database path configuration changes, you can submit another pull request and it will be reviewed. I'm not sure how that would be best implemented, but if you've thought about it then maybe you have a better idea. In the meanwhile, thank you for pointing out that our error messages and troubleshooting instructions could be more clear.

PR #3016, opened 2019-08-01
[TRIVIAL] Clarify online delete data error message.

mtrippled:
This was inspired by the conversation on https://github.com/ripple/rippled/pull/3015
Namely, the existing error message is unclear that rippled doesn't just automatically know how to move its data around when the config changes. This error message is more clear that rippled doesn't know how to work with data in a corrupted state.

### DIFF
--- a/src/ripple/app/misc/SHAMapStoreImp.cpp
+++ b/src/ripple/app/misc/SHAMapStoreImp.cpp
@@ -512,18 +512,20 @@ SHAMapStoreImp::dbPaths()
         stateDbPathName /= dbName_;
         stateDbPathName += "*";
 
-        journal_.error() << "state db error: " << std::endl
+        journal_.error() << "state db error:\n"
                 << "  writableDbExists " << writableDbExists
-                << " archiveDbExists " << archiveDbExists << std::endl
+                << " archiveDbExists " << archiveDbExists << '\n'
                 << "  writableDb '" << state.writableDb
-                << "' archiveDb '" << state.archiveDb << "'"
-                << std::endl << std::endl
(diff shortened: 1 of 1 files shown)


src/ripple/app/misc/SHAMapStoreImp.cpp
```diff
@@ -542,16 +542,20 @@ SHAMapStoreImp::dbPaths()
 
         journal_.error() << "state db error: " << std::endl
                 << "  writableDbExists " << writableDbExists
-                << " archiveDbExists " << archiveDbExists << std::endl
+                << " archiveDbExists " << archiveDbExists << '\n'
                 << "  writableDb '" << state.writableDb
                 << "' archiveDb '" << state.archiveDb << "'"
-                << std::endl << std::endl
```
MarkusTeufelberger: Doesn't this cause issues on Windows? Why mix `std::endl` and `\n` in the first place?

nbougalis: I don't think it's a problem. The `\n` works correctly when printing to the console; worse case scenario, it doesn't print over multiple lines in the file and that's sorta-kinda OK. The only difference here is that `std::endl` will flush `stdout` buffers.

mtrippled: @MarkusTeufelberger @nbougalis I've removed std::endl -- at the least, it's kind of ugly to mix the two types of line endings. And when rippled writes to the console, it's to stderr, which is unbuffered, so std::endl doesn't even have an effect different from '\n'.

mtrippled:
rebased

Issue #2719, opened 2018-10-14
Openssl 1.1.1 incompatibility


robcat:
On archlinux, Openssl 1.1.1 is found but rejected by cmake, even if its version is greater than the specified minimum (1.0.2).
Openssl 1.0.2.p is also installed in parallel (via the official package openssl-1.0), but not automatically detected.

Rippled affected version: 1.1.0 and develop-tip 

```
  Could NOT find OpenSSL, try to set the path to OpenSSL root folder in the
  system variable OPENSSL_ROOT_DIR (missing: OPENSSL_CRYPTO_LIBRARY)
  (found suitable version "1.1.1", minimum required is "1.0.2")
```

MarkusTeufelberger:
Are you sure that it's not their change to build a statically linked version by default (and Arch's "great" decision not to ship static libraries at all)? I had the same problems on Manjaro, try to add `-Dstatic=OFF` to the initial cmake call...

MarkusTeufelberger:
In fact the only builds that seem to semi-reliably work currently using only distro packages for dependencies are Arch nonstatic, Opensuse Tumbleweed nonstatic and Ubuntu 18.10 static and nonstatic. (https://github.com/MarkusTeufelberger/rippled-distrotest). I hope that Alpine edge nonstatic will start working (or at least fail elsewhere) once #2712 lands.

The error message is less than helpful, I'm not sure though if that's the fault of cmake (it would be much easier to spot if it said `libssl.a not found` for example) or the way it is used in this repo.

MarkusTeufelberger:
In case you want to reproduce:

Static (default) Arch build of 1.2.0-rc3: https://travis-ci.org/MarkusTeufelberger/rippled-distrotest/jobs/441293994
Nonstatic (otherwise default) Arch build of 1.2.0-rc3: https://travis-ci.org/MarkusTeufelberger/rippled-distrotest/jobs/441294015

Dockerfiles are in the repo mentioned above.

mellery451:
I think @MarkusTeufelberger is correct - this error message is the result of not having the static openssl libraries. Those deceptive/confusing error messages come from CMake's `find_package_handle_standard_args` which `FindOpenSSL` is invoking. `-Dstatic=OFF` is how you force our build to use dynamic boost/openssl/protobuf.

nbougalis:
Should we close this issue? Is this something actionable on our end, @mellery451?

mellery451:
[...]

PR #3019, opened 2019-08-02
Clarify Linux build instructions for configuring

mDuo13:
- Explain that Arch/Manjaro/etc. need `-Dstatic=OFF` during the configure step
- move configuration options closer to that step
- separate sub-headers for configuration and build

Addresses concerns raised in #2719

### DIFF
--- a/Builds/linux/README.md
+++ b/Builds/linux/README.md
@@ -91,7 +91,7 @@ export BOOST_ROOT=~/boost_1_70_0
 Alternatively, you can add `DBOOST_ROOT=~/boost_1_70_0` to the command line when
 invoking `cmake`.
 
-### Generate and Build
+### Generate Configuration
 
 All builds should be done in a separate directory from the source tree root 
 (a subdirectory is fine). For example, from the root of the ripple source tree:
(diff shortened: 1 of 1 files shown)


Builds/linux/README.md
```diff
@@ -107,6 +107,10 @@ followed by:
 cmake -DCMAKE_BUILD_TYPE=Debug ..
 ```
 
+If your operating system does not provide static libraries (Arch Linux, and
```
MarkusTeufelberger: Arch and Manjaro are closely related. Maybe mention Arch, Alpine and OpenSUSE instead? I'm not sure about Red Hat based distros.

mDuo13: Are you sure Alpine doesn't [provide static libs](https://pkgs.alpinelinux.org/package/edge/main/x86_64/openssl-libs-static) for the necessary stuff?

I'm not very knowledgeable about OpenSUSE, but it _does_ seem like they have [a policy against packaging static libs](https://en.opensuse.org/openSUSE:Shared_library_packaging_policy#Best_Practices), so maybe that can be added.

nbougalis: I think it'd be great to add additional details here, but I'd like to merge this. Perhaps we can further enhance this document in a follow-up?

PR #3022, opened 2019-08-06
Find boost policy

mellery451:
small fix for non-bleeding-edge cmake versions.


mellery451:
just to clarify - only c9eb8c30caf74cdac37e1eaa49b3afb511313bc0 needs to be reviewed, but this change only makes sense ontop of 1.4-b1 which is not in develop yet.

PR #3025, opened 2019-08-06
Add Shard SQLite support

miguelportilla:
This PR is a small step in preparation for self-contained portable shards. With this implementation, new shards will maintain their own SQLite ledger and transaction database instances.

Only the last commit should be reviewed.


MarkusTeufelberger:
Would it be possible to store all contents of shards (including the key-value pairs) in SQLite (with an index for the keys)? I suspect that performance will of course not be equal to NuDB, but it might be still fast enough?

miguelportilla:
@MarkusTeufelberger I've had the same thought and yes, it can be done. I can't speak to the performance implication but I think the idea is worth considering.

MarkusTeufelberger:
As far as I remember, SQLite was actually an option for `node_db` early on (I think it might have been the only one available in Windows), but it got relatively quickly kicked out when k/v-DBs like hyperleveldb or rocksdb were added to the code base.

src/ripple/nodestore/impl/Shard.cpp
```diff
@@ -424,29 +442,317 @@ Shard::valFetch(uint256 const& hash)
...
+                if (!accounts.empty())
+                {
+                    auto const sTxSeq {std::to_string(txMeta->getIndex())};
+                    std::string sql (
+                        "INSERT INTO AccountTransactions "
+                        "(TransID, Account, LedgerSeq, TxnSeq) VALUES ('");
+
+                    // Try to make an educated guess on how much space we'll
+                    // need for our arguments. In argument order we have:
+                    // 64 + 34 + 10 + 10 = 118 + 10 extra = 128 bytes
```
mellery451: possible alternative for your consideration:

```
std::string sql (
    "INSERT INTO AccountTransactions "
    "(TransID, Account, LedgerSeq, TxnSeq) VALUES ");
sql.reserve(sql.length() + (accounts.size() * 128));
boost::algorithm::join (accounts |
    boost::adaptors::transformed(
        [&](AccountID const& a){
            return boost::str (boost::format (
                "('%1%','%2%',%3%,%4%)")
                % sTxID
                % ripple::toBase58(a)
                % sSeq
                % sTxSeq); }),
    ",");
sql += ';';
session << sql;
```

src/ripple/nodestore/impl/Shard.cpp
```diff
@@ -424,29 +442,317 @@ Shard::valFetch(uint256 const& hash)
...
+        soci::use(to_string(ledger->info().drops)),
+        soci::use(ledger->info().closeTime.time_since_epoch().count()),
+        soci::use(ledger->info().parentCloseTime.time_since_epoch().count()),
+        soci::use(ledger->info().closeTimeResolution.count()),
+        soci::use(ledger->info().closeFlags),
+        soci::use(to_string(ledger->info().accountHash)),
+        soci::use(to_string(ledger->info().txHash));
+
+    tr.commit();
+    return true;
```
mellery451: should this method also catch `soci_error` and return false in the case of such and exception?

src/ripple/nodestore/impl/Shard.cpp
```diff
@@ -215,61 +210,70 @@ Shard::prepare()
...
     if (complete_)
         return true;
     return boost::icl::contains(storedSeqs_, seq);
 }
 
+void
+Shard::sweep()
+{
+    assert(backend_);
+    pCache_->sweep();
```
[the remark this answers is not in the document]
miguelportilla: The only place they are reset is in `open` on failure. `backend_`'s validity is always in sync with these caches.

src/ripple/nodestore/impl/Shard.cpp
miguelportilla: Nice use of join and transformed, thanks!

src/ripple/nodestore/impl/Shard.cpp
```diff
@@ -428,29 +443,324 @@ Shard::valFetch(uint256 const& hash)
...
+                CompleteShardDBPragma,
+                TxDBInit);
+            txSQLiteDB_->getSession() <<
+                boost::str(boost::format("PRAGMA cache_size=-%d;") %
+                kilobytes(Config::getSize(siTxnDBCache, 0)));
         }
+        else
+        {
+            // The incomplete shard uses a Write Ahead Log for performance
+            lgrSQLiteDB_ = std::make_unique <DatabaseCon>(
```
jwbusch: nit: these two blocks are pretty duplicated aside from a few pieces like `setupCheckpointing` and the Pragma parameter passed in during construction.

src/ripple/nodestore/impl/DatabaseShardImp.cpp
```diff
@@ -164,43 +213,37 @@ DatabaseShardImp::init()
...
-                "Insufficient disk space";
-        }
-        fdLimit_ = 1 + (fdLimit_ *
-            std::max<std::uint64_t>(1, maxDiskSpace_ / avgShardSz_));
+        // New Shard Store
+        if (maxFileSz_ > available())
+            return fail("insufficient storage space");
+
+        // Minimum number of file descriptors required
+        fdLimit_ = 15;
```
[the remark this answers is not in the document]
miguelportilla: I agree, `fdRequired_` is a much better name, I'll make it happen.

src/ripple/nodestore/impl/DatabaseShardImp.cpp
```diff
@@ -812,11 +847,11 @@ DatabaseShardImp::import(Database& source)
     init_ = false;
     complete_.clear();
     incomplete_.reset();
-    usedDiskSpace_ = 0;
+    fileSz_ = 0;
```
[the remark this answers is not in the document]
miguelportilla: It should be set back to 0.

PR #3028, opened 2019-08-07
Cleanups 17

seelabs:
Note: Only the top three commits need to be reviewed. This current is branched off of 1.4.0-b1 candidate (which I know is changing). I'll rebase it off the actual 1.4.0-b1 when that is released

Three patches to cleanup code using c++-17 features:
1) Use fold expressions to replace the "Sean Parent for_each_arg" trick
2) Use structured binding to replace some uses of `std::tie`
3) Use class template argument deduction for `std::lock_guard`

Structured bindings replaces some explicit type declarations with `auto`. If there are objections to this, I'm happy to revert the cases where people think the explicit type decl is important.


seelabs:
Closing this; will reopen later.

Issue #2990, opened 2019-06-28
ValidatorSite needs to honor ssl related config settings


mellery451:
a couple of config settings are relevant to ssl client handling (e.g. ssl_verify) - and are respected by two http client classes: `SSLHTTPDownloader` and `HTTPClient`. The `WorkSSL` class that is used by ValidatorSite needs to do the same.

PR #3036, opened 2019-08-09
Honor SSL config settings for ValidatorSites:

mellery451:
FIXES: #2990

* refactor common SSL client setup
* enable SSL in unit-test http server
* add tests for SSLHTTPDownloader
* misc test refactoring

### DIFF
--- a/src/ripple/app/main/Application.cpp
+++ b/src/ripple/app/main/Application.cpp
@@ -508,8 +508,7 @@ public:
             *validatorManifests_, *publisherManifests_, *timeKeeper_,
             logs_->journal("ValidatorList"), config_->VALIDATION_QUORUM))
 
-        , validatorSites_ (std::make_unique<ValidatorSite> (
-            get_io_service (), *validators_, logs_->journal("ValidatorSite")))
+        , validatorSites_ (std::make_unique<ValidatorSite> (*this))
 
         , serverHandler_ (make_ServerHandler (*this, *m_networkOPs, get_io_service (),
             *m_jobQueue, *m_networkOPs, *m_resourceManager,
--- a/src/ripple/app/misc/ValidatorSite.h
+++ b/src/ripple/app/misc/ValidatorSite.h
@@ -22,10 +22,13 @@
 
 #include <ripple/app/misc/ValidatorList.h>
 #include <ripple/app/misc/detail/Work.h>
+#include <ripple/app/main/Application.h>
 #include <ripple/basics/Log.h>
 #include <ripple/basics/StringUtilities.h>
+#include <ripple/core/Config.h>
 #include <ripple/json/json_value.h>
 #include <boost/asio.hpp>
+#include <boost/optional.hpp>
 #include <mutex>
 #include <memory>
 
@@ -106,9 +109,9 @@ private:
         boost::optional<Status> lastRefreshStatus;
     };
 
-    boost::asio::io_service& ios_;
-    ValidatorList& validators_;
+    Application& app_;
     beast::Journal j_;
+
     std::mutex mutable sites_mutex_;
     std::mutex mutable state_mutex_;
 
@@ -131,9 +134,8 @@ private:
 
 public:
     ValidatorSite (
-        boost::asio::io_service& ios,
-        ValidatorList& validators,
-        beast::Journal j,
+        Application& app,
+        boost::optional<beast::Journal> j = boost::none,
         std::chrono::seconds timeout = std::chrono::seconds{20});
     ~ValidatorSite ();
 
--- a/src/ripple/app/misc/detail/WorkSSL.h
+++ b/src/ripple/app/misc/detail/WorkSSL.h
@@ -21,8 +21,9 @@
 #define RIPPLE_APP_MISC_DETAIL_WORKSSL_H_INCLUDED
 
 #include <ripple/app/misc/detail/WorkBase.h>
-#include <ripple/net/RegisterSSLCerts.h>
 #include <ripple/basics/contract.h>
+#include <ripple/core/Config.h>
+#include <ripple/net/HTTPClientSSLContext.h>
 #include <boost/asio/ssl.hpp>
 #include <boost/bind.hpp>
 #include <boost/format.hpp>
@@ -30,24 +31,6 @@
 namespace ripple {
 
 namespace detail {
-
-class SSLContext : public boost::asio::ssl::context
-{
-public:
-    SSLContext(beast::Journal j)
-    : boost::asio::ssl::context(boost::asio::ssl::context::sslv23)
-    {
-        boost::system::error_code ec;
-        registerSSLCerts(*this, ec, j);
-        if (ec)
-        {
-            Throw<std::runtime_error> (
-                boost::str (boost::format (
-                    "Failed to set_default_verify_paths: %s") %
-                    ec.message ()));
-        }
-    }
-};
 
 // Work over SSL
 class WorkSSL : public WorkBase<WorkSSL>
@@ -58,7 +41,7 @@ class WorkSSL : public WorkBase<WorkSSL>
 private:
     using stream_type = boost::asio::ssl::stream<socket_type&>;
 
-    SSLContext context_;
+    HTTPClientSSLContext context_;
     stream_type stream_;
 
 public:
@@ -68,6 +51,7 @@ public:
         std::string const& port,
         boost::asio::io_service& ios,
         beast::Journal j,
+        Config const& config,
         callback_type cb);
     ~WorkSSL() = default;
 
@@ -83,15 +67,6 @@ private:
 
     void
     onHandshake(error_code const& ec);
-
-    static bool
-    rfc2818_verify(
-        std::string const& domain,
-        bool preverified,
-        boost::asio::ssl::verify_context& ctx)
-    {
-        return boost::asio::ssl::rfc2818_verification(domain)(preverified, ctx);
-    }
 };
 
 //------------------------------------------------------------------------------
@@ -102,24 +77,24 @@ WorkSSL::WorkSSL(
     std::string const& port,
     boost::asio::io_service& ios,
     beast::Journal j,
+    Config const& config,
     callback_type cb)
     : WorkBase(host, path, port, ios, cb)
-    , context_(j)
-    , stream_(socket_, context_)
+    , context_(config, j, boost::asio::ssl::context::tlsv12_client)
+    , stream_(socket_, context_.context())
 {
-    // Set SNI hostname
-    SSL_set_tlsext_host_name(stream_.native_handle(), host.c_str());
-    stream_.set_verify_mode (boost::asio::ssl::verify_peer);
-    stream_.set_verify_callback(    std::bind (
-            &WorkSSL::rfc2818_verify, host_,
-            std::placeholders::_1, std::placeholders::_2));
+    auto ec = context_.preConnectVerify(stream_, host_);
+    if (ec)
+        Throw<std::runtime_error> (
+            boost::str (boost::format ("preConnectVerify: %s") % ec.message ()));
 }
 
 void
 WorkSSL::onConnect(error_code const& ec)
 {
-    if (ec)
-        return fail(ec);
+    auto err = ec ? ec : context_.postConnectVerify(stream_, host_);
+    if (err)
+        return fail(err);
 
     stream_.async_handshake(
         boost::asio::ssl::stream_base::client,
--- a/src/ripple/net/HTTPClientSSLContext.h
+++ b/src/ripple/net/HTTPClientSSLContext.h
@@ -1,0 +1,193 @@
+//------------------------------------------------------------------------------
+/*
+    This file is part of rippled: https://github.com/ripple/rippled
+    Copyright (c) 2019 Ripple Labs Inc.
+
+    Permission to use, copy, modify, and/or distribute this software for any
+    purpose  with  or without fee is hereby granted, provided that the above
+    copyright notice and this permission notice appear in all copies.
+
+    THE  SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+    WITH  REGARD  TO  THIS  SOFTWARE  INCLUDING  ALL  IMPLIED  WARRANTIES  OF
+    MERCHANTABILITY  AND  FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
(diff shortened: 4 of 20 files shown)


src/ripple/app/main/Application.cpp
```diff
@@ -501,7 +501,7 @@ class ApplicationImp
             logs_->journal("ValidatorList"), config_->VALIDATION_QUORUM))
 
         , validatorSites_ (std::make_unique<ValidatorSite> (
-            get_io_service (), *validators_, logs_->journal("ValidatorSite")))
+            get_io_service (), *validators_, logs_->journal("ValidatorSite"), *config_))
```
miguelportilla: Radical concept... change the ctor to only accept an application instance reference, and store it rather than the other references. Yes, it could potentially limit the configuration of the object in some future use case, but not at this moment. Besides, the complexity can be added on an as needed basis.

src/ripple/app/misc/ValidatorSite.h
```diff
@@ -109,6 +110,10 @@ class ValidatorSite
     boost::asio::io_service& ios_;
     ValidatorList& validators_;
     beast::Journal j_;
+    // we store the config because it's needed
+    // to initialize the https client (WorkSSL)
+    Config const& config_;
```
miguelportilla: IMHO, the comment is unnecessary.

src/ripple/app/misc/detail/WorkSSL.h
```diff
@@ -102,24 +77,24 @@ WorkSSL::WorkSSL(
     std::string const& port,
     boost::asio::io_service& ios,
     beast::Journal j,
+    Config const& config,
     callback_type cb)
     : WorkBase(host, path, port, ios, cb)
-    , context_(j)
-    , stream_(socket_, context_)
+    , context_(config, j)
```
miguelportilla: I wonder if the default context method should be `tlsv12_client` rather than `sslv23`, which is quite old. If anything, you might want to consider passing in `sslv23_client`.

src/ripple/app/misc/detail/WorkSSL.h
```diff
@@ -102,24 +77,24 @@ WorkSSL::WorkSSL(
...
 }
 
 void
 WorkSSL::onConnect(error_code const& ec)
 {
-    if (ec)
-        return fail(ec);
+    auto err = ec ? ec : context_.postConnectVerify(stream_, host_);
+    if (err)
+        return fail(err);
```
miguelportilla: Outside the scope of this PR but we should be using the new stream timer expiration here.

src/ripple/net/HTTPClientSSLContext.h
```diff
@@ -0,0 +1,193 @@
...
+        JLOG (j.warn()) <<
+            "Outbound SSL connection to " << domain <<
+            " fails certificate verification";
+        return false;
+    }
+
+private:
+    boost::asio::ssl::context ssl_context_;
+    beast::Journal j_;
+    bool verify_;
```
miguelportilla: can be const

src/ripple/net/impl/HTTPClient.cpp
```diff
@@ -21,6 +21,7 @@
 #include <ripple/basics/Log.h>
 #include <ripple/basics/StringUtilities.h>
 #include <ripple/net/HTTPClient.h>
+#include <ripple/net/HTTPClientSSLContext.h>
 #include <ripple/net/AutoSocket.h>
 #include <ripple/net/RegisterSSLCerts.h>
```
miguelportilla: Can be removed

src/ripple/net/SSLHTTPDownloader.h
```diff
@@ -22,6 +22,7 @@
 
 #include <ripple/basics/Log.h>
 #include <ripple/core/Config.h>
+#include <ripple/net/HTTPClientSSLContext.h>
```
miguelportilla: `RegisterSSLCerts.h` no longer needs to be included from the source file.

src/ripple/net/impl/HTTPClient.cpp
```diff
@@ -266,24 +228,23 @@ class HTTPClientImp
     )
     {
         if (!mShutdown)
-            mShutdown   = ecResult;
+            mShutdown =
+                ecResult ?
+                    ecResult :
+                    httpClientSSLContext->preConnectVerify (
+                        mSocket.SSLSocket(), mDeqSites[0]);
```
miguelportilla: nit: I suggest curlies around this multi-line.

src/test/jtx/TrustedPublisherServer.h
```diff
@@ -45,31 +50,118 @@ class TrustedPublisherServer
...
 
     socket_type sock_;
+    endpoint_type ep_;
     boost::asio::ip::tcp::acceptor acceptor_;
     std::function<std::string(int)> getList_;
 
-public:
+    // The SSL context is required, and holds certificates
+    bool useSSL_;
+    boost::asio::ssl::context sslCtx_{boost::asio::ssl::context::tlsv12};
```
miguelportilla: I wonder if it would be a good idea to use a global constexpr so that all connections are synced on the method version.

src/test/jtx/TrustedPublisherServer.h
```diff
@@ -45,31 +50,118 @@ class TrustedPublisherServer
...
+        bool immediateStart = true,
+        int sequence = 1)
+        : sock_{ioc}
+        , ep_{beast::IP::Address::from_string(
+                ripple::test::getEnvLocalhostAddr()),
+            // 0 means let OS pick the port based on what's available
+            0}
+        , acceptor_{ioc}
+        , useSSL_{useSSL}
+        , sslCtx_{boost::asio::ssl::context::tlsv12}
```
miguelportilla: `sslCtx_` was already initialized

src/test/jtx/TrustedPublisherServer.h
```diff
@@ -118,26 +229,212 @@ class TrustedPublisherServer
...
+uvWEtg6EucGOszIH8O7eJWaJpVpAfZIX+c62MQWLpOLi/QIDAQABoyowKDAmBgNV
+HREEHzAdgglsb2NhbGhvc3SHEAAAAAAAAAAAAAAAAAAAAAEwDQYJKoZIhvcNAQEF
+BQADggEBAOhLAO/e0lGi9TZ2HiVi4sJ7KVQaBQHGhfsysILoQNHrNqDypPc/ZrSa
+WQ2OqyUeltMnUdN5S1h3MKRZlbAeBQlwkPdjTzlzWkCMWB5BsfIGy5ovqmNQ7zPa
+Khg5oxq3mU8ZLiJP4HngyU+hOOCt5tttex2S8ubjFT+3C3cydLKEOXCUPspaVkKn
+Eq8WSBoYTvyUVmSi6+m6HGiowWsM5Qgj93IRW6JCbkgfPeKXC/5ykAPQcFHwNaKT
+rpWokcavZyMbVjRsbzCQcc7n2j7tbLOu2svSLy6oXwG6n/bEagl5WpN2/TzQuwe7
+f5ktutc4DDJSV7fuYYCuGumrHAjcELE=
+-----END CERTIFICATE-----
+)cert"};
```
miguelportilla: nit: this might look better indented.

src/test/jtx/TrustedPublisherServer.h
```diff
@@ -94,22 +186,41 @@ class TrustedPublisherServer
...
                 &TrustedPublisherServer::on_accept, this, std::placeholders::_1));
     }
 
-    ~TrustedPublisherServer()
+    void stop()
     {
         error_code ec;
         acceptor_.close(ec);
+        // TODO consider making this join
+        // any running do_peer threads
```
miguelportilla: Probably a good idea, or maintain a reference count and have the dtor block.

src/test/jtx/TrustedPublisherServer.h
```diff
@@ -150,32 +447,52 @@ class TrustedPublisherServer
             return;
 
         static int id_ = 0;
-        std::thread{lambda{++id_, *this, std::move(sock_)}}.detach();
+        std::thread{lambda{++id_, *this, std::move(sock_), useSSL_}}.detach();
```
miguelportilla: Not sure we want to move `sock_` as `async_accept` needs it.

src/test/jtx/TrustedPublisherServer.h
```diff
@@ -150,32 +447,52 @@ class TrustedPublisherServer
...
         acceptor_.async_accept(
             sock_,
             std::bind(
                 &TrustedPublisherServer::on_accept, this, std::placeholders::_1));
     }
 
     void
-    do_peer(int id, socket_type&& sock0)
+    do_peer(int id, socket_type&& sock0, bool ssl)
```
miguelportilla: nit: can we rename to `sock`?

src/test/net/SSLHTTPDownloader_test.cpp
```diff
@@ -0,0 +1,269 @@
...
+        bool called = false;
+        boost::filesystem::path dest;
+
+        void operator ()(boost::filesystem::path dst)
+        {
+            std::unique_lock<std::mutex> lk(m);
+            called = true;
+            dest = std::move(dst);
+            lk.unlock();
+            cv.notify_one();
```
miguelportilla: I believe our past consensus has been to notify while holding the lock. 

```
std::lock_guard<std::mutex> lk(m);
called = true;
dest = std::move(dst);
cv.notify_one();
```

src/test/net/SSLHTTPDownloader_test.cpp
```diff
@@ -0,0 +1,269 @@
...
+              env.app().getIOService(), journal_, env.app().config())}
+        {}
+
+        SSLHTTPDownloader* operator->()
+        {
+            return ptr_.get();
+        }
+    };
+
+
```
miguelportilla: nit: extra lf

src/test/net/SSLHTTPDownloader_test.cpp
```diff
@@ -0,0 +1,269 @@
...
+            std::string("Basic download - SSL ") +
+            (verify ? "Verify" : "No Verify");
+
+        using namespace jtx;
+        ripple::test::detail::FileDirGuard cert {
+            *this, "_cert", "ca.pem", TrustedPublisherServer::ca_cert()};
+
+        Env env {*this, envconfig([&cert, &verify](std::unique_ptr<Config> cfg)
+            {
+                if ((cfg->SSL_VERIFY = verify)) //yes, this is assignment
```
miguelportilla: nit: minus the comment, this is less text.
```
cfg->SSL_VERIFY = verify;
if (verify)
```

src/test/net/SSLHTTPDownloader_test.cpp
```diff
@@ -0,0 +1,269 @@
...
+            dest = std::move(dst);
+            lk.unlock();
+            cv.notify_one();
+        };
+
+        bool waitComplete()
+        {
+            std::unique_lock<std::mutex> lk(m);
+            using namespace std::chrono_literals;
+            return cv.wait_for(lk, 1s, [this]{return called;});
```
miguelportilla: It might be worth considering a greater timeout on debug builds.

src/ripple/app/misc/detail/WorkSSL.h
mellery451: a timer for `postConnectVerify`, or just generally for all stream/async operations ? if the latter, I agree...if the former, I guess I need more clarification.

miguelportilla: for all stream/async operations.

src/ripple/app/main/Application.cpp
mellery451: done

src/ripple/app/misc/ValidatorSite.h
mellery451: removed

src/ripple/app/misc/detail/WorkSSL.h
mellery451: I've gone ahead and changed the WorkSSL to use `tlsv12_client`. You're probably right that the default value should be increased, but I'll leave that as a TBD item

src/ripple/app/misc/detail/WorkSSL.h
mellery451: agree..will make an issue for this.

src/ripple/net/HTTPClientSSLContext.h
mellery451: done

src/ripple/net/SSLHTTPDownloader.h
mellery451: done

src/test/jtx/TrustedPublisherServer.h
mellery451: fixed

src/test/jtx/TrustedPublisherServer.h
mellery451: I wanted to do this, but I think we'd have to convert to async calls. That's not horrible, but more work than I want to do here. I'll make a low priority issue

src/test/jtx/TrustedPublisherServer.h
mellery451: easier to paste from command line if we don't indent and leave it as a raw string block. I don't think leading spaces would be permitted by the current raw string decl (or openssl wouldn't like it..). That said, another  approach would be to figure out what openssl api calls could generate these strings at runtime (the equiv of the command lines I noted in the comment) and then we don't need the pasted wall of text at all.

src/test/jtx/TrustedPublisherServer.h
mellery451: I think what's happening is that sock_ is being stolen/moved for each session and then a new sock is effectively created at that moved-from location by the next accept. I don't necessarily like it, but I wasn't motivated to change it either.

src/test/jtx/TrustedPublisherServer.h
mellery451: done

src/test/net/SSLHTTPDownloader_test.cpp
mellery451: sounds reasonable - changed to this

src/test/net/SSLHTTPDownloader_test.cpp
mellery451: I'll keep an eye on tests, but this value seems fine so far (everything is tiny messages over localhost). If it ever causes spurious failures, I'll change it.

src/test/net/SSLHTTPDownloader_test.cpp
mellery451: I'm not sure what this comment applies to

PR #3039, opened 2019-08-10
Added sahyadri.isrdc.in to the bootstrap hubs

ivishwas:
ISRDC @ IIT Bombay is one of the four information security research centers spread across India. ISRDC @ IIT-Bombay does R&D in a wide range of topics in information security and privacy.


MarkusTeufelberger:
https://sahyadri.isrdc.in:51235/crawl doesn't seem to work for me... Can you give a little bit of information about your cluster and your setup, especially networking wise?

nbougalis:
Allowing a hub to be crawled is nice, but I don't know that we want to make that a requirement for including the hub in the bootstrap list.

I will note that this particular hub already appears on the network and, if I am right, already carries several hundred peers. If @ivishwas wishes to share general details about his system (CPU, RAM, bandwidth) that's up to him.

Issue #2955, opened 2019-05-30
CMake build system should optionally be able to make use of installed/packaged dependencies


mellery451:
see https://github.com/ripple/rippled/pull/2949.  We should be able to optionally locate/use dependencies installed on the build system if they are available and meet our requirements (static libs if requested, min versions, etc.).

mellery451:
I think @movitto did a bunch of work to make this mostly feasible. At this point, I'll close this issue and we can open issues for specific dependencies as needed to support installed/packaged deps.

PR #3044, opened 2019-08-13
Refactor cmake deps / pull in system libraries

movitto:
See #2955 and #2949. This is the start of a series of changes to simplify the build system, begining with a modularization of the control logic from residing in one large CMakeLists.txt file to many targeted subcomponents localed in the 'cmake/' subdir. The build logic is 100% preserved in the first patch and no changes are introduced outside of splitting up the large file.

This patchset also introduces the start of the logic conditionally checking for system dependencies before attempting to pull them as vendored static libs, specifically addressing the lz4 library. One edge case that will need to be explored is in the case of using system dependencies in debug builds as often linux distros and other environments ship this information seperately (on RedHat/CentOS platforms for example debugging symbols are found in 'debuginfo' packages in the same named repository).

If the direction looks good, we should be able to apply the same pattern to most if not all the vendored dependencies. Let me know!


mellery451:
@movitto thanks for these changes - looks interesting. I'll play with it a bit and comment where appropriate. Having worked on a number of large cmake builds, my personal preference is *not* to break up the CMakeLists into lots of sub-includes, but I understand that might be an unpopular opinion - so I don't object to giving your approach a try.

One note on organization - we already have a few cmake/build support files in `Builds/CMake`, so we probably need to figure out how to unify your new files with those existing files (move one or the other). I'd vote to move stuff into the pre-existing location and possibly move a few existing files into a further subdir as needed. Your call.

cmake/RippledBase.cmake
```diff
@@ -0,0 +1,21 @@
...
+endif ()
+include (ExternalProject)
+include (CMakeFuncs) # must come *after* ExternalProject b/c it overrides one function in EP
+include (ProcessorCount)
+if (target)
+  message (WARNING
+    "The target option is deprecated and will be removed in a future release")
+  parse_target()
+endif ()
+project (rippled)
```
mellery451: project declarations need to be in the top level cmake file I believe:  https://cmake.org/cmake/help/latest/command/project.html

cmake/deps/Lz4.cmake
```diff
@@ -0,0 +1,75 @@
+#[===================================================================[
+   NIH dep: lz4
+#]===================================================================]
+
+add_library (lz4_lib STATIC IMPORTED GLOBAL)
+find_library(lz4 NAMES lz4)
```
mellery451: Two main things are lacking with the simple `find_library` approach: (1) version requirement and (2) static/dynamic use. This currently is broken for me on macos, for example, because find_library finds the dylib even when I do static builds. Furthermore, we've declared a STATIC imported library, but then this searches for any lib which usually defaults to dynamic first. That's inconsistent.

lz4, like most projects, is not producing proper cmake configs. But it looks like it's instead installing package config modules, so I think we need to grab a find module that does the pkgconfig search/imports. https://github.com/scylladb/seastar/blob/master/cmake/Findlz4.cmake might be a good start.  That would give us a version check (I would make 1.8 minimum version because I'm not willing to test all the prior versions) and PkgConfig should provide static and dynamic vars when found. If the the case of static rippled builds, we need to make sure the static libs is being used here.

movitto:
@mellery451 +1 thanks for the feedback. As far as the single large CMakeLists file vs multiple modularized components, as a relatively new contributor I've found the large file a bit unwieldy to work around and feel a bit of modularization may go a long way. From what I've read subprojects are the way to go, but that would take a bit of undertaking with the current code base, and this may serve as a good first stage on that road. Of course I am down to go with team & community standards, if the large file is preferred I can revert the first patch and apply subsequent patches ontop of the single file.

Agree on merging the newly introduced modules into the existing Builds/CMake dir. I'll send a followup "fold" patch with that and the other feedback, the top level "project" deceleration & pkgconfig integration in a bit.

movitto:
Hey @thejohnfreeman, this is stemming from the discussion on #2949 and #2955. What we're going for consists of the first patch + subsequent patches in the same style of the second patch here, one for each vendored dependency that we want to try to pull in from the base OS (before falling back to the current default if not available)

Just checked out conan, seems interesting, kinda like rubygems/bundler for C++. Would we still be able to use OS level libraries in that case?

movitto:
@mellery451 @thejohnfreeman Just pushed some followup patches incorporating the feedback, including:

- merging the cmake modules into the existing utility dir
- moving the project definition and top level checks back to top level CMakeLists
- using pkgconfig to search for / setup lz4 lib paths (distinguishing between static & dyn libs)

If this looks correct I'll go in and update the remaining deps to work like this so that rippled can be built against system libraries.

mellery451:
@movitto thanks for those updates - these changes look good to me. If you proceed with the other deps, I'm not sure the pkg config lookup will work for all of them (only if they actually install pkg-config files). I would recommend hunting around for a Find* module for each dependency and try to mimic what others are doing to locate the specific library. Also, I would skip the ed25519 and secp256 libs for now - those are now part of openssl and we can switch to those versions by requiring a recent openssl.

Builds/CMake/deps/Lz4.cmake
```diff
@@ -0,0 +1,96 @@
+#[===================================================================[
+   NIH dep: lz4
+#]===================================================================]
+
+add_library (lz4_lib STATIC IMPORTED GLOBAL)
+
+find_package (PkgConfig REQUIRED)
```
mellery451: PkgConfig doesn't exist on any normal windows machine (maybe on cygwin or mingw...), so I would wrap all the search-for-lz4 stuff in an `if (NOT WIN32)` check.

movitto:
@mellery451 updated, defines IMPORTED_LOCATION_DEBUG and includes win32 platform guard.

Will start looking into other deps soon (if this is ack'd & merged before I get to it I'll send a new PR, else I'll just push ontop of this)

movitto:
@mellery451 remaining dependency checks implemented (where appropriate). At some point we should look at implementing some helper logic containing the abstracted pkgconfig logic but for the time being this should work. Let me know if there's anything else!

thejohnfreeman:
Even with Conan, you can still pull in system packages as you are now, bypassing the Find Module that Conan writes in favor of your own. However, I've always heard it considered bad practice to just take whatever is found on the system instead of installing from a controlled dependency. At some point, an incompatible version of a dependency will be installed (perhaps because it is a dependency of another unrelated system package), or you want to update to a new version but no package has yet been made for your system's package manager, or countless other problems.

I share the desire for a way to stop the build from installing every dependency via `ExternalProject_Add`. The best way to do that, in my opinion, is for the build to use the CMake facility for locating dependencies (`find_package`) first, and only after it fails, manually installing the dependency. `find_package` gives the consumer a chance to fill in the blank and point to the dependency installation they want to use, via a Find Module. Then, this PR can be about providing a default set of Find Modules (as @mellery451 has suggested too). The general pattern of use can be this (for all platforms):

```cmake
find_package(<package>)
if(NOT <package>_FOUND)
    ExternalProject_Add(<package> ...)
endif()
```

movitto:
@thejohnfreeman :+1: makes sense, I will update this to use the more 'pkgconfig' centric in the near future.

Agree that being able to tightly control deps leads to more project independence and control though supporting the option to build against system deps helps with compliance purposes, security maintenance, etc. Nice thing is we can have both!

thejohnfreeman:
Forgive me for being pedantic, I just want to make sure there is no confusion. `find_package` is unrelated to pkg-config, so I would not call it a "pkg-config centric" approach. `find_package` hunts for either a CMake Package Configuration File (written by dependency authors) or a CMake Find Module (written by dependency users). Inside of the Find Module, you can look for a pkg-config file or anything else. I would call the approach I'm recommending "`find_package` with default Find Modules".

MarkusTeufelberger:
> However, I've always heard it considered bad practice to just take whatever is found on the system instead of installing from a controlled dependency.

I'd argue that what is installed on my system is a controlled dependency. The reality is unfortunately, that the Nix philosophy for package management isn't as wide spread as it should be.

Still, eventually it would be nice to have rippled in mainstream distros. This means likely still that Boost needs to be vendored in and linked statically, but other dependencies really should also be provided externally. After all you're also not always the fastest when it comes to updating the pinned versions of your dependencies.

mellery451:
@movitto at your convenience, please rebase this on latest `develop` and let's see how the updated CI handles it. Thanks.

movitto:
@mellery451 doh! fixed

movitto:
> this looks great - thanks for making these changes. If you think you're done for now, we can mark this as passed (assuming travis passes) and merge it in the next beta. If you have more changes you'd like to get in, that's fine too.

Awesome @mellery451, many thanks for the review. Since this incorporates a large / fragile change to CMakeLists.txt I'd say lets go ahead and merge this. I'll take care of the additional dependencies in a follow-on PR.

Builds/CMake/deps/Boost.cmake
```diff
@@ -0,0 +1,92 @@
...
+  endif ()
+  message(STATUS "Adding [${Boost_INCLUDE_DIRS}] to sanitizer blacklist")
+  file (WRITE ${CMAKE_CURRENT_BINARY_DIR}/san_bl.txt "src:${Boost_INCLUDE_DIRS}/*")
+  target_compile_options (opts
+    INTERFACE
+      # ignore boost headers for sanitizing
+      -fsanitize-blacklist=${CMAKE_CURRENT_BINARY_DIR}/san_bl.txt)
+endif ()
+
+# workaround for xcode 10.2 and boost < 1.69
```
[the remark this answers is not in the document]
movitto: Are you referring to the comment or the entire workaround? (eg what about for xcode 10.2?)

Will remove in the follow-up PR.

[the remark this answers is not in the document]
mellery451: the entire workaround can be removed...a future PR would be just fine. Thanks.

nbougalis:
Merged in #3077.

PR #3049, opened 2019-08-16
Improve protocol-level handshaking protocol

nbougalis:
This commit restructures the HTTP based protocol negotiation that `rippled` executes and introduces support for negotiation of compression for peer links which, if implemented, should result in significant bandwidth savings for some server roles.

This commit also introduces the new `[network_id]` configuration option that administrators can use to specify which network the server is part of and intends to join. This makes it possible for servers from different networks to drop the link early.

The changeset also improves the log messages generated when negotiation of a peer link upgrade fails. In the past, no useful information would be logged, making it more difficult for admins to troubleshoot errors.


nbougalis:
@MarkusTeufelberger, I would appreciate a review from you as well, since this also serves to correct several issues you've previously identified or mentioned in passing, including updating documentation.

src/ripple/overlay/impl/ProtocolMessage.h
```diff
@@ -106,80 +214,102 @@ std::pair <std::size_t, boost::system::error_code>
...
         return result;
 
-    if (bs > Message::kMaxMessageSize)
+    /** The largest size that a message can be.
+
+        Sending a message whose size exceeds this may result in the connection
+        being dropped. A larger message size may be supported in the future or
+        negotiated as part of a protocol upgrade.
+    */
+    static std::size_t constexpr kMaxMessageSize = 64 * 1024 * 1024;
```
MarkusTeufelberger: Wasn't there something about support for `64MB` or `64 MiB` or `megabytes(64)` or so?

src/ripple/overlay/README.md
```diff
@@ -128,17 +111,21 @@ Content-Type: application/json
...
-    list of at least one element where each element is of the form "RTXP/"
-    followed by the dotted major and minor protocol version number. For
-    responses the value must be a single element matching one of the elements
-    provided in the corresponding request field. If the server does not
-    understand any of the requested protocols, the request is rejected.
+    list of at least one element where each element is of the form "XRPL/"
+    followed by a dotted major and minor protocol version number, where the
+    major number is greater than or equal to 2 and the minor is greater than
+    or equal to 0. The legacy version "RTXP/1.2" is also supported at this time
+    and is an alias for "XRPL/2.0". For responses the value **MUST** be a _single
```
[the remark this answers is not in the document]
nbougalis: It doesn't change any message types: what's removed was already not used (or in the case of `TMHello` message, used only internally).

src/ripple/overlay/README.md
```diff
@@ -147,41 +134,128 @@ Content-Type: application/json
...
+    The value is a 32-bit unsigned integer. The following well-known values are
+    in use:
+
+    - **0**: The "main net"
+    - **1**: The Ripple-operated [Test Net](https://xrpl.org/xrp-test-net-faucet.html).
+
+* `Network-Time`
+
+    If present, reports the current [time](https://xrpl.org/basic-data-types.html#specifying-time)
+    according to the other endpoint's internal clock.
```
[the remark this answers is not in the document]
nbougalis: The relevant context here is whether you're the *sending* or *receiving* the field. If you're sending, it's your clock. If you're receiving, it's the sender's clock. I'll ask @mDuo13 to help refine the wording here.

src/ripple/overlay/README.md
```diff
@@ -147,41 +134,128 @@ Content-Type: application/json
...
+    [parallel networks](https://xrpl.org/parallel-networks.html) the sending
+    server is joined.
+
+    The value is a 32-bit unsigned integer. The following well-known values are
+    in use:
+
+    - **0**: The "main net"
+    - **1**: The Ripple-operated [Test Net](https://xrpl.org/xrp-test-net-faucet.html).
+
+* `Network-Time`
```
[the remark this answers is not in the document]
nbougalis: It's optional. You can't really make it mandatory (it would break everyone!) and it can _only_ result in refused connections if both servers have a configured value and the values don't match.

src/ripple/overlay/README.md
```diff
@@ -147,41 +134,128 @@ Content-Type: application/json
...
+
+    If present, reports the current [time](https://xrpl.org/basic-data-types.html#specifying-time)
```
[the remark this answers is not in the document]
nbougalis: It's an unsigned integer. There's no floating time. That's just crazy talk! :sweat_smile:

src/ripple/overlay/README.md
```diff
@@ -147,41 +134,128 @@ Content-Type: application/json
...
+    The hash is encoded using **Base64** encoding, but a future release of the
+    software will switch to supporting hexadecimal encoding.
 
 * `Crawl` (optional)
 
     If present, and the value is "public" then neighbors will report the IP
     address to crawler requests. If absent, neighbor's default behavior is to
     not report IP addresses.
 
+* `Closed-Ledger` (optional)
```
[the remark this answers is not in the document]
nbougalis: Well, the handshake happens every time you get a new connection. I guess we could send a full ledger header, but I'm not sure it really helps very much. There's nothing that the server could do with that value at this time anyways.

src/ripple/overlay/README.md
```diff
@@ -24,32 +25,8 @@ messages are exchanged between peers and serialized using
...
-operate in both the leaf and superpeer role for different connections. One of
-the requirements 
-
-### Client Handler
-
-While not part of the responsibilities of the Overlay module, a peer
-operating in the Client Handler role accepts incoming connections from clients
-and services them through the JSON-RPC interface. A peer can operate in either
-the leaf or superpeer roles while also operating as a client handler.
+affects the behavior of message routing At present, only a single type is
```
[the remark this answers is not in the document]
nbougalis: Oops.

src/ripple/overlay/README.md
```diff
@@ -111,8 +94,8 @@ Content-Type: application/json
 
 * *HTTP Version*
```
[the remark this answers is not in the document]
nbougalis: I like the idea of a table. I'll see what I can do. Re: xrpl.org, certainly this could also be there, but I think it's good to have these "inline" markdown files (providing we keep them up to date) so that the documentation can travel _with_ the project.

src/ripple/overlay/impl/ProtocolMessage.h
```diff
@@ -106,80 +214,102 @@ std::pair <std::size_t, boost::system::error_code>
...
+    if (size == 0)
+        return result;
+
+    auto header = detail::parseMessageHeader(buffers, size);
 
-    // If we don't even have enough bytes for the header, there's no point
-    // in doing any work.
-    if (bs < Message::kHeaderBytes)
+    // If we can't parse the header then it may be that we don't have enough
+    // bytes yet, or because the message
```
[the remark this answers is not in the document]
nbougalis: I thought that comment was rather smart as is... then you had to go and ruin it. :laughing:

src/ripple/overlay/impl/ProtocolMessage.h
```diff
@@ -63,33 +61,143 @@ protocolMessageName (int type)
...
+        hdr.payload_wire_size &= 0x0FFFFFFF;
+
+        for (int i = 0; i != 4; i++)
+            hdr.payload_real_size = (hdr.payload_real_size << 8) + *iter++;
+
+        hdr.payload_real_size &= 0x0FFFFFFF;
+
+        hdr.total_wire_size = hdr.header_size + hdr.payload_wire_size;
+
+        for (int i = 0; i != 2; ++i)
```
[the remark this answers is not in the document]
nbougalis: I've gotten used to the `++variable` form unless I _specifically_ need the pre-increment value. I doubt it really matters in practice; compilers are [SMRT](https://www.youtube.com/watch?v=9scdimuHGh4)...

src/ripple/overlay/impl/ProtocolMessage.h
```diff
@@ -63,33 +61,143 @@ protocolMessageName (int type)
...
+            hdr.message_type = (hdr.message_type << 8) + *iter++;
+
+        return hdr;
+    }
+
+    // Version 2 header: compressed payload.
```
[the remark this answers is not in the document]
nbougalis: I hear you. I actually don't really like this binary prefix business, and am seriously considering migrating to a pure WebSocket-based transport (which would make the binary header unnecessary) or adopting an approach closer to what you describe (involving coded protobuf streams).

src/ripple/overlay/impl/ProtocolMessage.h
nbougalis: I think so. I'll go spelunking for it.

src/ripple/overlay/README.md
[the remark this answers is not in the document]
MarkusTeufelberger: What does it currently do with these values?

src/ripple/overlay/impl/ProtocolMessage.h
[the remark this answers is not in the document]
MarkusTeufelberger: There are only 2 hard problems in distributed systems:

2) Message Ordering
1) Exactly once delivery
1) Exactly once delivery

src/ripple/overlay/impl/ProtocolMessage.h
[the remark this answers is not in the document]
MarkusTeufelberger: gRPC might be worth a look too then...

src/ripple/overlay/README.md
```diff
@@ -147,41 +134,128 @@ Content-Type: application/json
...
+talking directly to each other over a single end-to-end SSL/TLS session instead
+of two separate SSL/TLS sessions, with an attacker acting as a proxy.
+
+The XRP Ledger protocol prevents this attack by leveraging the fact that the two
+servers each have a node identity, in the form of **`secp256k1`** keypairs, and
+use that to strongly bind the SSL/TLS session to the node identities of each of
+the two servers at the end of the SSL/TLS session.
+
+To do this we "reach into" the SSL/TLS session, and extract the **`finished`**
+messages for the local and remote endpoints, and combine them to generate a unique
```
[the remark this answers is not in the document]
MarkusTeufelberger: Urgh, I tried to implement this in Python with `pyOpenSSL` and `cryptography` (and `base58`, since the node public key is encoded that way - it would be nice to have _everything_ in either hex, base64 or base58...).

`rippled` seems to just drop connection attempts it doesn't like/understand. It would be nice to have at least a HTTP 400 error or similar returned instead.

src/ripple/overlay/README.md
```diff
@@ -147,41 +134,128 @@ Content-Type: application/json
...
-    This field value must be present, and contain a base 64 encoded value used
-    as a server public key identifier.
+    This field value must be present, and contain the server's public key
+    identifier. The value is encoded in base58 using the standard encoding
+    for node public keys.
 
 * `Session-Signature`
 
     This field must be present. It contains a cryptographic token formed
     from the SHA512 hash of the shared data exchanged during SSL handshaking.
```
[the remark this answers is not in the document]
MarkusTeufelberger: Apparently both...

SHA512full of the data, Sha512Half is used in the signature...

src/ripple/overlay/impl/ProtocolMessage.h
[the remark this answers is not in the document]
nbougalis: I think that gRPC is probably overkill and will require more rearchitecting than I'd like to do at this time. But I will take a look.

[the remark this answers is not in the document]
MarkusTeufelberger: I just started looking into libp2p as well (used by Ethereum 2.0, IPFS etc.), but there's no C++ implementation so far and I suspect they are also vulnerable to active MITM with their TLS handshake.

Cap'n'Proto might also be worth a look, gRPC has the advantage that it might also take care of the transport stuff and is probably more widely used and implemented in other software. Wouldn't solve the MITM handshake mitigation, but there were options discussed for that already.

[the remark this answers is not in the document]
MarkusTeufelberger: Ok, libp2p's handshake is not vulnerable and also prevents MITM attacks with a standard certificate based approach. The only extra part is that one needs to check that the host key embedded in the certificate did actually sign the public key of the certificate (and that the connection is using this key). That's more related to establishing a connection though, not so much to the stream of protobuf messages here.

src/ripple/overlay/README.md
```diff
@@ -147,41 +134,128 @@ Content-Type: application/json
...
+### Session Signature
+
+Even for SSL/TLS encrypted connections, it is possible for an attacker to mount
+relatively inexpensive MITM attacks that can be extremely hard to detect and
+may afford the attacker the ability to intelligently tamper with messages
+exchanged between the two endpoints.
+
+This risk can be mitigated if at least one side has a certificate from a certificate
+authority trusted by the other endpoint, but having a certificate is not always
+possible (or even desirable) in a decentralized and permissionless network.
```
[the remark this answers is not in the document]
nbougalis: It's possible, but additional code and additional code isn't always a good idea.

src/ripple/overlay/README.md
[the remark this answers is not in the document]
nbougalis: Try testing against this branch. You should _reliably_ get an error message that is descriptive. If you don't, let me know.

src/ripple/overlay/README.md
```diff
@@ -147,41 +134,128 @@ Content-Type: application/json
...
+
+To do this we "reach into" the SSL/TLS session, and extract the **`finished`**
+messages for the local and remote endpoints, and combine them to generate a unique
+"fingerprint". By design, this fingerprint should be the same for both SSL/TLS
+endpoints.
+
+That fingerprint, which is never shared over the wire (since each endpoint will
+calculate it indepedently), is then signed by each server using its public
+**`secp256k1`** node identity and the signature is transferred over the SSL/TLS
+encrypted link during the protocol handshake phase.
```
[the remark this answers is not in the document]
nbougalis: Correct.

src/ripple/protocol/impl/BuildInfo.cpp
```diff
@@ -49,46 +49,9 @@ char const* const versionString = "1.3.1"
...
-    );
-
-    return minimumProtocol;
-}
-
-//
 //
 // Don't touch anything below this line
 //
-//------------------------------------------------------------------------------
```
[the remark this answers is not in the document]
nbougalis: Really? To me, "this line" means "this line" (i.e. line 53) and not "the line two lines below."

src/ripple/overlay/impl/ProtocolVersion.cpp
```diff
@@ -0,0 +1,140 @@
...
+};
+
+constexpr
+ProtocolVersion const legacyProtocol { 1, 2 };
+
+std::string
+to_string(ProtocolVersion const& p)
+{
+    if (p == legacyProtocol)
+        return "RTXP/" + std::to_string(p.first) + "." + std::to_string(p.second);
```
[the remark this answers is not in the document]
nbougalis: I'm fine with micro-optimizing. Also, it's easier to read, I guess.

src/ripple/overlay/impl/ProtocolMessage.h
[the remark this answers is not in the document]
nbougalis: While you can certainly address this issue using certificates, I still think the extra work that users have to go through makes it unlikely that they will. The primary objection to the existing MITM mitigation technique seems to be "yuck, I don't like how this is done." And that's fine but it's not really disposative.

src/ripple/protocol/impl/BuildInfo.cpp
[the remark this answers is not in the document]
MarkusTeufelberger: Ah, I thought "this line" would refer to the actual line of `-----` below.

src/ripple/overlay/impl/ProtocolMessage.h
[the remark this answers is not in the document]
MarkusTeufelberger: No, it is "this is not possible without adding a patch to Python's stdlib" in my case. Accessing these messages is just not possible in a lot of implementations of TLS, verifying a special field in a certificate is.

The certs in libp2p are self-signed and can be created by the application ad-hoc, so there's no work on the user's end...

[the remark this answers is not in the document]
nbougalis: OK. I'll give this some further thought.

src/ripple/overlay/impl/ProtocolMessage.h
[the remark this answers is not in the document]
movitto: From my experience ++i is usually preferred over i++ unless needed anyways. A while back there was discussion pertaining to a rippled code style guide, did that ever materialize?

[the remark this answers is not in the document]
MarkusTeufelberger: I'm just wondering since in all 3 loops `i` isn't used, but 2 use post-increment and this one pre-increment. Seemed odd, but is apparently more personal style.

movitto:
> Phew, that took much longer to read through and understand than expected.
> 
> Thanks for the PR!
> 
> I'd really like to see at least some options besides signing the encoded TLS "finished" messages available. RFC5705 sounds more reasonable to me, or also just offering to use actual widely trusted certificates, if available (e.g. via Let's Encrypt).

Just jumping in here as I start to take a look at this. We've already implemented the handshake protocol in the[ ruby library](https://github.com/DevNullProd/XRBP/blob/master/lib/xrbp/overlay/handshake.rb) and I'd like to keep that up to date.

Reading RFC5705, it specified the necessity for a 'master_secret' and 'label' for the EKM context, what were you thinking for those?

I'd personally vote _against_ certificates as it's a centralizing component and the whole point on the P2P protocol is to be decentralized.

MarkusTeufelberger:
master_secret and label can be public values as far as I understand it.

Certificates don't have to be signed by a CA, they can be self-signed. Take a look at libp2p's TLS 1.3 handshake protocol for example. Nothing centralized about that one.

src/ripple/overlay/impl/ProtocolMessage.h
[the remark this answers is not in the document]
nbougalis: Huh, I didn't notice that some used pre and others did post. I agree that's a weird. I'll pick one and stick with it.

src/ripple/overlay/impl/ProtocolMessage.h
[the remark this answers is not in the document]
nbougalis: Hmm. This actually does give me an idea. I may do this, but it'll have to wait for a follow-up PR.

nbougalis:
> Just jumping in here as I start to take a look at this. We've already implemented the handshake protocol in the[ ruby library](https://github.com/DevNullProd/XRBP/blob/master/lib/xrbp/overlay/handshake.rb) and I'd like to keep that up to date.

That's neat. Even if we implement something different, it's likely that this mechanism will stick around for a while, and you'll have plenty of time to implement it. Also, _hopefully_ it'll be something that doesn't require custom ruby gems.

> Reading RFC5705, it specified the necessity for a 'master_secret' and 'label' for the EKM context, what were you thinking for those?

I'm not convinced the RFC5705 solution(s) prevent the kind of attack that our simple construct prevents anyways. I have technique that _can_ do it and which doesn't require reaching into the SSL session and pulling something out of it, but it's a more complicated process than what we do currently requiring at least 5 messages (3 in one direction, 2 in the other direction) to be exchanged.

> I'd personally vote _against_ certificates as it's a centralizing component and the whole point on the P2P protocol is to be decentralized.

I think that we could implement something where the server creates a self-signed certificate (if none exists) and it would be no more difficult for the user, but I need to carefully consider the ramifications and ensure that we get the security guarantees we want.

src/ripple/overlay/impl/Handshake.cpp
```diff
@@ -0,0 +1,302 @@
...
+    boost::beast::http::fields const& headers,
+    const ripple::uint256& sharedValue,
+    boost::optional<std::uint32_t> networkID,
+    beast::IP::Address public_ip,
+    beast::IP::Address remote,
+    Application& app)
+{
+    if (networkID)
+    {
+        auto const iter = headers.find("Network-ID");
```
seelabs: We could use C++-17's `if (init; cond)` for. I.e. `if (auto const iter = headers.find("Network-ID"); iter != headers.end())`. There are a couple of other places it could be used as well, I'll mark them with a terse: could use `if (init; cond)` here. No change needed if you prefer to keep the init outside.

src/ripple/overlay/impl/Handshake.cpp
```diff
@@ -0,0 +1,302 @@
...
+            if (!beast::lexicalCastChecked(nid, iter->value().to_string()))
+                throw std::runtime_error("Invalid peer network identifier");
+
+            if(nid != *networkID)
+                throw std::runtime_error("Peer is on a different network");
+        }
+    }
+
+    {
+        auto const iter = headers.find("Network-Time");
```
seelabs: Could use `if (init; cond)` here

src/ripple/overlay/impl/Handshake.cpp
```diff
@@ -0,0 +1,302 @@
...
+                throw std::runtime_error("Invalid Local-IP");
+
+            if (beast::IP::is_public(remote) && remote != local_ip)
+                throw std::runtime_error("Incorrect Local-IP: " +
+                    remote.to_string() + " instead of " + local_ip.to_string());
+        }
+    }
+
+    {
+        auto const iter = headers.find("Remote-IP");
```
seelabs: Could use `if (init; cond)` here

src/ripple/overlay/impl/PeerImp.cpp
```diff
@@ -113,8 +113,59 @@ void
...
+        return uint256{ s };
+    };
+
+    boost::optional<uint256> closed;
+    boost::optional<uint256> previous;
+
+    {
+        auto const iter = headers_.find("Closed-Ledger");
+
+        if (iter != headers_.end())
```
seelabs: Could use `if (init; cond)` here

src/ripple/overlay/impl/PeerImp.cpp
```diff
@@ -113,8 +113,59 @@ void
...
+        {
+            closed = parseLedgerHash(iter->value().to_string());
+
+            if (!closed)
+                fail("Malformed handshake data (1)");
+        }
+    }
+
+    {
+        auto const iter = headers_.find("Previous-Ledger");
```
seelabs: Could use `if (init; cond)` here

src/ripple/overlay/impl/Handshake.cpp
```diff
@@ -0,0 +1,302 @@
...
+
+    {
+        auto const iter = headers.find ("Public-Key");
+
+        if (iter != headers.end())
+            publicKey = parseBase58<PublicKey>(TokenType::NodePublic, iter->value().to_string());
+    }
+
+    if (!publicKey)
+        throw std::runtime_error("Bad node public key");
```
seelabs: Could use a lambda that throws and no optional. Something like:
```
    PublicKey const publicKey = [&headers]
    {
        if (auto const iter = headers.find ("Public-Key");
        iter != headers.end()){
            if (auto pk = parseBase58<PublicKey>(TokenType::NodePublic, iter->value().to_string())
              return *pk;
         }
         throw std::runtime_error("Bad node public key");
    }();
```

src/ripple/overlay/README.md
```diff
@@ -24,71 +25,69 @@ messages are exchanged between peers and serialized using
...
-The request uses the [_HTTP/1.1 Upgrade_][upgrade_header] mechanism with some
-custom fields to communicate protocol specific information:
+##### Example HTTP Upgrade Request
 
 ```
 GET / HTTP/1.1
-User-Agent: rippled-0.27.0
-Local-Address: 192.168.0.101:8421
-Upgrade: RTXP/1.2, RTXP/1.3
+User-Agent: rippled-1.4.0-b1+DEBUG
```
seelabs: I'd leave off the `-b1+DEBUG`

src/ripple/overlay/impl/Handshake.cpp
```diff
@@ -0,0 +1,302 @@
...
+        return boost::none;
+    }
+
+    return sha512Half (Slice (result.data(), result.size()));
+}
+
+void
+buildHandshake(
+    boost::beast::http::fields& h,
+    const ripple::uint256& sharedValue,
```
seelabs: Nit: use east const

src/ripple/overlay/impl/Handshake.cpp
```diff
@@ -0,0 +1,302 @@
...
+            cl->info().hash.begin(), cl->info().hash.size()));
+        h.insert ("Previous-Ledger", base64_encode(
+            cl->info().parentHash.begin(), cl->info().parentHash.size()));
+    }
+}
+
+PublicKey
+verifyHandshake(
+    boost::beast::http::fields const& headers,
+    const ripple::uint256& sharedValue,
```
seelabs: Nit: use east const

src/ripple/overlay/impl/Handshake.cpp
```diff
@@ -0,0 +1,302 @@
...
+            {
+                boost::optional<TimeKeeper::time_point> ret;
+
+                TimeKeeper::duration::rep val;
+
+                if (beast::lexicalCastChecked(val, str))
+                    ret.emplace(TimeKeeper::duration{val});
+
+                return ret;
+            }();
```
seelabs: Could move the `throw` inside the lambda and get rid of the `optional`. I'd also put a return type on the lambda, since with the re-write it's not immediately obviously what the return type is Something like:
```
            auto const netTime =
                    [&iter]() -> TimeKeeper::time_point {
                TimeKeeper::duration::rep val;
                if (beast::lexicalCastChecked(val, iter->value().to_string()))
                    return TimeKeeper::time_point{TimeKeeper::duration{val}};
                // It's not an error for the header field to not be present but if
                // it is present and it contains junk data, that is an error.
                throw std::runtime_error("Invalid peer clock timestamp");
            }();
```

src/ripple/overlay/impl/OverlayImpl.cpp
```diff
@@ -237,88 +237,84 @@ OverlayImpl::onHandoff (std::unique_ptr <beast::asio::ssl_bundle>&& ssl_bundle,
         }
     }
 
-    auto hello = parseHello (true, request, journal);
-    if(! hello)
+    auto negotiatedVersion = negotiateProtocolVersion(request["Upgrade"]);
```
seelabs: Could be `const`

src/ripple/overlay/impl/OverlayImpl.cpp
```diff
@@ -237,88 +237,84 @@ OverlayImpl::onHandoff (std::unique_ptr <beast::asio::ssl_bundle>&& ssl_bundle,
...
-            remote_endpoint.address(),
-            "Unable to parse HELLO message");
+            remote_endpoint.address(), "Unable to agree on a protocol version");
         handoff.keep_alive = false;
         return handoff;
     }
 
-    auto sharedValue = makeSharedValue(
-        ssl_bundle->stream.native_handle(), journal);
+    auto sharedValue = makeSharedValue(*ssl_bundle, journal);
```
seelabs: Could be `const`

src/ripple/overlay/impl/Handshake.cpp
```diff
@@ -0,0 +1,302 @@
...
+
+        if (iter != headers.end())
+        {
+            std::uint32_t nid;
+
+            if (!beast::lexicalCastChecked(nid, iter->value().to_string()))
+                throw std::runtime_error("Invalid peer network identifier");
+
+            if(nid != *networkID)
+                throw std::runtime_error("Peer is on a different network");
```
seelabs: In general, we don't throw on "external input". This throws based on the peer protocol. I understand this is a different level of trust than an RPC command. Are we OK with this throw (and others in this function)?

src/ripple/overlay/impl/OverlayImpl.cpp
```diff
@@ -237,88 +237,84 @@ OverlayImpl::onHandoff (std::unique_ptr <beast::asio::ssl_bundle>&& ssl_bundle,
...
         return handoff;
     }
 
-    auto publicKey = verifyHello (*hello,
-        *sharedValue,
-        setup_.public_ip,
-        beast::IPAddressConversion::from_asio(
-            remote_endpoint), journal, app_);
-    if(! publicKey)
+    try
```
seelabs: I believe this `try` block is based on the fact that `verifyHandshake` now throws. Since it has a much larger scope than just the `verifyHandshake` call, once we add it's much more difficult to remove it since we have to audit all the calls in the scope for throws.

Reducing the scope of the `try` makes the code uglier, I'm not advocating that; I am pointing out that once we decide `verifyHandshake` throws it's hard to reverse that decision.

src/ripple/overlay/impl/PeerImp.cpp
```diff
@@ -113,8 +113,59 @@ void
...
+            if (!previous)
+                fail("Malformed handshake data (2)");
+        }
+    }
+
+    if (previous && !closed)
+        fail("Malformed handshake data (3)");
+
+    {
+        std::lock_guard<std::mutex> sl(recentLock_);
```
seelabs: Just `std::lock_guard` works (class template argument deduction works well with `lock_guard`)

src/ripple/overlay/impl/PeerImp.cpp
```diff
@@ -709,47 +727,28 @@ void PeerImp::doAccept()
...
-    }
 
     onWriteResponse(error_code(), 0);
 }
 
 http_response_type
 PeerImp::makeResponse (bool crawl,
     http_request_type const& req,
-    beast::IP::Endpoint remote,
+    beast::IP::Address remote_ip,
```
seelabs: Nit: `snake_case` var. Tho I understand why with all the types in `snake_case`

src/ripple/overlay/impl/ProtocolMessage.h
```diff
@@ -106,80 +155,98 @@ std::pair <std::size_t, boost::system::error_code>
...
+    // We don't have the whole message yet. This isn't an error but we have
+    // nothing to do.
+    if (header->total_wire_size > size)
         return result;
 
-    auto const type = Message::type(buffers);
+    bool success;
 
-    switch (type)
+    switch (header->message_type)
```
seelabs: Could re-write this as an immediately invoked lambda that returns `success`.

src/ripple/overlay/impl/ProtocolMessage.h
```diff
@@ -106,80 +155,98 @@ std::pair <std::size_t, boost::system::error_code>
...
-    case protocol::mtLEDGER_DATA:           ec = detail::invoke<protocol::TMLedgerData> (type, buffers, handler); break;
-    case protocol::mtPROPOSE_LEDGER:        ec = detail::invoke<protocol::TMProposeSet> (type, buffers, handler); break;
-    case protocol::mtSTATUS_CHANGE:         ec = detail::invoke<protocol::TMStatusChange> (type, buffers, handler); break;
-    case protocol::mtHAVE_SET:              ec = detail::invoke<protocol::TMHaveTransactionSet> (type, buffers, handler); break;
-    case protocol::mtVALIDATION:            ec = detail::invoke<protocol::TMValidation> (type, buffers, handler); break;
-    case protocol::mtGET_OBJECTS:           ec = detail::invoke<protocol::TMGetObjectByHash> (type, buffers, handler); break;
+    case protocol::mtMANIFESTS:
+        success = detail::invoke<protocol::TMManifests>(*header, buffers, handler);
+        break;
+    case protocol::mtPING:
```
seelabs: Many of these case statements have trailing white space.

src/ripple/overlay/impl/ProtocolVersion.cpp
```diff
@@ -0,0 +1,140 @@
...
+
+    return "XRPL/" + std::to_string(p.first) + "." + std::to_string(p.second);
+}
+
+std::vector<ProtocolVersion>
+parseProtocolVersions(boost::beast::string_view const& value)
```
seelabs: Document that this function must return a sorted list (or `negotiateProtocolVersion` breaks)

src/ripple/overlay/impl/ProtocolVersion.cpp
```diff
@@ -0,0 +1,140 @@
...
+
+#include <ripple/overlay/impl/ProtocolVersion.h>
+#include <ripple/beast/core/LexicalCast.h>
+#include <ripple/beast/rfc2616.h>
+#include <boost/regex.hpp>
+#include <algorithm>
+
+namespace ripple {
+
+/** The list of protocol versions we speak and we prefer to use. */
```
seelabs: Document that this list must be sorted (or `negotiateProtocolVersion` breaks

src/ripple/overlay/impl/ProtocolVersion.cpp
```diff
@@ -0,0 +1,140 @@
...
+        "(0|[1-9][0-9]*)"    // a number (no leading zeroes unless exactly zero)
+        ")"
+        "$"                  // The end of the string
+        , boost::regex_constants::optimize);
+
+    static std::string const legacy = to_string(legacyProtocol);
+
+    std::vector<ProtocolVersion> result;
+
+    for (auto const& s : beast::rfc2616::split_commas(value))
```
seelabs: There's no error handling here. For example, if we pass a list of versions with "junk on the end" it ignores it. Is this OK?

src/ripple/overlay/impl/ProtocolVersion.cpp
```diff
@@ -0,0 +1,140 @@
...
+    result.erase(std::unique(result.begin(), result.end()), result.end());
+    return result;
+}
+
+boost::optional<ProtocolVersion>
+negotiateProtocolVersion(boost::beast::string_view const& versions)
+{
+    auto const them = parseProtocolVersions(versions);
+
+    std::vector<ProtocolVersion> common;
```
seelabs: `reserve` here (or maybe use a `boost::small_vector` to try to avoid the allocation in most cases)

src/ripple/overlay/impl/ProtocolVersion.cpp
```diff
@@ -0,0 +1,140 @@
...
+        return ret;
+    }();
+
+    return supported;
+}
+
+bool
+isProtocolSupported(ProtocolVersion const& v)
+{
+    return std::end(supportedProtocolList) != std::find(
```
seelabs: Nit: this reads like a yoda condtion. No change needed.

src/ripple/overlay/README.md
```diff
@@ -101,87 +100,252 @@ Content-Type: application/json
...
+specify the version of the software that is used.
+
+See [RFC2616 &sect;14.43](https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.43).
+
+| Field Name          	|  Request          	| Response          	|
+|---------------------	|:-----------------:	|:-----------------:	|
+| `Server`            	|                   	| :heavy_check_mark: 	|
+
+The `Server` field indicates the version of the software that the
+peer that is processig the HTTP request is using. No semantic meaning is
```
seelabs: sp: `processig` -> `processing`

src/ripple/overlay/impl/OverlayImpl.cpp
```diff
@@ -366,18 +359,18 @@ OverlayImpl::makeRedirectResponse (PeerFinder::Slot::ptr const& slot,
...
+OverlayImpl::makeErrorResponse (std::shared_ptr<PeerFinder::Slot> const& slot,
     http_request_type const& request,
     address_type remote_address,
     std::string text)
 {
-    boost::beast::http::response<boost::beast::http::string_body> msg;
+    boost::beast::http::response<boost::beast::http::empty_body> msg;
     msg.version(request.version());
     msg.result(boost::beast::http::status::bad_request);
+    msg.reason("Bad Request (" + text + ")");
```
seelabs: According to the docs on `reason` this function is deprecated. Ref:
```

    /** Set the response reason-phrase (deprecated)

        This function sets a custom reason-phrase to a copy of
        the string passed in. Normally it is not necessary to set
        the reason phrase on an outgoing response object; the
        implementation will automatically use the standard reason
        text for the corresponding status code.

        To clear a previously set custom phrase, pass an empty
        string. This will restore the default standard reason text
        based on the status code used when serializing.

        The reason-phrase is obsolete as of rfc7230.

        @param s The string to use for the reason-phrase.

        @note This function is only available when `isRequest == false`.
    */
 ```

File: src/ripple/overlay/impl/OverlayImpl.cpp [hunk 43]
```diff
@@ -366,18 +359,18 @@ OverlayImpl::makeRedirectResponse (PeerFinder::Slot::ptr const& slot,
...
     msg.insert("Server", BuildInfo::getFullVersionString());
     msg.insert("Remote-Address", remote_address.to_string());
     msg.insert(boost::beast::http::field::connection, "close");
-    msg.body() = text;
     msg.prepare_payload();
     return std::make_shared<SimpleWriter>(msg);
```
seelabs: `SimpleWriter` is also deprecated, tho I don't know what replaces it (and obviously not part of this patch).

src/ripple/overlay/impl/PeerImp.cpp
nbougalis: :open_mouth: I didn't know that.

src/ripple/overlay/impl/OverlayImpl.cpp
nbougalis: I understand that the reason phrase is deprecated, but we have no better way to communicate an actual error message (short of adding a custom header) that gets displayed to the user.

This is a legit problem with existing `rippled` instances; if you get disconnected for, say, the clock being too far off, nothing gets logged on your end which can make debugging difficult.

src/ripple/overlay/impl/Handshake.cpp
nbougalis: Fixed the ones that make sense. :+1:

src/ripple/overlay/impl/Handshake.cpp
```diff
@@ -0,0 +1,302 @@
...
+        }
+    }
+
+    {
+        auto const iter = headers.find("Network-Time");
+
+        if (iter != headers.end())
```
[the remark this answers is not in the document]
nbougalis: The field wasn't required before and should not be required. Thanks.

src/ripple/overlay/impl/ProtocolMessage.h
```diff
@@ -63,33 +62,83 @@ protocolMessageName (int type)
...
+    BufferSequence const& bufs,
+    std::size_t size)
+{
+    auto iter = boost::asio::buffers_iterator<BufferSequence, std::uint8_t>::begin(bufs);
+
+    MessageHeader hdr;
+
+    // Version 1 header: uncompressed payload.
+    // The top six bits of the first byte are 0.
+    if ((*iter & 0xFC) == 0)
```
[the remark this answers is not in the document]
nbougalis: We could certainly do that. I have a different plan for compression in mind though and I'll probably repurpose the high bit to indicate compression when I do that PR.

src/ripple/overlay/impl/ProtocolVersion.cpp
nbougalis: We already do, in the header file, but I've made the wording more explicit and added a comment here.

src/ripple/overlay/impl/ProtocolVersion.cpp
nbougalis: On purpose, it was. More difficult to read the other way around would be. And more error prone.

src/ripple/overlay/impl/ProtocolVersion.cpp
nbougalis: I'm not sure I follow. This code will crack a given input string at the commas, and for each item, it will call `boost::regex_match` and we only do work if the item matches the regex description. Only two kinds of entries will pass:

1. **`RTXP/1.2`**
2. **`XRPL/x.y`** where `x ≥ 2` and `y ≥ 0`.

src/ripple/overlay/impl/ProtocolVersion.cpp
```diff
@@ -0,0 +1,142 @@
...
+    std::is_sorted(supportedProtocolList.begin(), supportedProtocolList.end()),
+        "The list of supported protocols must be in sorted order.");
+#endif
+
+std::string
+to_string(ProtocolVersion const& p)
+{
+    // The legacy protocol uses a different name. This can be removed when we
+    // migrate away from it and require 2.0 or later.
+    if (p == { 1, 2 })
```
seelabs: Compile error here (tho `if (operator==(p, { 1, 2 }))` works). I'd just put `ProtocolVersion{1,2}`.

src/ripple/overlay/impl/ProtocolVersion.cpp
seelabs: I was considering the case: `RTXP/1.2, GOBBLYGOOK`. But now that I think about it, of course you want to silently ignore `GOBBLYGOOK`.

MarkusTeufelberger: What about `XRPL/10.0`?

seelabs: @MarkusTeufelberger I agree, that looks broken. Good find!

src/ripple/overlay/impl/ProtocolVersion.cpp
```diff
@@ -0,0 +1,142 @@
...
+        std::distance(
+            std::begin(supportedProtocolList),
+            std::end(supportedProtocolList)));
+    std::set_intersection(
+        them.begin(), them.end(),
+        std::begin(supportedProtocolList),
+        std::end(supportedProtocolList),
+        std::back_inserter(common));
+
+    auto x = std::max_element(common.begin(), common.end());
```
seelabs: The largest element will always be on the end, no need to search.

I just thought of something "cute". First, two observations:
1) We don't need to allocate at all. If we use the same algorithm as `std::set_intersection`, use reverse iterators, the result is always the first element that would be written. `std::set_intersection` isn't set up to do that, and it isn't worth it to write a custom version.
2) If we use `set_intersection` as-is and provide a custom iterator that calls a function, we can get most of the same thing. We could write to an optional on every iteration. The result is the final value of the optional. Something like:
```

boost::optional<ProtocolVersion>
negotiateProtocolVersion(boost::beast::string_view const& versions)
{
    auto const them = parseProtocolVersions(versions);
    boost::optional<ProtocolVersion> result;
    std::set_intersection(
        them.begin(),
        them.end(),
        std::begin(supportedProtocolList),
        std::end(supportedProtocolList),
        boost::make_function_output_iterator(
            [&result](auto const& e) { result = e; }));
    return result;
}
```
(we need to include `#include <boost/function_output_iterator.hpp>` for that to work).

BTW, I'm 100% fine with the current code (my "cute" code takes a minute to decipher while yours is clear). Just wanted to share the thought.

src/ripple/overlay/impl/ProtocolVersion.cpp
nbougalis: Oops. I squashed things wrong. Thanks.

src/ripple/overlay/impl/ProtocolVersion.cpp
nbougalis: I actually like that a lot.

src/ripple/overlay/impl/ProtocolVersion.cpp
nbougalis: I'm not sure I follow... **`XRPL/10.0`** is theoretically a valid protocol to request. The regular expression accounts for it:

    "XRPL/"              // the string "XRPL/"
    "([2-9][0-9]*)"      // a number (greater than 2 with no leading zeroes)
    "\\."                // a period
    "(0|[1-9][0-9]*)"    // a number (no leading zeroes unless exactly zero)

So, a string like `RTXP/1.2, GOBBLYGOOK, XRPL/2.0, XRPL/10.0` will, correctly, produce the following list: **`{ { 1, 2 }, { 2, 0 }, { 10, 0 } }`**

nbougalis: Fixed by upgrading the regex from:
`([2-9][0-9]*)` to:
`([2-9]|[1-9][0-9]+)`

Thanks @MarkusTeufelberger.

ximinez:
Just FYI, VS doesn't like the `make_function_output_iterator`
https://travis-ci.com/ripple/rippled/jobs/231590174#L860. I'm not sure how to fix it.

ximinez:
This commit works around the compile error I pointed out above by doing it "the hard way" for MSVC builds. https://github.com/ximinez/rippled/commit/c974971d3fa9c775cd08fc282fb7aeb78378f365

src/test/overlay/TMHello_test.cpp
```diff
@@ -1,70 +0,0 @@
...
-    }
-
-    void
-    run() override
-    {
-        test_protocolVersions();
-    }
-};
-
-BEAST_DEFINE_TESTSUITE(TMHello,overlay,ripple);
```
ximinez: I am always reluctant to get rid of unit tests without a really good reason. I don't think you've entirely removed the functionality being tested here, so why remove the tests?

src/test/protocol/BuildInfo_test.cpp
```diff
@@ -1,106 +0,0 @@
...
-
-        BEAST_EXPECT(current_protocol >= minimum_protocol);
-
-        log <<
-            "   Ripple Version: " << BuildInfo::getVersionString() << '\n' <<
-            " Protocol Version: " << to_string (current_protocol) << std::endl;
-    }
-};
-
-BEAST_DEFINE_TESTSUITE(BuildInfo,ripple_data,ripple);
```
ximinez: As above, I am always reluctant to get rid of unit tests without a really good reason. I don't think you've entirely removed the functionality being tested here, so why remove the tests?

nbougalis: This code changes `ProtocolVersion` to be a simple `std::pair` and all the relevant operators that were being tested here before are now provided by std so I don't believe that testing _that_ functionality is necessary.

The only test that is no longer there is BEAST_EXPECT(current_protocol >= minimum_protocol) and the new implementation of this code simply doesn't have such values in a way that makes sense to test. Rather the implementation requires that all supported versions are listed in a single array `supportedProtocolList`. I had a `static_assert` that was meant to test, but it turns out I had incorrectly gated it so that it would require compiling with C++20; I rewrote the `static_assert` to execute under C++17 and verified that it works correctly.

src/ripple/overlay/impl/ProtocolMessage.h
```diff
@@ -106,80 +155,98 @@ std::pair <std::size_t, boost::system::error_code>
...
+        break;
+    case protocol::mtHAVE_SET:
+        success = detail::invoke<protocol::TMHaveTransactionSet>(*header, buffers, handler);
+        break;
+    case protocol::mtVALIDATION:
+        success = detail::invoke<protocol::TMValidation>(*header, buffers, handler);
+        break;
+    case protocol::mtGET_OBJECTS:
+        success = detail::invoke<protocol::TMGetObjectByHash>(*header, buffers, handler);
+        break;
```
ximinez: I noticed that in rewriting this `switch`, you got rid of three deprecated message types: `mtHELLO`, `mtGET_PEERS`, and `mtPEERS`. While you're at it, why not go all the way and remove those entries from the enum and all instances in other places, too? At least on mainnet, any node running a version old enough to send these is surely going to be amendment locked by now, *and* the `handler.onMessageUnknown` doesn't do anything anyway, so even on other networks, it won't hurt anything.

nbougalis: Hmm, I thought I did. I'll double-check.

nbougalis:
> I also fixed the Windows build by ripping off the `boost::function_output_iterator` with an iterator that skips the lambda and does the assignment directly. It's probably not the best option, but it works.

The change in question is: https://github.com/ximinez/rippled/commit/878b1d8c14c8d79a03a7554891ec310727d54446

I appreciate you doing this, but it seems to me that it is overkill to address what is, at the end of the day, either a compiler deficiency or an issue with Boost against Windows. The fix that creates a `std::function` to wrap the lambda is more compact and easy to understand IMO.

Unless you feel particularly strong about your approach, I'll stick with what I have, @ximinez.

ximinez:
> I appreciate you doing this, but it seems to me that it is overkill to address what is, at the end of the day, either a compiler deficiency or an issue with Boost against Windows. The fix that I put in place after your initial report (creating a `std::function` to wrap the lambda) is more compact and easy to understand IMO. So unless you feel particularly strong about your approach, I'll stick with what I have, @ximinez.

Welp, egg on my face, because I'm having no problems building at all now. I have no idea what changed, but appears to be a moot point either way.

src/ripple/overlay/impl/ProtocolMessage.h
nbougalis: Removed some leftover instances of `mtGET_PEERS` and `mtPEERS` in response this comment.

src/ripple/overlay/impl/ProtocolVersion.cpp
```diff
@@ -0,0 +1,184 @@
...
+        return true;
+    }(), "The list of supported protocols isn't properly sorted.");
+
+
+std::string
+to_string(ProtocolVersion const& p)
+{
+    // The legacy protocol uses a different name. This can be removed when we
+    // migrate away from it and require 2.0 or later.
+    if (p == ProtocolVersion{ 1, 2 })
```
ximinez: It's only used three times, but you could define a `constexpr ProtocolVersion legacyVersion{ 1, 2 }` and use it in those few places. When we migrate away from it, remove the variable.

src/ripple/overlay/impl/ProtocolVersion.h
```diff
@@ -0,0 +1,79 @@
...
+using ProtocolVersion = std::pair<std::uint16_t, std::uint16_t>;
+
+inline
+ProtocolVersion
+make_protocol(std::uint16_t major, std::uint16_t minor)
+{
+    return { major, minor };
+}
+
+/** Print a protocol version a human-readable string. */
```
ximinez: `/** Return a protocol version as a human-readable string. */`

PR #3054, opened 2019-08-19
Cleanups 17

seelabs:
This is a set of miscellaneous refactoring using c++-17 features. This will be tedious to review. If reviewers think the added benefit isn't worth the time to review I'm happy to pull the offending patches.

Note the `Fix shadow warnings` commit turns on the `-Wshadow` warning. This is to flag any shadowed variables in the structured bindings refactoring (the biggest danger I can think of in that patch). I'm happy to turn this warning back off if there are objections.


src/ripple/app/tx/impl/ApplyContext.cpp
```diff
@@ -103,17 +103,16 @@ ApplyContext::checkInvariantsHelper(
...
-                            visitEntry(index, isDelete, before, after)), 0)...}
-                    };
+                    (..., std::get<Is>(checkers).visitEntry(index, isDelete, before, after));
                 });
 
-            // Sean Parent for_each_argument trick (a fold expression with `&&`
-            // would be really nice here when we move to C++-17)
-            std::array<bool, sizeof...(Is)> finalizers {{
-                std::get<Is>(checkers).finalize(tx, result, fee, journal)...}};
+            // Note: do not replace this logic with a `...&&` fold expression.
```
scottschurr: Nice comment!  Thanks!

src/ripple/app/ledger/Ledger.cpp
```diff
@@ -873,19 +875,20 @@ static bool saveValidatedLedger (
 
         std::string const ledgerSeq (std::to_string (seq));
 
-        for (auto const& vt : aLedger->getMap ())
+        for (auto const& [_, acceptedLedgerTx] : aLedger->getMap ())
         {
-            uint256 transactionID = vt.second->getTransactionID ();
+            (void)_;
```
JoeLoser: Is this due to older GCC not liking unused structured bindings? Do you prefer this over using `[[maybe_unused]]` on the whole structured decomposition to be clear which variable(s) are not used.

src/ripple/app/misc/Manifest.h
```diff
@@ -371,11 +372,11 @@ class ManifestCache
...
     for_each_manifest(PreFun&& pf, EachFun&& f) const
     {
-        std::lock_guard<std::mutex> lock{read_mutex_};
+        std::lock_guard lock{read_mutex_};
         pf(map_.size ());
-        for (auto const& m : map_)
+        for (auto const& [_, manifest] : map_)
         {
-            f(m.second);
+            f(manifest);
```
JoeLoser: Do we need to explicitly say `(void) _` here?

src/ripple/app/paths/impl/DirectStep.cpp
```diff
@@ -635,8 +631,7 @@ DirectStepI<TDerived>::fwdImp (
     std::tie (maxSrcToDst, srcDebtDir) =
```
JoeLoser: Any reason not to use structured bindings here?

src/ripple/app/tx/impl/ApplyContext.cpp
```diff
@@ -103,17 +103,16 @@ ApplyContext::checkInvariantsHelper(
...
+            // The fold expression will only run until the first check fails (it
+            // short-circuits). While the logic is still correct, the log
+            // message won't be. Every failed invariant should write to the log,
+            // not just the first one.
+            std::array<bool, sizeof...(Is)> finalizers{
```
JoeLoser: Can we use CTAD here to just say `std::array` and let the value type and size be deduced by the RHS?

src/ripple/basics/TaggedCache.h
```diff
@@ -61,7 +61,6 @@ class TaggedCache
     using mutex_type = Mutex;
     // VFALCO DEPRECATED The caller can just use std::unique_lock <type>
     using ScopedLockType = std::unique_lock <mutex_type>;
```
JoeLoser: Perhaps in a future PR it would be good to remove the `ScopedLockType` typedef from this class.

src/ripple/basics/impl/Log.cpp
```diff
@@ -161,11 +161,11 @@ std::vector<std::pair<std::string, std::string>>
...
 {
     std::vector<std::pair<std::string, std::string>> list;
-    std::lock_guard <std::mutex> lock (mutex_);
+    std::lock_guard lock (mutex_);
     list.reserve (sinks_.size());
-    for (auto const& e : sinks_)
-        list.push_back(std::make_pair(e.first,
-            toString(fromSeverity(e.second->threshold()))));
+    for (auto const& [name, sink] : sinks_)
+        list.push_back(std::make_pair(name,
```
JoeLoser: What are your thoughts on changing `std::make_pair` to `std::pair` thanks to CTAD? Similarly, if we have `std::make_tuple` floating around, we could replace with `std::tuple`.

src/ripple/core/JobTypes.h
```diff
@@ -138,14 +138,15 @@ add(    jtNS_WRITE,      "WriteNode",               0,        true,  0ms,     0m
     {
         assert (m_map.find (jt) == m_map.end ());
 
-        std::pair<Map::iterator,bool> result (m_map.emplace (
+        auto const [_, inserted] = m_map.emplace (
```
JoeLoser: In this case, I would probably use `[[maybe_unused]]` on the whole structured binding declaration rather than two `(void)` statements below.

src/ripple/rpc/impl/ServerHandlerImp.cpp
```diff
@@ -247,9 +247,9 @@ build_map(boost::beast::http::fields const& h)
         auto key (e.name_string().to_string());
         // TODO Replace with safe C++14 version
```
JoeLoser: (Not related to this PR), but what does this comment mean exactly? I remember a PR when @HowardHinnant explained the gotchas with the argument to `tolower` (requiring the cast to `unsigned char`), but I don't understand the `TODO` comment.

seelabs:
I see the issues in clang of the form:
```

File: src/ripple/app/ledger/Ledger.cpp (hunk 2 above)
Comment (seelabs) [reply src/ripple/app/ledger/Ledger.cpp]: It appears that neither gcc nor clang are warning about unused structured bindings. I just assumed that they would and pre-emptively added the `(void)_;` statements. I'll remove them.

File: src/ripple/app/misc/Manifest.h (hunk 3 above)
Comment (seelabs) [reply src/ripple/app/misc/Manifest.h]: Apparently gcc and clang aren't warning about this, so no `(void)_` needed, and I'll remove the ones I already added.

File: src/ripple/app/paths/impl/DirectStep.cpp (hunk 4 above)
Comment (seelabs) [reply src/ripple/app/paths/impl/DirectStep.cpp]: No reason, I just missed converting this one. The major reason for some of the ones I didn't convert is you can't capture a structured binding in a lambda. But that doesn't apply here. Will fix.

File: src/ripple/app/tx/impl/ApplyContext.cpp (hunk 5 above)
Comment (seelabs) [reply src/ripple/app/tx/impl/ApplyContext.cpp]: Doesn't work for me:
```
error: no viable constructor or deduction guide for deduction of template arguments of 'array'

src/ripple/basics/TaggedCache.h
seelabs: Agreed. Will fix in this PR.

src/ripple/app/ledger/Ledger.cpp
JoeLoser: Even better. 👍 

I recall running into issues with unused parts of a structured binding when I first started using C++17 on older compilers last year; IIRC it was only with GCC 6.

src/ripple/basics/impl/Log.cpp
seelabs: I'll change this to `emplace_back` and get rid of the `std::pair`. Will fix.

src/ripple/core/JobTypes.h
seelabs: I'm going to get rid of the `(void)` statements all together. I'll add them back (or add `[[maybe_unused]]` if compilers start warning about them.

src/ripple/app/paths/impl/DirectStep.cpp
JoeLoser: Yeah, it's unfortunate that we still can't capture structured bindings. It may get fixed in C++23 as there has been a [CWG issue](https://wg21.cmeerw.net/cwg/issue2308) for a while and there's no technical reason to disallow it AFAIK. [P1091](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p1091r3.html) will fix this if it makes it into C++23.

src/ripple/app/tx/impl/ApplyContext.cpp
JoeLoser: Hmm, perhaps the compilers on CI don't have full deduction guides implemented for all containers? As far as I can tell, it should work fine. Not a biggie, though. :)

src/ripple/app/ledger/Ledger.cpp
seelabs: I only tested on my local system. I'll have to add them back if one of the CI configurations triggers on them. But gcc 7 is the new min version, so if it's only gcc 6 we should be OK.

seelabs: Fixed in `71c328bbc * [fold] Remove "maybe unused" warning work around:`

src/ripple/core/JobTypes.h
seelabs: Fixed in 71c328bbc `* [fold] Remove "maybe unused" warning work around:`

src/ripple/basics/impl/Log.cpp
seelabs: Fixed in 03289817b `[fold] Use structured binding in another place & rm std::pair ctor:`

src/ripple/app/paths/impl/DirectStep.cpp
seelabs: Fixed in 03289817b `[fold] Use structured binding in another place & rm std::pair ctor:`

src/ripple/app/misc/Manifest.h
seelabs: Fixed in 71c328bbc `* [fold] Remove "maybe unused" warning work around:`

src/ripple/basics/TaggedCache.h
seelabs: Fixed in 3fd690eae "[fold] Rm ScopedLockType and use std::unique_lock directly"

src/ripple/app/tx/impl/ApplyContext.cpp
HowardHinnant: Nitpick on the comment:  "The fold expression" is ambiguous.  It could refer to `...,` or `...&&`.  How about:

// The `...&&` fold expression will ...

miguelportilla:
You might also want to change the lock in `beast_io_latency_probe_test.cpp` line 81.

CMakeLists.txt
```diff
@@ -548,6 +548,12 @@ if (MSVC)
 else ()
   # HACK : because these need to come first, before any warning demotion
   string (APPEND CMAKE_CXX_FLAGS " -Wall -Wdeprecated")
+  if (is_clang OR (is_gcc AND NOT unity))
+    # gcc flags many more instances with `-Wshadow` than clang. Many incouous
+    # shadowings are flagged in unity mode. Only turn on gcc warnings in nonunity
+    # builds
+    string (APPEND CMAKE_CXX_FLAGS " -Wshadow")
+  endif()
```
miguelportilla: Should we be enabling c++17 with MSVC?
eg.
```
set_property (
  TARGET ${PROJECT_NAME}
  PROPERTY CXX_STANDARD 17)```

File: src/ripple/app/paths/PathState.cpp [hunk 11]
```diff
@@ -197,7 +197,7 @@ TER PathState::pushNode (
 
     node.uFlags = iType;
     node.issue_.currency = hasCurrency ?
-            currency : backNode.issue_.currency;
+            currency : backNodeOuter.issue_.currency;
```
miguelportilla: Comment on line 266 needs to be updated.
eg.
```// Note: backNodeOuter may no longer be the immediately previous node.```

File: src/ripple/nodestore/impl/codec.h [hunk 12]
```diff
@@ -319,7 +319,7 @@ nodeobject_compress (void const* in,
     using std::runtime_error;
     using namespace nudb::detail;
 
-    std::size_t type = 1;
+    std::size_t const codecType = 1;
```
miguelportilla: Consider `constexpr ` and moving to line ~466.

src/test/app/ValidatorList_test.cpp
```diff
@@ -571,64 +571,64 @@ class ValidatorList_test : public beast::unit_test::suite
...
+        auto trustedKeysOuter = std::make_unique <ValidatorList> (
+            manifestsOuter, manifestsOuter, env.timeKeeper(), env.journal);
 
-        std::vector<std::string> cfgPublishers;
-        hash_set<NodeID> activeValidators;
+        std::vector<std::string> cfgPublishersOuter;
+        hash_set<NodeID> activeValidatorsOuter;
 
-        std::size_t const n = 40;
+        std::size_t const nConst = 40;
```
miguelportilla: Consider constexpr

src/test/csf/Peer.h
```diff
@@ -145,9 +145,9 @@ struct Peer
         }
 
         boost::optional<Ledger>
-        acquire(Ledger::ID const & id)
+        acquire(Ledger::ID const & lid)
```
miguelportilla: consider `lID`

src/ripple/app/consensus/RCLConsensus.cpp
```diff
@@ -551,18 +551,19 @@ RCLConsensus::Adaptor::doAccept(
         // in the previous consensus round.
         //
         bool anyDisputes = false;
-        for (auto& it : result.disputes)
+        for (auto& [_, dispute] : result.disputes)
```
miguelportilla: auto const&

src/ripple/consensus/Consensus.h
```diff
@@ -1635,15 +1633,15 @@ Consensus<Adaptor>::createDisputes(TxSet_t const& o)
 
     int dc = 0;
 
-    for (auto& id : differences)
+    for (auto& [txId, inThisSet] : differences)
```
miguelportilla: auto const&

src/ripple/consensus/Consensus.h
```diff
@@ -1394,27 +1392,27 @@ Consensus<Adaptor>::updateOurPositions()
     // Update votes on disputed transactions
     {
         boost::optional<typename TxSet_t::MutableTxSet> mutableSet;
-        for (auto& it : result_->disputes)
+        for (auto& [txId, dispute] : result_->disputes)
```
miguelportilla: auto const&

src/ripple/rpc/handlers/AccountTx.cpp
```diff
@@ -157,27 +157,26 @@ Json::Value doAccountTx (RPC::Context& context)
                 *account, uLedgerMin, uLedgerMax, bForward, resumeToken, limit,
                 isUnlimited (context.role));
 
-            for (auto& it: txns)
+            for (auto& [txn, txMeta]: txns)
```
miguelportilla: auto const&

src/ripple/shamap/impl/SHAMapSync.cpp
```diff
@@ -386,10 +386,10 @@ SHAMap::getMissingNodes(int max, SHAMapSyncFilter* filter)
             if (mn.stack_.empty() && ! mn.resumes_.empty())
             {
                 // Recheck nodes we could not finish before
-                for (auto& r : mn.resumes_)
-                    if (! r.first->isFullBelow (mn.generation_))
+                for (auto& [innerNode, nodeId] : mn.resumes_)
```
miguelportilla: auto const&

src/ripple/core/impl/TimeKeeper.cpp
```diff
@@ -118,7 +118,7 @@ class TimeKeeperImpl : public TimeKeeper
     std::chrono::duration<std::int32_t>
```
HowardHinnant: Missed one just above here in `nowOffset()`.

src/ripple/nodestore/impl/codec.h
```diff
@@ -301,13 +301,7 @@ template <class = void>
...
 zero32()
 {
-    static std::array<char, 32> v =
-        []
-        {
-            std::array<char, 32> v;
-            v.fill(0);
-            return v;
-        }();
+    static std::array<char, 32> v{};
```
HowardHinnant: Wow, good find!

HowardHinnant: There's another one of these at src/ripple/protocol/impl/STTx.cpp:449.

src/ripple/rpc/impl/ServerHandlerImp.cpp
HowardHinnant: I don't really understand this comment either.  Nor the use of `::tolower`.  We don't explicitly `#include <ctype.h>` (where `::tolower` lives), and we don't normally call the C versions of functions.

I recommend removing the comment and calling `std::tolower`.  There's another `::tolower` at src/ripple/beast/core/LexicalCaast.h:195.

HowardHinnant:
There's one more `// fall through` at src/ripple/app/tz/impl/applySteps.cpp:218.

src/ripple/app/misc/impl/LoadFeeTrack.cpp
```diff
@@ -81,47 +81,18 @@ LoadFeeTrack::lowerLocalFee ()
...
+        std::is_integral_v<T2>>
 >
 void lowestTerms(T1& a,  T2& b)
 {
-    if (a == 0 && b == 0)
-        return;
-
-    std::uint64_t x = a, y = b;
-    while (y != 0)
+    if (auto const gcd = std::gcd(a, b))
```
HowardHinnant: Needs `#include <numeric>`.

CMakeLists.txt
seelabs: @miguelportilla I believe this takes care of turning on c++-17 for us: https://github.com/seelabs/rippled/blob/cleanups-17/CMakeLists.txt#L482 Do I need to do something extra for msvc (I don't have a windows system to test, but CI seems to work).

src/ripple/app/paths/PathState.cpp
seelabs: Fixed in 42d4b481a `Misc fixups in shadow warnings commit`

src/ripple/nodestore/impl/codec.h
seelabs: Fixed in 42d4b481a `Misc fixups in shadow warnings commit`

src/test/app/ValidatorList_test.cpp
seelabs: Fixed in 42d4b481a `Misc fixups in shadow warnings commit`

src/test/csf/Peer.h
seelabs: Fixed in 42d4b481a `Misc fixups in shadow warnings commit`

src/ripple/app/consensus/RCLConsensus.cpp
seelabs: Fixed in a225b62e5 `Add const to some structured bindings`

src/ripple/consensus/Consensus.h
seelabs: Fixed in a225b62e5 `Add const to some structured bindings`

src/ripple/rpc/handlers/AccountTx.cpp
seelabs: Fixed in a225b62e5 `Add const to some structured bindings`

src/ripple/shamap/impl/SHAMapSync.cpp
seelabs: Fixed in a225b62e5 `Add const to some structured bindings`

src/ripple/core/impl/TimeKeeper.cpp
seelabs: Fixed in e43572149 `[fold] Use class template template deduction for lock_guard in TimeKeeper.h`

src/ripple/nodestore/impl/codec.h
seelabs: Thanks!

src/ripple/app/misc/impl/LoadFeeTrack.cpp
seelabs: Fixed in b5a58480b `Add missing include for std::gcd`

CMakeLists.txt
miguelportilla: I had not seen that, thanks. MSVC 2017 is working fine with this PR, so no further change required.

HowardHinnant:
C++17 has `std::optional`.  Why are we using `boost::optional`?

src/ripple/app/ledger/LedgerHistory.cpp
```diff
@@ -60,7 +60,7 @@ LedgerHistory::insert(
```
HowardHinnant: We could also get rid of `make_lock` (src/ripple/basics/make_lock.h) as the only reason this function existed was to deduce the mutex type (which CTAD now does).

seelabs: Nice! Fixed in 7846c24db `[fold] Remove make_lock`

seelabs:
@HowardHinnant I started making a patch for boost::optional -> std::optional. It's mostly mechanical, but there are issues with soci. They may be easy to deal with, but I'd rather punt for now.

HowardHinnant:
Still 👍  at 7846c24dbc4b1a730ebbe61958e6219ef519923d.

src/ripple/rpc/impl/ServerHandlerImp.cpp
seelabs: I don't understand the comment.I think I'm going to leave this issue alone in this PR. We can change it in another PR.

seelabs:
@JoeLoser Unfortunately, we're seeing `unused variable` warnings in CI (I'm not sure why I don't see them locally). At any rate, I'm going to put the `(void)_` checks back in.

seelabs:
Some notes on structured bindings and unused variables:

1) With this patch, gcc stopped warning about unused structured bindings if the underlying bound object is ever used. The motivation for this was since C++-17 doesn't have way to ignore a binding (i.e. like `tie` does) it caused too many false alarms. However, gcc versions before this patch do warn, so we need resolve them.

2) [[maybe_unused]] is not working on structured bindings with gcc 7.4, so the `(void)var;` for must be used.

PR #3056, opened 2019-08-20
Fix jenkins/travis CI:

mellery451:
* remove clang builds from jenkins
* make clang 7 min version
* disable windows travis cache
* limit make parallelism
* update linux CI image

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ services:
 
 env:
   global:
-    - DOCKER_IMAGE="mellery451/rippled-ci-builder:2019-07-18"
+    - DOCKER_IMAGE="mellery451/rippled-ci-builder:2019-08-19"
     - CMAKE_EXTRA_ARGS="-Dwerr=ON"
     - NINJA_BUILD=true
     # change this if we get more VM capacity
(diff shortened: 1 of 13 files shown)


.travis.yml
```diff
@@ -6,7 +6,7 @@ services:
 
 env:
   global:
-    - DOCKER_IMAGE="mellery451/rippled-ci-builder:2019-07-18"
+    - DOCKER_IMAGE="mellery451/rippled-ci-builder:2019-08-19"
```
seelabs: I'd rather get build artifacts from ripple "official" repositories rather than developer specific repositories. No changed needed in this PR, but we should look at changing this in another PR.

Builds/containers/shared/install_cmake.sh
```diff
@@ -2,9 +2,9 @@
 set -ex
 
 cd /tmp
-CM_INSTALLER=cmake-3.13.2-Linux-x86_64.sh
-CM_VER_DIR=/opt/local/cmake-3.13
-wget https://cmake.org/files/v3.13/$CM_INSTALLER
+CM_INSTALLER=cmake-3.15.2-Linux-x86_64.sh
```
seelabs: What is our min version of cmake? Do we have a CI configuration with this min version? No change needed, but the upgrade made me wonder.

CMakeLists.txt
```diff
@@ -61,16 +61,15 @@ endif ()
 
 if ("${CMAKE_CXX_COMPILER_ID}" MATCHES ".*Clang") # both Clang and AppleClang
   set (is_clang TRUE)
-  # TODO enable this version check
   if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang" AND
-         CMAKE_CXX_COMPILER_VERSION VERSION_LESS 5.0)
-    message (FATAL_ERROR "This project requires clang 5 or later")
+         CMAKE_CXX_COMPILER_VERSION VERSION_LESS 7.0)
+    message (FATAL_ERROR "This project requires clang 7 or later")
```
seelabs: I'm a little conflicted about this check. The problem with clang 5 is the interaction with gcc (as you know). I think it's reasonable to say "we only support clang 7 or newer". However, this is refusing to build with older clangs, even if they would work.

What do you think about making this a warning? Keeping it as an error is also reasonable. I suspect there are very few people building with clang 5 anyway.

CMakeLists.txt
```diff
@@ -61,16 +61,15 @@ endif ()
...
   endif ()
   # TODO min AppleClang version check ?
 elseif ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
   set (is_gcc TRUE)
-  if (CMAKE_CXX_COMPILER_VERSION VERSION_LESS 7.0)
-    message (FATAL_ERROR "This project requires GCC 7 or later")
+  if (CMAKE_CXX_COMPILER_VERSION VERSION_LESS 7.4)
+    message (FATAL_ERROR "This project requires GCC 7.4 or later")
```
seelabs: I think we can remove this check. The ICE with gcc 7.3 appears  to be a jenkins only issue.

.travis.yml
mellery451: I agree - it's on my TODO list

Builds/containers/shared/install_cmake.sh
mellery451: `cmake_minimum_required` says 3.9 currently. I would personally prefer 3.12 or higher, but I agree we should have a test for whatever that version is. I'll add another TODO item for that.

CMakeLists.txt
mellery451: I'm fine keeping it at 5, but I think we need the travis builds to cover it then...otherwise it's untested. The 5 and 6 builds work just fine in travis, so if we want 5 as min ver, then we add the builds back in...I think you agree?

PR #3059, opened 2019-08-21
Add shard thread safety

miguelportilla:
This PR is part of a larger goal in creating self-contained portable shards and replacing the node store with the shard store.

Do not review the "Implement Shard SQLite support" commit.


src/ripple/nodestore/impl/Shard.cpp
```diff
@@ -55,9 +55,8 @@ Shard::Shard(
 bool
 Shard::open(Scheduler& scheduler, nudb::context& ctx)
 {
+    std::lock_guard<std::mutex> lock(mutex_);
```
HowardHinnant: We no longer need the `<std::mutex>` in this declaration.  Indeed, @seelabs is  just finishing up a PR that removes all of them.
```
std::lock_guard lock{mutex_};
```

miguelportilla: I was going to wait for its merge before opening up the flood gates.

HowardHinnant: I wouldn't remove any `<std::mutex>`.  I just think the thing to do now is not add any more.

HowardHinnant:
This looks pretty good to me.  But I'm left unsure exactly what subset of `Shard` that `mutex_` is protecting.  Which functions are multi-thread safe and which aren't?  If I read the code I can figure it out.  But some comments in Shard.h would help confirm that your intent matches the code.

HowardHinnant:
I'm still seeing read accesses to `complete_` and `storedSeqs_`.  Also to `j_`, but I haven't checked if `Journal` itself is thread safe.

miguelportilla:
@HowardHinnant Feedback addressed, thank you. The `Journal` severity stream access functions (eg. debug(), error() etc...) are thread safe using our current `Logs::Sink` [implementation](https://github.com/ripple/rippled/blob/develop/src/ripple/basics/impl/Log.cpp#L177
).

As a side note... I believe the Journal design could be greatly improved. As is, thread safety isn't guaranteed. For instance, the `Journal:: Sink` `thresh_` and `m_console` members are mutable. Additionally, concurrent writes to file under a user lock will block, adversely affecting performance. :(

HowardHinnant:
How about this:  https://github.com/HowardHinnant/rippled/commit/7723e82db1c00308caf59362bdb2e0e8ad0b98e3

The diff looks much worse than it is.  Most of it is just un-indenting the scope for the `unique_lock`.  Once you look past that, the only thing I've done is transfer the locked ownership of `m_` from the `unique_lock` to the `lock_guard`, without unlocking `m_`.

This is both more efficient, and eliminates the possibility of another thread grabbing `m_` while it is unlocked within this narrow window in `validate()`.

miguelportilla:
@HowardHinnanrt Cherry-picked your mutex suggestion, thanks!

src/ripple/nodestore/impl/Shard.h
```diff
@@ -88,44 +90,50 @@ class Shard
...
-    std::shared_ptr<Backend> const&
-    getBackend() const {assert(backend_); return backend_;}
+    fdRequired() const;
 
     std::shared_ptr<Ledger const>
-    lastStored() {assert(backend_); return lastStored_;}
+    lastStored() const;
+
+    bool
+    validate();
```
scottschurr: Just a question... What is the non-`const` thing that `validate()` does?  If all it does is examine the `Shard` and return `true` or `false` about its validity, then it could be `const`.  I suspect that's not all that happens.  At any rate, it seems like `validate()` should either be `const` or have a comment explaining why it is not.

I think, maybe, you can declare `validate() const` if you also declare `valLedger(ledger, next) const` and `valFetch(hash) const`?

src/ripple/nodestore/impl/Shard.h
```diff
@@ -179,43 +187,52 @@ class Shard
...
     // Save the control file for an incomplete shard
+    // Lock over mutex_ required
     bool
-    saveControl();
+    saveControl(std::lock_guard<std::mutex> const& lock);
+
+    // Validate this ledger by walking its SHAMaps
+    // and verifying each merkle tree
+    bool
+    valLedger(
```
scottschurr: Could this be `valLedger(ledger, next) const`?

src/ripple/nodestore/impl/Shard.h
```diff
@@ -179,43 +187,52 @@ class Shard
...
+        std::shared_ptr<Ledger const> const& ledger,
+        std::shared_ptr<Ledger const> const& next);
+
+    // Fetches from the backend and will log
+    // errors based on status codes
+    std::shared_ptr<NodeObject>
```
scottschurr: Could this be `valFetch(hash) const`?

src/ripple/nodestore/impl/Shard.h
```diff
@@ -88,44 +90,50 @@ class Shard
...
+    validate();
 
 private:
     friend class boost::serialization::access;
-    template<class Archive>
-    void serialize(Archive & ar, const unsigned int version)
+
+    template <class Archive>
+    void
+    serialize(Archive & ar, const unsigned int version)
```
scottschurr: As `serialize(ar, version)` is currently implemented I think it's not thread safe.  I suspect it needs a `lock`.  But that's okay, it also looks like it is unused.  So my suggestion would be to delete the private method.

src/ripple/nodestore/impl/DatabaseShardImp.cpp
```diff
@@ -622,36 +622,36 @@ DatabaseShardImp::getCompleteShards()
 void
 DatabaseShardImp::validate()
 {
+    std::map<std::uint32_t, std::shared_ptr<Shard>> completeCpy;
```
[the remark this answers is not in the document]
miguelportilla: Agreed, will do! 👍

src/ripple/nodestore/impl/DatabaseShardImp.cpp
```diff
@@ -622,36 +622,36 @@ DatabaseShardImp::getCompleteShards()
...
-        if (incomplete_)
-            s += std::to_string(incomplete_->index());
-        else
-            s.pop_back();
-        JLOG(j_.debug()) << s;
+        completeCpy = complete_;
     }
 
-    for (auto& e : complete_)
+    for (auto const &[_, shard] : completeCpy)
```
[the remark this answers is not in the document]
miguelportilla: I think you'll like the new solution even better.

src/ripple/nodestore/impl/Shard.cpp
```diff
@@ -205,31 +205,100 @@ Shard::prepare()
...
+
 bool
 Shard::validate()
 {
     uint256 hash;
-    std::uint32_t seq;
+    std::uint32_t seq {0};
     std::shared_ptr<Ledger> ledger;
-    auto fail = [this](std::string const& msg)
+    auto fail = [this, &hash, seq](std::string const& msg)
```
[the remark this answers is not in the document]
miguelportilla: Will do.  Yes, I meant to capture 'seq' by ref. Good find!

src/ripple/nodestore/impl/Shard.cpp
```diff
@@ -790,5 +720,120 @@ Shard::saveControl(std::lock_guard<std::mutex> const&)
     return true;
 }
 
+bool
+Shard::valLedger(
+    std::shared_ptr<Ledger const> const& ledger,
+    std::shared_ptr<Ledger const> const& next)
+{
+    auto fail = [this, &ledger](std::string const& msg)
```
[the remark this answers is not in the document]
miguelportilla: Of course my OCD compels me to apply similar captures throughout the file.

src/ripple/nodestore/impl/Shard.h
miguelportilla: They can all be const.

src/ripple/nodestore/impl/Shard.h
miguelportilla: yes

src/ripple/nodestore/impl/Shard.h
miguelportilla: yes

src/ripple/nodestore/impl/Shard.h
miguelportilla: Its a remnant and can be removed. Good find.

miguelportilla:
@scottschurr Feedback addressed, thank you! Please let me know if I missed anything.

PR #3067, opened 2019-08-29
Make XRPAmount constructor explicit

seelabs:
Only the commits starting with `Make XRPAmount constructor explicit:` need to be reviewed (it is based on @ximinez PR https://github.com/ripple/rippled/pull/2910/, since that moved the `XRPAmount.h` file and made some changes).

Remove the implicit conversion from int64 to XRPAmount. The motivation for this was noticing that many calls to `to_string` with an integer parameter type were calling the wrong `to_string` function. Since the calls were not prefixed with `std::`, and there is no ADL to call `std::to_string`, this was converting the int to an `XRPAmount` and calling `to_string(XRPAmount)`.

Since `to_string(XRPAmount)` did the same thing as `to_string(int)` this error went undetected.

 @HowardHinnant @ximinez


HowardHinnant:
I like this change a lot!

These mods were necessary to compile on macOS:  https://github.com/HowardHinnant/rippled/commit/cbac51fcbea0103e0a52439b9b20ba46c7a0828d

seelabs:
Thanks @HowardHinnant! Cherry picked and pushed.

HowardHinnant:
I'm getting 4 failures in unit tests:
```
ripple.ledger.Invariants checks enabled - no zero escrow
#50 failed: Invariants_test.cpp(122)
#51 failed: Invariants_test.cpp(122)
#54 failed: Invariants_test.cpp(122)
#55 failed: Invariants_test.cpp(122)
ripple.ledger.PaymentSandbox selfFunding
```

seelabs:
@HowardHinnant Fixed the invariant check. However given the culprit (by making the `XRPAmount` constructor explicit, code that used to call that constructor now calls the non-explicit `STAmount` constructor in some cases). I need to think through the consequences of that and address them.

At any rate, the Invariant tests now pass.

HowardHinnant:
Yeah, I'm alarmed that this wasn't caught at compile-time, and that the generated code returned the wrong answer.

seelabs:
@HowardHinnant the behavior differs in how the two classes handle out of range values. `STAmount` throws and XRPAmount does not. That is what caused the test to fail. While I do need address this (I'll try to make sure the old code use the same class it originally used), it's not _quite_ as bad as giving wrong answers. Both `STAmount` and `XRPAmount` are going to do the same thing in most cases.

seelabs:
Pushed a commit to fix old implicit conversions for XRPAmount and STAmount

* Make STAmount std::uint64_t constructor explicit:
* Add user defined suffix for XRPAmount `_xrpa`
* Add user defined suffix for STAmount `_stamt`
* Resolve literals that used to call these non-explicit constructors

HowardHinnant:
Still 👍

nbougalis:
Ugh, I'm not sure how much I like these UDLs... :slightly_frowning_face:

seelabs:
@nbougalis I'm fine killing the the UDL's. I don't love them either. Will do.

seelabs:
Pushed a commit to remove the user defined literals for XRPAmount and STAmount.

seelabs:
Rebased onto @ximinez newly rebased branch (which was rebased onto 1.4.0-rc1). Also squashed commits fold commits.

src/ripple/protocol/SystemParameters.h
```diff
@@ -42,7 +42,7 @@ systemName ()
 /** Number of drops in the genesis account. */
 constexpr
 XRPAmount
-INITIAL_XRP{ 100'000'000'000 * DROPS_PER_XRP };
+INITIAL_XRP{ static_cast<std::int64_t>(100'000'000'000) * DROPS_PER_XRP };
```
ximinez: Is this cast necessary? If it is, could you change it to a `safe_cast`?

seelabs: I don't think it's needed. Fixed in `c2ef2bb5d [fold] Remove unneeded static_cast`

ximinez:
This branch is rebased onto 1.5.0-b1 that resolves conflicts and fixes build errors with in calls to `to_string` that incorrectly use the implicit constructor introduced by new code. https://github.com/ximinez/rippled/commits/sd--xrpamount-explicit

seelabs:
@ximinez I'm marking this as passed (tho I did push a slightly different fix to the rebase issues; If you do prefer the explicit `std::` prefix I'm fine with that too).

PR #3069, opened 2019-09-01
Automatically update NodeDB path if changed

yusufsahinhamza:
Fixes #3015

### DIFF
--- a/src/ripple/app/misc/SHAMapStoreImp.cpp
+++ b/src/ripple/app/misc/SHAMapStoreImp.cpp
@@ -488,6 +488,30 @@ SHAMapStoreImp::dbPaths()
     }
 
     SavedState state = state_db_.getState();
+
+    {
+        auto update = [&dbPath](std::string& sPath)
+        {
+            if (sPath.empty())
+                return false;
+
+            // Check if configured "path" matches stored directory path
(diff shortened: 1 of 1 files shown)


nbougalis:
Thanks for your contribution! I’m going to ask @miguelportilla, @mtrippled and @jwbusch to take a look at this and sign off

miguelportilla:
@yusufsahinhamza Considering narrowing the scope of work since it is only used in once place. Please take a look at [cabe1c41](https://github.com/miguelportilla/rippled/commit/cabe1c41743c9f630240ec7ccb48c66656ec5c8b)

yusufsahinhamza:
@miguelportilla Yes, it's absolutely looks better now!

src/ripple/app/misc/SHAMapStoreImp.cpp
```diff
@@ -554,6 +565,17 @@ SHAMapStoreImp::makeBackendRotating (std::string path)
     return backend;
 }
 
+std::string
+SHAMapStoreImp::updateDbPath(std::string path)
```
mtrippled: the path parameter should be const.

mtrippled:
:+1:

PR #3072, opened 2019-09-05
Propagate validator lists (VLs or UNLs) over the peer network:

### LINKED ISSUE #2953: Allow validator list propagation over the peer network
Any time a Server downloads or receives a validator list, it should broadcast a message reporting the contents of the validator list to all of its immediately connected peers, except those which have recently sent the same list to the Server.

The message should NOT be flooded throughout the peer network; just to a server's immediate peers. Optionally, servers could track (on an ephemeral per-connection basis) which sequence number a particular peer has and avoid sending the notification to servers known to have that list already. 

Additionally, the Server should save the updated validator list into its local data directory with a file name like: `unl.<publickey>`. On startup, the Server should read that file, validate it, and use it as its starting validator list.

Finally, the Server should send all of its trusted validator lists to new peers when they first connect.

Replaces Jira issue RIPD-1673.

ximinez:
* Whenever a node downloads a new VL, send it to all peers that
  haven't already sent or received it. It also saves it to the
  database_dir as a Json text file named "cache." plus the public key of
  the list signer. Any files that exist for public keys provided in
  [validator_list_keys] will be loaded and processed if any download
  from [validator_list_sites] fails or no [validator_list_sites] are
  configured.
* Whenever a node receives a broadcast VL message, it treats it as if
  it had downloaded it on it's own, broadcasting to other peers as
  described above.
* Because nodes normally download the VL once every 5 minutes, a single
  node downloading a VL with an updated sequence number could
  potentially propagate across a large part of a well-connected network
  before any other nodes attempt to download, decreasing the amount of
  time that different parts of the network are using different VLs.
* Broadcast all of our current valid VLs to new peers on connection.
  This is probably the "noisiest" part of this change, but will give
  poorly connected or poorly networked nodes the best chance of syncing
  quickly. Nodes which have no http(s) access configured or available
  can get a VL with no extra effort.
* Requests on the peer port to the /vl/<pubkey> endpoint will return
  that VL in the same JSON format as is used to download now, IF the
  node trusts and has a valid instance of that VL.
* Resolves #2953


MarkusTeufelberger:
How hard/easy is it to spam the network with a few GB/TB of validly signed VLs?

nbougalis:
>How hard/easy is it to spam the network with a few GB/TB of validly signed VLs?

This is a great question that underlies an important concern. The answer is "_**Not very**_".

First, a server will only ever relay lists from publishers that they trust, meaning that someone can't just generate a million keypairs and use them to sign and "publish" a million lists.

Second, the server will only ever relay the *latest* list from a trusted publisher. So, if someone just sends old lists with a sequence number earlier than what the server considers to be "current" they won't be relayed. 

I believe that these two safeguards make this feature unsuitable for spamming.

MarkusTeufelberger:
A widely trusted VL publisher can attack the network that way though, right? Just by publishing a huge number of lists instead of actively malicioulsly publishing a list that grants them full control over the network.

There might be legitimate use cases to publish a new list every ledger close for an automated VL provider (e.g. to know that the provider is still active and that it also is actively moitoring the network by including the latest ledger hash as extra data in the list). Also these lists can grow relatively large.
Probably still not enough to cause issues or to overwhelm a network link though.

Just to be clear: As I understand it, a server can receive lots of VLs from different issuers in lots of versions, but will only relay a single one (the most recent VL of the issuer it trusts) to its peers?

This leaks private information (which VL is trusted by a server) and makes it very hard to even have a second VL provider in the same network using this mechanism, since they can only get propagated via direct connections instead of hopping over peers. I also have concerns around leaking private in-cluster VLs (if one operates several servers, but wants to retain control over their UNL, the best way currently is to set up a private VL for that cluster and have the cluster members trust it). Just turning off the endpoint doesn't seem to be enough to stop this information (if a VL is trusted and if yes: which exact version by which provider) from leaking right after connecting to a server.

ximinez:
> > How hard/easy is it to spam the network with a few GB/TB of validly signed VLs?
> [...]
> I believe that these two safeguards make this functionality unsuitable for spamming.

There are additional protections at the individual node level. The big one is that `PeerImp::onMessage (std::shared_ptr <protocol::TMValidatorList> const& m)` will charge the peer a higher fee if the processing result is anything other than `accepted` (which means it's new, valid, and trusted). A normal peer won't have any problem with this extra fee, but one that is behaving badly will get booted pretty quickly. Each node also tracks what it's already sent to its peers so that it should never send the same VL more than once to the same peer.

ximinez:
> A widely trusted VL publisher can attack the network that way though, right? Just by publishing a huge number of lists instead of actively malicioulsly publishing a list that grants them full control over the network.

Are you talking about a scenario where the publisher creates a valid list, pushes it out through its node, then immediately creates another list, over and over?  I believe that would quickly lead that publisher losing trust. If enough nodes stop trusting the publisher the damage may be contained, since those nodes will not relay. I also think the standard peer fee may kick the "source" node out eventually if the flooding is fast enough. I can see it being worthwhile to increase the peer fee for "success" to mitigate this scenario, too.

Finally, as you mention yourself, there are a lot more effective ways for a trusted VL publisher to attack and damage the network.

That said, I could see value in some kind of outgoing rate limit for well-behaved peers so that they won't relay for a publisher if they've relayed too recently.

> Just to be clear: As I understand it, a server can receive lots of VLs from different issuers in lots of versions, but will only relay a single one (the most recent VL of the issuer it trusts) to its peers?

Mostly correct. If a server trusts more than one VL, it will relay the most recent VL of all the issuers it trusts. But that's splitting hairs.

> This leaks private information (which VL is trusted by a server) and makes it very hard to even have a second VL provider in the same network using this mechanism, since they can only get propagated via direct connections instead of hopping over peers. I also have concerns around leaking private in-cluster VLs (if one operates several servers, but wants to retain control over their UNL, the best way currently is to set up a private VL for that cluster and have the cluster members trust it). Just turning off the endpoint doesn't seem to be enough to stop this information (if a VL is trusted and if yes: which exact version by which provider) from leaking right after connecting to a server.

I could see value in a configuration option to disable sending VLs to peers if someone is concerned about their trust map or a private VL. I suspect a small minority of servers will use it, but even if it is popular, I bet having public hubs and other well-connect nodes doing the broadcast may be sufficient to keep the network healthy.

MarkusTeufelberger:
> Are you talking about a scenario where the publisher creates a valid list, pushes it out through its node, then immediately creates another list, over and over? I believe that would quickly lead that publisher losing trust.

I already described a valid (and maybe useful?) use case: Publish a VL for every closed ledger.

Anyways, all in all it only would allow someone to shoot themselves into the foot, but others would just drop connections after a while if that node tries to send a few gigabytes of VLs upon connection and not propagate them.

This still leaves my concern about VL centralization - how does this scheme act in the presence of several widely trusted VL publishers? After all, validations themselves are always forwarded (up to traffic limits etc.) - maybe recent VLs that have x% overlap (80%? 90%?) with a server's UNL also should be forwarded, even if they are not trusted?

nbougalis:
>Just to be clear: As I understand it, a server can receive lots of VLs from different issuers in lots of versions, but will only relay a single one (the most recent VL of the issuer it trusts) to its peers?

Correct: a server can receive lots of VLs from different issuers and multiple VLs from the same issuer but it will only ever forward the "most recent" VL from publishers that it trusts.

>This leaks private information (which VL is trusted by a server)

To be honest, I don't know that I consider that "private information" though, of course, your mileage may vary and that's just fine.

>I already described a valid (and maybe useful?) use case: Publish a VL for every closed ledger.

I find that scenario "broken" personally, but that's another topic entirely. A UNL publisher can already sorta-kinda do that: publish a list that expires 5 seconds into the future, thus forcing anyone that's trusting them to keep connecting to them to retrieve the list. 

>This still leaves my concern about VL centralization - how does this scheme act in the presence of several widely trusted VL publishers? 

It makes no real difference, as far as I can tell. It can cause a slight, bounded increase in the bandwidth used by the overlay, but it offers a _critical_ service to other servers in return and removes a potential point of centralization (a web server).

If the bandwidth concerns are significant, we could always flip the model on its head by implementing a "request" model ("Hey, I'm looking for the UNL signed by `ED2677ABFFD1B33AC6FBC3062B71F1E8397C1505E1C42C64D11AD1B28FF73F4734`. Do you have it?") which can be rate limited.

In that case, servers could simply opt to cache and serve UNLs from publishers they don't trust as well.

>After all, validations themselves are always forwarded (up to traffic limits etc.) - maybe recent VLs that have x% overlap (80%? 90%?) with a server's UNL also should be forwarded, even if they are not trusted?

A server only forwards validations from _trusted_ validators; validations from untrusted validators aren't regularly forwarded. We can have a discussion on if and how untrusted validations should be forwarded but this PR isn't the right place for that.

MarkusTeufelberger:
I think the request model might be better (in all cases except the first time this data is not needed anyways when connecting to a server), or a hybrid approach where servers only send the VL(s) + sequence number(s) they currently would offer instead of the full data all the time so it is easier to choose from where to request the latest one.

By the way, since this changes the p2p protocol by adding a message type, should this change to `RTXP/1.3` or `XRPL/2.1`?

ximinez:
> I already described a valid (and maybe useful?) use case: Publish a VL for every closed ledger.

I'm not sure that's a particularly _useful_ use case, but even if someone did want to implement this, publishing a new VL every 3-4 seconds is not as bad as what I thought you were talking about, which is a scenario of many VLs per second. Even so, for that to be a problem, the publisher would have to be pretty widely trusted - else the VLs aren't going to propagate very far - and the way I see it, if such a publisher _is_ widely trusted, then that's a significant chunk of the network saying that such behavior is desirable.

ximinez:
> I think the request model might be better (in all cases except the first time this data is not needed anyways when connecting to a server), or a hybrid approach where servers only send the VL(s) + sequence number(s) they currently would offer instead of the full data all the time so it is easier to choose from where to request the latest one.

One advantage of the push model is that VLs can potentially propagate as soon as they're updated and the first node grabs a copy.  Also, the `/vl/` endpoint implementation gives an out-of-protocol request model option for poorly connected nodes or publishers. It's not perfect, since you'd have to ask a specific node, obviously, but it's an option...

> By the way, since this changes the p2p protocol by adding a message type, should this change to `RTXP/1.3` or `XRPL/2.1`?

I don't _think_ so, since the behavior is backward compatible in the sense that any node receiving a VL that doesn't know about this message will just ignore it.

MarkusTeufelberger:
> One advantage of the push model is that VLs can potentially propagate as soon as they're updated and the first node grabs a copy.

This could also happen in the pull scenario by notifying peers about a new VL (unless VL versions are communicated only in the handshake) once a new one is detected.

> the behavior is backward compatible in the sense that any node receiving a VL that doesn't know about this message will just ignore it.

If it runs `rippled`, not so sure about re-implementations. Could there be even more message types added with the same protocol version? How does one distinguish between nodes that advertise the same protocol version, but can understand different message types?

PS: I'd really love to have an option to store VL versions semi-permanently, and as I already wrote in the Issue for this PR: Why are they not added to a database but dumped as JSON? If they got stored in e.g. `wallet.db`, they could be indexed by pubkey and sequence number...

nbougalis:
>PS: I'd really love to have an option to store VL versions semi-permanently, and as I already wrote in the Issue for this PR: Why are they not added to a database but dumped as JSON? If they got stored in e.g. wallet.db, they could be indexed by pubkey and sequence number...

I think that's a neat option for anyone interested in maintaining a "historical UNL" archive, but I'm not sure that there's much of a benefit in coding this functionality inside `rippled`. I think that such an archiving tool is pretty trivial to script.

What do you believe are the advantages of having this functionality _inside_ `rippled`?

nbougalis:
>>By the way, since this changes the p2p protocol by adding a message type, should this change to RTXP/1.3 or XRPL/2.1?
>I don't think so, since the behavior is backward compatible in the sense that any node receiving a VL that doesn't know about this message will just ignore it.

While you're right that a peer that receives this message will ignore it, I think that it's a good habit for us to get into the "bump up" the wire protocol version when changes are made to the wire protocol.

I don't think any changes are needed _now_, but if this goes in _after_ #3049, then we maybe want to gate this as an `XRPL/2.0` feature.

src/ripple/app/main/Application.cpp
```diff
@@ -1448,10 +1457,43 @@ bool ApplicationImp::setup()
...
     //             if (!config_.standalone())
-    m_overlay = make_Overlay (*this, setup_Overlay(*config_), *m_jobQueue,
+    overlay_ = make_Overlay (*this, setup_Overlay(*config_), *m_jobQueue,
         *serverHandler_, *m_resourceManager, *m_resolver, get_io_service(),
         *config_);
-    add (*m_overlay); // add to PropertyStream
+    add (*overlay_); // add to PropertyStream
+
+    validators_->setMessageCallback(
+        [&](uint256 const& hash,
```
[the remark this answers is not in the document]
ximinez: Rewritten to use local variables, and to get rid of the callbacks entirely.

src/ripple/app/main/Application.cpp
```diff
@@ -1448,10 +1457,43 @@ bool ApplicationImp::setup()
...
+        assert(overlay_);
+        assert(hashRouter_);
+        auto const toSkip = hashRouter_->shouldRelay(hash);
+
+        if (toSkip)
+        {
+            overlay_->foreach_mutable(send_and_modify_if_not(
+                std::make_shared<Message>(msg, protocol::mtVALIDATORLIST),
+                peer_in_set_or_condition(*toSkip,
+                    [&](std::shared_ptr<Peer> const& peer) {
```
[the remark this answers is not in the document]
ximinez: Done.

src/ripple/app/misc/ValidatorList.h
```diff
@@ -123,11 +129,27 @@ class ValidatorList
...
     TimeKeeper& timeKeeper_;
+    // Typically, the Overlay and ValidatorSite objects haven't been created
+    // when this object is created, so references to them can't be provided
+    // to the ctor. Instead, the creator will provide these callback functions
+    // to do the appropriate work when necessary.
+    std::function<void(uint256 const& hash,
+        protocol::TMValidatorList const& msg,
+        PublicKey const& publisherKey,
+        size_t sequence)> sendMessage_;
+    std::function<void(std::vector<std::string> const& sites)>
```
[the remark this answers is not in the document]
ximinez: Inspired by this comment, I completely removed the callbacks.

src/ripple/app/misc/impl/ValidatorList.cpp
```diff
@@ -190,33 +196,88 @@ ValidatorList::load (
...
+ValidatorList::PublisherListStats
+ValidatorList::applyListAndBroadcast(
+    std::string const& manifest,
+    std::string const& blob,
+    std::string const& signature,
+    std::uint32_t version,
+    std::string siteUri,
+    uint256 const& hash)
+{
+    auto const result = applyList(manifest, blob, signature,
```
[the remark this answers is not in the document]
ximinez: Done.

src/ripple/app/misc/ValidatorList.h
```diff
@@ -147,16 +169,45 @@ class ValidatorList
...
     ~ValidatorList () = default;
 
+    /** Describes the result of processing a Validator List (UNL),
+    including some of the information from the list which can
+    be used by the caller to know which list publisher is
+    involved.
+    */
+    struct PublisherListStats
+    {
+        explicit PublisherListStats(ListDisposition d)
```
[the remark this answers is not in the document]
ximinez: Yeah, that makes sense. Done.

src/ripple/app/misc/impl/ValidatorList.cpp
```diff
@@ -309,7 +370,101 @@ ValidatorList::applyList (
...
+                JLOG(j_.error()) <<
+                    "Problem writing " <<
+                    filename <<
+                    " " <<
+                    ec.value() <<
+                    ": " <<
+                    ec.message();
+            }
+        }
+        catch (std::exception const& e)
```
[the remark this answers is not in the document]
ximinez: You're right. I simply forgot that `ofstream` doesn't throw by default.

src/ripple/app/misc/impl/ValidatorList.cpp
```diff
@@ -309,7 +370,101 @@ ValidatorList::applyList (
         }
     }
 
-    return ListDisposition::accepted;
+    // Cache the validator list in a file
+    if (!dataPath_.empty())
```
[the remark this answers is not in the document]
ximinez: Done.

src/ripple/app/misc/impl/ValidatorList.cpp
```diff
@@ -309,7 +370,101 @@ ValidatorList::applyList (
...
+        boost::system::error_code ec;
+
+        if (publisher.available)
+            continue;
+
+        boost::filesystem::path const filename{
+            dataPath_ / (filePrefix_ + strHex(pubKey)) };
+
+        auto const fullPath{ canonical(filename, ec) };
+        if (!ec)
```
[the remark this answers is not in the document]
ximinez: Yeah, make sense. Done.

src/ripple/app/misc/impl/ValidatorList.cpp
```diff
@@ -309,7 +370,101 @@ ValidatorList::applyList (
...
+                    return "file://";
+                }
+#endif
+            }();
+            sites.emplace_back(prefix + fullPath.string());
+        }
+    }
+
+    // Then let the ValidatorSites do the rest of the work.
+    if (!sites.empty() && loadValidatorSites_)
```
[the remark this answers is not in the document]
ximinez: Done.

src/ripple/app/main/Application.cpp
```diff
@@ -571,6 +573,12 @@ class ApplicationImp
         m_nodeStoreScheduler.setJobQueue (*m_jobQueue);
 
         add (m_ledgerMaster->getPropertySource ());
+
+        validators_->setLoadSitesCallback(
+            [&](std::vector<std::string> const& sites) {
```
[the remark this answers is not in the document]
ximinez: No longer relevant

src/ripple/app/main/Application.cpp
```diff
@@ -1448,10 +1457,43 @@ bool ApplicationImp::setup()
...
         *config_);
-    add (*m_overlay); // add to PropertyStream
+    add (*overlay_); // add to PropertyStream
+
+    validators_->setMessageCallback(
+        [&](uint256 const& hash,
+            protocol::TMValidatorList const& msg,
+            PublicKey const& publisherKey,
+            size_t sequence) {
+        assert(overlay_);
```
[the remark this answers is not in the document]
ximinez: No longer relevant

ximinez:
I'm making this PR a dependent of #3049 so I don't have to reinvent the wheel with the protocol change. As of this moment, all I've done is rebase on @nbougalis 's branch and fix conflicts. I'll be adding the `XRPL/2.1`, soon, and checking that peers support that version before sending UNLs.

wilsonianb:
fwiw here's an ancient discussion of forwarding lists: https://github.com/ripple/rippled/pull/1842#issuecomment-248759417
with bonus code changes: https://github.com/wilsonianb/rippled/commit/0da3b99b8940428fb6653cf511790ce634eaf975

ximinez:
@wilsonianb Thanks for the pointer. I read over the discussion and code changes and other than a little bit of "I could have saved a chunk of time if I had seen this sooner", I don't see any fundamental / breaking differences between the two implementations. If I missed anything, I'd be happy to look closer.

ximinez:
Squashed and rebased on to 1.5.0-b1 to resolve conflicts.

src/ripple/app/misc/impl/ValidatorList.cpp
```diff
@@ -191,33 +200,164 @@ ValidatorList::load (
...
+                    // Don't send it next time.
+                    hashRouter.addSuppressionPeer(
+                        hash, peer->id());
+                    peer->setPublisherListSequence(
+                        publisherKey, sequence);
+                };
+            overlay.foreach_mutable(send_and_modify_if_not(
+                std::make_shared<Message>(msg, protocol::mtVALIDATORLIST),
+                peer_in_set_or_condition(*toSkip, peerNoSendCondition),
+                peerupdate));
```
thejohnfreeman: I noticed there was some previous conversation that led you to remove some callbacks elsewhere. Can we remove these too? `foreach_mutable`, which you added, just calls `getActivePeers`, which is a public method. Then we can get rid of the non-lambda function objects that end up splitting the body of this loop across several files with lots of distracting boilerplate cluttered around the juicy bits.

src/ripple/overlay/predicates.h
```diff
@@ -106,6 +106,45 @@ send_if_not_pred<Predicate> send_if_not (
...
+
+/** Helper function to aid in type deduction */
+template <typename Predicate, typename Update>
+send_and_modify_if_not_pred<Predicate, Update> send_and_modify_if_not (
+    std::shared_ptr<Message> const& m,
+    Predicate const &f,
+    Update const& u)
+{
+    return send_and_modify_if_not_pred<Predicate, Update>(m, f, u);
+}
```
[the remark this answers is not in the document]
thejohnfreeman: (but please see my later comment about removing this file entirely)

src/ripple/app/misc/impl/ValidatorList.cpp
ximinez: Nice catch. I got caught up in the copypasta, and didn't even think to check the visibility of `getActivePeers`.

src/ripple/overlay/predicates.h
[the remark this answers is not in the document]
ximinez: Removing this file might be a good idea, but I think it's out of the scope of this PR. However, I was able to remove all of the stuff that I added in this file.

src/ripple/overlay/impl/PeerImp.h
```diff
@@ -342,6 +345,29 @@ class PeerImp
...
+    supportsProtocolVersion(const char* target)
+    const override;
+
+    boost::optional<std::size_t>
+    publisherListSequence(PublicKey const& pubKey) const override
+    {
+        std::lock_guard<std::mutex> sl(recentLock_);
+
+        if (publisherListSequences_.count(pubKey))
+            return publisherListSequences_.at(pubKey);
```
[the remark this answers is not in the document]
ximinez: Fixed

src/ripple/overlay/predicates.h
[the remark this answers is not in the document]
thejohnfreeman: Oops, that's what I meant. I mistakenly thought this file was entirely new.

src/ripple/app/misc/impl/ValidatorList.cpp
```diff
@@ -191,33 +199,159 @@ ValidatorList::load (
...
+{
+    return dataPath_ / (filePrefix_ + strHex(pubKey));
+}
+
+void
+ValidatorList::CacheValidatorFile(PublicKey const& pubKey,
+    PublisherList const& publisher)
+{
+    if (!dataPath_.empty())
+    {
```
thejohnfreeman: I'm a fan of early return here, to reduce nesting for the rest of the block and let the reader know there is no (non-trivial) else-branch.

ximinez:
@thejohnfreeman I made the trivial change to do the early return, but went ahead and marked the PR as passed anyway.

src/ripple/overlay/impl/TrafficCount.h
```diff
@@ -187,6 +188,7 @@ class TrafficCount
         { "overhead: manifest" },                                 // category::manifests
         { "transactions" },                                       // category::transaction
         { "proposals" },                                          // category::proposal
+        { "validator lists" },                                    // category::validatorlist
```
nbougalis: This particular change breaks #3179. It's worth considering whether that needs to be fixed _there_ or here (although considering that it's already included in the proposed **1.5.0-b2** PR #3199, the fix will likely have to come in a separate PR and be coordinated with this).

src/ripple/overlay/impl/PeerImp.cpp
```diff
@@ -803,6 +817,36 @@ PeerImp::doProtocolStart()
 {
     onReadMessage(error_code(), 0);
 
+    // Send all the validator lists that have been loaded
+    if (supportsProtocolVersion("XRPL/2.1") )
```
nbougalis: I don't quite like this; it requires manual parsing of a string for no real reason. It could simply boil down to:

```if (protocol_ >= make_protocol(2,1))```

File: src/ripple/overlay/impl/PeerImp.cpp [hunk 18]
```diff
@@ -1965,6 +2009,134 @@ PeerImp::onMessage (std::shared_ptr <protocol::TMHaveTransactionSet> const& m)
...
+            // Charging this fee here won't hurt the peer in the normal
+            // course of operation (ie. refresh every 5 minutes), but
+            // will add up if the peer is misbehaving.
+            fee_ = Resource::feeUnwantedData;
+            break;
+        case ListDisposition::invalid:
+            JLOG (p_journal_.warn()) <<
+                "Invalid validator list from peer " << remote_address_;
+            // This shouldn't ever happen with a well-behaved peer
+            fee_ = Resource::feeInvalidSignature;
```
nbougalis: I think that this warrants disconnecting the peer on the spot. Rationale: the peer has _already_ sent us invalid data; what's the point of giving it a chance to send us more invalid data?

src/ripple/overlay/impl/PeerImp.cpp
```diff
@@ -1965,6 +2009,134 @@ PeerImp::onMessage (std::shared_ptr <protocol::TMHaveTransactionSet> const& m)
...
+        case ListDisposition::unsupported_version:
+            JLOG (p_journal_.warn()) <<
+                "Unsupported version validator list from peer " <<
+                remote_address_;
+            // During a version transition, this may be legitimate.
+            // If it happens frequently, that's probably bad.
+            fee_ = Resource::feeBadData;
+            break;
+        default:
+            BOOST_ASSERT(false);
```
nbougalis: Is there a point in using `BOOST_ASSERT` instead of good ole' `std::assert`?

src/ripple/overlay/impl/PeerImp.cpp
```diff
@@ -1965,6 +2009,134 @@ PeerImp::onMessage (std::shared_ptr <protocol::TMHaveTransactionSet> const& m)
...
+        switch (disp)
+        {
+        case ListDisposition::accepted:
+            JLOG (p_journal_.debug()) <<
+                "Applied new validator list from peer " << remote_address_;
+            {
+                // if the remote node uses any of the same lists
+                // as this node, treat it as sane.
+                auto expected = Sanity::unknown;
+                sanity_.compare_exchange_strong(expected, Sanity::sane);
```
nbougalis: I'm not sure I agree that we should treat such peers as sane. Sanity implies _more_.

src/ripple/overlay/impl/PeerImp.cpp
```diff
@@ -405,6 +405,20 @@ PeerImp::json()
...
+PeerImp::supportsProtocolVersion(const char* target) const
+{
+    auto const goal = parseProtocolVersions(target);
+    // The values are passed in as hard-coded strings. Make sure that
+    // string represents a single, valid supported version.
+    assert(goal.size() == 1 &&
+        negotiateProtocolVersion(goal));
+
+    return goal.size() == 1 &&
+        protocol_ >= goal.front();
```
nbougalis: I'm not quite sure I like this.

First, just because feature X is supported with `XRPL/2.1` doesn't mean it's supported on version `XRPL/2.2` or later. Second, the API is messy. It requires me (as the prorammer) to keep mental track of which features are supported with each version. This can and will quickly become unwieldy.

I do believe this functionality is good to have, but I propose a different way of accessing it:

```
enum class ProtocolFeature
{
    ValidatorListPropagation = 1,
    SomethingElse = 2,
};

bool PeerImp::supportsFeature(ProtocolFeature f)
{
    switch (f)
    {
    case ProtocolFeature::ValidatorListPropagation: 
        return protocol_ >= make_protocol(2, 1);
    case ProtocolFeature::SomethingElse:
        return protocol >= make_protocol(2, 0) && protocol < make_protocol(3,1);
    }
    return false;
}
```

If we felt like it (and I don't think we need to go the extra mile) with a little more work and the help of `constexpr` we could do a little better, and reduce this to an array index and/or map lookup.

src/ripple/overlay/impl/OverlayImpl.cpp
```diff
@@ -1036,6 +1036,64 @@ OverlayImpl::processRequest (http_request_type const& req,
...
 
+bool
+OverlayImpl::processValidatorList (http_request_type const& req,
+    Handoff& handoff)
+{
+    // If the target is in the form "/vl/<validator_list_public_key>",
+    // return the most recent validator list for that key.
+    // "unl" can be used as an alias for "vl", but this usage is discouraged
+    if ((!req.target().starts_with("/vl/") &&
+            !req.target().starts_with("/unl/")) ||
```
nbougalis: I wouldn't allow `/unl/` at all. No point to provide a moniker but "discourage" it's use (i.e. make it "deprecated") in the initial version of an API.

src/ripple/basics/StringUtilities.h
```diff
@@ -63,7 +64,59 @@ inline static std::string sqlEscape (Blob const& vecSrc)
 
 uint64_t uintFromHex (std::string const& strSrc);
 
-boost::optional<Blob> strUnHex (std::string const& strSrc);
+template<class iter_t>
```
nbougalis: Suggest: `template <class Iterator>` to follow our usual practice.

ximinez:
@nbougalis That's some good feedback. I removed the "Passed" label and will work on addressing them.

src/ripple/overlay/impl/TrafficCount.h
ximinez: I'm fine with waiting until b2 is pushed, and then resolving conflicts, but I'm looking at #3179, and I don't see how this breaks it. What am I missing?

src/ripple/overlay/impl/PeerImp.cpp
ximinez: Agreed. Not sure why I thought this was a good idea...

src/ripple/overlay/impl/PeerImp.cpp
ximinez: By that logic, shouldn't we drop the connection (almost?) everywhere we charge `feeInvalidSignature`? If that's the case, why not change that cost from 1000 to 3000?

src/ripple/overlay/impl/TrafficCount.h
nbougalis: It passes the metrics as one large array and _implicitly_ relies on ordering to know what's what.  This breaks because you've inserted something in the middle of the array.

With that said, it's #3179 that's, ultimately, the problem: it has an extremely fragile implementation and you just happened to the unlucky one to break it. So this needs to be fixed there.

src/ripple/overlay/impl/PeerImp.cpp
nbougalis: You're right that in some of those cases, dropping would be appropriate. It's nothing urgent, we can leave it as is now, and then decide how we want to tackle this (i.e. charge vs. drop) across the board in a subsequent PR.

src/ripple/overlay/impl/PeerImp.cpp
ximinez: Nope. Fixed.

src/ripple/overlay/impl/PeerImp.cpp
ximinez: That's fair.

src/ripple/overlay/impl/PeerImp.cpp
ximinez: I like this a lot better, too. I'll make the change.

src/ripple/overlay/impl/OverlayImpl.cpp
ximinez: Good point. Fixed.

src/ripple/overlay/impl/TrafficCount.h
ximinez: It looks like #3179 got pulled from b2, so I guess this isn't as pressing. It would be easy enough to move the new item to the end of the list, but is it necessary, or should I just leave it be and let #3179 get fixed?

src/ripple/overlay/impl/PeerImp.cpp
```diff
@@ -2014,7 +2011,7 @@ PeerImp::onMessage (std::shared_ptr <protocol::TMValidatorList> const& m)
 {
     try
     {
-        assert(supportsProtocolVersion("XRPL/2.1"));
+        assert(supportsFeature(ProtocolFeature::ValidatorListPropagation));
```
nbougalis: This is a fine check to have, but I'd turn it into this:

    if (!supportsFeature(ProtocolFeature::ValidatorListPropagation))
        return;

Rationale: as is, the `assert` would allow a remote server to crash a server running this changeset with asserts enabled just by sending an unexpected message.

src/ripple/overlay/impl/ProtocolVersion.h
```diff
@@ -37,6 +37,7 @@ namespace ripple {
 using ProtocolVersion = std::pair<std::uint16_t, std::uint16_t>;
 
 inline
+constexpr
```
nbougalis: :tada:

src/ripple/overlay/impl/PeerImp.cpp
ximinez: Good catch. Fixed, including charging a fee for the unwanted data.

ximinez:
Rebased on to 1.5.0-b3 and resolved conflicts.

ximinez:
Squashed and rebased on to 1.5.0-b4.

PR #3073, opened 2019-09-05
Cleanup usage of ScopedUnlock

mellery451:
I originally intended to replace GenericScopedUnlock with boost::reverse_lock, but scaled back my ambition to simply cleanup the usage (e.g. unlocking a lock instead of the mutex directly).


src/ripple/app/ledger/impl/LedgerMaster.cpp
```diff
@@ -1093,7 +1093,7 @@ LedgerMaster::findNewLedgersToPublish ()
     auto valLedger = mValidLedger.get ();
     std::uint32_t valSeq = valLedger->info().seq;
 
-    ScopedUnlockType sul(m_mutex);
+    ScopedUnlockType sul{sl};
```
scottschurr: Just use `ScopedUnlock` and class template argument deduction will do the rest.

src/ripple/app/ledger/impl/LedgerMaster.cpp
```diff
@@ -1617,9 +1617,10 @@ void
 LedgerMaster::fetchForHistory(
     std::uint32_t missing,
     bool& progress,
-    InboundLedger::Reason reason)
+    InboundLedger::Reason reason,
+    std::unique_lock<std::recursive_mutex>& sl)
 {
-    ScopedUnlockType sl(m_mutex);
+    ScopedUnlockType sul{sl};
```
scottschurr: Just use `ScopedUnlock` and class template argument deduction will do the rest.

src/ripple/app/ledger/impl/LedgerMaster.cpp
```diff
@@ -1809,7 +1810,7 @@ void LedgerMaster::doAdvance (std::lock_guard<std::recursive_mutex>& sl)
             for(auto ledger : pubLedgers)
             {
                 {
-                    ScopedUnlockType sul (m_mutex);
+                    ScopedUnlockType sul{sl};
```
scottschurr: Just use `ScopedUnlock` and class template argument deduction will do the rest.

src/ripple/app/ledger/impl/LedgerMaster.cpp
```diff
@@ -1818,7 +1819,7 @@ void LedgerMaster::doAdvance (std::lock_guard<std::recursive_mutex>& sl)
                 setPubLedger(ledger);
 
                 {
-                    ScopedUnlockType sul(m_mutex);
+                    ScopedUnlockType sul{sl};
```
scottschurr: Just use `ScopedUnlock` and class template argument deduction will do the rest.

src/ripple/basics/ScopedLock.h
```diff
@@ -51,7 +54,7 @@ namespace ripple
         {
             ... do some stuff with it locked ..
 
-            GenericScopedUnlock<std::mutex> unlocker{mut};
+            ScopedUnlock<std::mutex> unlocker{myScopedLock};
```
scottschurr: Given class template argument deduction, you should probably remove the `<std::mutex>` from the example.

src/ripple/app/ledger/LedgerMaster.h
```diff
@@ -256,7 +256,7 @@ class LedgerMaster
     }
 
 private:
-    using ScopedUnlockType = GenericScopedUnlock <std::recursive_mutex>;
+    using ScopedUnlockType = ScopedUnlock<std::recursive_mutex>;
```
scottschurr: With class template argument deduction in C++17 you no longer need this declaration.  I suggest you remove it and use class template argument deduction to construct `ScopedUnlock` objects directly.

Assuming you do that, you should also remove the `#include ... ScopedLock.h>` from this file. Then you'll need to add the include into LedgerMaster.cpp.

src/ripple/basics/ScopedLock.h
```diff
@@ -61,40 +64,40 @@ namespace ripple
...
+        when the ScopedLock object is deleted, the unique_lock will
         be re-locked.
 
         Make sure this object is created and deleted by the same thread,
         otherwise there are no guarantees what will happen! Best just to use it
-        as a local stack object, rather than creating one with the new() operator.
+        as a local stack object, rather than creating on the heap.
     */
-    explicit GenericScopedUnlock (MutexType& lock) noexcept
+    explicit ScopedUnlock (std::unique_lock<MutexType>& lock)
```
scottschurr: In C++17 this constructor implicitly provides the class template argument deduction guide that you need elsewhere.

mellery451:
@scottschurr thanks for those suggestions - I've implemented them in a60d1b424cd6525b938ac7f577757d3599e43b57.

src/ripple/basics/ScopedLock.h
```diff
@@ -23,26 +23,29 @@
...
 #define RIPPLE_BASICS_SCOPEDLOCK_H_INCLUDED
+#include <mutex>
 
 namespace ripple
 {
 
 //==============================================================================
 /**
-    Automatically unlocks and re-locks a mutex object.
+    Automatically unlocks and re-locks a unique_lock object.
```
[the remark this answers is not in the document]
mellery451: went with the anonymous namespace and removed the header file. thanks.

src/ripple/app/ledger/impl/LedgerMaster.cpp
```diff
@@ -51,6 +51,85 @@
...
+    */
+    ~ScopedUnlock() noexcept(false)
+    {
+        lock_.lock();
+    }
+};
+
+}
+
 using namespace std::chrono_literals;
```
scottschurr: _\<Looking sheepish\>_ Not on your watch, but I just noticed this `using namespace` at file scope.  That's usually considered okay in a .cpp file.  I have a minor aversion to them in _our_ .cpp files because we do unity builds.  So a `using namespace` at file scope in one .cpp file can infect a different .cpp file.

At any rate, if you wanted to get rid of this `using namespace` I'd be in favor of that.  But don't feel like that's a requirement at all.  If you're curious, here's a commit you could cherry-pick to make the appropriate changes: https://github.com/scottschurr/rippled/commit/735b9c5af8a8b0711d2b16f1b340b5e9c3f62887.

Again, this is totally at your discretion.  I'm approving your pull request regardless.

mellery451: thanks for that extra cleanup - I have folded your changes in

PR #3081, opened 2019-09-18
Fix VS2019 debug build

jwbusch:
This is a quick fix to replace the `/ZI` flag with `/Zi`. 

VS2019 has it's own CMake included, which chose different defaults than the flags being used previously (This restriction existed in 2017). These new defaults are incompatible and results in debug builds failing.

https://docs.microsoft.com/en-us/cpp/build/reference/z7-zi-zi-debug-information-format?view=vs-2019

### DIFF
--- a/Builds/CMake/RippledCompiler.cmake
+++ b/Builds/CMake/RippledCompiler.cmake
@@ -28,11 +28,16 @@ target_compile_definitions (common
 if (MSVC)
   # remove existing exception flag since we set it to -EHa
   string (REGEX REPLACE "[-/]EH[a-z]+" "" CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
-  # also remove dynamic runtime
+
   foreach (var_
       CMAKE_C_FLAGS_DEBUG CMAKE_C_FLAGS_RELEASE
       CMAKE_CXX_FLAGS_DEBUG CMAKE_CXX_FLAGS_RELEASE)
+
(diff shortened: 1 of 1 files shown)


Builds/CMake/RippledCompiler.cmake
```diff
@@ -28,6 +28,11 @@ target_compile_definitions (common
...
   # remove existing exception flag since we set it to -EHa
   string (REGEX REPLACE "[-/]EH[a-z]+" "" CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS}")
+
+  # /ZI (Edit & Continue Debugging Information) is incompatible with Gy-
+  string (REPLACE "/ZI" "/Zi" CMAKE_CXX_FLAGS_DEBUG  ${CMAKE_CXX_FLAGS_DEBUG})
+  string (REPLACE "/ZI" "/Zi" CMAKE_C_FLAGS_DEBUG  ${CMAKE_C_FLAGS_DEBUG})
+
   # also remove dynamic runtime
   foreach (var_
       CMAKE_C_FLAGS_DEBUG CMAKE_C_FLAGS_RELEASE
```
jwbusch: Alternatively I could re-purpose this loop to do the same thing

mellery451: slightly simpler if you move your replace inside this loop (presumably it won't match for the release flags anyhow..), so I agree with your suggestion.

PR #3082, opened 2019-09-18
Fix startup error with `--import`

mellery451:


src/ripple/app/main/Application.cpp
```diff
@@ -931,20 +932,42 @@ class ApplicationImp
...
+             * read-only instance used for import, and it has zero read threads.
+             * If we were to construct it with one of the main app stoppables
+             * (like jobqueue), then this temporary DB instance is left as a
+             * dangling child in the stoppable tree (because there is no way to
+             * unregister an object as a child).
+             *
+             * SOLUTION: create a dummy stoppable tree/parent for this
+             * particular source DB instance because it has nothing that needs
+             * stopping anyhow.
+             **/
```
miguelportilla: I don't think the comment is necessary as `dummyRoot` is self explanatory but it is up to you.

miguelportilla: Consider renaming `schedule` to `dummySchedule`

mellery451: done (both) - thx

Issue #2597, opened 2018-06-22
tx command fails to return transactions that can't be submitted


mDuo13:
_(Original issue: [RIPD-1098](https://ripplelabs.atlassian.net/browse/RIPD-1098))_

The `tx` method cannot fetch pseudo-transactions, nor others transactions that aren't currently considered "valid" for submission, even if those transactions were incorporated into a validated ledger under rules that applied at the time. Doing so returns a misleading "not found" error. Example of a `tx` websocket command that will always fail (use a full-history server to confirm):

```json
{
  "command": "tx",
  "transaction": "5B1F1E8E791A9C243DD728680F108FEF1F28F21BA3B202B8F66E7833CA71D3C3"
}
```

The `transaction_entry` method _can_ fetch such transactions, provided you know what ledger contains them. Example of a `transaction_entry` websocket command that succeeds at fetching the same:

```json
{
  "id": 2,
  "command": "transaction_entry",
  "tx_hash": "5B1F1E8E791A9C243DD728680F108FEF1F28F21BA3B202B8F66E7833CA71D3C3",
  "ledger_index": 21225473
}
```

Per @JoelKatz:
> This occurs because the "tx" command asks the TransactionMaster to "fetch" the transaction, which calls Transaction::load. This function calls "checkValidity" to assure the returned transaction is valid. But this makes no sense. You should be able to retrieve transactions that are illegal under current rules. And pseudo-transactions are always illegal to submit, which is what checkValidity checks.
>
> For historical retrieval, the only check should be that we can deserialize the binary transaction data, and we should return an error other than "not found" in that case (and we can include the ledger sequence in the RPC, which would be helpful).

So, the changes to make would be:

- Remove the validity check from the `tx` command. Return it if we can deserialize it.
- If we can't deserialize the transaction, return a different error. I suggest a new error code, `txnDeserializingError`. Also, this should be logged at the warning level, because `rippled` _should_ be able to deserialize any transaction in valid history (even if that transaction is no longer valid under current rules).

cjcobb23:
@mDuo13 This has been resolved here: https://github.com/ripple/rippled/commit/11cf27e00698dbfc099b23463927d1dac829ed19
Can you close this issue?

mDuo13:
I'll test this out on a 1.5.0 beta server soon and close it when I confirm it works as expected.

carlhua:
[...]

PR #3084, opened 2019-09-25
FIXES: #2597

ShangyanLi:
Historical Transaction retrieval no longer needs to pass current validity check;
Introduced additional RPC error code for database deserialization error.

### DIFF
--- a/src/ripple/app/ledger/TransactionMaster.h
+++ b/src/ripple/app/ledger/TransactionMaster.h
@@ -20,6 +20,7 @@
 #ifndef RIPPLE_APP_LEDGER_TRANSACTIONMASTER_H_INCLUDED
 #define RIPPLE_APP_LEDGER_TRANSACTIONMASTER_H_INCLUDED
 
+#include <ripple/protocol/ErrorCodes.h>
 #include <ripple/shamap/SHAMapItem.h>
 #include <ripple/shamap/SHAMapTreeNode.h>
 
@@ -39,7 +40,10 @@ public:
     TransactionMaster& operator= (TransactionMaster const&) = delete;
 
     std::shared_ptr<Transaction>
-    fetch (uint256 const& , bool checkDisk);
+    fetch_from_cache (uint256 const&);
+
+    std::shared_ptr<Transaction>
+    fetch (uint256 const& , error_code_i& ec);
 
     std::shared_ptr<STTx const>
     fetch (std::shared_ptr<SHAMapItem> const& item,
--- a/src/ripple/app/misc/impl/Transaction.cpp
+++ b/src/ripple/app/misc/impl/Transaction.cpp
@@ -25,6 +25,7 @@
 #include <ripple/app/main/Application.h>
 #include <ripple/app/misc/HashRouter.h>
 #include <ripple/basics/safe_cast.h>
+#include <ripple/protocol/ErrorCodes.h>
 #include <ripple/protocol/Feature.h>
 #include <ripple/protocol/jss.h>
 #include <boost/optional.hpp>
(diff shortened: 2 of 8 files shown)


src/ripple/app/misc/impl/Transaction.cpp
```diff
@@ -134,13 +117,27 @@ Transaction::pointer Transaction::load(uint256 const& id, Application& app)
...
 
-    return Transaction::transactionFromSQLValidated (
-        ledgerSeq, status, rawTxn, app);
+    std::shared_ptr<Transaction> txn;
+    try
+    {
+        txn = Transaction::transactionFromSQL(
+                ledgerSeq, status, rawTxn, app);
+    }
+    catch (std::exception& e)
```
mellery451: this windows (travis) build is warning about `e` being unused here - you can just remove it

mellery451: ..or add it to the log message..up to you

mellery451:
the `xcode11` failure in travis is because it's picking up boost 1.71 from homebrew instead of the 1.70 that we install. Please ignore this failure for now -- I'll submit a separate CI fix for that.

src/ripple/app/misc/impl/Transaction.cpp
ShangyanLi: Thanks for catching that!

mDuo13:
I'm very excited to see this bug fixed. Thank you for tackling it. :tada:

src/ripple/app/ledger/TransactionMaster.h
```diff
@@ -39,7 +39,10 @@ class TransactionMaster
     TransactionMaster& operator= (TransactionMaster const&) = delete;
 
     std::shared_ptr<Transaction>
-    fetch (uint256 const& , bool checkDisk);
+    fetch_from_cache (uint256 const&);
+
+    std::shared_ptr<Transaction>
+    fetch (uint256 const& , int& iError);
```
seelabs: I'd change the type to `error_code_i` and rename the parameter `ec`. (here and in other places). More descriptive param names is usually a good thing, but for error codes `ec` is usually used in the codebase (`boost::asio` uses `ec` too). Side note: We also don't use hungarian notation, so we'd usually leave off the `i` in `iError`.

src/ripple/rpc/handlers/Tx.cpp
```diff
@@ -97,11 +97,12 @@ Json::Value doTx (RPC::Context& context)
     if (!isHexTxID (txid))
         return rpcError (rpcNOT_IMPL);
 
+    int error_code;
```
seelabs: If there is no error then it appears `error_code` remains unitialized. I'd initialize it to `rpcSUCCESS`.

ShangyanLi:
@seelabs sorry took forever to see and resolve this. made the changes accordingly!

seelabs:
Looking now. I think the commit message still needs to be addressed tho, right? 

Ref: https://github.com/ripple/rippled/pull/3084#pullrequestreview-299518291

seelabs:
@ShangyanLi One minor thing, I find it easier to review things if they aren't squashed down until the review ends. With a patch like this it isn't too bad, but with larger patch I find it much easier if I can easily see what's changed since the last time I looked. Thanks!

ShangyanLi:
@seelabs updated message - will take note re squashing in the future

seelabs:
@ShangyanLi Thanks for fixing the commit message, but it's still not quite right. To see what I'm seeing, run the following command: `git log --oneline` Putting a blank line between the subject and the body should fix this.

ShangyanLi:
@seelabs 3rd time's the charm 😂

seelabs:
@ShangyanLi Commit looks good here. Thanks for fixing that. :+1:

PR #3085, opened 2019-09-26
Add deletion_blockers_only param to account_objects RPC command

jwbusch:
This change will add a new boolean parameter `deletion_blockers_only` to the `account_objects` RPC command to allow users to discover which account objects are blocking their accounts from deletion.

### DIFF
--- a/src/ripple/rpc/handlers/AccountObjects.cpp
+++ b/src/ripple/rpc/handlers/AccountObjects.cpp
@@ -77,7 +77,7 @@ Json::Value doAccountObjects (RPC::Context& context)
     if (params.isMember(jss::deletion_blockers_only) &&
         params[jss::deletion_blockers_only].asBool())
     {
-        struct NameType {
+        struct {
             Json::StaticString name;
             LedgerEntryType type;
         } static constexpr deletionBlockers[] = {
--- a/src/ripple/rpc/handlers/AccountObjects.cpp
+++ b/src/ripple/rpc/handlers/AccountObjects.cpp
@@ -72,12 +72,49 @@ Json::Value doAccountObjects (RPC::Context& context)
     if (! ledger->exists(keylet::account (accountID)))
         return rpcError (rpcACT_NOT_FOUND);
 
-    auto [rpcStatus, type] = RPC::chooseLedgerEntryType(params);
-    if (rpcStatus)
+    std::optional<std::vector<LedgerEntryType>> typeFilter;
+
+    if (params.isMember(jss::deletion_blockers_only) &&
+        params[jss::deletion_blockers_only].asBool())
     {
-        result.clear();
-        rpcStatus.inject(result);
-        return result;
+        struct {
+            Json::StaticString name;
+            LedgerEntryType type;
+        } static constexpr deletionBlockers[] = {
+            { jss::check,           ltCHECK },
+            { jss::escrow,          ltESCROW },
+            { jss::payment_channel, ltPAYCHAN },
+            { jss::state,           ltRIPPLE_STATE }
+        };
+
+        typeFilter.emplace();
+        typeFilter->reserve(std::size(deletionBlockers));
+
+        for (auto [name, type] : deletionBlockers)
+        {
+            if (params.isMember(jss::type) &&
+                name != params[jss::type])
+            {
+                continue;
+            }
+
+            typeFilter->push_back(type);
+        }
+    }
+    else
+    {
+        auto [rpcStatus, type] = RPC::chooseLedgerEntryType(params);
+
+        if (rpcStatus)
+        {
+            result.clear();
+            rpcStatus.inject(result);
+            return result;
+        }
+        else if (type != ltINVALID)
+        {
+            typeFilter.emplace({ type });
+        }
     }
 
     unsigned int limit;
@@ -107,7 +144,7 @@ Json::Value doAccountObjects (RPC::Context& context)
             return RPC::invalid_field_error (jss::marker);
     }
 
-    if (! RPC::getAccountObjects (*ledger, accountID, type,
+    if (! RPC::getAccountObjects (*ledger, accountID, typeFilter,
         dirIndex, entryIndex, limit, result))
     {
         result[jss::account_objects] = Json::arrayValue;
--- a/src/ripple/rpc/handlers/AccountObjects.cpp
+++ b/src/ripple/rpc/handlers/AccountObjects.cpp
@@ -72,7 +72,7 @@ Json::Value doAccountObjects (RPC::Context& context)
     if (! ledger->exists(keylet::account (accountID)))
         return rpcError (rpcACT_NOT_FOUND);
 
-    std::optional<std::vector<LedgerEntryType>> typeFilter;
+    boost::optional<std::vector<LedgerEntryType>> typeFilter;
 
     if (params.isMember(jss::deletion_blockers_only) &&
         params[jss::deletion_blockers_only].asBool())
@@ -113,7 +113,7 @@ Json::Value doAccountObjects (RPC::Context& context)
         }
         else if (type != ltINVALID)
         {
-            typeFilter.emplace({ type });
+            typeFilter = std::vector<LedgerEntryType>({ type });
         }
     }
 
--- a/src/ripple/rpc/handlers/AccountObjects.cpp
+++ b/src/ripple/rpc/handlers/AccountObjects.cpp
@@ -77,7 +77,7 @@ Json::Value doAccountObjects (RPC::Context& context)
     if (params.isMember(jss::deletion_blockers_only) &&
         params[jss::deletion_blockers_only].asBool())
     {
-        struct {
+        struct NameType {
             Json::StaticString name;
             LedgerEntryType type;
         } static constexpr deletionBlockers[] = {
--- a/src/ripple/rpc/impl/RPCHelpers.cpp
+++ b/src/ripple/rpc/impl/RPCHelpers.cpp
@@ -82,7 +82,7 @@ accountFromString(
 
 bool
 getAccountObjects(ReadView const& ledger, AccountID const& account,
-    std::optional<std::vector<LedgerEntryType>> const& typeFilter, uint256 dirIndex,
+    boost::optional<std::vector<LedgerEntryType>> const& typeFilter, uint256 dirIndex,
     uint256 const& entryIndex, std::uint32_t const limit, Json::Value& jvResult)
 {
     auto const rootDirIndex = getOwnerDirIndex (account);
--- a/src/ripple/rpc/impl/RPCHelpers.cpp
+++ b/src/ripple/rpc/impl/RPCHelpers.cpp
@@ -82,8 +82,8 @@ accountFromString(
 
 bool
 getAccountObjects(ReadView const& ledger, AccountID const& account,
-    LedgerEntryType const type, uint256 dirIndex, uint256 const& entryIndex,
-    std::uint32_t const limit, Json::Value& jvResult)
+    std::optional<std::vector<LedgerEntryType>> const& typeFilter, uint256 dirIndex,
+    uint256 const& entryIndex, std::uint32_t const limit, Json::Value& jvResult)
 {
     auto const rootDirIndex = getOwnerDirIndex (account);
     auto found = false;
@@ -117,7 +117,18 @@ getAccountObjects(ReadView const& ledger, AccountID const& account,
         for (; iter != entries.end (); ++iter)
         {
             auto const sleNode = ledger.read(keylet::child(*iter));
-            if (type == ltINVALID || sleNode->getType () == type)
+
+            auto typeMatchesFilter = [] (
+                std::vector<LedgerEntryType> const& typeFilter,
+                LedgerEntryType ledgerType)
+            {
+                auto it = std::find(typeFilter.begin(), typeFilter.end(),
+                    ledgerType);
+                return it != typeFilter.end();
+            };
+
+            if (!typeFilter.has_value() ||
+                typeMatchesFilter(typeFilter.value(), sleNode->getType()))
             {
                 jvObjects.append (sleNode->getJson (JsonOptions::none));
 
--- a/src/test/rpc/AccountObjects_test.cpp
+++ b/src/test/rpc/AccountObjects_test.cpp
@@ -529,7 +529,6 @@ public:
             if (BEAST_EXPECT(acct_objs_is_size(resp, 1u)))
             {
                 auto const& aobjs = resp[jss::result][jss::account_objects];
-                
                 BEAST_EXPECT(
                     aobjs[0u]["LedgerEntryType"] == jss::Escrow);
             }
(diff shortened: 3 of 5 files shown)


seelabs:
I agree we need this functionality: if an account delete fails because it owns obligations, users are going to want to ask "what obligations". However, looking at the changed interface to account objects, I'm not sure this is the right place for this functionality.

The problem is the "type" parameter currently lets me restrict returned values by ledger type: trust lines, offers, payment channels, etc. This PR adds a new type "blocks deletion" that is a different animal. This is not a type of ledger object, but is a category of ledger objects. This is a problem for two reasons:

1) It makes the "type" parameter do two things: filter by object type, and filter by category. This makes the API more confusing.
2) it makes the "account objects" RPC command inconsistent with the other RPC commands that return ledger objects. Those RPC commands can only restrict by ledger object types, not ledger object "categories". This also make the API less consistent.

We already have RPC commands for "account_lines", "account_offers", etc. I suggest making a new RPC command "account_deletion_blockers" (names are hard - feel free to bike-shed a better name). I would not modify the "account_objects" command.

Side note: one modification you could make to the "account_objects" command is to take an array of account types, and only return elements from that array. That could be used as a low level building block for higher level APIs that could implement "account_deletion_blockers". But if we did that I'd probably make that a separate PR.

src/test/rpc/AccountObjects_test.cpp
```diff
@@ -494,6 +494,31 @@ class AccountObjects_test : public beast::unit_test::suite
...
+            Json::Value params;
+            params[jss::account] = gw.human();
+            params[jss::type] = jss::blocks_deletion;
+            auto resp = env.rpc("json", "account_objects", to_string(params));
+
+            constexpr Json::StaticString const expectedLedgerTypes[] = {
+                jss::Escrow, jss::Check, jss::RippleState, jss::PayChannel
+            };
+            constexpr std::uint32_t expectedAccountObjects{
+                std::size(expectedLedgerTypes)
```
seelabs: MSVC warns about a narrowing conversion here.

src/ripple/rpc/impl/RPCHelpers.cpp
```diff
@@ -117,7 +117,18 @@ getAccountObjects(ReadView const& ledger, AccountID const& account,
...
+            auto typeMatchesFilter = [] (
+                std::vector<LedgerEntryType> const& typeFilter,
+                LedgerEntryType ledgerType)
+            {
+                auto it = std::find(typeFilter.begin(), typeFilter.end(),
+                    ledgerType);
+                return it != typeFilter.end();
+            };
+
+            if (typeFilter.empty() ||
```
seelabs: I'd expect an empty vector to return no values, not all the values. Since this is part of the interface to this function, I think we should clarify the interface. The two ways I can think of are:

1) Make `typeFilter` an `optional<vector>`, and an empty optional means "no filter".
2) Make `typeFilter` a bitset that can be indexed by an ascii character. The "no filter" case would be handled by setting all the bits in this bitset.

scottschurr:
@mDuo13, I'm hoping I can get you to weigh in on a preferred way to answer the question, "Is my account deletable and if not why not?"  I think three different approaches have been lofted:

1. Add a new argument "type : blocks_deletion" to `account_objects`.  This is the approach currently taken in the pull request.  This approach has the advantage that introduces minimal disruption to the API.  It has the disadvantage that "blocks_deletion" is not really a "type".  It also has the disadvantage that it introduces an asymmetry between `account_objects`, `ledger`, and `ledger_data` which all accept a `type` argument.

2. Create a new `account_deletion_blockers` RPC command.  This has the advantage that it leaves all of the existing RPC commands unchanged.  It also widens the API the most of any of the options, since it adds an entirely new RPC command.  I believe the design of the current `account_objects` command was intended to help reduce the proliferation of similar special purpose RPC commands.

3. Modify the "type" parameter of the `account_objects`, `ledger` and `ledger_data` commands to accept an array of ledger types.  This has the advantage of maintaining consistent behavior between the three RPC commands.  It has the disadvantage that if you want to know about account deletion blockers then you have to know which ledger types to ask about.  It's less fool proof than the other two choices.

Thanks for your thoughts.

mDuo13:
I support approach (1) from those options. It's likely that the list of things that block deletion may add new types in the future, so option (3) would be a mess. I agree that it's probably best not to add even more API methods.

A potential variation that may be worth considering is a separate argument in account_objects, `"blocks_deletion": true`. This would, effectively, work like the special type value, but without introducing weirdness into the definitions of the types. It would also allow for future extensions where some types of ledger objects might only block deletion in certain circumstances. You could combine it with `type` to get just the instances of that type that block deletion, not just all instances of that type.

scottschurr:
Thanks for your views @mDuo13. @seelabs, what is your current view?  In particular, what is your take on replacing...

* `"type": "blocks_deletion"` in the proposed implementation with
* `"blocks_deletion": true`?

Would you consider that an improvement?

seelabs:
@scottschurr If we have `"bocks_deletion": true`, then what does `"blocks_deletion": false` mean? This still doesn't feel like a clean API. I suppose we could have a parameter like `"filter_type": "deletion_blockers"` that could work like the proposed  `blocks_deletion` parameter, but that's still not my preferred solution.

Here's my preferences, in order:
1) Add a new RPC command. We already have them for `account_lines`, `account_offers` and others. An `account_deletion_blockers` seems like a natural extension to the API.
2) Add a `filter_type` or `filter_types` parameter to `account_objects`.
3) Allow `types: "deletion_blockers"`. 

I understand json is not strongly typed, but it's useful to think of the parameters of APIs to have types. What is the type of that parameter? Right now we have a good answer - it's a ledger type. If we go with (3), we no longer have a good answer. It's either a ledger type or this filter type.

As usual, the above is my vote - I'm not the final work on such things. I do weight both your vote and @mDuo13 vote heavily, but I'm still voting for (1).

nbougalis:
So I was looking at **`account_objects`**. Two questions:

* Why can't we add a **`"blocks_deletion":true`* entry to _individual_ entries that do block deletion? For example:

    {
        "Balance": {
          "currency": "EUR",
          "issuer": "rrrrrrrrrrrrrrrrrrrrBZbvji",
          "value": "0"
        },
        "Flags": 2162688,
        "HighLimit": {
          "currency": "EUR",
          "issuer": "rSomeAddress",
          "value": "0"
        },
        "HighNode": "0000000000000000",
        "LedgerEntryType": "RippleState",
        "LowLimit": {
          "currency": "EUR",
          "issuer": "rSomeAddress",
          "value": "0"
        },
        "LowNode": "0000000000000000",
        "PreviousTxnID": "...",
        "PreviousTxnLgrSeq": 123456,
        "index": "...",
        "block_deletion": true
      },

Or why not simply have a new RPC command called `check_delete` or somesuch which reports the list of objects that will prevent deletion? You can either include the objects or just report their ledger **`index`**.

scottschurr:
@nikb, yup, both of those approaches would work.

Interestingly your suggestion of adding a `"blocks_deletion" true` field to `account_objects` entries was the first approach that @jwbusch and @mDuo13 suggested.  I talked them out of it because I felt like it would be more work at the server end.  It also struck me as an awkward fit for the command.  But I could easily be wrong.

Your second suggestion, add a new RPC command, aligns with @seelab's preference.

My feeling at this point is that I'm not worried about the perfect solution.  I would like us to pick something everyone can live with and move forward.  I'm just not sure how we arrive at that compromise.

mDuo13:
After spending some time thinking about it in more depth, my preference is for **a `"deletion_blockers_only": true` filter at the top level of the `account_objects` request.** Here's my rationale.

## 1. New Object Types

It's extremely likely that new types of ledger objects will be added by future amendments. History shows that most ideas for what to do with the XRP Ledger involve storing new types of data.

So in addition to other concerns, a suggestion of using an "array" of types to search for things that block deletion would be awkward because it would require client code to be updated whenever such a new feature came out.

## 2. Semi-blocking Object Types

It's easy to envision future extensions to deletable accounts that let an account be deleted even if it holds certain types of objects that would in other circumstances prevent the account from being deleted. 

In the current proposed [deletable accounts spec](https://github.com/xrp-community/standards-drafts/issues/8), the rules for what makes an account deletable don't dig into the detailed state of the objects the account holds, but there are good reasons we might want to refine those rules in the future.  For example, maybe an account could be deletable even if it has a zero-balance trust line with nonzero limits, or a trust line with authorization set by either side (since trust line authorization can't be unset). This shows that simply looking at the "type" of the objects an account owns could become insufficient to know whether an account is deletable.

This adds another reason not to add a synthetic `blocks_deletion` type to the existing `type` filter.

## 3. Non-object Blockers

It's a bit more of a stretch, but we could also conceive of other reasons an account can't be deleted. Maybe some future setting in the `AccountRoot` itself could block an account from being deletable. At the very least, not having enough XRP to pay the transaction cost of the AccountDelete transaction

src/ripple/rpc/handlers/AccountObjects.cpp
```diff
@@ -72,12 +72,49 @@ Json::Value doAccountObjects (RPC::Context& context)
...
-        return result;
+        struct {
+            Json::StaticString name;
+            LedgerEntryType type;
+        } static constexpr deletionBlockers[] = {
+            { jss::check,           ltCHECK },
+            { jss::escrow,          ltESCROW },
+            { jss::payment_channel, ltPAYCHAN },
+            { jss::state,           ltRIPPLE_STATE }
+        };
```
seelabs: Heh. I kinda like that. But I think I'd just give the struct a name, otherwise it'll take a minute for people to decipher the code, since we rarely see that construct.

src/ripple/rpc/handlers/AccountObjects.cpp
```diff
@@ -72,12 +72,49 @@ Json::Value doAccountObjects (RPC::Context& context)
...
+
+        typeFilter.emplace();
+        typeFilter->reserve(std::size(deletionBlockers));
+
+        for (auto [name, type] : deletionBlockers)
+        {
+            if (params.isMember(jss::type) &&
+                name != params[jss::type])
```
seelabs: This is reasonable behavior. We could also consider not allowing both `type` and `deletion_blockers_only: true`.

src/test/rpc/AccountObjects_test.cpp
```diff
@@ -494,6 +494,47 @@ class AccountObjects_test : public beast::unit_test::suite
...
+            Json::Value params;
+            params[jss::account] = gw.human();
+            params[jss::deletion_blockers_only] = true;
+            params[jss::type] = jss::escrow;
+            auto resp = env.rpc("json", "account_objects", to_string(params));
+
+            if (BEAST_EXPECT(acct_objs_is_size(resp, 1u)))
+            {
+                auto const& aobjs = resp[jss::result][jss::account_objects];
+
```
seelabs: Extra whitespace here.

src/ripple/rpc/handlers/AccountObjects.cpp
seelabs: I think you misunderstood me. I'm objecting to putting the declaration of the struct with the declaration of the array. _Just_ adding a typename doesn't fix that. We also need to use that typename in the declaration.

Looking more closely, I'd probably just use `std::pair` here; we don't actually use the names in the struct, so something like:
```
        static constexpr std::pair<Json::StaticString, LedgerEntryType>
            deletionBlockers[] = {{jss::check, ltCHECK},
                                  {jss::escrow, ltESCROW},
                                  {jss::payment_channel, ltPAYCHAN},
                                  {jss::state, ltRIPPLE_STATE}};
```

If you really want to keep the struct, I'd do something like:
```
        static constexpr NameType deletionBlockers[] = {
            {jss::check, ltCHECK},
            {jss::escrow, ltESCROW},
            {jss::payment_channel, ltPAYCHAN},
            {jss::state, ltRIPPLE_STATE}};
```
I prefer `std::pair` as we don't actually make use of the struct, but I don't object to the struct if you really want to keep it.

src/ripple/rpc/handlers/AccountObjects.cpp
```diff
@@ -72,12 +72,49 @@ Json::Value doAccountObjects (RPC::Context& context)
...
+
+        if (rpcStatus)
+        {
+            result.clear();
+            rpcStatus.inject(result);
+            return result;
+        }
+        else if (type != ltINVALID)
+        {
+            typeFilter.emplace({ type });
```
seelabs: If we change `typeFilter` to `boost::optional` we need to change this to  `typeFilter.emplace(type);`

scottschurr:
Effectively all travis builds are currently failing.  A quick sampling of the failures suggests this is a problem:
```
1120 ../../src/ripple/rpc/handlers/AccountObjects.cpp:147:55: error: invalid initialization of reference of type 'const boost::optional<std::vector<ripple::LedgerEntryType> >&' from expression of type 'std::optional<std::vector<ripple::LedgerEntryType> >'
1121     if (! RPC::getAccountObjects (*ledger, accountID, typeFilter,
1122                                                       ^~~~~~~~~~
1123In file included from ../../src/ripple/rpc/handlers/AccountObjects.cpp:30:
1124../../src/ripple/rpc/impl/RPCHelpers.h:65:1: note: in passing argument 3 of 'bool ripple::RPC::getAccountObjects(const ripple::ReadView&, const AccountID&, const boost::optional<std::vector<ripple::LedgerEntryType> >&, ripple::uint256, const uint256&, uint32_t, Json::Value&)'
1125 getAccountObjects (ReadView const& ledger, AccountID const& account,
1126 ^~~~~~~~~~~~~~~~~
```

src/ripple/rpc/handlers/AccountObjects.cpp
jwbusch: That ended up calling the wrong constructor for `std::vector`, the `type` enum implicitly converted to a `size_t`. Should be good now.

src/ripple/rpc/handlers/AccountObjects.cpp
jwbusch: Oh, I'd rather keep the pattern as-is then (no name for the struct). It has advantages of not polluting any namespaces, and can't be mistakenly re-used for the wrong purpose (you might be able to `decltype(deletionBlockers[0])`, but only inside the same function. I think that `std::pair` is less readable. With the first iteration I know the first parameter is meant to be the "name" for whatever this entry is, `std::pair` loses that bit of information, I just know that it's a `Json::StaticString`. 

I think `std::pair/tuple` are great for generic programming where you don't know what the names should be. But we already have all the information, so I'd rather just use a struct.

seelabs: I'm fine with keeping the original.

nbougalis: I find `struct` superior to `std::pair` 99.9% of the time. For me, this happens to be the 0.1% of the time when it’s the other way around.

We never use the names of the `struct` directly and the code is very localized—all within 10-15 lines.

I’d much prefer if we did: `std::pair<Json::StaticString, LedgerEntryType> = { { blah, blah }, ... }`.

Rationale: the nameless struct declaration is cool but “exotic” enough that people will do a double take; the names introduced by the `struct` and the strong typing don’t help, especially given the limited scope.

I’m not speaking _ex cathedra_ and I’m not gonna hold this up over this; just my personal opinion.

PR #3086, opened 2019-09-26
Add metrics for PeerImp to track bandwidth usage of peers

jwbusch:
This change will allow server operators to see how much bandwidth each peer is consuming (both send/recv).

### DIFF
--- a/src/ripple/overlay/impl/PeerImp.cpp
+++ b/src/ripple/overlay/impl/PeerImp.cpp
@@ -375,6 +375,12 @@ PeerImp::json()
         }
     }
 
+    ret[jss::metrics] = Json::Value(Json::objectValue);
+    ret[jss::metrics][jss::bytes_recv_total] = Json::UInt(metrics_.recv.total_bytes());
+    ret[jss::metrics][jss::bytes_sent_total] = Json::UInt(metrics_.sent.total_bytes());
+    ret[jss::metrics][jss::bytes_recv_avg] = Json::UInt(metrics_.recv.average_bytes());
+    ret[jss::metrics][jss::bytes_sent_avg] = Json::UInt(metrics_.sent.average_bytes());
+
     return ret;
 }
 
@@ -846,6 +852,8 @@ PeerImp::onReadMessage (error_code ec, std::size_t bytes_transferred)
             stream << "onReadMessage";
     }
 
+    metrics_.recv.add_message(bytes_transferred);
+
     read_buffer_.commit (bytes_transferred);
 
     while (read_buffer_.size() > 0)
@@ -892,6 +900,8 @@ PeerImp::onWriteMessage (error_code ec, std::size_t bytes_transferred)
         else
             stream << "onWriteMessage";
     }
+
+    metrics_.sent.add_message(bytes_transferred);
 
     assert(! send_queue_.empty());
     send_queue_.pop();
@@ -2850,5 +2860,37 @@ PeerImp::isHighLatency() const
     std::lock_guard sl (recentLock_);
     return latency_ >= Tuning::peerHighLatency;
 }
+
+void
+PeerImp::Metrics::add_message(std::uint64_t bytes)
+{
+    using namespace std::chrono_literals;
+
+    totalBytes += bytes;
+    accumBytes += bytes;
+    auto const actualInterval = clock_type::now() - intervalStart;
+
+    if (actualInterval >= 1s)
+    {
+        auto const avgBytes = accumBytes / std::chrono::duration_cast<std::chrono::seconds>(actualInterval).count();
+        rollingAvg[currIndex] = avgBytes;
+        rollingAvgBytes = std::accumulate(rollingAvg.begin(), rollingAvg.end(), 0ull) / rollingAvg.size();
+
+        currIndex = (currIndex + 1) % rollingAvg.size();
+        intervalStart = clock_type::now();
+        accumBytes = 0;
+    }
+}
+
+std::uint64_t
+PeerImp::Metrics::average_bytes() const {
+    return rollingAvgBytes;
+}
+
+std::uint64_t
+PeerImp::Metrics::total_bytes() const {
+    return totalBytes;
+}
+
 
 } // ripple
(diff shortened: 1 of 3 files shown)


src/ripple/overlay/impl/PeerImp.cpp
```diff
@@ -2851,4 +2861,36 @@ PeerImp::isHighLatency() const
...
+
+        currIndex = (currIndex + 1) % rollingAvg.size();
+        intervalStart = clock_type::now();
+        accumBytes = 0;
+    }
+}
+
+std::uint64_t
+PeerImp::Metrics::average_bytes() const {
+    return rollingAvgBytes;
```
nbougalis: I know that 32-bit values are atomic, but this (and the next) probably needs to be protected by a lock.

src/ripple/protocol/jss.h
```diff
@@ -139,6 +139,10 @@ JSS ( both );                       // in: Subscribe, Unsubscribe
 JSS ( both_sides );                 // in: Subscribe, Unsubscribe
 JSS ( build_path );                 // in: TransactionSign
 JSS ( build_version );              // out: NetworkOPs
+JSS ( bytes_recv_avg );             // out: Peers
```
nbougalis: This (and the next) should probably be "avg_bps_recv" and "avg_bps_sent"

src/ripple/overlay/impl/PeerImp.cpp
scottschurr: FWIW, please never assume that any word size (even an octet) is atomic.  In C++11 and greater _any_ race condition is undefined behavior.

src/ripple/overlay/impl/PeerImp.h
```diff
@@ -39,6 +39,7 @@
 #include <deque>
 #include <queue>
 #include <shared_mutex>
+#include <numeric>
```
jwbusch: nit: move to cpp

src/ripple/overlay/impl/PeerImp.cpp
jwbusch: Wasn't assuming it was atomic, it's just not a critical code-path and didn't want to over complicate it. The only "bad thing" is you get a slightly older answer. But I can put it behind a lock for correctness :)

src/ripple/overlay/impl/PeerImp.h
```diff
@@ -198,6 +199,26 @@ class PeerImp
...
+    public:
+        void add_message(std::uint64_t bytes);
+        std::uint64_t average_bytes() const;
+        std::uint64_t total_bytes() const;
+
+    private:
+        std::uint64_t totalBytes = 0;
+        std::uint64_t accumBytes = 0;
+        std::uint64_t rollingAvgBytes = 0;
+        std::array<std::uint64_t, 30> rollingAvg = {};
```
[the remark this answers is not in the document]
jwbusch: Didn't know that existed (I should just assume `boost` has everything), but agreed, I'll take a look at it.

src/ripple/overlay/impl/PeerImp.cpp
nbougalis: >FWIW, please never assume that any word size (even an octet) is atomic. In C++11 and greater any race condition is undefined behavior.

I was talking about the underlying x86 semantics. But your point is well-made.

src/ripple/overlay/impl/PeerImp.h
nbougalis: A self-nit! I hereby award you: :1st_place_medal:

src/ripple/overlay/impl/PeerImp.h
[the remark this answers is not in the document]
HowardHinnant: I believe `boost::circular_buffer` is a heap-based container.  Since `Metrics` is not being moved, `std::array` as an advantage in being stack allocated.

src/ripple/overlay/impl/PeerImp.h
```diff
@@ -198,6 +199,26 @@ class PeerImp
 
     friend class OverlayImpl;
 
+    class Metrics {
+    public:
```
HowardHinnant: I recommend adding:
```
        Metrics() = default;
        Metrics(Metrics const&) = delete;
        Metrics& operator=(Metrics const&) = delete;
```

This helps reader know that this class will not be copied/moved.

src/ripple/protocol/jss.h
jwbusch: to mirror that I renamed the other ones to `total_bytes_recv` and `total_bytes_sent`

src/ripple/overlay/impl/PeerImp.h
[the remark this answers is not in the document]
jwbusch: I opt'd for readability this time around...

src/ripple/overlay/impl/PeerImp.cpp
jwbusch: I added a `shared_mutex` but still debating if we should just make the `uint64_t`'s atomic,

src/ripple/overlay/impl/PeerImp.cpp
```diff
@@ -375,6 +376,12 @@ PeerImp::json()
         }
     }
 
+    ret[jss::metrics] = Json::Value(Json::objectValue);
+    ret[jss::metrics][jss::total_bytes_recv] = Json::UInt(metrics_.recv.total_bytes());
+    ret[jss::metrics][jss::total_bytes_sent] = Json::UInt(metrics_.sent.total_bytes());
```
jwbusch: I did notice that `Json::UInt` is a 32-bit int, wondering if there's a better type here given that `total_bytes` is a 64-bit int.

src/ripple/overlay/impl/PeerImp.h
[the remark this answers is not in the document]
nbougalis: It makes sense that it's heap-based, but I expected a fixed-sized version exists. I think that's OK for our purposes, and simplifies things, so I'm ok with it. Happy to reconsider if you have strong objections, @HowardHinnant.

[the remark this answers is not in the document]
HowardHinnant: No objections here.

PR #3093, opened 2019-10-02
Deletable Accounts

scottschurr:
An implementation of the Deletable Accounts spec: https://github.com/xrp-community/standards-drafts/issues/8

Reviewers: @ximinez and @seelabs for transactor expertise.  Thanks folks.


src/ripple/app/tx/impl/DeleteAccount.cpp
```diff
@@ -0,0 +1,337 @@
...
+    if (mulDivResult.first)
+        return mulDivResult.second;
+
+    // If mulDiv returns false then overflow happened.  Punt by using the
+    // standard calculation.
+    return Transactor::calculateBaseFee (view, tx);
+}
+
+namespace 
+{
```
seelabs: In the past anon namespaces have been banned because of unity builds. I'm fine with them tho.

src/ripple/app/tx/impl/DeleteAccount.cpp
```diff
@@ -0,0 +1,337 @@
...
+        {
+            if (deletable.type == nodeType)
+            {
+                found = true;
+                break;
+            }
+        }
+        if (! found)
+            // There's at least one directory entry that we can't delete.  Stop.
+            return tecHAS_OBLIGATIONS;
```
seelabs: I'd refactor this into a function that takes the ledger entry type and returns the pointer to the deleter function (if available). It can be used here and in the function below. I'll provide a couple patches you can cherry pick if you want to do this (patch will be in the "summary" comment so all the patches are together).

This also lets us remove the `CanDelete` struct and `deletables` array

src/ripple/app/tx/impl/DepositPreauth.h
```diff
@@ -42,6 +42,12 @@ class DepositPreauth
     preclaim(PreclaimContext const& ctx);
 
     TER doApply () override;
+
+    // Interface used by DeleteAccount
+    static
+    TER
+    removeDepositPreauthFromLedger (Application& app, ApplyView& view,
```
seelabs: I would rename this `removeFromLedger` Rational: it's a class static function, it's should be called as `DepositPreauth::removeDepositPreauthFromLedger`. `DepositPreauth` is written twice, and there's no need for that.

Also, it's easier to use such functions in generic code, i.e. we can say `T::removeFromLedger` when `T` is a template arg, but there is no way to say `T::removeTFromLedger` without using macros. I'll provide a patch for this in the summary comment

src/ripple/app/tx/impl/SetSignerList.h
```diff
@@ -67,6 +67,12 @@ class SetSignerList : public Transactor
     TER doApply () override;
     void preCompute() override;
 
+    // Interface used by DeleteAccount
+    static
+    TER
+    removeSignersFromLedger (
```
seelabs: Rename this `removeFromLedger` (see comment on `removeDepositPreauthFromLedger`). I will provide a patch in the summary comment.

src/ripple/app/tx/impl/DepositPreauth.cpp
```diff
@@ -154,35 +156,50 @@ DepositPreauth::doApply ()
...
+DepositPreauth::removeDepositPreauthFromLedger (Application& app,
+    ApplyView& view, uint256 const& preauthIndex, beast::Journal j)
+{
+    // Verify that the Preauth entry they asked to remove is
+    // in the ledger.
+    std::shared_ptr<SLE> const slePreauth {
+        view.peek (keylet::depositPreauth (preauthIndex))};
+    if (! slePreauth)
+    {
+        // Error should have been caught in preclaim.
```
seelabs: This is a static public function. I don't think `DepositPreauth::preclaim` is still a precondition to calling this. I'd probably just remove the comment

src/ripple/app/tx/impl/DepositPreauth.cpp
```diff
@@ -154,35 +156,50 @@ DepositPreauth::doApply ()
...
+        JLOG(j.warn()) << "Selected DepositPreauth does not exist.";
+        return tecNO_ENTRY;
+    }
 
-        // Remove DepositPreauth from ledger.
-        view().erase (slePreauth);
+    AccountID const account {(*slePreauth)[sfAccount]};
+    auto viewJ = app.journal ("View");
```
seelabs: We're passing in a journal as a parameter. Why are we creating another one here?

src/ripple/app/tx/impl/SetSignerList.h
```diff
@@ -85,8 +91,11 @@ class SetSignerList : public Transactor
     TER replaceSignerList ();
     TER destroySignerList ();
 
-    TER removeSignersFromLedger (Keylet const& accountKeylet,
-        Keylet const& ownerDirKeylet, Keylet const& signerListKeylet);
+    static
+    TER removeSignersFromLedger (
```
seelabs: Why a private static? I'd expect this to be a file-scoped function in the `cpp`. Because of unity build issues or something else?

src/ripple/ledger/ApplyView.h
```diff
@@ -371,6 +371,19 @@ class ApplyView
...
+        @param directory the identifier of the directory node to be deleted
+        @return \c true if the directory was found was succesfully deleted
+                \c false otherwise.
+
+        @note The function should only be called with the root entry (i.e. with
+              the first page) of a directory. Calling it with an intermediate
+              page can result in undefined behavior.
+    */
+    bool
+    dirDelete(Keylet const& directory);
```
seelabs: So `dirRemove` removes an entry (and possibly the directory as well) and `dirDelete` removes the directory. I think those two names are too similar to eachother. How about `rmDirEntry` and `rmDir`?

As a side note: both implementation are similar. We could probably refactor them to use a helper function, tho we may want to punt on both of these for this PR and do a refactor in another PR if we want.

src/ripple/ledger/impl/ApplyStateTable.cpp
```diff
@@ -562,10 +562,10 @@ ApplyStateTable::getForMod (ReadView const& base,
             auto const& item = iter->second;
             if (item.first == Action::erase)
             {
-                // VFALCO We need to think about throwing
-                //        an exception or calling LogicError
-                JLOG(j.fatal()) <<
-                    "Trying to thread to deleted node";
+                // The Destination of an Escrow or a PayChannel may have been
```
seelabs: Dst of escrow cannot be deleted

src/ripple/ledger/impl/ApplyStateTable.cpp
```diff
@@ -596,14 +596,12 @@ ApplyStateTable::threadTx (ReadView const& base,
...
         keylet::account(to).key, mods, j);
-    assert(sle);
     if (! sle)
     {
-        // VFALCO We need to think about throwing
-        //        an exception or calling LogicError
-        JLOG(j.fatal()) <<
-            "Threading to non-existent account: " <<
-                toBase58(to);
+        // The Destination of an Escrow or PayChannel may have been deleted.
```
seelabs: Dst of escrow cannot be deleted

src/test/app/AccountDelete_test.cpp
```diff
@@ -0,0 +1,699 @@
...
+        env (offer (carol, gw["USD"](51), XRP (51)));
+        env (signers (carol, 1, {{alice, 1}, {becky, 1}}));
+
+        // Deleting should fail with TOO_SOON, which is a relatively
+        // cheap check compared to validating the contents of her directory.
+        env (acctdelete (alice, becky), fee (acctDelFee), ter (tecTOO_SOON));
+
+        // Close enough ledgers to almost be able to delete alice's account.
+        std::uint32_t const ledgerCount {
+            openLedgerSeq (env) + 256 - env.seq (alice)};
```
seelabs: This doesn't look right. The larger alice's account seq, the more ledger we should need to close. As written, that's backwards. I have a patch that has a function for what I think is the correct calculation (there are other instances of this bug in this file as well, the patch should address all of them.)

src/test/app/AccountDelete_test.cpp
```diff
@@ -0,0 +1,699 @@
...
+        // the amendment to get new-style PayChannels.
+        Env env {*this, supported_amendments() - fixPayChanRecipientOwnerDir};
+        Account const alice("alice");
+        Account const becky("becky");
+        Account const gw("gw");
+
+        env.fund(XRP(100000), alice, becky, gw);
+        env.close();
+
+        // Give alice and becky a whole passel of offers that we have to
```
seelabs: `passel` isn't a common word. I'm fine if you really want to keep it, but I'd probably lose it.

It is a cool word tho :-)

src/test/app/AccountDelete_test.cpp
```diff
@@ -0,0 +1,699 @@
...
+        env (payChanCreate (
+            alice, becky, XRP(57), 4s, env.now() + 2s, alice.pk()));
+        env.close();
+
+        // An old-style PayChannel does not add a back link from the
+        // destination.  So with the PayChannel in place becky should be
+        // able to delete her account, but alice should not.
+        //
+        // NOTE: this test will need to change after the
+        // fixPayChanRecipientOwnerDir amendment is introduced.
```
seelabs: the amendment is there. I'd remove the comment

scottschurr:
Added a unit test suggested by @ximinez and @mDuo13.

src/test/rpc/AccountTx_test.cpp
```diff
@@ -28,6 +28,78 @@ namespace test {
...
+    {
+        BEAST_EXPECT(txNode[jss::validated].asBool() == true);
+        BEAST_EXPECT(
+            txNode[jss::tx][sfTransactionType.jsonName].asString() == 
+            sane.txType);
+
+        // Make sure all of the expected node types are present.
+        std::vector<std::string> createdNodes {};
+        std::vector<std::string> deletedNodes {};
+        std::vector<std::string> modifiedNodes{};
```
seelabs: I'd replace the `std::vector` here and the `std::initializer_list` above with a boost `flat_set`

Using unsorted vectors and initializer_list complicates the test:
1) They require a lambea to sort the values before they are compared
2) One of the parameters to the lambda `cmpNodeTypes` is modified, something I
wouldn't expect from what looks like a pure function
3) The other parameter is copied

Replaceing these containers with a `flat_set` gets rid of the lambda, and lets
us use the clearer form: `BEAST_EXPECT(createdNodes == sane.created)`
 Here's a patch to cherry pick/review if you want (that patch also fixes some white space issues in this file)

Patch: https://github.com/seelabs/rippled/commit/bcda862d76498b9ddb78fe1b6a6653a70ce2e8c3

src/test/rpc/AccountTx_test.cpp
```diff
@@ -28,6 +28,78 @@ namespace test {
 
 class AccountTx_test : public beast::unit_test::suite
 {
+    // A data structure used to describe the basic structure of a
```
seelabs: Extra whitespace here

src/test/rpc/AccountTx_test.cpp
```diff
@@ -28,6 +28,78 @@ namespace test {
...
+        std::initializer_list<char const*> modified;
+    };
+
+    // A helper method tests can use to validate returned JSON vs NodeSanity.
+    void
+    checkSanity (Json::Value const& txNode, NodeSanity const& sane)
+    {
+        BEAST_EXPECT(txNode[jss::validated].asBool() == true);
+        BEAST_EXPECT(
+            txNode[jss::tx][sfTransactionType.jsonName].asString() ==
```
seelabs: Extra whitespace here

src/test/rpc/AccountTx_test.cpp
```diff
@@ -447,12 +450,107 @@ class AccountTx_test : public beast::unit_test::suite
...
+        env.close();
+
+        // Actually resurrect becky's account.
+        env (pay (alice, becky, XRP (45)));
+        env.close();
+
+        // becky's account root should be back.
+        BEAST_EXPECT (env.closed()->exists (beckyAcctKey));
+        BEAST_EXPECT (env.balance (becky) == XRP (45));
+
```
seelabs: Extra whitespace here

src/test/app/Escrow_test.cpp
```diff
@@ -1277,6 +1277,134 @@ struct Escrow_test : public beast::unit_test::suite
...
+
+        // Neither evita not fiona should be able to delete their accounts
+        // once this escrow is created.
+        std::uint32_t const evitaEscrowSeq {env.seq(evita)};
+        env(escrow(evita, fiona, XRP(1000)), finish_time(env.now() + 1s));
+        env.close();
+
+        // Close enough ledgers to be able to delete any of the accounts.
+        std::uint32_t const ledgerCount {
+            env.current()->seq() + 257 - env.seq (alice)};
```
seelabs: I don't believe this is the correct ledgerCount. See this comment: https://github.com/ripple/rippled/pull/3093#discussion_r331691858 and this patch: https://github.com/seelabs/rippled/commit/f310e26b5e73ce5f0097521a574e7e7bd7ec9e07

(note that patch fixes other instances of this bug, but does not fix it here)

src/test/app/Escrow_test.cpp
```diff
@@ -1277,6 +1277,134 @@ struct Escrow_test : public beast::unit_test::suite
...
+        // her account.
+        env (acctdelete (alice, grace), fee (acctDelFee), ter (tecHAS_OBLIGATIONS));
+
+        // The fact that the destination of carol's escrow has evaporated
+        // should have no impact on whether carol's escrow can be canceled.
+        env(cancel(carol, carol, carolEscrowSeq));
+        env.close();
+
+        // Now that carol's escrow is gone carol should be able to delete
+        // her account.
```
seelabs: Extra whitespace.

src/test/app/Escrow_test.cpp
```diff
@@ -1277,6 +1277,134 @@ struct Escrow_test : public beast::unit_test::suite
...
+        // should have no impact on whether carol's escrow can be canceled.
+        env(cancel(carol, carol, carolEscrowSeq));
+        env.close();
+
+        // Now that carol's escrow is gone carol should be able to delete
+        // her account. 
+        env (acctdelete (carol, grace), fee (acctDelFee));
+        env.close();
+
+        // We'll now resurrect bruce's account.  alice should then be able to
```
seelabs: Extra whitespace.

src/ripple/app/tx/impl/DeleteAccount.cpp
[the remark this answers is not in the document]
seelabs: Extra whitespace

src/ripple/app/tx/impl/InvariantCheck.cpp
```diff
@@ -283,19 +297,37 @@ NoZeroEscrow::finalize(STTx const& tx, TER const, XRPAmount const, beast::Journa
...
     std::shared_ptr <SLE const> const&)
 {
     if (isDelete && before && before->getType() == ltACCOUNT_ROOT)
-        accountDeleted_ = true;
+        accountsDeleted_++;
 }
 
 bool
-AccountRootsNotDeleted::finalize(STTx const&, TER const, XRPAmount const, beast::Journal const& j)
+AccountRootsNotDeleted::finalize(
```
seelabs: Extra whitespace

src/ripple/app/tx/impl/InvariantCheck.cpp
```diff
@@ -388,5 +428,58 @@ NoXRPTrustLines::finalize(STTx const&, TER const, XRPAmount const, beast::Journa
...
+{
+    if (!before && after->getType() == ltACCOUNT_ROOT)
+    {
+        accountsCreated_++;
+        accountSeq_ = (*after)[sfSequence];
+    }
+}
+
+bool
+ValidNewAccountRoot::finalize(
```
seelabs: Extra whitespace

src/ripple/app/tx/impl/DeleteAccount.cpp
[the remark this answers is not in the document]
scottschurr: Thanks for pointing out all the extra whitespace.  This surprised me, since up until recently Sublime Text has stripped trailing whitespace on exit for me.  I believe I've put a fix in my config file now, so I think things will hopefully be better.

src/ripple/app/tx/impl/DeleteAccount.cpp
```diff
@@ -0,0 +1,337 @@
...
+        (sleDst->getFlags() & lsfDepositAuth))
+    {
+        if (! ctx.view.exists (keylet::depositPreauth (dst, account)))
+            return tecNO_PERMISSION;
+    }
+
+    auto sleAccount = ctx.view.read (keylet::account(account));
+    assert(sleAccount);
+    if (! sleAccount)
+        return terNO_ACCOUNT;
```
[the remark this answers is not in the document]
scottschurr: Right.  My goal for this particular pass was to find all places where a missing account root in the ledger might crash the application.  For that pass I decided that overkill would be okay.  The checks are low cost.  It was easier for me to make the pass by simply thinking locally, inside of a given function, rather than worrying about currently guaranteed code flow between functions.

I don't feel strongly that this check is essential -- it's not.  But the presence of the check reduces my cognitive load somewhat when I'm thinking about what happens if an account root goes missing in the ledger.

src/ripple/app/tx/impl/DeleteAccount.cpp
scottschurr: Cherry-picked 442f3cab6480dfafb291c4352ace3157671e75d3.

src/ripple/ledger/impl/ApplyStateTable.cpp
seelabs: Follow-up on my own comment. I'm wrong about the above. I forgot about pre-fix escrows. So no change needed.

src/ripple/ledger/impl/ApplyStateTable.cpp
seelabs: Follow-up on my own comment. I'm wrong about the above. I forgot about pre-fix escrows. So no change needed.

scottschurr:
Rebased and most recent review comments addressed.

src/ripple/ledger/View.h
```diff
@@ -56,24 +56,24 @@ enum FreezeHandling
     fhZERO_IF_FROZEN
 };
 
-bool
+[[nodiscard]] bool
 isGlobalFrozen (ReadView const& view,
     AccountID const& issuer);
 
-bool
+[[nodiscard] ]bool
```
ximinez: Nit: `[[nodiscard]] bool`, just for consistency

scottschurr: Oooh!  Good eyes.  I'll go ahead and squash, but I'll fix that during the squash.

Thanks for all the help everybody!

PR #3094, opened 2019-10-03
Small bug fixes in BuildLedger.cpp

scottschurr:
As I was working in other areas I wandered across a couple of lines of code in BuildLedger.cpp that seemed not quite right.  This is my best stab at what I think the code intended to be doing.

Reviewers: @bachase and @nbougalis for previous experience in this file.  Thanks folks!


bachase:
LGTM. Good catch

PR #3095, opened 2019-10-04
Add option for -Wextra to build

mellery451:
fixed warnings that were reported by current compilers. Added a suppress for `unused-parameter`, although we can revisit that particular warning if we think there is any value.

### DIFF
--- a/src/ripple/app/main/Application.cpp
+++ b/src/ripple/app/main/Application.cpp
@@ -1566,19 +1566,8 @@ bool ApplicationImp::setup()
 
         Resource::Charge loadType = Resource::feeReferenceRPC;
         Resource::Consumer c;
-        RPC::Context context {
-            journal ("RPCHandler"),
-            jvCommand,
-            *this,
-            loadType,
-            getOPs (),
-            getLedgerMaster (),
-            c,
-            Role::ADMIN,
-            {},
-            {},
-            {{},{}}
-        };
+        RPC::Context context { journal ("RPCHandler"), jvCommand, *this,
+            loadType, getOPs (), getLedgerMaster (), c, Role::ADMIN};
 
         Json::Value jvResult;
         RPC::doCommand (context, jvResult);
--- a/src/ripple/app/main/Application.cpp
+++ b/src/ripple/app/main/Application.cpp
@@ -1566,8 +1566,19 @@ bool ApplicationImp::setup()
 
         Resource::Charge loadType = Resource::feeReferenceRPC;
         Resource::Consumer c;
-        RPC::Context context { journal ("RPCHandler"), jvCommand, *this,
-            loadType, getOPs (), getLedgerMaster(), c, Role::ADMIN };
+        RPC::Context context {
+            journal ("RPCHandler"),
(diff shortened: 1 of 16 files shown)


src/ripple/app/main/Application.cpp
```diff
@@ -1566,8 +1566,19 @@ bool ApplicationImp::setup()
...
+            jvCommand,
+            *this,
+            loadType,
+            getOPs (),
+            getLedgerMaster (),
+            c,
+            Role::ADMIN,
+            {},
+            {},
+            {{},{}}
```
HowardHinnant: What was the warning that motivated the addition of these last three member initializations?

mellery451: `-Wmissing-field-initializers` (clang). If you'd prefer I disable that one, I'm happy to do that.

HowardHinnant: Thanks.  What do you think about silencing them in Context.h instead, like this:
```
struct Context
{
    // ...
    std::shared_ptr<JobQueue::Coro> coro{};
    InfoSub::pointer infoSub{};
    Headers headers{};
};
```
It isn't a huge deal because we only construct like this in one place.  But when I read the code in Application.cpp, I had to look up in Context.h to see what it was actually doing.  If we do it in Context.h, it seems easier to read to me.

mellery451: yep - I like that better. will change.

HowardHinnant: Thanks.

src/ripple/app/misc/impl/Transaction.cpp
```diff
@@ -63,7 +63,9 @@ void Transaction::setStatus (TransStatus ts, std::uint32_t lseq)
 TransStatus Transaction::sqlTransactionStatus(
     boost::optional<std::string> const& status)
 {
-    char const c = (status) ? (*status)[0] : txnSqlUnknown;
+    char const c = (status) ?
+        (*status)[0] :
+        static_cast<std::underlying_type<TxnSql>::type>(txnSqlUnknown);
```
HowardHinnant: `safe_cast` lives for this kind of stuff. :-)

```
#include <ripple/basics/safe_cast.h>
...
safe_cast<char>(txnSqlUnknown);
```

mellery451: 👍  will do

src/ripple/overlay/PeerReservationTable.h
```diff
@@ -89,7 +89,7 @@ class PeerReservationTable final
     contains(PublicKey const& nodeId)
     {
         std::lock_guard lock(this->mutex_);
-        return table_.find({nodeId}) != table_.end();
+        return table_.find({nodeId,{}}) != table_.end();
```
HowardHinnant: Here too?
```
std::string description{};
```
And ditto...

scottschurr:
@HowardHinnant, thanks for suggesting `safe_cast` and moving the initializers.  I wish I'd thought of that.  :+1:

PR #3107, opened 2019-10-11
Proposed 1.4.0-b6

manojsdoshi:
If merged this PR will:

close: #3104 
close: #3096 
close: #3095 
close: #3094 
close: #3086 
close: #3083 
close: #3047


MarkusTeufelberger:
Could we get #3055 in there too?

nbougalis:
We should consider it. @seelabs, comments?

seelabs:
@nbougalis @MarkusTeufelberger my plan was to wait for 1.5-b1 before we merged 1.71 support. While that PR is relatively low-risk I don't think 1.4 needs it, and since we're getting close to a release-candidate I'd rather wait.

MarkusTeufelberger:
Well, then my node still will have to wait a few weeks/months(?) until I can compile `rippled` again. :-(

Your call, but I assume Alpine maintainers for example would be really happy to use their packaged version of boost (https://pkgs.alpinelinux.org/package/edge/main/x86_64/boost) or Arch linux ones too (https://www.archlinux.org/packages/extra/x86_64/boost/) instead of having to port your patch to 1.4.0.

nbougalis:
Hmm, interesting @MarkusTeufelberger. Do the Alpine maintainers publish a `rippled` package in their own repos?

MarkusTeufelberger:
They are actually the only distro I know of that do: https://pkgs.alpinelinux.org/package/edge/community/x86_64/rippled

seelabs:
@MarkusTeufelberger Thanks for the information on Alpine linux. After further discussion we've decided to add the boost 1.71 PR to this beta.

Issue #2924, opened 2019-05-01
tx - txnNotFound Error should include which ledger versions it searched


intelliot:
(Copied from RIPD-1705)

With online deletion, there is a race condition where a client could incorrectly conclude that a transaction is not in a validated ledger because that ledger was deleted after the client queried which ledgers were available on the server.

This is an edge case in reliable transaction submission: you have to know that a transaction is not in a range of validated ledgers before you conclude that it has failed. The scenario in question goes something like this (where "you" are a rippled client):

* You submit a transaction T.
* Transaction T becomes validated in ledger M, but you don't know that yet.
* You see that the server has ledger M in its list of available ledgers, either because you're subscribed to the server's `ledgerClosed` stream, or by checking `server_info`.
* After you've asked about the available ledgers but before you query for the transaction, the server deletes ledger M. (If you're following the `ledgerClosed` stream, this would have to be after a number of ledger versions have closed equal to or greater than `online_delete`, which is minimum ~15 minutes. If you asked via `server_info`, this could happen immediately if the ledger was in the range to be deleted when `online_delete` next ran.)
* You ask the server about transaction T and get a `txnNotFound` response from the server.
* You incorrectly conclude that the transaction T is not in ledger M, because you thought that the server had ledger M.

Even if you check server_info again after not finding the transaction, there's a small chance that the ledger in question was just finished downloading in between. Naturally, most of this stuff is unlikely, but becomes more plausible if there's an unusual amount of network latency between client and server.

One very straightforward and robust fix for this problem would be for the txnNotFound response to include a list of ledger indexes it checked. Essentially, paste the `complete_ledgers` field of `server_info`—all the validated ledger versions the server checked—into the error message. (It could also be in the form of `ledger_index_min` / `ledger_index_max` fields as the `account_tx` command returns.) That way, the client knows without risk of race conditions exactly which ledgers the transaction is not in.

intelliot:
https://github.com/ripple/rippled/pull/3145 -- included in 1.5.0-b1

PR #3112, opened 2019-10-15
Provide the range of ledgers searched when returning a 'txnNotFound' …

undertome:
…error.

FIXES: #2924

* Create TransactionEnvelope struct. (Encapsulates transaction with ledger info)
* Return a TransactionEnvelope when fetching transactions from the database.
* Provide ledger info when returning a 'txnNotFound' error.

### DIFF
--- a/src/ripple/app/ledger/impl/TransactionMaster.cpp
+++ b/src/ripple/app/ledger/impl/TransactionMaster.cpp
@@ -44,22 +44,24 @@ bool TransactionMaster::inLedger (uint256 const& hash, std::uint32_t ledger)
     return true;
 }
 
-std::shared_ptr<Transaction>
-TransactionMaster::fetch (uint256 const& txnID, bool checkDisk)
+auto
+TransactionMaster::fetch (uint256 const& txnID, bool checkDisk) -> TransactionEnvelope
 {
-    auto txn = mCache.fetch (txnID);
+    TransactionEnvelope ret;
 
-    if (!checkDisk || txn)
-        return txn;
+    ret.txn = mCache.fetch (txnID);
 
-    txn = Transaction::load (txnID, mApp);
+    if (!checkDisk || ret.txn)
+        return ret;
 
-    if (!txn)
-        return txn;
+    ret = Transaction::load (txnID, mApp);
 
-    mCache.canonicalize (txnID, txn);
+    if (!ret.txn)
+        return ret;
 
-    return txn;
+    mCache.canonicalize (txnID, ret.txn);
+
+    return ret;
 }
 
 std::shared_ptr<STTx const>
@@ -68,7 +70,7 @@ TransactionMaster::fetch (std::shared_ptr<SHAMapItem> const& item,
         bool checkDisk, std::uint32_t uCommitLedger)
 {
     std::shared_ptr<STTx const>  txn;
-    auto iTx = fetch (item->key(), false);
+    auto iTx = fetch (item->key(), false).txn;
 
     if (!iTx)
     {
--- a/src/ripple/app/ledger/impl/TransactionMaster.cpp
+++ b/src/ripple/app/ledger/impl/TransactionMaster.cpp
@@ -44,24 +44,22 @@ bool TransactionMaster::inLedger (uint256 const& hash, std::uint32_t ledger)
     return true;
 }
 
-auto
-TransactionMaster::fetch (uint256 const& txnID, bool checkDisk) -> TransactionEnvelope
+std::shared_ptr<Transaction>
+TransactionMaster::fetch (uint256 const& txnID, bool checkDisk)
 {
-    TransactionEnvelope ret;
+    auto txn = mCache.fetch (txnID);
 
-    ret.txn = mCache.fetch (txnID);
+    if (!checkDisk || txn)
+        return txn;
 
-    if (!checkDisk || ret.txn)
-        return ret;
+    txn = Transaction::load (txnID, mApp);
 
-    ret = Transaction::load (txnID, mApp);
+    if (!txn)
+        return txn;
 
-    if (!ret.txn)
-        return ret;
+    mCache.canonicalize (txnID, txn);
 
-    mCache.canonicalize (txnID, ret.txn);
-
-    return ret;
+    return txn;
 }
 
 std::shared_ptr<STTx const>
@@ -70,7 +68,7 @@ TransactionMaster::fetch (std::shared_ptr<SHAMapItem> const& item,
         bool checkDisk, std::uint32_t uCommitLedger)
 {
     std::shared_ptr<STTx const>  txn;
-    auto iTx = fetch (item->key(), false).txn;
+    auto iTx = fetch (item->key(), false);
 
     if (!iTx)
     {
--- a/src/ripple/app/misc/Transaction.h
+++ b/src/ripple/app/misc/Transaction.h
@@ -33,6 +33,7 @@ namespace ripple {
 // obtain a binary version.
 //
 
+class TransactionEnvelope;
 class Application;
 class Database;
 class Rules;
@@ -156,7 +157,7 @@ public:
 
     Json::Value getJson (JsonOptions options, bool binary = false) const;
 
-    static Transaction::pointer load (uint256 const& id, Application& app);
+    static TransactionEnvelope load (uint256 const& id, Application& app);
 
 private:
     uint256         mTransactionID;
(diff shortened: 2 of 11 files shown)


scottschurr:
FWIW, continuous integration is failing.  Here is one of the error messages:
```
In file included from ../../src/ripple/app/ledger/impl/LedgerMaster.cpp:31:
../../src/ripple/app/misc/Transaction.h:175:1: error: 'TransactionEnvelope' defined as a struct here but previously declared as a class; this is valid, but may result in linker errors under the Microsoft C++ ABI [-Werror,-Wmismatched-tags]
struct TransactionEnvelope
^
../../src/ripple/app/misc/Transaction.h:36:1: note: did you mean struct here?
class TransactionEnvelope;
^~~~~
struct
In file included from ../../src/ripple/unity/app_ledger_impl.cpp:32:
In file included from ../../src/ripple/app/ledger/impl/TransactionMaster.cpp:20:
../../src/ripple/app/ledger/TransactionMaster.h:28:1: error: class 'TransactionEnvelope' was previously declared as a struct; this is valid, but may result in linker errors under the Microsoft C++ ABI [-Werror,-Wmismatched-tags]
class TransactionEnvelope;
^
../../src/ripple/app/misc/Transaction.h:175:8: note: previous use is here
struct TransactionEnvelope
       ^
../../src/ripple/app/ledger/TransactionMaster.h:28:1: note: did you mean struct here?
class TransactionEnvelope;
^~~~~
struct
2 errors generated.
```

src/ripple/app/misc/Transaction.h
```diff
@@ -171,6 +172,21 @@ class Transaction
     beast::Journal  j_;
 };
 
+struct TransactionEnvelope
```
seelabs: `TransactionEnvelope` could be too many things, especially considering this is a transaction. I'd rename this.

src/ripple/app/ledger/impl/TransactionMaster.cpp
```diff
@@ -44,22 +44,24 @@ bool TransactionMaster::inLedger (uint256 const& hash, std::uint32_t ledger)
...
-    auto txn = mCache.fetch (txnID);
+    TransactionEnvelope ret;
 
-    if (!checkDisk || txn)
-        return txn;
+    ret.txn = mCache.fetch (txnID);
 
-    txn = Transaction::load (txnID, mApp);
+    if (!checkDisk || ret.txn)
+        return ret;
```
seelabs: If the transaction is in the cache we don't return the range but if it's not we do? That seems strange.

src/ripple/app/misc/Transaction.h
```diff
@@ -171,6 +172,21 @@ class Transaction
...
+{
+public:
+    TransactionEnvelope (std::shared_ptr<Transaction> txn,
+                         std::string const& completeLedgers)
+        : txn (txn)
+        , completeLedgers (completeLedgers)
```
seelabs: We're taking the txn by value, we may as well move it (i.e. `tnx(std::move(txn))` I'd do something similar for `completeLedgers` as well, take the param by value and move it.

Nit: I'd rename the constructor params so they don't match the member vars.

src/ripple/app/misc/Transaction.h
```diff
@@ -171,6 +172,21 @@ class Transaction
...
+    {}
+
+    TransactionEnvelope () = default;
```
seelabs: I probably wouldn't make this default constructable.

src/ripple/rpc/handlers/Tx.cpp
```diff
@@ -97,41 +97,47 @@ Json::Value doTx (RPC::Context& context)
     if (!isHexTxID (txid))
         return rpcError (rpcNOT_IMPL);
 
-    auto txn = context.app.getMasterTransaction ().fetch (
+    auto te = context.app.getMasterTransaction ().fetch (
```
seelabs: One issue we have is updating complete ledgers and updating the database table is not atomic. For example, in online delete, we take the lock for `complete ledgers` update it, release that lock, then take the lock for the database and update the table. We're going to have other issues when we add ledgers to that table as well.

Given that it's not atomic, let me propose two alternatives:

1) what if we saved the complete ledgers both before and after calling `fetch` and used the interesection of those two sets as our returned complete ledgers?

That should make the implementation very easy. Of course, it means a larger interval set may have been searched than what we returned. Given the threading issues with updating the database and updating `complete ledgers` it's going to take a much more extensive fix to actually return the correct ledgers if we use `complete ledgers` for this.

If we do find the transaction, I don't think we should return `complete_ledgers` since we might find a transaction outside of the `complete_ledgers` range.

2) maybe we could use the min and max ledger sequence from the transactions table. Since ledger sequence in indexed, this shouldn't be too expensive.

src/ripple/app/ledger/impl/TransactionMaster.cpp
undertome: Yeah, it seems strange initially, but the reason for this is because we don't need to report the range if the transaction is found. The client only needs the range when the transaction is not found in order to determine that the returned range of ledgers do not contain the requested transaction.

src/ripple/rpc/handlers/Tx.cpp
undertome: - So per my comments from the cpp-informal thread, can we just acquire the complete ledgers lock first?

- I agree with you about not returning `complete_ledgers` in when the transaction isn't found. Will do.

- I briefly considered using min and max as well, but my concern with that is that (based on the documentation provided for `complete_ledgers`) there may be gaps in the ranges of the completed ledgers stored which wouldn't be captured just by providing a single range based on these values.

src/ripple/app/ledger/impl/TransactionMaster.cpp
seelabs: If we only return the searched range when it's not found, what do you think about using a `std::variant` as the return type? I like that this makes the API clear: we either return the transaction or the searched range. This also solves renaming the `TransactionEnvelope` struct.

I'll admit I'm not a huge fan of `std::variant`. It's cumbersome to use. However in this case it _might_ work out OK.

undertome:
> FWIW, continuous integration is failing. Here is one of the error messages:
> 
> ```
> In file included from ../../src/ripple/app/ledger/impl/LedgerMaster.cpp:31:
> ../../src/ripple/app/misc/Transaction.h:175:1: error: 'TransactionEnvelope' defined as a struct here but previously declared as a class; this is valid, but may result in linker errors under the Microsoft C++ ABI [-Werror,-Wmismatched-tags]
> struct TransactionEnvelope
> ^
> ../../src/ripple/app/misc/Transaction.h:36:1: note: did you mean struct here?
> class TransactionEnvelope;
> ^~~~~
> struct
> In file included from ../../src/ripple/unity/app_ledger_impl.cpp:32:
> In file included from ../../src/ripple/app/ledger/impl/TransactionMaster.cpp:20:
> ../../src/ripple/app/ledger/TransactionMaster.h:28:1: error: class 'TransactionEnvelope' was previously declared as a struct; this is valid, but may result in linker errors under the Microsoft C++ ABI [-Werror,-Wmismatched-tags]
> class TransactionEnvelope;
> ^
> ../../src/ripple/app/misc/Transaction.h:175:8: note: previous use is here
> struct TransactionEnvelope
>        ^
> ../../src/ripple/app/ledger/TransactionMaster.h:28:1: note: did you mean struct here?
> class TransactionEnvelope;
> ^~~~~
> struct
> 2 errors generated.
> ```

Got rid of TransactionEnvelope.

src/ripple/app/misc/Transaction.h
undertome: Got rid of TransactionEnvelope.

src/ripple/app/ledger/impl/TransactionMaster.cpp
undertome: Reverted this file.

src/ripple/app/misc/Transaction.h
undertome: Removed TransactionEnvelope.

src/ripple/app/misc/Transaction.h
undertome: Removed.

src/ripple/rpc/handlers/Tx.cpp
undertome: Went with the first approach.

src/ripple/rpc/handlers/Tx.cpp
```diff
@@ -97,47 +96,55 @@ Json::Value doTx (RPC::Context& context)
     if (!isHexTxID (txid))
         return rpcError (rpcNOT_IMPL);
 
-    auto te = context.app.getMasterTransaction ().fetch (
+    std::array<RangeSet<std::uint32_t>, 2> ledgers;
+
+    ledgers[0] = context.app.getLedgerMaster ().getCompleteLedgersRange ();
+
+    auto txn = context.app.getMasterTransaction ().fetch (
```
seelabs: Unfortunately, this doesn't handle the case from `LedgerMaster::setFullLedger`. That adds the ledger to the history range and then queue the update to the sqlite database to the job queue. That can violate our invariant that the range computed here is a strict subset of the searched ledgers.

I thought we could address that by computing the min and max ledgers in the database, and intersecting with that as well. But that only works if `setFullLedger` is only ever called with ledgers the increase the sequence number. It looks like that assumption isn't true. I need to think about how to address this.

undertome: Ok

Issue #3105, opened 2019-10-11
Improve Ed25519 Support


mDuo13:
Ed25519 is great, and we should support it anywhere we support secp256k1. But right now, there are several places that it's possible to use secp256k1 keys and not Ed25519:

- The [`channel_authorize` method](https://xrpl.org/channel_authorize.html) can't sign claims on payment channels using Ed25519 keys. (Ripple internal issue: [RIPD-1474](https://ripplelabs.atlassian.net/browse/RIPD-1474).)
- Validators can't use Ed25519 keys as their ephemeral keys (encoded into the token)
- The [`wallet_propose` command](https://xrpl.org/wallet_propose.html)'s commandline format can't generate Ed25519 keys. I know we view wallet_propose as quasi-deprecated or likely to be deprecated soon, but I still think it would be straightforward and worthwhile to add an optional key type parameter to this command if we can.

MarkusTeufelberger:
Nodes can't really use ed25519 keys as node keys too afaik.

mDuo13:
Incidentally, #2647 adds support for Ed25519 in validator keys and node keys. Looks like it might need some updating, though...

nbougalis:
With #3116 merged, the `channel_authorize` method can now sign claims on payment channels using Ed25519 keys.

PR #3116, opened 2019-10-16
Allow `channel_authorize` to use Ed25519 keys

nbougalis:
This commit fixes a long-standing bug (internally tracked as RIPD-1474) which prevented using  *Ed25519* keys with the `channel_authorize` command.

This commit fixes the problem by allowing an optional key type when the command-line version of this function is used. Additionally, the WebSocket and JSON/RPC version of the function now accepts a `key_type` parameter, and can accept and parse the keying material as a `seed`, `seed_hex` and `passphrase`.

As always, when using server-side signing you are trusting the server to not divulge your keying material. **NEVER USE SERVER-SIDE SIGNING ON A MACHINE YOU DO NOT CONTROL**.

This commit addresses one part of #3105.

cc: @mDuo13 for documentation updates.

### DIFF
--- a/src/ripple/protocol/ErrorCodes.h
+++ b/src/ripple/protocol/ErrorCodes.h
@@ -92,7 +92,7 @@ enum error_code_i
     rpcALREADY_MULTISIG      = 36,
     rpcALREADY_SINGLE_SIG    = 37,
     // unused                  38,
-    // unused                  39,
+    rpcBAD_KEY_TYPE          = 39,
     rpcBAD_FEATURE           = 40,
     rpcBAD_ISSUER            = 41,
     rpcBAD_MARKET            = 42,
--- a/src/ripple/protocol/ErrorCodes.h
+++ b/src/ripple/protocol/ErrorCodes.h
@@ -92,7 +92,7 @@ enum error_code_i
     rpcALREADY_MULTISIG      = 36,
     rpcALREADY_SINGLE_SIG    = 37,
     // unused                  38,
-    rpcBAD_KEY_TYPE          = 39,
+    // unused                  39,
     rpcBAD_FEATURE           = 40,
     rpcBAD_ISSUER            = 41,
     rpcBAD_MARKET            = 42,
@@ -131,7 +131,8 @@ enum error_code_i
     rpcINTERNAL              = 73,  // Generic internal error.
     rpcNOT_IMPL              = 74,
     rpcNOT_SUPPORTED         = 75,
-    rpcLAST = rpcNOT_SUPPORTED   // rpcLAST should always equal the last code.
+    rpcBAD_KEY_TYPE          = 76,
+    rpcLAST = rpcBAD_KEY_TYPE      // rpcLAST should always equal the last code.
 };
 
 //------------------------------------------------------------------------------
--- a/src/ripple/rpc/handlers/PayChanClaim.cpp
+++ b/src/ripple/rpc/handlers/PayChanClaim.cpp
@@ -36,15 +36,22 @@ namespace ripple {
 
 // {
 //   secret_key: <signing_secret_key>
+//   key_type: optional; either ed25519 or secp256k1 (default to secp256k1)
 //   channel_id: 256-bit channel id
 //   drops: 64-bit uint (as string)
 // }
 Json::Value doChannelAuthorize (RPC::Context& context)
 {
     auto const& params (context.params);
-    for (auto const& p : {jss::secret, jss::channel_id, jss::amount})
+    for (auto const& p : {jss::channel_id, jss::amount})
         if (!params.isMember (p))
             return RPC::missing_field_error (p);
+
+    // Compatibility if a key type isn't specified. If it is, the
+    // keypairForSignature code will validate parameters and return
+    // the appropriate error.
+    if (!params.isMember(jss::key_type) && !params.isMember(jss::secret))
+        return RPC::missing_field_error (jss::secret);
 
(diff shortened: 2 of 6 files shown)


src/test/app/PayChan_test.cpp
```diff
@@ -1036,6 +1037,63 @@ struct PayChan_test : public beast::unit_test::suite
...
+                BEAST_EXPECT(!rv[jss::result][jss::signature_verified].asBool());
+            }
+
+            // Try to authorize using an unknown key type, except an error:
+            auto const rs1 =
+                env.rpc ("channel_authorize", "charlie", "nyx", chan, "1000");
+            BEAST_EXPECT(rs1[jss::error] == "badKeyType");
+
+            // Try to authorize using secp256k1; the authorization _should_
+            // succeed but the verification should fail:
```
seelabs: That's unfortunate. One way to fix this would be to optionally provide the public key to `channel_authorize` and either:

1) check that public key can be correctly derived from the secret key for the inferred type 
2) If the type isn't given, and the public key is, infer the key type by trying to derive the public key from the secret key for the different types.

However, since we can't require that without breaking the API, I'm not sure how helpful that really is.

src/ripple/protocol/ErrorCodes.h
```diff
@@ -92,7 +92,7 @@ enum error_code_i
     rpcALREADY_MULTISIG      = 36,
     rpcALREADY_SINGLE_SIG    = 37,
     // unused                  38,
-    // unused                  39,
+    rpcBAD_KEY_TYPE          = 39,
```
seelabs: I'm not sure that's kosher. See this comment on line 38:
```
// Please only append to this table. Do not "fill-in" gaps and do not re-use
// or repurpose error code values.
```

scottschurr: It's a question regarding the harm caused.  Presumably code `39` meant something in the past and is now unused.  If someone in the past wrote code that cared about the code `39` that part of their code has not been exercised in a while.  Now this re-use of `39` appears.  It's hard to say how their code would react.

The advice given in the comment on line 38 is the safest, most cautious, approach.

src/test/app/PayChan_test.cpp
nbougalis: I thought about/considered option (2) but it was too big a change. It would be possible to do by retrieving the channel and examining the public key, but I opted not to do that.

Generally, I’m allergic to commands like `channel_authorize` which require submitting things to the server. I’d rather deprecate it and provide a standalone tool to achieve this.

src/ripple/protocol/ErrorCodes.h
nbougalis: I tried to go back in time and look at the history; I think using 39 is safe in this instance. But I’m fine with using a “fresh” error code, if we so desire.

I’ll point out (following up on previous discussion) that the numeric value should not be canonical; it’s just an artifact, really. The string key should be what’s used by client apps.

mDuo13:
The proper "signing" methods use a predefined helper, [getSeedFromRPC](https://github.com/ripple/rippled/blob/efa926676c1b63fddb88f7686fb3e6edc26b169c/src/ripple/rpc/impl/RPCHelpers.cpp#L460-L520). Is it not feasible to reuse that code path for this method?

src/ripple/protocol/ErrorCodes.h
scottschurr: I agree that the error code values should not be canonical.  But after I did a refactor that compressed the codes (changing lots of these values) we found out that change broke at least one partner, so we backed out that change.  Hyrum's law for the win: https://www.hyrumslaw.com

Or, in other words, the only way to get clients to not use a value is to not make it visible.  Anywhere.  Or any time.  There are no take-backs.

src/ripple/rpc/handlers/PayChanClaim.cpp
```diff
@@ -36,16 +36,23 @@ namespace ripple {
...
-    for (auto const& p : {jss::secret, jss::channel_id, jss::amount})
+    for (auto const& p : {jss::channel_id, jss::amount})
         if (!params.isMember (p))
             return RPC::missing_field_error (p);
 
+    // Compatibility if a key type isn't specified. If it is, the
+    // keypairForSignature code will validate parameters and return
+    // the appropriate error.
+    if (!params.isMember(jss::key_type) && !params.isMember(jss::secret))
+        return RPC::missing_field_error (jss::secret);
```
scottschurr: FWIW, codecov indicates that the new unit tests in PayChan_test.cpp don't exercise this line or line 48 for that matter.  Also there are other lines in the vicinity that are not exercised by any of the unit tests.

If you would like, you can get coverage for all the lines in this method by cherry-picking this commit: https://github.com/scottschurr/rippled/commit/d9b2cc91ea1016f444fbb0d2f3896dfc39c8c396.  If you do choose to cherry-pick it, feel free to adapt it as you see fit.

nbougalis:
> The proper "signing" methods use a predefined helper, [getSeedFromRPC](https://github.com/ripple/rippled/blob/efa926676c1b63fddb88f7686fb3e6edc26b169c/src/ripple/rpc/impl/RPCHelpers.cpp#L460-L520). Is it not feasible to reuse that code path for this method?

We do, indirectly: we call ` keypairForSignature` which calls ` getSeedFromRPC`. The problem is that `getSeedFromRPC` doesn't respect the `secret` parameter, which we need for backwards compatibility.

Issue #3106, opened 2019-10-11
Unit Test failure when built with protobuf 3.10


mellery451:
observed on macos with homebrew managed protobuf. Current pb version is 3.10 and we get a single unit test failure here:

 https://github.com/ripple/rippled/blob/develop/src/test/protocol/STTx_test.cpp#L1431

Might be related to the kind of exception from protobuf or some other change. Needs investigation so that users on latest stable protobuf won't break.

PR #3119, opened 2019-10-17
Relax STTx test failure criterion:

mellery451:
FIXES: #3106

Different versions of protobuf produce subtly different
results when given invalid message payloads. This leads to
subtly different behavior when we try to deserialize these
invalid messages. As such, we can't tie success to a
particular exception.

### DIFF
--- a/src/test/protocol/STTx_test.cpp
+++ b/src/test/protocol/STTx_test.cpp
@@ -1419,6 +1419,17 @@ public:
         try
         {
             protocol::TMTransaction tx2;
+            // In all versions of protobuf, the parsing of this payload fails.
+            // However, different versions of protobuf will leave the parsed
+            // message (tx2) in different states. Versions 3.10+ seem to clear
+            // the message, whereas earlier versions would leave the message
+            // in a partial state. The varying nature of the failed-parsed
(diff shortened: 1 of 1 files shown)


src/test/protocol/STTx_test.cpp
```diff
@@ -1419,16 +1419,16 @@ class STTx_test : public beast::unit_test::suite
         try
         {
             protocol::TMTransaction tx2;
-            tx2.ParseFromArray(payload1, sizeof(payload1));
+            BEAST_EXPECT(! tx2.ParseFromArray(payload1, sizeof(payload1)));
```
scottschurr: Might be nice to add a comment along the line that, "Depending on the version of protobuf either an exception is thrown of a zero-length transaction is returned.  In either case an exception is thrown, either here or by the `STTx` constructor below."  Or if that's not accurate, then whatever the truth is...

mellery451: 👍 - let me know if the added comments make sense.

scottschurr: Excellent.  Nicely done. Thanks.  :+1:

Issue #2851, opened 2019-02-12
Augment "submit" command response


mDuo13:
When you submit a transaction, there's a bunch of information you'd like to know so you can track your transaction's status accurately. You can look up some of that info separately, but you have to be careful about race conditions if you don't do everything in precisely the correct order. We can simplify a lot of the challenges by providing useful information directly in the response to the `submit` command.

Here are some suggestions (with credit to @ximinez for suggesting most of these):
- `applied`: boolean. Indicates whether the transaction was applied to the open ledger
- `queued`: boolean. Indicates whether the transaction was put into the transaction queue
- `broadcast`: boolean. Indicates whether the transaction was broadcast to the network
- `kept` (name pending): One or more fields to indicate that a transaction was kept by the server and may be retried later. (Aside from the queue, there are two "holding pens", localtxns and heldtxns, which are used in cases of a ledger jump or normal ledger advancement respectively. But I'm not sure distinguishing between the two is relevant to end-users.)
- `available_sequence`: number. (Also not sure about this name.) Regardless of the disposition of the transaction, returns the sequence number that the account that owns the just-submitted transaction should use on its next transaction to have the largest chance of success. In short, it'll look at the account sequence number and the transaction queue, and look for the first "available" sequence. If the Tickets amendment is enabled, this should adjust for that.
- `account_sequence`: number. The sequence number of transaction's sending account, as of the "current" ledger.
- `open_ledger_reference_drops`: number. (Not sure if this is a good idea or a good name.) Number of drops required for a reference transaction to get into the open ledger.
- `validated_ledger_index`: number. The ledger index of the latest validated ledger. The latest validated ledger at the time of submission. The earliest ledger index that the submitted transaction could appear in¹ is this value +1, so you can use this to easily bound your tx queries to find out the final status of this transaction. Returning the value at this point saves a separate request and avoids possible race conditions. (Example race condition: You request the latest ledger index with the "ledger" command, then you submit, but the submission returns first, then the ledger index advances some number of times, then your "ledger" command returns with a higher ledger index than the earliest one your transaction could conceivably be in. If the delays are big enough it's possible this could cause you to miss your transaction.) I specifically say validated ledger index because getting the latest current or closed ledger is not a guarantee—your transaction could (theoretically) make it into an earlier ledger index that also hasn't been validated yet.

[...]

PR #3125, opened 2019-10-20
[#2851] added more RPC fields to reply of Submit command

p2peer:
Here is an example of new reply fields of Submit command:

"result" : {
      "accepted" : true,
      "account_sequence_available" : 11,
      "account_sequence_next" : 11,
      "applied" : true,
      "broadcast" : true,
      "kept" : true,
      "open_ledger_cost" : 10,
      "queued" : false,
      "validated_ledger_index" : 1388096
   }

### DIFF
--- a/src/ripple/app/misc/NetworkOPs.cpp
+++ b/src/ripple/app/misc/NetworkOPs.cpp
@@ -1062,7 +1062,7 @@ void NetworkOPsImp::apply (std::unique_lock<std::mutex>& batchLock)
         auto newOL = app_.openLedger().current();
         for (TransactionStatus& e : transactions)
         {
-            e.transaction->clearAccepted();
+            e.transaction->clearSubmitResult();
 
             if (e.applied)
             {
--- a/src/ripple/app/misc/NetworkOPs.cpp
+++ b/src/ripple/app/misc/NetworkOPs.cpp
@@ -937,6 +937,11 @@ void NetworkOPsImp::processTransaction (std::shared_ptr<Transaction>& transactio
         doTransactionSync (transaction, bUnlimited, failType);
     else
         doTransactionAsync (transaction, bUnlimited, failType);
+
+    auto const lpClosed = m_ledgerMaster.getValidatedLedger ();
+
+    if (lpClosed)
+        transaction->setValidatedLedgerIndex(lpClosed->info().seq);
 }
 
 void NetworkOPsImp::doTransactionAsync (std::shared_ptr<Transaction> transaction,
@@ -1058,6 +1063,8 @@ void NetworkOPsImp::apply (std::unique_lock<std::mutex>& batchLock)
         auto newOL = app_.openLedger().current();
         for (TransactionStatus& e : transactions)
         {
+            e.transaction->clearAccepted();
+
             if (e.applied)
             {
                 pubProposedTransaction (newOL,
@@ -1102,6 +1109,7 @@ void NetworkOPsImp::apply (std::unique_lock<std::mutex>& batchLock)
                         t, false, false, FailHard::no);
                     t->setApplying();
                 }
+                e.transaction->setApplied();
             }
             else if (e.result == tefPAST_SEQ)
             {
@@ -1118,6 +1126,8 @@ void NetworkOPsImp::apply (std::unique_lock<std::mutex>& batchLock)
                 // kicked out of the queue, and this will try to
                 // put it back.
                 m_ledgerMaster.addHeldTransaction(e.transaction);
+                e.transaction->setQueued();
+                e.transaction->setKept();
             }
             else if (isTerRetry (e.result))
             {
@@ -1132,6 +1142,7 @@ void NetworkOPsImp::apply (std::unique_lock<std::mutex>& batchLock)
                         << "Transaction should be held: " << e.result;
                     e.transaction->setStatus (HELD);
                     m_ledgerMaster.addHeldTransaction (e.transaction);
+                    e.transaction->setKept();
                 }
             }
             else
@@ -1146,6 +1157,7 @@ void NetworkOPsImp::apply (std::unique_lock<std::mutex>& batchLock)
                 m_localTX->push_back (
                     m_ledgerMaster.getCurrentLedgerIndex(),
                     e.transaction->getSTransaction());
+                e.transaction->setKept();
             }
 
             if (e.applied || ((mMode != OperatingMode::FULL) &&
@@ -1169,8 +1181,17 @@ void NetworkOPsImp::apply (std::unique_lock<std::mutex>& batchLock)
                     app_.overlay().foreach (send_if_not (
                         std::make_shared<Message> (tx, protocol::mtTRANSACTION),
                         peer_in_set(*toSkip)));
+                    e.transaction->setBroadcast();
                 }
             }
+
+            std::uint32_t accountSeq;
+            std::uint32_t availableSeq;
+            XRPAmount fee = app_.getTxQ().getTxRequiredFeeAndSeq(
+                *newOL, e.transaction->getSTransaction(),
+                accountSeq, availableSeq);
+            e.transaction->setRequiredFeeAndSeq(
+                fee, accountSeq, availableSeq);
         }
     }
 
--- a/src/ripple/app/misc/NetworkOPs.cpp
+++ b/src/ripple/app/misc/NetworkOPs.cpp
@@ -937,11 +937,6 @@ void NetworkOPsImp::processTransaction (std::shared_ptr<Transaction>& transactio
         doTransactionSync (transaction, bUnlimited, failType);
     else
         doTransactionAsync (transaction, bUnlimited, failType);
-
-    auto const lpClosed = m_ledgerMaster.getValidatedLedger ();
-
-    if (lpClosed)
-        transaction->setValidatedLedgerIndex(lpClosed->info().seq);
 }
 
 void NetworkOPsImp::doTransactionAsync (std::shared_ptr<Transaction> transaction,
@@ -1059,6 +1054,10 @@ void NetworkOPsImp::apply (std::unique_lock<std::mutex>& batchLock)
         }
         if (changed)
             reportFeeChange();
+
+        boost::optional<LedgerIndex> validatedLedgerIndex;
+        if (auto const l = m_ledgerMaster.getValidatedLedger())
+            validatedLedgerIndex = l->info().seq;
 
         auto newOL = app_.openLedger().current();
         for (TransactionStatus& e : transactions)
@@ -1185,13 +1184,14 @@ void NetworkOPsImp::apply (std::unique_lock<std::mutex>& batchLock)
                 }
             }
 
-            std::uint32_t accountSeq;
-            std::uint32_t availableSeq;
-            XRPAmount fee = app_.getTxQ().getTxRequiredFeeAndSeq(
-                *newOL, e.transaction->getSTransaction(),
-                accountSeq, availableSeq);
-            e.transaction->setRequiredFeeAndSeq(
-                fee, accountSeq, availableSeq);
+            if (validatedLedgerIndex)
+            {
+                auto [fee, accountSeq, availableSeq] =
+                    app_.getTxQ().getTxRequiredFeeAndSeq(
+                        *newOL, e.transaction->getSTransaction());
+                e.transaction->setCurrentLedgerState(
+                    *validatedLedgerIndex, fee, accountSeq, availableSeq);
+            }
         }
     }
 
--- a/src/ripple/app/misc/Transaction.h
+++ b/src/ripple/app/misc/Transaction.h
@@ -154,6 +154,118 @@ public:
         mApplying = false;
     }
 
+    /**
+     * @brief Detect if transaction was accepted
+     * @param applied ref to store if transaction was appiled
+     * @param broadcast ref to store if transaction was appiled
+     * @param queued ref to store if transaction was appiled
+     * @param kept ref to store if transaction was appiled
+     * @return true if (applied|broadcast|queued|kept)
+     */
+    bool getAccepted(bool &applied, bool &broadcast,
+        bool &queued, bool &kept) const
+    {
+        applied = mApplied;
+        broadcast = mBroadcast;
+        queued = mQueued;
+        kept = mKept;
+        return mApplied || mBroadcast || mQueued || mKept;
+    }
+
+    /**
+     * Set this flag once was applied to open ledger.
+     */
+    void setApplied() {
+        mApplied = true;
+    }
+
+    /**
+     * Set this flag once was put into heldtxns queue.
+     */
+    void setQueued() {
+        mQueued = true;
+    }
+
+    /**
+     * Set this flag once was broadcasted via network.
+     */
+    void setBroadcast() {
+        mBroadcast = true;
+    }
+
+    /**
+     * Set this flag once was put to localtxns queue.
+     */
+    void setKept() {
+        mKept = true;
+    }
+
+    /**
+     * @brief clear all accepted-related flags
+     */
+    void clearAccepted() {
+        mApplied = false;
+        mBroadcast = false;
+        mQueued = false;
+        mKept = false;
+    }
+
+    /**
+     * Get last account sequence in current ledger and first available.
+     *
+     * @param available ref to save first available account sequence
+     * @return last account sequence in current ledger or early.
+     */
+    std::uint32_t getAccountSequence(std::uint32_t &available) const
+    {
+        available = mAccountSeqAvail;
+        return mAccountSeqNext;
+    }
+
+    /**
+     * Returns minimum fee required at current moment.
+     *
+     * @return minimum fee measured in drops.
+     */
+    XRPAmount getMinimumFeeRequired() const
+    {
+        return mMinFeeRequired;
+    }
+
+    /**
+     * @brief Sets minimum required fee and sequences for the transaction
+     * @param fee minimum fee required for the transaction
+     * @param accountSeq first valid account sequence in current ledger
+     * @param availableSeq first available sequence for the transaction
+     */
+    void setRequiredFeeAndSeq(XRPAmount fee,
+        std::uint32_t accountSeq, std::uint32_t availableSeq)
+    {
+        mMinFeeRequired = fee;
+        mAccountSeqNext = accountSeq;
+        mAccountSeqAvail = availableSeq;
+    }
+
+    /**
+     * Sets last known index of validated ledger.
+     *
+     * @param index number of last validated ledger
+     */
+    void setValidatedLedgerIndex(LedgerIndex index)
+    {
+        mValidatedLedger = index;
+    }
+
+    /**
+     * Returns last known index of validated ledger.
+     *
+     * @return last index of validated ledger.
+     */
+    LedgerIndex getValidatedLedgerIndex() const
+    {
+        return mValidatedLedger;
+    }
+
     Json::Value getJson (JsonOptions options, bool binary = false) const;
 
     static Transaction::pointer load (uint256 const& id, Application& app);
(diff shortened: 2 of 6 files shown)


nbougalis:
Thanks for your submission. It's being reviewed.

src/ripple/app/misc/NetworkOPs.cpp
```diff
@@ -937,6 +937,11 @@ void NetworkOPsImp::processTransaction (std::shared_ptr<Transaction>& transactio
         doTransactionSync (transaction, bUnlimited, failType);
     else
         doTransactionAsync (transaction, bUnlimited, failType);
+
+    auto const lpClosed = m_ledgerMaster.getValidatedLedger ();
+
+    if (lpClosed)
+        transaction->setValidatedLedgerIndex(lpClosed->info().seq);
```
miguelportilla: Alternate suggestion:
```
if (auto const valLedger {m_ledgerMaster.getValidatedLedger()}; valLedger)
    transaction->setValidatedLedgerIndex(valLedger->info().seq);```

File: src/ripple/app/misc/impl/TxQ.cpp [hunk 2]
```diff
@@ -1393,6 +1393,58 @@ TxQ::getMetrics(OpenView const& view) const
...
+
+    if (sle)
+        accountSeq = (*sle)[sfSequence];
+    else
+        accountSeq = 0;
+
+    availableSeq = accountSeq;
+
+    auto accountIter = byAccount_.find(account);
+    if (accountIter != byAccount_.end())
```
miguelportilla: Alternate suggestion:
```
if (auto iter {byAccount_.find(account)}; iter != byAccount_.end())
```

src/ripple/app/misc/NetworkOPs.cpp
seelabs: Or even just
```
    if (auto const l = m_ledgerMaster.getValidatedLedger())
        transaction->setValidatedLedgerIndex(l->info().seq);
```
I.e. there's no need for the `; valLedger)` part.

src/ripple/app/misc/NetworkOPs.cpp
```diff
@@ -1169,8 +1181,17 @@ void NetworkOPsImp::apply (std::unique_lock<std::mutex>& batchLock)
...
                         peer_in_set(*toSkip)));
+                    e.transaction->setBroadcast();
                 }
             }
+
+            std::uint32_t accountSeq;
+            std::uint32_t availableSeq;
+            XRPAmount fee = app_.getTxQ().getTxRequiredFeeAndSeq(
+                *newOL, e.transaction->getSTransaction(),
+                accountSeq, availableSeq);
```
seelabs: I'd get rid of the return value combined with the out parameters and return a struct with all the value you want. You can then use the struct or use a structured binding here. Something like:
```
            auto [fee, accountSeq, availableSeq] =
                app_.getTxQ().getTxRequiredFeeAndSeq(
                    *newOL, e.transaction->getSTransaction());
```

src/ripple/app/misc/TxQ.h
```diff
@@ -319,6 +319,25 @@ class TxQ
...
+     * @param useFeeIncrease increase minimum required fee in the case
+     *        of multi tx and in the case of replace existing tx
+     * @return minimum required fee
+     */
+    XRPAmount
+    getTxRequiredFeeAndSeq(OpenView const& view,
+        std::shared_ptr<STTx const> const& tx,
+        std::uint32_t &accountSeq,
+        std::uint32_t &availableSeq,
+        bool useFeeIncrease = false) const;
```
seelabs: I'd get rid of the return value combined with the out parameters and just return a struct with the values you want to return. (Also, style nit: we usually write `uint32_t&` instead of `uint32_t &`)

src/ripple/rpc/handlers/Submit.cpp
```diff
@@ -142,12 +142,41 @@ Json::Value doSubmit (RPC::Context& context)
         {
             std::string sToken;
             std::string sHuman;
+            std::uint32_t mAccountSeqNext;
+            std::uint32_t mAccountSeqAvail;
+            bool mAccepted;
+            bool mApplied;
+            bool mBroadcast;
+            bool mQueued;
+            bool mKept;
```
seelabs: Style nit: these aren't member variables, so they shouldn't begin with `m`.

src/ripple/rpc/handlers/Submit.cpp
```diff
@@ -142,12 +142,41 @@ Json::Value doSubmit (RPC::Context& context)
...
 
             transResultInfo (tpTrans->getResult (), sToken, sHuman);
 
             jvResult[jss::engine_result]           = sToken;
             jvResult[jss::engine_result_code]      = tpTrans->getResult ();
             jvResult[jss::engine_result_message]   = sHuman;
+
+            mAccepted = tpTrans->getAccepted (mApplied, mBroadcast,
+                    mQueued, mKept);
```
seelabs: I'd get rid of the combined return value/out parameters for `getAccepted` and just return a struct with the values you want. You can add a function to the struct to return true if any of the values are set. Something like:
```
            auto const accepted = tpTrans->getAccepted();

            jvResult[jss::accepted] = accepted.any();
            jvResult[jss::applied] = accepted.applied;
            jvResult[jss::broadcast] = accepted.broadcast;
            jvResult[jss::queued] = accepted.queued;
            jvResult[jss::kept] = accepted.kept;
```

src/ripple/app/misc/Transaction.h
```diff
@@ -166,6 +278,18 @@ class Transaction
...
+    bool            mApplied = false;
+    bool            mQueued = false;
+    bool            mBroadcast = false;
+    bool            mKept = false;
+
+    /** tips on current state of the ledger */
+    LedgerIndex     mValidatedLedger = 0;
+    XRPAmount       mMinFeeRequired = 0;
+    std::uint32_t   mAccountSeqNext = 0;
+    std::uint32_t   mAccountSeqAvail = 0;
```
seelabs: Rather than using `0` as a `not set` value, consider using `boost::optional`

src/ripple/app/misc/Transaction.h
```diff
@@ -166,6 +278,18 @@ class Transaction
     TER             mResult = temUNCERTAIN;
     bool            mApplying = false;
 
+    /** different ways for transaction to be accepted */
+    bool            mApplied = false;
+    bool            mQueued = false;
+    bool            mBroadcast = false;
+    bool            mKept = false;
```
seelabs: Some of these values look mutually exclusive. Does it make sense to make an enum to represent the valid states instead of using the four bools here?

We should also look to see if some of these states are implied with the TER result. I.e, do we need a `mQueued` flag when we already have a `terQUEUED` `mResult`?

src/ripple/app/misc/NetworkOPs.cpp
seelabs: I'd like to package up all the the auxiliary ledger info and either set all of it or none of it. To support that, what do you think of setting the validated ledger index here. Something like:
```

            if (validatedLedgerIndex)
            {
                auto [fee, accountSeq, availableSeq] =
                    app_.getTxQ().getTxRequiredFeeAndSeq(
                        *newOL, e.transaction->getSTransaction());
                e.transaction->setCurrentLedgerState(
                    *validatedLedgerIndex, fee, accountSeq, availableSeq);
            }
```
Where `validatedLedgerIndex` is a `boost::optional` set outside the loop.

src/ripple/app/misc/NetworkOPs.cpp
seelabs: Can we set this in `apply` instead? That way all the auxiliary ledger information is set in the same place.

src/ripple/app/misc/Transaction.h
```diff
@@ -154,6 +154,118 @@ class Transaction
...
+    /**
+     * Returns last known index of validated ledger.
+     *
+     * @return last index of validated ledger.
+     */
+    LedgerIndex getValidatedLedgerIndex() const
+    {
+        return mValidatedLedger;
+    }
+
```
seelabs: I'd remove all the functions that set and get the auxiliary ledger information and replace them with two functions:
```
    struct CurrentLedgerState
    {
        CurrentLedgerState() = delete;

        CurrentLedgerState(
            LedgerIndex li,
            XRPAmount fee,
            std::uint32_t accSeqNext,
            std::uint32_t accSeqAvail)
            : validatedLedger{li}
            , minFeeRequired{fee}
            , accountSeqNext{accSeqNext}
            , accountSeqAvail{accSeqAvail}
        {
        }

        LedgerIndex validatedLedger;
        XRPAmount minFeeRequired;
        std::uint32_t accountSeqNext;
        std::uint32_t accountSeqAvail;
    };

    boost::optional<CurrentLedgerState>
    getCurrentLedgerState() const
    {
        return mCurrentLedgetState;
    }

    void
    setCurrentLedgerState(
        LedgerIndex validatedLedger,
        XRPAmount fee,
        std::uint32_t accountSeq,
        std::uint32_t availableSeq)
    {
        mCurrentLedgetState.emplace(
            validatedLedger, fee, accountSeq, availableSeq);
    }
```
Where `mCurrentLedgerState` is `boost::optional<CurrentLedgerState> mCurrentLedgetState;`

src/ripple/app/misc/Transaction.h
seelabs: What do you think about replacing these four member vars with a `boost::optional<CurrentLedgerState> mCurrentLedgetState;` where `CurrentLedgerState` just packages up these vars?

src/ripple/rpc/handlers/Submit.cpp
```diff
@@ -142,12 +142,41 @@ Json::Value doSubmit (RPC::Context& context)
...
+
+            jvResult[jss::account_sequence_next]
+                    = (Json::Value::UInt)mAccountSeqNext;
+            jvResult[jss::account_sequence_available]
+                    = (Json::Value::UInt)mAccountSeqAvail;
+
+            jvResult[jss::open_ledger_cost] = (Json::Value::UInt)
+                    tpTrans->getMinimumFeeRequired ().drops ();
+            jvResult[jss::validated_ledger_index] = (Json::Value::UInt)
+                    tpTrans->getValidatedLedgerIndex ();
```
seelabs: What do you think about something like:
```
            if (auto currentLedgerState = tpTrans->getCurrentLedgerState())
            {
                jvResult[jss::account_sequence_next]
                        = (Json::Value::UInt)currentLedgerState->accountSeqNext;
                jvResult[jss::account_sequence_available]
                        = (Json::Value::UInt)currentLedgerState->accountSeqAvail;

                jvResult[jss::open_ledger_cost] = (Json::Value::UInt)
                        currentLedgerState->minFeeRequired.drops ();
                jvResult[jss::validated_ledger_index] = (Json::Value::UInt)
                        currentLedgerState->validatedLedger;
            }
```
Where `currentLedgerState` is a `boost::optional`

src/ripple/app/misc/NetworkOPs.cpp
p2peer: Done, simplified, moved to apply.

src/ripple/app/misc/impl/TxQ.cpp
p2peer: Done.

src/ripple/app/misc/NetworkOPs.cpp
p2peer: Done.

src/ripple/app/misc/TxQ.h
p2peer: Done.

src/ripple/rpc/handlers/Submit.cpp
p2peer: Done.

src/ripple/rpc/handlers/Submit.cpp
p2peer: Done.

src/ripple/app/misc/Transaction.h
p2peer: Done.

src/ripple/app/misc/Transaction.h
p2peer: I think that per-bit representation of these flags is better than as enum, because these flags come to us as bits and go to output as the same bits. In the case of mutually exclusive enum values we need to code input bits into enum values, then to decode it from enum values to bits. This will require more computational steps. The queued flag is preserved because it is mentioned in the original task for this ticket.

src/ripple/app/misc/Transaction.h
p2peer: Done.

src/ripple/rpc/handlers/Submit.cpp
p2peer: Done.

src/ripple/app/misc/Transaction.h
```diff
@@ -166,6 +301,12 @@ class Transaction
     TER             mResult = temUNCERTAIN;
     bool            mApplying = false;
 
+    /** different ways for transaction to be accepted */
+    AcceptedState   mAcceptedState;
+
+    /** current ledger state */
+    boost::optional<CurrentLedgerState> mCurrentLedgetState;
```
ximinez: The `m` prefix is a deprecated standard for indicating member variables. The new hotness standard is a `_` suffix. So instead of `mAcceptedState`, use `acceptedState_`, etc.

src/ripple/app/misc/Transaction.h
```diff
@@ -154,6 +154,141 @@ class Transaction
...
 
+    struct AcceptedState
+    {
+        AcceptedState()
+            : applied{false}
+            , broadcast{false}
+            , queued{false}
+            , kept{false}
+        {
+        }
```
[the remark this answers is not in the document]
p2peer: Done

src/ripple/app/misc/Transaction.h
```diff
@@ -154,6 +154,141 @@ class Transaction
         mApplying = false;
     }
 
+    struct AcceptedState
```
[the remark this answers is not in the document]
p2peer: Done

src/ripple/app/misc/Transaction.h
```diff
@@ -154,6 +154,141 @@ class Transaction
...
+     */
+    void setKept()
+    {
+        mAcceptedState.kept = true;
+    }
+
+    /**
+     * @brief Clear all flags in mAcceptedState
+     */
+    void clearAccepted()
```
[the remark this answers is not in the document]
p2peer: Done

src/ripple/app/misc/Transaction.h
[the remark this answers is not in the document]
p2peer: Done

src/ripple/app/misc/TxQ.h
```diff
@@ -319,6 +319,23 @@ class TxQ
...
     getMetrics(OpenView const& view) const;
 
+    /**
+     * @brief Returns minimum required fee for tx and two sequences:
+     *        first vaild sequence for this account in current ledger
+     *        and first available sequence for transaction
+     * @param view current open ledger
+     * @param tx the transaction
+     * @param accountSeq reference to save first sequence in the ledger
+     * @param availableSeq reference to save firts available sequence
```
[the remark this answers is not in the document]
p2peer: Done

src/ripple/app/misc/TxQ.h
```diff
@@ -319,6 +319,23 @@ class TxQ
...
+     * @param useFeeIncrease increase minimum required fee in the case
+     *        of multi tx and in the case of replace existing tx
+     * @return minimum required fee
+     */
+    std::tuple<XRPAmount, std::uint32_t, std::uint32_t>
+    getTxRequiredFeeAndSeq(OpenView const& view,
+        std::shared_ptr<STTx const> const& tx,
+        bool useFeeIncrease = false) const;
```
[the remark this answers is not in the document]
p2peer: Done

src/ripple/rpc/handlers/Submit.cpp
```diff
@@ -148,6 +148,27 @@ Json::Value doSubmit (RPC::Context& context)
...
+            {
+                jvResult[jss::account_sequence_next]
+                        = (Json::Value::UInt)currentLedgerState->accountSeqNext;
+                jvResult[jss::account_sequence_available]
+                        = (Json::Value::UInt)currentLedgerState->accountSeqAvail;
+
+                jvResult[jss::open_ledger_cost] = (Json::Value::UInt)
+                        currentLedgerState->minFeeRequired.drops ();
+                jvResult[jss::validated_ledger_index] = (Json::Value::UInt)
+                        currentLedgerState->validatedLedger;
```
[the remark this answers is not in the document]
p2peer: Done

src/ripple/app/misc/Transaction.h
p2peer: Done

src/ripple/app/misc/TxQ.h
```diff
@@ -319,6 +319,19 @@ class TxQ
...
+    /**
+     * @brief Returns minimum required fee for tx and two sequences:
+     *        first vaild sequence for this account in current ledger
+     *        and first available sequence for transaction
+     * @param view current open ledger
+     * @param tx the transaction
+     * @return minimum required fee, first sequence in the ledger
+     *        and first available sequence
+     */
+    std::tuple<XRPAmount, std::uint32_t, std::uint32_t>
```
seelabs: I'd return a struct here, not a tuple. This tuple has type fields of the same type, both for sequences. I'd rather use a struct to help keep these fields straight.

src/ripple/app/misc/impl/TxQ.cpp
```diff
@@ -1393,6 +1393,41 @@ TxQ::getMetrics(OpenView const& view) const
...
+    auto const baseFee = calculateBaseFee(view, *tx);
+    auto fee = FeeMetrics::scaleFeeLevel(snapshot, view);
+
+    auto const sle = view.read(keylet::account(account));
+
+    std::uint32_t accountSeq;
+    if (sle)
+        accountSeq = (*sle)[sfSequence];
+    else
+        accountSeq = 0;
```
seelabs: We could replace this with something like:
```
    auto const accountSeq = [&view, &account]() -> std::uint32_t {
        auto const sle = view.read(keylet::account(account));
        if (sle)
            return (*sle)[sfSequence];
        return 0;
    }();
```
That restricts the scope of the `sle` and lets us make `accountSeq` `const`

src/ripple/app/misc/impl/TxQ.cpp
```diff
@@ -1393,6 +1393,41 @@ TxQ::getMetrics(OpenView const& view) const
...
+
+    auto availableSeq = accountSeq;
+
+    if (auto accIter {byAccount_.find(account)}; accIter != byAccount_.end())
+    {
+        auto& txQAcct = accIter->second;
+        for (auto iter : txQAcct.transactions)
```
seelabs: The variable is named `iter`, but it is a `std::pair` not a type of iterator. Also, `iter` is taken by value, this makes an unneeded copy (the first element of the pair is cheap to copy, the second is not).

However, I think the clearest change is to use structured binding to just get the first element. Something like:
```
        for (auto const& [seq, _] : txQAcct.transactions)
        {
            if (seq >= availableSeq)
                availableSeq = seq + 1;
        }
```

src/ripple/app/misc/Transaction.h
```diff
@@ -166,6 +293,12 @@ class Transaction
     TER             mResult = temUNCERTAIN;
     bool            mApplying = false;
 
+    /** different ways for transaction to be accepted */
+    SubmitResult    submitResult_;
+
+    /** current ledger state */
+    boost::optional<CurrentLedgerState> currentLedgerState_;
```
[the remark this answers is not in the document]
p2peer: Done

src/ripple/app/misc/TxQ.h
p2peer: Done

src/ripple/app/misc/impl/TxQ.cpp
p2peer: Done

src/ripple/app/misc/impl/TxQ.cpp
p2peer: Done

p2peer:
Changes after review part 3 applied.

src/ripple/app/misc/impl/TxQ.cpp
```diff
@@ -1419,6 +1419,7 @@ TxQ::getTxRequiredFeeAndSeq(OpenView const& view,
         auto& txQAcct = iter->second;
         for (auto const& [seq, _] : txQAcct.transactions)
         {
+            (void)_;
```
seelabs: It's a shame we can't use `std::ignore` in structured bindings.

PR #3139, opened 2019-10-30
Use pkgconfig to search for system deps before vendoring

movitto:
Continuing the effort started in #3044, this PR consists of a series of patches, each utilizing pkgconfig to search for system libraries to satisfy dependencies before falling back to the current default behaviour (downloading / checking out the dependency).

This PR contains patches for the following deps:
- soci
- sqlite3
- snappy
- secp256k1
- libarchive
- lz4

The following deps still need to be addressed after this PR:
- ed25519-donna
- rocksdb
- nudb

These patches (or similar) are needed for environments where vendored dependencies are prohibited.


Builds/CMake/deps/Secp256k1.cmake
```diff
@@ -2,34 +2,58 @@
...
+  add_library (NIH::secp256k1 ALIAS secp256k1_lib)
+
+else()
+  set(INSTALL_SECP256K1 true)
+
+  add_library (secp256k1 STATIC
+    src/secp256k1/src/secp256k1.c)
+  target_compile_definitions (secp256k1
+    PRIVATE
+      USE_NUM_NONE
```
mellery451: For libraries like this that have a number of preprocessor options/variants, I wonder how the installed library might differ. I suspect most of these are just speed optimizations, but it brings up the question of how the prebuilt/installed library could differ functionally from the one we build. Same situation for sqlite where there is the `SQLITE_THREADSAFE` option. I assume it impacts the binary output (not just headers) and it's not obvious to me which option a prebuilt library might choose for that setting.  I'm not too worried about all this, but it is something to keep in mind if we think any of those build options matter significantly to our application.

movitto: @mellery451 thanks for the review and agree. Many Linux distros include default flags for enhanced security, platform comparability, etc. Furthermore builds for different architectures (x86, ARM, etc) will often entail different compiler flags.

Unfortunately there is no way to know the impact until we actually try to compile (and use) the binaries on those platforms.

PR #3143, opened 2019-11-02
Change linear whitespace handling:

nbougalis:
When splitting a string, the existing code would compress linear whitespace away completely; this seems to not be intention of [RFC2616](https://www.w3.org/Protocols/rfc2616/rfc2616-sec2.html#sec2.2), which allows replacing any amount of linear whitespace with a single space.

Requesting an additional signoff from @vinniefalco.

### DIFF
--- a/src/ripple/beast/rfc2616.h
+++ b/src/ripple/beast/rfc2616.h
@@ -247,7 +247,7 @@ split(FwdIt first, FwdIt last, Char delim)
         }
         else if (*iter == delim)
         {
-            e = trim_right (e);
+            e = trim (e);
             if (! e.empty())
             {
                 result.emplace_back(std::move(e));
@@ -257,7 +257,10 @@ split(FwdIt first, FwdIt last, Char delim)
         }
         else if (is_lws (*iter))
         {
-            ++iter;
+            e.append (1, ' ');
+            do
+                iter++;
+            while (iter != last && is_lws (*iter));
         }
         else
         {
@@ -267,7 +270,7 @@ split(FwdIt first, FwdIt last, Char delim)
 
     if (! e.empty())
     {
-        e = trim_right (e);
+        e = trim (e);
         if (! e.empty())
             result.emplace_back(std::move(e));
     }
--- a/src/test/beast/beast_rfc2616_test.cpp
+++ b/src/test/beast/beast_rfc2616_test.cpp
@@ -1,0 +1,60 @@
+//------------------------------------------------------------------------------
+/*
+This file is part of rippled: https://github.com/ripple/rippled
+Copyright (c) 2012, 2013 Ripple Labs Inc.
+
+Permission to use, copy, modify, and/or distribute this software for any
+purpose  with  or without fee is hereby granted, provided that the above
+copyright notice and this permission notice appear in all copies.
+
+THE  SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+WITH  REGARD  TO  THIS  SOFTWARE  INCLUDING  ALL  IMPLIED  WARRANTIES  OF
+MERCHANTABILITY  AND  FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+ANY  SPECIAL ,  DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+WHATSOEVER  RESULTING  FROM  LOSS  OF USE, DATA OR PROFITS, WHETHER IN AN
+ACTION  OF  CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+
+//==============================================================================
+
+#include <ripple/basics/random.h>
+#include <ripple/beast/unit_test.h>
+#include <ripple/beast/rfc2616.h>
+#include <ripple/beast/xor_shift_engine.h>
+
+namespace beast {
+
+class rfc2616_test : public unit_test::suite
+{
+public:
+    void run() override
+    {
+        testcase("LWS compression & trimming during parsing");
+
+        beast::xor_shift_engine rng(0x243F6A8885A308D3);
+
+        std::vector<std::string> words;
+
+        for (int i = 0; i != 64; ++i)
+            words.push_back("X-" + std::to_string(ripple::rand_int(rng, 100, 1000)));
+
+        std::string question;
+
+        for (auto w : words)
+        {
+            if (!question.empty())
+                question += ",";
+
+            question.append(ripple::rand_int(rng, 0, 3), ' ');
+            question.append(w);
+            question.append(ripple::rand_int(rng, 0, 3), ' ');
+        }
+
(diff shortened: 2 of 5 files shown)


vinniefalco:
Tests?

nbougalis:
Sheesh! Everything fails. That’ll teach me to do a PR on a Friday night!

vinniefalco:
The tests would be nicer if you could tell by looking at them what the results should be for example
```
test( "apple,    pear,     peach", { "apple", "pear", "peach" });
```

nbougalis:
> The tests would be nicer if you could tell by looking at them what the results should be for example
> 
> ```
> test( "apple,    pear,     peach", { "apple", "pear", "peach" });
> ```

Probably, however to exercise all possible combinations (no spaces, single space, multiple spaces, in both prefix & suffix form) would require a lot of entries. I believe that this automatically generated string is better from that standpoint, but it is more difficult to troubleshoot in the case of failure.

If you feel strongly enough about manual checks, I can make that change.

vinniefalco:
I don't have strong feelings about it but to be honest I have not checked the HTTP spec to see if the change is correct

vinniefalco:
The latest RFC is https://tools.ietf.org/html/rfc7231 btw

vinniefalco:
\sigh you made me scrutinize the test. I think that the test is not exercising the modified code path. 

Consider this string:
    "the    rain, in    spain"

For the purpose of splitting into tokens we expect two entries:
    "the rain" and "in spain"

As you can see, the multiple whitespace in between words for each item in the list has been folded down to one whitespace. The test does not look like it exercises this case. The test also does not check that multiple whitespace inside a quoted string is preserved, i.e. the string `"the   rain", in    spain` should turn into `"the   rain"` and `in spain`.

nbougalis:
Fair enough. I’ll migrate to explicit tests. 🖖

src/test/beast/beast_rfc2616_test.cpp
```diff
@@ -0,0 +1,60 @@
...
+        beast::xor_shift_engine rng(0x243F6A8885A308D3);
+
+        std::vector<std::string> words;
+
+        for (int i = 0; i != 64; ++i)
+            words.push_back("X-" + std::to_string(ripple::rand_int(rng, 100, 1000)));
+
+        std::string question;
+
+        for (auto w : words)
```
miguelportilla: `auto const& w`

src/ripple/beast/rfc2616.h
```diff
@@ -257,7 +257,10 @@ split(FwdIt first, FwdIt last, Char delim)
         }
         else if (is_lws (*iter))
         {
-            ++iter;
+            e.append (1, ' ');
+            do
+                iter++;
+            while (iter != last && is_lws (*iter));
```
miguelportilla: Alternate suggestion:

```
e.append (1, ' ');
while (++iter != last && is_lws (*iter));
```

src/ripple/beast/rfc2616.h
```diff
@@ -257,7 +257,10 @@ split(FwdIt first, FwdIt last, Char delim)
         }
         else if (is_lws (*iter))
         {
-            ++iter;
+            e.append (1, ' ');
```
seelabs: Instead of unconditionally adding a space, what do you think about:
```
            if (!e.empty())
                e.append (1, ' ');
```
That way we would only add a space when we are not starting a new token and we could save the `trim_left` call (i.e. replace `trim` with `trim_right`).

nbougalis: I'm confused as to how that would work. Remember, we aren't splitting the word only at spaces (e.g. look for `split_comma` below, which splits on commas.

vinniefalco: I have a better idea, lets get those tests up first so we can visualize the cases that we're discussing!

vinniefalco:
>  to exercise all possible combinations 

You can keep the procedural tests, and just add a few select manual cases of literal strings. There's no need to "exercise all possible combinations" just make sure you get the corner cases and as close to 100% coverage as possible.

src/ripple/beast/rfc2616.h
seelabs: @nbougalis As written, this code collapses all linear white space into a single space. When we add a token to the result, we trim the left and right white space. I'm proposing we change this so when starting a new token we collapse all linear white space into nothing, so we no longer need the left trim.

I.e. there's no reason to add a space to a new token just to trim it away later.

thejohnfreeman: Here's a different [implementation](https://gist.github.com/thejohnfreeman/d1e298e0d860a1bf539e6b0178ae05d9) that's shorter, with fewer else-branches, more DRY, and more early exits, but with a `goto`.

src/test/beast/beast_rfc2616_test.cpp
```diff
@@ -0,0 +1,60 @@
...
+        {
+            if (!question.empty())
+                question += ",";
+
+            question.append(ripple::rand_int(rng, 0, 3), ' ');
+            question.append(w);
+            question.append(ripple::rand_int(rng, 0, 3), ' ');
+        }
+
+        BEAST_EXPECT(beast::rfc2616::split_commas(question) == words);
```
thejohnfreeman, line 54: As the author of [autocheck](https://github.com/thejohnfreeman/autocheck), I appreciate the property-based testing you're trying to do here, but I think we're going to want some example-based tests too. Makes it easier for humans to understand the edge cases, and they're deterministic. If we want to use property-based testing, we should lean on a library that offers deterministic generators.

miguelportilla:
@nbougalis I've written a [branch ](https://github.com/miguelportilla/rippled/commits/rfc2616) with a more extensive unit test, feel free to cherry pick.

seelabs:
@miguelportilla Did you mean to close this?

miguelportilla:
@seelabs Oops! I did not intend to close the PR.

src/ripple/beast/rfc2616.h
nbougalis: I appreciate that implementation @thejohnfreeman but I'm slightly allergic to `goto`. It _may_ be OK here, and I think it does help make the implementation a little clearer, but I need to think whether I want to adopt that approach.

thejohnfreeman:
I just pushed a branch [`rfc2616`](https://github.com/thejohnfreeman/rippled/commits/rfc2616) on my personal fork. It rebases this PR onto 1.6.0-b2 (020b28580816a295185aef8832a8fe07ecc40de2), then applies these changes:

- @miguelportilla's [tests](https://github.com/ripple/rippled/pull/3143#issuecomment-608724603)
- @miguelportilla's [fix for building RocksDB](https://github.com/ripple/rippled/pull/3356)
- my [alternative implementation](https://github.com/ripple/rippled/pull/3143#discussion_r403290429) of `rfc2616::split` (after fixing bugs)

It passes all the tests.

nbougalis:
Thanks @thejohnfreeman. I looked at your branch and while I am heavily allergic to `goto it's certainly used to good effect here.

With that said, I'm not sure how I want to handle this PR. I need to verify that the parsing changes I am proposing here are actually consistent with RFC2616 semantics, instead of consistent with how *I* think things should be parsed.

With that in mind, I'll close this PR for now and reopen, if necessary, at a later time.

PR #3145, opened 2019-11-04
Provide additional info with `txnNotFound` errors.

undertome:
FIXES: #2924

* Tx command now supports min_ledger and max_ledger fields.
* If the requested transaction isn't found and these fields are
  provided, the error response indicates whether or not every
  ledger in the the provided range was searched.

### DIFF
--- a/src/ripple/app/ledger/TransactionMaster.h
+++ b/src/ripple/app/ledger/TransactionMaster.h
@@ -23,6 +23,7 @@
 #include <ripple/protocol/ErrorCodes.h>
 #include <ripple/shamap/SHAMapItem.h>
 #include <ripple/shamap/SHAMapTreeNode.h>
+#include <ripple/basics/RangeSet.h>
 
 namespace ripple {
 
@@ -43,7 +44,9 @@ public:
     fetch_from_cache (uint256 const&);
 
     std::shared_ptr<Transaction>
-    fetch (uint256 const& , error_code_i& ec);
+    fetch (uint256 const& , error_code_i& ec,
+               ClosedInterval<uint32_t> const& range = {},
+                   bool* searchedAll = nullptr);
 
     std::shared_ptr<STTx const>
     fetch (std::shared_ptr<SHAMapItem> const& item,
--- a/src/ripple/app/ledger/TransactionMaster.h
+++ b/src/ripple/app/ledger/TransactionMaster.h
@@ -24,11 +24,11 @@
 #include <ripple/shamap/SHAMapItem.h>
 #include <ripple/shamap/SHAMapTreeNode.h>
 #include <ripple/basics/RangeSet.h>
+#include <ripple/app/misc/Transaction.h>
 
 namespace ripple {
 
 class Application;
-class Transaction;
 class STTx;
 
 // Tracks all transactions in memory
@@ -44,14 +44,14 @@ public:
     fetch_from_cache (uint256 const&);
 
     std::shared_ptr<Transaction>
-    fetch (uint256 const& , error_code_i& ec,
-               ClosedInterval<uint32_t> const& range = {},
-                   bool* searchedAll = nullptr);
+    fetch (uint256 const& , error_code_i& ec);
+
+    Transaction::variant
+    fetch (uint256 const& , error_code_i& ec, ClosedInterval<uint32_t> const& range);
 
     std::shared_ptr<STTx const>
     fetch (std::shared_ptr<SHAMapItem> const& item,
-        SHAMapTreeNode::TNType type, bool checkDisk,
-            std::uint32_t uCommitLedger);
+        SHAMapTreeNode::TNType type, std::uint32_t uCommitLedger);
 
     // return value: true = we had the transaction already
     bool inLedger (uint256 const& hash, std::uint32_t ledger);
--- a/src/ripple/app/ledger/TransactionMaster.h
+++ b/src/ripple/app/ledger/TransactionMaster.h
@@ -20,6 +20,7 @@
 #ifndef RIPPLE_APP_LEDGER_TRANSACTIONMASTER_H_INCLUDED
 #define RIPPLE_APP_LEDGER_TRANSACTIONMASTER_H_INCLUDED
 
+#include <ripple/protocol/ErrorCodes.h>
 #include <ripple/shamap/SHAMapItem.h>
 #include <ripple/shamap/SHAMapTreeNode.h>
 
@@ -39,7 +40,10 @@ public:
     TransactionMaster& operator= (TransactionMaster const&) = delete;
 
     std::shared_ptr<Transaction>
-    fetch (uint256 const& , bool checkDisk);
+    fetch_from_cache (uint256 const&);
+
+    std::shared_ptr<Transaction>
+    fetch (uint256 const& , error_code_i& ec);
 
     std::shared_ptr<STTx const>
     fetch (std::shared_ptr<SHAMapItem> const& item,
--- a/src/ripple/app/ledger/TransactionMaster.h
+++ b/src/ripple/app/ledger/TransactionMaster.h
@@ -55,7 +55,7 @@ public:
      *         the provided range were present in the database
      *         while the search was conducted.
      */
-    Transaction::variant
+    boost::variant<Transaction::pointer, bool>
     fetch (uint256 const& , ClosedInterval<uint32_t> const& range, error_code_i& ec);
 
     std::shared_ptr<STTx const>
--- a/src/ripple/app/ledger/TransactionMaster.h
+++ b/src/ripple/app/ledger/TransactionMaster.h
@@ -46,8 +46,17 @@ public:
     std::shared_ptr<Transaction>
     fetch (uint256 const& , error_code_i& ec);
 
+    /**
+     * Fetch transaction from the cache or database.
+     *
+     * @return A boost::variant that contains either a
+     *         shared_pointer to the retrieved transaction or a
+     *         bool indicating whether or not the all ledgers in
+     *         the provided range were present in the database
+     *         while the search was conducted.
+     */
     Transaction::variant
-    fetch (uint256 const& , error_code_i& ec, ClosedInterval<uint32_t> const& range);
+    fetch (uint256 const& , ClosedInterval<uint32_t> const& range, error_code_i& ec);
 
     std::shared_ptr<STTx const>
     fetch (std::shared_ptr<SHAMapItem> const& item,
--- a/src/ripple/app/ledger/impl/TransactionMaster.cpp
+++ b/src/ripple/app/ledger/impl/TransactionMaster.cpp
@@ -21,7 +21,6 @@
 #include <ripple/app/misc/Transaction.h>
 #include <ripple/app/main/Application.h>
 #include <ripple/protocol/STTx.h>
-#include <ripple/basics/Log.h>
 #include <ripple/basics/chrono.h>
 
 namespace ripple {
@@ -51,15 +50,14 @@ TransactionMaster::fetch_from_cache (uint256 const& txnID)
 }
 
 std::shared_ptr<Transaction>
-TransactionMaster::fetch (uint256 const& txnID, error_code_i& ec,
-    ClosedInterval<uint32_t> const& range, bool* searchedAll)
+TransactionMaster::fetch (uint256 const& txnID, error_code_i& ec)
 {
     auto txn = fetch_from_cache (txnID);
 
     if (txn)
         return txn;
 
-    txn = Transaction::load (txnID, mApp, ec, range, searchedAll);
+    txn = Transaction::load (txnID, mApp, ec);
 
     if (!txn)
         return txn;
@@ -69,10 +67,26 @@ TransactionMaster::fetch (uint256 const& txnID, error_code_i& ec,
     return txn;
 }
 
+Transaction::variant
+TransactionMaster::fetch (uint256 const& txnID, error_code_i& ec,
+    ClosedInterval<uint32_t> const& range)
+{
+    auto txn = mCache.fetch (txnID);
+
+    if (txn)
+        return txn;
+
+    auto v = Transaction::load (txnID, mApp, ec, range);
+
+    if (v.index () == 0 && std::get<0> (v))
+        mCache.canonicalize (txnID, std::get<0> (v));
+
+    return v;
+}
+
 std::shared_ptr<STTx const>
 TransactionMaster::fetch (std::shared_ptr<SHAMapItem> const& item,
-    SHAMapTreeNode::TNType type,
-        bool checkDisk, std::uint32_t uCommitLedger)
+    SHAMapTreeNode::TNType type, std::uint32_t uCommitLedger)
 {
     std::shared_ptr<STTx const>  txn;
     auto iTx = fetch_from_cache (item->key());
--- a/src/ripple/app/ledger/impl/TransactionMaster.cpp
+++ b/src/ripple/app/ledger/impl/TransactionMaster.cpp
@@ -45,14 +45,20 @@ bool TransactionMaster::inLedger (uint256 const& hash, std::uint32_t ledger)
 }
 
 std::shared_ptr<Transaction>
-TransactionMaster::fetch (uint256 const& txnID, bool checkDisk)
+TransactionMaster::fetch_from_cache (uint256 const& txnID)
 {
-    auto txn = mCache.fetch (txnID);
+    return mCache.fetch (txnID);
+}
 
-    if (!checkDisk || txn)
+std::shared_ptr<Transaction>
+TransactionMaster::fetch (uint256 const& txnID, error_code_i& ec)
+{
+    auto txn = fetch_from_cache (txnID);
+
+    if (txn)
         return txn;
 
-    txn = Transaction::load (txnID, mApp);
+    txn = Transaction::load (txnID, mApp, ec);
 
     if (!txn)
         return txn;
@@ -68,7 +74,7 @@ TransactionMaster::fetch (std::shared_ptr<SHAMapItem> const& item,
         bool checkDisk, std::uint32_t uCommitLedger)
 {
     std::shared_ptr<STTx const>  txn;
-    auto iTx = fetch (item->key(), false);
+    auto iTx = fetch_from_cache (item->key());
 
     if (!iTx)
     {
--- a/src/ripple/app/ledger/impl/TransactionMaster.cpp
+++ b/src/ripple/app/ledger/impl/TransactionMaster.cpp
@@ -51,14 +51,15 @@ TransactionMaster::fetch_from_cache (uint256 const& txnID)
 }
 
 std::shared_ptr<Transaction>
-TransactionMaster::fetch (uint256 const& txnID, error_code_i& ec)
+TransactionMaster::fetch (uint256 const& txnID, error_code_i& ec,
+    ClosedInterval<uint32_t> const& range, bool* searchedAll)
 {
     auto txn = fetch_from_cache (txnID);
 
     if (txn)
         return txn;
 
-    txn = Transaction::load (txnID, mApp, ec);
+    txn = Transaction::load (txnID, mApp, ec, range, searchedAll);
 
     if (!txn)
         return txn;
--- a/src/ripple/app/ledger/impl/TransactionMaster.cpp
+++ b/src/ripple/app/ledger/impl/TransactionMaster.cpp
@@ -67,7 +67,7 @@ TransactionMaster::fetch (uint256 const& txnID, error_code_i& ec)
     return txn;
 }
 
-Transaction::variant
+boost::variant<Transaction::pointer, bool>
 TransactionMaster::fetch (uint256 const& txnID, ClosedInterval<uint32_t> const& range,
     error_code_i& ec)
 {
@@ -78,7 +78,8 @@ TransactionMaster::fetch (uint256 const& txnID, ClosedInterval<uint32_t> const& 
     if (txn)
         return txn;
 
-    Transaction::variant v = Transaction::load (txnID, mApp, range, ec);
+    boost::variant<Transaction::pointer, bool> v = Transaction::load (
+        txnID, mApp, range, ec);
 
     if (v.which () == 0 && boost::get<pointer> (v))
         mCache.canonicalize (txnID, boost::get<pointer> (v));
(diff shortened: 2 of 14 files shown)


seelabs:
Can we write new unit tests for this to show that it's working as expected?

src/ripple/app/ledger/TransactionMaster.h
```diff
@@ -39,7 +40,9 @@ class TransactionMaster
     TransactionMaster& operator= (TransactionMaster const&) = delete;
 
     std::shared_ptr<Transaction>
-    fetch (uint256 const& , bool checkDisk);
+    fetch (uint256 const& , bool checkDisk,
+               ClosedInterval<uint32_t> const& range = {},
+                   bool* searchedAll = nullptr);
```
seelabs: What do you think of a new overload, something like:
```
std::variant<Transaction::pointer, bool> load(uint256 const&, bool, ClosedInterval const&)
```
Rational: I want to make clear that this function returns either the transaction or a bool saying if the entire range is searched. If it returns the transaction, I want to skip the `searchedAll` calculation, and I don't want callers to think that it's valid value.

src/ripple/app/misc/impl/Transaction.cpp
```diff
@@ -132,6 +133,23 @@ Transaction::pointer Transaction::load(uint256 const& id, Application& app)
         soci::blob sociRawTxnBlob (*db);
         soci::indicator rti;
 
+        if(searchedAll)
+        {
```
seelabs: I would only do this calculation if we don't find the txn. (See other comment about changing the function signature to make it clear this function either returns the txn or if the range was searched).

src/ripple/app/misc/impl/Transaction.cpp
```diff
@@ -132,6 +133,23 @@ Transaction::pointer Transaction::load(uint256 const& id, Application& app)
...
+
+            *db << "SELECT COUNT(DISTINCT LedgerSeq) FROM Transactions WHERE LedgerSeq BETWEEN "
+                << range.first ()
+                << " AND "
+                << range.last ()
+                << ";",
+                soci::into (count, rti);
+
+            if (!db->got_data () || rti != soci::i_ok)
+                return {};
```
seelabs: I don't think I'd return early here. Instead I'd set `searchedAll` to `false`

src/ripple/app/misc/impl/Transaction.cpp
```diff
@@ -132,6 +133,23 @@ Transaction::pointer Transaction::load(uint256 const& id, Application& app)
...
         soci::indicator rti;
 
+        if(searchedAll)
+        {
+            uint64_t count = 0;
+
+            *db << "SELECT COUNT(DISTINCT LedgerSeq) FROM Transactions WHERE LedgerSeq BETWEEN "
+                << range.first ()
+                << " AND "
+                << range.last ()
```
seelabs: I'm unsure how expensive this function is when the range is large. I think we want to restrict the range to a reasonable size. @intelliot has suggested a range of 200 would be sufficient. I'm fine making it somewhat bigger than that if we want. I'll suggest 500 or 1000.

src/ripple/net/impl/RPCCall.cpp
```diff
@@ -1008,7 +1008,13 @@ class RPCParser
                 jvRequest[jss::binary] = true;
         }
 
-        jvRequest["transaction"]    = jvParams[0u].asString ();
+        if (jvParams.size () > 3)
+        {
```
seelabs: I'm not sure this parses right. As written, if we specify a range we must specify binary. And something like `tx 100 200` would be misparsed an no error would be issued (at least by my reading of the code). I'd love to change this so we'd give an error if we can't parse correctly and we correctly make binary an optional parameter. We probably also want to give errors for too large a range or for min>max.

src/ripple/app/ledger/TransactionMaster.h
nbougalis: I like this approach.

src/ripple/app/misc/impl/Transaction.cpp
undertome: Should we return an error if the range exceeds the limit?

src/ripple/app/misc/impl/Transaction.cpp
undertome: You think its possible for us to have the next query succeed even if this one fails?

seelabs: @undertome I don't see a way for this to fail and the next one to succeed. I agree, code is fine as-is.

src/ripple/app/misc/impl/Transaction.cpp
seelabs: Yes, I would return an error.

src/ripple/app/ledger/TransactionMaster.h
undertome: I considered doing something similar with `boost::optional`, but the only drawback to this approach is that now we have to rely on the value of the `ClosedInterval` parameter to determine whether or not the user intended to perform the range search. We can assume that a default range of `[0,0]` means that the user didn't supply a range, but this means that in some imagined scenario where the user actually wants to check ledger 0, this won't be possible. But if we don't really care about that, then I'll just use this approach as is. The other option is to keep the variant return, but allow a bool parameter but then this is very similar to my original approach.

src/ripple/app/misc/impl/Transaction.cpp
undertome: Done

src/ripple/app/misc/impl/Transaction.cpp
undertome: Done

src/ripple/net/impl/RPCCall.cpp
undertome: Done

src/ripple/app/ledger/TransactionMaster.h
seelabs: The way this interface works is:
1) When we don't pass an interval, the `searchedAll` parameter isn't relevant. 
2) When we find the transaction, the `searchedAll` parameter isn't relevant. 
3) When we don't find the transaction, and we specify an interval, when we need to set `searchedAll`. 

I'm proposing that we keep the old function, and add an additional overload. I.e.:
```
// overload 1
std::shared_ptr<Transaction> 
fetch (uint256 const& , bool checkDisk);

// overload 2
std::variant<Transaction::pointer, bool> 
load(uint256 const&, bool checkDisk, ClosedInterval const&)
```
I think these two overload capture those semantics better than the single function with the default interval parameter and `searchedAll` out parameter. 

Note: My biggest complaint about the above interface is `std::variant` is a pain to use in C++, but in this case I prefer it to the alternatives. An overload returning `std::pair<shared_ptr, bool>` would be my second choice. I would avoid `tribool` or optional<bool>` as too error prone.

If we need to distinguish if the user passes a range or not, we could use an `optional<ClosedInterval>` for that. However, I prefer the two overloads to that.

undertome: Is overload 1 supposed to be load?

seelabs: @undertome Oops. That was a typo. Both should be `fetch`.

undertome: Understood - I’ll have think of a good way to distinguish wether or not the range should be used when you get to the load method though.

seelabs: Overload `load` as well.

undertome:
> Can we write new unit tests for this to show that it's working as expected?

Done.

undertome:
> Left some comments. Note that this PR is going to conflict with another PR in the queue (this one: #3084). You might want base this PR of the other one to save some trouble later.

Rebased. Done.

src/ripple/app/ledger/TransactionMaster.h
undertome: Right, but my concern was avoiding code duplication. So I went with overload stubs to mitigate that.

undertome: Done.

undertome: Is std::variant used elsewhere in the code? It seems to be causing build errors with clang...

mellery451: clang 5/6 and xcode 9 specifically seem to fail. No, I don't think we've used `<variant>` so far and it's looking like we'd need to trim our supported compilers or make accommodations for them (fallback to boost perhaps?)

HowardHinnant: I'm ok with requiring xcode 10.  Requiring latest tools on platforms we don't officially support doesn't seem that draconian to me.

undertome: Great, that solves the xcode failure - what should we do about clang 5/6?

seelabs: I'd use `boost::variant`, as @mellery451 suggests

src/ripple/app/misc/impl/Transaction.cpp
```diff
@@ -134,14 +132,48 @@ Transaction::pointer Transaction::load(uint256 const& id, Application& app)
...
-        if (!db->got_data () || rti != soci::i_ok)
-            return {};
 
-        convert(sociRawTxnBlob, rawTxn);
+        auto const got_data = db->got_data ();
+
+        if ((!got_data || rti != soci::i_ok) && !useRange)
+            return nullptr;
+
+        if (!got_data)
```
seelabs: May as well check `rti` here too.

src/ripple/rpc/handlers/Tx.cpp
```diff
@@ -97,41 +96,87 @@ Json::Value doTx (RPC::Context& context)
...
+        }
+        catch (...)
+        {
+            // One of the calls to `asUInt ()` failed.
+            return rpcError (rpcINVALID_LGR_RANGE);
+        }
+    }
+
+    auto ec = rpcSUCCESS;
+    Transaction::variant v;
```
seelabs: Instead of a variant here, how about a `Transaction::pointer`. Then in the `rangeProvided` branch, handle the `bool` alternative of the variant and return there, or set the pointer and continue. The non`rangeProvided` branch would just set the pointer.

Doing it this way probably require duplicating two lines of code, but I think that's an acceptable trade-off for the clearer code.

src/ripple/rpc/handlers/Tx.cpp
```diff
@@ -97,41 +96,87 @@ Json::Value doTx (RPC::Context& context)
...
+        v = context.app.getMasterTransaction().fetch(
+            from_hex_text<uint256>(txid), ec, range);
+    else
+        v = context.app.getMasterTransaction().fetch(
+            from_hex_text<uint256>(txid), ec);
+
+    if (ec == rpcDB_DESERIALIZATION)
+        return rpcError (ec);
+
+    if (v.index () != 0 || !std::get<0> (v))
```
seelabs: It's not clear what index 0 is here. I'd move this variant handling to the `rangeProvided` branch only and explicitly use the variant type rather than the type alias.

src/ripple/rpc/handlers/Tx.cpp
```diff
@@ -97,41 +96,87 @@ Json::Value doTx (RPC::Context& context)
...
+        if (rangeProvided)
+            jvResult[jss::searched_all] = std::get<1> (v);
+
+        return rpcError (rpcTXN_NOT_FOUND, jvResult);
+    }
+
+    Json::Value ret = std::get<0> (v)->getJson (JsonOptions::include_date, binary);
 
-    if (txn->getLedger () == 0)
+    if (std::get<0> (v)->getLedger () == 0)
```
seelabs: All these `get<0>` should be replaced with `txn` (see above about moving `v` to the `rangeProvided` branch only.

src/test/rpc/Transaction_test.cpp
```diff
@@ -0,0 +1,282 @@
...
+        const char* BINARY    = jss::binary.c_str();
+        const char* NOT_FOUND = RPC::get_error_info(rpcTXN_NOT_FOUND).token;
+        const char* INVALID   = RPC::get_error_info(rpcINVALID_LGR_RANGE).token;
+        const char* EXCESSIVE = RPC::get_error_info(rpcEXCESSIVE_LGR_RANGE).token;
+
+        Env env(*this);
+        auto const alice = Account("alice");
+        env.fund(XRP(1000), alice);
+        env.close();
+
```
seelabs: Extraneous white space.

src/ripple/app/misc/Transaction.h
```diff
@@ -61,8 +64,8 @@ class Transaction
 
     using pointer = std::shared_ptr<Transaction>;
     using ref = const pointer&;
+    using variant = std::variant<Transaction::pointer, bool>;
```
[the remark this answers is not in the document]
undertome: I want to use an alias mostly because the full typename is pretty long. Can we just put it somewhere else?

src/ripple/app/misc/impl/Transaction.cpp
undertome: I don't bother checking `rti` here since when the SQL statement fails to retrieve data `rti != soci::i_ok` is always true.

src/ripple/app/misc/Transaction.h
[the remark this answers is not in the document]
seelabs: I'd probably would write this type out. Type aliases and `auto` can be useful for long types when we don't really care about the actual type (container `iterator` types would be a good example). In this case, when we read the code we actually care that it's a variant and we'll want to know the types. If we use a type alias (or `auto`) we'll end up looking up that alias almost every time we read the code.

However, I do understand this is a judgement call. If you feel the code is better with the type alias, leave it in and we can see what other reviews think too.

src/test/rpc/Transaction_test.cpp
undertome: I thought breaking the statements at this point improves the readability and look of the code, but done anyway.

src/ripple/rpc/handlers/Tx.cpp
undertome: Done.

src/ripple/rpc/handlers/Tx.cpp
undertome: Done.

src/ripple/rpc/handlers/Tx.cpp
undertome: Done.

src/ripple/app/misc/Transaction.h
```diff
@@ -156,9 +151,18 @@ class Transaction
...
+    load (uint256 const& id, Application& app, error_code_i& ec);
+
+    static variant
+    load (uint256 const& id, Application& app, error_code_i& ec, ClosedInterval<uint32_t> const& range);
 
 private:
+
+    static variant
+    load (uint256 const& id, Application& app, error_code_i& ec, ClosedInterval<uint32_t> const& range,
+        bool useRange);
```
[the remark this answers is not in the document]
undertome: I didn't initially because you expressed reluctance at using optional, but done.

src/ripple/app/ledger/impl/TransactionMaster.cpp
```diff
@@ -62,13 +67,29 @@ TransactionMaster::fetch (uint256 const& txnID, bool checkDisk)
...
+    ClosedInterval<uint32_t> const& range)
+{
+    auto txn = mCache.fetch (txnID);
+
+    if (txn)
+        return txn;
+
+    auto v = Transaction::load (txnID, mApp, ec, range);
+
+    if (v.index () == 0 && std::get<0> (v))
```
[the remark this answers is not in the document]
undertome: Done (using the alias though).

src/ripple/app/ledger/TransactionMaster.h
```diff
@@ -39,12 +41,17 @@ class TransactionMaster
...
 
     std::shared_ptr<Transaction>
-    fetch (uint256 const& , bool checkDisk);
+    fetch_from_cache (uint256 const&);
+
+    std::shared_ptr<Transaction>
+    fetch (uint256 const& , error_code_i& ec);
+
+    Transaction::variant
+    fetch (uint256 const& , error_code_i& ec, ClosedInterval<uint32_t> const& range);
```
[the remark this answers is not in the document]
undertome: Done.

src/ripple/app/ledger/TransactionMaster.h
```diff
@@ -39,12 +41,17 @@ class TransactionMaster
     TransactionMaster& operator= (TransactionMaster const&) = delete;
 
     std::shared_ptr<Transaction>
-    fetch (uint256 const& , bool checkDisk);
+    fetch_from_cache (uint256 const&);
+
+    std::shared_ptr<Transaction>
+    fetch (uint256 const& , error_code_i& ec);
+
+    Transaction::variant
```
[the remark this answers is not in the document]
undertome: Added documentation.

src/ripple/app/misc/Transaction.h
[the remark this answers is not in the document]
undertome: Does anyone else have an opinion on this

undertome:
Yesterday I had a test case failing on certain Travis CI builds. I debugged it by pushing code with `std::cout` statements. Moving forward, is there a way to ssh into a machine that matches a Travis build configuration to debug on other platforms?

src/test/rpc/Transaction_test.cpp
seelabs: @undertome I don't mind the newline at all - I'm in favor of keeping it. I'm flagging the eight spaces that start the line.

src/ripple/app/misc/Transaction.h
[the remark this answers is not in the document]
seelabs: Ah, sorry for the misunderstanding. I'm reluctant to use `optional<bool>`, not `optional` in general.

src/ripple/app/misc/Transaction.h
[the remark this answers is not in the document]
scottschurr: Sure.  I agree with @seelabs.  I think in this case it's way better to see the type spelled out in the source code.  Unlike some of the other long templated type names, the information describing the `variant` is really useful for understanding the nature and use of the `variant`.  Just my take.

src/test/rpc/Transaction_test.cpp
undertome: Done.

src/ripple/app/misc/Transaction.h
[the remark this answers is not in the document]
undertome: Done. BTW is there a coding standard that I can consult? Would be helpful to avoid style comments in the future.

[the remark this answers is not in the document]
seelabs: @undertome Our coding standard is here: https://github.com/ripple/rippled/blob/develop/docs/CodingStyle.md

However, when to define type aliases and when to use `auto` are judgement calls. Unless we had a rule that said "never use auto" or "never use type aliases" (and no one wants that) it's hard to codify judgement calls into a style guide.

src/ripple/app/misc/impl/Transaction.cpp
```diff
@@ -99,28 +100,29 @@ Transaction::pointer Transaction::transactionFromSQL (
...
-    auto ret = transactionFromSQL(ledgerSeq, status, rawTxn, app);
+    using op = boost::optional<ClosedInterval<uint32_t>>;
 
-    if (checkValidity(app.getHashRouter(),
-            *ret->getSTransaction(), app.
-                getLedgerMaster().getValidatedRules(),
-                    app.config()).first !=
-                        Validity::Valid)
-        return {};
+    return boost::get<pointer> (load (id, app, op {}, ec));
```
seelabs: No need for the type alias and `op{}`. Use `boost::none` here instead.

src/ripple/app/misc/impl/Transaction.cpp
```diff
@@ -99,28 +100,29 @@ Transaction::pointer Transaction::transactionFromSQL (
...
+}
 
-    return ret;
+boost::variant<Transaction::pointer, bool>
+Transaction::load (uint256 const& id, Application& app, ClosedInterval<uint32_t> const& range,
+    error_code_i& ec)
+{
+    using op = boost::optional<ClosedInterval<uint32_t>>;
+
+    return load (id, app, op {range}, ec);
```
seelabs: No need the type alias and `op{range}`. Just `range` will work (a temporary `boost::optional` will be created for the parameter).

src/ripple/rpc/handlers/Tx.cpp
```diff
@@ -97,11 +96,68 @@ Json::Value doTx (RPC::Context& context)
...
+    pointer txn;
+
+    if (rangeProvided)
+    {
+        boost::variant<Transaction::pointer, bool> v =
+            context.app.getMasterTransaction().fetch(
+                from_hex_text<uint256>(txid), range, ec);
+
+        if (v.which () != 0 || !boost::get<pointer> (v))
+            searchedAll = boost::get<bool> (v);
```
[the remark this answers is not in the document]
undertome: Its actually not possible for both of these conditions to be true:

-  `v.which() != 0`
-  `boost::get<pointer> (v) == nullptr` 

But I will change the logic as you've indicated to simplify the code.

src/ripple/app/misc/impl/Transaction.cpp
undertome: It turns out that just using `range` doesn't work because the compiler will end up calling the current method which will recurse infinitely.

seelabs: Oh, I missed that! Good catch. Yup, you need the wrapper.

src/ripple/rpc/handlers/Tx.cpp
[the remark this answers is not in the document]
seelabs: I meant `v.which() == 0` and the pointer part of the variant is a nullpointer then line 141 throws 
(I suspect you meant that in your replay as well). If that can't happen (it looks like it only happens on a serialization error), then we shouldn't check for it. If we do check for it, we need to do something sensible when we detect it.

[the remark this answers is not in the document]
undertome: Yeah, that also can’t happen. Basically in this branch of the code we’ll never have a ‘nullptr’ because we’d be return a ‘bool’ in the ‘variant’ instead. But yeah, I’ll clean up the logic here to be cleaner.

src/ripple/app/misc/impl/Transaction.cpp
undertome: Done.

Issue #3121, opened 2019-10-18
Trailing comments `#...` in rippled config lines aren't handled


mellery451:
...currently trailing comment text probably gets interpreted as a value or some such thing and in-fact they just need to be stripped away before any real parsing.

alloyxrp:
A line which has a trailing comment gets ignored completely.

scottschurr:
I think the actual state of the world is a little worse than that.  If I remember correctly, it depends on the other content of the line as to what effect the trailing comment has.

mellery451:
Proposal: 
1. allow trailing comments on all config files lines (`#` marks the start of a comment anywhere on a line).
2. if you need a literal `#` in one of your fields/lines, escape the `#` with a `\`. (the only field types that could even possibly need a `#` might be passwords or file paths ?)
3. If any trailing comment is detected in a current config, emit a warning to the log letting operators know that the behavior with respect to trailing comments has changed. This warning would be removed in a subsequent release.  

Thoughts?

scottschurr:
I'm not keen on escaping literal `#` in lines that need it.  Escapes get really messy because you need to set up ways to escape the escape.

I'd like to suggest an alternative that I got from from @bachase:

- As you parse the config file check every line to see if all characters are consumed.  Only if there are leftover trailing characters...
- Check if the leftover characters look like a comment (optional whitespace followed by a `#` optionally followed by other characters).
- If the leftover characters look like a comment then ignore the comment and proceed.
- If there are leftover characters that don't look like a comment, then echo a diagnostic message to the console, stop parsing, and close rippled.

Some types of lines consume the entire line.  Examples are key/value pairs for passwords and file paths.  If we use the approach I suggest it will never be possible to add trailing comments to these lines.  But it's valid to put the comment on the previous line in these few cases.

The suggested approach has the advantage that people will not need to edit their config files to escape any required `#'` characters.

MarkusTeufelberger:
How about using an existing library to parse config files in some standard format (ini, yaml, toml...)? Surely this is not a unique problem...

movitto:
> How about using an existing library to parse config files in some standard format (ini, yaml, toml...)? Surely this is not a unique problem...

@MarkusTeufelberger technically rippled.cfg isn't ini as comments in the former are marked by '#'s and ';' in the later

https://en.wikipedia.org/wiki/INI_file#Comments_2

There are a few libraries in this domain we could use. I worked on this in a previous life: http://augeas.net/

[...]

PR #3151, opened 2019-11-06
Allow trailing comments in config file:

mellery451:
Fixes: #3121

Treat all `#` characters in config files as comments (and remove) *unless* the `#` is immediately preceded by `\`. Write a warning to log file when trailing comments are found/ignored in the config
to let operators know that the treatment of trailing `#` has changed.

### DIFF
--- a/src/ripple/basics/BasicConfig.h
+++ b/src/ripple/basics/BasicConfig.h
@@ -25,6 +25,7 @@
 #include <boost/beast/core/string.hpp>
 #include <boost/lexical_cast.hpp>
 #include <boost/optional.hpp>
+#include <algorithm>
 #include <map>
 #include <ostream>
 #include <string>
@@ -47,6 +48,7 @@ private:
     std::string name_;
     std::vector <std::string> lines_;
     std::vector <std::string> values_;
+    bool had_trailing_comments_ = false;
 
 public:
     /** Create an empty section. */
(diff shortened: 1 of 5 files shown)


src/ripple/basics/BasicConfig.h
```diff
@@ -364,12 +352,10 @@ get_if_exists<bool> (Section const& section,
...
     int intVal = 0;
-    if (get_if_exists (section, name, intVal))
-    {
+    auto stat = get_if_exists(section, name, intVal);
+    if (stat)
         v = bool (intVal);
-        return true;
-    }
-    return false;
+    return stat;
```
miguelportilla: Nit: I don't see the value in creating `stat`.

src/ripple/basics/impl/BasicConfig.cpp
```diff
@@ -53,14 +54,54 @@ Section::append (std::vector <std::string> const& lines)
...
+                    removed_trailing = true;
+                    break;
+                }
+
+                comment = val.find('#', comment);
+            }
+            return removed_trailing;
+        };
+
+        if (remove_comment(line) && !line.empty())
```
miguelportilla: Suggestion:

```if (!had_trailing_comments_ && remove_comment(line) && !line.empty())```

PR #3152, opened 2019-11-06
Return correct HTTP header Forwarded IP address with multiple proxies.

mtrippled:


src/ripple/rpc/impl/Role.cpp
```diff
@@ -108,14 +109,7 @@ requestInboundEndpoint (Resource::Manager& manager,
...
 {
-    auto it = request.find("X-Forwarded-For");
-    if (it != request.end())
-    {
-        return boost::beast::http::ext_list{
-            it->value()}.begin()->first;
-    }
-
-    it = request.find("Forwarded");
+    auto it = request.find(boost::beast::http::field::forwarded);
```
mellery451: you might want to `#include` the header that defines the fields, just for completeness.

src/ripple/rpc/impl/Role.cpp
```diff
@@ -137,10 +131,20 @@ forwardedFor(http_request_type const& request)
             return {};
 
         found += forStr.size();
-        auto pos{it->value().find(';', forStr.size())};
+        auto pos = std::string(found).find(';');
```
mellery451: can you do this without the temporary string ?(e.g. using the former string view methods, but perhaps correcting the start offset). I don't this is safe and in-fact the asan build is flagging it.

src/ripple/rpc/impl/Role.cpp
mtrippled: fixed

src/ripple/rpc/impl/Role.cpp
mtrippled: fixed

src/ripple/rpc/impl/Role.cpp
```diff
@@ -137,10 +131,20 @@ forwardedFor(http_request_type const& request)
...
         if (pos != boost::string_view::npos)
-            return {found, pos + 1};
-        return {found, it->value().size() - forStr.size()};
+        {
+            return *boost::beast::http::token_list(boost::string_view(
+                found, pos)).begin();
+        }
+        return *boost::beast::http::token_list(boost::string_view(
+            found, it->value().size() - forStr.size())).begin();
+    }
```
miguelportilla: An alternate to consider:

```
        auto const pos ([&]() {
            auto const pos {boost::string_view(
                found, it->value().end() - found).find(';')};
            if (pos == boost::string_view::npos)
                return it->value().size() - forStr.size();
            return pos;
        }());

        return *boost::beast::http::token_list(
            boost::string_view(found, pos)).begin();```

Comment (mtrippled) [reply src/ripple/rpc/impl/Role.cpp]: fixed

File: src/ripple/rpc/impl/Role.cpp [hunk 4]
```diff
@@ -137,10 +133,23 @@ forwardedFor(http_request_type const& request)
...
+        std::size_t const pos ([&]()
+        {
+            std::size_t const pos{boost::string_view(
+                found, it->value().end() - found).find(';')};
+            if (pos == boost::string_view::npos)
+                return it->value().size() - forStr.size();
+            return pos;
+        }());
+
+        return *boost::beast::http::token_list(
```
jwbusch: nit: does this container have a `front()` method? slightly clearer vs. dereferencing the begin() iterator

src/ripple/rpc/impl/Role.cpp
```diff
@@ -137,10 +133,23 @@ forwardedFor(http_request_type const& request)
...
+            boost::string_view(found, pos)).begin();
+    }
+
+    it = request.find("X-Forwarded-For");
+    if (it != request.end())
+    {
+        return *boost::beast::http::token_list(it->value()).begin();
```
jwbusch: This used to be `beast::http::ext_list` and now it's `beast::http::token_list` ... was that part of the bug fix?

mtrippled: @jwbusch Yes, "ext_list" returns extensions which are basically key=value pairs, each separated by comma. "token_list" returns single items (tokens), also separated by comma.

The X-Forwarded-For HTTP header contains IP addresses separated by columns. I don't know why I used ext_list originally, but it was a mistake.

PR #3154, opened 2019-11-12
Report queue time and execute time together always for [insights]

mayurbhandary:
The amount of time that a job spends in the queue and the amount of time that it takes to execute the job are reported independently. It would be good to report them together so that we don't see one without the other.

### DIFF
--- a/src/ripple/app/ledger/LedgerMaster.h
+++ b/src/ripple/app/ledger/LedgerMaster.h
@@ -371,6 +371,35 @@ private:
     // without first wiping the database.
     LedgerIndex const max_ledger_difference_ {1000000};
 
+private:
+    struct Stats
+    {
+        template <class Handler>
+        Stats (Handler const& handler, beast::insight::Collector::ptr const& collector)
+            : hook (collector->make_hook (handler)),
+            mValidatedLedgerAge (collector->make_gauge ("LedgerMaster", "Validated_Ledger_Age")),
+            mPublishedLedgerAge (collector->make_gauge ("LedgerMaster", "Published_Ledger_Age"))
+            { }
+
+        beast::insight::Hook hook;
+        beast::insight::Gauge mValidatedLedgerAge;
+        beast::insight::Gauge mPublishedLedgerAge;
+    };
+
+    Stats m_stats;
+
+private:
+    void collect_metrics()
+    {
+        std::lock_guard lock (m_mutex);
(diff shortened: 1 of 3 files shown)


src/ripple/core/impl/JobQueue.cpp
```diff
@@ -420,19 +420,27 @@ JobQueue::processTask (int instance)
...
-            perfLog_.jobStart(type, us, start_time, instance);
-            if (us >= 10ms)
-                getJobTypeData(type).dequeue.notify(us);
+            perfLog_.jobStart(type, q_time, start_time, instance);
+            
             job.doJob ();
+
+            //The amount of time it took to dequeue and execute the job
+            auto const x_time = date::ceil<microseconds>(
+                Job::clock_type::now() - job.queue_time());
```
nbougalis: `job.queue_time()` should be `start_time` and the comment should be updated to reflect the fact that this represents the execution and not the wait time.

mayurbhandary: Thanks. I made the changes.

src/ripple/app/ledger/LedgerMaster.h
```diff
@@ -371,6 +371,35 @@ class LedgerMaster
     // without first wiping the database.
     LedgerIndex const max_ledger_difference_ {1000000};
 
+private:
+    struct Stats
+    {
+        template <class Handler>
+        Stats (Handler const& handler, beast::insight::Collector::ptr const& collector)
+            : hook (collector->make_hook (handler)),
+            mValidatedLedgerAge (collector->make_gauge ("LedgerMaster", "Validated_Ledger_Age")),
```
nbougalis: Formatting micro-nit:
```
    template <class Handler>
    Stats (Handler const& handler, beast::insight::Collector::ptr const& collector)
        : hook (collector->make_hook (handler))
        , validatedLedgerAge (collector->make_gauge ("LedgerMaster", "Validated_Ledger_Age"))
        , publishedLedgerAge (collector->make_gauge ("LedgerMaster", "Published_Ledger_Age"))
    { }
```

Also, note how I avoided the `m` prefix by using just `validatedLedgerAge` and `publishedLedgerAge`.

mayurbhandary: Thanks! I fixed it.

PR #3155, opened 2019-11-12
API versioning

pwang200:
The requirements are:
https://github.com/pwang200/RippledRPCDesign/blob/API_versioning/requirement/requirements.md
The design document is:
https://github.com/pwang200/RippledRPCDesign/blob/API_versioning/design/design.md

Note that the documentation is for informational only. We will consolidate the documentation and create another PR after both API Versioning and gRPC are merged.

### DIFF
--- a/src/ripple/net/impl/RPCCall.cpp
+++ b/src/ripple/net/impl/RPCCall.cpp
@@ -1355,18 +1355,18 @@ rpcCmdLineToJson (std::vector<std::string> const& args,
     jvRequest   = rpParser.parseCommand (args[0], jvRpcParams, true);
 
     auto insert_api_version = [](Json::Value & jr){
-        if(!jr.isMember(jss::error) && !jr.isMember(jss::api_version))
+        if( jr.isObject() &&
+            !jr.isMember(jss::error) &&
+            !jr.isMember(jss::api_version))
+        {
             jr[jss::api_version] = RPC::ApiMaximumSupportedVersion;
+        }
     };
 
     if(jvRequest.isObject())
         insert_api_version(jvRequest);
-    else if(jvRequest.isArray()) {
-        for (Json::UInt j = 0; j < jvRequest.size(); ++j) {
-            if (jvRequest[j].isObject())
-                insert_api_version(jvRequest[j]);
-        }
-    }
+    else if(jvRequest.isArray())
+        std::for_each(jvRequest.begin(), jvRequest.end(), insert_api_version);
 
     JLOG (j.trace()) << "RPC Request: " << jvRequest << std::endl;
     return jvRequest;
--- a/src/ripple/net/impl/RPCCall.cpp
+++ b/src/ripple/net/impl/RPCCall.cpp
@@ -36,6 +36,7 @@
 #include <ripple/protocol/SystemParameters.h>
 #include <ripple/protocol/UintTypes.h>
 #include <ripple/rpc/ServerHandler.h>
+#include <ripple/rpc/impl/RPCHelpers.h>
 #include <ripple/beast/core/LexicalCast.h>
 
 #include <boost/algorithm/string/predicate.hpp>
@@ -1352,8 +1353,20 @@ rpcCmdLineToJson (std::vector<std::string> const& args,
     retParams[jss::params] = jvRpcParams;
 
     jvRequest   = rpParser.parseCommand (args[0], jvRpcParams, true);
-
+    if(jvRequest.isObject())
+    {
+        if(!jvRequest.isMember(jss::error))
+            jvRequest["api_version"] = RPC::APIVersionCommandLine;
+    }
+    else if(jvRequest.isArray()) {
+        for (Json::UInt j = 0; j < jvRequest.size(); ++j) {
+            if (jvRequest[j].isObject())
+                jvRequest[j]["api_version"] = RPC::APIVersionCommandLine;
+        }
+    }
     JLOG (j.trace()) << "RPC Request: " << jvRequest << std::endl;
+    //std::cout << "RPC Request: " << jvRequest << std::endl;//TODO Peng
+
     return jvRequest;
 }
 
@@ -1441,9 +1454,13 @@ rpcClient(std::vector<std::string> const& args,
                 jvRequest["admin_password"] = setup.client.admin_password;
 
             if (jvRequest.isObject())
+            //{
+                //jvRequest["api_version"] = RPC::APIVersionCommandLine;
                 jvParams.append (jvRequest);
+            //}
             else if (jvRequest.isArray())
             {
+                //jvParams.append(RPC::getAPIVersionString());
                 for (Json::UInt i = 0; i < jvRequest.size(); ++i)
                     jvParams.append(jvRequest[i]);
             }
--- a/src/ripple/net/impl/RPCCall.cpp
+++ b/src/ripple/net/impl/RPCCall.cpp
@@ -1353,19 +1353,22 @@ rpcCmdLineToJson (std::vector<std::string> const& args,
     retParams[jss::params] = jvRpcParams;
 
     jvRequest   = rpParser.parseCommand (args[0], jvRpcParams, true);
+
+    auto insert_api_version = [](Json::Value & jr){
+        if(!jr.isMember(jss::error) && !jr.isMember(jss::api_version))
+            jr[jss::api_version] = RPC::APIVersionCommandLine;
+    };
+
     if(jvRequest.isObject())
-    {
-        if(!jvRequest.isMember(jss::error))
-            jvRequest["api_version"] = RPC::APIVersionCommandLine;
-    }
+        insert_api_version(jvRequest);
     else if(jvRequest.isArray()) {
         for (Json::UInt j = 0; j < jvRequest.size(); ++j) {
             if (jvRequest[j].isObject())
-                jvRequest[j]["api_version"] = RPC::APIVersionCommandLine;
+                insert_api_version(jvRequest[j]);
         }
     }
+
     JLOG (j.trace()) << "RPC Request: " << jvRequest << std::endl;
-    //std::cout << "RPC Request: " << jvRequest << std::endl;//TODO Peng
 
     return jvRequest;
 }
@@ -1454,13 +1457,11 @@ rpcClient(std::vector<std::string> const& args,
                 jvRequest["admin_password"] = setup.client.admin_password;
 
             if (jvRequest.isObject())
-            //{
-                //jvRequest["api_version"] = RPC::APIVersionCommandLine;
+            {
                 jvParams.append (jvRequest);
-            //}
+            }
             else if (jvRequest.isArray())
             {
-                //jvParams.append(RPC::getAPIVersionString());
                 for (Json::UInt i = 0; i < jvRequest.size(); ++i)
                     jvParams.append(jvRequest[i]);
             }
--- a/src/ripple/net/impl/RPCCall.cpp
+++ b/src/ripple/net/impl/RPCCall.cpp
@@ -1369,7 +1369,6 @@ rpcCmdLineToJson (std::vector<std::string> const& args,
     }
 
     JLOG (j.trace()) << "RPC Request: " << jvRequest << std::endl;
-
     return jvRequest;
 }
 
@@ -1457,9 +1456,7 @@ rpcClient(std::vector<std::string> const& args,
                 jvRequest["admin_password"] = setup.client.admin_password;
 
             if (jvRequest.isObject())
-            {
                 jvParams.append (jvRequest);
-            }
             else if (jvRequest.isArray())
             {
                 for (Json::UInt i = 0; i < jvRequest.size(); ++i)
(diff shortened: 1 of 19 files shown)


MarkusTeufelberger:
How about a versioning scheme similar to Kubernetes objects where also a readiness is encoded (Alpha, Beta...)? Just an integer is a bit weak...

intelliot:
In my opinion, any API version that appears in a stable release of rippled is “ready”. I don’t think there’s a need for the additional complexity that comes with alpha/beta distinctions.

src/ripple/rpc/impl/RPCHelpers.h
```diff
@@ -114,10 +114,34 @@ parseRippleLibSeed(Json::Value const& params);
...
+ * Requests with a version number in the range
+ * [APIVersionRangeLow, APIVersionRangeHigh] are supported.
+ *
+ * Network Requests without explicit version numbers use
+ * APIVersionIfUnspecified.
+ * Command line Requests use APIVersionCommandLine.
+ */
+
+constexpr unsigned APIInvalidVersion = 0;
+constexpr unsigned APIFirstVersion = 1;
```
nbougalis: We only use this here to set `APIVersionIfUnspecified`. I'd remove it, and just set `APIVersionIfUnspecified` to be equal to `APIVersionSupportedRangeLow`.

src/ripple/rpc/impl/RPCHelpers.h
```diff
@@ -114,10 +114,34 @@ parseRippleLibSeed(Json::Value const& params);
...
+constexpr unsigned APIVersionSupportedRangeLow = 1;
+constexpr unsigned APIVersionSupportedRangeHigh = 1;
+constexpr unsigned APIVersionIfUnspecified = APIFirstVersion;
+constexpr unsigned APIVersionCommandLine = APIVersionSupportedRangeHigh;
+constexpr unsigned APIVersionRippledStartUp = APIVersionSupportedRangeHigh;
```
nbougalis: Considering that these two values are always likely to use a fixed version of the API, I wonder if defining values for them makes sense. We should _probably_ just use `RPC::APIVersionSupportedRangeLow` directly in the places where this is used and remove `APIVersionCommandLine` and `APIVersionRippledStartUp`.

src/ripple/rpc/impl/Handler.cpp
```diff
@@ -148,45 +154,59 @@ class HandlerTable {
...
-    std::map<std::string, Handler> table_;
+    std::array<std::map<std::string, Handler>, APINumberVersionSupported> table_;
 
     template <class HandlerImpl>
-    void addHandler()
+    void addHandler(unsigned version)
     {
-        assert (table_.find(HandlerImpl::name()) == table_.end());
+        assert (version >= RPC::APIVersionSupportedRangeLow && version <= RPC::APIVersionSupportedRangeHigh);
+        auto & innerTable = table_[versionToIndex(version)];
```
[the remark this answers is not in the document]
pwang200: I like "empty" table[0]. Probably not too much cost, but when we increase the value of APIVersionSupportedRangeLow, there will be more "empty" ones. Say APIVersionSupportedRangeLow==4, table[0], table[1], table[2], and table[3] will be empty.

src/ripple/rpc/impl/RPCHelpers.h
pwang200: -- I will remove "APIFirstVersion". As you said, it is only for setting (actually documenting) APIVersionIfUnspecified. 
-- APIVersionIfUnspecified should always be 1. Since all the RPC requests with a version >= 2 must explicitly specify the version in the requests. If agreed, I will add a comment in code to explain.

src/ripple/rpc/impl/RPCHelpers.h
```diff
@@ -114,10 +114,34 @@ parseRippleLibSeed(Json::Value const& params);
...
+ * [APIVersionRangeLow, APIVersionRangeHigh] are supported.
+ *
+ * Network Requests without explicit version numbers use
+ * APIVersionIfUnspecified.
+ * Command line Requests use APIVersionCommandLine.
+ */
+
+constexpr unsigned APIInvalidVersion = 0;
+constexpr unsigned APIFirstVersion = 1;
+constexpr unsigned APIVersionSupportedRangeLow = 1;
```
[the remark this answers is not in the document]
pwang200: ApiMinimumSupportedVersion and ApiMaximumSupportedVersion are better names. I will change to them. 

I used "ApiDefaultVersion" originally in the code too. But I feel we have two "Default Versions", (1) a network RPC request without specify a version which should default to version 1, (2) a command line request which I think should default to ApiMaximumSupportedVersion.

src/ripple/rpc/impl/RPCHelpers.h
pwang200: I will remove APIVersionCommandLine and APIVersionRippledStartUp. 

Just to confirm, we will use APIVersionSupportedRangeLow as the command line request version. I can see the benefit of both "high" and "low.

src/ripple/rpc/impl/Handler.cpp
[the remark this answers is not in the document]
nbougalis: Ah, hmm... there's that, which I didn't think about. This may be a better fit for a map, but that comes at a minor performance cost.

I think this is fine as is, for now and let's see if anyone else has a better idea. Thanks.

src/ripple/rpc/impl/RPCHelpers.h
nbougalis: I can argue for either low or high. I'm fine with using whatever is appropriate and leave that up to your judgement.

src/ripple/rpc/impl/RPCHelpers.h
nbougalis: I think that's fine. Thanks!

src/ripple/rpc/impl/RPCHelpers.cpp
```diff
@@ -716,5 +716,23 @@ beast::SemanticVersion const firstVersion("1.0.0");
...
+        {
+            unsigned apiVersion = jv[jss::api_version].asInt();
+            if(apiVersion >= RPC::APIVersionSupportedRangeLow &&
+               apiVersion <= RPC::APIVersionSupportedRangeHigh)
+            {
+                return apiVersion;
+            }
+        }
+        return RPC::APIInvalidVersion;
+    } else
```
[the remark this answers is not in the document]
nbougalis: Actually my simplifcation doesn't _quite_ work. It needs a further tweak.

pwang200:
Thanks, I will work on it.

src/ripple/rpc/impl/RPCHelpers.h
```diff
@@ -131,6 +163,20 @@ setVersion(Object& parent)
 std::pair<RPC::Status, LedgerEntryType>
     chooseLedgerEntryType(Json::Value const& params);
 
+/**
+ * Retrieve the api version number from the json value
+ *
+ * Note that APIInvalidVersion will be returned if
+ * 1) the version number field has a wrong format
+ * 2) the version number retrieved is out of the supported range
+ * 3) the version number is unspecified and
```
mellery451: seems like case 3 should be covered by the static checks that @nbougalis suggested above

src/ripple/rpc/impl/ServerHandlerImp.cpp
```diff
@@ -598,11 +610,40 @@ ServerHandlerImp::processRequest (Port const& port,
             continue;
         }
 
+        auto apiVersion = RPC::APIVersionIfUnspecified;
+        if (jsonRPC.isMember(jss::params) &&
+            jsonRPC[jss::params].isArray() &&
+            jsonRPC[jss::params].size() > 0 &&
+            jsonRPC[jss::params][Json::UInt(0)].isObject())
```
mellery451: `jsonRPC[jss::params][0u]....` might be little more concise (in general for indexing of json arrays with static values)

src/ripple/rpc/impl/RPCHelpers.h
```diff
@@ -114,10 +114,34 @@ parseRippleLibSeed(Json::Value const& params);
...
+constexpr unsigned APIInvalidVersion = 0;
+constexpr unsigned APIFirstVersion = 1;
+constexpr unsigned APIVersionSupportedRangeLow = 1;
+constexpr unsigned APIVersionSupportedRangeHigh = 1;
+constexpr unsigned APIVersionIfUnspecified = APIFirstVersion;
+constexpr unsigned APIVersionCommandLine = APIVersionSupportedRangeHigh;
+constexpr unsigned APIVersionRippledStartUp = APIVersionSupportedRangeHigh;
+constexpr unsigned APINumberVersionSupported = APIVersionSupportedRangeHigh -
+                                               APIVersionSupportedRangeLow + 1;
+
```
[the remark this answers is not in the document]
pwang200: Added two:
```
static_assert (ApiMinimumSupportedVersion >= APIVersionIfUnspecified);
static_assert (ApiMaximumSupportedVersion >= ApiMinimumSupportedVersion);
```

src/ripple/rpc/impl/RPCHelpers.cpp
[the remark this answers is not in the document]
pwang200: I modified it and added a unit test.

src/ripple/rpc/Context.h
```diff
@@ -55,6 +55,7 @@ struct Context
     LedgerMaster& ledgerMaster;
     Resource::Consumer& consumer;
     Role role;
+    unsigned apiVersion;
```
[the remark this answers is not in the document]
pwang200: Thanks. I changed `unsigned` to `unsigned int` in all files in this PR.

src/ripple/rpc/impl/Handler.cpp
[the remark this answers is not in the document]
pwang200: OK

src/ripple/rpc/impl/RPCHelpers.h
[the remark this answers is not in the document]
pwang200: `ApiMinimumSupportedVersion` and `ApiMaximumSupportedVersion` are changed.
`APIVersionIfUnspecified` is not changed.

src/ripple/rpc/impl/RPCHelpers.h
pwang200: Removed `APIFirstVersion`.
`APIVersionIfUnspecified` is set to 1.

src/ripple/rpc/impl/RPCHelpers.h
pwang200: I removed `APIVersionCommandLine` and `APIVersionRippledStartUp`, and used "high" for command line requests' versions.

src/test/rpc/RPCCall_test.cpp
```diff
@@ -77,6 +79,7 @@ static RPCCallTestData const rpcCallTestArray [] =
     "method" : "account_channels",
     "params" : [
       {
+         "api_version" : 1,
```
mellery451: if I understand correctly, this version text/field is always replaced with the max supported (?). If that's the case, I would prefer just a token string here (e.g. `%MAX_API_VER%`....) and then you search/replace that token in the update method (it gives a better sense of what's actually happening). If I've misunderstood what's going on, please ignore.

src/ripple/rpc/impl/RPCHelpers.h
pwang200: You are right. But I think we need both the static check (checking we don't make a mistake in coding) and the check in the function (checking network requests are acceptable).

src/ripple/net/impl/RPCCall.cpp
```diff
@@ -1353,6 +1354,20 @@ rpcCmdLineToJson (std::vector<std::string> const& args,
...
 
+    auto insert_api_version = [](Json::Value & jr){
+        if(!jr.isMember(jss::error) && !jr.isMember(jss::api_version))
+            jr[jss::api_version] = RPC::ApiMaximumSupportedVersion;
+    };
+
+    if(jvRequest.isObject())
+        insert_api_version(jvRequest);
+    else if(jvRequest.isArray()) {
+        for (Json::UInt j = 0; j < jvRequest.size(); ++j) {
```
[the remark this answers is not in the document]
pwang200: Thanks, I changed to `std::for_each(jvRequest.begin(), jvRequest.end(), insert_api_version);`

I modified `insert_api_version()` to `include isObject()`. I think that is still needed. Otherwise, `Value::operator[]` could be called on a non-object.

src/test/rpc/RPCCall_test.cpp
pwang200: Great, thanks! I changed the code.

src/ripple/rpc/impl/ServerHandlerImp.cpp
pwang200: Thanks! I changed the code.

scottschurr:
@pwang200, FWIW, I'm recusing myself from this code review.

I have misgivings about the on-going indefinite maintenance effort that this change represents versus the benefit to users.  Based on the conversations I've seen in other pull requests, I do not expect my concerns to be considered seriously at this point in time.  I find myself unable to focus only on whether "an implementation has successfully realized its associated design" when I have reservations about the design itself.

Rather than increase noise with no expected change in outcome I'll simply recuse myself.

intelliot:
> Note that the documentation is for informational only. We will consolidate the documentation and create another PR after both API Versioning and gRPC are merged.

@pwang200 Has the consolidated documentation been created yet?

PR #3161, opened 2019-11-20
Improve Json::Value:

scottschurr:
We use `Json::Value` pretty hard.  I recently discovered that it is not very well tested.  So I added tests.

In the process of testing I discovered a few interfaces that are not used.  I removed those.  If we haven't used them yet we're unlikely to need them.

I also found a feature that was seriously broken: `resize()`.  It turned out that `resize()` was used in a couple of places, but it did not do what was expected.  So I removed those two calls to `resize()` (which were unnecessary) and also removed the `resize()` method.

### DIFF
--- a/src/ripple/json/json_value.h
+++ b/src/ripple/json/json_value.h
@@ -290,6 +290,7 @@ public:
     ///
     /// Equivalent to jsonvalue[jsonvalue.size()] = value;
     Value& append ( const Value& value );
+    Value& append ( Value&& value );
 
     /// Access an object value by name, create a null member if it does not exist.
     Value& operator[] ( const char* key );
--- a/src/ripple/json/json_value.h
+++ b/src/ripple/json/json_value.h
@@ -46,17 +46,9 @@ enum ValueType
     objectValue    ///< object value (collection of name/value pairs).
 };
 
-enum CommentPlacement
-{
-    commentBefore = 0,        ///< a comment placed on the line before a value
-    commentAfterOnSameLine,   ///< a comment just after a value on the same line
-    commentAfter,             ///< a comment on the line after a value (only make sense for root value)
-    numberOfCommentPlacement
-};
-
 /** \brief Lightweight wrapper to tag static string.
  *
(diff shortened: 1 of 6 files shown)


src/ripple/json/json_value.h
```diff
@@ -284,13 +270,6 @@ class Value
...
-    /// Resize the array to size elements.
-    /// New elements are initialized to null.
-    /// May only be called on nullValue or arrayValue.
-    /// \pre type() is arrayValue or nullValue
-    /// \post type() is arrayValue
-    void resize ( UInt size );
-
     /// Access an array element (zero based index ).
     /// If the array contains less than index element, then null value are inserted
     /// in the array so that its size is index+1.
```
jwbusch: Another small part that could be improved is `append`, it only has one overload -> `append(Value const&)` but I've seen pieces of code that call it as `append(std::move(val))` meaning they're expecting an overload like `append(Value&&)`

That's declared on line 292, but I can't put a comment there :)

HowardHinnant: +1  This would actually get used a lot, right within the json lib itself.  Adding such a signature with `= delete` gives a preview of the first places the compiler finds where to use it.

src/test/json/json_value_test.cpp
```diff
@@ -30,6 +31,548 @@ namespace ripple {
 
 struct json_value_test : beast::unit_test::suite
 {
+    void test_StaticString()
```
HowardHinnant: Nice test additions!

src/ripple/json/json_value.h
vinniefalco: Coming Soon https://github.com/vinniefalco/json

scottschurr: Right, @vinniefalco, I'm aware that's on the horizon.  And, once it's available, I anticipate we'll adopt it.  So I'm intentionally limiting these changes to be focused on stability and removing broken / unused / untested interfaces.

PR #3163, opened 2019-11-26
Proposed 1.4.0

nbougalis:
Closes #3162


mellery451:
The xcode-9 failure can be safely ignored - it is a known homebrew/travis issue and will be fixed in the first 1.5 beta. All other variants are passing.

PR #3171, opened 2019-12-02
[TRIVIAL] Update comments in TER.h

intelliot:
The web address has changed (page moved under xrpl.org)

`ripple-binary-codec` currently hardcodes the int values. If the numbers here are changed, `ripple-binary-codec` should be updated to match.

### DIFF
--- a/src/ripple/protocol/TER.h
+++ b/src/ripple/protocol/TER.h
@@ -1,7 +1,7 @@
 //------------------------------------------------------------------------------
 /*
     This file is part of rippled: https://github.com/ripple/rippled
-    Copyright (c) 2012 - 2018 Ripple Labs Inc.
+    Copyright (c) 2012 - 2019 Ripple Labs Inc.
 
     Permission to use, copy, modify, and/or distribute this software for any
     purpose  with  or without fee is hereby granted, provided that the above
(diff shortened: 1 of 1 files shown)


scottschurr:
Thanks for updating the comments, @intelliot.  Your changes look good.  However, I'm thinking they may not go far enough.  The codec's dependency on specific numeric values suggests that rippled developers need to not be so fast and loose redefining these error codes.

So I'm going to suggest that we modify TER.h further by making the numeric values of all TER codes explicit.  And we should add a unit test that will help flag if any of the pre-existing codes change in the future.  I have a commit you could cherry-pick if you like that idea.  You'll find the commit here: https://github.com/scottschurr/rippled/commit/c80f535ba0cae2856c964ad13f480fa5be7a60c4

If you choose to cherry-pick that commit, you should add another reviewer, because I'll no longer be an impartial observer.

nbougalis:
>The codec's dependency on specific numeric values suggests that rippled developers need to not be so fast and loose redefining these error codes.

I think that we should consider generating these from a single source of truth: a file that gets processed and can produce both C++, JS and anything else we desire.

MarkusTeufelberger:
Maybe something that describes a buffer... for a protocol... ;-)

PR #3174, opened 2019-12-04
Do not restart server on deadlock detection:

seelabs:
Replace the logic error with a log message. It's possible an overloaded job
queue is causing false alarms on the deadlock detector.

### DIFF
--- a/src/ripple/app/main/LoadManager.cpp
+++ b/src/ripple/app/main/LoadManager.cpp
@@ -145,9 +145,12 @@ void LoadManager::run ()
                     {
                         JLOG(journal_.fatal())
                             << "Deadlock detected. Deadlocked time: "
-                            << timeSpentDeadlocked.count()
-                            << "s JobQueue Overloaded: "
-                            << app_.getJobQueue().isOverloaded();
+                            << timeSpentDeadlocked.count() << "s";
+                        if (app_.getJobQueue().isOverloaded())
+                        {
+                            JLOG(journal_.fatal())
+                                << app_.getJobQueue().getJson(0);
+                        }
                     }
                 }
 
@@ -158,10 +161,12 @@ void LoadManager::run ()
                 if (timeSpentDeadlocked >= deadlockLogicErrorTimeLimit)
                 {
                     JLOG(journal_.fatal())
-                            << "LogicError: Deadlock detected. Deadlocked time: "
-                            << timeSpentDeadlocked.count()
-                            << "s JobQueue Overloaded: "
-                            << app_.getJobQueue().isOverloaded();
+                        << "LogicError: Deadlock detected. Deadlocked time: "
+                        << timeSpentDeadlocked.count() << "s";
+                    if (app_.getJobQueue().isOverloaded())
+                    {
+                        JLOG(journal_.fatal()) << app_.getJobQueue().getJson(0);
+                    }
                     LogicError("Deadlock detected");
                 }
             }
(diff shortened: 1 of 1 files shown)


src/ripple/app/main/LoadManager.cpp
```diff
@@ -129,23 +129,32 @@ void LoadManager::run ()
...
+                        JLOG(journal_.fatal())
+                            << "Deadlock detected. Deadlocked time: "
+                            << timeSpentDeadlocked.count()
+                            << "s JobQueue Overloaded: "
+                            << app_.getJobQueue().isOverloaded();
+                    }
                 }
 
-                // If we go over 90 seconds spent deadlocked, it means that
+                // If we go over the deadlockTimeLimit spent deadlocked, it means that
```
miguelportilla: Nit: Is the comment better placed at the fatal log?

miguelportilla: We may also want to move the assert to line 145 eg. `assert(false)`

seelabs: @miguelportilla I think I'd rather keep the assert outside the reporting interval. But I don't really object if you want to move it.

src/ripple/app/main/LoadManager.cpp
```diff
@@ -129,23 +129,32 @@ void LoadManager::run ()
...
+                        JLOG(journal_.warn())
+                            << "Server stalled for "
+                            << timeSpentDeadlocked.count() << " seconds.";
+                    }
+                    else
+                    {
+                        JLOG(journal_.fatal())
+                            << "Deadlock detected. Deadlocked time: "
+                            << timeSpentDeadlocked.count()
+                            << "s JobQueue Overloaded: "
```
scottschurr: I think what @miguelportilla had in mind was, if the job queue is overloaded then write the job queue status json at fatal level.  Maybe something like this:
```
        if (app_.getJobQueue().isOverloaded())
        {
            JLOG(journal_.fatal()) << app_.getJobQueue().getJson(0);
        }
```

src/ripple/app/main/LoadManager.cpp
```diff
@@ -129,23 +129,32 @@ void LoadManager::run ()
...
+                            << app_.getJobQueue().isOverloaded();
```
miguelportilla: Unlikely but I wanted to note that if changes ever make `deadlockTimeLimit` not divisible by `reportingIntervalSeconds`, we could potentially miss a fatal log. Say for instance that `reportingIntervalSeconds` is 12, and a beat is heard at 95 seconds.

src/ripple/app/main/LoadManager.cpp
miguelportilla: @scottschurr Indeed.

src/ripple/app/main/LoadManager.cpp
miguelportilla: @seelabs I don't feel strongly about either one and okay as is.

src/ripple/app/main/LoadManager.cpp
scottschurr: I was thinking the same thing about the modulus.  But, yes, I also think it's unlikely.  This particular thread wakes itself up once every second using the `alertable_sleep_until` on line 189.

src/ripple/app/main/LoadManager.cpp
scottschurr: @seelabs, what do you think about revising the logging in this way?

seelabs: @miguelportilla @scottschurr Fixed in `b9b524917  [FOLD] Log more info about the job queue when overloaded`

src/ripple/app/main/LoadManager.cpp
seelabs: It's a good thought. I suppose we could keep track of last reported time and check if some duration has passed. But let's not do that in this PR; I'd like to keep this one relatively minimal.

PR #3179, opened 2019-12-05
Added support for statsD Traffic Counts reporting

mayurbhandary:
Traffic counts metrics summed across all categories and exported. These metrics include Bytes In Bytes Out, Messages In, and Messages Out.

### DIFF
--- a/src/ripple/app/main/Application.cpp
+++ b/src/ripple/app/main/Application.cpp
@@ -1463,7 +1463,7 @@ bool ApplicationImp::setup()
     //             if (!config_.standalone())
     m_overlay = make_Overlay (*this, setup_Overlay(*config_), *m_jobQueue,
         *serverHandler_, *m_resourceManager, *m_resolver, get_io_service(),
-        *config_);
+        *config_, m_collectorManager->collector ());
     add (*m_overlay); // add to PropertyStream
 
     if (!config_->standalone())
--- a/src/ripple/overlay/impl/OverlayImpl.cpp
+++ b/src/ripple/overlay/impl/OverlayImpl.cpp
@@ -156,7 +156,7 @@ OverlayImpl::OverlayImpl (
     , serverHandler_(serverHandler)
     , m_resourceManager (resourceManager)
     , m_peerFinder (PeerFinder::make_Manager (*this, io_service,
-        stopwatch(), app_.journal("PeerFinder"), config))
+        stopwatch(), app_.journal("PeerFinder"), config, collector))
     , m_resolver (resolver)
     , next_id_(1)
     , timer_count_(0)
--- a/src/ripple/overlay/impl/OverlayImpl.cpp
+++ b/src/ripple/overlay/impl/OverlayImpl.cpp
@@ -144,7 +144,8 @@ OverlayImpl::OverlayImpl (
     Resource::Manager& resourceManager,
     Resolver& resolver,
     boost::asio::io_service& io_service,
-    BasicConfig const& config)
+    BasicConfig const& config,
+    beast::insight::Collector::ptr const& collector)
     : Overlay (parent)
     , app_ (app)
     , io_service_ (io_service)
@@ -159,6 +160,7 @@ OverlayImpl::OverlayImpl (
     , m_resolver (resolver)
     , next_id_(1)
     , timer_count_(0)
+    , m_stats(std::bind (&OverlayImpl::collect_metrics, this),collector)
 {
     beast::PropertyStream::Source::add (m_peerFinder.get());
 }
@@ -1334,10 +1336,11 @@ make_Overlay (
     Resource::Manager& resourceManager,
     Resolver& resolver,
     boost::asio::io_service& io_service,
-    BasicConfig const& config)
+    BasicConfig const& config,
+    beast::insight::Collector::ptr const& collector)
 {
     return std::make_unique<OverlayImpl>(app, setup, parent, serverHandler,
-        resourceManager, resolver, io_service, config);
+        resourceManager, resolver, io_service, config, collector);
 }
 
 }
--- a/src/ripple/overlay/impl/OverlayImpl.cpp
+++ b/src/ripple/overlay/impl/OverlayImpl.cpp
@@ -168,9 +168,9 @@ OverlayImpl::OverlayImpl (
             std::vector<TrafficGauges> ret;
             ret.reserve(counts.size());
 
-            for (size_t i =0; i < counts.size(); ++i)
+            for (size_t i = 0; i < counts.size(); ++i)
             {
-                ret.push_back(TrafficGauges(counts[i].name,collector));
+                ret.push_back(TrafficGauges(counts[i].name, collector));
             }
 
             return ret;
--- a/src/ripple/overlay/impl/OverlayImpl.cpp
+++ b/src/ripple/overlay/impl/OverlayImpl.cpp
@@ -165,15 +165,12 @@ OverlayImpl::OverlayImpl (
         collector,
         [counts = m_traffic.getCounts(), collector]()
         {
-            std::vector<beast::insight::Gauge> ret;
-            ret.reserve(counts.size() * 4);
+            std::vector<TrafficGauges> ret;
+            ret.reserve(counts.size());
 
-            for (auto const& s : counts)
+            for (size_t i =0; i < counts.size(); ++i)
             {
-                ret.push_back(collector->make_gauge(s.name, "Bytes_In"));
-                ret.push_back(collector->make_gauge(s.name, "Bytes_Out"));
-                ret.push_back(collector->make_gauge(s.name, "Messages_In"));
-                ret.push_back(collector->make_gauge(s.name, "Messages_Out"));
+                ret.push_back(TrafficGauges(counts[i].name,collector));
             }
 
             return ret;
--- a/src/ripple/overlay/impl/OverlayImpl.cpp
+++ b/src/ripple/overlay/impl/OverlayImpl.cpp
@@ -160,7 +160,24 @@ OverlayImpl::OverlayImpl (
     , m_resolver (resolver)
     , next_id_(1)
     , timer_count_(0)
-    , m_stats(std::bind (&OverlayImpl::collect_metrics, this),collector)
+    , m_stats (
+        std::bind(&OverlayImpl::collect_metrics, this),
+        collector,
+        [counts = m_traffic.getCounts(), collector]()
+        {
+            std::vector<beast::insight::Gauge> ret;
+            ret.reserve(counts.size() * 4);
+
+            for (auto const& s : counts)
+            {
+                ret.push_back(collector->make_gauge(s.name, "Bytes_In"));
+                ret.push_back(collector->make_gauge(s.name, "Bytes_Out"));
+                ret.push_back(collector->make_gauge(s.name, "Messages_In"));
+                ret.push_back(collector->make_gauge(s.name, "Messages_Out"));
+            }
+
+            return ret;
+        }())
 {
     beast::PropertyStream::Source::add (m_peerFinder.get());
 }
--- a/src/ripple/overlay/impl/OverlayImpl.h
+++ b/src/ripple/overlay/impl/OverlayImpl.h
@@ -465,14 +465,14 @@ private:
         Stats (Handler const& handler, beast::insight::Collector::ptr const& collector)
             : hook (collector->make_hook (handler))
             , peerDisconnects (collector->make_gauge("Overlay","Peer_Disconnects"))
-            , m_collector(collector) 
+            , collector(collector) 
             { 
             }
 
         beast::insight::Hook hook;
         beast::insight::Gauge peerDisconnects;
         std::vector<beast::insight::Gauge> trafficGauges;
-        beast::insight::Collector::ptr m_collector;
+        beast::insight::Collector::ptr collector;
         
     };
     
@@ -488,13 +488,13 @@ private:
             for (auto const& s : stats)
             {   
                 m_stats.trafficGauges.push_back(
-                    beast::insight::Gauge (m_stats.m_collector->make_gauge(s.name,"Bytes_In")));
+                    beast::insight::Gauge (m_stats.collector->make_gauge(s.name,"Bytes_In")));
                 m_stats.trafficGauges.push_back(
-                    beast::insight::Gauge (m_stats.m_collector->make_gauge(s.name,"Bytes_Out")));
+                    beast::insight::Gauge (m_stats.collector->make_gauge(s.name,"Bytes_Out")));
                 m_stats.trafficGauges.push_back(
-                    beast::insight::Gauge (m_stats.m_collector->make_gauge(s.name,"Messages_In")));
+                    beast::insight::Gauge (m_stats.collector->make_gauge(s.name,"Messages_In")));
                 m_stats.trafficGauges.push_back(
-                    beast::insight::Gauge (m_stats.m_collector->make_gauge(s.name,"Messages_Out")));
+                    beast::insight::Gauge (m_stats.collector->make_gauge(s.name,"Messages_Out")));
             }
         }
 
--- a/src/ripple/overlay/impl/OverlayImpl.h
+++ b/src/ripple/overlay/impl/OverlayImpl.h
@@ -458,6 +458,22 @@ private:
     sendEndpoints();
 
 private:
+
+    struct TrafficGauges{
+        TrafficGauges (char const* name, beast::insight::Collector::ptr const& collector)
+        : bytesIn(collector->make_gauge(name,"Bytes_In"))
+        , bytesOut(collector->make_gauge(name,"Bytes_Out"))
+        , messagesIn(collector->make_gauge(name,"Messages_In"))
+        , messagesOut(collector->make_gauge(name,"Messages_Out"))
+        {   
+        }
+        beast::insight::Gauge bytesIn;
+        beast::insight::Gauge bytesOut;
+        beast::insight::Gauge messagesIn;
+        beast::insight::Gauge messagesOut;
+    };
+
+    
     struct Stats
     {
 
@@ -465,7 +481,7 @@ private:
         Stats (
                 Handler const& handler, 
                 beast::insight::Collector::ptr const& collector,
-                std::vector<beast::insight::Gauge>&& trafficGauges_)
+                std::vector<TrafficGauges>&& trafficGauges_)
             : peerDisconnects (collector->make_gauge("Overlay","Peer_Disconnects"))
             , trafficGauges (trafficGauges_)
             , hook (collector->make_hook (handler))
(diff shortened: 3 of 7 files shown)


src/ripple/app/main/Application.cpp
```diff
@@ -1463,7 +1463,7 @@ bool ApplicationImp::setup()
     //             if (!config_.standalone())
     m_overlay = make_Overlay (*this, setup_Overlay(*config_), *m_jobQueue,
         *serverHandler_, *m_resourceManager, *m_resolver, get_io_service(),
-        *config_);
+        *config_, m_collectorManager->collector ());
```
nbougalis: Technically you don't need this: `Application` (which you pass in as `this`) already exposes `getCollectorManager`.

src/ripple/overlay/impl/OverlayImpl.cpp
```diff
@@ -159,6 +160,7 @@ OverlayImpl::OverlayImpl (
     , m_resolver (resolver)
     , next_id_(1)
     , timer_count_(0)
+    , m_stats(std::bind (&OverlayImpl::collect_metrics, this),collector)
```
nbougalis: There's an alternative here which allows you to use (the heretofore unused) `Application::getCollectorManager` function:

```
    , m_stats(
        std::bind(&OverlayImpl::collect_metrics, this),
        app_.getCollectorManager().collector())
```

This _slightly_ simplifies `OverlayImpl` construction, but I'm not sure if it's a net gain. All these are codepaths we call once and only once. It also doesn't help that we never use `Application::getCollectorManager` anywhere currently, which means we could (_should_?) remove it, slimming down `Application`.

mayurbhandary: I think we should remove getCollectorManager since it's unused. If we decided to start using it, we would need to make changes wherever we pass the collector in through the constructor. I'm happy to do that, but I don't think it would really be a net gain.

src/ripple/overlay/impl/OverlayImpl.h
```diff
@@ -456,6 +456,41 @@ class OverlayImpl : public Overlay
...
+
+    Stats m_stats;
+
+private:
+    void collect_metrics()
+    {   
+        std::lock_guard lock (mutex_);
+        auto const stats = m_traffic.getCounts();
+        for (auto const& i : stats)
+        {
```
nbougalis: This PR has an issue: it never reports the _name_ of the counters. It merely coagulates all different categories into a single `struct`.

We need one set of counters for every message, which complicates things slighty. Consider changing `m_stats` to be `std::vector<Stats>` and adding a `name` to the `Stats` object.

mayurbhandary: Thanks, I'm on it.

mayurbhandary: @nbougalis I made the changes you requested.

src/ripple/overlay/impl/TrafficCount.h
```diff
@@ -223,6 +229,51 @@ class TrafficCount
         { "getobject (get)" },                                    // category::get_hash
         { "unknown" }                                             // category::unknown
     }};
+
+    std::array<std::string, category::unknown+1> monNames_
```
mayurbhandary: @nbougalis I had to make a new array with StatsD-friendly names corresponding to each original category name. I tried my best to keep them similar to the original names. Let me know if you had something else in mind.

src/ripple/overlay/impl/OverlayImpl.h
```diff
@@ -456,6 +456,60 @@ class OverlayImpl : public Overlay
...
+                    beast::insight::Gauge (m_stats.mcollector->make_gauge(i.name,"Bytes_In")));
+                m_stats.trafficGauges.push_back(
+                    beast::insight::Gauge (m_stats.mcollector->make_gauge(i.name,"Bytes_Out")));
+                m_stats.trafficGauges.push_back(
+                    beast::insight::Gauge (m_stats.mcollector->make_gauge(i.name,"Messages_In")));
+                m_stats.trafficGauges.push_back(
+                    beast::insight::Gauge (m_stats.mcollector->make_gauge(i.name,"Messages_Out")));
+            }
+        }
+
```
mayurbhandary: It is not ideal for trafficGauges to be partially initialized in the constructor and completed here. Since this is a nested struct and m_traffic is a member of the parent class, we can't do `auto const stats = m_traffic.getCounts();` in the constructor. Another option might be to have a copy of the array of categories in the Stats struct. Please advise if there are any alternative workarounds. 

For reference, this is the original array of TrafficStats objects with category names:
https://github.com/ripple/rippled/blob/90d9ca901d06b9da1653526c234bc4b00e27070b/src/ripple/overlay/impl/TrafficCount.h#L182

src/ripple/overlay/impl/OverlayImpl.h
```diff
@@ -456,6 +456,60 @@ class OverlayImpl : public Overlay
...
+private:
+    void collect_metrics()
+    {   
+        std::lock_guard lock (mutex_);
+
+        auto const stats = m_traffic.getCounts();
+
+        if(m_stats.trafficGauges.empty()){
+            for (auto const& i : stats)
```
thejohnfreeman: This variable is called `i`. It's not an integer, which is what most people expect an `i` to be. There is an integer named `i` declared just after this block; two variables with the same name very close in scope but with different types is inviting confusion. The same variable in the same iteration pattern just below is named `s`; can we name this one `s` too?

src/ripple/overlay/impl/TrafficCount.h
```diff
@@ -178,50 +178,51 @@ class TrafficCount
...
+        {"getobject_Transaction_node_get"},                  // category::get_hash_txnode
+        {"getobject_Account_State_node_share"},              // category::share_hash_asnode
+        {"getobject_Account_State_node_get"},                // category::get_hash_asnode
+        {"getobject_CAS_share"},                             // category::share_cas_object
+        {"getobject_CAS_get"},                               // category::get_cas_object
+        {"getobject_Fetch_Pack_share"},                      // category::share_fetch_pack
+        {"getobject_Fetch Pack_get"},                        // category::get_fetch_pack
+        {"getobject_share"},                                 // category::share_hash
+        {"getobject_get"},                                   // category::get_hash
+        {"unknown"}                                          // category::unknown
```
thejohnfreeman: Storing these names in the `TrafficStats` object is causing trouble: it has two constructors, neither of which is default, and there's a `std::string` being hauled around in every `TrafficStats` object, wasting space. I searched and no one is ever passing around a single `TrafficStats` object. They are written by their index (via the enum `category`), and only ever read while iterating through the whole collection.

If that is true, then we could pull the name out of `TrafficStats`, and use the index during iteration to access a `constexpr array` of the names, stored as `char const *` instead of `std::string`. The `TrafficStats` class can be given a default constructor. The `counts_` array can delegate to the default constructor by just initializing it with the size.

At the very least, the `name` member of `TrafficStats` can be changed to `char const *`.

src/ripple/overlay/impl/TrafficCount.h
```diff
@@ -178,50 +178,51 @@ class TrafficCount
...
+        {"shards"},                                          // category::shards
+        {"set_get"},                                         // category::get_set
+        {"set_share"},                                       // category::share_set
+        {"ledger_data_Transaction_Set_candidate_get"},       // category::ld_tsc_get
+        {"ledger_data_Transaction_Set_candidate_share"},     // category::ld_tsc_share
+        {"ledger_data_Transaction_Node_get"},                // category::ld_txn_get
+        {"ledger_data_Transaction_Node_share"},              // category::ld_txn_share
+        {"ledger_data_Account_State_Node_get"},              // category::ld_asn_get
+        {"ledger_data_Account_State_Node_share"},            // category::ld_asn_share
+        {"}ledger_data_get"},                                 // category::ld_get
```
thejohnfreeman: `}ledger_data_get` probably needs that leading curly removed.

src/ripple/overlay/impl/TrafficCount.h
```diff
@@ -178,50 +178,51 @@ class TrafficCount
         return counts_;
     }
 
+
 protected:
     std::array<TrafficStats, category::unknown + 1> counts_
```
thejohnfreeman: A `TrafficCount` object holds many `TrafficStats` objects under a member named `counts_`. This is incredibly confusing naming that we inherited. I would love to see it changed. Here's one idea:

`TrafficCount` becomes `TrafficCounters`, `TrafficStats` becomes `TrafficCounter`, and `counts_` becomes `counters_`. This borrows the "counter" metric name from StatsD, and uses a plural vs. singular distinction to reflect that one is a collection of the other.

src/ripple/overlay/impl/OverlayImpl.h
```diff
@@ -456,6 +456,60 @@ class OverlayImpl : public Overlay
...
+        Stats (Handler const& handler, beast::insight::Collector::ptr const& collector)
+            : hook (collector->make_hook (handler))
+            , peerDisconnects (collector->make_gauge("Overlay","Peer_Disconnects")) 
+            { 
+                mcollector=collector;
+            }
+
+        beast::insight::Hook hook;
+        beast::insight::Gauge peerDisconnects;
+        std::vector<beast::insight::Gauge> trafficGauges;
```
thejohnfreeman: I know you didn't choose the name of `beast::insight::Gauge`, but can we call this member `trafficCounters` instead of `trafficGauges`? Going by StatsD definitions, these are counters, not gauges. I'd love to see `beast::insight::Gauge` renamed to `Counter` as well, if it is in fact a counter and not a gauge, but I haven't looked.

src/ripple/overlay/impl/OverlayImpl.h
```diff
@@ -456,6 +456,60 @@ class OverlayImpl : public Overlay
...
+        template <class Handler>
+        Stats (Handler const& handler, beast::insight::Collector::ptr const& collector)
+            : hook (collector->make_hook (handler))
+            , peerDisconnects (collector->make_gauge("Overlay","Peer_Disconnects")) 
+            { 
+                mcollector=collector;
+            }
+
+        beast::insight::Hook hook;
+        beast::insight::Gauge peerDisconnects;
```
thejohnfreeman: Why not add a `peer_disconnects` value to the `enum TrafficCounter::category` and merge this member into `trafficGauges`?

src/ripple/overlay/impl/OverlayImpl.h
thejohnfreeman: Can we make this an `array` instead of a `vector`?

src/ripple/overlay/impl/TrafficCount.h
mayurbhandary: I think TrafficStats is a more appropriate name given its members. BytesIn, BytesOut, MessagesIn, and MessagesOut. I do agree that TrafficCount should reflect that it contains a collection of metrics.

src/ripple/overlay/impl/OverlayImpl.h
mayurbhandary: Keeping it a vector to make initialization easier.

src/ripple/overlay/impl/OverlayImpl.h
mayurbhandary: I believe that enum is only for TrafficStats. Each category in that enum has a bytesIn, bytesOut, messagesIn, and messagesOut. It made sense to me to keep this one separate.

src/ripple/overlay/impl/OverlayImpl.h
mayurbhandary: Since the value we are exporting is already an aggregate statistic I am exporting it as a gauge so that the metrics are not double counted. I understand that it is confusing to call this a gauge, but I will be sure to document the rationale for this in the monitoring tool.

src/ripple/overlay/impl/TrafficCount.h
mayurbhandary: Thanks, I changed the `name` member to a `char const* `

src/ripple/overlay/impl/OverlayImpl.h
mayurbhandary: Thanks, I made these changes.

src/ripple/overlay/impl/OverlayImpl.h
```diff
@@ -456,6 +456,60 @@ class OverlayImpl : public Overlay
...
+private:
+    struct Stats
+    {
+
+        template <class Handler>
+        Stats (Handler const& handler, beast::insight::Collector::ptr const& collector)
+            : hook (collector->make_hook (handler))
+            , peerDisconnects (collector->make_gauge("Overlay","Peer_Disconnects")) 
+            { 
+                mcollector=collector;
```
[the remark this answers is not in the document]
mayurbhandary: Good point. I made it match the style. Thanks.

src/ripple/overlay/impl/OverlayImpl.h
```diff
@@ -456,6 +456,60 @@ class OverlayImpl : public Overlay
...
+            }
+
+        beast::insight::Hook hook;
+        beast::insight::Gauge peerDisconnects;
+        std::vector<beast::insight::Gauge> trafficGauges;
+        beast::insight::Collector::ptr mcollector;
```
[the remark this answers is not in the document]
mayurbhandary: You're right, `collector` makes more sense. I changed it to `collector`. Thanks.

src/ripple/overlay/impl/OverlayImpl.h
```diff
@@ -456,6 +456,44 @@ class OverlayImpl : public Overlay
...
+                beast::insight::Collector::ptr const& collector,
+                std::vector<beast::insight::Gauge>&& trafficGauges_)
+            : peerDisconnects (collector->make_gauge("Overlay","Peer_Disconnects"))
+            , trafficGauges (trafficGauges_)
+            , hook (collector->make_hook (handler))
+            { 
+            }
+            
+        beast::insight::Gauge peerDisconnects;
+        std::vector<beast::insight::Gauge> trafficGauges;
```
jwbusch: I think this is better off as something like 
```
struct TrafficGauges {
    Gauge bytesIn;
    Gauge bytesOut;
    Gauge messagesIn;
    Gauge messagesOut;
};

std::vector<TrafficGauges> trafficGauges;
```

That way they're grouped w/ names instead of implicitly every 4 elements.

src/ripple/overlay/impl/OverlayImpl.cpp
```diff
@@ -155,10 +156,28 @@ OverlayImpl::OverlayImpl (
...
     , m_resolver (resolver)
     , next_id_(1)
     , timer_count_(0)
+    , m_stats (
+        std::bind(&OverlayImpl::collect_metrics, this),
+        collector,
+        [counts = m_traffic.getCounts(), collector]()
+        {
+            std::vector<beast::insight::Gauge> ret;
+            ret.reserve(counts.size() * 4);
```
jwbusch: This would be updated/simpler w/ the new struct as well.

src/ripple/overlay/impl/OverlayImpl.h
```diff
@@ -458,39 +458,56 @@ class OverlayImpl : public Overlay
...
 private:
     void collect_metrics()
     {   
-        std::lock_guard lock (mutex_);
-        std::size_t i = 0;
+        std::lock_guard lock (m_statsMutex);
+        auto counts= m_traffic.getCounts();
 
-        for (auto const& s : m_traffic.getCounts())
+        for (std::size_t i = 0; i < counts.size(); ++i)
```
jwbusch: nit: we should add an assert here that `assert(counts.size() == m_stats.trafficGauges.size())` We know it is the same from the constructor, but it's just good to put those assertions into code.

src/ripple/overlay/impl/OverlayImpl.cpp
```diff
@@ -165,15 +165,12 @@ OverlayImpl::OverlayImpl (
         collector,
         [counts = m_traffic.getCounts(), collector]()
         {
-            std::vector<beast::insight::Gauge> ret;
-            ret.reserve(counts.size() * 4);
+            std::vector<TrafficGauges> ret;
+            ret.reserve(counts.size());
 
-            for (auto const& s : counts)
+            for (size_t i =0; i < counts.size(); ++i)
```
jwbusch: nit: spacing/formatting throughout these files. `i = 0` and `counts[i].name, collector` etc. (spaces after commas, spaces in-between operator= operands)

mayurbhandary: Fixed, thanks!

src/ripple/overlay/impl/OverlayImpl.h
mayurbhandary: Added, thanks!

src/ripple/overlay/impl/OverlayImpl.h
```diff
@@ -456,6 +456,44 @@ class OverlayImpl : public Overlay
...
+    struct Stats
+    {
+
+        template <class Handler>
+        Stats (
+                Handler const& handler, 
+                beast::insight::Collector::ptr const& collector,
+                std::vector<beast::insight::Gauge>&& trafficGauges_)
+            : peerDisconnects (collector->make_gauge("Overlay","Peer_Disconnects"))
+            , trafficGauges (trafficGauges_)
```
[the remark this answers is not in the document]
mayurbhandary: Added, thanks!

src/ripple/overlay/impl/OverlayImpl.h
```diff
@@ -456,6 +456,63 @@ class OverlayImpl : public Overlay
...
+    };
+    
+    Stats m_stats;
+    std::mutex m_statsMutex;
+
+private:
+    void collect_metrics()
+    {   
+        std::lock_guard lock (m_statsMutex);
+        auto counts = m_traffic.getCounts();
```
jwbusch: nit: I think you can grab `counts` before the lock. (m_statsMutex is needed to protect m_stats, but not m_traffic).

src/ripple/overlay/impl/OverlayImpl.h
```diff
@@ -456,6 +456,63 @@ class OverlayImpl : public Overlay
...
+
+        for (std::size_t i = 0; i < counts.size(); ++i)
+        {
+            assert(counts.size() == m_stats.trafficGauges.size());
```
jwbusch: nit: can move this assert to a few lines up, outside of the loop (it only needs to be checked once)

mayurbhandary: Thanks, I fixed it.

src/ripple/overlay/impl/OverlayImpl.h
mayurbhandary: Thanks, fixed.

src/ripple/overlay/impl/OverlayImpl.h
nbougalis: :vulcan_salute:

src/ripple/overlay/impl/OverlayImpl.h
```diff
@@ -456,6 +456,62 @@ class OverlayImpl : public Overlay
 
     void
     sendEndpoints();
+
+private:
+
+    struct TrafficGauges{
```
nbougalis: Micro-nit: this is, technically, a `TrafficGauge` not `TrafficGauges`. I don't think this is a particularly big deal, but I think the change is basically a search-and-replace so it shouldn't be a big deal.

MarkusTeufelberger: Isn't it 4 gauges (bytes/messages in/out) in one structure?

mayurbhandary: @nbougalis Since there are multiple gauges in the struct, shouldn't it remain TrafficGauges?

PR #3183, opened 2019-12-11
Improve platform detection and reduce includes:

nbougalis:
The existing platform detection code was derived from the old **Beast** library, which was, itself, derived from [JUCE](https://juce.com/).

This commit removes that code and replaces it with the **`Boost.Predef`** library which defines a consistent set of compiler, architecture, operating system, library, and other version numbers.

For more on **`Boost.Predef`**, please see the Boost documentation. The documentation for the current version as of this writing is at: https://www.boost.org/doc/libs/1_71_0/doc/html/predef.html


seelabs:
1000 lines removed? Love it! I'll take a look tomorrow, but it looks like there are some legit compile error (easy to solve I'm sure). Here's one I saw:
```
../../src/test/beast/beast_CurrentThreadName_test.cpp:44:13: error: no viable conversion from 'const std::__cxx11::basic_string<char>' to 'bool'

        if (initialThreadName)

            ^~~~~~~~~~~~~~~~~

/usr/bin/../lib/gcc/x86_64-linux-gnu/9/../../../../include/c++/9/bits/basic_string.h:816:7: note: candidate function

      operator __sv_type() const noexcept

      ^

1 error generated.
```

src/ripple/crypto/impl/RFC1751.cpp
```diff
@@ -393,25 +393,22 @@ int RFC1751::wsrch (std::string const& strWord, int iMin, int iMax)
...
 {
-    int i, p, v, l;
-    char b[9];
-
     if (6 != vsHuman.size ())
         return -1;
 
-    memset (b, 0, sizeof (b));
+    int i, p = 0;
+    char b[9] = { 0 };
```
seelabs: I'd write `char b[9]{};` for things like this (it's fine as-is, I'm just not used to seeing `= {0}`.

scottschurr: I think there's a difference between what `{}` and `{ 0 }` produce.  I should test this before writing this comment of course, but if I'm remembering correctly...

- `{}` will default initialize all nine members of the array, and
- `{ 0 }` will only initialize the first member of the array.

So, if that's correct, `{}` is more conservative and potentially better for long term maintenance.  For example, if you initialize only the first member of the `char` array, then it _might_ be undefined behavior if someone reads an element of the array other than the first.

Explaining the _might_: I'm pretty sure it would not be undefined for an array of `std::byte` or `unsigned char`; it's possible that `char` is okay?  I don't have a copy of the standard with me right now...

But `{ 0 }` is less expensive if we're counting cycles.

seelabs: I spoke with Scott and Howard this morning, and `char b[9] = {0};` and `char b[9]{}` both do the same thing, so it is fine as-is. The rule is that the elements without an initial value will be zero initialized, so `char b[9] = {1,2}` sets the first two elements and the rest will be zero. However, I do think this is a more obscure rule than `char b[9]{};`. Neither Scott S. nor I knew that off the top of our heads (Howard did of course :-)

src/ripple/app/tx/impl/Transactor.cpp
```diff
@@ -635,7 +635,7 @@ Transactor::operator()()
 {
     JLOG(j_.trace()) << "apply: " << ctx_.tx.getTransactionID ();
 
-#ifdef BEAST_DEBUG
+#ifdef RIPPLED_DEBUG
```
mellery451: I don't think this macro is actually defined anywhere ? We already have `DEBUG` and `_DEBUG`, but you can add `RIPPLED_DEBUG` here: https://github.com/ripple/rippled/blob/e76f19a6715335bfac2a8f13ba45843cb4869ec2/Builds/CMake/RippledCompiler.cmake#L19

src/ripple/crypto/impl/RFC1751.cpp
nbougalis: That's interesting. I thought that this `b[9]{}` was more obscure; I knew that would "default-initialize" but I didn't know off-hand if "default" in this case means 0, but I knew that `b[9] = { /* something */ };` initializes the first element to "something" and zero-initializes the rest.

I'm fine with doing `char b[9]{}`

src/ripple/app/tx/impl/Transactor.cpp
nbougalis: I'll use `DEBUG` instead.

src/ripple/app/main/Main.cpp
```diff
@@ -37,30 +37,39 @@
...
+#if !BOOST_OS_LINUX && !BOOST_OS_WINDOWS && !BOOST_OS_MACOS
+#error Supported platforms are: Linux, Windows and MacOS
+#endif
+
+// Ensure that precisely one platform is detected.
+#if (BOOST_OS_LINUX   && (BOOST_OS_WINDOWS || BOOST_OS_MACOS)) || \
+    (BOOST_OS_MACOS   && (BOOST_OS_WINDOWS || BOOST_OS_LINUX)) || \
+    (BOOST_OS_WINDOWS && (BOOST_OS_LINUX   || BOOST_OS_MACOS))
+#error Multiple supported platforms appear active at once
+#endif
```
ximinez: At the risk of bikeshedding, what do you think about
```
constexpr int boost_os_check = []() {
    size_t check{0};
#if BOOST_OS_LINUX
    ++check;
#endif
#if BOOST_OS_WINDOWS
    ++check;
#endif
#if BOOST_OS_MACOS
    ++check;
#endif
    return check;
}();
static_assert(
    boost_os_check > 0,
    "Supported platforms are: Linux, Windows and MacOS");
static_assert(
    boost_os_check < 2,
    "Multiple supported platforms appear active at once");
```
It would make adding future platforms a little bit easier.

nbougalis: Looks reasonable, but if it's all the same to you, I think I'll punt on this for now.

ximinez: Not a problem.

PR #3188, opened 2019-12-16
Fix node auto-configuration code

nbougalis:
The `node_size` configuration option is used to automatically configure various parameters (cache sizes, timeouts, etc) for the server.

A previous commit included changes that caused incorrect values to be returned which can result in sub-optimal performance that can manifest as difficulty syncing to the network, or increased disk I/O and/or memory usage. The problem was introduced with commit 66fad62.

This commit, if merged, fixes the code to ensure that the correct values are returned and introduces a compile-time check to prevent this issue from reoccurring.

### DIFF
--- a/src/ripple/app/main/BasicApp.cpp
+++ b/src/ripple/app/main/BasicApp.cpp
@@ -24,20 +24,22 @@ BasicApp::BasicApp(std::size_t numberOfThreads)
 {
     work_.emplace (io_service_);
     threads_.reserve(numberOfThreads);
+
     while(numberOfThreads--)
-        threads_.emplace_back(
-            [this, numberOfThreads]()
+    {
+        threads_.emplace_back([this, numberOfThreads]()
             {
-                beast::setCurrentThreadName(
-                    std::string("io_service #") +
-                        std::to_string(numberOfThreads));
-                this->io_service_.run();
+              beast::setCurrentThreadName("io svc #" +
+                  std::to_string(numberOfThreads));
+              this->io_service_.run();
             });
+    }
 }
 
 BasicApp::~BasicApp()
 {
     work_ = boost::none;
-    for (auto& _ : threads_)
-        _.join();
+
+    for (auto& t : threads_)
+        t.join();
 }
--- a/src/ripple/core/Config.h
+++ b/src/ripple/core/Config.h
@@ -29,6 +29,7 @@
 #include <boost/filesystem.hpp> // VFALCO FIX: This include should not be here
 #include <boost/lexical_cast.hpp>
 #include <boost/optional.hpp>
+#include <algorithm>
 #include <cstdint>
 #include <map>
 #include <string>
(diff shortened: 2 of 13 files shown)


src/ripple/core/Config.h
```diff
@@ -202,19 +226,35 @@ class Config : public BasicConfig
...
-        assert(item < sizedItems.size());
-        return getSize(item, NODE_SIZE);
+        auto const index = static_cast<std::underlying_type_t<SizedItem>>(item);
+        assert(index < sizedItems.size());
+
+        if (!node)
+            node = NODE_SIZE;
+
+        assert(*node <= 4);
+        return sizedItems.at(index).second.at(*node);
```
miguelportilla: Consider...

{
auto const index = static_cast<std::underlying_type_t<SizedItem>>(item); 
assert(index < sizedItems.size());
       
assert(!node || *node <= 4);
return sizedItems.at(index).second.at(node.value_or(NODE_SIZE));
}

src/ripple/app/main/BasicApp.cpp
```diff
@@ -24,20 +24,22 @@ BasicApp::BasicApp(std::size_t numberOfThreads)
...
             {
-                beast::setCurrentThreadName(
-                    std::string("io_service #") +
-                        std::to_string(numberOfThreads));
-                this->io_service_.run();
+              beast::setCurrentThreadName("io svc #" +
+                  std::to_string(numberOfThreads));
+              this->io_service_.run();
             });
+    }
```
miguelportilla: Nit: Threads are named in reverse order to how they are emplaced and run. I am okay as is but it could be to:
```
    for(std::size_t i = 1; i <= numberOfThreads; ++i)
    {
        threads_.emplace_back([this, i]()
            {
              beast::setCurrentThreadName("io svc #" + std::to_string(i));
              this->io_service_.run();
            });
    }
```

src/ripple/core/impl/Config.cpp
```diff
@@ -330,13 +378,13 @@ void Config::loadFromString (std::string const& fileContents)
             NODE_SIZE = 4;
         else
         {
-            NODE_SIZE = beast::lexicalCastThrow <int> (strTemp);
+            NODE_SIZE = beast::lexicalCastThrow <std::size_t>(strTemp);
```
mellery451: use `std::clamp` here?

src/ripple/core/impl/Config.cpp
```diff
@@ -330,13 +378,13 @@ void Config::loadFromString (std::string const& fileContents)
...
 
-            if (NODE_SIZE < 0)
-                NODE_SIZE = 0;
-            else if (NODE_SIZE > 4)
+            if (NODE_SIZE > 4)
                 NODE_SIZE = 4;
         }
+
+        assert (NODE_SIZE <= 4);
```
mellery451: I don't see any way for this assert to fire...am I missing something?

mellery451:
windows debug build unit test are crashing (repeatable), so I think that needs investigation.

nbougalis:
> windows debug build unit test are crashing (repeatable), so I think that needs investigation.

Hmm. OK, I'll take a closer look, thanks! Probably something silly.

src/ripple/core/impl/Config.cpp
nbougalis: Good idea.

src/ripple/core/impl/Config.cpp
nbougalis: Not anymore. I'll remove it.

src/ripple/core/impl/Config.cpp
nbougalis: Used `std::min` instead.

PR #3189, opened 2019-12-20
Report StateAccounting metrics and fixed StatsD bug

mayurbhandary:
StateAccounting metrics report the amount of time the server spends in each state and the number of transitions into that state. 

A StatsD reporting bug that reports gauges as counters was also fixed in this PR.

### DIFF
--- a/src/ripple/app/misc/NetworkOPs.cpp
+++ b/src/ripple/app/misc/NetworkOPs.cpp
@@ -638,11 +638,11 @@ private:
             , syncing_duration(collector->make_gauge("State_Accounting", "Syncing_duration"))
             , tracking_duration (collector->make_gauge("State_Accounting", "Tracking_duration"))
             , full_duration (collector-> make_gauge("State_Accounting", "Full_duration"))
-            , disconnected_trasitions (collector->make_gauge("State_Accounting","Disconnected_trasitions"))
-            , connected_trasitions (collector->make_gauge("State_Accounting","Connected_trasitions"))
-            , syncing_trasitions(collector->make_gauge("State_Accounting", "Syncing_trasitions"))
-            , tracking_trasitions (collector->make_gauge("State_Accounting", "Tracking_trasitions"))
-            , full_trasitions (collector-> make_gauge("State_Accounting", "Full_trasitions"))
+            , disconnected_transitions (collector->make_gauge("State_Accounting","Disconnected_transitions"))
+            , connected_transitions (collector->make_gauge("State_Accounting","Connected_transitions"))
+            , syncing_transitions(collector->make_gauge("State_Accounting", "Syncing_transitions"))
+            , tracking_transitions (collector->make_gauge("State_Accounting", "Tracking_transitions"))
+            , full_transitions (collector-> make_gauge("State_Accounting", "Full_transitions"))
             { }
 
         beast::insight::Hook hook;
@@ -652,11 +652,11 @@ private:
         beast::insight::Gauge tracking_duration;
         beast::insight::Gauge full_duration;
 
-        beast::insight::Gauge disconnected_trasitions;
-        beast::insight::Gauge connected_trasitions;
-        beast::insight::Gauge syncing_trasitions;
-        beast::insight::Gauge tracking_trasitions;
-        beast::insight::Gauge full_trasitions;
+        beast::insight::Gauge disconnected_transitions;
+        beast::insight::Gauge connected_transitions;
+        beast::insight::Gauge syncing_transitions;
+        beast::insight::Gauge tracking_transitions;
+        beast::insight::Gauge full_transitions;
     };
 
     Stats m_stats;
@@ -675,11 +675,11 @@ private:
         m_stats.tracking_duration.set(counters[static_cast<std::size_t>(OperatingMode::TRACKING)].dur.count());
         m_stats.full_duration.set(counters[static_cast<std::size_t>(OperatingMode::FULL)].dur.count());
 
-        m_stats.disconnected_trasitions.set(counters[static_cast<std::size_t>(OperatingMode::DISCONNECTED)].transitions);
-        m_stats.connected_trasitions.set(counters[static_cast<std::size_t>(OperatingMode::CONNECTED)].transitions);
-        m_stats.syncing_trasitions.set(counters[static_cast<std::size_t>(OperatingMode::SYNCING)].transitions);
-        m_stats.tracking_trasitions.set(counters[static_cast<std::size_t>(OperatingMode::TRACKING)].transitions);
-        m_stats.full_trasitions.set(counters[static_cast<std::size_t>(OperatingMode::FULL)].transitions);
+        m_stats.disconnected_transitions.set(counters[static_cast<std::size_t>(OperatingMode::DISCONNECTED)].transitions);
+        m_stats.connected_transitions.set(counters[static_cast<std::size_t>(OperatingMode::CONNECTED)].transitions);
+        m_stats.syncing_transitions.set(counters[static_cast<std::size_t>(OperatingMode::SYNCING)].transitions);
+        m_stats.tracking_transitions.set(counters[static_cast<std::size_t>(OperatingMode::TRACKING)].transitions);
+        m_stats.full_transitions.set(counters[static_cast<std::size_t>(OperatingMode::FULL)].transitions);
     }
 };
 
--- a/src/ripple/app/misc/NetworkOPs.cpp
+++ b/src/ripple/app/misc/NetworkOPs.cpp
@@ -170,9 +170,16 @@ class NetworkOPsImp final
          */
         StateCountersJson json() const;
 
-        auto getCounterData() const {
+
+        struct CounterData {
+            decltype(counters_) counters;
+            decltype(mode_) mode;
+            decltype(start_) start;
+        };
+
+        CounterData getCounterData() const{ 
             std::lock_guard lock(mutex_);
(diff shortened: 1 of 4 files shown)


src/ripple/app/misc/NetworkOPs.cpp
```diff
@@ -169,6 +169,25 @@ class NetworkOPsImp final
          * @return JSON object.
          */
         StateCountersJson json() const;
+
+        std::array<Counters, 5> getCounters(){
```
jwbusch: nit: just make this return type `decltype(counters_)`

src/ripple/app/misc/NetworkOPs.cpp
```diff
@@ -619,6 +639,63 @@ class NetworkOPsImp final
...
+
+    Stats m_stats;
+
+private:
+    void collect_metrics()
+    {   
+        auto counters = accounting_.getCounters();
+        auto start = accounting_.getStart();
+        auto mode = accounting_.getMode();
+
```
jwbusch: You could use a small lambda to clean-up the casts. 
```
auto getCounters = [&counters_](auto mode) -> auto& {
    return counters_[static_cast<std::size_t>(mode)];
}
```
Then you can use it like...
```
getCounter(mode).dur += current;
m_stats.disconnected_duration.set(getCounter(OperatingMode::Disconnected).dur.count());
```

Not sure how much better that really is though, I'll let you decide. I think we try to keep the columns <= 80 chars.

src/ripple/beast/insight/impl/StatsDCollector.cpp
```diff
@@ -589,7 +589,7 @@ void StatsDGaugeImpl::flush ()
         ss <<
             m_impl->prefix() << "." <<
             m_name << ":" <<
-            m_value << "|c" <<
+            m_value << "|g" <<
```
jwbusch: Just curious, what's the difference between `c` and `g` here?

src/ripple/app/misc/NetworkOPs.cpp
```diff
@@ -619,6 +639,63 @@ class NetworkOPsImp final
...
+        auto const current = std::chrono::duration_cast<std::chrono::microseconds>(std::chrono::system_clock::now() - start);
+        counters[static_cast<std::size_t>(mode)].dur += current;
+
+        m_stats.disconnected_duration.set(counters[static_cast<std::size_t>(OperatingMode::DISCONNECTED)].dur.count());
```
jwbusch: does `m_stats` or `accounting_` need to be under a lock? Seems like they might need to be, although I'm not sure when this callback is triggered (`beast::insight::Hook`?)

src/ripple/app/misc/NetworkOPs.cpp
```diff
@@ -169,6 +169,25 @@ class NetworkOPsImp final
          * @return JSON object.
          */
         StateCountersJson json() const;
+
+        std::array<Counters, 5> getCounters(){
+            std::lock_guard lock(mutex_);
+            auto counters = counters_;
+            return counters;
```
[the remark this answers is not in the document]
mayurbhandary: Thanks, that's a good point. I will redo it with a tuple. I had initially thought I might want to use these getters individually later.

src/ripple/beast/insight/impl/StatsDCollector.cpp
mayurbhandary: The g gets parsed as a gauge and the c gets parsed as a counter by the statsD collector.

jwbusch: Ah, so this was just a bug fix :)

mayurbhandary: Just a small bug :D

src/ripple/app/misc/NetworkOPs.cpp
mayurbhandary: This call back is triggered once a second via the StatsDCollector. See here:

https://github.com/ripple/rippled/blob/90d9ca901d06b9da1653526c234bc4b00e27070b/src/ripple/beast/insight/impl/StatsDCollector.cpp#L418

From what I can tell, I don't think they need to be under a lock.

src/ripple/app/misc/NetworkOPs.cpp
```diff
@@ -619,6 +626,61 @@ class NetworkOPsImp final
...
+        template <class Handler>
+        Stats (Handler const& handler, beast::insight::Collector::ptr const& collector)
+            : hook (collector->make_hook (handler))
+            , disconnected_duration (collector->make_gauge("State_Accounting","Disconnected_duration"))
+            , connected_duration (collector->make_gauge("State_Accounting","Connected_duration"))
+            , syncing_duration(collector->make_gauge("State_Accounting", "Syncing_duration"))
+            , tracking_duration (collector->make_gauge("State_Accounting", "Tracking_duration"))
+            , full_duration (collector-> make_gauge("State_Accounting", "Full_duration"))
+            , disconnected_trasitions (collector->make_gauge("State_Accounting","Disconnected_trasitions"))
+            , connected_trasitions (collector->make_gauge("State_Accounting","Connected_trasitions"))
```
mellery451: `transitions` is misspelled in both the member and gauge names.

src/ripple/app/misc/NetworkOPs.cpp
```diff
@@ -619,6 +626,61 @@ class NetworkOPsImp final
...
+
+    Stats m_stats;
+
+private:
+    void collect_metrics()
+    {     
+        auto [counters, mode, start] = accounting_.getCounterData();
+
+        auto const current = std::chrono::duration_cast<std::chrono::microseconds>(std::chrono::system_clock::now() - start);
+        counters[static_cast<std::size_t>(mode)].dur += current;
```
mellery451: can you explain why you need to update the local copy of data in `counters` here? Since this data is being obtained from the `StateAccounting` instance, I would have expected its duration data to be up-to-date.

src/ripple/app/misc/NetworkOPs.cpp
mayurbhandary: Oops, thanks.

src/ripple/app/misc/NetworkOPs.cpp
mayurbhandary: The current mode's duration is not up to date. It has to be manually incremented when observed. Here's another spot where we do the same thing:

https://github.com/ripple/rippled/blob/90d9ca901d06b9da1653526c234bc4b00e27070b/src/ripple/app/misc/NetworkOPs.cpp#L3490

In this case, we are observing and updating the duration of the current mode before returning a JSON object.

mellery451: yes - but in your case, you are updating a local *copy* of the array (`getCounterData` returns a tuple that has a copy of the data). Is that what you want, or do you want to update the array stored in `accounting_` ?

mayurbhandary: Yes, I want to update the local copy and not the copy in accounting_.

src/ripple/app/misc/NetworkOPs.cpp
```diff
@@ -169,6 +169,12 @@ class NetworkOPsImp final
          * @return JSON object.
          */
         StateCountersJson json() const;
```
jwbusch: nit: I think I saw that `json` member function could use the new `getCounterData`

src/ripple/app/misc/NetworkOPs.cpp
```diff
@@ -169,6 +169,12 @@ class NetworkOPsImp final
...
+
+        auto getCounterData() const {
```
jwbusch: I think I'd prefer the version with the 

```
struct CounterData {
    decltype(counters_) counters;
    decltype(mode_) mode;
    decltype(start_) start;
};
```

Then just `return { counters_, mode_, start_ };`

I realized that `tuple/pair` is great for generic programming, but if you already know the types and can give them reasonable names, then it's better to just declare new struct. Makes it easier to read.

src/ripple/app/misc/NetworkOPs.cpp
jwbusch: So currently `accounting_` is safe, since it protects it's members, and `m_stats` is only modified here. But you don't control when it's called/from what thread, so it might be better to protect `m_stats` as well just to protect against future changes.

src/ripple/app/misc/NetworkOPs.cpp
nbougalis: I'm confused. Can you explain _why_?

mayurbhandary: I want to update the local copy because it is stale when I get it (the duration of the _current_ state in `accounting_` is stale therefore the local copy is stale too). This is because the durations of each mode in `accounting_` are only updated when the state changes and a new mode is set see [here](https://github.com/ripple/rippled/blob/cc4cefaa4b42c82da4f249eb3f54a77d2bcdc2d6/src/ripple/app/misc/NetworkOPs.cpp#L3464).  I could have also updated the array stored in `accounting_`, but it wouldn't benefit us because other places in the code similarly make a local copy and update it to get the _most_ up-to-date duration. An example if this is [here](https://github.com/ripple/rippled/blob/cc4cefaa4b42c82da4f249eb3f54a77d2bcdc2d6/src/ripple/app/misc/NetworkOPs.cpp#L3488). If I updated the duration of the current state in the array in `accounting_`, it would immediately become stale and need to be updated when observed again anyways.

Issue #2983, opened 2019-06-21
Command line parser issues


ximinez:
1. Some of the commands which use the `parseAccountItems` and `parseAccountCurrencies` functions to interpret the command line options are too restrictive. Specifically `account_info` and `account_currencies` are limited to 2 params max, when they should allow three. (account, ledger, "strict").
2. Not all of the commands which use the above functions are documented as allowing the `"strict"` parameter. Audit.
3. A command that returns the rpc error `rpcBAD_SYNTAX` does not generate any output. Fix `fromCommandLine` so that it does.

ximinez:
Addressed by #3196 (which was merged by #3259). Commit message didn't auto-close this issue. Oops.

PR #3196, opened 2019-12-30
Command line parser issues

movitto:
This pull request addresses issue #2983 with the following changes:

- Expand the signatures of cli commands corresponding to parseAccountItems and parseAccountCurrencies to accept a 3rd 'strict' parameter
- Remove the edge case omitting output in the case of bad syntax, now output is always produced

As far as the 2nd bullet point in issue #2983, the aforementioned commands were inspected and found to output documentation pertaining to the 'strict' parameter as expected (see attached screenshot). If I missed something please let me know and I can look into further.

rippled-commands-strict

### DIFF
--- a/Builds/CMake/deps/Protobuf.cmake
+++ b/Builds/CMake/deps/Protobuf.cmake
@@ -10,6 +10,7 @@ if (static)
 endif ()
 find_package (Protobuf 3.8)
 if (local_protobuf OR NOT Protobuf_FOUND)
+  include (GNUInstallDirs)
   message (STATUS "using local protobuf build.")
   if (WIN32)
     # protobuf prepends lib even on windows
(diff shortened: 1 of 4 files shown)


nbougalis:
LGTM :+1:

ximinez:
`account_offers` calls the same parsing functions as `account_info`, etc., and so can take the `strict` parameter. Could you update the usage to indicate that? (`owner_info` also can take `strict`, but it's not deprecated, so don't add it.)

movitto:
@ximinez updated the account_offers usage output to reflect support for the 'strict' param & updated parseAccountItems comments pertaining to other relavant commands. Let me know if there is anything else!

scottschurr:
It would be good to update RPCCall_test.cpp.

For example with this pull request the comment here needs to be updated: https://github.com/ripple/rippled/blob/develop/src/test/rpc/RPCCall_test.cpp#L472

And here's a test case that maybe should work but doesn't: https://github.com/ripple/rippled/blob/develop/src/test/rpc/RPCCall_test.cpp#L490

movitto:
@scottschurr Fixed the broken tests and updated the outdated comments

movitto:
@scottschurr cherry picked and tried it out, LGTM

mDuo13:
For the `strict` parameter, what do you type on the commandline to set it to true? Is it `1`, `true` or the word `strict`?

(I'm just curious, as this has been a point of confusion on past commandline syntax and I don't have a clear answer on which one you should use.)

scottschurr:
@mDuo13, for the command line you type the word "strict" I believe.  @movitto, please correct me if I'm wrong.

movitto:
@scottschurr yes on the command line "strict" (by itself) should be specified

scottschurr:
@mDuo13, I haven't seen any objections from you yet.  So you're good with this pull request?  If so I think it can be marked `PASSED`

mDuo13:
:+1:

scottschurr:
Sorry, my mistake.  All of the CI build targets that run the regular unit tests are showing unit test failures.  Here's the text of the failures as I'm seeing them for both clang and gcc:
```
#14 failed: account_currencies: current ledger.: RPCCall_test.cpp(320)
#25 failed: account_info: with ledger index and strict.: RPCCall_test.cpp(529)
#252 failed: owner_info: with ledger index and strict.: RPCCall_test.cpp(5032)
#257 failed: owner_info: invalid ledger selection and strict.: RPCCall_test.cpp(5131)
```
We need to understand the reason for the unit test failure first.  Assuming the test failures are caused by something benign, once the unit tests are patched up I think this pull request will be good to go.  Thanks for being patient.

movitto:
@scottschurr it's hard to determine from the CI engine why exactly those tests are failing, would it be possible to enable more verbose output?

mellery451:
you might want to try running the unit tests locally (`rippled --unittest` does the trick), but FWIW the location of the failures is indicated by SOURCE(line_number) here: https://travis-ci.com/ripple/rippled/jobs/275448524#L914-L917

movitto:
@mellery451 ok thanks, I was able to reproduce the error locally. @scottschurr pushed a followup patch fixing the issue

scottschurr:
🎉 Thanks for the persistence @movitto.  Looks like we're ready for all of this to be squashed.

Issue #3099, opened 2019-10-07
Allow server to report its current manifest


nbougalis:
We need `rippled` to improve its manifest support:

1) A `rippled` instance configured as a validator should offer an admin-only RPC command that can report the master public key, the ephemeral public key, the manifest sequence number and, if present, the domain associated with the manifest that the server is using. The command can be `validator_info`. If the server is not configured as a validator, the command should report an error.
2) Any `rippled` instance should offer an RPC command that, given a master or ephemeral public key, should report the master public key, the ephemeral public key, the manifest sequence number the domain associated with the given manifest (if present), and the manifest in base64-encoded format. The command can be `manifest`. If the server does not have the given manifest, the command should report an error.

Optionally, for (1) a `--validator-info` command line should parse the config file, load the **`validator_token`** if one is present, and report the master and ephemeral public keys, the manifest sequence number and, if present, the domain associated with the manifest. Additionally, the command should report the base64 encoded manifest. Once completed, the program would terminate (similar to, e.g. `--version`)

intelliot:
1. [`validator_info`](https://xrpl.org/validator_info.html) has been implemented
2. [`manifest`](https://xrpl.org/manifest.html) has been implemented*

*(`manifest` was implemented as an admin-only command, but I think it would be fine to make it public)

I'm not certain of the status of the command line feature idea, but if it is still needed, we should create a new issue for it. The title of this issue (`Allow server to report its current manifest`) is done.

PR #3197, opened 2019-12-30
New validator_info and manifest RPC methods

movitto:
This pull request addresses issue #3099 by adding two new RPC methods to allow the inspection of validator state.

- validator_info will return information pertaining to the current validators keys, manifest sequence, and domain

- manifest looks up manifest information for the specified key (either master or ephemeral)

Utility methods have been added to other modules to support this functionality

### DIFF
--- a/Builds/CMake/RippledCore.cmake
+++ b/Builds/CMake/RippledCore.cmake
@@ -650,6 +650,7 @@ else ()
     src/ripple/rpc/handlers/LedgerRequest.cpp
     src/ripple/rpc/handlers/LogLevel.cpp
     src/ripple/rpc/handlers/LogRotate.cpp
+    src/ripple/rpc/handlers/Manifest.cpp
     src/ripple/rpc/handlers/NoRippleCheck.cpp
     src/ripple/rpc/handlers/OwnerInfo.cpp
     src/ripple/rpc/handlers/PathFind.cpp
@@ -676,6 +677,7 @@ else ()
     src/ripple/rpc/handlers/ValidationCreate.cpp
     src/ripple/rpc/handlers/ValidatorListSites.cpp
     src/ripple/rpc/handlers/Validators.cpp
+    src/ripple/rpc/handlers/ValidatorInfo.cpp
     src/ripple/rpc/handlers/WalletPropose.cpp
     src/ripple/rpc/impl/DeliveredAmount.cpp
     src/ripple/rpc/impl/Handler.cpp
@@ -967,6 +969,7 @@ else ()
     src/test/rpc/LedgerData_test.cpp
     src/test/rpc/LedgerRPC_test.cpp
     src/test/rpc/LedgerRequestRPC_test.cpp
+    src/test/rpc/ManifestRPC_test.cpp
     src/test/rpc/NoRippleCheck_test.cpp
     src/test/rpc/NoRipple_test.cpp
     src/test/rpc/OwnerInfo_test.cpp
@@ -983,6 +986,7 @@ else ()
     src/test/rpc/TransactionEntry_test.cpp
     src/test/rpc/TransactionHistory_test.cpp
     src/test/rpc/Tx_test.cpp
+    src/test/rpc/ValidatorInfo_test.cpp
     src/test/rpc/ValidatorRPC_test.cpp
     src/test/rpc/Version_test.cpp
     #[===============================[
--- a/src/ripple/app/main/Application.h
+++ b/src/ripple/app/main/Application.h
@@ -53,6 +53,7 @@ class AcceptedLedger;
 class LedgerMaster;
 class LoadManager;
 class ManifestCache;
+class ValidatorKeys;
 class NetworkOPs;
 class OpenLedger;
 class OrderBookDB;
--- a/src/ripple/app/misc/Manifest.h
+++ b/src/ripple/app/misc/Manifest.h
@@ -257,6 +257,30 @@ public:
     PublicKey
     getMasterKey (PublicKey const& pk) const;
 
+    /** Returns master key's current manifest sequence.
+
+        @return sequence corresponding to Master public key
+          if configured or boost::none otherwise
+    */
+    boost::optional<std::uint32_t>
+    getSequence (PublicKey const& pk) const;
+
+    /** Returns domain claimed by a given public key
+
+        @return domain corresponding to Master public key
+          if present, otherwise boost::none
+    */
+    boost::optional<std::string>
+    getDomain (PublicKey const& pk) const;
+
+    /** Returns mainfest corresponding to a given public key
+
+        @return manifest corresponding to Master public key
+          if present, otherwise boost::none
+    */
+    boost::optional<std::string>
+    getManifest (PublicKey const& pk) const;
+
     /** Returns `true` if master key has been revoked in a manifest.
 
         @param pk Master public key
--- a/src/ripple/app/misc/impl/Manifest.cpp
+++ b/src/ripple/app/misc/impl/Manifest.cpp
@@ -309,6 +309,42 @@ ManifestCache::getMasterKey (PublicKey const& pk) const
     return pk;
 }
 
+boost::optional<std::uint32_t>
+ManifestCache::getSequence (PublicKey const& pk) const
+{
+    std::lock_guard lock{read_mutex_};
+    auto const iter = map_.find (pk);
+
+    if (iter != map_.end () && !iter->second.revoked ())
+        return iter->second.sequence;
+
+    return boost::none;
+}
+
+boost::optional<std::string>
+ManifestCache::getDomain (PublicKey const& pk) const
+{
+    std::lock_guard lock{read_mutex_};
+    auto const iter = map_.find (pk);
+
+    if (iter != map_.end () && !iter->second.revoked ())
+        return iter->second.domain;
+
+    return boost::none;
+}
+
+boost::optional<std::string>
+ManifestCache::getManifest (PublicKey const& pk) const
+{
+    std::lock_guard lock{read_mutex_};
+    auto const iter = map_.find (pk);
+
+    if (iter != map_.end () && !iter->second.revoked ())
+        return iter->second.serialized;
+
+    return boost::none;
+}
+
 bool
 ManifestCache::revoked (PublicKey const& pk) const
 {
--- a/src/ripple/net/impl/RPCCall.cpp
+++ b/src/ripple/net/impl/RPCCall.cpp
@@ -145,6 +145,21 @@ private:
         }
     }
 
+    static bool validPublicKey (std::string const& strPk)
+    {
+        if (parseBase58<PublicKey> (TokenType::AccountPublic, strPk))
+            return true;
+
+        auto pkHex = strUnHex (strPk);
+        if (!pkHex)
+            return false;
+
+        if (!publicKeyType(makeSlice(*pkHex)))
+            return false;
+
+        return true;
+    }
+
 private:
     using parseFuncPtr = Json::Value (RPCParser::*) (Json::Value const& jvParams);
 
@@ -201,6 +216,24 @@ private:
         v[jss::params] = params;
 
         return v;
+    }
+
+    Json::Value parseManifest (Json::Value const& jvParams)
+    {
+        if (jvParams.size () == 1)
+        {
+            Json::Value jvRequest (Json::objectValue);
+
+            std::string const strPk = jvParams[0u].asString ();
+            if (!validPublicKey (strPk))
+              return rpcError (rpcPUBLIC_MALFORMED);
+
+            jvRequest[jss::public_key] = strPk;
+
+            return jvRequest;
+        }
+
+        return rpcError (rpcINVALID_PARAMS);
     }
 
     // fetch_info [clear]
@@ -764,21 +797,7 @@ private:
     {
         std::string const strPk = jvParams[0u].asString ();
 
-        bool const validPublicKey = [&strPk]{
-            if (parseBase58<PublicKey> (TokenType::AccountPublic, strPk))
-                return true;
-
-            auto pkHex = strUnHex (strPk);
-            if (!pkHex)
-                return false;
-
-            if (!publicKeyType(makeSlice(*pkHex)))
-                return false;
-
-            return true;
-        }();
-
-        if (!validPublicKey)
+        if (!validPublicKey(strPk))
             return rpcError (rpcPUBLIC_MALFORMED);
 
         Json::Value jvRequest (Json::objectValue);
@@ -1171,6 +1190,7 @@ public:
             {   "ledger_request",       &RPCParser::parseLedgerId,              1,  1   },
             {   "log_level",            &RPCParser::parseLogLevel,              0,  2   },
             {   "logrotate",            &RPCParser::parseAsIs,                  0,  0   },
+            {   "manifest",             &RPCParser::parseManifest,              1,  1   },
             {   "owner_info",           &RPCParser::parseAccountItems,          1,  2   },
             {   "peers",                &RPCParser::parseAsIs,                  0,  0   },
             {   "ping",                 &RPCParser::parseAsIs,                  0,  0   },
@@ -1195,6 +1215,7 @@ public:
             {   "tx_history",           &RPCParser::parseTxHistory,             1,  1   },
             {   "unl_list",             &RPCParser::parseAsIs,                  0,  0   },
             {   "validation_create",    &RPCParser::parseValidationCreate,      0,  1   },
+            {   "validator_info",       &RPCParser::parseAsIs,                  0,  0   },
             {   "version",              &RPCParser::parseAsIs,                  0,  0   },
             {   "wallet_propose",       &RPCParser::parseWalletPropose,         0,  1   },
             {   "internal",             &RPCParser::parseInternal,              1, -1   },
--- a/src/ripple/rpc/handlers/Manifest.cpp
+++ b/src/ripple/rpc/handlers/Manifest.cpp
@@ -25,7 +25,7 @@
 #include <ripple/rpc/Context.h>
 
 namespace ripple {
-Json::Value doManifest (RPC::Context& context)
+Json::Value doManifest (RPC::JsonContext& context)
 {
     auto& params = context.params;
 
--- a/src/ripple/rpc/handlers/Manifest.cpp
+++ b/src/ripple/rpc/handlers/Manifest.cpp
@@ -1,0 +1,75 @@
+//------------------------------------------------------------------------------
+/*
+    This file is part of rippled: https://github.com/ripple/rippled
+    Copyright (c) 2019 Dev Null Productions
+
+    Permission to use, copy, modify, and/or distribute this software for any
+    purpose  with  or without fee is hereby granted, provided that the above
+    copyright notice and this permission notice appear in all copies.
+
+    THE  SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+    WITH  REGARD  TO  THIS  SOFTWARE  INCLUDING  ALL  IMPLIED  WARRANTIES  OF
+    MERCHANTABILITY  AND  FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+    ANY  SPECIAL ,  DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+    WHATSOEVER  RESULTING  FROM  LOSS  OF USE, DATA OR PROFITS, WHETHER IN AN
+    ACTION  OF  CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+    OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+//==============================================================================
+
+#include <ripple/app/main/Application.h>
+#include <ripple/basics/base64.h>
+#include <ripple/json/json_value.h>
+#include <ripple/protocol/ErrorCodes.h>
+#include <ripple/protocol/jss.h>
+#include <ripple/rpc/Context.h>
+
+namespace ripple {
+Json::Value doManifest (RPC::Context& context)
+{
+    auto& params = context.params;
+
+    if (!params.isMember(jss::public_key))
+        return RPC::missing_field_error (jss::public_key);
+
+    auto const requested = params[jss::public_key].asString();
+
+    Json::Value ret;
+    ret[jss::requested] = requested;
+
+    auto const pk = parseBase58<PublicKey>(TokenType::NodePublic, requested);
+    if (!pk)
+    {
+        RPC::inject_error(rpcINVALID_PARAMS, ret);
+        return ret;
+    }
+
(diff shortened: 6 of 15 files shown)


src/ripple/app/misc/Manifest.h
```diff
@@ -257,6 +257,32 @@ class ManifestCache
     PublicKey
     getMasterKey (PublicKey const& pk) const;
 
+    /** Returns master key's current manifest sequence.
+
+        @return sequence corresponding to Master public key
```
nbougalis: Micro-nit: no need for multiple **`@return`** statements:

```
@return sequence corresponding to master public key if
        configured or 0 otherwise.
```

I also appreciate the thread-safety notation, but I'm not sure that it adds a lot. Of course, I'm biased (since I understand the code) and perhaps others will find it useful.

src/ripple/app/misc/Manifest.h
```diff
@@ -257,6 +257,32 @@ class ManifestCache
...
+        @par Thread Safety
+
+        May be called concurrently
+    */
+    std::uint32_t
+    getSequence (PublicKey const& pk) const;
+
+    /** Returns master key's current domain
+
+        @return domain corresponding to Master public key
```
nbougalis: Similar comment as above.

src/ripple/app/misc/Manifest.h
```diff
@@ -257,6 +257,32 @@ class ManifestCache
...
+        @return 0 if no known sequence from a manifest
+
+        @par Thread Safety
+
+        May be called concurrently
+    */
+    std::uint32_t
+    getSequence (PublicKey const& pk) const;
+
+    /** Returns master key's current domain
```
nbougalis: Not to get into a grammar war, but this possessive is likely to irritate someone at some point. It also reads slightly weird, imo. Consider:

`/** Return the domain claimed by a given public key

src/ripple/app/misc/impl/Manifest.cpp
```diff
@@ -309,6 +309,30 @@ ManifestCache::getMasterKey (PublicKey const& pk) const
     return pk;
 }
 
+std::uint32_t
+ManifestCache::getSequence (PublicKey const& pk) const
+{
+    std::lock_guard lock{read_mutex_};
+    auto const iter = map_.find (pk);
+
+    if (iter != map_.end () && !iter->second.revoked ())
```
nbougalis: I'm not sure if returning `0` is appropriate for a revoked manifest.

src/ripple/net/impl/RPCCall.cpp
```diff
@@ -144,6 +145,21 @@ class RPCParser
         }
     }
 
+    static Json::Value validPublicKey (std::string const& strPk)
```
nbougalis: This function should return a `bool` and not a `Json::Value`.

src/ripple/net/impl/RPCCall.cpp
```diff
@@ -144,6 +145,21 @@ class RPCParser
...
+            return true;
+
+        auto pkHex = strUnHex (strPk);
+        if (!pkHex)
+            return false;
+
+        if (!publicKeyType(makeSlice(*pkHex)))
+            return false;
+
+        return true;
```
nbougalis: An alternative here is this:

```
if (auto k = strUnHex(strPk))
    return static_cast<bool>(publicKeyType(makeSlice(*k)));
return false;
```

Note that this rewrite returns `true` from the branch that _positively_ identifies a key as being of a given type, instead of by default. I think that's a net positive. Apart from that, I don't know if it's better; it is _shorter_, but that isn't always a positive.

src/ripple/net/impl/RPCCall.cpp
```diff
@@ -202,6 +218,25 @@ class RPCParser
...
+    {
+        if (1 == jvParams.size ())
+        {
+          Json::Value     jvRequest (Json::objectValue);
+
+          std::string const strPk = jvParams[0u].asString ();
+          if (!validPublicKey (strPk))
+            return rpcError (rpcPUBLIC_MALFORMED);
+
+          jvRequest[jss::public_key] = strPk;
```
nbougalis: If `validPublicKey` returned `boost::optional<PublicKey>` then we could return the public key in "canonical" format (whether base58 or hex) instead of whichever format the user provided it for us.

src/ripple/rpc/handlers/ValidatorInfo.cpp
```diff
@@ -0,0 +1,46 @@
...
+Json::Value doValidatorInfo (RPC::Context& context)
+{
+    // return error if not configured as validator
+    if(context.app.validatorKeys().publicKey.empty())
+      return RPC::not_validator_error ();
+
+    Json::Value ret;
+
+    auto pk = context.app.validatorKeys().publicKey;
+    ret[jss::master_key]        = toBase58(TokenType::NodePublic, pk);
```
nbougalis: Similarly, why not:

```
ret[jss::master_key] = toBase58(TokenType::NodePublic,
    context.app.getValidationPublicKey());
```

Also hate to complain about whitespace, but (here and elsewhere) I'd put one space before and one space after the `=` and avoid attempting to "align" the code manually.

src/ripple/rpc/handlers/Manifest.cpp
```diff
@@ -0,0 +1,60 @@
...
+    // first attempt to use params as ephemeral key,
+    // if this lookup succeeds master key will be returned,
+    // else pk will just be returned and we will assume that
+    // is master key anyways
+    auto mk = context.app.validatorManifests().getMasterKey(*pk);
+
+    auto ek = context.app.validatorManifests().getSigningKey(mk);
+
+    // if ephemeral key not found, we don't have specified manifest
+    if(ek == mk)
```
nbougalis: This is a particularly confusing way of determining that we don't have a given manifest; not your fault, the API for this kind of thing is just awkward and it's something we ought to look at improving.

It also results in returning an empty JSON object and that _is_ something we can address here and now: we should, at the very least, return something helpful, e.g. the key that we searched for.

Consider if the returned JSON was:

```
{
    "requested": "nHDpbBXLzgktPjgcEEpb1Q9YgNDnZBPdHRXHza8PNLbUpGFxBZxV",
    /** These appear only if the manifest was found */
    "manifest" : "the actual manifest in hex",
    "details" : {
        "master_key" : "nHDpbBXLzgktPjgcEEpb1Q9YgNDnZBPdHRXHza8PNLbUpGFxBZxV",
        "ephemeral_key" : "n9LdwcMvLtsgpVcNeQJDCTQWbd46AY1i71pQPiJPsjnjBYuPWMu7",
        "sequence" : 1,
        "domain" : "area51.number8.us"
    }
}
```

src/ripple/rpc/handlers/ValidatorInfo.cpp
```diff
@@ -0,0 +1,46 @@
...
+*/
+//==============================================================================
+
+#include <ripple/app/misc/ValidatorKeys.h>
+
+namespace ripple {
+Json::Value doValidatorInfo (RPC::Context& context)
+{
+    // return error if not configured as validator
+    if(context.app.validatorKeys().publicKey.empty())
```
nbougalis: Why not: `if(context.app.getValidationPublicKey().empty())`?

src/ripple/app/main/Application.h
```diff
@@ -137,6 +138,7 @@ class Application : public beast::PropertyStream::Source
     virtual TxQ&                        getTxQ() = 0;
     virtual ValidatorList&              validators () = 0;
     virtual ValidatorSite&              validatorSites () = 0;
+    virtual ValidatorKeys const&        validatorKeys () = 0;
```
nbougalis: This seems unnecessary to me for this patch. See why further below.

src/ripple/rpc/handlers/ValidatorInfo.cpp
```diff
@@ -0,0 +1,46 @@
...
+      return RPC::not_validator_error ();
+
+    Json::Value ret;
+
+    auto pk = context.app.validatorKeys().publicKey;
+    ret[jss::master_key]        = toBase58(TokenType::NodePublic, pk);
+
+    auto ek = context.app.validatorManifests().getSigningKey(pk);
+    ret[jss::ephemeral_key]     = toBase58(TokenType::NodePublic, ek);
```
nbougalis: You should detect and account for the use case where a validator is configured with a fixed key and not a manifest. If that happens, this field and those that follow are not relevant and should not be included.

src/ripple/app/misc/impl/Manifest.cpp
```diff
@@ -309,6 +309,30 @@ ManifestCache::getMasterKey (PublicKey const& pk) const
...
+        return iter->second.sequence;
+
+    return 0;
+}
+
+std::string
```
nbougalis: This and `ManifestCache::getSequence` should probably wrap their return in a `boost::optional`, making it possible to distinguish a "I can't find the key you asked for" from "the key you asked for doesn't have the attribute you request".

src/ripple/app/misc/Manifest.h
movitto: Merged the @return statements into one. As far as thread safety, was following the style demonstrated in the other accessor methods but removed in the followup patch.

src/ripple/app/misc/Manifest.h
movitto: Done

src/ripple/app/misc/Manifest.h
movitto: Done

src/ripple/app/misc/impl/Manifest.cpp
movitto: Changed this to boost::optional and boost::none as suggested below

src/ripple/net/impl/RPCCall.cpp
movitto: Done

src/ripple/net/impl/RPCCall.cpp
movitto: @nbougalis do you mind if we leave this as is for this patch? The reason I suggest this is that this method is simply extracted verbatim from the **parseFetchInfo** method below. This way we can just focus on extracting the common logic here, and refactoring / cleaning this up in a separate PR.

src/ripple/net/impl/RPCCall.cpp
movitto: Done

src/ripple/rpc/handlers/ValidatorInfo.cpp
movitto: Done

src/ripple/rpc/handlers/Manifest.cpp
movitto: Updated the return JSON to match the suggested format and to always return the 'requested' key even in the case of the failed lookup

src/ripple/rpc/handlers/ValidatorInfo.cpp
movitto: Done

src/ripple/app/main/Application.h
movitto: Done

src/ripple/rpc/handlers/ValidatorInfo.cpp
movitto: I **believe** I addressed this by checking if the ephemeral key does exists, and if not using this to imply the manifest is not present. Please correct if this is the wrong assumption

src/ripple/app/misc/impl/Manifest.cpp
movitto: Done

src/ripple/net/impl/RPCCall.cpp
nbougalis: Not at all. Whether review comments should be "resolved" is typically up to the PR author. Obviously, exceptions for bugs/problems exist, but this wasn't that. I'm fine with leaving this as is.

src/ripple/net/impl/RPCCall.cpp
```diff
@@ -114,6 +114,7 @@ class RPCParser
         return true;
     }
 
+
```
seelabs: Extra space snuck in here.

src/ripple/net/impl/RPCCall.cpp
```diff
@@ -202,6 +218,25 @@ class RPCParser
         return v;
     }
 
+    Json::Value parseManifest (Json::Value const& jvParams)
+    {
+        if (1 == jvParams.size ())
```
seelabs: Style: We don't use yoda conditions in rippled. I understand the reasoning, but I'd like to stay consistent with the style of the rest of rippled.

src/ripple/net/impl/RPCCall.cpp
```diff
@@ -202,6 +218,25 @@ class RPCParser
...
+        {
+          Json::Value     jvRequest (Json::objectValue);
```
seelabs: Two style nits:
1) should indent four spaces
2) should be only one space after the type

src/ripple/rpc/handlers/Manifest.cpp
```diff
@@ -0,0 +1,75 @@
...
+      return ret;
+
+    auto manifest = context.app.validatorManifests().getManifest(mk);
+    if(manifest)
+      ret[jss::manifest] = base64_encode((std::uint8_t const*)manifest->c_str(),
```
seelabs: Two style nits:
1) should be indented four spaces
2) we usually use `static_cast` over c-style casts

src/ripple/rpc/handlers/Manifest.cpp
```diff
@@ -0,0 +1,75 @@
...
+    auto mk = context.app.validatorManifests().getMasterKey(*pk);
+
+    auto ek = context.app.validatorManifests().getSigningKey(mk);
+
+    // if ephemeral key not found, we don't have specified manifest
+    if(ek == mk)
+      return ret;
+
+    auto manifest = context.app.validatorManifests().getManifest(mk);
+    if(manifest)
```
seelabs: If a variable is only used in the if statement, I prefer to put the declaration in the if statment. `if (auto manifest = context....)`. No change needed if you really hate that style, but I do prefer to keep variable scopes as narrow as we can (in general). This is done in several places in this file (and ValidatorInfo.cpp). I'll briefly tag the others as well.

src/ripple/rpc/handlers/Manifest.cpp
```diff
@@ -0,0 +1,75 @@
...
+      ret[jss::manifest] = base64_encode((std::uint8_t const*)manifest->c_str(),
+                                                              manifest->length());
+
+    Json::Value details;
+
+    details[jss::master_key] = toBase58(TokenType::NodePublic, mk);
+    details[jss::ephemeral_key] = toBase58(TokenType::NodePublic, ek);
+
+    auto seq = context.app.validatorManifests().getSequence(mk);
+    if(seq)
```
seelabs: `seq`'s scope could be narrowed to the `if` statement.

Also, four spaces indent.

src/ripple/rpc/handlers/Manifest.cpp
```diff
@@ -0,0 +1,75 @@
...
+      details[jss::seq] = *seq;
+
+    auto domain = context.app.validatorManifests().getDomain(mk);
+    if(domain)
```
seelabs: `domain` scope could be narrowed to the if statement

src/ripple/rpc/handlers/ValidatorInfo.cpp
```diff
@@ -0,0 +1,61 @@
...
+    // pk is maskter key, eg no ephemeral key, eg no manifest, just return
+    if(mk == pk)
+      return ret;
+
+    // lookup ephemeral key
+    auto ek = context.app.validatorManifests().getSigningKey(mk);
+    ret[jss::ephemeral_key] = toBase58(TokenType::NodePublic, ek);
+
+    auto manifest = context.app.validatorManifests().getManifest(mk);
+    if(manifest)
```
seelabs: `manifest` scope could be narrowed to the `if` statement

Also, indent four spaces.

src/ripple/rpc/handlers/ValidatorInfo.cpp
```diff
@@ -0,0 +1,61 @@
...
+      ret[jss::manifest] = base64_encode((std::uint8_t const*)manifest->c_str(),
+                                                             manifest->length());
+
+    auto seq = context.app.validatorManifests().getSequence(mk);
+    if(seq)
```
seelabs: `seq` scope could be narrowed to the if statement

Also one space after `if` and condition and indent four spaces

src/ripple/rpc/handlers/ValidatorInfo.cpp
```diff
@@ -0,0 +1,61 @@
...
+      ret[jss::seq] = *seq;
+
+    auto domain = context.app.validatorManifests().getDomain(mk);
+    if(domain)
```
seelabs: `domain` scope could be narrowed to the if statement.

Also, four spaces indent

seelabs:
I also noticed there are no unit tests for this. It would be good to have some simple tests for this as well.

seelabs:
Non-unity build is broken. I think if both ValidatorInfo and Manifest have the following includes it should fix it:
```
 
+#include <ripple/app/main/Application.h>
 #include <ripple/basics/base64.h>
+#include <ripple/json/json_value.h>
+#include <ripple/protocol/ErrorCodes.h>
+#include <ripple/protocol/jss.h>
+#include <ripple/rpc/Context.h>
```

movitto:
@seelabs thanks for the review. Resolved the style issues and added missing includes for the unity build.

As far as static_cast, when making the change results in the following error:

```
/home/mmorsi/Workspace/rippled/src/ripple/rpc/handlers/Manifest.cpp:55:94: error: invalid static_cast from type ‘const char*’ to type ‘const uint8_t*’ {aka ‘const unsigned char*’}
   55 |         ret[jss::manifest] = base64_encode(static_cast<std::uint8_t const*>(manifest->c_str()),
```
Could you advice a fix or should this be left as is?

Will look into adding some tests in the near future.

src/ripple/rpc/handlers/Manifest.cpp
seelabs: @movitto (responding to your comment here https://github.com/ripple/rippled/pull/3197#issuecomment-572272128)

`static_cast` won't work - we need a `reinterpret_cast` so something like:
```
        ret[jss::manifest] = base64_encode(
            reinterpret_cast<std::uint8_t const*>(manifest->data()),
            manifest->length());
```

*However*, there's an overload of `base64_encode` that takes a `std::string`; I'd use that. So we can change this to:
```
        ret[jss::manifest] = base64_encode(*manifest);
```

jwbusch:
> /home/mmorsi/Workspace/rippled/src/ripple/rpc/handlers/Manifest.cpp:55:94: error: invalid static_cast from type ‘const char*’ to type ‘const uint8_t*’ {aka ‘const unsigned char*’}
>    55 |         ret[jss::manifest] = base64_encode(static_cast<std::uint8_t const*>(manifest->c_str()),
> ```
> 
> Could you advice a fix or should this be left as is?
> 
> Will look into adding some tests in the near future.

` ret[jss::manifest] = base64_encode(*manifest);`

Should work as @seelabs mentioned.

movitto:
@seelabs @jwbusch cool thanks for the feedback. Called the simpler base64_encode method, removed the cast, and added some tests. Let me know if there is anything else!

src/ripple/rpc/handlers/ValidatorInfo.cpp
```diff
@@ -0,0 +1,63 @@
...
+namespace ripple {
+Json::Value doValidatorInfo (RPC::Context& context)
+{
+    // return error if not configured as validator
+    if (context.app.getValidationPublicKey().empty())
+      return RPC::not_validator_error ();
+
+    Json::Value ret;
+
+    auto pk = context.app.getValidationPublicKey();
```
jwbusch: nit: Would like to see more `const` used if possible. i.e. `const auto pk = ...`

Makes it clear that the variable won't be modified later (which is the case for a lot of these).

src/ripple/rpc/handlers/ValidatorInfo.cpp
```diff
@@ -0,0 +1,63 @@
...
+    auto mk = context.app.validatorManifests().getMasterKey(pk);
+    ret[jss::master_key] = toBase58(TokenType::NodePublic, mk);
+
+    // pk is maskter key, eg no ephemeral key, eg no manifest, just return
+    if (mk == pk)
+        return ret;
+
+    // lookup ephemeral key
+    auto ek = context.app.validatorManifests().getSigningKey(mk);
+    ret[jss::ephemeral_key] = toBase58(TokenType::NodePublic, ek);
```
jwbusch: this one could be scoped like the others down below

```
if (const auto ek = ...)
   ret[...] = ...;
```

movitto:
@jwbusch pushed a patch constantizing the fields. As far as the scope ephimeral_key, that method (getSigningKey, already existing before this PR) works a bit different than the new ones in the sense that if the key is not found, the input param (the master key) is simply returned. Since this is not a boost::optional we can't check it's 'truthyness' like the other fields.

Builds/CMake/RippledCore.cmake
```diff
@@ -979,6 +981,7 @@ else ()
     src/test/rpc/LedgerData_test.cpp
     src/test/rpc/LedgerRPC_test.cpp
     src/test/rpc/LedgerRequestRPC_test.cpp
+    src/test/rpc/Manifest_test.cpp
```
seelabs: I think this needs to be `ManifestRPC_test.cpp`

movitto: @seelabs doh! :man_facepalming: fixed

seelabs:
I think this is ready to go, we just need to fix the non-unity build issues. I think we just need to patch two files. I think this patch should fix things: https://github.com/seelabs/rippled/commit/b47e3742cdff7062d56add2ac3d2a97d541f82a5

movitto:
> I think this is ready to go, we just need to fix the non-unity build issues. I think we just need to patch two files. I think this patch should fix things: [seelabs@b47e374](https://github.com/seelabs/rippled/commit/b47e3742cdff7062d56add2ac3d2a97d541f82a5)

@seelabs cherry picked onto this branch, thanks!

seelabs:
@movitto FYI: Rebasing should be pretty straight-forward. One change to look out for is the parameters for `doValidatorInfo` and `doManifest` need to be changed from `RPC::Context` to `RPC::JsonContext`.

nbougalis:
@movitto: I can handle the rebase and squashing down if you want, and would be happy to jump on a call with you to walk you through the commands and explain what I'm doing so you know for the next time.

movitto:
@nbougalis sure that works. With the XRP meetup tomorrow won't have time to do the call until later on in the week, but perhaps we can schedule it for then (lets coordinate on that one via email)

movitto:
@seelabs @nbougalis rebased and pushed a followup patch w/ the 'Context' -> 'JsonContext' substitution. Using @mellery451's protobuf fix from #3244 (not included in this PR) I was able to build and test locally.

seelabs:
@movitto Things looks good; I just added a "passed" label. Thanks for (yet another) contribution!

PR #3209, opened 2020-01-06
Check XRP endpoints for circular paths (RIPD-1781):

seelabs:
The payment engine restricts payment paths so two steps do not input the same
Currency/Issuer or output the same Currency/Issuer. This check was skipped when
the path started or ended with XRP. An example of a path that was incorrectly
accepted was: XRP -> //USD -> //XRP -> EUR

This patch enables the path loop check for paths that start or end with XRP.

### DIFF
--- a/src/ripple/app/paths/impl/XRPEndpointStep.cpp
+++ b/src/ripple/app/paths/impl/XRPEndpointStep.cpp
@@ -25,6 +25,7 @@
 #include <ripple/basics/Log.h>
 #include <ripple/basics/XRPAmount.h>
 #include <ripple/ledger/PaymentSandbox.h>
+#include <ripple/protocol/Feature.h>
 #include <ripple/protocol/Quality.h>
 
 #include <boost/container/flat_set.hpp>
@@ -358,6 +359,18 @@ XRPEndpointStep<TDerived>::check (StrandContext const& ctx) const
     auto ter = checkFreeze (ctx.view, src, dst, xrpCurrency ());
     if (ter != tesSUCCESS)
         return ter;
+
+    if (ctx.view.rules().enabled(fix1781))
+    {
+        auto const issuesIndex = isLast_ ? 0 : 1;
+        if (!ctx.seenDirectIssues[issuesIndex].insert(xrpIssue()).second)
+        {
+            JLOG(j_.debug())
+                << "XRPEndpointStep: loop detected: Index: " << ctx.strandSize
+                << ' ' << *this;
+            return temBAD_PATH_LOOP;
+        }
+    }
 
     return tesSUCCESS;
 }
--- a/src/test/app/Flow_test.cpp
+++ b/src/test/app/Flow_test.cpp
@@ -1180,6 +1180,100 @@ struct Flow_test : public beast::unit_test::suite
             ter(temBAD_PATH));
     }
 
+    void
+    testXRPPathLoop()
+    {
+        testcase("Circular XRP");
+
+        using namespace jtx;
+        auto const alice = Account("alice");
+        auto const bob = Account("bob");
+        auto const gw = Account("gw");
+        auto const USD = gw["USD"];
+        auto const EUR = gw["EUR"];
+
+        for (auto const withFix : {true, false})
+        {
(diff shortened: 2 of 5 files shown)


yxxyun:
I think should support circular payment not disable it
https://github.com/ripple/rippled/issues/1257

MarkusTeufelberger:
Me too, but currently this is not supported, so until they are properly enabled I guess it is fine to make sure they are strictly not happening and not just most of the time unless there's some special edge case.

tuloski:
All the forward and backward problems and then it seems that rippled can already handle that, so why not just enabling circular payments. It's a feature more than a drawback IMO.

src/ripple/app/paths/impl/XRPEndpointStep.cpp
```diff
@@ -360,6 +361,18 @@ XRPEndpointStep<TDerived>::check (StrandContext const& ctx) const
...
 
+    if (ctx.view.rules().enabled(fix1781))
+    {
+        auto const issuesIndex = isLast_ ? 0 : 1;
+        if (!ctx.seenDirectIssues[issuesIndex].insert(xrpIssue()).second)
+        {
+            JLOG(j_.debug())
+                << "XRPEndpointStep: loop detected: Index: " << ctx.strandSize
+                << ' ' << *this;
+            return temBAD_PATH_LOOP;
```
scottschurr: When I run code_cov over the new code, this line returning `temBAD_PATH_LOOP` isn't hit.  Is there a (not too painful) way to hit this line with a unit test?

seelabs: I don't think I can make a test that will hit this. The reason is the book steps also check for a path, and the book step will detect the loop before the endpoint will.

I'd still like to leave this code in, but in practice I don't think it can be hit.

seelabs:
Rebased onto 1.5.0-b5 and added two more tests. Since this has already been approved in it's current version, I'm still going to consider this approved (but of course feel free to review the new tests). I'm going to aim to get this into 1.6 rather than 1.5, since we're so late into the 1.5 cycle.

src/test/app/Flow_test.cpp
```diff
@@ -1248,17 +1297,18 @@ struct Flow_test : public beast::unit_test::suite
 
     void run() override
     {
-        testLimitQuality();
+        // testXXX();
+        // testLimitQuality();
```
scottschurr: You left most of the old tests commented out when you added the new tests.

src/test/app/Flow_test.cpp
```diff
@@ -1248,17 +1297,18 @@ struct Flow_test : public beast::unit_test::suite
...
-        testEmptyStrand(sa);
+        // testRIPD1443();
+        // testRIPD1449();
+
+        // using namespace jtx;
+        // auto const sa = supported_amendments();
+        // testWithFeats(sa - fix1373 - featureFlowCross);
+        // testWithFeats(sa           - featureFlowCross);
+        // testWithFeats(sa);
+        // testEmptyStrand(sa);
```
scottschurr: While you're in un-commenting these tests, consider tidying up the closing parenthesis on line 1330.  Just for extra credit :wink:

seelabs: Ug! Good catch!!! Will fix.

src/test/app/Flow_test.cpp
seelabs: Fixed in `c5b605e2b [fold] Uncomment existing tests`

src/test/app/Flow_test.cpp
seelabs: Fixed in `c5b605e2b [fold] Uncomment existing tests`

seelabs:
@scottschurr I don't mind wrapping at 72 characters. Will do. I know you've requested changes like this in the past. The reason I keep posting 80 character commit messages is my editor wraps git commit messages at 80 characters by default and I haven't taken the time to figure out what I need to do to change that.

intelliot:
This introduces a new amendment called `fix1781`.

PR #3212, opened 2020-01-06
Remove time based switches

scottschurr:
Some of the earliest amendment-like things were time-based switches that enabled a particular new behavior once a parent ledger had passed a particular date and time.  The most recent time-based switch was enabled on March 30, 2017.  Well over 2 years ago.

If merged this pull request will have two effects:

1. If someone is replaying a ledger that is from before January 1 2018, a warning (at `FTL` level) is printed to the log.

2. All of the conditional execution based on time-based switches is removed. In effect, the ledger now presumes that the threshold time is passed.  As a consequence all of the time-based switches are removed.

The first commit adds the log message.  Each remaining commit removes an individual time-based switch.  My preference is that the commits be merged independently (not squashed) since each change really is independent of the others.

Thanks for the help.

### DIFF
--- a/src/ripple/app/main/Application.cpp
+++ b/src/ripple/app/main/Application.cpp
@@ -1987,6 +1987,20 @@ bool ApplicationImp::loadOldLedger (
                 }
             }
         }
+        using namespace std::chrono_literals;
+        using namespace date;
+        static constexpr NetClock::time_point ledgerWarnTimePoint{
+                          sys_days{January/1/2018} - sys_days{January/1/2000}};
+        if (loadLedger->info().closeTime < ledgerWarnTimePoint)
+        {
+            JLOG(m_journal.fatal()) <<
+                "\n\n***  WARNING   ***\n"
+                "You are replaying a ledger from before " <<
+                to_string(ledgerWarnTimePoint) << " UTC.\n"
+                "This replay will not handle your ledger as it was originally "
+                "handled.\nConsider running an earlier version of rippled to "
+                "get the older rules.\n*** CONTINUING ***\n";
+        }
 
         JLOG(m_journal.info()) <<
             "Loading ledger " << loadLedger->info().hash <<
--- a/src/ripple/app/misc/impl/TxQ.cpp
+++ b/src/ripple/app/misc/impl/TxQ.cpp
@@ -278,9 +278,6 @@ TxQ::MaybeTx::MaybeTx(
 std::pair<TER, bool>
 TxQ::MaybeTx::apply(Application& app, OpenView& view, beast::Journal j)
 {
-    boost::optional<STAmountSO> saved;
-    if (view.rules().enabled(fix1513))
-        saved.emplace(view.info().parentCloseTime);
     // If the rules or flags change, preflight again
     assert(pfresult);
     if (pfresult->rules != view.rules() ||
@@ -539,13 +536,7 @@ TxQ::tryClearAccountQueue(Application& app, OpenView& view,
     }
     // Apply the current tx. Because the state of the view has been changed
     // by the queued txs, we also need to preclaim again.
-    auto txResult = [&]{
-        boost::optional<STAmountSO> saved;
-        if (view.rules().enabled(fix1513))
-            saved.emplace(view.info().parentCloseTime);
-        auto const pcresult = preclaim(pfresult, app, view);
-        return doApply(pcresult, app, view);
-    }();
+    auto const txResult = doApply (preclaim (pfresult, app, view), app, view);
 
     if (txResult.second)
     {
@@ -633,11 +624,6 @@ TxQ::apply(Application& app, OpenView& view,
     auto const account = (*tx)[sfAccount];
     auto const transactionID = tx->getTransactionID();
     auto const tSeq = tx->getSequence();
-
-    boost::optional<STAmountSO> saved;
-    if (view.rules().enabled(fix1513))
-        saved.emplace(view.info().parentCloseTime);
-
     // See if the transaction is valid, properly formed,
     // etc. before doing potentially expensive queue
     // replace and multi-transaction operations.
--- a/src/ripple/app/paths/impl/BookStep.cpp
+++ b/src/ripple/app/paths/impl/BookStep.cpp
@@ -714,14 +714,12 @@ BookStep<TIn, TOut, TDerived>::revImp (
             result.out = out;
             this->consumeOffer (sb, offer, ofrAdjAmt, stpAdjAmt, ownerGivesAdj);
 
-            // When the mantissas of two iou amounts differ by less than ten, then
-            // subtracting them leaves a result of zero. This can cause the check for
-            // (stpAmt.out > remainingOut) to incorrectly think an offer will be funded
-            // after subtracting remainingIn.
-            if (fix1298(sb.parentCloseTime()))
-                return offer.fully_consumed();
-            else
-                return false;
+            // Explicitly check whether the offer is funded.  Given that we have
+            // (stpAmt.out > remainingOut), it's natural to assume the offer
+            // will still be funded after consuming remainingOut but that is
+            // not always the case.  If the mantissas of two IOU amounts differ
+            // by less than ten, then subtracting them leaves a zero.
+            return offer.fully_consumed();
         }
     };
 
@@ -884,10 +882,7 @@ BookStep<TIn, TOut, TDerived>::fwdImp (
         // subtracting them leaves a result of zero. This can cause the check for
         // (stpAmt.in > remainingIn) to incorrectly think an offer will be funded
         // after subtracting remainingIn.
-        if (fix1298(sb.parentCloseTime()))
-            processMore = processMore || offer.fully_consumed();
-
-        return processMore;
+        return processMore || offer.fully_consumed();
     };
 
     {
--- a/src/ripple/app/paths/impl/BookStep.cpp
+++ b/src/ripple/app/paths/impl/BookStep.cpp
@@ -1015,23 +1015,20 @@ BookStep<TIn, TOut, TDerived>::check(StrandContext const& ctx) const
         return tecNO_ISSUER;
     }
 
-    if (fix1443(ctx.view.info().parentCloseTime))
+    if (ctx.prevStep)
     {
-        if (ctx.prevStep)
+        if (auto const prev = ctx.prevStep->directStepSrcAcct())
         {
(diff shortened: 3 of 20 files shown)


src/ripple/protocol/Feature.h
```diff
@@ -46,6 +46,23 @@ namespace ripple {
 
 namespace detail {
 
+// *NOTE*
+//
+// Features, or Amendments as they are called elsewhere, are enabled on the
+// network at some specific time based on Validator voting.  Once an
+// Amendment has been enabled for, say, more than two years then retaining
+// the conditional code based on the Amendment is pointless.
```
seelabs: "pointless" is a little strong. They are still useful when replaying old transaction. I agree we should remove them, but I do actually replay old transactions now and then.

src/ripple/app/main/Application.cpp
```diff
@@ -1976,6 +1976,17 @@ bool ApplicationImp::loadOldLedger (
                 }
             }
         }
+        using namespace std::chrono_literals;
+        static constexpr char ledgerWarnTime[] {"Jan 01, 2018 00:00:00 UTC"};
```
miguelportilla: Consider inlining `Jan 01, 2018 00:00:00 UTC` into the log message versus declaring a local.

src/ripple/app/misc/impl/TxQ.cpp
```diff
@@ -536,9 +533,6 @@ TxQ::tryClearAccountQueue(Application& app, OpenView& view,
     // Apply the current tx. Because the state of the view has been changed
     // by the queued txs, we also need to preclaim again.
     auto txResult = [&]{
```
nbougalis: Do we still need this lambda? Can this become:

```
auto txResult = doApply(
    preclaim(pfresult, app, view),
    app, view);
```

src/ripple/app/paths/impl/BookStep.cpp
```diff
@@ -718,10 +718,7 @@ BookStep<TIn, TOut, TDerived>::revImp (
             // subtracting them leaves a result of zero. This can cause the check for
             // (stpAmt.out > remainingOut) to incorrectly think an offer will be funded
             // after subtracting remainingIn.
-            if (fix1298(sb.parentCloseTime()))
-                return offer.fully_consumed();
-            else
-                return false;
+            return offer.fully_consumed();
```
nbougalis: Since you're here, I'm really confused about this comment. What is it telling me that's relevant _here_? Is this an artifact of older code?

src/ripple/protocol/STAmount.h
```diff
@@ -405,38 +405,6 @@ inline bool isXRP(STAmount const& amount)
     return isXRP (amount.issue().currency);
 }
 
-extern LocalValue<bool> stAmountCalcSwitchover;
```
nbougalis: Can also remove the `#include <ripple/basics/LocalValue.h>` in this file.

src/ripple/protocol/impl/STAmount.cpp
```diff
@@ -1240,9 +1229,9 @@ mulRound (STAmount const& v1, STAmount const& v2, Issue const& issue,
     STAmount result (issue, amount, offset, resultNegative);
 
     // Control when bugfixes that require switchover dates are enabled
```
nbougalis: Comment is no longer relevant?

src/ripple/protocol/impl/STAmount.cpp
```diff
@@ -1312,9 +1301,9 @@ divRound (STAmount const& num, STAmount const& den,
 
     STAmount result (issue, amount, offset, resultNegative);
     // Control when bugfixes that require switchover dates are enabled
```
nbougalis: Comment is no longer relevant?

src/ripple/app/main/Application.cpp
```diff
@@ -1976,6 +1976,17 @@ bool ApplicationImp::loadOldLedger (
...
+        static constexpr NetClock::time_point ledgerWarnTimePoint {568080000s};
```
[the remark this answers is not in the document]
scottschurr: Boy, I wish I had a better answer for this question.  And I will once we're building with C++20.  In C++20 `chrono` provides calendar support.  Once we have calendar support then the string can be generated from the `time_point`.  In the meantime I think this is the best we can do.  But possibly @HowardHinnant has a better approach?

src/ripple/app/main/Application.cpp
```diff
@@ -1976,6 +1976,17 @@ bool ApplicationImp::loadOldLedger (
...
+        if (loadLedger->info().closeTime < ledgerWarnTimePoint)
+        {
+            JLOG(m_journal.fatal()) <<
+                "\n\n*** You are replaying a ledger from before " <<
+                ledgerWarnTime << " ***\n"
+                "*** You may want to run an earlier version of rippled ***\n"
+                "*** that supports the older rules.  Continuing. ***\n";
+        }
```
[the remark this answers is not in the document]
scottschurr: @seelabs, how would you feel about just refusing to process old ledgers?  Personally, I can go either way since I seldom replay old ledgers.  But I can imagine situations where we want to replay an old ledger with newer code to see whether an old bug is fixed.  In those cases blocking old ledger replay would be counterproductive.

But I'll implement whatever you two agree to.

src/ripple/app/main/Application.cpp
scottschurr: My personal preference is to have the `WarnTime` string be immediately adjacent to the `WarnTimePoint`.  I think that increases the likelihood that, if the are changed, they will be maintained together.

Of course the best option is to generate the string from the `time_point` as suggested here: https://github.com/ripple/rippled/pull/3212#discussion_r364531737. We'll have to see if that can be done.

src/ripple/app/main/Application.cpp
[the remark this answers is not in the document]
HowardHinnant: Here's a tested alternative: https://github.com/HowardHinnant/rippled/commit/3b1a9b7acbea9afc7a8d5a76a3bb2ee7fdc4c878

It isn't great.  I think it could be better but would involve a change to the draft C++20 spec (which this has motivated me to try for).

src/ripple/app/main/Application.cpp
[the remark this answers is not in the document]
nbougalis: I'm OK with allowing such replay **if** we make it clear that the result may be different because of different rules. I don't think the message in place now achieves that.

src/ripple/app/main/Application.cpp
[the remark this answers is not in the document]
nbougalis: This is a minor nit really; no change is necessary to this for me. But looking forward to what Howard comes up with ;)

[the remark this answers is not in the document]
HowardHinnant: We can do `to_string(ledgerWarnTimePoint)` today.  The output will be `2018-Jan-01 00:00:00 UTC` instead of `Jan 01, 2018 00:00:00 UTC`, which I think is insignificant.  That gets rid of `ledgerWarnTime`.

Now the only question is how to best construct `ledgerWarnTimePoint`:

Status quo:
```
constexpr NetClock::time_point ledgerWarnTimePoint {568080000s};
```
Proposal we can do today:
```
constexpr NetClock::time_point ledgerWarnTimePoint{
                          sys_days{January/1/2018} - sys_days{January/1/2000}};
```

In both of the above cases `static` is unnecessary.

What I believe I'm going to try to do:

```
constexpr auto ledgerWarnTimePoint = clock_cast<NetClock>(sys_days{January/1/2018});
```
We can almost do this today, except we can't make it `constexpr`.  Making it `constexpr` is what I think I want to try to do.  I've already got a PR from someone requesting this change.

src/ripple/app/paths/impl/BookStep.cpp
seelabs: I wrote that comment. It is still relevant. I agree it's not very clear. Is this clearer? 
```
            // Given that (stpAmt.out > remainingOut), it's natural to expect
            // that the offer will still be funded after consuming remainingOut.
            // However, this is not always the case. When the mantissas of two
            // iou amounts differ by less than ten, then subtracting them leaves
            // a result of zero. So instead of declaring the offer is still funded,
            // we need to check explicitly with a call to offer.fully_consumed
            // (see RIPD-1298)
```

src/ripple/app/main/Application.cpp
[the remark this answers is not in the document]
seelabs: @scottschurr It's not a killer. Replaying old transactions is rare enough that I can hack the code myself. I understand Nik's concern about not replaying ledgers with different rules. I'm OK refusing to replay.

src/ripple/app/paths/impl/BookStep.cpp
nbougalis: Thank you, that's better. Consider instead:

```
// Explicitly check whether an offer is funded: given that we have
// (stpAmt.out > remainingOut), it's natural to expect the offer
// will still be funded after consuming remainingOut but that is
// not always the case: if the mantissas of two IOU amounts differ
// by less than ten, then subtracting them result in a zero.
```

seelabs: Yep, that's better. Thanks @nbougalis . Let's go with that ^^^.

src/ripple/app/main/Application.cpp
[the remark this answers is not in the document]
scottschurr: @seelabs, I'm going with a new-and-improved warning that @nbougalis and I have agreed to:
```
***  WARNING   ***
You are replaying a ledger from before 2018-Jan-01 00:00:00.000000 UTC.
This replay will not handle your ledger as it was originally handled.
Consider running an earlier version of rippled to get the older rules.
*** CONTINUING ***
```

scottschurr:
Thanks for all the great review comments, folks.  In particular, thanks to @HowardHinnant who showed me that chrono date capabilities are available in our code base.  I had no idea.  Very cool.

I've pushed three `FOLD` commits that I believe address all of the comments so far.  Please look them over at your convenience and let me know act you think.  Thanks.

HowardHinnant:
Already looked, looks great! 👍

scottschurr:
Fold commits squashed into their appropriate commits and rebased to 1.5.0-b3.  I'll see if it survives Travis before I mark it as passed.

scottschurr:
Travis failure was vc2019, which I believe is expected or at least unsurprising.  I think this is good to go.

PR #3216, opened 2020-01-08
Streamline pkg and travis CI:

mellery451:
* use tagged containers for pkg build
* update build images
* continue to build container images in pipeline, but allow
  failure (non-block)
* limit travis macos cache
* add vs2019 windows to travis
* remove xcode 9 travis build
* remove clang5/6 from CI and update min version of Clang required in
  cmake
* break windows CI build into stages to reduce timeouts
* update datelib
* add if condition to travis builds to allow commit message to limit
  builds by platform


mellery451:
thanks @scottschurr  - yes, that's a disk space failure and currently working with the CI provider to see about increasing the fs allocation.

PR #3220, opened 2020-01-10
Rm preflow payments

seelabs:
This removes the old payment engine from rippled. The changes in RippleCalc should be carefully reviewed. The commits are split as they are so they can be easily reviewed. The first commit removes the engine, but does as little refactoring/reformatting as possible. The other commits clean up the code a bit.


src/ripple/app/paths/RippleCalc.cpp
```diff
@@ -100,17 +114,17 @@ RippleCalc::Output RippleCalc::rippleCalculate (
...
-            ", dstAmtReq: " << saDstAmountReq <<
-            ", sendMax: " << saMaxAmountReq <<
-            ", algo: " << "V2";
+    j.debug() << "RippleCalc Result> "
+              << " actualIn: " << flowOut.actualAmountIn
+              << ", actualOut: " << flowOut.actualAmountOut
+              << ", result: " << flowOut.result()
+              << ", dstAmtReq: " << saDstAmountReq
+              << ", sendMax: " << saMaxAmountReq << ", algo: "
+              << "V2";
```
scottschurr: Is the "algo: V2" still relevant?

miguelportilla:
Due to the changes, the following comments may no longer apply:

https://github.com/seelabs/rippled/blob/rm-preflow-payments/src/ripple/ledger/View.h#L302

https://github.com/seelabs/rippled/blob/rm-preflow-payments/src/ripple/protocol/jss.h#L180
https://github.com/seelabs/rippled/blob/rm-preflow-payments/src/ripple/protocol/jss.h#L239
https://github.com/seelabs/rippled/blob/rm-preflow-payments/src/ripple/protocol/jss.h#L276
https://github.com/seelabs/rippled/blob/rm-preflow-payments/src/ripple/protocol/jss.h#L535

https://github.com/seelabs/rippled/blob/rm-preflow-payments/src/ripple/protocol/jss.h#L265
https://github.com/seelabs/rippled/blob/rm-preflow-payments/src/ripple/protocol/jss.h#L376

miguelportilla:
The loop can be reduce to a single offer this part of the test:

https://github.com/seelabs/rippled/blob/rm-preflow-payments/src/test/app/Offer_test.cpp#L273-L275

eg.

```     ...
        env (pay (gw, alice, USD (100)));
        env (pay (gw, carol, EUR (100)));
        env (offer (carol, USD (1), EUR (2)));```

miguelportilla:
`tecExpect` may no longer be needed...

https://github.com/seelabs/rippled/blob/rm-preflow-payments/src/test/app/Offer_test.cpp#L3572-L3573

src/ripple/app/paths/RippleCalc.cpp
seelabs: I have some old scripts that use it, but on re-consideration, those scripts are obsolete now. Removed.

seelabs:
@miguelportilla Nice finds, all. I addressed them all, except I kept the loop in the regression test (with a comment).

PR #3225, opened 2020-01-13
Warn operators about upcoming unknown amendments:

ximinez:
* When an unknown amendment reaches majority, log an error-level
  message, and return a `warnings` stanza on all successful
  admin-level RPC calls with a message describing the problem, and
  the expected deadline.
* Check on every flag ledger to see if the amendment(s) lose majority.
  Logs again if they don't, resumes normal operations if they did.
* The intention is to give operators earlier warning that their
  instances are in danger of being amendment blocked, which will
  hopefully motivate them to update ahead of time.

### DIFF
--- a/src/ripple/app/misc/impl/AmendmentTable.cpp
+++ b/src/ripple/app/misc/impl/AmendmentTable.cpp
@@ -158,6 +158,10 @@ protected:
 
     // True if an unsupported amendment is enabled
     bool unsupportedEnabled_;
+    // Unset if no unsupported amendments reach majority,
+    // else set to the earliest time an unsupported amendment
+    // will be enabled.
+    boost::optional<NetClock::time_point> firstUnsupportedExpected_;
 
     beast::Journal const j_;
 
@@ -190,15 +194,19 @@ public:
     bool isSupported (uint256 const& amendment) override;
 
     bool hasUnsupportedEnabled () override;
+    boost::optional<NetClock::time_point>
+    firstUnsupportedExpected() override;
 
     Json::Value getJson (int) override;
     Json::Value getJson (uint256 const&) override;
 
     bool needValidatedLedger (LedgerIndex seq) override;
 
-    void doValidatedLedger (
+    void
+    doValidatedLedger(
         LedgerIndex seq,
-        std::set<uint256> const& enabled) override;
+        std::set<uint256> const& enabled,
+        majorityAmendments_t const& majority) override;
 
     std::vector <uint256>
     doValidation (std::set<uint256> const& enabledAmendments) override;
@@ -392,6 +400,13 @@ AmendmentTableImpl::hasUnsupportedEnabled ()
     return unsupportedEnabled_;
 }
 
+boost::optional<NetClock::time_point>
+AmendmentTableImpl::firstUnsupportedExpected()
+{
+    std::lock_guard sl(mutex_);
+    return firstUnsupportedExpected_;
+}
+
 std::vector <uint256>
 AmendmentTableImpl::doValidation (
     std::set<uint256> const& enabled)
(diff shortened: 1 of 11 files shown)


ximinez:
Rebased on to 1.5.0-b3 and resolved conflicts

src/ripple/app/misc/impl/AmendmentTable.cpp
```diff
@@ -539,12 +554,36 @@ AmendmentTableImpl::needValidatedLedger (LedgerIndex ledgerSeq)
...
+
+        if (s->enabled)
+            break;
+
+        if (!s->supported)
+        {
+            JLOG(j_.info()) << "Unsupported amendment " << m.first
+                             << " reached majority at " << to_string(m.second);
+            if (!firstUnsupportedExpected_ || firstUnsupportedExpected_ > m.second)
+                firstUnsupportedExpected_ = m.second;
```
nbougalis: It's a pity that the [semantics of relational operators for `boost::optional`](https://www.boost.org/doc/libs/1_72_0/libs/optional/doc/html/boost_optional/tutorial/relational_operators.html) are the reverse of what we want here, so they won't allow us to simplify this:

>In a similar manner, type `optional<T>` is `LessThanComparable` whenever `T` is `LessThanComparable`. The optional object containing no value is compared less than any value of `T`.

src/ripple/rpc/impl/RPCHandler.cpp
```diff
@@ -173,6 +174,30 @@ Status callMethod (
             jtGENERIC, "cmd:" + name);
 
         auto ret = method (context, result);
+        // Handle system-level warnings on admin connections
+        if (context.role == Role::ADMIN)
```
nbougalis: I don't know that we want this _here_, touching _every_ command. I think it should be in the `server_info` handler.

src/ripple/app/misc/impl/AmendmentTable.cpp
```diff
@@ -539,12 +554,36 @@ AmendmentTableImpl::needValidatedLedger (LedgerIndex ledgerSeq)
...
 {
     for (auto& e : enabled)
         enable(e);
+
+    std::lock_guard sl(mutex_);
+    // Since we have the whole list in `majority`, reset the time flag, even if
+    // it's currently set. If it's not set when the loop is done, then any
+    // prior unknown amendments have lost majority.
+    firstUnsupportedExpected_.reset();
+    for (auto& m : majority)
```
seelabs: What do you think about `for (auto const& [amd, t] : majority)` instead. No worries if you prefer this instead.

src/ripple/app/misc/impl/AmendmentTable.cpp
```diff
@@ -539,12 +554,36 @@ AmendmentTableImpl::needValidatedLedger (LedgerIndex ledgerSeq)
...
+    {
+        auto s = add(m.first);
+
+        if (s->enabled)
+            break;
```
seelabs: Don't we want to check if other amendments may have reached and are unsupported. I.e. Shouldn't this be `continue`?

src/ripple/rpc/impl/RPCHandler.cpp
seelabs: Yeah, I have concerns about putting the warning in every command.

src/ripple/app/misc/impl/AmendmentTable.cpp
ximinez: Agreed, though I suppose one way to look at it is that they (presumably) have to put unassigned at one end of the value space or the other. If I was looking for the largest, it would have been just fine.

src/ripple/rpc/impl/RPCHandler.cpp
ximinez: My thinking. rippled currently has the following indications that a server is amendment blocked (https://xrpl.org/amendments.html#amendment-blocked)
1. When submitting a transaction, it returns `amendmentBlocked`.
2. It logs a `LedgerMaster:ERR` message.
3. `server_info` returns a flag.

I figure that being amendment blocked is pretty drastic, and the user is likely to notice pretty quickly, so those notifications are sufficient. But the warning is less drastic, because everything will continue to work normally until it "suddenly" doesn't, so I wanted the warnings to be more aggressive so the user has more of a chance to notice them.  Thus, logging the `LedgerMaster:ERR` message every flag ledger, and including `warnings` on every admin-level RPC command instead of just `submit` and `server_info`. I'm not married to the idea, though, so if you still think it's not a good idea to have it on every comment, I'll move the check, but I would like to still maintain at minimum returning the `warnings` on both `submit` and `server_info`.

src/ripple/app/misc/impl/AmendmentTable.cpp
ximinez: Fixed

src/ripple/app/misc/impl/AmendmentTable.cpp
ximinez: Yes. I totally derped here. Fixed.

src/ripple/rpc/impl/RPCHandler.cpp
ximinez: @nbougalis Per our conversation, the warning is now only included in the results for `server_info` and `server_state`, and added a second warning for amendment blocked with a more verbose description of the problem.

PR #3226, opened 2020-01-14
Account channels RPC bug fixes

seelabs:
This PR fixes two bugs in the `account_channels` RPC command:

1) The marker functionality did not work at all
2) This command returns the channels where a given account is the channel's source. With the new amendment that adds channels to the owner account of the channel's destination, this command would have returned channels where the account was either the source of destination. This patch fixes the command so it behaves as originally intended.

### DIFF
--- a/src/test/app/PayChan_test.cpp
+++ b/src/test/app/PayChan_test.cpp
@@ -1047,6 +1047,41 @@ struct PayChan_test : public beast::unit_test::suite
     }
 
     void
+    testAccountChannelsRPCSenderOnly()
+    {
+        // Check that the account_channels command only returns channels owned by the account
+        testcase("Account channels RPC owner only");
+
+        using namespace test::jtx;
+        using namespace std::literals;
+
+
+        auto const alice = Account("alice");
+        auto const bob = Account("bob");
+        Env env(*this);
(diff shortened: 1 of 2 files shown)


src/test/app/PayChan_test.cpp
```diff
@@ -870,9 +870,221 @@ struct PayChan_test : public beast::unit_test::suite
...
+            chan1PkStr =
+                r[jss::result][jss::channels][0u][jss::public_key].asString();
+        }
+        {
+            auto const r = env.rpc("account_channels", alice.human());
+            BEAST_EXPECT(r[jss::result][jss::channels].size() == 1);
+            BEAST_EXPECT(
+                r[jss::result][jss::channels][0u][jss::channel_id] == chan1Str);
+            BEAST_EXPECT(r[jss::result][jss::validated]);
+            chan1PkStr =
```
mellery451: you assign this value (twice) bet never do anything with it...

seelabs: Fixed in `ee122e24b [fold] rm write-only var`

seelabs:
Rebased onto b4. Ping @thejohnfreeman

PR #3229, opened 2020-01-15
Add unittests for peer_reservations command line interface

scottschurr:
While reviewing a different pull request I noticed a small hole in our command-line interface unit testing.  Adding these tests fills that particular hole.

### DIFF
--- a/src/test/rpc/RPCCall_test.cpp
+++ b/src/test/rpc/RPCCall_test.cpp
@@ -5168,6 +5168,137 @@ static RPCCallTestData const rpcCallTestArray [] =
     })"
 },
 
+// peer_reservations_add -------------------------------------------------------
+{
+    "peer_reservations_add: minimal.", __LINE__,
+    {
+        "peer_reservations_add",
+        "public_key_string"
+    },
+    RPCCallTestData::no_exception,
+    R"({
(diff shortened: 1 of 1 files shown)


jwbusch:
Do there need to be tests for "This has the correct number of arguments, but the arguments are invalid" ?

scottschurr:
@jwbusch, good question.  That set of unit tests only exercises the code that transforms a command line into an RPC call.  In particular, the tests I'm adding target these two functions: https://github.com/ripple/rippled/blob/develop/src/ripple/net/impl/RPCCall.cpp#L875-L893

Since everything comes in from the (actual) command line as a string, it's not really possible to send through an argument that will be detected as invalid at the level we're working.  It will be detected as invalid when the actual RPC call gets sent.  But these tests just return the contents of the RPC call.

Does that make sense?  If not let me know and we can continue the discussion.

jwbusch:
> Since everything comes in from the (actual) command line as a string, it's not really possible to send through an argument that will be detected as invalid at the level we're working. It will be detected as invalid when the actual RPC call gets sent. But these tests just return the contents of the RPC call.

Got it, I wasn't familiar with the specifics of this test, but I see that now, looks good.

PR #3235, opened 2020-01-28
Append default port (51235) in [ips] and [ips_fixed] when no port is specified

gregtatcam:
FIXES: #2861

### DIFF
--- a/src/ripple/overlay/impl/OverlayImpl.cpp
+++ b/src/ripple/overlay/impl/OverlayImpl.cpp
@@ -577,14 +577,9 @@ OverlayImpl::onPrepare()
             for (auto const& addr : addresses)
             {
                 if (addr.port() == 0)
-                {
-                    Throw<std::runtime_error>(
-                        "Port not specified for "
-                        "address:" +
-                        addr.to_string());
-                }
-
-                ips.push_back(to_string(addr));
+                    ips.push_back(to_string(addr.at_port(DEFAULT_PEER_PORT)));
+                else
+                    ips.push_back(to_string(addr));
             }
 
             std::string const base("config: ");
@@ -600,8 +595,19 @@ OverlayImpl::onPrepare()
             [this](
                 std::string const& name,
                 std::vector<beast::IP::Endpoint> const& addresses) {
-                if (!addresses.empty())
-                    m_peerFinder->addFixedPeer(name, addresses);
+                std::vector<beast::IP::Endpoint> ips;
+                ips.reserve(addresses.size());
+
+                for (auto& addr : addresses)
+                {
+                    if (addr.port() == 0)
+                        ips.emplace_back(addr.address(), DEFAULT_PEER_PORT);
+                    else
+                        ips.emplace_back(addr);
+                }
+
+                if (!ips.empty())
+                    m_peerFinder->addFixedPeer(name, ips);
             });
     }
 }
--- a/src/ripple/rpc/handlers/Connect.cpp
+++ b/src/ripple/rpc/handlers/Connect.cpp
@@ -22,6 +22,7 @@
 #include <ripple/net/RPCErr.h>
 #include <ripple/overlay/Overlay.h>
 #include <ripple/protocol/ErrorCodes.h>
+#include <ripple/protocol/SystemParameters.h>
 #include <ripple/protocol/jss.h>
 #include <ripple/rpc/Context.h>
 #include <ripple/rpc/impl/Handler.h>
@@ -53,7 +54,7 @@ doConnect(RPC::JsonContext& context)
     if (context.params.isMember(jss::port))
         iPort = context.params[jss::port].asInt();
     else
-        iPort = 6561;
+        iPort = DEFAULT_PEER_PORT;
 
     auto ip =
         beast::IP::Endpoint::from_string(context.params[jss::ip].asString());
--- a/src/ripple/protocol/SystemParameters.h
+++ b/src/ripple/protocol/SystemParameters.h
@@ -61,4 +61,7 @@ static std::uint32_t constexpr XRP_LEDGER_EARLIEST_SEQ{32570};
 
 }  // namespace ripple
 
+/** Default peer port (IANA registered) */
+inline std::uint16_t constexpr DEFAULT_PEER_PORT{2459};
+
 #endif


MarkusTeufelberger:
Shouldn't this be the IANA registered port and no longer 51235?

src/ripple/overlay/impl/OverlayImpl.cpp
```diff
@@ -538,11 +538,16 @@ OverlayImpl::onPrepare()
             {
                 if (addr.port () == 0)
                 {
-                    Throw<std::runtime_error> ("Port not specified for "
-                        "address:" + addr.to_string ());
+                    JLOG(journal_.debug()) << "Port not specified (bootstrapIps) for address:"
```
nbougalis: The  comment is no longer relevant. I'd remove it.

src/ripple/overlay/impl/OverlayImpl.cpp
```diff
@@ -538,11 +538,16 @@ OverlayImpl::onPrepare()
             {
                 if (addr.port () == 0)
                 {
-                    Throw<std::runtime_error> ("Port not specified for "
-                        "address:" + addr.to_string ());
+                    JLOG(journal_.debug()) << "Port not specified (bootstrapIps) for address:" 
+                                           << addr.to_string ()
+                                           << ", using default port 51235";
+                    beast::IP::Endpoint addr_default(addr.address(), 51235);
+                    ips.push_back (to_string (addr_default));
```
nbougalis: Prefer: `ips.push_back(to_string(addr.at_port(51235))`

src/ripple/overlay/impl/OverlayImpl.cpp
```diff
@@ -558,8 +563,28 @@ OverlayImpl::onPrepare()
                 std::string const& name,
                 std::vector <beast::IP::Endpoint> const& addresses)
             {
-                if (!addresses.empty ())
-                    m_peerFinder->addFixedPeer (name, addresses);
+                std::vector <beast::IP::Endpoint> addresses_update;
+                addresses_update.reserve(addresses.size());
+                for (auto const& addr : addresses)
+                {
+                    if (addr.port () == 0)
```
nbougalis: Same as above. I also don't quite like the introduction of a temporary here. Thought: remove `const&` from the lambda argument list (so `addresses` is passed by copy, potentially leveraging move semantics) and then:

    for(auto& addr : addresses)
    {
        if (addr.port() == 0)
            addr = addr.at_port(51235);
    }

nbougalis:
> Shouldn't this be the IANA registered port and no longer 51235?

I guess for the fixed IPs that's true, since people had to manually specify the default, but I don't know of any servers that listen on 2459 atm, except for mine... so defaulting to it also seems wrong

We need to think of a transition plan. An option might be a config setting that opens 2459 and 51235 to support both new style and legacy connections.

movitto:
> We need to think of a transition plan. An option might be a config setting that opens 2459 and 51235 to support both new style and legacy connections.

Could work but seems redundant, listening on two ports for the same protocol / use case.

Related to the discussion of #2931, perhaps a good solution would be to try both ports on the client side and select the first one that works? As you mentioned there it's not ideal but would provide an interim solution until the majority of the network switched over to the new port at which point we could simplify / just use new standardized port.

src/ripple/beast/net/IPEndpoint.h
```diff
@@ -104,8 +106,13 @@ class Endpoint
...
 
 private:
+    friend class ripple::OverlayImpl;
+    /** Updates the port */
+    void set_port (Port port) const
+    { m_port = port; }
+
     Address m_addr;
-    Port m_port;
+    mutable Port m_port;
```
HowardHinnant: How about something like this to avoid setting `m_port` to `mutable`? https://github.com/HowardHinnant/rippled/commit/f31addfbe5fce0c83e309b26e209437b1705bab6

seelabs: I understand why you (@gregtatcam) made this change. We save copying about 20 bytes per address in `OverlayImpl`. However, to do this we had to:

1) Make `OverlayImpl` a friend. This makes code harder to audit, as I can't assume `OverlayImpl` only uses a public interface. 
2) Make a `set_port(...) const`. It's unusual to have a `set` function that is also `const`
3) Make the `port` member `mutable`. Usually, when something is mutable it's a mutex or cache or somesuch. It's unusual to make a "regular" data member mutable.

I saw @HowardHinnant's patch that solves some of these problems (and I like it). But it still leaves us with a friend function and makes the `addresses` parameter in/out. Those aren't killer, but I'm not sure it's worth it to save copying 20 bytes. I'd probably take the (small) performance hit in the copy rather than change the interfaces.

gregtatcam:
Yes, the main motivation for `set_port` was to reduce the copy but I see you point. Updated with no interface changes and some extra copying.

src/ripple/overlay/impl/OverlayImpl.cpp
gregtatcam: Removed.

src/ripple/overlay/impl/OverlayImpl.cpp
gregtatcam: Changed.

src/ripple/overlay/impl/OverlayImpl.cpp
gregtatcam: This requires change to the `Resolve` class/subclasses to remove the `const`; i.e. semantics changes. Current implementation doesn't require the change at the expense of copying. But this only happens at start up and most likely the number of fixed peers is not substantial.

src/ripple/overlay/impl/OverlayImpl.cpp
```diff
@@ -553,12 +553,9 @@ OverlayImpl::onPrepare()
             for (auto const& addr : addresses)
             {
                 if (addr.port () == 0)
-                {
-                    Throw<std::runtime_error> ("Port not specified for "
-                        "address:" + addr.to_string ());
-                }
-
-                ips.push_back (to_string (addr));
+                    ips.push_back(to_string(addr.at_port(51235)));
```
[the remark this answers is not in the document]
gregtatcam: Done

src/ripple/overlay/impl/OverlayImpl.cpp
```diff
@@ -572,10 +569,21 @@ OverlayImpl::onPrepare()
         m_resolver.resolve (app_.config().IPS_FIXED,
             [this](
                 std::string const& name,
-                std::vector <beast::IP::Endpoint> const& addresses)
+                std::vector <beast::IP::Endpoint> const & addresses)
```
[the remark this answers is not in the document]
gregtatcam: Done

src/ripple/overlay/impl/OverlayImpl.cpp
```diff
@@ -572,10 +569,21 @@ OverlayImpl::onPrepare()
...
-                    m_peerFinder->addFixedPeer (name, addresses);
+                std::vector <beast::IP::Endpoint> ips;
+                ips.reserve(addresses.size());
+
+                for (auto& addr : addresses)
+                {
+                    if (addr.port () == 0)
+                        ips.emplace_back(addr.address(), 51235);
+                    else
+                        ips.emplace_back(addr.address(), addr.port());
```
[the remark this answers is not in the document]
gregtatcam: Done

src/ripple/rpc/handlers/Connect.cpp
```diff
@@ -52,7 +53,7 @@ Json::Value doConnect (RPC::JsonContext& context)
     if(context.params.isMember (jss::port))
         iPort = context.params[jss::port].asInt ();
     else
-        iPort = 6561;
+        iPort = DEFAULT_PEER_PORT;
```
ximinez, line 56: This is going to require a documentation change. https://xrpl.org/connect.html
Not a deal breaker, but important to bring to @mDuo13's attention. We also need to consider if this is going to cause backward-compatibility issues. I doubt it will, because no nodes I know of run on that port.

I'll add the "Documentation" label now.

PR #3237, opened 2020-01-30
Issue 3166: Removing old Ripple.Beast code

p2peer:
1) Removed workaround for waitable timers
2) SSL detection replaced by Boost.Beast functionality
3) ssl_bundle replaced by boost::asio::ssl::stream


nbougalis:
Requesting an additional review by the Beast himself, @vinniefalco.

vinniefalco:
Looking

src/ripple/overlay/Overlay.h
```diff
@@ -45,6 +47,9 @@ class Overlay
     , public beast::PropertyStream::Source
 {
 protected:
+    using socket_type = boost::asio::ip::tcp::socket;
```
vinniefalco: If you're using Boost 1.72, note that the `tcp::socket` type uses the type-erasing executor wrapper `net::executor`. You might want to just use a socket with an explicit strand, which would be declared thusly:
```
using socket_type = net::basic_stream_socket<
    net::ip::tcp, net::strand<net::io_context::executor_type> >;
```

src/ripple/overlay/Overlay.h
```diff
@@ -45,6 +47,9 @@ class Overlay
...
+    using stream_type = boost::asio::ssl::stream <socket_type>;
```
vinniefalco: There's `beast::ssl_stream` which offers significant performance improvements for both HTTP and WebSocket purposes, by consolidating buffers (asio's SSL implementation does not support scatter/gather).

src/ripple/overlay/impl/PeerImp.h
```diff
@@ -554,11 +553,11 @@ PeerImp::PeerImp (Application& app, std::unique_ptr<beast::asio::ssl_bundle>&& s
...
     , p_journal_ (p_sink_)
-    , ssl_bundle_(std::move(ssl_bundle))
-    , socket_ (ssl_bundle_->socket)
-    , stream_ (ssl_bundle_->stream)
+    , stream_ptr_(std::move(stream_ptr))
+    , socket_ (stream_ptr_->next_layer())
+    , stream_ (*stream_ptr_)
     , strand_ (socket_.get_executor())
-    , timer_ (beast::create_waitable_timer<waitable_timer>(socket_))
+    , timer_ (waitable_timer{socket_.get_executor()})
```
vinniefalco: `beast::tcp_stream` has built-in timeouts so if you ever want to get rid of this timer, it can be done.

vinniefalco:
Looks like pretty straightforward refactoring with no gotchas.

src/ripple/overlay/Overlay.h
p2peer: The asio::ssl::stream class has lowest_layer() method which is used by rippled. Is there such a method in beast::ssl_stream?

vinniefalco: There's a free function which does the same thing
https://www.boost.org/doc/libs/1_72_0/libs/beast/doc/html/beast/ref/boost__beast__get_lowest_layer.html

src/ripple/overlay/Overlay.h
p2peer: Am I right that keeping in mind this problem it is recommended to use

`beast::basic_stream< net::ip::tcp, net::strand<net::io_context::executor_type>, unlimited_rate_policy >`

instead of `beast::tcp_stream` as a base for `beast::ssl_stream`?

vinniefalco: Yes

src/ripple/overlay/impl/PeerImp.h
p2peer: I got rid of several timers. But another ones such as the timer in PeerImp and the timer in websockets stuff control of ping-pong activity. So I don't think that such a timers can be removed.

src/ripple/overlay/Overlay.h
p2peer: Since we are using Boost 1.70 as current version and 1.71 as planned to migrate to I think that this change can be postponed.

src/ripple/overlay/Overlay.h
p2peer: Migration to beast::ssl_stream as stream_type and beast::tcp_stream as socket_type done.

PR #3247, opened 2020-02-04
Changed default node_size to huge and default online_delete to 30000

mayurbhandary:
The node_size setting should be `huge` by default so that those who run production servers don't have to worry about changing it. An `online_delete` setting of `2000` is too small and system resources are used to complete deletion too frequently. Setting this to 30000 is more reasonable. Reference ranges for these settings can be found [here](https://xrpl.org/capacity-planning.html#node-size).

### DIFF
--- a/cfg/rippled-example.cfg
+++ b/cfg/rippled-example.cfg
@@ -1158,7 +1158,7 @@ protocol = ws
 #-------------------------------------------------------------------------------
 
 [node_size]
-medium
+huge
 
 # This is primary persistent datastore for rippled.  This includes transaction
 # metadata, account states, and ledger headers.  Helpful information can be
@@ -1173,7 +1173,7 @@ filter_bits=12
 cache_mb=256
 file_size_mb=8
 file_size_mult=2
-online_delete=2000
+online_delete=30000
 advisory_delete=0
 
 # This is the persistent datastore for shards. It is important for the health


MarkusTeufelberger:
Looking at how many issues people seem to run into with their nodes, I'd rather remove this custom setting and let nodes default to the _actual_ default of "tiny".

mayurbhandary:
Hi @MarkusTeufelberger, I created a topic on forum.xpring.io to discuss this more. I listed some arguments for and against these changes it it would be great if we could bounce ideas around on there. https://forum.xpring.io/t/proposed-changes-to-the-default-rippled-cfg-file/82

MarkusTeufelberger:
It would be nice to first open an issue in this repository and discuss changes there before opening pull requests - I don't see how this additional external forum (which seems not to be about `rippled` at all) is supposed to help.

Anyways - the default for node size is actually "tiny" (see https://github.com/ripple/rippled/blob/cc4cefaa4b42c82da4f249eb3f54a77d2bcdc2d6/cfg/rippled-example.cfg#L601) - you just want to set yet another different value that is not the default. Besides that by the way my "huge" full history node has less than 10 GB out of 32 GB RAM used, so I'm not sure how/if that much memory is actually needed.

About increase of online_delete numbers: A shard has 32768 ledgers, so I'd rather see this number go up to ~35k instead of 30000, so there's a good chance that at least the most recent full shard can be served from a node. I expect this kind of settings to be pased out sooner or later anyways though and being replaced by a "this is how much storage the server can use" setting with the whole node store revolving around shards instead of a single big database.

Tl;dr: I'd rather remove the node_size reference in the example config than change it and change the online_delete number to something that's larger than a shard (32768 ledgers).

cfg/rippled-example.cfg
```diff
@@ -1158,7 +1158,7 @@ protocol = ws
 #-------------------------------------------------------------------------------
 
 [node_size]
-medium
+huge
```
nbougalis: I don't think that **`tiny`** or **`huge`** are sensible defaults but, between the two, **`tiny`** is the safer option and may be more appropriate. Ideally, we would try to auto-detect the appropriate value for this instead of just picking a "default" value in the config.

At a minimum we need to pick a value and stick with it. As @MarkusTeufelberger pointed out, we claim the default is "tiny" earlier in the config and, true enough, it is the default if nothing else is specified but something _is_ specified here.

cfg/rippled-example.cfg
```diff
@@ -1173,7 +1173,7 @@ filter_bits=12
 cache_mb=256
 file_size_mb=8
 file_size_mult=2
-online_delete=2000
+online_delete=30000
```
nbougalis: I can take or leave this. I'm curious what the rationale for this value is.

mayurbhandary: I picked that number based on the Capacity Planning Guide [here](https://xrpl.org/capacity-planning.html#recommendation-1) so that more than 1 day of ledger versions is stored.

cfg/rippled-example.cfg
mayurbhandary: That makes sense. I will close this PR.

mayurbhandary:
@MarkusTeufelberger I agree with not changing the default node_size setting, and will close this pull request. Changing the online_delete setting to larger than a shard sounds like a good idea. The post on forum.xpring.io has also been moved to the "Protocols" category instead of "Uncategorized" to be more relevant. There is a [concern](https://forum.xpring.io/t/proposed-changes-to-the-default-rippled-cfg-file/82/2?u=mayurbhandary) that increasing the online_delete setting won't have an impact on the total time spent for deletion so I might not reopen this as an issue.

PR #3249, opened 2020-02-05
Use native cmake unity support:

mellery451:
* Remove hand-rolled unity sources
* unity unavailable for cmake 3.15 and earlier
* remove deprecated target configure option


mellery451:
note to reviewers: you will need cmake 3.16 to test unity builds. Anything earlier should issue a warning and then fall-back to non-unity builds.

src/test/resource/Logic_test.cpp
```diff
@@ -93,11 +94,11 @@ class ResourceManager_test : public beast::unit_test::suite
...
 
-        using namespace std::placeholders;
-
         std::function<Consumer(beast::IP::Endpoint)> ep = limited ?
-            std::bind(&TestLogic::newInboundEndpoint, &logic, _1) :
-            std::bind(&TestLogic::newUnlimitedEndpoint, &logic, _1);
+            std::bind(
+                &TestLogic::newInboundEndpoint, &logic, std::placeholders::_1) :
+            std::bind(
+                &TestLogic::newUnlimitedEndpoint, &logic, std::placeholders::_1);
```
seelabs: What's the motivation for this change? I don't object, I just don't understand why it's needed to support the cmake unity build or removing the deprecated `target` variable.

Builds/CMake/RippledSettings.cmake
```diff
@@ -3,7 +3,18 @@
 #]===================================================================]
 
 option (assert "Enables asserts, even in release builds" OFF)
-option (unity "Creates a build based on unity sources. This is the default" ON)
+
+option (unity "Creates a build using UNITY support in cmake. This is the default" ON)
+if (unity)
+  if (CMAKE_VERSION VERSION_LESS 3.16)
+    message (WARNING "unity option only supported for with cmake 3.16+ (please upgrade)")
+    set (unity OFF CACHE BOOL "unity only available for cmake 3.16+" FORCE)
```
seelabs: We might want to give an error here instead. They asked for unity build, and we can't give it to them. They can always make a non-unity build themselves. What do you think? I do see the arguments to reverting to a non-unity build as you do here, so if you prefer that I'm OK with that too.

mellery451: if `unity` were NOT the default, I might agree....as it is, I'd hate to fail the configure step for default configurations for anyone using cmake less than 3.16.

src/test/resource/Logic_test.cpp
mellery451: one of the compilers (lost track of which one) got confused between `_1` in a boost namespace and in stdlib for unity..obviously this happens now because different TUs are being grouped together that had not previously been.

Builds/CMake/deps/Protobuf.cmake
```diff
@@ -41,6 +41,7 @@ if (local_protobuf OR NOT Protobuf_FOUND)
       -Dprotobuf_DEBUG_POSTFIX=_d
       -Dprotobuf_WITH_ZLIB=$<IF:$<BOOL:${has_zlib}>,ON,OFF>
       $<$<BOOL:${CMAKE_VERBOSE_MAKEFILE}>:-DCMAKE_VERBOSE_MAKEFILE=ON>
+      $<$<BOOL:${unity}>:-DCMAKE_UNITY_BUILD=ON}>
```
thejohnfreeman: I don't feel like ABI-agnostic build options specific to one package should affect build options of other packages, but I'm not going to contest strongly.

mellery451: This is our private/local/nih build of protobuf. It's certainly possible that some external project will not be ready for unity builds (and fail spectacularly), in which case we'd want to not pass the option, or force it to OFF - but so far these few seem to be fine. The goal of unity is to speed up compilation. If you asked for unity AND you doing local deps building, why wouldn't you want unity there too? As always, you can bypass the local build of most deps by pre-building/installing.

seelabs:
Looks like there is a build failure: `src/test/beast/beast_asio_error_test.cpp`

intelliot:
For those looking at this in the future, if you get an error like this:
```
% cmake -G "Unix Makefiles" -DCMAKE_BUILD_TYPE=Debug ..
CMake Error at CMakeLists.txt:23 (message):
  The target option has been removed - use native cmake options to control
  build


-- Configuring incomplete, errors occurred!
See also "/Users/user/ripple/rippled/my_build/CMakeFiles/CMakeOutput.log".
```

You may need to delete your build folder (with the build files) and/or create a new one.

Issue #3042, opened 2019-08-12
Amend Protocol to Always Require Fully-Canonical Signatures


mDuo13:
## Summary
Despite being mostly fixed in 2014, [transaction malleability](https://xrpl.org/transaction-malleability.html) is still theoretically a problem today, because you have to opt in to protection from it by setting **tfFullyCanonicalSig**. We can fix this with an amendment that makes non-fully-canonical signatures invalid regardless of transaction flags.

## Background

The tfFullyCanonicalSig flag was created to allow backwards compatibility with transactions signed with old signing logic that didn't create _fully_-canonical signatures. This design leaves all transactions valid, but makes full protection from transaction malleability (sort-of) opt-in.

As far as I know, every tool in use today generates fully-canonical signatures, but malicious actors could still convert transactions to non-fully-canonical signatures, resulting in a different hash, if those transactions don't have tfFullyCanonicalSig enabled. This could be used [to trick naive transaction submission logic](https://xrpl.org/transaction-malleability.html#exploit-with-malleable-transactions) into sending the same payment multiple times.

By default, `rippled`'s `sign` method supplies a `Flags` field with tfFullyCanonicalSig enabled if you don't provide a `Flags` field. However, tfFullyCanonicalSig **won't be enabled** in the following cases:

- If you use ripple-lib's signing functionality without "preparing" the transaction first (see also: [ripple-lib PR#1028](https://github.com/ripple/ripple-lib/pull/1028))
- If you multi-sign using `rippled`'s `sign_for` method and didn't explicitly set the tfFullyCanonicalFlag
- If you use `rippled`'s `sign` method but explicitly supply a `Flags` field that doesn't include tfFullyCanonicalSig—for example, if you're using another flag, and you don't bitwise-OR it with the tfFullyCanonicalSig value.
- If you're using your own signing logic that doesn't supply tfFullyCanonicalSig.

This state of affairs requires users to take extra steps to protect themselves against transaction malleability in some cases, and leaves many transactions vulnerable because users forgot or misunderstood the necessary steps.

## Proposed Solution

Create an amendment named RequireFullyCanonicalSig (or something along those lines). With this amendment enabled, non-fully-canonical transaction signatures are always considered invalid regardless of whether tfFullyCanonicalSig is specified. (Basically, transactions are treated like tfFullyCanonicalSig is enabled even if it isn't.)

### Compatibility with Legacy Transactions

[...]

PR #3256, opened 2020-02-11
Protocol Amendment: Always Require Fully-Canonical Signatures

movitto:
Addresses #3042 by adding an amendment which if enabled would require full-canonical signatures regardless of whether or not the **tfFullyCanonicalSig** flag is set. Existing functionality will be preserved if amendment is not enabled.

### DIFF
--- a/src/ripple/app/tx/impl/apply.cpp
+++ b/src/ripple/app/tx/impl/apply.cpp
@@ -47,7 +47,7 @@ checkValidity(HashRouter& router,
     if (!(flags & SF_SIGGOOD))
     {
         // Don't know signature state. Check it.
-        auto const sigVerify = tx.checkSign();
+        auto const sigVerify = tx.checkSign(rules);
         if (! sigVerify.first)
         {
             router.setFlags(id, SF_SIGBAD);
--- a/src/ripple/app/tx/impl/apply.cpp
+++ b/src/ripple/app/tx/impl/apply.cpp
@@ -47,7 +47,7 @@ checkValidity(HashRouter& router,
     if (!(flags & SF_SIGGOOD))
     {
         // Don't know signature state. Check it.
-        STTx::RequireFullyCanonicalSig const requireCanonicalSig =
+        auto const requireCanonicalSig =
             rules.enabled(featureRequireFullyCanonicalSig) ?
             STTx::RequireFullyCanonicalSig::yes :
             STTx::RequireFullyCanonicalSig::no;
--- a/src/ripple/app/tx/impl/apply.cpp
+++ b/src/ripple/app/tx/impl/apply.cpp
@@ -47,7 +47,12 @@ checkValidity(HashRouter& router,
     if (!(flags & SF_SIGGOOD))
     {
         // Don't know signature state. Check it.
-        auto const sigVerify = tx.checkSign(rules);
+        STTx::RequireFullyCanonicalSig const requireCanonicalSig =
+            rules.enabled(featureRequireFullyCanonicalSig) ?
+            STTx::RequireFullyCanonicalSig::yes :
+            STTx::RequireFullyCanonicalSig::no;
+
+        auto const sigVerify = tx.checkSign(requireCanonicalSig);
         if (! sigVerify.first)
         {
             router.setFlags(id, SF_SIGBAD);
--- a/src/ripple/protocol/impl/STTx.cpp
+++ b/src/ripple/protocol/impl/STTx.cpp
@@ -187,7 +187,11 @@ std::pair<bool, std::string> STTx::checkSign(Rules const& rules) const
         // at the SigningPubKey.  It it's empty we must be
         // multi-signing.  Otherwise we're single-signing.
         Blob const& signingPubKey = getFieldVL (sfSigningPubKey);
-        ret = signingPubKey.empty () ? checkMultiSign (rules) : checkSingleSign (rules);
+        bool require_fully_canonical =
+            rules.enabled(featureRequireFullyCanonicalSig);
+        ret = signingPubKey.empty () ?
+            checkMultiSign (require_fully_canonical) : checkSingleSign (require_fully_canonical);
+
     }
     catch (std::exception const&)
     {
@@ -251,7 +255,7 @@ STTx::getMetaSQL (Serializer rawTxn,
                 % getSequence () % inLedger % status % rTxn % escapedMetaData);
 }
 
-std::pair<bool, std::string> STTx::checkSingleSign (Rules const& rules) const
+std::pair<bool, std::string> STTx::checkSingleSign (bool require_fully_canonical) const
 {
     // We don't allow both a non-empty sfSigningPubKey and an sfSigners.
     // That would allow the transaction to be signed two ways.  So if both
@@ -262,7 +266,7 @@ std::pair<bool, std::string> STTx::checkSingleSign (Rules const& rules) const
     bool validSig = false;
     try
     {
-        bool const fullyCanonical = (getFlags() & tfFullyCanonicalSig) || rules.enabled(featureRequireFullyCanonicalSig);
+        bool const fullyCanonical = (getFlags() & tfFullyCanonicalSig) || require_fully_canonical;
         auto const spk = getFieldVL (sfSigningPubKey);
 
         if (publicKeyType (makeSlice(spk)))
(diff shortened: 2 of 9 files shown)


movitto:
Incidentally, first #XRPCommunity amendment?

src/ripple/protocol/impl/STTx.cpp
```diff
@@ -186,7 +187,7 @@ std::pair<bool, std::string> STTx::checkSign() const
         // at the SigningPubKey.  It it's empty we must be
         // multi-signing.  Otherwise we're single-signing.
         Blob const& signingPubKey = getFieldVL (sfSigningPubKey);
-        ret = signingPubKey.empty () ? checkMultiSign () : checkSingleSign ();
+        ret = signingPubKey.empty () ? checkMultiSign (rules) : checkSingleSign (rules);
```
nbougalis: I think that for these two **`private`** member functions, it's a better idea to pass an actual `bool` instead of the actual `Rules` object.

MarkusTeufelberger:
> Incidentally, first #XRPCommunity amendment?

Only if you consider Ripple not as a part of the community. ;-)

scottschurr:
The non-unity CI builds are failing.  It looks like, at a minimum, STTx.cpp is missing an include of Feature.h.  Here the the output from the one of the non-unity builds:
```
../../src/ripple/protocol/impl/STTx.cpp:265:89: error: 'featureRequireFullyCanonicalSig' was not declared in this scope
         bool const fullyCanonical = (getFlags() & tfFullyCanonicalSig) || rules.enabled(featureRequireFullyCanonicalSig);
                                                                                         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../../src/ripple/protocol/impl/STTx.cpp:265:89: note: suggested alternative: 'tfFullyCanonicalSig'
         bool const fullyCanonical = (getFlags() & tfFullyCanonicalSig) || rules.enabled(featureRequireFullyCanonicalSig);
                                                                                         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
                                                                                         tfFullyCanonicalSig
../../src/ripple/protocol/impl/STTx.cpp: In member function 'std::pair<bool, std::__cxx11::basic_string<char> > ripple::STTx::checkMultiSign(const ripple::Rules&) const':
../../src/ripple/protocol/impl/STTx.cpp:318:85: error: 'featureRequireFullyCanonicalSig' was not declared in this scope
     bool const fullyCanonical = (getFlags() & tfFullyCanonicalSig) || rules.enabled(featureRequireFullyCanonicalSig);
                                                                                     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../../src/ripple/protocol/impl/STTx.cpp:318:85: note: suggested alternative: 'tfFullyCanonicalSig'
     bool const fullyCanonical = (getFlags() & tfFullyCanonicalSig) || rules.enabled(featureRequireFullyCanonicalSig);
                                                                                     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
                                                                                     tfFullyCanonicalSig

```

src/ripple/protocol/impl/STTx.cpp
movitto: Done.

movitto:
> Lean and mean. I like this. But it needs more thorough unit tests.

@nbougalis could you advise as to how to generate non-canonical signatures for testing purposes? Can ripple-lib be modified to do so?

movitto:
> > Incidentally, first #XRPCommunity amendment?
> 
> Only if you consider Ripple not as a part of the community. ;-)

D'oh! Should've said non-ripple #XRPCommunity ammendment! :man_facepalming:

movitto:
> The non-unity CI builds are failing. It looks like, at a minimum, STTx.cpp is missing an include of Feature.h. Here the the output from the one of the non-unity builds:

Pushed a followon patch fixing the unity build

intelliot:
I don't think ripple-lib can generate non-canonical signatures. It may be possible to modify it to do so, but I'm not knowledgeable about how to do that. Searches turned up the following resources, in case they are helpful. I imagine @nbougalis knows more.
- https://github.com/ripple/ripple-lib/commit/9d6ccdcab1fc237dbcfae41fc9e0ca1d2b7565ca
- https://xrpl.org/transaction-malleability.html#alternate-secp256k1-signatures

scottschurr:
First and foremost, thanks for doing this!  I think you've made a great change.  This is yet another case where you're doing something that we should have taken care of years ago.  I appreciate it.

That said, there are a couple of changes that I think need to be made.  They are:

1. Introducing `Rules` into STTx breaks levelization.  I wouldn't have noticed, but fortunately it caused the validator-keys CI build to break.  I think a good way to fix this would be to check the `Rules` in apply.cpp before the call into STTx.  Then pass a flag into STTx, rather than passing all of the `Rules`.  You can find an example of how I'd approach that problem here: https://github.com/scottschurr/rippled/commit/ff6aa77b30b79e1d76bb4a595891cbfdb7b68990. You're welcome to cherry-pick that, but you can also implement it yourself without much trouble.

2. The harder problem is that we need a unit test to verify the functionality that you're putting in.  This is tricky, since you need a non-fully-canonical transaction signature and none of our tools currently produce that (as far as I know).

    I don't think all is lost, however.  I talked with @nbougalis and possibly, with some multi precision math, you can turn a fully-canonical signature into a valid non-fully-canonical signature.  If you're interested in making a run at that, here are a couple of resources to checkout:

    * This documentation page gives an overview of what the difference is between a canonical signature and a fully-canonical signature: https://xrpl.org/transaction-malleability.html

    * Here's the code that checks how canonical a signature is: https://github.com/ripple/rippled/blob/develop/src/ripple/protocol/impl/PublicKey.cpp#L115-L152. It may give you a sense for how to organize the math.

I haven't actually tried turning a fully-canonical signature into a non-fully-canonical signature using this technique.  So I can't guarantee that it will work.  Also, if you feel like figuring this out is above and beyond the call of duty then let me know and I'll try to put together a couple of unit tests.

I think you have a good change here.  It's very much worth my time to help you get this change into the code base.

Let me know if you want any help.  And thanks again.

movitto:
@nbougalis @scottschurr OK I was able to generate a TX w/ a non-fully-canonical signature (also w/out the tfFullyCanonicalSig flag). I pushed a follow on patch adding a test case w/ that, verifying the transaction fails to be verified when the amendment is enabled, and succeeds when that amendment is disabled.

Also cherry picked @scottschurr's patch introducing the new "RequireFullyCanonicalSig" enum and added a followon patch updating the test case to use that.

Let me know if there is anything else!

movitto:
@scottschurr pushed a patch testing the acceptability of non-canonical txs via the enablement of the amendment vs the state of the boolean in the internal helpers. Also fixes the style issues you pointed out. I believe this addresses the items you pointed out, let me know if otherwise

src/ripple/app/tx/impl/apply.cpp
```diff
@@ -47,7 +47,12 @@ checkValidity(HashRouter& router,
     if (!(flags & SF_SIGGOOD))
     {
         // Don't know signature state. Check it.
-        auto const sigVerify = tx.checkSign();
+        STTx::RequireFullyCanonicalSig const requireCanonicalSig =
+            rules.enabled(featureRequireFullyCanonicalSig) ?
+            STTx::RequireFullyCanonicalSig::yes :
+            STTx::RequireFullyCanonicalSig::no;
+
+        auto const sigVerify = tx.checkSign(requireCanonicalSig);
```
nbougalis: I'd prefer: `auto const requireCanonicalSig = ... whatever ...` instead of `STTx::RequireFullyCanonicalSig const requireCanonicalSig = ... whatever ...`.

Rationale: the precise type of the variable here is irrelevant and doesn't add much. This is, ultimately, a `bool` with some type-safety and user-friendliness sprinked on it, and a programmer can deduce the actual type by examining the ternary.

src/ripple/protocol/impl/STTx.cpp
```diff
@@ -261,7 +266,10 @@ std::pair<bool, std::string> STTx::checkSingleSign () const
     bool validSig = false;
     try
     {
-        bool const fullyCanonical = (getFlags() & tfFullyCanonicalSig);
+        bool const fullyCanonical =
+            (getFlags() & tfFullyCanonicalSig) ||
+            (requireCanonicalSig != RequireFullyCanonicalSig::no);
```
nbougalis: I'd prefer `requireCanonicalSig == RequireFullyCanonicalSig::yes` here (and below in **`checkMultiSign`**). It makes it easier to read and inverting the check just makes it harder (at least for me) to reason about this.

scottschurr: I hear the argument.  Personally, I picked `!= no` because I know that it's possible for a `bool` to have values other than 0 and 1.  So in case there's something unexpected in the `bool` it will set `fullyCanonical == true`.  Using `== yes` doesn't give that guarantee.

But I don't feel strongly about it.  The case is pretty unlikely.

movitto:
@nbougalis pushed a followup patch incorporating your feedback

p2peer:
Linux build OK after rebasing on 1.5.0-b5, unit tests passed.

PR #3258, opened 2020-02-12
Cooperate with CMAKE_MODULE_PATH set by Conan

thejohnfreeman:
Trying to make changes bit-by-bit until I can use Conan to fetch and build dependencies for rippled. This first set of changes:

- Move all Find Modules into `Builds/CMake/deps`. Not sure why they were divided between two directories, but I would like to consolidate.
- Execute the `project` command as early as possible to read the toolchain file. This lets our configuration react to the toolchain file generated by Conan. The first side affect is to put our Find Modules in the `CMAKE_MODULE_PATH` _after_ the ones generated by Conan.


CMakeLists.txt
```diff
@@ -4,6 +4,8 @@ if (POLICY CMP0074)
   cmake_policy(SET CMP0074 NEW)
 endif ()
 
+project (rippled)
```
mellery451: moving `project` here breaks the hacky way of overriding the compiler using the `target` argument. Since I've removed that target stuff in https://github.com/ripple/rippled/pull/3249, I think this should be fine as long as both changes get merged around the same time.

thejohnfreeman:
One failure [with this error](https://travis-ci.com/ripple/rippled/jobs/286467265#L909-L910) while configuring soci:

```
Boost 1.33.1 found.
Found Boost components:
   date_time
CMake Error in /opt/local/cmake-3.9/share/cmake-3.9/Modules/FindBoost.cmake:
  cmake_policy PUSH without matching POP
```

Any ideas? Why is it using the Boost Find Module from CMake 3.9?

mellery451:
the job that is failing is the "min cmake" job, which explicitly builds using the min version of cmake that we require - that's 3.9 currently. I haven't looked closely at the failure, but I will do so tomorrow and see if I can think of something to make this work.

mellery451:
@thejohnfreeman I think https://github.com/mellery451/rippled/commit/107cba33f0ad18b63ca998b29f56a71a6c7d5a3b will fix the 3.9 issue, although I don't fully understand why we only see this now with your small change.

PR #3260, opened 2020-02-12
Add getInfoRippled to packages:

mellery451:
* include info gathering script in rippled linux pkgs
* modify script to output results to single markdown file


MarkusTeufelberger:
How about calling the script `getRippledInfo` instead of `getInfoRippled`?

mellery451:
@MarkusTeufelberger good suggestion. I think I'll drop the `.sh` from the name as well.

MarkusTeufelberger:
I'd add an `ioping` and a `fio` run to the report, if these are available on the machine - similar to `iostat`

PR #3264, opened 2020-02-19
Include field name when diagnosing field not found

### LINKED ISSUE #3263: Include field name when diagnosing "Field not found"
There are 10 instances (in `STObject.{h,cpp}`) of throwing an exception with the description "Field not found" in the presence of an `SField`, which has a name, but without including that name. We should include that name.

thejohnfreeman:
Closes #3263

### DIFF
--- a/src/ripple/protocol/STObject.h
+++ b/src/ripple/protocol/STObject.h
@@ -41,7 +41,11 @@ namespace ripple {
 
 class STArray;
 
-//------------------------------------------------------------------------------
+inline void
+throwFieldNotFound(SField const& field)
+{
+    Throw<std::runtime_error>("Field not found: " + field.getName());
+}
 
 class STObject
     : public STBase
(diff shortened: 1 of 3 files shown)


thejohnfreeman:
It would be nice to amend the "Wrong field type" messages to include the type, but I haven't dug into how to stringify the type. Pointers appreciated.

scottschurr:
If you have an `SField` at hand it has a `fieldName` data member that you can use for the name.  Is that what you are looking for?

thejohnfreeman:
Right, that's what I used (through the method `getName`). The problem is that the previous throws failed to include the name in the exception message.

seelabs:
@thejohnfreeman  Looks like we need to fix the STTx unit tests in two places:
1. https://github.com/thejohnfreeman/rippled/blob/field-not-found/src/test/protocol/STTx_test.cpp#L1286-L1288
2. https://github.com/thejohnfreeman/rippled/blob/field-not-found/src/test/protocol/STTx_test.cpp#L1351-L1353

Other than that, nice change! I'll give a thumbs up after the tests are fixed.

PR #3273, opened 2020-02-28
Simplify HashPrefix by converting to an enum

nbougalis:


src/ripple/protocol/impl/Serializer.cpp
```diff
@@ -52,6 +53,17 @@ int Serializer::add32 (std::uint32_t i)
...
 
+
+int Serializer::add32 (HashPrefix p)
+{
+    // This should never trigger; the size & type of a hash prefix are
+    // integral parts of the protocol and unlikely to ever change.
+    static_assert(std::is_same_v<std::uint32_t,
+        std::underlying_type_t<decltype(p)>>);
+
+    return add32(static_cast<std::uint32_t>(p));
```
HowardHinnant: I believe this should use `safe_cast` as well:
```c++
return add32(safe_cast<std::uint32_t>(p));
```
This change effectively makes the preceding `static_assert` superfluous and it could be removed.  If the underlying type of `HashPrefix` ever changes in such a way that a `uint32_t` could not hold a value of `HashPrefix`, the `safe_cast` will fail to compile.

PR #3274, opened 2020-02-28
Use ephemeral validator public keys in lag ratchet:

nbougalis:
A review of the lag ratchet code revealed that we were using the long-term master public keys of trusted validators, when we should have been using the ephemeral public keys instead.

As a result, the lag ratchet code would be effectively inoperable.


mtrippled:
:+1:

PR #3280, opened 2020-03-02
Clarify the code a comment applies to:

seelabs:
The `seq & 0xff` does not depend of `diff`, so does not belong in the block that
defines `diff`.

There should be no logic changes in this patch.


KrystianM28:
Clarify the code

PR #3282, opened 2020-03-04
Allow non-root users to build packages.

mtrippled:


thejohnfreeman:
I don't understand why `-f` is needed either, or how it is related to the title of the PR. Can you add an explanation to the files themselves (or a relevant README, if one exists)?

mtrippled:
> Looks reasonable, but I always get a little concerned with **`-f`** needs to be passed to anything. Are we leaving junk behind and we just forcibly overwrite them here?

The files are being forcibly overwritten only if the "-u" condition is met (from the man page):
      -u, --update
              copy only when the SOURCE file is  newer  than  the  destination
              file or when the destination file is missing

mtrippled:
> I don't understand why `-f` is needed either, or how it is related to the title of the PR. Can you add an explanation to the files themselves (or a relevant README, if one exists)?

The docker container's files are root owned, so if the user of the container is not root, then the user has no permission to over-write them. As seen in the existing comments, the intention with the "-u" flag that is already present is to update the files if they are newer.

thejohnfreeman:
If the user doesn't have permission to overwrite the file, how does `-f` change that?

mtrippled:
> If the user doesn't have permission to overwrite the file, how does `-f` change that?

They do if the directory mode is 777.

If you want to see what's wrong currently, build an RPM from the tip of develop as a non-root user like so: https://github.com/ripple/rippled/tree/develop/Builds/containers

Then try with this branch.

mtrippled:
Try on Linux.

PR #3287, opened 2020-03-06
Protocol message compression support

gregtatcam:
Peers negotiate compression via HTTP Headers `X-Offer-Compression:lz4`. If the client and the server agree on the compression/algorithm (only lz4 is currently supported) then the protocol messages between the peers are compressed when possible. Messages greater than 70 bytes and MANIFESTS, ENDPOINTS, TRANSACTION, GET_LEDGER, LEDGER_DATA, GET_OBJECT, and VALIDATORLIST messages are compressed. If the compressed message is larger than the uncompressed message then the uncompressed message is sent.  Compression flag and the compression algorithm type are included in the payload header.

### DIFF
--- a/src/ripple/basics/CompressionAlgorithms.h
+++ b/src/ripple/basics/CompressionAlgorithms.h
@@ -1,0 +1,153 @@
+//------------------------------------------------------------------------------
+/*
+    This file is part of rippled: https://github.com/ripple/rippled
+    Copyright (c) 2020 Ripple Labs Inc.
+
+    Permission to use, copy, modify, and/or distribute this software for any
+    purpose  with  or without fee is hereby granted, provided that the above
+    copyright notice and this permission notice appear in all copies.
+
+    THE  SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+    WITH  REGARD  TO  THIS  SOFTWARE  INCLUDING  ALL  IMPLIED  WARRANTIES  OF
+    MERCHANTABILITY  AND  FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+    ANY  SPECIAL ,  DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+    WHATSOEVER  RESULTING  FROM  LOSS  OF USE, DATA OR PROFITS, WHETHER IN AN
+    ACTION  OF  CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+    OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+//==============================================================================
+
+#ifndef RIPPLED_COMPRESSIONALGORITHMS_H_INCLUDED
+#define RIPPLED_COMPRESSIONALGORITHMS_H_INCLUDED
+
+#include <ripple/basics/contract.h>
+#include <lz4.h>
+#include <algorithm>
+
+namespace ripple {
+
+namespace compression_algorithms {
+
+/** Convenience wrapper for Throw
+ * @param message Message to log/throw
+ */
+inline void doThrow(const char *message)
+{
+    Throw<std::runtime_error>(message);
+}
+
+/** LZ4 block compression.
+ * @tparam BufferFactory Callable object or lambda.
+ *     Takes the requested buffer size and returns allocated buffer pointer.
+ * @param in Data to compress
+ * @param inSize Size of the data
+ * @param bf Compressed buffer allocator
+ * @return Size of compressed data, or zero if failed to compress
+ */
+template<typename BufferFactory>
+std::size_t
+lz4Compress(void const* in,
+            std::size_t inSize, BufferFactory&& bf)
+{
+    if (inSize > UINT32_MAX)
+        doThrow("lz4 compress: invalid size");
+
+    auto const outCapacity = LZ4_compressBound(inSize);
+
+    // Request the caller to allocate and return the buffer to hold compressed data
+    auto compressed = bf(outCapacity);
+
+    auto compressedSize = LZ4_compress_default(
+            reinterpret_cast<const char*>(in),
+            reinterpret_cast<char*>(compressed),
+            inSize,
+            outCapacity);
+    if (compressedSize == 0)
+        doThrow("lz4 compress: failed");
+
+    return compressedSize;
+}
+
+/**
+ * @param in Compressed data
+ * @param inSize Size of compressed data
+ * @param decompressed Buffer to hold decompressed data
+ * @param decompressedSize Size of the decompressed buffer
+ * @return size of the decompressed data
+ */
+inline
+std::size_t
+lz4Decompress(std::uint8_t const* in, std::size_t inSize,
+              std::uint8_t* decompressed, std::size_t decompressedSize)
+{
+    auto ret = LZ4_decompress_safe(reinterpret_cast<const char*>(in),
+                               reinterpret_cast<char*>(decompressed), inSize, decompressedSize);
+
+    if (ret <= 0 || ret != decompressedSize)
+        doThrow("lz4 decompress: failed");
+
+    return decompressedSize;
+}
+
+/** LZ4 block decompression.
+ * @tparam InputStream ZeroCopyInputStream
+ * @param in Input source stream
+ * @param inSize Size of compressed data
+ * @param decompressed Buffer to hold decompressed data
+ * @param decompressedSize Size of the decompressed buffer
+ * @return size of the decompressed data
+ */
+template<typename InputStream>
+std::size_t
+lz4Decompress(InputStream& in, std::size_t inSize,
+              std::uint8_t* decompressed, std::size_t decompressedSize)
+{
+    std::vector<std::uint8_t> compressed;
+    std::uint8_t const* chunk = nullptr;
+    int chunkSize = 0;
+    int copiedInSize = 0;
+    auto const currentBytes = in.ByteCount();
+
+    // Use the first chunk if it is >= inSize bytes of the compressed message.
+    // Otherwise copy inSize bytes of chunks into compressed buffer and
+    // use the buffer to decompress.
+    while (in.Next(reinterpret_cast<void const**>(&chunk), &chunkSize))
+    {
+        if (copiedInSize == 0)
+        {
+            if (chunkSize >= inSize)
+            {
+                copiedInSize = inSize;
+                break;
+            }
+            compressed.resize(inSize);
+        }
+
+        chunkSize = chunkSize < (inSize - copiedInSize) ? chunkSize : (inSize - copiedInSize);
+
+        std::copy(chunk, chunk + chunkSize, compressed.data() + copiedInSize);
+
+        copiedInSize += chunkSize;
+
+        if (copiedInSize == inSize)
+        {
+            chunk = compressed.data();
+            break;
+        }
+    }
+
+    // Put back unused bytes
+    if (in.ByteCount() > (currentBytes + copiedInSize))
+        in.BackUp(in.ByteCount() - currentBytes - copiedInSize);
+
+    if ((copiedInSize == 0 && chunkSize < inSize) || (copiedInSize > 0 && copiedInSize != inSize))
+        doThrow("lz4 decompress: insufficient input size");
+
+    return lz4Decompress(chunk, inSize, decompressed, decompressedSize);
+}
+
+} // compression
+
+} // ripple
+
+#endif //RIPPLED_COMPRESSIONALGORITHMS_H_INCLUDED
--- a/src/ripple/overlay/Compression.h
+++ b/src/ripple/overlay/Compression.h
@@ -1,0 +1,103 @@
+//------------------------------------------------------------------------------
+/*
+    This file is part of rippled: https://github.com/ripple/rippled
+    Copyright (c) 2020 Ripple Labs Inc.
+
+    Permission to use, copy, modify, and/or distribute this software for any
+    purpose  with  or without fee is hereby granted, provided that the above
+    copyright notice and this permission notice appear in all copies.
+
+    THE  SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+    WITH  REGARD  TO  THIS  SOFTWARE  INCLUDING  ALL  IMPLIED  WARRANTIES  OF
+    MERCHANTABILITY  AND  FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+    ANY  SPECIAL ,  DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+    WHATSOEVER  RESULTING  FROM  LOSS  OF USE, DATA OR PROFITS, WHETHER IN AN
+    ACTION  OF  CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+    OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+//==============================================================================
+
+#ifndef RIPPLED_COMPRESSION_H_INCLUDED
+#define RIPPLED_COMPRESSION_H_INCLUDED
+
+#include <ripple/basics/CompressionAlgorithms.h>
+#include <ripple/basics/Log.h>
+#include <lz4frame.h>
+
+namespace ripple {
+
+namespace compression {
+
+std::size_t constexpr headerBytes = 6;
+std::size_t constexpr headerBytesCompressed = 10;
+
+enum class Algorithm : std::uint8_t {
+    None = 0x00,
+    LZ4 = 0x01
+};
+
+enum class Compressed : std::uint8_t {
+    On,
+    Off
+};
+
+/** Decompress input stream.
+ * @tparam InputStream ZeroCopyInputStream
+ * @param in Input source stream
+ * @param inSize Size of compressed data
+ * @param decompressed Buffer to hold decompressed message
+ * @param algorithm Compression algorithm type
+ * @return Size of decompressed data or zero if failed to decompress
+ */
+template<typename InputStream>
+std::size_t
+decompress(InputStream& in, std::size_t inSize, std::uint8_t* decompressed,
+           std::size_t decompressedSize, Algorithm algorithm = Algorithm::LZ4) {
+    try
+    {
+        if (algorithm == Algorithm::LZ4)
+            return ripple::compression_algorithms::lz4Decompress(in, inSize,
+                                                                 decompressed, decompressedSize);
+        else
+        {
+            JLOG(debugLog().warn()) << "decompress: invalid compression algorithm "
+                                    << static_cast<int>(algorithm);
+            assert(0);
+        }
+    }
+    catch (...) {}
+    return 0;
+}
+
+/** Compress input data.
+ * @tparam BufferFactory Callable object or lambda.
+ *     Takes the requested buffer size and returns allocated buffer pointer.
+ * @param in Data to compress
+ * @param inSize Size of the data
+ * @param bf Compressed buffer allocator
+ * @param algorithm Compression algorithm type
+ * @return Size of compressed data, or zero if failed to compress
+ */
+template<class BufferFactory>
+std::size_t
+compress(void const* in,
+         std::size_t inSize, BufferFactory&& bf, Algorithm algorithm = Algorithm::LZ4) {
+    try
+    {
+        if (algorithm == Algorithm::LZ4)
+            return ripple::compression_algorithms::lz4Compress(in, inSize, std::forward<BufferFactory>(bf));
+        else
+        {
+            JLOG(debugLog().warn()) << "compress: invalid compression algorithm"
+                                    << static_cast<int>(algorithm);
+            assert(0);
+        }
+    }
+    catch (...) {}
+    return 0;
+}
+} // compression
+
+} // ripple
+
+#endif //RIPPLED_COMPRESSION_H_INCLUDED
--- a/src/ripple/overlay/Message.h
+++ b/src/ripple/overlay/Message.h
@@ -20,6 +20,7 @@
 #ifndef RIPPLE_OVERLAY_MESSAGE_H_INCLUDED
 #define RIPPLE_OVERLAY_MESSAGE_H_INCLUDED
 
+#include <ripple/overlay/Compression.h>
 #include <ripple/protocol/messages.h>
 #include <boost/asio/buffer.hpp>
 #include <boost/asio/buffers_iterator.hpp>
@@ -47,27 +48,61 @@ namespace ripple {
 
 class Message : public std::enable_shared_from_this <Message>
 {
+    using Compressed = compression::Compressed;
+    using Algorithm = compression::Algorithm;
 public:
+    /** Constructor
+     * @param message Protocol message to serialize
+     * @param type Protocol message type
+     */
     Message (::google::protobuf::Message const& message, int type);
 
-public:
-    /** Retrieve the packed message data. */
+    /** Retrieve the packed message data. If compressed message is requested but the message
+     * is not compressible then the uncompressed buffer is returned.
+     * @param compressed Request compressed (Compress::On) or
+     *     uncompressed (Compress::Off) payload buffer
+     * @return Payload buffer
+     */
     std::vector <uint8_t> const&
-    getBuffer () const
-    {
-        return mBuffer;
-    }
+    getBuffer (Compressed tryCompressed);
 
     /** Get the traffic category */
     std::size_t
     getCategory () const
     {
-        return mCategory;
+        return category_;
     }
 
 private:
-    std::vector <uint8_t> mBuffer;
-    std::size_t mCategory;
+    std::vector <uint8_t> buffer_;
+    std::vector <uint8_t> bufferCompressed_;
+    std::size_t category_;
+    std::once_flag once_flag_;
+
+    /** Set the payload header
+     * @param in Pointer to the payload
+     * @param payloadBytes Size of the payload excluding the header size
+     * @param type Protocol message type
+     * @param comprAlgorithm Compression algorithm used in compression,
+     *   currently LZ4 only. If None then the message is uncompressed.
+     * @param uncompressedBytes Size of the uncompressed message
+     */
+    void setHeader(std::uint8_t* in, std::uint32_t payloadBytes, int type,
+            Algorithm comprAlgorithm, std::uint32_t uncompressedBytes);
+
+    /** Try to compress the payload.
+     * Can be called concurrently by multiple peers but is compressed once.
+     * If the message is not compressible then the serialized buffer_ is used.
+     */
+    void compress();
+
+    /** Get the message type from the payload header.
+     * First four bytes are the compression/algorithm flag and the payload size.
+     * Next two bytes are the message type
+     * @param in Payload header pointer
+     * @return Message type
+     */
+    int getType(std::uint8_t const* in) const;
 };
 
 }
--- a/src/ripple/overlay/impl/Message.cpp
+++ b/src/ripple/overlay/impl/Message.cpp
@@ -17,7 +17,6 @@
 */
 //==============================================================================
 
-#include <ripple/basics/safe_cast.h>
 #include <ripple/overlay/Message.h>
 #include <ripple/overlay/impl/TrafficCount.h>
 #include <cstdint>
@@ -25,8 +24,9 @@
 namespace ripple {
 
 Message::Message (::google::protobuf::Message const& message, int type)
-    : mCategory(TrafficCount::categorize(message, type, false))
+    : category_(TrafficCount::categorize(message, type, false))
 {
+    using namespace ripple::compression;
 
 #if defined(GOOGLE_PROTOBUF_VERSION) && (GOOGLE_PROTOBUF_VERSION >= 3011000)
     auto const messageBytes = message.ByteSizeLong ();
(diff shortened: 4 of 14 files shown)


src/ripple/overlay/Message.h
```diff
@@ -48,14 +48,17 @@ namespace ripple {
 class Message : public std::enable_shared_from_this <Message>
 {
 public:
-    Message (::google::protobuf::Message const& message, int type);
+    Message (::google::protobuf::Message const& message, int type, bool compression_enabled);
 
 public:
```
seelabs: Redundant `public`

src/ripple/overlay/Message.h
```diff
@@ -48,14 +48,17 @@ namespace ripple {
...
     /** Retrieve the packed message data. */
     std::vector <uint8_t> const&
-    getBuffer () const
+    getBuffer (bool get_compressed) const
```
seelabs: I'd use an enum class instead of a bool param here.

src/ripple/overlay/Message.h
```diff
@@ -48,14 +48,17 @@ namespace ripple {
...
     {
-        return mBuffer;
+        if (get_compressed && mBufferCompressed.size() > 0)
+            return mBufferCompressed;
+        else
+            return mBuffer;
```
seelabs: I can specify `get_compressed` and still return an uncompressed buffer. We need to document that with a comment, and also rename the variable so it's clear it can return uncompressed data even when `compressed` is specified.

src/ripple/overlay/impl/Message.cpp
```diff
@@ -20,11 +20,12 @@
 #include <ripple/basics/safe_cast.h>
 #include <ripple/overlay/Message.h>
 #include <ripple/overlay/impl/TrafficCount.h>
+#include <ripple/overlay/Compression.h>
```
seelabs: Style nit: includes in alphabetical order (within their categories)

src/ripple/overlay/impl/Message.cpp
```diff
@@ -41,18 +42,53 @@ Message::Message (::google::protobuf::Message const& message, int type)
 
     mBuffer.resize (headerBytes + messageBytes);
 
-    auto ptr = mBuffer.data();
+    auto set_header = [](uint8_t *ptr, decltype(messageBytes) messageBytes, int type, bool compressed=false,
```
seelabs: Consider a type alias for `decltype(messageBytes)`. I don't love `decltype(messageBytes) messageBytes`, it mixes variables from different scopes.

src/ripple/overlay/impl/Message.cpp
```diff
@@ -41,18 +42,53 @@ Message::Message (::google::protobuf::Message const& message, int type)
...
+            uint8_t compr_algorithm=ripple::compression::Algorithm::LZ4) {
+        uint8_t compression = (compressed?0xF0:0) & (0x80 | (compr_algorithm << 4));
+        *ptr++ = static_cast<std::uint8_t>(((messageBytes >> 24) | compression) & 0xFF);
```
seelabs: Is the way we are using the flag here to distinguish compressed/uncompressed data documented anywhere (I didn't see it, but I may have missed it). If not, we need to document it.

src/ripple/overlay/impl/Message.cpp
```diff
@@ -41,18 +42,53 @@ Message::Message (::google::protobuf::Message const& message, int type)
...
 
     if (messageBytes != 0)
-        message.SerializeToArray(ptr, messageBytes);
+        message.SerializeToArray(mBuffer.data() + headerBytes, messageBytes);
+
+    bool compressible = compression_enabled &&
+            (type == protocol::mtMANIFESTS || type == protocol::mtENDPOINTS ||
+             type == protocol::mtTRANSACTION || type == protocol::mtGET_LEDGER || type == protocol::mtLEDGER_DATA ||
+             type == protocol::mtGET_OBJECTS || type == protocol::mtVALIDATORLIST) &&
+            messageBytes > 70;
```
seelabs: What do you think about a lambda and a switch to reduce all the `type ==` noise? Something like:
```

    bool const compressible = [&]{
        if (!compression_enabled || messageBytes <= 70)
            return false;
        switch(type)
        {
            case protocol::mtMANIFESTS:
            case protocol::mtENDPOINTS:
            case protocol::mtTRANSACTION:
            case protocol::mtGET_LEDGER:
            case protocol::mtLEDGER_DATA:
            case protocol::mtGET_OBJECTS:
            case protocol::mtVALIDATORLIST:
                return true;
            default:
                return false;
        }
    }();
```

src/ripple/overlay/impl/Message.cpp
```diff
@@ -41,18 +42,53 @@ Message::Message (::google::protobuf::Message const& message, int type)
...
+
+    if (compressible)
+    {
+        auto *payload = static_cast<void const*>(mBuffer.data() + headerBytes);
+
+        auto res = ripple::compression::compress(
```
seelabs: Maybe a structured binding here? Something like:
```
auto [_, compressedSize] = ...
```

src/ripple/overlay/impl/Message.cpp
```diff
@@ -41,18 +42,53 @@ Message::Message (::google::protobuf::Message const& message, int type)
...
+                payload,
+                messageBytes,
+                [&](std::size_t in_size) {
+                    mBufferCompressed.resize(in_size + headerBytes);
+                    return (mBufferCompressed.data() + headerBytes);
```
seelabs: I'd love to make it clearer how this lambda parameter is used. I need to think about this. It isn't a reason to hold up the PR if we can't think of a better way, but it at least needs a comment.

src/ripple/overlay/impl/ProtocolMessage.h
```diff
@@ -105,6 +127,8 @@ boost::optional<MessageHeader> parseMessageHeader(
 
         for (int i = 0; i != 4; ++i)
             hdr.payload_wire_size = (hdr.payload_wire_size << 8) + *iter++;
+        // clear the compression bits
+        hdr.payload_wire_size &= 0x03FFFFFF;
```
seelabs: The fact that there's a flog like this in the `payload_wire_size` needs to be documented somewhere (it's also possible I missed the comment documenting it).

src/ripple/overlay/impl/ProtocolMessage.h
```diff
@@ -130,7 +154,26 @@ invoke (
     ZeroCopyInputStream<Buffers> stream(buffers);
     stream.Skip(header.header_size);
 
-    if (! m->ParseFromZeroCopyStream(&stream))
+    if (header.compressed)
+    {
+        std::vector<std::uint8_t> serialized_message;
+
+        auto res = ripple::compression::decompress(
```
seelabs: Maybe use a structured binding here?
```
auto [payload, payload_size] = ...
```

src/ripple/app/consensus/RCLConsensus.cpp
```diff
@@ -163,7 +163,7 @@ RCLConsensus::Adaptor::share(RCLCxTx const& tx)
         msg.set_receivetimestamp(
             app_.timeKeeper().now().time_since_epoch().count());
         app_.overlay().foreach (send_always(
-            std::make_shared<Message>(msg, protocol::mtTRANSACTION)));
+            std::make_shared<Message>(msg, protocol::mtTRANSACTION, app_.config().COMPRESSION)));
```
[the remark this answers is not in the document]
gregtatcam: Good idea. The compression could be moved to `getBuffer()`. We will compress/cache the message when `getBuffer()` is called first time requesting compressed message.

src/ripple/app/misc/ValidatorList.h
```diff
@@ -255,7 +257,8 @@ class ValidatorList
         std::string siteUri,
         uint256 const& hash,
         Overlay& overlay,
-        HashRouter& hashRouter);
+        HashRouter& hashRouter,
+        bool compressionEnabled);
```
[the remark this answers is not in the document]
gregtatcam: What is the rational? Avoid passing the wrong value?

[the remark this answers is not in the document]
seelabs: It's easier to see what the parameters mean at the call site (for me). I.e. when I see a call like `applyListBroadcast(manifest, blob, ..., true)` I have to look up what that `true` means. If I see `applyListBroadcast(manifest, blob, ..., compress::enable)` I know what's going on. There is also a strong typing argument to make, tho in this case it's a minor issue.

I understand this is a preference, and other people dislike using enums. If you prefer the bool parameter over an enum I don't object. However, if the function takes more than one `bool` param, then the argument for the enum class gets stronger, as it prevents swapping the parameters by mistake. I.e. I prefer `foo(compress::enable, logging::disable)` over `foo(true, false)` for that reason.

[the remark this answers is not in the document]
gregtatcam: Makes sense. Will do. Would you rather have the enum go into its own include?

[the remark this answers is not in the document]
seelabs: I'm not sure where the best place for this is. I'd usually put it with the function declaration, but this enum will be used in multiple places. Whatever you decide is fine with me.

src/ripple/overlay/impl/ProtocolMessage.h
gregtatcam: It's covered in the requirements: https://github.com/gregtatcam/rippled-specs/blob/master/0002-protocol-compression/README.md

seelabs: Got it. I'd like to document this in the code as well.

src/ripple/app/consensus/RCLConsensus.cpp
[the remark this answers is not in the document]
gregtatcam: Done. Moved compression to `getBuffer()`.

src/ripple/app/misc/ValidatorList.h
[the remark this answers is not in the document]
gregtatcam: Done. Added enum to Compression.h.

src/ripple/basics/CompressionAlgorithms.h
```diff
@@ -0,0 +1,233 @@
...
+    ACTION  OF  CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+    OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+//==============================================================================
+
+#ifndef RIPPLED_COMPRESSIONALGORITHMS_H_INCLUDED
+#define RIPPLED_COMPRESSIONALGORITHMS_H_INCLUDED
+
+#include <ripple/basics/varint_common.h>
+#include <ripple/basics/contract.h>
```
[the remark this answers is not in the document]
gregtatcam: Done.

src/ripple/basics/CompressionAlgorithms.h
```diff
@@ -0,0 +1,233 @@
...
+#include <lz4frame.h>
+#include <array>
+#include <cstring>
+
+namespace ripple {
+
+namespace compression_algorithms {
+
+inline void do_throw(const char *message)
```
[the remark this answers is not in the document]
gregtatcam: Updated.

src/ripple/basics/CompressionAlgorithms.h
```diff
@@ -0,0 +1,233 @@
...
+{
+    Throw<std::runtime_error>(message);
+}
+
+template<typename BufferFactory>
+std::pair<void const*, std::size_t>
+lz4f_compress(void const * in,
+              std::size_t in_size, BufferFactory&& bf)
+{
+    std::pair<void const*, std::size_t> result;
```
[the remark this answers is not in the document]
gregtatcam: Done.

src/ripple/overlay/impl/Message.cpp
gregtatcam: Done.

src/ripple/overlay/impl/Message.cpp
gregtatcam: Added comments.

src/ripple/overlay/impl/ProtocolMessage.h
gregtatcam: Added comments to `setHeader()`.

src/ripple/overlay/impl/ProtocolMessage.h
gregtatcam: Done.

src/ripple/overlay/impl/Message.cpp
gregtatcam: done

src/ripple/basics/CompressionAlgorithms.h
```diff
@@ -0,0 +1,272 @@
...
+#include <ripple/basics/varint_common.h>
+#include <lz4frame.h>
+#include <array>
+#include <cstring>
+
+namespace ripple {
+
+namespace compression_algorithms {
+
+inline void doThrow(const char *message)
```
nbougalis: This function is invoked from multiple places and I don't know that they're all from within a `try/catch` block, or that the `catch` is appropriately scoped. We need to be careful when throwing to ensure that not only will the exception be caught but that continuing after the handler doesn't cause weirdness.

src/ripple/basics/CompressionAlgorithms.h
```diff
@@ -0,0 +1,272 @@
...
+ */
+template<typename InputStream>
+bool
+nextChunk(InputStream &in, std::size_t inSize,
+          void const *& compressedChunk, int &chunkSize,
+          std::vector<std::uint8_t> &buffer)
+{
+    bool res = in.Next(&compressedChunk, &chunkSize);
+
+    if (!res && buffer.size() == 0)
```
nbougalis: I think that `!buffer.empty()` reads better both here and on line 167.

src/ripple/overlay/Message.h
```diff
@@ -45,17 +46,35 @@ namespace ripple {
...
+     */
     std::vector <uint8_t> const&
-    getBuffer () const
+    getBuffer (Compressed compressed)
     {
-        return mBuffer;
+        if (compressed == Compressed::Off)
+            return mBuffer;
+
+        if (!mCompressedRequested)
```
nbougalis: Minor nit, but given the usage, this is probably better named: `alreadyCompressed_`.

Also I'm concerned about thread-safety here; I'm not sure it's a problem, but we need to triple-check that it's not. I'd almost feel better if we just did the compression during construction of the message, so that it never became an issue.

src/test/overlay/compression_test.cpp
```diff
@@ -0,0 +1,374 @@
+//------------------------------------------------------------------------------
```
nbougalis: :+1: for the unit tests here.

src/ripple/app/misc/impl/ValidatorSite.cpp
```diff
@@ -402,7 +402,8 @@ ValidatorSite::parseJsonResponse (
         uri,
         hash,
         app_.overlay(),
-        app_.getHashRouter());
+        app_.getHashRouter(),
+        app_.config().COMPRESSION);
```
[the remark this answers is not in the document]
gregtatcam: You are right it's not needed. It's a left-over from the original compression implementation.

src/ripple/basics/CompressionAlgorithms.h
```diff
@@ -0,0 +1,272 @@
...
+/** LZ4 frame compression. Compressed data is prefixed with varint header containing the size of the original data.
+ * @tparam BufferFactory Callable object or lambda. Takes the required buffer size and returns buffer pointer.
+ * @param in Data to compress
+ * @param inSize Size of the data
+ * @param bf Compressed buffer allocator
+ * @return Pointer and size of compressed data
+ */
+template<typename BufferFactory>
+std::pair<void const*, std::size_t>
+lz4fCompress(void const * in,
```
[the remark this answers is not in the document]
gregtatcam: The problem is that the message is sent as one buffer but received into multiple buffers. If we use `lz4_compress` (block compression) on send then on receive the multiple buffers have to be copied into one buffer to pass to `lz4_decompress` which is undesirable. lz4 frame compression can compress the whole message and decompress one chunk at a time so the multiple buffers can be passed in as zero copy input stream. The only issue with lz4 frame decompression is that it may not consume all input bytes from a chunk. In this case the remaining bytes have to be copied and concatenated with the next input chunk. I have not seen this happen though in my test on 32 million of protocol messages captured from the mainnet. But this case still has to be coded. On another hand the captured sample data also shows that 1) 69% of the messages are received into one buffer; 2) 29% of the messages are received into two buffers; 3) 98% out of the 29% have the message size less than 500 bytes. So the copying might not be a substantial problem after all. Perhaps more data should be captured before we can decide on what compression mode is best to use - block or frame.

src/ripple/overlay/impl/Message.cpp
```diff
@@ -36,23 +40,86 @@ Message::Message (::google::protobuf::Message const& message, int type)
...
-    if (messageBytes != 0)
-        message.SerializeToArray(ptr, messageBytes);
+    auto type = getType(mBuffer.data());
+
+    bool const compressible = [&]{
+        if (messageBytes <= 70)
+            return false;
+        switch(type)
+        {
+            case protocol::mtMANIFESTS:
```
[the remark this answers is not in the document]
gregtatcam: Majority of the messages are not compressible (88% in the sample). The filtering is desirable because both CPU and memory (have to allocate compressed message buffer) will be used. Perhaps we could mitigate maintainability concern by adding compiler warning when a case is missing in the switch?

src/ripple/overlay/impl/Message.cpp
```diff
@@ -36,23 +40,86 @@ Message::Message (::google::protobuf::Message const& message, int type)
...
+
+        auto compressedSize = ripple::compression::compress(
+                payload,
+                messageBytes,
+                [&](std::size_t in_size) { // size of required compressed buffer
+                    mBufferCompressed.resize(in_size + headerBytes);
+                    return (mBufferCompressed.data() + headerBytes);
+                });
+
+        double ratio = 1.0 - (double)compressedSize / (double)messageBytes;
```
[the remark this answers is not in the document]
gregtatcam: `LZ4_compress_default` may return non-zero and still compress into larger buffer. I changed to checking if the compressed size is less than uncompressed.

src/ripple/overlay/impl/Message.cpp
```diff
@@ -36,23 +40,86 @@ Message::Message (::google::protobuf::Message const& message, int type)
...
+ * 46-44    Compression algorithm, value 1-7. Set to 1 to indicate LZ4 compression
+ * 43-42    Set to 0
+ * 41-16    Payload size
+ * 15-0	    Message Type
+*/
+void
+Message::setHeader(std::uint8_t *in, uint32_t messageBytes, int type,
+                   Compressed compressed, std::uint8_t comprAlgorithm)
+{
+    uint8_t compression = (compressed == Compressed::On?0xF0:0) & (0x80 | (comprAlgorithm << 4));
```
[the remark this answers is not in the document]
gregtatcam: The main motivation for using `varint` is to save a couple of bytes since the vast majority of the messages are small. We could move the original size into the message's header but then it has to be fixed at 4 bytes.

src/ripple/basics/CompressionAlgorithms.h
```diff
@@ -0,0 +1,233 @@
...
+{
+    std::pair<void const*, std::size_t> result;
+    std::array<std::uint8_t, varint_traits<std::uint32_t>::max> vi;
+
+    if (in_size > UINT32_MAX)
+        do_throw("lz4f compress: invalid size");
+
+    auto const original_size_bytes = write_varint(vi.data(), in_size);
+
+    std::size_t const out_capacity = LZ4F_compressFrameBound(in_size, NULL);
```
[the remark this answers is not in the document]
gregtatcam: updated

src/ripple/basics/CompressionAlgorithms.h
gregtatcam: updated

src/ripple/basics/CompressionAlgorithms.h
```diff
@@ -0,0 +1,233 @@
...
+        void const *& compressed_chunk, int &chunk_size,
+        std::vector<std::uint8_t> &buffer)
+{
+    bool res = in.Next(&compressed_chunk, &chunk_size);
+
+    if (!res && buffer.size() == 0)
+        return false;
+
+    if (buffer.size() != 0)
+    {
```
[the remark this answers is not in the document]
gregtatcam: updated

src/ripple/basics/CompressionAlgorithms.h
```diff
@@ -0,0 +1,233 @@
...
+        auto sz = buffer.size();
+        chunk_size = ((chunk_size + sz) <= in_size) ? chunk_size : (in_size - sz);
```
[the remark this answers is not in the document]
gregtatcam: We use ZeroCopyInputStream to get the data from the boost's multiple buffers. We are only going to read the payload size of bytes from the stream and return to the caller (PeerImp calls invokeProtocolMessage) the number of bytes that we read. PeerImp then calls `consume` to consume the number of bytes read from the multiple buffer.

src/ripple/basics/CompressionAlgorithms.h
```diff
@@ -0,0 +1,233 @@
...
+        buffer.resize(sz + chunk_size);
+        std::memcpy(buffer.data() + sz, compressed_chunk, chunk_size);
+        compressed_chunk = buffer.data();
+        chunk_size = buffer.size();
+    }
+    else
+        chunk_size = (chunk_size <= in_size) ? chunk_size : in_size;
```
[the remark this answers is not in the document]
gregtatcam: Same as above.

src/ripple/basics/CompressionAlgorithms.h
```diff
@@ -0,0 +1,233 @@
...
+
+    return true;
+}
+
+inline
+void
+update_buffer(std::size_t src_size,
```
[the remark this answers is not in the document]
gregtatcam: updated

src/ripple/basics/CompressionAlgorithms.h
```diff
@@ -0,0 +1,233 @@
...
+        copied += sz;
+        used_size.push_back(data_size);
+    }
+
+    return copied == size;
+}
+
+template<typename InputStream>
+std::size_t
+get_original_size(InputStream& in)
```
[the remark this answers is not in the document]
gregtatcam: It looks good. The only thing is that it makes assumptions about `read_varint` implementation (`val & 0x80`).  What if we overload `read_varint`: `read_varint(void const* buf, std::size_t buflen, std::size_t& n, std::size_t& t)`
It takes extra argument `n` which is the number of bytes read so far from the `buf`. `buf` always points to the start of `varint` header. The function returns -1 if the `buf` is too small. Then `getOriginalSize` is something like:

    template<typename InputStream>
    std::size_t
    getOriginalSize1(InputStream& in)
    {
        constexpr size_t maxSize = varint_traits<std::uint32_t>::max;
        boost::container::static_vector<std::uint8_t, maxSize> vi;
        std::size_t originalSize = 0;
        std::uint8_t const* data = nullptr;
        int dataSize = 0;
        std::size_t used = 0;
        int originalSizeBytes = 0;
        std::size_t chunksBytes = 0;

        while (in.Next(reinterpret_cast<void const**>(&data), &dataSize)) {

            if (dataSize == 0)
                continue;

            auto size = dataSize > (maxSize - vi.size()) ? (maxSize - vi.size()) : dataSize;

            if (!vi.empty())
            {
                // save bytes consumed so far from the stream - need to know how many bytes to back up
                // from the last data chunk
                chunksBytes = vi.size();
                vi.insert(vi.begin() + vi.size(), data, data + size);
                data = vi.data();
                size = vi.size();
            }

            originalSizeBytes = read_varint(data, size, used, originalSize);

            if (originalSizeBytes == 0)
                doThrow("lz4f decompress:: originalSizeBytes == 0");
            else if (originalSizeBytes > 0)
                break;

            if (vi.empty())
                vi.insert(vi.begin(), data, data + size);
        }

        if (dataSize > (originalSizeBytes - chunksBytes))
            in.BackUp(dataSize - (originalSizeBytes - chunksBytes));

        return originalSize;
    }
If the first chunk is large enough then `read_varint` parses the header in first iteration. This is the vast majority of cases. Otherwise the chunks are copied into the `vi`. We only `BackUp` the last chunk.

src/ripple/basics/CompressionAlgorithms.h
gregtatcam: Added try/catch in Compression.h module.

src/ripple/overlay/Message.h
gregtatcam: Changed to `alreadyRequested_`. The message may or may not be compressed.

src/ripple/overlay/impl/Message.cpp
```diff
@@ -17,15 +17,18 @@
...
 
-#include <ripple/basics/safe_cast.h>
 #include <ripple/overlay/Message.h>
 #include <ripple/overlay/impl/TrafficCount.h>
 #include <cstdint>
 
 namespace ripple {
 
+/** Number of bytes in a message header. */
+std::size_t constexpr headerBytes = 6;
```
seelabs: I'd put this variable in an anon namespace

src/ripple/overlay/impl/ProtocolMessage.h
```diff
@@ -83,28 +84,51 @@ struct MessageHeader
 
     /** The type of the message. */
     std::uint16_t message_type = 0;
+
+    /** Indicates whether the message is compressed. */
+    bool compressed = false;
```
seelabs: Now that `Algorithm::None` is added, do we still need this `compressed` member?

src/ripple/overlay/Compression.h
```diff
@@ -0,0 +1,98 @@
...
+std::size_t
+decompress(InputStream& in,
+           std::size_t in_size, BufferFactory&& bf, Algorithm algorithm = Algorithm::LZ4) {
+    try
+    {
+        if (algorithm == Algorithm::LZ4)
+            return ripple::compression_algorithms::lz4fDecompress(in, in_size, std::forward<BufferFactory>(bf));
+        else
+            Throw<std::runtime_error>(
+                    "decompress: invalid compression algorithm");
```
seelabs: I wouldn't throw here. I'd probably `assert`, log the error, and return 0; (ditto for the throw on line 86 too)

src/ripple/overlay/Message.h
```diff
@@ -45,29 +46,79 @@ namespace ripple {
...
+
+    /** Get the message type from the payload header.
+     * First four bytes are the compression/algorithm flag and the payload size.
+     * Next two bytes are the message type
+     * @param in Payload header pointer
+     * @return Message type
+     */
+    int getType(std::uint8_t const* in) const
+    {
+        int type = ((int)*(in + 4) << 8) + *(in + 5);
```
seelabs: Nit: we usually avoid c-style casts.

src/ripple/overlay/impl/Message.cpp
```diff
@@ -36,23 +39,102 @@ Message::Message (::google::protobuf::Message const& message, int type)
...
+}
 
-    *ptr++ = static_cast<std::uint8_t>((messageBytes >> 24) & 0xFF);
-    *ptr++ = static_cast<std::uint8_t>((messageBytes >> 16) & 0xFF);
-    *ptr++ = static_cast<std::uint8_t>((messageBytes >> 8) & 0xFF);
-    *ptr++ = static_cast<std::uint8_t>(messageBytes & 0xFF);
+void
+Message::compress()
+{
+    std::lock_guard<std::mutex> lock(mutex_);
```
seelabs: C++-17 can infer the template parameter for this. Just `std::lock_guard lock(mutex_);` should work.

src/ripple/overlay/Compression.h
```diff
@@ -0,0 +1,98 @@
...
+#define RIPPLED_COMPRESSION_H_INCLUDED
+
+#include <ripple/basics/CompressionAlgorithms.h>
+#include <lz4frame.h>
+
+namespace ripple {
+
+namespace compression {
+
+enum Algorithm : std::uint8_t {
```
seelabs: I prefer enum classes for most things. It's OK if you want to keep it this way tho - it's already in the compression namespace.

src/ripple/basics/CompressionAlgorithms.h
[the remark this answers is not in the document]
seelabs: Yup, I think that looks good. I don't have time to take a close look now, but go ahead and push it and I'll look at it closely when I hit a good stopping point.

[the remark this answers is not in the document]
gregtatcam: Thanks. Will do.

src/ripple/overlay/Message.h
seelabs: Names are hard. I'm fine with `tryCompressed`.

src/ripple/overlay/impl/Message.cpp
seelabs: FYI, I considered suggesting using a streambuf for the compressed data. I think I prefer your solution over that, but I still wish I could think of a better way to do this besides the lambda. At any rate, since I can't think of a better way a comment is good enough for now. Resolving this.

src/ripple/basics/varint_common.h
```diff
@@ -48,24 +48,42 @@ struct varint_traits<T, true>
...
+//          {
+//              std::copy(buf.begin(), buf.begin() + 2, data); // save the chunk
+//              in.Next(&chunk, &chunkSize); // get more data, chunkSize is 100
+//              std::copy(buf.begin() + 2, buf.begin() + 5, chunk);
+//              n = read_varint(buf.data(), 5, n, t); // is good now
+//          }
 //
 template <class = void>
-std::size_t
+int
```
seelabs: Returning an `int` here and using `-1` as a "buffer too small" code may be an issue. It's too easy to assign the result of this function to an unsigned variable (`size_t` in particular).

src/ripple/basics/CompressionAlgorithms.h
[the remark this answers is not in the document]
seelabs: On closer look, the fact that `read_varint` now returns a signed type worries me. If the result is mistaken assigned to an `size_t` (as easy mistake to make, I think), then the code compiles without warning, but does the wrong thing. (left a comment in the code too).

src/ripple/basics/varint_common.h
HowardHinnant: Note that the original signature is still here (below).  This new signature is used only in one place in CompressionAlgorithms.h, where the algorithm needs to make a distinction between overflow and buffer-too-small.

I have the same concern, but I'm so far not coming up with a better solution.   Return `variant<size_t, enum_error_code>`?

gregtatcam: How about instead of overloading changing the name to `read_varint_continue` in order to stress the difference with `read_varint`?

src/ripple/overlay/impl/Message.cpp
gregtatcam: How about passing in callable object instead?

src/ripple/basics/varint_common.h
HowardHinnant: Sounds good to me.

gregtatcam: Would you still prefer returning `variant<size_t, enum_error_code>` or `int` is ok in this case?

seelabs: @gregtatcam @HowardHinnant I think my preference would be to either:

1) Lose the `varint` and always send a 4 byte uint (or maybe 3 bytes). Yes, I'm wasting 2 or 3 bytes much of the time, but it's still not much overhead compared to the expected message size (which I admit I don't know, but it's not going to be _tiny_). But I understand if there's pushback on this suggestion. The point of this PR is compression and I'm suggesting adding bytes. Still, I think the extra 2 or 3 bytes is OK.

2) Instead of reading from a buffer, have an overload of `read_varint` that reads from input stream (i.e. a this implementation, and then send the buffer to the other overload).
```
auto const vi = [&] {
        // read up to maxSize bytes. Stop at the first byte without its high bit
        // set
        constexpr size_t maxSize = varint_traits<std::uint32_t>::max;
        boost::container::static_vector<std::uint8_t, maxSize> buf;

        std::uint8_t const* data = nullptr;
        int data_size = 0;
        while (1)
        {
            if (!in.Next(reinterpret_cast<void const**>(&data), &data_size))
                do_throw("lz4f decompress: header");

            for (int i = 0; i < data_size; ++i)
            {
                if (buf.size() == maxSize)
                {
                    in.BackUp(data_size - i);
                    do_throw("lz4f decompress: invalid input size");
                }
                auto const val = data[i];
                buf.push_back(val);
                if (!(val & 0x80))
                {
                    in.BackUp(data_size - (i + 1));
                    return buf;
                }
            }
        }
    }();
```
3) @HowardHinnant suggestion of returning an error code. I don't care if it's returned as a variant or it's an out parameter.

HowardHinnant: The `variant` solution feels like overkill to me.  But I don't have terribly strong feelings about it either way.  `std::ptrdiff_t` might be better than `int` though.  It is the "signed equivalent" of `std::size_t`.

src/ripple/overlay/impl/Message.cpp
seelabs: I'm not sure I understand. Do you might mean changing the interfaces of `compress` and `decompress` to not take buffer factory as template, but instead take an abstract class who's interface we define and document? And then we'd subclass that abstract class for our use case?

It's not a bad idea, but honestly I think your current "just pass in a lambda" is probably easier to understand than that (unless I misunderstood your idea).

src/ripple/basics/varint_common.h
gregtatcam: I'll go with opt 3 for now. There is still ongoing discussion about what compression algorithm to use (snappy vs lz4) and if it is better to put the fixed uncompressed size into the message header instead of the payload. So opt 1 is still on the table.

src/ripple/overlay/impl/Message.cpp
gregtatcam: Right, the buffer factory could be either the lambda that takes the size of the buffer and returns allocated buffer (current implementation) or the callable object - the abstract class and subclass that does basically the same. The `nodestore` uses the latter approach: https://github.com/ripple/rippled/blob/develop/src/ripple/nodestore/backend/NuDBFactory.cpp#L157. I could implement the same. I do like lambda better. Another thought is to have the buffer factory  to have `resize` method. It's a bit more obscure but then we can pass in the std::vector's reference and it still gives enough flexibility of defining another buffer implementation, just as long as it has resize method.

src/ripple/overlay/Message.h
```diff
@@ -45,29 +46,79 @@ namespace ripple {
...
-public:
-    /** Retrieve the packed message data. */
+    /** Retrieve the packed message data. If compressed message is requested but the message
+     * is not compressible then the uncompressed buffer is returned.
+     * @param compressed Request compressed (Compress::On) or uncompressed (Compress::Off) payload buffer
+     * @return Payload buffer
+     */
     std::vector <uint8_t> const&
-    getBuffer () const
+    getBuffer (Compressed compressed)
```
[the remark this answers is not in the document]
gregtatcam: Moved implementation to the cpp. I can see the point of the `logical const` but I'd rather not use `const` in this case.

src/ripple/overlay/Message.h
gregtatcam: Done

src/ripple/overlay/impl/Message.cpp
gregtatcam: Done.

src/ripple/overlay/impl/Message.cpp
gregtatcam: Done

src/ripple/overlay/impl/ProtocolMessage.h
gregtatcam: Removed `compressed`.

src/ripple/app/misc/ValidatorList.h
```diff
@@ -239,6 +239,8 @@ class ValidatorList
         @param hashRouter HashRouter object which will determine which
             peers not to send to
 
+        @param compressionEnabled is true if compression is enabled in rippled.cfg
+
```
HowardHinnant: Looks like this documentation got lost.

src/ripple/overlay/impl/PeerImp.cpp
```diff
@@ -86,6 +86,7 @@ PeerImp::PeerImp (Application& app, id_t id,
     , slot_ (slot)
     , request_(std::move(request))
     , headers_(request_)
+    , compressionEnabled_(headers_["X-Offer-Compression"] == "lz4" ? Compressed::On : Compressed::Off)
```
HowardHinnant: Does this need `&& app_.config().COMPRESSION` as was done in the other `PeerImp` constructor?

src/ripple/overlay/Compression.h
seelabs: Passing an unsupported algorithm is a logic error. If it happens it'd like to the debug build to hit an assert. Right now, this can never happen, but if a new algorithm type is added and we forget to update this code then it returns "failed to decompress", which sounds like the input data was bad, which is different from a programming logic error.

I'm not suggesting that every possible logic error needs an assert, but adding new values to an enum is a common occurrence, if that silently breaks code I like to put asserts where it would break.

src/ripple/basics/varint_common.h
```diff
@@ -0,0 +1,162 @@
...
+template <class T>
+struct varint_traits<T, true>
+{
+    explicit varint_traits() = default;
+
+    static std::size_t constexpr max =
+            (8 * sizeof(T) + 6) / 7;
+};
+
+// Returns: Number of bytes consumed or -1 if the buffer is small
```
seelabs: This documentation needs to be updated to reflect the new return type.

src/ripple/app/misc/ValidatorList.h
gregtatcam: It's no longer needed. Will delete.

src/ripple/overlay/impl/PeerImp.cpp
gregtatcam: This is constructor for a peer making outbound connection. The header is not going to have the `X-Offer-Compression` header if the compression is not enabled so no need to check for `COMPRESSION`. The other constructor is for the inbound peer. We are not going to send compressed message to that peer if the compression is not enabled so we check for `COMPRESSION`.

src/ripple/overlay/impl/Message.cpp
```diff
@@ -36,23 +36,129 @@ Message::Message (::google::protobuf::Message const& message, int type)
...
+ * 31-0     Uncompressed message size
+*/
+void
+Message::setHeader(std::uint8_t* in, std::uint32_t payloadBytes, int type,
+                   Algorithm comprAlgorithm, std::uint32_t uncompressedBytes)
+{
+    auto h = in;
+
+    auto pack = [](std::uint8_t*& in, std::uint32_t size) {
+        *in++ = static_cast<std::uint8_t>((size >> 24) & 0x0F); // leftmost 4 are ompression bits
```
seelabs: `ompression` -> `compression`?

src/ripple/basics/CompressionAlgorithms.h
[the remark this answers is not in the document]
seelabs: I guess I'm still confused on this issue. It seems like we're leaving `InputStream` in an inconsistent state in the sense if `Next` is called again and we don't rewind here then we skipped bytes.

I understand the current callers won't run across this issue. They will never call `Next` after this function returns. That might be OK for some private function of limited scope, but since this is a general-purpose decompressor I'd this we should leave the stream in a consistent state.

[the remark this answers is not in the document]
gregtatcam: Fixed.

gregtatcam:
Thanks for good comments!

intelliot:
Released in 1.6.0

PR #3291, opened 2020-03-09
Harden validations, report server version & deprecate TTL fields

nbougalis:
This PR contains three separate commits:

### Remove obsolete TTL fields
Proposals and validations include a TTL field. This was part of an earlier idea which never really materialized. The fields were never really used in practice (a TTL value is not set by default). This change simply makes this official: the fields are deprecated, should not be used and are not coming back.

Ripple engineers will be implementing the system outlined by @JoelKatz in this [post](https://www.xrpchat.com/topic/33075-suggestion-reduce-relaying/).

### Include version in validations

Currently, it is impossible to know what version of the software a given validator is running. This information is useful for server operators, for developing who are putting together dashboards, and the community at large.

This change adds a "packed" version in a validation message. The message will, by default, only be sent every 256 ledgers.

### Harden validations

Several changes are included in this PR.

1. Track how many validations a given server publishes for a given ledger. This should be precisely one. When such validations are detected, the server will now log a message to notify server operators of this issue.
2. The commit adds a 64-bit "cookie" in every validation which can be used to _attempt_ to decipher whether multiple validations are the result of accidental misconfiguration vs. outright byzantine behavior. It is important to note that a malicious validator operator can manipulate the cookie.
3. Adds the hash of the last ledger a validator considers to be fully validated in the validation message.

### DIFF
--- a/src/ripple/app/consensus/RCLConsensus.cpp
+++ b/src/ripple/app/consensus/RCLConsensus.cpp
@@ -32,6 +32,7 @@
 #include <ripple/app/misc/TxQ.h>
 #include <ripple/app/misc/ValidatorKeys.h>
 #include <ripple/app/misc/ValidatorList.h>
+#include <ripple/basics/random.h>
 #include <ripple/beast/core/LexicalCast.h>
 #include <ripple/consensus/LedgerTiming.h>
 #include <ripple/nodestore/DatabaseShard.h>
@@ -85,7 +86,14 @@ RCLConsensus::Adaptor::Adaptor(
     , nodeID_{validatorKeys.nodeID}
     , valPublic_{validatorKeys.publicKey}
     , valSecret_{validatorKeys.secretKey}
+    , valCookie_{
+          rand_int<std::uint64_t>(1, std::numeric_limits<std::uint64_t>::max())}
 {
+    assert(valCookie_ != 0);
+
+    JLOG(j_.info()) << "Consensus engine started"
+                    << " (Node: " << to_string(nodeID_)
+                    << ", Cookie: " << valCookie_ << ")";
 }
 
 boost::optional<RCLCxLedger>
@@ -753,41 +761,63 @@ RCLConsensus::Adaptor::validate(
     bool proposing)
 {
     using namespace std::chrono_literals;
+
     auto validationTime = app_.timeKeeper().closeTime();
     if (validationTime <= lastValidationTime_)
         validationTime = lastValidationTime_ + 1s;
     lastValidationTime_ = validationTime;
 
-    STValidation::FeeSettings fees;
-    std::vector<uint256> amendments;
-
-    auto const& feeTrack = app_.getFeeTrack();
-    std::uint32_t fee =
-        std::max(feeTrack.getLocalFee(), feeTrack.getClusterFee());
-
-    if (fee > feeTrack.getLoadBase())
-        fees.loadFee = fee;
-
-    // next ledger is flag ledger
-    if (((ledger.seq() + 1) % 256) == 0)
-    {
-        // Suggest fee changes and new features
-        feeVote_->doValidation(ledger.ledger_, fees);
-        amendments = app_.getAmendmentTable().doValidation(
-            getEnabledAmendments(*ledger.ledger_));
-    }
-
     auto v = std::make_shared<STValidation>(
-        ledger.id(),
-        ledger.seq(),
-        txns.id(),
-        validationTime,
+        lastValidationTime_,
         valPublic_,
         valSecret_,
         nodeID_,
-        proposing /* full if proposed */,
-        fees,
-        amendments);
+        [&](STValidation& v) {
+            v.setFieldH256(sfLedgerHash, ledger.id());
+            v.setFieldH256(sfConsensusHash, txns.id());
+
+            v.setFieldU32(sfLedgerSequence, ledger.seq());
+
+            if (proposing)
+                v.setFlag(vfFullValidation);
+
+            if (ledger.ledger_->rules().enabled(featureHardenedValidations))
+            {
+                // Attest to the hash of what we consider to be the last fully
+                // validated ledger. This may be the hash of the ledger we are
+                // validating here, and that's fine.
+                if (auto const vl = ledgerMaster_.getValidatedLedger())
+                    v.setFieldH256(sfValidatedHash, vl->info().hash);
+
+                v.setFieldU64(sfCookie, valCookie_);
+            }
+
+            // Report our load
+            {
+                auto const& ft = app_.getFeeTrack();
+                auto const fee = std::max(ft.getLocalFee(), ft.getClusterFee());
+                if (fee > ft.getLoadBase())
+                    v.setFieldU32(sfLoadFee, fee);
+            }
+
+            // If the next ledger is a flag ledger, suggest fee changes and
+            // new features:
+            if ((ledger.seq() + 1) % 256 == 0)
+            {
+                // Fees:
+                feeVote_->doValidation(ledger.ledger_->fees(), v);
+
+                // Amendments
+                // FIXME: pass `v` and have the function insert the array
+                // directly?
+                auto const amendments = app_.getAmendmentTable().doValidation(
+                    getEnabledAmendments(*ledger.ledger_));
+
+                if (!amendments.empty())
+                    v.setFieldV256(
+                        sfAmendments, STVector256(sfAmendments, amendments));
+            }
+        });
 
     // suppress it if we receive it
     app_.getHashRouter().addSuppression(
--- a/src/ripple/app/consensus/RCLConsensus.cpp
+++ b/src/ripple/app/consensus/RCLConsensus.cpp
@@ -38,6 +38,7 @@
 #include <ripple/nodestore/DatabaseShard.h>
 #include <ripple/overlay/Overlay.h>
 #include <ripple/overlay/predicates.h>
+#include <ripple/protocol/BuildInfo.h>
 #include <ripple/protocol/Feature.h>
 #include <ripple/protocol/digest.h>
 
@@ -790,6 +791,11 @@ RCLConsensus::Adaptor::validate(
                     v.setFieldH256(sfValidatedHash, vl->info().hash);
 
                 v.setFieldU64(sfCookie, valCookie_);
+
+                // Report our server version every flag ledger:
+                if ((ledger.seq() + 1) % 256 == 0)
+                    v.setFieldU64(
+                        sfServerVersion, BuildInfo::getEncodedVersion());
             }
 
             // Report our load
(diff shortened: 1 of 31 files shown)


nbougalis:
(an unofficial lookover of 99f907b by @bachase would be awesome!)

MarkusTeufelberger:
> When such validations are detected, the server will now log a message to notify server operators of this issue.

Could we optionally also get a full log (or sqlite database, if you feel fancy) of all validation messages received? As far as I know, this data is currently only available on WebSocket streams.

nbougalis:
> > When such validations are detected, the server will now log a message to notify server operators of this issue.
> 
> Could we optionally also get a full log (or sqlite database, if you feel fancy) of all validation messages received? As far as I know, this data is currently only available on WebSocket streams.

We used to record validations in an SQL table, and we removed that functionality (see https://github.com/ripple/rippled/commit/c5a95f1eb5a5eedc7ece46ad26505209c1703486), because it was not really used by anyone/anything.

I think (and this is just my personal position) that the best option is to use the subscription and to index the received data separately. But would we, for example, report conflicting validations in the subscription stream?

Happy to discuss what to do here.

src/ripple/consensus/Validations.h
```diff
@@ -167,7 +167,11 @@ enum class ValStatus {
     /// Not current or was older than current from this node
     stale,
     /// A validation violates the increasing seq requirement
-    badSeq
+    badSeq,
+    /// Multiple validations for the same sequence for same from multiple validators
```
bachase: Typo, extra `for same`?

src/ripple/consensus/Validations.h
```diff
@@ -590,9 +606,45 @@ class Validations
...
+                        std::min(it->second.signTime(), val.signTime());
+
+                    if (diff > std::chrono::minutes{10} &&
+                        val.signTime() > it->second.signTime())
+                        it->second = val;
+                }
+
+                // If the ledger hash matches, it must be a validation for
+                // the same ledger:
+                bool const same = (it->second.ledgerID() == val.ledgerID());
```
bachase: I'm unclear how these conditions wouldn't already be blocked by the `badSeq` one above. That is, the `enforcer` requires a strictly increasing sequence number seen by a validator. So I think it already filters these alternate outcomes.

src/ripple/consensus/Validations.h
```diff
@@ -590,9 +606,45 @@ class Validations
...
+                if (!inserted)
+                {
+                    // Check if it's expired; if it is, replace it. We need to
+                    // calculate the difference between the sign times for the
+                    // two validations:
+                    auto const diff =
+                        std::max(it->second.signTime(), val.signTime()) -
+                        std::min(it->second.signTime(), val.signTime());
+
+                    if (diff > std::chrono::minutes{10} &&
```
bachase: Consider using one of the other parameterized durations over this hardcoded 10 minutes.

src/ripple/consensus/Validations.h
bachase: If you do want to check these, an option would be to move them above before `badSeq` returns.

src/ripple/app/consensus/RCLConsensus.cpp
```diff
@@ -745,40 +752,60 @@ RCLConsensus::Adaptor::validate(RCLCxLedger const& ledger,
...
     using namespace std::chrono_literals;
+
     auto validationTime = app_.timeKeeper().closeTime();
     if (validationTime <= lastValidationTime_)
         validationTime = lastValidationTime_ + 1s;
     lastValidationTime_ = validationTime;
 
-    STValidation::FeeSettings fees;
-    std::vector<uint256> amendments;
+    auto v = std::make_shared<STValidation>(
```
bachase: :+1: Much cleaner

src/test/consensus/Validations_test.cpp
```diff
@@ -1080,6 +1080,14 @@ class Validations_test : public beast::unit_test::suite
         }
     }
 
+    void
+    testDifferentCookies()
```
bachase: I think testing this would be useful; I'm a little rusty on the test harness here but I think it should be feasible and can try to help.

src/ripple/consensus/Validations.h
carlhua: Interesting comment here @bachase . I would argue that this is a valid check due to malicious/byzantine behavior. An attacker can increment the seq without changing the content. So this check can help us catch those byzantine behaviors is what I am thinking. I was thinking maybe the signature check would block the logic here but it seems to me that sfLedgerHash is just one field within the digest so a byzantine node can still sign the same validation with a different sequence and this logic here would catch it

src/ripple/app/consensus/RCLValidations.cpp
```diff
@@ -191,9 +192,26 @@ handleNewValidation(Application& app,
     if (masterKey)
     {
         ValStatus const outcome = validations.add(calcNodeID(*masterKey), val);
+
         if(j.debug())
             dmp(j.debug(), to_string(outcome));
 
+        if (outcome == ValStatus::conflicting && j.fatal())
```
carlhua: Couple thoughts here:
1. Consider this condition to be one of the criteria to be added to nUNL @pwang200  (future, but if we agree we should open an issue to track)
2. Consider this condition to be folded into peer selection/overlay enhancement - this is a byzantine behavior and we should consider disconnecting from this node or at least not connect to this node. Currently, this might not be a good idea but in the future when we are actively relaying validations from untrusted peers, we should consider keep records and potentially not connecting to these bad nodes. My argument here is that this allows us to eliminate bad nodes from the network.

src/test/consensus/Validations_test.cpp
```diff
@@ -1080,6 +1080,14 @@ class Validations_test : public beast::unit_test::suite
...
+    {
+        testcase("Multiple Validations");
```
carlhua: I think it would be good to have more unit tests here. In addition, I would ask @manojsdoshi to consider to have automated tests for this PR to force the logic to execute on QE network. This should never happen naturally on the real network but it would be useful to see it in action in test.

src/ripple/consensus/Validations.h
bachase: @carlhua but doesn't this check presume the two validations being compared have the same sequence number? They both come from the `bySequence_` map in [Line 614](https://github.com/ripple/rippled/pull/3291/files/31796963fa7ca6226751366b78d324775f0fe584..330f35b5f5712c758057fed0a962d5d831960ad8#diff-46f83c3cd15dd825b63068c0ab4dea00R616).

carlhua: @bachase you are right!!  These checks would still guard against the scenarios depicted in the code right?
1. Same sequence, hashs match but cookies don't so we say its multiple
2. Same sequence, hashes don't match, we say its conflicting

mDuo13:
I don't see any documentation impacts beyond the new amendment. Am I missing anything? Are there changes to the fields provided in any of the API subscription streams for example?

src/ripple/consensus/Validations.h
nbougalis: I sorted this out already. I this I misfolded some of my work before pushing. See updated commit.

src/test/consensus/Validations_test.cpp
nbougalis: Appreciate the offer. I'll work with someone else to try and get these tests so that more people learn about consensus instead of having all the information locked up in the heads of three or four people.

src/ripple/app/consensus/RCLValidations.cpp
```diff
@@ -191,13 +193,30 @@ handleNewValidation(Application& app,
     if (masterKey)
     {
         ValStatus const outcome = validations.add(calcNodeID(*masterKey), val);
+
         if(j.debug())
             dmp(j.debug(), to_string(outcome));
 
-        if (outcome == ValStatus::badSeq && j.warn())
+        if (outcome == ValStatus::conflicting && j.fatal())
```
bachase, line 200: Note that this checks `j.fatal()`, but logs below with `j.warn()`.  Same for line 208

src/ripple/consensus/Validations.h
```diff
@@ -587,11 +603,46 @@ class Validations
...
+                    seqit->second = val;
+            }
+
+            // Enforce monotonically increasing sequences for validations
+            // by a given node:
+            if (auto& enf = seqEnforcers_[nodeID]; !enf(now, val.seq(), parms_))
+            {
+                // If the ledger hash matches, it must be a validation for
+                // the same ledger:
+                bool const same = (seqit->second.ledgerID() == val.ledgerID());
```
bachase, line 631: Consider renaming to `sameID` to make it clear what "same" thing you are referring to.

src/ripple/consensus/Validations.h
```diff
@@ -587,11 +603,46 @@ class Validations
...
+
+                // Two validations for the same ledger but with different
+                // cookies. This is probably accidental misconfiguration.
+                if (same && seqit->second.cookie() != val.cookie())
+                    return ValStatus::multiple;
+
+                // Two validations for the same sequence but for different
+                // ledgers. This could be the result of misconfiguration
+                // but it can also mean a Byzantine validator.
+                if (!same)
```
bachase, line 641: If they have different ledger IDs, what guarantees they still have the same sequence number at this spot in the code? Do you also have to check `seqInserted`?

src/ripple/proto/ripple.proto
```diff
@@ -210,9 +214,14 @@ message TMValidatorList
...
-    optional bool checkedSignature  = 2;    // node vouches signature is correct
-    optional uint32 hops            = 3;    // Number of hops traveled
+    // The serialized validation
+    required bytes validation = 1;
+
+    // node vouches signature is correct
+    optional bool checkedSignature = 2 [deprecated = true];
+
+    // Number of hops traveled
+    optional uint32 hops = 3           [deprecated = true];
```
gregtatcam, line 224: Could `hops` still be useful for the `reduce relaying` task? We can infer the preferred peer to receive TMValidation and TMPropose from as the ones with the smallest number of hops.

src/ripple/consensus/Validations.h
nbougalis: I'll double-check this. Thanks.

nbougalis:
I _think_ I've addressed all your concerns, except for improved unit test @bachase. I've avoided rebasing this to make reviewing easier; once you sign-off, I'll do it and force-push the branch.

In the meantime, I'm running a rebased version against the network.

src/test/consensus/Validations_test.cpp
seelabs: I agree that this is important. I'm not sure it needs to be addressed before this is merged (that's up to Brad and Nik), but it would be good to assign this task to someone before this is merged, just so we don't lose track of the task.

pwang200: I will be happy to work on it, probably after the negative UNL network tests, if that is OK.

nbougalis:
I'm going to rebase this against *b4* and squash.

PR #3292, opened 2020-03-09
Unconditionalize 2017 amendments

scottschurr:
Amendments that were enabled in 2017 have now been on the books and active for over two years.  Once an amendment is enabled it cannot be revoked.  So the only situations where these amendments are disabled is when replaying ledgers (and their contained transactions) from before 2018.

It is uncommon, but not unheard of, to want to replay ledgers that are more than two years old.  If these changes are accepted into the code base then a user wanting to replay a ledger from before 2018 will need to build an old version of the rippled code.  This requirement is reinforced by a message that the rippled software generates if it is asked to replay a ledger from before 2018.

This pull request consists of a collection of small commits.  Each commit unconditionalizes one specific amendment.  Each commit is, to all intents and purposes, standalone.  Reviewing the commits one at a time may make the review process easier.  My intention is that these separate commits will not be squashed before merging.  That way, in case one of the commits has an error, identifying the problem using a binary search will be easier.


src/ripple/protocol/Feature.h
```diff
@@ -366,24 +365,23 @@ foreachFeature(FeatureBitset bs, F&& f)
 
 extern uint256 const featureTickets;
 extern uint256 const featureOwnerPaysFee;
-extern uint256 const featureCompareFlowV1V2;
-extern uint256 const featurePayChan;
+extern uint256 const retiredPayChan;
```
nbougalis: I think it makes sense to group the unconditionalized amendments together at the bottom, with a comment explaining that they are amendments that have been activated and are live and the "pre-amendment" code has been removed.

scottschurr:
@nbougalis, thanks for the review comments.  I believe all your comments have been addressed.  If I missed anything please let me know.

BTW, despite how Github displays them, the [FOLD] commits are actually placed immediately adjacent to their target in the commit stack.  I dunno why Github shows them in time order not in commit order, but here we are...

scottschurr:
Hi @pwang200, thanks for the review.

> About the warning in ApplicationImp::loadOldLedger(), when an really old ledger is replayed, the code logs a warning message and continues. What are the possible outcomes?

First, it's important to know that the warning is associated with replaying ledgers more than two years old.  Replaying a ledger is a debugging technique that must be actively requested by someone using rippled.  And ledger replay happens on a stand-alone rippled instance; it never happens on the live network.

With that background, let me get back to your original question.  There are three possible outcomes.  From most likely to least likely they are:

1. The ledger replays exactly like it did the first time it ran on the network.  Most amendments affect bugs in corner cases or enable new features.  Most transactions don't live in the corner cases and old (validated) transactions can't use features that were enabled after the transaction was validated.

2. All transactions in the replayed ledger are replayed, but the outcomes of some of the transactions are different from the outcomes when the ledger was originally produced.  The difference in outcomes can be seen by comparing the metadata.  This would happen because amendments that were not in force at the time the original transaction was validated are now unconditionally in force.  So a transaction that was working in one of those amendment-affected corner cases would exhibit the new behavior.

3. The server could crash or show some other undesirable behavior.  I'm only including this case to be complete.  We don't currently know of any way to cause this, but it can't be ruled out.  The server might crash because the old ledger contains some data configuration that is no longer allowed and is not present in current ledgers.  The ledger is robust to most unexpected data configurations, but there is no way to guarantee that it can correctly handle all unexpected data configurations.

Even in the crashing case nothing bad happens to the network, since the replay is done on a stand-alone instance

If someone wants to understand why an old transaction originally played the way it did, and they run across case 2 or 3, then they would need to build an old version of the server from source and use it for the replay.  That's also the advice that the warning provides.

src/ripple/protocol/Feature.h
```diff
@@ -369,7 +369,7 @@ extern uint256 const featurePayChan;
 extern uint256 const featureFlow;
 extern uint256 const featureCompareTakerFlowCross;
 extern uint256 const featureFlowCross;
-extern uint256 const featureCryptoConditions;
+extern uint256 const retiredCryptoConditions;
```
miguelportilla: I noticed no rule enabled calls were removed. I assume the amendment was never proposed?

src/ripple/protocol/impl/Feature.cpp
```diff
@@ -161,7 +161,7 @@ uint256 const featureFlow = *getRegisteredFeature("Flow");
 uint256 const featureCompareTakerFlowCross = *getRegisteredFeature("CompareTakerFlowCross");
 uint256 const featureFlowCross = *getRegisteredFeature("FlowCross");
 uint256 const retiredCryptoConditions = *getRegisteredFeature("CryptoConditions");
-uint256 const featureTickSize = *getRegisteredFeature("TickSize");
+uint256 const retiredTickSize = *getRegisteredFeature("TickSize");
```
miguelportilla: I am guessing a retired tick size is `full` :)

src/ripple/protocol/Feature.h
scottschurr: I'm not entirely sure about the history.  Apparently the amendment was already unconditionalized before I got here.  It looks like it had to do with `SusPay`, which was never enabled (https://xrpl.org/known-amendments.html#cryptoconditions).  But the CryptoConditions amendment is live and on the ledger.

src/ripple/ledger/impl/PaymentSandbox.cpp
```diff
@@ -194,16 +193,12 @@ PaymentSandbox::balanceHook (AccountID const& account,
...
-        // to compute usable balance just slightly above what the ledger
-        // calculates (but always less than the actual balance).
-        adjustedAmt = std::min(adjustedAmt, minBal);
-    }
 
+    // The adjusted amount should never be larger than the balance. In
+    // some circumstances, it is possible for the deferred credits table
+    // to compute usable balance just slightly above what the ledger
+    // calculates (but always less than the actual balance).
+    auto adjustedAmt = std::min({amount, lastBal - delta, minBal});
```
miguelportilla: Excellent comment!

src/ripple/app/paths/impl/PaySteps.cpp
```diff
@@ -133,255 +131,7 @@ toStep (
...
-        }
-    }
-
-    return {tesSUCCESS, std::move (result)};
-}
-
-
-std::pair<TER, Strand>
-toStrandV2 (
+toStrand (
```
miguelportilla: Great reduction in unnecessary code!

src/ripple/ledger/impl/View.cpp
```diff
@@ -713,13 +713,10 @@ dirAdd (ApplyView& view,
...
-    {
-        if (strictOrder)
-            return view.dirAppend(dir, uLedgerIndex, fDescriber);
-        else
-            return view.dirInsert(dir, uLedgerIndex, fDescriber);
-    }
+    if (strictOrder)
+        return view.dirAppend(dir, uLedgerIndex, fDescriber);
+    else
+        return view.dirInsert(dir, uLedgerIndex, fDescriber);
```
miguelportilla: Consider eliminating the else statement

`if (strictOrder)
    return view.dirAppend(dir, uLedgerIndex, fDescriber);
return view.dirInsert(dir, uLedgerIndex, fDescriber);`

src/ripple/app/tx/impl/Escrow.cpp
```diff
@@ -488,10 +477,10 @@ EscrowFinish::doApply()
     }
 
     // Remove escrow from recipient's owner directory, if present.
-    if (ctx_.view ().rules().enabled(fix1523) && (*slep)[~sfDestinationNode])
+    if (auto const optPage = (*slep)[~sfDestinationNode]; optPage)
```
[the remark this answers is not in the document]
miguelportilla: I prefer the first form and lately have started using it.

src/ripple/app/tx/impl/Escrow.cpp
```diff
@@ -564,11 +559,10 @@ EscrowCancel::doApply()
...
     {
-        auto const page = (*slep)[sfDestinationNode];
         if (! ctx_.view().dirRemove(
-                keylet::ownerDir((*slep)[sfDestination]), page, k.key, true))
+                keylet::ownerDir((*slep)[sfDestination]), *optPage, k.key, true))
```
miguelportilla: Nit: beyond 80 cols, consider removing one tab indentation.

src/test/rpc/AccountSet_test.cpp
```diff
@@ -33,6 +33,8 @@ class AccountSet_test : public beast::unit_test::suite
 
     void testNullAccountSet()
     {
+        testcase ("No AccountSet");
```
miguelportilla: Great job giving these names.

src/test/rpc/AccountSet_test.cpp
```diff
@@ -284,45 +303,36 @@ class AccountSet_test : public beast::unit_test::suite
...
     }
 
     void testGateway()
     {
+        testcase ("Gateway");
+
         using namespace test::jtx;
-        auto runTest = [](Env&& env, double tr)
+
+        auto runTest = [this](double tr)
```
miguelportilla: I think `transferRate` or `rate` might be an improvement over `tr`.

src/test/rpc/AccountSet_test.cpp
```diff
@@ -331,53 +341,82 @@ class AccountSet_test : public beast::unit_test::suite
...
-        runTest (Env{*this, noFix1201}, 1.02);
-        runTest (Env{*this, noFix1201}, 1);
-        runTest (Env{*this, noFix1201}, 2);
-        runTest (Env{*this, noFix1201}, 2.1);
-        runTest (Env{*this, supported_amendments()}, 1.02);
-        runTest (Env{*this, supported_amendments()}, 2);
+        // Test gateway with a variety of allowed transfer rates
+        for (double tr = 1.0; tr <= 2.0; tr += 0.03125)
+        {
+            runTest (tr);
```
miguelportilla: Considering the `runTest` lambda is only called from here, why not just move the code inside the loop?

src/test/rpc/AccountSet_test.cpp
```diff
@@ -331,53 +341,82 @@ class AccountSet_test : public beast::unit_test::suite
...
 
-        // Test gateway when amendment is set after transfer rate
+        // Since fix1201 was enabled on Nov 14 2017 a rate in excess of
+        // 2.0 has been blocked by the transactor.  But there are a few
+        // accounts on the MainNet that have larger-than-currently-allowed
+        // TransferRates.  We'll bypass the transactor so we can check
+        // operation of these legacy TransferRates.
+        //
+        // Two out-of-bound values are currently in the ledger (March 2020)
+        // They are 4.0 and 4.294967295.  So those are the values we test.
```
miguelportilla: Excellent comments!

src/test/rpc/AccountSet_test.cpp
```diff
@@ -331,53 +341,82 @@ class AccountSet_test : public beast::unit_test::suite
...
+        for (double tr : {4.0, 4.294967295})
```
miguelportilla: Consider renaming `tr` to `transferRate` or `rate`.

src/test/rpc/AccountSet_test.cpp
```diff
@@ -331,53 +341,82 @@ class AccountSet_test : public beast::unit_test::suite
...
 
+            Env env (*this);
             env.fund(XRP(10000), gw, alice, bob);
-            env.trust(USD(3), alice, bob);
-            env(rate(gw, tr));
             env.close();
-            env.enableFeature(fix1201);
+            env.trust(USD(10), alice, bob);
+            env.close();
+            env(rate(gw, 2.0));
```
miguelportilla: Should the rate be set here to `tr`?

src/ripple/ledger/impl/View.cpp
miguelportilla: Ah never mind, I see you did that in a later commit.

scottschurr:
Squashed all fixes into their appropriate commits and rebased to 1.6.0-b1.

PR #3293, opened 2020-03-10
Set applied field of submit response to true on tec errors

cjcobb23:
Fixes: https://github.com/ripple/rippled/issues/3285

---
This change is [Reviewable](https://reviewable.io/reviews/ripple/rippled/3293)


mDuo13:
Code diff looks good, but I'll test it to be sure

PR #3296, opened 2020-03-11
Attempt to maintain history to the earliest persisted ledger.

mtrippled:


mtrippled:
Currently, with online_delete disabled, ledger history is maintained only based on the ledger_history value. This means that hosts that have ledger gaps that exceed that value will essentially lose old data because rippled will not fetch that far back. However, with online_delete enabled, ledger history is maintained back to the point at which online_delete is configured to maintain ledgers. This makes behavior consistent both with and without online_delete enabled.

src/ripple/app/ledger/impl/LedgerMaster.cpp
```diff
@@ -1701,17 +1711,21 @@ bool
...
-    // is requested by the advisory delete setting, or
-    // is within our configured history range
+    // is the greater of both the minimum persisted ledger, if any,
+    // and the minimum ledger for online deletion to retain, or
+    // is within our configured history range.
 
     bool ret (candidateLedger >= currentLedger ||
-        ((ledgerHistoryIndex > 0) &&
-            (candidateLedger > ledgerHistoryIndex)) ||
+        candidateLedger >= std::max(
```
carlhua: @mtrippled  and I discussed this, and this comment is used to capture our discussion. Here we are changing the default behavior of rippled. Before rippled would not go out there to grab gaps and now it will. if online delete is not enabled, users can see an increase in disk usage. I suggested a cfg enhancement to make this behavior opt-in, but I also think leaving gaps not desireable. I do think this has a doc impact @mDuo13 @nbougalis

src/ripple/app/misc/SHAMapStoreImp.cpp
```diff
@@ -171,7 +171,7 @@ SHAMapStoreImp::SHAMapStoreImp(
     , scheduler_ (scheduler)
     , journal_ (journal)
     , working_(true)
-    , canDelete_ (std::numeric_limits <LedgerIndex>::max())
+    , canDelete_ (0)
```
cjcobb23: This change would seem to affect the behavior of this bit of code: https://github.com/ripple/rippled/blob/develop/src/ripple/app/misc/SHAMapStoreImp.cpp#L368

Is the behavior of that bit of code still correct?

mtrippled: Yes, it's still correct. If advisory_delete is enabled, that value is set to 0 initially. The only reason that it was set to std::numeric_limits <LedgerIndex>::max() originally was to be fed to shouldAcquire() that way if advisory_delete was not set.

src/ripple/app/ledger/impl/LedgerMaster.cpp
mDuo13: I believe that filling in the gaps is preferable in all cases. The disk impact is probably not noticeable on well-resourced servers and may be insignificant enough to get lost in the noise of ledger size variance in general. I would vote against making this a config option.

mtrippled: Also, @mDuo13 I think it would simplify this document: https://xrpl.org/ledger-history.html
Particularly the part about "With Advisory Deletion." Going forward, that will be the normal case, not just with advisory deletion.

mtrippled:
@nbougalis @carlhua @cjcobb23 @mDuo13 @scottschurr I'm closing this PR for now. Unit tests uncovered a bug in it. I need to think it through again, and will resubmit.

PR #3299, opened 2020-03-13
Attempt to maintain history back to the earliest persisted ledger.

mtrippled:
This makes behavior consistent with configurations both with and
without online delete.


scottschurr:
I haven't really started the review, but I thought I'd mention something about the commit message while I was thinking about it.  I suggest you change the commit message to something like this:
```
Maintain history back to the earliest persisted ledger:

This makes behavior consistent with configurations both with and
without online delete.
```
Three important things about that formatting:
1. Keeping the first line under 60 characters makes the formatting better when looking at history.  History often adds levels of indentation.  Keeping that first line under 60 characters keeps the line from wrapping even with significant indentation.
2. The colon at the end of the first line tells folks that there's more to read in case they only see the first line.
3. The blank line between the first and second lines keeps git from sucking all of the lines together and undoing your careful work with the first line.

Here's a widely recognized recommendation on how to format a git commit message: https://chris.beams.io/posts/git-commit/  It wants you to keep the first line under _50_ characters.  I don't believe that's reasonable.  I struggle to keep mine at 60 or below.  But you get the idea.

Hope that helps.  Thanks.

mtrippled:
> I haven't really started the review, but I thought I'd mention something about the commit message while I was thinking about it. I suggest you change the commit message to something like this:
> 
> ```
> Maintain history back to the earliest persisted ledger:
> 
> This makes behavior consistent with configurations both with and
> without online delete.

fixed

src/ripple/app/ledger/LedgerMaster.h
```diff
@@ -283,7 +285,8 @@ class LedgerMaster
     bool shouldAcquire(
```
scottschurr: `shouldAcquire()` is only called from one place, and that is in LedgerMaster.cpp.  I think you should remove this declaration and have the implementation be a file scoped function in ledgerMaster.cpp.  More details in the LedgerMaster.cpp comments.

src/ripple/app/ledger/impl/LedgerMaster.cpp
```diff
@@ -1439,6 +1440,15 @@ LedgerMaster::newPFWork (const char *name, std::unique_lock<std::recursive_mutex
...
 }
 
+boost::optional<LedgerIndex>
+LedgerMaster::minSqlSeq() const
+{
+    boost::optional<LedgerIndex> seq;
+    *app_.getLedgerDB().checkoutDb()
+        << "SELECT MIN(LedgerSeq) FROM Ledgers", soci::into(seq);
+    return seq;
+}
```
scottschurr: See suggestion in the header file comments...

src/ripple/app/ledger/LedgerMaster.h
```diff
@@ -295,7 +298,8 @@ class LedgerMaster
     // The passed lock is a reminder to callers.
     bool newPFWork(const char *name, std::unique_lock<std::recursive_mutex>&);
 
-private:
+    boost::optional<LedgerIndex> minSqlSeq() const;
```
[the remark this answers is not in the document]
mtrippled: > Did you intend to lose the `private:` with this change? I'm guessing that was accidental...

Yes, private (and no subsequent public or protected) is already set on line 259.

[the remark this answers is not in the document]
scottschurr: Good point.  Sorry for the noise.

mtrippled:
> Generally looks like it should do what you want. I left some comments. I don't think any of them are critical. If you disagree with any of them I'm happy to talk.
> 
> It would be great if someone could write a unit test to exercise this functionality. But that may be a pretty big ask, since I can't find any unit tests that exercise online delete at all. So I'm not insisting, just leaving it for you to consider.

There's unit tests for online delete, but not for shouldAcquire(), doAccept(), and such.

src/ripple/app/ledger/LedgerMaster.h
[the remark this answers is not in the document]
mtrippled: > `minSqlSeq()` is only called from one place, and that is in LedgerMaster.cpp. I'm thinking you could turn this into a lambda in the (admittedly large) function where it is called. Possibly better would be to make `minSqlSeq()` a file scoped function in the .cpp file. It might look like this:
> 
> ```
> // Helper function for LedgerMaster::doAdvance()
> static boost::optional<LedgerIndex>
> minSqlSeq(Application& app)
> {
>     boost::optional<LedgerIndex> seq;
>     *app.getLedgerDB().checkoutDb()
>         << "SELECT MIN(LedgerSeq) FROM Ledgers", soci::into(seq);
>     return seq;
> }
> ```

fixed

src/ripple/app/ledger/LedgerMaster.h
```diff
@@ -34,11 +34,13 @@
 #include <ripple/basics/StringUtilities.h>
 #include <ripple/beast/insight/Collector.h>
 #include <ripple/beast/utility/PropertyStream.h>
+#include <ripple/core/DatabaseCon.h>
```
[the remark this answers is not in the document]
mtrippled: fixed

src/ripple/app/ledger/impl/LedgerMaster.cpp
```diff
@@ -1701,17 +1711,23 @@ bool
...
-    // is within our configured history range
-
+    // Fetch ledger if:
+    // 1) it might be the current ledger, or
+    // 2) is greater than or equal to both:
+    //     a. the minimum persisted ledger or
+    //        the maximum possible sequence value, if no persisted ledger, and
+    //     b. minimum ledger that will be persisted as of the next
+    //        online deletion interval, or 1 if online deletion is disabled, or
+    // 3) is within our configured history range.
```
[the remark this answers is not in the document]
mtrippled: fixed

src/ripple/app/ledger/impl/LedgerMaster.cpp
```diff
@@ -1701,17 +1711,23 @@ bool
 LedgerMaster::shouldAcquire (
```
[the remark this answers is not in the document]
mtrippled: fixed

src/ripple/app/ledger/impl/LedgerMaster.cpp
```diff
@@ -1701,17 +1711,23 @@ bool
...
+    //     a. the minimum persisted ledger or
+    //        the maximum possible sequence value, if no persisted ledger, and
+    //     b. minimum ledger that will be persisted as of the next
+    //        online deletion interval, or 1 if online deletion is disabled, or
+    // 3) is within our configured history range.
     bool ret (candidateLedger >= currentLedger ||
-        ((ledgerHistoryIndex > 0) &&
-            (candidateLedger > ledgerHistoryIndex)) ||
+        candidateLedger >= std::max(
+            minSeq.has_value() ?
```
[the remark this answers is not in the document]
mtrippled: fixed

src/ripple/app/misc/SHAMapStoreImp.h
```diff
@@ -159,7 +166,7 @@ class SHAMapStoreImp : public SHAMapStore
     LedgerIndex
     getLastRotated() override
     {
-        return state_db_.getState().lastRotated;
+        return lastRotated_;
```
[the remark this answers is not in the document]
mtrippled: fixed

src/ripple/app/misc/SHAMapStoreImp.cpp
```diff
@@ -356,20 +356,20 @@ SHAMapStoreImp::run()
...
-                && canDelete_ >= lastRotated - 1)
+        // will delete up to (not including) lastRotated_)
+        if (validatedSeq >= lastRotated_ + deleteInterval_
+                && canDelete_ >= lastRotated_ - 1)
         {
             JLOG(journal_.warn()) << "rotating  validatedSeq " << validatedSeq
-                    << " lastRotated " << lastRotated << " deleteInterval "
-                    << deleteInterval_ << " canDelete_ " << canDelete_;
+                                  << " lastRotated_ " << lastRotated_ << " deleteInterval "
+                                  << deleteInterval_ << " canDelete_ " << canDelete_;
```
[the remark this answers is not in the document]
mtrippled: fixed, my IDE (clion) did this in a copy & paste and I didn't look at it.

src/ripple/app/ledger/LedgerMaster.h
mtrippled: fixed

mtrippled:
> @mtrippled Thanks for the update here, I reviewed the change again and since @scottschurr already gave a thorough review, I only have one comment.
> This update although desirable IMO, it changes the default behavior of the software. Namely now if I had a really old ledger with online delete off, my software will go retrieve all the ledgers. I think one potential side effect is we might run into potential disk issues. Should we consider a cfg file update to make this behavior explict? (I do not have a strong opinion here, i think either way is fine. maybe @nbougalis and @mDuo13 can chime in here). I do think minmally, we need to update the example cfg to call out this behavior. Also, do we need XRPL.org documentation update?

Here's a link to the previous PR where a similar conversation took place: https://github.com/ripple/rippled/pull/3296

Here's the conversation we had:

carlhua 6 days ago Member
@mtrippled and I discussed this, and this comment is used to capture our discussion. Here we are changing the default behavior of rippled. Before rippled would not go out there to grab gaps and now it will. if online delete is not enabled, users can see an increase in disk usage. I suggested a cfg enhancement to make this behavior opt-in, but I also think leaving gaps not desireable. I do think this has a doc impact @mDuo13 @nbougalis

 @mDuo13
mDuo13 6 days ago Member
I believe that filling in the gaps is preferable in all cases. The disk impact is probably not noticeable on well-resourced servers and may be insignificant enough to get lost in the noise of ledger size variance in general. I would vote against making this a config option.

 @mtrippled
mtrippled 6 days ago Author Member
Also, @mDuo13 I think it would simplify this document: https://xrpl.org/ledger-history.html
Particularly the part about "With Advisory Deletion." Going forward, that will be the normal case, not just with advisory deletion.

src/ripple/app/ledger/impl/LedgerMaster.cpp
```diff
@@ -140,6 +142,50 @@ static constexpr std::chrono::minutes MAX_LEDGER_AGE_ACQUIRE {1};
...
+minSqlSeq(Application& app)
+{
+    boost::optional<LedgerIndex> seq;
+    auto db = app.getLedgerDB().checkoutDb();
+    *db << "SELECT MIN(LedgerSeq) FROM Ledgers", soci::into(seq);
+    return seq;
+}
+
+// Helper function for LedgerMaster::doAdvance()
+// Return true if a ledger should be fetched from the network.
```
scottschurr: Nice comment, but I have a small suggestion.  Consider replacing "a ledger" with "candidateLedger" in the comment.

mtrippled: @scottschurr 
Spacing fixed, comment fixed.

For the unit tests for this specific behavior, there's no coverage for several generations of callers of shouldAcquire(), and none for what I can see for make sure that we maintain history. Quite difficult to implement.

src/ripple/app/misc/SHAMapStoreImp.h
```diff
@@ -84,6 +85,13 @@ class SHAMapStoreImp : public SHAMapStore
     static std::uint32_t const minimumDeletionInterval_ = 256;
     // minimum # of ledgers required for standalone mode.
     static std::uint32_t const minimumDeletionIntervalSA_ = 8;
+    // Ledger sequence at which the last deletion interval was triggered,
+    // or the current validated sequence as of first use
+    // if there have been no prior deletions. Deletion occurs up to (but
+    // not including) this value. All ledgers past this value are accumulated
+    // until the next online deletion. This value is persisted to SQLite
+    // nearly immediately after modification.
+    std::atomic<LedgerIndex> lastRotated_ {};
```
cjcobb23: Why does this need to be atomic? Maybe add to the comment explaining briefly why it needs to be atomic

mtrippled: It's read from multiple threads, so it needs to be atomic. For synchronized things, atomics, things in mutexes, and such, I think that's generally pretty implicit.

src/ripple/app/ledger/LedgerMaster.h
```diff
@@ -39,6 +39,7 @@
 #include <ripple/protocol/Protocol.h>
 #include <ripple/protocol/RippleLedgerHash.h>
 #include <ripple/protocol/STValidation.h>
+#include <boost/optional.hpp>
```
nbougalis: I _think_ this isn't needed here and can move to the **`LedgerMaster.cpp`** file.

src/ripple/app/ledger/impl/LedgerMaster.cpp
```diff
@@ -140,6 +142,50 @@ static constexpr std::chrono::minutes MAX_LEDGER_AGE_ACQUIRE {1};
...
+{
+    boost::optional<LedgerIndex> seq;
+    auto db = app.getLedgerDB().checkoutDb();
+    *db << "SELECT MIN(LedgerSeq) FROM Ledgers", soci::into(seq);
+    return seq;
+}
+
+// Helper function for LedgerMaster::doAdvance()
+// Return true if candidateLedger should be fetched from the network.
+bool
```
nbougalis: Similarly, `static bool` here to hint to the compiler that this method is not accessed from anywhere else.

src/ripple/app/ledger/impl/LedgerMaster.cpp
```diff
@@ -140,6 +142,50 @@ static constexpr std::chrono::minutes MAX_LEDGER_AGE_ACQUIRE {1};
...
+{
+    // Fetch ledger if:
+    // 1) it might be the current ledger, or
+    // 2) is within our configured history range, or
+    // 3) is greater than or equal to both:
+    //     a. the minimum persisted ledger or
+    //        the maximum possible sequence value, if no persisted ledger, and
+    //     b. minimum ledger that will be persisted as of the next
+    //        online deletion interval, or 1 if online deletion is disabled.
+    bool ret (candidateLedger >= currentLedger ||
```
nbougalis: Can we please use sensible syntax for initializing this variable?

```
bool ret = candidateLedger >= currentLedger ||
    (currentLedger - candidateLedger) <= ledgerHistory ||
        candidateLedger >= std::max(
            minSeq.value_or(std::numeric_limits<LedgerIndex>::max()),
            lastRotated + 1));
```

Also, please consider the following:

```
bool ret = [&]()
{
    // Fetch ledger if it may be the current ledger
    if (candidateLedger >= currentLedger)
        return true;

    // Or if it is within our configured history range:
    if (currentLedger - candidateLedger <= ledgerHistory)
        return true;

    // Or it's greater than or equal to both:
    // - the minimum persisted ledger or the maximum possible
    //   sequence value, if no persisted ledger, and
    // - minimum ledger that will be persisted as of the next online
    //   deletion interval, or 1 if online deletion is disabled.
    return 
        candidateLedger >= std::max(
            minSeq.value_or(std::numeric_limits<LedgerIndex>::max()),
            lastRotated + 1);
}();
```

src/ripple/app/ledger/impl/LedgerMaster.cpp
```diff
@@ -140,6 +142,50 @@ static constexpr std::chrono::minutes MAX_LEDGER_AGE_ACQUIRE {1};
...
+    auto db = app.getLedgerDB().checkoutDb();
+    *db << "SELECT MIN(LedgerSeq) FROM Ledgers", soci::into(seq);
+    return seq;
+}
+
+// Helper function for LedgerMaster::doAdvance()
+// Return true if candidateLedger should be fetched from the network.
+bool
+shouldAcquire (
+    std::uint32_t const currentLedger,
```
nbougalis: Making "by value" function arguments `const` is a bit unusual and I don't think we do this anywhere else.

src/ripple/app/ledger/impl/LedgerMaster.cpp
```diff
@@ -140,6 +142,50 @@ static constexpr std::chrono::minutes MAX_LEDGER_AGE_ACQUIRE {1};
 // Don't acquire history if write load is too high
 static constexpr int MAX_WRITE_LOAD_ACQUIRE {8192};
 
+// Helper function for LedgerMaster::doAdvance()
+// Returns the minimum ledger sequence in SQL database, if any.
+boost::optional<LedgerIndex>
```
nbougalis:
```
static
boost::optional<LedgerIndex>
```

src/ripple/app/ledger/LedgerMaster.h
mtrippled: A bunch of function declarations return boost::optional in this file.

src/ripple/app/ledger/impl/LedgerMaster.cpp
mtrippled: This is a free function in the .cpp file, not part of the class.

src/ripple/app/ledger/impl/LedgerMaster.cpp
mtrippled: I like const on parameters because I know it's not going to change in the function itself. This function has been around forever like this. The reason it looks new is because as a result of the review process I made it a free function in the cpp file, so I copy & pasted it.

src/ripple/app/ledger/impl/LedgerMaster.cpp
mtrippled: Same, this is a free function, not a class member.

src/ripple/app/ledger/impl/LedgerMaster.cpp
mtrippled: fixed

mtrippled:
@scottschurr @nbougalis Those functions are now static.

src/ripple/app/ledger/impl/LedgerMaster.cpp
```diff
@@ -140,6 +142,57 @@ static constexpr std::chrono::minutes MAX_LEDGER_AGE_ACQUIRE {1};
...
+static bool
+shouldAcquire (
+    std::uint32_t const currentLedger,
+    std::uint32_t const ledgerHistory,
+    boost::optional<LedgerIndex> minSeq,
+    std::uint32_t const lastRotated,
+    std::uint32_t const candidateLedger,
+    beast::Journal j)
+{
+    bool ret = [&]()
```
ximinez, line 167: `bool const ret`?

PR #3302, opened 2020-03-17
Add GitHub workflow to publish docs on GitHub

thejohnfreeman:
This will update https://ripple.github.io/rippled automatically any time we push to the `develop` branch. We can always push to the `gh-pages` branch manually to override it.

You can see an example of the current output at https://thejohnfreeman.github.io/rippled.

I experimented with Read The Docs before this. RTD supports browsing multiple versions of documentation via its Sphinx theme, but the HTML output of Doxygen is completely unaware of this theme. Even if we were ok with supporting just a single version, getting Sphinx to use the raw HTML output of Doxygen requires a custom builder, but RTD only lets us choose one of the three built-in builders (`html`, `dirhtml`, and `singlehtml`).


Builds/CMake/RippledDocs.cmake
```diff
@@ -3,29 +3,70 @@
...
+  src/README.md)
+set (dependencies "${doxygen_input}" "${doxyfile}")
+
+find_path (doxygen_plantuml_jar_path plantuml.jar PATH_SUFFIXES share/plantuml)
+if (NOT doxygen_plantuml_jar_path)
+  set (doxygen_plantuml_jar_path)
+endif ()
+
+# https://en.cppreference.com/w/Cppreference:Archives
+# https://stackoverflow.com/questions/60822559/how-to-move-a-file-download-from-configure-step-to-build-step
```
cjcobb23: Nice comment!

docs/Doxyfile
```diff
@@ -104,71 +104,17 @@ WARN_LOGFILE           =
...
-    ../src/ripple/app/ledger/README.md \
-    ../src/ripple/app/paths/README.md \
-    ../src/ripple/app/tx/README.md \
-    ../src/ripple/proto/README.md \
-    ../src/ripple/shamap/README.md \
-    ../src/ripple/protocol/README.md \
-    ../src/ripple/json/README.md \
-    ../src/ripple/json/TODO.md \
-    ../src/ripple/resource/README.md \
-    ../src/ripple/rpc/README.md \
```
cjcobb23: Are these READMEs still included in the documentation? I can't seem to find them at https://thejohnfreeman.github.io/rippled/. For instance, where do I find src/ripple/rpc/README.md?

thejohnfreeman: They're all under Related Pages: https://thejohnfreeman.github.io/rippled/pages.html

To find a given document, you need to know its title. I'd like to go through and re-title them all to make them easier to find in the Related Pages list. `src/ripple/rpc/README.md` is at https://thejohnfreeman.github.io/rippled/md_ripple_rpc_README.html

docs/Doxyfile
```diff
@@ -104,71 +104,17 @@ WARN_LOGFILE           =
...
+    src/test \
+    src/README.md \
+    README.md \
+    RELEASENOTES.md \
 
 
 INPUT_ENCODING         = UTF-8
-FILE_PATTERNS          =
-RECURSIVE              = NO
+FILE_PATTERNS          = *.h *.cpp *.md
```
ximinez: Do we also want to get the `*.hpp` files, and any other source files that have unusual extensions?

docs/README.md
```diff
@@ -40,22 +34,28 @@ functionality for docs, so it's largely optional. Requirements:
    Set a system environment variable named `PLANTUML_JAR` with a value of the fullpath
```
ximinez: Should the environment variable name be updated to be `DOXYGEN_PLANTUML_JAR_PATH` to match the "Configuration" section below?

docs/Doxyfile
thejohnfreeman: We don't have any `.hpp` in our source tree. I think we should limit our extensions to these; some of the docs will like `@include` other files, and I don't want Doxygen treating them as independent documents.

docs/README.md
thejohnfreeman: Good catch, thank you. Done.

docs/Doxyfile
ximinez: I thought we did, but I must have been thinking of beast. Ok, no problem.

docs/README.md
ximinez: You're welcome!

.github/workflows/main.yml
```diff
@@ -0,0 +1,23 @@
...
+      - name: set OUTPUT_DIRECTORY
+        run: echo 'OUTPUT_DIRECTORY = html' >> docs/Doxyfile
+      - name: build
+        uses: mattnotmitt/doxygen-action@v1
+        with:
+          doxyfile-path: 'docs/Doxyfile'
+      - name: publish
+        uses: peaceiris/actions-gh-pages@v3
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
```
carlhua, line 22: What secret are we using? is it your personal token or an org token? Before this is pushed I think we should make sure we are using an org token vs your personal.

thejohnfreeman: I didn't have to set anything. I'm pretty sure this is a built-in token for every project.

.github/workflows/doxygen.yml
```diff
@@ -0,0 +1,23 @@
+name: Build and publish Doxygen documentation
+on:
+  push:
+    branches:
+      - develop
+
+jobs:
+  job:
+    runs-on: ubuntu-18.04
```
cjcobb23, line 9: Why not ubuntu-latest?

PR #3306, opened 2020-03-19
Adds new 'ledgers' RPC command - return multiple ledgers with one request

movitto:
Ledgers to be retrieved can be specified via the following params:

- ledger_hashes: array of specific hashes to retrieve
- ledger_indexes: array of specific indexes to retrieve
- ledger_index_range: range of indexes specified by first/last

This would be of use in the "ledgers" section in [xrp1ntel](https://xrp1ntel.com/ledgers) where we would use it to render statistics pertaining to a range of ledgers in a terse manner.

### DIFF
--- a/src/ripple/rpc/handlers/LedgersHandler.cpp
+++ b/src/ripple/rpc/handlers/LedgersHandler.cpp
@@ -1,0 +1,85 @@
+//------------------------------------------------------------------------------
+/*
+    This file is part of rippled: https://github.com/ripple/rippled
+    Copyright (c) 2020 Dev Null Productions, LLC
+
+    Permission to use, copy, modify, and/or distribute this software for any
+    purpose  with  or without fee is hereby granted, provided that the above
+    copyright notice and this permission notice appear in all copies.
+
+    THE  SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+    WITH  REGARD  TO  THIS  SOFTWARE  INCLUDING  ALL  IMPLIED  WARRANTIES  OF
+    MERCHANTABILITY  AND  FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+    ANY  SPECIAL ,  DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+    WHATSOEVER  RESULTING  FROM  LOSS  OF USE, DATA OR PROFITS, WHETHER IN AN
+    ACTION  OF  CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+    OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+//==============================================================================
+
+#include <ripple/rpc/handlers/LedgersHandler.h>
+#include <ripple/app/ledger/LedgerToJson.h>
+#include <ripple/app/main/Application.h>
+#include <ripple/app/misc/LoadFeeTrack.h>
+#include <ripple/json/Object.h>
+#include <ripple/protocol/ErrorCodes.h>
+#include <ripple/protocol/jss.h>
+#include <ripple/resource/Fees.h>
+#include <ripple/rpc/impl/RPCHelpers.h>
+#include <ripple/rpc/Role.h>
+
+namespace ripple {
+namespace RPC {
+
+LedgersHandler::LedgersHandler (JsonContext& context) : context_ (context)
+{
+}
+
+Status LedgersHandler::check()
+{
+    auto const& params = context_.params;
+    bool needsLedger =
+            params.isMember (jss::ledger_hashes) ||
+            params.isMember (jss::ledger_indexes) ||
+            params.isMember (jss::ledger_index_range);
+    if (! needsLedger)
+        return Status::OK;
+
+    if (auto s = lookupLedgers (ledgers_, context_))
+        return s;
+
+    bool const full = params[jss::full].asBool();
+    bool const transactions = params[jss::transactions].asBool();
+    bool const accounts = params[jss::accounts].asBool();
+    bool const expand = params[jss::expand].asBool();
+    bool const binary = params[jss::binary].asBool();
+    bool const owner_funds = params[jss::owner_funds].asBool();
+
+    options_ = (full ? LedgerFill::full : 0)
+            | (expand ? LedgerFill::expand : 0)
+            | (transactions ? LedgerFill::dumpTxrp : 0)
+            | (accounts ? LedgerFill::dumpState : 0)
+            | (binary ? LedgerFill::binary : 0)
+            | (owner_funds ? LedgerFill::ownerFunds : 0);
+
+    if (full || accounts)
+    {
+        // Until some sane way to get full ledgers has been implemented,
+        // disallow retrieving all state nodes.
+        if (! isUnlimited (context_.role))
+            return rpcNO_PERMISSION;
+
+        if (context_.app.getFeeTrack().isLoadedLocal() &&
+            ! isUnlimited (context_.role))
+        {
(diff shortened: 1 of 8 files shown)


src/ripple/rpc/impl/RPCHelpers.cpp
```diff
@@ -256,6 +256,81 @@ ledgerFromRequest(T& ledger, JsonContext& context)
 
     return Status::OK;
 }
+
+template <class T>
+Status
+ledgersFromRequest(std::vector<T>& ledgers, JsonContext& context)
```
cjcobb23: Instead of writing a new function, why can't you just call `ledgerFromRequest` in a loop? There is a lot of code duplication between `ledgersFromRequest` and `ledgerFromRequest`

src/ripple/rpc/handlers/LedgersHandler.cpp
```diff
@@ -0,0 +1,85 @@
...
+#include <ripple/rpc/Role.h>
+
+namespace ripple {
+namespace RPC {
+
+LedgersHandler::LedgersHandler (JsonContext& context) : context_ (context)
+{
+}
+
+Status LedgersHandler::check()
```
cjcobb23: A lot of code duplication between this function and `LedgerHandler::check()`. I think we should refactor to eliminate the code duplication

MarkusTeufelberger:
About a lot of `ledger` RPC traffic: In many cases the returned data is static (unless the current open ledger is being queried). This is an excellent candidate for putting a smart caching proxy in front of `rippled` and only a small fraction of requests should actually ever reach the server.

Something like the functionality in this PR could also be implemented in such a proxy (especially if there is no way to query open ledgers, only closed ones) as the backend would only need to query and combine the missing data per single ledger from `rippled` itself.

cjcobb23:
> About a lot of `ledger` RPC traffic: In many cases the returned data is static (unless the current open ledger is being queried). This is an excellent candidate for putting a smart caching proxy in front of `rippled` and only a small fraction of requests should actually ever reach the server.
> 
> Something like the functionality in this PR could also be implemented in such a proxy (especially if there is no way to query open ledgers, only closed ones) as the backend would only need to query and combine the missing data per single ledger from `rippled` itself.

@MarkusTeufelberger I agree that the data is static and a great candidate for caching. We don't have that yet though, so I oppose this functionality until we have the mentioned caching. Also, there is a limit to what we can cache; this PR allows for clients to query for large ledger ranges, which we may not be able to cache.

@movitto I also don't really understand why clients can't build on the existing `ledger` rpc, and just call that method in a loop, for each ledger desired.

MarkusTeufelberger:
The advantage would probably to use the existing cache inside rippled, so you're much closer to the data you want to serve. It creates load on a system that should be real time as much as possible though, so there is a tradeoff.

cjcobb23:
> The advantage would probably to use the existing cache inside rippled, so you're much closer to the data you want to serve. It creates load on a system that should be real time as much as possible though, so there is a tradeoff.

What existing cache are you referring to?

MarkusTeufelberger:
https://github.com/ripple/rippled/blob/90d9ca901d06b9da1653526c234bc4b00e27070b/src/ripple/core/impl/Config.cpp - NodeCache, TreeCache etc. in addition to the existing disk cache of the operating system that rippled runs on which likely contains parts of the node database too.

MarkusTeufelberger:
Agreed, still these would be closer to the database where the data comes from. I started on a reimplementation of a RPC node that would only hook into the NuDB of a running rippled instance a while back that would answer calls like this, so it would only mildly impact the server running alongside  - I gave up on that though. In theory `rippled` could do some smarter decisions regarding to caching when it knows that a client will want to request 100 consecutive ledgers at once compared to 100 potentially unrelated RPC calls, but this PR doesn't seem to contain much more than the simplest possible case.

cjcobb23:
> Agreed, still these would be closer to the database where the data comes from. I started on a reimplementation of a RPC node that would only hook into the NuDB of a running rippled instance a while back that would answer calls like this, so it would only mildly impact the server running alongside - I gave up on that though. In theory `rippled` could do some smarter decisions regarding to caching when it knows that a client will want to request 100 consecutive ledgers at once compared to 100 potentially unrelated RPC calls, but this PR doesn't seem to contain much more than the simplest possible case.

I see. Makes sense. Your comment about the RPC node that you were working on is interesting - it seems similar to a current ongoing effort to move the bulk of RPC traffic to another process. Here's the latest spec, if you want to check it out and/or leave comments : https://github.com/ripple/rippled-specs/pull/13

MarkusTeufelberger:
I would need to join Ripple to read this, sorry.

movitto:
@cjcobb23 @MarkusTeufelberger yes this is a convenience mechanism that would provide quick access to a set of ledgers w/out the overhead of multiple RPC calls (network time, disparate packet bandwidth, etc).

I understand the concern about not wanting to impose a mechanism that could put the server under undue strain but on the flip side this tradeoff may be desired in certain server roles / operating environments (read-only nodes, clusters, etc) and as @MarkusTeufelberger mentioned this could leverage optimizations to make multi-ledger lookups more efficient (this is simply a first draft to get the discussion started and explore the acceptability of the feature).

Perhaps this would be acceptable as an optional, "opt-in", feature where this RPC method would not be available via a stock deployment but only be provided if the user has a config option enabling it?

MarkusTeufelberger:
Historically, expensive RPC calls or options were just made "admin only" (e.g. requesting `full: true` on a `ledger` RPC call). This might be an option here too, so it can be used by people operating their own servers, but not abused by random users.

cjcobb23:
> @cjcobb23 @MarkusTeufelberger yes this is a convenience mechanism that would provide quick access to a set of ledgers w/out the overhead of multiple RPC calls (network time, disparate packet bandwidth, etc).
> 
> I understand the concern about not wanting to impose a mechanism that could put the server under undue strain but on the flip side this tradeoff may be desired in certain server roles / operating environments (read-only nodes, clusters, etc) and as @MarkusTeufelberger mentioned this could leverage optimizations to make multi-ledger lookups more efficient (this is simply a first draft to get the discussion started and explore the acceptability of the feature).
> 
> Perhaps this would be acceptable as an optional, "opt-in", feature where this RPC method would not be available via a stock deployment but only be provided if the user has a config option enabling it?

@movitto @MarkusTeufelberger In general, I think this functionality is better suited for a second layer (calling `ledger` in a loop client side), and does not belong in rippled. In my eyes, this PR would contribute to code bloat.
I do not agree with the cache argument that this would make multi ledger lookups more efficient; I think the cache behavior will be the same in both cases. Whether the lookup for each ledger is triggered client side (calling `ledger` in a loop) or server side (via this `ledgers` rpc call that is being proposed, which would result in a server side loop), makes no difference to the cache. The cache is going to be populated with the same entries regardless.
Also, this PR would create the possibility of a very long running job (looking up say 100 ledgers at once). This long running job is going to park itself on the JobQueue for a long period of time, which degrades the multitasking ability of rippled.
As to performance concerns, if you use a websocket connection (so the connection is persistent), and call `ledger` in a loop, the only overhead is the additional requests being sent. The total size of the response data is the same in both cases (`ledger` in a loop vs `ledgers`), and the requests themselves are very small compared to the size of the response. Furthermore, I think calling `ledger` in a loop can lead to better performance than fetching all ledgers at once, because you can leverage parallelism: fetch a ledger, and then process the received data while asynchronously fetching the next ledger. In this way, you can parallelize the client side work (processing the received data) and the server side work (fetching the ledger, serializing it, and sending over the network). Contrast this to sending a request for a bunch of ledgers, waiting for rippled to read, serialize and send all of them, and then process the data client side. The parallelized algorithm is going to lead to better performance.
For all of these reasons, I oppose this PR.

cjcobb23:
> I would need to join Ripple to read this, sorry.

@MarkusTeufelberger sorry, I forgot that repo was private. We might open it up eventually.

movitto:
@cjcobb23 ok this is fine. Closing this one.

Issue #2809, opened 2018-12-14
Create new "health check" method


mDuo13:
The `server_info` command is very cluttered and in many cases it's not easy to diagnose whether the server is healthy or not from that. A simple "health check" method could make it easier to monitor `rippled` with industry-standard tooling and also easier to diagnose manually as well.

Ideally:

- It would be an HTTP GET call, to maximize compatibility with common monitoring software. (Maybe using the same port as "crawl"?)
- It would report a health score ("Healthy", "Warning", or "Critical") depending on several factors:
    - Age of last validated ledger is < 7s (healthy), <=20s (warning) or >20s (critical)
    - Not amendment blocked (critical) and amendments unknown to the server don't currently have a majority in the network (warning)
    - Number of peers is >7 (healthy) <5 (warning unless configured w/ `peer_private`), or 0 (critical)
    - `server_state` is `full`/`validating`/`proposing` (healthy), `syncing`/`tracking`/`connected` (warning), or `disconnected` (critical)
    - Maybe a `load_factor` based warning, too? Not sure what thresholds to use there.
- If the health score is not "healthy", it reports which factor(s) are not healthy. If it's healthy, the word "HEALTHY" is the only thing in the response.

ximinez:
Speaking as an engineer, not an end-user, I think that reporting the values of the individual factors _along with_ the health score could be valuable. 
Reasons:
1. Some people just like to watch the blinkenlites, and having these top factors only may be interesting.
2. Advanced users may want to monitor _how_ healthy their node is. For example, if I'm running a hub, I would expect hundreds of peers, not just 7. It would be handy to get this information in one place so I can confirm the node thinks itself is healthy, but I can keep an extra watch on the number of peers, and have some local alert if it drops below my own threshold. Yes, an advanced user can get this info from `server_info`, but including it here, when we also return it when there are problems, saves an RPC call.

MarkusTeufelberger:
It might be useful to split this up into "liveness" and "readiness" (see https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/ for example) and I'm not convinced that a server that has constant 6 seconds lag is "healthy".

garay-daniel:
Agree with Ed's point. I think this also pushes operators to understand what metrics are important to monitor. 

If we make this compatible with crawl, then it would be easy for anyone to scrape this data to build dashboards and monitoring of the network, which would be great.

MarkusTeufelberger:
[...]

PR #3308, opened 2020-03-19
Respond 200 on empty HTTP request to support GCE ingress health check

tlongwell-ripple:
https://github.com/ripple/rippled/issues/3307

### DIFF
--- a/src/ripple/rpc/impl/ServerHandlerImp.cpp
+++ b/src/ripple/rpc/impl/ServerHandlerImp.cpp
@@ -563,6 +563,12 @@ ServerHandlerImp::processRequest (Port const& port,
 {
     auto rpcJ = app_.journal ("RPC");
 
+    if (request.size () == 0)
+    {
+        HTTPReply (200, "Okay", output, rpcJ);
+        return;
+    }
+
     Json::Value jsonOrig;
     {
         Json::Reader reader;


tlongwell-ripple:
For https://github.com/ripple/rippled/issues/3307

MarkusTeufelberger:
Why wouldn't you do your health checks on an actual health check endpoint instead?

tlongwell-ripple:
Having a public HTTP health check endpoint would be a great feature. It should use some heuristic (like info from server_info) to return a meaningful response in addition to a 200.

This 200 is just an indication that the rippled server is running and accepting connections on a particular port. It is not meant to denote health wrt network synchronization/load/etc.

This was a suggestion from @nbougalis as a one-off build for a particular use case I have. My biggest motivation for opening the PR is really to have Travis run and see if this unexpectedly breaks anything.

MarkusTeufelberger:
A non 200 response would also indicate that it is running and accepting connections.

tlongwell-ripple:
Responding with a 200 allows this dumb endpoint to be used with a GKE ingress https://cloud.google.com/kubernetes-engine/docs/concepts/ingress#health_checks

The ingress allows changing the path of the health check, but not the accepted response code. The response code must be exactly 200.

MarkusTeufelberger:
You can set a path for readiness probes there, the /crawl endpoint for example would be a good workaround for this patch.

MarkusTeufelberger:
See https://github.com/ripple/rippled/issues/2809 for some discussion on this topic by the way.

intelliot:
Markus, that's a good idea. Docs are here: https://xrpl.org/peer-crawler.html

Example request: `https://s1.ripple.com:51235/crawl`

carlhua:
@MarkusTeufelberger  @intelliot  I really appreciate the feedback. I will prioritize #2809 to be discussed in the near future.  This fix is simply to provide a 200 response for the load balancer. Lets review this PR while #2809 undergoes discussion.

intelliot:
If this is for a load balancer, I think it might make sense to return 200 only if the server is synced with the network.

Sidenote: the `/crawl` endpoint does provide a 200 response:
```
% curl -I https://s1.ripple.com:51235/crawl
HTTP/1.1 200 OK
Server: rippled-1.4.0
Content-Type: application/json
Connection: close
Transfer-Encoding: chunked
```

HowardHinnant:
Removing myself as reviewer.  The C++ looks fine.  But I'm not qualified to say if this is something that should be done for rippled.

carlhua:
@nbougalis do we want this as part of 1.6?

nbougalis:
I think we should fold the idea behind this into #3365 and consider following the guidelines provided by @MarkusTeufelberger in that thread. I'd suggest that we just close this.

PR #3310, opened 2020-03-20
Housecleaning: small fixes here and there

thejohnfreeman:
A collection of small changes I've made while studying the code. I wanted to pull them out of hiding on my machine. It might be easier to review the commits separately instead of all jumbled together. There is no urgency to this PR. I chose reviewers I thought might be interested, but am not expecting anyone in particular to review this.

### DIFF
--- a/src/ripple/app/consensus/RCLConsensus.cpp
+++ b/src/ripple/app/consensus/RCLConsensus.cpp
@@ -113,7 +113,7 @@ RCLConsensus::Adaptor::acquireLedger(LedgerHash const& hash)
         return boost::none;
     }
 
-    assert(!built->open() && built->isImmutable());
+    assert(!built->open() && !built->isMutable());
     assert(built->info().hash == hash);
 
     // Notify inbound transactions of the new ledger sequence number
--- a/src/ripple/app/ledger/Ledger.cpp
+++ b/src/ripple/app/ledger/Ledger.cpp
@@ -75,8 +75,7 @@ calculateLedgerHash (LedgerInfo const& info)
 
 //------------------------------------------------------------------------------
 
-class Ledger::sles_iter_impl
-    : public sles_type::iter_base
+class sles_iter_impl : public Ledger::sles_type::iter_base
 {
 private:
     SHAMap::const_iterator iter_;
@@ -113,7 +112,7 @@ public:
         ++iter_;
     }
 
-    sles_type::value_type
+    Ledger::sles_type::value_type
     dereference() const override
     {
         auto const item = *iter_;
@@ -125,8 +124,7 @@ public:
 
 //------------------------------------------------------------------------------
 
-class Ledger::txs_iter_impl
-    : public txs_type::iter_base
+class txs_iter_impl : public Ledger::txs_type::iter_base
 {
 private:
     bool metadata_;
@@ -165,7 +163,7 @@ public:
         ++iter_;
     }
 
-    txs_type::value_type
+    Ledger::txs_type::value_type
     dereference() const override
     {
         auto const item = *iter_;
--- a/src/ripple/app/ledger/Ledger.cpp
+++ b/src/ripple/app/ledger/Ledger.cpp
@@ -175,16 +175,14 @@ public:
 
 //------------------------------------------------------------------------------
 
-Ledger::Ledger (
-        create_genesis_t,
-        Config const& config,
-        std::vector<uint256> const& amendments,
-        Family& family)
-    : mImmutable (false)
-    , txMap_ (std::make_shared <SHAMap> (SHAMapType::TRANSACTION,
-        family))
-    , stateMap_ (std::make_shared <SHAMap> (SHAMapType::STATE,
-        family))
+Ledger::Ledger(
+    create_genesis_t,
+    Config const& config,
+    std::vector<uint256> const& amendments,
+    Family& family)
+    : mMutable(true)
+    , txMap_(std::make_shared<SHAMap>(SHAMapType::TRANSACTION, family))
+    , stateMap_(std::make_shared<SHAMap>(SHAMapType::STATE, family))
     , rules_{config.features}
 {
     info_.seq = 1;
@@ -213,20 +211,22 @@ Ledger::Ledger (
     setImmutable(config);
 }
 
-Ledger::Ledger (
-        LedgerInfo const& info,
-        bool& loaded,
-        bool acquire,
-        Config const& config,
-        Family& family,
-        beast::Journal j)
-    : mImmutable (true)
-    , txMap_ (std::make_shared <SHAMap> (SHAMapType::TRANSACTION,
-        info.txHash, family))
-    , stateMap_ (std::make_shared <SHAMap> (SHAMapType::STATE,
-        info.accountHash, family))
-    , rules_ (config.features)
-    , info_ (info)
+Ledger::Ledger(
+    LedgerInfo const& info,
+    bool& loaded,
+    bool acquire,
+    Config const& config,
+    Family& family,
+    beast::Journal j)
+    : mMutable(false)
+    , txMap_(std::make_shared<SHAMap>(
+          SHAMapType::TRANSACTION,
+          info.txHash,
+          family))
+    , stateMap_(
+          std::make_shared<SHAMap>(SHAMapType::STATE, info.accountHash, family))
+    , rules_(config.features)
+    , info_(info)
 {
     loaded = true;
 
@@ -259,13 +259,12 @@ Ledger::Ledger (
 }
 
 // Create a new ledger that follows this one
-Ledger::Ledger (Ledger const& prevLedger,
-    NetClock::time_point closeTime)
-    : mImmutable (false)
-    , txMap_ (std::make_shared <SHAMap> (
-        SHAMapType::TRANSACTION,
-        prevLedger.stateMap_->family()))
-    , stateMap_ (prevLedger.stateMap_->snapShot (true))
+Ledger::Ledger(Ledger const& prevLedger, NetClock::time_point closeTime)
+    : mMutable(true)
+    , txMap_(std::make_shared<SHAMap>(
+          SHAMapType::TRANSACTION,
+          prevLedger.stateMap_->family()))
+    , stateMap_(prevLedger.stateMap_->snapShot(true))
     , fees_(prevLedger.fees_)
     , rules_(prevLedger.rules_)
 {
@@ -291,29 +290,28 @@ Ledger::Ledger (Ledger const& prevLedger,
     }
 }
 
-Ledger::Ledger (
-        LedgerInfo const& info,
-        Config const& config,
-        Family& family)
-    : mImmutable (true)
-    , txMap_ (std::make_shared <SHAMap> (SHAMapType::TRANSACTION,
-        info.txHash, family))
-    , stateMap_ (std::make_shared <SHAMap> (SHAMapType::STATE,
-        info.accountHash, family))
+Ledger::Ledger(LedgerInfo const& info, Config const& config, Family& family)
+    : mMutable(false)
+    , txMap_(std::make_shared<SHAMap>(
+          SHAMapType::TRANSACTION,
+          info.txHash,
+          family))
+    , stateMap_(
+          std::make_shared<SHAMap>(SHAMapType::STATE, info.accountHash, family))
     , rules_{config.features}
-    , info_ (info)
+    , info_(info)
 {
     info_.hash = calculateLedgerHash (info_);
 }
 
-Ledger::Ledger (std::uint32_t ledgerSeq,
-        NetClock::time_point closeTime, Config const& config,
-            Family& family)
-    : mImmutable (false)
-    , txMap_ (std::make_shared <SHAMap> (
-          SHAMapType::TRANSACTION, family))
-    , stateMap_ (std::make_shared <SHAMap> (
-          SHAMapType::STATE, family))
+Ledger::Ledger(
+    std::uint32_t ledgerSeq,
+    NetClock::time_point closeTime,
+    Config const& config,
+    Family& family)
+    : mMutable(true)
+    , txMap_(std::make_shared<SHAMap>(SHAMapType::TRANSACTION, family))
+    , stateMap_(std::make_shared<SHAMap>(SHAMapType::STATE, family))
     , rules_{config.features}
 {
     info_.seq = ledgerSeq;
@@ -326,7 +324,7 @@ void Ledger::setImmutable (Config const& config)
 {
     // Force update, since this is the only
     // place the hash transitions to valid
-    if (! mImmutable)
+    if (mMutable)
     {
         info_.txHash = txMap_->getHash ().as_uint256();
         info_.accountHash = stateMap_->getHash ().as_uint256();
@@ -334,7 +332,7 @@ void Ledger::setImmutable (Config const& config)
 
     info_.hash = calculateLedgerHash (info_);
 
-    mImmutable = true;
+    mMutable = false;
     txMap_->setImmutable ();
     stateMap_->setImmutable ();
     setup(config);
@@ -1001,7 +999,7 @@ bool pendSaveValidated (
         }
     }
 
-    assert (ledger->isImmutable ());
+    assert(!ledger->isMutable());
 
     if (! app.pendingSaves().shouldWork (ledger->info().seq, isSynchronous))
     {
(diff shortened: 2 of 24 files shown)


src/ripple/core/impl/Stoppable.cpp
```diff
@@ -85,11 +85,11 @@ void Stoppable::onStart ()
 
 void Stoppable::onStop ()
 {
-    stopped();
```
seelabs: I'd expect `stopped` to be called from `onStop` and not from `onChildrenStopped`. If a class derives from Stoppable, then it overrides `onChildrenStopped` but not `onStopped` it has to know to called `stopped()`? That doesn't seem intuitive.

I think the motivation for this change is to simplify the implementation of the `JobQueue`, is that right? I think I'd prefer to keep `Stoppable::onStop` as it is, and maybe add a comment in `onStop` to say why the base class does the wrong thing for this case.

If we do decide to keep this change, we need to update the docs on `onChildrenStopped` and `onStopped` to reflect the new behavior (the current docs says `onStopped` calls `stopped` and `onChildrenStopped` does nothing)

src/ripple/app/ledger/Ledger.h
```diff
@@ -255,6 +255,12 @@ class Ledger final
         return mImmutable;
     }
 
+    bool
+    isMutable() const
```
seelabs: I understand the motivation: `if (!isImmutable())` is a double negative and reads weird. I'd be OK if `isImmutable` was replaced with `!isMutable` and we removed that function, but I'm don't love an interface that has both `isImmutable` and `isMutable`.

src/ripple/core/impl/Stoppable.cpp
thejohnfreeman: > If a class derives from `Stoppable`, and it overrides `onChildrenStopped` but not `onStop`, then it has to know to called `stopped()`? That doesn't seem intuitive.

I think this is just a matter of naming. We can change the names, and I'd be happy to. It's a harder sell to change public API method names, so that's why I left them alone. I think better names would be `beforeChildrenStopped` and `afterChildrenStopped`. The documentation suggests that a parent can expect to no longer be called by its children after they have stopped, which suggests to me that that is the only safe time for a parent to stop, which means that's when it should call `stopped()`.

> I think the motivation for this change is to simplify the implementation of the `JobQueue`, is that right?

The primary motivation is to simplify the uses of the API, which includes `JobQueue`. I think the default implementation is wrong, considering the semantics of the class:

- Children depend on their parents.
- Dependents should be stopped before dependencies are stopped.
- `onChildrenStopped` is called after all of a `Stoppable`'s dependents have stopped.
- Thus, `onChildrenStopped` is when `stopped()` should be called.

> If we do decide to keep this change, we need to update the docs

Indeed, that was my oversight. Thanks for reminding!

seelabs: I have to think about this one. Unlike the other changes in this PR, this changes in a critical API.

scottschurr: Yeah, I suspect there's something thread-related wrong with this pull request.  I fired off a start-sync-for-60-seconds-stop loop on this branch.  The 48th start-stop died with a crash.  The actual crash was here:
```
11  rippled                       	0x000000010ac2911b ripple::OverlayImpl::collect_metrics() + 139 (vector:657)
12  rippled                       	0x000000010b06b31a beast::insight::detail::StatsDCollectorImp::on_timer(boost::system::error_code) + 234
```
but it is also interesting to note that the `ApplicationImp` destructor was running at the time:
```
Thread 0:: rippled: main  Dispatch queue: com.apple.main-thread
0   libsystem_kernel.dylib        	0x00007fff63f41f06 __psynch_mutexwait + 10
1   libsystem_pthread.dylib       	0x00007fff63ffed52 _pthread_mutex_firstfit_lock_wait + 96
2   libsystem_pthread.dylib       	0x00007fff63ffc4cd _pthread_mutex_firstfit_lock_slow + 222
3   libc++.1.dylib                	0x00007fff6103c0bf std::__1::recursive_mutex::lock() + 9
4   rippled                       	0x000000010b05cd6b beast::insight::detail::StatsDHookImpl::~StatsDHookImpl() + 59 (List.h:151)
5   rippled                       	0x000000010ac2950e ripple::OverlayImpl::~OverlayImpl() + 190 (memory:3587)
6   rippled                       	0x000000010ac29c2e ripple::OverlayImpl::~OverlayImpl() + 14 (OverlayImpl.cpp:183)
7   rippled                       	0x000000010af66525 ripple::ApplicationImp::~ApplicationImp() + 325 (memory:2649)
8   rippled                       	0x000000010af5dafe ripple::ApplicationImp::~ApplicationImp() + 14 (Application.cpp:250)
```
If it would be useful I can share the scripts I use to automate that test loop.  I run them on a Mac, so you'd have to tweak them...

nbougalis: @scottschurr can you try to the same test against the version on top of which this PR is built on.

src/ripple/app/ledger/impl/LedgerMaster.cpp
```diff
@@ -1181,12 +1180,7 @@ LedgerMaster::findNewLedgersToPublish(std::unique_lock<std::recursive_mutex>& sl
             "Gap in validated ledger stream " << mPubLedgerSeq <<
             " - " << mValidLedgerSeq - 1;
 
-        auto valLedger = mValidLedger.get ();
-        ret.push_back (valLedger);
-        setPubLedger (valLedger);
-        app_.getOrderBookDB().setup(valLedger);
```
nbougalis: Why are we just removing this?

src/ripple/app/ledger/impl/LedgerMaster.cpp
```diff
@@ -1211,24 +1211,21 @@ LedgerMaster::findNewLedgersToPublish(std::unique_lock<std::recursive_mutex>& sl
...
-            if (! hash)
-                hash = beast::zero; // kludge
-            if (seq == valSeq)
-            {
-                // We need to publish the ledger we just fully validated
-                ledger = valLedger;
-            }
-            else if (hash->isZero())
+            // VFALCO TODO Restructure this code so that zero is not used.
+            if (!hash || hash->isZero())
```
nbougalis: I think that we can probably just fix this TODO for Vinnie.

src/ripple/app/ledger/impl/LedgerMaster.cpp
seelabs: @nbougalis This work is now done in `setFullLedger`. The conditional in that function was changed to handle this. The only way I could see this being an issue is the refactored code calls these functions later than the original code, but as far as I can tell the refactored code is correct.

thejohnfreeman: The only place this function, `findNewLedgersToPublish`, is called is [here](https://github.com/ripple/rippled/blob/2341710e91f2626001e5cd3abfda4d25e82e4ad1/src/ripple/app/ledger/impl/LedgerMaster.cpp#L1849) (all links are to the same commit as this PR) in `LedgerMaster::doAdvance()`. This line is returning a non-empty `pubLedgers` to `doAdvance()`, so it takes the else branch ([here](https://github.com/ripple/rippled/blob/2341710e91f2626001e5cd3abfda4d25e82e4ad1/src/ripple/app/ledger/impl/LedgerMaster.cpp#L1909)) of the if-statement [here](https://github.com/ripple/rippled/blob/2341710e91f2626001e5cd3abfda4d25e82e4ad1/src/ripple/app/ledger/impl/LedgerMaster.cpp#L1850).

In that else branch, we unconditionally call `setPubLedger(ledger)` ([here](https://github.com/ripple/rippled/blob/2341710e91f2626001e5cd3abfda4d25e82e4ad1/src/ripple/app/ledger/impl/LedgerMaster.cpp#L1923)) for the ledger we returned, which is the same as the call that used to be on line 1186.

For the call to `OrderBookDB::setup()` that was on line 1187, we must look into the call of `setFullLedger(ledger)` ([here](https://github.com/ripple/rippled/blob/2341710e91f2626001e5cd3abfda4d25e82e4ad1/src/ripple/app/ledger/impl/LedgerMaster.cpp#L1920)), which makes the same call [here](https://github.com/ripple/rippled/blob/2341710e91f2626001e5cd3abfda4d25e82e4ad1/src/ripple/app/ledger/impl/LedgerMaster.cpp#L856) (after no early returns).

src/ripple/app/ledger/impl/LedgerMaster.cpp
thejohnfreeman: Are you suggesting to just remove it? I don't know enough to confirm that `hashOfSeq` never returns zero.

src/ripple/core/impl/Stoppable.cpp
scottschurr: @nbougalis, yes I'll try the same against -rc3.   FWIW a few years back, when I was working on thread safety, I was able to get well over 1000 start-sync-stop loops with no crashes.  In my opinion if we're not able to attain that then there's some work to do.

scottschurr: @nbougalis, I ran the same test against -rc3 and it showed approximately the same level of stability as this branch.  This time there was a crash on the 38th iteration.  Again the crash was in `OverlayImpl::collect_metrics()`:

```
9   libc++.1.dylib                	0x00007fff789b50e7 std::__1::__throw_system_error(int, char const*) + 77
10  libc++.1.dylib                	0x00007fff7896f96d std::__1::mutex::lock() + 29
11  rippled                       	0x00000001083366bb ripple::OverlayImpl::collect_metrics() + 139 (vector:657)
12  rippled                       	0x0000000108778ada beast::insight::detail::StatsDCollectorImp::on_timer(boost::system::error_code) + 234
13  rippled                       	0x0000000108778e53 boost::asio::detail::wait_handler<std::__1::__bind<void (beast::insight::detail::StatsDCollectorImp::*)(boost::system::error_code), beast::insight::detail::StatsDCollectorImp*, std::__1::placeholders::__ph<1> const&>, boost::asio::detail::io_object_executor<boost::asio::executor> >::do_complete(void*, boost::asio::detail::scheduler_operation*, boost::system::error_code const&, unsigned long) + 227 (type_traits:4302)
14  rippled                       	0x00000001074132c8 boost::asio::detail::scheduler::do_run_one(boost::asio::detail::conditionally_enabled_mutex::scoped_lock&, boost::asio::detail::scheduler_thread_info&, boost::system::error_code const&) + 920
```
It looks like the `ApplicationImp` destructor is running in another thread and it's trying to destroy `StatsDHookImpl`:
```
3   libc++.1.dylib                	0x00007fff789740bf std::__1::recursive_mutex::lock() + 9
4   rippled                       	0x000000010876a52b beast::insight::detail::StatsDHookImpl::~StatsDHookImpl() + 59 (List.h:151)
5   rippled                       	0x0000000108336aae ripple::OverlayImpl::~OverlayImpl() + 190 (memory:3587)
6   rippled                       	0x00000001083371ce ripple::OverlayImpl::~OverlayImpl() + 14 (OverlayImpl.cpp:183)
7   rippled                       	0x0000000108673d45 ripple::ApplicationImp::~ApplicationImp() + 325 (memory:2649)
8   rippled                       	0x000000010866b31e ripple::ApplicationImp::~ApplicationImp() + 14 (Application.cpp:250)
9   rippled                       	0x00000001086696bd ripple::run(int, char**) + 16781 (memory:2649)
```
My take is that we shouldn't be making any changes to `Stoppable` until the stability is fixed.  Or maybe another way to put it is that any changes to `Stoppable` should be specifically associated with fixing this stability issue.  And the stability should be significantly improved by any changes to `Stoppable`.

The stability problem may not be in `Stoppable`, but until the stability is fixed we have no way of validating any threading-related changes.

thejohnfreeman: I'm digging deeper into `Stoppable` to analyze the start and stop order. I'll see if I can find the cause of this.

src/ripple/app/ledger/impl/LedgerMaster.cpp
nbougalis: :+1: thanks both

src/ripple/app/ledger/impl/LedgerMaster.cpp
nbougalis: I _think_ we've already fixed that but on second thought, I'm not 100% sure, so let's not do that right now.

cjcobb23:
Just sneaking in a comment here that is only mildly related to the PR: The `GRPCServer` class is not part of the stoppable hierarchy. It's constructed as part of `ApplicationImpl` construction, started in `ApplicationImpl::setup()` (immediately after the `JobQueue`'s thread count is set), and stopped+destroyed as part of `ApplicationImpl` destruction. I am not sure if this is okay or not. @thejohnfreeman based on your observations of `Stoppable`, do you think the `GRPCServer` should be part of the `Stoppable` hierarchy?

thejohnfreeman:
@cjcobb23 Does the `GRPCServer` access any other `Stoppable` component in an asynchronous operation? If not, then I don't think it needs to be part of the hierarchy.

cjcobb23:
> @cjcobb23 Does the `GRPCServer` access any other `Stoppable` component in an asynchronous operation? If not, then I don't think it needs to be part of the hierarchy.

@thejohnfreeman It posts coroutines to the JobQueue (via `postCoro()`)

thejohnfreeman:
Then I suspect it needs to be `Stoppable`, but I'm not sure how it needs to be implemented.

cjcobb23:
> Then I suspect it needs to be `Stoppable`, but I'm not sure how it needs to be implemented.

I'll create an issue for this

src/ripple/app/consensus/RCLConsensus.cpp
```diff
@@ -113,7 +113,7 @@ RCLConsensus::Adaptor::acquireLedger(LedgerHash const& hash)
         return boost::none;
     }
 
-    assert(!built->open() && built->isImmutable());
+    assert(!built->open() && !built->isMutable());
```
nbougalis, line 116: I love this change. Thanks!

src/ripple/app/ledger/Ledger.cpp
```diff
@@ -261,13 +259,12 @@ Ledger::Ledger (
 }
 
 // Create a new ledger that follows this one
```
nbougalis, line 261: Apropos of our conversation about terminology and vocabulary, this comment is slightly confusing in multiple ways. First, it probably belongs in the header and it should read: "create a successor ledger to the one provided".

src/ripple/app/ledger/Ledger.cpp
```diff
@@ -293,29 +290,28 @@ Ledger::Ledger (Ledger const& prevLedger,
...
-Ledger::Ledger (std::uint32_t ledgerSeq,
-        NetClock::time_point closeTime, Config const& config,
-            Family& family)
-    : mImmutable (false)
-    , txMap_ (std::make_shared <SHAMap> (
-          SHAMapType::TRANSACTION, family))
-    , stateMap_ (std::make_shared <SHAMap> (
-          SHAMapType::STATE, family))
+Ledger::Ledger(
+    std::uint32_t ledgerSeq,
```
nbougalis, line 308: Not actionable here, but just a comment for you to think about: if possible, I'd love to see a `clang-format` configuration that either indents the arguments or the constructor initializers one extra level, to visually distinguish argument lists from constructor initializers.

src/ripple/core/Stoppable.h
```diff
@@ -97,9 +97,10 @@ class RootStoppable;
 
     7.  stop()
 
-        This first calls stopAsync(), and then blocks on each child Stoppable in
-        the in the tree from the bottom up, until the Stoppable indicates it has
-        stopped. This will usually be called from the main thread of execution
+        This first calls stopAsync(), then blocks on each child Stoppable in
```
nbougalis, line 100: I wonder if it would be better if we extracted this massive comment into `Stoppable.md` under `src/ripple/core` and simply reference it from here.

src/ripple/core/impl/Stoppable.cpp
nbougalis: So, I spent some time looking at this. We _recently_ changed this code (https://github.com/ripple/rippled/commit/b784988caf757d919abe6ddc30fe62ee28a9d779) and I suspect that the problem is that the collector keeps trying to collect metrics.

Tagging @mayurbhandary for visibility.

thejohnfreeman:
I'm going to close this review for now. It has some quality-of-life changes I would still like to see merged, but, while I ended up learning and writing much about the `Stoppable` hierarchy, I did not replicate or unwind the crash that Scott discovered. I may circle back to it at a later date, but it's not on the roadmap right now.

PR #3316, opened 2020-03-25
[TRIVIAL] Fix unneeded copies in range some range for loops:

seelabs:
clang 10 warns about an unneeded copy for these range for
loops (range-loop-construct warnings)

### DIFF
--- a/src/test/app/PayChan_test.cpp
+++ b/src/test/app/PayChan_test.cpp
@@ -946,7 +946,7 @@ struct PayChan_test : public beast::unit_test::suite
 
         Env env(*this);
         env.fund(XRP(10000), alice);
-        for (auto const a : bobs)
+        for (auto const& a : bobs)
         {
             env.fund(XRP(10000), a);
             env.close();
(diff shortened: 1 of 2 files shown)


src/test/app/PayChan_test.cpp
```diff
@@ -946,7 +946,7 @@ struct PayChan_test : public beast::unit_test::suite
 
         Env env(*this);
         env.fund(XRP(10000), alice);
-        for (auto const a : bobs)
+        for (auto const& a : bobs)
```
cjcobb23: Nit: Can you change `a` to `b` here? Same for line 993. Since the collection is called `bobs`, I think `b` is a better name. Also, on line 959, we iterate through `bobs` and name the element `b`.

seelabs: `a` for "account", but I don't mind changing it to `b`. Although that may take this out of the running for shortest patch - four characters!

cjcobb23: Ha! It's your call then. I don't mind either way.

PR #3320, opened 2020-03-26
More housecleaning in PeerSet type hierarchy

thejohnfreeman:
Lots of commits more easily reviewed separately.

You can see the PlantUML sequence diagram rendered [here](http://www.plantuml.com/plantuml/uml/lPHDZzem48Rl-HKM9oDj8oL7ggqMTQKkLUc7uDQJWwV19R63FQQjn8-l7TF8u4ISkZaGOJ_pyRuP-yaWqsZR9glKRpBRW3w19iKi8-xft9ROYOv992XQq0LXXimCyiMY_vzYFu-1-QYgRnOif8Ie0h6IbRAIRpMJXkrGA6bI65c08sgvErgX8MKx4TYl07hLThCC92TELNObhHmBBAy3Kr7bNL7bbAYe3sPH-TKJGo1APj_VlEljM90QS0D8jLBumin-tXVmMuv0vedKWHjQ0rejqq0X1L9fMw_EF9OU3W-oY6ASzm1jH2imnX9KIKykEGTEEm0Tq4aSjQXhq0cQu9tmk079wNRz0V0Jy7eg918RFnm6UrtIxDvMkZogm01k_RJJSLFybiT-CO6KplrQ-h7oenySDq1qFcH9bAJzH4Hu4MWx_4CsByirwx8ip_Ei4nVxolxZhC4OLaFYeGSSvuSvV6_VJ6ZtClXv_I-0jY79_6gqejvShb3SPnkAf_5nOL6LKvD-Ff-ZGXTrSQn_3RCHQZmpP27wfg-nDeSqHUnHciiJmQJYrgIHevtN2dqetJib3-I-QKNencSVZlvuZlGS8pl0PyN1KEb-Ly_cWLIDganSZLW3Fsn3paJ84si4ZykBwbA-dzWeyp2Q8HMon8BfiNPrr06-o7109gKZ6wH7XM-2i_IgYRRkNrh5QrrQKY4GH-nnGThtgkqQG22FBih2s_zmAILBYRLF8Bbjcxy0).

### DIFF
--- a/src/ripple/app/ledger/InboundLedger.h
+++ b/src/ripple/app/ledger/InboundLedger.h
@@ -31,9 +31,9 @@
 namespace ripple {
 
 // A ledger we are trying to acquire
-class InboundLedger : public PeerSet,
-                      public std::enable_shared_from_this<InboundLedger>,
-                      public CountedObject<InboundLedger>
+class InboundLedger final : public PeerSet,
+                            public std::enable_shared_from_this<InboundLedger>,
+                            public CountedObject<InboundLedger>
 {
 public:
     using clock_type = beast::abstract_clock<std::chrono::steady_clock>;
--- a/src/ripple/app/ledger/InboundLedger.h
+++ b/src/ripple/app/ledger/InboundLedger.h
@@ -66,7 +66,7 @@ public:
 
     // Called when the PeerSet timer expires
     void
-    execute() override;
+    queueJob() override;
 
     // Called when another attempt is made to fetch this same ledger
     void
@@ -157,7 +157,7 @@ private:
     onTimer(bool progress, ScopedLockType& peerSetLock) override;
 
     void
-    newPeer(std::shared_ptr<Peer> const& peer) override
+    onPeerAdded(std::shared_ptr<Peer> const& peer) override
     {
         // For historical nodes, do not trigger too soon
         // since a fetch pack is probably coming
--- a/src/ripple/app/ledger/InboundLedger.h
+++ b/src/ripple/app/ledger/InboundLedger.h
@@ -72,6 +72,27 @@ public:
     void
     update(std::uint32_t seq);
 
+    /** Returns true if we got all the data. */
+    bool
+    isComplete() const
+    {
+        return mComplete;
+    }
+
+    /** Returns false if we failed to get the data. */
+    bool
+    isFailed() const
+    {
+        return mFailed;
+    }
+
+    /** Returns the number of times we timed out. */
+    int
+    getTimeouts() const
+    {
+        return mTimeouts;
+    }
+
     std::shared_ptr<Ledger const>
     getLedger() const
     {
--- a/src/ripple/app/ledger/InboundLedger.h
+++ b/src/ripple/app/ledger/InboundLedger.h
@@ -118,9 +118,6 @@ public:
     void
     runData();
 
-    static LedgerInfo
-    deserializeHeader(Slice data, bool hasPrefix);
-
     void
     touch()
     {
@@ -225,6 +222,14 @@ private:
     std::vector<PeerDataPairType> mReceivedData;
     bool mReceiveDispatched;
 };
+
+/** Deserialize a ledger header from a byte array. */
+LedgerInfo
+deserializeHeader(Slice data);
+
+/** Deserialize a ledger header (prefixed with 4 bytes) from a byte array. */
+LedgerInfo
+deserializePrefixedHeader(Slice data);
 
 }  // namespace ripple
 
--- a/src/ripple/app/ledger/InboundLedger.h
+++ b/src/ripple/app/ledger/InboundLedger.h
@@ -86,13 +86,6 @@ public:
         return mFailed;
     }
 
-    /** Returns the number of times we timed out. */
-    int
-    getTimeouts() const
-    {
-        return mTimeouts;
-    }
-
     std::shared_ptr<Ledger const>
     getLedger() const
     {
@@ -103,12 +96,6 @@ public:
     getSeq() const
     {
         return mSeq;
-    }
-
-    Reason
-    getReason() const
-    {
-        return mReason;
     }
 
     bool
@@ -144,12 +131,6 @@ public:
     getLastAction() const
     {
         return mLastAction;
-    }
-
-    bool
-    isActive() const
-    {
-        return !isDone();
     }
 
 private:
--- a/src/ripple/app/ledger/InboundLedger.h
+++ b/src/ripple/app/ledger/InboundLedger.h
@@ -36,6 +36,8 @@ class InboundLedger : public PeerSet,
                       public CountedObject<InboundLedger>
 {
 public:
+    using clock_type = beast::abstract_clock<std::chrono::steady_clock>;
+
     static char const*
     getCountedObjectName()
     {
@@ -111,6 +113,24 @@ public:
     static LedgerInfo
     deserializeHeader(Slice data, bool hasPrefix);
 
+    void
+    touch()
+    {
+        mLastAction = m_clock.now();
+    }
+
+    clock_type::time_point
+    getLastAction() const
+    {
+        return mLastAction;
+    }
+
+    bool
+    isActive() const
+    {
+        return !isDone();
+    }
+
 private:
     enum class TriggerReason { added, reply, timeout };
 
@@ -145,6 +165,9 @@ private:
             trigger(peer, TriggerReason::added);
     }
 
+    std::size_t
+    getPeerCount() const;
+
     std::weak_ptr<PeerSet>
     pmDowncast() override;
 
@@ -178,6 +201,9 @@ private:
 
     std::vector<uint256>
     neededStateHashes(int max, SHAMapSyncFilter* filter) const;
+
+    clock_type& m_clock;
+    clock_type::time_point mLastAction;
 
     std::shared_ptr<Ledger> mLedger;
     bool mHaveHeader;
--- a/src/ripple/app/ledger/InboundLedger.h
+++ b/src/ripple/app/ledger/InboundLedger.h
@@ -64,10 +64,6 @@ public:
 
     ~InboundLedger();
 
-    // Called when the PeerSet timer expires
-    void
-    queueJob() override;
-
     // Called when another attempt is made to fetch this same ledger
     void
     update(std::uint32_t seq);
@@ -154,6 +150,9 @@ private:
 
     void
     onTimer(bool progress, ScopedLockType& peerSetLock) override;
+
+    void
+    queueJob() override;
 
     void
     onPeerAdded(std::shared_ptr<Peer> const& peer) override
--- a/src/ripple/app/ledger/InboundTransactions.uml
+++ b/src/ripple/app/ledger/InboundTransactions.uml
@@ -1,0 +1,81 @@
+@startuml
+box "Server 1"
+    participant timer1 as "std::timer"
+    participant jq1 as "JobQueue"
+    participant ib1 as "InboundTransactions"
+    participant ta1 as "TransactionAcquire"
+    participant pi1 as "PeerImp"
+end box
+
+box "Server 2"
+    participant pi2 as "PeerImp"
+    participant jq2 as "JobQueue"
+    participant ib2 as "InboundTransactions"
+end box
+
+autoactivate on
+[-> ib1 : getSet(rootHash)
+    ib1 -> ta1 : TransactionAcquire(rootHash)
+    return
+    ib1 -> ta1 : init(numPeers=2)
+        ta1 -> ta1 : addPeers(limit=2)
+            ta1 -> ta1 : onPeerAdded(peer)
+                ta1 -> ta1 : trigger(peer)
+                    ta1 -> pi1 : send(TMGetLedger)
+                    return
+                deactivate ta1
+            deactivate ta1
+        deactivate ta1
+        ta1 -> ta1 : setTimer()
+            ta1 -> timer1 : wait(() -> TransactionAcquire.queueJob())
+            return
+        deactivate ta1
+    return
+return empty SHAMap
+
+...
+pi1 -> pi2 : onMessage(TMGetLedger)
+    pi2 -> jq2 : addJob(() -> PeerImp.getLedger(message))
+    return
+deactivate pi2
+
+...
+jq2 -> pi2 : getLedger(message)
+    pi2 -> ib2 : getSet(rootHash)
+    |||
+    return SHAMap
+    pi2 -> pi2 : send(TMLedgerData)
+    deactivate pi2
+deactivate pi2
+
+...
+pi2 -> pi1 : onMessage(TMLedgerData)
+    pi1 -> jq1 : addJob(() -> InboundTransactions.gotData(hash, message))
+    return
+deactivate pi1
+
+...
+jq1 -> ib1 : gotData(hash, message)
+    ib1 -> ta1 : takeNodes(nodeIDs, blobs)
+    return useful | invalid
+deactivate ib1
+
+...
+timer1 -> ta1 : queueJob()
+    ta1 -> jq1 : addJob(() -> TransactionAcquire.invokeOnTimer())
+    return
+deactivate ta1
+
+...
+jq1 -> ta1 : invokeOnTimer()
+    ta1 -> ta1 : onTimer()
+        ta1 -> ta1 : addPeers(limit=1)
+            ta1 -> ta1 : onPeerAdded(peer)
+                ta1 -> ta1 : trigger(peer)
+                note right: mComplete = true;
+                deactivate ta1
+            deactivate ta1
+        deactivate ta1
+    deactivate ta1
+deactivate ta1
+@enduml
--- a/src/ripple/app/ledger/impl/InboundLedger.cpp
+++ b/src/ripple/app/ledger/impl/InboundLedger.cpp
@@ -77,12 +77,8 @@ InboundLedger::InboundLedger(
     std::uint32_t seq,
     Reason reason,
     clock_type& clock)
-    : PeerSet(
-          app,
-          hash,
-          ledgerAcquireTimeout,
-          clock,
-          app.journal("InboundLedger"))
+    : PeerSet(app, hash, ledgerAcquireTimeout, app.journal("InboundLedger"))
+    , m_clock(clock)
     , mHaveHeader(false)
     , mHaveState(false)
     , mHaveTransactions(false)
@@ -93,6 +89,7 @@ InboundLedger::InboundLedger(
     , mReceiveDispatched(false)
 {
     JLOG(m_journal.trace()) << "Acquiring ledger " << mHash;
+    touch();
 }
 
 void
@@ -156,6 +153,20 @@ InboundLedger::init(ScopedLockType& collectionLock)
         app_.getLedgerMaster().checkAccept(mLedger);
 }
 
+std::size_t
+InboundLedger::getPeerCount() const
+{
+    std::size_t ret(0);
+
+    for (auto id : mPeers)
+    {
+        if (app_.overlay().findPeerByShortID(id))
+            ++ret;
+    }
+
+    return ret;
+}
+
 void
 InboundLedger::execute()
 {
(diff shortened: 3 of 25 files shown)


src/ripple/app/ledger/InboundTransactions.uml
```diff
@@ -0,0 +1,81 @@
...
+    ta1 -> jq1 : addJob(() -> TransactionAcquire.invokeOnTimer())
+    return
+deactivate ta1
+
+...
+jq1 -> ta1 : invokeOnTimer()
+    ta1 -> ta1 : onTimer()
+        ta1 -> ta1 : addPeers(limit=1)
+            ta1 -> ta1 : onPeerAdded(peer)
+                ta1 -> ta1 : trigger(peer)
```
carlhua: I am a huge fan of PlantUML sequence diagrams - just one nit here. Generally, I find it more helpful when the steps are called out with reader-friendly English vs function names. It gives more context.  For example, trigger(peer) although explicit, gives little context. Can we consider adding more descriptive language to the sequence diagram?

thejohnfreeman: I wrote this to help me visualize the call graph, so I could take a step back and see the forest:

- who are the classes involved?
- how does data flow?
- what are the dependency edges?
- what is the concurrency?

I just didn't want to throw it away, so I tucked it in this file. I'm not sure it's worth expanding because I don't think it will remain correct for much longer, since I'm trying to redesign these components.

I think the problem with `trigger` is just that it's a bad method name, along with `onTimer`, `invokeOnTimer`, and `gotData`. I had already renamed `onPeerAdded` and `queueJob` (formerly `execute`) to be more understandable.

src/ripple/app/ledger/impl/InboundLedger.cpp
```diff
@@ -440,14 +461,14 @@ void InboundLedger::onTimer (bool wasProgress, ScopedLockType&)
 /** Add more peers to the set, if possible */
 void InboundLedger::addPeers ()
 {
-    app_.overlay().selectPeers (*this,
+    PeerSet::addPeers(
         (getPeerCount() == 0) ? peerCountStart : peerCountAdd,
-        ScoreHasLedger (getHash(), mSeq));
+        ScoreHasLedger(mHash, mSeq));
```
nbougalis: I think I wrote `ScoreHasLedger` and friends. Back then we didn't have lambdas, but now that we do, I think that we should consider using them to "inline" the code instead of requiring the user to look up a hand-rolled functor. One minor caveat: to do so _may_ require introducing functions that are currently "tucked away" in the `PeerImp` interface to the `Peer` interface. If that's the case, then it's probably not a good trade-off.

src/ripple/app/ledger/InboundLedger.h
```diff
@@ -99,6 +123,24 @@ class InboundLedger
     LedgerInfo
     deserializeHeader(Slice data, bool hasPrefix);
 
+    void
+    touch()
```
nbougalis: I'd recommend having this accept a `clock_type::time_point` and eliminating the need to keep `m_clock` as a member. You'd also need to change the constructor to take a `clock_type::time_point` as well.

src/ripple/overlay/impl/PeerSet.cpp
```diff
@@ -104,47 +121,26 @@ void PeerSet::invokeOnTimer ()
...
     ScopedLockType sl (mLock);
 
     if (mPeers.empty ())
         return;
 
-    auto packet = std::make_shared<Message>(tmGL, protocol::mtGET_LEDGER);
-
     for (auto id : mPeers)
     {
         if (auto peer = app_.overlay ().findPeerByShortID (id))
```
nbougalis: Post-refactoring, `auto peer` will shadow the function argument. Consider using `p`.

src/ripple/app/ledger/InboundLedger.h
```diff
@@ -99,6 +123,24 @@ class InboundLedger
     LedgerInfo
     deserializeHeader(Slice data, bool hasPrefix);
```
nbougalis: Generic comment, just because I came across this code:

1. This should not be a `static` member of this class; it has very little do with the class and it's just here because it's just where it happened to be put at some point. It should be a free function.
2. The `bool` argument to "skip" bytes is ugly. Callers that need to skip the header should just skip the first 4 bytes and pass the correct `Slice` to this function.

No need to act on this, really. it may be nice to document it or, potentially, open an issue so that it can tackled in the future.

src/ripple/app/ledger/impl/InboundLedger.cpp
```diff
@@ -144,7 +149,22 @@ InboundLedger::init(ScopedLockType& collectionLock)
         app_.getLedgerMaster().checkAccept(mLedger);
 }
 
-void InboundLedger::execute ()
+std::size_t
+InboundLedger::getPeerCount() const
+{
+    std::size_t ret(0);
```
nbougalis: Is this clearer?

```
std::size_t
InboundLedger::getPeerCount() const
{
    return std::count_if(mPeers.begin(), mPeers.end(), [this](auto id)
        { return app_.overlay().findPeerByShortID(id) != nullptr; });
}
```

src/ripple/app/ledger/InboundLedger.h
thejohnfreeman: I've done both (push incoming). Minor thought: it might be nice to prefer static functions in classes as an extra namespace layer, to avoid polluting the `ripple` namespace. I don't feel too strongly right now, but I like when I can tell where a function comes from by reading its call site or the import statements at the top of the file. C++ makes this difficult with header inclusion and huge, open namespaces.

src/ripple/app/ledger/InboundLedger.h
thejohnfreeman: I'm going to skip this one for now. I'm first trying to redesign the whole `InboundLedgers` + `InboundLedger` architecture.

src/ripple/app/main/Application.cpp
```diff
@@ -491,7 +491,7 @@ class ApplicationImp
             *m_jobQueue, m_collectorManager->collector ()))
 
         , m_inboundTransactions (make_InboundTransactions
-            ( *this, stopwatch()
+            ( *this
```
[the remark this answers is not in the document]
thejohnfreeman: I don't understand this question. The `clock_type` in `InboundLedger` after this changeset is the same as the one before, as far as I can tell. I just pushed it down from `PeerSet`.

src/ripple/overlay/impl/PeerSet.cpp
```diff
@@ -46,15 +46,41 @@ PeerSet::PeerSet(
...
+        });
+
+    std::size_t accepted = 0;
+    ScopedLockType sl(mLock);
+    for (auto const& pair : pairs)
+    {
+        auto const peer = pair.second;
+        if (!mPeers.insert(peer->id()).second)
+            continue;
+        newPeer(peer);
```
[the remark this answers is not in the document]
thejohnfreeman: Is "this" referring to the call to `newPeer()`? The old implementation of `selectPeers()` [called](https://github.com/ripple/rippled/blob/0c6d380780ae368a2236a2e8e3e42efa4a1d2b46/src/ripple/overlay/impl/OverlayImpl.cpp#L873) `PeerSet::insert` (not `std::set::insert` directly), [which](https://github.com/ripple/rippled/blob/4a1148eb2849513dd1e7ae080288fd47ab57a376/src/ripple/overlay/impl/PeerSet.cpp#L58-L67) acquired the lock and called `newPeer()`. Instead of toggling the lock on and off, I figured I'd move it out of the loop and acquire it once for the whole set.

src/ripple/nodestore/impl/DatabaseShardImp.cpp
```diff
@@ -510,8 +510,8 @@ DatabaseShardImp::fetchLedger(uint256 const& hash, std::uint32_t seq)
         return {};
     };
 
-    auto ledger {std::make_shared<Ledger>(
-        InboundLedger::deserializeHeader(makeSlice(nObj->getData()), true),
+    auto ledger{std::make_shared<Ledger>(
+        deserializeHeader(makeSlice(nObj->getData()) + 4),
```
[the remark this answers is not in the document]
thejohnfreeman: I personally hate Boolean parameters. I'll add a second function that calls the first.

src/ripple/overlay/PeerSet.h
```diff
@@ -74,12 +68,10 @@ class PeerSet
...
 
-    bool isActive ();
-
-    /** Called to indicate that forward progress has been made. */
-    void progress ()
+    bool
+    isActive() const
     {
-        mProgress = true;
+        return !isDone();
```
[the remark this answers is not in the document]
thejohnfreeman: That's a great question. `isActive()`, `isComplete()`, `isFailed()`, `getTimeouts()`, `getSeq()`, and `getLastAction()` are only ever called in one place, by `InboundLedgers`. `isActive()` acquired the lock, but none of the others did, including `getTimeouts()` which is called right after `isActive()`. I'll have to dig deeper to see if those fields are ever accessed concurrently.

src/ripple/overlay/impl/PeerSet.cpp
```diff
@@ -62,7 +59,7 @@ bool PeerSet::insert (std::shared_ptr<Peer> const& ptr)
 
 void PeerSet::setTimer ()
 {
-    mTimer.expires_from_now(mTimerInterval);
+    mTimer.expires_after(mTimerInterval);
```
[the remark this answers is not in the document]
thejohnfreeman: `expires_from_now` is [deprecated since at least 1.66](https://www.boost.org/doc/libs/1_66_0/doc/html/boost_asio/reference/basic_waitable_timer/expires_from_now.html), and the suggested alternative is `expires_after`.

src/ripple/overlay/PeerSet.h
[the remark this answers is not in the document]
mtrippled: isDone() is the thing that was locked by isActive() previously. mComplete and mFailed are the variables in that function. They are each called by a whole bunch of things.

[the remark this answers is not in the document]
thejohnfreeman: I'm not sure what your point is. `isDone()` is only ever called within the type hierarchy, hence why it is protected in `PeerSet` (at least after these changes). `mComplete` and `mFailed` are protected as well. I was assuming that all of their calls and accesses were thread-safe going into these changes; if they were, then they remain so after these changes.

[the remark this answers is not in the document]
mtrippled: Before:

PeerSet::mComplete and PeerSet::mFailed are locked by PeerSet::mLock in PeerSet::isActive()

https://github.com/ripple/rippled/blob/b54d672710227e494dd774c3ca3a4bc0a9976233/src/ripple/overlay/impl/PeerSet.cpp#L107-L111

bool PeerSet::isActive ()
{
    ScopedLockType sl (mLock);
    return !isDone ();
}

https://github.com/ripple/rippled/blob/b54d672710227e494dd774c3ca3a4bc0a9976233/src/ripple/overlay/PeerSet.h#L101-L104

    virtual bool isDone () const
    {
        return mComplete || mFailed;
    }

-------------------------------------------------

After:

PeerSet::mComplete and PeerSet::mFailed are not locked by PeerSet::mLock when PeerSet::isActive() is called.

https://github.com/thejohnfreeman/rippled/blob/peerset/src/ripple/app/ledger/InboundLedger.h#L128-L132

    bool
    isActive() const
    {
        return !isDone();
    }

https://github.com/thejohnfreeman/rippled/blob/peerset/src/ripple/overlay/PeerSet.h#L81-L85

    bool
    isDone() const
    {
        return mComplete || mFailed;
    }

------------------------------------------------------

Solution:
The 1 caller of isActive() in turn has no callers, so both can be removed: https://github.com/mtrippled/rippled/commit/8b6ba458c3464f98a06041509bc8a06073e98fff

src/ripple/app/ledger/InboundTransactions.uml
```diff
@@ -0,0 +1,81 @@
...
+    participant pi1 as "PeerImp"
+end box
+
+box "Server 2"
+    participant pi2 as "PeerImp"
+    participant jq2 as "JobQueue"
+    participant ib2 as "InboundTransactions"
+end box
+
+autoactivate on
```
cjcobb23, line 16: What is the activation supposed to represent in this sequence diagram?

src/ripple/app/ledger/impl/InboundLedger.cpp
```diff
@@ -66,20 +66,25 @@ enum
...
-    , mSeq (seq)
-    , mReason (reason)
-    , mReceiveDispatched (false)
+InboundLedger::InboundLedger(
+    Application& app,
+    uint256 const& hash,
+    std::uint32_t seq,
+    Reason reason,
+    clock_type& clock)
+    : PeerSet(app, hash, ledgerAcquireTimeout, app.journal("InboundLedger"))
```
nbougalis, line 75: Is it possible to tweak `clang-format` here so that _either_ the argument list or the initializer list gets an extra indent?

thejohnfreeman: We probably want to get that change into the format PR before it gets merged. I'll note the argument list and initializer list are already distinguishable by the leading separators for each initializer (either `:` or `,`).

thejohnfreeman: We can set `ConstructorInitializerIndentWidth` so that the initializers _always_ get an extra indent (even if there no parameters).

src/ripple/overlay/Overlay.h
```diff
@@ -174,73 +172,22 @@ class Overlay
...
+     *
+     * @param f the invocable to call with every peer
+     */
     template <class Function>
-    std::enable_if_t <
-        std::is_void <typename Function::return_type>::value,
-        typename Function::return_type
-    >
-    foreach(Function f)
+    // REVIEWER: Should we use enable_if in place of a concept for now?
```
nbougalis, line 183: I’d prefer a constraint via `std::enable_if`, personally but up to you. Either way please either remove this comment; if you want to leave it please change it to a TODO.

thejohnfreeman: I removed the comment. There's only one overload now. A signature incompatibility will be diagnosed by a substitution error instead of a missing declaration error.

src/ripple/overlay/impl/PeerSet.cpp
```diff
@@ -26,19 +26,22 @@ namespace ripple {
...
-    , mHash (hash)
-    , mTimerInterval (interval)
-    , mTimeouts (0)
-    , mComplete (false)
-    , mFailed (false)
-    , mProgress (false)
-    , mTimer (app_.getIOService ())
+    : app_(app)
+    , m_journal(journal)
+    , mHash(hash)
```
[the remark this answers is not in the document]
thejohnfreeman: I tried this, but there are about 30 member variables that need to be renamed, leading to changes on nearly 1000 lines. I'm going to defer this.

thejohnfreeman:
I've rebased onto 1.6.0-b3. It should apply cleanly to 1.6.0-b4. I've rearranged the commits a little bit. I'd like to get this into the next beta, so pinging @mtrippled and @nbougalis for re-review.

Small point of pride: over 200 fewer lines of code!

thejohnfreeman:
Note: the commits as listed on the [Commits tab](https://github.com/ripple/rippled/pull/3320/commits) are in the wrong order. If you want to see the correct order, check the [source branch](https://github.com/thejohnfreeman/rippled/commits/peerset).

src/ripple/app/ledger/impl/InboundLedger.cpp
```diff
@@ -452,8 +452,7 @@ InboundLedger::onTimer(bool wasProgress, ScopedLockType&)
 void
 InboundLedger::addPeers()
 {
-    app_.overlay().selectPeers(
-        *this,
+    PeerSet::addPeers(
```
mtrippled, line 452: This is not moving selectPeers, instead it removes selectPeers and seems to put addPeers in its place. But these functions don't do the same thing. What's the intention for this commit?

thejohnfreeman: `selectPeers` was a method (that could have been declared const) on `Overlay` that would take a `PeerSet` by non-const reference and insert peers into it (up to some limit, ordered by a scoring function). Instead, I moved the body of that method to `PeerSet` (which takes `this` as a pointer-to-non-const). I like mutating functions to be defined in the object they mutate. These functions should do the same thing. What is the difference in behavior that you spotted?

mtrippled: 2 calls to newPeer(), and the lock that are in addPeers() but not selectPeers()

thejohnfreeman: I only see one call to `newPeer()` on [line 80 of `PeerSet.cpp`](https://github.com/thejohnfreeman/rippled/blob/e2d06dd496b271c3bbda860d7ca9cd160ed3b23d/src/ripple/overlay/impl/PeerSet.cpp#L80). Where do you see the second?

The lock was transitively acquired by `selectPeers` whenever it called `PeerSet::insert`. `PeerSet::insert` has been inlined into `PeerSet::addPeers`.

mtrippled: I just see the one, I C&P'd the function into a note and must have duplicated that call. Is newPeer() new for this function?

thejohnfreeman: The call to `newPeer()` is not new. It was [called](https://github.com/thejohnfreeman/rippled/commit/e2d06dd496b271c3bbda860d7ca9cd160ed3b23d#diff-6d55b06f2011cfb72aec36becb4e15c6L57) in `PeerSet::insert()`, which I inlined into `PeerSet::addPeers()`.

src/ripple/overlay/impl/PeerSet.cpp
```diff
@@ -26,61 +26,74 @@ namespace ripple {
...
+    overlay.foreach([&](auto const& peer) {
+        auto const score = peer->getScore(hasItem(peer));
+        pairs.emplace_back(score, std::move(peer));
+    });
+
+    std::sort(
+        pairs.begin(),
+        pairs.end(),
+        [](ScoredPeer const& lhs, ScoredPeer const& rhs) {
+            return lhs.first > rhs.first;
```
nbougalis, line 70: Would it be better if we deferred to the default `std::pair` comparison operator? Used naively, it would result in the things being sorted the "wrong" way, but that can be fixed trivially:

```
    // Sort based on scores, high to low:
    std::sort(pairs.rbegin(), pairs.rend());
```

src/ripple/overlay/impl/PeerSet.cpp
```diff
@@ -110,53 +123,26 @@ PeerSet::invokeOnTimer()
...
-{
-    ScopedLockType sl(mLock);
-
-    if (mPeers.empty())
-        return;
-
     auto packet = std::make_shared<Message>(tmGL, protocol::mtGET_LEDGER);
 
-    for (auto id : mPeers)
+    if (peer)
```
nbougalis, line 133: No change needed (and not much you can do at this point, since other code locks you into this API) but just noting that I think that it's best to be explicit about things:

```
/** Send request to all peers */
void
PeerSet::broadcastRequest(
    const protocol::TMGetLedger& tmGL);

/** Send request to the specified peer */
void
PeerSet::sendRequest(
    const protocol::TMGetLedger& tmGL,
    std::shared_ptr<Peer> const& peer);
```

src/ripple/app/ledger/impl/TransactionAcquire.cpp
```diff
@@ -223,23 +207,24 @@ TransactionAcquire::takeNodes(
             if (nodeIDit->isRoot())
             {
                 if (mHaveRoot)
-                    JLOG(j_.debug()) << "Got root TXS node, already have it";
+                    JLOG(m_journal.debug())
```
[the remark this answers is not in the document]
thejohnfreeman: Could maybe do even better with ranges...

carlhua:
@thejohnfreeman can we put the plantuml diagrams in readme.md so they are preserved with the source code?

thejohnfreeman:
@carlhua the UML is presently merged at `src/ripple/app/ledger/InboundTransactions.uml`. Is that enough for preservation, or do you really want it inlined into a README?

PR #3325, opened 2020-03-30
Proposed 1.5.0

carlhua:
- Updated version to 1.5.0

- Updated release notes

---
This change is [Reviewable](https://reviewable.io/reviews/ripple/rippled/3325)


RELEASENOTES.md
```diff
@@ -1,18 +1,40 @@
...
+ 
+Have new ideas? Need help with setting up your node? Come visit us [here](https://github.com/ripple/rippled/issues/new/choose)
 
-## Updating `rippled`
+# Releases
 
-If you are using Red Hat Enterprise Linux 7 or CentOS 7, you can [update using `yum`](https://ripple.com/build/rippled-setup/#updating-rippled). For other platforms, please [compile from source](https://wiki.ripple.com/Rippled_build_instructions).
+## Version 1.5.0
+
+The `rippled` 1.5.0 release introduces several improvements and new features, including support for gRPC API, API versioning, UNL propagation via the peer network, new RPC methods `validator_info` and `manifest`, augmented `submit` method, improved `tx` method, improved CLI parsing, improved protocol-level handshaking protocol, improved package building and various other minor bug fixes and improvements.
```
scottschurr: "improved protocol-level handshaking protocol" reads oddly.  I'm also not at all sure what it means, so I can't suggest an improvement.

Does everyone know what CLI means?  Perhaps replace that with "improved command line parsing"?

RELEASENOTES.md
```diff
@@ -1,18 +1,40 @@
...
+
+This release also introduces two new amendments: `fixQualityUpperBound` and `RequireFullyCanonicalSig`.
+ 
+**New and Updated Features**
+- Native [gRPC API](https://grpc.io/) API support. Currently, this API provides a subset of the full `rippled` [API](https://xrpl.org/rippled-api.html). You can enable the gRPC API on your server with a new configuration stanza.
```
scottschurr: "Native `gRPC API` API ..." reads oddly to me.  Consider removing the second "API".

RELEASENOTES.md
```diff
@@ -1,18 +1,40 @@
...
+This release also introduces two new amendments: `fixQualityUpperBound` and `RequireFullyCanonicalSig`.
+ 
+**New and Updated Features**
+- Native [gRPC API](https://grpc.io/) API support. Currently, this API provides a subset of the full `rippled` [API](https://xrpl.org/rippled-api.html). You can enable the gRPC API on your server with a new configuration stanza. 
+- API Versioning which allows for future breaking change of RPC methods to co-exists with existing versions.
+- Nodes now receive and broadcast UNLs over the peer network under various conditions. see `rippled` [#3072](https://github.com/ripple/rippled/pull/3072)
+- Augmented `submit` method to include additional details on the status of the command.
+- Improved `tx` method to allow the search to specify a range of ledgers using the `min_ledger` and `max_ledger` fields in the request.
+- new `validator_info` method which returns information pertaining to the current validators keys, manifest sequence, and domain.
+- new `manifest` method which looks up manifest information for the specified key (either master or ephemeral).
```
scottschurr: The other sentences in the bullet list have initial caps.  Consider giving "new" an initial cap on this line and the previous one.

RELEASENOTES.md
```diff
@@ -1,18 +1,40 @@
...
+
+This release also introduces two new amendments: `fixQualityUpperBound` and `RequireFullyCanonicalSig`.
+ 
+**New and Updated Features**
+- Native [gRPC API](https://grpc.io/) API support. Currently, this API provides a subset of the full `rippled` [API](https://xrpl.org/rippled-api.html). You can enable the gRPC API on your server with a new configuration stanza. 
+- API Versioning which allows for future breaking change of RPC methods to co-exists with existing versions.
+- Nodes now receive and broadcast UNLs over the peer network under various conditions. see `rippled` [#3072](https://github.com/ripple/rippled/pull/3072)
+- Augmented `submit` method to include additional details on the status of the command.
+- Improved `tx` method to allow the search to specify a range of ledgers using the `min_ledger` and `max_ledger` fields in the request.
+- new `validator_info` method which returns information pertaining to the current validators keys, manifest sequence, and domain.
```
scottschurr: I'm guessing "validators keys..." refers to the keys of the validators?  If that's the case then validators should probably be possessive.  Consider spelling it as "validator's".

RELEASENOTES.md
```diff
@@ -1,18 +1,40 @@
...
+- new `manifest` method which looks up manifest information for the specified key (either master or ephemeral).
+- Introduce handshake protocol for compression negotiation (compression is not implemented at this point) and other minor improvements. 
+- Remove various old conditionals introduced by amendments.
+- Add `getRippledInfo` to packages to include info gathering script in `rippled` Linux packages.
```
scottschurr: This sentence reads oddly to me.  Consider instead...

"Add `getRippledInfo` info gathering script to `rippled` Linux packages."

RELEASENOTES.md
```diff
@@ -1,18 +1,40 @@
...
+- The `RequireFullyCanonicalSig` which changes the signature requirements for the XRP Ledger protocol so that non-fully-canonical signatures are no longer valid in any case. This protects against transaction malleability on all transactions, instead of just transactions with the tfFullyCanonicalSig flag enabled. Without this amendment, a transaction is malleable if it uses a secp256k1 signature and does not have tfFullyCanonicalSig enabled. Most signing utilities enable tfFullyCanonicalSig by default, but there are exceptions. With this amendment, no single-signed transactions are malleable. (Multi-signed transactions may still be malleable if signers provide more signatures than are necessary.) All transactions must use the fully canonical form of the signature, regardless of the tfFullyCanonicalSig flag. Signing utilities that do not create fully canonical signatures are not supported. All of Ripple's signing utilities have been providing fully-canonical signatures exclusively since at least 2014. For more information, see `rippled` issue [#3042](https://github.com/ripple/rippled/issues/3042).
```
scottschurr: Perhaps append "amendment" to `RequireFullyCanonicalSig`.  Also remove "in any case" from the first sentence.  So the first sentence would read...

"The `RequireFullyCanonicalSig` amendment which changes the signature requirements for the XRP Ledger protocol so that non-fully-canonical signatures are no longer valid."

RELEASENOTES.md
```diff
@@ -1,18 +1,40 @@
...
+- `tx` method now properly fetches all historical tx if they are incorporated into a validated ledger under rules that applied at the time. 
+- Fix to how `fail_hard` flag is handled with the `submit` method - transactions that are submitted with the `fail_hard` flag that result in any TER code besides tesSUCCESS is neither queued nor held.
+- Remove unused `Beast` code.
+- Lag ratchet code fix to use proper ephemeral public keys instead of the long-term master public keys.
```
scottschurr: I don't believe "Lag ratchet" is a well known term of art; it certainly isn't for me and Google doesn't seem to recognize it either.  So for many readers (including me) this sentence will make no sense.

mtrippled: A lag ratchet is something that reduces lag incrementally.

RELEASENOTES.md
```diff
@@ -1,18 +1,45 @@
...
+At this time, the developers recommend running with sharding disabled, pending the improvements scheduled to be introduced with 1.6. For more information on how to disable sharding, please visit https://xrpl.org/configure-history-sharding.html
+
+ 
+**New and Updated Features**
+- Native [gRPC API](https://grpc.io/) API support. Currently, this API provides a subset of the full `rippled` [API](https://xrpl.org/rippled-api.html). You can enable the gRPC API on your server with a new configuration stanza. 
+- API Versioning which allows for future breaking change of RPC methods to co-exists with existing versions.
+- Nodes now receive and broadcast UNLs over the peer network under various conditions. see `rippled` [#3072](https://github.com/ripple/rippled/pull/3072)
+- Augmented `submit` method to include additional details on the status of the command.
+- Improved `tx` method to allow the search to specify a range of ledgers using the `min_ledger` and `max_ledger` fields in the request.
```
cjcobb23: Nit: I'm not sure that this sentence is correct. See the conversation here: https://github.com/ripple/xrpl-dev-portal/pull/809#discussion_r399539244. The semantics of `min_ledger` and `max_ledger` are kind of confusing (the method only uses those parameters for error handling; they don't affect the search). The semantics are a bit messy to put in the release notes, but I don't want the release notes to be misleading. So it's up to you whether you want to change this sentence or leave as is.

RELEASENOTES.md
carlhua: Updated the wording per your comments - also changed "protocol-level handshaking protocol" to "improved handshake protocol"

RELEASENOTES.md
carlhua: Updated per your comment

RELEASENOTES.md
carlhua: Good catch, updated.

RELEASENOTES.md
carlhua: Thanks! Updated.

RELEASENOTES.md
carlhua: Good catch! Updated

RELEASENOTES.md
carlhua: Yes, updated. It reads better now!

RELEASENOTES.md
carlhua: I will keep "lag ratchet" since its mentioned in the comments - I added a reference to that commit.

RELEASENOTES.md
carlhua: Updated to correctly describe the improvement.

RELEASENOTES.md
```diff
@@ -1,18 +1,45 @@
...
+This release also introduces two new amendments: `fixQualityUpperBound` and `RequireFullyCanonicalSig`.
+
+Several improvements to the sharding system are currently being evaluated for inclusion into the upcoming 1.6 release of `rippled`. These changes are incompatible with shards generated by previous versions of the code. 
+Additionally, an issue with the existing sharding engine can result in a server running versions 1.4 or 1.5 of the software to experience a deadlock and automatically restart when running with the sharding feature enabled. 
+At this time, the developers recommend running with sharding disabled, pending the improvements scheduled to be introduced with 1.6. For more information on how to disable sharding, please visit https://xrpl.org/configure-history-sharding.html
+
+ 
+**New and Updated Features**
+- Native [gRPC API](https://grpc.io/) support. Currently, this API provides a subset of the full `rippled` [API](https://xrpl.org/rippled-api.html). You can enable the gRPC API on your server with a new configuration stanza. 
+- API Versioning which allows for future breaking change of RPC methods to co-exists with existing versions.
```
scottschurr: Sorry, here's one I missed on the previous pass.  Consider replacing "co-exists" with "co-exist", so "... to co-exist with ..."

carlhua: Good catch! Updated!

RELEASENOTES.md
```diff
@@ -1,18 +1,45 @@
...
+- The `RequireFullyCanonicalSig` amendment which changes the signature requirements for the XRP Ledger protocol so that non-fully-canonical signatures are no longer valid. This protects against transaction malleability on all transactions, instead of just transactions with the tfFullyCanonicalSig flag enabled. Without this amendment, a transaction is malleable if it uses a secp256k1 signature and does not have tfFullyCanonicalSig enabled. Most signing utilities enable tfFullyCanonicalSig by default, but there are exceptions. With this amendment, no single-signed transactions are malleable. (Multi-signed transactions may still be malleable if signers provide more signatures than are necessary.) All transactions must use the fully canonical form of the signature, regardless of the tfFullyCanonicalSig flag. Signing utilities that do not create fully canonical signatures are not supported. All of Ripple's signing utilities have been providing fully-canonical signatures exclusively since at least 2014. For more information. [`ec137044a`](https://github.com/ripple/rippled/commit/ec137044a014530263cd3309d81643a5a3c1fdab)
```
ximinez: I asked the same question over in https://github.com/ripple/dev-blog/pull/40#discussion_r400512220:

Is this really a bug being fixed? Am I wrong that allowing non-canonical sigs was an early intentional design decision?

mDuo13: _(Copying my comment over here, where the public can see it.)_

Allowing non-fully-canonical signatures was not an intentional design decision, to my understanding. There have definitely been significant advances in the state of elliptic curve cryptography since 2009 (when Bitcoin's signing code, which the XRP ledger copied, was written) and 2011 (when @JoelKatz started writing the XRPL code). For example, deterministic nonces¹, which have since become an important security precaution for ECDSA signatures, weren't specified [until 2013](https://tools.ietf.org/html/rfc6979).

I believe the way it happened was that only after the signature code had been put in place did it become apparent that the rules for "canonical" signatures were not sufficient to make sure any given signature could not be malleable. When this became known, the devs added the tfFullyCanonical flag to provide protection, and had the official signing code turn this flag on by default whenever possible. I don't know if the idea at the time was for this to be a transitional step to provide time to upgrade any code that made non-fully-canonical signatures; but I think it served that purpose. At this point, the benefits of maintaining that legacy support are outweighed by the benefits of protecting more transactions against malleability.

Ultimately I would suggest categorizing this as an "improvement" that is neither a feature nor a bug fix.

¹ In contrast to fully-canonical-signatures, which you can recognize with only the signature and the public key necessary to confirm it, I think there is no way to know whether a signature uses a deterministic nonce unless you know the private key. So, unfortunately, the ledger cannot enforce deterministic nonces in the way it enforces fully-canonical signatures.

carlhua: Updated and moved it to the section above.

RELEASENOTES.md
```diff
@@ -1,18 +1,45 @@
...
+![XRP](docs/images/xrp-text-mark-black-small@2x.png)
 
-This document contains the release notes for `rippled`, the reference server implementation of the Ripple protocol. To learn more about how to build and run a `rippled` server, visit https://ripple.com/build/rippled-setup/
+This document contains the release notes for `rippled`, the reference server implementation of the Ripple protocol. To learn more about how to build, run or update a `rippled` server, visit https://xrpl.org/install-rippled.html
 
-> **Do you work at a digital asset exchange or wallet provider?** 
-> 
-> Please [contact us](mailto:support@ripple.com). We can help guide your integration.
+ 
+Have new ideas? Need help with setting up your node? Come visit us [here](https://github.com/ripple/rippled/issues/new/choose)
```
MarkusTeufelberger: So the issue tracker is now the official point for support requests?

PR #3326, opened 2020-03-30
Cleanup

nbougalis:
This PR includes commits that:

1. address an issue that can occur during the loading of validator tokens, where a deliberately malformed token could cause the server to crash during startup. The issue was identified by @GuidoVranken.
2. Cleanup & modernize code, remove obsolete workarounds, code and comments.


src/ripple/app/misc/Manifest.h
```diff
@@ -167,18 +167,11 @@ struct ValidatorToken
 {
     std::string manifest;
     SecretKey validationSecret;
-
-private:
-    ValidatorToken(std::string const& m, SecretKey const& valSecret);
-
-public:
-    ValidatorToken(ValidatorToken const&) = delete;
```
undertome: It seems like protection against certain methods of construction (default, and from a reference) is now absent. Is this intended?

src/ripple/protocol/impl/STArray.cpp
```diff
@@ -26,10 +26,6 @@ namespace ripple {
 
 STArray::STArray()
 {
-    // VFALCO NOTE We need to determine if this is
```
undertome: With this comment removed, seems like we can just make defaulted in the header.

src/test/basics/TaggedCache_test.cpp
```diff
@@ -101,8 +101,8 @@ class TaggedCache_test : public beast::unit_test::suite
             BEAST_EXPECT(! c.insert (3, "three"));
 
             {
-                Cache::mapped_ptr const p1 (c.fetch (3));
-                Cache::mapped_ptr p2 (std::make_shared <Value> ("three"));
+                auto p1 = c.fetch (3);
```
undertome: const qualifier wasn't preserved

src/test/basics/TaggedCache_test.cpp
```diff
@@ -123,7 +123,7 @@ class TaggedCache_test : public beast::unit_test::suite
 
             {
                 // Keep a strong pointer to it
-                Cache::mapped_ptr p1 (c.fetch (4));
+                auto p1 = c.fetch (4);
```
undertome: Might as well add the const qualifier here since we're doing a cleanup anyway

src/ripple/protocol/impl/STArray.cpp
```diff
@@ -53,7 +49,7 @@ STArray::STArray (int n)
 STArray::STArray (SField const& f)
     : STBase (f)
 {
-    v_.reserve(reserveSize);
+
```
HowardHinnant, line 52: This should now be `= default` in the header.

nbougalis: So if I do: `STArray(SField const& f) = default` then the compiler will _automatically_ find the `STBase::STBase(SField const& f)` constructor and call it?

HowardHinnant: Wow, I completely misread this!  For some reason I thought I was talking about the `STArray` default constructor.  Please ignore.  Sorry for the noise.

But adding `explicit` does sound like a good idea.

src/ripple/app/misc/Manifest.h
nbougalis: Yeah. The type is now basically a glorified `std::pair` and can use uniform initialization. I'd have used `std::pair` but I think the consensus of everyone on our team is that `.manifest` and `.validationSecret` are more helpful that `.first` and `.second`, which is a fair point.

src/ripple/protocol/impl/Indexes.cpp
```diff
@@ -141,102 +108,52 @@ getQuality (uint256 const& uBase)
...
 }
 
-Keylet child (uint256 const& key)
+Keylet child (uint256 const& key) noexcept
 {
     return { ltCHILD, key };
 }
 
-Keylet skip_t::operator()() const
+Keylet const& skip() noexcept
```
seelabs, line 131: Using a `static` and returning a `const&` instead of value may be a pessimization because of RVO, it's going to depend on how it's called. If it's used in a function call and bound to a `const&` paramter (`foo(skip())`) then this is an optimization. If it's used to initialize a value (`auto const a = skip()`) then it's a pessimization (there's extra coping because of missed RVO).

I think I'd keep this as `Keylet skip() noexcept`, but I'm OK if you want to keep the `const&`. I really just wanted to make sure you knew there was a trade-off here to consider. (same somment for `fees` as well)

src/ripple/ledger/impl/View.cpp
```diff
@@ -1132,19 +1096,45 @@ rippleCredit (ApplyView& view,
...
+                             index.key,
+                             sleAccount,
+                             false,
+                             noRipple,
+                             false,
+                             saBalance,
+                             saReceiverLimit,
+                             0,
+                             0,
+                             j);
```
seelabs, line 1137: Nit: formatting

src/ripple/protocol/impl/UintTypes.cpp
```diff
@@ -45,59 +59,50 @@ std::string to_string(Currency const& currency)
...
     }
 
-    static const int CURRENCY_CODE_LENGTH = 3;
-    if (code.size () == CURRENCY_CODE_LENGTH)
+    // Handle ISO-4217-like 3-digit character codes.
+    if (code.size () == detail::isoCodeLength)
     {
-        Blob codeBlob (CURRENCY_CODE_LENGTH);
+        if (code.find_first_not_of(detail::isoCharSet) != std::string::npos)
+            return false;
```
seelabs, line 90: Making a note that I need to look at this change more closely. It looks like the new code does extra error checking that the old code does not and I want to make sure we don't break anything with the new check.

src/ripple/protocol/impl/Indexes.cpp
nbougalis: The thing is that these particular keylets are **actually** constants; they accept no arguments and their value is a fixed part of the protocol and will never change. From that perspective, I think that the return value being a `const&` is relevant.

However, you raise a good point that it may, actually, result in a pessimization. I'll take a look.

src/ripple/protocol/impl/UintTypes.cpp
seelabs: I looked and I don't see how the extra error checking could be transaction breaking. So resolving this comment.

nbougalis:
I am going to rebase this against **1.6.0-b4**

src/ripple/app/consensus/RCLConsensus.cpp
```diff
@@ -753,41 +761,63 @@ RCLConsensus::Adaptor::validate(
...
+            {
+                auto const& ft = app_.getFeeTrack();
+                auto const fee = std::max(ft.getLocalFee(), ft.getClusterFee());
+                if (fee > ft.getLoadBase())
+                    v.setFieldU32(sfLoadFee, fee);
+            }
+
+            // If the next ledger is a flag ledger, suggest fee changes and
+            // new features:
+            if ((ledger.seq() + 1) % 256 == 0)
```
p2peer, line 811: I heard that (x & 0xff) works faster than (x % 256).

src/ripple/protocol/STValidation.h
```diff
@@ -25,15 +25,20 @@
...
 
 // Validation flags
-const std::uint32_t vfFullyCanonicalSig =
-    0x80000000;  // signature is fully canonical
+
+// This is a full (as opposed to a partial) validation
+constexpr std::uint32_t vfFullValidation = 0x00000001;
+
+// The signature is fully canonical
+constexpr std::uint32_t vfFullyCanonicalSig = 0x80000000;
```
p2peer, line 41: May be have a sense to use prefix 'u' because 0x80000000 is too big?

src/ripple/app/consensus/RCLConsensus.cpp
```diff
@@ -790,6 +791,11 @@ RCLConsensus::Adaptor::validate(
                     v.setFieldH256(sfValidatedHash, vl->info().hash);
 
                 v.setFieldU64(sfCookie, valCookie_);
+
+                // Report our server version every flag ledger:
+                if ((ledger.seq() + 1) % 256 == 0)
```
p2peer, line 796: One more x % 256.

src/ripple/protocol/impl/BuildInfo.cpp
```diff
@@ -76,6 +77,77 @@ getFullVersionString()
     return value;
 }
 
+std::uint64_t
+getEncodedVersion()
+{
+    static std::uint64_t const cookie = []() {
+        std::uint64_t c = 0x183B000000000000;
```
p2peer, line 84: May be 'ull' prefix?

src/ripple/app/consensus/RCLConsensus.cpp
nbougalis: Thanks for pointing this out. I would prefer not to make this change for three reasons:

1. I feel that it would sacrifice readability for humans, which is an important consideration, especially in this code.
2. A decent compiler would _probably_ perform this optimization on our behalf if it were advantageous on the underlying platform.
3. Lastly, even if this is suboptimal and the compiler doesn't address it, it's called so infrequently in the grand scheme of things (roughly once every 4 seconds) that I doubt the gains would be measurable.

src/ripple/protocol/STValidation.h
nbougalis: Hmm, I think that's safe, but let me ask the gurus: @HowardHinnant, @scottschurr, any suggestions here?

src/ripple/protocol/impl/BuildInfo.cpp
nbougalis: @HowardHinnant, @scottschurr, any suggestions here?

src/ripple/app/consensus/RCLConsensus.cpp
seelabs: FYI, both clang and gcc generate the same code for both versions: https://godbolt.org/z/a7y3mq

src/ripple/protocol/STValidation.h
HowardHinnant: It's safe as is, but I like using the `u` suffix when initializing unsigned.  But I also alphabetize my socks.  So you can't always go by me. :-)

src/ripple/protocol/impl/BuildInfo.cpp
HowardHinnant: I like the `ull` suffix here.  You could try it as a prefix, but I don't think it will compile. :-)  That being said, it's optional.  Just a readability issue.

src/ripple/protocol/STValidation.h
nbougalis: :socks:

scottschurr: 🤣 All of my socks are identical so I don't have to alphabetize them.  I also never need to match them.

PR #3328, opened 2020-04-01
[TRIVIAL] Add README for gRPC protobuf folder

cjcobb23:
This README is meant to assist any developer that is modifying the gRPC protocol buffer definitions, to ensure that the developer does not break the wire format, as well as adheres to the current pattern the protocol buffer definitions follow. The latter is not very important, but the former is very important.

---
This change is [Reviewable](https://reviewable.io/reviews/ripple/rippled/3328)

### DIFF
--- a/src/ripple/proto/org/xrpl/rpc/v1/README.md
+++ b/src/ripple/proto/org/xrpl/rpc/v1/README.md
@@ -1,0 +1,35 @@
+This folder contains the protocol buffer definitions used by the rippled gRPC API.
+The gRPC API attempts to mimic the JSON/Websocket API as much as possible.
+As of April 2020, the gRPC API supports a subset of the full rippled API:
+tx, account_tx, account_info, fee and submit.
+
+### Making Changes
+#### Wire Format and Backwards Compatibility
+When making changes to the protocol buffer definitions in this folder, care must
+be taken to ensure the changes do not break the wire format, which would break
+backwards compatibility. At a high level, do not change any existing fields.
+This includes the field's name, type and field number. Do not remove any
+existing fields. It is always safe to add fields; just remember to give each of
+the new fields a unique field number. The field numbers don't have to be in any
+particular order and there can be gaps. More info about what changes break the
+wire format can be found
+[here](https://developers.google.com/protocol-buffers/docs/proto3#updating).
+#### Conventions
+For fields that are reused across different message types, we define the field as a unique
+message type in common.proto. The name of the message type is the same as the
+field name, with the exception that the field name itself is snake case, whereas
+the message type is in Pascal case. The message type has one field, called
+`value`. This pattern does not need to be strictly followed across the entire API,
+but should be followed for transactions and ledger objects, since there is a high rate
+of field reuse across different transactions and ledger objects.
+The motivation for this pattern is two-fold. First, we ensure the field has the
+same type everywhere that the field is used. Second, wrapping primitive types in
+their own message type prevents default initialization of those primitive types.
+For example, `uint32` is always initialized to `0`, even if never explicitly set;
+there is no way to tell if the client or server set the field to `0` (which may be
+a valid value for the field) or the field was default initialized.
+
+Refer to the Protocol Buffers [language
+guide](https://developers.google.com/protocol-buffers/docs/proto3)
+for more detailed information about Protocol Buffers.
+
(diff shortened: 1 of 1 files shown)


src/ripple/proto/org/xrpl/rpc/v1/README.md
```diff
@@ -0,0 +1,35 @@
...
+the message type is in Pascal case. The message type has one field, called
+`value`. This pattern does not need to be strictly followed across the entire API,
+but should be followed for transactions and ledger objects, since there is a high rate
+of field reuse across different transactions and ledger objects.
+The motivation for this pattern is two-fold. First, we ensure the field has the
+same type everywhere that the field is used. Second, wrapping primitive types in
+their own message type prevents default initialization of those primitive types.
+For example, `uint32` is always initialized to `0`, even if never explicitly set;
+there is no way to tell if the client or server set the field to `0` (which may be
+a valid value for the field) or the field was default initialized.
```
carlhua, line 30: Good stuff - it calls out what we should/should not do. In addition, can we add content to describe how to add/modify a message step by step? I think @scottschurr just went through this, and there were some comments. We do expect contributors to read the code but it would also be helpful to give a general guideline

cjcobb23: @carlhua Are you referring to what needs to change in the C++ code as well, or just the protobuf files themselves?

carlhua: Ideally, we give an overview of what would it take to add an RPC call end to end including protobuf so that other members of the team can help contribute

cjcobb23:
> @cjcobb23, thanks. Looks good as is.
> 
> I'll mention that when I was adding tickets support there was a name collision. There's a `message Ticket` type defined in common.proto. There's also a `message TicketObject` defined in ledger_objects.proto. When I first coded it up they had the same name. Confusing build messages ensued. But I followed a pattern I found elsewhere in ledger_objects.proto (`DepositPreauthObject`) and appended `Object` to avoid the collision.
> 
> Would it be worth saying something about name collisions and the recommended ways of working around them? Not saying it would be, just asking the question.

@scottschurr good point, I think it would be worth adding that just because the build messages are confusing. I'll say something about it.

src/ripple/proto/org/xrpl/rpc/v1/README.md
```diff
@@ -0,0 +1,35 @@
...
+message type in common.proto. The name of the message type is the same as the
+field name, with the exception that the field name itself is snake case, whereas
+the message type is in Pascal case. The message type has one field, called
+`value`. This pattern does not need to be strictly followed across the entire API,
+but should be followed for transactions and ledger objects, since there is a high rate
+of field reuse across different transactions and ledger objects.
+The motivation for this pattern is two-fold. First, we ensure the field has the
+same type everywhere that the field is used. Second, wrapping primitive types in
+their own message type prevents default initialization of those primitive types.
+For example, `uint32` is always initialized to `0`, even if never explicitly set;
```
thejohnfreeman, line 28: This might read better as "`uint32` is initialized to `0` if not explicitly set;"

src/ripple/proto/org/xrpl/rpc/v1/README.md
```diff
@@ -0,0 +1,35 @@
+This folder contains the protocol buffer definitions used by the rippled gRPC API.
+The gRPC API attempts to mimic the JSON/Websocket API as much as possible.
+As of April 2020, the gRPC API supports a subset of the full rippled API:
+tx, account_tx, account_info, fee and submit.
+
+### Making Changes
```
thejohnfreeman, line 6: I'd like at least one blank line after each header. With syntax highlighting, the distinction between headers and paragraphs is easy to see, but in situations without, it can be easy to miss without vertical separation.

src/ripple/proto/org/xrpl/rpc/v1/README.md
thejohnfreeman: On that topic, it might be good to merge this content: https://ripplelabs.atlassian.net/wiki/spaces/~189280334/pages/543031357/How+to+add+a+method+to+rippled

cjcobb23: @thejohnfreeman that doc talks about how to add non gRPC methods. I don't think that belongs in this folder with the protobuf files. My main motivation for adding this README is to attempt to prevent breaking changes, since we do not have breaking change detection as part of CI yet. I can extend this README to further talk about adding/modifying gRPC stuff in general, but I don't want to add non gRPC things into this README.
I do think the content of the doc you linked belongs in a README somewhere in the source code though.

PR #3333, opened 2020-04-06
Hosting a signers list public keys to a new location

manojsdoshi:
Changing the location where we can store our public keys used to sign our packages.


MarkusTeufelberger:
As these are public keys, maybe these could be made publicly available anyways?

PR #3342, opened 2020-04-08
Improve online delete backend lock

miguelportilla:
Addresses a possibility in the online delete process where one or more backend shared pointer references may become invalid during rotation.

### DIFF
--- a/src/ripple/app/misc/SHAMapStoreImp.cpp
+++ b/src/ripple/app/misc/SHAMapStoreImp.cpp
@@ -427,14 +427,19 @@ SHAMapStoreImp::run()
             }
 
             lastRotated_ = validatedSeq;
-            clearCaches(validatedSeq);
 
-            SavedState savedState;
-            savedState.writableDb = newBackend->getName();
-            savedState.archiveDb = dbRotating_->rotateBackends(
-                std::move(newBackend));
-            savedState.lastRotated = lastRotated_;
-            state_db_.setState(savedState);
+            dbRotating_->rotateWithLock(
+                [&](std::string const& writableBackendName) {
+                    SavedState savedState;
+                    savedState.writableDb = newBackend->getName();
+                    savedState.archiveDb = writableBackendName;
+                    savedState.lastRotated = lastRotated_;
+                    state_db_.setState(savedState);
+
+                    clearCaches(validatedSeq);
+
+                    return std::move(newBackend);
+                });
 
             JLOG(journal_.warn()) << "finished rotation " << validatedSeq;
         }
--- a/src/ripple/app/misc/SHAMapStoreImp.cpp
+++ b/src/ripple/app/misc/SHAMapStoreImp.cpp
@@ -426,22 +426,16 @@ SHAMapStoreImp::run()
                 default:;
             }
 
-            std::string nextArchiveDir =
-                dbRotating_->getWritableBackend()->getName();
             lastRotated_ = validatedSeq;
-            std::shared_ptr<NodeStore::Backend> oldBackend;
-            {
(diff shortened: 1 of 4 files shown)


nbougalis:
@miguelportilla, this branch is based off of the wrong **1.6.0-b1**; your commit needs to apply on top of 4f422f6f393b12.

src/ripple/nodestore/impl/DatabaseRotatingImp.cpp
```diff
@@ -102,19 +148,49 @@ DatabaseRotatingImp::sweep()
 std::shared_ptr<NodeObject>
 DatabaseRotatingImp::fetchFrom(uint256 const& hash, std::uint32_t seq)
 {
-    Backends b = getBackends();
-    auto nObj = fetchInternal(hash, b.writableBackend);
+    auto backends = [&] {
```
seelabs, line 151: Might be a good place for a structured binding (here and in the similar lambda in `for_each`). But fine as-is as well.

src/ripple/app/misc/SHAMapStoreImp.cpp
```diff
@@ -446,23 +446,16 @@ SHAMapStoreImp::run()
...
-                                                nextArchiveDir, lastRotated_});
-                clearCaches (validatedSeq);
-                oldBackend = dbRotating_->rotateBackends(
-                    std::move(newBackend),
-                    lock);
-            }
-            JLOG(journal_.warn()) << "finished rotation " << validatedSeq;
+            state_db_.setState(SavedState {
+                newBackend->getName(),
+                dbRotating_->getName(),
```
seelabs, line 452: Since we don't hold a lock here, we can't guarantee that the `dbRotating->getName()` is the same db as the one that's rotated out on line 456.

Maybe `rotateBackends` should return the name of the rotated db and we set the `state_db_` just after the rotation?

I'll also note that the old code had the same issue, so it's possible that the scenario above can't happen for some other reason, but I did want to point it out.

src/ripple/nodestore/impl/DatabaseRotatingImp.cpp
miguelportilla: Good idea, thanks!

src/ripple/app/misc/SHAMapStoreImp.cpp
miguelportilla: Nice find, fixed.

src/ripple/app/misc/SHAMapStoreImp.cpp
```diff
@@ -446,23 +446,17 @@ SHAMapStoreImp::run()
...
-            JLOG(journal_.warn()) << "finished rotation " << validatedSeq;
+            clearCaches(validatedSeq);
 
-            oldBackend->setDeletePath();
+            SavedState savedState;
+            savedState.writableDb = newBackend->getName();
+            savedState.archiveDb = dbRotating_->rotateBackends(
+                std::move(newBackend));
+            savedState.lastRotated = lastRotated_;
+            state_db_.setState(savedState);
```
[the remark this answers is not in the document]
nbougalis: Curious: could we "wrap" this logic inside a `TRANSACTION`?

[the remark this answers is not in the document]
mtrippled: If you mean a SQLIte transaction, no. Because rotateBackends() is not involved with SQLite, but affects the Backend objects that write to the nodestore.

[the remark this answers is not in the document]
miguelportilla: @mtrippled Good point, addressed.

PR #3344, opened 2020-04-09
Reduce calls to std::random_device:

nbougalis:
Existing per-thread PRNGs are individually initialized using calls to `std::random_device`.

If merged, this commit will use a single PRNG, initialized from `std::random_device` on startup, to seed the thread-specific PRNGs.

Acknowledgements:
    Thomas Snider, who reported this issue to Ripple on April 8, 2020.


src/ripple/basics/random.h
```diff
@@ -67,24 +68,41 @@ inline
...
+    {
+        std::random_device rng;
+
+        std::uint64_t seed = 0;
+
+        for (int i = 0; i < 16; ++i)
+            seed = (seed << 4) ^ rng();
+
+        // Exceedingly unlikely to happen, but just in case
+        if (seed == 0)
```
seelabs, line 82: No change needed, but this could be a `while`. But honestly, this is never going to be hit unless `random_device` is broken.

src/ripple/basics/random.h
```diff
@@ -67,24 +68,41 @@ inline
...
     boost::thread_specific_ptr<beast::xor_shift_engine> engine;
 
     if (!engine.get())
     {
-        std::random_device rng;
-
-        std::uint64_t seed = rng();
-
-        for (int i = 0; i < 6; ++i)
+        std::uint64_t seed = []
```
seelabs, line 98: Nit: `const`

HowardHinnant:
Consider this simplification:  https://github.com/HowardHinnant/rippled/commit/3b959fda16d4f56f13fdfbf621778e23a1ea3c1c

*  Uses `thread_local` instead of `boost::thread_specific_ptr`.
*  Uses the functionality of `uniform_int_distribution` to distribute randomness across all 64 bits of the `seed`.  `uniform_int_distribution` will call the `random_device` multiple times to achieve this.

seelabs:
@HowardHinnant That looks good to me.

HowardHinnant:
I withdraw my suggestion except for the `thread_local` subbed in for `thread_specific_ptr`.  My suggestion was based on a misunderstanding on my part that `random_device` only consumed a file handle on creation, not use.  But I can't find any evidence to support that.

nbougalis:
I've pushed an update, incorporating some of your suggestions @HowardHinnant.

nbougalis:
Ack, I seem to have accidentally deleted a comment by you Howard! Sorry!

src/ripple/basics/random.h
```diff
@@ -67,27 +67,37 @@ inline
...
         std::random_device rng;
-
-        std::uint64_t seed = rng();
-
-        for (int i = 0; i < 6; ++i)
+        std::uniform_int_distribution<std::uint64_t> distribution;
+        std::uint64_t seed;
+        do
+            seed = distribution(rng);
+        while (seed == 0);
```
HowardHinnant, line 78: These lines can be reduced down to:
```c++
std::uniform_int_distribution<std::uint64_t> distribution{1};
auto seed = distribution(rng);
```
And ditto for the same code below.

src/ripple/basics/random.h
```diff
@@ -67,27 +67,37 @@ inline
...
+        return beast::xor_shift_engine(seed);
+    }();
+
+    // This protects the seeder
+    static std::mutex m;
```
seelabs, line 83: I'd move this inside the lambda

src/ripple/basics/random.h
```diff
@@ -67,27 +67,37 @@ inline
...
-    if (!engine.get())
+    // This is used to seed the thread-specific PRNGs on demand
+    static beast::xor_shift_engine seeder = []
     {
         std::random_device rng;
-
-        std::uint64_t seed = rng();
-
-        for (int i = 0; i < 6; ++i)
+        std::uniform_int_distribution<std::uint64_t> distribution;
```
seelabs, line 74: Nit: but I'd just rename this `d`

src/ripple/basics/random.h
```diff
@@ -67,27 +67,37 @@ inline
...
+    thread_local beast::xor_shift_engine engine = []
+    {
+        std::uint64_t seed;
         {
-            if (seed == 0)
-                seed = rng();
-
-            seed ^= (seed << (7 - i)) * rng();
+            std::lock_guard lk(m);
+            std::uniform_int_distribution<std::uint64_t> distribution;
```
seelabs, line 91: Nit, but I'd rename this `d`

src/ripple/basics/random.h
```diff
@@ -67,27 +67,37 @@ inline
...
+            std::uint64_t seed;
```
seelabs, line 92: This is a bug. This shadows the outer `seed`

PR #3345, opened 2020-04-09
Fix Doxygen Workflow

thejohnfreeman:


.github/workflows/doxygen.yml
```diff
@@ -7,17 +7,19 @@ on:
 jobs:
   job:
     runs-on: ubuntu-18.04
+    container:
+      image: docker://thejohnfreeman/rippled-build-ubuntu:2944b78d22db
```
nbougalis, line 11: This is fine for now, but we should probably move to hosting this container under a Ripple account.

PR #3347, opened 2020-04-10
[Trivial] Reduce strand re-execute log message severity from error to warning:

scottschurr:
Historically strand re-execute log messages have been treated as errors.  However in the vast majority of cases these log messages are caused by well understood mechanics in the payment engine.  So usually these log messages should be treated as warnings.


mDuo13:
If these errors are "normally occurring" in normal transaction operation, shouldn't they should be an even lower level than warning? I would think warnings should be carefully reserved for messages that likely indicate a problem.

Have these errors also occurred in cases that indicate a serious server problem? Or is that purely theoretical?

seelabs:
@mDuo13 These are "normally occurring" in that there are numeric issues that cause very minor discrepancies. I would like to resolve these discrepancies, but they are do not indicate a serious server problem, no. "Warning" seems like a reasonable level for this, but I'm fine demoting them further too.

PR #3348, opened 2020-04-10
Cleanup move semantics and other misc cleanups

seelabs:
Clean up some code flagged by a static analysis run

### DIFF
--- a/src/ripple/app/ledger/Ledger.cpp
+++ b/src/ripple/app/ledger/Ledger.cpp
@@ -499,7 +499,6 @@ Ledger::rawInsert(std::shared_ptr<SLE> const& sle)
     Serializer ss;
     sle->add(ss);
     auto item = std::make_shared<SHAMapItem const>(sle->key(), std::move(ss));
-    // addGiveItem should take ownership
     if (!stateMap_->addGiveItem(std::move(item), false, false))
         LogicError("Ledger::rawInsert: key already exists");
 }
@@ -510,7 +509,7 @@ Ledger::rawReplace(std::shared_ptr<SLE> const& sle)
     Serializer ss;
     sle->add(ss);
     auto item = std::make_shared<SHAMapItem const>(sle->key(), std::move(ss));
-    // updateGiveItem should take ownership
+
     if (!stateMap_->updateGiveItem(std::move(item), false, false))
         LogicError("Ledger::rawReplace: key not found");
 }
--- a/src/ripple/app/ledger/Ledger.cpp
+++ b/src/ripple/app/ledger/Ledger.cpp
@@ -499,7 +499,7 @@ Ledger::rawInsert(std::shared_ptr<SLE> const& sle)
     Serializer ss;
     sle->add(ss);
     auto item = std::make_shared<SHAMapItem const>(sle->key(), std::move(ss));
-    // VFALCO NOTE addGiveItem should take ownership
+    // addGiveItem should take ownership
     if (!stateMap_->addGiveItem(std::move(item), false, false))
         LogicError("Ledger::rawInsert: key already exists");
 }
@@ -510,7 +510,7 @@ Ledger::rawReplace(std::shared_ptr<SLE> const& sle)
     Serializer ss;
     sle->add(ss);
     auto item = std::make_shared<SHAMapItem const>(sle->key(), std::move(ss));
-    // VFALCO NOTE updateGiveItem should take ownership
+    // updateGiveItem should take ownership
     if (!stateMap_->updateGiveItem(std::move(item), false, false))
         LogicError("Ledger::rawReplace: key not found");
 }
--- a/src/ripple/app/main/Application.cpp
+++ b/src/ripple/app/main/Application.cpp
@@ -587,7 +587,8 @@ public:
         , m_loadManager(
               make_LoadManager(*this, *this, logs_->journal("LoadManager")))
 
-        , txQ_(make_TxQ(setup_TxQ(*config_), logs_->journal("TxQ")))
+        , txQ_(
+              std::make_unique<TxQ>(setup_TxQ(*config_), logs_->journal("TxQ")))
 
         , sweepTimer_(get_io_service())
 
@@ -1832,8 +1833,8 @@ ApplicationImp::fdRequired() const
     // Standard handles, config file, misc I/O etc:
     int needed = 128;
 
-    // 1.5 times the configured peer limit for peer connections:
-    needed += static_cast<int>(0.5 + (1.5 * overlay_->limit()));
+    // 2x the configured peer limit for peer connections:
+    needed += 2 * overlay_->limit();
 
     // the number of fds needed by the backend (internally
     // doubled if online delete is enabled).
--- a/src/ripple/app/misc/NetworkOPs.cpp
+++ b/src/ripple/app/misc/NetworkOPs.cpp
@@ -81,7 +81,7 @@ class NetworkOPsImp final : public NetworkOPs
         bool const admin;
         bool const local;
         FailHard const failType;
-        bool applied;
+        bool applied = false;
         TER result;
 
         TransactionStatus(
--- a/src/ripple/app/misc/NetworkOPs.cpp
+++ b/src/ripple/app/misc/NetworkOPs.cpp
@@ -2414,8 +2414,8 @@ NetworkOPsImp::getTxsAccount(
     auto bound = [&ret, &app](
                      std::uint32_t ledger_index,
                      std::string const& status,
-                     Blob const& rawTxn,
-                     Blob const& rawMeta) {
+                     Blob&& rawTxn,
+                     Blob&& rawMeta) {
         convertBlobsToTxResult(ret, ledger_index, status, rawTxn, rawMeta, app);
     };
 
@@ -2453,8 +2453,8 @@ NetworkOPsImp::getTxsAccountB(
     auto bound = [&ret](
                      std::uint32_t ledgerIndex,
                      std::string const& status,
-                     Blob const& rawTxn,
-                     Blob const& rawMeta) {
+                     Blob&& rawTxn,
+                     Blob&& rawMeta) {
         ret.emplace_back(std::move(rawTxn), std::move(rawMeta), ledgerIndex);
     };
 
--- a/src/ripple/app/misc/impl/AccountTxPaging.cpp
+++ b/src/ripple/app/misc/impl/AccountTxPaging.cpp
@@ -65,11 +65,9 @@ accountTxPage(
     DatabaseCon& connection,
     AccountIDCache const& idCache,
     std::function<void(std::uint32_t)> const& onUnsavedLedger,
-    std::function<void(
-        std::uint32_t,
-        std::string const&,
-        Blob const&,
-        Blob const&)> const& onTransaction,
+    std::function<
+        void(std::uint32_t, std::string const&, Blob&&, Blob&&)> const&
+        onTransaction,
     AccountID const& account,
     std::int32_t minLedger,
     std::int32_t maxLedger,
@@ -247,11 +245,21 @@ accountTxPage(
                 if (rawMeta.size() == 0)
                     onUnsavedLedger(ledgerSeq.value_or(0));
 
+                // `rawData` and `rawMeta` will be used after they are moved.
+                // That's OK.
                 onTransaction(
                     rangeCheckedCast<std::uint32_t>(ledgerSeq.value_or(0)),
                     *status,
-                    rawData,
-                    rawMeta);
+                    std::move(rawData),
+                    std::move(rawMeta));
+                // Note some callbacks will move the data, some will not. Clear
+                // them so code doesn't depend on if the data was actually moved
+                // or not. The code will be more efficient if `rawData` and
+                // `rawMeta` don't have to allocate in `convert`, so don't
+                // refactor my moving these variables into loop scope.
+                rawData.clear();
+                rawMeta.clear();
+
                 --numberOfResults;
             }
         }
--- a/src/ripple/app/misc/impl/TxQ.cpp
+++ b/src/ripple/app/misc/impl/TxQ.cpp
@@ -1552,7 +1552,7 @@ setup_TxQ(Config const& config)
 std::unique_ptr<TxQ>
 make_TxQ(TxQ::Setup const& setup, beast::Journal j)
 {
-    return std::make_unique<TxQ>(setup, std::move(j));
+    return std::make_unique<TxQ>(setup, j);
 }
 
 }  // namespace ripple
--- a/src/ripple/app/misc/impl/TxQ.cpp
+++ b/src/ripple/app/misc/impl/TxQ.cpp
@@ -1549,10 +1549,4 @@ setup_TxQ(Config const& config)
     return setup;
 }
 
-std::unique_ptr<TxQ>
-make_TxQ(TxQ::Setup const& setup, beast::Journal j)
-{
-    return std::make_unique<TxQ>(setup, j);
-}
-
 }  // namespace ripple
--- a/src/ripple/app/ledger/impl/LedgerReplay.cpp
+++ b/src/ripple/app/ledger/impl/LedgerReplay.cpp
@@ -29,7 +29,7 @@ LedgerReplay::LedgerReplay(
 {
     for (auto const& item : replay_->txMap())
     {
-        auto const txPair = replay_->txRead(item.key());
+        auto txPair = replay_->txRead(item.key());  // non-const so can be moved
         auto const txIndex = (*txPair.second)[sfTransactionIndex];
         orderedTxns_.emplace(txIndex, std::move(txPair.first));
     }
--- a/src/ripple/app/misc/FeeVoteImpl.cpp
+++ b/src/ripple/app/misc/FeeVoteImpl.cpp
@@ -245,7 +245,7 @@ FeeVoteImpl::doVoting(
 
         auto tItem = std::make_shared<SHAMapItem>(txID, s.peekData());
 
-        if (!initialPosition->addGiveItem(tItem, true, false))
+        if (!initialPosition->addGiveItem(std::move(tItem), true, false))
         {
             JLOG(journal_.warn()) << "Ledger already had fee change";
         }
--- a/src/ripple/app/misc/TxQ.h
+++ b/src/ripple/app/misc/TxQ.h
@@ -774,12 +774,6 @@ private:
 TxQ::Setup
 setup_TxQ(Config const&);
 
-/**
-    @ref TxQ object factory.
-*/
-std::unique_ptr<TxQ>
-make_TxQ(TxQ::Setup const&, beast::Journal);
-
 template <class T>
 std::pair<bool, XRPAmount>
 toDrops(FeeLevel<T> const& level, XRPAmount const& baseFee)
--- a/src/ripple/app/misc/impl/AccountTxPaging.h
+++ b/src/ripple/app/misc/impl/AccountTxPaging.h
@@ -47,11 +47,9 @@ accountTxPage(
     DatabaseCon& connection,
     AccountIDCache const& idCache,
     std::function<void(std::uint32_t)> const& onUnsavedLedger,
-    std::function<void(
-        std::uint32_t,
-        std::string const&,
-        Blob const&,
-        Blob const&)> const& onTransaction,
+    std::function<
+        void(std::uint32_t, std::string const&, Blob&&, Blob&&)> const&
+        onTransaction,
     AccountID const& account,
     std::int32_t minLedger,
     std::int32_t maxLedger,
--- a/src/ripple/app/misc/impl/ValidatorKeys.cpp
+++ b/src/ripple/app/misc/impl/ValidatorKeys.cpp
@@ -39,7 +39,8 @@ ValidatorKeys::ValidatorKeys(Config const& config, beast::Journal j)
 
     if (config.exists(SECTION_VALIDATOR_TOKEN))
     {
-        if (auto const token = loadValidatorToken(
+        // token is non-const so it can be moved from
+        if (auto token = loadValidatorToken(
                 config.section(SECTION_VALIDATOR_TOKEN).lines()))
         {
             auto const pk =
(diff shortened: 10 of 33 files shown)


src/ripple/app/main/GRPCServer.cpp
```diff
@@ -55,7 +55,7 @@ GRPCServerImpl::CallData<Request, Response>::CallData(
     , responder_(&ctx_)
     , bindListener_(std::move(bindListener))
     , handler_(std::move(handler))
-    , requiredCondition_(std::move(requiredCondition))
+    , requiredCondition_(requiredCondition)
```
HowardHinnant, line 58: Intended for wider team review:

I'm uncomfortable removing a `std::move` in a context like this because it will cause future readers to wonder if the missing `move` on this one argument is perhaps an oversight and should be added.  Future readers will continually have to look up the underlying type to discover that a `move` on it will neither help nor hurt.

Move semantics was designed from day 1 to interoperate with "copy only" types with no negative consequences so that programmers would not have to worry about this issue so much.  Especially in generic code, but more generally, even in situations like this.

http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2002/n1377.htm#Cast%20to%20rvalue

> The "request to move" semantics turn out to be very handy in generic code. One can request that a type move itself without having to know whether or not the type is really movable. If the type is movable it will move, else if the type is copyable, it will copy, else you will get a compile-time error.

scottschurr: I'm not sure there's a good universal answer for this.  I agree that uses of `std::move` in copy-only situations are benign.  I think there are two additional considerations however.

1. Does the presence of the `std::move` cause an otherwise useful static checker to produce enough noise that the static checker becomes less useful?  [E.g., it generates`std::move has no effect`]. We have to think about keeping our tools useful.

2. It is possible for there to be a situation where if a `move` took place there would be a use after `move`, but that `move` doesn't currently happen because the type is only copyable.  (That's not the situation right here.)  So leaving that `move` in place would become a bug if the type actually becomes `move`able in the future.

I think consideration 1 is the most important.  If consideration 1 can be worked around then tools and code reviews will _probably_ manage issue 2 for us?  But consideration 2 would add risk to making a type `move`able if it hasn't been before.

HowardHinnant: Scott D tells me that this warning can be disabled, without disabling a warning about use after move, or a warning about moves that cause the code to become more expensive.

I agree that case 2 is a bug that should be fixed (except when that use has no preconditions, e.g. assign-to).  And I believe that the tools will catch that.

Another aspect that has come up between Scott and I in private conversations is that this warning is flagging cases where the client signals his intention that he no longer needs the value with move.  But the `std::move` is only a _request_ to move, not a requirement.  It is up to the far away client code whether to actually pilfer the object or not.  And that can change with the implementation of the far away code, without needing to survey clients.  Clients silently get the optimization when the far away code changes, with no semantic or syntax change to the client code.  Example:

https://github.com/ripple/rippled/pull/3348/commits/06740e6759eeeb2e08bedcb4f226fab84b95d9e7#diff-9647085a1d5d0d2cf7b0b835b0e8c742R147

In this case the client is saying I don't need the value of `ledger` any more, but the `RCLValidatedLedger` constructor doesn't move from it.  But in the `RCLValidatedLedger` might change to move from it.  If that happens, the client gets the optimization automatically and silently.  And in the mean time, the `move` in the client code causes no harm, and signals to the reader the intent:  last use.

nbougalis: I'm a bit torn here. On the one hand, I understand that `std::move` is just a request, and programmers can't expect it to _actually_ mean that the given instance is being moved from. On the other hand, specifying it, especially in a non-generic-programming context, when you know it's **not** applicable only serves to confuse.

Either way, a future programmer has to go looking at the code to attempt to decipher what's going on and if this is a case where the move was missed or explicitly omitted. In that context, I wonder if it makes sense to use a marker similar to `std::move` that _expressly_ indicates that the programmer expliictly chose to not use `std::move`?

seelabs: In the vast majority of cases `move` is used for two things: 1) as an optimization technique; 2) To pass ownership of unique resources.

Code that's missing `move` (that compiles) is almost never a bug. I don't think anyone is going to look at code where a statement has a missing `move` and wonder if there's a bug (although they may wondering if there's a missed optimization opportunity).

Up to now, we haven't put `move` in code that didn't have a benefit (at least intentionally). If we see a `move`, we can infer that the type would benefit from a move. Seeing `move(requiredCondition)`, gives a false signal about the type of `requiredCondition`. I would not have guessed that someone would put a `move` around an enum (of course, that's a circular argument - it sends a signal about the type because that's how we've used it; you are proposing to change that). The other "cost" of putting in a unneeded `move` is readability suffers. `foo(std::move(a), std::move(b), std::move(c))` is more cluttered `foo(a, b, c)`. If we put a move, I'd like it to have a potential benefit more than the surrounding code uses `move`.

The `move` in the `RCLValidatedLedger(ledger)` that you pointed out is a different case. It would not send the wrong signal about the `ledger` type, and there is a realistic chance that a future change would take advantage of the move.

My vote is I don't want the move for the enum. I'm _weakly_ against the move for `RCLValidatedLedger`, but only weakly (I'd prefer moves only where it makes a difference); I do see legitimate advantages of putting it there.

Of course, Howard has thought about move semantics move deeply that any of us. I think I'll add the `move` back in to the `RCLValidatedLedger` call - Howard's vote should count move than my "weakly against".

I'll also don't think putting a `move` around an enum is _that_ big a deal. The "costs" I've mentioned are all minor. While I'm more strongly against that change, I'm OK doing it if you (Howard) feel strongly.

Note: I'm not worried about making the static analyzer happy. It would be a large task and I suspect the code would be worse for the effort.

cjcobb23: >Move semantics was designed from day 1 to interoperate with "copy only" types with no negative consequences so 
> that programmers would not have to worry about this issue so much.

This is generally how I feel. I see the use of `std::move` as a request to move, written as an optimization. If the move doesn't happen, its not usually a bug or a problem, unless I really, really need that optimization. For this reason, I am hesitant to remove a `move`.

However, I agree that a `move` around an enum, or any type that clearly has no heap storage (int, bool, etc), is confusing to the reader; an actual move could never happen, so using `move` makes it seem like the programmer made a mistake, makes the intentions of the code unclear and might cause the reader to question whether they know what is going on. However, I think this situation is somewhat of an exception; the other arguments to the constructor are being moved from, so the intention seems clear to me: "Move each argument, if possible". I'm fine if you want to remove the `move` here, though I don't think its necessary to do so.

However, I am confused why the move is being removed from here: https://github.com/ripple/rippled/commit/06740e6759eeeb2e08bedcb4f226fab84b95d9e7#diff-3f6df86ad506fd7e45db86b186e102e2R2273 . 
This seems like a perfect use case for `move`. Why is this being removed? Is the move not actually happening? If so, instead of removing the `std::move`, we should figure out how to make the move actually happen.

seelabs: @cjcobb23 Re: moving `Journal`: An explicit goal for is to be light weight enough to be copied by value (see this comment: https://github.com/ripple/rippled/blob/develop/src/ripple/beast/utility/Journal.h#L52) The class has a single member - a pointer to a `Sink`. It wouldn't benefit from a move, (and if we change `Journal` so it would be cheaper to move than copy we'd have to change a lot of code). We don't move other `Journals` either.

cjcobb23: @seelabs Sorry, maybe my link isn't working correctly. I was referring to NetworkOps.cpp:2273, where we are removing two `std::move`s that each wrap a `Blob`, when I asked why the `move` was being removed. I'll leave a comment on that line.

HowardHinnant: This is basically the motivation for why move semantics is designed the way it is.  It is too complicated for programmers to keep track of which types really benefit from a `move` and which don't.  Therefore the design is that you can `move` any type.  Whether it helps or not is an implementation detail of the type.  But it won't hurt.

So the only question the programmer need ask himself is:  Do I need the value of this type after this point?  If not `move` it.  If so, don't `move` it.  Same coding guideline no matter the type.

src/ripple/app/misc/NetworkOPs.cpp
```diff
@@ -2270,7 +2270,7 @@ NetworkOPsImp::getTxsAccountB(
                      std::string const& status,
                      Blob const& rawTxn,
                      Blob const& rawMeta) {
-        ret.emplace_back(std::move(rawTxn), std::move(rawMeta), ledgerIndex);
+        ret.emplace_back(rawTxn, rawMeta, ledgerIndex);
```
cjcobb23, line 2273: @seelabs Why are these `move`s being removed? A `Blob` seems to be a great candidate for `move`.

src/ripple/app/main/GRPCServer.cpp
seelabs: @cjcobb23 Ah, I was looking at the wrong diff, sorry. The blobs aren't moved because the moved-from values are const.

HowardHinnant: That being said (about what types to `move`), when a type is clearly marked `const` within the same scope that it is moved, then there is a clear misunderstanding and a clarification would make the code more readable.

But is the `const` wrong or is the `move` wrong?  I haven't studied the code enough to know which.

cjcobb23: @seelabs I don't think the moved-from values need to be const. Unless I am mistaken, the values are local to the function, are overwritten in each iteration of the loop, and are not read from between the move and being overwritten.

cjcobb23: Also, about moving the `ledger` parameter of the `RCLValidatedLedger` constructor, I think that `move` should remain as well. I think across the board, in my opinion, when we discover a `move` statement that does not actually move, but a move could occur if we implemented the required logic, we should implement the required logic or leave a TODO saying something like "TODO: This does not actually move. Achieving a move here requires implementing X". As opposed to removing the `move` statement.

seelabs: @cjcobb23 For the `NetworkOps.cpp:2273` case, we might be able to do that, I'll take a look. 

But note that optimizing for the case on line 2273 may make other calls run more slowly (depending on the internals of soci). We'd have to change the parameter in the lambda from a `const&` to a value type and move from the `accountsTxPage` function. No big deal, but note not all callers will move from these parameters. For example, the callback on line 2234.

Whether this will hurt calls that don't move from the parameters depends on how `soci` deals with reusing the vectors. If it can reuse the space without allocating, then this may be a net loss - we'll move into the callback, the callback won't move variable, and soci may reallocate when it didn't need to before (depending on the internals of soci - I haven't looked).

For the `RCLValidatedLedger` constructor case the `Ledger` parameter is never saved within that object so there's no reason to move it - so there's nothing to change to make the move work. However, if that ever does change so the ledger is save we wouldn't need to change the callers is we keep the (currently) unneeded move.

seelabs: @HowardHinnant I re-added the two moves you flagged (See: https://github.com/ripple/rippled/pull/3348/commits/d3684aa8668eb84c041684e33a22a19ad9d2df85). I'm still against making this the new style, but I don't want to get too hung on on these two cases and it's hard not to weigh your opinions heavily on a topic like `move`.

cjcobb23: @seelabs Maybe I am misunderstanding how `move` works, but can't we do something like this?: https://github.com/cjcobb23/rippled/commit/92a60d55781e6febafb2bd237271e92e6c3b8463

Based on my understanding of move semantics, the actual move is only going to happen in the callback at [NetworkOPs.cpp:2273](https://github.com/cjcobb23/rippled/commit/92a60d55781e6febafb2bd237271e92e6c3b8463#diff-3f6df86ad506fd7e45db86b186e102e2R2273); specifically, instead of copying the `Blob`s into the vector, the data will be moved. For the callback that doesn't move the parameters ( [NetworkOPs.cpp:2234](https://github.com/cjcobb23/rippled/commit/92a60d55781e6febafb2bd237271e92e6c3b8463#diff-3f6df86ad506fd7e45db86b186e102e2R2234)), no actual move will occur; the arguments will be passed by reference. So, the callback that does not move will not behave any differently. Feel free to correct me if I am wrong (@HowardHinnant ).

Re: reallocation within `soci`. The parameters used by `soci` are actually the `txnData` and `txnMeta` variables, which have type `soci::blob`( [link](https://github.com/cjcobb23/rippled/commit/92a60d55781e6febafb2bd237271e92e6c3b8463#diff-19022b996a8937da021845d984070f15R202)). We then, call the `convert` function (defined [here](https://github.com/cjcobb23/rippled/blob/92a60d55781e6febafb2bd237271e92e6c3b8463/src/ripple/core/impl/SociDB.cpp#L150)), to copy data from a `soci::blob` into a `Blob`. This `convert` function first performs a `resize` on the `Blob`, and then copies the data from `soci::blob` into `Blob`. 
The move I am proposing would then move those `Blob`s into the vector (which is captured by reference in the lambdas), instead of copying into the vector.

Let's consider each case. First, without the move: The `Blob` objects are being reused, which means that the `resize` in `convert` might not perform a reallocation; the object will already own some heap storage from the previous iteration of the loop. Then, a copy is performed (inside `convert`). Then, the `Blob` is copied into the result vector, resulting in a copy as well as an allocation; the allocation occurs during the copy constructor of `Blob`. In summary, there is a possible but unlikely reallocation, followed by a copy, another copy and an allocation.

Now, with the move: The `Blob` objects that are being reused will not own any heap storage,  since at the end of each iteration of the loop, the `Blob`s will be moved from. So, the `resize` in `convert` will always perform an allocation. Then, a copy is performed (inside `convert`). Then, `Blob` is moved into the result vector, meaning no copy and no allocation.

Comparing the two:
Without move: possible but unlikely allocation, copy, copy, allocation
With move: allocation, copy
To keep things simple, we can ignore the possible but unlikely allocation in the first case. In this way, the two approaches are identical except for an additional copy if the move is not used.

Sorry in advance if I am mistaken and this is all wrong. I would really like the move to occur in this situation, since the account_tx call is rather expensive, and it would be nice to increase the performance.

seelabs: @cjcobb23 Very nice! I did not consider passing rvalue references to the callback (I almost always pass sink parameters as values - that way callers can choose to move or copy at their discretion). In this case there's only one caller, and we know we can move there. Love it! (And smacking myself on the head for not thinking of it myself).

seelabs: @cjcobb23 Just a note: For the case of the callback on NetworkOPs.cpp:2234, if we did pass by value, that change would be pessimizing for that case (but of course, not for pass by rvalue ref). So we really wouldn't want to do this if you didn't suggest passing by rvalue ref.

seelabs: @cjcobb23 fixed in `334821850 [fold] Use rvalue refs on `accountTxPage` callbacks
`

src/ripple/app/misc/NetworkOPs.cpp
seelabs: (I know you know this, but leaving a comment for future searches)
The parameters were `const`, so they won't be moved from.

src/ripple/app/ledger/Ledger.cpp
```diff
@@ -527,7 +527,7 @@ Ledger::rawInsert(std::shared_ptr<SLE> const& sle)
     auto item = std::make_shared<
         SHAMapItem const>(sle->key(),
             std::move(ss));
-    // VFALCO NOTE addGiveItem should take ownership
+    // addGiveItem should take ownership
```
nbougalis, line 530: This comment makes no sense to me. It _should_ but _does it_? A better comment might be:

    // addGiveItem takes ownership

But maybe that isn't even needed, given the `std::move`. It may be clearer still to cut out the `std::move` and eliminate the `item` temporary entirely by inlining the call into `addGiveItem` which makes it extra obvious that the pointer is being handed off to the map for safe-keeping.

src/ripple/app/ledger/Ledger.cpp
```diff
@@ -541,7 +541,7 @@ Ledger::rawReplace(std::shared_ptr<SLE> const& sle)
     auto item = std::make_shared<
         SHAMapItem const>(sle->key(),
             std::move(ss));
-    // VFALCO NOTE updateGiveItem should take ownership
+    // updateGiveItem should take ownership
```
nbougalis, line 544: Same comment as above.

src/ripple/app/ledger/Ledger.cpp
```diff
@@ -527,7 +527,7 @@ Ledger::rawInsert(std::shared_ptr<SLE> const& sle)
     auto item = std::make_shared<
         SHAMapItem const>(sle->key(),
             std::move(ss));
```
nbougalis, line 529: So `Serializer` doesn't have a move constructor and this is a prime use case for having one. @HowardHinnant, should we add one to it (along with a move assignment, I guess), even if it's just `= default`.

src/ripple/app/misc/NetworkOPs.cpp
```diff
@@ -82,7 +82,7 @@ class NetworkOPsImp final
         bool const admin;
         bool const local;
         FailHard const failType;
-        bool applied;
+        bool applied{false};
```
nbougalis, line 85: This uniform initialization stuff makes my eyes bleed. It's just so jarring. Unless there's some really cool, esoteric advantage to this, can we just use `bool applied = false` please?

src/ripple/app/misc/impl/AccountTxPaging.cpp
```diff
@@ -68,8 +68,8 @@ accountTxPage(
     std::function<void(
         std::uint32_t,
         std::string const&,
-        Blob const&,
-        Blob const&)> const& onTransaction,
+        Blob&&,
+        Blob&&)> const& onTransaction,
```
nbougalis, line 72: No action needed; just admiring the `accountTxPage` signature... 12 arguments, two of which are callbacks. :grimacing:

src/ripple/app/misc/impl/TxQ.cpp
```diff
@@ -1599,7 +1599,7 @@ setup_TxQ(Config const& config)
 std::unique_ptr<TxQ>
 make_TxQ(TxQ::Setup const& setup, beast::Journal j)
 {
-    return std::make_unique<TxQ>(setup, std::move(j));
+    return std::make_unique<TxQ>(setup, j);
```
nbougalis, line 1602: Since you're here: this `make_TxQ` function is entirely unnecessary; we usually use this idiot when we have an interface class and the implementation is hidden and inaccessible. This isn't the case here, so it could go away by directly inlining the `std::make_unique<TxQ>(setup, j)` in **Application.cpp**

src/ripple/app/main/Application.cpp
```diff
@@ -1726,7 +1727,7 @@ int ApplicationImp::fdRequired() const
     int needed = 128;
 
     // 1.5 times the configured peer limit for peer connections:
-    needed += static_cast<int>(0.5 + (1.5 * overlay_->limit()));
+    needed += lround(1.5 * overlay_->limit());
```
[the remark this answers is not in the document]
nbougalis: :point_up:

src/ripple/app/ledger/Ledger.cpp
seelabs: @nbougalis I believe the compiler will supply a move constructor for `Serializer`

src/ripple/app/main/Application.cpp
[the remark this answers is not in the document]
seelabs: Fixed in `bd4cc1711 [fold] Misc fixes:`

src/ripple/app/ledger/Ledger.cpp
seelabs: Fixed in `bd4cc1711 [fold] Misc fixes:`

src/ripple/app/ledger/Ledger.cpp
seelabs: Fixed in `bd4cc1711 [fold] Misc fixes:`

src/ripple/app/misc/NetworkOPs.cpp
seelabs: Fixed in `bd4cc1711 [fold] Misc fixes:`

src/ripple/app/misc/impl/TxQ.cpp
seelabs: Fixed in `5087498fd [fold] Rm make_TxQ factory function`

Lol'd at the mistype of `idiot` for `idiom`

src/ripple/app/main/GRPCServer.cpp
thejohnfreeman: I feel like the only person who doesn't find `std::move` around a small value type confusing. I'd be more confused by seeing a list of constructor initializers where some but not all arguments are moved.

I think Howard has already said everything I feel:

- A `std::move` that doesn't move doesn't hurt anything, and could help if, one day, it starts to move.
- A `std::move` serves as a good indicator of "I'm done with this value", even if it doesn't move.
- A `std::move` is not a guarantee of a move.

In my opinion, any confusion surrounding an appearance of `std::move` is likely caused by the reader's misinterpretation of `std::move` as an expression of "this value will be moved" instead of "I'm done with this value, and don't care if it is moved or not". Perhaps we could define an alias like `give`.

HowardHinnant:
Here, https://github.com/ripple/rippled/blob/develop/src/ripple/app/misc/FeeVoteImpl.cpp#L254 , `tItem` should now be moved.

HowardHinnant:
Here, https://github.com/ripple/rippled/blob/develop/src/ripple/shamap/impl/SHAMapTreeNode.cpp#L99 , `item` should now be moved (in 12 similar places).

seelabs:
@HowardHinnant I can't comment on top level comments, but all of those top level comments should be fixed in `4e58b1edc [fold] Add some missing `std::move`s`

BTW, I know this needs to be rebased. I expect there will be a new beta tomorrow (the re-formatting beta), so I'll rebase after that beta gets merged.

seelabs:
Squashed and rebased onto b3

seelabs:
rebased onto b4

src/ripple/app/misc/impl/TxQ.cpp
nbougalis: LOL... I just noticed that typo. :facepalm:

seelabs:
rebased on b5

PR #3351, opened 2020-04-13
[TRIVIAL] Improve reporting of missing node exceptions

nbougalis:


src/ripple/shamap/SHAMapMissingNode.h
```diff
@@ -34,33 +36,44 @@ enum class SHAMapType
...
-    SHAMapType mType;
-    SHAMapHash mNodeHash;
-    uint256    mNodeID;
 public:
-    SHAMapMissingNode (SHAMapType t,
-                       SHAMapHash const& nodeHash)
-        : std::runtime_error ("SHAMapMissingNode")
-        , mType (t)
-        , mNodeHash (nodeHash)
+    SHAMapMissingNode (SHAMapType t, SHAMapHash const& hash)
```
seelabs, line 56: What do you think of adding the stack trace here when in debug mode?

src/ripple/app/consensus/RCLConsensus.cpp
```diff
@@ -830,7 +830,7 @@ RCLConsensus::timerEntry(NetClock::time_point const& now)
     catch (SHAMapMissingNode const& mn)
     {
         // This should never happen
-        JLOG(j_.error()) << "Missing node during consensus process " << mn;
+        JLOG(j_.error()) << "During consensus process: " << mn.what();
```
miguelportilla, line 833: To determine which function caught, consider adding the function name to the log.
eg.
```JLOG(j_.error()) << "During consensus process: " << mn.what() << ". Caught in " << __func__;```

File: src/ripple/app/consensus/RCLConsensus.cpp [hunk 3]
```diff
@@ -846,7 +846,7 @@ RCLConsensus::gotTxSet(NetClock::time_point const& now, RCLTxSet const& txSet)
...
```
miguelportilla, line 849: Consider adding the function name.
eg.
```JLOG(j_.error()) << "During consensus process: " << mn.what() << ". Caught in " << __func__;```

File: src/ripple/shamap/SHAMapMissingNode.h [hunk 4]
```diff
@@ -34,33 +36,44 @@ enum class SHAMapType
...
+to_string(SHAMapType t)
+{
+    if (t == SHAMapType::TRANSACTION)
+        return "Transaction Tree";
+    if (t == SHAMapType::STATE)
+        return "State Tree";
+    if (t == SHAMapType::FREE)
+        return "Free Tree";
+    return std::to_string(safe_cast<std::underlying_type_t<SHAMapType>>(t));
+}
```
miguelportilla, line 50: Consider a switch statement instead.
eg.
```
switch (t)
{
case SHAMapType::TRANSACTION:
        return "Transaction Tree";
case SHAMapType::STATE:
        return "State Tree";
case SHAMapType::FREE:
        return "Free Tree";
default:
    return std::to_string(safe_cast<std::underlying_type_t<SHAMapType>>(t));
}
```

src/ripple/shamap/SHAMapMissingNode.h
nbougalis: I thought about doing this in `LogThrow` actually.

src/ripple/app/consensus/RCLConsensus.cpp
nbougalis: I can't wait for C++20 and [`std::source_location`](https://en.cppreference.com/w/cpp/utility/source_location). :)

thejohnfreeman: Don't need the `.what()` (here and below) if you're still defining the stream operator (it still used in other places in the code).

src/ripple/shamap/SHAMapMissingNode.h
HowardHinnant: On Mac and Linux, the existing terminate handler creates a very good stack dump.  I think we used to have a custom terminate handlers and got rid of them in order to _improve_ our stack dumps.

src/ripple/app/consensus/RCLConsensus.cpp
HowardHinnant: Personally I wouldn't dare to write such a proposal until I had some field experience with it.

scottschurr: I can hear the cries of ABI compatibility already.  A change like that would need to be more carefully crafted than I know how to do.

src/ripple/shamap/SHAMapMissingNode.h
nbougalis: The question is *where* do we get the usable stack dumps from? I know on Mac you can get them in the log, but what about Linux?

src/ripple/app/consensus/RCLConsensus.cpp
nbougalis: I'm thinking of just removing the `<<` and printing `what()` instead. Not sure if that's better.

nbougalis: Appreciate both of your feedback. :vulcan_salute:

src/ripple/shamap/SHAMapMissingNode.h
HowardHinnant: As not-a-Linux-user, I don't know.  I thought we had that knowledge within the rippled team though.  We used to get them mailed to us whenever a production server went down (hasn't happened in awhile).

Perhaps I'm misunderstanding this thread...

src/ripple/app/consensus/RCLConsensus.cpp
thejohnfreeman: I look at the stream `operator<<` as the C++ version of `__str__` in Python or `Show` in Haskell, generally considered good practice, so I like it.

nbougalis: Hmmm. That seems reasonable.

src/ripple/shamap/SHAMapMissingNode.h
nbougalis: That was caused by a script we were running on production servers ran by Ripple that would detect the crash and automatically ran `gdb` on the resulting core file. The question is what happens if there is no core file? On the Mac we do get a very pretty call stack in the logs. We'll have to research what happens on Linux.

PR #3355, opened 2020-04-14
Automate project management 

carlhua:
This PR if merged allows the automation of project management to move card between "In Review" and "Approved for Merging" when the "Passed" label is added/removed

---
This change is [Reviewable](https://reviewable.io/reviews/ripple/rippled/3355)

### DIFF
--- a/.github/workflows/pr_automation.yml
+++ b/.github/workflows/pr_automation.yml
@@ -1,0 +1,61 @@
+name: PR automation
+on:
+  pull_request:
+    types: [labeled, unlabeled]
+
+jobs:
+  job:
+    runs-on: ubuntu-latest
+    steps: 
+      - name: retrieve label
+        run: |
+          jq '.' $GITHUB_EVENT_PATH
+          LABEL_NAME1=$(jq --raw-output ".label.name" $GITHUB_EVENT_PATH)
+          echo "Label in action is $LABEL_NAME1"
+          echo ::set-env name=LABEL_NAME::$LABEL_NAME1
+      #inspect the PR to get the project name
+      - name: retrieve project name
+        uses: octokit/graphql-action@v2.x
+        id: get_project_name
+        with:
+          query: |
+            query  get_project_name($owner:String!, $repo:String!, $number:Int!) {
+              repository(owner:$owner,name:$repo) {
+                  pullRequest(number: $number) {
+                    projectCards(first: 1) {
+                      nodes {
+                        project {
+                          name
+                        }
+                      }
+                    }
+                  }
+                }
+              }
+          owner: ${{ github.event.repository.owner.login }}
+          repo: ${{ github.event.repository.name }}
+          number: ${{ github.event.number }}
+        env:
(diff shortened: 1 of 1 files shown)


nbougalis:
Need to be squashed down into a single commit prior to merging.

.github/workflows/pr_automation.yml
```diff
@@ -0,0 +1,61 @@
...
+      - name: perform automation for label
+        if: env.LABEL_NAME == 'Passed' && steps.extract.outputs.name != '' && github.event.action == 'labeled'
+        uses: alex-page/github-project-automation-plus@v0.2.2
+        with:
+          project: ${{ steps.extract.outputs.name }}
+          #if we change the column name, this automation needs to be updated
+          column: 'Approved for Merging'
+          repo-token: ${{ secrets.PR_TOKEN }}
+      - name: perform automation for unlabel
+        if: env.LABEL_NAME == 'Passed' && steps.extract.outputs.name != '' &&  github.event.action == 'unlabeled'
```
thejohnfreeman, line 56: It doesn't seem possible to cancel a job through any means other than failing a step, but you can get the same effect by setting an `if` on every step after which the condition is known. So maybe set `if: env.LABEL_NAME == 'Passed'` on _every_ step after the first one (which sets `env.LABEL_NAME`).

carlhua: I actually thought about doing that but decided not to because in the future I plan to use this workflow to further automate other labels so I think we should keep them as is.

carlhua:
> Need to be squashed down into a single commit prior to merging.

Updated and squashed

.github/workflows/pr_automation.yml
```diff
@@ -0,0 +1,61 @@
...
+    runs-on: ubuntu-latest
+    steps: 
+      - name: retrieve label
+        run: |
+          jq '.' $GITHUB_EVENT_PATH
+          LABEL_NAME1=$(jq --raw-output ".label.name" $GITHUB_EVENT_PATH)
+          echo "Label in action is $LABEL_NAME1"
+          echo ::set-env name=LABEL_NAME::$LABEL_NAME1
+      #inspect the PR to get the project name
+      - name: retrive project name
```
thejohnfreeman, line 17: Misspelling here.

carlhua: updated.

.github/workflows/pr_automation.yml
```diff
@@ -0,0 +1,61 @@
...
+                    projectCards(first: 1) {
+                      nodes {
+                        project {
+                          name
+                        }
+                      }
+                    }
+                  }
+                }
+              }
```
cjcobb23, line 34: Is this formatting necessary? It kind've looks ugly. Can we put some things on the same line? Like `nodes { project { name } }`. We don't want one really long line, but this looks a bit excessive. Just a style concern, non-blocking.

carlhua:
It is a false positive that the automation failed. This automation uses a token which is not passed to a PR that is submitted via a fork - so failure is expected. Once merged, it will have access to the token. see the screenshot.
screenshot_28

PR #3360, opened 2020-04-15
Adjust timing of Validator Site tests

ximinez:
* Adjust timeouts in Validator Site tests to prevent spurious failures due to DNS resolution.
* Close UNL download socket before result is passed to the callback in `ValidatorSite`.

---
This change is [Reviewable](https://reviewable.io/reviews/ripple/rippled/3360)


seelabs:
It's easy to see that this patch does what it advertises to do. However, I'd like to get confirmation that this actually fixes the issue before we approve it.

thejohnfreeman:
I tested with this command:

```
while true; do ./test_timeout --unittest ripple.app.ValidatorSite --unittest-jobs $(nproc) 2>&1 | tee -a logs; if [ $? -eq 0 ]; then echo good; else echo bad; fi; done
```

Left it running overnight. 1939 suite passes, 0 failures.

src/test/app/ValidatorSite_test.cpp
```diff
@@ -197,8 +197,12 @@ class ValidatorSite_test : public beast::unit_test::suite
             emptyLocalKey, emptyCfgKeys, cfgPublishers));
 
         using namespace std::chrono_literals;
+        // Normally, tests will almost never need all this time,
+        // but sometimes DNS resolution takes an inordinate amount
+        // of time. This appears to be an OS-level issue, so the
+        // test will just wait.
```
seelabs, line 203: I'd probably leave out the "this appears to be an OS-level issue". We never really dug into this, we just know that it take a long time.

seelabs:
We might want to remove the `trivial` tag from this. Changing the timeout is trivial, chaning `WorkBase` is probably over the `trivial` threashold.

src/test/app/ValidatorSite_test.cpp
ximinez: Good point. Will change.

ximinez:
Squashed and ready for merge. I think it makes sense to leave these two commits separated when merging.

Issue #2822, opened 2019-01-09
systemd - start rippled after network.target


jnr101:
Hi,

I have my rippled server running on a VM in a cloud and noticed that rippled failed to start up automatically after reboot. It turns out that the network is not available yet when the rippled server is being started. A fix that works for me is by adjusting the `/lib/systemd/system/rippled.service`, adding the line `After=syslog.target network.target` in the `[Unit]` section as follows:

```
[Unit]
Description=Ripple Daemon
After=syslog.target network.target
 
[Service]
Type=simple
ExecStart=/opt/ripple/bin/rippled --net --silent --conf /etc/opt/ripple/rippled.cfg
# Default KillSignal can be used if/when rippled handles SIGTERM
KillSignal=SIGINT
Restart=no
User=rippled
Group=rippled
 
[Install]
WantedBy=multi-user.target
```

MarkusTeufelberger:
Why do you start the server silenced instead of logging to stdout?

jnr101:
I can't remember to have changed it after installation, except for the 'After' line .. So it must have been in there from the start

jnr101:
ok so there it was. Couldn't find it :-) I will do that and close this one. Thanks

PR #3361, opened 2020-04-16
Remove the built-in "sustain" watchdog:

nbougalis:
The built-in watchdog is simplistic and can, sometimes, cause problems especially on systems that have the ability to automatically start and monitor processes.

This commit removes the sustain system entirely, changes the handling of the `SIGTERM` signal to properly terminate the process and improves the error message reported to the user when the command line used to start `rippled` is incorrect and malformed.


MarkusTeufelberger:
Thanks a lot, this makes it much easier to monitor `rippled` in production as one can actually see if a server crash-loops without having to query its uptime constantly via API calls! :-)

src/ripple/app/main/Main.cpp
```diff
@@ -403,6 +401,8 @@ int run (int argc, char** argv)
         "so this option is not needed for users")
     ("unittest-child",
         "For internal use only when spawning child unit test processes.")
+    ("fg",
+        "Depreceated: server always in foreground mode.")
```
ximinez, line 405: "Deprecated"
* https://www.dictionary.com/browse/deprecate
* https://en.wikipedia.org/wiki/Deprecation

nbougalis: D’oh! I know better than that.

nbougalis:
If merged, this PR also addresses #2822, which has been closed for a while.

Builds/containers/shared/rippled.service
```diff
@@ -1,12 +1,12 @@
...
+After=network-online.target
+Wants=network-online.target
 
 [Service]
 Type=simple
 ExecStart=/opt/ripple/bin/rippled --net --silent --conf /etc/opt/ripple/rippled.cfg
-# Default KillSignal can be used if/when rippled handles SIGTERM
-KillSignal=SIGINT
-Restart=no
+Restart=yes
```
carlhua, line 9: What is the intention of this line? Based on the documentation of systemd.service, it did not list =yes as an option. There are various configuration we can pass to Restart= depending on what we want. do we want to restart the service even if it has a clean exit code initiated by an operator? I used this link as my reference http://manpages.ubuntu.com/manpages/cosmic/man5/systemd.service.5.html

nbougalis: Yeah, sorry. Brain shortcircuit, this should be `on-failure`

src/ripple/app/main/Application.cpp
```diff
@@ -1296,10 +1276,33 @@ class ApplicationImp
 bool ApplicationImp::setup()
 {
     // We want to intercept and wait for CTRL-C to terminate the process
-    m_signals.add (SIGINT);
+    m_signals.add(SIGINT);
+    m_signals.add(SIGTERM);
+    m_signals.add(SIGHUP);
```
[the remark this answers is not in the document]
nbougalis: Ahh, good ole' Windows. I'll fix this.

nbougalis:
This PR has been cherry-picked into into #3326. It will be merged together with that PR.

PR #3362, opened 2020-04-16
[TRIVIAL] Fix a build issue involving Ubuntu Docker containers:

manojsdoshi:
Upgrade the build pipeline to use ubuntu 19.10 container
in order for the build pipeline to successfully complete builds.


seelabs:
Ubuntu 20.04 comes out next week, and that's a LTS version. Does it make sense to wait a week and use that instead?

manojsdoshi:
Yeh makes sense waiting for 20.04 especially since it has LTS.  

Although one concern I have is we don't have packages for 1.6.0-b2 since the package build is broken. If we are ok living with that, I think it's ok to wait for a week.

Issue #2688, opened 2018-09-18
Deterministic shard databases/files


MarkusTeufelberger:
As discussed in #2625 I'd like to propose a few changes to how shard databases are currently being built with the goal to have them created deterministically.

These changes would be:

1) Only allow NuDB for shard databases (at least the deterministic kind). RocksDB is too much of a moving target.
2) Values that are inserted into the shard database must be deterministically compressed (this could include not compressing them at all...)
3) salt, app_num, key_size, block_size, load_factor and the hashing algorithm must be fixed and not derived from properties of the current system (sane defaults would probably be an easily recognizable number + the current shard number as salt, 1 for the app_num, 32 bytes for the key size [as is currently standard anyways], 4 KiB for the block_size, 0.5 for the load_factor and xxhash for the hashing algorithm)
4) Deterministic shard database files can only be written once the last (highest) ledger sequence of that shard has been closed and only if all nodes of the 16384 ledgers they contain are available
5) To write such a database file, all nodes contained in the shard must be written single-threaded in sorted order (ascending or descending, whatever is easier to implement) to a NuDB file. I haven't checked yet if there are other caches or optimizations going on, but hopefully that should already be enough to guarantee a deterministic result. If there's a lot of RAM available, it might be possible to load the whole shard into memory and sort it there, otherwise it could mean hitting the node store twice (once for getting all the keys and after sorting them for the values again).

Once such a file has been generated (ideally also with a canonical name...), it then is possible to compare hashes with others, immediately help seeding torrents or IPFS swarms or use this knowledge to implement a trustless way of quickly getting a lot of history without hitting your peer's node stores or shard databases. If you want to get very fancy, it would even be possible to add IPFS or BTinfo hashes to the ledger itself, if validators are willing (and able) to download and verify their contents (e.g. by extending rippled into a historic-headers-only mode and then asserting that a shard database contains all nodes necessary to walk all SHAmaps in the specified range with an unbroken header chain up to the latest one).

An alternative to forcing determinism upon NuDB would be to not use a database file format at all and just specify a deterministic export/import format for shards. An example would be a CSV file containing sorted key-value pairs in simple hex or base64 encoding. If these are independently verifiable, it might be already enough. As far as I understand the shard import code though, it kinda expects being able to randomly query the import file like a database.

[...]

PR #3363, opened 2020-04-16
[#2688] Deterministic Database Shards

p2peer:
# Deterministic Database Shards

This doc describes the standard way to assemble the database shard. A shard assembled using this approach becomes deterministic i.e. if two independent sides assemble the shard consists of the same ledgers, accounts and transactions, then they will obtain the same shard files `nudb.dat` and `nudb.key`. The approach deals with the `NuDB` database format only, refer to `https://github.com/vinniefalco/NuDB`.

## Headers

Due to NuDB database definition, the following headers are using for database files:

nudb.key:
```
char[8]         Type            The characters "nudb.key"
uint16          Version         Holds the version number
uint64          UID             Unique ID generated on creation
uint64          Appnum          Application defined constant
uint16          KeySize         Key size in bytes
uint64          Salt            A random seed
uint64          Pepper          The salt hashed
uint16          BlockSize       Size of a file block in bytes
uint16          LoadFactor      Target fraction in 65536ths
uint8[56]       Reserved        Zeroes
uint8[]         Reserved        Zero-pad to block size
```

nudb.dat:
```
char[8]         Type            The characters "nudb.dat"
uint16          Version         Holds the version number
uint64          UID             Unique ID generated on creation
uint64          Appnum          Application defined constant
uint16          KeySize         Key size in bytes
uint8[64]       (reserved)      Zeroes
```
there all fields are saved using network byte order (most significant byte first).

To make the shard deterministic the following parameters are used as values of header field both for `nudb.key` and `nudb.dat` files.
```
Version         2
UID             digest(1)
Appnum          1
KeySize         32
Salt            digest(0)
Pepper          xxhasher(Salt)
BlockSize       0x1000
LoadFactor      0.5 (numeric 0x8000)
```
Note: xxhasher is native hasher of NuDB, refer to NuDB code.

The `digest(i)` mentioned above defined as the follows:

First, SHA256 hash `H` calculated of the following structure
```
uint256         lastHash        Hash of last ledger in shard
uint32          index           Index of the shard
uint32          firstSeq        Sequence number of first ledger in the shard
uint32          lastSeq         Sequence number of last ledger in the shard
uint32          version         Version of shard, 2 at the present
```

Then, `digest(i)` is defined as `i`-th 64-bit portion of the above hash `H` with swapped order of bytes, i.e.
```
digest(i) = H[i*8+0] << 56 | H[i*8+1] << 48 | ... | H[i*8+7] << 0.
```
where `H[i]` denotes `i`-th byte of hash `H`.

## Contents

After deterministic shard is created using the above mentioned headers, it filled with objects. First, all objects of the shard are collected and sorted in according to their hashes. Here the objects are: ledgers, SHAmap tree nodes including accounts and transactions, and final key object with hash 0. Objects are sorted by increasing of their hashes, precisely, by increasing of hex representations of hashes in lexicographic order. 

For example, the following is an example of sorted hashes in their hex representation:
```
0000000000000000000000000000000000000000000000000000000000000000
154F29A919B30F50443A241C466691B046677C923EE7905AB97A4DBE8A5C2423
2231553FC01D37A66C61BBEEACBB8C460994493E5659D118E19A8DDBB1444273
272DCBFD8E4D5D786CF11A5444B30FB35435933B5DE6C660AA46E68CF0F5C447
3C062FD9F0BCDCA31ACEBCD8E530D0BDAD1F1D1257B89C435616506A3EE6CB9E
58A0E5AE427CDDC1C7C06448E8C3E4BF718DE036D827881624B20465C3E1334F
...
```

Finally, objects added to the shard one by one in the sorted order from low to high hashes.

## Tests

To perform test to deterministic shards implementation one can enter the following command:
```
rippled --unittest ripple.NodeStore.DatabaseShard
```

The following is the right output of deterministic shards test:
```
ripple.NodeStore.DatabaseShard DatabaseShard deterministic_shard with backend nudb
Iteration 0: md5[nudb.key] = 59ABA6E282AB88C632CB8870BAC1B132
Iteration 0: md5[nudb.dat] = 36F6526D79DE4E01307615CDCFA279F6
Iteration 1: md5[nudb.key] = 59ABA6E282AB88C632CB8870BAC1B132
Iteration 1: md5[nudb.dat] = 36F6526D79DE4E01307615CDCFA279F6
Iteration 2: md5[nudb.key] = 59ABA6E282AB88C632CB8870BAC1B132
Iteration 2: md5[nudb.dat] = 36F6526D79DE4E01307615CDCFA279F6
Iteration 3: md5[nudb.key] = 59ABA6E282AB88C632CB8870BAC1B132
Iteration 3: md5[nudb.dat] = 36F6526D79DE4E01307615CDCFA279F6
```

### DIFF
--- a/src/ripple/nodestore/Backend.h
+++ b/src/ripple/nodestore/Backend.h
@@ -58,6 +58,25 @@ public:
     virtual void
     open(bool createIfMissing = true) = 0;
 
+    /** Open the backend.
+        @param createIfMissing Create the database files if necessary.
+        @param appType Deterministic appType used to create a backend.
+        @param uid Deterministic uid used to create a backend.
+        @param salt Deterministic salt used to create a backend.
+        This allows the caller to catch exceptions.
+    */
+    virtual void
+    open(
+        bool createIfMissing,
+        boost::optional<uint64_t> appType,
+        boost::optional<uint64_t> uid,
+        boost::optional<uint64_t> salt)
+    {
+        Throw<std::runtime_error>(std::string(
+            "Deterministic appType/uid/salt not supported by backend " +
+            getName()));
+    }
+
     /** Close the backend.
         This allows the caller to catch exceptions.
     */
--- a/src/ripple/nodestore/DeterministicShard.md
+++ b/src/ripple/nodestore/DeterministicShard.md
@@ -1,0 +1,103 @@
+# Deterministic Database Shards
+
+This doc describes the standard way to assemble the database shard. A shard assembled using this approach becomes deterministic i.e. if two independent sides assemble the shard consists of the same ledgers, accounts and transactions, then they will obtain the same shard files `nudb.dat` and `nudb.key`. The approach deals with the `NuDB` database format only, refer to `https://github.com/vinniefalco/NuDB`.
+
+
+## Headers
+
+Due to NuDB database definition, the following headers are used for database files:
+
+nudb.key:
+```
+char[8]         Type            The characters "nudb.key"
+uint16          Version         Holds the version number
+uint64          UID             Unique ID generated on creation
+uint64          Appnum          Application defined constant
+uint16          KeySize         Key size in bytes
+uint64          Salt            A random seed
+uint64          Pepper          The salt hashed
+uint16          BlockSize       Size of a file block in bytes
+uint16          LoadFactor      Target fraction in 65536ths
+uint8[56]       Reserved        Zeroes
+uint8[]         Reserved        Zero-pad to block size
+```
+
+nudb.dat:
+```
+char[8]         Type            The characters "nudb.dat"
+uint16          Version         Holds the version number
+uint64          UID             Unique ID generated on creation
+uint64          Appnum          Application defined constant
+uint16          KeySize         Key size in bytes
+uint8[64]       (reserved)      Zeroes
+```
+there all fields are saved using network byte order (most significant byte first).
+
+To make the shard deterministic the following parameters are used as values of header field both for `nudb.key` and `nudb.dat` files.
+```
+Version         2
+UID             digest(0)
+Appnum          digest(2) | 0x5348524400000000 /* 'SHRD' */
+KeySize         32
+Salt            digest(1)
+Pepper          XXH64(Salt)
+BlockSize       0x1000 (4096 bytes)
+LoadFactor      0.5 (numeric 0x8000)
+```
+Note: XXH64() is well-known hash algorithm.
+
+The `digest(i)` mentioned above defined as the follows:
+
+First, RIPEMD160 hash `H` calculated of the following structure
+```
+uint256         lastHash        Hash of last ledger in shard
+uint32          index           Index of the shard
+uint32          firstSeq        Sequence number of first ledger in the shard
+uint32          lastSeq         Sequence number of last ledger in the shard
+uint32          version         Version of shard, 2 at the present
+```
+there all 32-bit integers are hashed in network byte order.
+
+Then, `digest(i)` is defined as the following portion of the above hash `H`:
+```
+digest(0) = H[0] << 56 | H[2] << 48 | ... | H[14] << 0,
+digest(1) = H[1] << 56 | H[3] << 48 | ... | H[15] << 0,
+digest(2) = H[19] << 24 | H[18] << 16 | ... | H[16] << 0,
+```
+where `H[i]` denotes `i`-th byte of hash `H`.
+
+
+## Contents
+
+After deterministic shard is created using the above mentioned headers, it filled with objects. First, all objects of the shard are collected and sorted in according to their hashes. Here the objects are: ledgers, SHAmap tree nodes including accounts and transactions, and final key object with hash 0. Objects are sorted by increasing of their hashes, precisely, by increasing of hex representations of hashes in lexicographic order. 
+
+For example, the following is an example of sorted hashes in their hex representation:
+```
+0000000000000000000000000000000000000000000000000000000000000000
+154F29A919B30F50443A241C466691B046677C923EE7905AB97A4DBE8A5C2423
+2231553FC01D37A66C61BBEEACBB8C460994493E5659D118E19A8DDBB1444273
+272DCBFD8E4D5D786CF11A5444B30FB35435933B5DE6C660AA46E68CF0F5C447
+3C062FD9F0BCDCA31ACEBCD8E530D0BDAD1F1D1257B89C435616506A3EE6CB9E
+58A0E5AE427CDDC1C7C06448E8C3E4BF718DE036D827881624B20465C3E1334F
+...
+```
+
+Finally, objects added to the shard one by one in the sorted order from low to high hashes.
+
+
+## Tests
+
+To perform test to deterministic shards implementation one can enter the following command:
+```
+rippled --unittest ripple.NodeStore.DatabaseShard
+```
+
+The following is the right output of deterministic shards test:
+```
+ripple.NodeStore.DatabaseShard DatabaseShard deterministic_shard with backend nudb
+Iteration 0: RIPEMD160[nudb.key] = 4CFA8985836B549EC99D2E9705707F488DC91E4E
+Iteration 0: RIPEMD160[nudb.dat] = 8CC61F503C36339803F8C2FC652C1102DDB889F1
+Iteration 1: RIPEMD160[nudb.key] = 4CFA8985836B549EC99D2E9705707F488DC91E4E
+Iteration 1: RIPEMD160[nudb.dat] = 8CC61F503C36339803F8C2FC652C1102DDB889F1
+```
+
--- a/src/ripple/nodestore/backend/NuDBFactory.cpp
+++ b/src/ripple/nodestore/backend/NuDBFactory.cpp
@@ -38,7 +38,10 @@ namespace NodeStore {
 class NuDBBackend : public Backend
 {
 public:
-    static constexpr std::size_t currentType = 1;
+    static constexpr std::uint64_t currentType = 1;
+    static constexpr std::uint64_t deterministicType = 0x5348524400000000ull;
+    /* "SHRD" in ASCII */
+    static constexpr std::uint64_t deterministicMask = 0xFFFFFFFF00000000ull;
 
     beast::Journal const j_;
     size_t const keyBytes_;
@@ -93,7 +96,11 @@ public:
     }
 
     void
-    open(bool createIfMissing) override
+    open(
+        bool createIfMissing,
+        boost::optional<uint64_t> appType,
+        boost::optional<uint64_t> uid,
+        boost::optional<uint64_t> salt) override
     {
         using namespace boost::filesystem;
         if (db_.is_open())
@@ -114,8 +121,9 @@ public:
                 dp,
                 kp,
                 lp,
-                currentType,
-                nudb::make_salt(),
+                appType.value_or(currentType),
+                uid.value_or(nudb::make_uid()),
+                salt.value_or(nudb::make_salt()),
                 keyBytes_,
                 nudb::block_size(kp),
                 0.50,
@@ -128,8 +136,25 @@ public:
         db_.open(dp, kp, lp, ec);
         if (ec)
             Throw<nudb::system_error>(ec);
-        if (db_.appnum() != currentType)
+
+        /** Old value currentType is accepted for appnum in traditional
+         *  databases, new value is used for deterministic shard databases.
+         *  New 64-bit value is constructed from fixed and random parts.
+         *  Fixed part is bounded by bitmask deterministicMask,
+         *  and the value of fixed part is deterministicType.
+         *  Random part depends on the contents of the shard and may be any.
+         *  The contents of appnum field should match either old or new rule.
+         */
+        if (db_.appnum() != appType.value_or(currentType) &&
+            (appType ||
+             (db_.appnum() & deterministicMask) != deterministicType))
             Throw<std::runtime_error>("nodestore: unknown appnum");
+    }
+
+    void
+    open(bool createIfMissing) override
+    {
+        open(createIfMissing, boost::none, boost::none, boost::none);
     }
 
     void
--- a/src/ripple/nodestore/impl/DeterministicShard.cpp
+++ b/src/ripple/nodestore/impl/DeterministicShard.cpp
@@ -1,0 +1,206 @@
+//------------------------------------------------------------------------------
+/*
+    This file is part of rippled: https://github.com/ripple/rippled
+    Copyright (c) 2020 Ripple Labs Inc.
+
+    Permission to use, copy, modify, and/or distribute this software for any
+    purpose  with  or without fee is hereby granted, provided that the above
+    copyright notice and this permission notice appear in all copies.
+
+    THE  SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+    WITH  REGARD  TO  THIS  SOFTWARE  INCLUDING  ALL  IMPLIED  WARRANTIES  OF
+    MERCHANTABILITY  AND  FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+    ANY  SPECIAL ,  DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+    WHATSOEVER  RESULTING  FROM  LOSS  OF USE, DATA OR PROFITS, WHETHER IN AN
+    ACTION  OF  CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+    OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+//==============================================================================
+
+#include <ripple/app/main/Application.h>
+#include <ripple/beast/hash/hash_append.h>
+#include <ripple/core/ConfigSections.h>
+#include <ripple/nodestore/Manager.h>
+#include <ripple/nodestore/impl/DeterministicShard.h>
+#include <ripple/nodestore/impl/Shard.h>
+#include <ripple/protocol/digest.h>
+#include <fstream>
+#include <nudb/detail/format.hpp>
+#include <nudb/nudb.hpp>
+#include <openssl/ripemd.h>
+
+namespace ripple {
+namespace NodeStore {
+
+DeterministicShard::DeterministicShard(
+    Application& app,
+    DatabaseShard const& db,
+    std::uint32_t index,
+    uint256 const& lastHash,
+    beast::Journal j)
+    : inited_(false)
+    , nodeset_([](nodeptr l, nodeptr r) { return l->getHash() < r->getHash(); })
+    , app_(app)
+    , db_(db)
+    , index_(index)
+    , hash_(hash(lastHash))
+    , tempdir_(db.getRootDir() / (std::to_string(index_) + ".tmp"))
+    , finaldir_(db.getRootDir() / std::to_string(index_))
+    , ctx_(std::make_unique<nudb::context>())
+    , j_(j)
+{
+}
+
+DeterministicShard::~DeterministicShard()
+{
+    close(true);
+}
+
+uint160
+DeterministicShard::hash(uint256 const& lastHash) const
+{
+    using beast::hash_append;
+    ripemd160_hasher h;
+
+    hash_append(h, lastHash);
+    hash_append(h, index_);
+    hash_append(h, db_.firstLedgerSeq(index_));
+    hash_append(h, db_.lastLedgerSeq(index_));
+    hash_append(h, Shard::version);
+
+    auto const result = static_cast<ripemd160_hasher::result_type>(h);
+    return uint160::fromVoid(result.data());
+}
+
+std::uint64_t
+DeterministicShard::digest(int n) const
+{
+    auto const data = hash_.data();
+
+    if (n == 2)
+    {  // Extract 32 bits:
+        return (static_cast<std::uint64_t>(data[19]) << 24) +
+            (static_cast<std::uint64_t>(data[18]) << 16) +
+            (static_cast<std::uint64_t>(data[17]) << 8) +
+            (static_cast<std::uint64_t>(data[16]));
+    }
+
+    std::uint64_t ret = 0;
+
+    if (n == 0 || n == 1)
+    {  // Extract 64 bits
+        for (int i = n; i < 16; i += 2)
+            ret = (ret << 8) + data[i];
+    }
+
+    return ret;
+}
+
+bool
+DeterministicShard::init()
+{
+    if (index_ < db_.earliestShardIndex())
+    {
+        JLOG(j_.error()) << "shard " << index_ << " is illegal";
+        return false;
+    }
+
+    Config const& config{app_.config()};
+
+    Section section{config.section(ConfigSection::shardDatabase())};
+    std::string const type{get<std::string>(section, "type", "nudb")};
+
+    if (type != "nudb")
+    {
+        JLOG(j_.error()) << "shard " << index_ << " backend type " << type
+                         << " not supported";
+        return false;
+    }
+
+    auto factory{Manager::instance().find(type)};
+    if (!factory)
+    {
+        JLOG(j_.error()) << "shard " << index_
+                         << " failed to create factory for backend type "
+                         << type;
+        return false;
+    }
+
+    ctx_->start();
+
+    section.set("path", tempdir_.string());
+    backend_ = factory->createInstance(
+        NodeObject::keyBytes, section, scheduler_, *ctx_, j_);
+
+    if (!backend_)
+    {
+        JLOG(j_.error()) << "shard " << index_
+                         << " failed to create backend type " << type;
+        return false;
+    }
+
+    // Open or create the NuDB key/value store
+    bool preexist = exists(tempdir_);
+    if (preexist)
+    {
+        remove_all(tempdir_);
+        preexist = false;
+    }
+
+    backend_->open(
+        !preexist,
+        digest(2) | 0x5348524400000000ll, /* appType */
+        digest(0),                        /* uid */
+        digest(1)                         /* salt */
+    );
+
+    inited_ = true;
+
+    return true;
+}
+
+void
+DeterministicShard::close(bool cancel)
+{
+    if (!inited_)
+        return;
+
+    backend_->close();
+    if (cancel)
+    {
+        remove_all(tempdir_);
+    }
+    else
+    {
+        flush();
+        remove_all(finaldir_);
+        rename(tempdir_, finaldir_);
+    }
+    inited_ = false;
+}
+
+void
+DeterministicShard::store(nodeptr nObj)
+{
+    if (!inited_)
+        return;
+
+    nodeset_.insert(nObj);
+}
+
+void
+DeterministicShard::flush()
+{
+    if (!inited_)
+        return;
+
+    for (auto nObj : nodeset_)
+    {
+        backend_->store(nObj);
+    }
+
+    nodeset_.clear();
+}
+
+}  // namespace NodeStore
+}  // namespace ripple
--- a/src/ripple/nodestore/impl/DeterministicShard.h
+++ b/src/ripple/nodestore/impl/DeterministicShard.h
@@ -1,0 +1,144 @@
+//------------------------------------------------------------------------------
+/*
+    This file is part of rippled: https://github.com/ripple/rippled
+    Copyright (c) 2020 Ripple Labs Inc.
+
+    Permission to use, copy, modify, and/or distribute this software for any
+    purpose  with  or without fee is hereby granted, provided that the above
+    copyright notice and this permission notice appear in all copies.
+
+    THE  SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+    WITH  REGARD  TO  THIS  SOFTWARE  INCLUDING  ALL  IMPLIED  WARRANTIES  OF
+    MERCHANTABILITY  AND  FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+    ANY  SPECIAL ,  DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+    WHATSOEVER  RESULTING  FROM  LOSS  OF USE, DATA OR PROFITS, WHETHER IN AN
+    ACTION  OF  CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+    OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+//==============================================================================
+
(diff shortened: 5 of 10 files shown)


p2peer:
Depend on https://github.com/ripple/rippled/pull/3332

MarkusTeufelberger:
Please don't use MD5 for anything. Even if its use might be OK in this instance, just keeping/using it anywhere in the code base is the wrong move in my opinon.

p2peer:
> Please don't use MD5 for anything. Even if its use might be OK in this instance, just keeping/using it anywhere in the code base is the wrong move in my opinon.

No problem if this is official position of Ripple.

src/ripple/nodestore/DeterministicShard.md
```diff
@@ -0,0 +1,104 @@
...
+uint32          version         Version of shard, 2 at the present
+```
+
+Then, `digest(i)` is defined as `i`-th 64-bit portion of the above hash `H` with swapped order of bytes, i.e.
+```
+digest(i) = H[i*8+0] << 56 | H[i*8+1] << 48 | ... | H[i*8+7] << 0.
+```
+where `H[i]` denotes `i`-th byte of hash `H`.
+
+
```
nbougalis, line 66: First question, and this is more theoretical: should we modify `NuDB` to increase the size of the headers of the key and data files so that they both are _precisely_ 512 bytes, increase the size of some fields (specifically the UID field that should be 256 bits) and allocate a portion of the header as an area of the database owner to store his own metadata? 

Now on to more practical matters: I really don't like the way that we're taking a cryptographic hash function, chopping its output up and manipulating it.

I recommend the following: replace `SHA256` with `RIPEMD-160` and use it to calculate `H` over the following data:

```
uint256         lastHash        Hash of last ledger in shard
uint32          index           Index of the shard
uint32          firstSeq        Sequence number of first ledger in the shard
uint32          lastSeq         Sequence number of last ledger in the shard
uint32          version         Version of shard, 2 at the present
```

Crack the 160 bit hash `H` apart into **2** 64-bit values `A` and `B` and **1** 32-bit value `C` and set: 

1. `UID := A`;
2. `SALT := B`; and
3. `APPNUM := C | 0x5348524400000000`. (`0x53585244 == 'S' 'H' 'R' 'D'`).

src/ripple/nodestore/backend/NuDBFactory.cpp
```diff
@@ -114,7 +114,7 @@ class NuDBBackend
         {
             create_directories(folder);
             nudb::create<nudb::xxhasher>(dp, kp, lp,
-                currentType, nudb::make_salt(), keyBytes_,
+                currentType, (salt ? *salt : nudb::make_salt()), keyBytes_,
```
nbougalis, line 117: Consider `salt.value_or(nudb::make_salt())` instead.

src/ripple/nodestore/Backend.h
```diff
@@ -56,6 +56,13 @@ class Backend
     */
     virtual void open(bool createIfMissing = true) = 0;
 
+    /** Open the backend.
+        @param createIfMissing Create the database files if necessary.
+        @param salt Deterministic salt used to create a backend.
+        This allows the caller to catch exceptions.
+    */
+    virtual void open(bool createIfMissing, boost::optional<uint64_t> salt) {}
```
nbougalis, line 64: This isn't ideal: it allows an implementation detail to "bleed" into the API, but I don't quite have a better idea on how to handle it.

At a _minimum_ the default implementation of this function, here in `Backend` should throw an `std::runtime_error` exception, with an appropriate error message that explains that the specific backend does not support deterministic salts.

src/ripple/nodestore/impl/DeterministicShard.cpp
```diff
@@ -0,0 +1,256 @@
...
+            }
+        };
+
+        stream_ keyfile(keypath.string());
+        stream_ datfile(datpath.string());
+
+        if (!keyfile || !datfile)
+            return false;
+
+        nudb::detail::key_file_header keyhead;
```
nbougalis, line 187: @p2peer, @miguelportilla we are reaching inside `NuDB` internals here; we absolutely **do not** want to do this. We need to modify `NuDB` so that it cleanly supports the operations that we need to perform.

src/test/jtx/impl/Env.cpp
```diff
@@ -124,11 +125,17 @@ Env::close(NetClock::time_point closeTime,
         app().getOPs().acceptLedger(consensusDelay);
     else
     {
-        rpc("ledger_accept");
+        auto resp = rpc("ledger_accept");
+        if (resp["result"]["status"] != std::string("success"))
+        {
+            std::cerr << "Env::close() failed: " << resp["result"]["status"] << std::endl;
+            res = false;
```
nbougalis, line 132: I can't imagine any situation where `ledger_accept` would not return success, but I know that you encountered at least one instance; I think it's worth understanding what happened there, because other that not running in standalone mode, there's no way the operation can fail and the unit tests essentially "force" standalone mode.

I would definitely not use `std::cerr` to log from here; use `test.log << "whatever"` instead.

src/ripple/nodestore/impl/Shard.cpp
```diff
@@ -556,7 +567,10 @@ Shard::finalize(const bool writeSQLite)
             return fail("missing root TXN node");
         }
 
-        if (!valLedger(ledger, next))
+        if (dsh)
+            dsh->store(nObj);
+
+        if (!valLedger(ledger, next, dsh))
```
nbougalis, line 573: While you're here, please rename this function `verifyLedger`. Rationale: `valLedger` is a poor name and the term `validate` has a very different meaning in our code; we should avoid confusing. Similarly, please reame the message below: "verification check failed"

src/ripple/nodestore/DeterministicShard.md
p2peer: Saving hash of shard as UID/salt was not required by this ticket. It is my proposition. Ticket recommends to set UID/salt to known constant + shard_index. Your proposition of using RIPEMD-160 instead of SHA256 is better. I think that changing NuDB standard is much more difficult because current standard is used is main net if I am not mistaking.

src/ripple/nodestore/impl/DeterministicShard.cpp
p2peer: Which way do you want to modify NuDB: 1) include NuDB to rippled and change the code; 2) connect to maintainers of NuDB and ask them to push a modification? I don't know if the way 2 will be effective.

src/test/jtx/impl/Env.cpp
p2peer: Several failures of Env::close() appears on almost every local run or Travis run, you can see the Travis log for example: [https://travis-ci.com/github/ripple/rippled/jobs/320158724)](https://travis-ci.com/github/ripple/rippled/jobs/320158724), lines 713-721 and 773-775. I used std::cerr to see these messages on Travis log with --quiet option.

src/ripple/nodestore/impl/Shard.cpp
p2peer: No problem if the author (@miguelportilla if I am not mistaking) agree with it.

miguelportilla: `verifyLedger` is okay with me.

vinniefalco:
None of this is necessary, all you need to do is store a key/value pair with a known key constant (say, the number 1 expressed as a 256-bit little endian integer) and with the value containing all the data you want.

p2peer:
> None of this is necessary, all you need to do is store a key/value pair with a known key constant (say, the number 1 expressed as a 256-bit little endian integer) and with the value containing all the data you want.

These data are already saved as a finalKey under key 0. But the problem is that UID and salt should be deterministic. I can pass custom salt to the NuDB constructor, but I can't pass UID. That is the problem, so the modification of NuDB constructor is required.

vinniefalco:
> I can't pass UID

The UID is not user-modifiable. The purpose of this field is to trivially detect when a data file does not match the key file.

vinniefalco:
Oh, I see... you want two shards generated by different machines to have the same NuDB file contents. In this case, yes I think we can add UID as a parameter to `nudb::create`, and a new function `make_uid` which produces the necessary value. But this could open rippled up to algorithmic complexity attacks. I will research this.

MarkusTeufelberger:
I guess to prevent these complexity attacks the hash of the last header in the shard is added to the salt, so there's no real way to mount such an attack until it is already too late to know how the hash will be modified for permanent storage.

I am somewhat concerned about compression of values stored in there. Which compression scheme (if any?) must be used?

vinniefalco:
> I guess to prevent these complexity attacks the hash of the last header in the shard is added to the salt

There is no way to mount an attack, because each node builds the shard using a random salt. And when the shard is preprocessed for sharing, it is rewritten (to put the key/value pairs in order and thus, be deterministic). But since the shard at this point is read-only, no attack is possible.

MarkusTeufelberger:
Wouldn't an algorithmic attack also affect a database file that is only being read from (and that only contains a single bucket that has tons of spill entries due to hash collisions in the extreme case)?

vinniefalco:
> Wouldn't an algorithmic attack also affect a database file that is only being read from (and that only contains a single bucket that has tons of spill entries due to hash collisions in the extreme case)?

An algorithmic attack on the database works by trying to get a client to build or accept a database whose keys collide. A client only accepts a database if the contents have been signed by validators. The deterministic salt cannot be predicted in advance, so the attacker has no idea what ledger entries she must create in order to cause collisions. And once the shard is built, an attacker cannot insert anything so no attack is possible.

MarkusTeufelberger:
> The deterministic salt cannot be predicted in advance

Yes, due to the header hash of the last header that gets mixed in, as I wrote already. I probably misunderstood your initial comment about the "just use a constant key of 1" somewhere.

As far as i understand this here, so far it will create a deterministic header and a deterministic order of keys. I'm unsure if (e.g. after an update of `lz4`) the actual stored values will be deterministic though. There are 2 options: Either mandate (and document) a certain compression scheme or store all data uncompressed. Both options have their advantages and disadvantages - if one assumes that during transport content will get compressed anyways there's not much to gain from compressing at insertion (and disk space is relatively cheap for these files - they get written once and then ingested elsewhere once too, where they can be compressed by the receiver). If one assumes/proves that compression will result in huge benefits or is inherently deterministic (which might be true for several runs with the same `lz4` library in single threaded mode, but I'm not even 100% certain if it would be the case for different (future) versions of that library or if that's a guarantee that they give) then these benefits of course should also be taken.

As a data point, storing these shards with uncompressed values and then compressing the resulting files with `zstd` was about as good as storing `lz4` compressed values and compressed inner nodes at least for the very early shards that I experimented with a while back. This means there's at least the option to create uncompressed deterministic files, compress them for longer term storage or distribution and recompress them upon ingestion into whatever format or database `rippled` in the distant future implements. Shard contents are anyways re-verified upon import these days, right?

vinniefalco:
Compressing the entire uncompressed data file instead of compressing each individual key/value pair makes a lot more sense (and is easily extensible too, if you put a header in front of the resulting compressed file).

src/ripple/nodestore/DeterministicShard.md
p2peer: Done

src/ripple/nodestore/backend/NuDBFactory.cpp
p2peer: Done

src/ripple/nodestore/Backend.h
p2peer: Done

src/ripple/nodestore/impl/DeterministicShard.cpp
p2peer: Author of NuDB said that he will research this, see his comment below. While he is researching, I put code under macro NUDB_SUPPORTS_CUSTOM_UID.

p2peer:
> Please don't use MD5 for anything. Even if its use might be OK in this instance, just keeping/using it anywhere in the code base is the wrong move in my opinon.

Changed to RIPEMD-160.

p2peer:
> I am somewhat concerned about compression of values stored in there. Which compression scheme (if any?) must be used?

For now, compression by default is used. I can change this to no compression, if this is better.

src/ripple/nodestore/impl/Shard.cpp
p2peer: Done

src/test/jtx/impl/Env.cpp
p2peer: Writing to std::cerr replaced by writing to journal.

src/ripple/nodestore/Backend.h
```diff
@@ -58,6 +58,25 @@ class Backend
...
+    */
+    virtual void
+    open(
+        bool createIfMissing,
+        boost::optional<uint64_t> appType,
+        boost::optional<uint64_t> uid,
+        boost::optional<uint64_t> salt)
+    {
+        Throw<std::runtime_error>(std::string(
+            "Deterministic appType/uid/salt not supported by bachend " +
```
HowardHinnant, line 76: Spelling error:  bachend -> backend

src/ripple/nodestore/backend/NuDBFactory.cpp
```diff
@@ -38,7 +38,9 @@ namespace NodeStore {
 class NuDBBackend : public Backend
 {
 public:
-    static constexpr std::size_t currentType = 1;
+    static constexpr std::uint64_t currentType = 1;
+    static constexpr std::uint64_t deterministicType = 0x5348524400000000ll;
+    static constexpr std::uint64_t deterministicMask = 0xFFFFFFFF00000000ll;
```
HowardHinnant, line 43: Nitpick:  A `ull` suffix would be more consistent.

src/ripple/nodestore/impl/DeterministicShard.h
```diff
@@ -0,0 +1,104 @@
...
+#include <set>
+
+namespace ripple {
+namespace NodeStore {
+
+class DeterministicShard
+{
+public:
+    using nodeptr = std::shared_ptr<NodeObject>;
+
```
HowardHinnant, line 35: Consider adding:

```c++
DeterministicShard(DeterministicShard const&) = delete;
DeterministicShard& operator=(DeterministicShard const&) = delete;
```

This won't change anything.  `DeterministicShard` is neither copyable nor movable, constructible nor assignable, due to the combination of the user-declared destructor and the datamembers.  But that fact is subtle.  By explicitly deleting the copy members you make it more clear that this class has no copy nor move members.

Some people would go further and delete the move members too.  My personal taste is that that is overly verbose because the user-declared copy members suppress the move members.

src/ripple/nodestore/DeterministicShard.md
```diff
@@ -0,0 +1,102 @@
...
+Salt            digest(1)
+Pepper          xxhasher(Salt)
+BlockSize       0x1000
+LoadFactor      0.5 (numeric 0x8000)
+```
+Note: xxhasher is native hasher of NuDB, refer to NuDB code.
+
+The `digest(i)` mentioned above defined as the follows:
+
+First, SHA256 hash `H` calculated of the following structure
```
MarkusTeufelberger, line 51: This still needs to be changed to the actual RIPEMD160 implementation.

src/ripple/nodestore/impl/DeterministicShard.cpp
```diff
@@ -0,0 +1,303 @@
...
+        if (ec)
+            return false;
+        nudb::detail::write(datfile, dathead, ec);
+        if (ec)
+            return false;
+    }
+
+    backend_->open(
+        false,
+        digest(2) | 0x5348524400000000ll, /* appType */
```
MarkusTeufelberger, line 248: Maybe a small explanation in the comment to this magic number? ("SHRD" in ASCII)

src/ripple/nodestore/impl/Shard.h
```diff
@@ -126,9 +127,10 @@ class Shard final
 
         @param writeSQLite If true, SQLite entries will be rewritten using
         verified backend data.
+        @param writeSQLite If true, shard will be rewritten deterministically.
```
MarkusTeufelberger, line 130:
```suggestion
        @param writeDeterministicShard If true, shard will be rewritten deterministically.
```

src/ripple/nodestore/backend/NuDBFactory.cpp
```diff
@@ -38,7 +38,9 @@ namespace NodeStore {
 class NuDBBackend : public Backend
 {
 public:
-    static constexpr std::size_t currentType = 1;
+    static constexpr std::uint64_t currentType = 1;
+    static constexpr std::uint64_t deterministicType = 0x5348524400000000ll;
```
MarkusTeufelberger, line 42: Also here a comment on what this number means would be useful

src/ripple/nodestore/DeterministicShard.md
```diff
@@ -0,0 +1,102 @@
+# Deterministic Database Shards
+
+This doc describes the standard way to assemble the database shard. A shard assembled using this approach becomes deterministic i.e. if two independent sides assemble the shard consists of the same ledgers, accounts and transactions, then they will obtain the same shard files `nudb.dat` and `nudb.key`. The approach deals with the `NuDB` database format only, refer to `https://github.com/vinniefalco/NuDB`.
+
+
+## Headers
+
+Due to NuDB database definition, the following headers are using for database files:
```
MarkusTeufelberger, line 8:
```suggestion
Due to NuDB database definition, the following headers are used for database files:
```

src/ripple/nodestore/DeterministicShard.md
```diff
@@ -0,0 +1,102 @@
...
+Version         2
+UID             digest(0)
+Appnum          digest(2) | 0x5348524400000000 /* 'SHRD' */
+KeySize         32
+Salt            digest(1)
+Pepper          xxhasher(Salt)
+BlockSize       0x1000
+LoadFactor      0.5 (numeric 0x8000)
+```
+Note: xxhasher is native hasher of NuDB, refer to NuDB code.
```
MarkusTeufelberger, line 47: Btw. it is the "XXH64" algorithm, might be worth documenting (though I don't think NuDB would change that anytime soon) to save one lookup in a different repository.

src/ripple/nodestore/DeterministicShard.md
```diff
@@ -0,0 +1,102 @@
...
+
+To make the shard deterministic the following parameters are used as values of header field both for `nudb.key` and `nudb.dat` files.
+```
+Version         2
+UID             digest(0)
+Appnum          digest(2) | 0x5348524400000000 /* 'SHRD' */
+KeySize         32
+Salt            digest(1)
+Pepper          xxhasher(Salt)
+BlockSize       0x1000
```
MarkusTeufelberger, line 44:
```suggestion
BlockSize       0x1000 (4096 bytes)
```

src/ripple/nodestore/Backend.h
p2peer: Done

src/ripple/nodestore/backend/NuDBFactory.cpp
p2peer: Done

src/ripple/nodestore/impl/DeterministicShard.h
p2peer: Done

src/ripple/nodestore/DeterministicShard.md
p2peer: Done

src/ripple/nodestore/impl/DeterministicShard.cpp
p2peer: Done

src/ripple/nodestore/impl/Shard.h
p2peer: Done

src/ripple/nodestore/backend/NuDBFactory.cpp
p2peer: Done

src/ripple/nodestore/DeterministicShard.md
p2peer: Done

src/ripple/nodestore/DeterministicShard.md
p2peer: Done

src/ripple/nodestore/DeterministicShard.md
p2peer: Done

HowardHinnant:
Needs rebasing to 1.6.0-b5.

MarkusTeufelberger:
Having an option for storing data uncompressed would be really nice. Might be something for a future iteration though.

src/ripple/nodestore/impl/Shard.cpp
```diff
@@ -627,6 +649,14 @@ Shard::finalize(
             std::string("exception ") + e.what() + " in function " + __func__);
     }
 
+    if (dsh)
+    {
+        backend_->close();
+        dsh->close();
+        backend_->open(false);
+        return finalize(false, expectedHash, false);
+    }
```
pwang200, line 658: Could you please explain the rational of this block of code?

src/ripple/nodestore/backend/NuDBFactory.cpp
```diff
@@ -128,10 +138,18 @@ class NuDBBackend : public Backend
         db_.open(dp, kp, lp, ec);
         if (ec)
             Throw<nudb::system_error>(ec);
-        if (db_.appnum() != currentType)
+        if (db_.appnum() != appType.value_or(currentType) &&
+            (appType ||
+             (db_.appnum() & deterministicMask) != deterministicType))
```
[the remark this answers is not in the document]
p2peer: In the past only currentType value was used as a valid values for the appnum field. By the changes of this PR old value currentType is still accepted for traditional databases, but new value is used for deterministic databases. New 64-bit value is constructed from two bit parts: fixed and random. Boundaries of fixed part is determined by bitmask deterministicMask, and the value of fixed part is deterministicType. Random part depends on the contents of the shard and may be any. Summarising this, to determine if the database is valid we should check, is the contents of appnum filed match old rule or new rule. So the right condition "field is not accepted" can be coded as "old rule is not applied" and "new rule is not applied". That's why we have a condition db_.appnum() != currentType && (db_.appnum() & deterministicMask) != deterministicType .

src/ripple/nodestore/impl/Shard.cpp
```diff
@@ -938,11 +969,14 @@ Shard::valLedger(
...
     bool error{false};
-    auto visit = [this, &error](SHAMapAbstractNode& node) {
+    auto visit = [this, &error, dsh](SHAMapAbstractNode& node) {
         if (stop_)
             return false;
-        if (!valFetch(node.getNodeHash().as_uint256()))
+        auto nObj = valFetch(node.getNodeHash().as_uint256());
+        if (!nObj)
             error = true;
+        if (dsh)
```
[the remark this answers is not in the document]
p2peer: Done

src/ripple/nodestore/backend/NuDBFactory.cpp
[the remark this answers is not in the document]
pwang200: Thanks for the explanation.

src/ripple/nodestore/impl/DeterministicShard.h
```diff
@@ -0,0 +1,108 @@
...
+    store(nodeptr nobj);
+
+    void
+    flush();
+
+    uint160
+    hash(const uint256& lastHash) const;
+
+    std::uint64_t
+    digest(int n) const;
```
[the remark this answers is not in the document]
p2peer: Done

src/ripple/nodestore/impl/DeterministicShard.cpp
```diff
@@ -0,0 +1,303 @@
...
+    backend_->close();
+    if (cancel)
+    {
+        remove_all(tempdir_);
+    }
+    else
+    {
+        flush();
+        remove_all(finaldir_);
+        rename(tempdir_, finaldir_);
```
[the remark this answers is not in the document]
p2peer: 1. The init() method creates temporary folder tempdir_, and the deterministic shard is initialized in that folder.
2. The store() method adds object to memory pool.
3. The flush() method stores all objects from memory pool to the shard located in tempdir_ in sorted order.
4. The close(true) method finalizes the shard and moves it from tempdir_ temporary folder to filandir_ permanent folder, deleting old (non-deterministic) shard located in finaldir_.

Done: added this comment to code.

Document is available in the file src/ripple/nodestore/DeterministicShard.md

src/ripple/nodestore/impl/Shard.cpp
p2peer: backend_->close(); // Close non-deterministic shard database.
dsh->close(); // Replace non-deterministic shard by deterministic one.
backend_->open(false); // Re-open deterministic shard database.

The finalize() function verifies the shard and, if third parameter is true, then replaces the shard by deterministic copy of the shard. After deterministic shard is created it verifies again, The finalize() function called here to verify deterministic shard, third parameter is false.

Done: comment added to code

src/ripple/nodestore/impl/DeterministicShard.h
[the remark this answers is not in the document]
pwang200: Thanks. Please edit the document of `flush()`.

src/ripple/nodestore/impl/DeterministicShard.cpp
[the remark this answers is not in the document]
pwang200: Thanks for the explanation!

src/ripple/nodestore/impl/Shard.cpp
pwang200: Thanks so much!

p2peer:
> Please edit the document of the `flush()` function in DeterministicShard.h

Done

src/ripple/nodestore/impl/DeterministicShard.cpp
```diff
@@ -0,0 +1,198 @@
...
+    std::uint64_t res = 0;
+    for (int i = sizeof(std::uint64_t) - 1; i >= 0; --i)
+    {
+        res <<= 8;
+        ind = i + n * sizeof(std::uint64_t);
+        if (ind < 160 / 8)
+            res += hash_.data()[ind] & 0xff;
+    }
+    return res;
+}
```
nbougalis, line 89: This function works but seems a little overly complicated for what it's trying to do. You don't need to change it, but I'm wondering how you feel about this:

```
std::uint64_t
DeterministicShard::digest(int n) const
{
    auto const data = hash_.data();

    if (n == 2)
    { // Extract 32 bits:
        return (static_cast<std::uint64_t>(data[19]) << 24) +
            (static_cast<std::uint64_t>(data[18]) << 16) +
            (static_cast<std::uint64_t>(data[17]) << 8) +
            (static_cast<std::uint64_t>(data[16]));
    }

    std::uint64_t ret = 0;

    if (n == 0 || n == 1)
    { // Extract 64 bits 
        for(int i = n; i < 16; i += 2)
            ret = (ret << 8) + data[i];
    }

    return ret;
}
```

Note that `digest(2)` now returns 32 bits (as a 64-bit value), while `digest(0)` and `digest(1)` return 64-bit values.

src/test/nodestore/DatabaseShard_test.cpp
```diff
@@ -27,13 +27,125 @@
...
+#include <fstream>
+#include <iostream>
+#include <openssl/ripemd.h>
 #include <test/jtx.h>
 #include <test/nodestore/TestBase.h>
 
 namespace ripple {
 namespace NodeStore {
 
+/* std::uniform_int_distribution is platform dependent */
```
nbougalis, line 41: Ugh... I don't know that I care for all this. I understand the need, given the platform-specific implementation details of `std::uniform_int_distribution`. I wonder if there's a better option here...

src/test/nodestore/DatabaseShard_test.cpp
```diff
@@ -978,6 +1090,93 @@ class DatabaseShard_test : public TestBase
         }
     }
 
+    std::string
+    ripemd160File(std::string filename)
```
nbougalis, line 1094: Same comment as earlier: please use the hasher.

src/test/nodestore/DatabaseShard_test.cpp
seelabs: What's the motivation for doing this? It it's to make sure we can repro unit test bugs on all platforms, then I'd just live with the platform dependence and remove this class. There is a very small possibility that we'll see failures on some platforms that we won't be able to repro on other platforms in the first place. If this happens, we'll just have to debug on the platforms where we can repro.

(if there's another reason for the class then nvm, I only took a very quick look before weighing in).

p2peer: My unit test for deterministic shards is the following: it generates predictable accounts and transactions, packs them into ledgers and makes the shard. The hash of this shard should be equal to the given value. I saw that on different platforms (precisely, linux and mac) hashes of the resulting shard was different. I unvestigated this case and determine that the problem is in the class std::uniform_int_distribution which generates different pseudorandom sequences on different platforms, but we need predictable sequence.

seelabs: @p2peer That is a different motivation that what I was envisioning. Thanks for the clarification!

src/test/nodestore/DatabaseShard_test.cpp
p2peer: Done

src/test/nodestore/DatabaseShard_test.cpp
p2peer: Done: added comment to the code

src/ripple/nodestore/impl/DeterministicShard.cpp
p2peer: Done

src/ripple/nodestore/DeterministicShard.md
```diff
@@ -0,0 +1,102 @@
...
+Salt            digest(1)
+Pepper          XXH64(Salt)
+BlockSize       0x1000 (4096 bytes)
+LoadFactor      0.5 (numeric 0x8000)
+```
+Note: XXH64() is well-known hash algorithm.
+
+The `digest(i)` mentioned above defined as the follows:
+
+First, RIPEMD160 hash `H` calculated of the following structure
```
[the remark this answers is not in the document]
p2peer: Done

src/ripple/nodestore/impl/DeterministicShard.cpp
```diff
@@ -0,0 +1,303 @@
...
+
+    RIPEMD160_Init(&c);
+    RIPEMD160_Update(&c, &lastHash, sizeof(lastHash));
+    RIPEMD160_Update(&c, &index_, sizeof(index_));
+    auto firstSeq = db_.firstLedgerSeq(index_);
+    RIPEMD160_Update(&c, &firstSeq, sizeof(firstSeq));
+    auto lastSeq = db_.lastLedgerSeq(index_);
+    RIPEMD160_Update(&c, &lastSeq, sizeof(lastSeq));
+    RIPEMD160_Update(&c, &Shard::version, sizeof(Shard::version));
+    RIPEMD160_Final(result, &c);
```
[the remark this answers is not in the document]
p2peer: Done

PR #3364, opened 2020-04-17
[TRIVIAL] Reduce visibility of retired amendment identifiers:

nbougalis:
Identifiers for retired amendments should not generally be used in the codebase.

This commit reduces their visibility down to one translation unit and marks them as unused and deprecated to prevent accidental reuse.

### DIFF
--- a/src/ripple/protocol/impl/Feature.cpp
+++ b/src/ripple/protocol/impl/Feature.cpp
@@ -154,58 +154,52 @@ bitsetIndexToFeature(size_t i)
     return featureCollections.bitsetIndexToFeature(i);
 }
 
-uint256 const featureTickets = *getRegisteredFeature("Tickets");
-uint256 const featureOwnerPaysFee = *getRegisteredFeature("OwnerPaysFee");
-uint256 const featureFlow = *getRegisteredFeature("Flow");
-uint256 const featureCompareTakerFlowCross =
-    *getRegisteredFeature("CompareTakerFlowCross");
-uint256 const featureFlowCross = *getRegisteredFeature("FlowCross");
-uint256 const featureCryptoConditionsSuite =
-    *getRegisteredFeature("CryptoConditionsSuite");
-uint256 const fix1513 = *getRegisteredFeature("fix1513");
-uint256 const featureDepositAuth = *getRegisteredFeature("DepositAuth");
-uint256 const featureChecks = *getRegisteredFeature("Checks");
-uint256 const fix1571 = *getRegisteredFeature("fix1571");
-uint256 const fix1543 = *getRegisteredFeature("fix1543");
-uint256 const fix1623 = *getRegisteredFeature("fix1623");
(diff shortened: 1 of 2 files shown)


scottschurr:
@nbougalis, I think reducing the visibility of these amendment identifiers is a great idea.  I've incorporated your suggested changes into the retiring amendments pull request:  https://github.com/ripple/rippled/pull/3368.

nbougalis:
I am going to close this PR, since you've folded these changes in 3368.

nbougalis:
@scottschurr: considering that #3368 is closed for now, I'm going to reopen this to reduce the visibility of retired amendments. Would appreciate a review.

src/ripple/protocol/impl/Feature.cpp
```diff
@@ -154,58 +154,52 @@ bitsetIndexToFeature(size_t i)
...
-uint256 const retiredSortedDirectories =
-    *getRegisteredFeature("SortedDirectories");
-uint256 const retiredFix1201 = *getRegisteredFeature("fix1201");
-uint256 const retiredFix1512 = *getRegisteredFeature("fix1512");
-uint256 const retiredFix1523 = *getRegisteredFeature("fix1523");
-uint256 const retiredFix1528 = *getRegisteredFeature("fix1528");
+// clang-format off
+
+uint256 const
+    featureTickets                  = *getRegisteredFeature("Tickets"),
```
scottschurr, line 160: Just as a note, if `getRegisteredFeature()` is called with a misspelled name it returns a null pointer which then crashes the app on startup.  I had this happen to me while messing with features for retiring amendments.  So I created different function that would create a good error message if the passed in name was not found.

Here's a commit you can cherry-pick if you'd like to pick up that change: https://github.com/scottschurr/rippled/commit/9277217d150b4a29528f1035105905300fea4143  That commit also includes some improved comments.

src/ripple/protocol/impl/Feature.cpp
```diff
@@ -154,58 +154,52 @@ bitsetIndexToFeature(size_t i)
...
+    fixQualityUpperBound            = *getRegisteredFeature("fixQualityUpperBound"),
+    featureRequireFullyCanonicalSig = *getRegisteredFeature("RequireFullyCanonicalSig"),
+    fix1781                         = *getRegisteredFeature("fix1781"),
+    featureHardenedValidations      = *getRegisteredFeature("HardenedValidations");
+
+// The following amendments have been active for at least two years. Their
+// pre-amendment code has been removed and the identifiers are deprecated.
+[[deprecated("The referenced amendment has been retired"), maybe_unused]]
+uint256 const
+    retiredPayChan           = *getRegisteredFeature("PayChan"),
```
scottschurr, line 190: This list is incomplete because when I first started unconditionalizing amendments I completely removed their registered features.  I have a commit you could cherry-pick (the same one I mention elsewhere) which fills in the retired amendments that are missing from this list: https://github.com/scottschurr/rippled/commit/9277217d150b4a29528f1035105905300fea4143

PR #3368, opened 2020-04-21
Add featureRetire2017Amendments:

scottschurr:
If the Retire2017Amendments amendment is activated, then those amendments that are ...

1. Enabled,
2. No longer dependent on runtime amendment checks, and
3. Listed in detail::retiringAmendments() ...

will be removed from the Amendment ledger object and the AmendmentTable.

The motivation for this change is to reduce the number of amendments stored in the Amendment ledger object.  It also reduces noise in the feature RPC command (and source code) regarding amendments that have been enabled for a long time.

### DIFF
--- a/src/ripple/app/misc/README.md
+++ b/src/ripple/app/misc/README.md
@@ -62,32 +62,32 @@ of the administrator to set these values.
 
 # Amendment
 
-An Amendment is a new or proposed change to a ledger rule. Ledger rules affect 
-transaction processing and consensus; peers must use the same set of rules for 
-consensus to succeed, otherwise different instances of rippled will get 
-different results. Amendments can be almost anything but they must be accepted 
-by a network majority through a consensus process before they are utilized. An 
-Amendment must receive at least an 80% approval rate from validating nodes for 
-a period of two weeks before being accepted. The following example outlines the 
-process of an Amendment from its conception to approval and usage. 
+An Amendment is a new or proposed change to a ledger rule. Ledger rules affect
+transaction processing and consensus; peers must use the same set of rules for
+consensus to succeed, otherwise different instances of rippled will get
+different results. Amendments can be almost anything but they must be accepted
+by a network majority through a consensus process before they are utilized. An
+Amendment must receive at least an 80% approval rate from validating nodes for
+a period of two weeks before being accepted. The following example outlines the
+process of an Amendment from its conception to approval and usage.
 
-*  A community member makes proposes to change transaction processing in some 
-  way. The proposal is discussed amongst the community and receives its support 
-  creating a community or human consensus. 
+*  A community member proposes to change transaction processing in some way.
+  The proposal is discussed amongst the community and receives its support
+  creating a community or human consensus.
 
 *  Some members contribute their time and work to develop the Amendment.
 
-*  A pull request is created and the new code is folded into a rippled build 
+*  A pull request is created and the new code is folded into a rippled build
   and made available for use.
 
 *  The consensus process begins with the validating nodes.
 
-*  If the Amendment holds an 80% majority for a two week period, nodes will begin 
+*  If the Amendment holds an 80% majority for a two week period, nodes will begin
   including the transaction to enable it in their initial sets.
 
-Nodes may veto Amendments they consider undesirable by never announcing their 
-support for those Amendments. Just a few nodes vetoing an Amendment will normally 
-keep it from being accepted. Nodes could also vote yes on an Amendments even 
+Nodes may veto Amendments they consider undesirable by never announcing their
+support for those Amendments. Just a few nodes vetoing an Amendment will normally
+keep it from being accepted. Nodes could also vote yes on an Amendments even
 before it obtains a super-majority. This might make sense for a critical bug fix.
 
 Validators that support an amendment that is not yet enabled announce their
@@ -101,7 +101,7 @@ the majority status from the ledger.
 If an amendment holds majority status for two weeks, validators will
 introduce a pseudo-transaction to enable the amendment.
 
-All amednements are assumed to be critical and irreversible. Thus there
+All amendments are assumed to be critical and irreversible. Thus there
 is no mechanism to disable or revoke an amendment, nor is there a way
 for a server to operate while an amendment it does not understand is
 enabled.
--- a/src/ripple/app/misc/impl/AmendmentTable.cpp
+++ b/src/ripple/app/misc/impl/AmendmentTable.cpp
@@ -21,6 +21,7 @@
 #include <ripple/app/misc/AmendmentTable.h>
 #include <ripple/core/ConfigSections.h>
 #include <ripple/core/DatabaseCon.h>
+#include <ripple/protocol/Feature.h>
 #include <ripple/protocol/STValidation.h>
 #include <ripple/protocol/TxFlags.h>
 #include <ripple/protocol/jss.h>
@@ -28,6 +29,7 @@
 #include <boost/regex.hpp>
 #include <algorithm>
 #include <mutex>
+#include <utility>
 
 namespace ripple {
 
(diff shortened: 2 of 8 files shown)


src/ripple/app/misc/README.md
```diff
@@ -62,32 +62,32 @@ of the administrator to set these values.
...
+An Amendment is a new or proposed change to a ledger rule. Ledger rules affect
+transaction processing and consensus; peers must use the same set of rules for
+consensus to succeed, otherwise different instances of rippled will get
+different results. Amendments can be almost anything but they must be accepted
+by a network majority through a consensus process before they are utilized. An
+Amendment must receive at least an 80% approval rate from validating nodes for
+a period of two weeks before being accepted. The following example outlines the
+process of an Amendment from its conception to approval and usage.
+
+*  A community member makes proposes to change transaction processing in some
```
seelabs, line 74: Typo? `makes proposes` -> `proposes`

src/ripple/app/tx/impl/Change.cpp
```diff
@@ -189,6 +189,26 @@ Change::applyAmendment()
...
+        if (amendment == featureRetire2017Amendments)
+        {
+            // For every retiring amendment, if we find it remove it from
+            // amendmentObject.
+            //
+            // Note that we can't remove the amendments from the AmendmentTable
+            // yet since the AmendmentTable is not a ledger object.  Just
+            // because we do the Change doesn't guarantee enough other
+            // validators will.
+            for (uint256 const& retire : detail::retiringAmendments())
```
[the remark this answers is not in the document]
scottschurr: Good spotting, and almost.  But `set_difference` operates on sorted ranges and neither `detail::retiringAmendments` nor `amendments` are sorted.

src/ripple/app/tx/impl/Change.cpp
```diff
@@ -189,6 +189,26 @@ Change::applyAmendment()
...
             ctx_.app.getOPs ().setAmendmentBlocked ();
         }
+
+        // See if the just enabled amendment retires already-enabled amendments.
+        if (amendment == featureRetire2017Amendments)
+        {
+            // For every retiring amendment, if we find it remove it from
+            // amendmentObject.
+            //
+            // Note that we can't remove the amendments from the AmendmentTable
```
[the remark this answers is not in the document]
scottschurr: I see your perspective.  In a practical sense all validators _should_ be doing this on the same ledger and the outcome should, under ordinary circumstances, be a foregone conclusion.

What I'm attempting to say here is that, in general, we don't assume that all validators move in lock step.  We expect them to produce different ledgers initially and then negotiate a ledger they can all agree on.  Just because the `Change` transaction is applied locally doesn't allow us to assume that the `Change` will be in the validated ledger.  We have to wait for all of the validators to weigh in before we can assume that.

Once we see `featureRetire2017Amendments` in a validated ledger then we're allowed to modify `AmendmentTable`.  But at this point we're not allowed to assume that everyone will agree on `featureRetire2017Amendments`.  So we wait and make the change in `AmendmentTable:: doValidatedLedger`.

If you find the comment confusing I'm happy to remove it or reword it.  Or maybe I'm just confused...

src/ripple/app/tx/impl/Change.cpp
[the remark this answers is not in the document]
nbougalis: Which raises the interesting question of whether they _should_ be. The existing "temporal" order has one benefit: it shows what activated in what order, but sorting offers benefits too.

src/ripple/app/misc/impl/AmendmentTable.cpp
```diff
@@ -554,40 +587,88 @@ AmendmentTableImpl::needValidatedLedger (LedgerIndex ledgerSeq)
...
+    if (rules.enabled(featureRetire2017Amendments))
+    {
+        // Remove retired amendments from the AmendmentMap_ so we stop
+        // reporting about amendments that are retired.  This should only
+        // need to be done once, so check whether it needs to be done
+        // before actually doing the work.
+        bool const needsRetiring = [this, ledgerSeq]()
+        {
+            std::lock_guard sl (mutex_);
+            lastUpdateSeq_ = ledgerSeq;
```
pwang200, line 637: Not in Scott's new code, but in existing code, I don't see lastUpdateSeq_ get updated. It is initialized to 0 and seems never changed. If so, the function `bool needValidatedLedger (LedgerIndex seq)` always return true, right? Again, it is in the existing code.

pwang200:
The spec and in code comments are really helpful! To confirm my understanding, all nodes will remove the retired amendments from the ledger at the same ledger sequence. But the fixing of AmendmentTable data structure is done later and nodes may do it at different times, which is ok.

src/ripple/app/misc/impl/AmendmentTable.cpp
scottschurr: Yes, this fixes a minor bug in the existing code.  I found the bug while testing the new code.  `needValidatedLedger()` always returning true was a pessimization of this code:  https://github.com/ripple/rippled/blob/develop/src/ripple/app/misc/AmendmentTable.h#L80  It would always call `doValidatedLedger()`.

pwang200: Right. And it is minor performance bug. I just did not get why the fix will only take effect after featureRetire2017Amendments is enabled.

scottschurr: I don't remember the original motivation.  Perhaps I was being conservative.  Since I have an amendment readily available, putting it on the amendment means that the behavior change happens across the network at the same time.  I don't _think_ the change should impact behavior noticeably but, just in case, it's a bit more cautious to make the change on an amendment.

I'd be happy to move the change outside of the amendment if you think that's better.  I don't believe there's any risk to that.  But we would want to make the change carefully.  `lastUpdateSeq_` should only be touched under the lock.  And, for efficiency, we probably don't want to do the work inside the preceding `for` loop under its lock.  Where it is right now it a place where it's under a lock and it will only be executed once per call.

pwang200: Thanks for the explanation. I think it is fine as is. It is a minor performance bug and it should not be very long to get the amendment. I was just curious about your thinking.

src/ripple/protocol/impl/Feature.cpp
```diff
@@ -153,45 +168,92 @@ uint256 bitsetIndexToFeature(size_t i)
...
+//
+// The static retired amendments could hypothetically be moved so they are only
+// inside the detail::retiringAmendments() implementation.  However doing so
+// would postpone the discovery of any construction problems until the first
+// call to retiringAmendments().  By leaving these definitions at file
+// scope any run-time build problems will be revealed before main() is called.
+static uint256 const retiredFeeEscalation = getMandatoryFeature("FeeEscalation");
+static uint256 const retiredMultiSign = getMandatoryFeature("MultiSign");
+static uint256 const retiredTrustSetAuth = getMandatoryFeature("TrustSetAuth");
+static uint256 const retiredFlow = getMandatoryFeature("Flow");
```
pwang200, line 220: A minor question: why the 4 retired* variable above? They did not exist in the last version. I thought it could be for documentation. But according to step 12 of the spec, they will be removed later.

scottschurr: Those four `retired` variables are the first four features that I unconditionalized.  When I unconditionalized them I also removed the `feature` definitions because they were no longer referenced in the code base.  Then I started to research how to retire amendments and I discovered that I needed the values in order to retire them.  So I had to bring them back.

All of the `retired` features, including these four, are listed in the `retiringAmendments` list, below: https://github.com/ripple/rippled/pull/3368/files#diff-df6c689515140ea3ed30f8db7a9978b8R239-R242. That list is used to determine which amendments should be removed from the ledger and the `AmendmentTable`.

pwang200: Make sense. Thanks for the explanation!

scottschurr:
Squashed, clang-formatted, and rebased to 1.6.0-b4.

nbougalis:
Having spoken with @scottschurr, I think that we don't quite feel comfortable merging this in yet. The code is fine, but this is a "breaking change" from a user's standpoint. The spec behind this will remain open, to solicit further feedback. Please see https://github.com/xrp-community/standards-drafts/issues/14

PR #3370, opened 2020-04-22
Add delivered amount to GetAccountTransactionHistory responses

cjcobb23:
The delivered_amount field was not being populated when calling GetAccountTransactionHistory. In contrast, the delivered_amount field was being populated when calling GetTransaction. This change populates delivered_amount in the response to GetAccountTransactionHistory, and adds a unit test to make sure the results delivered by GetTransaction and GetAccountTransactionHistory match each other.

---
This change is [Reviewable](https://reviewable.io/reviews/ripple/rippled/3370)

### DIFF
--- a/src/ripple/rpc/handlers/AccountTx.cpp
+++ b/src/ripple/rpc/handlers/AccountTx.cpp
@@ -373,19 +373,21 @@ populateProtoResponse(
                             closeTime->time_since_epoch().count());
                     if (txnMeta)
                     {
-                        if (!txnMeta->hasDeliveredAmount())
+                        RPC::convert(*txnProto->mutable_meta(), txnMeta);
+                        if (!txnProto->meta().has_delivered_amount())
                         {
-                            std::optional<STAmount> amount = getDeliveredAmount(
-                                context,
-                                txn->getSTransaction(),
-                                *txnMeta,
-                                txn->getLedger());
-                            if (amount)
+                            if (auto amt = getDeliveredAmount(
+                                    context,
+                                    txn->getSTransaction(),
+                                    *txnMeta,
+                                    txn->getLedger()))
                             {
-                                txnMeta->setDeliveredAmount(*amount);
+                                RPC::convert(
+                                    *txnProto->mutable_meta()
+                                         ->mutable_delivered_amount(),
+                                    *amt);
                             }
                         }
-                        RPC::convert(*txnProto->mutable_meta(), txnMeta);
                     }
                 }
             }
--- a/src/test/rpc/Tx_test.cpp
+++ b/src/test/rpc/Tx_test.cpp
@@ -505,6 +505,25 @@ class Tx_test : public beast::unit_test::suite
         }
     };
 
+    class GrpcAccountTxClient : public GRPCTestClientBase
+    {
+    public:
+        org::xrpl::rpc::v1::GetAccountTransactionHistoryRequest request;
+        org::xrpl::rpc::v1::GetAccountTransactionHistoryResponse reply;
+
+        explicit GrpcAccountTxClient(std::string const& port)
+            : GRPCTestClientBase(port)
+        {
+        }
+
+        void
+        AccountTx()
+        {
+            status =
+                stub_->GetAccountTransactionHistory(&context, request, &reply);
+        }
+    };
+
     void
     testTxGrpc()
     {
@@ -674,22 +693,65 @@ class Tx_test : public beast::unit_test::suite
                     cmpPaymentTx(result.second.transaction(), tx);
                 }
 
-                if (ledger && !b)
-                {
-                    auto rawMeta = ledger->txRead(id).second;
-                    if (rawMeta)
-                    {
-                        auto txMeta = std::make_shared<TxMeta>(
-                            id, ledger->seq(), *rawMeta);
+                if (!ledger || b)
+                    continue;
 
-                        cmpMeta(result.second.meta(), txMeta);
-                        cmpDeliveredAmount(
-                            result.second.meta(),
-                            result.second.transaction(),
-                            txMeta,
-                            tx);
-                    }
-                }
+                auto rawMeta = ledger->txRead(id).second;
+                if (!rawMeta)
+                    continue;
+
+                auto txMeta =
+                    std::make_shared<TxMeta>(id, ledger->seq(), *rawMeta);
+
+                cmpMeta(result.second.meta(), txMeta);
+                cmpDeliveredAmount(
+                    result.second.meta(),
+                    result.second.transaction(),
+                    txMeta,
+                    tx);
+
+                auto grpcAccountTx = [&grpcPort](
+                                         uint256 const& id,
+                                         bool binary,
+                                         AccountID const& account)
+                    -> std::
+                        pair<bool, org::xrpl::rpc::v1::GetTransactionResponse> {
+                            GrpcAccountTxClient client(grpcPort);
+                            client.request.set_binary(binary);
+                            client.request.mutable_account()->set_address(
+                                toBase58(account));
+                            client.AccountTx();
+                            org::xrpl::rpc::v1::GetTransactionResponse res;
+
+                            for (auto const& tx : client.reply.transactions())
+                            {
+                                if (uint256::fromVoid(tx.hash().data()) == id)
+                                {
+                                    return {client.status.ok(), tx};
+                                }
+                            }
+                            return {false, res};
+                        };
+
+                // Compare result to result from account_tx
+                auto mentioned = tx->getMentionedAccounts();
+
+                if (!BEAST_EXPECT(mentioned.size()))
+                    continue;
+
+                auto account = *mentioned.begin();
+                auto const accountTxResult = grpcAccountTx(id, b, account);
+
+                if (!BEAST_EXPECT(accountTxResult.first))
+                    continue;
+
+                cmpPaymentTx(accountTxResult.second.transaction(), tx);
+                cmpMeta(accountTxResult.second.meta(), txMeta);
+                cmpDeliveredAmount(
+                    accountTxResult.second.meta(),
+                    accountTxResult.second.transaction(),
+                    txMeta,
+                    tx);
             }
             index++;
         }
--- a/src/ripple/rpc/impl/GRPCHelpers.cpp
+++ b/src/ripple/rpc/impl/GRPCHelpers.cpp
@@ -1567,6 +1567,9 @@ convert(org::xrpl::rpc::v1::Meta& to, std::shared_ptr<TxMeta> const& from)
     to.mutable_transaction_result()->set_result(
         transToken(from->getResultTER()));
 
+    if (from->hasDeliveredAmount())
+        convert(*to.mutable_delivered_amount(), from->getDeliveredAmount());
+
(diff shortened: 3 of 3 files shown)


src/ripple/rpc/handlers/AccountTx.cpp
```diff
@@ -373,19 +373,20 @@ populateProtoResponse(
...
-                                *txnMeta,
-                                txn->getLedger());
-                            if (amount)
-                            {
-                                txnMeta->setDeliveredAmount(*amount);
-                            }
+                            RPC::convert(
+                                *txnProto->mutable_meta()
+                                     ->mutable_delivered_amount(),
+                                *amt);
```
seelabs, line 388: I'm a little confused. The old code looks correct to me. Is the problem that `RPC::convert` does not convert the `delivered_amount`? If so, why don't we fix `RPC::convert`? (If not, can you explain what's wrong with the old code?)

src/test/rpc/Tx_test.cpp
```diff
@@ -528,6 +546,26 @@ class Tx_test : public beast::unit_test::suite
                 client.status.ok(), client.reply);
         };
 
+        auto grpcAccountTx = [&grpcPort](auto hash, auto binary, auto account) {
```
seelabs, line 549: I don't think this needs to be a template, I'd replace the `auto`s with their actual parameter types. I'd also pass the first and third parameter by `const&`

src/ripple/rpc/handlers/AccountTx.cpp
scottschurr: The way I read this is that in the old code if `DeliveredAmount` was present, then `delivered_amount` was not inserted.  I think the intention is that `delivered_amount` is always provided so client code doesn't have to check two places.  No guarantee I got that right, though.

src/test/rpc/Tx_test.cpp
```diff
@@ -688,6 +726,27 @@ class Tx_test : public beast::unit_test::suite
                             result.second.transaction(),
                             txMeta,
                             tx);
+
+                        // Compare result to result from account_tx
+                        // The transactions in later ledgers do not include A1
+                        if (index < startLegSeq + 14)
```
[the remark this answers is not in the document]
cjcobb23: I'll fix these magic numbers

src/ripple/rpc/handlers/AccountTx.cpp
cjcobb23: `RPC::convert` does not convert `delivered_amount`. I thought about adding a conversion for `delivered_amount` inside `RPC::convert(org::xrpl::rpc::v1::Meta&,std::shared_ptr<TxMeta>)`, but the issue is that `delivered_amount` is not always present in the `TxMeta` object, but is sometimes populated from the `sfAmount` field of the transaction object. I didn't like the idea of having to remember to call `TxMeta::setDeliveredAmount()` before calling `RPC::convert()`. The code needs to call the `getDeliveredAmount()` helper function anyways, so I preferred setting `delivered_amount` directly, instead of altering the `TxMeta` object and passing it to `RPC::convert()`.
I don't feel strongly about this, so am willing to change the code if the other way is clearer.

seelabs: @scottschurr When I say "the old code looks correct to me" I mean I'd expect `RPC::convert` to handle `delivered_amount`. The fix here appears to compensate for a bug in `RPC::convert`. So unless there's a reason `RPC::convert` doesn't handle this, I'd keep this code as it was before and fix `RPC::convert`.

seelabs: @cjcobb23 It does seem weird that `RPC::convert` doesn't convert `delivered_amount` even if present. I vote that we change that.

cjcobb23: fixed in `d10a4a02`; `RPC::convert` converts `delivered_amount` if present.

src/test/rpc/Tx_test.cpp
cjcobb23: fixed in `d10a4a02`

src/test/rpc/Tx_test.cpp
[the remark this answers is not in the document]
cjcobb23: Removed this magic number in `d10a4a02`

src/ripple/rpc/handlers/AccountTx.cpp
```diff
@@ -374,18 +374,19 @@ populateProtoResponse(
...
                     {
                         RPC::convert(*txnProto->mutable_meta(), txnMeta);
-
-                        auto amt = getDeliveredAmount(
-                            context,
-                            txn->getSTransaction(),
-                            *txnMeta,
-                            txn->getLedger());
-                        if (amt)
+                        if (!txnMeta->hasDeliveredAmount())
```
seelabs, line 377: Nit: the thing we _really_ care about is if `txnProto` has delivered amount. If that is easy to check, I'd rather check that.

src/test/rpc/Tx_test.cpp
```diff
@@ -688,6 +730,29 @@ class Tx_test : public beast::unit_test::suite
                             result.second.transaction(),
                             txMeta,
                             tx);
+
+                        // Compare result to result from account_tx
+                        auto mentioned = tx->getMentionedAccounts();
+
+                        if (BEAST_EXPECT(mentioned.size()))
+                        {
+                            auto a = *mentioned.begin();
```
thejohnfreeman, line 739: Can we call this `account`?

src/test/rpc/Tx_test.cpp
```diff
@@ -528,6 +547,29 @@ class Tx_test : public beast::unit_test::suite
                 client.status.ok(), client.reply);
         };
 
+        auto grpcAccountTx = [&grpcPort](
+                                 uint256 const& hash,
```
thejohnfreeman, line 551: The parameter is called `hash`, but the argument is called `id`. I prefer `id`: a hash is a function that returns a digest which is used as an identifier.

src/test/rpc/Tx_test.cpp
```diff
@@ -688,6 +730,29 @@ class Tx_test : public beast::unit_test::suite
                             result.second.transaction(),
                             txMeta,
                             tx);
+
+                        // Compare result to result from account_tx
+                        auto mentioned = tx->getMentionedAccounts();
+
+                        if (BEAST_EXPECT(mentioned.size()))
```
thejohnfreeman, line 737: My preference: reduce nesting by inverting these conditionals and turning them into early exits (`continue` in this situation). This one on line 737, but also the ones on lines 719, 722, and 744.

src/test/rpc/Tx_test.cpp
```diff
@@ -688,6 +730,29 @@ class Tx_test : public beast::unit_test::suite
...
+                        {
+                            auto a = *mentioned.begin();
+                            auto const accountTxResult =
+                                grpcAccountTx(id, b, a);
+
+                            BEAST_EXPECT(accountTxResult.first);
+                            if (accountTxResult.first)
```
thejohnfreeman, line 744: `BEAST_EXPECT` was put in the condition on line 737 but appears separately here. Could it be moved in?

cjcobb23:
Rebased and squashed

src/ripple/rpc/handlers/AccountTx.cpp
cjcobb23: changed this

cjcobb23:
> My comments are all minor preferences.

@thejohnfreeman I implemented all of your minor preferences

src/test/rpc/Tx_test.cpp
```diff
@@ -674,22 +716,42 @@ class Tx_test : public beast::unit_test::suite
...
-
-                        cmpMeta(result.second.meta(), txMeta);
-                        cmpDeliveredAmount(
-                            result.second.meta(),
-                            result.second.transaction(),
-                            txMeta,
-                            tx);
-                    }
-                }
+                if (not ledger or b)
```
thejohnfreeman, line 719: First time I've seen the alternative logical operators in this codebase. I'm not opposed, it's got a nice Pythonic feel, but I'm wondering if this was deliberate or accidental.

src/test/rpc/Tx_test.cpp
```diff
@@ -528,6 +547,29 @@ class Tx_test : public beast::unit_test::suite
                 client.status.ok(), client.reply);
         };
 
+        auto grpcAccountTx = [&grpcPort](
+                                 uint256 const& id,
+                                 auto binary,
```
seelabs, line 552: I think you missed this generic param when you fixed the other two. `auto` -> `bool`

src/test/rpc/Tx_test.cpp
```diff
@@ -528,6 +547,29 @@ class Tx_test : public beast::unit_test::suite
...
+                                 AccountID const& account) {
+            GrpcAccountTxClient client(grpcPort);
+            client.request.set_binary(binary);
+            client.request.mutable_account()->set_address(toBase58(account));
+            client.AccountTx();
+            org::xrpl::rpc::v1::GetTransactionResponse res;
+
+            for (auto tx : client.reply.transactions())
```
seelabs, line 560: `auto const&` or makes a copy

src/test/rpc/Tx_test.cpp
seelabs: I'd prefer that we stick to what the rest of the code base does.

src/test/rpc/Tx_test.cpp
```diff
@@ -528,6 +547,29 @@ class Tx_test : public beast::unit_test::suite
                 client.status.ok(), client.reply);
         };
 
+        auto grpcAccountTx = [&grpcPort](
+                                 uint256 const& id,
+                                 auto binary,
+                                 AccountID const& account) {
```
seelabs, line 553: I'd put a return type on the lambda. You should be able leave off the long return types in the body if you do and it makes it easier to see the return type.

src/test/rpc/Tx_test.cpp
cjcobb23: Ha, this was deliberate. I wanted to see what people would say. I like the alternative logical operators better than the traditional logical operators. I think they are clearer. But, consistency matters, so I can change them back.

src/test/rpc/Tx_test.cpp
```diff
@@ -528,6 +547,29 @@ class Tx_test : public beast::unit_test::suite
                 client.status.ok(), client.reply);
         };
 
+        auto grpcAccountTx = [&grpcPort](
```
scottschurr, line 550: This lambda doesn't get used until much further down in the file.  Maybe it should be moved down closer to where it is used, around line 693.

src/test/rpc/Tx_test.cpp
cjcobb23: I think these operators are actually not supported on windows

cjcobb23:
Addressed final comments. Rebased and squashed again.

cjcobb23:
@seelabs @thejohnfreeman @scottschurr Can someone approve this and add a passed label, if everything looks good? I'm finished making changes.

PR #3377, opened 2020-04-27
Adding support to build rippled packages for ubuntu 20.04

manojsdoshi:


Builds/containers/gitlab-ci/verify_head_commit.sh
```diff
@@ -1,6 +1,7 @@
 #!/usr/bin/env sh
 set -ex
 apt -y update
+DEBIAN_FRONTEND="noninteractive" apt-get -y install tzdata
```
thejohnfreeman, line 4: I think the `DEBIAN_FRONTEND=noninteractive` is only required when you pipe the output to another program, but granted, it won't hurt anything.

PR #3356, opened 2020-04-14
Update to RocksDB 6.7.3

miguelportilla:

PR #3379, opened 2020-04-27
Proposed 1.6.0-b4

manojsdoshi:
If merged this PR will:

close #3355
close #3356
close #3360
close #3328
close #3377


mDuo13:
Looks like no documentation-impacting changes in this beta, other than maybe Ubuntu 20.04 support. :+1:

PR #3382, opened 2020-04-28
[Trivial] Support boost 1.73

seelabs:
Resolves a new warning in boost 1.73 about adding the `_1`-like bind variables at global scope.

### DIFF
--- a/src/ripple/app/misc/detail/WorkSSL.h
+++ b/src/ripple/app/misc/detail/WorkSSL.h
@@ -25,8 +25,9 @@
 #include <ripple/core/Config.h>
 #include <ripple/net/HTTPClientSSLContext.h>
 #include <boost/asio/ssl.hpp>
-#include <boost/bind.hpp>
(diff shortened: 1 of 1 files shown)


src/ripple/app/misc/detail/WorkSSL.h
```diff
@@ -25,7 +25,7 @@
 #include <ripple/core/Config.h>
 #include <ripple/net/HTTPClientSSLContext.h>
 #include <boost/asio/ssl.hpp>
-#include <boost/bind.hpp>
+#include <boost/bind/bind.hpp>
```
nbougalis, line 28: This is the _only_ place where we call `boost::bind`. We should be able to just replace `boost::bind` with `std::bind` and remove the dependence.

seelabs: Great idea! Done (force pushed since it's such a small PR).

seelabs:
reformatted and re-pushed

PR #3391, opened 2020-05-06
Tune relaying of untrusted proposals & validations: 

nbougalis:
In deciding whether to relay a proposal or validation, a server would consider whether it was issued by a validator on that server's UNL.

While both trusted proposals and validations were always relayed, the code prioritized relaying of untrusted proposals over untrusted validations. While not technically incorrect, validations are generally more "valuable" because they are required during the consensus process, whereas proposals are not, strictly, required.

The commit introduces two new configuration options, allowing server operators to fine-tune the relaying behavior:

The `[relay_proposals]` option controls the relaying behavior for proposals received by this server. It has two settings: "**trusted**" and "**all**" and the default is "**trusted**".

The `[relay_validations]` options controls the relaying behavior for validations received by this server. It has two settings: "**trusted**" and "**all**" and the default is "**all**".

This change does not require an amendment as it does not affect transaction processing.


MarkusTeufelberger:
Proposals are necessary to be able to even do validations though, right?

I would prefer if both untrusted proposals and validations are by default relayed (as it is already the case). If server operators want to limit this for some reason, they now have the option to do so. Otherwise this creates incentives to collaborate more than maybe healthy or necessary with other node operators to even be able to reliably broadcast proposals.

nbougalis:
>Proposals are necessary to be able to even do validations though, right?

Even if you never see a single trusted proposal, as long as you see the validation you will continue. Of course, if *nobody else* sees a trusted proposal either, that's a problem.

>I would prefer if both untrusted proposals and validations are by default relayed (as it is already the case).

This is not already the case: currently trusted proposals and validations are always related. Untrusted validations are, at present, **never** relayed and untrusted proposal are only if they're from a cluster peer or they're proposals on top of the last ledger we consider part of the consensus set. 

The not relaying untrusted validations but relaying _some_ untrusted proposals is topsy-turvy. Again, validations are more important than proposals.

With that said, we can certainly make the default for proposals be `all` instead of `trusted`. 

>If server operators want to limit this for some reason, they now have the option to do so. Otherwise this creates incentives to collaborate more than maybe healthy or necessary with other node operators to even be able to reliably broadcast proposals.

I don't know that I agree completely, but I can certainly understand the point you're trying to make.

src/ripple/app/consensus/RCLValidations.cpp
```diff
@@ -152,7 +152,8 @@ bool
 handleNewValidation(
     Application& app,
     std::shared_ptr<STValidation> const& val,
-    std::string const& source)
+    std::string const& source,
+    bool relayUntrusted)
```
gregtatcam, line 156: `relayUntrusted` (basically is `RELAY_UNTRUSTED_VALIDATIONS`) is used to override the return of `handleNewValidation`. I think it'll be more clear just to evaluate `RELAY_UNTRUSTED_VALIDATIONS` when deciding whether relay or not. It also simplifies the function's signature and reduces the required changes.

src/ripple/app/consensus/RCLConsensus.cpp
```diff
@@ -828,12 +828,15 @@ RCLConsensus::Adaptor::validate(
...
     app_.getHashRouter().addSuppression(
         sha512Half(makeSlice(v->getSerialized())));
-    handleNewValidation(app_, v, "local");
-    Blob validation = v->getSerialized();
-    protocol::TMValidation val;
-    val.set_validation(&validation[0], validation.size());
-    // Send signed validation to all of our directly connected peers
-    app_.overlay().send(val);
+
+    if (handleNewValidation(app_, v, "local", true))
```
gregtatcam, line 832: Don't need to check for return (see other comment in RCVValidations.cpp).

src/ripple/app/misc/NetworkOPs.cpp
```diff
@@ -2482,7 +2473,8 @@ NetworkOPsImp::recvValidation(
     JLOG(m_journal.debug())
         << "recvValidation " << val->getLedgerHash() << " from " << source;
     pubValidation(val);
-    return handleNewValidation(app_, val, source);
+    return handleNewValidation(
+        app_, val, source, app_.config().RELAY_UNTRUSTED_VALIDATIONS);
```
gregtatcam, line 2477: Consider not passing `RELAY_UNTRUSTED_VALIDATIONS` and add it to the expression in `PeerImp::checkValidation()`: 

> if (app_.getOPs().recvValidation(val, std::to_string(id())) ||
            cluster() || app_.config().RELAY_UNTRUSTED_VALIDATIONS)
    {
            auto const suppression =
                sha512Half(makeSlice(val->getSerialized()));
            overlay_.relay(*packet, suppression);
    }

src/ripple/protocol/STValidation.h
```diff
@@ -88,6 +88,13 @@ class STValidation final : public STObject, public CountedObject<STValidation>
             Throw<std::runtime_error>("Invalid signature in validation");
         }
 
+        if (!isFieldPresent(sfLedgerSequence))
+        {
+            JLOG(debugLog().error()) << "Signature with no ledger sequence: "
+                                     << getJson(JsonOptions::none);
```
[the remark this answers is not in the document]
nbougalis: Oops, good catch.

src/ripple/overlay/impl/PeerImp.cpp
```diff
@@ -2460,25 +2461,15 @@ PeerImp::checkPropose(
...
-            // relay untrusted proposal
-            JLOG(p_journal_.trace()) << "relaying UNTRUSTED proposal";
-            overlay_.relay(*packet, peerPos.suppressionID());
-        }
-        else
-        {
-            JLOG(p_journal_.debug()) << "Not relaying UNTRUSTED proposal";
-        }
-    }
+        relay = app_.config().RELAY_UNTRUSTED_PROPOSALS || cluster();
```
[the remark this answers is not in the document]
nbougalis: We could. The benefits of doing so are minimal, especially if we end up not relaying the resulting validation.

src/ripple/app/consensus/RCLValidations.cpp
```diff
@@ -182,50 +182,35 @@ handleNewValidation(
...
                     " (likely from a Byzantine validator)");
         }
 
         if (outcome == ValStatus::multiple && j.warn())
         {
-            auto const seq = val->getFieldU32(sfLedgerSequence);
             dmp(j.warn(),
                 "multiple validations issued for " + to_string(seq) +
                     " (multiple validators operating with the same key?)");
         }
```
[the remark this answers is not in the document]
nbougalis: `badSeq` can occur for any number of reasons, almost all of which are bening.

src/ripple/app/consensus/RCLValidations.cpp
nbougalis: Thanks, that's nice. Adopted.

src/ripple/overlay/impl/PeerImp.cpp
```diff
@@ -2460,25 +2461,15 @@ PeerImp::checkPropose(
         return;
     }
 
+    bool relay;
+
     if (isTrusted)
-    {
-        app_.getOPs().processTrustedProposal(peerPos, packet);
-    }
+        relay = app_.getOPs().processTrustedProposal(peerPos);
```
pwang200, line 2467: `processTrustedProposal` eventually calls `Consensus<Adaptor>::peerProposalInternal` and checks things like `if (phase_ == ConsensusPhase::accepted)`. That means this proposal is too late for me. At this point, most likely the network doesn't need it anymore. But someone might. I think either way is fine. Just wanted to point it out. 

Also this part seems asymmetric. Ie. if RELAY_UNTRUSTED_PROPOSALS is "all", untrusted proposals get relayed no matter what, but trusted proposals must go through additional checks.

cfg/rippled-example.cfg
```diff
@@ -593,6 +593,23 @@
...
+#
+#   Controls the relaying behavior for proposals received by this server that
+#   are issued by validators that are not on the server's UNL.
+#
+#   Legal values are: "trusted" and "all". The default is "trusted".
+#
+#
+# [relay_validations]
+#
+#   Controls the relaying behavior for proposals received by this server that
```
pwang200, line 607: typo: proposals -> validations

src/ripple/overlay/impl/PeerImp.cpp
nbougalis: Good eyes. Yes, "all" in this case doesn't _really_ mean "all" because of this additional filtering that we perform. But as you note, proposals are, generally, not needed if the corresponding _validation_ is seen.

cfg/rippled-example.cfg
nbougalis: :+1:

nbougalis:
@pwang200:

>`sfLedgerSequence` is still marked as `soeOPTIONAL` in `STValidation::validationFormat`, but is required now. The reason is we don't want to change the API of RPC etc, right? Maybe add a comment in `STValidation::validationFormat` to explain?

I don't think there's any reason to not change it to `soeREQUIRED`. I'll double-check and do that.

pwang200:
Right, changing from `optional` to `required` is not a breaking change.

nbougalis:
@pwang200: addressed the typo and marked the `sfLedgerSequence` field as required in validations. Courtesy unit test included.

src/ripple/protocol/STValidation.h
```diff
@@ -122,15 +122,19 @@ class STValidation final : public STObject, public CountedObject<STValidation>
...
 
-        // This is a logic error because we should never assemble a
-        // validation without a ledger sequence number.
-        if (!isFieldPresent(sfLedgerSequence))
-            LogicError("Missing ledger sequence in validation");
-
         // Finally, sign the validation and mark it as trusted:
         setFlag(vfFullyCanonicalSig);
         setFieldVL(sfSignature, signDigest(pk, sk, getSigningHash()));
         setTrusted();
```
gregtatcam, line 128: Should we sign after the validation is complete?

src/ripple/protocol/STValidation.h
```diff
@@ -126,6 +126,15 @@ class STValidation final : public STObject, public CountedObject<STValidation>
...
+
+        // Check to ensure that all required fields are present.
+        for (auto const& e : validationFormat())
+        {
+            if (e.style() == soeREQUIRED && !isFieldPresent(e.sField()))
+                LogicError(
+                    "Required field '" + e.sField().getName() +
+                    "' missing from validation.");
+        }
```
gregtatcam, line 137: Should we sign after the validation is complete?

src/ripple/core/Config.h
```diff
@@ -136,6 +136,8 @@ class Config : public BasicConfig
     std::size_t NETWORK_QUORUM = 1;
 
     // Peer networking parameters
+    bool RELAY_UNTRUSTED_VALIDATIONS = true;
+    bool RELAY_UNTRUSTED_PROPOSALS = false;
```
MarkusTeufelberger, line 140:
```suggestion
    bool RELAY_UNTRUSTED_PROPOSALS = true;
```

Issue #3321, opened 2020-03-26
Issues with online_delete configuration (Version: 1.5.0-rc3)


ximinez:
## Several issues with the configuration options for `online_delete`

1. There are several undocumented features for the `[node_db]` config section which can be used to tune `online_delete` performance. These need to be documented at least in rippled-example.cfg.
    * `delete_batch` - number of records to delete per query
    * `backOff` - milliseconds to sleep between deletes
    * `age_threshold` - maximum age of the latest validated ledger before the online_delete process abends.
2. **BUG:** The `age_threshold` option is ignored. Instead, `SHAMapStore::health()` uses a `constexpr` value.
   * This needs to be fixed so the configured value is used. Since it's not functional now anyway, also change the name to `age_threshold_seconds`, and change the `SHAMapStore::ageThreshold_` variable to a `chrono::seconds`.
3. Because `backOff` is a ms value, but is functional, add a preferred `back_off_milliseconds` config option, and only document that one. Also change `SHAMapStore::backOff_` to be a `chrono::milliseconds`. Leave `backOff` available for backward compatibility for anyone using the undocumented feature.
4. (Edited to add:) Add some begin/end log messages around the execution of the `DELETE` SQL query at `trace` level to allow for more detailed analysis if desired.

## Steps to Reproduce
* Define `age_threshold` to a value different than the default 60. Something really low, like "1" would be good for this test.
* Run rippled until `online_delete` runs.
* Observe that the process doesn't abend after the LVL reaches the specified age.

## Expected Result
* The `online_delete` process abends after `age_threshold` seconds.

## Actual Result
* Process only abends if the LVL gets to be 60 seconds old.

## Environment
n/a

## Supporting Files
n/a

mtrippled:
Also consider an optional parameter to set the SQLIte database to not be synchronous. What this should do is allow delete operations to return quickly, releasing the lock. The actual data would be deleted by the operating system asynchronously. This creates a risk of data problems if the server suddenly crashes, but it would likely reduce their symptoms. It should not be done on full history servers or on systems where we care too much about keeping that data in tact. I think these 2 fixes would go a long way towards making things better for bitso the next time they run online delete.

Basically, PRAGMA SYNCHRONOUS=OFF
somewhere in the initialization code. It is per-session, so every time rippled starts up and after opening the transaction.db.

mtrippled:
[...]

PR #3397, opened 2020-05-11
Make server_info report consistent with internal

mtrippled:
evaluations for validated ledger age.

### DIFF
--- a/src/ripple/app/misc/NetworkOPs.cpp
+++ b/src/ripple/app/misc/NetworkOPs.cpp
@@ -2760,16 +2760,26 @@ NetworkOPsImp::getServerInfo(bool human, bool admin, bool counters)
             if (std::abs(closeOffset.count()) >= 60)
                 l[jss::close_time_offset] = closeOffset.count();
 
-            auto lCloseTime = lpClosed->info().closeTime;
-            auto closeTime = app_.timeKeeper().closeTime();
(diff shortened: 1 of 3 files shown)


src/ripple/app/misc/NetworkOPs.cpp
```diff
@@ -2760,16 +2760,26 @@ NetworkOPsImp::getServerInfo(bool human, bool admin, bool counters)
...
+                auto lCloseTime = lpClosed->info().closeTime;
+                auto closeTime = app_.timeKeeper().closeTime();
+                if (lCloseTime <= closeTime)
+                {
+                    using namespace std::chrono_literals;
+                    auto age = closeTime - lCloseTime;
+                    if (age < HIGH_AGE_THRESHOLD)
+                        l[jss::age] = Json::UInt(age.count());
+                    else
+                        l[jss::age] = 0;
```
ximinez, line 2781: Suggestion: Rewrite this using the `?:` operator as you did above

ximinez:
I can't imagine why, but one of the Travis MacOS builds failed, and Windows builds didn't even run. It's complaining about missing symbols, but I can't tell if that's a real problem or not. 

While I'm here, I wanted to let you know that I'm using this change as the base for my work on #3321.

mtrippled:
This duplicates part of what Ed is working on.

PR #3398, opened 2020-05-12
Add comments to protobuf files about specifying ledgers

cjcobb23:
Specifying a ledger has a lot of gotchas. For `GetTransaction`, the ledger range is only used for error handling, and searched_all is part of the error message, as opposed to the protobuf response. For `GetAccountInfo`, `SHORTCUT_VALIDATED` returns the most up to date data; however, using `SHORTCUT_VALIDATED` for `GetAccountTransactionHistory` results in only transactions from the most recently validated ledger being returned. I added some comments to point out these nuances more clearly.

---
This change is [Reviewable](https://reviewable.io/reviews/ripple/rippled/3398)

### DIFF
--- a/src/ripple/proto/org/xrpl/rpc/v1/get_transaction.proto
+++ b/src/ripple/proto/org/xrpl/rpc/v1/get_transaction.proto
@@ -23,7 +23,11 @@ message GetTransactionRequest {
     // if true, return data in binary format. defaults to false
     bool binary = 2;
 
-    // search only specified range. optional
+    // If the transaction was not found, server will report whether the entire
+    // specified range was searched. The value is contained in the error message.
+    // The error message is of the form:
+    // "txn not found. searched_all = [true,false]"
+    // If the transaction was found, this parameter is ignored.
     LedgerRange ledger_range = 3;
 }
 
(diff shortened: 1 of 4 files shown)


cjcobb23:
Requested the devx people for review, since they pointed this out.

cjcobb23:
@mduo13 gRPC and the JSON API use the same code underneath. If the comments are true for JSON, they are true for gRPC.

src/ripple/proto/org/xrpl/rpc/v1/get_transaction.proto
```diff
@@ -23,7 +23,11 @@ message GetTransactionRequest {
     // if true, return data in binary format. defaults to false
     bool binary = 2;
 
-    // search only specified range. optional
+    // If the transaction is not found, server will report whether the entire
+    // specified range is searched. The value is contained in the error message.
```
amiecorso, line 27: I might change "is" to "was"

cjcobb23: changed to was

cjcobb23:
@amiecorso @Stormtv  can you approve this if it looks good?

PR #3399, opened 2020-05-13
Find date::date in CMake package configuration file

thejohnfreeman:
Came across this problem while trying to build validator-keys-tool.


manojsdoshi:
@thejohnfreeman  the CI job is failing for your changes

thejohnfreeman:
It's failing for everyone on MacOS because c-ares 1.16.1 added a link dependency on libresolv that our CMake configuration is not picking up.

PR #3404, opened 2020-05-15
AmendmentTable improvements

scottschurr:
While working on retiring amendments I made a number of improvements to the AmendmentTable and its unit tests.  Since we're not moving forward with retiring amendments, this pull request contains the improvements to the AmendmentTable and its unit tests that I believe we want to keep.

This code should look very familiar to the reviewers.  It's just selected favorite highlights from the retiring amendments pull request.  Thanks for looking it over again.

### DIFF
--- a/src/ripple/app/misc/impl/AmendmentTable.cpp
+++ b/src/ripple/app/misc/impl/AmendmentTable.cpp
@@ -139,7 +139,7 @@ public:
 class AmendmentTableImpl final : public AmendmentTable
 {
 protected:
-    std::mutex mutex_;
+    mutable std::mutex mutex_;
 
     hash_map<uint256, AmendmentState> amendmentMap_;
     std::uint32_t lastUpdateSeq_;
@@ -157,6 +157,7 @@ protected:
 
     // True if an unsupported amendment is enabled
     bool unsupportedEnabled_;
+
     // Unset if no unsupported amendments reach majority,
     // else set to the earliest time an unsupported amendment
     // will be enabled.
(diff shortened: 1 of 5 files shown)


src/ripple/app/misc/impl/AmendmentTable.cpp
```diff
@@ -375,58 +397,45 @@ AmendmentTableImpl::enable(uint256 const& amendment)
...
+AmendmentTableImpl::doValidation(std::set<uint256> const& enabled) const
 {
     // Get the list of amendments we support and do not
     // veto, but that are not already enabled
     std::vector<uint256> amendments;
-    amendments.reserve(amendmentMap_.size());
 
     {
         std::lock_guard sl(mutex_);
+        amendments.reserve(amendmentMap_.size());
```
seelabs, line 438: Nice catch on this one.

src/ripple/app/misc/impl/AmendmentTable.cpp
```diff
@@ -297,14 +305,28 @@ AmendmentTableImpl::AmendmentTableImpl(
...
 
 AmendmentState*
-AmendmentTableImpl::get(uint256 const& amendmentHash)
+AmendmentTableImpl::get(
+    uint256 const& amendmentHash,
+    std::lock_guard<std::mutex> const& sl)
+{
+    // Forward to the const version of get.
+    return const_cast<AmendmentState*>(
+        std::as_const(*this).get(amendmentHash, sl));
```
nbougalis, line 323: What's this now? `std::as_const`? Neat.

PR #3407, opened 2020-05-18
[TRIVIAL] Adjust frequency of mtENDPOINTS messages

nbougalis:
Previously servers would send their peers `mtENDPOINTS` messages every 5 seconds. This is too often and for no real reason. Increase the frequency to 61 seconds (a prime number).

### DIFF
--- a/src/ripple/peerfinder/impl/Tuning.h
+++ b/src/ripple/peerfinder/impl/Tuning.h
@@ -129,7 +129,8 @@ enum {
 };
 
 // How often we send or accept mtENDPOINTS messages per peer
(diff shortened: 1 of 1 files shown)


src/ripple/peerfinder/impl/Tuning.h
```diff
@@ -129,7 +129,8 @@ enum {
 };
 
 // How often we send or accept mtENDPOINTS messages per peer
-static std::chrono::seconds const secondsPerMessage(5);
+// (we use a prime number of purpose)
```
MarkusTeufelberger, line 132: Which purpose?

PR #3410, opened 2020-05-21
[TRIVIAL] Disable formatting of operator&

seelabs:
Clang 10.0.0 and clang 10.0.1 disagree on formatting operator&. Disable for now.


seelabs:
Only the last commit needs to be reviewed.

thejohnfreeman:
Minor point: it's not a difference between 10.0.0 and 10.0.1, it's a difference between nightly versions of 10.0.1. They update their PPA every day it seems.

PR #3412, opened 2020-05-21
Reduce Validation and Proposal message relaying

gregtatcam:
- The motivation for the reduce relay is to reduce CPU and bandwidth cost and improve link latency.
- Every node independently goes through the peer selection rounds at some random intervals.
- MAX_PEERS are randomly selected from the pool of directly connected peers with the Validation and Proposal message count from a validator in {MESSAGE_LOW_THRESHOLD,MESSAGE_UPPER_THRESHOLD}. The peers are chosen to be the source of Validation and Proposal messages from this validator. Protocol message is sent to the rest of directly connected peers instructing them to squelch (stop relaying) the message for some random period of time in {MAX_UNSQUELCH_EXPIRE,MIN_UNSQUELCH_EXPIRE}.
- When the squelch expires or when a new peer connects to the node, another peer selection round starts.
- When a peer, which is selected to be the source of the messages for the given validator disconnects, or is idled then unsquelch message is sent to all squelched peers and another peer selection round starts.

### DIFF
--- a/src/ripple/app/misc/HashRouter.cpp
+++ b/src/ripple/app/misc/HashRouter.cpp
@@ -50,8 +50,7 @@ HashRouter::addSuppression(uint256 const& key)
 bool
 HashRouter::addSuppressionPeer(uint256 const& key, PeerShortID peer)
 {
-    auto [added, _] = addSuppressionPeerWithStatus(key, peer);
-    return added;
+    return addSuppressionPeerWithStatus(key, peer).first;
 }
 
 std::pair<bool, std::optional<Stopwatch::time_point>>
--- a/src/ripple/app/misc/HashRouter.cpp
+++ b/src/ripple/app/misc/HashRouter.cpp
@@ -50,11 +50,18 @@ HashRouter::addSuppression(uint256 const& key)
 bool
 HashRouter::addSuppressionPeer(uint256 const& key, PeerShortID peer)
 {
+    auto [added, _] = addSuppressionPeerWithStatus(key, peer);
+    return added;
+}
+
+std::pair<bool, std::optional<Stopwatch::time_point>>
+HashRouter::addSuppressionPeerWithStatus(const uint256& key, PeerShortID peer)
+{
     std::lock_guard lock(mutex_);
 
     auto result = emplace(key);
     result.first.addPeer(peer);
-    return result.second;
+    return {result.second, result.first.relayed()};
 }
 
 bool
@@ -110,14 +117,14 @@ HashRouter::setFlags(uint256 const& key, int flags)
 
 auto
 HashRouter::shouldRelay(uint256 const& key)
-    -> boost::optional<std::set<PeerShortID>>
+    -> std::optional<std::set<PeerShortID>>
 {
     std::lock_guard lock(mutex_);
 
     auto& s = emplace(key).first;
 
     if (!s.shouldRelay(suppressionMap_.clock().now(), holdTime_))
-        return boost::none;
+        return {};
 
     return s.releasePeerSet();
 }
--- a/src/ripple/overlay/Message.h
+++ b/src/ripple/overlay/Message.h
@@ -21,6 +21,7 @@
 #define RIPPLE_OVERLAY_MESSAGE_H_INCLUDED
 
 #include <ripple/overlay/Compression.h>
+#include <ripple/protocol/PublicKey.h>
 #include <ripple/protocol/messages.h>
 #include <boost/asio/buffer.hpp>
 #include <boost/asio/buffers_iterator.hpp>
@@ -55,8 +56,13 @@ public:
     /** Constructor
      * @param message Protocol message to serialize
      * @param type Protocol message type
+     * @param validator Public Key of the source validator for Validation or
+     * Proposal message. Used to check if the message should be squelched.
      */
-    Message(::google::protobuf::Message const& message, int type);
+    Message(
+        ::google::protobuf::Message const& message,
+        int type,
+        boost::optional<PublicKey> const& validator = {});
 
     /** Retrieve the packed message data. If compressed message is requested but
      * the message is not compressible then the uncompressed buffer is returned.
@@ -74,11 +80,19 @@ public:
         return category_;
     }
 
+    /** Get the validator's key */
+    boost::optional<PublicKey> const&
+    getValidatorKey() const
+    {
+        return validatorKey_;
+    }
+
 private:
     std::vector<uint8_t> buffer_;
     std::vector<uint8_t> bufferCompressed_;
     std::size_t category_;
     std::once_flag once_flag_;
+    boost::optional<PublicKey> validatorKey_;
 
     /** Set the payload header
      * @param in Pointer to the payload
--- a/src/ripple/overlay/Slot.h
+++ b/src/ripple/overlay/Slot.h
@@ -430,7 +430,10 @@ void
 Slot<clock_type>::resetCounts()
 {
     for (auto& [_, peer] : peers_)
+    {
+        (void)_;
         peer.count = 0;
+    }
 }
 
 template <typename clock_type>
@@ -685,8 +688,11 @@ Slots<clock_type>::updateSlotAndSquelch(
     {
         JLOG(journal_.debug())
             << "updateSlotAndSquelch: new slot " << Slice(validator);
-        auto [it, _] = slots_.emplace(std::make_pair(
-            validator, Slot<clock_type>(handler_, app_.journal("Slot"))));
+        auto it = slots_
+                      .emplace(std::make_pair(
+                          validator,
+                          Slot<clock_type>(handler_, app_.journal("Slot"))))
+                      .first;
         it->second.update(validator, id, type);
     }
     else
--- a/src/ripple/overlay/Slot.h
+++ b/src/ripple/overlay/Slot.h
@@ -1,0 +1,728 @@
+//------------------------------------------------------------------------------
+/*
+    This file is part of rippled: https://github.com/ripple/rippled
+    Copyright (c) 2012, 2013 Ripple Labs Inc.
+
+    Permission to use, copy, modify, and/or distribute this software for any
+    purpose  with  or without fee is hereby granted, provided that the above
+    copyright notice and this permission notice appear in all copies.
+
+    THE  SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+    WITH  REGARD  TO  THIS  SOFTWARE  INCLUDING  ALL  IMPLIED  WARRANTIES  OF
+    MERCHANTABILITY  AND  FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+    ANY  SPECIAL ,  DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+    WHATSOEVER  RESULTING  FROM  LOSS  OF USE, DATA OR PROFITS, WHETHER IN AN
+    ACTION  OF  CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+    OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+//==============================================================================
+
+#ifndef RIPPLE_OVERLAY_SLOT_H_INCLUDED
+#define RIPPLE_OVERLAY_SLOT_H_INCLUDED
+
+#include <ripple/app/main/Application.h>
+#include <ripple/basics/chrono.h>
+#include <ripple/beast/container/aged_unordered_map.h>
+#include <ripple/beast/utility/Journal.h>
+#include <ripple/overlay/Peer.h>
+#include <ripple/overlay/Squelch.h>
+#include <ripple/overlay/SquelchCommon.h>
+#include <ripple/protocol/PublicKey.h>
+#include <ripple.pb.h>
+
+#include <algorithm>
+#include <memory>
+#include <optional>
+#include <set>
+#include <tuple>
+#include <unordered_map>
+#include <unordered_set>
+
+namespace ripple {
+
+namespace squelch {
+
+template <typename clock_type>
+class Slots;
+
+/** Peer's State */
+enum class PeerState : uint8_t {
+    Counting,   // counting messages
+    Selected,   // selected to relay, counting if Slot in Counting
+    Squelched,  // squelched, doesn't relay
+};
+/** Slot's State */
+enum class SlotState : uint8_t {
+    Counting,  // counting messages
+    Selected,  // peers selected, stop counting
+};
+
+template <typename Unit, typename TP>
+Unit
+epoch(TP const& t)
+{
+    return duration_cast<Unit>(t.time_since_epoch());
+}
+
+/** Abstract class. Declares squelch and unsquelch handlers.
+ * OverlayImpl inherits from this class. Motivation is
+ * for easier unit tests to facilitate on the fly
+ * changing callbacks. */
+class SquelchHandler
+{
+public:
+    virtual ~SquelchHandler()
+    {
+    }
+    /** Squelch handler
+     * @param validator Public key of the source validator
+     * @param id Peer's id to squelch
+     * @param duration Squelch duration in seconds
+     */
+    virtual void
+    squelch(PublicKey const& validator, Peer::id_t id, std::uint32_t duration)
+        const = 0;
+    /** Unsquelch handler
+     * @param validator Public key of the source validator
+     * @param id Peer's id to unsquelch
+     */
+    virtual void
+    unsquelch(PublicKey const& validator, Peer::id_t id) const = 0;
+};
+
+/**
+ * Slot is associated with a specific validator via validator's public key.
+ * Slot counts messages from a validator, selects peers to be the source
+ * of the messages, and communicates the peers to be squelched. Slot can be
+ * in the following states: 1) Counting. This is the peer selection state
+ * when Slot counts the messages and selects the peers; 2) Selected. Slot
+ * doesn't count messages in Selected state. A message received from
+ * unsquelched, disconnected peer, or idling peer may transition Slot to
+ * Counting state.
+ */
+template <typename clock_type>
+class Slot final
+{
+private:
+    friend class Slots<clock_type>;
+    using id_t = Peer::id_t;
+    using time_point = typename clock_type::time_point;
+
+    /** Constructor
+     * @param journal Journal for logging
+     * @param handler Squelch/Unsquelch implementation
+     */
+    Slot(SquelchHandler const& handler, beast::Journal journal)
+        : reachedThreshold_(0)
+        , lastSelected_(clock_type::now())
+        , state_(SlotState::Counting)
+        , handler_(handler)
+        , journal_(journal)
+    {
+    }
+
+    /** Update peer info. If the message is from a new
+     * peer or from a previously expired squelched peer then switch
+     * the peer's and slot's state to Counting. If time of last
+     * selection round is > 2 * MAX_UNSQUELCH_EXPIRE then switch the slot's
+     * state to Counting. If the number of messages for the peer
+     * is > MIN_MESSAGE_THRESHOLD then add peer to considered peers pool.
+     * If the number of considered peers who reached MAX_MESSAGE_THRESHOLD is
+     * MAX_SELECTED_PEERS then randomly select MAX_SELECTED_PEERS from
+     * considered peers, and call squelch handler for each peer, which is not
+     * selected and not already in Squelched state. Set the state for those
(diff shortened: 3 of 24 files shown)


src/ripple/overlay/impl/PeerImp.cpp
```diff
@@ -2309,6 +2313,23 @@ PeerImp::onMessage(std::shared_ptr<protocol::TMGetObjectByHash> const& m)
     }
 }
 
+void
+PeerImp::onMessage(std::shared_ptr<protocol::TMSquelch> const& m)
+{
+    auto validator = m->validatorpubkey();
+    PublicKey key(Slice(validator.data(), validator.size()));
```
pwang200, line 2320: Constructing `key` may throw if `validator` has bad data. Please see this example:
https://github.com/ripple/rippled/blob/2d79e593ee1dc54092d25ffe9d3d461037f76965/src/ripple/overlay/impl/PeerImp.cpp#L1237

src/ripple/overlay/impl/OverlayImpl.h
```diff
@@ -364,6 +373,26 @@ class OverlayImpl : public Overlay
     void
     lastLink(std::uint32_t id);
 
+    /** Updates message count for validator/peer. Sends TMSquelch if the number
+     * of messages for N peers reaches threshold T.
```
pwang200, line 377: "Sends TMSquelch if the number of messages for N peers reaches threshold T." 
Is the number of messages >= T or N >= T? Or there are two thresholds?

src/ripple/overlay/impl/PeerImp.cpp
```diff
@@ -2499,9 +2524,13 @@ PeerImp::checkValidation(
         if (app_.getOPs().recvValidation(val, std::to_string(id())) ||
             cluster())
         {
+            overlay_.checkForSquelch(
+                val->getSignerPublic(),
+                shared_from_this(),
+                protocol::mtVALIDATION);
```
pwang200, line 2530: For proposal, `overlay_.checkForSquelch` is called no matter if the node decides to forward or not. For validation, it is called only if the node decides to forward. Why the difference?

src/ripple/overlay/Slot.h
```diff
@@ -0,0 +1,607 @@
...
+
+    if (now - lastSelected_ > 2 * MAX_UNSQUELCH_EXPIRE)
+    {
+        initCounting();
+        return;
+    }
+
+    if (
+        considered_.size() > MAX_SELECTED_PEERS &&
+        considered_.size() >= 3 * peers_.size() / 4)
```
pwang200, line 251: discussed with @gregtatcam 
Checking `considered_.size() >= 3 * peers_.size() / 4` could delay peer selection and result in a saw tooth traffic pattern with high point being 75%.

src/ripple/overlay/Slot.h
```diff
@@ -0,0 +1,607 @@
...
+    F&& f)
+{
+    auto now = clock_type::now();
+    auto it = peers_.find(id);
+    // First message from this peer
+    if (it == peers_.end())
+    {
+        peers_.emplace(std::make_pair(
+            id,
+            std::move(PeerInfo{peerPtr, PeerState::Counting, 0, now, now})));
```
seelabs, line 220: `move` is not needed here.

src/ripple/overlay/Slot.h
```diff
@@ -0,0 +1,607 @@
...
+    const time_point&
+    getLastSelected() const
+    {
+        return lastSelected_;
+    }
+
+    /** Return number of peers in state/not in state */
+    template <typename Comp = std::equal_to<>>
+    std::uint16_t
+    inState(PeerState state, Comp comp = {}) const;
```
seelabs, line 131: The `comp` parameter here is a consequence of making this class generic. If we could solve that we could get rid of this parameter as well (here and below).

src/ripple/overlay/Squelch.h
```diff
@@ -0,0 +1,121 @@
...
+
+template <typename clock_type>
+bool
+Squelch<clock_type>::isSquelched(PublicKey const& validator)
+{
+    auto now = clock_type::now();
+
+    if (squelched_.find(validator) == squelched_.end())
+        return false;
+    else if (squelched_[validator] > now)
```
seelabs, line 100: The `find` function has already done this lookup. Instead of doing it again here, save the iterator from `find` and use that to assign the value.

src/ripple/overlay/Squelch.h
```diff
@@ -0,0 +1,121 @@
...
+        return true;
+
+    squelched_.erase(validator);
```
seelabs, line 103: Ditto here: Save it iterator from the `find` function to avoid looking up the value again.

src/ripple/overlay/Slot.h
```diff
@@ -0,0 +1,607 @@
...
+/**
+ * Slot is associated with a specific validator via validator's public key.
+ * Slot counts messages from a validator, selects peers to be the source
+ * of the messages, and communicates the peers to be squelched. Slot can be
+ * in the following states: 1) Counting. This is the peer selection state
+ * when Slot counts the messages and selects the peers; 2) Selected. Slot
+ * doesn't count messages in this state but a received message
+ * may transition Slot to Counting state.
+ */
+template <typename Peer, typename clock_type>
```
seelabs, line 63: I understand the motivation to make this class generic - it help with testing. However, making this generic makes the code much harder to navigate and understand. We really need to think hard if there's another way to test the code that doesn't necessitate making this class generic.

This is not just a nit to me. Making this class use the concrete peer type, getting rid of all the generic callback functions, and getting rid of the comparator parameter would be a huge improvement in code understandably.

I do know we did something similar in the consensus code. Please think hard if there's another, better solution to the testing problem;

src/ripple/overlay/Slot.h
gregtatcam: Changed the logic to start the selection when MAX_SELECTED_PEERS reach MAX_MESSAGE_THRESHOLD (20) and include into the selection pool all peers that reached MIN_MESSAGE_THRESHOLD (17).

src/ripple/overlay/Slot.h
gregtatcam: We can  implement `Peer` class in unit tests with most of the methods simply returning some value. This is not going to be the test framework or a step towards implementing the test framework but a way to avoid  the extra allocation in (4). We can then pass `std::weak_ptr<Peer>` to all applicable slot methods as we do today with `Peer` being a concrete type so no template is required. We will have abstract class `SquelchHandler` with `squelch` and `unsquelch` methods. The class is going to be implemented in `Overlay` and `UnitTest`. Reference to the base class is going to be passed to `Slots` and `Slot` constructors and `squelch` and `unsquelch` called as needed. `UnitTest` can overwrite the handling on a fly providing flexibility of different handlers as needed. Does this work?

src/ripple/overlay/Slot.h
```diff
@@ -0,0 +1,635 @@
...
+        std::weak_ptr<Peer> const peer;  // peer's weak_ptr
+        PeerState state;                 // peer's state
+        std::size_t count;               // message count
+        time_point expire;               // squelch expiration time
+        time_point lastMessage;          // time last message received
+    };
+    std::unordered_map<id_t, PeerInfo> peers_;  // peer's data
+    std::unordered_set<id_t> considered_;       // pool of peers considered
+                                                // as the source of messages
+                                                // from validator - peers
```
seelabs, line 212: Putting very short comments to the right is OK, but when the comment on the right spans several lines I'd rather put the comment on it's own line above the code it's commenting, where it would only take one line instead of three. If you prefer this style then no change needed. Mine is only a preference.

src/ripple/overlay/Slot.h
```diff
@@ -0,0 +1,635 @@
...
+            if (it == slots_.end())
+            {
+                auto [it, _] = slots_.emplace(std::make_pair(
+                    validator, std::move(Slot<clock_type>(handler_))));
+                return it;
+            }
+            else
+                return it;
+        }()
+            ->second;
```
seelabs, line 595: Consider folding this `->second` into the lambda.

src/ripple/overlay/Slot.h
```diff
@@ -0,0 +1,635 @@
...
+    void
+    deletePeer(id_t id, bool erase);
+
+    hash_map<PublicKey, Slot<clock_type>> slots_;
+    SquelchHandler const& handler_;  // squelch/unsquelch handler
+};
+
+template <typename clock_type>
+void
+Slots<clock_type>::checkForSquelch(
```
seelabs, line 577: A function called `checkForX` sounds like a pure function that either returns a `bool` or an `X`. Consider re-naming this.

src/ripple/overlay/Slot.h
```diff
@@ -0,0 +1,635 @@
...
+    std::weak_ptr<Peer> const& peerPtr,
+    protocol::MessageType type)
+{
+    auto& peer =
+        [&]() {
+            auto it = slots_.find(validator);
+            if (it == slots_.end())
+            {
+                auto [it, _] = slots_.emplace(std::make_pair(
+                    validator, std::move(Slot<clock_type>(handler_))));
```
seelabs, line 589: No need for a `move` here.

src/ripple/overlay/Slot.h
```diff
@@ -0,0 +1,635 @@
...
+    // First message from this peer
+    if (it == peers_.end())
+    {
+        peers_.emplace(std::make_pair(
+            id, PeerInfo{peerPtr, PeerState::Counting, 0, now, now}));
+        initCounting();
+        return;
+    }
+    // Message from a peer with expired squelch
+    else if (
```
seelabs, line 257: No need for the else; the previous `if` returns.

src/ripple/overlay/Slot.h
```diff
@@ -0,0 +1,635 @@
...
+    {
+        auto const& it = slots_.find(validator);
+        if (it != slots_.end())
+            return it->second.notInState(state);
+        return {};
+    }
+
+    /** Return true if Slot is in State */
+    template <typename Comp = std::equal_to<>>
+    boost::optional<bool>
```
seelabs, line 518: `optional<bool>` is highly error prone to use, and the callers don't appear to need to distinguish between the two levels for `false`. Consider changing the return type to just `bool` (or a struct if it really is necessary to distinguish between the two `false`'s).

src/ripple/overlay/Slot.h
```diff
@@ -0,0 +1,635 @@
...
+            auto it = std::next(considered_.begin(), i);
+            auto id = *it;
+            considered_.erase(it);
+            if (peers_.find(id) == peers_.end())
+            {
+                // TBD have to log this
+                continue;
+            }
+            // TBD : should we be stricter and include < x% of IDLED? (50%?)
+            if (now - peers_[id].lastMessage < IDLED)
```
seelabs, line 305: We already searched for `id` above. Instead of looking up again save an iterator above and use the iterator.

src/ripple/overlay/Slot.h
```diff
@@ -0,0 +1,635 @@
...
+};
+
+template <typename Unit, typename TP>
+Unit
+epoch(TP const& t)
+{
+    return duration_cast<Unit>(t.time_since_epoch());
+}
+
+/** Abstract class. Declares squelch and unsquelch handlers */
```
seelabs, line 55: I'd note that OverlayImpl inherits from this class and the motivation for this is for easier unit tests.

src/ripple/overlay/Slot.h
```diff
@@ -0,0 +1,635 @@
...
+
+#ifndef RIPPLE_OVERLAY_SLOT_H_INCLUDED
+#define RIPPLE_OVERLAY_SLOT_H_INCLUDED
+
+#include <ripple/overlay/Peer.h>
+#include <ripple/overlay/Squelch.h>
+#include <ripple/overlay/SquelchCommon.h>
+#include <ripple/protocol/PublicKey.h>
+#include <ripple.pb.h>
+
```
seelabs, line 28: Some includes are missing:
```
#include <boost/optional.hpp>

#include <algorithm>
#include <memory>
#include <set>
#include <tuple>
#include <unordered_map>
#include <unordered_set>

```

src/ripple/overlay/Slot.h
```diff
@@ -0,0 +1,635 @@
...
+ * Slot is associated with a specific validator via validator's public key.
+ * Slot counts messages from a validator, selects peers to be the source
+ * of the messages, and communicates the peers to be squelched. Slot can be
+ * in the following states: 1) Counting. This is the peer selection state
+ * when Slot counts the messages and selects the peers; 2) Selected. Slot
+ * doesn't count messages in Selected state. A message received from
+ * unsquelched, disconnected peer, or idling peer may transition Slot to
+ * Counting state.
+ */
+template <typename clock_type>
```
seelabs, line 91: Now that this class is simplified to take a single template parameter, consider moving the implementation to a `.cpp` file and explicitly instantiating the clocks we need for linking.

src/ripple/overlay/Message.h
```diff
@@ -21,6 +21,7 @@
 #define RIPPLE_OVERLAY_MESSAGE_H_INCLUDED
```
seelabs, line 21: This file needs to be reformatted with clang format.

src/ripple/overlay/Slot.h
seelabs: We'd have to move the ManualClock to a common header file, and we normally would not include a test file like that in the main code (tho I think we have a good reason in this case). But let's not worry about that now, we can do that later in another PR if we want.

src/ripple/overlay/Slot.h
```diff
@@ -0,0 +1,636 @@
...
+#define RIPPLE_OVERLAY_SLOT_H_INCLUDED
+
+#include <ripple/overlay/Peer.h>
+#include <ripple/overlay/Squelch.h>
+#include <ripple/overlay/SquelchCommon.h>
+#include <ripple/protocol/PublicKey.h>
+#include <algorithm>
+#include <memory>
+#include <optional>
+#include <ripple.pb.h>
```
seelabs, line 30: This include snuck in with the wrong files. While you're there, we usually put a space between the different include groups. I.e
```
#include <ripple/overlay/SquelchCommon.h>

#include <algorithm>
#include <memory>
```

src/test/overlay/reduce_relay_test.cpp
```diff
@@ -0,0 +1,1301 @@
...
+#include <ripple/beast/unit_test.h>
+#include <ripple/overlay/Message.h>
+#include <ripple/overlay/Peer.h>
+#include <ripple/overlay/Slot.h>
+#include <ripple/protocol/SecretKey.h>
+#include <boost/optional.hpp>
+#include <boost/thread.hpp>
+#include <numeric>
+#include <optional>
+#include <ripple.pb.h>
```
seelabs, line 29: Standard includes should come last. We need to move <ripple.pb.h> up. We also usually leave a space between the different include groups (usually groups: ripple boost std).

src/ripple/overlay/Slot.h
```diff
@@ -0,0 +1,636 @@
...
+    std::optional<std::uint16_t>
+    inState(PublicKey const& validator, PeerState state) const
+    {
+        auto const& it = slots_.find(validator);
+        if (it != slots_.end())
+            return it->second.inState(state);
+        return {};
+    }
+
+    /** Return number of peers in state in state */
```
seelabs, line 514: Typo: I think you mean "not in state `state`"

src/ripple/overlay/Slot.h
```diff
@@ -0,0 +1,636 @@
...
+        {
+            auto i =
+                considered_.size() == 1 ? 0 : rand_int(considered_.size() - 1);
+            auto it = std::next(considered_.begin(), i);
+            auto id = *it;
+            considered_.erase(it);
+            auto const& itpeers = peers_.find(id);
+            if (itpeers == peers_.end())
+            {
+                // TBD have to log this
```
pwang200, line 309: yeah, should log this.

src/ripple/overlay/Slot.h
```diff
@@ -0,0 +1,636 @@
...
+                continue;
+            }
+            // TBD : should we be stricter and include < x% of IDLED? (50%?)
```
pwang200, line 312: I think it is good as is.

src/ripple/overlay/Slot.h
```diff
@@ -0,0 +1,636 @@
...
+    void
+    unsquelch(id_t id);
+
+    /** Check if peers stopped relaying messages
+     * and if slots stopped receiving messages from the validator.
+     */
+    void
+    deleteIdlePeers();
+
+    /** Return number of peers in state in state */
```
pwang200, line 504: typo: two 'in state', here and below

src/ripple/overlay/Squelch.h
```diff
@@ -0,0 +1,122 @@
...
+{
+    if (squelch)
+    {
+        squelched_[validator] = [squelchDuration]() {
+            seconds duration = seconds(squelchDuration);
+            return clock_type::now() +
+                ((duration >= MIN_UNSQUELCH_EXPIRE &&
+                  duration <= MAX_UNSQUELCH_EXPIRE)
+                     ? duration
+                     : getSquelchDuration());  // TBD should we disconnect if
```
pwang200, line 84: I don't think we should disconnect.

src/ripple/overlay/Slot.h
```diff
@@ -0,0 +1,750 @@
...
+ * changing callbacks. */
+class SquelchHandler
+{
+public:
+    virtual ~SquelchHandler()
+    {
+    }
+    /** Squelch handler
+     * @param validator Public key of the source validator
+     * @param peer Peer to squelch
```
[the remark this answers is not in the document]
gregtatcam: Done

src/ripple/overlay/Slot.h
```diff
@@ -0,0 +1,750 @@
...
+     * @param duration Squelch duration in seconds
+     */
+    virtual void
+    squelch(
+        PublicKey const& validator,
+        Peer::id_t const id,
+        std::uint32_t duration) const = 0;
+    /** Unsquelch handler
+     * @param validator Public key of the source validator
+     * @param peer Peer to unsquelch
```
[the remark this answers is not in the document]
gregtatcam: Done

src/ripple/overlay/Slot.h
```diff
@@ -0,0 +1,750 @@
...
+private:
+    friend class Slots<clock_type>;
+    using id_t = Peer::id_t;
+    using time_point = typename clock_type::time_point;
+
+    /** Constructor
+     * @param app Application reference
+     * @param handler Squelch/Unsquelch implementation
+     */
+    Slot(Application& app, SquelchHandler const& handler)
```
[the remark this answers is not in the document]
seelabs: We usually make `journal` the last parameter and pass by value.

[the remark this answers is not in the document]
gregtatcam: Done.

src/ripple/app/misc/HashRouter.cpp
```diff
@@ -49,12 +49,19 @@ HashRouter::addSuppression(uint256 const& key)
 
 bool
 HashRouter::addSuppressionPeer(uint256 const& key, PeerShortID peer)
+{
+    auto [added, _] = addSuppressionPeerWithStatus(key, peer);
+    return added;
+}
+
+std::pair<bool, std::optional<Stopwatch::time_point>>
+HashRouter::addSuppressionPeerWithStatus(const uint256& key, PeerShortID peer)
```
nbougalis, line 58: The `HashRouter` API was already confusing enough. It's now become a little _too_ confusing. I don't have an improvement to offer, but we need to think about this.

src/ripple/overlay/impl/PeerImp.cpp
```diff
@@ -2324,6 +2353,45 @@ PeerImp::onMessage(std::shared_ptr<protocol::TMGetObjectByHash> const& m)
     }
 }
 
+void
+PeerImp::onMessage(std::shared_ptr<protocol::TMSquelch> const& m)
+{
+    if (!m->has_validatorpubkey())
+    {
+        JLOG(p_journal_.debug()) << "onMessage: TMSquelch, missing public key";
```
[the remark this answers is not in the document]
gregtatcam: Done

src/ripple/overlay/Message.h
```diff
@@ -56,7 +57,10 @@ class Message : public std::enable_shared_from_this<Message>
      * @param message Protocol message to serialize
      * @param type Protocol message type
      */
-    Message(::google::protobuf::Message const& message, int type);
+    Message(
+        ::google::protobuf::Message const& message,
+        int type,
+        boost::optional<PublicKey> validator = {});
```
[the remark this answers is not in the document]
gregtatcam: It is documented. This is what it currently states:
    `@param validator Public Key of the source validator for Validation or
     Proposal message. Used to check if the message should be squelched.`

PR #3419, opened 2020-05-29
Improve compression support:

nbougalis:
* Optimize parsing of compressed message headers
* Enforce protocol-defined message size maxima
* Update comments

### DIFF
--- a/src/ripple/overlay/Compression.h
+++ b/src/ripple/overlay/Compression.h
@@ -31,7 +31,9 @@ namespace compression {
 std::size_t constexpr headerBytes = 6;
 std::size_t constexpr headerBytesCompressed = 10;
 
-enum class Algorithm : std::uint8_t { None = 0x00, LZ4 = 0x01 };
+// All values other than 'none' must have the high bit. The low order four bits
+// must be 0.
+enum class Algorithm : std::uint8_t { None = 0x00, LZ4 = 0x90 };
 
 enum class Compressed : std::uint8_t { On, Off };
 
--- a/src/ripple/overlay/impl/Message.cpp
+++ b/src/ripple/overlay/impl/Message.cpp
@@ -143,7 +143,7 @@ Message::setHeader(
     std::uint8_t* in,
     std::uint32_t payloadBytes,
     int type,
-    Algorithm comprAlgorithm,
+    Algorithm compression,
     std::uint32_t uncompressedBytes)
 {
     auto h = in;
@@ -161,10 +161,10 @@ Message::setHeader(
     *in++ = static_cast<std::uint8_t>((type >> 8) & 0xFF);
     *in++ = static_cast<std::uint8_t>(type & 0xFF);
 
-    if (comprAlgorithm != Algorithm::None)
+    if (compression != Algorithm::None)
     {
         pack(in, uncompressedBytes);
-        *h |= 0x80 | (static_cast<uint8_t>(comprAlgorithm) << 4);
+        *h |= static_cast<std::uint8_t>(compression);
     }
 }
 
--- a/src/ripple/overlay/impl/Message.cpp
+++ b/src/ripple/overlay/impl/Message.cpp
@@ -109,16 +109,35 @@ Message::compress()
 }
 
 /** Set payload header
- * Uncompressed message header
- * 47-42    Set to 0
- * 41-16    Payload size
- * 15-0	    Message Type
- * Compressed message header
- * 79       Set to 0, indicates the message is compressed
- * 78-76    Compression algorithm, value 1-7. Set to 1 to indicate LZ4
- * compression 75-74    Set to 0 73-48    Payload size 47-32	Message Type
- * 31-0     Uncompressed message size
- */
+
+    The header is a variable-sized structure that contains information about
+    the type of the message and the length and encoding of the payload.
+
+    The first 6 bits determine whether a message is compressed or uncompressed
+    and determine the type of compression:
+
+    All multi-byte values are represented in big endian.
+
+    For uncompressed messages (6 bytes), numbering bits from left to right:
+
+        - The first 6 bits are set to 0.
+        - The next 26 bits represent the payload size.
+        - The remaining 16 bits represent the message type.
+
+    For compressed messages (10 bytes), numbering bits from left to right:
+
+        - The first 32 bits, together, represent the compression algorithm
+          and payload size:
+            - The first bit is set to 1 to indicate the message is compressed.
+            - The next 3 bits indicate the compression algorithm.
+            - The remaining 28 bits represent the payload size.
+        - The next 16 bits represent the message type.
+        - The remaining 32 bits are the uncompressed message size.
+
+    @note While nominally a part of the wire protocol, the framing is subject
+          to change; future versions of the code may negotiate the use of
+          substantially different framing.
+*/
 void
 Message::setHeader(
     std::uint8_t* in,
--- a/src/ripple/overlay/impl/Message.cpp
+++ b/src/ripple/overlay/impl/Message.cpp
@@ -113,8 +113,9 @@ Message::compress()
     The header is a variable-sized structure that contains information about
     the type of the message and the length and encoding of the payload.
 
-    The first 6 bits determine whether a message is compressed or uncompressed
-    and determine the type of compression:
+    The first bit determines whether a message is compressed or uncompressed;
+    for compressed messages, the next three bits identify the compression
+    algorithm.
 
     All multi-byte values are represented in big endian.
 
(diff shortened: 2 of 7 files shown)


src/ripple/overlay/impl/PeerImp.h
```diff
@@ -428,6 +428,12 @@ class PeerImp : public Peer,
     boost::optional<hash_map<PublicKey, ShardInfo>>
     getPeerShardInfo() const;
 
+    bool
+    compressionEnabled () const override
```
seelabs, line 432: Clang format complains about this: no space before paran

src/ripple/overlay/impl/ProtocolMessage.h
```diff
@@ -226,23 +263,35 @@ invokeProtocolMessage(Buffers const& buffers, Handler& handler)
...
     if (!header)
         return result;
 
     // We implement a maximum size for protocol messages. Sending a message
     // whose size exceeds this may result in the connection being dropped. A
     // larger message size may be supported in the future or negotiated as
     // part of a protocol upgrade.
-    if (header->payload_wire_size > megabytes(64))
+    if (header->payload_wire_size > megabytes(64) ||
+            header->uncompressed_size > megabytes(64))
```
seelabs, line 281: Nice catch on this one.

Also, nit: Clang format complains about this: no extra indent for second line of the condition.

src/test/overlay/compression_test.cpp
```diff
@@ -84,21 +84,14 @@ class compression_test : public beast::unit_test::suite
         std::shared_ptr<T> proto,
         protocol::MessageType mt,
         uint16_t nbuffers,
-        const char* msg,
-        bool log = false)
+        std::string msg)
```
seelabs, line 87: `const&`

seelabs:
These changes are non-trivial enough that we should probably assign another reviewer as well.

src/ripple/overlay/Compression.h
```diff
@@ -31,7 +31,9 @@ namespace compression {
 std::size_t constexpr headerBytes = 6;
 std::size_t constexpr headerBytesCompressed = 10;
 
-enum class Algorithm : std::uint8_t { None = 0x00, LZ4 = 0x01 };
+// All values other than 'none' must have the high bit. The low order four bits
+// must be 0.
+enum class Algorithm : std::uint8_t { None = 0x00, LZ4 = 0x90 };
```
MarkusTeufelberger, line 36: Why the change from `0x01` to `0x90`?

src/ripple/overlay/impl/Message.cpp
```diff
@@ -109,22 +109,41 @@ Message::compress()
...
+
+    The first 6 bits determine whether a message is compressed or uncompressed
+    and determine the type of compression:
+
+    All multi-byte values are represented in big endian.
+
+    For uncompressed messages (6 bytes), numbering bits from left to right:
+
+        - The first 6 bits are set to 0.
+        - The next 26 bits represent the payload size.
```
gregtatcam, line 124: Strictly speaking the payload size is 28 bits because we only set 4 bits to 0. Perhaps we should set 6 bits to 0? And use the same rule for the compressed message payload  size and uncompressed message size.

src/ripple/overlay/impl/ProtocolMessage.h
```diff
@@ -120,51 +120,88 @@ buffersBegin(BufferSequence const& bufs)
...
     using namespace ripple::compression;
-    auto iter = buffersBegin(bufs);
 
     MessageHeader hdr;
-    auto const compressed = (*iter & 0x80) == 0x80;
+    auto iter = buffersBegin(bufs);
 
     // Check valid header
-    if ((*iter & 0xFC) == 0 || compressed)
+    if (*iter & 0x80)
```
gregtatcam, line 143: Should we also check here that `(*iter & 0x0C) == 0`?

src/ripple/overlay/impl/Message.cpp
```diff
@@ -109,22 +109,41 @@ Message::compress()
...
 void
 Message::setHeader(
     std::uint8_t* in,
     std::uint32_t payloadBytes,
     int type,
-    Algorithm comprAlgorithm,
+    Algorithm compression,
     std::uint32_t uncompressedBytes)
 {
     auto h = in;
```
gregtatcam, line 149: Should we set 6 bits to 0 in `pack`? And the comment should be updated because we use `pack` to pack the payload size and the uncompressed message size. The comment doesn't apply in the latter case.

src/ripple/overlay/impl/ProtocolMessage.h
nbougalis: Done.

src/ripple/overlay/impl/Message.cpp
nbougalis: We can't do so willy-nilly: it may truncate a message. Of course, any such message would be over 64MB, so we should never send it, but that's a larger change that I'd like, at this point.

src/ripple/overlay/impl/ProtocolMessage.h
```diff
@@ -120,51 +120,94 @@ buffersBegin(BufferSequence const& bufs)
...
 
+/** Parse a message header
+ * @return a seated optional if the message header was successfully
+ *         parsed. An unseated optional otherwise, in which case
+ *         @param ec contains more information:
+ *         - set to `errc::success` if not enough bytes were present
+ *         - set to `errc::no_message` if a valid header was not present
+ */
 template <class BufferSequence>
 boost::optional<MessageHeader>
```
cjcobb23, line 131: Can we change this to `std::optional`? I think we are trying to migrate away from `boost::optional`

src/ripple/overlay/impl/ProtocolMessage.h
cjcobb23: What do you think of putting the bitwise comparison inside of a named lambda, where the name describes what is being checked? I don't really like bitwise operations in a conditional directly, because it doesn't read that easily. It always takes me a second to figure out whats going on, and I think sometimes subtle bugs can occur. Remember this: https://github.com/ripple/rippled/issues/3417? At the least, I think we should add a comment saying exactly what is being checked.

cjcobb23: Actually, after reading through the whole function, I think we should abstract away these bitwise operations to a class, or a bunch of helper functions. Bitwise operations are pretty low level, and I think it's hard to read the code and easy to mess up. We could compromise on more judicious use of comments (I would put a clear comment before every bitwise operation), but I'd really prefer named functions; it's pretty easy for a developer to change the code and forget to update the comment, but it's harder to change the function implementation and forget to update the function name.

src/ripple/overlay/impl/ProtocolMessage.h
```diff
@@ -120,51 +120,94 @@ buffersBegin(BufferSequence const& bufs)
...
-parseMessageHeader(BufferSequence const& bufs, std::size_t size)
+parseMessageHeader(
```
cjcobb23, line 132: What do you think of putting the bitwise comparison inside of a named lambda, where the name describes what is being checked? I don't really like bitwise operations in a conditional directly, because it doesn't read that easily. It always takes me a second to figure out whats going on, and I think sometimes subtle bugs can occur. Remember this: https://github.com/ripple/rippled/issues/3417? At the least, I think we should add a comment saying exactly what is being checked.

cjcobb23, line 132: What do you think of putting the bitwise comparison inside of a named lambda, where the name describes what is being checked? I don't really like bitwise operations in a conditional directly, because it doesn't read that easily. It always takes me a second to figure out whats going on, and I think sometimes subtle bugs can occur. Remember this: https://github.com/ripple/rippled/issues/3417? At the least, I think we should add a comment saying exactly what is being checked.

cjcobb23: Actually, after reading through the whole function, I think we should abstract away these bitwise operations to a class, or a bunch of helper functions. Bitwise operations are pretty low level, and I think it's hard to read the code and easy to mess up. We could compromise on more judicious use of comments (I would put a clear comment before every bitwise operation), but I'd really prefer named functions; it's pretty easy for a developer to change the code and forget to update the comment, but it's harder to change the function implementation and forget to update the function name.

src/ripple/overlay/impl/ProtocolMessage.h
nbougalis: We can. Shouldn't be a problem.

src/ripple/overlay/impl/ProtocolMessage.h
nbougalis: I hear what you're saying, but I think that adding additional lambas would only add confusion. Better comments would probably help; I'll try and follow-up.

cjcobb23:
Looks good to me

PR #3421, opened 2020-05-29
Cleanup SHAMap and simplify interfaces

nbougalis:
* Improve error reporting (more readable exception messages)
* Reduce function complexity (split oversized function to smaller pieces)
* Reduce code duplication
* Reduce buffer copying

### DIFF
--- a/src/ripple/app/ledger/impl/InboundLedger.cpp
+++ b/src/ripple/app/ledger/impl/InboundLedger.cpp
@@ -1147,12 +1147,14 @@ InboundLedger::processData(
 
         // Verify node IDs and data are complete
         for (auto const& node : packet.nodes())
+        {
             if (!node.has_nodeid() || !node.has_nodedata())
             {
                 JLOG(m_journal.warn()) << "Got bad node";
                 peer->charge(Resource::feeInvalidRequest);
                 return -1;
             }
+        }
 
         SHAMapAddNode san;
         receiveNode(packet, san);
--- a/src/ripple/app/ledger/impl/InboundLedger.cpp
+++ b/src/ripple/app/ledger/impl/InboundLedger.cpp
@@ -898,33 +898,40 @@ InboundLedger::takeTxNode(
         return true;
     }
 
-    auto nodeIDit = nodeIDs.cbegin();
-    auto nodeDatait = data.begin();
-    TransactionStateSF filter(
-        mLedger->txMap().family().db(), app_.getLedgerMaster());
-
-    while (nodeIDit != nodeIDs.cend())
+    try
     {
-        if (nodeIDit->isRoot())
-        {
-            san += mLedger->txMap().addRootNode(
-                SHAMapHash{mLedger->info().txHash},
-                makeSlice(*nodeDatait),
-                snfWIRE,
-                &filter);
-            if (!san.isGood())
-                return false;
-        }
-        else
-        {
-            san += mLedger->txMap().addKnownNode(
-                *nodeIDit, makeSlice(*nodeDatait), &filter);
-            if (!san.isGood())
-                return false;
-        }
+        auto nodeIDit = nodeIDs.cbegin();
+        auto nodeDatait = data.begin();
+        TransactionStateSF filter(
+            mLedger->txMap().family().db(), app_.getLedgerMaster());
 
-        ++nodeIDit;
-        ++nodeDatait;
+        while (nodeIDit != nodeIDs.cend())
+        {
+            if (nodeIDit->isRoot())
+            {
+                san += mLedger->txMap().addRootNode(
+                    SHAMapHash{mLedger->info().txHash},
+                    makeSlice(*nodeDatait),
+                    &filter);
+                if (!san.isGood())
+                    return false;
+            }
+            else
+            {
+                san += mLedger->txMap().addKnownNode(
+                    *nodeIDit, makeSlice(*nodeDatait), &filter);
+                if (!san.isGood())
+                    return false;
+            }
+
+            ++nodeIDit;
+            ++nodeDatait;
+        }
+    }
+    catch (std::exception const& ex)
+    {
+        JLOG(m_journal.error()) << "Peer sent bad tx node data: " << ex.what();
+        return false;
     }
 
     if (!mLedger->txMap().isSynching())
@@ -972,39 +979,47 @@ InboundLedger::takeAsNode(
         return true;
     }
 
-    auto nodeIDit = nodeIDs.cbegin();
-    auto nodeDatait = data.begin();
-    AccountStateSF filter(
-        mLedger->stateMap().family().db(), app_.getLedgerMaster());
-
-    while (nodeIDit != nodeIDs.cend())
+    try
     {
-        if (nodeIDit->isRoot())
-        {
-            san += mLedger->stateMap().addRootNode(
-                SHAMapHash{mLedger->info().accountHash},
-                makeSlice(*nodeDatait),
-                snfWIRE,
-                &filter);
-            if (!san.isGood())
-            {
-                JLOG(m_journal.warn()) << "Bad ledger header";
-                return false;
-            }
-        }
-        else
-        {
-            san += mLedger->stateMap().addKnownNode(
-                *nodeIDit, makeSlice(*nodeDatait), &filter);
-            if (!san.isGood())
-            {
-                JLOG(m_journal.warn()) << "Unable to add AS node";
-                return false;
-            }
-        }
+        auto nodeIDit = nodeIDs.cbegin();
+        auto nodeDatait = data.begin();
+        AccountStateSF filter(
+            mLedger->stateMap().family().db(), app_.getLedgerMaster());
 
-        ++nodeIDit;
-        ++nodeDatait;
+        while (nodeIDit != nodeIDs.cend())
+        {
+            if (nodeIDit->isRoot())
+            {
+                san += mLedger->stateMap().addRootNode(
+                    SHAMapHash{mLedger->info().accountHash},
+                    makeSlice(*nodeDatait),
+                    &filter);
+                if (!san.isGood())
+                {
+                    JLOG(m_journal.warn()) << "Unable to add AS root node";
+                    return false;
+                }
+            }
+            else
+            {
+                san += mLedger->stateMap().addKnownNode(
+                    *nodeIDit, makeSlice(*nodeDatait), &filter);
+                if (!san.isGood())
+                {
+                    JLOG(m_journal.warn()) << "Unable to add AS node";
+                    return false;
+                }
+            }
+
+            ++nodeIDit;
+            ++nodeDatait;
+        }
+    }
+    catch (std::exception const& ex)
+    {
+        JLOG(m_journal.error())
+            << "Peer sent bad account state node data: " << ex.what();
+        return false;
     }
 
     if (!mLedger->stateMap().isSynching())
@@ -1042,7 +1057,7 @@ InboundLedger::takeAsRootNode(Slice const& data, SHAMapAddNode& san)
     AccountStateSF filter(
         mLedger->stateMap().family().db(), app_.getLedgerMaster());
     san += mLedger->stateMap().addRootNode(
-        SHAMapHash{mLedger->info().accountHash}, data, snfWIRE, &filter);
+        SHAMapHash{mLedger->info().accountHash}, data, &filter);
     return san.isGood();
 }
 
@@ -1067,7 +1082,7 @@ InboundLedger::takeTxRootNode(Slice const& data, SHAMapAddNode& san)
     TransactionStateSF filter(
         mLedger->txMap().family().db(), app_.getLedgerMaster());
     san += mLedger->txMap().addRootNode(
-        SHAMapHash{mLedger->info().txHash}, data, snfWIRE, &filter);
+        SHAMapHash{mLedger->info().txHash}, data, &filter);
     return san.isGood();
 }
 
@@ -1156,28 +1171,38 @@ InboundLedger::processData(
 
         SHAMapAddNode san;
 
-        if (!mHaveHeader)
+        try
         {
-            if (takeHeader(packet.nodes(0).nodedata()))
-                san.incUseful();
-            else
+            if (!mHaveHeader)
             {
-                JLOG(m_journal.warn()) << "Got invalid header data";
-                peer->charge(Resource::feeInvalidRequest);
-                return -1;
+                if (!takeHeader(packet.nodes(0).nodedata()))
+                {
+                    JLOG(m_journal.warn()) << "Got invalid header data";
+                    peer->charge(Resource::feeInvalidRequest);
+                    return -1;
+                }
+
+                san.incUseful();
+            }
+
+            if (!mHaveState && (packet.nodes().size() > 1) &&
+                !takeAsRootNode(makeSlice(packet.nodes(1).nodedata()), san))
+            {
+                JLOG(m_journal.warn()) << "Included AS root invalid";
+            }
+
+            if (!mHaveTransactions && (packet.nodes().size() > 2) &&
+                !takeTxRootNode(makeSlice(packet.nodes(2).nodedata()), san))
+            {
+                JLOG(m_journal.warn()) << "Included TX root invalid";
             }
         }
-
-        if (!mHaveState && (packet.nodes().size() > 1) &&
-            !takeAsRootNode(makeSlice(packet.nodes(1).nodedata()), san))
+        catch (std::exception const& ex)
         {
-            JLOG(m_journal.warn()) << "Included AS root invalid";
-        }
-
-        if (!mHaveTransactions && (packet.nodes().size() > 2) &&
-            !takeTxRootNode(makeSlice(packet.nodes(2).nodedata()), san))
-        {
-            JLOG(m_journal.warn()) << "Included TX root invalid";
+            JLOG(m_journal.warn())
+                << "Included AS/TX root invalid: " << ex.what();
+            peer->charge(Resource::feeBadData);
+            return -1;
         }
 
         if (san.isUseful())
--- a/src/ripple/app/ledger/impl/InboundLedger.cpp
+++ b/src/ripple/app/ledger/impl/InboundLedger.cpp
@@ -47,7 +47,7 @@ enum {
     ,
     peerCountAdd = 2
 
-    // how many timeouts before we giveup
+    // how many timeouts before we give up
     ,
     ledgerTimeoutRetriesMax = 10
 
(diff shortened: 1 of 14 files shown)


src/ripple/shamap/SHAMapTreeNode.h
```diff
@@ -176,14 +176,43 @@ class SHAMapAbstractNode
...
+        bool hashValid,
+        beast::Journal j);
+
+    static std::shared_ptr<SHAMapAbstractNode>
+    makeFromWire(
+        Slice rawNode,
         std::uint32_t seq,
-        SHANodeFormat format,
         SHAMapHash const& hash,
         bool hashValid,
```
seelabs, line 191: It looks like all the current callers of this always pass in an invalid hash, could remove the `hash` and `hashValid` parameters.

There are several functions that use this `hash`, `hashValid` pair of params. It appears this is done to avoid a copy. I'm fine with that, but it would be good to document the rational for this on one of the functions or someone may refactor this to use `optional`.

src/ripple/shamap/impl/SHAMapTreeNode.cpp
```diff
@@ -76,220 +76,229 @@ SHAMapTreeNode::SHAMapTreeNode(
...
+    if (hashValid)
+        return std::make_shared<SHAMapTreeNode>(
+            std::move(item), tnTRANSACTION_NM, seq, hash);
+
+    return std::make_shared<SHAMapTreeNode>(
+        std::move(item), tnTRANSACTION_NM, seq);
+}
+
+std::shared_ptr<SHAMapAbstractNode>
+SHAMapAbstractNode::makeTransactionWithMeta(
```
seelabs, line 100: `makeTransactionWithMeta` and `makeAccountState` look identical except for the exception messages and `TNTType`, could make a single helper function they both call.

src/ripple/shamap/SHAMapTreeNode.h
seelabs: We could also consider using a raw pointer to the SHAMapHash as a stand in for an optional that doesn't copy.

src/ripple/basics/Slice.h
```diff
@@ -117,6 +126,21 @@ class Slice
     }
     /** @} */
 
+    /** Shrinks the slice by moving its start forward by n characters. */
+    void
+    remove_prefix(std::size_t n)
```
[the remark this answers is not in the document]
nbougalis: I opted to duplicate the interface of `std::string_view` since `Slice` is, by any other name, a `std::string_view`. The intention is to eventually eliminate the need for `Slice` altogether.

src/ripple/basics/Slice.h
```diff
@@ -117,6 +126,21 @@ class Slice
...
+    {
+        data_ += n;
+        size_ -= n;
```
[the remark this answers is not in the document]
nbougalis: Same as above: this actually replicates the semantics of `std::string_view` which, I will admit, are somewhat peculiar. I think I'm OK with doing this _but_ I'm concerned that if we come to rely on this extra error-checking, if we ever migrate to `std::string_view` we'll face problems.

src/ripple/overlay/impl/PeerImp.cpp
```diff
@@ -1184,7 +1184,15 @@ PeerImp::onMessage(std::shared_ptr<protocol::TMGetPeerShardInfo> const& m)
                 reply.set_lastlink(true);
 
             if (m->peerchain_size() > 0)
+            {
+                for (int i = 0; i < m->peerchain_size(); ++i)
+                {
+                    if (!publicKeyType(makeSlice(m->peerchain(i).nodepubkey())))
+                        return badData("Invalid peer chain public key");
```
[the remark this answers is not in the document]
nbougalis: A good idea.

src/ripple/shamap/SHAMapTreeNode.h
```diff
@@ -176,14 +176,43 @@ class SHAMapAbstractNode
...
         beast::Journal j,
-        SHAMapNodeID const& id = SHAMapNodeID{});
+        SHAMapNodeID const& id);
```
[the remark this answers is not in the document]
nbougalis: Good point - now that the `id` isn't defaulted, I can reshuffle things.

src/ripple/shamap/impl/SHAMapTreeNode.cpp
nbougalis: Will look, thanks!

src/ripple/shamap/SHAMapTreeNode.h
nbougalis: Hmm, interesting. Let me double-check that.

src/ripple/basics/Slice.h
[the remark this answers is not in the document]
seelabs: I hear you; I still vote for making the class consistent by adding the throw. I'm fine if you decide not to tho.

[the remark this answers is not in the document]
miguelportilla: I think the throw is a good idea and if not, we should at least assert.

src/ripple/basics/Slice.h
```diff
@@ -117,6 +126,21 @@ class Slice
...
+    }
+
+    /** Shrinks the slice by moving its end backward by n characters. */
+    void
+    remove_suffix(std::size_t n)
+    {
+        size_ -= n;
```
miguelportilla, line 141: Consider a throw or assert on n larger than size.

src/ripple/basics/Slice.h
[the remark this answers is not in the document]
miguelportilla: Can be noexcept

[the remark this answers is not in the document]
miguelportilla: Can be 
`    constexpr void
    remove_prefix(std::size_t n) noexcept`

src/ripple/basics/Slice.h
miguelportilla: Consider making it `constexpr void remove_suffix(std::size_t n)` with `noexcept` if you decide not to throw.

src/ripple/basics/Slice.h
```diff
@@ -117,6 +126,21 @@ class Slice
...
+    remove_prefix(std::size_t n)
+    {
+        data_ += n;
+        size_ -= n;
+    }
+
+    /** Shrinks the slice by moving its end backward by n characters. */
+    void
+    remove_suffix(std::size_t n)
+    {
```
miguelportilla, line 140: Consider making it constexpr void remove_suffix(std::size_t n) with noexcept if you decide not to throw.

src/ripple/basics/Slice.h
```diff
@@ -140,6 +164,28 @@ class Slice
...
+        @count requested length
+
+        @returns The requested subslice, if the request is valid.
+        @throws std::out_of_range if pos > size()
+     */
+    Slice
+    substr(
+        std::size_t pos,
+        std::size_t count = std::numeric_limits<std::size_t>::max()) const
+    {
```
miguelportilla, line 183: Consider making this `constexpr` and `pos` default to zero, as `string_view::substr` does.

src/ripple/basics/Slice.h
```diff
@@ -73,12 +74,20 @@ class Slice
...
     size() const noexcept
     {
         return size_;
     }
 
+    std::size_t
+    length() const noexcept
+    {
+        return size_;
+    }
```
miguelportilla, line 88: Consider making this `constexpr`.

src/ripple/basics/Slice.h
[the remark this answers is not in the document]
miguelportilla: Why not eliminate this class and use `std::string_view`?

src/ripple/basics/Slice.h
miguelportilla: Please update `Slice_test.cpp` to cover new functions.

src/ripple/app/ledger/impl/InboundLedger.cpp
```diff
@@ -898,33 +898,40 @@ InboundLedger::takeTxNode(
...
-                return false;
-        }
+            if (nodeIDit->isRoot())
+            {
+                san += mLedger->txMap().addRootNode(
+                    SHAMapHash{mLedger->info().txHash},
+                    makeSlice(*nodeDatait),
+                    &filter);
+                if (!san.isGood())
+                    return false;
```
miguelportilla, line 917: Consider logging here and below as we do with an Account State node.
eg:
`
                if (!san.isGood())
                {
                    JLOG(m_journal.warn()) << "Unable to add TX root node";
                    return false;
                }
`

src/ripple/app/ledger/impl/InboundLedger.cpp
```diff
@@ -898,33 +898,40 @@ InboundLedger::takeTxNode(
...
+            }
+            else
+            {
+                san += mLedger->txMap().addKnownNode(
+                    *nodeIDit, makeSlice(*nodeDatait), &filter);
+                if (!san.isGood())
```
miguelportilla, line 923: Consider logging here as we do with an Account State node.
eg:
```
                if (!san.isGood())
                {
                    JLOG(m_journal.warn()) << "Unable to add TX node";
                    return false;
                }
```

src/ripple/app/ledger/impl/InboundLedger.cpp
```diff
@@ -898,33 +898,40 @@ InboundLedger::takeTxNode(
...
+                    return false;
```
miguelportilla, line 924: Consider logging here as we do with an Account State node.
eg:
```
                if (!san.isGood())
                {
                    JLOG(m_journal.warn()) << "Unable to add TX node";
                    return false;
                }
```

src/ripple/app/ledger/impl/InboundLedger.cpp
```diff
@@ -898,33 +898,40 @@ InboundLedger::takeTxNode(
...
 
-        ++nodeIDit;
-        ++nodeDatait;
+            ++nodeIDit;
+            ++nodeDatait;
+        }
+    }
+    catch (std::exception const& ex)
+    {
+        JLOG(m_journal.error()) << "Peer sent bad tx node data: " << ex.what();
```
miguelportilla, line 933: OCD nit: consider either `Peer sent bad **TX** node data:` or `Peer sent bad transaction node data:`

src/ripple/app/ledger/impl/InboundLedger.cpp
```diff
@@ -972,39 +979,47 @@ InboundLedger::takeAsNode(
...
             {
-                JLOG(m_journal.warn()) << "Unable to add AS node";
-                return false;
+                san += mLedger->stateMap().addKnownNode(
+                    *nodeIDit, makeSlice(*nodeDatait), &filter);
+                if (!san.isGood())
+                {
+                    JLOG(m_journal.warn()) << "Unable to add AS node";
+                    return false;
+                }
```
miguelportilla, line 1011: Might be a good idea to condense these loops to 
```
        auto& stateMap{mLedger->stateMap()};
        AccountStateSF filter(stateMap.family().db(), app_.getLedgerMaster());

        while (nodeIDit != nodeIDs.cend())
        {
            if (nodeIDit->isRoot())
                san += stateMap.addRootNode(
                    SHAMapHash{mLedger->info().accountHash},
                    makeSlice(*nodeDatait),
                    &filter);
            else
                san += stateMap.addKnownNode(
                    *nodeIDit, makeSlice(*nodeDatait), &filter);

            if (!san.isGood())
            {
                JLOG(m_journal.warn()) << "Unable to add AS node";
                return false;
            }
```

miguelportilla: Lots of code duplication between `takeTxNode` and `takeAsNode`, might be a good idea to combine.

src/ripple/app/ledger/impl/InboundLedger.cpp
```diff
@@ -898,33 +898,40 @@ InboundLedger::takeTxNode(
...
-                SHAMapHash{mLedger->info().txHash},
-                makeSlice(*nodeDatait),
-                snfWIRE,
-                &filter);
-            if (!san.isGood())
-                return false;
-        }
-        else
+        auto nodeIDit = nodeIDs.cbegin();
+        auto nodeDatait = data.begin();
```
miguelportilla, line 904: I realize we pass (and assume) same sized vectors but a better signature would be 
```
bool
InboundLedger::takeTxNode(
    std::vector<std::pair<SHAMapNodeID, Blob>> const& nodes,
    SHAMapAddNode& san)
```

src/ripple/app/ledger/impl/InboundLedger.cpp
```diff
@@ -972,39 +979,47 @@ InboundLedger::takeAsNode(
         return true;
     }
 
-    auto nodeIDit = nodeIDs.cbegin();
-    auto nodeDatait = data.begin();
-    AccountStateSF filter(
-        mLedger->stateMap().family().db(), app_.getLedgerMaster());
-
-    while (nodeIDit != nodeIDs.cend())
+    try
```
miguelportilla, line 982: Consider
```
bool
InboundLedger::takeAsNode(
    std::vector<std::pair<SHAMapNodeID, Blob>> const& nodes,
    SHAMapAddNode& san)
```

src/ripple/app/ledger/impl/InboundLedger.cpp
miguelportilla: Using the function signature I suggested would clean it up further...
```
        auto& stateMap{mLedger->stateMap()};
        AccountStateSF filter(stateMap.family().db(), app_.getLedgerMaster());
        for (auto& [nodeID, data] : nodes)
        {
            if (nodeID.isRoot())
                san += stateMap.addRootNode(
                    SHAMapHash{mLedger->info().accountHash},
                    makeSlice(data),
                    &filter);
            else
                san += stateMap.addKnownNode(
                    nodeID, makeSlice(data), &filter);

            if (!san.isGood())
            {
                JLOG(m_journal.warn()) << "Unable to add AS node";
                return false;
            }
        }
```

src/ripple/basics/Slice.h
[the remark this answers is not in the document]
seelabs: I'm not sure I agree that `Slice` should be replaced `string_view`; `Slice` is for bytes (opaque binary data), `string_view` is for strings. We do sometimes use string-like containers for binary data, but I always think that sends the wrong signal as-to what type of data is stored. We probably want `std::span`, but that doesn't come until c++-20.

Having said that, it's only a weak objection. If both of you want to use `string_view` for this I'm OK with it.

src/ripple/app/ledger/impl/InboundLedger.cpp
miguelportilla, line 1011: The loops in `TakeTxNode` and `TakeAsNode` can be improved....

```
        auto& stateMap{mLedger->stateMap()};
        AccountStateSF filter(stateMap.family().db(), app_.getLedgerMaster());
        for (auto& [nodeID, data] : nodes)
        {
            if (nodeID.isRoot())
                san += stateMap.addRootNode(
                    SHAMapHash{mLedger->info().accountHash},
                    makeSlice(data),
                    &filter);
            else
                san += stateMap.addKnownNode(nodeID, makeSlice(data), &filter);

            if (!san.isGood())
            {
                JLOG(m_journal.warn()) << "Unable to add AS node";
                return false;
            }
        }
```

You can cherry-pick my changes here: https://github.com/miguelportilla/rippled/commits/catchy-plus-plus

src/ripple/shamap/SHAMapTreeNode.h
miguelportilla: As @seelabs  mentioned, current callers of `makeFromWire` pass in an invalid hash, both params can be removed.

Conversely, current callers of `makeFromPrefix` pass in a valid hash and `hashValid` can be removed.

I don't agree with the raw pointer idea, `SHAMapHash const&` is the convention.

src/ripple/shamap/SHAMapTreeNode.h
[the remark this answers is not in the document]
miguelportilla: The `Journal` param isn't used in either function and should be removed.

src/ripple/shamap/impl/SHAMap.cpp
```diff
@@ -181,20 +176,32 @@ SHAMap::fetchNodeFromDB(SHAMapHash const& hash) const
...
-        if (node)
+        try
         {
-            filter->gotNode(
-                true, hash, ledgerSeq_, std::move(*nodeData), node->getType());
-            if (backed_)
-                canonicalize(hash, node);
+            auto node = SHAMapAbstractNode::makeFromPrefix(
+                makeSlice(*nodeData), 0, hash, true, f_.journal());
+            if (node)
```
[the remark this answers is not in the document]
miguelportilla: Or
```
        try
        {
            if (auto node = SHAMapAbstractNode::makeFromPrefix(
                    makeSlice(*nodeData), 0, hash, true, f_.journal()))
            {
                filter->gotNode(
                    true,
                    hash,
                    ledgerSeq_,
                    std::move(*nodeData),
                    node->getType());
                if (backed_)
                    canonicalize(hash, node);
            }
        }
```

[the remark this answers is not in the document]
miguelportilla: Another possibility 

```
        try
        {
            if (auto node = SHAMapAbstractNode::makeFromPrefix(
                    makeSlice(*nodeData), 0, hash, true, f_.journal()))
            {
                ...
                return node;
            }
        }
```

src/ripple/basics/Slice.h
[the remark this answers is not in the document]
miguelportilla: @seelabs good point and convincing enough for me to stay as is until `span`.

src/ripple/basics/Slice.h
HowardHinnant: Not a big deal, but string_view has `size` too.  `length` is for legacy support.

miguelportilla:
@nbougalis On Windows I get a build fail with libarchive
archive_read_support_format_warc.c(344): error C4061: enumerator 'WT_NONE' in switch of enum 'warc_type_t' is not explicitly handled by a case label

The problem goes away after a rebase to 1.6.0-b7

src/ripple/basics/Slice.h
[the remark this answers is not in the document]
HowardHinnant: I agree that `span` is closer to our semantics.  But I don't yet have the experience to know that `span`'s API is a good fit for replacing `Splice`.  It _probably_ is.  If it is, we should be leaning towards that API as it is different than `string_view`.

src/ripple/app/ledger/impl/InboundLedger.cpp
miguelportilla: `san.incInvalid()` should probably be called here.

src/ripple/app/ledger/impl/InboundLedger.cpp
```diff
@@ -972,39 +979,47 @@ InboundLedger::takeAsNode(
...
-        ++nodeIDit;
-        ++nodeDatait;
+            ++nodeIDit;
+            ++nodeDatait;
+        }
+    }
+    catch (std::exception const& ex)
+    {
+        JLOG(m_journal.error())
+            << "Peer sent bad account state node data: " << ex.what();
```
miguelportilla, line 1021: `san.incInvalid()` should probably be called here.

src/ripple/basics/Slice.h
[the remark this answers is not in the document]
nbougalis: There's nothing about `std::string_view`, as far as I can tell, that requires it be used for strings, except perhaps its unfortunate name. I believe it was misnamed and should have been `std::byte_sequence` or somesuch, with a std aliases for a `const char*` instantiation and/or a specialization for `std::string`. 

We could replace `Slice` with `string_view` now but I don't know if that's a good idea. I agree that using `string_view` as the name implies something that may not be true for arbitrary binary blobs. Beyond the bad name it has a dangerous API, which I'm surprised survived review: some of the APIs respect boundaries, others don't. 

My hope is that we can remove custom code in favor of more `std::` stuff; the less code we need to write and maintain the better, especially for boilerplate stuff like this, but I don't know that we have a good `std::` alternative.

We could use [`boost::asio::const_buffer`](https://www.boost.org/doc/libs/1_70_0/doc/html/boost_asio/reference/const_buffer.html) and [`boost::asio::mutable_buffer`](https://www.boost.org/doc/libs/1_70_0/doc/html/boost_asio/reference/mutable_buffer.html) instead, but they lack some the nice functions that `std::string_view` offers.

Maybe we can just punt on further changes to `Slice` for now until we find something good.

src/ripple/basics/Slice.h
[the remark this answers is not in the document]
nbougalis: I think it's best to make the class API consistent, instead of repeating the committee's mistakes.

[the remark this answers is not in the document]
seelabs: I still think making `+=` throw and `remove_prefix` not throw is an inconsistent API (if anything, `remove_prefix` sounds like the higher-level api that should error check, not the other way around). But we don't need to re-litigate this, we've got bigger fish to fry.

src/ripple/shamap/SHAMapTreeNode.h
seelabs: @miguelportilla Side note: The idea was to use a raw pointer to replace an optional with an expensive copy. We can't use `SHAMapHash const&` to do that (well, not without another bool parameter). But it's moot anyway, the params can be removed.

Issue #3396, opened 2020-05-11
EnableAmendment timer starts at 77.78% (Version: [rippled 1.5.0])


gnurag:
## Issue Description
I may be reading this incorrectly - but I think the 14 day timer to enable an amendment starts at `77.78%` (28 Ayes out of 36 as of today). 28 Ayes of 36 is only `77.78%` agreement. [Amendment process](https://xrpl.org/amendments.html#amendment-process) doc says this must be `80%` or above.

Observed this behavior with [`RequireFullyCanonicalSig`](https://xrpl.org/known-amendments.html#requirefullycanonicalsig) amendment which is hovering around 27 Ayes at the moment. Once it achieved 28 Ayes (=`77.78%` agreement), the validator inserted an [EnableAmendment transaction](https://xrpcharts.ripple.com/#/transactions/0F699B1AC8AF0B112E7AE855ACAD14F3009245CB5D4582EFCE720CB86132C5E6) and the 14 day timer started.

![fullycanonicalsig-vote](https://user-images.githubusercontent.com/13650/81534663-0ced4b00-9386-11ea-9b1d-126d2339578e.png)

## Steps to Reproduce
Run:
```
# /opt/ripple/bin/rippled feature
```
on the validator when an amendment has 28 Ayes out of 36. 

## Expected Result
1. [EnableAmendment transaction](https://xrpcharts.ripple.com/#/transactions/0F699B1AC8AF0B112E7AE855ACAD14F3009245CB5D4582EFCE720CB86132C5E6) is not inserted into the ledger.
2. `majority` property is not set.

## Actual Result
1. [EnableAmendment transaction](https://xrpcharts.ripple.com/#/transactions/0F699B1AC8AF0B112E7AE855ACAD14F3009245CB5D4582EFCE720CB86132C5E6) is inserted into the ledger.
2. `majority` property is set (see screenshot above).

## Environment
rippled-1.5.0 

## Supporting Files

![fullycanonicalsig-votes](https://user-images.githubusercontent.com/13650/81535679-dc0e1580-9387-11ea-9f94-f3da8e034358.png)

nbougalis:
Thanks for opening an issue. A couple of people had already reached out to me about the DeletableAccounts amendment, wondering how come it reached majority, since 28 validators seem to support it but (a) the consensus quorum is presently 29 and (b) 28/36 is strictly less than 80%, which is the amendment activation threshold, so I already had a chance to triage this..

The answer boils down to two things:

[...]

PR #3428, opened 2020-06-02
Amendment table clean up

gregtatcam:
Use base 10 for majority vote calculation
Add majority timer configuration
FIXES: #3396

### DIFF
--- a/src/ripple/app/main/Application.cpp
+++ b/src/ripple/app/main/Application.cpp
@@ -78,7 +78,9 @@
 namespace ripple {
 
 // 204/256 about 80%
-static int const MAJORITY_FRACTION(204);
+static int const MAJORITY_FRACTION_OLD(204);
+// 8/10 = 80%
+static int const MAJORITY_FRACTION(8);
 
 //------------------------------------------------------------------------------
 
@@ -1533,8 +1535,8 @@ ApplicationImp::setup()
         Section enabledAmendments = config_->section(SECTION_AMENDMENTS);
 
         m_amendmentTable = make_AmendmentTable(
-            weeks{2},
-            MAJORITY_FRACTION,
+            config().AMENDMENT_MAJORITY_TIME,
+            {MAJORITY_FRACTION_OLD, MAJORITY_FRACTION},
             supportedAmendments,
             enabledAmendments,
             config_->section(SECTION_VETO_AMENDMENTS),
--- a/src/ripple/app/main/Application.cpp
+++ b/src/ripple/app/main/Application.cpp
@@ -77,11 +77,6 @@
 
 namespace ripple {
 
-// 204/256 about 80%
-static int const MAJORITY_FRACTION_OLD(204);
-// 8/10 = 80%
-static int const MAJORITY_FRACTION(8);
-
 //------------------------------------------------------------------------------
 
 namespace detail {
@@ -1536,7 +1531,7 @@ ApplicationImp::setup()
 
         m_amendmentTable = make_AmendmentTable(
             config().AMENDMENT_MAJORITY_TIME,
-            {MAJORITY_FRACTION_OLD, MAJORITY_FRACTION},
+            {amendmentMajorityFractionOld, amendmentMajorityFraction},
             supportedAmendments,
             enabledAmendments,
             config_->section(SECTION_VETO_AMENDMENTS),
--- a/src/ripple/app/main/Application.cpp
+++ b/src/ripple/app/main/Application.cpp
@@ -1531,7 +1531,6 @@ ApplicationImp::setup()
 
         m_amendmentTable = make_AmendmentTable(
             config().AMENDMENT_MAJORITY_TIME,
-            {amendmentMajorityFractionOld, amendmentMajorityFraction},
             supportedAmendments,
             enabledAmendments,
             config_->section(SECTION_VETO_AMENDMENTS),
--- a/src/ripple/app/misc/impl/AmendmentTable.cpp
+++ b/src/ripple/app/misc/impl/AmendmentTable.cpp
@@ -132,15 +132,15 @@ public:
             ? std::max(
                   1L,
                   static_cast<long>(
-                          (trustedValidations_ *
-                           preFixAmendmentMajorityCalcThreshold.num) /
-                          preFixAmendmentMajorityCalcThreshold.den))
+                      (trustedValidations_ *
+                      preFixAmendmentMajorityCalcThreshold.num) /
+                      preFixAmendmentMajorityCalcThreshold.den))
             : std::max(
                   1L,
                   static_cast<long>(
-                          (trustedValidations_ *
-                           postFixAmendmentMajorityCalcThreshold.num) /
-                          postFixAmendmentMajorityCalcThreshold.den));
+                      (trustedValidations_ *
+                      postFixAmendmentMajorityCalcThreshold.num) /
+                      postFixAmendmentMajorityCalcThreshold.den));
     }
 
     bool
--- a/src/ripple/app/misc/impl/AmendmentTable.cpp
+++ b/src/ripple/app/misc/impl/AmendmentTable.cpp
@@ -93,12 +93,16 @@ struct AmendmentState
 };
 
 /** The status of all amendments requested in a given window. */
-struct AmendmentSet
+class AmendmentSet
 {
 private:
     // How many yes votes each amendment received
     hash_map<uint256, int> votes_;
     Rules const& rules_;
+    // number of trusted validations
+    int trustedValidations_ = 0;
+    // number of votes needed
+    int threshold_ = 0;
 
 public:
     AmendmentSet(
@@ -120,23 +124,23 @@ public:
                         [&](auto const& amendment) { ++votes_[amendment]; });
                 }
 
-                ++mTrustedValidations;
+                ++trustedValidations_;
             }
         }
 
-        mThreshold = !rules_.enabled(fixAmendmentMajorityCalc)
+        threshold_ = !rules_.enabled(fixAmendmentMajorityCalc)
             ? std::max(
                   1L,
                   static_cast<long>(
-                      (mTrustedValidations *
-                       amendmentSuperMajorityThresholdPre3396.num) /
-                      amendmentSuperMajorityThresholdPre3396.den))
+                          (trustedValidations_ *
+                           preFixAmendmentMajorityCalcThreshold.num) /
+                          preFixAmendmentMajorityCalcThreshold.den))
             : std::max(
                   1L,
                   static_cast<long>(
-                      (mTrustedValidations *
-                       amendmentSuperMajorityThresholdPost3396.num) /
-                      amendmentSuperMajorityThresholdPost3396.den));
+                          (trustedValidations_ *
+                           postFixAmendmentMajorityCalcThreshold.num) /
+                          postFixAmendmentMajorityCalcThreshold.den));
     }
 
     bool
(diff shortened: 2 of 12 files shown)


src/ripple/core/impl/Config.cpp
```diff
@@ -480,6 +480,31 @@ Config::loadFromString(std::string const& fileContents)
...
+        using namespace std::chrono;
+        boost::regex const re(
+            "^\\s*(\\d+)\\s*(seconds|minutes|hours|days|weeks)\\s*(\\s+.*)?$");
+        boost::smatch match;
+        if (!boost::regex_match(strTemp, match, re))
+            Throw<std::runtime_error>(
+                "Invalid " SECTION_AMENDMENT_MAJORITY_TIME
+                ", must be: [0-9]+ [seconds|minutes|hours|days|weeks]");
+        std::uint32_t duration =
+            beast::lexicalCastThrow<std::uint32_t>(match[1].str());
```
nbougalis, line 495: I would not allow `seconds`: there's no real plausible scenario for such a short interval. In fact, I think I'd reject any interval smaller than 15 minutes.

src/ripple/app/main/Application.cpp
```diff
@@ -78,7 +78,9 @@
 namespace ripple {
 
 // 204/256 about 80%
-static int const MAJORITY_FRACTION(204);
+static int const MAJORITY_FRACTION_OLD(204);
+// 8/10 = 80%
+static int const MAJORITY_FRACTION(8);
```
nbougalis, line 83: These values, along with the default value for the majority time should probably migrate into `SystemParametersInfo.h`:

```
/** The minimum amount of support an amendment should have.

    @note This value is used by legacy code and will become obsolete
          once the fix3396 amendment activates.
*/
constexpr int const AmendmentMajorityFractionOld = 204;

/** The minimum amount of support an amendment must have */
constexpr int const AmendmentMajorityPercent = 80;

/** The minimum amount of time an amendment must hold a majority */
constexpr std::chrono::seconds const AmendmentMajorityTime = std::chrono::weeks{2};

```

src/ripple/core/impl/Config.cpp
gregtatcam: Done

src/ripple/app/main/Application.cpp
gregtatcam: Done, except for `amendmentMajorityPercent`, which is currently not used. Added instead `amendmentMajorityFraction`.

src/ripple/core/impl/Config.cpp
```diff
@@ -485,24 +485,29 @@ Config::loadFromString(std::string const& fileContents)
...
+        if (boost::iequals(match[2], "minutes"))
             AMENDMENT_MAJORITY_TIME = minutes(duration);
         else if (boost::iequals(match[2], "hours"))
             AMENDMENT_MAJORITY_TIME = hours(duration);
         else if (boost::iequals(match[2], "days"))
             AMENDMENT_MAJORITY_TIME = days(duration);
         else if (boost::iequals(match[2], "weeks"))
             AMENDMENT_MAJORITY_TIME = weeks(duration);
+
+        if (AMENDMENT_MAJORITY_TIME < defaultAmendmentMajorityTime)
```
nbougalis, line 507: I think this check is wrong. It will prevent anything less than 2 weeks. I'd simply hard-code `std::chrono::minutes(15)` here.

gregtatcam: Right!!!

src/test/app/AmendmentTable_test.cpp
```diff
@@ -713,8 +726,8 @@ class AmendmentTable_test final : public beast::unit_test::suite
                 enabled,
                 majority);
 
-            if (i < 8)
-            {
+            if (i < 4)  // 16 - 3 = 13 => 13/16 = 0.8125 => > 80%
+            {           // 16 - 4 = 12 => 12/16 = 0.75 => < 80%
```
scottschurr, line 730: Nice comment.

src/ripple/app/misc/impl/AmendmentTable.cpp
```diff
@@ -489,8 +492,10 @@ AmendmentTableImpl::doVoting(
         }
     }
 
-    vote->mThreshold =
-        std::max(1, (vote->mTrustedValidations * majorityFraction_) / 256);
+    vote->mThreshold = rules.enabled(ripple::fix3396)
+        ? std::max(1, (vote->mTrustedValidations * majorityFraction_.new_) / 10)
```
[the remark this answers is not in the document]
nbougalis: I appreciate the well-thought out comment (and agree with the 8 -> 80 change), but I find the division code confusing to read. Once I read it carefully, I understand what's happening, but I'm not confident it behaves _identically_ pre-amendment. I am hoping we can do something even better:

We should push all this logic into `AmendmentSet` and reduce it's public interface to a _single_ function: `bool passes(uint256) const`:

```
struct AmendmentSet
{
private:
    // How many yes votes each amendment received
    hash_map<uint256, int> votes_;

    // number of trusted validations
    int mTrustedValidations = 0;

    // number of votes needed
    int mThreshold;

    Rules const& rules;

public:
    AmendmentSet(
        Rules const& rules,
        MajorityFraction const& majorityFraction,
        std::vector<std::shared_ptr<STValidation>> const& valSet)
    {
        for (auto const& val : valSet)
        {
            if (val->isTrusted())
            {
                std::set<uint256> ballot;

                if (val->isFieldPresent(sfAmendments))
                {
                    auto const choices = val->getFieldV256(sfAmendments);
                    ballot.insert(choices.begin(), choices.end());
                }

                ++mTrustedValidations;

                for (auto const& amendment : ballot)
                    ++votes_[amendment];
            }
        }

        mThreshold = rules.enabled(fix3396)
            ? std::max(1, (vote->mTrustedValidations * majorityFraction.pre3396.num) / majorityFraction.pre3396.den)
             : std::max(1, (vote->mTrustedValidations * majorityFraction.post3396.num) / majorityFraction.post3396.den);
    }

    bool
    passes(uint256 const& amendment) const
    {
        auto const& it = votes_.find(amendment);

        if (it == votes_.end())
            return false;

        // Before this fix, it was possible for an amendment to activate with a percentage
        // slightly less than 80% because we compared for "greater than or equal to"
        // instead of strictly "greater than".
        if (!rules.enabled(fix3396))
            return it->second >= fix3396;

        return it->second > fix3396;
    }
};
```

In reviewing this code, I spotted something else that we may want to include: we "import" the `std::vector` of validations into a `std::set` to "de-duplicate" and make it easier to search.

I propose, instead, that, if this amendment is enabled, we skip that and instead call `std::sort` and `std::unique` and if a duplicate vote is present, that validator's vote for this round should be discarded in toto, and a warning should be produced.

I'll code it up.

src/ripple/protocol/SystemParameters.h
```diff
@@ -59,6 +60,18 @@ systemCurrencyCode()
 /** The XRP ledger network's earliest allowed sequence */
 static std::uint32_t constexpr XRP_LEDGER_EARLIEST_SEQ{32570};
 
+/** The minimum amount of support an amendment should have.
+
+    @note This value is used by legacy code and will become obsolete
+          once the fixAmendmentMajorityCalc amendment activates.
+*/
+constexpr std::ratio<205, 256> amendmentSuperMajorityThresholdPre3396;
```
nbougalis, line 68: Shouldn't the numerator be **`204`** to precisely match the semantics of the old code?

src/ripple/core/Config.h
```diff
@@ -23,6 +23,7 @@
 #include <ripple/basics/BasicConfig.h>
 #include <ripple/basics/FeeUnits.h>
 #include <ripple/basics/base_uint.h>
+#include <ripple/basics/chrono.h>
```
nbougalis, line 26: This should probably just be `#include <chrono>`

src/ripple/protocol/SystemParameters.h
```diff
@@ -21,6 +21,7 @@
 #define RIPPLE_PROTOCOL_SYSTEMPARAMETERS_H_INCLUDED
 
 #include <ripple/basics/XRPAmount.h>
+#include <ripple/basics/chrono.h>
```
nbougalis, line 24: `#include <chrono>`

src/ripple/core/Config.h
gregtatcam: Done

src/ripple/protocol/SystemParameters.h
gregtatcam: We need `ripple/basics/chrono.h` because of `weeks`.

src/ripple/protocol/SystemParameters.h
gregtatcam: Thanks. Nasty typo!

src/ripple/protocol/SystemParameters.h
nbougalis: Aha! OK.

nbougalis:
Weird unit test failures happening...

gregtatcam:
> Weird unit test failures happening...

I have been testing AmendmentTable only. It does fail on other tests. I'll take a look.

gregtatcam:
> > Weird unit test failures happening...
> 
> I have been testing AmendmentTable only. It does fail on other tests. I'll take a look.

The problem is that in some tests (for inst Feature_test::testWithMajorities()), we have one trusted validator and one vote. Consequently the threshold is max(1, 1*80/100) = 1. And because of the change for post 3396 fix the votes should be greater than the threshold, which is false and the test fails. Pre 3396 fix, the votes should be greater or equal to the threshold, which was true and the test succeeded. I'm investigating how to address this issue in the unit tests.

nbougalis:
@gregtatcam, please cherry-pick 9f6aefc28163e9d29f1a8f00e6bffd7dd06a52e5 to remove the "per-256" reporting in the `feature` RPC call. This is nominally an API change (removing the `vote`) response, but it's [not documented](https://xrpl.org/feature.html#feature) and the value (a) is confusing and (b) can be derived from the other information returned.

nbougalis:
The issue identified by @ximinez can be fixed by cherry-picking c228f9f439aa3086f34517ff0f6bdd9574b7402a

src/ripple/protocol/SystemParameters.h
```diff
@@ -59,6 +60,18 @@ systemCurrencyCode()
...
 static std::uint32_t constexpr XRP_LEDGER_EARLIEST_SEQ{32570};
 
+/** The minimum amount of support an amendment should have.
+
+    @note This value is used by legacy code and will become obsolete
+          once the fixAmendmentMajorityCalc amendment activates.
+*/
+constexpr std::ratio<204, 256> amendmentSuperMajorityThresholdPre3396;
+
+constexpr std::ratio<80, 100> amendmentSuperMajorityThresholdPost3396;
```
scottschurr, line 70: Now that the `fix` has been renamed these names don't make as much sense as they used to.  There aren't any other mentions of 3396 in the source code.  I don't have any great suggestions, but maybe...

- `preFixAmendmentMajorityCalcThreshold` and `postFixAmendmentMajorityCalcThreshold` or
- `preFixAmendmentMajorityCalcSuperMajority` and `postFixAmendmentMajorityCalcSuperMajority`

would be a little easier to figure out.  You may have better ideas than these.

src/ripple/app/misc/impl/AmendmentTable.cpp
```diff
@@ -97,25 +98,73 @@ struct AmendmentSet
 private:
     // How many yes votes each amendment received
     hash_map<uint256, int> votes_;
+    Rules const& rules_;
```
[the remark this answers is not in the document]
gregtatcam: Done

src/ripple/protocol/SystemParameters.h
gregtatcam: Done

intelliot:
Released in 1.6.0 https://xrpl.org/blog/2020/rippled-1.6.0.html

Amendment activated in April 2021

PR #3429, opened 2020-06-02
Improve online_delete configuration and DB tuning:

### LINKED ISSUE #3321: Issues with online_delete configuration (Version: 1.5.0-rc3)

ximinez:
* Document delete_batch, back_off_milliseconds, age_threshold_seconds.
* Convert those time values to chrono types.
* Fix bug that ignored age_threshold_seconds.
* Add a "recovery buffer" to the config that gives the node a chance to
  recover before aborting online delete.
* Add begin/end log messages around the SQL queries.
* Add a new configuration section: [sqlite] to allow tuning the sqlite
  database operations. Ignored on full/large history servers.
* Update documentation of [node_db] and [sqlite] in the
  rippled-example.cfg file.
* Resolves ripple#3321

I added the "recovery buffer" function that wasn't described in the original issue to scratch my own itch, because my local machine was losing sync for a fraction of a second while the ledger was being copied over. This resolved that issue. I defaulted it to not run, so it doesn't cause unexpected behavior.

Note this includes two commits, one from @mtrippled . Both need to be reviewed.

---
This change is [Reviewable](https://reviewable.io/reviews/ripple/rippled/3429)

### DIFF
--- a/cfg/rippled-example.cfg
+++ b/cfg/rippled-example.cfg
@@ -975,7 +975,11 @@
 #   particularly in failure scenarios. It is strongly recommended that they
 #   be left at their defaults unless the server is having performance issues
 #   during normal operation or during automatic purging (online_delete)
-#   operations.
+#   operations. A warning will be logged on startup if 'ledger_history'
+#   is configured to store more than 10,000,000 ledgers and any of these
+#   settings are less safe than the default. This is due to the inordinate
+#   amount of time and bandwidth it will take to safely rebuild a corrupted
+#   database from other peers.
 #
 #   Optional keys:
 #
--- a/cfg/rippled-example.cfg
+++ b/cfg/rippled-example.cfg
@@ -36,7 +36,7 @@
 #   For more information on where the rippled server instance searches for the
 #   file, visit:
 #
-#     https://developers.ripple.com/commandline-usage.html#generic-options
+#     https://xrpl.org/commandline-usage.html#generic-options
 #
 #   This file should be named rippled.cfg. This file is UTF-8 with DOS, UNIX,
 #   or Mac style end of lines. Blank lines and lines beginning with '#' are
@@ -939,7 +939,7 @@
 #   [import_db]     Settings for performing a one-time import (optional)
 #   [database_path]   Path to the book-keeping databases.
 #
-#   There are 4 or 5 bookkeeping SQLite database that the server creates and
+#   There are 4 or 5 bookkeeping SQLite databases that the server creates and
 #   maintains. If you omit this configuration setting, it will default to
 #   creating a directory called "db" located in the same place as your
 #   rippled.cfg file. Partial pathnames will be considered relative to
@@ -1335,7 +1335,7 @@ advisory_delete=0
 # This is the persistent datastore for shards. It is important for the health
 # of the ripple network that rippled operators shard as much as practical.
 # NuDB requires SSD storage. Helpful information can be found at
-# https://ripple.com/build/history-sharding
+# https://xrpl.org/history-sharding.html
 #[shard_db]
 #path=/var/lib/rippled/db/shards/nudb
 #max_size_gb=500
@@ -1354,7 +1354,8 @@ time.apple.com
 time.nist.gov
 pool.ntp.org
 
-# To use the XRP test network (see https://ripple.com/build/xrp-test-net/),
+# To use the XRP test network
+# (see https://xrpl.org/connect-your-rippled-to-the-xrp-test-net.html),
 # use the following [ips] section:
 # [ips]
 # r.altnet.rippletest.net 51235
--- a/cfg/rippled-example.cfg
+++ b/cfg/rippled-example.cfg
@@ -917,13 +917,16 @@
 #                           The online delete process checks periodically
 #                           that rippled is still in sync with the network,
 #                           and that the validated ledger is less than
-#                           *age_threshold_seconds* old. By default, if it
+#                           'age_threshold_seconds' old. By default, if it
 #                           is not the online delete process aborts and
-#                           tries again later. If *recovery_buffer_seconds*
-#                           is set, online delete will wait this number of
-#                           seconds for rippled to recover before it aborts.
+#                           tries again later. If 'recovery_buffer_seconds'
+#                           is set and rippled is out of sync, but likely to
+#                           recover quickly, then online delete will wait
+#                           this number of seconds for rippled to get back
+#                           into sync before it aborts.
 #                           Set this value if the node is otherwise staying
-#                           in sync, or recovering quickly.
+#                           in sync, or recovering quickly, but the online
+#                           delete process is unable to finish.
 #                           Default is unset.
 #
 #   Notes:
--- a/cfg/rippled-example.cfg
+++ b/cfg/rippled-example.cfg
@@ -913,13 +913,13 @@
 #                           number of seconds.
 #                           Default is 60.
 #
-#       recovery_buffer_seconds
+#       recovery_wait_seconds
 #                           The online delete process checks periodically
 #                           that rippled is still in sync with the network,
 #                           and that the validated ledger is less than
 #                           'age_threshold_seconds' old. By default, if it
 #                           is not the online delete process aborts and
-#                           tries again later. If 'recovery_buffer_seconds'
+#                           tries again later. If 'recovery_wait_seconds'
 #                           is set and rippled is out of sync, but likely to
 #                           recover quickly, then online delete will wait
 #                           this number of seconds for rippled to get back
@@ -967,30 +967,37 @@
 #       <key> '=' <value>
 #       ...
 #
-#   Example:
+#   Example 1:
 #       sync_level=low
+#
+#   Example 2:
 #       journal_mode=off
+#       synchronous=off
 #
 #   WARNING: These settings can have significant effects on data integrity,
-#   particularly in failure scenarios. It is strongly recommended that they
-#   be left at their defaults unless the server is having performance issues
-#   during normal operation or during automatic purging (online_delete)
-#   operations. A warning will be logged on startup if 'ledger_history'
-#   is configured to store more than 10,000,000 ledgers and any of these
-#   settings are less safe than the default. This is due to the inordinate
-#   amount of time and bandwidth it will take to safely rebuild a corrupted
-#   database from other peers.
+#   particularly in systemic failure scenarios. It is strongly recommended
+#   that they be left at their defaults unless the server is having
+#   performance issues during normal operation or during automatic purging
+#   (online_delete) operations. A warning will be logged on startup if
+#   'ledger_history' is configured to store more than 10,000,000 ledgers and
+#   any of these settings are less safe than the default. This is due to the
+#   inordinate amount of time and bandwidth it will take to safely rebuild a
+#   corrupted database of that size from other peers.
 #
 #   Optional keys:
 #
 #       safety_level        Valid values: high, low
-#                           The default is "high", and tunes the SQLite
-#                           databases in the most reliable mode. "low"
-#                           is equivalent to
+#                           The default is "high", which tunes the SQLite
+#                           databases in the most reliable mode, and is
+#                           equivalent to:
+#                             journal_mode=wal
+#                             synchronous=normal
+#                             temp_store=file
+#                           "low" is equivalent to:
 #                             journal_mode=memory
 #                             synchronous=off
 #                             temp_store=memory
-#                           These settings trade speed and reduced I/O
+#                           These "low" settings trade speed and reduced I/O
 #                           for a higher risk of data loss. See the
 #                           individual settings below for more information.
 #                           This setting may not be combined with any of the
--- a/cfg/rippled-example.cfg
+++ b/cfg/rippled-example.cfg
@@ -869,18 +869,62 @@
 #
 #       These keys are possible for any type of backend:
 #
+#       earliest_seq        The default is 32570 to match the XRP ledger
+#                           network's earliest allowed sequence. Alternate
+#                           networks may set this value. Minimum value of 1.
+#                           If a [shard_db] section is defined, and this
+#                           value is present either [node_db] or [shard_db],
+#                           it must be defined with the same value in both
+#                           sections.
+#
 #       online_delete       Minimum value of 256. Enable automatic purging
 #                           of older ledger information. Maintain at least this
 #                           number of ledger records online. Must be greater
 #                           than or equal to ledger_history.
 #
-#       advisory_delete     0 for disabled, 1 for enabled. If set, then
-#                           require administrative RPC call "can_delete"
-#                           to enable online deletion of ledger records.
+#       These keys modify the behavior of online_delete, and thus are only
+#       relevant if online_delete is defined and non-zero:
 #
-#       earliest_seq        The default is 32570 to match the XRP ledger
-#                           network's earliest allowed sequence. Alternate
-#                           networks may set this value. Minimum value of 1.
+#       advisory_delete     0 for disabled, 1 for enabled. If set, the
+#                           administrative RPC call "can_delete" is required
+#                           to enable online deletion of ledger records.
+#                           Online deletion does not run automatically if
+#                           non-zero and the last deletion was on a ledger
+#                           greater than the current "can_delete" setting.
+#                           Default is 0.
+#
+#       delete_batch        When automatically purging, SQLite database
+#                           records are deleted in batches. This value
+#                           controls the maximum size of each batch. Larger
+#                           batches keep the databases locked for more time,
+#                           which may cause other functions to fall behind,
+#                           and thus cause the node to lose sync.
+#                           Default is 100.
+#
+#       back_off_milliseconds
+#                           Number of milliseconds to wait between
+#                           online_delete batches to allow other functions
+#                           to catch up.
+#                           Default is 100.
+#
+#       age_threshold_seconds
+#                           The online delete process will only run if the
+#                           latest validated ledger is younger than this
+#                           number of seconds.
+#                           Default is 60.
+#
+#       recovery_buffer_seconds
+#                           The online delete process checks periodically
+#                           that rippled is still in sync with the network,
+#                           and that the validated ledger is less than
+#                           *age_threshold_seconds* old. By default, if it
+#                           is not the online delete process aborts and
+#                           tries again later. If *recovery_buffer_seconds*
+#                           is set, online delete will wait this number of
+#                           seconds for rippled to recover before it aborts.
+#                           Set this value if the node is otherwise staying
+#                           in sync, or recovering quickly.
+#                           Default is unset.
 #
 #   Notes:
 #       The 'node_db' entry configures the primary, persistent storage.
@@ -891,6 +935,12 @@
 #
 #   [import_db]     Settings for performing a one-time import (optional)
 #   [database_path]   Path to the book-keeping databases.
+#
+#   There are 4 or 5 bookkeeping SQLite database that the server creates and
+#   maintains. If you omit this configuration setting, it will default to
+#   creating a directory called "db" located in the same place as your
+#   rippled.cfg file. Partial pathnames will be considered relative to
+#   the location of the rippled executable.
 #
 #   [shard_db]      Settings for the Shard Database (optional)
 #
@@ -907,12 +957,64 @@
 #
 #       max_size_gb         Maximum disk space the database will utilize (in gigabytes)
 #
+#   [sqlite]       Tuning settings for the SQLite databases (optional)
 #
-#   There are 4 bookkeeping SQLite database that the server creates and
-#   maintains. If you omit this configuration setting, it will default to
-#   creating a directory called "db" located in the same place as your
-#   rippled.cfg file. Partial pathnames will be considered relative to
-#   the location of the rippled executable.
+#   Format (without spaces):
+#       One or more lines of case-insensitive key / value pairs:
+#       <key> '=' <value>
+#       ...
+#
+#   Example:
+#       sync_level=low
+#       journal_mode=off
+#
+#   WARNING: These settings can have significant effects on data integrity,
+#   particularly in failure scenarios. It is strongly recommended that they
+#   be left at their defaults unless the server is having performance issues
+#   during normal operation or during automatic purging (online_delete)
+#   operations.
+#
+#   Optional keys:
+#
+#       safety_level        Valid values: high, low
+#                           The default is "high", and tunes the SQLite
+#                           databases in the most reliable mode. "low"
+#                           is equivalent to
+#                             journal_mode=memory
+#                             synchronous=off
+#                             temp_store=memory
+#                           These settings trade speed and reduced I/O
+#                           for a higher risk of data loss. See the
+#                           individual settings below for more information.
+#
+#       journal_mode        Valid values: delete, truncate, persist, memory, wal, off
+#                           The default is "wal", which uses a write-ahead
+#                           log to implement database transactions.
+#                           Alternately, "memory" saves disk I/O, but if
+#                           rippled crashes during a transaction, the
+#                           database is likely to be corrupted.
+#                           See https://www.sqlite.org/pragma.html#pragma_journal_mode
+#                           for more details about the available options.
+#
+#       synchronous         Valid values: off, normal, full, extra
+#                           The default is "normal", which works well with
+#                           the "wal" journal mode. Alternatively, "off"
+#                           allows rippled to continue as soon as data is
+#                           passed to the OS, which can significantly
+#                           increase speed, but risks data corruption if
+#                           the host computer crashes before writing that
+#                           data to disk.
+#                           See https://www.sqlite.org/pragma.html#pragma_synchronous
+#                           for more details about the available options.
+#
+#       temp_store          Valid values: default, file, memory
+#                           The default is "file", which will use files
+#                           for temporary database tables and indices.
+#                           Alternatively, "memory" may save I/O, but
+#                           rippled does not currently use many, if any,
+#                           of these temporary objects.
+#                           See https://www.sqlite.org/pragma.html#pragma_temp_store
+#                           for more details about the available options.
 #
 #
 #
@@ -1212,23 +1314,24 @@ medium
 
 # This is primary persistent datastore for rippled.  This includes transaction
 # metadata, account states, and ledger headers.  Helpful information can be
-# found here: https://ripple.com/wiki/NodeBackEnd
-# delete old ledgers while maintaining at least 2000. Do not require an
-# external administrative command to initiate deletion.
+# found at https://xrpl.org/capacity-planning.html#node-db-type
+# type=NuDB is recommended for non-validators with fast SSDs. Validators or
+#    slow / spinning disks should use RocksDB.
+# online_delete=512 is recommended to delete old ledgers while maintaining at
+#    least 512.
+# advisory_delete=0 allows the online delete process to run automatically
+#    when the node has approximately two times the "online_delete" value of
+#    ledgers. No external administrative command is required to initiate
+#    deletion.
 [node_db]
-type=RocksDB
-path=/var/lib/rippled/db/rocksdb
-open_files=2000
-filter_bits=12
-cache_mb=256
-file_size_mb=8
-file_size_mult=2
-online_delete=2000
+type=NuDB
+path=/var/lib/rippled/db/nudb
+online_delete=512
 advisory_delete=0
 
 # This is the persistent datastore for shards. It is important for the health
 # of the ripple network that rippled operators shard as much as practical.
-# NuDB requires SSD storage. Helpful information can be found here
+# NuDB requires SSD storage. Helpful information can be found at
 # https://ripple.com/build/history-sharding
 #[shard_db]
 #path=/var/lib/rippled/db/shards/nudb
--- a/cfg/rippled-example.cfg
+++ b/cfg/rippled-example.cfg
@@ -993,6 +993,9 @@
 #                           These settings trade speed and reduced I/O
 #                           for a higher risk of data loss. See the
 #                           individual settings below for more information.
+#                           This setting may not be combined with any of the
+#                           other tuning settings: "journal_mode",
+#                           "synchronous", or "temp_store".
 #
 #       journal_mode        Valid values: delete, truncate, persist, memory, wal, off
 #                           The default is "wal", which uses a write-ahead
@@ -1002,6 +1005,8 @@
 #                           database is likely to be corrupted.
 #                           See https://www.sqlite.org/pragma.html#pragma_journal_mode
 #                           for more details about the available options.
+#                           This setting may not be combined with the
+#                           "safety_level" setting.
 #
 #       synchronous         Valid values: off, normal, full, extra
 #                           The default is "normal", which works well with
@@ -1013,6 +1018,8 @@
 #                           data to disk.
 #                           See https://www.sqlite.org/pragma.html#pragma_synchronous
 #                           for more details about the available options.
+#                           This setting may not be combined with the
+#                           "safety_level" setting.
 #
 #       temp_store          Valid values: default, file, memory
 #                           The default is "file", which will use files
@@ -1022,6 +1029,8 @@
 #                           of these temporary objects.
 #                           See https://www.sqlite.org/pragma.html#pragma_temp_store
 #                           for more details about the available options.
+#                           This setting may not be combined with the
+#                           "safety_level" setting.
 #
 #
 #
--- a/cfg/rippled-example.cfg
+++ b/cfg/rippled-example.cfg
@@ -939,11 +939,11 @@
 #   [import_db]     Settings for performing a one-time import (optional)
 #   [database_path]   Path to the book-keeping databases.
 #
-#   There are 4 or 5 bookkeeping SQLite databases that the server creates and
-#   maintains. If you omit this configuration setting, it will default to
-#   creating a directory called "db" located in the same place as your
-#   rippled.cfg file. Partial pathnames will be considered relative to
-#   the location of the rippled executable.
+#   The server creates and maintains 4 to 5 bookkeeping SQLite databases in
+#   the 'database_path' location. If you omit this configuration setting,
+#   the server creates a directory called "db" located in the same place as
+#   your rippled.cfg file.
+#   Partial pathnames are relative to the location of the rippled executable.
 #
 #   [shard_db]      Settings for the Shard Database (optional)
 #
@@ -1339,7 +1339,9 @@ medium
 # metadata, account states, and ledger headers.  Helpful information can be
 # found at https://xrpl.org/capacity-planning.html#node-db-type
 # type=NuDB is recommended for non-validators with fast SSDs. Validators or
-#    slow / spinning disks should use RocksDB.
+#    slow / spinning disks should use RocksDB. Caution: Spinning disks are
+#    not recommended. They do not perform well enough to consistently remain
+#    synced to the network.
 # online_delete=512 is recommended to delete old ledgers while maintaining at
 #    least 512.
 # advisory_delete=0 allows the online delete process to run automatically
--- a/src/ripple/app/ledger/Ledger.cpp
+++ b/src/ripple/app/ledger/Ledger.cpp
@@ -228,14 +228,14 @@ Ledger::Ledger(
         !txMap_->fetchRoot(SHAMapHash{info_.txHash}, nullptr))
     {
         loaded = false;
-        JLOG(j.warn()) << "Don't have TX root for ledger";
+        JLOG(j.warn()) << "Don't have TX root for ledger" << info_.seq;
     }
 
     if (info_.accountHash.isNonZero() &&
         !stateMap_->fetchRoot(SHAMapHash{info_.accountHash}, nullptr))
     {
         loaded = false;
-        JLOG(j.warn()) << "Don't have AS root for ledger";
+        JLOG(j.warn()) << "Don't have AS root for ledger" << info_.seq;
     }
 
     txMap_->setImmutable();
--- a/src/ripple/app/ledger/Ledger.cpp
+++ b/src/ripple/app/ledger/Ledger.cpp
@@ -228,14 +228,14 @@ Ledger::Ledger(
         !txMap_->fetchRoot(SHAMapHash{info_.txHash}, nullptr))
     {
         loaded = false;
-        JLOG(j.warn()) << "Don't have TX root for ledger" << info_.seq;
+        JLOG(j.warn()) << "Don't have transaction root for ledger" << info_.seq;
     }
 
     if (info_.accountHash.isNonZero() &&
         !stateMap_->fetchRoot(SHAMapHash{info_.accountHash}, nullptr))
     {
         loaded = false;
-        JLOG(j.warn()) << "Don't have AS root for ledger" << info_.seq;
+        JLOG(j.warn()) << "Don't have state data root for ledger" << info_.seq;
     }
 
     txMap_->setImmutable();
--- a/src/ripple/app/main/DBInit.h
+++ b/src/ripple/app/main/DBInit.h
@@ -71,21 +71,13 @@ inline constexpr std::array<char const*, 5> LgrDBInit{
 // Transaction database holds transactions and public keys
 inline constexpr auto TxDBName{"transaction.db"};
 
-inline constexpr
-#if (ULONG_MAX > UINT_MAX) && !defined(NO_SQLITE_MMAP)
-    std::array<char const*, 4>
-        TxDBPragma
+inline constexpr std::array TxDBPragma
 {
-    {
-#else
-    std::array<char const*, 3> TxDBPragma {{
-#endif
-        "PRAGMA page_size=4096;", "PRAGMA journal_size_limit=1582080;",
-            "PRAGMA max_page_count=2147483646;",
+    "PRAGMA page_size=4096;", "PRAGMA journal_size_limit=1582080;",
+        "PRAGMA max_page_count=2147483646;",
 #if (ULONG_MAX > UINT_MAX) && !defined(NO_SQLITE_MMAP)
-            "PRAGMA mmap_size=17179869184;"
+        "PRAGMA mmap_size=17179869184;"
 #endif
-    }
 };
 
 inline constexpr std::array<char const*, 8> TxDBInit{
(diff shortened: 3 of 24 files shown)


cfg/rippled-example.cfg
```diff
@@ -1212,23 +1314,24 @@ medium
...
-path=/var/lib/rippled/db/rocksdb
-open_files=2000
-filter_bits=12
-cache_mb=256
-file_size_mb=8
-file_size_mult=2
-online_delete=2000
+type=NuDB
+path=/var/lib/rippled/db/nudb
+online_delete=512
```
MarkusTeufelberger, line 1329: If such a small value is the default, please consider also adding a shard_db section with a limit of a few dozen/hundred GiB by default.

cfg/rippled-example.cfg
```diff
@@ -1212,23 +1314,24 @@ medium
 
 # This is primary persistent datastore for rippled.  This includes transaction
 # metadata, account states, and ledger headers.  Helpful information can be
-# found here: https://ripple.com/wiki/NodeBackEnd
-# delete old ledgers while maintaining at least 2000. Do not require an
-# external administrative command to initiate deletion.
+# found at https://xrpl.org/capacity-planning.html#node-db-type
+# type=NuDB is recommended for non-validators with fast SSDs. Validators or
```
MarkusTeufelberger, line 1318: I would use a `none` database backend on a validator. They don't need to concern themselves with historic data at all.

cfg/rippled-example.cfg
ximinez: @MarkusTeufelberger Thanks for that feedback. Those default settings are definitely not yet set in stone. My understanding that the suggestion for the smaller default was to make it easier for more people to participate without requiring huge resources. (The original ticket didn't have a lot of detail.) Of course, there's a balance to be found, because we already have pretty high hardware requirements, and we do want as many nodes as possible to contribute to history storage as a way to help the network.

I'm not sure if shards are ready to be set by default. I'll defer to @miguelportilla on that question.

cfg/rippled-example.cfg
ximinez: The linked capacity planning page recommends RocksDB, but I can definitely see your point about using `none` (or `memory`). My biggest concern with using those options on the individual node level, is that it means it will take much longer to start and restart, because the node will have to download the entire ledger every time.  On the global level, I'm thinking about a failsafe / worst case situation. For example, if every validator shuts down simultaneously :scream:, they will need to have some data on disk so they can bootstrap restarting the network.

MarkusTeufelberger: Anecdotally, my full history node takes about half an hour to start with `--load` vs. a few minutes at most to get synced with `--net`.

Bootstrapping can be done safely with `--ledgerfile` (https://xrpl.org/commandline-usage.html#initial-ledger-options) and that's probably better anyways a better way forward than trusting that data was stored to disk correctly in such a catastrophic scenario. The file itself could be hashed and communicated out-of-band in a lot of different ways compared to a node database that's not even a full shard in most cases.

cfg/rippled-example.cfg
```diff
@@ -1212,23 +1317,24 @@ medium
...
+type=NuDB
+path=/var/lib/rippled/db/nudb
+online_delete=512
 advisory_delete=0
 
 # This is the persistent datastore for shards. It is important for the health
 # of the ripple network that rippled operators shard as much as practical.
-# NuDB requires SSD storage. Helpful information can be found here
+# NuDB requires SSD storage. Helpful information can be found at
 # https://ripple.com/build/history-sharding
```
nbougalis, line 1338: Please update this URL

src/ripple/core/impl/Config.cpp
```diff
@@ -454,7 +454,7 @@ Config::loadFromString(std::string const& fileContents)
         if (boost::iequals(strTemp, "none"))
             FETCH_DEPTH = 0;
         else if (boost::iequals(strTemp, "full"))
-            FETCH_DEPTH = 1000000000u;
+            FETCH_DEPTH = 1'000'000'000u;
```
nbougalis, line 457: Consider [`std::numeric_limits<std::uint32_t>::max`](https://en.cppreference.com/w/cpp/types/numeric_limits/max) instead for `FETCH_DEPTH` and `LEDGER_HISTORY`.

cfg/rippled-example.cfg
```diff
@@ -892,6 +939,12 @@
 #   [import_db]     Settings for performing a one-time import (optional)
 #   [database_path]   Path to the book-keeping databases.
 #
+#   There are 4 or 5 bookkeeping SQLite database that the server creates and
```
nbougalis, line 942: Typo: database -> databases.

src/ripple/core/impl/Config.cpp
ximinez: Do you know if there was a reason this was defined as 1 billion instead of `max()` in the first place?

cfg/rippled-example.cfg
ximinez: Good catch. Fixed this and several other ripple.com URLs.

cfg/rippled-example.cfg
ximinez: Fixed.

cfg/rippled-example.cfg
nbougalis: @MarkusTeufelberger: re: slow start times with `--load`: are you using SSDs?

src/ripple/core/impl/Config.cpp
nbougalis: There was no particular reason.

src/ripple/core/DatabaseCon.h
```diff
@@ -89,12 +89,17 @@ class DatabaseCon
...
+        /// Shortcut used by the database connections that ignore the common
+        /// pragma strings
+        static const std::vector<std::string> NoCommonPragma;
     };
 
     template <std::size_t N, std::size_t M>
     DatabaseCon(
         Setup const& setup,
         std::string const& DBName,
+        bool useCommonPragma,
```
nbougalis, line 102: I don't know if I like this; I'm generally allergic to functions that accept a `bool` and use it to do one or the other thing. Why can't we set the right values in `setup` directly?

src/ripple/core/DatabaseCon.h
```diff
@@ -89,12 +89,17 @@ class DatabaseCon
         Config::StartUpType startUp = Config::NORMAL;
         bool standAlone = false;
         boost::filesystem::path dataDir;
+        static std::unique_ptr<std::vector<std::string> const> CommonPragma;
+        /// Shortcut used by the database connections that ignore the common
+        /// pragma strings
+        static const std::vector<std::string> NoCommonPragma;
```
nbougalis, line 95: μNit: weird case for the name(s) here. Also confused as to why two separate things are needed.

src/ripple/app/main/DBInit.h
```diff
@@ -26,13 +26,23 @@ namespace ripple {
...
+// connections, unless otherwise noted.
+inline constexpr char const* CommonDBPragmaJournal{"PRAGMA journal_mode=%s;"};
+inline constexpr char const* CommonDBPragmaSync{"PRAGMA synchronous=%s;"};
+inline constexpr char const* CommonDBPragmaTemp{"PRAGMA temp_store=%s;"};
+// Default values will always be used for the common pragmas if
+// at least this much ledger history is configured. This includes
+// full history nodes. This is because such a large amount of data will
+// be more difficult to recover if a rare failure occurs, which are
+// more likely with some of the other available tuning settings.
+inline constexpr std::uint32_t SQLITE_TUNING_CUTOFF = 100'000'000;
```
nbougalis, line 39: I understand the rationale behind this, but I'm not sure if it's a good idea. I think that it's reasonable to warn in the configuration file and even on every startup about these options, but making the options entirely unavailable is not. Ironically, the existing logic that enforces this will prevent choosing options that are safer than the default even if the server operator _really_ wants them.

I'll give this some more thought, but right now, it seems like a bad idea.

nbougalis (commented):
Left some comments. Still thinking about this.

src/ripple/core/DatabaseCon.h
ximinez: It would be pretty simple to set the values in the `Setup` object, now that you mention it. I might need to define a member function or two, just to make the internals easier, and to make it easier switch between the options for `Setup` objects that are used to create multiple `DatabaseCon`s

ximinez: PS. I went down this path because I initially thought the other ctor override would need to use the "common pragma" but without a `Setup` object being passed in. I later realized none of them do, but didn't backtrack to see if the `bool` was still a good idea. So thanks for pointing this out!

src/ripple/core/DatabaseCon.h
ximinez: A lot of other static class variables have capitalized names, but I don't necessarily have to keep doing it.

cfg/rippled-example.cfg
MarkusTeufelberger: Of course. Over a dozen TB of 'em.

src/ripple/app/main/DBInit.h
ximinez: I can change it to warning log message(s) pretty easily, and update the config file.

src/ripple/app/misc/SHAMapStoreImp.cpp
```diff
@@ -666,16 +705,32 @@ SHAMapStoreImp::health()
...
+            JLOG(journal_.warn())
+                << "Waiting " << recoveryBuffer_->count()
+                << "s for node to get back into sync with network. state: "
+                << app_.getOPs().strOperatingMode(mode, false) << ". age "
+                << age.count() << 's';
+            std::this_thread::sleep_for(*recoveryBuffer_);
+
+            age = ledgerMaster_->getValidatedLedgerAge();
+            mode = netOPs_->getOperatingMode();
+        }
```
mtrippled, line 726: Notwithstanding recovery_buffer_seconds, the online_delete activity is triggered each time a new ledger is validated. And desyncs seem to usually happen for an entire consensus interval, of ~3-4 seconds. This really seems superfluous, especially adding a new config option to tune this.

mtrippled (changes requested):

src/ripple/app/misc/SHAMapStoreImp.cpp
```diff
@@ -566,19 +573,34 @@ SHAMapStoreImp::clearSql(
...
     }
 
     if (min > lastRotated || health() != Health::ok)
         return false;
+    if (min == lastRotated)
+    {
+        // Micro-optimization mainly to clarify logs
+        JLOG(journal_.trace()) << "Nothing to delete from " << deleteQuery;
+        return true;
+    }
```
mtrippled, line 603: Actually, we want to delete up to or equal to lastRotated.
For example, the loop on line 609 should be while (min <= lastRotated)
And if min == lastRotated, then break out of the loop/return after the query is executed. Though there may be something more elegant.

src/ripple/core/impl/DatabaseCon.cpp
```diff
@@ -38,9 +40,119 @@ setup_DatabaseCon(Config const& c)
...
 
+    if (!setup.globalPragma)
+    {
+        setup.globalPragma = [&c, &j]() {
+            auto const& sqlite = c.section("sqlite");
+            auto result = std::make_unique<std::vector<std::string>>();
+            result->reserve(3);
+
+            // defaults
+            std::string safety_level = "high";
```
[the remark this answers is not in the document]
ximinez: I disagree, but I think it's because of how I'm viewing the settings overall. I think of the `safety_level` as a baseline of settings, then all the others as overrides of the `safety_level`. In other words, the default is "high", but if I want transactions to be asyncronous, I can set `synchronous=off`. Similarly, if I want the lower safety level, but want to make sure that logging is safer, I could set
```
safety_level=low
journal_mode=wal
```

It's entirely possible I'm overthinking this. I could update the documentation in the example config file to be clearer that safety_level is the baseline and the others override.

> I think part of what you're doing with showWarning addresses this

The intention of `showWarning` is specific to when the config is set up to store a _lot_ of history. Think full history node, but since it's just a warning message, I set the cutoff lower.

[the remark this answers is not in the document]
ximinez: > Why would anybody choose only 1 or 2, and not all 3?

Excellent question. Some of the options can actually lower the data integrity risk, at the cost of extra I/O. I figured if I'm going to provide the options, I should go all the way, since I'm sure there are some folks who would use them.

> Perhaps safety_level should be removed

The more I think about it, the more I think it makes sense to make `safety_level` mutually exclusive from the other options. Someone who wants a simple option can use `safety_level=low`, and someone who wants to tune with more detail can whatever combos of the other three that they want. In other words, the example I gave above is starting to feel contrived and unrealistic.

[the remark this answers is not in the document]
mtrippled: I agree.

src/ripple/app/main/Main.cpp
```diff
@@ -555,7 +556,9 @@ run(int argc, char** argv)
             session << "PRAGMA temp_store_directory=\"" << tmpPath.string()
                     << "\";";
             session << "VACUUM;";
-            session << "PRAGMA journal_mode=WAL;";
```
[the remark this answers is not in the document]
ximinez: Turns out that `temp_store_directory` is deprecated: https://www.sqlite.org/pragma.html#pragma_temp_store_directory.

It looks like you're right about the `journal_mode`: https://www.sqlite.org/lang_vacuum.html
> The VACUUM command works by copying the contents of the database into a temporary database file and then overwriting the original with the contents of the temporary file. When overwriting the original, a rollback journal or write-ahead log WAL file is used just as it would be for any other database transaction.

So I'll update that.

src/ripple/app/misc/SHAMapStoreImp.cpp
```diff
@@ -348,23 +359,14 @@ SHAMapStoreImp::run()
 
         // will delete up to (not including) lastRotated
         if (validatedSeq >= lastRotated + deleteInterval_ &&
-            canDelete_ >= lastRotated - 1)
+            canDelete_ >= lastRotated - 1 && !health())
```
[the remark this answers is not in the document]
ximinez: Primarily, it's to prevent writing the log message if no actual work is going to get done. Compare to what it was doing before: log the "rotating" message, then do a `health()` check, then `clearPrior()`.

If any change should be made, I'd think it would be to take the first `health()` check out of `clearPrior()`. What do you think?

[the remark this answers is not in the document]
mtrippled: Yes, removing the first health() from clearPrior() makes sense.

src/ripple/app/misc/SHAMapStoreImp.cpp
```diff
@@ -566,19 +573,34 @@ SHAMapStoreImp::clearSql(
...
+        JLOG(journal_.trace())
+            << "Begin: Look up lowest value of: " << minQuery;
+        if (auto db = database.checkoutDb())
+            *db << minQuery, soci::into(m);
+        else
+        {
+            assert(false);
+            return false;
+        }
+        JLOG(journal_.trace()) << "End: Look up lowest value of: " << minQuery;
```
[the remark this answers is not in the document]
ximinez: I honestly don't remember for sure why I added the `if (auto db = database.checkoutDb())` check... I think it was so that it would go out of scope and release the lock sooner, and then I added the `else` because it "wouldn't hurt". It would make more sense to rewrite it as a simple local block.

src/ripple/app/misc/SHAMapStoreImp.cpp
```diff
@@ -566,19 +573,34 @@ SHAMapStoreImp::clearSql(
...
         boost::optional<std::uint64_t> m;
-        *db << minQuery, soci::into(m);
+        JLOG(journal_.trace())
+            << "Begin: Look up lowest value of: " << minQuery;
+        if (auto db = database.checkoutDb())
+            *db << minQuery, soci::into(m);
+        else
+        {
+            assert(false);
+            return false;
```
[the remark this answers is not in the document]
ximinez: Good catch! Will fix.

src/ripple/app/misc/SHAMapStoreImp.cpp
ximinez: I'm not married to it, but on my local node, it was taking so long for the ledger to copy that the node would switch to `syncing` before it could finish, and thus it would abort only to almost immediately start rotating again. The thing is that as far as I could tell, it wasn't losing sync because of the ledger copy, but because it build a different ledger during consensus than the UNL validators. Here's an example that includes some log messages that I added only to figure out what was going on:

```
2020-May-27 19:55:17.7525561 UTC SHAMapStore:TRC Nothing to delete from DELETE FROM AccountTransactions WHERE LedgerSeq < %u;
2020-May-27 19:55:17.7547395 UTC SHAMapStore:TRC copying ledger 55752529
2020-May-27 19:55:35.0709935 UTC NetworkOPs:NFO consensusViewChange
2020-May-27 19:55:35.0718717 UTC NetworkOPs:NFO STATE->syncing
2020-May-27 19:55:35.1798535 UTC SHAMapStore:WRN Not deleting. state: syncing. age 1s
2020-May-27 19:55:35.1803785 UTC SHAMapStore:WRN Not deleting. state: syncing. age 1s
2020-May-27 19:55:37.7108585 UTC SHAMapStore:WRN Not deleting. state: syncing. age 0s
2020-May-27 19:55:38.6848155 UTC NetworkOPs:NFO STATE->tracking
2020-May-27 19:55:38.6850136 UTC NetworkOPs:NFO STATE->full
2020-May-27 19:55:41.6406001 UTC SHAMapStore:WRN rotating  validatedSeq 55752535 lastRotated_ 55461357 deleteInterval 30000 canDelete_ 55461356 state full age 0s
```
Note the timestamps. At 19:55:17, it starts copying the ledger. 28 seconds later, the consensus view changes and it drops to syncing. 0.1 seconds later, the ledger copy aborts in `health()`. About 3 seconds after that (the next ledger, I presume), we're back to tracking, then full. 3 seconds later, we start rotating again. My node was doing this for _hours_. It's a beefy machine, but a Windows debug build.

The reason I added it as a config option is so that it _wouldn't_ do this by default. 

The obvious question is whether it's worth adding a feature to support marginal nodes. It seemed like a good idea at the time, and it at least helped me get this feature developed and finished.

I can think of two options:
* Remove it entirely.
* Make the feature `#if DEBUG`, remove the config option, and use a hard-coded value.

I'll go with whatever you think makes more sense.

mtrippled: It could be useful in a troubleshooting context--since you've already done the work, may as well keep it, as an option, but not by default. It's not plausible from that vantage point to determine whether the node itself is flaky, or if online_delete triggered the desync. So, to be safe, assume the latter, which means the default behavior should be to not have a buffer of time to wait for the node to recover.

src/ripple/app/misc/SHAMapStoreImp.cpp
ximinez: Are you sure about that? Every other check runs up to, but doesn't include `lastRotated`, and it's documented that way. https://github.com/ripple/rippled/blob/9173a6a7d3a08be4a7e8345bd763802693a8dcdb/src/ripple/app/misc/SHAMapStoreImp.cpp#L360__

src/ripple/app/misc/SHAMapStoreImp.cpp
ximinez: To confirm, the default time is to not have the buffer by default. :+1:

src/ripple/app/misc/SHAMapStoreImp.cpp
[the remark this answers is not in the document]
ximinez: Done

src/ripple/app/misc/SHAMapStoreImp.cpp
mtrippled: To be precise, the code should actually execute the query as it is for all values <=  lastRotated
            DELETE FROM <table> WHERE LedgerSeq < <lastRotated>;

The existing code actually does not call the query when min==lastRotated. It should execute the query. Consider that lastRotated is 1000000. We want to delete everything from 1..999999. But imagine that the function is called initially with min set to 1000000. The delete query will not be called. This is in the existing code. Your opimization made it apparent to me upon close inspection.

Once everything < lastRotated has been deleted, perform no more queries. With the existing code, it would mean checking if min==lastRotated after the delete query, then breaking out of the while loop, and checking for min <= lastRotated in the while condition.

Alternately, something like this:

    while (min <= lastRotated)
    {
        LedgerIndex minRange = std::min(lastRotated, min + deleteBatch);
        {
            auto db =  database.checkoutDb ();
            *db << boost::str (formattedDeleteQuery % minRange);
        }
        if (health())
            return true;
        min = minRange;
        if (min < lastRotated)
            std::this_thread::sleep_for (
                    std::chrono::milliseconds (backOff_));
    }

ximinez: But `min` is initialized from the data in the table.
```
    LedgerIndex min = std::numeric_limits<LedgerIndex>::max();

    {
        boost::optional<std::uint64_t> m;
        JLOG(journal_.trace())
            << "Begin: Look up lowest value of: " << minQuery;
        {
            auto db = database.checkoutDb();
            *db << minQuery, soci::into(m);
        }
        JLOG(journal_.trace()) << "End: Look up lowest value of: " << minQuery;
        if (!m)
            return;
        min = *m;
    }
```
So if `lastRotated` is 1,000,000, and the `minQuery` returns `1,000,000`, that means that there are no records in that particular table from 1..999,999. That's why I added the extra check, and why the loop doesn't need to run. We _know_ that the DELETE query won't delete any rows, because we just discovered that there are no rows to delete.

mtrippled: You're correct. The change I suggested doesn't need to be made. Also, keeping your original change is OK.

src/ripple/app/main/DBInit.h
```diff
@@ -63,15 +73,14 @@ inline constexpr auto TxDBName{"transaction.db"};
...
 inline constexpr
 #if (ULONG_MAX > UINT_MAX) && !defined(NO_SQLITE_MMAP)
-    std::array<char const*, 6>
+    std::array<char const*, 4>
         TxDBPragma
 {
     {
 #else
-    std::array<char const*, 5> TxDBPragma {{
+    std::array<char const*, 3> TxDBPragma {{
```
seelabs, line 81: The preprocessor guards make this declaration hard to read. Now that we're on c++-17 I think we can simplify this to:
```

inline constexpr std::array TxDBPragma
{
    "PRAGMA page_size=4096;", "PRAGMA journal_size_limit=1582080;",
        "PRAGMA max_page_count=2147483646;",
#if (ULONG_MAX > UINT_MAX) && !defined(NO_SQLITE_MMAP)
        "PRAGMA mmap_size=17179869184;",
#endif
};
```

src/ripple/app/main/Main.cpp
```diff
@@ -522,23 +519,24 @@ run(int argc, char** argv)
...
-                std::cerr << "A valid directory for vacuuming must be "
-                             "specified on a filesystem with at least "
-                             "as much free space as the size of "
-                          << dbPath.string() << ", which is " << dbSize
-                          << " bytes. The filesystem for " << tmpPath.string()
-                          << " only has " << space(tmpPath).available
-                          << " bytes.\n";
-                return -1;
+                auto available = space(dbPath.parent_path()).available;
+                if (available < dbSize)
```
seelabs, line 530: Could move the decl into the `if` and lose the extra scope:
```
            if (auto available = space(dbPath.parent_path()).available;
                available < dbSize)
```

src/ripple/app/misc/SHAMapStoreImp.h
```diff
@@ -106,8 +107,9 @@ class SHAMapStoreImp : public SHAMapStore
     std::uint32_t deleteInterval_ = 0;
     bool advisoryDelete_ = false;
     std::uint32_t deleteBatch_ = 100;
-    std::uint32_t backOff_ = 100;
-    std::int32_t ageThreshold_ = 60;
+    std::chrono::milliseconds backOff_{100};
+    std::chrono::seconds ageThreshold_{60};
+    boost::optional<std::chrono::seconds> recoveryBuffer_{};
```
seelabs, line 112: When I see something called `xxxBuffer` I think of memory buffers. We're using "buffer" in a different sense here. Maybe cal lit "Margin"? Or something else?

We should also leave a brief comment explaining what this does. I know the others don't, but it would be good to add comment for the new fields at least.
Also, nit: No need for the extra `{}`

src/ripple/app/misc/SHAMapStoreImp.h
```diff
@@ -218,7 +220,7 @@ class SHAMapStoreImp : public SHAMapStore
      *  @return true if any deletable rows were found (though not
```
seelabs, line 220: Outdated comment

src/ripple/app/misc/SHAMapStoreImp.cpp
```diff
@@ -587,17 +606,25 @@ SHAMapStoreImp::clearSql(
...
+            return;
         if (min < lastRotated)
-            std::this_thread::sleep_for(std::chrono::milliseconds(backOff_));
+            std::this_thread::sleep_for(backOff_);
+        if (health())
+            return;
     }
     JLOG(journal_.debug()) << "finished: " << deleteQuery;
-    return true;
+    return;
```
seelabs, line 627: Could remove this return.

src/ripple/app/misc/SHAMapStoreImp.cpp
```diff
@@ -666,16 +694,32 @@ SHAMapStoreImp::health()
...
+        OperatingMode mode = netOPs_->getOperatingMode();
+        if (recoveryBuffer_ && mode == OperatingMode::SYNCING &&
+            age < ageThreshold_)
+        {
+            JLOG(journal_.warn())
+                << "Waiting " << recoveryBuffer_->count()
+                << "s for node to get back into sync with network. state: "
+                << app_.getOPs().strOperatingMode(mode, false) << ". age "
+                << age.count() << 's';
+            std::this_thread::sleep_for(*recoveryBuffer_);
```
seelabs, line 711: Sleeping in a call to "health" is a bit of a surprising interface. I can see why you did this. Since `health` is always called from functions on the main `run` thread, this should be OK. Still, we should probably add a comment to "health" saying the function can sleep and that it should never be called from a function outside of the `run` thread including functions that part of the public interface.

src/ripple/core/impl/DatabaseCon.cpp
```diff
@@ -38,9 +40,134 @@ setup_DatabaseCon(Config const& c)
...
+            std::string safety_level;
+            std::string journal_mode = "wal";
+            std::string synchronous = "normal";
+            std::string temp_store = "file";
+            bool showRiskWarning = false;
+
+            if (set(safety_level, "safety_level", sqlite))
+            {
+                showRiskWarning = boost::iequals(safety_level, "low");
+                if (showRiskWarning)
```
seelabs, line 60: I'd rewrite this to make the condition `boost::iequals(safety_level, "low")` and then setting `showRiskWarning` in the if statement. That way it's clearer that we're running that code if the `safety_level` is low, it's only incidentally related to `showRiskWarning`.

src/ripple/nodestore/impl/Shard.cpp
```diff
@@ -683,6 +684,7 @@ Shard::initSQLite(std::lock_guard<std::recursive_mutex> const&)
 
         if (backendComplete_)
         {
+            setup.noPragma();
```
seelabs, line 687: Maybe initialize `setup` so it's `const` and not change the `setup` object in the arms of the if statement? Something like:
```
    DatabaseCon::Setup const setup = [&] {
        DatabaseCon::Setup result;
        result.startUp = config.START_UP;
        result.standAlone = config.standalone();
        result.dataDir = dir_;
        if (backendComplete_)
            result.noPragma();
        else
            result.usePragma();
        return result;
    }();
```

src/test/jtx/CheckMessageLogs.h
```diff
@@ -0,0 +1,74 @@
...
+        }
+    };
+
+public:
+    /** Constructor
+
+        @param msg The message string to search for
+        @param found The variable to set to true if the message is found
+    */
+    CheckMessageLogs(std::string msg, bool& found)
```
seelabs, line 59: I think the interface is clearer if we use raw pointers for this interface. I.e. the caller would look like:
```
CheckMessageLogs("MISMATCH ", &found);
```
If you didn't write the test code, I think it's easier to guess what `CheckMessageLogs` does if we pass by pointer.

seelabs (commented):
Left some minor comments, but overall this looks good.

src/ripple/app/main/Main.cpp
```diff
@@ -552,9 +549,6 @@ run(int argc, char** argv)
             std::cout << "VACUUM beginning. page_size: " << pageSize
                       << std::endl;
 
-            session << "PRAGMA journal_mode=OFF;";
-            session << "PRAGMA temp_store_directory=\"" << tmpPath.string()
```
mtrippled, line 555: The purpose of the temp_store_directory is to make sure there's enough space to perform a VACUUM, which essentially requires as much as the entire database being VACUUMed. The temp_store in memory will likely cause a system to run out of RAM. Imagine transaction DB with 2TB+. (https://sqlite.org/tempfiles.html)

cfg/rippled-example.cfg
nbougalis: I think that `--net` should become the default now, unless a different command line option (for which I don't have a good name) is specified.

ximinez: @nbougalis I think making `--net` the default is out of scope for this PR. Do you agree?

src/ripple/app/misc/NetworkOPs.cpp
```diff
@@ -2757,16 +2757,24 @@ NetworkOPsImp::getServerInfo(bool human, bool admin, bool counters)
             if (std::abs(closeOffset.count()) >= 60)
                 l[jss::close_time_offset] = closeOffset.count();
 
-            auto lCloseTime = lpClosed->info().closeTime;
-            auto closeTime = app_.timeKeeper().closeTime();
-            if (lCloseTime <= closeTime)
+            constexpr std::chrono::seconds HIGH_AGE_THRESHOLD{1000000};
```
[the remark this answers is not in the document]
ximinez: Fixed

cfg/rippled-example.cfg
```diff
@@ -907,12 +960,77 @@
...
+#   settings are less safe than the default. This is due to the inordinate
+#   amount of time and bandwidth it will take to safely rebuild a corrupted
+#   database from other peers.
+#
+#   Optional keys:
+#
+#       safety_level        Valid values: high, low
+#                           The default is "high", and tunes the SQLite
+#                           databases in the most reliable mode. "low"
+#                           is equivalent to
```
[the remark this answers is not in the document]
ximinez: Good suggestion. Done.

cfg/rippled-example.cfg
```diff
@@ -907,12 +960,77 @@
...
+#                           passed to the OS, which can significantly
+#                           increase speed, but risks data corruption if
+#                           the host computer crashes before writing that
+#                           data to disk.
+#                           See https://www.sqlite.org/pragma.html#pragma_synchronous
+#                           for more details about the available options.
+#                           This setting may not be combined with the
+#                           "safety_level" setting.
+#
+#       temp_store          Valid values: default, file, memory
```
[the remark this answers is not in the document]
ximinez: I'm honestly not sure. It came up in my research as potentially useful, but I couldn't really find any examples of rippled using temporary objects. However, that doesn't mean that it doesn't, and I figured that if someone is willing to take the risk, it doesn't hurt to give them the option. I'm weakly in favor of adding/keeping it unless we can definitively prove that it doesn't matter.

src/ripple/core/DatabaseCon.h
```diff
@@ -139,11 +156,20 @@ class DatabaseCon
     void
     init(
```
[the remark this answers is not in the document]
ximinez: The only reason I can see is that the `pPath` is a little ugly to do inline. Otherwise, it seems to work fine.

cfg/rippled-example.cfg
```diff
@@ -892,6 +939,12 @@
 #   [import_db]     Settings for performing a one-time import (optional)
 #   [database_path]   Path to the book-keeping databases.
 #
+#   There are 4 or 5 bookkeeping SQLite databases that the server creates and
```
mDuo13, line 942: Why is it "4 or 5"? Is one of them only added with certain configurations? Without more context, the phrase "4 or 5" feels uncertain and vague. Maybe "4 to 5" would carry the right implication.

cfg/rippled-example.cfg
```diff
@@ -892,6 +939,12 @@
...
+#   maintains. If you omit this configuration setting, it will default to
+#   creating a directory called "db" located in the same place as your
+#   rippled.cfg file. Partial pathnames will be considered relative to
+#   the location of the rippled executable.
```
mDuo13, line 946: nit: instead of "it will default to creating a directory", just say, "the server creates a directory." Similarly, instead of "will be considered relative" just say "are relative".

cfg/rippled-example.cfg
```diff
@@ -1212,24 +1330,25 @@ medium
 
 # This is primary persistent datastore for rippled.  This includes transaction
 # metadata, account states, and ledger headers.  Helpful information can be
-# found here: https://ripple.com/wiki/NodeBackEnd
-# delete old ledgers while maintaining at least 2000. Do not require an
-# external administrative command to initiate deletion.
+# found at https://xrpl.org/capacity-planning.html#node-db-type
+# type=NuDB is recommended for non-validators with fast SSDs. Validators or
+#    slow / spinning disks should use RocksDB.
```
mDuo13, line 1335: Consider adding a note like this:

> Caution: Spinning disks are not recommended. They do not perform well enough to consistently remain synced to the network.

src/ripple/app/ledger/Ledger.cpp
```diff
@@ -228,14 +228,14 @@ Ledger::Ledger(
         !txMap_->fetchRoot(SHAMapHash{info_.txHash}, nullptr))
     {
         loaded = false;
-        JLOG(j.warn()) << "Don't have TX root for ledger";
+        JLOG(j.warn()) << "Don't have TX root for ledger" << info_.seq;
```
mDuo13, line 231: Nit: as long as we're changing these messages, we could make them clearer by removing the abbreviations "TX" (here) and "AS" (below, line 238). Instead, just say "transaction" and "state data".

mDuo13 (commented):
I looked over the documentation changes. Definitely a big improvement over the previous version! I left a few optional suggestions to tune up the language in a few places, but :+1: with or without those changes.

src/ripple/core/DatabaseCon.h
```diff
@@ -116,7 +133,7 @@ class DatabaseCon
         std::array<char const*, N> const& pragma,
         std::array<char const*, M> const& initSQL)
     {
-        init((dataDir / DBName), pragma, initSQL);
+        init((dataDir / DBName), {}, pragma, initSQL);
```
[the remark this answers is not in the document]
ximinez: Fixed

src/ripple/core/DatabaseCon.h
```diff
@@ -89,6 +90,22 @@ class DatabaseCon
         Config::StartUpType startUp = Config::NORMAL;
         bool standAlone = false;
         boost::filesystem::path dataDir;
+        // If unseated, then the `globalPragma` are not used,
+        // otherwise should point to `globalPragma`
+        std::shared_ptr<std::vector<std::string> const> commonPragma;
```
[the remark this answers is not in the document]
ximinez: Yes, I belive I could. This is different than the `bool` param that @nbougalis objected to in https://github.com/ripple/rippled/pull/3429#discussion_r435656253, because it's a member with a clearer purpose. I can also abstract that away so the caller only sees a function call.

src/ripple/core/DatabaseCon.h
```diff
@@ -89,6 +90,22 @@ class DatabaseCon
...
+            commonPragma.reset();
+        }
+        void
+        usePragma()
+        {
+            assert(globalPragma);
+            commonPragma = globalPragma;
+        }
+
+        static std::shared_ptr<std::vector<std::string> const> globalPragma;
```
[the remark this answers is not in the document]
ximinez: Alas. But I'm changing it to a `unique_ptr`.

src/ripple/app/main/DBInit.h
ximinez: Nice. Done.

src/ripple/app/misc/SHAMapStoreImp.h
ximinez: `recoveryWaitTime_` and renamed the config setting to `recovery_wait_seconds`.

src/ripple/core/impl/DatabaseCon.cpp
ximinez: That's how I had it the first time, but I changed it to match the pattern on the other items. But it's not quite the same pattern, and it does make sense to do it your way here.

src/ripple/nodestore/impl/Shard.cpp
ximinez: Done

src/test/jtx/CheckMessageLogs.h
ximinez: Makes sense. Done.

src/ripple/app/main/Main.cpp
ximinez: Good catch! I'll change that.

cfg/rippled-example.cfg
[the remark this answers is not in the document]
seelabs: Ok let's keep it.

src/ripple/core/DatabaseCon.h
```diff
@@ -115,8 +131,8 @@ class DatabaseCon
         std::string const& DBName,
         std::array<char const*, N> const& pragma,
         std::array<char const*, M> const& initSQL)
+        : DatabaseCon({}, dataDir / DBName, {}, pragma, initSQL)
```
seelabs, line 134: See the comment below about removing the first parameter to this function. However, _if_ we wanted to use this idiom, I'd give the full class name `DatabaseCon(Base{}, ...)` to make the code easier to navigate; I'd probably rename the `Base` parameter as well, but it's not worth bike-shedding now since I think we should remove it anyway.

src/ripple/core/DatabaseCon.h
```diff
@@ -135,15 +151,28 @@ class DatabaseCon
...
 private:
+    class Base
+    {
+    };
+
     template <std::size_t N, std::size_t M>
-    void
-    init(
+    DatabaseCon(
+        Base,
```
seelabs, line 160: I don't understand the motivation for adding the `Base` parameter. The function doesn't appear to be ambiguous without it. Unless I'm missing the reason, can we remove this parameter?

seelabs (approved):
:+1: I did leave one nit that I think we should resolve (the `Base` parameter on the `DatabaseCon` constructor). However since that's so minor I'll give a thumbs up now.

src/ripple/app/main/Main.cpp
```diff
@@ -517,33 +517,34 @@ run(int argc, char** argv)
...
 
             auto txnDB = std::make_unique<DatabaseCon>(
                 dbSetup, TxDBName, TxDBPragma, TxDBInit);
             auto& session = txnDB->getSession();
             std::uint32_t pageSize;
 
+            // Only the most trivial databases will fit in memory on typical
+            // (recommended) software. Force temp files to be written to disk
+            // regardless of the config settings.
+            session << boost::format(CommonDBPragmaTemp) % "file";
```
mtrippled, line 547: The original behavior needs to come back--namely, that the actual temp directory be specified. Otherwise, SQLIte will by default use the directory in which the database is.

Alternately, I think that getting rid of this feature altogether makes a lot of sense. If somebody *really* wants to VACUUM a database, it's easier to use the command line tool.

mtrippled (changes requested):

src/ripple/core/DatabaseCon.h
ximinez: Basically, if I continue to use `{}` to specify the `commonPragma` param, then it is ambiguous. However, because of the rewrite in the last commit to use a raw pointer, it is not ambiguous if I use `nullptr`, so I've removed `Base`, and changed the caller to use `nullptr`.

mtrippled (approved):
LGTM
:+1:

cfg/rippled-example.cfg
ximinez: I think `state.db` is created if you're using `online_delete`, so that would be the 5th one. I'll make the change.

cfg/rippled-example.cfg
ximinez: Changed and reworded to be less passive tone.

mDuo13:
New changes  :+1:  from me.

PR #3449, opened 2020-06-11
Update cmake so that rippled can build as a submodule

ximinez:
This is a relatively trivial fix to the build system which allows rippled or parts of it to be built when included as a submodule to another project. There is also the option used in dbc8f147c92bf60f06bbed4eabcd93fd206d33dc, but this seemed simpler.

---
This change is [Reviewable](https://reviewable.io/reviews/ripple/rippled/3449)


ximinez:
@cjcobb23 In short, if you can build rippled in all of the current configurations, then it works.

ximinez:
Squashed and set the "passed" flag.

PR #3450, opened 2020-06-11
crawl_shards comment fix

mDuo13:
Corrects the public_key parameter name in the comment.
See https://github.com/ripple/xrpl-dev-portal/pull/854 for context.


mDuo13:
It's a multiline comment. But I think the problem is I accidentally opened this PR to the wrong branch (should be to develop)

PR #3453, opened 2020-06-15
Remove CryptoConditionsSuite amendment:

seelabs:
* This was always a stub and enabled no functional changes

### DIFF
--- a/src/ripple/protocol/impl/Feature.cpp
+++ b/src/ripple/protocol/impl/Feature.cpp
@@ -102,7 +102,7 @@ detail::supportedAmendments()
         "TickSize",
         "fix1368",
         "Escrow",
-        // "CryptoConditionsSuite", NO NOT REUSE
+        // "CryptoConditionsSuite", DO NOT REUSE
         "fix1373",
         "EnforceInvariants",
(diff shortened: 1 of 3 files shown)


src/ripple/protocol/impl/Feature.cpp
```diff
@@ -102,7 +102,7 @@ detail::supportedAmendments()
         "TickSize",
         "fix1368",
         "Escrow",
-        "CryptoConditionsSuite",
+        // "CryptoConditionsSuite", NO NOT REUSE
```
intelliot, line 105:
```suggestion
        // "CryptoConditionsSuite", DO NOT REUSE
```

seelabs: Thanks Elliot! Fixed in `10f000c89 [fold] Fix typo
`

mDuo13:
It's a little weird (and perhaps inadvisable) to think about this way, but it's actually not really a technical problem to reuse the amendment name in the future so long as all servers that had the old version of the amendment are already amendment blocked by something else.

Still probably better for everyone's sanity not to reuse the name, though.

PR #3456, opened 2020-06-18
Update tecUNFUNDED & tecUNFUNDED_ADD messages:

mDuo13:
The tecUNFUNDED code is actively used [when attempting to create payment
channels](https://github.com/ripple/rippled/blob/023f5704d07d09e70091f38a0d4e5df213a3144b/src/ripple/app/tx/impl/PayChan.cpp#L205) and [escrows](https://github.com/ripple/rippled/blob/023f5704d07d09e70091f38a0d4e5df213a3144b/src/ripple/app/tx/impl/Escrow.cpp#L212); the messages incorrectly describes it as deprecated. Meanwhile, the tecUNFUNDED_ADD code actually is an unused legacy code, dating back to [when there was a WalletAdd transactor](https://github.com/ripple/rippled/blob/1017adf7438017f520d0539b6d0c116c0207439e/src/ripple_app/tx/WalletAddTransactor.cpp#L73).

Engine result messages are (I think) not part of the binary format and are [documented as subject to change without notice](https://xrpl.org/transaction-results.html#immediate-response), so this should not require an amendment nor a new API version.

Aside: the clang-formatted version of `TER.cpp` is fugly.


nbougalis:
I think it's fine to bypass **`clang-format`** for tabular data structures, like this table. To that end, consider cherry-picking this commit to your branch: 7010ce1f81f518fe9d5eb1c51f6a41c01958ee3b.

I also made 4788d7626abb1e105022460f0f81123a39ab43d4 (which applies on top of the above commit) which reduces the verbosity further and makes this more readable for humans at the cost of a macro.

I don't know if these will pass muster with your reviewers, but my _personal_ opinion is that they're both improvements over the current formatting.

nbougalis:
Consider marking `terLAST` and `terFUNDS_SPENT` as deprecated as well. We should propose a change that simply removes them.

scottschurr:
FWIW, I took a look at @nbougalis's suggestions and I'm good with all of them, including the macro.

mDuo13:
I appreciate the macro and clang formatting; pulled those commits in. And I've added c8f2c6a to mark those errors as deprecated. I didn't want to remove them outright because I think it would change the remaining ter codes' `engine_result_code` numbers. Even though we tell people [the numbers are subject to change](https://xrpl.org/transaction-results.html#immediate-response), I'd rather keep the scope of this PR from growing further and affecting other codes that are unrelated.

scottschurr:
As I was giving this one final look-over, I did have one question occur to me.  `tecUNFUNDED_ADD` is currently deprecated and unused.  However, it's possible that `tecUNFUNDED_ADD` is a result code in old metadata.  Would that be sufficient reason to preserve the old description?  Or do we think that if anyone wants to look at metadata that old they should use an old version of rippled?

At the moment I'm leaning toward preserving the old description, "Insufficient XRP balance for WalletAdd."  But I'm really unsure.

Thoughts?

mDuo13:
I am >90% certain that ` tecUNFUNDED_ADD` does not appear in _any_ historical ledger metadata in the XRPL mainnet history. (Nor in any modern testnet, devnet, etc.)

Based on a search of the Git history, this result code was only ever used with the `WalletAdd` transactor, and it is my understanding that the "WalletAdd" transactor was never used in the current mainnet ledger chain; if it was used at all, it would've been in a pre-release dev chain, before the "final reset". (Mayyyybe @JoelKatz remembers if that's true.)

As far as I know the Data API's importer can't even deserialize `WalletAdd` so it must not have encountered any in the full history data it has available. And if there's no WalletAdd, there's no tecUNFUNDED_ADD. So, to answer your question, @scottschurr, I think we have no reason to preserve the old description.

scottschurr:
Excellent.  Thank you for your thorough research.

PR #3463, opened 2020-06-27
Feature/shard path multiplexing

undertome:
This PR introduces changes that allow a user to specify multiple paths for storing historical shards. The changes are based in the `DatabaseShardImp` class and in the shards unit tests. Please read [the documentation](https://github.com/undertome/rippled/blob/feature/shard-path-multiplexing/src/ripple/nodestore/ShardStoragePaths.md) to get an idea of the changes involved.

### DIFF
--- a/src/ripple/net/ShardDownloader.md
+++ b/src/ripple/net/ShardDownloader.md
@@ -177,7 +177,7 @@ The `download_path` field of the `shard_db` entry is used to determine where to 
 type=NuDB
 path=/var/lib/rippled/db/shards/nudb
 download_path=/var/lib/rippled/db/shards/
-max_size_gb=50
+max_historical_shards=50
 ```
 
 ##### Resuming Partial Downloads
--- a/src/ripple/nodestore/ShardStoragePaths.md
+++ b/src/ripple/nodestore/ShardStoragePaths.md
@@ -1,0 +1,127 @@
+# Shard Storage Paths
+
+## Overview
+
+The shard database stores validated ledgers in logical groups called shards. As
+of June 2020, a shard stores 16384 ledgers by default. In order to allow users
+to store shards on multiple devices, the shard database can be configured with
+several file system paths. Each path provided should refer to a directory on a
+distinct filesystem, and no two paths should ever correspond to the same
+filesystem. Violating this restriction will cause the server to inaccurately
+estimate the amount of space available for storing shards. In the absence of a
+suitable platform agnostic solution, this requirement is enforced only on
+Linux. However, on other platforms we employ a heuristic that issues a warning
+if we suspect that this restriction is violated.
+
+## Configuration
+
+The `shard_db` and `historical_shard_paths` sections of the server's
+configuration file will be used to determine where the server stores shards.
+Minimally, the `shard_db` section must contain a single `path` key.
+If this is the only storage path provided, all shards will be stored at this
+location. If the configuration also lists one or more lines in the
+`historical_shard_paths` section, all older shards will be stored at these
+locations, and the `path` will be used only to store the current
+and previous shards. The goal is to allow users to provide an efficient SSD for
+storing recent shards, as these will be accessed more frequently, while using
+large mechanical drives for storing older shards that will be accessed less
+frequently.
+
+Below is a sample configuration snippet that provides a path for main storage
+and several paths for historical storage:
+
+```dosini
+# This is the persistent datastore for shards. It is important for the health
+# of the ripple network that rippled operators shard as much as practical.
+# NuDB requires SSD storage. Helpful information can be found here
+# https://ripple.com/build/history-sharding
+[shard_db]
+type=NuDB
+
+# A single path for storing
+# the current and previous
+# shards:
+# -------------------------
+path=/var/lib/rippled/db/shards/nudb
+
+# Path where shards are stored
+# while being downloaded:
+# ----------------------------
+download_path=/var/lib/rippled/db/shards/
+
+# The number of historical shards to store.
+# The default value is 0, which means that
+# the server won't store any historical
+# shards - only the current and previous
+# shards will be stored.
+# ------------------------------------
+max_historical_shards=100
+
+# List of paths for storing older shards.
+[historical_shard_paths]
+/mnt/disk1
+/mnt/disk2
+/mnt/disk3
+
+```
+## Shard Migration
+
+When a new shard (*current shard*) is confirmed by the network, the recent
+shards will shift. The *previous shard* will become a *historical shard*, the
+*current shard* will become the *previous shard*, and the new shard will become
+the *current shard*. These are just logical labels, and the shards themselves
+don't change to reflect being current, previous, or historical. However, if the
+server's configuration specifies one or more paths for historical storage,
+during this shift the formerly *previous shard* will be migrated to one of the
+historical paths. If multiple paths are provided, the server dynamically
+chooses one with sufficient space for storing the shard.
+
+**Note:** As of June 2020, the shard database does not store the partial shard
+currently being built by live network transactions, but this is planned to
+change. When this feature is implemented, the *current shard* will refer to this
+partial shard, and the *previous shard* will refer to the most recently
+validated shard.
+
+### Selecting a Historical Storage Path
+
+When storing historical shards, if multiple historical paths are provided, the
+path to use for each shard will be selected in a random fashion. By using all
+available storage devices, we create a uniform distribution of disk utilization
+for disks of equivalent size, (provided that the disks are used only to store
+shards). In theory, selecting devices in this manner will also increase our
+chances for concurrent access to stored shards, however as of June 2020
+concurrent shard access is not implemented. Lastly, a storage path is included
+in the random distribution only if it has enough storage capacity to hold the
+next shard.
+
+## Shard Acquisition
+
+When the server is acquiring shard history, these acquired shards will be stored
+at a path designated for historical storage (`historical_storage_path`). If no
+such path is provided, acquired shards will be stored at the
+`path`.
+
+## Storage capacity
+
+### Filesystem Capacity
+
+When the shard database updates its record of disk utilization, it trusts that
+the provided historical paths refer to distinct devices, or at least distinct
+filesystems. If this requirement is violated, the database will operate with an
+inaccurate view of how many shards it can store. Violation of this requirement
+won't necessarily impede database operations, but the database will fail to
+identify scenarios wherein storing the maximum number of historical shards (as
+per the 'historical_shard_count' parameter in the configuration file) would
+exceed the amount of storage space available.
+
+### Shard Migration
+
+During a "recent shard shift", if the server has already reached the configured
+limit of stored historical shards, instead of moving the formerly *previous
+shard* to a historical drive (or keeping it at the 'path') the
+shard will be dropped and removed from the filesystem.
+
+### Shard Acquisition
+
+Once the configured limit of stored historical shards has been reached, shard
+acquisition halts, and no additional shards will be acquired.
--- a/src/ripple/nodestore/impl/DatabaseShardImp.cpp
+++ b/src/ripple/nodestore/impl/DatabaseShardImp.cpp
@@ -32,6 +32,10 @@
 
 #include <boost/algorithm/string/predicate.hpp>
 
+#if BOOST_OS_LINUX
+#include <sys/statvfs.h>
+#endif
+
 namespace ripple {
 namespace NodeStore {
 
@@ -82,95 +86,135 @@ DatabaseShardImp::init()
         try
         {
             using namespace boost::filesystem;
-            if (exists(dir_))
+
+            // Consolidate the main storage path and all
+            // historical paths
+            std::vector<path> paths{dir_};
+            paths.insert(
+                paths.end(), historicalPaths_.begin(), historicalPaths_.end());
+
+            for (auto const& path : paths)
             {
-                if (!is_directory(dir_))
+                try
                 {
-                    JLOG(j_.error()) << "'path' must be a directory";
+                    if (exists(path))
+                    {
+                        if (!is_directory(path))
+                        {
+                            JLOG(j_.error()) << path << " must be a directory";
+                            return false;
+                        }
+                    }
+                    else if (!create_directories(path))
+                    {
+                        JLOG(j_.error())
+                            << "failed to create path: " + path.string();
+                        return false;
+                    }
+                }
+                catch (...)
+                {
+                    JLOG(j_.error())
+                        << "failed to create path: " + path.string();
                     return false;
                 }
             }
-            else
-                create_directories(dir_);
+
+            if (!app_.config().standalone() && !historicalPaths_.empty())
+            {
+                // Check historical paths for duplicated
+                // filesystems
+                if (!checkHistoricalPaths())
+                    return false;
+            }
 
             ctx_ = std::make_unique<nudb::context>();
             ctx_->start();
 
             // Find shards
-            for (auto const& d : directory_iterator(dir_))
+            for (auto const& path : paths)
             {
-                if (!is_directory(d))
-                    continue;
+                for (auto const& d : directory_iterator(path))
+                {
+                    if (!is_directory(d))
+                        continue;
 
-                // Check shard directory name is numeric
-                auto dirName = d.path().stem().string();
-                if (!std::all_of(dirName.begin(), dirName.end(), [](auto c) {
-                        return ::isdigit(static_cast<unsigned char>(c));
-                    }))
-                {
-                    continue;
-                }
+                    auto const shardDir = d.path();
 
-                auto const shardIndex{std::stoul(dirName)};
-                if (shardIndex < earliestShardIndex())
-                {
-                    JLOG(j_.error()) << "shard " << shardIndex
-                                     << " comes before earliest shard index "
-                                     << earliestShardIndex();
-                    return false;
-                }
+                    // Check shard directory name is numeric
+                    auto dirName = shardDir.stem().string();
+                    if (!std::all_of(
+                            dirName.begin(), dirName.end(), [](auto c) {
+                                return ::isdigit(static_cast<unsigned char>(c));
+                            }))
+                    {
+                        continue;
+                    }
 
-                auto const shardDir{dir_ / std::to_string(shardIndex)};
-
-                // Check if a previous import failed
-                if (is_regular_file(shardDir / importMarker_))
-                {
-                    JLOG(j_.warn()) << "shard " << shardIndex
-                                    << " previously failed import, removing";
-                    remove_all(shardDir);
-                    continue;
-                }
-
-                auto shard{
-                    std::make_unique<Shard>(app_, *this, shardIndex, j_)};
-                if (!shard->open(scheduler_, *ctx_))
-                {
-                    // Remove corrupted or legacy shard
-                    shard->removeOnDestroy();
-                    JLOG(j_.warn())
-                        << "shard " << shardIndex << " removed, "
-                        << (shard->isLegacy() ? "legacy" : "corrupted")
-                        << " shard";
-                    continue;
-                }
-
-                if (shard->isFinal())
-                {
-                    shards_.emplace(
-                        shardIndex,
-                        ShardInfo(std::move(shard), ShardInfo::State::final));
-                }
-                else if (shard->isBackendComplete())
-                {
-                    auto const result{shards_.emplace(
-                        shardIndex,
-                        ShardInfo(std::move(shard), ShardInfo::State::none))};
-                    finalizeShard(
-                        result.first->second, true, lock, boost::none);
-                }
-                else
-                {
-                    if (acquireIndex_ != 0)
+                    auto const shardIndex{std::stoul(dirName)};
+                    if (shardIndex < earliestShardIndex())
                     {
                         JLOG(j_.error())
-                            << "more than one shard being acquired";
+                            << "shard " << shardIndex
+                            << " comes before earliest shard index "
+                            << earliestShardIndex();
                         return false;
                     }
 
-                    shards_.emplace(
-                        shardIndex,
-                        ShardInfo(std::move(shard), ShardInfo::State::acquire));
-                    acquireIndex_ = shardIndex;
+                    // Check if a previous import failed
+                    if (is_regular_file(shardDir / importMarker_))
+                    {
+                        JLOG(j_.warn())
+                            << "shard " << shardIndex
+                            << " previously failed import, removing";
+                        remove_all(shardDir);
+                        continue;
+                    }
+
+                    auto shard{std::make_unique<Shard>(
+                        app_, *this, shardIndex, shardDir.parent_path(), j_)};
+                    if (!shard->open(scheduler_, *ctx_))
+                    {
+                        // Remove corrupted or legacy shard
+                        shard->removeOnDestroy();
+                        JLOG(j_.warn())
+                            << "shard " << shardIndex << " removed, "
+                            << (shard->isLegacy() ? "legacy" : "corrupted")
+                            << " shard";
+                        continue;
+                    }
+
+                    if (shard->isFinal())
+                    {
+                        shards_.emplace(
+                            shardIndex,
+                            ShardInfo(
+                                std::move(shard), ShardInfo::State::final));
+                    }
+                    else if (shard->isBackendComplete())
+                    {
+                        auto const result{shards_.emplace(
+                            shardIndex,
+                            ShardInfo(
+                                std::move(shard), ShardInfo::State::none))};
+                        finalizeShard(
+                            result.first->second, true, lock, boost::none);
+                    }
+                    else
+                    {
+                        if (acquireIndex_ != 0)
+                        {
+                            JLOG(j_.error())
+                                << "more than one shard being acquired";
+                            return false;
+                        }
+
+                        shards_.emplace(
+                            shardIndex,
+                            ShardInfo(
+                                std::move(shard), ShardInfo::State::acquire));
+                        acquireIndex_ = shardIndex;
+                    }
                 }
             }
         }
@@ -178,6 +222,7 @@ DatabaseShardImp::init()
         {
             JLOG(j_.error())
                 << "exception " << e.what() << " in function " << __func__;
+            return false;
         }
 
         updateStatus(lock);
@@ -209,20 +254,6 @@ DatabaseShardImp::prepareLedger(std::uint32_t validLedgerSeq)
         if (!canAdd_)
             return boost::none;
 
-        // Check available storage space
-        if (fileSz_ + avgShardFileSz_ > maxFileSz_)
-        {
-            JLOG(j_.debug()) << "maximum storage size reached";
-            canAdd_ = false;
-            return boost::none;
-        }
-        if (avgShardFileSz_ > available())
-        {
-            JLOG(j_.error()) << "insufficient storage space available";
-            canAdd_ = false;
-            return boost::none;
-        }
-
         shardIndex = findAcquireIndex(validLedgerSeq, lock);
     }
 
@@ -236,7 +267,27 @@ DatabaseShardImp::prepareLedger(std::uint32_t validLedgerSeq)
         return boost::none;
     }
 
-    auto shard{std::make_unique<Shard>(app_, *this, *shardIndex, j_)};
+    auto const pathDesignation = [this, shardIndex = *shardIndex]() {
+        std::lock_guard lock(mutex_);
+        return prepareForNewShard(shardIndex, numHistoricalShards(lock), lock);
+    }();
+
+    if (!pathDesignation)
+        return false;
+
+    auto const needsHistoricalPath =
+        *pathDesignation == PathDesignation::historical;
+
+    auto shard = [this, shardIndex, needsHistoricalPath] {
+        std::lock_guard lock(mutex_);
+        return std::make_unique<Shard>(
+            app_,
+            *this,
+            *shardIndex,
+            (needsHistoricalPath ? chooseHistoricalPath(lock) : ""),
+            j_);
+    }();
+
     if (!shard->open(scheduler_, *ctx_))
         return boost::none;
 
@@ -292,10 +343,19 @@ DatabaseShardImp::prepareShard(std::uint32_t shardIndex)
         return false;
     }
 
-    // Check available storage space
-    if (fileSz_ + avgShardFileSz_ > maxFileSz_)
-        return fail("maximum storage size reached");
-    if (avgShardFileSz_ > available())
+    // Any shard earlier than the two most recent shards
+    // is a historical shard
+    bool const isHistoricalShard = shardIndex < shardBoundaryIndex(lock);
+    auto const numHistShards = numHistoricalShards(lock);
+
+    // Check shard count and available storage space
+    if (isHistoricalShard && numHistShards >= maxHistoricalShards_)
+        return fail("maximum number of historical shards reached");
+    if (!sufficientStorage(
+            1,
+            isHistoricalShard ? PathDesignation::historical
+                              : PathDesignation::none,
+            lock))
         return fail("insufficient storage space available");
 
     shards_.emplace(shardIndex, ShardInfo(nullptr, ShardInfo::State::import));
@@ -392,15 +452,30 @@ DatabaseShardImp::importShard(
             return false;
         }
 
-        dstDir = dir_ / std::to_string(shardIndex);
+        auto const pathDesignation =
+            prepareForNewShard(shardIndex, numHistoricalShards(lock), lock);
+
+        if (!pathDesignation)
+        {
+            JLOG(j_.error()) << "shard " << shardIndex << " failed to import";
+            return false;
+        }
+
+        auto const needsHistoricalPath =
+            *pathDesignation == PathDesignation::historical;
+        dstDir = needsHistoricalPath ? chooseHistoricalPath(lock) : dir_;
     }
+
+    dstDir /= std::to_string(shardIndex);
 
     // Rename source directory to the shard database directory
     if (!renameDir(srcDir, dstDir))
         return false;
 
     // Create the new shard
-    auto shard{std::make_unique<Shard>(app_, *this, shardIndex, j_)};
+    auto shard{std::make_unique<Shard>(
+        app_, *this, shardIndex, dstDir.parent_path(), j_)};
+
     if (!shard->open(scheduler_, *ctx_) || !shard->isBackendComplete())
     {
         JLOG(j_.error()) << "shard " << shardIndex << " failed to import";
@@ -688,23 +763,21 @@ DatabaseShardImp::import(Database& source)
             }
         }
 
+        auto numHistShards = this->numHistoricalShards(lock);
+
         // Import the shards
         for (std::uint32_t shardIndex = earliestIndex;
              shardIndex <= latestIndex;
              ++shardIndex)
         {
-            if (fileSz_ + avgShardFileSz_ > maxFileSz_)
-            {
-                JLOG(j_.error()) << "maximum storage size reached";
-                canAdd_ = false;
+            auto const pathDesignation =
+                prepareForNewShard(shardIndex, numHistShards, lock);
+
+            if (!pathDesignation)
                 break;
-            }
-            if (avgShardFileSz_ > available())
-            {
-                JLOG(j_.error()) << "insufficient storage space available";
-                canAdd_ = false;
-                break;
-            }
+
+            auto const needsHistoricalPath =
+                *pathDesignation == PathDesignation::historical;
 
             // Skip if already stored
             if (shardIndex == acquireIndex_ ||
@@ -741,13 +814,18 @@ DatabaseShardImp::import(Database& source)
                     continue;
             }
 
+            auto const path =
+                needsHistoricalPath ? chooseHistoricalPath(lock) : dir_;
+
             // Create the new shard
-            auto shard{std::make_unique<Shard>(app_, *this, shardIndex, j_)};
+            auto shard =
+                std::make_unique<Shard>(app_, *this, shardIndex, path, j_);
+
             if (!shard->open(scheduler_, *ctx_))
                 continue;
 
             // Create a marker file to signify an import in progress
-            auto const shardDir{dir_ / std::to_string(shardIndex)};
+            auto const shardDir{path / std::to_string(shardIndex)};
             auto const markerFile{shardDir / importMarker_};
             {
                 std::ofstream ofs{markerFile.string()};
@@ -818,6 +896,9 @@ DatabaseShardImp::import(Database& source)
                         ShardInfo(std::move(shard), ShardInfo::State::none))};
                     finalizeShard(
                         result.first->second, true, lock, boost::none);
+
+                    if (shardIndex < shardBoundaryIndex(lock))
+                        ++numHistShards;
                 }
                 catch (std::exception const& e)
                 {
@@ -1074,19 +1155,28 @@ DatabaseShardImp::initConfig(std::lock_guard<std::mutex>&)
         return fail("'path' missing");
 
     {
-        std::uint64_t sz;
-        if (!get_if_exists<std::uint64_t>(section, "max_size_gb", sz))
-            return fail("'max_size_gb' missing");
+        get_if_exists(section, "max_historical_shards", maxHistoricalShards_);
 
-        if ((sz << 30) < sz)
-            return fail("'max_size_gb' overflow");
+        Section const& historicalShardPaths =
+            config.section(SECTION_HISTORICAL_SHARD_PATHS);
 
-        // Minimum storage space required (in gigabytes)
-        if (sz < 10)
-            return fail("'max_size_gb' must be at least 10");
+        auto values = historicalShardPaths.values();
 
-        // Convert to bytes
-        maxFileSz_ = sz << 30;
+        std::sort(values.begin(), values.end());
+        values.erase(std::unique(values.begin(), values.end()), values.end());
+
+        for (auto const& s : values)
+        {
+            auto const dir = path(s);
+            if (dir_ == dir)
+            {
+                return fail(
+                    "the 'path' cannot also be in the  "
+                    "'historical_shard_path' section");
+            }
+
+            historicalPaths_.push_back(s);
+        }
     }
 
     if (section.exists("ledgers_per_shard"))
@@ -1245,6 +1335,43 @@ DatabaseShardImp::finalizeShard(
                 return;
             it->second.state = ShardInfo::State::final;
             updateStatus(lock);
+
+            auto const boundaryIndex = shardBoundaryIndex(lock);
+            auto const isHistoricalShard = shardIndex < boundaryIndex;
+
+            if (isHistoricalShard)
+            {
+                if (!historicalPaths_.empty() &&
+                    shard->getDir().parent_path() == dir_)
+                {
+                    // This is a historical shard that wasn't
+                    // placed at a separate historical path
+                    JLOG(j_.warn()) << "shard " << shardIndex
+                                    << " is not stored at a historical path";
+                }
+            }
+
+            else
+            {
+                // Not a historical shard. Shift recent shards
+                // if necessary
+                relocateOutdatedShards(lock);
+                assert(!boundaryIndex || shardIndex - boundaryIndex <= 1);
+
+                auto& recentShard = shardIndex == boundaryIndex
+                    ? secondLatestShardIndex_
+                    : latestShardIndex_;
+
+                // Set the appropriate recent shard
+                // index
+                recentShard = shardIndex;
+
+                if (shard->getDir().parent_path() != dir_)
+                {
+                    JLOG(j_.warn()) << "shard " << shard->index()
+                                    << " is not stored at the path";
+                }
+            }
         }
 
         setFileStats();
@@ -1298,12 +1425,16 @@ DatabaseShardImp::setFileStats()
     fdRequired_ = sumFd;
     avgShardFileSz_ = (numShards == 0 ? fileSz_ : fileSz_ / numShards);
 
-    if (fileSz_ >= maxFileSz_)
+    if (auto const count = numHistoricalShards(lock);
+        count >= maxHistoricalShards_)
     {
-        JLOG(j_.warn()) << "maximum storage size reached";
+        JLOG(j_.warn()) << "maximum number of historical shards reached";
         canAdd_ = false;
     }
-    else if (maxFileSz_ - fileSz_ > available())
+    else if (!sufficientStorage(
+                 maxHistoricalShards_ - count,
+                 PathDesignation::historical,
+                 lock))
     {
         JLOG(j_.warn())
             << "maximum shard store size exceeds available storage space";
@@ -1350,19 +1481,59 @@ DatabaseShardImp::getCache(std::uint32_t seq)
     return std::make_pair(pCache, nCache);
 }
 
-std::uint64_t
-DatabaseShardImp::available() const
+bool
+DatabaseShardImp::sufficientStorage(
+    std::uint32_t numShards,
+    PathDesignation pathDesignation,
+    std::lock_guard<std::mutex> const&) const
 {
     try
     {
-        return boost::filesystem::space(dir_).available;
+        std::vector<std::uint64_t> capacities;
+
+        if (pathDesignation == PathDesignation::historical &&
+            !historicalPaths_.empty())
+        {
+            capacities.reserve(historicalPaths_.size());
+
+            for (auto const& path : historicalPaths_)
+            {
+                // Get the available storage for each historical path
+                auto const availableSpace =
+                    boost::filesystem::space(path).available;
+
+                capacities.push_back(availableSpace);
+            }
+        }
+        else
+        {
+            // Get the available storage for the main shard path
+            capacities.push_back(boost::filesystem::space(dir_).available);
+        }
+
+        for (std::uint64_t const capacity : capacities)
+        {
+            // Leverage all the historical shard paths to
+            // see if collectively they can fit the specified
+            // number of shards. For this to work properly,
+            // each historical path must correspond to a separate
+            // physical device or filesystem.
+
+            auto const shardCap = capacity / avgShardFileSz_;
+            if (numShards <= shardCap)
+                return true;
+
+            numShards -= shardCap;
+        }
     }
     catch (std::exception const& e)
     {
         JLOG(j_.error()) << "exception " << e.what() << " in function "
                          << __func__;
-        return 0;
+        return false;
     }
+
+    return false;
 }
 
 bool
@@ -1402,7 +1573,7 @@ DatabaseShardImp::storeLedgerInShard(
 }
 
 void
-DatabaseShardImp::removeFailedShard(std::shared_ptr<Shard> shard)
+DatabaseShardImp::removeFailedShard(std::shared_ptr<Shard>& shard)
 {
     {
         std::lock_guard lock(mutex_);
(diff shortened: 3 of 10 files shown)


src/ripple/nodestore/ShardStoragePaths.md
```diff
@@ -0,0 +1,56 @@
...
+# The number of historical shards to
+# store. The default value is 0, which
+# means that only the current and
+# previous shards will be stored.
+# ------------------------------------
+max_historical_shards=100
+
+# List of paths for storing older shards.
+[historical_shard_paths]
+/dev/sdb1
```
MarkusTeufelberger, line 41: I doubt that shards will be stored on raw devices? Might need a better path like `/mnt/disk1` for example.

src/ripple/nodestore/ShardStoragePaths.md
```diff
@@ -0,0 +1,56 @@
...
+/dev/sdb3
+
+```
+## Shard Migration
+
+When a new shard (*current shard*) is confirmed by the network, the recent shards will shift. The next shard to build will become the *current shard* and the *current shard* will become the *previous shard*. These are just logical labels, and the shards themselves don't change to reflect being current or previous. However, if the server's configuration specifies one or more paths for historical storage, during this shift the formerly *previous shard* will be migrated to one of the historical paths. If multiple paths are provided, the server dynamically chooses one with sufficient space for storing the shard.
+
+### Selecting a Historical Storage Path
+
+When storing historical shards, if multiple historical paths are provided, the paths will be selected in a round-robin fashion. By using all available storage devices, we increase our chances for concurrent access to stored shards. Naturally, a storage path is included in the round-robin distribution only if it has enough storage capacity to hold the next shard.
```
MarkusTeufelberger, line 52: This would cause issues if I for example have 3 paths configured, the respective disks start getting full and I add a 4th one. I would recommend to store the next shard at the disk that has the least space used by shards (still an issue if disk sizes are different, but less problematic than the "proper" solution of choosing the disk that has the most free space) - or, if getting sizes is too difficult, then at least choosing the disk that has the least amount of shard files stored, with a random selection (or just the order in the config file) as tie-breaker.

src/ripple/nodestore/ShardStoragePaths.md
undertome: Done

src/ripple/nodestore/ShardStoragePaths.md
undertome: The selection process outlined here is actually outdated - during the implementation I decided to go with random selection. The doc in the most recent push has been updated with that change.

I think the random selection is sufficient for our needs. It's simple (which is good) and does the job well enough. If the motivation for using capacity-based selection is to increase performance by reducing file fragmentation, I don't think its worth implementing. For any server that stores limited history, shard fragmentation is inevitable and eventually all of its historical shards will be fragmented. And for a server that stores full history, in theory shard fragmentation won't occur. In either case I don't see a reason for space-based selection.

src/ripple/nodestore/ShardStoragePaths.md
```diff
@@ -0,0 +1,73 @@
+# Shard Storage Paths
+
+## Overview
+
+The shard database stores validated ledgers in logical groups called shards. As of June 2020, a shard stores 16384 ledgers by default. In order to allow users to store shards on multiple devices, the shard database can be configured with several file system paths. Each path provided should refer to a directory on a distinct filesystem, and no two paths should ever correspond to the same filesystem. Violating this restriction will cause the server to inaccurately estimate the amount of space available for storing shards.
```
seelabs, line 5: Please limit lines to 80 characters. Rational: one advantage of markdown is it may be read in raw form. The long lines make it more difficult to read in the terminal and in some editors. Also, long lines make it more difficult to leave review comments localized to the area of the issue.

seelabs: For the sentence:
```
Each path provided should refer to a directory on a distinct filesystem, 
and no two paths should ever correspond to the same filesystem.
```
We should be able to check this condition in code - for linux at least - and either error out or warn if the condition isn't met.

src/ripple/nodestore/impl/DatabaseShardImp.cpp
```diff
@@ -1419,6 +1595,266 @@ DatabaseShardImp::removeFailedShard(std::shared_ptr<Shard> shard)
...
+                    }
+                }
+
+                cur = boost::none;
+            }
+        }
+    }
+}
+
+boost::optional<bool>
```
seelabs, line 1788: `optional<bool>` is error prone to use. It's too easy to use the outer `bool` instead of the inner `bool`. I'd recommend replacing the inner `bool` with an enum.

src/ripple/nodestore/impl/DatabaseShardImp.cpp
```diff
@@ -1350,12 +1484,48 @@ DatabaseShardImp::getCache(std::uint32_t seq)
     return std::make_pair(pCache, nCache);
 }
 
-std::uint64_t
-DatabaseShardImp::available() const
+bool
+DatabaseShardImp::sufficientStorage(
+    std::uint32_t numShards,
+    bool isHistorical,
```
seelabs, line 1490: Replacing the `bool` with an enum would make the calling sites clearer.

src/ripple/nodestore/impl/DatabaseShardImp.cpp
```diff
@@ -1350,12 +1484,48 @@ DatabaseShardImp::getCache(std::uint32_t seq)
...
+                   // number of shards. For this to work properly,
+                   // each historical path must correspond to a separate
+                   // physical device.
+                   [&numShards, this](std::uint64_t const& capacity) {
+                       numShards -= std::min(
+                           capacity / avgShardFileSz_,
+                           std::uint64_t(numShards));
+
+                       return numShards == 0;
+                   }) != capacities.end();
```
seelabs, line 1528: While clever, we aren't using `find_if` to find anything. It's likely people will need to decode this. What do you think about calculating the available number of shards and checking if there's enough space that way instead. I.e.
```

        std::uint64_t availSpacsShards = 0;
        for (auto const c : capacities)
        {
            availSpacsShards += c / avgShardFileSz_;
        }
        return availSpacsShards >= numShards;
```
That code does not exit early (tho it's easy enough to do that). But given the size of `capacities` it won't matter.

src/ripple/nodestore/impl/DatabaseShardImp.cpp
```diff
@@ -1350,12 +1484,48 @@ DatabaseShardImp::getCache(std::uint32_t seq)
...
+        // Get the available storage for each historical path
+        if (isHistorical && !historicalPaths_.empty())
+        {
+            std::for_each(
+                historicalPaths_.begin(),
+                historicalPaths_.end(),
+                [&capacities](boost::filesystem::path const& path) {
+                    capacities.push_back(
+                        boost::filesystem::space(path).available);
+                });
```
seelabs, line 1506: A range for loop feels simpler for this. I.e.
```
            capacities.reserve(historicalPaths_.size());
            for (auto const& p : historicalPaths_)
                capacities.push_back(boost::filesystem::space(p).available);
```
Note, even if we end up keeping the `for_each`, we should still `reserve`

src/ripple/nodestore/impl/DatabaseShardImp.cpp
```diff
@@ -1350,12 +1484,48 @@ DatabaseShardImp::getCache(std::uint32_t seq)
...
+    std::uint32_t numShards,
+    bool isHistorical,
+    std::lock_guard<std::mutex> const&) const
 {
     try
     {
-        return boost::filesystem::space(dir_).available;
+        std::vector<std::uint64_t> capacities;
+
+        // Get the available storage for each historical path
```
seelabs, line 1497: Nit: We usually put comment pertaining to the block inside the block, not before the block (for the `else` block as well)

src/ripple/nodestore/impl/DatabaseShardImp.cpp
seelabs, line 1528: Since you're modifying this file, Line 1534 (not part of this changeset) should be return false not return 0.

src/ripple/nodestore/impl/DatabaseShardImp.cpp
```diff
@@ -292,10 +309,15 @@ DatabaseShardImp::prepareShard(std::uint32_t shardIndex)
...
     }
 
-    // Check available storage space
-    if (fileSz_ + avgShardFileSz_ > maxFileSz_)
-        return fail("maximum storage size reached");
-    if (avgShardFileSz_ > available())
+    // Any shard earlier than the two most recent shards
+    // is a historical shard
+    bool isHistoricalShard = shardIndex < getShardBoundaryIndex(lock);
+    auto numHistoricalShards = countHistoricalShards(lock);
```
seelabs, line 315: The above two vars could be `const`

src/ripple/nodestore/impl/DatabaseShardImp.cpp
```diff
@@ -392,15 +414,28 @@ DatabaseShardImp::importShard(
...
     // Rename source directory to the shard database directory
-    if (!renameDir(srcDir, dstDir))
+    if (!renameDir(srcDir, dstDir / std::to_string(shardIndex)))
         return false;
 
     // Create the new shard
-    auto shard{std::make_unique<Shard>(app_, *this, shardIndex, j_)};
+    auto shard{std::make_unique<Shard>(app_, *this, shardIndex, j_, dstDir)};
+
+    dstDir /= std::to_string(shardIndex);
```
seelabs, line 437: We convert `shardIndex` to a string in two places. We may want to create a local var for this (we do it here and line 431).

src/ripple/nodestore/impl/DatabaseShardImp.cpp
```diff
@@ -1070,23 +1110,74 @@ DatabaseShardImp::initConfig(std::lock_guard<std::mutex>&)
     }
 
     using namespace boost::filesystem;
-    if (!get_if_exists<path>(section, "path", dir_))
-        return fail("'path' missing");
+    if (!get_if_exists<path>(section, "main_storage_path", dir_))
```
seelabs, line 1113: Does this break existing config files?

src/ripple/nodestore/impl/DatabaseShardImp.cpp
```diff
@@ -1070,23 +1110,74 @@ DatabaseShardImp::initConfig(std::lock_guard<std::mutex>&)
...
-        if (sz < 10)
-            return fail("'max_size_gb' must be at least 10");
+        std::sort(values.begin(), values.end());
+        values.erase(std::unique(values.begin(), values.end()), values.end());
 
-        // Convert to bytes
-        maxFileSz_ = sz << 30;
+        std::string hPathError = "";
+
+        if (std::find_if(
```
seelabs, line 1132: I'd rewrite this so it doesn't use `find_if`. Rational: 1) we aren't finding anything, so it's easy to misread the code, 2) Normally the lambdas to `find_if` are pure, here the lambda is creating directories 3) Even if the lambda were pure, an explicit range for loop is something most developers are used to seeing for code like this. Unless there's a compelling reason, it's good to use idioms other developers expect.

src/ripple/nodestore/impl/DatabaseShardImp.cpp
```diff
@@ -1298,12 +1430,14 @@ DatabaseShardImp::setFileStats()
     fdRequired_ = sumFd;
     avgShardFileSz_ = (numShards == 0 ? fileSz_ : fileSz_ / numShards);
 
-    if (fileSz_ >= maxFileSz_)
+    if (auto count = countHistoricalShards(lock); count >= maxHistoricalShards_)
```
seelabs, line 1433: Nit: `const`

src/ripple/nodestore/impl/DatabaseShardImp.cpp
```diff
@@ -1419,6 +1595,266 @@ DatabaseShardImp::removeFailedShard(std::shared_ptr<Shard> shard)
...
+
+std::uint32_t
+DatabaseShardImp::countHistoricalShards(std::lock_guard<std::mutex> const& lock)
+{
+    auto const recentShardBoundaryIndex = getShardBoundaryIndex(lock);
+
+    return std::count_if(
+        shards_.begin(),
+        shards_.end(),
+        [recentShardBoundaryIndex](decltype(shards_)::value_type const& entry) {
```
seelabs, line 1619: Could use `auto const& entry` for the param type here and get rid of the `decltype`

src/ripple/nodestore/impl/DatabaseShardImp.cpp
```diff
@@ -1419,6 +1595,266 @@ DatabaseShardImp::removeFailedShard(std::shared_ptr<Shard> shard)
...
+        cur || prev)
+    {
+        auto const latestShardIndex = NodeStore::seqToShardIndex(
+            app_.getLedgerMaster().getValidLedgerIndex(), ledgersPerShard_);
+
+        auto const separateHistoricalPath = !historicalPaths_.empty();
+
+        //////////////////////////////////////////////////////////////////
+        // Lambdas used inside of this method:
+        auto const removeShard = [this](std::uint32_t const shardIndex) {
```
seelabs, line 1637: Consider adding return types for all these lambdas. We often omit return types, for when there are several helper functions like this I'd prefer to add them.

src/ripple/nodestore/impl/DatabaseShardImp.cpp
```diff
@@ -1419,6 +1595,266 @@ DatabaseShardImp::removeFailedShard(std::shared_ptr<Shard> shard)
...
+    std::array<boost::filesystem::path, 1> historicalShardPath;
+    std::vector<boost::filesystem::path> potentialPaths;
+
+    std::for_each(
+        historicalPaths_.begin(),
+        historicalPaths_.end(),
+        [this, &potentialPaths](boost::filesystem::path const& path) {
+            if (boost::filesystem::space(path).available >= avgShardFileSz_)
+                potentialPaths.push_back(path);
+        });
```
seelabs, line 1840: Consider using a range based for loop instead. It's shorter, and more common than `for_each` in our code.

src/ripple/nodestore/impl/DatabaseShardImp.cpp
```diff
@@ -1419,6 +1595,266 @@ DatabaseShardImp::removeFailedShard(std::shared_ptr<Shard> shard)
...
+
+boost::filesystem::path
+DatabaseShardImp::chooseHistoricalPath(std::lock_guard<std::mutex> const&)
+{
+    // If not configured with separate historical paths,
+    // use the main path (dir_) by default.
+    if (historicalPaths_.empty())
+        return dir_;
+
+    std::array<boost::filesystem::path, 1> historicalShardPath;
```
seelabs, line 1831: Instead of a single element array, consider just `boost::filesystem::path historicalShardPath;` When you need an iterator for `std::sample` take the address. I.e.
```
    std::sample(
        potentialPaths.begin(),
        potentialPaths.end(),
        &historicalShardPath,
        1,
        default_prng());
```

src/ripple/nodestore/impl/DeterministicShard.h
```diff
@@ -61,7 +61,8 @@ class DeterministicShard
         DatabaseShard const& db,
         std::uint32_t index,
         uint256 const& lastHash,
-        beast::Journal j);
+        beast::Journal j,
+        boost::filesystem::path const& dir);
```
seelabs, line 65: `journal` should be the last parameter

src/ripple/nodestore/impl/Shard.h
```diff
@@ -56,13 +56,17 @@ class Shard final
         Application& app,
         DatabaseShard const& db,
         std::uint32_t index,
-        beast::Journal j);
+        beast::Journal j,
+        boost::filesystem::path const& dir = "");
```
seelabs, line 60: `journal` are usually last parameter. In this, default parameters won't work with that. We'll either two functions or we can keep `journal` where it is. Your call.

src/ripple/rpc/impl/ShardArchiveHandler.cpp
```diff
@@ -37,9 +37,9 @@ boost::filesystem::path
 ShardArchiveHandler::getDownloadDirectory(Config const& config)
 {
     return get(config.section(ConfigSection::shardDatabase()),
-               "download_path",
+               "partial_shard_download_path",
```
seelabs, line 40: Does this break existing config files?

src/ripple/nodestore/ShardStoragePaths.md
undertome: > Please limit lines to 80 characters. Rational: one advantage of markdown is it may be read in raw form. The long lines make it more difficult to read in the terminal and in some editors. Also, long lines make it more difficult to leave review comments localized to the area of the issue.

Done

undertome: > For the sentence:
> 
> ```
> Each path provided should refer to a directory on a distinct filesystem, 
> and no two paths should ever correspond to the same filesystem.
> ```
> 
> We should be able to check this condition in code - for linux at least - and either error out or warn if the condition isn't met.

Added this check for Linux. Errors out during initialization.

src/ripple/nodestore/impl/DatabaseShardImp.cpp
undertome: Done

src/ripple/nodestore/impl/DatabaseShardImp.cpp
undertome: Done

src/ripple/nodestore/impl/DatabaseShardImp.cpp
undertome: Done

src/ripple/nodestore/impl/DatabaseShardImp.cpp
undertome: Done

src/ripple/nodestore/impl/DatabaseShardImp.cpp
undertome: Done

src/ripple/nodestore/impl/DatabaseShardImp.cpp
undertome: Yes

src/ripple/rpc/impl/ShardArchiveHandler.cpp
undertome: I guess that depends on what you mean by break. A server whose config file only provides the oudated property name (`download_path`) will use a default path as the `partial_download_path` which may or may not reflect whatever path was specified with the outdated name.

src/ripple/nodestore/impl/DeterministicShard.h
undertome: I think this will get removed since `DeterministicShards` are getting pulled from the release.

src/ripple/nodestore/impl/DatabaseShardImp.cpp
undertome: Done

src/ripple/nodestore/impl/DatabaseShardImp.cpp
undertome: Changed to a for loop.

src/ripple/nodestore/impl/DatabaseShardImp.cpp
undertome: Added reserve.

undertome:
@mDuo13 This PR contains changes to two `shard_db` properties:

1.  `path` -> `main_storage_path` **(breaking change)**
2. `download_path` -> `partial_shard_download_path`

src/ripple/nodestore/impl/DatabaseShardImp.cpp
undertome: Done

src/ripple/nodestore/impl/DatabaseShardImp.cpp
undertome: Done

src/ripple/nodestore/impl/DatabaseShardImp.cpp
undertome: Done

src/ripple/nodestore/impl/DatabaseShardImp.cpp
undertome: Done

src/ripple/nodestore/impl/DatabaseShardImp.cpp
undertome: Done

src/ripple/nodestore/impl/DatabaseShardImp.cpp
undertome: Done

src/ripple/nodestore/impl/DatabaseShardImp.cpp
undertome: Done

src/test/nodestore/DatabaseShard_test.cpp
```diff
@@ -162,6 +160,7 @@ randInt(Engine& engine, Integral max)
 class DatabaseShard_test : public TestBase
 {
     static constexpr std::uint32_t maxSizeGb = 10;
+    static constexpr std::uint32_t maxHistoricalShards = 100;
```
miguelportilla, line 163: This file needs to `#include <numeric>`

src/test/nodestore/DatabaseShard_test.cpp
```diff
@@ -1185,6 +1191,320 @@ class DatabaseShard_test : public TestBase
...
+
+            auto const mainPathCount = std::distance(
+                boost::filesystem::directory_iterator(shardDir.path()),
+                boost::filesystem::directory_iterator());
+
+            // Only the two most recent shards
+            // should be stored at the main path
+            BEAST_EXPECT(mainPathCount == 2);
+
+            auto const historicalPathCount = std::accumulate(
```
miguelportilla, line 1262: This file needs to `#include <numeric>`

src/ripple/nodestore/impl/Shard.h
undertome: Added a second constructor

src/test/nodestore/DatabaseShard_test.cpp
ximinez: The Windows build is broken without it. I put this branch on top of the PR that fixes the Travis builds, and you can see all the Windows builds fail. https://travis-ci.com/github/ximinez/rippled/builds/177519102

undertome: Oh, good catch. Done.

src/ripple/nodestore/impl/DatabaseShardImp.cpp
```diff
@@ -1419,6 +1627,273 @@ DatabaseShardImp::removeFailedShard(std::shared_ptr<Shard> shard)
...
+                            dst / std::to_string(shardIndex));
+                    }
+                    catch (std::exception const& e)
+                    {
+                        JLOG(j_.error())
+                            << "shard " << shardIndex
+                            << " failed to move to historical storage";
+
+                        return;
+                    }
```
ximinez, line 1741: `e` is unused. Either remove it, or output `e.what()` as part of the error message.

src/ripple/nodestore/ShardStoragePaths.md
```diff
@@ -0,0 +1,123 @@
...
+When storing historical shards, if multiple historical paths are provided, the
+path to use for each shard will be selected in a random fashion. By using all
+available storage devices, we create a uniform distribution of disk utilization
+for disks of equivalent size, (provided that the disks are used only to store
+shards). In theory, selecting devices in this manner will also increase our
+chances for concurrent access to stored shards, however as of June 2020
+concurrent shard access is not implemented. Lastly, a storage path is included
+in the random distribution only if it has enough storage capacity to hold the
+next shard.
+## Shard Acquisition
```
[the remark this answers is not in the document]
undertome: Done

src/ripple/net/ShardDownloader.md
```diff
@@ -175,9 +175,9 @@ The `download_path` field of the `shard_db` entry is used to determine where to
 # https://ripple.com/build/history-sharding
 [shard_db]
 type=NuDB
-path=/var/lib/rippled/db/shards/nudb
-download_path=/var/lib/rippled/db/shards/
-max_size_gb=50
+main_storage_path=/var/lib/rippled/db/shards/nudb
+partial_shard_download_path=/var/lib/rippled/db/shards/
+max_historical_shards=50
```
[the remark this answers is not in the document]
undertome: > I don't like some of these name changes:
> 
>     1. `path` to `main_storage_path` feels redundant. A path is clearly talking about storage, and if it doesn't have a prefix, it makes sense to assume that it's the "main" path. Additionally, some operators were already using shards, and now they have to make an arbitrary change that doesn't really add any value.
> 
>     2. `download_path` to `partial_shard_download_path` isn't as eggregious, but we're already in the `[shard_db]` section, so "shard" in the name is also redundant. "partial" also feels unnecessary, though I suppose you could make an argument that it's more descriptive since fully downloaded shards aren't left there, but that feels like a really unnecessary addition since we don't have a competing `complete_download_path`.
> 
> 
> For both `path` and `download_path`, since existing configs use the old values, and these name changes don't seem to add anything that can't be added with updated and clarified docs, I think these name changes should be rolled back
> 

I don't like renaming the two path keys either. It was suggested by @scottschurr on #3927 and I disagreed with the suggestion. See [this comment thread](https://github.com/ripple/rippled/pull/3297#discussion_r402649904). I'm fine with reverting the path names but I think we should arrive at a consensus to avoid further thrashing. Pinging @miguelportilla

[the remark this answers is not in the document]
undertome: > I'm OK with switching from a size to a count, but I'm curious what the motivation for the change was. The reason it might be a problem is that it's pretty easy for an operator to look at their device like, "I have 100Gb free on this disk. I'll let shards have 60Gb of it." Whereas there's no real way to definitively say how much space 50 shards will take up, especially since they vary by size. Would it be worth adding an optional "minimum free space" field so that any download stops if the disk gets less than that amount of space?

The decision to control storage with a shard count rather than a byte count came from @miguelportilla. This was partly motivated by a desire to mimic the node store's approach of specifying ledger counts to limit db size, placing the burden of ensuring the availability of sufficient storage capacity on the user. Also note that the `DatabaseShardImp` does provide some help by issuing a warning if the available capacity is insufficient to store the full quantity of shards specified by the shard count. However, this warning is just meant to be supplementary and we expect the user to be responsible for providing ample storage for any server storing shard history.

Currently, the `DatabaseShardImp` already implements some notion of "minimum free space" by estimating the average size of a shard and minimally requiring that this much space must be available before storing an additional shard. Unfortunately this messy calculation is not ideal as shard sizes vary and are positively correlated with the shard index (later shards are typically larger in size). We have tentative plans to improve this somewhat by adding a sizeable buffer to this minimum size requirement (this change is partly inspired by the addition of `DeterministicShards`, as the capacity requirements during the finalization of shards may be double the size of the finalized shard). So in short, we're in the process of improving capacity management, but as of yet we have no plans of doing this with configuration properties.

src/ripple/nodestore/impl/DatabaseShardImp.cpp
```diff
@@ -1070,23 +1122,89 @@ DatabaseShardImp::initConfig(std::lock_guard<std::mutex>&)
...
-        return fail("'path' missing");
+    if (!get_if_exists<path>(section, "main_storage_path", dir_))
+        return fail("'main_storage_path' missing");
 
     {
-        std::uint64_t sz;
-        if (!get_if_exists<std::uint64_t>(section, "max_size_gb", sz))
-            return fail("'max_size_gb' missing");
+        std::uint64_t count = 0;
+        get_if_exists<std::uint64_t>(section, "max_historical_shards", count);
```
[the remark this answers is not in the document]
undertome: Done

src/ripple/nodestore/impl/DatabaseShardImp.cpp
```diff
@@ -1070,23 +1122,89 @@ DatabaseShardImp::initConfig(std::lock_guard<std::mutex>&)
...
 
-        // Minimum storage space required (in gigabytes)
-        if (sz < 10)
-            return fail("'max_size_gb' must be at least 10");
+            auto const dir = path(s);
+            if (dir_ == dir)
+            {
+                return fail(
+                    "a 'historical_shard_path' value == "
+                    "'main_storage_path'");
```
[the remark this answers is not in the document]
undertome: Done

src/ripple/nodestore/impl/DatabaseShardImp.cpp
```diff
@@ -1070,23 +1122,89 @@ DatabaseShardImp::initConfig(std::lock_guard<std::mutex>&)
...
+                        dir.string());
+                }
+            }
+            catch (...)
+            {
+                return fail(
+                    "failed to create 'historical_shard_path': " +
+                    dir.string());
+            }
+
```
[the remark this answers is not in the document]
undertome: Moved creation to init

src/ripple/nodestore/impl/DatabaseShardImp.cpp
```diff
@@ -1070,23 +1122,89 @@ DatabaseShardImp::initConfig(std::lock_guard<std::mutex>&)
...
+            config.section(SECTION_HISTORICAL_SHARD_PATHS);
 
-        if ((sz << 30) < sz)
-            return fail("'max_size_gb' overflow");
+        auto values = historicalShardPaths.values();
+
+        std::sort(values.begin(), values.end());
+        values.erase(std::unique(values.begin(), values.end()), values.end());
+
+        std::unordered_set<int> filesystemIDs;
```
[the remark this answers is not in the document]
undertome: I thought of that but I didn't want to include another preprocessor directive just for that.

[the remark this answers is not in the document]
undertome: Done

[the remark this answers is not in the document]
ximinez: I suppose it could go either way. I didn't notice a compiler warning about an unused variable, so maybe it optimized it away.

src/ripple/nodestore/impl/DatabaseShardImp.cpp
```diff
@@ -1245,6 +1363,47 @@ DatabaseShardImp::finalizeShard(
...
+                {
+                    // This is a historical shard that wasn't
+                    // placed at a separate historical path
+                    if (shard->getDir().parent_path() == dir_)
+                    {
+                        JLOG(j_.warn())
+                            << "shard " << shardIndex
+                            << " is not stored at a historical path";
+                    }
+                }
```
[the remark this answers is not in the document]
undertome: Done

src/ripple/nodestore/impl/DatabaseShardImp.cpp
```diff
@@ -1245,6 +1363,47 @@ DatabaseShardImp::finalizeShard(
...
+            }
+
+            // Not a historical shard
+            else
+            {
```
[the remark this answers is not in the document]
undertome: Done

src/ripple/nodestore/impl/DatabaseShardImp.cpp
```diff
@@ -1298,12 +1457,16 @@ DatabaseShardImp::setFileStats()
...
+    if (auto const count = countHistoricalShards(lock);
+        count >= maxHistoricalShards_)
     {
-        JLOG(j_.warn()) << "maximum storage size reached";
+        JLOG(j_.warn()) << "maximum number of historical shards reached";
         canAdd_ = false;
     }
-    else if (maxFileSz_ - fileSz_ > available())
+    else if (
+        maxHistoricalShards_ > count &&
```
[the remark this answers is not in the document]
undertome: Done

src/ripple/nodestore/impl/DatabaseShardImp.cpp
```diff
@@ -1350,19 +1513,58 @@ DatabaseShardImp::getCache(std::uint32_t seq)
...
+            // Leverage all the historical shard paths to
+            // see if collectively they can fit the specified
+            // number of shards. For this to work properly,
+            // each historical path must correspond to a separate
+            // physical device or filesystem.
+            numShards -=
+                std::min(capacity / avgShardFileSz_, std::uint64_t(numShards));
+
+            if (numShards == 0)
+                return true;
```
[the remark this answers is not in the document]
undertome: Done

src/ripple/nodestore/impl/DatabaseShardImp.cpp
```diff
@@ -1419,6 +1627,273 @@ DatabaseShardImp::removeFailedShard(std::shared_ptr<Shard> shard)
...
+
+        // Move a shard from the main shard path to a historical shard
+        // path by copying the contents, and creating a new shard.
+        auto const moveShard = [this,
+                                &lock](std::uint32_t const shardIndex) -> void {
+            auto const dst = chooseHistoricalPath(lock);
+
+            if (auto it = shards_.find(shardIndex); it != shards_.end())
+            {
+                if (auto& shard = it->second.shard; shard)
```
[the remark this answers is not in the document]
undertome: Done

src/ripple/nodestore/impl/DatabaseShardImp.cpp
```diff
@@ -1419,6 +1627,273 @@ DatabaseShardImp::removeFailedShard(std::shared_ptr<Shard> shard)
...
+                    // that would attempt to close the fds after the
+                    // directory has been moved.
+                    shard->closeAll();
+
+                    try
+                    {
+                        // Move the shard directory to the new path
+                        boost::filesystem::rename(
+                            shard->getDir().string(),
+                            dst / std::to_string(shardIndex));
```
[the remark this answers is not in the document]
undertome: This concern relates to a problem the `DatabaseShardImp` had wherein validating completed shards could cause the server to seize up. This was fixed by creating a background thread that would handle shard finalization. This same background thread also handles shard migration, so it shouldn't interfere with consensus call paths.

src/ripple/nodestore/impl/DatabaseShardImp.h
```diff
@@ -222,8 +227,11 @@ class DatabaseShardImp : public DatabaseShard
     // The name associated with the backend used with the shard store
     std::string backendName_;
 
-    // Maximum storage space the shard store can utilize (in bytes)
-    std::uint64_t maxFileSz_;
+    // Maximum number of historical shards to store.
+    std::uint64_t maxHistoricalShards_;
```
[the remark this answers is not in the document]
undertome: Done

src/ripple/nodestore/impl/DatabaseShardImp.h
```diff
@@ -286,9 +307,16 @@ class DatabaseShardImp : public DatabaseShard
     std::pair<std::shared_ptr<PCache>, std::shared_ptr<NCache>>
     getCache(std::uint32_t seq);
 
-    // Returns available storage space
-    std::uint64_t
-    available() const;
+    // Returns true if the filesystem has enough storage
+    // available to hold the specified number of shards.
+    // If isHistorical is set, that means that the shard(s)
```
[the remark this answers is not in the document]
undertome: Done

src/ripple/nodestore/impl/DatabaseShardImp.cpp
undertome: Done

src/ripple/nodestore/impl/DatabaseShardImp.cpp
```diff
@@ -1419,6 +1627,273 @@ DatabaseShardImp::removeFailedShard(std::shared_ptr<Shard> shard)
...
+auto
+DatabaseShardImp::expectNewShard(
+    std::uint32_t shardIndex,
+    std::uint32_t numHistoricalShards,
+    std::lock_guard<std::mutex> const& lock) -> boost::optional<PathDesignation>
+{
+    // Any shard earlier than the two most recent shards
+    // is a historical shard
+    auto const boundaryIndex = getShardBoundaryIndex(lock);
+    auto const isHistoricalShard = shardIndex < boundaryIndex;
```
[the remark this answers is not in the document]
undertome: Done

src/ripple/net/ShardDownloader.md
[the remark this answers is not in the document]
undertome: > I don't like some of these name changes:
> 
>     1. `path` to `main_storage_path` feels redundant. A path is clearly talking about storage, and if it doesn't have a prefix, it makes sense to assume that it's the "main" path. Additionally, some operators were already using shards, and now they have to make an arbitrary change that doesn't really add any value.
> 
>     2. `download_path` to `partial_shard_download_path` isn't as eggregious, but we're already in the `[shard_db]` section, so "shard" in the name is also redundant. "partial" also feels unnecessary, though I suppose you could make an argument that it's more descriptive since fully downloaded shards aren't left there, but that feels like a really unnecessary addition since we don't have a competing `complete_download_path`.
> 
> 
> For both `path` and `download_path`, since existing configs use the old values, and these name changes don't seem to add anything that can't be added with updated and clarified docs, I think these name changes should be rolled back
> 

Rolling back.

src/ripple/nodestore/impl/DatabaseShardImp.cpp
```diff
@@ -292,10 +314,19 @@ DatabaseShardImp::prepareShard(std::uint32_t shardIndex)
         return false;
     }
 
-    // Check available storage space
-    if (fileSz_ + avgShardFileSz_ > maxFileSz_)
-        return fail("maximum storage size reached");
-    if (avgShardFileSz_ > available())
+    // Any shard earlier than the two most recent shards
+    // is a historical shard
+    bool const isHistoricalShard = shardIndex < getShardBoundaryIndex(lock);
```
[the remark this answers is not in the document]
undertome: Since it's only a one liner in just two spots, I'm inclined to leave as is.

src/ripple/nodestore/impl/DeterministicShard.h
undertome: Removed.

src/ripple/rpc/impl/ShardArchiveHandler.cpp
undertome: Rolled back.

src/ripple/nodestore/impl/DatabaseShardImp.cpp
undertome: Rolled back.

ximinez:
@undertome In the future, please do not rewrite history and force push a single commit over and over. Instead, add additional commits onto your branch, starting the commit message with "[FOLD]" to indicate that these commits will need to be squashed once the PR is passed or merged. This makes my job as reviewer exponentially more manageable for several reasons, including I can confirm that you addressed all issues, I can see precisely what changed since the last time I looked at the PR, and I can focus on those changes and lower the risk of missing something that gets "buried" in the original change set. Thanks!

src/ripple/nodestore/impl/DatabaseShardImp.cpp
```diff
@@ -82,95 +86,128 @@ DatabaseShardImp::init()
...
             {
-                if (!is_directory(dir_))
+                try
+                {
+                    if (exists(path))
+                    {
+                        if (!is_directory(path))
+                        {
+                            JLOG(j_.error()) << "each path must be a "
+                                                "directory";
```
ximinez, line 105: How about
```
JLOG(j_.error()) << path << " must be a directory";
```
?

src/ripple/nodestore/impl/DatabaseShardImp.cpp
```diff
@@ -1245,6 +1355,43 @@ DatabaseShardImp::finalizeShard(
...
+                                    << " is not stored at a historical path";
+                }
+            }
+
+            else
+            {
+                // Not a historical shard. Shift recent shards
+                // if necessary
+                checkRecentShards(lock);
+                assert(shardIndex - boundaryIndex <= 1 || !boundaryIndex);
```
ximinez, line 1379: I know it's just an assert, but when using short-circuit operators, I usually like to do the "easier" check first. So `assert(!boundaryIndex || shardIndex - boundaryIndex <= 1);`

src/ripple/nodestore/impl/DatabaseShardImp.cpp
```diff
@@ -1350,19 +1501,59 @@ DatabaseShardImp::getCache(std::uint32_t seq)
...
+            capacities.push_back(boost::filesystem::space(dir_).available);
+        }
+
+        for (std::uint64_t const capacity : capacities)
+        {
+            // Leverage all the historical shard paths to
+            // see if collectively they can fit the specified
+            // number of shards. For this to work properly,
+            // each historical path must correspond to a separate
+            // physical device or filesystem.
```
ximinez, line 1540: It's not a perfect check, but how hard would it be to check for duplicates when you're populating `capacities`? If two different paths have the same number of bytes available, there's a chance that they're the same device. It's not guaranteed, so don't fail or anything, but you could log a warning message. Something like `JLOG(j_.warn()) << path1 << " and " << path2 << " both have " << capacity << " bytes free, and may be the same device. Each configured historical storage path should be on a unique device.`

You could even make it `#if ! BOOST_OS_LINUX`

src/ripple/nodestore/impl/DatabaseShardImp.cpp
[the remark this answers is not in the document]
ximinez: That's exactly what I was hoping to hear. :grin:

src/ripple/nodestore/impl/DatabaseShardImp.cpp
undertome: Done

src/ripple/nodestore/impl/DatabaseShardImp.cpp
undertome: Done

src/ripple/nodestore/impl/DatabaseShardImp.cpp
undertome: Done

undertome:
> @undertome In the future, please do not rewrite history and force push a single commit over and over. Instead, add additional commits onto your branch, starting the commit message with "[FOLD]" to indicate that these commits will need to be squashed once the PR is passed or merged. This makes my job as reviewer exponentially more manageable for several reasons, including I can confirm that you addressed all issues, I can see precisely what changed since the last time I looked at the PR, and I can focus on those changes and lower the risk of missing something that gets "buried" in the original change set. Thanks!

Sure thing!

src/ripple/nodestore/impl/DatabaseShardImp.cpp
```diff
@@ -1350,19 +1500,106 @@ DatabaseShardImp::getCache(std::uint32_t seq)
...
+
+#if !BOOST_OS_LINUX
+            // Map byte counts to each path that
+            // shares that byte count.
+            std::unordered_map<std::uint64_t, std::vector<std::string>>
+                uniqueCapacities(historicalPaths_.size());
+#endif
+
+            // Get the available storage for each historical path
+            std::for_each(
```
seelabs, line 1526: What is your criteria for using `std::for_each` vs. a range based for loop? Most of rippled uses range based for loops (and I prefer them as well). Unless there's a non-stylistic reason for this, I'd prefer we stick to range based for loops. (To be clear, there's nothing wrong with `std::for_each`, it just seems like there's a better way to do that now that we have range based for loops).

src/ripple/nodestore/impl/DatabaseShardImp.cpp
```diff
@@ -1410,6 +1647,12 @@ DatabaseShardImp::removeFailedShard(std::shared_ptr<Shard> shard)
         if (shard->index() == acquireIndex_)
             acquireIndex_ = 0;
 
+        if (shard->index() == latestShardIndex_)
+            latestShardIndex_ = boost::none;
+
+        if (shard->index() == secondLatestShardIndex_)
+            secondLatestShardIndex_ = boost::none;
+
```
seelabs, line 1655: Not in this changeset, but it looks like the `shard` parameter should be passed by `const&` and not value (unless I'm missing the reason).

src/ripple/nodestore/impl/DatabaseShardImp.cpp
```diff
@@ -1419,6 +1662,271 @@ DatabaseShardImp::removeFailedShard(std::shared_ptr<Shard> shard)
     setFileStats();
```
seelabs, line 1662: Also not in this changeset, but what is the point of the `reset` on the previous line? If there is one it should be documented, if there isn't it should be removed.

src/ripple/nodestore/impl/DatabaseShardImp.cpp
```diff
@@ -1419,6 +1662,271 @@ DatabaseShardImp::removeFailedShard(std::shared_ptr<Shard> shard)
...
+    if (auto& cur = latestShardIndex_, &prev = secondLatestShardIndex_;
+        cur || prev)
+    {
+        auto const latestShardIndex = NodeStore::seqToShardIndex(
+            app_.getLedgerMaster().getValidLedgerIndex(), ledgersPerShard_);
+
+        auto const separateHistoricalPath = !historicalPaths_.empty();
+
+        //////////////////////////////////////////////////////////////////
+        // Lambdas used inside of this method:
```
seelabs, line 1703: It's a bit of a personal preference, but I prefer not to include comments that are obvious from the code. I.e there's no need to point out that these are lambdas used in this method.

src/ripple/nodestore/impl/DatabaseShardImp.cpp
```diff
@@ -1419,6 +1662,271 @@ DatabaseShardImp::removeFailedShard(std::shared_ptr<Shard> shard)
...
+        JLOG(j_.error()) << "failed to select a historical shard path";
+        return "";
+    }
+
+    std::sample(
+        potentialPaths.begin(),
+        potentialPaths.end(),
+        &historicalShardPath,
+        1,
+        default_prng());
```
seelabs, line 1925: Instead of a random sample, why not choose the one with the most space available?

src/ripple/nodestore/ShardStoragePaths.md
```diff
@@ -0,0 +1,124 @@
...
+
+## Overview
+
+The shard database stores validated ledgers in logical groups called shards. As
+of June 2020, a shard stores 16384 ledgers by default. In order to allow users
+to store shards on multiple devices, the shard database can be configured with
+several file system paths. Each path provided should refer to a directory on a
+distinct filesystem, and no two paths should ever correspond to the same
+filesystem. Violating this restriction will cause the server to inaccurately
+estimate the amount of space available for storing shards.
```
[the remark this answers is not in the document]
undertome: Done

src/ripple/nodestore/impl/DatabaseShardImp.h
```diff
@@ -242,6 +250,19 @@ class DatabaseShardImp : public DatabaseShard
     // File name used to mark shards being imported from node store
     static constexpr auto importMarker_ = "import";
 
+    //////////////////////////////////////////////////////////////////
```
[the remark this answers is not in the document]
undertome: Done

src/ripple/nodestore/impl/DatabaseShardImp.h
```diff
@@ -297,6 +326,38 @@ class DatabaseShardImp : public DatabaseShard
 
     void
     removeFailedShard(std::shared_ptr<Shard> shard);
+
+    // Returns the index that represents the logical
+    // partition between historical and recent shards
+    std::uint32_t
+    getShardBoundaryIndex(std::lock_guard<std::mutex> const&);
```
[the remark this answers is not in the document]
undertome: Done

src/ripple/nodestore/impl/DatabaseShardImp.h
```diff
@@ -297,6 +326,38 @@ class DatabaseShardImp : public DatabaseShard
...
+
+    std::uint32_t
+    countHistoricalShards(std::lock_guard<std::mutex> const&);
```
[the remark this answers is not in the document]
undertome: Base on this and another renaming suggestion you made, it seems that the convention is to came the method after the variable returned, rather than the action its performing - is that the case?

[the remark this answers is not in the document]
undertome: Done

src/ripple/nodestore/impl/DatabaseShardImp.h
```diff
@@ -297,6 +326,38 @@ class DatabaseShardImp : public DatabaseShard
...
+
+    // Shifts the recent and second most recent (by index)
+    // shards as new shards become available on the network.
+    // Older shards are moved to a historical shard path.
+    void
+    checkRecentShards(std::lock_guard<std::mutex> const& lock);
```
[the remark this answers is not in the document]
undertome: I think that's misleading as well since the majority of calls to this function will not result in any relocation.

src/ripple/nodestore/impl/DatabaseShardImp.h
```diff
@@ -297,6 +326,38 @@ class DatabaseShardImp : public DatabaseShard
...
+
+    // Checks whether the shard can be stored. If
+    // the new shard can't be stored, returns
+    // boost::none. Otherwise returns a bool
```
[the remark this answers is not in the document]
undertome: Done

src/ripple/nodestore/impl/DatabaseShardImp.h
```diff
@@ -297,6 +326,38 @@ class DatabaseShardImp : public DatabaseShard
...
+    // indicating whether the new shard should be
+    // placed in a separate directory for historical
+    // shards.
+    boost::optional<PathDesignation>
+    expectNewShard(
```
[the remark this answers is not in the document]
undertome: I agree that's a better name.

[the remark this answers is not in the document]
undertome: Done

src/ripple/nodestore/impl/DatabaseShardImp.h
```diff
@@ -297,6 +326,38 @@ class DatabaseShardImp : public DatabaseShard
...
+        std::uint32_t shardIndex,
+        std::uint32_t numHistoricalShards,
+        std::lock_guard<std::mutex> const& lock);
+
+    boost::filesystem::path
+    chooseHistoricalPath();
```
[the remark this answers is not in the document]
undertome: Done

src/ripple/nodestore/impl/DatabaseShardImp.h
```diff
@@ -297,6 +326,38 @@ class DatabaseShardImp : public DatabaseShard
...
+
+    boost::filesystem::path
+    chooseHistoricalPath(std::lock_guard<std::mutex> const&);
```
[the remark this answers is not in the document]
undertome: Done

src/ripple/nodestore/impl/DatabaseShardImp.cpp
```diff
@@ -82,95 +86,127 @@ DatabaseShardImp::init()
...
+                    {
+                        JLOG(j_.error())
+                            << "failed to create path: " + path.string();
+                        return false;
+                    }
+                }
+                catch (...)
+                {
+                    JLOG(j_.error())
+                        << "failed to create path: " + path.string();
```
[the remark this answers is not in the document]
undertome: The try loops are a preexisting convention. Seems like a case of six of one... so I'm inclined to leave them as they are.

src/ripple/nodestore/impl/DatabaseShardImp.cpp
```diff
@@ -82,95 +86,127 @@ DatabaseShardImp::init()
...
-                // Check shard directory name is numeric
-                auto dirName = d.path().stem().string();
-                if (!std::all_of(dirName.begin(), dirName.end(), [](auto c) {
-                        return ::isdigit(static_cast<unsigned char>(c));
-                    }))
+                for (auto const& d : directory_iterator(path))
                 {
-                    continue;
-                }
+                    if (!is_directory(d))
```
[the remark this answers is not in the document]
undertome: This is already protected by a `try` and `catch` block. However, I'm wondering if the catch statement should return false. I didn't write that part so I will have to look into it.

src/ripple/nodestore/impl/DatabaseShardImp.cpp
undertome: Done

src/ripple/nodestore/impl/DatabaseShardImp.cpp
undertome: It should be a reference.

undertome: Done

src/ripple/nodestore/impl/DatabaseShardImp.cpp
undertome: This causes the shard to be removed from the `DatabaseShardImp`. This is why it's important for the shard to be passed by reference.

undertome: Added a comment.

src/ripple/nodestore/impl/DatabaseShardImp.cpp
undertome: It's not necessary - see [this comment thread](https://github.com/ripple/rippled/pull/3463#discussion_r447239944).

src/ripple/nodestore/impl/DatabaseShardImp.cpp
undertome: Done

src/ripple/nodestore/impl/DatabaseShardImp.cpp
```diff
@@ -1074,19 +1143,59 @@ DatabaseShardImp::initConfig(std::lock_guard<std::mutex>&)
...
 
-        // Minimum storage space required (in gigabytes)
-        if (sz < 10)
-            return fail("'max_size_gb' must be at least 10");
+        std::sort(values.begin(), values.end());
+        values.erase(std::unique(values.begin(), values.end()), values.end());
 
-        // Convert to bytes
-        maxFileSz_ = sz << 30;
+#if BOOST_OS_LINUX
```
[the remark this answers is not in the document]
undertome: Done

src/ripple/nodestore/impl/DatabaseShardImp.cpp
```diff
@@ -1350,19 +1500,106 @@ DatabaseShardImp::getCache(std::uint32_t seq)
...
+            !historicalPaths_.empty())
+        {
+            capacities.reserve(historicalPaths_.size());
+
+#if !BOOST_OS_LINUX
+            // Map byte counts to each path that
+            // shares that byte count.
+            std::unordered_map<std::uint64_t, std::vector<std::string>>
+                uniqueCapacities(historicalPaths_.size());
+#endif
```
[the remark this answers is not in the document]
undertome: Done

src/ripple/nodestore/impl/DatabaseShardImp.cpp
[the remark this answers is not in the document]
undertome: I added a `return false` statement to the catch statement

src/ripple/nodestore/impl/DatabaseShardImp.h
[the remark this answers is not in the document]
undertome: Changed to `relocateOutdatedShards`

src/ripple/net/ShardDownloader.md
```diff
@@ -177,7 +177,7 @@ The `download_path` field of the `shard_db` entry is used to determine where to
 type=NuDB
 path=/var/lib/rippled/db/shards/nudb
 download_path=/var/lib/rippled/db/shards/
-max_size_gb=50
+max_historical_shards=50
```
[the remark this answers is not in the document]
undertome: Yes, the `max_size_gb` key is being removed. The decision came from @miguelportilla. This is a preexisting convention BTW, since the node store limits are determined by ledger counts and not drive capacity.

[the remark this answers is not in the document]
seelabs: I defer to you and @miguelportilla on this. It's probably OK to break backward compatibility as few people are using shards right now.

src/ripple/nodestore/impl/DatabaseShardImp.h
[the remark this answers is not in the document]
seelabs: Yep, that's a better name.

src/ripple/nodestore/impl/DatabaseShardImp.cpp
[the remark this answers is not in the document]
seelabs: I'm OK leaving them.

src/ripple/nodestore/impl/DatabaseShardImp.cpp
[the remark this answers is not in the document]
seelabs: That's great, thanks. One minor change I'd still like: instead of one function with the preprossor directives inside the function, can we change it so there are two separate functions and the preprossor directives control which function is compiled in?

src/ripple/nodestore/impl/DatabaseShardImp.cpp
```diff
@@ -1402,23 +1573,363 @@ DatabaseShardImp::storeLedgerInShard(
...
+    return std::count_if(
+        shards_.begin(),
+        shards_.end(),
+        [recentShardBoundaryIndex](auto const& entry) {
+            return entry.first < recentShardBoundaryIndex;
+        });
+}
+
+void
+DatabaseShardImp::relocateOutdatedShards(std::lock_guard<std::mutex> const& lock)
```
seelabs, line 1630: clang format will reformat this line

src/ripple/nodestore/impl/DatabaseShardImp.cpp
```diff
@@ -292,10 +335,19 @@ DatabaseShardImp::prepareShard(std::uint32_t shardIndex)
...
     }
 
-    // Check available storage space
-    if (fileSz_ + avgShardFileSz_ > maxFileSz_)
-        return fail("maximum storage size reached");
-    if (avgShardFileSz_ > available())
+    // Any shard earlier than the two most recent shards
+    // is a historical shard
+    bool const isHistoricalShard = shardIndex < shardBoundaryIndex(lock);
+    auto const numHistoricalShards = this->numHistoricalShards(lock);
```
seelabs, line 341: Now that we've renamed the function, we should probably rename this variables. Maybe `numHistShards`.

src/ripple/nodestore/impl/DatabaseShardImp.cpp
```diff
@@ -688,23 +755,21 @@ DatabaseShardImp::import(Database& source)
             }
         }
 
+        auto numHistoricalShards = this->numHistoricalShards(lock);
```
seelabs, line 758: Consider renaming the var.

src/ripple/nodestore/impl/DatabaseShardImp.cpp
undertome: Done

src/ripple/nodestore/impl/DatabaseShardImp.cpp
undertome: Done

src/ripple/nodestore/impl/DatabaseShardImp.cpp
undertome: Done

src/ripple/nodestore/impl/DatabaseShardImp.h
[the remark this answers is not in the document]
ximinez: The new `relocateOutdatedShards` name is better, since it makes it clear _which_ shards will be relocated, but just FYI another convention I've seen is "maybe", so `maybeRelocateShards`.

src/ripple/nodestore/impl/DatabaseShardImp.cpp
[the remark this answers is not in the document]
undertome: I missed this comment. What's the benefit vs having the directives inside the function?

src/ripple/nodestore/impl/DatabaseShardImp.h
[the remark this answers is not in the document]
undertome: That's funny I actually considered "maybe" but wasn't sure if that would be accepted.

src/ripple/nodestore/impl/DatabaseShardImp.cpp
```diff
@@ -1402,23 +1573,362 @@ DatabaseShardImp::storeLedgerInShard(
...
+            JLOG(j_.error())
+                << "failed to acquire stats for a 'historical_shard_path'";
+            return false;
+        }
+
+        if (!filesystemIDs.insert(buffer.f_fsid).second)
+        {
+            JLOG(j_.error())
+                << "each 'historical_shard_path' must correspond to a "
+                   "distinct device or filesystem";
```
ximinez, line 1882: I think it would be worthwhile to output the duplicate path here, too. But now that I look at it, it would be even better to output both paths that are on the same device so the user can pick one. I think the best way to implement this would be to use a `map` similar to how you checked below for non-linux filesystems.

src/ripple/nodestore/impl/DatabaseShardImp.cpp
```diff
@@ -1402,23 +1573,362 @@ DatabaseShardImp::storeLedgerInShard(
...
+        uniqueCapacities(historicalPaths_.size());
+
+    for (auto const& path : historicalPaths_)
+    {
+        auto const availableSpace = boost::filesystem::space(path).available;
+
+        if (uniqueCapacities.find(availableSpace) != uniqueCapacities.end())
+            uniqueCapacities[availableSpace].push_back(path.string());
+        else
+            uniqueCapacities[availableSpace] = {path.string()};
```
ximinez, line 1906: You can do this in one line because the `[]` operator creates a default value (i.e. an empty vector) as soon as you invoke it.
```
uniqueCapacities[boost::filesystem::space(path).available].push_back(path.string());
```

src/ripple/nodestore/impl/DatabaseShardImp.h
```diff
@@ -242,6 +250,16 @@ class DatabaseShardImp : public DatabaseShard
...
 
+    // latestShardIndex_ and secondLatestShardIndex hold the indexes
+    // of the shards most recently confirmed by the network. These
+    // values are not updated in real time and are modified only
+    // when adding shards to the database, in order to determine where
+    // pending shards will be stored on the filesystem. A value of
+    // boost::none indicates that the corresponding shard is not held
+    // by the database.
+    boost::optional<std::uint32_t> latestShardIndex_;
+    boost::optional<std::uint32_t> secondLatestShardIndex_;
```
ximinez, line 261: Instead of `latest` and `secondLatest`, what do you think about `last` and `penultimate`? I think it's fine and clear the way it is, but "penultimate" means "next to last", and it's a little shorter, sooooooooo......

src/ripple/nodestore/impl/DatabaseShardImp.cpp
```diff
@@ -1402,23 +1573,362 @@ DatabaseShardImp::storeLedgerInShard(
...
+    // Linux.
+
+    std::unordered_set<int> filesystemIDs;
+    for (auto const& path : historicalPaths_)
+    {
+        struct statvfs buffer;
+        if (statvfs(path.c_str(), &buffer))
+        {
+            JLOG(j_.error())
+                << "failed to acquire stats for a 'historical_shard_path'";
```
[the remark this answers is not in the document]
undertome: Done

src/ripple/nodestore/impl/DatabaseShardImp.cpp
undertome: Done

src/ripple/nodestore/impl/DatabaseShardImp.cpp
undertome: Done

src/ripple/nodestore/impl/DatabaseShardImp.h
undertome: Think I'll opt to keep the names as they are.

src/ripple/nodestore/impl/DatabaseShardImp.cpp
```diff
@@ -1864,25 +1864,39 @@ DatabaseShardImp::checkHistoricalPaths() const
...
+            // Two or more historical storage paths
+            // correspond to the same filesystem.
             JLOG(j_.error())
-                << "each 'historical_shard_path' must correspond to a "
-                   "distinct device or filesystem";
-            return false;
+                << "The following paths correspond to the same filesystem: "
+                << boost::algorithm::join(entry.second, ", ")
+                << ". Each configured historical storage path should"
+                   " be on a unique device or filesystem.";
```
ximinez, line 1896: Need to `return false;`

src/ripple/nodestore/impl/DatabaseShardImp.h
ximinez: > Think I'll opt to keep the names as they are.

I'm fine with that.

src/ripple/nodestore/impl/DatabaseShardImp.cpp
undertome: Done

undertome:
> Looks like there's an issue with the VS build:
> 
> ```
> (WarnCompileDuplicatedFilename target) -> 
> 
>          C:\Program Files (x86)\Microsoft Visual Studio\2017\BuildTools\Common7\IDE\VC\VCTargets\Microsoft.CppBuild.targets(944,5): warning MSB8027: Two or more files with the name of Manifest.cpp will produce outputs to the same location. This can lead to an incorrect build result.  The files involved are C:\Users\travis\build\ripple\rippled\src\ripple\app\misc\impl\Manifest.cpp, C:\Users\travis\build\ripple\rippled\src\ripple\rpc\handlers\Manifest.cpp. [C:\Users\travis\build\ripple\rippled\build.ms\rippled.vcxproj]
> ```
> 
> We've run into similar issues in the past and have solved it by renaming one of the files.

Let's address this in a separate PR since this one didn't change those files.

ximinez:
@seelabs I think that error is a non-fatal warning, and the actual CI failure was due to Travis being Travis. When I rebuilt that step, it succeeded. https://travis-ci.com/github/ripple/rippled/jobs/375010094

PR #3465, opened 2020-06-29
Travis CI Windows fix

carlhua:
This fix allows Travis CI Windows to pass. Previously, `wget` on Windows instance cannot verify some of the certificates and therefore causing the download of Boost to fail. This workaround bypasses the checks for cert. This is not a security issue because it only bypasses on windows CI and not the actual build process.

---
This change is [Reviewable](https://reviewable.io/reviews/ripple/rippled/3465)


carlhua:
Instead of merging in a workaround - lets wait for Travis to fix their instance. Closing PR.

PR #3487, opened 2020-07-06
Allow a SLE to be initialized on creation

nbougalis:
Not for merging. Just soliciting comments on this for now.

### DIFF
--- a/src/ripple/app/ledger/Ledger.cpp
+++ b/src/ripple/app/ledger/Ledger.cpp
@@ -187,19 +187,20 @@ Ledger::Ledger(
     static auto const id = calcAccountID(
         generateKeyPair(KeyType::secp256k1, generateSeed("masterpassphrase"))
             .first);
-    {
-        auto const sle = std::make_shared<SLE>(keylet::account(id));
-        sle->setFieldU32(sfSequence, 1);
-        sle->setAccountID(sfAccount, id);
-        sle->setFieldAmount(sfBalance, info_.drops);
-        rawInsert(sle);
-    }
+
+    rawInsert(std::make_shared<SLE>(
+        keylet::account(id), [&, this](SLE& sle) {
+            sle.setFieldU32(sfSequence, 1);
+            sle.setAccountID(sfAccount, id);
+            sle.setFieldAmount(sfBalance, info_.drops);
+        }));
 
     if (!amendments.empty())
     {
-        auto const sle = std::make_shared<SLE>(keylet::amendments());
-        sle->setFieldV256(sfAmendments, STVector256{amendments});
-        rawInsert(sle);
+        rawInsert(std::make_shared<SLE>(
+            keylet::amendments(), [&amendments](SLE& sle) {
+                sle.setFieldV256(sfAmendments, STVector256{amendments});
+            }));
     }
 
     stateMap_->flushDirty(hotACCOUNT_NODE, info_.seq);
--- a/src/ripple/app/main/DBInit.h
+++ b/src/ripple/app/main/DBInit.h
@@ -21,6 +21,7 @@
 #define RIPPLE_APP_DATA_DBINIT_H_INCLUDED
 
 #include <array>
+#include <cstdint>
 
 namespace ripple {
 
--- a/src/ripple/app/misc/NetworkOPs.cpp
+++ b/src/ripple/app/misc/NetworkOPs.cpp
@@ -3745,8 +3745,7 @@ NetworkOPsImp::getBookPage(
 
     while (iLimit-- > 0 && obIterator.nextOffer())
     {
-        SLE::pointer sleOffer = obIterator.getCurrentOffer();
-        if (sleOffer)
+        if (auto sleOffer = obIterator.getCurrentOffer())
         {
             auto const uOfferOwnerID = sleOffer->getAccountID(sfAccount);
             auto const& saTakerGets = sleOffer->getFieldAmount(sfTakerGets);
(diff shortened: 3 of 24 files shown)


src/ripple/app/ledger/Ledger.cpp
```diff
@@ -187,19 +187,20 @@ Ledger::Ledger(
...
-        sle->setFieldAmount(sfBalance, info_.drops);
-        rawInsert(sle);
-    }
+
+    rawInsert(std::make_shared<SLE>(
+        keylet::account(id), [&, this](SLE& sle) {
+            sle.setFieldU32(sfSequence, 1);
+            sle.setAccountID(sfAccount, id);
+            sle.setFieldAmount(sfBalance, info_.drops);
+        }));
```
seelabs, line 196: I like the idea of completely initializing a SLE like this. The lambda constructor trick isn't bad, but it's also a pretty big hammer. Just brainstorming...
My preferred syntax would be something like:
```
    rawInsert(std::make_shared<SLE>(
        keylet::account(id),
        {sfSequence, 1},
        {sfAccount, id},
        {sfBalance, info_.drops}));
```
I couldn't quite make that work, tho I might be able to get closer with deduction guides and a dedicated key/value struct. I _can_ get something like this to work
```
    rawInsert(std::make_shared<SLE>(
        keylet::account(id),
        sfSequence,
        1,
        sfAccount,
        id,
        sfBalance,
        info_.drops));
```
I wish that formatted better, but that's not too bad. Here's a WIP patch: https://github.com/seelabs/rippled/commit/148d5dc35d686c19b049f2263fcab3dbda5789c9
I don't claim the WIP patch is better than the lambda, but if I could modify that to get to my preferred syntax I'd like it better than a lambda (well, depending on how ugly the implementation turned out).

nbougalis: I actually like the idea of restricting it to passing a bunch of `SField`s.

seelabs: Here's another cut that uses a `SLEKV` (sle key/value) struct to initialize the fields. The new syntax would look like this:
```
    rawInsert(std::make_shared<SLE>(
        keylet::account(id),
        SLEKV{sfSequence, 1},
        SLEKV{sfAccount, id},
        SLEKV{sfBalance, info_.drops}));
```
Patch: https://github.com/seelabs/rippled/commit/c303c655ba5271fc14da33f28233defed20fb555 (uses the other WIP branch as a parent).

`SLEKV` is a placeholder. If we did use this I'd want to brainstorm another name. BTW, I still think the lambda solution is very much on the table.

cjcobb23: I like the solution using a `SLEKV` struct

nbougalis: I tried to rig something similar:

```
    template<class T>
    void foo(std::pair<SField, T> const& x)
    {
        (*this)[x.first] = x.second;
    }

    template<class ...Value>
    STLedgerEntry(int x, Keylet const& k, std::pair<SField, Value> ...fields)
        : STLedgerEntry(k)
    {
        foo(&fields...);
    }
```

I wanted to call it like this:

```
auto test = std::make_shared<SLE>(keylet::account(id),
    { sfSequence, 1 },
    { sfAccount, id },
    { sfBalance, info_.drops });
```

But `std::make_shared` doesn't like this and I do understand why.

scottschurr: I haven't run any experiments.  Just pondering out loud.  I wonder if replacing the `make_shared` with a custom variadic factory function that returns a `shared_ptr<SLE>` might be a way to work around this...

seelabs: I don't think C++'s template argument inference is smart enough to do what we want here. Replacing `make_shared` isn't going to help. I'm pretty sure we need a named type like `SLEVK` as a work around.

scottschurr: Yup.  I poked at this a little harder and I concede that @seelabs is right.  There's no getting around the named type to make the pair-like things.  To my eyes that looks like the most readable working approach.

nbougalis: I have a variant that works but don't "wrap" the field/value pair, so it may be a little difficult to read:

```
auto test = std::make_shared<SLE>(
    keylet::account(id),
    sfSequence,
    1,
    sfAccount,
    id,
    sfBalance,
    info_.drops);
```

I _think_ this is harder to read and maybe more error-prone. The ideal would be to have `{ field, value }` style syntax which makes the intent clear as a bell. I'll poke around this some more.

cjcobb23: I also think that this is harder to read and more error prone; I think it's pretty clunky to pass what are logically pairs (field, value) as individual parameters to a constructor, especially when there can be many such pairs. This approach also makes it difficult to put the pairs in any sort of local/temporary variables prior to construction, since you can't actually pass the pairs to the constructor. I think using a named struct is preferable. While I agree that something like
```    
    rawInsert(std::make_shared<SLE>(
        keylet::account(id),
        {sfSequence, 1},
        {sfAccount, id},
        {sfBalance, info_.drops}));
```
is the ideal solution here, using a named struct is really the same thing with just a little bit more typing:
```
    rawInsert(std::make_shared<SLE>(
        keylet::account(id),
        SLEKV{sfSequence, 1},
        SLEKV{sfAccount, id},
        SLEKV{sfBalance, info_.drops}));
```

carlhua:
@nbougalis do you intend to receive more comments? Can we close this PR?

PR #3488, opened 2020-07-06
Added public_key to ValidatorList::getAvailable response

### LINKED ISSUE #3392: vl method should return public key (Version: 1.5.0)
## Summary
The recently-added `/vl/{pubkey}` method (#3072) should include a `public_key` field in the response to be consistent with vl.ripple.com.

## Motivation
It would be nice if the vl method were exactly comparable to the format served from vl.ripple.com, for two reasons:

- Consistency makes things easier to understand.
- If someone needed to publish their own validator list, they could use a rippled server with a valid TLS key, or pool thereof, to publish the validator list. This reduces the amount of unique infrastructure strictly necessary to run the network.

## Solution
When you use [the /vl/ method of a rippled server](https://s1.ripple.com:51235/vl/ED2677ABFFD1B33AC6FBC3062B71F1E8397C1505E1C42C64D11AD1B28FF73F4734), the resulting JSON object should have the exact same fields as [vl.ripple.com](https://vl.ripple.com/). The only difference, currently, is the `public_key` field, which is known from the request URL but not copied back in rippled's response.

natenichols:
Public key from the `/vl/` request URL is now copied back in rippled's response.

### Context of Change
API change Fixes #3392, vl method should return public key (Version: 1.5.0). The response from rippled at the endpoint `/vl/<public_key>` now includes the public_key.

### Type of Change
- [ ] Documentation Updates
- [ ] Release

## Before / After
Request: `https://localhost:51235/vl/ED2677ABFFD1B33AC6FBC3062B71F1E8397C1505E1C42C64D11AD1B28FF73F4734`
Response:
**Before**
`{
    "blob": "eyJzZ...J9XX0=",
    "manifest": "JAAAA....sAg==",
    "signature": "4C59F...883E03",
    "version": 1
}`
**After**
`{
    "blob": "eyJzZ...J9XX0=",
    "manifest": "JAAAA....sAg==",
    "public_key": "ED2677ABFFD1B33AC6FBC3062B71F1E8397C1505E1C42C64D11AD1B28FF73F4734",
    "signature": "4C59F...883E03",
    "version": 1
}`

## Test Plan
None currently, and am open to suggestions for how to go about writing tests.

### DIFF
--- a/src/ripple/app/misc/impl/ValidatorList.cpp
+++ b/src/ripple/app/misc/impl/ValidatorList.cpp
@@ -820,6 +820,7 @@ ValidatorList::getAvailable(boost::beast::string_view const& pubKey)
 
     Json::Value value(Json::objectValue);
 
+    value["public_key"] = std::string{pubKey};
     value["manifest"] = iter->second.rawManifest;
     value["blob"] = iter->second.rawBlob;
     value["signature"] = iter->second.rawSignature;
--- a/src/ripple/app/misc/impl/ValidatorList.cpp
+++ b/src/ripple/app/misc/impl/ValidatorList.cpp
@@ -820,11 +820,11 @@ ValidatorList::getAvailable(boost::beast::string_view const& pubKey)
 
     Json::Value value(Json::objectValue);
 
-    value["public_key"] = std::string{pubKey};
-    value["manifest"] = iter->second.rawManifest;
-    value["blob"] = iter->second.rawBlob;
-    value["signature"] = iter->second.rawSignature;
-    value["version"] = iter->second.rawVersion;
+    value[jss::public_key] = std::string{pubKey};
+    value[jss::manifest] = iter->second.rawManifest;
+    value[jss::blob] = iter->second.rawBlob;
+    value[jss::signature] = iter->second.rawSignature;
+    value[jss::version] = iter->second.rawVersion;
 
     return value;
 }
--- a/src/test/app/ValidatorList_test.cpp
+++ b/src/test/app/ValidatorList_test.cpp
@@ -570,6 +570,25 @@ private:
             BEAST_EXPECT(trustedKeys->listed(val.signingPublic));
         }
 
+        const auto hexPublic =
+            strHex(publisherPublic.begin(), publisherPublic.end());
+
+        const auto available = trustedKeys->getAvailable(hexPublic);
+
+        BEAST_EXPECT(available);
+        if (available)
+        {
+            BEAST_EXPECT(
+                available->get(jss::public_key, Json::nullValue) == hexPublic);
+            BEAST_EXPECT(available->get(jss::blob, Json::nullValue) == blob2);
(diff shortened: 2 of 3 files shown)


src/ripple/app/misc/impl/ValidatorList.cpp
```diff
@@ -820,6 +820,7 @@ ValidatorList::getAvailable(boost::beast::string_view const& pubKey)
 
     Json::Value value(Json::objectValue);
 
+    value["public_key"] = std::string{pubKey};
```
scottschurr, line 823: This change looks great, as far as it goes.  However the area that this change is in has some bugs.  Rather than let the bugs spread I'd like you to squash these bugs while you are in the vicinity.

Our interfaces are set up so that programmers can use literal strings to supply parameters to our interfaces.  But literal strings are error prone and subject to things like typos.  So we have places that we like to get common user interface strings from.  This helps to assure that the always spell (and capitalize) strings in our interfaces the same way.

So, in place of `"public_key"` I'd like you to use `jss::public_key`.

That same sort of change needs to also be made for the four other near-by quoted strings.  So the final result will look something like...
```
    value[jss::public_key] = std::string{pubKey};
    value[jss::manifest] = iter->second.rawManifest;
    value[jss::blob] = iter->second.rawBlob;
    value[jss::signature] = iter->second.rawSignature;
    value[jss::version] = iter->second.rawVersion;
```
Once you've done this you'll get a compile error.  It turns out that a `jss` string (jss stands for JSON Static String) has not yet been defined for "blob".  So you'll need to add that.  To do so navigate to here:

https://github.com/ripple/rippled/blob/develop/src/ripple/protocol/jss.h#L146

This file contains all of the `jss` strings in alphabetical order.  You'll add one for "blob" right above the line that says `JSS(books);`. The final result should look something like this:
```
JSS(blob);                   // out: ValidatorList
JSS(books);                  // in: Subscribe, Unsubscribe
```
Then, assuming everything builds and passes unit tests, I think we're good.

natenichols: I've changed the strings to `jss` and added "blob" to `jss.h`. I'll talk to @ximinez about adding `Validator_List` unit tests for this PR. 

Thanks Scott!

src/test/app/ValidatorList_test.cpp
```diff
@@ -570,6 +570,25 @@ class ValidatorList_test : public beast::unit_test::suite
...
 
+        const auto hexPublic =
+            strHex(publisherPublic.begin(), publisherPublic.end());
+
+        const auto available = trustedKeys->getAvailable(hexPublic);
+
+        BEAST_EXPECT(available);
+        if (available)
+        {
+            BEAST_EXPECT(
```
ximinez, line 581: You're missing an include for the non-unity builds, so `jss` is unknown. You need `#include <ripple/protocol/jss.h>`

PR #3518, opened 2020-07-09
SQL interface

p2peer:
# SQL Interface

The following changes were made to the rippled code:

1) all SQL hard code (except of sqlite-specific tests) moved to the single place;
2) provided pure interface class which can have several implementations for different SQL database types;
3) done substitution of databases for the future use: if the node database is absent then shard databases used instead.

## Configuration

The following changes to rippled configuration file is proposed. New parameter `sql_backend` will be added to  `[node_db]` and  `[shard_db]` sections. The value of this parameter is the name of SQL implementation used for node or shard databases. At the present, the only valid value of this parameter is `sqlite`. 

## Files

These source files are new or moved from other location:

- `ripple/core/SQLInterface.h` - new file, contains definition of main pure classes of the interface: `SQLInterface` and `SQLDatabase`;
- `ripple/core/impl/SQLInterface.cpp` - new file, implementation of static methods of the class `SQLInterface`;
- `ripple/core/sql_backend/SQLInterface_sqlite.cpp` - new file, implementation of `SQLInterface` and `SQLDatabase` classes for the case of sqlite databases;
- `ripple/core/sql_backend/DatabaseCon.h` - moved from `ripple/core/DatabaseCon.h` because this file is specific to sqlite implementation;
- `ripple/core/sql_backend/DatabaseCon.cpp` - moved from `ripple/core/impl/DatabaseCon.h` because this file is specific to sqlite implementation;
- `ripple/core/sql_backend/SociDB.h` - moved from `ripple/core/SociDB.h` because this file is specific to sqlite implementation;
- `ripple/core/sql_backend/SociDB.cpp`- moved from `ripple/core/impl/SociDB.cpp` because this file is specific to sqlite implementation.

## Classes

The main class of the interface is `class SQLInterface` defined in the file `ripple/core/SQLInterface.h`. This class has both static methods aplied to all SQL implementations and non-static methods which should be provided by each implementation. Static methods are implemented in the file `ripple/core/impl/SQLInterface.cpp`. At present there is one implementation of the interface - `class SQLInterface_sqlite` for sqlite database. It is located in the file `ripple/core/sql_backend/SQLInterface_sqlite.cpp`.

Methods of the SQL interface deal with SQL databases by using the class `class SQLDatabase`. This class represents pointer to single SQL database of unspecified internal format. Class `SQLDatabase` is implemented as `std::unique_ptr` to abstract class  `SQLDatabase_`. The class `SQLDatabase_` represents single SQL database of unspecified internal format. There is one implementation of class `SQLDatabase_` - class `SQLDatabase_sqlite`  located in the file `ripple/core/sql_backend/SQLInterface_sqlite.cpp`. 

Class `SQLDatabase_sqlite` is designed as `std::variant` with alternative types `DatabaseCon`,  `soci::session` and `bool`. `DatabaseCon` is used in most cases as basic implementation of sqlite database. In rare cases `soci::session` is used directly instead of `DatabaseCon`. For the future implementation of absent node database the class `SQLDatabase_sqlite` for such an absent database should still be created using the `bool` alternative type.

## Methods 

There are 3 types of non-static methods os SQLInterface:

1) Methods for creation/initialization of the database. Methods for the creation of database create one or several `SQLDatabase` objects and return it (note that  `SQLDatabase` is `unique_ptr` to the actual `SQLDatabase_` class). Methods for database initialization obtain link to `SQLDatabase` as  parameter, create `SQLDatabase_` object and change the pointer in given `SQLDatabase` to created object.

2) Methods for usage of the database. Generally such a method obtains one or several existing databases as parameters. Specifically parameters are of the type `SQLDatabase &`.

3) Internal methods for the case of absent node database. Generally, such a method is similar to the corresponding method of type 2). It has the same name as the corresponding method and almost the same list of parameters. The exception is that the method of type 3) has  `SQLDatabase_ *` link type to the database instead of `SQLDatabase &`. 

These internal methods of type 3) are used as follows. If the external method of type 2) is called for node database and this database is present, then the corresponding method of class 3) is called for the same database to complete the operation. If the node database is absent, then the static internal maps of the class `SQLInterface` are used to determine which shard databases corresponds to this node database. 

Finally, the corresponding internal methods of type 3) are called for found shard databases to complete the operation. Note that shard databases may have another SQL implementation in comparison with node database. That's why methods of the type 3) are called for shard databases using the pointer to appropriate `SQLInterface` implementation, the pointer is stored in each  `SQLDatabase`

## Methods lists

Here is the list of methods of type 1. (for puprose of demonstration, not as a documentation):

```
makeLedgerDBs
makeAcquireDB
makeWalletDB
makeArchiveDB
initStateDB
openDatabaseBodyDb
makeVacuumDB
initPeerFinderDB
updatePeerFinderDB
```

Here is the list of the methods which have both type-2) and type-3) implementations:

```
getMinLedgerSeq
getMaxLedgerSeq
deleteByLedgerSeq
deleteBeforeLedgerSeq
getRows
getRowsMinMax
saveValidatedLedger
loadLedgerInfoByIndex
loadLedgerInfoByIndexSorted
loadLedgerInfoByIndexLimitedSorted
loadLedgerInfoByHash
getHashByIndex
getHashesByIndex
loadTxHistory
getAccountTxs
getAccountTxsB
accountTxPage
loadTransaction
```

Here is the list of the methods which have only type-2) implementation. The reason that this functions have no need in type-3) implementation is as follows: these functions are not called for node databases.

```
insertAcquireDBIndex
selectAcquireDBLedgerSeqs
selectAcquireDBLedgerSeqsHash
updateLedgerDBs
updateAcquireDB
loadManifest
saveManifest
checkDBSpace
loadNodeIdentity
databaseBodyDoPut
databaseBodyFinish
addValidatorManifest
loadPeerReservationTable
insertPeerReservation
deletePeerReservation
readArchiveDB
insertArchiveDB
deleteFromArchiveDB
getCanDelete
setCanDelete
getSavedState
setSavedState
setLastRotated
dropArchiveDB
getKBUsedAll
getKBUsedDB
readPeerFinderD
savePeerFinderDB
```


mtrippled:
>> 2b) A DatabaseCon is really just a soci::session wrapped in a lock. However, SQLDatabase_sqlite never uses this locking functionality (at least that I could see). Could we use a soci::session instead?

SQLite doesn't support concurrent anything with a writer. I'm not sure if soci's session is aware of this.

> If the motivation is "to eventually rid of soci", then I would make these more substantial changes when we decide to do that. If the motivation is to support postgres, then I would use soci to support postgres. If the motivation is something I'm missing please let me know.

The postgres implementation for the RPC refactoring project is not using SOCI, but instead the C libpq interface directly from the postgres project, mainly for flexibility. The postgres is also different from that used for SQLite. Also concurrency is done server-side with postgres.

We had a call a few weeks ago about SQLite, Postgres, and Sharding, and decided at the time it was simpler for all concerned to implement SQLite for this particular project for sharding and not worry about postgres until both of these projects are at a point where it makes sense to merge them.

mtrippled:
> > > 2b) A DatabaseCon is really just a soci::session wrapped in a lock. However, SQLDatabase_sqlite never uses this locking functionality (at least that I could see). Could we use a soci::session instead?
> 
> SQLite doesn't support concurrent anything with a writer. I'm not sure if soci's session is aware of this.
> 

Actually, it looks like some circumstances allow concurrent writes with SQLite. However, our current implementation has an exclusive lock around all SQLite accesses. Any changes to that model need to be considered carefully in light of how SQLite manages concurrency.

p2peer:
>     1. There is the class hierarchy SQLDatabase -> SQLDatabase_sqlite. This is done to support other database backends. However, soci already supports other database backends. I'd expect that the SQL statements can be written so they are database agnostic (sqlite even supports upserts now). There may be some statements that are different (pragmas in particular), but these minor differences can be handled by switch on some `backendType` enum.

I think that Mark is already answered this question. As far as I understand the answer is that there is planned some other approach to interface with SQL databases not based on soci.

>     2. The variant in `SQLDatabase_sqlite` supports `bool`, `soci::session`,  and `DatabaseCon`.
>        2a) What is the `bool` parameter for?

As far as I know it is planned to remove the node database at all and remain only shard databases. In this case we need in some fictive database object which we can pass as the first parameter to methods of SQL Interface instead of real node database object. To construct such an fictive database objects a constructor with bool type should be used.
 
> 2b) A `DatabaseCon` is really just a `soci::session` wrapped in a lock. However, `SQLDatabase_sqlite` never uses this locking functionality (at least that I could see). Could we use a `soci::session` instead?

Locking of DatabaseCon is actually used in many methods. The function static LockedSociSession checkoutDb(SQLDatabase& db) is intended for locking, and this function used by many methods in SQlInterface_sqlite.

cjcobb23:
>     1. There is the class hierarchy SQLDatabase -> SQLDatabase_sqlite. This is done to support other database backends. However, soci already supports other database backends. I'd expect that the SQL statements can be written so they are database agnostic (sqlite even supports upserts now). There may be some statements that are different (pragmas in particular), but these minor differences can be handled by switch on some `backendType` enum.

I would prefer a higher level abstraction than simply using soci with different database backends. I think the aim of this project is to create a layer of abstraction similar to the abstraction layer we use for the different nodestore backends. Simply wrapping soci limits the possibilities for what can be used in place of SQLite. For instance, I may want to use an in memory datastructure in place of SQLite, or even a non-relational database. Or maybe soci will work with my choice of backend, but a different library provides a much better API than soci. Using soci as the abstraction layer would make it difficult to do such things, and would make the abstraction layer very weak.

cjcobb23:
I will do a thorough review of this code by the end of the week.

seelabs:
Here's how I would approach this:

1) Add free functions that look a lot like the public interface to interface class. I.e. replace `SQLInterface` member functions like:

```
    virtual void
    insertAcquireDBIndex(SQLDatabase& db, std::uint32_t index_) override;
```
with free functions like:
```
    void
    insertAcquireDBIndex(soci::session& session, std::uint32_t index);
```

That approach should handle all the cases we have now, and is very easy to understand and navigate. We may want to put these free functions in their own namespace.

2) When a new non-sql database is introduced, add overloads for each high level function that it supports. I.e., add a free function like:
```
    void
    insertAcquireDBIndex(shardDBType& shardSession, std::uint32_t index);
```
Add a variant type: `using DBBackend = std::variant<soci::session, shardDBType>`. Add an overload that dispatches the the correct overload:
```
    void
    insertAcquireDBIndex(DBBackend& shardSession, std::uint32_t index){std::visit(...);}
```
Change the application to use `DBBackend` variant instead of `soci::session`.

3) When a database goes away - say all the sql databases go away and we no longer need `soci:session`, we remove that type from the `DBBackend` variant, (if there's only one variant left, we remove the variant all together and replace it with the single type), remove the overloads and change the visitors.

What I like about the above is it keeps the software simple for the case we have today - free functions are very easy to understand and navigate. Adding new database backends is going to be a very rare thing, but when we do it's easy to change the software to support it. When we remove any backends, the software can evolve back to the very simple state (1).

I think that evolves well and is simpler than the approach in this PR.

Are there important use-cases that the above doesn't address or are there advantages to the approach in this PR that I'm missing?

cjcobb23:
@seelabs I think your approach would work fine, but I think the `SQLInterface` solution would be simpler and easier to use. `SQLInterface` will force the programmer to implement all of the required methods, and won't compile if the programmer does not. Also, all of the call sites will auto dispatch to the correct database. With free functions, the programmer needs to be sure to implement every required free function for the new database implementation, and to also manually change every call site.  I think using an interface makes it a lot easier to add a new database implementation.

The nodestore backend has an interface `Nodestore::Backend` that is simple, intuitive and easy to use. There are RocksDB and NuDB backends, as well as a MemoryBackend. I think we should follow that pattern, not only because it works well, but also for consistency throughout the codebase. 

More generally, an interface seems like the most natural solution here: all of the databases need to implement a certain set of functions, and there can be many possible different databases. To me, free functions will work, but are actually going to be harder to navigate, understand and work with than an interface. What exactly about the interface solution do you find complicated?

p2peer:
> Are there important use-cases that the above doesn't address or are there advantages to the approach in this PR that I'm missing?

Just want to point one more time that DatabaseCon was already used in the rippled code, and the checkoutDb() locking method of DatabaseCon was already used. It is not clear to me what will happen if we just remove DatabaseCon and the locking functionality, and remain pure soci session. If this removal is OK than we can do this as a separate cleaning code task prior to introducing SQL Interface.

seelabs:
@p2peer I would not remove the `DatabaseCon` class. The session would be checked out of the class before feeding it to the free functions.

p2peer:
> Are there important use-cases that the above doesn't address or are there advantages to the approach in this PR that I'm missing?

Also the idea which Mickey told me was to remove node databases in the future. In this case we would not have any soci session for node database and will use many soci sessions for shard databases instead.

seelabs:
@cjcobb23 The free function approach would also not compile if a required method is not implemented. For the current non-variant implementation, it just calls the free function directly, so it will fail if you don't have free function. For the variant approach, we would use `std::visit` or equivalent. If one of the variant options were missing, it would also fail to compile.

We would not need to change the call sites either. The calls would look something like:
```
auto const db = app.getWalletDb().checkoutDb();
insertSomeFoo(*db, foo);
```
In the first implementation where we just `soci::session`, `*db` returns the `session` it calls the `soci::session` overload. If we change this to add a variant, `*db` returns a `std::variant` and calls the variant overload. The call site stays the same.

The nodestore classes are close to what `soci` does. It needs to manage the connection to the keystore (like soci manages the connection to the database) and it needs to provide the query language to the key/value store: fetch and store (like soci lets send sql queries). If we were building a replacement for soci, I would use classes and inheritance.

But look at the interface proposed in this PR. There is a member function on the `SQLInterface_sqlite` class:
```
    virtual void
    insertAcquireDBIndex(SQLDatabase& db, std::uint32_t index_) = 0;
```
The `SQLDatabase` parameter is essentially a `soci::session`. That's a class that wraps `std::variant<bool, soci::session, DatabaseCon>`. It then gets the `soci::session` and from either the second or third variant and executes the sql through the session. If it's the first variant it throws - so this already isn't type safe (but I'm sure we could fix that).

The `SQLInterface_sqlite` has no non static data members and inherits no non static data members (unless I missed them). This is essentially a free function now, we're just using virtual functions in anticipation of a future change. Right now, we don't need overloading and virtual functions at all - we can easily use a single free function.

Once we do add a new backend type - the shard example - then we might want to use a class hierarchy. But this is a straight forward change if we decide to do that (we do need to change the callers in that case, but it's not error prone). I'd probably still want to use overloads, tho I could see how an inheritance based solution could work too (tho I probably wouldn't separate out the database manager from the interface, like in this PR). In the mean time, we don't pay for that complexity until we actually need it.

p2peer:
In general I think that rippled is stable working software and no changes is required except of bug fixing. For this particular case I think that we should wait for Mickey come back from vacation to decide if we should keep already implemented schema of the interface or use soci::session as a parameter instead.

seelabs:
How does the current interface handle adding a new database backend? I assume would make a subclass of `SQLDatabase`. What happens if I call a virtual function that resolves to a `SQLInterface_sqlite` member function and the `SQLDatabase` resolves to (say) a Casandra database? It looks like we'd probably throw. Even if we don't use free functions, I think we want an API where it's impossible to call a function with the wrong database wrapper type. Free functions can do that, and an inheritance-based solution where the class owns the connection can do this too. But I think that can be an issue with the current API.

p2peer:
> How does the current interface handle adding a new database backend? I assume would make a subclass of `SQLDatabase`. What happens if I call a virtual function that resolves to a `SQLInterface_sqlite` member function and the `SQLDatabase` resolves to (say) a Casandra database? It looks like we'd probably throw. Even if we don't use free functions, I think we want an API where it's impossible to call a function with the wrong database wrapper type. Free functions can do that, and an inheritance-based solution where the class owns the connection can do this too. But I think that can be an issue with the current API.

If speaking on current implementation, each SQLDatabase_ object has a pointer to corresponding SQLInterface class implementation. This pointer used to call the method from appropriate backend. Correctness of the pointer is not checked now, but it can be fixed.

src/ripple/core/SQLInterface.h
```diff
@@ -0,0 +1,674 @@
...
+private:
+    static std::unordered_map<SQLInterface::DatabaseType, SQLInterface*>
+        type2iface_;
+    static std::unordered_map<SQLDatabase_*, DatabaseIndex> db2ind_;
+    static std::map<LedgerIndex, SQLDatabase_*> txInd2db_;
+    static std::map<LedgerIndex, SQLDatabase_*> lgrInd2db_;
+    static std::mutex maps_mutex_;
+};
+
+class SQLDatabase_
```
cjcobb23, line 632: I don't think the codebase suffixes any class names with underscores. I'd remove this underscore

src/ripple/core/SQLInterface.h
```diff
@@ -0,0 +1,674 @@
...
+        SQLInterface::removeDatabase(this);
+    }
+
+    SQLInterface*
+    getInterface()
+    {
+        return iface_;
+    }
+
+    SQLInterface* iface_;
```
cjcobb23, line 650: I think the indirection between `SQLInterace` and `SQLDatabase` is confusing. `SQLDatabase` has a `SQLInterface` data member, but a bunch of methods of `SQLInterface` take a `SQLDatabase` as a parameter. I think we should merge these two classes/interfaces

src/ripple/core/SQLInterface.h
```diff
@@ -0,0 +1,674 @@
...
+        beast::Journal& j,
+        AccountID const& account,
+        NetworkOPs::AccountTxs& ret,
+        std::int32_t minLedger,
+        std::int32_t maxLedger,
+        bool descending,
+        std::uint32_t offset,
+        int limit,
+        int limit_used,
+        bool bUnlimited) = 0;
```
cjcobb23, line 427: Some of these functions have a lot of parameters, some of which are the same types. I would consider packing some of these parameters together inside of a struct, which can then be passed to the function.

src/ripple/core/SQLInterface.h
```diff
@@ -0,0 +1,674 @@
...
+
+namespace ripple {
+
+class SQLDatabase_;
+
+class SQLDatabase : public std::unique_ptr<SQLDatabase_>
+{
+};
+
+class SQLInterface
```
cjcobb23, line 41: I think this interface has way too many responsibilities. I would consider breaking up the responsibilities into separate classes. The static methods and static members could go into some manager type of class. Most of these methods only read from one of several tables (such as ledgers or transactions). I would consider creating a different class to manage access to each of those tables. For instance, we could have a class `LedgerDB` and `TransactionDB`. I think this would make the code more readable and easier to manage.

src/ripple/core/SQLInterface.h
```diff
@@ -0,0 +1,674 @@
...
+        SQLInterface::SQLLedgerInfo& info,
+        beast::Journal const& j,
+        LedgerIndex ledgerSeq) = 0;
+
+    virtual bool
+    loadLedgerInfoByIndex(
+        SQLDatabase_* db,
+        SQLInterface::SQLLedgerInfo& info,
+        beast::Journal const& j,
+        LedgerIndex ledgerSeq) = 0;
```
cjcobb23, line 301: There are a bunch of functions with the same name, that do the same thing, that differ in one parameter. Can we merge these functions together?

src/ripple/core/sql_backend/SQLInterface_sqlite.cpp
```diff
@@ -0,0 +1,3747 @@
...
+        bool count,
+        bool bUnlimited);
+};
+
+static bool
+exists(SQLDatabase& db)
+{
+    if (db)
+    {
+        SQLDatabase_sqlite* sdb = dynamic_cast<SQLDatabase_sqlite*>(&*db);
```
cjcobb23, line 576: The `SQLDatabase` type is odd. It's a pointer in disguise. Working with a type thats secretly a pointer is error prone and hard to read. I think we can get rid of the `SQLDatabase_` and just use `std::unique_ptr<SQLDatabase>` when we need a pointer to a `SQLDatabase`. `&*db` is pretty confusing.

src/ripple/core/sql_backend/SQLInterface_sqlite.cpp
```diff
@@ -0,0 +1,3747 @@
...
+        if (sdb)
+            switch (sdb->db_.index())
+            {
+                case 1:
+                    return true;
+                case 2:
+                    return true;
```
cjcobb23, line 583: These both return true. We can use case fall-through here

src/ripple/core/sql_backend/SQLInterface_sqlite.cpp
```diff
@@ -0,0 +1,3747 @@
...
+        case VACUUM:
+            return stateDBName;
+        case PEER_FINDER:
+            return "peerfinder";
+    }
+    return "Unknown";
+}
+
+std::tuple<bool, SQLDatabase, SQLDatabase>
+SQLInterface_sqlite::makeLedgerDBs(
```
cjcobb23, line 683: This function makes the ledger database and the transaction database. I would change the name

src/ripple/core/sql_backend/SQLInterface_sqlite.cpp
```diff
@@ -0,0 +1,3747 @@
...
+        boost::filesystem::path const& dataDir,
+        std::string const& DBName,
+        std::array<char const*, N> const& pragma,
+        std::array<char const*, M> const& initSQL)
+        : SQLDatabase_(iface)
+        , db_(std::in_place_index<2>, dataDir, DBName, pragma, initSQL)
+    {
+    }
+
+    std::variant<bool, soci::session, DatabaseCon> db_;
```
cjcobb23, line 75: This data member should be private

src/ripple/core/sql_backend/SQLInterface_sqlite.cpp
```diff
@@ -0,0 +1,3747 @@
...
+               "  Key                    INTEGER PRIMARY KEY,"
+               "  WritableDb             TEXT,"
+               "  ArchiveDb              TEXT,"
+               "  LastRotatedLedger      INTEGER"
+               ");";
+
+    session << "CREATE TABLE IF NOT EXISTS CanDelete ("
+               "  Key                    INTEGER PRIMARY KEY,"
+               "  CanDeleteSeq           INTEGER"
+               ");";
```
cjcobb23, line 867: I think all of the table schemas should live together, in DBInit.h

src/ripple/core/SQLInterface.h
```diff
@@ -0,0 +1,674 @@
...
+
+    static LedgerIndex ledgersPerShard_;
+
+private:
+    static std::unordered_map<SQLInterface::DatabaseType, SQLInterface*>
+        type2iface_;
+    static std::unordered_map<SQLDatabase_*, DatabaseIndex> db2ind_;
+    static std::map<LedgerIndex, SQLDatabase_*> txInd2db_;
+    static std::map<LedgerIndex, SQLDatabase_*> lgrInd2db_;
+    static std::mutex maps_mutex_;
```
cjcobb23, line 629: As I mentioned before, I think these static methods should go into a manager class.

p2peer:
> I would like to use a different name than SQLInterface and SQLDatabase. This interface defines a bunch of functions, but those functions don't actually need to be implemented in SQL. SQL is an implementation detail. If we break each of the dbs out into separate classes, we could just call them `LedgerDatabase` , `TransactionDatabase` and so on. I think the name SQLInterface is too specific.

Am I right that you want to split SQLInterface class into several ones like LedgerDatabase, TransactionDatabase and other? Note that Ledger and Transactions are used simultaneously in some methods. But I agree, any other database types (for ex, AcquireDB, ArchiveDB etc) can be moved into separate class. Also, do you propose to merge functionality of SQLInterface and SQLDatabase? I mean, class LedgerDatabase represents the database and also has methods to work with such a database.

cjcobb23:
Yes I want to split SQLInterface into several smaller interfaces/classes. And yes, I think SQLDatabase should be merged with SQLInterface.

p2peer:
As for now we have mutually exclusive propositions from 2 reviewers:
1) use global functions with soci::session as parameter
2) use several classes with Inderface=Database.
It would be great if reviewers come with common proposition.

scottschurr:
I'm wondering whether this is still a viable pull request.  It looks like this may have been superseded by https://github.com/ripple/rippled/pull/3590?  If this is no longer a viable pull request it would be great if it were closed.  Thanks.

PR #3532, opened 2020-07-13
Reorder the Travis build

ximinez:
* Builds Windows dependencies first.
* Builds ALL OSs in the last stage.
* Windows dependency stages are allowed to fail so ALL configurations will
      attempt to build. Windows builds will probably fail if dependencies fail
      (caching may allow them to succeed), but they will at least be attempted.
* Remove broken AppVeyor config file, so it stops trying.

---
This change is [Reviewable](https://reviewable.io/reviews/ripple/rippled/3532)


scottschurr:
I don't see how these changes could cause this symptom, but with these changes a local build for me (macOS) doesn't seem to see the already build local object files.  It rebuilds them.  I've now run a unity build two times consecutively.  Both builds appear to build all of the local files (the grpc_pbufs and unity_X_cxx files) even though the first build should have left good object files behind.

It looks like debug build is similarly affected.

ximinez:
@scottschurr Well that's weird. Can you try the same build experiment without this commit, just to confirm that it's new behavior?

scottschurr:
Sure thing.  I just started the experiment now.  I'll let you know what I see.

scottschurr:
@ximinez, good call.  I removed your commit and re-ran the experiment.  It look like this behavior was not introduced by your change.

scottschurr:
Even more interesting, the problem is local to my laptop.  My Mac mini does not exhibit the same behavior.

scottschurr:
I'm watching the build run on my personal Github CI.  I'm noticing that the four dependencies run first and they run serially.  So it takes quite a while before any useful build results (actual builds of rippled) show up.

For the longer term, not necessarily in this pull request, I wonder if we could get the four dependences to build in parallel.  It seems like that might be possible, since I don't think there are any interactions between the four dependencies.  Building the four dependencies in parallel could noticeably improve the CI time.

In the mean time, I'm happy to pay the price of slower start ups in order to have all builds complete successfully.

ximinez:
@scottschurr Yes, I'm definitely making trade-offs. Those deps each check the cache, and short circuit if their artifacts are already there. So this build is probably very slow for you, but I'll bet subsequent builds will be a lot faster. Most importantly, though, since they all write to the cache, that's the interaction. I'm not 100% sure, but I think one job could clobber the cache from another.

Now, if there was a way to have a "dependency check" job that made the others abort, and not even take the time to startup and check, that would save a small chunk of time.

Builds/CMake/deps/Lz4.cmake
```diff
@@ -75,5 +77,7 @@ endif()
 
 add_dependencies (lz4_lib lz4)
```
gregtatcam, line 78: I know it's part of the original script. But shouldn't we have this dependency only when lz4 is added via ExternalProject_Add? lz4 is a custom target created by ExternalProject_Add.

Builds/CMake/deps/Lz4.cmake
```diff
@@ -75,5 +77,7 @@ endif()
...
 target_link_libraries (ripple_libs INTERFACE lz4_lib)
-exclude_if_included (lz4)
+if (${lz4_project})
+  exclude_if_included (lz4)
+endif()
 exclude_if_included (lz4_lib)
```
gregtatcam, line 83: Shouldn't lz4_lib also be excluded only when it's added via ExternalProject_Add?

ximinez: `lz4_lib` is added on line 5 regardless. `exclude_if_included` may not be the best name - it basically keeps the library from building if the whole project is included in another project. So, no, it should always be checked for exclusion.

Builds/CMake/deps/Lz4.cmake
ximinez: My suspicion is that it's going to be needed by all builds because `lz4` can also be discovered by `find_package` on line 8.

There's an easy way to find out, though instead of just speculating: https://github.com/ripple/rippled/pull/3532/commits/4773981698ed3a754f1b5496a9710eda730956fa
And I'll check back on the Travis build later to see if it succeeds. If it does, then the dependency isn't necessary when found. If it doesn't then I'll rollback that commit. https://travis-ci.com/github/ripple/rippled/builds/176060773

gregtatcam: Sounds good

ximinez: Build worked, so I'll keep that commit.

Issue #3502, opened 2020-07-09
Improve serialization of the "FeeSettings" ledger node


carlhua:
The JSON serialization of the "FeeSettings" ledger node is inconsistent:

```
{
   "BaseFee": "000000000000000A",
   "Flags": 0,
   "LedgerEntryType": "FeeSettings",
   "ReferenceFeeUnits": 10,
   "ReserveBase": 20000000,
   "ReserveIncrement": 5000000,
   "index": "4BC50C9B0D8515D3EAAE1E74B29A95804346C491EE1A95BF25E4AAB854A6A651"
}
```
The BaseFee, ReserveBase, and ReserveIncrement should all be integer strings since they represent XRP.

The ReferenceFeeUnits can stay as a JSON number since it's in fee units, not XRP, so it's limited to 32 bits.

Exported from RIPD-1550

movitto:
This is a duplicate of #3165

carlhua:
Thanks @movitto - closing

Issue #3165, opened 2019-11-27
Fix serialization of FeeSettings object


mDuo13:
The [`FeeSettings` ledger object type](https://xrpl.org/feesettings.html) has some weird behavior when it's serialized to JSON (In `ledger_entry` or `ledger_data` responses, or in the metadata of [a SetFee transaction](https://xrpl.org/websocket-api-tool.html?server=wss%3A%2F%2Fs2.ripple.com%2F&req=%7B%22id%22%3A%22setfee_nonunique_hash_1%22%2C%22command%22%3A%22transaction_entry%22%2C%22tx_hash%22%3A%221C15FEA3E1D50F96B6598607FC773FF1F6E0125F30160144BE0C5CBC52F5151B%22%2C%22ledger_index%22%3A3715073%7D)):

- `BaseFee` is serialized as hex despite being an XRP amount (UInt64)
- `ReferenceFeeUnits`, `ReserveBase`, and `ReserveIncrement` are serialized as JSON numbers, not strings as XRP usually is. (This is because they are actually UInt32 types instead of the usual UInt64.)

It would be nice to change the binary format of the `FeeSettings` object to make it use the proper data type for XRP amounts in all its XRP fields, but that's harder to fix because it would require an amendment.

mDuo13:
Note that with the given data structures the [maximum theoretical values for fee settings](https://xrpl.org/fee-voting.html#maximum-fee-values) are, shall we say, strange. The reserve settings, which would make sense to be much larger than the costs of sending a transaction, are more limited.

intelliot:
Related: https://github.com/XRPLF/rippled/issues/4425

intelliot:
@mDuo13 can you confirm this issue has been fixed as of what's in `develop` now?

intelliot:
**mDuo13** noted that this is probably fine. If anyone sees any potential issues, please comment here or create a new issue.

PR #3546, opened 2020-07-17
Changes baseFee, reserveBase, reserveIncrement to XRPAmount

movitto:
Addresses weird custom serialization of these fields, see
issues #3165 and #3502 on github.

Work in progress, submitting for initial feedback, will require
an amendment to transaction from the old representation to the
new.

Tests updated to pass.

### DIFF
--- a/src/ripple/app/ledger/Ledger.cpp
+++ b/src/ripple/app/ledger/Ledger.cpp
@@ -549,16 +549,16 @@ Ledger::setup(Config const& config)
             // VFALCO NOTE Why getFieldIndex and not isFieldPresent?
 
             if (sle->getFieldIndex(sfBaseFee) != -1)
-                fees_.base = sle->getFieldU64(sfBaseFee);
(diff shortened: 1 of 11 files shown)


nbougalis:
This change, as is, would cause massive breakage. Implementing this properly would, *at the very least*, require an amendment, and support for "dual-type" fields (U64 prior to the amendment, Amount afterwards).

If we go that far—and make no mistake, it's a **long** journey—there are other serialization improvements we could make, including the ledger header.

I don't mean to dissuade you from pursuing this, but I think that the cost-benefit ratio is simply not favorable.

carlhua:
@movitto Thank you for your contribution! rippled can always use more contributions! However, like @nbougalis pointed out, the changes would be a breaking change. I am going to convert this PR to draft as we iterate on this.

nbougalis:
@movitto, I'm going to close this PR. Please feel free to reopen it at a later time if you decide to pursue this further. :vulcan_salute:

PR #3553, opened 2020-07-17
[TRIVIAL] Add missing cstdint include:

seelabs:
The gcc non-unity build breaks without this include.


nbougalis:
This should be included in **1.6.0**.

nbougalis:
@manojsdoshi: this needs to be included in **rc2**.

PR #3554, opened 2020-07-17
Added GRPCServer to Stoppable Hierarchy

### LINKED ISSUE #3359: Add GRPCServer to Stoppable hierarchy
## Summary
`GRPCServer` is not part of the `Stoppable` hierarchy, yet it uses the `JobQueue` (which is part of the `Stoppable` hierarchy) in an asynchronous manner. This issue is to investigate whether `GRPCServer` should be part of the `Stoppable` hierarchy, and to implement the change if necessary.

natenichols:
* Fixes #3359, adding `GRPCServer` class to `Stoppable` Hierarchy.

### Context of Change
* Started `GRPCServer` using `onStart` that is triggered by `Application::doStart()`, instead of directly calling `GRPCServer::run()` in `Application::setup()`
* Removed `~GRPCServer`, and instead tear down the server in `GRPCServer::onStop()`
* Make `JobQueue` parent to `GRPCServer`

### Type of Change
- [ ] Documentation Updates
- [ ] Release

## Test Plan
Passes all existing unit test. When messages were logged in `onStop` in both `JobQueue` and `GRPCServer`, the GRPC Server is stopped before the JobQueue, as is expected.

### DIFF
--- a/src/ripple/app/main/GRPCServer.cpp
+++ b/src/ripple/app/main/GRPCServer.cpp
@@ -441,10 +441,11 @@ void
 GRPCServer::onStart()
 {
     // Start the server and setup listeners
-    if ((running_ = impl_.start()))
+    if (running_ = impl_.start(); running_)
     {
         thread_ = std::thread([this]() {
             // Start the event loop and begin handling requests
+            beast::setCurrentThreadName("rippled: grpc");
             this->impl_.handleRpcs();
         });
     }
@@ -457,9 +458,15 @@ GRPCServer::onStop()
     {
         impl_.shutdown();
         thread_.join();
+        running_ = false;
     }
 
     stopped();
+}
+
+GRPCServer::~GRPCServer()
+{
+    assert(!running_);
 }
 
 }  // namespace ripple
--- a/src/ripple/app/main/GRPCServer.cpp
+++ b/src/ripple/app/main/GRPCServer.cpp
@@ -450,13 +450,16 @@ GRPCServer::run()
     }
 }
 
-GRPCServer::~GRPCServer()
+void
+GRPCServer::onStop() 
 {
     if (running_)
     {
         impl_.shutdown();
         thread_.join();
     }
+
+    stopped();
 }
 
 }  // namespace ripple
--- a/src/ripple/app/main/GRPCServer.cpp
+++ b/src/ripple/app/main/GRPCServer.cpp
@@ -438,7 +438,7 @@ GRPCServerImpl::start()
 }
 
 void
-GRPCServer::run()
+GRPCServer::onStart()
 {
     // Start the server and setup listeners
     if ((running_ = impl_.start()))
--- a/src/ripple/app/main/GRPCServer.cpp
+++ b/src/ripple/app/main/GRPCServer.cpp
@@ -451,7 +451,7 @@ GRPCServer::onStart()
 }
 
 void
-GRPCServer::onStop() 
+GRPCServer::onStop()
 {
     if (running_)
     {
--- a/src/ripple/app/main/GRPCServer.cpp
+++ b/src/ripple/app/main/GRPCServer.cpp
@@ -18,6 +18,7 @@
 //==============================================================================
 
 #include <ripple/app/main/GRPCServer.h>
+#include <ripple/beast/core/CurrentThreadName.h>
 #include <ripple/resource/Fees.h>
 
 namespace ripple {
--- a/src/ripple/app/main/GRPCServer.h
+++ b/src/ripple/app/main/GRPCServer.h
@@ -254,7 +254,7 @@ public:
     void
     onStop() override;
 
-    ~GRPCServer(){};
+    ~GRPCServer() = default;
 
 private:
     GRPCServerImpl impl_;
--- a/src/ripple/app/main/GRPCServer.h
+++ b/src/ripple/app/main/GRPCServer.h
@@ -15,7 +15,7 @@
     ACTION  OF  CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
     OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
 */
-//======================================================å========================
+//==============================================================================
 
 #ifndef RIPPLE_CORE_GRPCSERVER_H_INCLUDED
 #define RIPPLE_CORE_GRPCSERVER_H_INCLUDED
--- a/src/ripple/app/main/GRPCServer.h
+++ b/src/ripple/app/main/GRPCServer.h
@@ -22,6 +22,7 @@
 
 #include <ripple/app/main/Application.h>
 #include <ripple/core/JobQueue.h>
+#include <ripple/core/Stoppable.h>
 #include <ripple/net/InfoSub.h>
 #include <ripple/protocol/ErrorCodes.h>
 #include <ripple/resource/Charge.h>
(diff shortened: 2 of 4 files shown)


src/ripple/core/Stoppable.h
```diff
@@ -186,9 +186,9 @@ class RootStoppable;
                                                              |
                                                          JobQueue
                                                              |
-        +--------+-----------+-----------+-----------+-------+---+----------+
-        |        |           |           |           |           |          |
-        |    NetworkOPs      |     InboundLedgers    |      OrderbookDB     |
+        +--------+-----------+-----------+-----------+-------+---+----------+----------+
+        |        |           |           |           |           |          |          |
+        |    NetworkOPs      |     InboundLedgers    |      OrderbookDB     |     GRPCServer
```
natenichols, line 191: `clang-format` is failing on my addition to this chart, not quite sure how to fix it.

src/ripple/core/Stoppable.h
```diff
@@ -186,13 +186,14 @@ class RootStoppable;
...
-        |                                            |                      |
-    PeerFinder                                LedgerCleaner            TaskQueue
+        +--------+-----------+---------+------------+---+----+---------+---+
+        |        |           |         |            |   |              |   |
+        |    NetworkOPs      |    InboundLedgers    |   |      OrderbookDB |
+        |                    |                      |  GRPCServer          |
+        |                    |                      |                  Database
+     Overlay        InboundTransactions        LedgerMaster                |
+        |                                           |                      |
+    PeerFinder                                LedgerCleaner           TaskQueue
```
seelabs, line 196: Unfortunately, clang-format still reformat this tree for me.

cjcobb23:
Let's just disable clang format for the stoppable tree: https://releases.llvm.org/3.6.0/tools/clang/docs/ClangFormatStyleOptions.html#disabling-formatting-on-a-piece-of-code

natenichols:
I changed the tree to be <80 characters before I saw @cjcobb23's comment. 
I can just as easily disable clang-format, if that's a better alternative.

cjcobb23:
> I changed the tree to be <80 characters before I saw @cjcobb23's comment.
> I can just as easily disable clang-format, if that's a better alternative.

If you got it to fit, that's preferable.

src/ripple/app/main/GRPCServer.cpp
```diff
@@ -438,7 +438,7 @@ GRPCServerImpl::start()
 }
 
 void
-GRPCServer::run()
+GRPCServer::onStart()
 {
     // Start the server and setup listeners
     if ((running_ = impl_.start()))
```
scottschurr, line 444: I get what this line is doing, but using an extra set of parentheses to prevent the compiler warning is pretty subtle.  I think this line can be made easier to read if we use C++17's if statement with initializer.  So I'll suggest changing the line to look like this:
```
if (running_ = impl_.start(); running_)
```

scottschurr: Someone on the team was recently muttering about how we don't give threads names.  Well, we're just about to create a thread; let's give it a name.  One way to do that would be to change the code inside this `if` to look like this:
```
        thread_ = std::thread([this]() {
            // Start the event loop and begin handling requests
            beast::setCurrentThreadName("rippled: grpc");
            this->impl_.handleRpcs();
        });
```
If you choose to make this change you'll need to add a
```
#include <ripple/beast/core/CurrentThreadName.h>
```
at the top of the file.

src/ripple/app/main/GRPCServer.cpp
```diff
@@ -450,13 +450,16 @@ GRPCServer::run()
...
 }
 
-GRPCServer::~GRPCServer()
+void
+GRPCServer::onStop()
 {
     if (running_)
     {
         impl_.shutdown();
         thread_.join();
```
scottschurr, line 459: Once we're past the `join`, then we're no longer running.  We should note that in member variable `running_`.  So I suggest adding
```
running_ = false;
```
immediately below the call to `join`.

src/ripple/app/main/GRPCServer.cpp
```diff
@@ -450,13 +450,16 @@ GRPCServer::run()
     }
 }
 
-GRPCServer::~GRPCServer()
```
scottschurr, line 453: Since we're now tracking `running_` correctly, I suggest we give `~GRPCServer()` a non-trivial body:
```
GRPCServer::~GRPCServer()
{
    assert (!running_);
}
```

src/ripple/core/Stoppable.h
```diff
@@ -186,13 +186,14 @@ class RootStoppable;
...
-        |                                            |                      |
-    PeerFinder                                LedgerCleaner            TaskQueue
+        +------+---------+--------+----------+---+----+---------+---+
+        |      |         |        |          |   |              |   |
+        |  NetworkOPs    |   InboundLedgers  |   |      OrderbookDB |
+        |                |                   |  GRPCServer          |
+        |                |                   |                  Database
+     Overlay    InboundTransactions     LedgerMaster                |
+        |                                    |                      |
+    PeerFinder                          LedgerCleaner          TaskQueue
```
scottschurr, line 196: Nicely done, updating the diagram!

src/ripple/core/Stoppable.h
```diff
@@ -186,13 +186,14 @@ class RootStoppable;
...
                                                          JobQueue
                                                              |
-        +--------+-----------+-----------+-----------+-------+---+----------+
-        |        |           |           |           |           |          |
-        |    NetworkOPs      |     InboundLedgers    |      OrderbookDB     |
-        |                    |                       |                      |
-     Overlay        InboundTransactions        LedgerMaster             Database
-        |                                            |                      |
-    PeerFinder                                LedgerCleaner            TaskQueue
+        +------+---------+--------+----------+---+----+---------+---+
```
scottschurr, line 189: The smallest of tweaks: consider moving the third `+` from the end so its vertical line aligns with the `|` immediately below `JobQueue`.

src/ripple/app/main/GRPCServer.h
```diff
@@ -21,6 +21,7 @@
 #define RIPPLE_CORE_GRPCSERVER_H_INCLUDED
 
 #include <ripple/app/main/Application.h>
+#include <ripple/beast/core/CurrentThreadName.h>
```
scottschurr, line 24: This may be a little counter intuitive, but in my opinion this include file is actually better in the `.cpp` file than in the `.h` file.  There are two reasons for that:

1. Nothing in the `.h` file is dependent on the contents of CurrentThreadName.h.  You should limit include files to only those places where they are needed.

2. Any `#include` in a header file is promiscuous.  Because header files are included lots of places, including CurrentThreadName.h in GRPCServer.h means that any file that includes GRPCServer.h will also `#include` CurrentThreadName.h.  And then there are includes of includes of includes.  You get the picture.

    Contrariwise, people don't (typically) do a `#include` of a cpp file.  So if you put the `#include` in GRPCServer.cpp then no additional files will be infiltrated by CurrentThreadName.h.

I hope that makes sense.  If not, then ask away...

src/ripple/app/main/GRPCServer.h
```diff
@@ -15,13 +15,12 @@
     ACTION  OF  CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
     OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
 */
-//==============================================================================
+//======================================================å========================
```
scottschurr, line 18: Looks like maybe a non-ascii character crept in here?  Not sure how...

src/ripple/app/main/GRPCServer.h
```diff
@@ -234,20 +235,26 @@ class GRPCServerImpl final
...
 
     void
-    run();
+    onStart() override;
+
+    void
+    onStop() override;
 
-    ~GRPCServer();
+    ~GRPCServer() = default;
```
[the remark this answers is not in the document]
seelabs: Interesting note: the core guidelines used to advise against adding `override` to destructors. They've since changed the advice to adding it. My own feeling: it's OK to add it, but it clearly isn't as useful as for regular functions (there's no danger of introducing an overload by mistake, for example), but it still tags it as a virtual function. So probably good to add to new code, but I'm not in a hurry to fix existing code.

PR #3555, opened 2020-07-17
Added counters for fetch pack requests

mayurbhandary:
I added StatsD counters for inbound and outbound fetch pack requests.


src/ripple/app/ledger/LedgerMaster.h
```diff
@@ -422,12 +422,18 @@ class LedgerMaster : public Stoppable, public AbstractFetchPackContainer
                   collector->make_gauge("LedgerMaster", "Validated_Ledger_Age"))
             , publishedLedgerAge(
                   collector->make_gauge("LedgerMaster", "Published_Ledger_Age"))
+            , outboundFetchPackRequests( 
+                  collector->make_counter("LedgerMaster", "Outbound_Fetch_Pack_Requests"))
+            , inboundFetchPackRequests(
+                  collector->make_counter("LedgerMaster", "Inbound_Fetch_Pack_Requests"))
```
ximinez, line 428: I almost missed this. clang-format wants to wrap these lines: https://github.com/ripple/rippled/pull/3555/checks?check_run_id=883715497. Please fix.

pwang200:
Clang-format check failed. The following gist might help. I will send you more information. 
https://gist.github.com/pwang200/f81824aa1d8b1c01e64553f72d2deb32

src/ripple/app/ledger/LedgerMaster.h
mayurbhandary: Thanks for catching

src/ripple/app/ledger/impl/LedgerMaster.cpp
```diff
@@ -2173,6 +2174,7 @@ LedgerMaster::makeFetchPack(
             << "Built fetch pack with " << reply.objects().size() << " nodes";
         auto msg = std::make_shared<Message>(reply, protocol::mtGET_OBJECTS);
         peer->send(msg);
+        m_stats.inboundFetchPackRequests++;
```
pwang200, line 2177: The meaning of `outboundFetchPackRequests` is clear to me. But I am not sure about `inboundFetchPackRequests`. Since it is increased here, we don't count all inbound requests. We only count the requests with replies. 

I just want to confirm what is being measured, the count of inbound requests or the count of replies?

mayurbhandary: Yes, my intention was to only count requests with replies. I will be sure to note this in the documentation of the counter. Would you recommend that I change the name of the variable? I could also make a separate counter for all inbound requests if that is required.

ximinez: Now that @pwang200 has asked the question, I'm a little confused. When you say "requests with replies", I take that to mean "inbound requests for which we were able to generate a reply". Is that correct? If so, I'd suggest renaming this counter to `outboundFetchPackReplies`.

And I do think it would be worthwhile to create a different `inboundFetchPathReplies` which counts the number of replies to our outbound queries.

pwang200: @mayurbhandary I think the variable should be renamed. And please document the counters clearly.

mayurbhandary: Closing this PR and creating a new one to address these comments. New PR will also include the full set of fetchpack metrics for inbound and outbound requests/replies.

PR #3557, opened 2020-07-21
reserve 22 as ttHOOK_SET

RichardAH:
We'd like to reserve 22 as SetHook transaction type. This tt will eventually allow configurable behaviour on accounts for send and receive events.

### DIFF
--- a/src/ripple/protocol/TxFormats.h
+++ b/src/ripple/protocol/TxFormats.h
@@ -56,6 +56,8 @@ enum TxType {
     ttTRUST_SET = 20,
     ttACCOUNT_DELETE = 21,
 
+    ttHOOK_SET [[maybe_unused]] = 22,
+
     ttAMENDMENT = 100,
     ttFEE = 101,
     ttUNL_MODIFY = 102,
--- a/src/ripple/protocol/TER.h
+++ b/src/ripple/protocol/TER.h
@@ -274,6 +274,7 @@ enum TECcodes : TERUnderlyingType {
     tecKILLED = 150,
     tecHAS_OBLIGATIONS = 151,
     tecTOO_SOON = 152,
+    tecHOOK_ERROR [[maybe_unused]] = 153
 };
 
 //------------------------------------------------------------------------------


src/ripple/protocol/TxFormats.h
```diff
@@ -56,6 +56,8 @@ enum TxType {
     ttTRUST_SET = 20,
     ttACCOUNT_DELETE = 21,
 
+    ttHOOK_SET = 22,
```
nbougalis, line 59: Consider:

    ttHOOK_SET = 22 [[maybe_unused]],

Since the code would be currently unused.

codetsunami: I think the correct syntax for setting an attribute on enum member is
```c++
ttHOOK_SET [[maybe_unused]] = 22,
```
there's a few other unused tx types, should we attribute all of them?

nbougalis: Which other tx types are unused?

MarkusTeufelberger: 6, 9, 10, 11, 102 I guess?

codetsunami: I decided to let you guys work out which others to attribute (if any) in another PR

mDuo13:
I'm not concerned about burning one ID (number 153) out of 64,000. :laughing: 

Hooks is an interesting idea. I think the scope will be paramount. Too big, and you end up with Solidity¹; too small, and you need to keep growing it with more features later. Sometimes having a bunch of specialized tools is useful, sometimes it's too complex and having one all-purpose tool is better.

¹ Solidity is cool, but bugs in smart contracts are a big problem (even the DAO couldn't get their right on the first try, and most of us don't have the option of hardforking the network if we screw up) and the more complexity, the more opportunity for bugs. Also, the more logic and state data you store on-ledger, the harder your scaling problems become.

carlhua:
I am giving a Passed label as this is trivial.

PR #3558, opened 2020-07-23
Boost 1.74

seelabs:
Tested on boost 1.74 beta 1. Note that there appears to be an issue in 1.74 b1, the file `boost/asio/ts/netfwd.hpp` needed to be modified to change line 61 from `typedef execution any_io_executor;` to `typedef executor any_io_executor;` (note: executor, not execution). We should not merge this PR until boost 1.74 is officially released.

This PR also includes another trivial patch to resolve some warnings from Howard's date library (not an issue with the date library itself).

### DIFF
--- a/Builds/CMake/RippledCore.cmake
+++ b/Builds/CMake/RippledCore.cmake
@@ -128,6 +128,7 @@ target_include_directories (xrpl_core
 
 target_compile_definitions(xrpl_core
   PUBLIC
+    BOOST_ASIO_USE_TS_EXECUTOR_AS_DEFAULT
     HAS_UNCAUGHT_EXCEPTIONS=1)
 target_compile_options (xrpl_core
   PUBLIC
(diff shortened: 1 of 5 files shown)


seelabs:
I submitted an issue to asio here: https://github.com/chriskohlhoff/asio/issues/519

HowardHinnant:
I'm getting a unit test crash under `rippled --unittest=ValidatorSite`.  Here's the dump:  https://gist.github.com/HowardHinnant/a1dd60ef9bbaf16cf592ff812dd0c662

This is happening only against the (fixed) boost 1.74b1.  I also tested against boost 1.73 and did not see this.

seelabs:
Thanks @HowardHinnant FYI, runs clean here, but clearly there's an issue. Will look into it.

seelabs:
Thanks @HowardHinnant Obviously there's a problem, will look into it. FYI, it does successfully run that test locally with both boost 1.74 b1 and 1.70.

seelabs:
The `ValidatorSite` failure appears to be an existing bug in the test server. I submitted a PR here with a proposed fix: https://github.com/ripple/rippled/pull/3561

It is curious that it's so repeatable on your machine with boost 1.74 b1, but not previous versions.

miguelportilla:
On Windows using VS2019 with latest update...
*   Using 1.70
    *   Okay
    
*   Using 1.73 and 1.74 B1
    *   Got the following error: `Consensus.h(591): error C2976: 'std::deque': too few template arguments`
    *   Several CMake warnings from FindBoost because of the newer version `New Boost version may have incorrect or missing dependencies and imported targets`

PR #3559, opened 2020-07-23
Try out random DNS resolved endpoints in case of a failure.

### LINKED ISSUE #3494: DNS resolution for validator lists
Validator list site are configured by specifying a URL from which the validator list can be retrieved. The URL can specify either a hostname or an IP address.

If a hostname is specified then we should resolve it every time we need to perform a check against that site; if the resolver response includes a single IP, then we can just use it normally: attempt to connect to that IP and retrieve the list. Store the IP address and the success/failure status in the object and produce a log entry.

If the resolver response includes multiple IPs we need to use more advanced logic. Resolve the hostname again and:

If the previous attempt was successful and if the previously used IP is still valid use that, following the "single IP" logic above.
If this isn't the first attempt and the previous attempt failed, remove the failed IP from the list of resolved addresses if it still exists. Select an IP at random from the remaining choices and follow the "single IP" logic above.

Exported from RIPD-1671

gregtatcam:
If the resolver response includes a single IP, then we can just use it normally: attempt to connect to that IP and retrieve the list. Store the IP address and the success/failure status in the object and produce a log entry.

If the resolver response includes multiple IPs we need to use more advanced logic. Resolve the hostname again and:

If the previous attempt was successful and if the previously used IP is still valid use that, following the "single IP" logic above.
If this isn't the first attempt and the previous attempt failed, remove the failed IP from the list of resolved addresses if it still exists. Select an IP at random from the remaining choices and follow the "single IP" logic above.

FIXES #3494


src/test/app/DNS_test.cpp
```diff
@@ -0,0 +1,129 @@
...
+        // Should have more than one but some endpoints can repeat since
+        // selected at random.
+        BEAST_EXPECT(resolved_.size() > 1);
+    }
+};
+
+BEAST_DEFINE_TESTSUITE_MANUAL_PRIO(DNS, ripple_data, ripple, 20);
+
+}  // namespace test
+}  // namespace ripple
```
nbougalis, line 129: Please add a trailing newline here.

gregtatcam: Done

src/ripple/app/misc/detail/WorkBase.h
```diff
@@ -182,8 +195,30 @@ WorkBase<Impl>::onResolve(error_code const& ec, resolver_type::iterator it)
     if (ec)
         return fail(ec);
 
+    resolver_type::iterator end;
+    std::vector<
+        std::reference_wrapper<resolver_type::iterator::value_type const>>
+        entries;
```
seelabs, line 201: I can see why this is written this way. It's a straight forward way to choose an random endpoint that was not previously failing. However, this code creates a vector or entries to randomly choose from, even when we won't be choosing from the vector. This will allocate memory, which we try to avoid. Even when we will be making a random choice, I don't think we need to create a vector to do so. What do you think about this refactoring?
```
template <class Impl>
void
WorkBase<Impl>::onResolve(error_code const& ec, resolver_type::iterator it)
{
    if (ec)
        return fail(ec);

    lastEndpoint_ = [&]() -> endpoint_type {
        // Use last endpoint if it is successfully connected and is in the list,
        // otherwise pick a random endpoint from the list. If there is only one
        // endpoint and it is the last endpoint then use the last endpoint.

        resolver_type::iterator end;
        auto const foundIt = std::find(it, end, lastEndpoint_);
        if (lastStatus_ && foundIt != end)
            return lastEndpoint_;

        switch (auto const n = std::distance(it, end); n)
        {
            case 0:
                return lastEndpoint_;
            case 1:
                return *it;
            default: {
                // need to choose a random element from the collection that does
                // not equal `lastEndpoint_`` if `lastEndpoint_` is part of the
                // collection.
                if (foundIt == end)
                    return *std::next(it, rand_int(n - 1));
                // lastEndpoint_ is part of the collection
                // Pick a random number from the n-1 valid choices, if we use
                // this as an index, note the last element will never be chosen
                // and the `lastEndpoint_` index may be chosen. So when the
                // `lastEndpoint_` index is chosen, that is treated as if the
                // last element was chosen.
                auto randIndex = (n > 2) ? rand_int(n - 2) : 0;
                auto const foundIndex = std::distance(it, foundIt);
                if (randIndex == foundIndex)
                    randIndex = n - 1;
                return *std::next(it, randIndex);
            }
        }
    }();

    socket_.async_connect(
        lastEndpoint_,
        strand_.wrap(std::bind(
            &Impl::onConnect,
            impl().shared_from_this(),
            std::placeholders::_1)));
}
```

src/test/app/DNS_test.cpp
```diff
@@ -0,0 +1,129 @@
...
+                resolved_.size() == 1 && resolved_.begin()->second == i);
+        }
+        if (!isMultipleEndpoints())
+            return;
+        // Run with the "failed" status. In this case endpoints are selected at
+        // random.
+        for (int i = 0; i < 4; ++i)
+            makeRequest(lastEndpoint_, false);
+        // Should have more than one but some endpoints can repeat since
+        // selected at random.
```
scottschurr, line 121: I ran the test repeatedly and I feel like I have to agree with your conclusion.  However I don't understand _why_.  If the endpoint selection is really random it seems like occasionally all of the random numbers would be the same.  With 4 possible endpoints being selected randomly 5 times I think you get a probability of 4:1024 (or 1:256) that you'd get the same value all five times.

Still, I've run the test over 4096 times and it has not failed.  Why is that?  You should put that explanation in the comments or else someone else will be just as puzzled as me when they maintain the test.  If it's an artifact of the pseudo random number generator, then the test is fragile.

src/ripple/app/misc/detail/WorkBase.h
scottschurr: FWIW, when I saw the `vector` being populated and usually not used I had concerns similar to those expressed by @seelabs.  I'd prefer an approach which either does not use the `vector` or only populates the `vector` if it will actually use the contents.

src/ripple/app/misc/detail/WorkBase.h
```diff
@@ -190,32 +191,39 @@ WorkBase<Impl>::fail(error_code const& ec)
...
+            return *results.begin();
+        else if (foundIt == results.end())
+            return *std::next(results.begin(), rand_int(results.size() - 1));
+
+        // We got more than 1 endpoint and last endpoint is in the list.
+        // If rand is equal to last endpoint then we select either
+        // previous or next endpoint.
+        auto r = rand_int(results.size() - 1);
+        return *std::next(
+            results.begin(), r != foundIndex ? r : (r > 0 ? --r : ++r));
```
seelabs, line 225: I think this is a biased sample. The `foundIndex`'s alternate element has two chances of being selected, while every other element has only one chance of being selected. Consider make the `r` not include the last element. If `foundIndex` is selected, then select the last element instead. This way every element has the same probability of being selected.

gregtatcam: You are right. Changed.

PR #3562, opened 2020-07-27
Maximum Transactions in JobQueue read from Config

### LINKED ISSUE #3556: max_transactions (i.e. mempool) should be a config option rather than hardcoded
https://github.com/ripple/rippled/blob/7b048b423e8ae08a54018a89231d050b9f562855/src/ripple/overlay/impl/PeerImp.cpp#L1510-L1514

Not sure why this isn't a config option already?

natenichols:
Fixes #3556. max_transactions is now read from the `rippled.cfg` instead of hardcoded.

### Context of Change
The maximum capacity of `JobQueue` was hardcoded in `PeerImp.cpp`. Now the maximum capacity is read from the configuration file. `max_transactions` defaults to 250, and has a hardcoded minimum of 100 and a hardcoded maximum of 1000, as suggested in the issue description. 

This option is added to the configuration with: 
```
[max_transactions]
    500
```

### Type of Change
- [ ] Documentation Updates
- [ ] Release

## Before/After
**Before:**
max_transactions was hardcoded.

**After:**
`MAX_TRANSACTIONS` is read from `rippled.cfg` in `Config.cpp`. `std::clamp` is used to ensure the configuration value is between 100 and 1000. Member variable `max_transactions_` was created in `PeerImp`, and initialized to the value read by the configuration.

## Test Plan
All unit tests pass. No tests added.

### DIFF
--- a/src/ripple/overlay/impl/PeerImp.cpp
+++ b/src/ripple/overlay/impl/PeerImp.cpp
@@ -1507,9 +1507,8 @@ PeerImp::onMessage(std::shared_ptr<protocol::TMTransaction> const& m)
             }
         }
 
-        // The maximum number of transactions to have in the job queue.
-        constexpr int max_transactions = 250;
-        if (app_.getJobQueue().getJobCount(jtTRANSACTION) > max_transactions)
+        if (app_.getJobQueue().getJobCount(jtTRANSACTION) >
+            app_.config().MAX_TRANSACTIONS)
         {
             overlay_.incJqTransOverflow();
             JLOG(p_journal_.info()) << "Transaction queue is full";
--- a/cfg/rippled-example.cfg
+++ b/cfg/rippled-example.cfg
@@ -453,6 +453,13 @@
 #
 #
 #
+# [max_transactions]
+#
+#   Configure the maximum number of transactions to have in the job queue
+#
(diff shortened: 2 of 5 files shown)


src/ripple/overlay/impl/PeerImp.cpp
```diff
@@ -1507,9 +1507,7 @@ PeerImp::onMessage(std::shared_ptr<protocol::TMTransaction> const& m)
             }
         }
 
-        // The maximum number of transactions to have in the job queue.
-        constexpr int max_transactions = 250;
-        if (app_.getJobQueue().getJobCount(jtTRANSACTION) > max_transactions)
+        if (app_.getJobQueue().getJobCount(jtTRANSACTION) > max_transactions_)
```
nbougalis, line 1510: Consider instead:

    if (app_.getJobQueueCount().getJobCount(jtTRANSACTION) > app_.config().MAX_TRANSACTIONS)

Rationale: the variable is only set in one place and isn't "per-peer" so this lets you eliminate the unnecessary member variable, and highlights the intent of the code.

natenichols:
Sorry about all the clang-formats, my VSCode on-save linting apparently isn't consistent.

PR #3563, opened 2020-07-28
Fix memory management issues with checkpointers:

seelabs:
The checkpointer class had assumed that the database would exist for the
lifetime of the application. This is no long true. These changes resolve bugs
involving dangling pointers.


seelabs:
Note: only the top commit needs to be reviewed

src/ripple/core/impl/SociDB.cpp
```diff
@@ -196,44 +197,33 @@ namespace {
...
+        sqlite_api::sqlite3& conn,
+        JobQueue& q,
+        Logs& logs)
+        : Id_(id)
+        , conn_(conn)
+        , jobQueue_(q)
+        , j_(logs.journal("WALCheckpointer"))
     {
-        sqlite_api::sqlite3_wal_hook(&conn_, &sqliteWALHook, this);
+        sqlite_api::sqlite3_wal_hook(&conn_, &sqliteWALHook, (void*)Id_);
```
nbougalis, line 214: I'd prefer if we used: `reinterpret_cast<void*>(id_)` instead because this cast is "sufficiently weird" that it should stand out.

src/ripple/core/impl/SociDB.cpp
```diff
@@ -196,44 +197,33 @@ namespace {
...
     class.
 */
+
 class WALCheckpointer : public Checkpointer
 {
 public:
-    WALCheckpointer(sqlite_api::sqlite3& conn, JobQueue& q, Logs& logs)
-        : conn_(conn), jobQueue_(q), j_(logs.journal("WALCheckpointer"))
+    WALCheckpointer(
+        std::uint64_t id,
```
nbougalis, line 205: Should be `std::uintptr_t`

src/ripple/core/impl/SociDB.cpp
nbougalis: Also, do we **enforce** the SQLite requirements for [`sqlite3_wal_hook`](https://www.sqlite.org/c3ref/wal_hook.html)?

>**A single database handle may have at most a single write-ahead log callback registered at one time. Calling sqlite3_wal_hook() replaces any previously registered write-ahead log callback.** Note that the sqlite3_wal_autocheckpoint() interface and the wal_autocheckpoint pragma both invoke sqlite3_wal_hook() and will overwrite any prior sqlite3_wal_hook() settings.

src/ripple/core/SociDB.h
```diff
@@ -131,10 +131,17 @@ convert(std::vector<std::uint8_t> const& from, soci::blob& to);
...
-class Checkpointer
+class Checkpointer : public std::enable_shared_from_this<Checkpointer>
 {
 public:
+    virtual std::uintptr_t
+    id() const = 0;
     virtual ~Checkpointer() = default;
+
+    virtual void
+    scheduleCheckpoint() = 0;
```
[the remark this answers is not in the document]
seelabs: Agreed. Fixed in `1a499c7b4 Styles fixes and use correct int type to cast to pointer`

src/ripple/core/impl/SociDB.cpp
```diff
@@ -272,15 +268,50 @@ class WALCheckpointer : public Checkpointer
         std::lock_guard lock(mutex_);
         running_ = false;
     }
+
+protected:
+    std::uint64_t const Id_;
```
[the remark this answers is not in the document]
seelabs: Agreed. Fixed in `1a499c7b4 Styles fixes and use correct int type to cast to pointer`

src/ripple/core/impl/SociDB.cpp
seelabs: Fixed the cast in `1a499c7b4 Styles fixes and use correct int type to cast to pointer`

I don't think it's an error to call `DatabaseCon::setupCheckpointing` more than once, it would just replace the current checkpointer and everything will continue to work. However, it doesn't really make sense to ever do that. It shouldn't be too hard to make `DatabaseCon::setupCheckpointing` private and make it part of the constructor. I'll look at doing that.

seelabs: @nbougalis I pushed a patch that ensures `setupCheckpointing` is called at most once: `3a6e5ab8a Ensure checkpointing is setup at most once` Take a look. It doubles the number of constructors, but with delegating constructors that's not too bad.

src/ripple/core/impl/SociDB.cpp
seelabs: Agreed. Fixed in `1a499c7b4 Styles fixes and use correct int type to cast to pointer`

src/ripple/core/SociDB.h
```diff
@@ -143,8 +150,8 @@ class Checkpointer
     The Checkpointer contains references to the session and job queue
     and so must outlive them both.
  */
-std::unique_ptr<Checkpointer>
-makeCheckpointer(soci::session&, JobQueue&, Logs&);
+std::shared_ptr<Checkpointer>
+makeCheckpointer(std::uintptr_t id, soci::session&, JobQueue&, Logs&);
```
miguelportilla, line 155: Nit: capitalize checkpointer in comment.

src/ripple/core/impl/DatabaseCon.cpp
```diff
@@ -27,6 +27,61 @@
...
+    // Each checkpointer is given a unique id. All the checkpointers that are
+    // part of a DatabaseCon are part of this collection. When the DatabaseCon
+    // is destroyed, its checkpointer is removed from the collection
+    std::map<std::uintptr_t, std::shared_ptr<Checkpointer>> checkpointers_;
+
+public:
+    std::shared_ptr<Checkpointer>
+    fromId(std::uintptr_t id)
+    {
+        std::lock_guard l{m_};
```
miguelportilla, line 44: Nit: for consistency, name this local 'lock' rather than 'l'.

src/ripple/core/impl/DatabaseCon.cpp
```diff
@@ -27,6 +27,61 @@
...
+        auto it = checkpointers_.find(id);
+        if (it != checkpointers_.end())
+            return it->second;
+        return {};
+    }
+
+    void
+    erase(std::uintptr_t id)
+    {
+        std::lock_guard l{m_};
```
miguelportilla, line 54: Nit: same as above

src/ripple/core/impl/DatabaseCon.cpp
```diff
@@ -27,6 +27,61 @@
...
+        checkpointers_.erase(id);
+    }
+
+    std::shared_ptr<Checkpointer>
+    create(soci::session& s, JobQueue& q, Logs& l)
```
miguelportilla, line 59: Nit: these locals could alternatively be named '(soci::session& session, JobQueue& jobQueue, Logs& logs)'

src/ripple/core/DatabaseCon.h
```diff
@@ -105,10 +105,16 @@ class DatabaseCon
         static std::unique_ptr<std::vector<std::string> const> globalPragma;
     };
 
+    struct CheckpointerSetup
+    {
+        JobQueue* jq;
+        Logs* logs;
+    };
+
```
miguelportilla, line 113: Instead of adding another setup struct to the ctor, consider moving this struct or ptrs into the already existing Setup struct.

src/ripple/core/impl/DatabaseCon.cpp
miguelportilla: Also, consider renaming m_ to mutex_

src/ripple/core/impl/SociDB.cpp
```diff
@@ -243,15 +233,21 @@ class WALCheckpointer : public Checkpointer
         }
 
         // If the Job is not added to the JobQueue then we're not running_.
-        if (!jobQueue_.addJob(jtWAL, "WAL", [this](Job&) { checkpoint(); }))
+        if (!jobQueue_.addJob(
+                jtWAL,
+                "WAL",
+                [wp = std::weak_ptr<Checkpointer>{shared_from_this()}](Job&) {
```
cjcobb23, line 240: Why do we capture the `weak_ptr` instead of `shared_ptr`? The way this is written, if the Checkpointer was deleted before the job is run, the job is a noop. When does this occur? Maybe add a comment talking about this.

src/ripple/core/SociDB.h
seelabs: Fixed in `50335297d [fold] Longer variable names`

src/ripple/core/impl/DatabaseCon.cpp
seelabs: Fixed in `50335297d [fold] Longer variable names`

src/ripple/core/impl/DatabaseCon.cpp
seelabs: Fixed in `50335297d [fold] Longer variable names`

src/ripple/core/impl/DatabaseCon.cpp
seelabs: Fixed in `50335297d [fold] Longer variable names`

src/ripple/core/DatabaseCon.h
seelabs: Fixed in `50335297d [fold] Longer variable names`

seelabs: @miguelportilla I'd rather keep the separate setup struct. There are two sets of constructors, one that takes a setup and one that does not. I'd like to make setting up a checkpointer the same for both types of constructors. The reason it's a struct instead of just two extra params, it I want to make it obvious at the caller that a checkpointer is being setup (because there's a local struct with `checkpointer` in its name).

src/ripple/core/impl/DatabaseCon.cpp
```diff
@@ -27,6 +27,61 @@
...
 
+class CheckpointersCollection
+{
+    std::uintptr_t nextId_{0};
+    // Mutex protects the CheckpointersCollection
+    std::mutex m_;
+    // Each checkpointer is given a unique id. All the checkpointers that are
+    // part of a DatabaseCon are part of this collection. When the DatabaseCon
+    // is destroyed, its checkpointer is removed from the collection
+    std::map<std::uintptr_t, std::shared_ptr<Checkpointer>> checkpointers_;
```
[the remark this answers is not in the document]
seelabs: Fixed in `c28a08286 [fold] Use unordered map`. I agree `unordered_map` is almost always a better choice unless we care about iterating in order. Although in my defense, in this case we're storing a handful of objects a handful of times a day, so the overhead of `map` doesn't _really_ matter.

src/ripple/core/impl/SociDB.cpp
seelabs: This occurs when the `DatabaseCon` object has been destroyed. When this happens, the `conn_` member on the `Checkpointer` is no longer valid, so we can't call the checkpointer.

However, while I was writing this, I realized this patch still has a memory bug (tho much less likely than before). If the weak pointer is locked in the callback and then the database object is destoryed we'll access a bad connection. Will fix.

src/ripple/core/impl/DatabaseCon.cpp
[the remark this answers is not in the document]
cjcobb23: I agree that the overhead of `map` really doesn't matter. I think it's just good practice to use `unordered_map` unless there is a reason not to (iteration, as you mentioned, or something like upper bound).

seelabs:
@cjcobb23 @miguelportilla  I just pushed a new patch that handles a case where the checkpointer could have used an already destroyed sqlite database connection (I also did some minor refactoring in the same patch, but the meat is the new `weak_ptr` to a connection in the checkpointer class). Please review when you get a chance.

PR #3564, opened 2020-07-30
Load transaction and meta-data from SQLite

natenichols:
The Tx RPC call now fetches both the transaction and transaction meta-data from SQLite.

### Context of Change
Previously, the Tx RPC call was querying SQLite for the Transaction, and then querying the key-value store for transaction metadata. Tx now queries SQLite for both transaction and meta-data, since the meta-data was already stored in SQLite along with the transaction. 

### Type of Change
- [ ] Documentation Updates
- [ ] Release

## Before/After
**Before:**
`doTxHelp()` queries `TransactionMaster::Fetch` to get Transaction, then queries the key-value store to get the meta-data. `Transaction::Load()` only fetches `txn` from SQLite.

**After**
`doTxHelp()` queries `TransactionMaster::fetch` for both Transaction and meta-data. TransactionMaster::fetch checks the cache for transaction, if it exists and is validated, the meta-data is queried from SQLite, otherwise only the transaction is fetched.  `Transaction::Load()` returns both `txn` and `txnMeta` from SQLite.

## Test Plan
All unit tests pass. No new tests added.

### DIFF
--- a/src/ripple/app/ledger/impl/TransactionMaster.cpp
+++ b/src/ripple/app/ledger/impl/TransactionMaster.cpp
@@ -56,20 +56,19 @@ TransactionMaster::fetch_from_cache(uint256 const& txnID)
 
 std::variant<
     std::pair<std::shared_ptr<Transaction>, std::shared_ptr<TxMeta>>,
-    TxSearchedAll>
+    TxSearched>
 TransactionMaster::fetch(uint256 const& txnID, error_code_i& ec)
 {
     using TxPair =
         std::pair<std::shared_ptr<Transaction>, std::shared_ptr<TxMeta>>;
 
-    // txn->getLedger() == 0 txn is not validated and does not have metadata
-    if (auto txn = fetch_from_cache(txnID); txn && txn->getLedger() == 0)
+    if (auto txn = fetch_from_cache(txnID); txn && !txn->isValidated())
         return std::pair{std::move(txn), nullptr};
 
     auto v = Transaction::load(txnID, mApp, ec);
 
-    if (auto e = std::get_if<TxSearchedAll>(&v))
-        return *e;
+    if (std::holds_alternative<TxSearched>(v))
+        return v;
 
     auto [txn, txnMeta] = std::get<TxPair>(v);
 
@@ -81,7 +80,7 @@ TransactionMaster::fetch(uint256 const& txnID, error_code_i& ec)
 
 std::variant<
     std::pair<std::shared_ptr<Transaction>, std::shared_ptr<TxMeta>>,
-    TxSearchedAll>
+    TxSearched>
 TransactionMaster::fetch(
     uint256 const& txnID,
     ClosedInterval<uint32_t> const& range,
@@ -90,14 +89,13 @@ TransactionMaster::fetch(
     using TxPair =
         std::pair<std::shared_ptr<Transaction>, std::shared_ptr<TxMeta>>;
 
-    // txn->getLedger() == 0 txn is not validated and does not have metadata
-    if (auto txn = fetch_from_cache(txnID); txn && txn->getLedger() == 0)
+    if (auto txn = fetch_from_cache(txnID); txn && !txn->isValidated())
         return std::pair{std::move(txn), nullptr};
 
     auto v = Transaction::load(txnID, mApp, range, ec);
 
-    if (auto e = std::get_if<TxSearchedAll>(&v))
-        return *e;
+    if (std::holds_alternative<TxSearched>(v))
+        return v;
 
     auto [txn, txnMeta] = std::get<TxPair>(v);
 
--- a/src/ripple/app/ledger/impl/TransactionMaster.cpp
+++ b/src/ripple/app/ledger/impl/TransactionMaster.cpp
@@ -56,7 +56,7 @@ TransactionMaster::fetch_from_cache(uint256 const& txnID)
 
 std::variant<
     std::pair<std::shared_ptr<Transaction>, std::shared_ptr<TxMeta>>,
-    SearchedAll>
+    TxSearchedAll>
 TransactionMaster::fetch(uint256 const& txnID, error_code_i& ec)
 {
     using TxPair =
@@ -68,7 +68,7 @@ TransactionMaster::fetch(uint256 const& txnID, error_code_i& ec)
 
     auto v = Transaction::load(txnID, mApp, ec);
 
-    if (auto e = std::get_if<SearchedAll>(&v))
+    if (auto e = std::get_if<TxSearchedAll>(&v))
         return *e;
 
     auto [txn, txnMeta] = std::get<TxPair>(v);
@@ -81,7 +81,7 @@ TransactionMaster::fetch(uint256 const& txnID, error_code_i& ec)
 
 std::variant<
     std::pair<std::shared_ptr<Transaction>, std::shared_ptr<TxMeta>>,
-    SearchedAll>
+    TxSearchedAll>
 TransactionMaster::fetch(
     uint256 const& txnID,
     ClosedInterval<uint32_t> const& range,
@@ -96,7 +96,7 @@ TransactionMaster::fetch(
 
     auto v = Transaction::load(txnID, mApp, range, ec);
 
-    if (auto e = std::get_if<SearchedAll>(&v))
+    if (auto e = std::get_if<TxSearchedAll>(&v))
         return *e;
 
     auto [txn, txnMeta] = std::get<TxPair>(v);
--- a/src/ripple/app/ledger/impl/TransactionMaster.cpp
+++ b/src/ripple/app/ledger/impl/TransactionMaster.cpp
@@ -54,44 +54,57 @@ TransactionMaster::fetch_from_cache(uint256 const& txnID)
     return mCache.fetch(txnID);
 }
 
-std::shared_ptr<Transaction>
+std::optional<std::pair<std::shared_ptr<Transaction>, std::shared_ptr<TxMeta>>>
 TransactionMaster::fetch(uint256 const& txnID, error_code_i& ec)
 {
-    auto txn = fetch_from_cache(txnID);
+    // txn->getLedger() == 0 then txn is not validated and does not contain
+    // metadata
+    if (auto txn = fetch_from_cache(txnID); txn && txn->getLedger() == 0)
+        return std::pair{std::move(txn), nullptr};
+
+    auto v = Transaction::load(txnID, mApp, ec);
+
+    if (!v)
+        return std::nullopt;
+
+    auto [txn, txnMeta] = v.value();
 
     if (txn)
-        return txn;
+        mCache.canonicalize_replace_client(txnID, txn);
 
-    txn = Transaction::load(txnID, mApp, ec);
-
-    if (!txn)
-        return txn;
-
-    mCache.canonicalize_replace_client(txnID, txn);
-
-    return txn;
+    return std::pair{std::move(txn), std::move(txnMeta)};
 }
 
-boost::variant<Transaction::pointer, bool>
+std::optional<std::variant<
+    std::pair<std::shared_ptr<Transaction>, std::shared_ptr<TxMeta>>,
+    bool>>
 TransactionMaster::fetch(
     uint256 const& txnID,
     ClosedInterval<uint32_t> const& range,
     error_code_i& ec)
 {
-    using pointer = Transaction::pointer;
+    using TxPair =
+        std::pair<std::shared_ptr<Transaction>, std::shared_ptr<TxMeta>>;
 
-    auto txn = mCache.fetch(txnID);
+    // txn->getLedger() == 0 then txn is not validated and does not contain
+    // metadata
+    if (auto txn = fetch_from_cache(txnID); txn && txn->getLedger() == 0)
+        return std::pair{std::move(txn), nullptr};
+
+    auto v = Transaction::load(txnID, mApp, range, ec);
+
+    if (!v)
+        return std::nullopt;
+
+    if (auto b = std::get_if<bool>(&v.value()))
+        return *b;
+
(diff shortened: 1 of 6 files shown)


src/ripple/app/misc/impl/Transaction.cpp
```diff
@@ -169,11 +180,20 @@ Transaction::load(
...
     }
 
     try
     {
-        return Transaction::transactionFromSQL(ledgerSeq, status, rawTxn, app);
+        auto txn =
+            Transaction::transactionFromSQL(ledgerSeq, status, rawTxn, app);
+
+        std::uint32_t const inLedger =
+            rangeCheckedCast<std::uint32_t>(ledgerSeq.value_or(0));
```
seelabs, line 192: I'm not sure `ledgerSeq` should ever be unseated here. We should check if we ever store metadata without a ledger seq (I don't think we do, but it should be easy enough to audit). If we don't then it's an error to get metadata without a ledger seq, and we shouldn't fill in a dummy value here. I'd treat it as if the metadata was not in the database.

src/ripple/app/ledger/impl/TransactionMaster.cpp
```diff
@@ -54,44 +54,57 @@ TransactionMaster::fetch_from_cache(uint256 const& txnID)
...
+    if (auto txn = mCache.fetch(txnID); txn && txn->getLedger() == 0)
+        return std::pair{std::move(txn), nullptr};
 
-    if (txn)
-        return txn;
+    auto v = Transaction::load(txnID, mApp, ec);
 
-    txn = Transaction::load(txnID, mApp, ec);
+    if (!v)
+        return v;
```
mtrippled, line 68: For clarity, return {} or std::nullopt.

src/ripple/app/ledger/impl/TransactionMaster.cpp
```diff
@@ -54,44 +54,57 @@ TransactionMaster::fetch_from_cache(uint256 const& txnID)
...
 }
 
-std::shared_ptr<Transaction>
+std::optional<std::pair<std::shared_ptr<Transaction>, std::shared_ptr<TxMeta>>>
 TransactionMaster::fetch(uint256 const& txnID, error_code_i& ec)
 {
-    auto txn = fetch_from_cache(txnID);
+    // txn->getLedger() == 0 then txn is not validated and does not contain
+    // metadata
+    if (auto txn = mCache.fetch(txnID); txn && txn->getLedger() == 0)
```
mtrippled, line 62: There's a possible race condition here. If txn->getLedger() is not set before https://github.com/ripple/rippled/blob/develop/src/ripple/app/ledger/Ledger.cpp#L1010-L1014 then there will be no metadata. I'm not sure if that's the case, but the race condition cannot be introduced.

src/ripple/app/ledger/impl/TransactionMaster.cpp
```diff
@@ -54,44 +54,57 @@ TransactionMaster::fetch_from_cache(uint256 const& txnID)
...
     error_code_i& ec)
 {
-    using pointer = Transaction::pointer;
+    using TxPair =
+        std::pair<std::shared_ptr<Transaction>, std::shared_ptr<TxMeta>>;
 
-    auto txn = mCache.fetch(txnID);
+    // txn->getLedger() == 0 then txn is not validated and does not contain
+    // metadata
+    if (auto txn = mCache.fetch(txnID); txn && txn->getLedger() == 0)
```
mtrippled, line 91: Same race condition possibility as described above.

src/ripple/app/ledger/impl/TransactionMaster.cpp
cjcobb23: If there is a race condition here, I think it already existed in the code prior to this PR. Previously, the code called txn->getLedger() to get the containing the ledger, and then traversed said ledger to get the metadata. If txn->getLedger() was 0, then no metadata was returned. The intention of this PR is to simply change where we get the metadata from.

mtrippled: :+1:

src/ripple/app/ledger/impl/TransactionMaster.cpp
```diff
@@ -54,44 +54,57 @@ TransactionMaster::fetch_from_cache(uint256 const& txnID)
...
+        mCache.canonicalize_replace_client(txnID, txn);
 
-    return txn;
+    return std::pair{std::move(txn), std::move(txnMeta)};
 }
 
-boost::variant<Transaction::pointer, bool>
+std::optional<std::variant<
+    std::pair<std::shared_ptr<Transaction>, std::shared_ptr<TxMeta>>,
+    bool>>
```
seelabs, line 80: Wow, that's a complex return type: an optional variant of a pair and a bool?!? Yuck.

 I don't think it's worth holding up this PR, but let's make a note to refactor this in the near future. This return type needs to be simplified. One idea is to return a variant of the pair and some enum.

src/ripple/rpc/handlers/Tx.cpp
```diff
@@ -120,38 +120,45 @@ doTxHelp(RPC::Context& context, TxArgs const& args)
...
     }
 
     if (ec == rpcDB_DESERIALIZATION)
     {
         return {result, ec};
     }
-    if (!txn)
-    {
-        return {result, rpcTXN_NOT_FOUND};
-    }
```
seelabs, line 154: I wouldn't lose the checks for null txn and null meta. Right now the code is correct, but given the return type of `fetch` is likely to be refactored, and we don't know what that refactoring will be, I'd prefer to leave the checks for null.

src/ripple/app/ledger/impl/TransactionMaster.cpp
natenichols: Been thinking about it and I think the way to go here is to have a return type of 
```
std::variant<std::pair<std::shared_ptr<Transaction>, std::shared_ptr<TxMeta>>, SearchedAll>
```
given that the `optional` is representing a `SearchedAll::Unkown`. Thoughts?

src/ripple/app/misc/Transaction.h
```diff
@@ -52,6 +53,8 @@ enum TransStatus {
     INCOMPLETE = 8   // needs more signatures
 };
 
+enum class SearchedAll { no, yes, unknown };
+
```
seelabs, line 57: Now that this lives in a header file, I think this needs to be renamed or put into another namespace. `SearchedAll` is too generic a name to live in the main rippled namespace. Maybe `TxSearchedAll`?

src/ripple/app/ledger/impl/TransactionMaster.cpp
cjcobb23: I think we should use `std::monostate` here: https://en.cppreference.com/w/cpp/utility/variant/monostate

mtrippled:
:+1: LGTM

src/ripple/app/ledger/impl/TransactionMaster.cpp
```diff
@@ -54,44 +54,52 @@ TransactionMaster::fetch_from_cache(uint256 const& txnID)
...
 
-    boost::variant<Transaction::pointer, bool> v =
-        Transaction::load(txnID, mApp, range, ec);
+    if (std::holds_alternative<bool>(v))
+        return v;
 
-    if (v.which() == 0 && boost::get<pointer>(v))
-        mCache.canonicalize_replace_client(txnID, boost::get<pointer>(v));
+    auto txnPair = std::get<tuple>(v);
+    txn = std::get<0>(txnPair);
```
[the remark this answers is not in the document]
thejohnfreeman: Compare before:

```cpp
    if (std::holds_alternative<bool>(v))
        return v;
```

with after:

```cpp
    if (auto e = std::get_if<TxSearchedAll>(&v))
        return *e;
```

I prefer before and think it is cleaner. No address-of and dereference, no extra variable. `holds_alternative` is for exactly this use-case, checking which type is resident; `get_if` is a `?.`-like getter.

src/ripple/app/ledger/impl/TransactionMaster.cpp
```diff
@@ -54,44 +54,52 @@ TransactionMaster::fetch_from_cache(uint256 const& txnID)
...
 TransactionMaster::fetch(uint256 const& txnID, error_code_i& ec)
 {
-    auto txn = fetch_from_cache(txnID);
-
-    if (txn)
-        return txn;
+    auto txn = mCache.fetch(txnID);
 
-    txn = Transaction::load(txnID, mApp, ec);
+    if (txn && txn->getLedger() == 0)
```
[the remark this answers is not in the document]
thejohnfreeman: How about an `isValidated(Transaction) -> bool` helper that wraps the inverse of this expression?

src/ripple/app/misc/Transaction.h
thejohnfreeman: How about `boost::tribool`?

Or this:

```cpp
enum class TxSearched { all, some, unknown };
```

src/ripple/rpc/handlers/Tx.cpp
```diff
@@ -119,31 +117,31 @@ doTxHelp(RPC::Context& context, TxArgs const& args)
             args.ledgerRange->first, args.ledgerRange->second);
     }
 
-    std::shared_ptr<Transaction> txn;
     auto ec{rpcSUCCESS};
 
-    result.searchedAll = SearchedAll::unknown;
+    using TxPair =
+        std::pair<std::shared_ptr<Transaction>, std::shared_ptr<TxMeta>>;
```
thejohnfreeman, line 123: Might want to elevate this type alias and use it in more places.

src/ripple/app/misc/Transaction.h
seelabs: The `all, some, unknown` renaming is good. I like that. I don't think `tribool` is a good fit for this tho.

cjcobb23: I kind've think we should just remove the `unknown` option here. I think `Transaction::load` should return `std::variant<std::monostate, std::pair<std::shared_ptr<Transaction>, std::shared_ptr<TxMeta>>, TxSearchedAll>`. `TxSearchedAll` could change to a bool then, though I think it's easier to understand what the meaning of the returned value is when the enum is used (as opposed to just a bool). 

There's really no need for `TxSearchedAll::unknown`; it just means the transaction wasn't found, and we didn't check the range of ledgers (because the caller didn't ask to check the range). The caller knows that they didn't ask for the range of ledgers to be checked, so `TxSearchedAll::unknown` doesn't actually provide any additional info to the caller. I think simply returning an empty variant (via `std::monostate`) is the most natural thing here. Or just wrapping the whole return type in an `std::optional`.

Another option would be to break this up into two separate functions. Right now `Transaction::load` is doing two things: looking for the transaction, and also reporting info about the range of ledgers in the database, in the event that the transaction wasn't found. We wouldn't have to hem and haw about return types if this was broken up into two separate functions.

seelabs: @cjcobb23 Just commenting on the `monostate` solution. I think I prefer the enum over that. The nice thing about the enum is there's one place to look when we don't get the result we want. With the `optional` or the `monostate` there's two places: is the `std::monostate` the active part of the `variant` - that means one kind of "we didn't get what we want' or if the `bool` part of the `variant` is active - that means another kind of "we didn't get what we want". I think it's cleaner to put that information in one place.

src/ripple/app/ledger/impl/TransactionMaster.cpp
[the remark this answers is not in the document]
natenichols: Moving this to a `Transaction::isValidated()` that compares `getLedger != 0`

src/ripple/rpc/handlers/Tx.cpp
natenichols: Originally we had this alias in `Transaction` namespace, and would use `Transaction::TxPair`. We chose to write out types in the to clearly describe what return types are. 

If you think it's better to use a type alias we can discuss.

src/ripple/app/misc/Transaction.h
cjcobb23: @seelabs that's a fair point about having all of the "not found" results as the same element of the variant.

I think ultimately I would prefer to split this into two functions. One to get the transaction, and one to check the ledger range. Both would be called from the handler, and the range check would only be called if the client asked for that information (and the transaction wasn't found).

seelabs: @cjcobb23 Splitting into two functions sounds attractive, I like that. We'd have to look to make sure that ledger range can't change between the not finding the transaction and checking the range we'd give faulty information (or make sure we hold the appropriate locks).

While I like the idea, what do you think about punting on it for now and tackling it in another PR? I'd like to get this one in and @natenichols's solution is pretty good as-is.

cjcobb23: @seelabs @natenichols I'm good with the code as-is as well

src/ripple/rpc/handlers/Tx.cpp
cjcobb23: I don't like this type alias, because there's `Transaction` and there's `STTx` which are both used as datatypes for transactions (as well as `TxMeta` and `STObject` for the metadata). So when I see `Transaction::TxPair`, it's not immediately clear to me what the type of the pair elements are.

seelabs (approved):
Still :+1: from me

cjcobb23 (approved):

scottschurr:
Ticket_test.cpp no longer compiles because of the changed return type of `Transaction::load()`.  Here's a patch to fix the unit test so it conforms to the new interface: 

https://github.com/scottschurr/rippled/commit/863946a26396c0f574db9f9adff72e514918ae53

Could I get a quick set of eyes for a thumbs up (or down) on the patch?  Thanks.

PR #3565, opened 2020-08-01
Remove legacy support for endpoint dissemination:

nbougalis:
Support for IPv6 messages was added with commit 08382d866b116e3ec9d50193bcf45caea8ca09a2
and version 1.1.0. No peer presently connected to the network in a useful capacity fails
to understand v2 messages.

This commit removes the code that generates and processes v1 messages and deletes legacy
messages from the protocol buffer definition file.

### DIFF
--- a/src/ripple/overlay/impl/PeerImp.cpp
+++ b/src/ripple/overlay/impl/PeerImp.cpp
@@ -1376,13 +1376,11 @@ PeerImp::onMessage(std::shared_ptr<protocol::TMEndpoints> const& m)
 
     for (auto const& tm : m->endpoints_v2())
     {
-        auto result =
-            beast::IP::Endpoint::from_string_checked(tm.endpoint());
+        auto result = beast::IP::Endpoint::from_string_checked(tm.endpoint());
         if (!result)
         {
-            JLOG(p_journal_.error())
-                << "failed to parse incoming endpoint: {" << tm.endpoint()
-                << "}";
+            JLOG(p_journal_.error()) << "failed to parse incoming endpoint: {"
+                                     << tm.endpoint() << "}";
             continue;
         }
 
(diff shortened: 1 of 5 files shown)


src/ripple/overlay/impl/PeerImp.h
```diff
@@ -656,24 +649,13 @@ void
 PeerImp::sendEndpoints(FwdIt first, FwdIt last)
 {
     protocol::TMEndpoints tm;
-    for (; first != last; ++first)
+
+    while (first != last)
```
gregtatcam, line 653: Is there any advantage to using `while` vs `for` in this case?

nbougalis: Not really.

src/ripple/overlay/impl/PeerImp.h
```diff
@@ -656,24 +656,13 @@ void
...
-        tme.mutable_ipv4()->set_ipv4port(ep.address.port());
-        tme.set_hops(ep.hops);
-
-        // add v2 endpoints (strings)
         auto& tme2(*tm.add_endpoints_v2());
-        tme2.set_endpoint(ep.address.to_string());
-        tme2.set_hops(ep.hops);
+        tme2.set_endpoint(first->address.to_string());
+        tme2.set_hops(first->hops);
+        first++;
```
miguelportilla, line 658: nit: Can be ++first

src/ripple/overlay/impl/PeerImp.cpp
```diff
@@ -1366,84 +1366,37 @@ PeerImp::onMessage(std::shared_ptr<protocol::TMPeerShardInfo> const& m)
 void
 PeerImp::onMessage(std::shared_ptr<protocol::TMEndpoints> const& m)
 {
-    if (sanity_.load() != Sanity::sane)
-    {
-        // Don't allow endpoints from peer not known sane
+    // Don't allow endpoints from peers that are not known sane or are
+    // not using a version of the message that we support:
+    if (sanity_.load() != Sanity::sane || m->version() != 2)
```
miguelportilla, line 1371: Nit: Can be `if (m->version() != 2 || sanity_.load() != Sanity::sane)`

PR #3566, opened 2020-08-03
NegativeUNL SField Changes

pwang200:
Several declarations in SField are changed so that the symbolic names and the text representations are consistent. 
-- the SField declarations are sfDisabledValidator, sfValidatorToDisable, and sfValidatorToReEnable
-- their corresponding gRPC declarations are also changed
-- the related function names are also changed
-- changed negativeUnl to negativeUNL too

Tested with unit test and local network tests.

### DIFF
--- a/src/ripple/app/ledger/Ledger.cpp
+++ b/src/ripple/app/ledger/Ledger.cpp
@@ -604,9 +604,9 @@ Ledger::negativeUNL() const
 {
     hash_set<PublicKey> negUnl;
     if (auto sle = read(keylet::negativeUNL());
-        sle && sle->isFieldPresent(sfNegativeUNL))
+        sle && sle->isFieldPresent(sfDisabledValidators))
     {
-        auto const& nUnlData = sle->getFieldArray(sfNegativeUNL);
+        auto const& nUnlData = sle->getFieldArray(sfDisabledValidators);
         for (auto const& n : nUnlData)
         {
             if (n.isFieldPresent(sfPublicKey))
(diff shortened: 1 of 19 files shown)


src/ripple/proto/org/xrpl/rpc/v1/ledger_objects.proto
```diff
@@ -333,11 +333,11 @@ message SignerList
...
+message NegativeUNL
 {
-    repeated NegativeUnlEntry negative_unl_entries = 1;
+    repeated DisabledValidator disabled_validators = 1;
 
-    NegativeUnlToDisable validator_to_disable = 2;
+    ValidatorToDisable validator_to_disable = 2;
 
-    NegativeUnlToReEnable validator_to_re_enable = 3;
+    ValidatorToReEnable validator_to_re_enable = 3;
```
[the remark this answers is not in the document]
pwang200: Added, also in `void convert(org::xrpl::rpc::v1::NegativeUNL& to, STObject const& from)`.

pwang200:
Changed sfNegativeUNL to sfDisabledValidators

src/ripple/app/ledger/Ledger.cpp
```diff
@@ -600,13 +600,13 @@ Ledger::peek(Keylet const& k) const
 }
 
 hash_set<PublicKey>
-Ledger::negativeUnl() const
+Ledger::negativeUNL() const
```
scottschurr, line 603: This could be addressed later, but I wonder whether this method should be named `disabledValidators()` now?

pwang200: I thought about it, but did not change it. My reasoning was that the same thing (the `negativeUNL`) was accessed here and used somewhere else. If I change here, I probably should change other places as well, which I felt confusing. I see the `disabledValidator` more as a type and `disabledValidators` as its plural (or container). And I see negativeUNL as a usage of the container. Iirc, the skip list has type hashes, but it is protocol meaning is the skip list.

scottschurr: That's fine.  It's one that we can think about.  I've been thinking of the `NegativeUNL` as a heterogeneous container of three elements:

- DisabledValidators,
- ValidatorToDisable, and
- ValidatorToReEnable

This specific entry point returns only DisabledValidators.  That's why I think the rename is appropriate.  But this name doesn't leak out into the API.  So we can change it, or not, later.

PR #3573, opened 2020-08-07
[TRIVIAL] Revert "Remove CryptoConditionsSuite stub amendment:"

seelabs:
This reverts commit d3798f62903df1bd24a5eb00a221963bf28f45f0.

### DIFF
--- a/src/ripple/app/tx/impl/Escrow.cpp
+++ b/src/ripple/app/tx/impl/Escrow.cpp
@@ -147,7 +147,10 @@ EscrowCreate::preflight(PreflightContext const& ctx)
             return temMALFORMED;
         }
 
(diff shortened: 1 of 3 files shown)


src/ripple/app/tx/impl/Escrow.cpp
```diff
@@ -147,7 +147,10 @@ EscrowCreate::preflight(PreflightContext const& ctx)
             return temMALFORMED;
         }
 
-        if (condition->type != Type::preimageSha256)
+        // Conditions other than PrefixSha256 require the
```
scottschurr, line 150: We could consider not reverting this part of the commit.  Then if someone enables CryptoConditionsSuite it will have no effect on the code.  But I don't feel strongly about it.

PR #3575, opened 2020-08-09
Update Base58 codec and remove Bitcoin support:

cryptobrad:
Use C++17 constant expressions to calculate the inverse
alphabet map at compile time instead of at runtime.

Remove support for encoding & decoding tokens using the
Bitcoin alphabet.

### DIFF
--- a/src/ripple/protocol/impl/tokens.cpp
+++ b/src/ripple/protocol/impl/tokens.cpp
@@ -30,13 +30,22 @@
 
 namespace ripple {
 
-static char rippleAlphabet[] =
+static constexpr char const* const xrplAlphabetForward =
     "rpshnaf39wBUDNEGHJKLM4PQRST7VWXYZ2bcdeCg65jkm8oFqi1tuvAxyz";
 
-static char bitcoinAlphabet[] =
-    "123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz";
+static constexpr std::array<int, 256> const xrplAlphabetReverse = []() constexpr
+{
+    std::array<int, 256> map{};
 
-//------------------------------------------------------------------------------
+    for (auto& m : map)
+        m = -1;
+
+    for (int i = 0, j = 0; xrplAlphabetForward[i] != 0; ++i)
+        map[static_cast<unsigned char>(xrplAlphabetForward[i])] = j++;
+
+    return map;
+}
+();
 
 template <class Hasher>
 static typename Hasher::result_type
@@ -66,7 +75,7 @@ digest2(Args const&... args)
     return digest<Hasher>(digest<Hasher>(args...));
 }
 
-/*  Calculate a 4-byte checksum of the data
+/** Calculate a 4-byte checksum of the data
 
     The checksum is calculated as the first 4 bytes
     of the SHA256 digest of the message. This is added
@@ -75,32 +84,28 @@ digest2(Args const&... args)
 
     @note This checksum algorithm is part of the client API
 */
-void
+static void
 checksum(void* out, void const* message, std::size_t size)
 {
     auto const h = digest2<sha256_hasher>(message, size);
     std::memcpy(out, h.data(), 4);
 }
 
-//------------------------------------------------------------------------------
+namespace detail {
 
-// Code from Bitcoin: https://github.com/bitcoin/bitcoin
-// Copyright (c) 2014 The Bitcoin Core developers
-// Distributed under the MIT software license, see the accompanying
-// file COPYING or http://www.opensource.org/licenses/mit-license.php.
-//
-// Modified from the original
-//
-// WARNING Do not call this directly, use
-//         encodeBase58Token instead since it
-//         calculates the size of buffer needed.
+/* The base58 encoding & decoding routines in this namespace are taken from
+ * Bitcoin but have been modified from the original.
+ *
+ * Copyright (c) 2014 The Bitcoin Core developers
+ * Distributed under the MIT software license, see the accompanying
+ * file COPYING or http://www.opensource.org/licenses/mit-license.php.
+ */
 static std::string
 encodeBase58(
     void const* message,
     std::size_t size,
     void* temp,
-    std::size_t temp_size,
-    char const* const alphabet)
+    std::size_t temp_size)
 {
     auto pbegin = reinterpret_cast<unsigned char const*>(message);
     auto const pend = pbegin + size;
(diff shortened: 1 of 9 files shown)


src/ripple/protocol/impl/tokens.cpp
```diff
@@ -30,13 +30,22 @@
 
 namespace ripple {
 
-static char rippleAlphabet[] =
+static constexpr char const* const xrplAlphabetForward =
```
MarkusTeufelberger, line 33:
```suggestion
static constexpr char const* const rippleAlphabetForward =
```

MarkusTeufelberger: Also in other places...

cryptobrad: Can you explain what you mean here? I am not understanding. You think it should use ripple? But the name 'ripple' is belong to the company and not the software.

MarkusTeufelberger: The software is called `rippled`, the company is called "Ripple Labs" and it just holds the copyright for the name "Ripple" (the project named "Ripple" is much older than that). Heck, even that folder is `src/ripple/...`. Previous code used `rippleAlphabet`, changing it to `xrplAlphabet` just puts the ISO-4217 shorthand of a cryptocurrency in the name for no good reason.

cryptobrad: Aha
Yes I understand you now and I prefer the name ripple also but the company is using that now and so to continue use of ripple seems causes a lot of confusion.
Maybe the company can answer on what this should be because the name get used again elsewhere.

MarkusTeufelberger: The company is actually still called "Ripple Labs Inc", it just uses the name "Ripple" in its branding. See for example https://ripple.com/terms-of-use/ ("These Terms of Use (“Terms”) apply to your access to, and use of, the websites of **Ripple Labs Inc.** and its subsidiaries and affiliated companies (“Ripple,” “we” or “us”), including www.ripple.com (collectively, the “Site”)."), in their privacy policy you also see the full company name (something you should maybe be aware of, if you are impersonating their CEO... ;-) ).

I guess someone working there will eventually review and resolve this.

src/test/rpc/AccountCurrencies_test.cpp
```diff
@@ -69,20 +69,6 @@ class AccountCurrencies_test : public beast::unit_test::suite
...
-        {  // strict mode, using properly formatted bitcoin token
-            Json::Value params;
-            params[jss::account] = base58EncodeTokenBitcoin(
-                TokenType::AccountID, alice.id().data(), alice.id().size());
-            params[jss::strict] = true;
-            auto const result = env.rpc(
-                "json",
-                "account_currencies",
-                boost::lexical_cast<std::string>(params))[jss::result];
-            BEAST_EXPECT(result[jss::error] == "actBitcoin");
```
MarkusTeufelberger, line 81: Probably the `actBitcoin` Error can be removed too then.

src/ripple/protocol/impl/tokens.cpp
nbougalis: I think we all like the name 'ripple' for this wonderful, collaborative project and feel an attachment to it. And if it were up to me I'd have kept it for the project and this could stay `rippleAlphabet`. But, feelings aside, there's reality. And, as has been pointed out, the naming can cause confusion.

So, since I work "there", let me play Solomon and split the proverbial baby so that we can all move forward: how about simply calling these variables `alphabet` and `reverseAlphabet` and avoiding this whole situation?

Rationale: your car doesn't use "FordNuts" or "FordBolts". It just uses "nuts" and "bolts".

cryptobrad: I changed variable names to 'alphabetForward' and 'alphabetReverse' as you have asked

Maybe Ripple can see the problem the name 'Xrp Ledger' has and can allow the name 'ripple' to be used again in future?

src/ripple/protocol/impl/tokens.cpp
```diff
@@ -30,13 +30,22 @@
 
 namespace ripple {
 
-static char rippleAlphabet[] =
+static constexpr char const* const alphabetForward =
```
scottschurr, line 33: Not something I feel strongly about, but the `const` that follows `*` is redundant since `alphabetForward` is `constexpr`.  So, if you want, you can omit the second `const`.

src/ripple/protocol/impl/tokens.cpp
```diff
@@ -30,13 +30,22 @@
...
     "rpshnaf39wBUDNEGHJKLM4PQRST7VWXYZ2bcdeCg65jkm8oFqi1tuvAxyz";
 
-static char bitcoinAlphabet[] =
-    "123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz";
+static constexpr std::array<int, 256> const alphabetReverse = []() constexpr
```
scottschurr, line 36: Another point I don't feel strongly about, but the trailing `constexpr` is unnecessary.  As of C++17 any lambda that can be `constexpr` automatically is.  To quote from cppreference:

> constexpr: explicitly specifies that the function call operator is a constexpr function. When this specifier is not present, the function call operator will be constexpr anyway, if it happens to satisfy all constexpr function requirements

https://en.cppreference.com/w/cpp/language/lambda

src/ripple/protocol/impl/tokens.cpp
```diff
@@ -246,16 +198,38 @@ decodeBase58(std::string const& s, InverseArray const& inv)
...
-    and removed from the returned result.
-*/
-template <class InverseArray>
-static std::string
-decodeBase58Token(std::string const& s, TokenType type, InverseArray const& inv)
+std::string
+encodeBase58Token(TokenType type, void const* token, std::size_t size)
 {
-    std::string const ret = decodeBase58(s, inv);
+    assert(size != 0);
```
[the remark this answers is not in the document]
cryptobrad: That would change output of functoin I think and didn't wnat to change behaviour that but having the assert makes sense to me.  I can do this if the change is not the problem.

src/ripple/protocol/impl/tokens.cpp
cryptobrad: OK that looks good

src/ripple/protocol/impl/tokens.cpp
cryptobrad: Yes i didn't know that change! Very nice.

cryptobrad:
Added the fold message and responded to review comments.  ccan you reopen please?

nbougalis:
I don't think we can reopen because you deleted your branch and now GitHub is confused. Please feel free to open a new PR for this! Thanks.

PR #3578, opened 2020-08-11
Allow shard downloads via HTTP or HTTPS

undertome:
This change allows the `HTTPDownloader` (formerly `SSLHTTPDownloader`) to download files via HTTP or HTTPS. When downloading shards, the `ShardArchiveHandler` determines which protocol to use based on the URL scheme and instructs the `HTTPDownloader` to download the shard with or without SSL.

### DIFF
--- a/src/ripple/net/HTTPDownloader.h
+++ b/src/ripple/net/HTTPDownloader.h
@@ -17,12 +17,12 @@
 */
 //==============================================================================
 
-#ifndef RIPPLE_NET_SSLHTTPDOWNLOADER_H_INCLUDED
-#define RIPPLE_NET_SSLHTTPDOWNLOADER_H_INCLUDED
+#ifndef RIPPLE_NET_HTTPDOWNLOADER_H_INCLUDED
+#define RIPPLE_NET_HTTPDOWNLOADER_H_INCLUDED
 
 #include <ripple/basics/Log.h>
 #include <ripple/core/Config.h>
-#include <ripple/net/HTTPClientSSLContext.h>
+#include <ripple/net/HTTPStream.h>
 
 #include <boost/asio/connect.hpp>
 #include <boost/asio/io_service.hpp>
@@ -39,14 +39,14 @@
 
 namespace ripple {
 
-/** Provides an asynchronous HTTPS file downloader
+/** Provides an asynchronous HTTP[S] file downloader
  */
-class SSLHTTPDownloader
+class HTTPDownloader
 {
 public:
     using error_code = boost::system::error_code;
 
-    SSLHTTPDownloader(
+    HTTPDownloader(
         boost::asio::io_service& io_service,
         beast::Journal j,
         Config const& config);
@@ -58,12 +58,13 @@ public:
         std::string const& target,
         int version,
         boost::filesystem::path const& dstPath,
-        std::function<void(boost::filesystem::path)> complete);
+        std::function<void(boost::filesystem::path)> complete,
+        bool ssl = true);
 
     void
     onStop();
 
-    virtual ~SSLHTTPDownloader() = default;
+    virtual ~HTTPDownloader() = default;
 
 protected:
     using parser = boost::beast::http::basic_parser<false>;
@@ -79,10 +80,9 @@ protected:
         std::shared_ptr<parser> parser);
 
 private:
-    HTTPClientSSLContext ssl_ctx_;
+    Config const& config_;
     boost::asio::io_service::strand strand_;
-    boost::optional<boost::asio::ssl::stream<boost::asio::ip::tcp::socket>>
-        stream_;
+    std::unique_ptr<HTTPStream> stream_;
     boost::beast::flat_buffer read_buf_;
     std::atomic<bool> cancelDownloads_;
 
@@ -99,6 +99,7 @@ private:
         int version,
         boost::filesystem::path dstPath,
         std::function<void(boost::filesystem::path)> complete,
+        bool ssl,
         boost::asio::yield_context yield);
 
     virtual std::shared_ptr<parser>
--- a/src/ripple/net/HTTPStream.h
+++ b/src/ripple/net/HTTPStream.h
@@ -1,0 +1,158 @@
+//------------------------------------------------------------------------------
+/*
+    This file is part of rippled: https://github.com/ripple/rippled
+    Copyright (c) 2020 Ripple Labs Inc.
+
+    Permission to use, copy, modify, and/or distribute this software for any
+    purpose  with  or without fee is hereby granted, provided that the above
+    copyright notice and this permission notice appear in all copies.
+
+    THE  SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+    WITH  REGARD  TO  THIS  SOFTWARE  INCLUDING  ALL  IMPLIED  WARRANTIES  OF
+    MERCHANTABILITY  AND  FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+    ANY  SPECIAL ,  DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+    WHATSOEVER  RESULTING  FROM  LOSS  OF USE, DATA OR PROFITS, WHETHER IN AN
+    ACTION  OF  CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+    OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+//==============================================================================
+
+#ifndef RIPPLE_NET_HTTPSTREAM_H_INCLUDED
+#define RIPPLE_NET_HTTPSTREAM_H_INCLUDED
+
+#include <ripple/core/Config.h>
+#include <ripple/net/HTTPClientSSLContext.h>
+
+#include <boost/asio/spawn.hpp>
+#include <boost/asio/strand.hpp>
+#include <boost/beast/core.hpp>
+#include <boost/beast/http.hpp>
+
+namespace ripple {
+
+class HTTPStream
+{
+public:
+    using request = boost::beast::http::request<boost::beast::http::empty_body>;
+    using parser = boost::beast::http::basic_parser<false>;
+
+    virtual ~HTTPStream() = default;
+
+    template <class T>
+    static std::unique_ptr<HTTPStream>
+    makeUnique(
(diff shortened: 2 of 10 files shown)


src/ripple/net/HTTPStream.h
```diff
@@ -0,0 +1,148 @@
...
+class HTTPStream
+{
+public:
+    using request = boost::beast::http::request<boost::beast::http::empty_body>;
+    using parser = boost::beast::http::basic_parser<false>;
+
+    virtual ~HTTPStream() = default;
+
+    virtual boost::asio::ip::tcp::socket&
+    getStream() = 0;
```
scottschurr, line 42: Consider decorating `getStream` with the `[[nodiscard]]` attribute.

src/ripple/net/HTTPStream.h
```diff
@@ -0,0 +1,148 @@
...
+
+    virtual bool
+    connect(
```
scottschurr, line 45: Consider decorating `connect` with the `[[nodiscard]]` attribute.

src/ripple/net/HTTPStream.h
```diff
@@ -0,0 +1,148 @@
...
+{
+public:
+    SSLStream(
+        Config const& config,
+        boost::asio::io_service::strand& strand,
+        beast::Journal j);
+
+    virtual ~SSLStream() = default;
+
+    boost::asio::ip::tcp::socket&
```
scottschurr, line 76: It _seems_ like `SSLStream` and `RawStream` provide implementations for `HTTPStream`.  And you might want for people to never directly call methods on, for example, `SSLStream`.  You might instead want them to only call through the base class.  If that's the case, you can consider making all of these `override` methods in `SSLStream` and `RawStream` `private`.  That will force access to alway be through the base class.

It's just an option for you to consider.  Since the pattern is a little unusual you could legitimately choose not to use it.

src/ripple/rpc/impl/ShardArchiveHandler.cpp
```diff
@@ -375,13 +375,17 @@ ShardArchiveHandler::next(std::lock_guard<std::mutex> const& l)
     // sleeps.
     auto const& url{archives_.begin()->second};
     auto wrapper = jobCounter_.wrap([this, url, dstDir](Job&) {
+        auto const ssl = url.scheme == "https";
```
scottschurr, line 378: In my opinion the `auto const ssl` only obscures what you're up to.  My preference would be `bool const ssl`.  But do what you want...

src/ripple/net/ShardDownloader.md
```diff
@@ -2,7 +2,7 @@
 
 ## Overview
 
-This document describes mechanics of the `SSLHTTPDownloader`, a class that performs the task of downloading shards from remote web servers via
+This document describes mechanics of the `HTTPDownloader`, a class that performs the task of downloading shards from remote web servers via
 SSL HTTP. The downloader utilizes a strand (`boost::asio::io_service::strand`) to ensure that downloads are never executed concurrently. Hence, if a download is in progress when another download is initiated, the second download will be queued and invoked only when the first download is completed.
```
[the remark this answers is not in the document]
undertome: Nope - removed

src/ripple/net/HTTPDownloader.h
```diff
@@ -17,12 +17,13 @@
...
 //==============================================================================
 
-#ifndef RIPPLE_NET_SSLHTTPDOWNLOADER_H_INCLUDED
-#define RIPPLE_NET_SSLHTTPDOWNLOADER_H_INCLUDED
+#ifndef RIPPLE_NET_HTTPDOWNLOADER_H_INCLUDED
+#define RIPPLE_NET_HTTPDOWNLOADER_H_INCLUDED
 
 #include <ripple/basics/Log.h>
 #include <ripple/core/Config.h>
 #include <ripple/net/HTTPClientSSLContext.h>
```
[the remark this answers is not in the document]
undertome: Done

src/ripple/net/impl/HTTPDownloader.cpp
```diff
@@ -118,17 +122,18 @@ SSLHTTPDownloader::do_session(
...
             ec.assign(0, ec.category());
         if (ec)
         {
             // Most web servers don't bother with performing
             // the SSL shutdown handshake, for speed.
-            JLOG(j_.trace()) << "async_shutdown: " << ec.message();
+            JLOG(j_.trace()) << "shutdown: " << ec.message();
         }
+
         // The socket cannot be reused
```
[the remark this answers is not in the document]
undertome: Done

src/ripple/net/impl/HTTPDownloader.cpp
```diff
@@ -161,36 +166,13 @@ SSLHTTPDownloader::do_session(
     // connection:
     std::uint64_t const rangeStart = size(p);
 
-    ip::tcp::resolver resolver{strand_.context()};
-    auto const results = resolver.async_resolve(host, port, yield[ec]);
-    if (ec)
-        return failAndExit("async_resolve", p);
+    stream_.reset(
+        ssl ? static_cast<HTTPStream*>(new SSLStream(config_, strand_, j_))
+            : static_cast<HTTPStream*>(new RawStream(config_, strand_, j_)));
```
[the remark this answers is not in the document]
undertome: Done

src/ripple/net/HTTPStream.h
undertome: Done

src/ripple/net/HTTPStream.h
undertome: Done

src/ripple/net/HTTPStream.h
undertome: For the moment these classes are called only via `HTTPStream`, but I don't want to force access to them through that interface.

src/ripple/net/impl/HTTPStream.cpp
```diff
@@ -0,0 +1,207 @@
...
+    boost::asio::io_service::strand& strand,
+    beast::Journal j)
+{
+    return std::make_unique<RawStream>(config, strand, j);
+}
+
+boost::asio::ip::tcp::socket&
+RawStream::getStream()
+{
+    return *stream_;
```
ximinez, line 142: Same here as the other `getStream`, please add some kind of check here that `stream_` is defined. At minimum, `assert(stream_)`.

src/ripple/net/impl/HTTPStream.cpp
```diff
@@ -0,0 +1,207 @@
...
+
+    boost::system::error_code ec;
+
+    auto fail = [&errorOut](std::string const& errorIn) {
+        errorOut = errorIn;
+        return false;
+    };
+
+    ip::tcp::resolver resolver{strand_.context()};
+    auto const results = resolver.async_resolve(host, port, yield[ec]);
```
ximinez, line 163: As above, `results` is a very generic name, and since it's coming in the middle of the function, and isn't the result that's actually returned by the function, I think another name would be better. How about "resolved", "ips" (or something similarly descriptive), or "connectTargets".

src/ripple/rpc/impl/ShardArchiveHandler.cpp
ximinez, line 378: I sometimes get confused when I see a bunch of `=` in a single command. Could you add parenthesis so that it's clearer what's going on? `auto const ssl = (url.scheme == "https");`. This doesn't affect functionality, just readability.

src/ripple/net/HTTPStream.h
```diff
@@ -0,0 +1,160 @@
...
+        boost::asio::io_service::strand& strand,
+        beast::Journal j);
+
+    virtual ~SSLStream() = default;
+
+    static std::unique_ptr<HTTPStream>
+    getUnique(
+        Config const& config,
+        boost::asio::io_service::strand& strand,
+        beast::Journal j);
```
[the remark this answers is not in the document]
undertome: Done

src/ripple/net/ShardDownloader.md
```diff
@@ -18,13 +22,15 @@ This document was created to document the changes introduced by this change.
...
 
-   This is a generic class designed for serially executing downloads via HTTP SSL.
+   This is a generic class designed for serially executing downloads via HTTP.
 
 - `ShardArchiveHandler`
 
-   This class uses the `SSLHTTPDownloader` to fetch shards from remote web servers. Additionally, the archive handler performs sanity checks on the downloaded files and imports the validated files into the local shard store.
+   This class uses the `HTTPDownloader` to fetch shards from remote web servers.
+   Additionally, the archive handler performs sanity checks on the downloaded
+   files and imports the validated files into the local shard store.
```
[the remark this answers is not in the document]
undertome: Done

src/ripple/net/impl/HTTPStream.cpp
```diff
@@ -0,0 +1,207 @@
...
+    boost::asio::io_service::strand& strand,
+    beast::Journal j)
+{
+    return std::make_unique<SSLStream>(config, strand, j);
+}
+
+boost::asio::ip::tcp::socket&
+SSLStream::getStream()
+{
+    return stream_->next_layer();
```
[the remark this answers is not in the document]
undertome: Done

src/ripple/net/impl/HTTPStream.cpp
```diff
@@ -0,0 +1,207 @@
...
+
+    boost::system::error_code ec;
+
+    auto fail = [&errorOut](std::string const& errorIn) {
+        errorOut = errorIn;
+        return false;
+    };
+
+    ip::tcp::resolver resolver{strand_.context()};
+    auto const results = resolver.async_resolve(host, port, yield[ec]);
```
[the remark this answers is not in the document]
undertome: Done

src/ripple/net/impl/HTTPStream.cpp
undertome: Done

src/ripple/net/impl/HTTPStream.cpp
undertome: Done

src/ripple/rpc/impl/ShardArchiveHandler.cpp
undertome: Done

src/ripple/net/impl/HTTPStream.cpp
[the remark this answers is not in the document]
ximinez: `endpoints` was a much better choice than my suggestions. :)

ximinez (approved):
Looks good!

PR #3579, opened 2020-08-11
Simplify the SetSignerList transactor

cryptobrad:

PR #3580, opened 2020-08-12
Simplify the 'SetSignerList' transactor

cryptobrad:
This is new version of #3579 which was closed by mistake

### DIFF
--- a/src/ripple/app/tx/impl/SignerEntries.h
+++ b/src/ripple/app/tx/impl/SignerEntries.h
@@ -25,49 +25,41 @@
 #include <ripple/protocol/STTx.h>           // STTx::maxMultiSigners
 #include <ripple/protocol/TER.h>            // temMALFORMED
 #include <ripple/protocol/UintTypes.h>      // AccountID
+#include <optional>
 
 namespace ripple {
 
 // Forward declarations
 class STObject;
 
-// Support for SignerEntries that is needed by a few Transactors
-class SignerEntries
+/** An individual entry in a signer list */
+struct SignerEntry
 {
-public:
-    explicit SignerEntries() = default;
+    AccountID account;
+    std::uint16_t weight;
 
-    struct SignerEntry
+    SignerEntry(AccountID const& inAccount, std::uint16_t inWeight)
+        : account(inAccount), weight(inWeight)
     {
-        AccountID account;
-        std::uint16_t weight;
+    }
 
-        SignerEntry(AccountID const& inAccount, std::uint16_t inWeight)
-            : account(inAccount), weight(inWeight)
-        {
-        }
+    // For sorting to look for duplicate accounts
+    friend bool
+    operator<(SignerEntry const& lhs, SignerEntry const& rhs)
+    {
+        return lhs.account < rhs.account;
+    }
 
-        // For sorting to look for duplicate accounts
-        friend bool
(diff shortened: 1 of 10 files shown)


src/ripple/app/tx/impl/SignerEntries.h
```diff
@@ -25,50 +25,42 @@
...
 
 namespace ripple {
 
 // Forward declarations
 class STObject;
 
-// Support for SignerEntries that is needed by a few Transactors
-class SignerEntries
+/** An individual entry in a signer list */
+struct SignerEntry
```
nbougalis, line 36: Having this structure makes code self-documenting. I'm curious if anyone thinks that `std::pair<AccountID, std::uint16_t>` would be better; it gives us the two comparison operators we want for "free" (not that the cost we're paying now is a lot).

seelabs:
Thanks for the contribution!

Before I do an in-depth review, can I ask you to make three changes:

1) It currently doesn't pass unittests. This appears to be caused by the assumption that when the `featureSimplifiedSetSignerList` is active then the `featureMultiSignReserve` is also active (a reasonable assumption, as multiSignReserve has been active for over a year). However, the `Checks` unit tests does not make this assumption. We either need to change the `Checks` tests or make it explicit that both features need to be turned on. (I'll provide a patch for this).
2) We started enforcing code formatting with clang-format. The patch needs to be reformatted wit clang-format (I'll provide a patch for this).
3) This patch is on an amendment. This usually means this is transaction breaking. If this is transaction breaking, can you document how it differs from the existing behaviour. If it's not transaction breaking we probably shouldn't put it on an amendment (tho there are some very rare cases where we'll do that anyway to mitigate risk).

Patch for your review: https://github.com/seelabs/rippled/commit/d93d7414bfd54227aa833fdc031227ca74641c9e (note one minor edit of a comment slipped in by mistake)

intelliot:
@cryptobrad Could you edit the PR description to explain this change? [Here is a pull request template](https://raw.githubusercontent.com/ripple/rippled/0bc4906451e4cd0a19a3bea84e2a278da7af747e/.github/pull_request_template.md) that you can use. You can copy/paste it as a starting point, and then fill in the relevant sections.

carlhua:
@cryptobrad have you gotten a chance to look over the comments? What are your thoguths? Would you like to help addressing the comments? Thanks.

nbougalis:
I suspect the amendment is needed, since the code returns subtly different errors in some cases: see https://github.com/ripple/rippled/pull/3580/files#diff-14dfe58ed326a3b9dcc4d9ff19f9700cf469ecd9fc5d268dffd20d5440b227e3R186

nbougalis:
I suspect the amendment is needed, since the code returns subtly different errors in some cases: see https://github.com/ripple/rippled/pull/3580/files#diff-14dfe58ed326a3b9dcc4d9ff19f9700cf469ecd9fc5d268dffd20d5440b227e3R186

carlhua:
@cryptobrad pinging you once more to see if we can have some of the comments considered here. Otherwise, I will go ahead and close this PR - we can always open another one in the future.

carlhua:
Stale PR - Closing for now.

cryptobrad:
This is second time you close my PR and it is one that improves the code and cleans it up. Yes I didn't answr the questoin. I used the amendment to be safe juts in case we missed something that was a transaction change.

PR #3582, opened 2020-08-13
Currently, the ledger command only takes int for ledger_index - update the logic to allow string as well.

### LINKED ISSUE #3533: Update ledger command to accept string as ledger_index
Currently, the ledger command only takes int for ledger_index - update the logic to allow string as well.

gregtatcam:
Assume ledger_index is quoted int if it is not "current", "closed", or "validated".

FIXES #3533

### DIFF
--- a/src/ripple/json/json_value.h
+++ b/src/ripple/json/json_value.h
@@ -136,7 +136,7 @@ operator!=(StaticString x, std::string const& y)
  * The sequence of an #arrayValue will be automatically resize and initialized
  * with #nullValue. resize() can be used to enlarge or truncate an #arrayValue.
  *
- * The get() methods can be used to obtanis default value in the case the
+ * The get() methods can be used to obtain a default value in the case the
  * required element does not exist.
  *
  * It is possible to iterate over the list of a #objectValue values using
--- a/src/ripple/rpc/impl/RPCHelpers.cpp
+++ b/src/ripple/rpc/impl/RPCHelpers.cpp
@@ -228,31 +228,23 @@ ledgerFromRequest(T& ledger, JsonContext& context)
             return {rpcINVALID_PARAMS, "ledgerHashMalformed"};
         return getLedger(ledger, ledgerHash, context);
     }
-    else if (indexValue.isNumeric())
-    {
-        return getLedger(ledger, indexValue.asInt(), context);
-    }
-    else
-    {
-        auto const index = indexValue.asString();
-        if (index == "validated")
-        {
-            return getLedger(ledger, LedgerShortcut::VALIDATED, context);
-        }
-        else
-        {
-            if (index.empty() || index == "current")
-                return getLedger(ledger, LedgerShortcut::CURRENT, context);
-            else if (index == "closed")
-                return getLedger(ledger, LedgerShortcut::CLOSED, context);
-            else
-            {
-                return {rpcINVALID_PARAMS, "ledgerIndexMalformed"};
-            }
-        }
-    }
 
-    return Status::OK;
+    auto const index = indexValue.asString();
+
+    if (index.empty() || index == "current")
+        return getLedger(ledger, LedgerShortcut::CURRENT, context);
+
+    if (index == "validated")
+        return getLedger(ledger, LedgerShortcut::VALIDATED, context);
+
+    if (index == "closed")
+        return getLedger(ledger, LedgerShortcut::CLOSED, context);
+
+    std::uint32_t iVal;
+    if (beast::lexicalCastChecked(iVal, index))
+        return getLedger(ledger, iVal, context);
+
+    return {rpcINVALID_PARAMS, "ledgerIndexMalformed"};
 }
 }  // namespace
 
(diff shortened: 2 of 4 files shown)


src/ripple/rpc/impl/RPCHelpers.cpp
```diff
@@ -241,10 +241,13 @@ ledgerFromRequest(T& ledger, JsonContext& context)
         }
         else
         {
+            int iVal;
```
nbougalis, line 244: I don't think it matters in the short term, but this should be a `std::uint32_t` to match the argument for `getLedger` and also catch "mistakes" like using negative ledger indices.

Also, I guess, to ensure this doesn't break sometime around 2235 when we hit 2 billion ledgers, because we will, invariably, be on vacation and it sucks to have your vacation interrupted because of an integer over/underflow.

src/ripple/rpc/impl/RPCHelpers.cpp
```diff
@@ -241,10 +241,13 @@ ledgerFromRequest(T& ledger, JsonContext& context)
...
             if (index.empty() || index == "current")
                 return getLedger(ledger, LedgerShortcut::CURRENT, context);
             else if (index == "closed")
                 return getLedger(ledger, LedgerShortcut::CLOSED, context);
+            else if (beast::lexicalCastChecked(iVal, indexValue.asString()))
```
nbougalis, line 249: You don't need `indexValue.asString()` since you already have it: it's `index`.

Also, we should simplify the code a bit. There's a weird `if/else{if/else if/else}` hierarchy here that seems confusing. Please consider whether simplifying things helps make lines 238-255 more readable:

```
if (index.empty() || index == "current")
    return getLedger(ledger, LedgerShortcut::CURRENT, context);

if (index == "validated")
    return getLedger(ledger, LedgerShortcut::VALIDATED, context);

if (index == "closed")
    return getLedger(ledger, LedgerShortcut::CLOSED, context);

if (std::uint32_t x; beast::lexicalCastChecked(x, index))
    return getLedger(ledger, x, context);

return {rpcINVALID_PARAMS, "ledgerIndexMalformed"};
```

Note too that there's an `isNumeric` check on lines 231-233 could simply be eliminated and made to flow through to this code, simplifying the logic further (albeit at the cost of a little more work to convert the integer to a string and then back into an integer).

src/test/rpc/LedgerRPC_test.cpp
```diff
@@ -109,14 +125,6 @@ class LedgerRPC_test : public beast::unit_test::suite
         env.trust(USD(1000), bob);
         env.close();
 
-        {
```
nbougalis, line 112: I'd leave this here and change it to something like a garbage string, like "potato". I'd also consider adding a test for a negative ledger index (you can pass it as either a string or an integer)

gregtatcam:
> Looks pretty good, but I'd like to request a couple of changes.
> 
> 1. The DepositAuthorized unit tests are failing as a result of your change.  You can choose to fix them yourself, or you could cherry-pick this commit which has a fix: [scottschurr@3461cd4](https://github.com/scottschurr/rippled/commit/3461cd4a57e634326e91fe6511492c477074fb17)
> 2. The commit messages need some work.  Here is generally good advice to follow: https://chris.beams.io/posts/git-commit/  But there are two specific changes I'd like to see
> 
> * The first line of the first commit message is too long.  It should be 60 characters or less.  If there's more stuff coming then that first line should end with a colon followed by a blank line followed by the additional stuff.  Consider something like this for that first commit message...
> 
> > RPC ledger_index argument accepts quoted numeric strings:
> > Some RPC commands return ledger_index as a quoted numeric string.
> > This change allows the returned value to be directly copied and
> > used for follow-on RPC commands.
> > FIXES 3533
> 
> * Presumably the remaining commits will want to be squashed into the first commit when this pull request is merged.  Please start the commit message of any commits you want squashed with `[FOLD]`.  The person assembling a release may not have been following the code review.  Including `[FOLD]` helps the person building the release to know whether they have additional work to do. So the second commit message might look like this...
> 
> > [FOLD] Update unit-test
> 
> The additional bug-fix and tuning commits should all include `[FOLD]` at the front. So, for example, the commit message for the third commit might look like...
> 
> > [FOLD] Simplify ledger_index parsing:
> > Change lexicalCastChecked() argument to std::unit32_t.
> > Update unit-test.
> 
> I'm sure you get the idea. Thanks.

Thank you. I fixed the unit test.

I'll update the commit message and squash per your recommendations once the PR is approved.

src/ripple/json/json_value.h
```diff
@@ -136,7 +136,7 @@ operator!=(StaticString x, std::string const& y)
  * The sequence of an #arrayValue will be automatically resize and initialized
  * with #nullValue. resize() can be used to enlarge or truncate an #arrayValue.
  *
- * The get() methods can be used to obtanis default value in the case the
+ * The get() methods can be used to obtain default value in the case the
```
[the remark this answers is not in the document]
gregtatcam: Thank you. Updated.

PR #3584, opened 2020-08-14
Improve shard finalize memory utilization

miguelportilla:
This pull request addresses memory utilization during the finalization stage of a shard. The current process visits the SHAmaps nodes of every ledger in a shard causing the family shard and backend tagged caches to grow quicker than they are pruned. The proposed solution resets the appropriate caches between ledgers to drastically reduce memory usage.

### DIFF
--- a/src/ripple/nodestore/impl/Shard.cpp
+++ b/src/ripple/nodestore/impl/Shard.cpp
@@ -219,7 +219,17 @@ Shard::open(Scheduler& scheduler, nudb::context& ctx)
             std::string("exception ") + e.what() + " in function " + __func__);
     }
 
-    setBackendCache(lock);
+    // Set backend caches
+    {
+        auto const size{config.getValueFor(SizedItem::nodeCacheSize, 0)};
+        auto const age{std::chrono::seconds{
+            config.getValueFor(SizedItem::nodeCacheAge, 0)}};
+        auto const name{"shard " + std::to_string(index_)};
+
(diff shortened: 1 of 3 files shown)


src/ripple/nodestore/impl/Shard.cpp
```diff
@@ -219,7 +219,17 @@ Shard::open(Scheduler& scheduler, nudb::context& ctx)
             std::string("exception ") + e.what() + " in function " + __func__);
     }
 
-    setBackendCache(lock);
+    // Set backend caches
+    {
+        auto const size{config.getValueFor(SizedItem::nodeCacheSize, 0)};
```
nbougalis, line 224: I get why you chose `0` here but is auto-sizing to `tiny` a good idea here? The caches would end up up holding ~16k entries. Perhaps having a more intelligent scheme could afford better usage of resources; for example, if the configured `NODE_SIZE` is `large` or `huge` we configure these caches to be `medium`, otherwise we opt for `tiny`.

miguelportilla: A shard's cache can be much smaller than the node store cache, especially completed shards. A possible optimization is to use a larger cache size with the 'current' shard as it will need to have better performance. We can easily add that when we offer the ability to use shards as the primary store.

miguelportilla:
> We should consider whether finalizing in a forked process makes sense, as it would allow us to more finely control resource usage.

Thanks for the feedback, Nik. I am not sure I understand how the forked process would work and benefit us. What did you have in mind?

PR #3585, opened 2020-08-14
Database downloader fix

undertome:
This PR implements a fix for an issue discovered by @seelabs and @ximinez wherein the `DatabaseDownloader` unit test crashes when a `boost::asio` coroutine is invoked for an object (namely an instance of `DatabaseDownloader`) that has already been destroyed.

The problem was solved by delaying destruction of the object until the coroutine has completed. This was achieved by calling `SSLHTTPDownloader::onStop` from the `DatabaseDownloader` desctructor.

To trigger the crash, I modified the following line in the unit test:

`auto constexpr timeout = 2s;`

I changed the timeout value to `10us` in order to consistently produce a crash. Each time the crash occurred, the offending line was inside of `SSLHTTPDownloader::do_session` and the downloader had already been destroyed, which is consistent with the stack trace from the crash observed by Scott. Naturally, I was able to use this miniature timeout value to confirm the fix.

**FYI:** This PR is stacked on two unmerged PRs. The relevant code is in commit 56c0224.

### DIFF
--- a/src/ripple/net/impl/HTTPDownloader.cpp
+++ b/src/ripple/net/impl/HTTPDownloader.cpp
@@ -24,12 +24,12 @@ namespace ripple {
 
 HTTPDownloader::HTTPDownloader(
     boost::asio::io_service& io_service,
-    beast::Journal j,
-    Config const& config)
+    Config const& config,
+    beast::Journal j)
     : j_(j)
     , config_(config)
     , strand_(io_service)
-    , cancelDownloads_(false)
+    , stop_(false)
     , sessionActive_(false)
 {
 }
@@ -47,19 +47,18 @@ HTTPDownloader::download(
     if (!checkPath(dstPath))
         return false;
 
+    if (stop_)
+        return true;
+
     {
         std::lock_guard lock(m_);
-
-        if (cancelDownloads_)
-            return true;
-
         sessionActive_ = true;
     }
 
     if (!strand_.running_in_this_thread())
         strand_.post(std::bind(
             &HTTPDownloader::download,
-            this,
+            shared_from_this(),
             host,
             port,
             target,
@@ -72,7 +71,7 @@ HTTPDownloader::download(
             strand_,
             std::bind(
                 &HTTPDownloader::do_session,
-                this,
+                shared_from_this(),
                 host,
                 port,
                 target,
@@ -82,20 +81,6 @@ HTTPDownloader::download(
                 ssl,
                 std::placeholders::_1));
     return true;
-}
-
-void
-HTTPDownloader::onStop()
-{
-    std::unique_lock lock(m_);
-
-    cancelDownloads_ = true;
-
-    if (sessionActive_)
-    {
-        // Wait for the handler to exit.
-        c_.wait(lock, [this]() { return !sessionActive_; });
-    }
 }
 
 void
@@ -140,21 +125,24 @@ HTTPDownloader::do_session(
     // because the server is shutting down,
     // this method notifies a 'Stoppable'
     // object that the session has ended.
-    auto exit = [this]() {
-        std::lock_guard<std::mutex> lock(m_);
+    auto exit = [this, &dstPath, complete] {
+        if (!stop_)
+            complete(std::move(dstPath));
+
+        std::lock_guard lock(m_);
         sessionActive_ = false;
         c_.notify_one();
     };
 
     auto failAndExit = [&exit, &dstPath, complete, &ec, this](
                            std::string const& errMsg, auto p) {
+        fail(dstPath, ec, errMsg, p);
         exit();
-        fail(dstPath, complete, ec, errMsg, p);
     };
     // end lambdas
     ////////////////////////////////////////////////////////////
 
-    if (cancelDownloads_.load())
+    if (stop_.load())
         return exit();
 
     auto p = this->getParser(dstPath, complete, ec);
@@ -274,7 +262,7 @@ HTTPDownloader::do_session(
     // Download the file
     while (!p->is_done())
     {
-        if (cancelDownloads_.load())
+        if (stop_.load())
         {
             close(p);
             return exit();
@@ -287,15 +275,24 @@ HTTPDownloader::do_session(
 
     close(p);
     exit();
+}
 
-    // Notify the completion handler
-    complete(std::move(dstPath));
+void
+HTTPDownloader::onStop()
+{
+    stop_ = true;
+
+    std::unique_lock lock(m_);
+    if (sessionActive_)
+    {
+        // Wait for the handler to exit.
+        c_.wait(lock, [this]() { return !sessionActive_; });
+    }
 }
 
 void
 HTTPDownloader::fail(
     boost::filesystem::path dstPath,
-    std::function<void(boost::filesystem::path)> const& complete,
     boost::system::error_code const& ec,
     std::string const& errMsg,
     std::shared_ptr<parser> parser)
@@ -321,7 +318,6 @@ HTTPDownloader::fail(
         JLOG(j_.error()) << "exception: " << e.what()
                          << " in function: " << __func__;
     }
-    complete(std::move(dstPath));
 }
 
 }  // namespace ripple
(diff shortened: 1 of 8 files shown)


undertome:
> This patch solves the problem of managing the lifetime of the DatabaseDownloader that was causing the unit test crash. I do have a couple issues I'd like to discuss:
> 
>     * This problem is often solved by inheriting from `std::enable_shared_from_this` and passing a `shared_ptr` into the asio callbacks. The asio documentation uses this and most of rippled also uses this. Using existing idioms helps keep our software understandable. There are often reasons not to use existing idioms, but they should be documented and discussed. Is there a reason not to use the existing idiom here?
> 
>     * Apart from using a different idiom, this solution requires setting an `sessionAcitve_` member and making sure no member functions are ever called after `sessionActive_` is set to false (or the object may be deleted). It also requires that derived classes call `onStop` in their destructors. The `shared_from_this` solution is does automatically what this solution is doing manually.
> 
> 
> Unless there are rationals for using `sessionActive_` that I'm missing, I'd prefer that we use the standard `shared_from_this` solution to keep objects alive in asio callbacks.



> This patch solves the problem of managing the lifetime of the DatabaseDownloader that was causing the unit test crash. I do have a couple issues I'd like to discuss:
> 
>     * This problem is often solved by inheriting from `std::enable_shared_from_this` and passing a `shared_ptr` into the asio callbacks. The asio documentation uses this and most of rippled also uses this. Using existing idioms helps keep our software understandable. There are often reasons not to use existing idioms, but they should be documented and discussed. Is there a reason not to use the existing idiom here?
> 
>     * Apart from using a different idiom, this solution requires setting an `sessionAcitve_` member and making sure no member functions are ever called after `sessionActive_` is set to false (or the object may be deleted). It also requires that derived classes call `onStop` in their destructors. The `shared_from_this` solution is does automatically what this solution is doing manually.
> 
> 
> Unless there are rationals for using `sessionActive_` that I'm missing, I'd prefer that we use the standard `shared_from_this` solution to keep objects alive in asio callbacks.

The current approach was implemented during #3297 where instead of arbitrarily extending the life of objects with `shared_ptr` we exercised more precision over the lifespan of objects.

If we just use `shared_ptr` and remove `sessionActive_`, the `ShardArchiveHandler` and the `SSLHTTPDownloader` will be destroyed independently at shutdown. We'll have no guarantee that the handler will outlive the downloader, and we'll have no way to ensure that the downloader isn't accessing destroyed objects. The current approach allows us to comply with graceful shutdowns.

Additionally, while the downloader is currently used only by the `ShardArchiveHandler`, it's still a generic class. If it gets used by some other module that queues several downloads directly to the downloader's, strand, just using a `shared_ptr` actually works against us, as we'll have the thing continuously trying to issue new downloads while the server is supposed to be stopping.

thejohnfreeman:
I dislike the pervasive use of `std::shared_ptr` in our code, but I think it is warranted here for a few reasons.

1. It is the established, well-understood practice for callbacks in Boost.Asio. People familiar with Boost.Asio have fewer surprises with this approach.

2. I don't like the idea of blocking destructors to wait for callbacks dependent on the destroyed object. I think we generally prefer to just cancel the callbacks, and one way to do that is through the `weak_ptr` pattern prescribed by Boost.Asio. I would like to find a better approach to cancellable async operations, but this is not going that direction.

src/ripple/rpc/impl/ShardArchiveHandler.cpp
```diff
@@ -307,15 +307,15 @@ ShardArchiveHandler::release()
...
+
     if (archives_.empty())
     {
         doRelease(l);
         return false;
     }
 
-    if (isStopping())
-        return false;
-
```
seelabs, line 318: On line 384 of this file, we're passing `this` into a callback used by `downloader_->download`. We need to make sure that `this` lives long enough or we'll get a use after free error. This does appear to be the case. `ShardArchiveHandler` is created once in the application and then never changed. Can you do a quick dummy check me for me to make sure that's true?

(apologies: I cannot leave a comment on that line directly since it is not part of this changeset).

src/ripple/net/SSLHTTPDownloader.h
```diff
@@ -41,7 +41,7 @@ namespace ripple {
 
 /** Provides an asynchronous HTTPS file downloader
  */
-class SSLHTTPDownloader
+class SSLHTTPDownloader : public std::enable_shared_from_this<SSLHTTPDownloader>
```
seelabs, line 44: Consider making the constructor private and adding a friend function:
```
    friend std::shared_ptr<DatabaseDownloader>
    make_DatabaseDownloader(
        boost::asio::io_service& io_service,
        Config const& config,
        beast::Journal j);
```
Rational: this class *must* be accessed through a shared pointer, so it would be good if the interface only allowed creating shared pointers. One downside: we can't use `make_shared` since it doesn't have access to the private constructor. This will cost us an extra allocation. However, given how infrequently this class is created that's probably an OK tradeoff.

seelabs (commented):
This looks good! I left one minor suggestion and one thing I'd like a dummy check on. I also noticed that the `journal` parameter is not the last parameter in some of the functions (not in this change set, but something we should cleanup). Here's a patch with my suggested changes for easy reference: https://github.com/seelabs/rippled/commit/0715dd78a2490e37c863515bd14c5c3616d92d33

src/ripple/rpc/impl/ShardArchiveHandler.cpp
undertome: Yes, that's true. Per the current design the `ShardArchiveHandler` will always outlive the downloader.

src/ripple/net/SSLHTTPDownloader.h
undertome: Cherry picked this change into my branch.

seelabs (approved):
:+1: Nice job on this and thanks for making the large requested change!

thejohnfreeman (approved):

src/ripple/net/impl/HTTPDownloader.cpp
```diff
@@ -140,22 +126,25 @@ HTTPDownloader::do_session(
     // because the server is shutting down,
     // this method notifies a 'Stoppable'
     // object that the session has ended.
-    auto exit = [this]() {
+    auto exit = [this, &dstPath, complete](bool invokeComplete) {
+        if (invokeComplete)
```
miguelportilla, line 130: The only time ```complete``` isn't called is when  ```cancelDownloads_``` is true.
Consider this simplification...
```
    auto exit = [this, &dstPath, complete]() {
        if (!cancelDownloads_)
            complete(std::move(dstPath));

        std::lock_guard<std::mutex> lock(m_);
        sessionActive_ = false;
        c_.notify_one();
    };
```

src/ripple/net/impl/HTTPDownloader.cpp
```diff
@@ -140,22 +126,25 @@ HTTPDownloader::do_session(
...
+            complete(std::move(dstPath));
+
         std::lock_guard<std::mutex> lock(m_);
```
miguelportilla, line 133: nit: can be ```std::lock_guard lock(m_);```

src/ripple/net/impl/HTTPDownloader.cpp
```diff
@@ -286,16 +275,26 @@ HTTPDownloader::do_session(
...
+}
+
+void
+HTTPDownloader::onStop()
+{
+    std::unique_lock lock(m_);
 
-    // Notify the completion handler
-    complete(std::move(dstPath));
+    cancelDownloads_ = true;
```
miguelportilla, line 286: Can this assignment occur before the lock? Same with evaluation in ```download```.

src/ripple/net/impl/HTTPDownloader.cpp
miguelportilla: Consider renaming ```cancelDownloads_``` to ```stop_```

undertome: Done

src/ripple/net/impl/HTTPDownloader.cpp
undertome: Done

src/ripple/net/impl/HTTPDownloader.cpp
undertome: Yes, done

miguelportilla (approved):
👍

undertome:
Force pushed to remove merge conflicts with develop

PR #3587, opened 2020-08-18
Refactor AccountLinesRPC defaultRipple response

### LINKED ISSUE #3245: account_lines method does not return no_ripple / no_ripple_peer in non-default state
Migrated from internal issue [RIPD-1031](https://ripplelabs.atlassian.net/browse/RIPD-1031).

By convention, the [account_lines method](https://xrpl.org/account_lines.html) returns booleans for flags that are in their non-default states. In the vast majority of cases, this means it returns flags when the values are `true` and omits them when the values are `false`. The NoRipple flags of a trust line are exceptions, and the API doesn't handle them properly. This is probably because the API was not changed after DefaultRipple was added, because DefaultRipple _changes what's considered the default state_ of the NoRipple flag.

The API should explicitly include the `no_ripple` and `no_ripple_peer` flags when their current value is non-default as determined by the lsfDefaultRipple flags of their respective accounts. This would make it far, far easier to recognize when a NoRipple flag (on either side) is causing a trust line to be in a non-default state.

For an example of a trust line like this, see [rPRvq4qwRgtsUy7JM911swhESfDDc5tm1D←→rGatEqm5MDi5bk8FSZEi9xmJiiNpdZx5VU](https://xrpl.org/websocket-api-tool.html?server=wss%3A%2F%2Fs2.ripple.com%2F&req=%7B%22id%22%3A2%2C%22command%22%3A%22account_lines%22%2C%22account%22%3A%22rPRvq4qwRgtsUy7JM911swhESfDDc5tm1D%22%2C%22ledger_index%22%3A%22validated%22%7D).

```json
{
  "id": 2,
  "result": {
    "account": "rPRvq4qwRgtsUy7JM911swhESfDDc5tm1D",
    "ledger_hash": "6467B8AE06B83BFE136CD2E7039CCB116800123EB03AE8A654E44BFDC7A9D4A1",
    "ledger_index": 53188407,
    "lines": [
      {
        "account": "rGatEqm5MDi5bk8FSZEi9xmJiiNpdZx5VU",
        "balance": "0",
        "currency": "USD",
        "limit": "0",
        "limit_peer": "0",
        "quality_in": 0,
        "quality_out": 0
      }
    ],
    "validated": true
  },
  "status": "success",
  "type": "response"
}
```

It's impossible to tell from just this information, but this trust line counts against rPRv...'s owner reserve, even though it appears that the trust line has no balances, no limits, no quality settings, and no flags enabled. So why's it considered to be in a non-default state? It's because rPRv...'s NoRipple setting for this trust line is _off_, but the default for rPRv...'s trust lines should be for NoRipple to be _on_, because rPRv's DefaultRipple setting is _off_. (All these double negatives make it confusing, which is kind of a different problem.)

Anyway, the point here is that a more user-friendly API would provide `no_ripple` and `no_ripple_peer` in the response whenever their values are preventing a trust line from being in its default state. Of course, right now clients are instructed to assume that any flags which are omitted are false, so we can't omit them if they're in a default state of `true`. So, to be backwards-compatible with the existing API, I propose the following logic:

For a trust line between accounts A and B, where A is the "perspective account" used in the API request's `account` field and B is the "peer account" for a given trust line:

- If A's side of the trust line has NoRipple enabled, return `"no_ripple": true` always.
- Else, if A has DefaultRipple disabled, explicitly return `"no_ripple": false`. (This is the case where "false" is the non-default value.)
- Else, omit `no_ripple`. (NoRipple is off and that's the default.)

Also apply the exact same logic to the peer account "B" for the field `no_ripple_peer`.

natenichols:
Fixes #3245. The AccountLinesRPC takes default no_ripple into the API response.

### Context of Change
The API will return the `no_ripple` field when `no_ripple: true` or when `no_ripple: false` and defaultRipple is disabled.

From issue #3245: 
For a trust line between accounts A and B, where A is the "perspective account" used in the API request's account field and B is the "peer account" for a given trust line:

    1) If A's side of the trust line has NoRipple enabled, return "no_ripple": true always.
    2) Else, if A has DefaultRipple disabled, explicitly return "no_ripple": false. 
    3) Else, omit no_ripple. (NoRipple is off and that's the default.)

### Type of Change
- [ ] Documentation Updates
- [ ] Release

## Before / After
**Before:**
The API only returns `no_ripple` when it is `true`

**After:**
The API only returns `no_ripple` when it is `true`. It also returns `no_ripple: false` when defaultRipple disabled. This is the case where `false` is the non-default value.

## Test Plan
Modified three tests to expect `no_ripple` field to be `false`.

### DIFF
--- a/src/ripple/app/paths/RippleState.h
+++ b/src/ripple/app/paths/RippleState.h
@@ -88,6 +88,12 @@ public:
     }
 
     bool
+    getDefaultRipple() const
+    {
+        return mFlags & lsfDefaultRipple;
+    }
+
+    bool
     getNoRipple() const
     {
         return mFlags & (mViewLowest ? lsfLowNoRipple : lsfHighNoRipple);
--- a/src/ripple/rpc/handlers/AccountLines.cpp
+++ b/src/ripple/rpc/handlers/AccountLines.cpp
@@ -62,10 +62,10 @@ addLine(Json::Value& jsonLines, RippleState const& line)
         jPeer[jss::authorized] = true;
     if (line.getAuthPeer())
         jPeer[jss::peer_authorized] = true;
-    if (line.getNoRipple())
-        jPeer[jss::no_ripple] = true;
-    if (line.getNoRipplePeer())
-        jPeer[jss::no_ripple_peer] = true;
+    if (line.getNoRipple() || !line.getDefaultRipple())
+        jPeer[jss::no_ripple] = line.getNoRipple();
+    if (line.getNoRipplePeer() || !line.getDefaultRipple())
+        jPeer[jss::no_ripple_peer] = line.getNoRipplePeer();
     if (line.getFreeze())
         jPeer[jss::freeze] = true;
     if (line.getFreezePeer())
--- a/src/test/rpc/NoRipple_test.cpp
+++ b/src/test/rpc/NoRipple_test.cpp
@@ -135,12 +135,12 @@ public:
             {
                 auto const aliceLines = getAccountLines(alice);
                 BEAST_EXPECT(aliceLines.size() == 1);
-                BEAST_EXPECT(!aliceLines[0u].isMember(jss::no_ripple));
+                BEAST_EXPECT(aliceLines[0u][jss::no_ripple].asBool() == false);
 
                 auto const bobLines = getAccountLines(bob);
                 BEAST_EXPECT(bobLines.size() == 2);
-                BEAST_EXPECT(!bobLines[0u].isMember(jss::no_ripple));
-                BEAST_EXPECT(!bobLines[1u].isMember(jss::no_ripple));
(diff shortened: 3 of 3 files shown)


seelabs:
I'm tagging this as "API Change", even though it's minor, just to make sure the docs team is aware of this.

mDuo13:
The convention to omit fields in many cases is (unfortunately) used throughout the API, so it would be a bit inconsistent to provide the field in all cases, as @cjcobb23 suggests.

Also, I think it's most appropriate to consider this a **non-breaking** change. The response adds a field with the value `false` to some responses, specifically in cases where that field previously would have been omitted _because_ its value was `false`. I can't think of a way that this could break a reasonable API client. (This purely adds information: if you see `"no_ripple": false` in a response it indicates a non-default `false` value.)

mDuo13 (approved):

cjcobb23 (approved):

PR #3589, opened 2020-08-20
Add unit test of correlation between KnownFormats and gRPC

scottschurr:
This commit adds a unit test that does its best to correlate RPC and gRPC representation of...

- Ledger objects and
- Transactions.

The point of performing the correlation is to provide early warning when the RPC and gRPC implementations get out of sync.  Typically new features or modifications to the XRP Ledger are made in an RPC-aware way, and the gRPC implementation is left behind.  The new unit test should fail when that happens.  Correcting the problem may involve fixes to the gRPC implementation, the KnownFormats or SFields, and/or to the unit test.

The test uses two key runtime introspection devices:

1. The XRP Ledger provides `KnowFormat`s which describe the required structure of important composite data types used by the ledger

2. Protobuf (and by extension gRPC, which is built on Protobuf) provides Descriptors (https://developers.google.com/protocol-buffers/docs/reference/cpp/google.protobuf.descriptor#Descriptor) which offer a way to query the structure and constraints of an interface built on Protobuf.

The test works by iterating through KnownFormats and the corresponding gRPC Descriptors.  During the iteration the names and data types of fields in the KnownFormat are compared to those in the Protobuf Descriptor.

The gRPC implementation is not completely aligned with the KnownFormat implementation.  So the unit test provides mappings where mappings are necessary and possible.  There are also a couple of points where the comparison is necessarily limited.  When the comparison hits these limits the comparison stops short before fully reaching bottom.

But the point of the test is not to validate the current implementations.  The point is for the test to fail when future changes or additions cause the implementations to get out of sync.  The test has already served that purpose once.  It helped to identify deficiencies in the KnownFormat and gRPC implementations of the nUNL.  The hope is that it will provide similar assistance in the future.

### DIFF
--- a/src/ripple/protocol/KnownFormats.h
+++ b/src/ripple/protocol/KnownFormats.h
@@ -130,6 +130,19 @@ public:
         return itr->second;
     }
 
+    // begin() and end() are provided for testing purposes.
+    typename std::forward_list<Item>::const_iterator
+    begin() const
+    {
+        return formats_.begin();
+    }
+
+    typename std::forward_list<Item>::const_iterator
+    end() const
+    {
+        return formats_.end();
+    }
+
 protected:
     /** Retrieve a format based on its name.
      */
--- a/src/test/protocol/KnownFormatToGRPC_test.cpp
+++ b/src/test/protocol/KnownFormatToGRPC_test.cpp
@@ -1,0 +1,962 @@
+//------------------------------------------------------------------------------
+/*
+    This file is part of rippled: https://github.com/ripple/rippled
+    Copyright (c) 2020 Ripple Labs Inc.
+
+    Permission to use, copy, modify, and/or distribute this software for any
+    purpose  with  or without fee is hereby granted, provided that the above
+    copyright notice and this permission notice appear in all copies.
+
+    THE  SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+    WITH  REGARD  TO  THIS  SOFTWARE  INCLUDING  ALL  IMPLIED  WARRANTIES  OF
+    MERCHANTABILITY  AND  FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+    ANY  SPECIAL ,  DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+    WHATSOEVER  RESULTING  FROM  LOSS  OF USE, DATA OR PROFITS, WHETHER IN AN
+    ACTION  OF  CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+    OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+//==============================================================================
+
+#include <ripple/basics/safe_cast.h>
+#include <ripple/beast/unit_test.h>
+#include <ripple/protocol/InnerObjectFormats.h>
+#include <ripple/protocol/LedgerFormats.h>
+#include <ripple/protocol/TxFormats.h>
+
+#include "org/xrpl/rpc/v1/ledger_objects.pb.h"
+#include "org/xrpl/rpc/v1/transaction.pb.h"
+
+#include <cctype>
+#include <map>
+#include <string>
+#include <type_traits>
+
+namespace ripple {
+
+// This test suite uses the google::protobuf::Descriptor class to do runtime
+// reflection on our gRPC stuff.  At the time of this writing documentation
+// for Descriptor could be found here:
+//
+// https://developers.google.com/protocol-buffers/docs/reference/cpp/google.protobuf.descriptor#Descriptor
+
+class KnownFormatToGRPC_test : public beast::unit_test::suite
+{
+private:
+    static constexpr auto fieldTYPE_UINT32 =
+        google::protobuf::FieldDescriptor::Type::TYPE_UINT32;
+
+    static constexpr auto fieldTYPE_UINT64 =
+        google::protobuf::FieldDescriptor::Type::TYPE_UINT64;
+
+    static constexpr auto fieldTYPE_BYTES =
+        google::protobuf::FieldDescriptor::Type::TYPE_BYTES;
+
+    static constexpr auto fieldTYPE_STRING =
+        google::protobuf::FieldDescriptor::Type::TYPE_STRING;
+
+    static constexpr auto fieldTYPE_MESSAGE =
+        google::protobuf::FieldDescriptor::Type::TYPE_MESSAGE;
+
+    // Format names are CamelCase and FieldDescriptor names are snake_case.
+    // Convert from CamelCase to snake_case.  Do not be fooled by consecutive
+    // capital letters like in NegativeUNL.
+    static std::string
+    formatNameToEntryTypeName(std::string const& fmtName)
+    {
+        std::string entryName;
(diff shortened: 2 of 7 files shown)


src/ripple/protocol/KnownFormats.h
```diff
@@ -130,6 +130,19 @@ class KnownFormats
         return itr->second;
     }
 
+    // begin() and end() are provided for testing purposes.
+    typename std::forward_list<Item>::const_iterator
+    begin() const
+    {
+        return std::as_const(formats_).begin();
```
seelabs, line 137: No need for `as_const` (for `end` as well).

src/test/protocol/KnownFormatToGRPC_test.cpp
```diff
@@ -0,0 +1,924 @@
...
+
+            prevUpper = upper;
+            entryName.push_back(std::tolower(ch));
+        }
+        return entryName;
+    };
+
+    // Create a map of (most) all the SFields in an SOTemplate.  This map
+    // can be used to correlate a gRPC Descriptor to its corresponding SField.
+    template <typename KeyType = int>
```
seelabs, line 84: The template parameter is going to be inferred from the function argument. I don't think the default parameter will be used. I'd remove it (unless there's a rational for it that I'm missing).

src/test/protocol/KnownFormatToGRPC_test.cpp
```diff
@@ -0,0 +1,924 @@
...
+
+            // Fields that gRPC never includes.  This is probably intentional.
+            using FieldCode_t =
+                std::remove_const<decltype(SField::fieldCode)>::type;
+            static const std::set<FieldCode_t> excludedSFields{
+                sfLedgerIndex.fieldCode,
+                sfLedgerEntryType.fieldCode,
+                sfOperationLimit.fieldCode};
+
+            if (excludedSFields.count(sField.fieldCode) == 1)
```
seelabs, line 103: There are lots of places that follow this pattern. Either `if (someSetOrMap(). count(key) == 1)` or ` != 1`. Instead of this, I'd treat the return value of `count` as a bool and leave off the `== 1` or `!= 1`. Rational: we're using `count` as a stand-in for `contains`. We don't care about the actual count, just if the collection contains the key or not.

src/test/protocol/KnownFormatToGRPC_test.cpp
```diff
@@ -0,0 +1,924 @@
...
+                {"Signature",         "PaymentChannelSignature"},
+                {"SigningPubKey",     "SigningPublicKey"},
+                {"TakerGetsCurrency", "TakerGetsCurreny"},
+                {"TxnSignature",      "TransactionSignature"},
+            };
+            // clang-format on
+
+            std::string gRPCName = sFieldToGRPC.count(sField.getName()) == 1
+                ? sFieldToGRPC.at(sField.getName())
+                : sField.getName();
```
seelabs, line 174: This does the lookup twice (once for `count` and once for `at`). We could use `find` instead to save a lookup. I honestly don't care in code like this, it's fine as-is, it just caught my eye.

src/test/protocol/KnownFormatToGRPC_test.cpp
```diff
@@ -0,0 +1,924 @@
...
+        }
+        return sFields;
+    }
+
+    // Given a Descriptor for a KnownFormat and a map of the SFields of that
+    // KnownFormat, make sure the fields are aligned.
+    void
+    validateDescriptorAgainstSFields(
+        google::protobuf::Descriptor const* const pbufDescriptor,
+        std::optional<google::protobuf::Descriptor const* const> commonFields,
```
seelabs, line 186: Could `commonFields` just be a pointer instead of an optional pointer? It looks like a null pointer works just as well as an un-seated optional in the code.

src/test/protocol/KnownFormatToGRPC_test.cpp
```diff
@@ -0,0 +1,924 @@
...
+        std::string const& knownFormatName,
+        std::map<std::string, SField const*> sFields)
```
seelabs, line 188: What do you think about making `sFields` an rvalue ref? I know we often make parameters like this value-types, so no change needed.

src/test/protocol/KnownFormatToGRPC_test.cpp
scottschurr: Yup.  Good call.  My initial take was to use a `nullptr`.  But I struggled with whether `nullptr` was acceptable stylistically these days, so I changed it.  The `nullptr` is more straight forward.

scottschurr:
Great review comments @seelabs!  I believe I've implemented all of your suggestions. Thanks!

src/test/protocol/KnownFormatToGRPC_test.cpp
```diff
@@ -0,0 +1,924 @@
...
+                                return *omit == sField;
+                            }) != omits.end())
+                    {
+                        // This is one of the fields that gRPC omits.
+                        continue;
+                    }
+                }
+            }
+
+            // The SFields and gRPC disagree on the names of some fields.
```
cjcobb23, line 154: I think we should explain why these differences exist, in the comment here. Most of these are the result of the gRPC API not abbreviating anything. 

`Fee -> XRPDropsAmount` is mainly the result of adding additional typing. 

There's the obvious misspelling with `TakerGetsCurreny`.

And then lastly, `Signature -> PaymentChannelSignature` was a deliberate deviation, with the rational being that `PaymentChannelClaim` transactions, when represented as JSON, have a `TxnSignature` field at the top level, and then a `Signature` field at a lower level. I thought this was confusing, so I changed `Signature` to `PaymentChannelSignature` to be more explicit.

Let's explain this in a comment.

src/test/protocol/KnownFormatToGRPC_test.cpp
```diff
@@ -0,0 +1,924 @@
...
+                //
+                // Some of these may be bugs.
+                static const std::
+                    map<LedgerEntryType, std::vector<SField const*>>
+                        gRPCOmitFields{
+                            {ltACCOUNT_ROOT, {&sfWalletLocator, &sfWalletSize}},
+                            {ltDIR_NODE, {&sfExchangeRate}},
+                            {ltLEDGER_HASHES, {&sfFirstLedgerSequence}},
+                            {ltPAYCHAN, {&sfDestinationNode}},
+                        };
```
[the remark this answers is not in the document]
cjcobb23: Ah, so this is an actual bug. Here's a patch that adds `DestinationNode` to `PayChannel` and updates the test: https://github.com/cjcobb23/rippled/commit/d7ac8ce8b0ead033400a10fd045e19374ebcbe60 . Would you mind cherry-picking this patch into the PR?

@mDuo13 `sfDestinationNode` is not documented on [https://xrpl.org/paychannel.html](https://xrpl.org/paychannel.html).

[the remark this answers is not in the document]
scottschurr: Thanks, @cjcobb23.  I have cherry-picked the patch.

scottschurr:
Thanks for the reviews, gentlefolk.  I've squashed the commits and I'll mark it `passed` after CI passes.

scottschurr:
Rebased to 1.7.0-b1.  I had to tweak the test a bit, since 1.7.0-b1 included new support for tickets.  It would probably be good for someone to take a look a the changes.  Thanks.

cjcobb23 (approved):
Latest commit looks good to me

scottschurr:
Thanks @cjcobb23.  The branch is now squashed.

PR #3591, opened 2020-08-24
Allow 100 signers

shaikezr:
Kava's bridge use case requires them to be able to set quorums of 67 on signer lists with 100 members. This PR would increase the max size of the signer list from 8 to 100.

### DIFF
--- a/src/ripple/protocol/Feature.h
+++ b/src/ripple/protocol/Feature.h
@@ -113,7 +113,8 @@ class FeatureCollections
                     // payment check
         "HardenedValidations",
         "fixAmendmentMajorityCalc",  // Fix Amendment majority calculation
-        "NegativeUNL"};
+        "NegativeUNL",
+        "increaseMaxSignersTo100"};
     std::vector<uint256> features;
     boost::container::flat_map<uint256, std::size_t> featureToIndex;
     boost::container::flat_map<std::string, uint256> nameToFeature;
@@ -370,6 +371,7 @@ extern uint256 const fix1781;
 extern uint256 const featureHardenedValidations;
 extern uint256 const fixAmendmentMajorityCalc;
 extern uint256 const featureNegativeUNL;
+extern uint256 const increaseMaxSignersTo100;
 
 }  // namespace ripple
 
--- a/src/ripple/protocol/STTx.h
+++ b/src/ripple/protocol/STTx.h
@@ -191,6 +191,6 @@ sterilize(STTx const& stx);
 bool
 isPseudoTx(STObject const& tx);
 
-}  // namespace ripple
+}  // namespace ripple 
 
 #endif
--- a/src/ripple/protocol/STTx.h
+++ b/src/ripple/protocol/STTx.h
@@ -191,6 +191,6 @@ sterilize(STTx const& stx);
 bool
 isPseudoTx(STObject const& tx);
 
-}  // namespace ripple 
+}  // namespace ripple
 
 #endif
--- a/src/ripple/protocol/STTx.h
+++ b/src/ripple/protocol/STTx.h
@@ -49,7 +49,7 @@ public:
     }
 
     static std::size_t const minMultiSigners = 1;
-    static std::size_t const maxMultiSigners = 8;
+    static std::size_t const maxMultiSigners = 100;
 
 public:
     STTx() = delete;
--- a/src/ripple/protocol/impl/Feature.cpp
+++ b/src/ripple/protocol/impl/Feature.cpp
@@ -132,6 +132,7 @@ detail::supportedAmendments()
         "fix1781",
         "HardenedValidations",
         "fixAmendmentMajorityCalc",
+        "increaseMaxSignersTo100",
         //"NegativeUNL"      // Commented out to prevent automatic enablement
     };
     return supported;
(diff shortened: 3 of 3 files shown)


nbougalis:
Thanks for your contribution. I haven't yet looked at the code but, at a minimum, this change would require an amendment.

I also think that if we are talking about increasing the list size to 100 it makes sense to think about whether lists should be paged, so that instead of a single object containing a list of 100 you have multiple pages of, say, 32. This has some advantages but one big disadvantage is that you need to load and manipulate multiple objects for every operation involving multisigning.

shaikezr:
Ok, that's a bit more complex. On the read side, I would need to at least:

1. make signers keylet in indexes.cpp able to [accept a page param passed to the static signers keylet](https://github.com/ripple/rippled/blob/develop/src/ripple/protocol/impl/Indexes.cpp#L257)
2. Add logic wherever the signers keylet is used to try paging through

On the write side, I'm a little bit more confused. It looks like right now we're creating a signer array using the size of the obj set in SetSignerList ([here](https://github.com/ripple/rippled/blob/develop/src/ripple/app/tx/impl/SetSignerList.cpp#L391)). To enable paging, would you suggest splitting the signers into 32 entry chunks and calling `writeSignersToSLE` multiple times, once for each chunk, [here](https://github.com/ripple/rippled/blob/develop/src/ripple/app/tx/impl/SetSignerList.cpp#L331)?

nbougalis:
I don't think we should use multiple pages; it complicates the situation significantly and I feel that it outweighs any benefits. We can just increase the size of the list, but that needs to be done under an amendment.

MarkusTeufelberger:
I don't see how "100" is a better/more meaningful number than "8" - it at least makes it far easier to spam the ledger with huge multisigned transactions (and apparently there are already people out there who actually want to use 67 of 100 multisign, so there's a real risk that they are going to start spamming huge transactions right away).

How did you come up with this number and why wouldn't an increase to e.g. 20 (so you can sign 14/20) instead be enough?

nbougalis:
>I don't see how "100" is a better/more meaningful number than "8" - it at least makes it far easier to spam the ledger with huge multisigned transactions (and apparently there are already people out there who actually want to use 67 of 100 multisign, so there's a real risk that they are going to start spamming huge transactions right away).
>
>How did you come up with this number and why wouldn't an increase to e.g. 20 (so you can sign 14/20) instead be enough?

I agree: 100 seems like overkill. We can do it, of course, but have other solutions, like threshold signatures, been considered?

shaikezr:
The questions about threshold signing and why 100 are valid questions.

I made this PR on behalf of Kava Labs because Brian Kerr indicated they needed more signers to support their use case. I chose 100 after asking Brian how many he needed and he replied "We ideally want at least 100 signer capability for our use case. Enabling 67/100 signers to approve of a transaction before it goes through"

I have sent Brian a link to this PR and have asked if he'd be willing to weigh in here because unfortunately I don't have enough info about his use case or what he's already tried.

src/ripple/protocol/Feature.h
```diff
@@ -113,7 +113,8 @@ class FeatureCollections
                     // payment check
         "HardenedValidations",
         "fixAmendmentMajorityCalc",  // Fix Amendment majority calculation
-        "NegativeUNL"};
+        "NegativeUNL",
+        "increaseMaxSignersTo100"};
```
nbougalis, line 117: I'd recommend: `ExpandedSignerLists` instead.

src/ripple/protocol/STTx.h
```diff
@@ -49,7 +49,7 @@ class STTx final : public STObject, public CountedObject<STTx>
     }
 
     static std::size_t const minMultiSigners = 1;
-    static std::size_t const maxMultiSigners = 8;
+    static std::size_t const maxMultiSigners = 100;
```
nbougalis, line 52: This is a problem: this change needs to be gated under an amendment; if the amendment is not active, the old value is used. The new only value only kicks in if the amendment is active.

Take a look at https://github.com/ripple/rippled/pull/3580 which makes some changes to the way signer lists are handled. It will show how to properly gate your code.

intelliot:
> what other options they have evaluated before settling for doing this on-ledger.

I'm certainly no expert on this, but [threshold signatures](https://en.wikipedia.org/wiki/Threshold_cryptosystem) may be an option. I have seen threshold signatures used to achieve multisig functionality/security without needing any special support from the underlying blockchain/ledger.

nbougalis:
> > what other options they have evaluated before settling for doing this on-ledger.
> 
> I'm certainly no expert on this, but [threshold signatures](https://en.wikipedia.org/wiki/Threshold_cryptosystem) may be an option. I have seen threshold signatures used to achieve multisig functionality/security without needing any special support from the underlying blockchain/ledger.

Right, that's exactly what I was wondering about.

carlhua:
My personal take is that allowing 100 signers is way too excessive and the existing allowance is sufficient. Per @MarkusTeufelberger 's comments that by allowing an excessive amounts of signers we make it easy for spams.  I see 2 potential options:

1. Per @intelliot 's suggestion that you scale offchain via multi-sig.
2. Consider BLS signature and charge XRP fee depending on how large the signer's list is. The actual signature's size would be constant.

shaikezr:
I spoke with Kava and they're exploring alternative signing paradigms. I've confirmed with them that they won't be blocked by the current limit of 8 signers per list. I'm going to close this pull request

PR #3597, opened 2020-08-27
Report server domain to other servers:

nbougalis:
This commit introduces a new configuration option that server operators can set. The value is communicated to other servers and is also reported via the `server_info` API.

The value is meant to allow third-party applications or tools to group servers together. For example, a tool that visualizes the network's topology can group servers together.

Similar to the `Domain` field in validator manifests, an operator can claim any domain. Prior to relying on the value returned, the domain should be verified by retrieving the xrp-ledger.toml file from the domain and looking for the server's public key in the`nodes` array.

### DIFF
--- a/src/ripple/basics/StringUtilities.h
+++ b/src/ripple/basics/StringUtilities.h
@@ -148,6 +148,15 @@ trim_whitespace(std::string str);
 boost::optional<std::uint64_t>
 to_uint64(std::string const& s);
 
+/** Determines if the given string looks like a TOML-file hosting domain.
+
+    Do not use this function to determine if a particular string is a valid
+    domain, as this function may reject domains that are otherwise valid and
+    doesn't check whether the TLD is valid.
+ */
+bool
+isProperlyFormedTomlDomain(std::string const& domain);
+
 }  // namespace ripple
 
 #endif
--- a/src/ripple/overlay/README.md
+++ b/src/ripple/overlay/README.md
@@ -243,6 +243,24 @@ encoded in base58 using the standard encoding for node public keys.
 
 See: https://xrpl.org/base58-encodings.html
 
+
+| Field Name          	|  Request          	| Response          	|
+|---------------------	|:-----------------:	|:-----------------:	|
+| `Server-Domain`   	| :white_check_mark: 	| :white_check_mark: 	|
+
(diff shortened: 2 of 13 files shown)


nbougalis:
(Tagging @mDuo13 for documentation purposes)

src/ripple/basics/StringUtilities.h
```diff
@@ -148,6 +148,15 @@ trim_whitespace(std::string str);
...
 to_uint64(std::string const& s);
 
+/** Determines if the given string looks like a valid TOML-file hosting domain.
+
+    Do not use this function to determine if a particular string is a valid
+    domain, as this function may reject domains that are otherwise valid and
+    doesn't check whether the TLD is valid.
+ */
+bool
+isValidTomlDomain(std::string const& domain);
```
undertome, line 158: The comment and the function name seem to be at odds with one another. Maybe a name like `isPossibly...` or `isSeemingly...` might communicate the purpose of the function more clearly. Up to you.

src/ripple/overlay/README.md
```diff
@@ -243,6 +243,24 @@ encoded in base58 using the standard encoding for node public keys.
...
+| `Server-Domain`   	| :white_check_mark: 	| :white_check_mark: 	|
+
+The optional `Server-Domain` field allows a server to report the domain that
+it is operating under. The value is configured by the server administrator in
+the configuration file using the `[server_domain]` key.
+
+The value is advisory and is not used by the code at this time, except for
+reporting purposes. External tools should verify this value prior to using
+it by attempting to locate a [TOML file](https://xrpl.org/xrp-ledger-toml.html)
+under the specified domain and locate the public key of this server under the
```
undertome, line 258: Should this be "locating the public key"  ?

undertome:
Some nit comments, overall looks good. Not worth blocking

HowardHinnant (approved):

src/ripple/basics/StringUtilities.h
nbougalis: You're right. I'll improve the naming and/or comment.

src/ripple/overlay/README.md
nbougalis: Yep. While I talk the England greatly some time the write goodly not. Will fix.

PR #3602, opened 2020-09-08
Improve reporting of unusual unit test failure conditions:

### LINKED ISSUE #3474: Fail if no unit tests are run when --unittest is specified
If rippled is compiled without unit tests, and rippled --unittest is run, the exit code indicates success, and the output is.

0ms, 0 suites, 0 cases, 0 tests total, 0 failures
To prevent this from being overlooked in a CI situation, the exit code should indicate failure. I don't see any need to change the output.

Example: https://travis-ci.org/ximinez/rippled/jobs/159977989#L1006

Exported from RIPD-1296

### LINKED ISSUE #3600: Unit test jobs: Crashed processes should be counted as "failed" (Version: 1.6.0)
## Summary
If a subprocess crashes when running multiple unit tests using the `--unittest-jobs` argument, no results are reported to the main "runner". This results in output that looks like:
```
183.6s, 194 suites, 1241 cases, 1310395 tests total, 0 failures

Suite: ripple.net.DatabaseDownloader failed to complete. The child process may have crashed.
```

## Motivation
Obviously something failed, but the summary message doesn't indicate that.

## Solution
Include the number of possible crashed suites in the summary, so the example output would look like:
```
183.6s, 194 suites, 1241 cases, 1310395 tests total, 1 failures

Suite: ripple.net.DatabaseDownloader failed to complete. The child process may have crashed.
```

ximinez:

* Jobs with no unit tests are counted as failures. Resolves #3474
* Crashed processes are counted as failures. Resolves #3600
* Any tests specified on the command line test do not have matching
  suites are counted as failures.
* Remove unused CI manual test.

This is another "scratching an itch" PR that is the result of debugging / tracking down another problem, in this case, a test crash. In the process, Travis CI found an entry in manual unit test list that no longer exists.

### Context of Change

Enhances unit test code to improve reporting, and catch more "failure" situations. Other than removing a manual test, all changes were made to testing code, including the test running code in Main.cpp.

### Type of Change

---
This change is [Reviewable](https://reviewable.io/reviews/ripple/rippled/3602)

### DIFF
--- a/src/ripple/app/main/Main.cpp
+++ b/src/ripple/app/main/Main.cpp
@@ -213,10 +213,38 @@ public:
                 return true;
         return false;
     }
+
+    std::size_t
+    size() const
+    {
+        return selectors_.size();
+    }
 };
 
 namespace test {
 extern std::atomic<bool> envUseIPv4;
+}
+
+template <class Runner>
+static bool
+anyMissing(Runner& runner, multi_selector const& pred)
+{
+    if (runner.tests() == 0)
+    {
+        runner.add_failures(1);
+        std::cout << "Failed: No tests run" << std::endl;
+        return true;
+    }
+    if (runner.suites() < pred.size())
+    {
+        auto const missing = pred.size() - runner.suites();
+        runner.add_failures(missing);
+        std::cout << "Failed: " << missing
+                  << " filters did not match any existing test suites"
+                  << std::endl;
+        return true;
+    }
+    return false;
 }
 
 static int
@@ -242,7 +270,9 @@ runUnitTests(
 
         multi_runner_child child_runner{num_jobs, quiet, log};
         child_runner.arg(argument);
-        auto const any_failed = child_runner.run_multi(multi_selector(pattern));
+        multi_selector pred(pattern);
+        auto const any_failed =
+            child_runner.run_multi(pred) || anyMissing(child_runner, pred);
 
         if (any_failed)
             return EXIT_FAILURE;
@@ -281,6 +311,9 @@ runUnitTests(
                 ++bad_child_exits;
             }
         }
+
+        parent_runner.add_failures(bad_child_exits);
+        anyMissing(parent_runner, multi_selector(pattern));
 
         if (parent_runner.any_failed() || bad_child_exits)
             return EXIT_FAILURE;
--- a/src/ripple/app/main/Main.cpp
+++ b/src/ripple/app/main/Main.cpp
@@ -271,8 +271,8 @@ runUnitTests(
         multi_runner_child child_runner{num_jobs, quiet, log};
         child_runner.arg(argument);
         multi_selector pred(pattern);
-        auto const any_failed =
-            child_runner.run_multi(pred) || anyMissing(child_runner, pred);
+        auto any_failed = child_runner.run_multi(pred);
+        any_failed = anyMissing(child_runner, pred) || any_failed;
 
         if (any_failed)
             return EXIT_FAILURE;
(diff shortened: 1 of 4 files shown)


src/test/unit_test/multi_runner.cpp
```diff
@@ -348,6 +364,30 @@ multi_runner_base<IsParent>::message_queue_send(
...
+template <bool IsParent>
+std::size_t
+multi_runner_base<IsParent>::suites() const
+{
+    return inner_->suites();
+}
+
+template <bool IsParent>
+void
+multi_runner_base<IsParent>::add_failures(std::size_t failures)
```
natenichols, line 383: Can/Should we make `failures` pass by const ref?

ximinez: The general rule of thumb that the team uses is that if the item is the same size or smaller than a pointer, especially if it's a simple type, just pass by value. So we can, but we generally don't.

src/test/unit_test/multi_runner.cpp
```diff
@@ -348,6 +364,30 @@ multi_runner_base<IsParent>::message_queue_send(
...
+{
+    results results;
+    results.failed += failures;
```
miguelportilla, line 386: Nit: this can be an assignment.

src/ripple/app/main/Main.cpp
```diff
@@ -242,7 +270,9 @@ runUnitTests(
 
         multi_runner_child child_runner{num_jobs, quiet, log};
         child_runner.arg(argument);
-        auto const any_failed = child_runner.run_multi(multi_selector(pattern));
+        multi_selector pred(pattern);
+        auto const any_failed =
+            child_runner.run_multi(pred) || anyMissing(child_runner, pred);
```
miguelportilla, line 275: Nit: consider...
```
        if (child_runner.run_multi(pred) || anyMissing(child_runner, pred))
            return EXIT_FAILURE;
```

miguelportilla, line 275: I noticed that `run_multi` returns `true` on failure. Should call `anyMissing` to report any missing filters if `run_multi` fails?

src/ripple/app/main/Main.cpp
```diff
@@ -213,12 +213,40 @@ class multi_selector
...
+        std::cout << "Failed: No tests run" << std::endl;
+        return true;
+    }
+    if (runner.suites() < pred.size())
+    {
+        auto const missing = pred.size() - runner.suites();
+        runner.add_failures(missing);
+        std::cout << "Failed: " << missing
+                  << " filters did not match any existing test suites"
+                  << std::endl;
```
miguelportilla, line 244: Would it be worth mentioning which filters?

miguelportilla (approved):
Nice improvement to the multi unit tests. Left some minor comments but not holding back.

src/test/unit_test/multi_runner.cpp
ximinez: Fixed and slightly optimized.

src/ripple/app/main/Main.cpp
ximinez: Oooh. I didn't think about the short-circuit. Good catch. I'll fix it so both functions are called every time.

src/ripple/app/main/Main.cpp
ximinez: I think it would be nice to mention which filters, but I was trying to keep these changes relatively simple, and my initial investigation into ways to either get the list of run tests or remove the run tests from the predicate was not simple, so I punted. I think that if it becomes enough of an issue, we can find a way to add it, but this seems sufficient for now. For now, it's easy enough to take the list of tests from the command line and rewrite it as a shell `for` loop to find the test that doesn't run anything.

PR #3604, opened 2020-09-09
Shard download space fix

undertome:

With this change, the `DatabaseShardImp` will confirm that sufficient space is available to store _all_ the shards queued for download, instead of confirming that each shard will fit without considering the additional space required by the others.

### Context of Change

This PR is stacked on two unmerged PRs. The relevant code starts with commit 6e1b74e.

### Type of Change

- [ X] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/src/ripple/net/HTTPStream.h
+++ b/src/ripple/net/HTTPStream.h
@@ -38,24 +38,14 @@ public:
 
     virtual ~HTTPStream() = default;
 
-    template <class T>
-    static std::unique_ptr<HTTPStream>
-    makeUnique(
-        Config const& config,
-        boost::asio::io_service::strand& strand,
-        beast::Journal j)
-    {
-        return std::make_unique<T>(config, strand, j);
-    }
-
     [[nodiscard]] virtual boost::asio::ip::tcp::socket&
     getStream() = 0;
 
     [[nodiscard]] virtual bool
     connect(
         std::string& errorOut,
-        std::string const host,
-        std::string const port,
+        std::string const& host,
+        std::string const& port,
         boost::asio::yield_context& yield) = 0;
 
     virtual void
@@ -68,7 +58,13 @@ public:
     asyncRead(
         boost::beast::flat_buffer& buf,
         parser& p,
-        bool readSome,
+        boost::asio::yield_context& yield,
+        boost::system::error_code& ec) = 0;
+
+    virtual void
+    asyncReadSome(
+        boost::beast::flat_buffer& buf,
+        parser& p,
         boost::asio::yield_context& yield,
         boost::system::error_code& ec) = 0;
 };
@@ -89,8 +85,8 @@ public:
     bool
     connect(
         std::string& errorOut,
-        std::string const host,
-        std::string const port,
+        std::string const& host,
+        std::string const& port,
         boost::asio::yield_context& yield) override;
 
     void
@@ -103,7 +99,13 @@ public:
     asyncRead(
         boost::beast::flat_buffer& buf,
         parser& p,
-        bool readSome,
+        boost::asio::yield_context& yield,
+        boost::system::error_code& ec) override;
+
+    void
+    asyncReadSome(
+        boost::beast::flat_buffer& buf,
+        parser& p,
         boost::asio::yield_context& yield,
         boost::system::error_code& ec) override;
 
@@ -117,10 +119,7 @@ private:
 class RawStream : public HTTPStream
 {
 public:
-    RawStream(
-        Config const& config,
-        boost::asio::io_service::strand& strand,
-        beast::Journal j);
+    RawStream(boost::asio::io_service::strand& strand);
 
     virtual ~RawStream() = default;
 
@@ -130,8 +129,8 @@ public:
     bool
     connect(
         std::string& errorOut,
-        std::string const host,
-        std::string const port,
+        std::string const& host,
+        std::string const& port,
         boost::asio::yield_context& yield) override;
 
     void
@@ -144,7 +143,13 @@ public:
     asyncRead(
         boost::beast::flat_buffer& buf,
         parser& p,
-        bool readSome,
+        boost::asio::yield_context& yield,
+        boost::system::error_code& ec) override;
+
+    void
+    asyncReadSome(
+        boost::beast::flat_buffer& buf,
+        parser& p,
         boost::asio::yield_context& yield,
         boost::system::error_code& ec) override;
 
--- a/src/ripple/net/impl/HTTPDownloader.cpp
+++ b/src/ripple/net/impl/HTTPDownloader.cpp
@@ -152,10 +152,10 @@ HTTPDownloader::do_session(
     //////////////////////////////////////////////
     // Prepare for download and establish the
     // connection:
-    std::uint64_t const rangeStart = size(p);
-
-    stream_ = ssl ? HTTPStream::makeUnique<SSLStream>(config_, strand_, j_)
-                  : HTTPStream::makeUnique<RawStream>(config_, strand_, j_);
+    if (ssl)
+        stream_ = std::make_unique<SSLStream>(config_, strand_, j_);
+    else
+        stream_ = std::make_unique<RawStream>(strand_);
 
     std::string error;
     if (!stream_->connect(error, host, port, yield))
@@ -165,6 +165,8 @@ HTTPDownloader::do_session(
     http::request<http::empty_body> req{http::verb::head, target, version};
     req.set(http::field::host, host);
     req.set(http::field::user_agent, BOOST_BEAST_VERSION_STRING);
+
+    std::uint64_t const rangeStart = size(p);
 
     // Requesting a portion of the file
     if (rangeStart)
@@ -182,7 +184,7 @@ HTTPDownloader::do_session(
         // Read the response
         http::response_parser<http::empty_body> connectParser;
         connectParser.skip(true);
-        stream_->asyncRead(read_buf_, connectParser, false, yield, ec);
+        stream_->asyncRead(read_buf_, connectParser, yield, ec);
         if (ec)
             return failAndExit("async_read", p);
 
@@ -198,7 +200,7 @@ HTTPDownloader::do_session(
             http::response_parser<http::empty_body> rangeParser;
             rangeParser.skip(true);
 
-            stream_->asyncRead(read_buf_, rangeParser, false, yield, ec);
+            stream_->asyncRead(read_buf_, rangeParser, yield, ec);
             if (ec)
                 return failAndExit("async_read_range_verify", p);
 
@@ -268,7 +270,7 @@ HTTPDownloader::do_session(
             return exit();
         }
 
-        stream_->asyncRead(read_buf_, *p, true, yield, ec);
+        stream_->asyncReadSome(read_buf_, *p, yield, ec);
     }
 
     JLOG(j_.trace()) << "download completed: " << dstPath.string();
--- a/src/ripple/net/impl/HTTPStream.cpp
+++ b/src/ripple/net/impl/HTTPStream.cpp
@@ -40,8 +40,8 @@ SSLStream::getStream()
 bool
 SSLStream::connect(
     std::string& errorOut,
-    std::string const host,
-    std::string const port,
+    std::string const& host,
+    std::string const& port,
     boost::asio::yield_context& yield)
 {
     using namespace boost::asio;
@@ -49,8 +49,10 @@ SSLStream::connect(
 
     boost::system::error_code ec;
 
-    auto fail = [&errorOut](std::string const& errorIn) {
-        errorOut = errorIn;
+    auto fail = [&errorOut, &ec](
+                    std::string const& errorIn,
+                    std::string const& message = "") {
+        errorOut = errorIn + ": " + (message.empty() ? ec.message() : message);
         return false;
     };
 
@@ -65,7 +67,7 @@ SSLStream::connect(
     }
     catch (std::exception const& e)
     {
-        return fail(std::string("exception: ") + e.what());
+        return fail("exception", e.what());
     }
 
     ec = ssl_ctx_.preConnectVerify(*stream_, host);
@@ -101,21 +103,23 @@ void
 SSLStream::asyncRead(
     boost::beast::flat_buffer& buf,
     parser& p,
-    bool readSome,
     boost::asio::yield_context& yield,
     boost::system::error_code& ec)
 {
-    if (readSome)
-        boost::beast::http::async_read_some(*stream_, buf, p, yield[ec]);
-    else
-        boost::beast::http::async_read(*stream_, buf, p, yield[ec]);
+    boost::beast::http::async_read(*stream_, buf, p, yield[ec]);
 }
 
-RawStream::RawStream(
-    Config const& config,
-    boost::asio::io_service::strand& strand,
-    beast::Journal j)
-    : strand_(strand)
+void
+SSLStream::asyncReadSome(
+    boost::beast::flat_buffer& buf,
+    parser& p,
+    boost::asio::yield_context& yield,
+    boost::system::error_code& ec)
+{
+    boost::beast::http::async_read_some(*stream_, buf, p, yield[ec]);
+}
+
+RawStream::RawStream(boost::asio::io_service::strand& strand) : strand_(strand)
 {
 }
 
@@ -129,8 +133,8 @@ RawStream::getStream()
 bool
 RawStream::connect(
     std::string& errorOut,
-    std::string const host,
-    std::string const port,
+    std::string const& host,
+    std::string const& port,
     boost::asio::yield_context& yield)
 {
     using namespace boost::asio;
@@ -138,8 +142,10 @@ RawStream::connect(
 
     boost::system::error_code ec;
 
-    auto fail = [&errorOut](std::string const& errorIn) {
-        errorOut = errorIn;
+    auto fail = [&errorOut, &ec](
+                    std::string const& errorIn,
+                    std::string const& message = "") {
+        errorOut = errorIn + ": " + (message.empty() ? ec.message() : message);
         return false;
     };
 
@@ -154,7 +160,7 @@ RawStream::connect(
     }
     catch (std::exception const& e)
     {
-        return fail(std::string("exception: ") + e.what());
+        return fail("exception", e.what());
     }
 
     boost::asio::async_connect(
@@ -178,14 +184,20 @@ void
 RawStream::asyncRead(
     boost::beast::flat_buffer& buf,
     parser& p,
-    bool readSome,
     boost::asio::yield_context& yield,
     boost::system::error_code& ec)
 {
-    if (readSome)
-        boost::beast::http::async_read_some(*stream_, buf, p, yield[ec]);
-    else
-        boost::beast::http::async_read(*stream_, buf, p, yield[ec]);
+    boost::beast::http::async_read(*stream_, buf, p, yield[ec]);
+}
+
+void
+RawStream::asyncReadSome(
+    boost::beast::flat_buffer& buf,
+    parser& p,
+    boost::asio::yield_context& yield,
+    boost::system::error_code& ec)
+{
+    boost::beast::http::async_read_some(*stream_, buf, p, yield[ec]);
 }
 
 }  // namespace ripple
--- a/src/ripple/nodestore/impl/DatabaseShardImp.cpp
+++ b/src/ripple/nodestore/impl/DatabaseShardImp.cpp
@@ -293,38 +293,31 @@ DatabaseShardImp::prepareLedger(std::uint32_t validLedgerSeq)
 }
 
 bool
-DatabaseShardImp::prepareShard(std::uint32_t shardIndex)
+DatabaseShardImp::prepareShards(std::vector<std::uint32_t> const& shardIndexes)
 {
-    auto fail = [j = j_, shardIndex](std::string const& msg) {
-        JLOG(j.error()) << "shard " << shardIndex << " " << msg;
+    auto fail = [j = j_, &shardIndexes](
+                    std::string const& msg,
+                    boost::optional<std::uint32_t> shardIndex = boost::none) {
+        auto multipleIndexPrequel = [&shardIndexes] {
+            std::vector<std::string> indexesAsString(shardIndexes.size());
+            std::transform(
+                shardIndexes.begin(),
+                shardIndexes.end(),
+                indexesAsString.begin(),
+                [](uint32_t const index) { return std::to_string(index); });
+
+            return std::string("shard") +
+                (shardIndexes.size() > 1 ? "s " : " ") +
+                boost::algorithm::join(indexesAsString, ", ");
+        };
+
+        std::string const prequel = shardIndex
+            ? "shard " + std::to_string(*shardIndex)
+            : multipleIndexPrequel();
+
+        JLOG(j.error()) << prequel << " " << msg;
         return false;
     };
-
-    if (shardIndex < earliestShardIndex())
-    {
-        return fail(
-            "comes before earliest shard index " +
-            std::to_string(earliestShardIndex()));
-    }
-
-    // If we are synced to the network, check if the shard index is
-    // greater or equal to the current or validated shard index.
-    auto seqCheck = [&](std::uint32_t ledgerSeq) {
-        if (ledgerSeq >= earliestLedgerSeq() &&
-            shardIndex >= seqToShardIndex(ledgerSeq))
-        {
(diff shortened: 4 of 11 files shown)


scottschurr:
Hi @undertome, it would be useful for me to have a little more background on this bug fix.  It would be handy to know:

- The user observed nature of the bug before the fix.
- The nature of the improved user experience after the bug fix.

Also it looks like the fix is associated with being low on resources.  Does the user have any recourse once they discover that they are too low on resources to handle the shards?

Thanks.

undertome:
> Hi @undertome, it would be useful for me to have a little more background on this bug fix. It would be handy to know:
> 
>     * The user observed nature of the bug before the fix.
> 
>     * The nature of the improved user experience after the bug fix.
> 
> 
> Also it looks like the fix is associated with being low on resources. Does the user have any recourse once they discover that they are too low on resources to handle the shards?
> 
> Thanks.

Hi Scott,

- Before the fix, the user would be able to initiate a batch of shard downloads when there was insufficient space to store all the shards, but sufficient space to store each shard in isolation. The user would see the first download complete, but subsequent downloads would fail.

- With the fix, a request to download a batch of shards will fail immediately if there is insufficient space to store all the shards and a descriptive log message will be created.

If the user is low on resources, the `DatabaseShardImp` will log a warning, discontinue acquiring shards, and reject requests to download shards. At this point, it would be the user's responsibility to provide more storage to accommodate additional shards. 

Hope that helps.

scottschurr:
A couple of your commit messages don't follow these guidelines: https://chris.beams.io/posts/git-commit/  May I suggest rewordings?  Consider:
```
Delay SSLHTTPDownloader dtor until current session finishes
```
and
```
Download queued shards only if there is space for them all
```
Or perhaps you can come up with something better.

src/ripple/nodestore/impl/DatabaseShardImp.cpp
```diff
@@ -303,62 +303,103 @@ DatabaseShardImp::prepareLedger(std::uint32_t validLedgerSeq)
...
+        };
+        if (!seqCheck(app_.getLedgerMaster().getValidLedgerIndex() + 1) ||
+            !seqCheck(app_.getLedgerMaster().getCurrentLedgerIndex()))
+        {
+            return false;
+        }
+
+        if (shards_.find(shardIndex) != shards_.end())
+        {
+            JLOG(j_.debug()) << "shard " << shardIndex
```
scottschurr, line 362: FWIW, not exercised in unit tests.

src/ripple/nodestore/impl/DatabaseShardImp.cpp
```diff
@@ -303,62 +303,103 @@ DatabaseShardImp::prepareLedger(std::uint32_t validLedgerSeq)
...
-            return fail("has an invalid index");
-        return true;
-    };
-    if (!seqCheck(app_.getLedgerMaster().getValidLedgerIndex() + 1) ||
-        !seqCheck(app_.getLedgerMaster().getCurrentLedgerIndex()))
+    auto const numHistShards = numHistoricalShards(lock);
+
+    // Check shard count and available storage space
+    if (numHistShards + historicalShardsToPrepare > maxHistoricalShards_)
+        return fail("maximum number of historical shards reached");
```
scottschurr, line 377: Not exercised in unit tests.

src/ripple/nodestore/impl/DatabaseShardImp.cpp
```diff
@@ -303,62 +303,103 @@ DatabaseShardImp::prepareLedger(std::uint32_t validLedgerSeq)
...
+
+    if (historicalShardsToPrepare)
     {
-        return false;
+        // Check available storage space for historical shards
+        if (!sufficientStorage(
+                historicalShardsToPrepare, PathDesignation::historical, lock))
+            return fail("insufficient storage space available");
```
scottschurr, line 384: Not exercised in unit tests.

src/ripple/nodestore/impl/DatabaseShardImp.cpp
```diff
@@ -303,62 +303,103 @@ DatabaseShardImp::prepareLedger(std::uint32_t validLedgerSeq)
...
+            shardIndexes.size() - historicalShardsToPrepare;
+        recentShardsToPrepare)
     {
-        JLOG(j_.debug()) << "shard " << shardIndex
-                         << " is already stored or queued for import";
-        return false;
+        // Check available storage space for recent shards
+        if (!sufficientStorage(
+                recentShardsToPrepare, PathDesignation::none, lock))
+            return fail("insufficient storage space available");
```
scottschurr, line 394: Not exercised in unit tests.

scottschurr (approved):
:+1: Looks good.  I left notes about a few things for you to think about.  Nothing critical, although covering a few more of the error cases in the unit tests would be really nice.  All comments are at your discretion.  Thanks for making life a little easier for ledger users!

src/ripple/nodestore/impl/DatabaseShardImp.cpp
```diff
@@ -303,62 +303,103 @@ DatabaseShardImp::prepareLedger(std::uint32_t validLedgerSeq)
...
+        std::transform(
+            shardIndexes.begin(),
+            shardIndexes.end(),
+            indexesAsString.begin(),
+            [](uint32_t const index) { return std::to_string(index); });
+
+        std::string const prequel = shardIndex
+            ? "shard " + std::to_string(*shardIndex)
+            : std::string("shard") + (shardIndexes.size() > 1 ? "s " : " ") +
+                boost::algorithm::join(indexesAsString, ", ");
```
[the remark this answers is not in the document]
undertome: Done

src/ripple/net/impl/HTTPStream.cpp
```diff
@@ -0,0 +1,191 @@
...
+{
+    assert(stream_);
+    return stream_->next_layer();
+}
+
+bool
+SSLStream::connect(
+    std::string& errorOut,
+    std::string const host,
+    std::string const port,
```
miguelportilla, line 44: `host` and `port` can be const references.

src/ripple/net/impl/HTTPStream.cpp
```diff
@@ -0,0 +1,191 @@
...
+{
+    assert(stream_);
+    return *stream_;
+}
+
+bool
+RawStream::connect(
+    std::string& errorOut,
+    std::string const host,
+    std::string const port,
```
miguelportilla, line 133: `host` and `port` can be const references.

src/ripple/net/HTTPStream.h
```diff
@@ -0,0 +1,158 @@
...
+    }
+
+    [[nodiscard]] virtual boost::asio::ip::tcp::socket&
+    getStream() = 0;
+
+    [[nodiscard]] virtual bool
+    connect(
+        std::string& errorOut,
+        std::string const host,
+        std::string const port,
```
miguelportilla, line 58: `host` and `port` can be const references.

src/ripple/net/HTTPStream.h
```diff
@@ -0,0 +1,158 @@
...
+    virtual ~SSLStream() = default;
+
+    boost::asio::ip::tcp::socket&
+    getStream() override;
+
+    bool
+    connect(
+        std::string& errorOut,
+        std::string const host,
+        std::string const port,
```
miguelportilla, line 93: `host` and `port` can be const references.

src/ripple/net/HTTPStream.h
```diff
@@ -0,0 +1,158 @@
...
+    virtual ~RawStream() = default;
+
+    boost::asio::ip::tcp::socket&
+    getStream() override;
+
+    bool
+    connect(
+        std::string& errorOut,
+        std::string const host,
+        std::string const port,
```
miguelportilla, line 134: `host` and `port` can be const references.

src/ripple/net/impl/HTTPStream.cpp
```diff
@@ -0,0 +1,191 @@
...
+    assert(stream_);
+    return stream_->next_layer();
+}
+
+bool
+SSLStream::connect(
+    std::string& errorOut,
+    std::string const host,
+    std::string const port,
+    boost::asio::yield_context& yield)
```
miguelportilla, line 45: We've lost `ec` error message loging. Consider passing in an `ec` ref as done in the other functions.

src/ripple/net/impl/HTTPDownloader.cpp
```diff
@@ -161,36 +166,12 @@ SSLHTTPDownloader::do_session(
     // connection:
     std::uint64_t const rangeStart = size(p);
```
miguelportilla, line 155: This local should be declared after the `connect` call.

src/ripple/net/HTTPStream.h
```diff
@@ -0,0 +1,158 @@
...
+    boost::asio::io_service::strand& strand_;
+};
+
+class RawStream : public HTTPStream
+{
+public:
+    RawStream(
+        Config const& config,
+        boost::asio::io_service::strand& strand,
+        beast::Journal j);
```
miguelportilla, line 123: `config` and `journal` are unused. These variable should be remove from the ctor. I recommend eliminating the `makeUnique` and modifying `HTTPDownloader` as such.
eg.
```
    if (ssl)
        stream_ = std::make_unique<SSLStream>(config_, strand_, j_);
    else
        stream_ = std::make_unique<RawStream>(strand_);
```

src/ripple/net/impl/HTTPStream.cpp
```diff
@@ -0,0 +1,191 @@
...
+    using namespace boost::beast;
+
+    boost::system::error_code ec;
+
+    auto fail = [&errorOut](std::string const& errorIn) {
+        errorOut = errorIn;
+        return false;
+    };
+
+    ip::tcp::resolver resolver{strand_.context()};
```
miguelportilla, line 57: Can this ctor throw?

src/ripple/net/impl/HTTPStream.cpp
```diff
@@ -0,0 +1,191 @@
...
```
miguelportilla, line 146: Can this ctor throw?

src/ripple/net/impl/HTTPStream.cpp
```diff
@@ -0,0 +1,191 @@
...
+    parser& p,
+    bool readSome,
+    boost::asio::yield_context& yield,
+    boost::system::error_code& ec)
+{
+    if (readSome)
+        boost::beast::http::async_read_some(*stream_, buf, p, yield[ec]);
+    else
+        boost::beast::http::async_read(*stream_, buf, p, yield[ec]);
+}
```
miguelportilla, line 189: Consider adding `HTTPStream::ayncReadSome` and removing the `::asyncRead` `readSome` params.

src/ripple/nodestore/impl/DatabaseShardImp.cpp
```diff
@@ -303,62 +303,103 @@ DatabaseShardImp::prepareLedger(std::uint32_t validLedgerSeq)
...
+    auto historicalShardsToPrepare = 0;
 
-    if (shardIndex < earliestShardIndex())
+    for (auto const shardIndex : shardIndexes)
     {
-        return fail(
-            "comes before earliest shard index " +
-            std::to_string(earliestShardIndex()));
+        if (!canAdd_)
+            return fail("cannot be stored at this time");
```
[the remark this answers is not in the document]
undertome: Added unit tests to exercise additional paths

src/ripple/nodestore/impl/DatabaseShardImp.cpp
undertome: Added unit tests to exercise additional paths

src/ripple/nodestore/impl/DatabaseShardImp.cpp
undertome: Added unit tests to exercise additional paths

src/ripple/nodestore/impl/DatabaseShardImp.cpp
undertome: Added unit tests to exercise additional paths, but this specific error condition is not triggered.

src/ripple/nodestore/impl/DatabaseShardImp.cpp
undertome: Added unit tests to exercise additional paths, but this specific error condition is not triggered.

src/ripple/nodestore/impl/DatabaseShardImp.cpp
```diff
@@ -303,62 +303,103 @@ DatabaseShardImp::prepareLedger(std::uint32_t validLedgerSeq)
...
+            return fail(
+                "comes before earliest shard index " +
+                    std::to_string(earliestShardIndex()),
+                shardIndex);
+        }
+
+        // If we are synced to the network, check if the shard index
+        // is greater or equal to the current shard.
+        auto seqCheck = [&](std::uint32_t seq) {
+            // seq will be greater than zero if valid
```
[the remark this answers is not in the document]
undertome: Lines 345 and 346 still seem relevant to me - but maybe line 348 should say ">= earliest ledger sequence".

scottschurr:
FWIW, I was running the unit tests for code coverage on this branch using macOS and I got this error message during the test run:
```
ERR:ShardArchiveHandler async_connect
ERR:ShardArchiveHandler Downloading shard id 10 from URL 127.0.0.1/10.tar.lz4
rippled(2843,0x13c67c5c0) malloc: *** error for object 0x13eed4000: pointer being freed was not allocated
rippled(2843,0x13c67c5c0) malloc: *** set a breakpoint in malloc_error_break to debug
make[3]: *** [CMakeFiles/coverage_report] Abort trap: 6
make[2]: *** [CMakeFiles/coverage_report.dir/all] Error 2
make[1]: *** [CMakeFiles/coverage_report.dir/rule] Error 2
make: *** [coverage_report] Error 2
```
It did not leave a stack trace behind.  I'll see if I can reproduce the problem.

scottschurr:
Not surprisingly the bug is not easy to reproduce.

- Running the unit tests alone (once) did not recreate it.
- Running code coverage on just `ShardArchiveHandler` did not reproduce it.
- Running all of the unit tests under code coverage again did not reproduce it.

I'm going to start a loop running the `ShardArchiveHandler` test to see if I can reproduce the bug with enough repetitions.

scottschurr (approved):
Thanks for the improved unit test coverage.

I'm guessing the non-reproducible unit test failure that I saw is not because of these changes.

scottschurr:
I continued running a test over the weekend to try and get more information about the crash observed on Friday.  I ran just the ShardArchiveHandler unit test in a loop.  After 5928 iterations I got a different crash from the originally reported crash.  Here's link to the stack trace of this most recent crash in case it's useful: https://gist.github.com/scottschurr/cec16dc035d9ec83fe61e6411cee0ddf

scottschurr:
Success!  By running just the ShardArchiveHandler under code coverage in a loop I was able to reproduce the original symptom quickly.  And this time I got a stack trace.  On the console I see:
```
ERR:ShardArchiveHandler Downloading shard id 9 from URL 127.0.0.1/9.tar.lz4
ERR:ShardArchiveHandler async_connect
ERR:ShardArchiveHandler Downloading shard id 10 from URL 127.0.0.1/10.tar.lz4
rippled(2685,0x148d915c0) malloc: *** error for object 0x13f548000: pointer being freed was not allocated
rippled(2685,0x148d915c0) malloc: *** set a breakpoint in malloc_error_break to debug
make[3]: *** [CMakeFiles/coverage_report] Abort trap: 6
make[2]: *** [CMakeFiles/coverage_report.dir/all] Error 2
make[1]: *** [CMakeFiles/coverage_report.dir/rule] Error 2
make: *** [coverage_report] Error 2
```
You can find the stack trace here: https://gist.github.com/scottschurr/f43be7a61290562738728bb4b668a6e8

undertome:
> I continued running a test over the weekend to try and get more information about the crash observed on Friday. I ran just the ShardArchiveHandler unit test in a loop. After 5928 iterations I got a different crash from the originally reported crash. Here's link to the stack trace of this most recent crash in case it's useful: https://gist.github.com/scottschurr/cec16dc035d9ec83fe61e6411cee0ddf

Are you sure you're running code from this branch? That stack trace references `SSLHTTPDownloader` which no longer exists.

scottschurr:
> Are you sure you're running code from this branch? That stack trace references SSLHTTPDownloader which no longer exists.

That's a fair question.  Here's the git log of the branch I've been testing:

```
commit fb1ca671319fcfdbeb3943f5d1fc554c3d7d230c (HEAD -> devon-shard-download-space-fix, devon/shard-download-space-fix)
Author: Devon White <dwhite@ripple.com>
Date:   Thu Sep 17 17:26:02 2020 -0400

    [FOLD] Clang format patch

commit 09827143f62683907745a23842f6876d3612ade4
Author: Devon White <dwhite@ripple.com>
Date:   Thu Sep 17 17:21:28 2020 -0400

    [FOLD] Add unit tests

commit f9b9647a5bd6b47133f2e58e1eb6585944c2d0fa
Author: Devon White <dwhite@ripple.com>
Date:   Wed Sep 9 17:15:13 2020 -0400

    Download queued shards only if there is space for them all

commit e31c02ba89b6b7727eb2a8246d3d4ad4f227b104
Author: Devon White <dwhite@ripple.com>
Date:   Fri Aug 14 15:45:56 2020 -0400

    Delay HTTPDownloader dtor until current session
    finishes

commit 649d4a741e112777575e9f9d10100c6588037d08
Author: Devon White <dwhite@ripple.com>
Date:   Tue Aug 11 12:10:10 2020 -0400

    Allow shard downloads via HTTP or HTTPS

commit a8d481c2a54267d80bd539003f56481424b9ae63 (upstream/develop, origin/develop, develop)
Author: Nik Bougalis <nikb@bougalis.net>
Date:   Tue Sep 1 10:43:26 2020 -0700

    Set version to 1.7.0-b1
```

scottschurr:
That said, even though that's the source code currently in place, it's possible (but seems unlikely) that I did not do a proper build before running the ShardArchiveHandler unit test in a loop over the weekend.  If that's the case please accept my apologies.

For the `malloc` failure under Codecov, however, I'm pretty confident that I have both the correct sources and that those are the source files I'm running.  The rebuild is part of how Codecov is invoked.

undertome:
> That said, even though that's the source code currently in place, it's possible (but seems unlikely) that I did not do a proper build before running the ShardArchiveHandler unit test in a loop over the weekend. If that's the case please accept my apologies.
> 
> For the `malloc` failure under Codecov, however, I'm pretty confident that I have both the correct sources and that those are the source files I'm running. The rebuild is part of how Codecov is invoked.

The `malloc` failure seems current.

src/ripple/net/HTTPStream.h
undertome: Done

src/ripple/net/HTTPStream.h
undertome: Done

src/ripple/net/HTTPStream.h
undertome: Done

src/ripple/net/impl/HTTPDownloader.cpp
undertome: Done

src/ripple/net/impl/HTTPStream.cpp
undertome: Passing the error code to the write and read functions is easy because in each case, there's only one function that could be the cause of an error, whereas in the connect functions, there are many functions that could cause an error. So to provide info to the caller about which function errored, we'd have to provide additional context with the `ec` parameter. Instead of that I'd rather just provide the context (`errorOut`) without the additional `ec` parameter.

src/ripple/net/HTTPStream.h
undertome: Done

miguelportilla:
I took a quick look at the stack @scottschurr produced and observed the following. The crash comes from within a journal. From `ShardArchiveHandler_test::testRedundantShardFailure`, on line 612, the test has only created `Env` called `close()` which calls `RPCCall::fromNetwork`. A possible culprit I see is that inside `fromNetwork` a local journal `j` is declared and passed into the callbacks and the HTTPClient object. This might be a problem as `HTTPClient::request` is non blocking and `j`'s sink object may have gone out of scope when the callbacks or `invokeComplete` are executed. Its possible the sink may be still be valid at this point as it might be kept alive by someone but I am not sure.

undertome:
> I took a quick look at the stack @scottschurr produced and observed the following. The crash comes from within a journal. From `ShardArchiveHandler_test::testRedundantShardFailure`, on line 612, the test has only created `Env` called `close()` which calls `RPCCall::fromNetwork`. A possible culprit I see is that inside `fromNetwork` a local journal `j` is declared and passed into the callbacks and the HTTPClient object. This might be a problem as `HTTPClient::request` is non blocking and `j`'s sink object may have gone out of scope when the callbacks or `invokeComplete` are executed. Its possible the sink may be still be valid at this point as it might be kept alive by someone but I am not sure.

Thanks for looking into it, I will see if this checks out.

src/ripple/net/impl/HTTPStream.cpp
miguelportilla: While reporting "which function errored" may be helpful, we are throwing away the cause. Typically if the `ec` is set, further progress on the process is aborted, the call stack is unwind and the `ec` is floated up. If we choose to keep things as they, it may be helpful to append its message into `errorOut`. 

eg.
```return fail("async_resolve. Error: " + ec.message());```

As is, if any of these functions fail, we don't know why.

Comment (miguelportilla) [reply src/ripple/net/impl/HTTPStream.cpp:45]: For example, if the call to `resolver.async_resolve` fails, `errorOut` will contain "async_resolve" but the reason it failed will be discarded and we will have no idea why it failed.

Comment (undertome) [reply src/ripple/net/impl/HTTPStream.cpp:45]: Done

File: src/ripple/net/impl/HTTPStream.cpp [hunk 19]
```diff
@@ -65,7 +65,7 @@ SSLStream::connect(
     }
     catch (std::exception const& e)
     {
-        return fail(std::string("exception: ") + e.what());
+        return fail("exception");
```
miguelportilla, line 68: If `ec` isn't set, we'd end up with an error message that says "exception".  I recommend keeping the exception message.

src/ripple/net/impl/HTTPStream.cpp
```diff
@@ -156,7 +156,7 @@ RawStream::connect(
...
```
miguelportilla, line 159: See prior comment.

src/ripple/net/impl/HTTPStream.cpp
```diff
@@ -49,8 +49,8 @@ SSLStream::connect(
 
     boost::system::error_code ec;
 
-    auto fail = [&errorOut](std::string const& errorIn) {
-        errorOut = errorIn;
+    auto fail = [&errorOut, &ec](std::string const& errorIn) {
+        errorOut = errorIn + ": " + ec.message();
```
miguelportilla, line 53: If ec is not set, we append ": ". It may be a good ide to check ec before appending it.
eg.
```
        errorOut = errorIn;
        if (ec)
            errorOut += ". Error code: " + ec.message();
```

src/ripple/net/impl/HTTPStream.cpp
```diff
@@ -140,8 +140,8 @@ RawStream::connect(
...
```
miguelportilla, line 144: See prior comment.

src/ripple/net/impl/HTTPStream.cpp
undertome: Done

src/ripple/net/impl/HTTPStream.cpp
undertome: Done

src/ripple/net/impl/HTTPStream.cpp
undertome: Got this done with an additional defaulted parameter to the lambda

src/ripple/net/impl/HTTPStream.cpp
undertome: Done

miguelportilla (approved):
👍

undertome:
New commit fixes a Travis error caused by an "unused" (it's used in an `assert` statement) variable. There was also a bug I discovered that was caused by the `Checkpointer` outliving a `DatabaseCon`.

miguelportilla (approved):

PR #3610, opened 2020-09-17
Improve Windows and MacOS test performance in Travis CI:

ximinez:
* Increases hard-coded number of parallel unit test processes for
  Windows and MacOS builds from 1 to 2.
* Reduces Travis job time to well under the timeout value of 1.5 hours.
* Continue using a hard-coded value rather than `nprocs` because higher
  values cause some jobs to run out of memory.

Windows debug builds in Travis timeout more and more often, especially in personal branches. Prior to this change, the Windows and MacOS builds override the number of parallel test jobs (`--unittest-jobs`) to only use 1, instead of the number of available processors (`nproc`). This PR changes the override to 2 jobs, which decreases the job time enough to not time out, and have a decent amount of time left over. (Though we do need to put some work into decreasing our unit test run times, that's a separate issue.)

More than 2 test processes causes these same Windows debug jobs to have OOM issues and fail for that reason, so 2 "threads the needle" or is in the "Goldilocks zone" to keep everything working for now.

### Context of Change

Travis CI is resource limited, and has a maximum job time that Windows debug builds, especially in personal branches, frequently runs in to.

### Type of Change

## Test Plan

Compare
* https://travis-ci.com/github/ximinez/rippled/builds/183923361
which failed because two of the test jobs timed out after 1 hour 30 minutes and
* https://travis-ci.com/github/ximinez/rippled/builds/184732255
which succeeded.

---
This change is [Reviewable](https://reviewable.io/reviews/ripple/rippled/3610)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -438,9 +438,7 @@ cache:
     - $CACHE_DIR
 
 before_install:
-  # NUM_PROCESSORS was restricted due to problems in parallel launch of
-  # unit tests on Mac platform
-  - if [ "$(uname)" = "Darwin" ] ; then export NUM_PROCESSORS=2; else export NUM_PROCESSORS=$(nproc); fi
+  - export NUM_PROCESSORS=$(nproc)
   - echo "NUM PROC is ${NUM_PROCESSORS}"
   - if [ "$(uname)" = "Linux" ] ; then docker pull ${DOCKER_IMAGE}; fi
   - if [ "${MATRIX_EVAL}" != "" ] ; then eval "${MATRIX_EVAL}"; fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -438,7 +438,8 @@ cache:
     - $CACHE_DIR
 
 before_install:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -439,7 +439,7 @@ cache:
 
 before_install:
   # NUM_PROCESSORS was set to 1 due to problems in parallel launch of unit tests on Mac platform
```
MarkusTeufelberger, line 441: This comment doesn't seem to apply any more.

Since it isn't related to the issue on Windows, maybe this restriction isn't needed at all any more?

ximinez: I pushed two commits. The first just fixes the comment. The second removes the MacOS restriction. If it works, I'll keep both, else I'll drop the second one.

gregtatcam (approved):
LGTM

.travis.yml
ximinez: Looks like it worked: https://travis-ci.com/github/ximinez/rippled/builds/184981619

miguelportilla (approved):
LGTM 👍

Issue #3611, opened 2020-09-17
Move manifest API call to the public API


tlongwell-ripple:
## Summary
Move the Manifest call (https://xrpl.org/manifest.html) from the admin API to the public API

## Motivation
This endpoint is used by the XRPL Explorer application. By moving the Manifest endpoint to the public API, it will be easier and safer to run the explorer since it will not require admin API access to a rippled node.

intelliot:
Done in rippled 1.7.0-b2 https://github.com/ripple/rippled/pull/3612

PR #3612, opened 2020-09-18
Update manifest RPC to be public

carlhua:
Trivial change to update the manifest RPC to be a public method.

### Context of Change

This change allows the access of manifest without an admin connection

---
This change is [Reviewable](https://reviewable.io/reviews/ripple/rippled/3612)


seelabs (approved):
:+1: LGTM

intelliot:
Fixes #3611

intelliot (approved):
It's not _necessarily_ trivial since we do need to make sure that there can't be any confidential info returned by the command, and that it can't be abused by non-admins. But I agree, it looks fine in this case.

PR #3615, opened 2020-09-22
Update shard db sections in the example config

undertome:

Update the example config file to reflect recent changes to the shard database.

### Context of Change

The example config file has had obsolete keys removed and new sections/keys added.

### Type of Change

- [x ] Documentation Updates

### DIFF
--- a/cfg/rippled-example.cfg
+++ b/cfg/rippled-example.cfg
@@ -385,7 +385,7 @@
 #
 #   [ips]
 #   192.168.0.1
-#   192.168.0.1 2459 
+#   192.168.0.1 2459
 #   r.ripple.com 51235
 #
 #
@@ -458,7 +458,7 @@
 #   Configure the maximum number of transactions to have in the job queue
 #
 #   Must be a number between 100 and 1000, defaults to 250
-# 
+#
 #
 # [overlay]
 #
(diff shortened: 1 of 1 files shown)


cfg/rippled-example.cfg
```diff
@@ -1359,7 +1365,14 @@ advisory_delete=0
 # https://xrpl.org/history-sharding.html
 #[shard_db]
 #path=/var/lib/rippled/db/shards/nudb
-#max_size_gb=500
```
MarkusTeufelberger, line 1362: Was this setting removed or just renamed?

undertome: I'd say it was removed since `max_historical_shards` has a different meaning.

miguelportilla (approved):
👍

cfg/rippled-example.cfg
MarkusTeufelberger: Then having "500" there is going to be unexpectedly large for a lot of people.

Also `max_historical_shards` isn't documented around line 960, only here.

Lastly having a section in the "actual" configuration section yet completely commenting it out seems like a strange choice. If you're only concerned about giving an example on how to configure the `[shard_db]` section, you already have the (nearly) full documentation in line ~950. If this is really that important for the network health, how about uncommenting this section and storing let's say a dozen historic shards by default?

undertome: Set the commented value for `max_historical_shards` to 50 and added documentation for this optional field. There's already a precedent for providing commented-out sections as starter templates so I'm inclined to leave that as is.

mDuo13 (approved):

PR #3619, opened 2020-10-05
Support UNLs with future effective dates:

### LINKED ISSUE #3470: Validator List Expiration Improvements
Some proposed improvements to internal behavior when our validator list has expired:

1. Make sure that the server generates an event log at FATAL and enters a fail-safe mode so as to not provide incorrect or outdated information; client queries should fail gracefully. (similar to amendment blocking - but needs more investigation)
2. Avoid processing any transactions received from peers, so as to not clutter the log files.

Currently, when the validator list expires, an ERROR (instead of FATAL) message is logged and the validator considers itself not synced and will not send validation message for ledgers. 

If a newer list is retrieved, the server should log a message at INFO level, resync to the network and resume normal operation.

Exported from RIPD-1757

### LINKED ISSUE #3548: Allow prepublishing of a UNL with a future activation date
Validator lists currently take immediate effect; that is, as soon as a list with a sequence number greater than the one presently configured is loaded becomes the authoritative list.

Ideally, a validator list should include a new field, tentatively named effective_date, which specifies when the list's validity period should begin.

Servers should be able to track multiple lists from a single source, each with a different sequence number. Then, as the server operates, the server should use the close time of the last validated ledger to decide whether a list with an effective_date should be activated.

The list format should be augmented to allow multiple lists, each with its own signature, to be contained within a single file.

If multiple lists with different sequence numbers but the same effective_date are retrieved, the server should silently choose the list with the highest sequence number.

If multiple lists with the same sequence number are retrieved but they have different effective_date values, the server should generate a diagnostic at ERROR level and use the list with the earliest effective date.

Exported from RIPD-1660

ximinez:

* Creates a version 2 of the UNL file format allowing publishers to
  pre-publish the next UNL while the current one is still valid.
* Version 1 of the UNL file format is still valid and backward
  compatible.
* Also causes rippled to lock down if it has no valid UNLs, similar to
  being amendment blocked, except reversible.
* Resolves #3548
* Resolves #3470

### Context of Change

[Internal Spec](https://github.com/ripple/rippled-specs/tree/master/0012-prepublish-and-expired-unls)

### Type of Change

## Test Plan

Set up an internal site similar to https://vl.ripple.com. Publish a version 2 UNL file on that site, as described in the spec. (Replace `blob` and `signature` with a `blobs_v2` array containing one or more blob/signature pairs.) Include one or more VLs in the array containing future effective dates (using Ripple-epoch time value). Verify that nodes switch to the new VL at the appropriate time.

## Future Tasks

Some relatively minor deviations and additions were made from the spec in the development process. I'll update the spec ASAP.

---
This change is [Reviewable](https://reviewable.io/reviews/ripple/rippled/3619)


nbougalis (commented):
Partial review, some quick comments. Will follow-up.

src/ripple/app/misc/ValidatorList.h
```diff
@@ -596,6 +816,63 @@ class ValidatorList
...
+{
+    for (auto const& item : blobs)
+        hash_append(h, item);
+}
+
+template <class Hasher>
+void
+hash_append(Hasher& h, std::map<std::size_t, ValidatorBlobInfo> const& blobs)
+{
+    for (auto const& [_, item] : blobs)
```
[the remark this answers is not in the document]
seelabs: `std::ignore` doesn't work with structured bindings. (I'd *love* if we got a standard "ignore" variable like `_` in other languages).

ximinez:
Rebased on to 1.7.0-b3

src/ripple/app/misc/ValidatorList.h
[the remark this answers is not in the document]
nbougalis: Oh right, it's only for `std::tie` :man_facepalming:

src/ripple/app/misc/impl/ValidatorList.cpp
```diff
@@ -177,15 +187,16 @@ ValidatorList::load(
             JLOG(j_.warn()) << "Duplicate node identity: " << match[1];
             continue;
         }
-        auto it = publisherLists_.emplace(
+        auto [it, inserted] = publisherLists_.emplace(
```
undertome, line 190: I have a few observations - it seems like the variable `local` is unnecessary as it isn't being modified to reflect information from `configKeys`. Additionally, if more than one validator is listed under the `[validators]` config section, only one entry will be added to `publisherLists_` because of the way `local` is being used. In any case, the entry added to `publisherLists_` won't reflect the key(s) present in the `validators` config section.

undertome:
The node crashes for me when I provide validators under the `validators` config section.

src/ripple/app/misc/impl/ValidatorList.cpp
```diff
@@ -195,29 +206,88 @@ ValidatorList::load(
...
+            {
+                (void)_;
+                add(pending);
+            }
+
+            value[jss::blobs_v2] = std::move(blobs);
+            break;
+        }
+        default:
+            value = Json::nullValue;
```
undertome, line 268: Am I correct in assuming that we should never hit this case? Maybe its worth logging.

src/ripple/app/misc/impl/ValidatorList.cpp
```diff
@@ -195,29 +206,88 @@ ValidatorList::load(
...
+ValidatorList::buildFileData(
+    std::string const& pubKey,
+    ValidatorList::PublisherListCollection const& pubCollection,
+    boost::optional<std::uint32_t> forceVersion /* = {} */)
+{
+    Json::Value value(Json::objectValue);
+
+    auto const effectiveVersion = forceVersion
+        ? *forceVersion
+        : (pubCollection.remaining.empty() ? pubCollection.rawVersion : 2);
```
undertome, line 225: Is there a reason for not simply using `pubCollection.rawVersion` whenever `forceVersion` is unseated? It seems a bit simpler and I'm not sure of why we'd opt for the downgraded format since the goal is to eventually migrate to version 2.

src/ripple/app/misc/impl/ValidatorList.cpp
```diff
@@ -229,152 +299,653 @@ ValidatorList::CacheValidatorFile(
...
+        blob.set_signature(blobInfo.signature);
+        if (blobInfo.manifest)
+            blob.set_manifest(*blobInfo.manifest);
+    }
+    assert(msg.blobs_size() > 0);
+    auto message =
+        std::make_shared<Message>(msg, protocol::mtVALIDATORLISTCOLLECTION);
+    if (message->getBuffer(compression::Compressed::Off).size() > maxSize)
+    {
+        // throw away the Message and split into smaller messages
```
undertome, line 542: Seems like you can avoid creating `message` in this case by comparing the sum of `msg.ByteSize()` and `compression::headerBytes` with `maxSize` and creating `message` only if it will be sent as a single message.

src/ripple/app/misc/impl/ValidatorList.cpp
```diff
@@ -229,152 +299,653 @@ ValidatorList::CacheValidatorFile(
...
+        boost::optional<protocol::TMValidatorListCollection> smallMsg;
+        smallMsg.emplace();
+        smallMsg->set_version(largeMsg.version());
+        smallMsg->set_manifest(largeMsg.manifest());
+
+        for (std::size_t i = begin; i < end; ++i)
+        {
+            *smallMsg->add_blobs() = largeMsg.blobs(i);
+        }
+        auto message = std::make_shared<Message>(
```
undertome, line 461: See [this comment](https://github.com/ripple/rippled/pull/3619/files#r507995736).

src/ripple/app/misc/impl/ValidatorList.cpp
```diff
@@ -229,152 +299,653 @@ ValidatorList::CacheValidatorFile(
...
+            smallMsg.set_manifest(blob.manifest());
+
+        auto message =
+            std::make_shared<Message>(smallMsg, protocol::mtVALIDATORLIST);
+
+        if (message->getBuffer(compression::Compressed::Off).size() <= maxSize)
+        {
+            messages.emplace_back(message, sha512Half(smallMsg), 1);
+            return messages.back().numVLs;
+        }
```
undertome, line 448: Suppose we land in the "implied" else clause (the message was too large) for some of the _VLs_ in a message, while other _VLs_ from the same original message were successfully appended to `messages`. In this case we can end up sending an incomplete collection of validator lists. Is this a scenario we care about? Based on the implementation it seems that we're opting to silently send an incomplete message, rather than failing loudly, or somehow denoting the deficiency in the outgoing message.

src/ripple/app/misc/impl/ValidatorList.cpp
undertome: In order to conveniently get the message size you can add a static function to the `Message` class that encapsulates [this logic](https://github.com/ximinez/rippled/blob/6ebf80b80734867cb7781d021838b88f28988f17/src/ripple/overlay/impl/Message.cpp#L35).

src/ripple/app/misc/impl/ValidatorList.cpp
```diff
@@ -229,152 +299,653 @@ ValidatorList::CacheValidatorFile(
...
+    auto const manifest =
+        currentBlob.manifest ? *currentBlob.manifest : rawManifest;
+    auto const version = 1;
+    msg.set_manifest(manifest);
+    msg.set_blob(currentBlob.blob);
+    msg.set_signature(currentBlob.signature);
+    // Override the version
+    msg.set_version(version);
+
+    auto message = std::make_shared<Message>(msg, protocol::mtVALIDATORLIST);
```
undertome, line 501: See [this comment](https://github.com/ripple/rippled/pull/3619/files#r507995736).

src/ripple/app/misc/NetworkOPs.cpp
```diff
@@ -1561,6 +1579,13 @@ NetworkOPsImp::setAmendmentBlocked()
     setMode(OperatingMode::TRACKING);
 }
 
+void
+NetworkOPsImp::setExpiredValidatorList()
+{
+    expiredValidatorList_ = true;
+    setMode(OperatingMode::TRACKING);
```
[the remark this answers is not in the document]
ximinez: Yes, that makes sense. I was paralleling amendment blocked, but this is a good place for a difference.

src/ripple/app/misc/ValidatorList.h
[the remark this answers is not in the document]
ximinez: Yeah, I definitely wish there was a more elegant way to do this...

src/ripple/app/main/Application.cpp
```diff
@@ -2143,6 +2143,12 @@ ApplicationImp::serverOkay(std::string& reason)
         return false;
     }
 
+    if (getOPs().isExpiredValidatorList())
```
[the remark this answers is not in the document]
ximinez: Hm. 
* `hasExpiredValidatorList`
* `isAnyValidatorListExpired`
* `needUpdatedUNL`
* `isConsensusBlocked`
Any preferences or suggestions?

src/ripple/app/misc/ValidatorList.h
```diff
@@ -26,40 +26,80 @@
...
 #include <ripple/protocol/PublicKey.h>
 #include <boost/iterator/counting_iterator.hpp>
 #include <boost/range/adaptors.hpp>
 #include <mutex>
 #include <numeric>
 #include <shared_mutex>
 
+namespace protocol {
+// predeclaration
+class TMValidatorList;
```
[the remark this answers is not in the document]
ximinez: I only did this so I could reference these classes in some function declarations below without needing to `#include` the entire `protocol/messages.h`. If there's a better way, I'm all ears.

src/ripple/app/misc/impl/ValidatorList.cpp
ximinez: You're partially right - I was able to get rid of `local` entirely, and `emplace` default objects. The reason it's not modified is because it is used as a "pseudo master key" for a UNL _signer_, not a validator. The validator keys are iterated with `n`, which is used to make `id`. On repeated calls, the `emplace` function simply returns the existing entry. So the first call uses a default `PublicKey`, and creates a new `PublisherListCollection`. Then `id` is added to the `PublisherListCollection`'s current list. On later calls, `emplace` returns the same `PublisherListCollection` that was created the first time, so `id` is added to the list. At the end of the process, the `list` contains all the `PublicKeys` corresponding to `configKeys`. You can look at `testConfigLoad` in ValidatorList_test.cpp to see this working in action.

src/ripple/app/main/Application.cpp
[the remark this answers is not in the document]
ximinez: Until we decide on a better name, I'm going to use `isUNLBlocked`.

src/ripple/app/misc/impl/ValidatorList.cpp
ximinez: We can hit this case based on user input. The `/vl/` endpoint will parse a number and pass it in.
```
$ curl -D - -k https://localhost:51235/vl/999/ED2677ABFFD1B33AC6FBC3062B71F1E8397C1505E1C42C64D11AD1B28FF73F4734
HTTP/1.1 400 Bad Request
Server: rippled-1.7.0-b3+DEBUG
Content-Type: application/json
Connection: close
Transfer-Encoding: chunked

null
```
The 400 result code indicates that the public key is valid, but the request is bad - in this case, because it requested version 999. If the public key is not valid, you get a 404.

Is it worth logging? Eh, couldn't hurt, but I think only at trace level.

src/ripple/app/misc/impl/ValidatorList.cpp
ximinez: Actually, there is one reason, but it's extremely unlikely. It's possible for a peer to send a version 1 protocol message which contains a VL with a future effective date. That would cause `rawVersion` to be 1, even though we _could_ be building a version 2 value.

What I'm going to do is change the code here to always use `rawVersion`, but also include an assert to make sure `rawVersion` is "correct". I'm also going to change the `rawVersion` assignment so that if we get into this situation, it forces the version to 2.

src/ripple/app/misc/impl/ValidatorList.cpp
ximinez: Great tip! I didn't dig enough into it to figure out how it was computed. I assumed it was way more complicated than that.

src/ripple/app/misc/impl/ValidatorList.cpp
ximinez: This is true. I'm going to sidestep your question, and I'm just going to send any v1 messages regardless of their size.

The current UNL is about 20Kb. The message size limit is 64Mb. I think I was being overly paranoid when I spec'd this out in the first place, because to come anywhere near that, the UNL file size would have to increase over 3000x. I'm tempted to remove the `splitMessage` logic entirely, but I'll leave it for now, and have these v1 messages send regardless.

src/ripple/app/misc/impl/ValidatorList.cpp
```diff
@@ -229,152 +298,653 @@ ValidatorList::CacheValidatorFile(
...
+    auto const manifest =
+        currentBlob.manifest ? *currentBlob.manifest : rawManifest;
+    auto const version = 1;
+    msg.set_manifest(manifest);
+    msg.set_blob(currentBlob.blob);
+    msg.set_signature(currentBlob.signature);
+    // Override the version
+    msg.set_version(version);
+
+    assert(Message::totalSize(msg) <= maxSize || maxSize < maximiumMessageSize);
```
undertome, line 501: Please see [this comment](https://github.com/ripple/rippled/pull/3619/files#r514512968).

src/ripple/app/misc/impl/ValidatorList.cpp
```diff
@@ -229,152 +298,653 @@ ValidatorList::CacheValidatorFile(
...
+
+        auto const& blob = largeMsg.blobs(begin);
+        smallMsg.set_blob(blob.blob());
+        smallMsg.set_signature(blob.signature());
+        // This is only possible if "downgrading" a v2 UNL to v1.
+        if (blob.has_manifest())
+            smallMsg.set_manifest(blob.manifest());
+
+        assert(
+            Message::totalSize(smallMsg) <= maxSize ||
```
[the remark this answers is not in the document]
ximinez: Good catch!

undertome (approved):

src/ripple/app/misc/impl/ValidatorList.cpp
```diff
@@ -229,152 +298,651 @@ ValidatorList::CacheValidatorFile(
...
-    using namespace std::string_literals;
-
-    if (version != requiredListVersion)
+    if (std::count(
+            std::begin(supportedListVersions),
+            std::end(supportedListVersions),
+            version) != 1)
         return PublisherListStats{ListDisposition::unsupported_version};
 
     std::unique_lock<std::shared_timed_mutex> lock{mutex_};
```
seelabs, line 876: Not in this change set, but it seems like this should be a `lock_guard`, not a `unique_lock`.

src/ripple/overlay/impl/PeerImp.h
```diff
@@ -43,6 +43,9 @@
 
 namespace ripple {
 
+// predeclaration
```
seelabs, line 46: Consider removing this comment

src/test/jtx/TrustedPublisherServer.h
```diff
@@ -58,6 +58,7 @@ class TrustedPublisherServer
     endpoint_type ep_;
     boost::asio::ip::tcp::acceptor acceptor_;
     std::function<std::string(int)> getList_;
+    std::function<std::string(int)> getList2_;
```
seelabs, line 61: Needs comment

src/test/jtx/TrustedPublisherServer.h
```diff
@@ -170,29 +176,77 @@ class TrustedPublisherServer
         auto const manifest = makeManifestString(
             publisherPublic_, publisherSecret_, keys.first, keys.second, 1);
 
-        std::string data = "{\"sequence\":" + std::to_string(sequence) +
-            ",\"expiration\":" +
-            std::to_string(expiration.time_since_epoch().count()) +
-            ",\"validators\":[";
+        std::vector<std::pair<std::string, std::string>> blobInfo;
```
seelabs, line 179: Consider a struct instead of `pair<string,string>`. Pairs where both elements have the same type are error prone.

src/test/jtx/TrustedPublisherServer.h
```diff
@@ -170,29 +176,77 @@ class TrustedPublisherServer
...
+        blobInfo.reserve(futures.size() + 1);
+        auto const [data, blob] = [&]() {
```
seelabs, line 181: Consider a comment and adding a return type to the lambda. (I don't mind leaving off return types of most lamdas, but when the return type is not obvious and the lambda is more than a couple lines I think they improve the code).

src/test/jtx/TrustedPublisherServer.h
```diff
@@ -618,14 +693,15 @@ inline std::shared_ptr<TrustedPublisherServer>
 make_TrustedPublisherServer(
     boost::asio::io_context& ioc,
     std::vector<TrustedPublisherServer::Validator> const& validators,
-    NetClock::time_point expiration,
+    NetClock::time_point validUntil,
+    std::vector<std::pair<NetClock::time_point, NetClock::time_point>> futures,
```
seelabs, line 697: Pass by `const&`

src/test/net/DatabaseDownloader_test.cpp
```diff
@@ -40,6 +40,7 @@ class DatabaseDownloader_test : public beast::unit_test::suite
             env.app().getIOService(),
             list,
             env.timeKeeper().now() + std::chrono::seconds{3600},
+            {},
```
seelabs, line 43: Probably could use a comment

src/test/rpc/ShardArchiveHandler_test.cpp
```diff
@@ -45,6 +45,7 @@ class ShardArchiveHandler_test : public beast::unit_test::suite
             env.app().getIOService(),
             list,
             env.timeKeeper().now() + std::chrono::seconds{3600},
+            {},
```
seelabs, line 48: Probably could use a comment

src/ripple/app/misc/NetworkOPs.cpp
```diff
@@ -1561,6 +1579,13 @@ NetworkOPsImp::setAmendmentBlocked()
     setMode(OperatingMode::TRACKING);
```
nbougalis, line 1579: While we're here, this should probably change too: `TRACKING` means "convinced we agree with the network" per the documentation. If we are amendment blocked we *might* transiently agree with the network, but the fact that we are actually **UNABLE** to process transactions the network is able to generate suggests that this should be `CONNECTED` instead.

src/ripple/app/misc/NetworkOPs.cpp
```diff
@@ -2154,7 +2183,9 @@ NetworkOPsImp::setMode(OperatingMode om)
             om = OperatingMode::CONNECTED;
     }
 
-    if ((om > OperatingMode::TRACKING) && amendmentBlocked_)
+    if ((om > OperatingMode::CONNECTED) && isUNLBlocked())
+        om = OperatingMode::CONNECTED;
+    else if ((om > OperatingMode::TRACKING) && isAmendmentBlocked())
         om = OperatingMode::TRACKING;
```
nbougalis, line 2189: Same here... `CONNECTED` seems more appropriate.

nbougalis (approved):
:+1: pending resolutions of a couple of things below and @seelabs's comments.

nbougalis:
@ximinez: please rebase this to the latest develop, resolving all conflicts and force-pushing here. Thanks.

src/ripple/app/misc/NetworkOPs.cpp
```diff
@@ -426,6 +426,11 @@ class NetworkOPsImp final : public NetworkOPs
     void
     setMode(OperatingMode om) override;
 
+    bool
+    isFunctionallyBlocked() override
```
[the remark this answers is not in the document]
ximinez: 👍  Renamed to `isBlocked`

src/ripple/app/misc/NetworkOPs.cpp
```diff
@@ -426,6 +426,11 @@ class NetworkOPsImp final : public NetworkOPs
...
+    {
+        return isAmendmentBlocked() || isUNLBlocked();
```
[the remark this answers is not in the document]
ximinez: Done.

src/ripple/app/misc/ValidatorList.h
```diff
@@ -26,40 +26,80 @@
...
+#include <ripple/overlay/Message.h>
 #include <ripple/protocol/PublicKey.h>
 #include <boost/iterator/counting_iterator.hpp>
 #include <boost/range/adaptors.hpp>
 #include <mutex>
 #include <numeric>
 #include <shared_mutex>
 
+namespace protocol {
+// predeclaration
```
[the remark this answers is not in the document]
ximinez: Done

src/ripple/app/misc/ValidatorList.h
```diff
@@ -73,6 +113,14 @@ struct TrustChanges
     hash_set<NodeID> removed;
 };
 
+/** Used to represent the information stored in the blobs_v2 Json array */
+struct ValidatorBlobInfo
+{
+    std::string blob;
+    std::string signature;
```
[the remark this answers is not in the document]
ximinez: Done

src/ripple/app/misc/ValidatorList.h
```diff
@@ -122,18 +172,30 @@ class ValidatorList
...
         std::string siteUri;
-        std::string rawManifest;
         std::string rawBlob;
         std::string rawSignature;
-        std::uint32_t rawVersion;
+        boost::optional<std::string> rawManifest;
         uint256 hash;
     };
 
+    struct PublisherListCollection
```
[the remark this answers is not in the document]
ximinez: Two of them are private, so I'd like to keep those. Of the two that are public, I notice that they have several functions defined inline. If I moved those definitions outside of the class, as I did for `NetworkOPsImp` above, I think that'll tighten things up. If you don't like the result, then I'll move them.

src/ripple/app/misc/ValidatorList.h
```diff
@@ -122,18 +172,30 @@ class ValidatorList
...
+    {
+        PublisherStatus status;
+        PublisherList current;
```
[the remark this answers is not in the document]
ximinez: Done. And for `remaining`.

src/ripple/app/misc/ValidatorList.h
```diff
@@ -122,18 +172,30 @@ class ValidatorList
...
+        boost::optional<std::size_t> maxSequence;
+        // The hash of the full set if sent in a single message
+        uint256 fullHash;
+        std::string rawManifest;
+        std::uint32_t rawVersion = 0;
+    };
+
     ManifestCache& validatorManifests_;
     ManifestCache& publisherManifests_;
     TimeKeeper& timeKeeper_;
```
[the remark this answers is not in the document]
ximinez: Not that unusual. `beast::Journal const` appears over 100 times in the code, and all of the member functions are defined as `const`, so it works just fine.

src/ripple/app/misc/ValidatorList.h
```diff
@@ -163,8 +225,12 @@ class ValidatorList
     // The master public keys of the current negative UNL
     hash_set<PublicKey> negativeUNL_;
 
-    // Currently supported version of publisher list format
-    static constexpr std::uint32_t requiredListVersion = 1;
+    // Currently supported versions of publisher list format
+    static constexpr std::uint32_t supportedListVersions[]{1, 2};
```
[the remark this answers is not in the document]
ximinez: The downside of `std::array` for me for something this trivially simple is that you have to specify the size in the template arguments.

src/ripple/app/misc/ValidatorList.h
```diff
@@ -163,8 +225,12 @@ class ValidatorList
...
+    // In the initial release, to prevent potential abuse and attacks, any VL
+    // collection with more than 5 entries will be considered malformed.
+    static constexpr std::size_t MAX_SUPPORTED_BLOBS = 5;
```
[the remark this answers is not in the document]
ximinez: Fixed.

src/ripple/app/misc/ValidatorList.h
```diff
@@ -184,114 +250,175 @@ class ValidatorList
     */
     struct PublisherListStats
     {
-        explicit PublisherListStats(ListDisposition d) : disposition(d)
+        explicit PublisherListStats() = default;
+
+        explicit PublisherListStats(ListDisposition d)
         {
+            ++dispositions[d];
```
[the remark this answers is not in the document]
ximinez: Done.

src/ripple/app/misc/ValidatorList.h
```diff
@@ -184,114 +250,175 @@ class ValidatorList
...
+        }
+        std::shared_ptr<Message> message;
+        uint256 hash;
+        std::size_t numVLs = 0;
     };
 
     /** Load configured trusted keys.
 
-        @param localSigningKey This node's validation public key
+            @param localSigningKey This node's validation public key
```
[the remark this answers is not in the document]
ximinez: Fixed

src/ripple/app/misc/ValidatorList.h
```diff
@@ -184,114 +250,175 @@ class ValidatorList
...
+    sendValidatorList(
+        Peer& peer,
+        std::uint64_t peerSequence,
+        PublicKey const& publisherKey,
+        std::size_t maxSequence,
+        std::uint32_t rawVersion,
+        std::string const& rawManifest,
+        std::map<std::size_t, ValidatorBlobInfo> const& blobInfos,
+        HashRouter& hashRouter,
+        beast::Journal const j);
```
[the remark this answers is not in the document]
ximinez: Fixed

src/ripple/app/misc/ValidatorList.h
```diff
@@ -184,114 +250,175 @@ class ValidatorList
...
-            Defaults to uninitialized
-
-        @return `ListDisposition::accepted`, plus some of the publisher
-            information, if list was successfully applied
-
-        @par Thread Safety
-
-        May be called concurrently
-    */
+    /** TODO: Document this once the interface settles */
```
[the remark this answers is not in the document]
ximinez: Documented

src/ripple/app/misc/ValidatorList.h
```diff
@@ -541,15 +675,104 @@ class ValidatorList
...
+
+    static void
+    broadcastBlobs(
+        PublicKey const& publisherKey,
+        PublisherListCollection const& lists,
+        std::size_t maxSequence,
+        uint256 const& hash,
+        Overlay& overlay,
+        HashRouter& hashRouter,
+        beast::Journal const j);
```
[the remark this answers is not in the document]
ximinez: fixed

src/ripple/app/misc/ValidatorList.h
```diff
@@ -541,15 +675,104 @@ class ValidatorList
...
+        Peer& peer,
+        std::uint64_t peerSequence,
+        PublicKey const& publisherKey,
+        std::size_t maxSequence,
+        std::uint32_t rawVersion,
+        std::string const& rawManifest,
+        std::map<std::size_t, ValidatorBlobInfo> const& blobInfos,
+        std::vector<MessageWithHash>& messages,
+        HashRouter& hashRouter,
+        beast::Journal const j);
```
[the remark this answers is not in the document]
ximinez: fixed

src/ripple/app/misc/ValidatorList.h
```diff
@@ -541,15 +675,104 @@ class ValidatorList
...
+    getCacheFileName(PublicKey const& pubKey) const;
+
+    /** Build a Json representation of the collection, suitable for
+        writing to a cache file, or serving to a /vl/ query
+    */
+    static Json::Value
+    buildFileData(
+        std::string const& pubKey,
+        PublisherListCollection const& pubCollection,
+        beast::Journal const j,
```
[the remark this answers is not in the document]
ximinez: Done

src/ripple/app/misc/ValidatorList.h
```diff
@@ -596,6 +819,63 @@ class ValidatorList
...
+        hash_append(h, item);
+}
+
+template <class Hasher>
+void
+hash_append(Hasher& h, std::map<std::size_t, ValidatorBlobInfo> const& blobs)
+{
+    for (auto const& [_, item] : blobs)
+    {
+        (void)_;
```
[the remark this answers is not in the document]
ximinez: CI is still building with older versions. I assume it's needed there.

src/ripple/app/misc/impl/ValidatorList.cpp
ximinez: That's gonna be fun to merge with develop, but, yeah, it makes sense.

src/ripple/overlay/impl/PeerImp.h
ximinez: Done

src/test/jtx/TrustedPublisherServer.h
ximinez: Done

src/ripple/app/misc/impl/ValidatorList.cpp
ximinez: Fixed after rebasing

src/test/jtx/TrustedPublisherServer.h
ximinez: Done

src/test/jtx/TrustedPublisherServer.h
ximinez: Done.

src/test/jtx/TrustedPublisherServer.h
ximinez: Fixed

src/test/net/DatabaseDownloader_test.cpp
ximinez: Fixed

src/test/rpc/ShardArchiveHandler_test.cpp
ximinez: Fixed

ximinez:
I forgot to mention this yesterday, but that last push included both rebasing onto 1.7.0-b8 and the latest round of feedback from @seelabs.

src/ripple/app/misc/ValidatorList.h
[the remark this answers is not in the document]
ximinez: I figure 2 is probably going to be what we're most likely to see, but I wanted to leave a little wiggle room in case someone comes up with some interesting use cases we haven't considered.

src/ripple/app/misc/NetworkOPs.cpp
ximinez: Good point. Done.

src/ripple/app/misc/NetworkOPs.cpp
ximinez: Done

src/ripple/app/misc/ValidatorList.h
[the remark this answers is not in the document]
seelabs: That works. Thanks!

src/ripple/app/misc/ValidatorList.h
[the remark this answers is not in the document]
seelabs: You're right! (I was looking at the `Sink` inner class where `write` is not `const` and thought this `journal` was never written to).

seelabs (approved):
Thanks for resolving those comments. LGTM :+1:

intelliot:
Merged via 4b9d3ca7de7c5758fbd6ae774c74196fa6b4268c

ckniffen:
While effective date is documented there is no notes on XRPL.org that you need to specify version as 2.

src/ripple/app/misc/ValidatorList.h
```diff
@@ -122,18 +170,30 @@ class ValidatorList
     {
         explicit PublisherList() = default;
 
-        bool available;
         std::vector<PublicKey> list;
+        std::vector<std::string> manifests;
         std::size_t sequence;
         TimeKeeper::time_point expiration;
+        TimeKeeper::time_point effective;
```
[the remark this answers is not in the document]
intelliot: `validFrom` and `validUntil` are used internally for the logic within `rippled` (in the C++ code). Validator list creators/maintainers use `effective` and `expiration`.

ximinez:
> While effective date is documented there is no notes on XRPL.org that you need to specify version as 2.

Strictly speaking, I think a version 1 blob with an effective date would work fine. It would not be best practice because the current VL blob wouldn't be included with it.

PR #3620, opened 2020-10-08
Temper excessive warnings about historical shard

undertome:

When a node already has historical shards stored, but is subsequently configured to not store historical shards, the shard database produces excessive warnings about the historical shard limit being exceeded. This PR fixes this issue, and also replaces several calls to a `Nodestore` function with roughly equivalent calls to a `DatabaseShardImp` function.

### DIFF
--- a/src/ripple/nodestore/impl/DatabaseShardImp.cpp
+++ b/src/ripple/nodestore/impl/DatabaseShardImp.cpp
@@ -1579,7 +1579,7 @@ DatabaseShardImp::shardBoundaryIndex() const
     auto const validIndex = app_.getLedgerMaster().getValidLedgerIndex();
 
     if (validIndex < earliestLedgerSeq())
-        return validIndex;
+        return 0;
 
     // Shards with an index earlier than the recent shard boundary index
     // are considered historical. The three shards at or later than
--- a/src/ripple/nodestore/impl/DatabaseShardImp.cpp
+++ b/src/ripple/nodestore/impl/DatabaseShardImp.cpp
@@ -891,6 +891,9 @@ DatabaseShardImp::import(Database& source)
                             true,
                             boost::none);
                         success = true;
+
+                        if (shardIndex < shardBoundaryIndex())
+                            ++numHistShards;
                     }
                     catch (std::exception const& e)
                     {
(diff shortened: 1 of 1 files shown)


src/ripple/nodestore/impl/DatabaseShardImp.cpp
```diff
@@ -1604,12 +1611,15 @@ DatabaseShardImp::shardBoundaryIndex(std::lock_guard<std::mutex> const&) const
 {
     auto const validIndex = app_.getLedgerMaster().getValidLedgerIndex();
 
+    if (validIndex < earliestLedgerSeq())
```
miguelportilla, line 1614: `DatabaseShardImp::shardBoundaryIndex` doesn't need a lock.

src/ripple/nodestore/impl/DatabaseShardImp.cpp
```diff
@@ -1428,7 +1428,14 @@ DatabaseShardImp::setFileStats()
...
     {
-        JLOG(j_.warn()) << "maximum number of historical shards reached";
+        if (maxHistoricalShards_)
+        {
+            // In order to avoid excessive output, don't produce
+            // this warning if the server isn't configured to
+            // store historical shards.
+            JLOG(j_.warn()) << "maximum number of historical shards reached";
+        }
+
```
miguelportilla, line 1438: If `canAdd_` is true, we can return early.
eg.
```
    if (!canAdd_)
        return;

    auto const count{numHistoricalShards(lock)};
    if (count >= maxHistoricalShards_)
    {
        JLOG(j_.warn()) << "maximum number of historical shards reached";
        canAdd_ = false;
    }
    else if (!sufficientStorage(
                 maxHistoricalShards_ - count,
                 PathDesignation::historical,
                 lock))
    {
        JLOG(j_.warn())
            << "maximum shard store size exceeds available storage space";
    }
```

src/ripple/nodestore/impl/DatabaseShardImp.cpp
undertome: Done

src/ripple/nodestore/impl/DatabaseShardImp.cpp
undertome: Good idea, done

miguelportilla (approved):
LGTM 👍

src/ripple/nodestore/impl/DatabaseShardImp.cpp
```diff
@@ -1561,14 +1576,17 @@ DatabaseShardImp::removeFailedShard(std::shared_ptr<Shard>& shard)
...
     // Shards with an index earlier than the recent shard boundary index
     // are considered historical. The three shards at or later than
     // this index consist of the two most recently validated shards
     // and the shard still in the process of being built by live
     // transactions.
-    return NodeStore::seqToShardIndex(
-               app_.getLedgerMaster().getValidLedgerIndex(), ledgersPerShard_) -
-        1;
+    return seqToShardIndex(validIndex) - 1;
```
ximinez, line 1589: I think you're mixing metaphors here. The `return validIndex` is returning a ledger index. The `return seqToShardIndex(validIndex) - 1` is returning a shard index. I'm guessing the idea was to avoid returning a negative / underflowed value, but that first return now doesn't make sense. Does it make more sense to `return 0` if the ledger index is too small?

ximinez (changes requested):

src/ripple/nodestore/impl/DatabaseShardImp.cpp
undertome: The idea was to avoid passing an index that's too early to `seqToShardIndex` which would cause an `assert` to fail. But I agree it makes more sense to just return zero. Done

ximinez (approved):

PR #3632, opened 2020-10-14
Always check the sequence when adding to the transaction queue

ximinez:

#3271 made a change to transaction processing in the Transaction Queue (`TxQ`) for transactions that are being considered for queuing, to validate the sequence/ticket on the `TxQ` side, and bypass the sequence check in the `Transactor`. Because the sequence didn't need to be checked, it was thought no longer necessary to "bump" the account's sequence number if there were multiple transactions queued, so that code was removed. Unfortunately, there are cases in `OfferCreate` and `OfferCancel` where the `OfferSequence` field is checked against the account sequence number. If the offer to be cancelled is in the queue, and the account sequence doesn't get bumped, then when the transaction is processed it will return `temBAD_SEQUENCE` from `preflight`. Worse, because that's a `tem` code, the transaction is cached as "bad" and won't be reconsidered unless the transaction is in a validated ledger (fortunately, no validators are currently running 1.7, so that's fine).

This PR accomplishes three separate things, left for now as three separate commits. When it's ready to be merged, we may want to squash them into a single commit, or only squash the first two (more on that below).
1. The first commit adds a unit test reproducing the problem. It also add a new engine result code `terTICKET`, which indicates that the transaction can't be processed until the ticket it refers to is consumed. No code outside of the test refers to this new code, so I think it's safe.
2. Is the actual fix. First, and most importantly, it changes the `TxQ` logic to go back to the old behavior of "bumping" the account's sequence number, and always checks the sequence in `preflight`. Second, it checks all transactions that refer to another transaction using its sequence number (including `OfferCreate`, `OfferCancel`, `EscrowFinish`, and `EscrowCancel`, but the check is generalized to all transactions that have the right fields) for an existing ticket in the ledger for that sequence. If a pending ticket is found, the new `terTICKET` result is returned. As long as this PR is included in 1.7 release, this does not need a new amendment because ticket functionality is new with 1.7 and without the `TicketBatch` amendment, tickets can not exist in the ledger. It also adds a `STObject::at()` which functions identically to `STObject::operator[]`, and splits the `TxQ` unit tests into two suites to improve speed.
3. This one is a little more optional. The original symptom of the problem was because `temBAD_SEQUENCE` was returned by `preclaim` based on data in the ledger. IMHO, that check can and should be done in `preflight` based just on the data in the transaction. This adds a new amendment `fixOfferSequence`. If the amendment is enabled and the `OfferSequence` is not smaller than the `Sequence`, `preflight` returns `temBAD_SEQUENCE`. This is the exact same behavior, only the check is sooner in the process. I only put it behind an amendment because they changed order may have some unexpected consequences (and it _did_ require updating a unit test). I still think it's useful because catching this malformation earlier will save time and resources.

### Type of Change

### Test the `temBAD_SEQUENCE` fix

With version 1.6.0 and the version that contains this PR.
1. Get enough transactions into the ledger to cause fee to escalate.
2. Submit several `OfferCreate` and `OfferCancel` transactions with low fees (e.g. 10 drops) in short order (before the ledger closes). Each transaction except the first should cancel the prior. The first must be an `OfferCreate`. The others can be some kind of mix.
3. Verify that all the transactions return `terQUEUED`.
4. Verify that all the transactions eventually make it into a validated ledger.

With a beta of 1.7 before this is merged, all of the transactions submitted after the first will return `temBAD_SEQUENCE`.

### Test `terTICKET`

1. Enable the "TicketBatch" amendment.
2. Create an arbitrary number of tickets using `TicketCreate`.
3. Submit an `OfferCreate`, `OfferCancel`, `EscrowFinish`, and/or `EscrowCancel` that use the sequence number of one of the just created tickets as the `OfferSequence`. For the escrow transactions, you can either use the same account in the `Owner` field, or even better, submit those from an account different than the one that created the tickets and use the ticket creator as `Owner`.
4. All of the transactions in step 2 should return `terTICKET`.
5. Submit _any_ kind of transaction to consume the tickets that the transactions from step 2 referred to. Some of them can be `EscrowCreate`, but that is not necessary.
6. Resubmit all of the transactions from step 2.
7. All of the offer-related transaction should succeed. The escrow-related transactions should only succeed if a corresponding escrow was created in step 4, otherwise, they should return `tecNO_TARGET`.

### Test `fixOfferSequence`

1. Without the amendment, submit an `OfferCreate` or `OfferCancel` where the `Sequence` field is smaller or equal to the `OfferSequence` field. Note that `OfferSequence` is optional in the `OfferCreate` transaction. It will return `temBAD_SEQUENCE`.
2. With the amendment, submit an `OfferCreate` or `OfferCancel` where the `Sequence` field is smaller or equal to the `OfferSequence` field. Note that `OfferSequence` is optional in the `OfferCreate` transaction. It will also return `temBAD_SEQUENCE`.

## Future Tasks

Only `preflight` should return `tem` engine results. It should not require referring to the ledger (except the `Rules` which is provided to `preflight`) to determine if a transaction is properly formed. Similar to the `NotTEC` type, consider a `NotTEM` type that is returned from `preclaim` and `doApply`. Once the `fixOfferSequence` is enabled, it should be safe to remove the check in `preflight` that first triggered this issue (or return a different result). Any other `tem` codes returned by `preclaim` or `doApply` will probably need similar amendments.

---
This change is [Reviewable](https://reviewable.io/reviews/ripple/rippled/3632)

### DIFF
--- a/src/ripple/app/misc/impl/TxQ.cpp
+++ b/src/ripple/app/misc/impl/TxQ.cpp
@@ -1120,6 +1120,14 @@ TxQ::apply(
                 std::min(balance - std::min(balance, reserve), potentialSpend);
             assert(potentialTotalSpend > XRPAmount{0});
             sleBump->setFieldAmount(sfBalance, balance - potentialTotalSpend);
+            // The transaction's sequence/ticket will be valid when the other
+            // transactions in the queue have been processed. If the tx has a
+            // sequence, set the account to match it. If it has a ticket, use
+            // the next queueable sequence, which is the closest approximation
+            // to the most successful case.
+            sleBump->at(sfSequence) = txSeqProx.isSeq()
+                ? txSeqProx.value()
+                : nextQueuableSeqImpl(sleAccount, lock).value();
         }
     }
 
@@ -1134,8 +1142,8 @@ TxQ::apply(
     // Note that earlier code has already verified that the sequence/ticket
     // is valid.  So we use a special entry point that runs all of the
     // preclaim checks with the exception of the sequence check.
-    auto const pcresult = ForTxQ::preclaimWithoutSeqCheck(
-        pfresult, app, multiTxn ? multiTxn->openView : view);
+    auto const pcresult =
+        preclaim(pfresult, app, multiTxn ? multiTxn->openView : view);
     if (!pcresult.likelyToClaimFee)
         return {pcresult.ter, false};
 
--- a/src/ripple/app/tx/impl/CancelOffer.cpp
+++ b/src/ripple/app/tx/impl/CancelOffer.cpp
@@ -40,8 +40,7 @@ CancelOffer::preflight(PreflightContext const& ctx)
         return temINVALID_FLAG;
     }
 
-    auto const seq = ctx.tx.getFieldU32(sfOfferSequence);
-    if (!seq)
+    if (!ctx.tx[sfOfferSequence])
     {
         JLOG(ctx.j.trace()) << "CancelOffer::preflight: missing sequence";
         return temBAD_SEQUENCE;
--- a/src/ripple/app/tx/impl/CreateOffer.cpp
+++ b/src/ripple/app/tx/impl/CreateOffer.cpp
@@ -75,9 +75,8 @@ CreateOffer::preflight(PreflightContext const& ctx)
         return temBAD_EXPIRATION;
     }
 
-    bool const bHaveCancel(tx.isFieldPresent(sfOfferSequence));
-
-    if (bHaveCancel && (tx.getFieldU32(sfOfferSequence) == 0))
+    if (auto const cancelSequence = tx[~sfOfferSequence];
+        cancelSequence && *cancelSequence == 0)
     {
         JLOG(j.debug()) << "Malformed offer: bad cancel sequence";
         return temBAD_SEQUENCE;
--- a/src/ripple/app/tx/impl/Transactor.cpp
+++ b/src/ripple/app/tx/impl/Transactor.cpp
@@ -247,7 +247,7 @@ Transactor::checkSeqProxy(
     {
         JLOG(j.trace())
             << "applyTransaction: delay: source account does not exist "
-            << toBase58(tx.getAccountID(sfAccount));
+            << toBase58(id);
         return terNO_ACCOUNT;
     }
 
@@ -305,7 +305,7 @@ Transactor::checkPriorTxAndLastLedger(PreclaimContext const& ctx)
     {
         JLOG(ctx.j.trace())
             << "applyTransaction: delay: source account does not exist "
-            << toBase58(ctx.tx.getAccountID(sfAccount));
+            << toBase58(id);
         return terNO_ACCOUNT;
     }
 
--- a/src/ripple/protocol/STObject.h
+++ b/src/ripple/protocol/STObject.h
@@ -477,31 +477,35 @@ public:
     const STArray&
     getFieldArray(SField const& field) const;
 
-    /** Return the value of a field.
-
-        Throws:
-
-            STObject::FieldErr if the field is
-            not present.
+    /** Get the value of a field.
+        @param A TypedField built from an SField value representing the desired
+            object field. In typical use, the TypedField will be implicitly
+            constructed.
+        @return The value of the specified field.
+        @throws STObject::FieldErr if the field is not present.
     */
     template <class T>
     typename T::value_type
     operator[](TypedField<T> const& f) const;
 
-    /** Return the value of a field as boost::optional
+    /** Get the value of a field as boost::optional
 
-        @return boost::none if the field is not present.
+        @param An OptionaledField built from an SField value representing the
+           desired object field. In typical use, the OptionaledField will be
+           constructed by using the ~ operator on an SField.
+        @return boost::none if the field is not present, else the value of the
+           specified field.
     */
     template <class T>
     boost::optional<std::decay_t<typename T::value_type>>
     operator[](OptionaledField<T> const& of) const;
 
-    /** Return a modifiable field value.
-
-        Throws:
-
-            STObject::FieldErr if the field is
-            not present.
+    /** Get a modifiable field value.
+        @param A TypedField built from an SField value representing the desired
+            object field. In typical use, the TypedField will be implicitly
+            constructed.
+        @return A modifiable reference to the value of the specified field.
+        @throws STObject::FieldErr if the field is not present.
     */
     template <class T>
     ValueProxy<T>
@@ -509,12 +513,63 @@ public:
 
     /** Return a modifiable field value as boost::optional
 
-        The return value equals boost::none if the
-        field is not present.
+        @param An OptionaledField built from an SField value representing the
+            desired object field. In typical use, the OptionaledField will be
+            constructed by using the ~ operator on an SField.
+        @return Transparent proxy object to an `optional` holding a modifiable
+            reference to the value of the specified field. Returns boost::none
+            if the field is not present.
     */
     template <class T>
     OptionalProxy<T>
     operator[](OptionaledField<T> const& of);
+
+    /** Get the value of a field.
+        @param A TypedField built from an SField value representing the desired
+            object field. In typical use, the TypedField will be implicitly
+            constructed.
+        @return The value of the specified field.
+        @throws STObject::FieldErr if the field is not present.
+    */
+    template <class T>
+    typename T::value_type
+    at(TypedField<T> const& f) const;
+
+    /** Get the value of a field as boost::optional
+
+        @param An OptionaledField built from an SField value representing the
+           desired object field. In typical use, the OptionaledField will be
+           constructed by using the ~ operator on an SField.
+        @return boost::none if the field is not present, else the value of the
+           specified field.
+    */
+    template <class T>
+    boost::optional<std::decay_t<typename T::value_type>>
+    at(OptionaledField<T> const& of) const;
+
+    /** Get a modifiable field value.
+        @param A TypedField built from an SField value representing the desired
+            object field. In typical use, the TypedField will be implicitly
+            constructed.
+        @return A modifiable reference to the value of the specified field.
+        @throws STObject::FieldErr if the field is not present.
+    */
+    template <class T>
+    ValueProxy<T>
+    at(TypedField<T> const& f);
+
+    /** Return a modifiable field value as boost::optional
+
+        @param An OptionaledField built from an SField value representing the
+            desired object field. In typical use, the OptionaledField will be
+            constructed by using the ~ operator on an SField.
+        @return Transparent proxy object to an `optional` holding a modifiable
+            reference to the value of the specified field. Returns boost::none
+            if the field is not present.
+    */
+    template <class T>
+    OptionalProxy<T>
+    at(OptionaledField<T> const& of);
 
     /** Set a field.
         if the field already exists, it is replaced.
@@ -927,6 +982,34 @@ template <class T>
 typename T::value_type
 STObject::operator[](TypedField<T> const& f) const
 {
+    return at(f);
+}
+
+template <class T>
+boost::optional<std::decay_t<typename T::value_type>>
+STObject::operator[](OptionaledField<T> const& of) const
+{
+    return at(of);
+}
+
+template <class T>
+inline auto
+STObject::operator[](TypedField<T> const& f) -> ValueProxy<T>
+{
+    return at(f);
+}
+
+template <class T>
+inline auto
+STObject::operator[](OptionaledField<T> const& of) -> OptionalProxy<T>
+{
+    return at(of);
+}
+
+template <class T>
+typename T::value_type
+STObject::at(TypedField<T> const& f) const
+{
     auto const b = peekAtPField(f);
     if (!b)
         // This is a free object (no constraints)
@@ -951,7 +1034,7 @@ STObject::operator[](TypedField<T> const& f) const
 
 template <class T>
 boost::optional<std::decay_t<typename T::value_type>>
-STObject::operator[](OptionaledField<T> const& of) const
+STObject::at(OptionaledField<T> const& of) const
 {
     auto const b = peekAtPField(*of.f);
     if (!b)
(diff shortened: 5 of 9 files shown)


ximinez:
Rebased on to 1.7.0-b3

scottschurr:
I spent some time looking at the bug that tickets introduced.  I compared the value of the messed up error check against the added complexity in the TxQ.  Given that comparison, I'd like to propose an alternative approach to a fix.

The `OfferSequence` validation on canceled offers is pretty rudimentary.  It checks for zero (which would never be valid) and it checks for the `OfferSequence` being less than the account root `Sequence`.  This validation only acts as a rudimentary sanity check.  There has never been any validation that the offer being canceled is now or ever was in the ledger.  Canceling an offer that is not in the ledger is handled as a successful transaction.

My suggestion is that we remove the rudimentary validation.  The rudimentary validation used to be cheap.  It's no longer cheap because it requires the TxQ to jump through hoops for every transaction the TxQ validates (not just those that cancel offers).  I think those rudimentary checks no longer hold up to a cost/benefit analysis.

I did an audit and, beyond offer canceling, there's one additional transaction that is sensitive to the TxQ not having the right account root `Sequence` during `preclaim`: that's `AccountDelete`.  That check is important, but it returns a `tec` code.  So that check can be moved from `preclaim` to `doApply` with no loss of efficiency.

Here's a commit that shows the alternative I'm proposing: https://github.com/scottschurr/rippled/commit/b5c793200b03c77983f8af364dadb5fbad7649fb

Thoughts?  Thanks.

ximinez:
@scottschurr Thanks for thinking this through and going through the effort to come up with an alternative solution. I don't think it's the way to go, though. I have a few reasons:
* I don't think the changes I made to `TxQ` are all that complicated or have it jumping through hoops. Basically, all that changed is that `TxQ` now calls `nextQueuableSeqImpl` earlier and unconditionally, changes the sequence (a uint32) on an SLE that is already loaded, and removes the code that bypasses the sequence check in the `Transactor`. That has the effect of making the validation cheap again. It's also more resilient - if new functionality is introduced or changed that depends on account sequence numbers, the transactor doesn't have to know or care under what circumstances it's being invoked under. 
* Without your new amendment, `fixOfferCancelSeqErr`, being enabled, the broken behavior will still exist in 1.7, and there's no telling how long it'll take to get that enabled. The reason this behavior is so bad is https://github.com/ripple/rippled/blob/develop/src/ripple/app/misc/NetworkOPs.cpp#L1344-L1345 - The transaction is marked as *bad* and will not be considered by that node again unless it's in a proposed transaction set. And if the validators are all running 1.7, it may not be in any proposed set.
* My changes fix the bug right out of the gate, no amendment required. The amendment I introduced really is optional, and simply does the same check - basically preserving the same `Transactor` behavior - earlier in the process.
* As for `terTICKET`, which you propose not adding, that's not a core part of the bug fix, and it could be removed without breaking transaction processing. It would make the potential results of two related transactions (one consuming a ticket to create a ledger entry, the other operating on that ledger entry - this could include Offers, but _also_ Escrows) dependent on the order in which the transactions are executed, but the validators will choose one order and everything will be fine in terms of the ledger. I imagine that may cause some users to get unexpected results if an order they thought they had cancelled was still on the books. Ordering has always a potential issue with tickets, but in most cases it didn't _really_ matter. Some general things might be affected, like not having enough XRP to pay fees, but that can always be an issue. The difference IMHO is that these two transactions are _explicitly_ linked, so it doesn't really make sense to allow the cancel/finish to process before the create.

src/ripple/app/tx/impl/CreateOffer.cpp
```diff
@@ -75,9 +75,12 @@ CreateOffer::preflight(PreflightContext const& ctx)
...
     }
 
-    bool const bHaveCancel(tx.isFieldPresent(sfOfferSequence));
+    auto const cancelSequence = tx[~sfOfferSequence];
 
-    if (bHaveCancel && (tx.getFieldU32(sfOfferSequence) == 0))
+    if (cancelSequence &&
+        (*cancelSequence == 0 ||
+         (ctx.rules.enabled(fixOfferSequence) &&
+          tx.getSeqProxy() <= SeqProxy::sequence(*cancelSequence))))
```
scottschurr, line 83: I'm okay with, but not excited about, the extra checks provided by the `fixOfferSequence` amendment.  The checks have a low compute cost and that's great.  I'm bugged that the extra checks are on an amendment.  Because this is a low-benefit amendment, the amendment needs to be postponed so it comes on line at a time when it won't cause servers to become amendment blocked.

If we decide that the change can be safely made without an amendment I'd be fine with that.  Otherwise we need to decide whether the overhead of managing when the amendment becomes active is justified by the (in my opinion) relatively small benefit that the amendment provides.

scottschurr (commented):
Nicely done!  Your fix for the bug I introduced is elegant.  You've also added good unit tests.

I have some questions about the value of a couple of the enhancements that came along for the ride on this pull request.  If we decide that those enhancements have sufficient motivation then charge ahead.  But I would like to at least discuss the enhancements first.  Thanks.

src/ripple/app/tx/impl/CancelOffer.cpp
```diff
@@ -46,6 +47,14 @@ CancelOffer::preflight(PreflightContext const& ctx)
         JLOG(ctx.j.trace()) << "CancelOffer::preflight: missing sequence";
```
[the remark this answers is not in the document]
ximinez: Yeah, that's a good idea.

src/ripple/app/tx/impl/Transactor.cpp
```diff
@@ -290,6 +290,29 @@ Transactor::checkSeqProxy(
...
+            // order can be significant, don't allow this one to be processed
+            // until the ticket has been consumed. This is analagous to
+            // "terPRE_SEQ" for tickets.
+            JLOG(j.debug())
+                << "Can not cancel offer " << *referenceSequence
+                << " because the corresponding ticket hasn't been used yet, "
+                << "and the order of Ticket processing can not be guaranteed.";
+            return terTICKET;
+        }
+    }
```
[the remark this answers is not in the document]
ximinez: I didn't consider the performance implications, so thank you for checking that.

I like your solution, but I am wide open to a discussion of whether the "problem" is a problem worth solving with either option. If we decide to keep this check, your commit is definitely the way to go. If not, it's super easy to remove this check.

In "real world" terms, I predict that it will be pretty unlikely that any use case that submits and cancels offers so quickly that this could be an issue is probably going to be a bot, and I think a bot is really unlikely to use tickets. And most escrows last much longer than a couple of consensus rounds, AND an escrow cancel or finish that uses a "bad" sequence will get a `tecNO_TARGET`, which is a perfectly acceptable result IMHO.

src/ripple/app/misc/impl/TxQ.cpp
```diff
@@ -1120,6 +1120,13 @@ TxQ::apply(
                 std::min(balance - std::min(balance, reserve), potentialSpend);
             assert(potentialTotalSpend > XRPAmount{0});
             sleBump->setFieldAmount(sfBalance, balance - potentialTotalSpend);
+            // The transaction's sequence/ticket will be valid when the other
+            // transactions in the queue have been processed. If the tx has a
+            // sequence, set the account to match it. If it has a ticket, use
+            // the next queueable sequence, which is the closest approximation
+            // to the most successful case.
+            sleBump->at(sfSequence) =
+                txSeqProx.isSeq() ? txSeqProx.value() : nextSeqProx.value();
```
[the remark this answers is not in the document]
ximinez: Thanks!

src/ripple/app/tx/impl/Transactor.cpp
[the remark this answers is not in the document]
scottschurr: That's kinda my feeling as well.  Agree with your assessment about how frequently this check will pay off.  There are definitely situations where the check can fire, but they seem low likelihood.  And if the check is removed the consequences seem acceptable.  My inclination is to remove the check altogether.  But I'm open to further discussion.

src/ripple/protocol/impl/Feature.cpp
```diff
@@ -186,7 +187,8 @@ uint256 const
     featureHardenedValidations      = *getRegisteredFeature("HardenedValidations"),
     fixAmendmentMajorityCalc        = *getRegisteredFeature("fixAmendmentMajorityCalc"),
     featureNegativeUNL              = *getRegisteredFeature("NegativeUNL"),
-    featureTicketBatch              = *getRegisteredFeature("TicketBatch");
+    featureTicketBatch              = *getRegisteredFeature("TicketBatch"),
+    fixOfferSequence                = *getRegisteredFeature("fixOfferSequence");
```
miguelportilla, line 191: Outside the scope of this PR but I wanted to note the concern. These string literals are used in three different places, which leaves room for misspells or incorrect camel-case. It may be a good idea to define and reference the text in a single place.

src/ripple/app/tx/impl/CreateOffer.cpp
miguelportilla: Would it be worth covering this case in the new unit test? Here is a [branch](https://github.com/miguelportilla/rippled/commit/85a58fb086ea88f53da2cf68ac7e0ea3fa06ae9e) checking for the bad sequence.

src/ripple/app/misc/impl/TxQ.cpp
```diff
@@ -998,6 +998,7 @@ TxQ::apply(
             // transaction, if one does.
             TxQAccount::TxMap::const_iterator const prevIter =
                 txQAcct.getPrevTx(txSeqProx);
+            auto const nextSeqProx = nextQueuableSeqImpl(sleAccount, lock);
```
miguelportilla, line 1001: Consider deferring the assignment. Eg. [commit](https://github.com/miguelportilla/rippled/commit/80150588718bd68d567c520e5f90e54bb0eb460c) here.

src/ripple/app/tx/impl/Transactor.cpp
```diff
@@ -247,7 +247,7 @@ Transactor::checkSeqProxy(
     {
         JLOG(j.trace())
             << "applyTransaction: delay: source account does not exist "
-            << toBase58(tx.getAccountID(sfAccount));
+            << toBase58(id);
```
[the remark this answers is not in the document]
miguelportilla: `Transactor::checkPriorTxAndLastLedger` can benefit from the same change.

src/ripple/app/tx/impl/Transactor.cpp
```diff
@@ -290,6 +290,29 @@ Transactor::checkSeqProxy(
...
+        // current account. Validation of those fields is done elsewhere.
+        auto const owner = tx[~sfOwner].value_or(id);
+        if (view.exists(keylet::ticket(owner, *referenceSequence)))
+        {
+            // Ticket execution order is non-deterministic. The owner account
+            // may have a transaction pending which consumes the existing ticket
+            // to create an object that this transaction depends on. Because the
+            // consequences of executing the two transactions in a different
+            // order can be significant, don't allow this one to be processed
+            // until the ticket has been consumed. This is analagous to
```
miguelportilla, line 307: `analagous` -> `analogous`

miguelportilla: Thank you for the excellent comments!

src/ripple/app/tx/impl/Transactor.cpp
```diff
@@ -290,6 +290,29 @@ Transactor::checkSeqProxy(
...
+            // "terPRE_SEQ" for tickets.
+            JLOG(j.debug())
+                << "Can not cancel offer " << *referenceSequence
```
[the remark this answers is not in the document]
miguelportilla: Nit: Did you mean `debug` severity? All others errors here use `trace`.

src/ripple/protocol/STObject.h
```diff
@@ -516,6 +516,45 @@ class STObject : public STBase, public CountedObject<STObject>
     OptionalProxy<T>
     operator[](OptionaledField<T> const& of);
 
+    /** Return the value of a field.
+
+        Throws:
+
+            STObject::FieldErr if the field is
+            not present.
```
miguelportilla, line 524: Nit:
`@throws STObject::FieldErr if the field is not present.`

src/ripple/protocol/STObject.h
```diff
@@ -516,6 +516,45 @@ class STObject : public STBase, public CountedObject<STObject>
...
+    boost::optional<std::decay_t<typename T::value_type>>
+    at(OptionaledField<T> const& of) const;
+
+    /** Return a modifiable field value.
+
+        Throws:
+
+            STObject::FieldErr if the field is
+            not present.
+    */
```
miguelportilla, line 544: Same as the previous comment.

src/ripple/app/misc/impl/TxQ.cpp
ximinez: Thanks for this code! As I looked it over, I realized that there's no need for the `nextSeqProx` variable at all. Here's why:
* At https://github.com/miguelportilla/rippled/commit/80150588718bd68d567c520e5f90e54bb0eb460c#diff-5872318276bb51e9a07456270eb57f6760f41902adbad461db66b51a70a176afR1031, we know that `txSeqProx` is a sequence.
* At https://github.com/miguelportilla/rippled/commit/80150588718bd68d567c520e5f90e54bb0eb460c#diff-5872318276bb51e9a07456270eb57f6760f41902adbad461db66b51a70a176afR1134, we only look at `nextSeqProx` if `txSeqProx` is a ticket (not a sequence). That means that `nextSeqProx` will *always* be unseated, and thus `nextQueueableSeqImpl` will always get called.

So what I can do is get rid of `nextSeqProx`, change the first check back to the original code, and call `nextQueueableSeqImpl` when bumping the account sequence.

src/ripple/app/tx/impl/CreateOffer.cpp
ximinez: After some further discussions and examination, this new amendment is really just controlling a minor optimization, and really isn't worth it.

miguelportilla (approved):
LGTM 👍

PR #3640, opened 2020-10-16
Severa improvements / simplifications to the codebase

nbougalis:


nbougalis:
So, some of our build matrix (e.g. g++7) compiles with `-std=c++1z` and is missing full C++17 support (e.g., no `<charconv>`). Since we've gone to full C++17 we should remove any builds from the matrix that don't support it..

src/ripple/app/ledger/impl/InboundLedger.cpp
```diff
@@ -947,14 +947,20 @@ InboundLedger::receiveNode(protocol::TMLedgerData& packet, SHAMapAddNode& san)
...
-            if (nodeID.isRoot())
+            auto const nodeID = deserializeSHAMapNodeID(node.nodeid());
+
+            if (!nodeID)
+            {
+                san.incInvalid();
+                return;
+            }
+
+            if (nodeID->isRoot())
```
cjcobb23, line 958: Suggest combining these two `if`s into an `if` `else`

cjcobb23 (approved):
Left some non-blocking suggestions. Code looks good to me

src/ripple/shamap/impl/SHAMapTreeNode.cpp
```diff
@@ -351,112 +359,88 @@ SHAMapTreeNode::updateHash()
 }
 
 void
-SHAMapInnerNode::addRaw(Serializer& s, SHANodeFormat format) const
+SHAMapInnerNode::serializeForWire(Serializer& s) const
 {
-    assert((format == snfPREFIX) || (format == snfWIRE) || (format == snfHASH));
-
-    if (mType == tnERROR)
-        Throw<std::runtime_error>("invalid I node type");
```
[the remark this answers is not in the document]
nbougalis: The value `tnERROR` was unused, so it's impossible for this to ever trigger.

src/ripple/app/ledger/impl/InboundLedger.cpp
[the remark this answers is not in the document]
nbougalis: I avoid `else` when there's a `return` inside the body of a previous `if`. I think that it's cleaner to understand.

src/ripple/app/ledger/impl/InboundLedger.cpp
```diff
@@ -947,14 +947,20 @@ InboundLedger::receiveNode(protocol::TMLedgerData& packet, SHAMapAddNode& san)
...
                 san += map.addRootNode(
                     rootHash, makeSlice(node.nodedata()), filter.get());
             else
                 san += map.addKnownNode(
-                    nodeID, makeSlice(node.nodedata()), filter.get());
+                    *nodeID, makeSlice(node.nodedata()), filter.get());
```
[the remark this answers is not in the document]
nbougalis: Agreed. I may fix this, but I don't know if `clang-format` will remove it.

src/ripple/app/ledger/Ledger.cpp
```diff
@@ -1173,13 +1173,13 @@ loadLedgerHelper(std::string const& sqlSuffix, Application& app, bool acquire)
...
+        (void)ledgerHash.parseHex(*sLedgerHash);
     if (sPrevHash)
-        prevHash.SetHexExact(*sPrevHash);
+        (void)prevHash.parseHex(*sPrevHash);
     if (sAccountHash)
-        accountHash.SetHexExact(*sAccountHash);
+        (void)accountHash.parseHex(*sAccountHash);
     if (sTransHash)
-        transHash.SetHexExact(*sTransHash);
+        (void)transHash.parseHex(*sTransHash);
```
seelabs, line 1182: Given how often we ignore the return value of `parseHex` (here and in many other files as well), what do you think about losing the `[[no discard]]` attribute on this function?

src/ripple/basics/CountedObject.h
```diff
@@ -45,47 +46,60 @@ class CountedObjects
 
         @internal
     */
-    class CounterBase
+    class Counter
     {
     public:
-        CounterBase() noexcept;
+        Counter(std::string name) noexcept : name_(name), count_(0)
```
seelabs, line 52: `std::move(name)`

src/ripple/basics/base_uint.h
```diff
@@ -363,99 +373,62 @@ class base_uint
...
-        When bStrict is false, the parsing is done in non-strict mode, and, if
-        present, leading whitespace and the 0x prefix will be skipped.
-    */
-    bool
-    SetHex(const char* psz, bool bStrict = false)
+        @param sv A null-terminated string of hexadecimal characters
+        @return true if the input was parsed properly; false otherwise.
+     */
+    [[nodiscard]] bool
+    parseHex(std::string_view sv)
```
seelabs, line 384: `parseHex` is a replacement for `SetHexExact`, but not for the more lenient `SetHex`. This breaks backward compatibility in many RPC functions. It's OK with me if we want to do that, but we would definitely document the change and note the break with backward compatibility. 

I'll tag all the places I found that would break backward compatibility in case it helps with updating docs.

src/ripple/basics/base_uint.h
```diff
@@ -363,99 +373,62 @@ class base_uint
...
-        unsigned char* pOut = end() - ((pEnd - pBegin + 1) / 2);
+        auto parseHexChar = [](char c) {
+            if (auto x = charUnHex(c); x != -1)
+                return static_cast<unsigned char>(x);
 
-        *this = beast::zero;
-
-        if ((pEnd - pBegin) & 1)
-            *pOut++ = charUnHex(*pBegin++);
+            throw c;
```
seelabs, line 400: You already know what I'm going to say :-).

We don't throw on user input. I agree the code is cleaner with the exception, but I'd still probably re-write to lose the `throw`.

src/ripple/protocol/impl/AccountID.cpp
```diff
@@ -143,11 +103,8 @@ noAccount()
 bool
 to_issuer(AccountID& issuer, std::string const& s)
 {
-    if (s.size() == (160 / 4))
-    {
-        issuer.SetHex(s);
+    if (issuer.parseHex(s))
```
seelabs, line 106: Breaks backward compatibility

src/ripple/protocol/impl/STLedgerEntry.cpp
```diff
@@ -25,22 +25,29 @@
...
-    if (!(0u <= type_ &&
-          type_ <= std::min<unsigned>(
-                       std::numeric_limits<std::uint16_t>::max(),
-                       std::numeric_limits<
-                           std::underlying_type_t<LedgerEntryType>>::max())))
+    // The on-ledger representation of a key type is a 16-bit unsigned integer
+    // but the LedgerEntryType enum has a larger range (including negative
+    // values), so catch obviously wrong values:
+    constexpr auto const minValidValue =
+        static_cast<LedgerEntryType>(std::numeric_limits<std::uint16_t>::min());
```
seelabs, line 39: I'd probably just say "0", but I can see why you might prefer this way.

src/ripple/protocol/impl/STPathSet.cpp
```diff
@@ -156,7 +156,6 @@ Json::Value STPath::getJson(JsonOptions) const
         auto const iType = it.getNodeType();
 
         elem[jss::type] = iType;
-        elem[jss::type_hex] = strHex(iType);
```
seelabs, line 159: Breaks backward compatibility

src/ripple/rpc/handlers/AccountChannels.cpp
```diff
@@ -121,7 +121,7 @@ doAccountChannels(RPC::JsonContext& context)
         if (!marker.isString())
             return RPC::expected_field_error(jss::marker, "string");
 
-        if (!startAfter.SetHex(marker.asString()))
+        if (!startAfter.parseHex(marker.asString()))
```
seelabs, line 124: Breaks backward compatibility

src/ripple/rpc/handlers/AccountLines.cpp
```diff
@@ -139,7 +139,9 @@ doAccountLines(RPC::JsonContext& context)
         if (!marker.isString())
             return RPC::expected_field_error(jss::marker, "string");
 
-        startAfter.SetHex(marker.asString());
+        if (!startAfter.parseHex(marker.asString()))
```
seelabs, line 142: Breaks backward compatibility

src/ripple/rpc/handlers/AccountObjects.cpp
```diff
@@ -134,13 +134,13 @@ doAccountObjects(RPC::JsonContext& context)
         if (!std::getline(ss, s, ','))
             return RPC::invalid_field_error(jss::marker);
 
-        if (!dirIndex.SetHex(s))
+        if (!dirIndex.parseHex(s))
```
seelabs, line 137: Breaks backward compatibility

src/ripple/rpc/handlers/AccountObjects.cpp
```diff
@@ -134,13 +134,13 @@ doAccountObjects(RPC::JsonContext& context)
...
             return RPC::invalid_field_error(jss::marker);
 
         if (!std::getline(ss, s, ','))
             return RPC::invalid_field_error(jss::marker);
 
-        if (!entryIndex.SetHex(s))
+        if (!entryIndex.parseHex(s))
```
seelabs, line 143: Breaks backward compatibility

src/ripple/rpc/handlers/AccountOffers.cpp
```diff
@@ -101,7 +101,9 @@ doAccountOffers(RPC::JsonContext& context)
         if (!marker.isString())
             return RPC::expected_field_error(jss::marker, "string");
 
-        startAfter.SetHex(marker.asString());
+        if (!startAfter.parseHex(marker.asString()))
```
seelabs, line 104: Breaks backward compatibility

src/ripple/rpc/handlers/AccountTx.cpp
```diff
@@ -165,7 +165,7 @@ parseLedgerArgs(Json::Value const& params)
         }
 
         LedgerHash hash;
-        if (!hash.SetHex(hashValue.asString()))
+        if (!hash.parseHex(hashValue.asString()))
```
seelabs, line 168: Breaks backward compatibility

src/ripple/rpc/handlers/LedgerData.cpp
```diff
@@ -55,7 +55,7 @@ doLedgerData(RPC::JsonContext& context)
     if (isMarker)
     {
         Json::Value const& jMarker = params[jss::marker];
-        if (!(jMarker.isString() && key.SetHex(jMarker.asString())))
+        if (!(jMarker.isString() && key.parseHex(jMarker.asString())))
```
seelabs, line 58: Breaks backward compatibility

src/ripple/rpc/handlers/LedgerEntry.cpp
```diff
@@ -50,7 +50,11 @@ doLedgerEntry(RPC::JsonContext& context)
 
     if (context.params.isMember(jss::index))
     {
-        uNodeIndex.SetHex(context.params[jss::index].asString());
+        if (!uNodeIndex.parseHex(context.params[jss::index].asString()))
```
seelabs, line 53: Breaks backward compatibility

src/ripple/rpc/handlers/LedgerEntry.cpp
```diff
@@ -65,7 +69,12 @@ doLedgerEntry(RPC::JsonContext& context)
     else if (context.params.isMember(jss::check))
     {
         expectedType = ltCHECK;
-        uNodeIndex.SetHex(context.params[jss::check].asString());
+
+        if (!uNodeIndex.parseHex(context.params[jss::check].asString()))
```
seelabs, line 73: Breaks backward compatibility

src/ripple/rpc/handlers/LedgerEntry.cpp
```diff
@@ -74,7 +83,7 @@ doLedgerEntry(RPC::JsonContext& context)
         if (!context.params[jss::deposit_preauth].isObject())
         {
             if (!context.params[jss::deposit_preauth].isString() ||
-                !uNodeIndex.SetHex(
+                !uNodeIndex.parseHex(
```
seelabs, line 86: Breaks backward compatibility

src/ripple/rpc/handlers/LedgerEntry.cpp
```diff
@@ -115,7 +124,11 @@ doLedgerEntry(RPC::JsonContext& context)
         }
         else if (!context.params[jss::directory].isObject())
         {
-            uNodeIndex.SetHex(context.params[jss::directory].asString());
+            if (!uNodeIndex.parseHex(context.params[jss::directory].asString()))
```
seelabs, line 127: Breaks backward compatibility

src/ripple/rpc/handlers/LedgerEntry.cpp
```diff
@@ -132,18 +145,22 @@ doLedgerEntry(RPC::JsonContext& context)
...
             if (context.params[jss::directory].isMember(jss::dir_root))
             {
+                uint256 uDirRoot;
+
                 if (context.params[jss::directory].isMember(jss::owner))
                 {
                     // May not specify both dir_root and owner.
                     jvResult[jss::error] = "malformedRequest";
                 }
+                else if (!uDirRoot.parseHex(
```
seelabs, line 155: Breaks backward compatibility

src/ripple/rpc/handlers/LedgerEntry.cpp
```diff
@@ -173,7 +190,12 @@ doLedgerEntry(RPC::JsonContext& context)
         expectedType = ltESCROW;
         if (!context.params[jss::escrow].isObject())
         {
-            uNodeIndex.SetHex(context.params[jss::escrow].asString());
+            // CHECKME We should return an error here
```
seelabs, line 193: Yes, I believe we should return an error here. There were a couple other places where you made similar changes and I agree those are correct as well.

src/ripple/rpc/handlers/LedgerEntry.cpp
```diff
@@ -173,7 +190,12 @@ doLedgerEntry(RPC::JsonContext& context)
...
+            if (!uNodeIndex.parseHex(context.params[jss::escrow].asString()))
```
seelabs, line 194: Breaks backward compatibility

src/ripple/rpc/handlers/LedgerEntry.cpp
```diff
@@ -200,7 +222,11 @@ doLedgerEntry(RPC::JsonContext& context)
         expectedType = ltOFFER;
         if (!context.params[jss::offer].isObject())
         {
-            uNodeIndex.SetHex(context.params[jss::offer].asString());
+            if (!uNodeIndex.parseHex(context.params[jss::offer].asString()))
```
seelabs, line 225: Breaks backward compatibility

src/ripple/rpc/handlers/LedgerEntry.cpp
```diff
@@ -271,7 +303,11 @@ doLedgerEntry(RPC::JsonContext& context)
         expectedType = ltTICKET;
         if (!context.params[jss::ticket].isObject())
         {
-            uNodeIndex.SetHex(context.params[jss::ticket].asString());
+            if (!uNodeIndex.parseHex(context.params[jss::ticket].asString()))
```
seelabs, line 306: Breaks backward compatibility

src/ripple/rpc/handlers/LedgerRequest.cpp
```diff
@@ -55,7 +55,7 @@ doLedgerRequest(RPC::JsonContext& context)
     if (hasHash)
     {
         auto const& jsonHash = context.params[jss::ledger_hash];
-        if (!jsonHash.isString() || !ledgerHash.SetHex(jsonHash.asString()))
+        if (!jsonHash.isString() || !ledgerHash.parseHex(jsonHash.asString()))
```
seelabs, line 58: Breaks backward compatibility

src/ripple/rpc/handlers/TransactionEntry.cpp
```diff
@@ -57,7 +57,11 @@ doTransactionEntry(RPC::JsonContext& context)
         uint256 uTransID;
         // XXX Relying on trusted WSS client. Would be better to have a strict
         // routine, returning success or failure.
-        uTransID.SetHex(context.params[jss::tx_hash].asString());
+        if (!uTransID.parseHex(context.params[jss::tx_hash].asString()))
```
seelabs, line 60: Breaks backward compatibility

src/ripple/rpc/impl/RPCHelpers.cpp
```diff
@@ -221,7 +221,7 @@ ledgerFromRequest(T& ledger, JsonContext& context)
             return {rpcINVALID_PARAMS, "ledgerHashNotString"};
 
         uint256 ledgerHash;
-        if (!ledgerHash.SetHex(hashValue.asString()))
+        if (!ledgerHash.parseHex(hashValue.asString()))
```
seelabs, line 224: Breaks backward compatibility

src/ripple/shamap/SHAMapNodeID.h
```diff
@@ -20,160 +20,130 @@
...
-    int mDepth;
+    uint256 id_;
+    unsigned int depth_ = 0;
 
 public:
-    SHAMapNodeID();
-    SHAMapNodeID(int depth, uint256 const& hash);
-    SHAMapNodeID(void const* ptr, int len);
+    SHAMapNodeID() = default;
+    SHAMapNodeID(SHAMapNodeID const& other) = default;
```
seelabs, line 41: This file needs to be reformatted (clang format modifies this funct90n).

src/ripple/shamap/SHAMapNodeID.h
```diff
@@ -20,160 +20,130 @@
...
-}
-
-inline uint256 const&
-SHAMapNodeID::getNodeID() const
-{
-    return mNodeID;
-}
+    /** Comparison operators */
+    bool
+    operator<(const SHAMapNodeID& n) const
```
seelabs, line 73: Should be east `const`.

src/ripple/shamap/impl/SHAMapNodeID.cpp
```diff
@@ -49,106 +48,88 @@ SHAMapNodeID::Masks(int depth)
...
+
+    // A SHAMap has exactly 65 levels, so nodes must not exceed that
+    // depth; if they do, this breaks the invariant of never allowing
+    // the construction of a SHAMapNodeID at an invalid depth. We assert
+    // to catch this in debug builds.
+    //
+    // We throw (but never assert) if the node is at level 64, since
+    // entries at that depth are leaf nodes and have no children and even
+    // constructing a child node from them would break the above invariant.
+    assert(depth_ <= SHAMap::leafDepth);
```
seelabs, line 86: I did read the comment, but I'm still not following why this is `<=` and not `<`. I'd expect anything we'd throw a `logic_error` for is something we'd also want to `assert` on.

src/test/rpc/LedgerRPC_test.cpp
```diff
@@ -249,10 +249,10 @@ class LedgerRPC_test : public beast::unit_test::suite
 
         Env env{*this, envconfig(no_admin)};
 
-        env.close();
+        //        env.close();
```
seelabs, line 252: May as well just remove it.

src/test/rpc/TransactionEntry_test.cpp
```diff
@@ -66,7 +66,7 @@ class TransactionEntry_test : public beast::unit_test::suite
             auto const result =
                 env.client().invoke("transaction_entry", params)[jss::result];
             BEAST_EXPECT(!result[jss::ledger_hash].asString().empty());
-            BEAST_EXPECT(result[jss::error] == "transactionNotFound");
+            BEAST_EXPECT(result[jss::error] == "malformedRequest");
```
seelabs, line 69: I think we want to give a valid (but not present) value for the hash, rather than `DEADBEEF` or we're changing what this test is testing.

seelabs (approved):
:+1: Love it, great cleanups!

90% of my comments are of the form "This breaks backward compatibility" due to the stricter requirements around hex strings. I flagged them so hopefully it will be easier to update the docs.

src/ripple/protocol/impl/STPathSet.cpp
nbougalis: Does not: these parameters are deprecated in the documentation and marked as "subject to being removed".

src/ripple/basics/base_uint.h
HowardHinnant: https://github.com/ripple/rippled/pull/3646/commits/d67aa43edf4227d1d844b1a49e7c1323d310e3ba#r518330716

src/ripple/shamap/SHAMapNodeID.h
```diff
@@ -21,159 +21,128 @@
...
 
     bool
-    operator<(const SHAMapNodeID& n) const;
-    bool
-    operator>(const SHAMapNodeID& n) const;
-    bool
-    operator<=(const SHAMapNodeID& n) const;
-    bool
-    operator>=(const SHAMapNodeID& n) const;
+    isRoot() const
```
HowardHinnant, line 47: Stylistic:  This is reverting a style that I find better, but I'm not going to loose sleep over it.

I think it is better to define member functions outside of the class declaration even when they are inline so that the reader can more easily get an overview of the class API.

src/ripple/shamap/impl/SHAMapNodeID.cpp
```diff
@@ -49,106 +48,88 @@ SHAMapNodeID::Masks(int depth)
             entry[mask_size - 1] = selector;
         }
     };
+
     static masks_t const masks;
     return masks.entry[depth];
```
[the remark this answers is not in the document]
HowardHinnant: I think we're ok here.  A reference to a function-local static is a pretty common idiom, often used with singletons.

src/ripple/protocol/impl/AccountID.cpp
nbougalis: Kinda: so, the code previously checked for `s.size() == (160/4)` so the only case where this would break is if someone passed in a string with the right length but which included a non-hex character: the previous code would detect that as the end of the hex string. After processing all the hex input the function would execute: `return !*pEnd;` which should duplicate the semantics of `SetHexExact`: if `*pEnd` is a `NUL` character it returns `true`, othrwise it returns `false`.

src/ripple/rpc/handlers/LedgerEntry.cpp
nbougalis: We do, so the comment is obsolete.

src/ripple/shamap/impl/SHAMapNodeID.cpp
nbougalis: You know, I had convinced myself that this was right but now I don't get it either.

nbougalis:
I've addressed additional review concerns and will close this as it's included in #3646 and will be further tracked there. Thanks! :vulcan_salute:

PR #3644, opened 2020-10-23
Add validation/proposal reduce-relay feature negotiation to the handshake, number of improvements and fixes

gregtatcam:

Add validation/proposal reduce-relay feature negotiation to the handshake
Make squelch duration proportional to a number of peers that can be squelched
Refactor makeRequest()/makeResponse() to facilitate handshake unit-testing
Fix compression enable flag for inbound peer
Fix compression algorithm parsing in the header parser
Fix squelch duration in onMessage(TMSquelch)
FIXES 3624, 3639, 3641

### Context of Change

Add validation/proposal reduce-relay HTTP handshake. This is required to exclude peers not supporting this feature from the 
reduce-relay algorithm and reduce a number of redundant `TMSquelch` messages sent to the peers not supporting the feature.

Random squelch duration of {5, 10} minutes works well with a fewer number of peers. As the number of peers, which a node has increases, the time lapse between peer's squelches decreases.  For instance, if a node has 100 peers then the time lapse between squelches could be in average 5 seconds. Because it takes more than 5 seconds for a node to select new peers as the source of validation and proposal messages, it means there will be load spikes. To mitigate the load spike the squelch duration is made proportional to a number of peers that can be squelched.

`AttemptConnect::makeRequest()` and `PeerImp::makeResponse()` are refactored as the standalone functions in `Handshake` module. This enables unit-testing of HTTP handshake.

To enable compression flag for an inbound peer, the code is checking the HTTP request's compression header. This is not sufficient since an inbound peer may support the compression but the server might have the compression disabled. In which case the link to the peer should not have the compression enabled. This issue is introduced in 1.6.0.

Compression algorithm is fetched from the first byte of the protocol's message header. It should be first 4 bits instead. This issue is introduced in 1.6.0.

`TMSquelch` protocol message has the squelch duration set to milliseconds. In `onMessage(TMSquelch)` the squelch duration is fetched in seconds. The issue is flagged by re-calculating the squelch duration when the duration is invalid. This issue is introduced in 1.7.0-b1.

### Type of Change

- [x ] Tests (You added tests for code that already exists, or your new feature included in this PR)

## Test Plan

Added new unit-tests in `reduce_relay` to test squelch duration and reduce-relay handshake.
Added new unit-test in `compression` to test compression handshake


src/ripple/overlay/impl/Handshake.cpp
```diff
@@ -291,4 +292,57 @@ verifyHandshake(
...
+    m.method(boost::beast::http::verb::get);
+    m.target("/");
+    m.version(11);
+    m.insert("User-Agent", BuildInfo::getFullVersionString());
+    m.insert("Upgrade", supportedProtocolVersions());
+    m.insert("Connection", "Upgrade");
+    m.insert("Connect-As", "Peer");
+    m.insert("Crawl", crawl ? "public" : "private");
+    if (config.COMPRESSION)
+        m.insert("X-Offer-Compression", "lz4");
```
pwang200, line 308: Placeholder for the "X-offer-Feature" design to be discussed.

pwang200 (commented):
Look good. I will continue the review after the "X-offer-Feature" discussion.

src/ripple/overlay/impl/OverlayImpl.cpp
```diff
@@ -1385,7 +1385,7 @@ void
 OverlayImpl::unsquelch(PublicKey const& validator, Peer::id_t id) const
 {
     if (auto peer = findPeerByShortID(id);
-        peer && app_.config().REDUCE_RELAY_SQUELCH)
+        peer && app_.config().VP_REDUCE_RELAY_SQUELCH)
```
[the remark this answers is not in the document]
gregtatcam: This is useful for debugging really. Neither `squelch()` or `unsquelch()` will be ever called if the feature is disabled at the protocol level.  If the feature is enabled and `VP_REDUCE_RELAY_SQUELCH` is disabled then squelch/unsquelch are never sent to peers but the logic otherwise works. For instance can connect to the mainnet with the feature enabled but squelch disabled and see all the interactions but without affecting any connected peers.

src/ripple/overlay/impl/PeerImp.cpp
```diff
@@ -2127,10 +2107,10 @@ PeerImp::onMessage(std::shared_ptr<protocol::TMValidation> const& m)
             // peer receives within IDLED seconds since the message has been
             // relayed. Wait WAIT_ON_BOOTUP time to let the server establish
             // connections to peers.
-            if (app_.config().REDUCE_RELAY_ENABLE && (bool)relayed &&
-                (stopwatch().now() - *relayed) < squelch::IDLED &&
-                squelch::epoch<std::chrono::minutes>(UptimeClock::now()) >
-                    squelch::WAIT_ON_BOOTUP)
+            if (vpReduceRelayEnabled_ && (bool)relayed &&
+                (stopwatch().now() - *relayed) < reduce_relay::IDLED &&
```
nbougalis, line 2111: Random thought: these all-uppercase values are a little too much visual noise. Also, statements like `                reduce_relay::epoch<std::chrono::minutes>(UptimeClock::now()) > reduce_relay::WAIT_ON_BOOTUP` will be `false` and then permanently switch to `true`. It's almost better to have that as a separate gate.

src/ripple/overlay/ReduceRelayCommon.h
```diff
@@ -0,0 +1,115 @@
...
+#define RIPPLE_OVERLAY_REDUCERELAYCOMMON_H_INCLUDED
+
+#include <chrono>
+#include <stdlib.h>
+
+namespace ripple {
+
+namespace reduce_relay {
+
+using namespace std::chrono;
```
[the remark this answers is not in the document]
HowardHinnant: Agreed.  using declarations should be limited to function scope.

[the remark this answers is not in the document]
gregtatcam: Changed.

src/ripple/overlay/Slot.h
```diff
@@ -382,6 +388,22 @@ Slot<clock_type>::update(
     }
 }
 
+template <typename clock_type>
+seconds
+Slot<clock_type>::getSquelchDuration(std::size_t npeers)
+{
+    long const maxExpire = UNSQUELCH_EXPIRE_MULTIPLIER * npeers;
```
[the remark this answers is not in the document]
HowardHinnant: If `UNSQUELCH_EXPIRE_MULTIPLIER` had units of `seconds` and maybe was named `squelch_duration_per_peer` or something like that, I think that would help.

Maybe something like:
```c++
template <typename clock_type>
std::chrono::seconds
Slot<clock_type>::getSquelchDuration(std::size_t npeers)
{
    using namespace std::chrono;
    auto m = std::max(MAX_UNSQUELCH_EXPIRE, SQUELCH_PER_PEER * npeers);
    if (m > OVERALL_MAX_UNSQUELCH_EXPIRE)
    {
        m = OVERALL_MAX_UNSQUELCH_EXPIRE;
        JLOG(journal_.warn())
            << "getSquelchDuration: unexpected squelch duration " << npeers;
    }
    auto d = seconds{ripple::rand_int(MIN_UNSQUELCH_EXPIRE/1s, m/1s)};
    return d;
}
```

src/ripple/overlay/impl/PeerImp.cpp
```diff
@@ -2127,10 +2103,8 @@ PeerImp::onMessage(std::shared_ptr<protocol::TMValidation> const& m)
             // peer receives within IDLED seconds since the message has been
             // relayed. Wait WAIT_ON_BOOTUP time to let the server establish
             // connections to peers.
-            if (app_.config().REDUCE_RELAY_ENABLE && (bool)relayed &&
-                (stopwatch().now() - *relayed) < squelch::IDLED &&
-                squelch::epoch<std::chrono::minutes>(UptimeClock::now()) >
-                    squelch::WAIT_ON_BOOTUP)
+            if (reduceRelayReady() && (bool)relayed &&
```
seelabs, line 2106: No need to cast optionals to bool.

src/ripple/overlay/impl/PeerImp.h
```diff
@@ -603,6 +606,22 @@ class PeerImp : public Peer,
 
     void
     getLedger(std::shared_ptr<protocol::TMGetLedger> const& packet);
+
+    compression::Compressed
+    enableCompression()
```
seelabs, line 611: I'd prefer if we don't put implementations in the class definition. It's easier to see interfaces without them.

I would not make `enableCompression` and `enableVPReduceRelay` members. Consider putting them in the cpp file as free functions, and pass the needed info to them as parameters. Part of the motivation for that is these functions are only used in the constructor, and they depend on `headers_` being initialized before they can be used. By passing the parameter explicitly it makes this dependency clear.

src/ripple/overlay/impl/ProtocolMessage.h
```diff
@@ -141,7 +142,11 @@ parseMessageHeader(
     MessageHeader hdr;
     auto iter = buffersBegin(bufs);
```
seelabs, line 143: Check for an empty buffer sequence?

seelabs (changes requested):

src/ripple/overlay/impl/OverlayImpl.cpp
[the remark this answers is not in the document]
pwang200: So `VP_REDUCE_RELAY_SQUELCH` is only used for testing, right? 
Do we plan to remove it after testing? Either way, we probably should document it more clearly in config.h.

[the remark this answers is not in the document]
gregtatcam: It is useful for testing and I'd like to keep since it provides an ability to test the feature's logic without affecting the relay. Will update the comments.

pwang200 (approved):
Looks good!

src/ripple/overlay/impl/Handshake.cpp
```diff
@@ -25,15 +25,76 @@
 #include <ripple/beast/rfc2616.h>
 #include <ripple/overlay/impl/Handshake.h>
 #include <ripple/protocol/digest.h>
-#include <boost/regex.hpp>
 #include <algorithm>
 #include <chrono>
+#include <regex>
```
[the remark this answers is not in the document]
seelabs: Thanks. Now that it's changed to `boost::regex` we can remove this `#include <regex>` too.

src/ripple/overlay/impl/Handshake.cpp
```diff
@@ -25,15 +25,76 @@
...
+isFeatureValue(
+    boost::beast::http::fields const& headers,
+    std::string const& feature,
+    std::string const& value)
+{
+    auto const [fvalue, present] = getFeatureValue(headers, feature);
+    if (!present)
+        return false;
+    std::regex rx(value);
+    return std::regex_search(fvalue, rx);
```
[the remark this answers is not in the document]
seelabs: That's going to give us some false positives. I.e. if value is `1` and `feature=10`, then we falsely return "true".

src/ripple/overlay/impl/Handshake.cpp
```diff
@@ -25,15 +25,76 @@
...
+bool
+featureEnabled(
+    boost::beast::http::fields const& headers,
+    std::string const& feature)
+{
+    return isFeatureValue(headers, feature, "1");
+}
+
+std::string
+makeFeaturesRequestHeader(Config const& config)
```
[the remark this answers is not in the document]
seelabs: Maybe make a struct for the parameters we need, initialize it from the config and pass that in?

src/ripple/overlay/impl/ProtocolMessage.h
seelabs: Since that's a precondition of this function, I'd document that as part of the docstring. You could also consider adding an `assert`

src/ripple/overlay/impl/Handshake.cpp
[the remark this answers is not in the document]
gregtatcam: Thanks, good catch. Fixed.

src/ripple/overlay/impl/Handshake.cpp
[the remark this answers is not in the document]
gregtatcam: Done

src/ripple/overlay/impl/Handshake.cpp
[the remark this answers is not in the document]
gregtatcam: Changed to passing enable flags for each feature instead.

src/ripple/overlay/impl/ProtocolMessage.h
gregtatcam: Done

src/ripple/overlay/impl/Handshake.cpp
```diff
@@ -34,6 +34,73 @@
...
+    auto const fvalue = getFeatureValue(headers, feature);
+    if (!fvalue)
+        return false;
+    auto test = [&](auto r) {
+        boost::regex rx(r);
+        return boost::regex_search(fvalue.value(), rx);
+    };
+    // could be a single value or a comma separated list of values
+    return test("^" + value + "$") || test("^" + value + ",") ||
+        test("," + value + ",") + test("," + value + "$");
```
seelabs, line 68: That works, but it ends up searching the string four times, doing basically the same search every time. We could do a couple of things here:
1. Tokenize `fvalue`, using comma as a delimiter, and check the token for equality with value. We'd have to write our own tokenizer (strtok requires a non-const `char*`), but that should be straight-forward.
2. Use a more complex regular expression, where the first char can either be a comma or the start, and the last char can either be a comma or the end.
3. Do a single search and check the boundaries.

But I don't think we need to do any of that. There's already a function in the code-base that's perfect for this: `beast::rfc2616::split_commas`. It even correctly handles cases where commas appear in quotes.

src/ripple/overlay/impl/Handshake.cpp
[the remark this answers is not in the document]
seelabs: Thanks! We can revisit this if the number of parameters grows to large.

src/ripple/overlay/impl/Handshake.cpp
gregtatcam: Thanks, updated.

src/test/overlay/handshake_test.cpp
```diff
@@ -0,0 +1,66 @@
+//------------------------------------------------------------------------------
+/*
+    This file is part of rippled: https://github.com/ripple/rippled
+    Copyright (c) 2012, 2020 Ripple Labs Inc.
```
seelabs, line 4: Weird copyright.

src/test/overlay/handshake_test.cpp
```diff
@@ -0,0 +1,66 @@
...
+namespace ripple {
+
+namespace test {
+
+class handshake_test : public beast::unit_test::suite
+{
+public:
+    handshake_test()
+    {
+    }
```
seelabs, line 32: Remove dead code.

seelabs (commented):
Latest patches look good. Let me do one final pass and then I'll give a thumbs up.

src/test/overlay/handshake_test.cpp
gregtatcam: Fixed

src/test/overlay/handshake_test.cpp
gregtatcam: Fixed

src/ripple/overlay/impl/Handshake.cpp
seelabs: I know I suggested `split_commas`, but take a look at rfc2616.h and see if `token_in_list` would be better, as that does not allocate.

seelabs (approved):
:+1: LGTM & nice job on this!. Tho do take a look to see if `token_in_list` would be better than `split_commas`.

src/ripple/overlay/impl/Handshake.cpp
gregtatcam: Changed.

src/ripple/overlay/ReduceRelayCommon.h
```diff
@@ -17,22 +17,31 @@
...
 
 // Peer's squelch is limited in time to
-// rand{MIN_UNSQUELCH_EXPIRE, MAX_UNSQUELCH_EXPIRE}
-static constexpr seconds MIN_UNSQUELCH_EXPIRE = seconds{300};
-static constexpr seconds MAX_UNSQUELCH_EXPIRE = seconds{600};
+// rand{MIN_UNSQUELCH_EXPIRE, max_squelch},
+// where max_squelch is
+// min(max(MAX_UNSQUELCH_EXPIRE_DEFAULT, SQUELCH_PER_PEER * number_of_peers),
+//     MAX_UNSQUELCH_EXPIRE_PEERS)
+static constexpr std::chrono::seconds MIN_UNSQUELCH_EXPIRE =
```
nbougalis, line 35: These could be `static constexrp auto` but no big deal.

src/ripple/overlay/impl/ProtocolMessage.h
```diff
@@ -150,21 +166,21 @@ parseMessageHeader(
...
 
         if (*iter & 0x0C)
         {
             ec = make_error_code(boost::system::errc::protocol_error);
-            return boost::none;
+            return std::nullopt;
         }
 
-        hdr.algorithm = static_cast<compression::Algorithm>(*iter);
+        hdr.algorithm = static_cast<compression::Algorithm>(*iter & 0xF0);
```
nbougalis, line 178: Yeah, nice catch here.

nbougalis (approved):
Left a minor comment about a header.

src/ripple/overlay/ReduceRelayCommon.h
```diff
@@ -17,22 +17,31 @@
 */
 //==============================================================================
 
-#ifndef RIPPLE_OVERLAY_SQUELCHCOMMON_H_INCLUDED
-#define RIPPLE_OVERLAY_SQUELCHCOMMON_H_INCLUDED
+#ifndef RIPPLE_OVERLAY_REDUCERELAYCOMMON_H_INCLUDED
+#define RIPPLE_OVERLAY_REDUCERELAYCOMMON_H_INCLUDED
+
 #include <chrono>
+#include <stdlib.h>
```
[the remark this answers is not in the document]
gregtatcam: :roll_eyes:  Removed.

src/ripple/overlay/ReduceRelayCommon.h
gregtatcam: Updated.

pwang200 (approved):
:+1:

seelabs (approved):
:+1:

PR #3659, opened 2020-11-11
Implement Ledger Replay

pwang200:

This PR implements the [Play Forward Ledger](https://www.xrpchat.com/topic/33071-suggestion-play-forward-ledger/). 

### Details of Change
* Added the `LedgerReplayer` class (parallel to the `InboundLedgers` class) that manages `LedgerReplayTask`s for retrieving ledger deltas from the network and building the ledgers. A `LedgerReplayTask` usually has a range of continuous ledgers. And the ranges of different `LedgerReplayTask`s may overlap. To reduce repeated work, `LedgerReplayTask`s share subtasks,  `SkipListAcquire` (for acquiring skiplists) and `LedgerDeltaAcquire` (for acquiring ledger deltas).
* Split the `PeerSet` class to (1) `PeerSet` that manages a set of peers for retrieving ledger data from peers, and (2) `TimeoutCounter` for counting timeouts.
* Added proof path retrieve and verify to `SHAMap`.
* Added four protobuf messages.
* Added a section to the config file so that this experimental feature can be turned off. Peers can negotiate the feature during handshake (to be refactored once #3644 is merged). 
* Modified the ledger publish logic (in `LedgerMaster`) to use this feature when retrieving ledgers from peers.
* Added three new unit tests. Most of the new code can be tested with them. The ledger publish modification was tested with QE network. 

### Type of Change

### Description and Diagram
Please see the [description of the pattern](https://github.com/ripple/rippled/pull/3659#issuecomment-726275545
) and [diagram](https://github.com/ripple/rippled/pull/3659#issuecomment-726424834) in the comments from @thejohnfreeman 

### Unit Tests
Three new unit tests are added, LedgerReplayer, LedgerReplayerTimeout, and LedgerReplayerLong which is a manual test. 

### Network Tests
Several cases were tested with the QE network, such as different replay task sizes, different overlapping among the tasks, tasks with some ledgers available locally, fallback to InboundLedger in cases that peers did not support the feature. The publish use case was tested by periodically disable/re-enable a node's network to force re-sync. 

## Future Tasks
Use this feature for other use cases, such as consensus, history, etc.


src/ripple/app/ledger/impl/TimeoutCounter.h
```diff
@@ -0,0 +1,140 @@
...
+
+    virtual ~TimeoutCounter() = default;
+
+    /** Schedule a call to queueJob() after mTimerInterval. */
+    void
+    setTimer();
+
+    /** Queue a job to call invokeOnTimer(). */
+    virtual void
+    queueJob();
```
thejohnfreeman, line 95: This can now be private and non-virtual. It could even be a static function in `TimeoutCounter.cpp`, or just inlined (it's only called in one place, `setTimer()`), but if you do that, don't forget to update the comments mentioning it in this file.

src/ripple/app/ledger/impl/InboundLedger.cpp
```diff
@@ -160,25 +167,12 @@ InboundLedger::init(ScopedLockType& collectionLock)
 std::size_t
 InboundLedger::getPeerCount() const
 {
-    return std::count_if(mPeers.begin(), mPeers.end(), [this](auto id) {
-        return app_.overlay().findPeerByShortID(id) != nullptr;
+    std::size_t count = 0;
+    mPeerSet->visitAddedPeers([this, &count](auto id) {
+        if (app_.overlay().findPeerByShortID(id) != nullptr)
+            ++count;
```
thejohnfreeman, line 173: We could go back to using standard algorithms if `PeerSet` was a container, which I think it should be. Short of that, it could just expose a `const` view of the privately-held container.

`PeerSet` is never shared, especially now that it is composed instead of derived. Only the surrounding types are shared, and they already have a lock for their internals that could be used to guard `PeerSet` just like it guards any other member. Then we could get rid of `PeerSet`'s lock.

src/ripple/overlay/impl/PeerSet.cpp
```diff
@@ -84,65 +105,112 @@ PeerSet::addPeers(
...
+}
 
-    if (isDone())
-        return;
+std::size_t
+PeerSetImpl::countAddedPeers()
+{
+    ScopedLockType sl(mLock);
+    return mPeers.size();
+}
```
thejohnfreeman, line 143: If we make `PeerSet` a container, this should be called `size()`.

src/ripple/app/ledger/impl/InboundLedger.cpp
```diff
@@ -644,7 +642,7 @@ InboundLedger::trigger(std::shared_ptr<Peer> const& peer, TriggerReason reason)
             tmGL.set_ledgerseq(mSeq);
         JLOG(m_journal.trace()) << "Sending header request to "
                                 << (peer ? "selected peer" : "all peers");
-        sendRequest(tmGL, peer);
+        mPeerSet->sendRequest(tmGL, protocol::mtGET_LEDGER, peer);
```
thejohnfreeman, line 645: It's possible to use some metaprogramming to compute the type code from the message type. It makes `sendRequest` a little more convenient and reduces the opportunities for error. This patch compiles against `912474b`.

```diff
diff --git a/src/ripple/app/ledger/impl/InboundLedger.cpp b/src/ripple/app/ledger/impl/InboundLedger.cpp
index 84d2e7c1a..c5d96fde6 100644
--- a/src/ripple/app/ledger/impl/InboundLedger.cpp
+++ b/src/ripple/app/ledger/impl/InboundLedger.cpp
@@ -645 +645 @@ InboundLedger::trigger(std::shared_ptr<Peer> const& peer, TriggerReason reason)
-        mPeerSet->sendRequest(tmGL, protocol::mtGET_LEDGER, peer);
+        mPeerSet->sendRequest(tmGL, peer);
@@ -682 +682 @@ InboundLedger::trigger(std::shared_ptr<Peer> const& peer, TriggerReason reason)
-            mPeerSet->sendRequest(tmGL, protocol::mtGET_LEDGER, peer);
+            mPeerSet->sendRequest(tmGL, peer);
@@ -727,2 +727 @@ InboundLedger::trigger(std::shared_ptr<Peer> const& peer, TriggerReason reason)
-                        mPeerSet->sendRequest(
-                            tmGL, protocol::mtGET_LEDGER, peer);
+                        mPeerSet->sendRequest(tmGL, peer);
@@ -755 +754 @@ InboundLedger::trigger(std::shared_ptr<Peer> const& peer, TriggerReason reason)
-            mPeerSet->sendRequest(tmGL, protocol::mtGET_LEDGER, peer);
+            mPeerSet->sendRequest(tmGL, peer);
@@ -792 +791 @@ InboundLedger::trigger(std::shared_ptr<Peer> const& peer, TriggerReason reason)
-                    mPeerSet->sendRequest(tmGL, protocol::mtGET_LEDGER, peer);
+                    mPeerSet->sendRequest(tmGL, peer);
diff --git a/src/ripple/app/ledger/impl/LedgerDeltaAcquire.cpp b/src/ripple/app/ledger/impl/LedgerDeltaAcquire.cpp
index e722dfaed..3c40feb9c 100644
--- a/src/ripple/app/ledger/impl/LedgerDeltaAcquire.cpp
+++ b/src/ripple/app/ledger/impl/LedgerDeltaAcquire.cpp
@@ -116,2 +116 @@ LedgerDeltaAcquire::trigger(std::size_t limit, ScopedLockType& sl)
-                peerSet_->sendRequest(
-                    request, protocol::mtREPLAY_DELTA_REQ, peer);
+                peerSet_->sendRequest(request, peer);
diff --git a/src/ripple/app/ledger/impl/SkipListAcquire.cpp b/src/ripple/app/ledger/impl/SkipListAcquire.cpp
index c0cdb28a9..9c0b39f46 100644
--- a/src/ripple/app/ledger/impl/SkipListAcquire.cpp
+++ b/src/ripple/app/ledger/impl/SkipListAcquire.cpp
@@ -109,2 +109 @@ SkipListAcquire::trigger(std::size_t limit, ScopedLockType& sl)
-                peerSet_->sendRequest(
-                    request, protocol::mtPROOF_PATH_REQ, peer);
+                peerSet_->sendRequest(request, peer);
diff --git a/src/ripple/app/ledger/impl/TransactionAcquire.cpp b/src/ripple/app/ledger/impl/TransactionAcquire.cpp
index cddd059ce..146bc7f75 100644
--- a/src/ripple/app/ledger/impl/TransactionAcquire.cpp
+++ b/src/ripple/app/ledger/impl/TransactionAcquire.cpp
@@ -26,0 +27 @@
+#include <ripple/overlay/impl/ProtocolMessage.h>
@@ -138 +139 @@ TransactionAcquire::trigger(std::shared_ptr<Peer> const& peer)
-        mPeerSet->sendRequest(tmGL, protocol::mtGET_LEDGER, peer);
+        mPeerSet->sendRequest(tmGL, peer);
@@ -172 +173 @@ TransactionAcquire::trigger(std::shared_ptr<Peer> const& peer)
-        mPeerSet->sendRequest(tmGL, protocol::mtGET_LEDGER, peer);
+        mPeerSet->sendRequest(tmGL, peer);
diff --git a/src/ripple/overlay/PeerSet.h b/src/ripple/overlay/PeerSet.h
index cce5e6b35..aa30f63bc 100644
--- a/src/ripple/overlay/PeerSet.h
+++ b/src/ripple/overlay/PeerSet.h
@@ -27,0 +28 @@
+#include "overlay/impl/ProtocolMessage.h"
@@ -62,0 +64,7 @@ public:
+    template <typename MessageType>
+    void
+    sendRequest(MessageType const& message, std::shared_ptr<Peer> const& peer)
+    {
+        this->sendRequest(message, protocolMessageType(message), peer);
+    }
+
diff --git a/src/ripple/overlay/impl/ProtocolMessage.h b/src/ripple/overlay/impl/ProtocolMessage.h
index 593b2ffd3..78fc7350a 100644
--- a/src/ripple/overlay/impl/ProtocolMessage.h
+++ b/src/ripple/overlay/impl/ProtocolMessage.h
@@ -33,0 +34 @@
+#include <ripple.pb.h>
@@ -38,0 +40,18 @@ namespace ripple {
+inline protocol::MessageType
+protocolMessageType(protocol::TMGetLedger const&)
+{
+    return protocol::mtGET_LEDGER;
+}
+
+inline protocol::MessageType
+protocolMessageType(protocol::TMReplayDeltaRequest const&)
+{
+    return protocol::mtREPLAY_DELTA_REQ;
+}
+
+inline protocol::MessageType
+protocolMessageType(protocol::TMProofPathRequest const&)
+{
+    return protocol::mtPROOF_PATH_REQ;
+}
+
```

src/ripple/app/ledger/impl/LedgerReplayTask.cpp
```diff
@@ -0,0 +1,271 @@
...
+void
+LedgerReplayTask::trigger(ScopedLockType& sl)
+{
+    JLOG(m_journal.trace()) << "trigger " << mHash;
+    if (isDone() || !parameter_.full)
+        return;
+
+    if (!parent_)
+    {
+        parent_ = app_.getLedgerMaster().getLedgerByHash(parameter_.startHash);
```
thejohnfreeman, line 134: I think `startHash` can be zero here. Consider this sequence of events in pseudocode:

```cpp
LedgerReplayer::replay() {
  skipList = new SkipListAcquire(...);
  task = new LedgerReplayTask(...);
  assert skipList.addTask(task) == true;
  task.init()
}
```

If we initialize the task before the skip list has returned, how could we possibly know the start ledger ID? It must be empty.

When that happens, I think that `LedgerMaster::getLedgerByHash` returns `nullptr`, but could not verify on a quick check. However, `InboundLedgers::acquire` _asserts_ that the ledger ID is not zero.

`InboundLedgers` does not let us register a callback to notify this task when the start ledger has arrived. Instead, this task checks upon each timeout. I think we need an early exit if `parameter_.startHash` is zero.

thejohnfreeman: Oh, this is guarded by the check on line 130 for `parameter_.full`. Never mind!

src/ripple/app/ledger/impl/LedgerReplayTask.cpp
```diff
@@ -0,0 +1,271 @@
...
+        ScopedLockType sl(mLock);
+        if (!parameter_.update(hash, seq, sList))
+        {
+            JLOG(m_journal.error()) << "Parameter update failed " << mHash;
+            mFailed = true;
+            return;
+        }
+    }
+
+    replayer_.createDeltas(shared_from_this());
```
thejohnfreeman, line 216: Perhaps `this->trigger(sl)` should be called here to start fetching the start ledger if need be.

thejohnfreeman:
Here's my attempt at explaining / understanding the pattern deployed here (and in the existing `InboundLedgers` / `InboundTransactions` architecture). Writing it has helped me understand the control flow, and I hope it can help other reviewers. Please help me revise and correct it. Perhaps we can add it as documentation in the code somewhere.

---

`LedgerReplayer` is a new `Stoppable`. Patterned after two other `Stoppable`s under `JobQueue`---`InboundLedgers` and `InboundTransactions`---it acts like a dispatcher that fires off asynchronous workers. These workers are derived from `TimeoutCounter` to give them timeouts. Think of these workers like asynchronous functions or promises. Like functions, they take a set of parameters. For most of them, that set is just one parameter: a ledger ID. For `LedgerReplayTask`, the set is {reason, finish ledger ID, number of ledgers}. The `Stoppable` maintains a table associating each parameter set with the worker currently working on that parameter set (or a superset of it).

Each `Stoppable` has an entry point. For `LedgerReplayer`, it is `replay`. `replay` creates two workers: a `LedgerReplayTask` and a `SkipListAcquire`. Whenever a worker is created, first the `Stoppable` searches its table to see if there is an existing worker with the same or overlapping parameter set. If it exists, then it is used. If not, then one is created, initialized, and added to the table.

Initialization fires off the underlying asyncrhonous network request and starts the timeout. The parameter set identifying the worker is included in the network request, and copied to the network response. `SkipListAcquire` sends a request for a proof path for the skip list of the desired ledger.

When a response arrives, the `Peer` object knows to send it to the `Stoppable`, which looks up the worker waiting for that response based on the identifying parameter set included in the response.

`LedgerReplayTask` may ask `InboundLedgers` to send requests to acquire the start ledger, but there is no way to attach a callback or be notified when the `InboundLedger` worker completes. All the responses for its messages will be directed to `InboundLedgers`, not `LedgerReplayer`. Instead, `LedgerReplayTask` checks whether the start ledger has arrived every time its timeout expires.

Like a promise, each worker keeps track of whether it is pending (`!isDone()`) or whether it has resolved successfully (`mComplete == true`) or unsuccessfully (`mFailed == true`). It will never exist in both resolved states at once, nor will it return to a pending state after reaching a resolved state.

Like promises, some workers can accept continutations to be called when they reach a resolved state, or immediately if they are already resolved. `SkipListAcquire` and `LedgerDeltaAcquire` both accept continuations of exactly one type, `LedgerReplayTask`, both via a method named `addTask` (compare to `then` on a JavaScript promise). Instead of calling the same function with different arguments based on the resolved state like promises in JavaScript (or like callbacks in Boost.Asio), `SkipListAcquire` calls different methods on the continuation object: `updateSkipList(...)` for success and `cancel()` for failure. `LedgerDeltaAcquire` calls `deltaReady(...)` and `cancel()`, respectively. Continuations cannot be removed explicitly, but they are held by `std::weak_ptr` so they can be removed implicitly.

`LedgerReplayTask` is simultaneously:

1. an asynchronous function,
2. a continuation to one `SkipListAcquire` asynchronous function,
3. a continuation to zero or more `LedgerDeltaAcquire` asynchronous functions, and
4. a continuation to its own timeout.

Each of these roles corresponds to different entry points:

1. `init()`
2. `updateSkipList(...)` or `cancel()`
3. `deltaReady(...)` or `cancel()`
4. `onTimer()`

Each of these entry points does something unique to that entry point. They either (a) transition `LedgerReplayTask` to a terminal failed resolved state (`cancel()` and `onTimer()`) or (b) try to make progress toward the successful resolved state. `init()` and `updateSkipList(...)` call `trigger()` while `deltaReady(...)` calls `tryAdvance()`, but these might as well be called `stage1()` and `stage2()`. There's a similarity between this pattern and the way coroutines are implemented, where every yield saves the spot in the code where it left off and every resume jumps back to that spot.

thejohnfreeman:
I made these diagrams in case they help others visualize the class structure and control flow.

ledger-replay-class-graph
ledger-replay-sequence

cjcobb23 (changes requested):
This is a great piece of work. I am not finished reviewing, but wanted to leave some comments as a first pass.

I would like to remove the use of `std::recursive_mutex` and just use `std::mutex`. I think we should avoid using `recursive_mutex` at all costs. It can almost always be replaced with a regular `mutex` and a clearer concurrency pattern.

There is a considerable amount of complexity that arises from having multiple replay tasks active at the same time, which could be overlapping. I haven't yet thought of a way to reduce this complexity. In the situation where we are only replaying validated ledgers, we could use the sequence numbers to slice up the ranges that are being replayed such that they are not overlapping. However, I think we plan to use forward ledger replay in places where the ledgers might not yet be validated. No action here yet, but just wanted to throw it out there in case anyone has any ideas.

There are also some circular dependencies (which are exacerbated due to the overlapping ranges) that I would like to see removed. Mainly between `LedgerReplayTask` and `SkipListAcquire`, as well as between `LedgerReplayTask` and `LedgerDeltaAcquire`; I think we could remove this circular dependency by just going through the `LedgerReplayer` object, instead of `LedgerDeltaAcquire` and `SkipListAcquire` holding `weak_ptr`s to `LedgerReplayTask`; `LedgerReplayer` could contain some `hash_map`s that map each of these objects to their corresponding `LedgerReplayTask`, or we could just loop through the `LedgerReplayTask`s to find the associated object. I understand why the code was written the way it was, but I think the circular dependencies are a bit hard to follow and reason about, especially in regards to lifetime. Related to that, I would like to see the usage of `std::weak_ptr` reduced, as I think it makes the code brittle and hard to reason about. I am going to keep thinking about how to do this, but wanted to just share my thoughts right now.

Lastly, given the complexity mentioned above, I think the code could use more in function comments, especially whenever objects are shared across multiple `LedgerReplayTask`s.

Overall, the code looks good. It's just that the problem that it's trying to solve is pretty complex.

src/ripple/shamap/impl/SHAMapSync.cpp
```diff
@@ -838,4 +838,83 @@ SHAMap::getFetchPack(
...
+    std::vector<Blob> const& path)
+{
+    if (path.empty() || path.size() > 65)
+        return false;
+
+    SHAMapHash hash{rootHash};
+    auto pathLen = path.size();
+    try
+    {
+        for (int depth = 0; depth < pathLen; ++depth)
```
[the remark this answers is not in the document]
thejohnfreeman: `depth` is still referenced in two statements, once in each branch of the last `if`.

src/ripple/app/ledger/impl/LedgerReplayer.cpp
```diff
@@ -0,0 +1,269 @@
...
+{
+    assert(
+        finishLedgerHash.isNonZero() && totalNumLedgers > 0 &&
+        totalNumLedgers <= LedgerReplayParameters::MAX_TASK_SIZE);
+
+    LedgerReplayTask::TaskParameter parameter(
+        r, finishLedgerHash, totalNumLedgers);
+
+    std::shared_ptr<LedgerReplayTask> task = {};
+    std::shared_ptr<SkipListAcquire> skipList = {};
```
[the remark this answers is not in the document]
thejohnfreeman: That works when the types are not primitive, but for consistency, I like `type identifier{};` before C++17 and `type identifier = {};` after.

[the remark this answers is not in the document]
seelabs: I agree with @cjcobb23 here. I'd rather leave off the `= {}`.

cjcobb23:
I've thought more about how to handle the memoization, shared ownership and circular dependencies in a way that's clearer, and I think it might be better to encapsulate these relationships inside of a separate class that manages several collections, instead of managing the relationships via composition. I wrote some pseudocode to try to demonstrate what I have in mind. This code won't compile as is, and leaves out some details like tasks failing, but I think the general pattern could work. I also left out any mutex locking, though I think that could easily be added to the class, and would make the locking clearer as well.

Basically, the idea is to replace the composition relationships with various maps, and to keep all of the logic related to the collections of the different objects (including the necessary locking, which I omitted in the below pseudocode) inside of one class, which would be a member of `LedgerReplayer`. The `LedgerReplayer` member variables `tasks_`, `deltas_` and `skipLists_` would be moved to this class. We wouldn't need any `std::weak_ptr`s either. I _think_ this is clearer and more maintainable than doing the same thing via composition. Let me know what you think @pwang200 
```
    class TaskMapper
    {

        // all outstanding tasks
        std::vector<std::shared_ptr<LedgerReplayTask>> tasks_;

        // maps a LedgerReplayTask to all of its associated SkipListAcquire
        // objects
        std::unordered_map<
            std::shared_ptr<LedgerReplayTask>,
            std::shared_ptr<SkipListAcquire>>
            skipLists_;

        // maps a LedgerReplayTask to all of its associated LedgerDeltaAcquire
        // objects
        std::unordered_map<
            std::shared_ptr<LedgerReplayTask>,
            std::vector<std::shared_ptr<LedgerDeltaAcquire>>>
            deltas_;

        // maps the skiplist finish hash to all associated tasks
        // this could probably also be a std::multimap
        hash_map<uint256, std::vector<std::shared_ptr<LedgerReplayTask>>>
            skipListToTasks_;

        // maps the ledger hash of the delta to all associated tasks
        // this could probably also be a std::multimap
        hash_map<uint256, std::vector<std::shared_ptr<LedgerReplayTask>>>
            deltaToTasks_;

        // TODO mutexs

        public:

        void
        addSkiplist(std::shared_ptr<LedgerReplayTask> const& task)
        {
            auto finishHash = task.getTaskParameter().finishHash;
            std::shared_ptr<SkipListAcquire> skiplist;

            if (auto it = skipListToTasks_.find(finishHash);
                it != skipListToTasks_.end())
            {
                // The desired skiplist is already being fetched. Get the
                // existing object
                auto othertask = it->second[0];
                skiplist = skipLists_[otherTask];
            }
            else
            {
                // Desired skiplist is not currently being fetched. Start the
                // fetch process
                skiplist = std::make_shared(/* appropriate constructor args */);
                skiplist->init();
                // Initialize with empty vector
                skipListsToTasks_[finishHash] = {};
            }
            // Update LedgerReplayTask to SkipListAcquire mapping
            skipLists_[task] = skiplist;
            // Update SkipListAcquire to LedgerReplayTask mapping
            skipListsToTasks_[finishHash].push_back(task);
        }

        void
        addDeltas(
            std::shared_ptr<LedgerReplayTask> const& task,
            std::vector<uint256> const& skiplist)
        {
            // Initialize empty vector to hold created LedgerDeltaAcquire's
            deltas_[task] = {};
            for (auto const& hash : skiplist)
            {
                std::shared_ptr<LedgerDeltaAcquire> delta;
                if (auto it = deltaToTasks_.find(hash))
                {
                    // The desired delta is already being fetched. Get the
                    // existing object
                    auto otherTask = it->second[0];
                    delta = deltas_[otherTask];
                }
                else
                {
                    // The desired delta is not currently being fetched. Start
                    // the fetch process
                    delta = std::make_shared(/* appropriate constructor args */);
                    delta->init();
                    // Initialize with empty vector
                    deltaToTasks_[hash] = {};
                }
                // Update LedgerReplayTask to LedgerDeltaAcquire mapping
                deltas_[task].push_back(delta);
                // Update LedgerDeltaAcquire to LedgerReplayTask mapping
                deltaToTasks_[hash].push_back(task);
            }
        }

        void
        notifyTasks(std::shared_ptr<SkipListAcquire> const& skiplist)
        {
            for (auto const& task : skipListToTasks_[skiplist->mHash])
            {
                task->updateSkipList(
                    skiplist->mHash, skiplist->ledgerSeq_, skiplist->skipList_);
            }
        }

        void
        notifyTasks(std::shared_ptr<LedgerDeltaAcquire> const& delta)
        {
            for (auto const& task : deltaToTasks_[delta->mHash])
            {
                task->deltaReady(delta->mHash);
            }
        }

        void
        finish(std::shared_ptr<LedgerReplayTask> const& task)
        {
            auto finishHash = task->getTaskParameter().finishHash;
            skipListsToTasks_[finishHash].erase(task);
            if(skipListsToTasks_[finishHash].size() == 0)
               skipListsToTasks_.erase(finishHash);
                
            for (auto const& hash : deltas_[task])
            {
                deltaToTasks_[hash].erase(task);
                if(deltaToTasks_[hash].size() == 0)
                    deltaToTasks_.erase(hash);
            }
            skipLists_.erase(task);
            deltas_.erase(task);
            tasks_.erase(task);
        }
    };

```

src/ripple/app/ledger/impl/TimeoutCounter.h
pwang200: Thanks. I removed `virtual`. It is still a `protested` since the `InboundLedger` calls it. 
BTW, `invokeOnTimer()` is moved to `private`.

src/ripple/app/ledger/impl/InboundLedger.cpp
pwang200: Thanks, now the privately-held container is exposed as a `const&`. 

The lock is removed as discussed.

src/ripple/overlay/impl/PeerSet.cpp
pwang200: Now the privately-held container is exposed as a const&. Please see the comment above.

src/ripple/app/ledger/impl/InboundLedger.cpp
pwang200: Thanks so much! changed.

src/ripple/app/ledger/impl/TimeoutCounter.cpp
```diff
@@ -0,0 +1,110 @@
...
+    , mQueueJobParameter(std::move(jobParameter))
+    , mTimer(app_.getIOService())
+{
+    assert((mTimerInterval > 10ms) && (mTimerInterval < 30s));
+}
+
+void
+TimeoutCounter::setTimer()
+{
+    mTimer.expires_after(mTimerInterval);
```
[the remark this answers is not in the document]
pwang200: Thanks, added. Actually all callers of `setTimer()` already hold the lock.

src/ripple/app/ledger/impl/TimeoutCounter.cpp
```diff
@@ -0,0 +1,110 @@
...
+
+            if (auto ptr = wptr.lock())
+                ptr->queueJob();
+        });
+}
+
+void
+TimeoutCounter::queueJob()
+{
+    if (mQueueJobParameter.jobLimit &&
```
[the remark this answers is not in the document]
pwang200: Thanks, added too.

src/ripple/app/ledger/impl/LedgerReplayTask.cpp
pwang200: You are right! Added.

src/ripple/app/ledger/LedgerReplayTask.h
```diff
@@ -0,0 +1,188 @@
...
+        InboundLedger::Reason reason;
+        uint256 finishHash;
+        std::uint32_t totalLedgers;  // including the start and the finish
+
+        // to be updated
+        std::uint32_t finishSeq = 0;
+        std::vector<uint256> skipList = {};  // including the finishHash
+        uint256 startHash = {};
+        std::uint32_t startSeq = 0;
+        bool full = false;
```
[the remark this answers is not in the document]
pwang200: Thanks! changed.
It started simple, then grew...

src/ripple/app/ledger/LedgerReplayTask.h
```diff
@@ -0,0 +1,188 @@
...
+
+        TaskParameter(
+            InboundLedger::Reason r,
+            uint256 const& finishLedgerHash,
+            std::uint32_t totalNumLedgers);
+
+        /**
```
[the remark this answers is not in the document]
pwang200: Thanks, added.

src/ripple/overlay/PeerSet.h
```diff
@@ -38,93 +38,61 @@ namespace ripple {
...
 
-    This class is an "active" object. It maintains its own timer
-    and dispatches work to a job queue. Implementations derive
-    from this class and override the abstract hook functions in
-    the base.
-
     The data is represented by its hash.
+
+    The original PeerSet is split to two. The logic for counting timeouts
+    is move to the new TimeoutCounter class.  // TODO remove after PR
```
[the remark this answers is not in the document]
pwang200: Will do.

src/ripple/proto/ripple.proto
```diff
@@ -348,3 +353,39 @@ message TMSquelch
     optional uint32 squelchDuration = 3; // squelch duration in milliseconds
 }
 
+enum TMLedgerMapType
+{
+    lmTX = 1;        // transaction map
+    lmAS = 2;        // account state map
```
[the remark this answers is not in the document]
pwang200: Thanks, changed to:
```
enum TMLedgerMapType
{
    lmTRANASCTION   = 1;        // transaction map
    lmACCOUNT_STATE = 2;        // account state map
}
```

src/ripple/shamap/SHAMap.h
```diff
@@ -238,6 +238,27 @@ class SHAMap
         bool fatLeaves,
         std::uint32_t depth) const;
 
+    /**
+     * Get the proof path (from the leaf to the root with siblings) of the key
```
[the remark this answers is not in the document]
pwang200: Thanks, changed to suggested.

src/ripple/shamap/impl/SHAMapSync.cpp
```diff
@@ -838,4 +838,83 @@ SHAMap::getFetchPack(
...
+                           ->getChildHash(nodeId.selectBranch(key));
+            }
+            else
+            {
+                // should exhaust all the blobs now
+                return depth + 1 == pathLen;
+            }
+        }
+    }
+    catch (std::exception const&)
```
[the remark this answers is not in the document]
pwang200: Thanks, added.

src/ripple/shamap/impl/SHAMapSync.cpp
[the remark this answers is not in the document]
pwang200: Thanks, changed to suggested. 
I found I stared at ```int depthToIndex = pathLen - 1 - depth;``` for too long.

src/ripple/app/ledger/LedgerReplayer.h
```diff
@@ -0,0 +1,153 @@
...
+auto constexpr SUB_TASK_FALLBACK_TIMEOUT = std::chrono::milliseconds{1000};
+auto constexpr MAX_NO_FEATURE_PEER_COUNT = 2;
+
+// for LedgerReplayer to limit the number of LedgerReplayTask
+std::uint32_t constexpr MAX_TASKS = 10;
+// for LedgerReplayer to limit the number of ledgers to replay in one task
+std::uint32_t constexpr MAX_TASK_SIZE = 256;
+// to limit the number of LedgerReplay related jobs in JobQueue
+std::uint32_t constexpr MAX_QUEUED_TASKS = 100;
+}  // namespace LedgerReplayParameters
```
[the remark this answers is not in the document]
pwang200: Thanks, rearranged and added more comments. hopefully better.

src/ripple/app/ledger/impl/LedgerReplayer.cpp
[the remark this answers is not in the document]
pwang200: Thanks, removed `={}`.

src/ripple/app/ledger/impl/LedgerReplayer.cpp
```diff
@@ -0,0 +1,269 @@
...
+    bool skipListNeedInit = false;
+    {
+        std::unique_lock<std::recursive_mutex> lock(lock_);
```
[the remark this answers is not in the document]
pwang200: Thanks, changed after changing the lock type to `std::mutex`

src/ripple/app/ledger/LedgerReplayer.h
```diff
@@ -0,0 +1,153 @@
...
+
+    /** Remove completed tasks */
+    void
+    sweep();
+
+    void
+    onStop() override;
+
+private:
+    mutable std::recursive_mutex lock_;
```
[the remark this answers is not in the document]
pwang200: Right, changed. I will running a longer network test later.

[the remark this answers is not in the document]
pwang200: right, changed.

src/ripple/overlay/impl/PeerSet.cpp
```diff
@@ -24,32 +24,53 @@
...
+
+private:
+    // Used in this class for access to boost::asio::io_service and
+    // ripple::Overlay.
+    Application& app_;
+    beast::Journal m_journal;
+    mutable std::recursive_mutex mLock;
+
+    /** The identifiers of the peers we are tracking. */
+    std::set<Peer::id_t> mPeers;
```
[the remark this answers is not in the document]
pwang200: Thanks, changed the names, and the lock is removed after discussing with @thejohnfreeman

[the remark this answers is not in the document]
pwang200: I have the same question. This is from existing code. For the new usages, ie. `LedgerDeltaAcquire` and `SkipListAcquire`, it doesn't matter much since I expect very a small number of items (e.g. 2). For the existing usages, I think it will be a good change. But I don't feel I know enough.

src/ripple/app/ledger/impl/TimeoutCounter.h
```diff
@@ -0,0 +1,140 @@
...
+
+   This loop executes concurrently with another asynchronous sequence,
+   implemented by the subtype, that is trying to make progress and eventually
+   set `mComplete = true`. While it is making progress but not complete, it
+   should set `mProgress = true`, which is passed to onTimer so it can decide
+   whether to postpone failure and reset the timeout. However, if it can
+   complete all its work in one synchronous step (while it holds the lock), then
+   it can ignore `mProgress`.
+
+   This class is split from the PeerSet class.  // TODO remove after PR
```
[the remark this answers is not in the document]
pwang200: Will do.

src/ripple/app/ledger/impl/TimeoutCounter.h
```diff
@@ -0,0 +1,140 @@
...
+    uint256 const mHash;
+    int mTimeouts;
+    bool mComplete;
+    bool mFailed;
+    /** Whether forward progress has been made. */
+    bool mProgress;
+    /** The minimum time to wait between calls to execute(). */
+    std::chrono::milliseconds mTimerInterval;
+
+    QueueJobParameter mQueueJobParameter;
```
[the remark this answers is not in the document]
pwang200: Yes, it is from existing code, but now or never...

src/ripple/overlay/impl/PeerSet.cpp
```diff
@@ -24,32 +24,53 @@
...
+
+    PeerSetImpl(Application& app);
+
+    void
+    addPeers(
+        std::size_t limit,
+        std::function<bool(std::shared_ptr<Peer> const&)> hasItem,
+        std::function<void(std::shared_ptr<Peer> const&)> onPeerAdded) override;
+
+    /** Send a GetLedger message to one or all peers. */
```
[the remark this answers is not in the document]
pwang200: You are right. Changed.

src/ripple/app/ledger/impl/LedgerReplayTask.cpp
```diff
@@ -0,0 +1,271 @@
...
+            JLOG(m_journal.debug())
+                << "Task " << mHash << " got ledger " << l->info().hash
+                << " deltaIndex=" << deltaToBuild
+                << " totalDeltas=" << deltas_.size();
+            parent_ = l;
+            ++deltaToBuild;
+        }
+        else
+            break;
+    }
```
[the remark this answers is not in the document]
pwang200: This will loop all the deltas, instead of yet-to-be-built deltas, right?

src/ripple/app/ledger/LedgerReplayer.h
```diff
@@ -0,0 +1,153 @@
...
+    std::list<std::shared_ptr<LedgerReplayTask>> tasks_;
```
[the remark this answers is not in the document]
pwang200: Right, changed. 
I did a simple test with a single access pattern. They break even at about 22. In real life, the number could be a bit bigger.

pwang200:
@thejohnfreeman thanks for the description and diagram. I linked them from the top PR comment.

@cjcobb23 thanks for the suggestions. Especially the big one from yesterday. I need to think about it.

src/ripple/app/ledger/impl/LedgerReplayTask.cpp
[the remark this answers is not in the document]
cjcobb23: Oh yes, you're right. I think I missed that `deltaToBuild` was a member variable (let's append underscore to the name). I think then I would write this as:
```
    for (; deltaToBuild < deltas_.size(); ++deltaToBuild)
    {
        auto& delta = deltas_[deltaToBuild];
        assert(parent_->seq() + 1 == delta->ledgerSeq_);
        if (auto l = delta->tryBuild(parent_); l)
        {
            JLOG(m_journal.debug())
                << "Task " << mHash << " got ledger " << l->info().hash
                << " totalDeltas=" << deltas_.size();
            parent_ = l;
        }
        else
            return;
    }

    mComplete = true;
    JLOG(m_journal.info()) << "Completed " << mHash;
```
You could also change `deltaToBuild` to an iterator initialized to `deltas_.begin()`, and then write the loop like:
```
    for (; deltaToBuild != deltas_.end(); ++deltaToBuild)
    {
        assert(parent_->seq() + 1 == (*deltaToBuild)->ledgerSeq_);
        if (auto l = (*deltaToBuild)->tryBuild(parent_); l)
        {
            JLOG(m_journal.debug())
                << "Task " << mHash << " got ledger " << l->info().hash
                << " totalDeltas=" << deltas_.size();
            parent_ = l;
        }
        else
            return;
    }

    mComplete = true;
    JLOG(m_journal.info()) << "Completed " << mHash;
```
Either way, I prefer the `for` loop to the `while` loop, and prefer `return` from the else clause to eliminate the `if` statement after the loop.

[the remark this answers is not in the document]
pwang200: Thanks, changed.

pwang200:
@cjcobb23 I see your concern about the circular dependencies and the idea behind the `TaskMapper` suggestion. I discussed with @thejohnfreeman too. 
I did not go that far to remove the composition relationship between the tasks and the subtasks. But I removed the "explicit" `std::weak_ptr`s of the tasks held by the subtasks as an attempt to remove the circular dependencies, as in the "[FOLD] subtasks hold callbacks instead of weak_ptr of tasks" commit. The subtasks don't need to know about the tasks, they just call the callbacks. Please take a look and see if it looks clearer to you.

src/ripple/app/ledger/impl/LedgerDeltaAcquire.cpp
```diff
@@ -174,38 +173,37 @@ LedgerDeltaAcquire::processData(
...
-LedgerDeltaAcquire::addTask(std::shared_ptr<LedgerReplayTask>& task)
+LedgerDeltaAcquire::addDataCallback(
+    InboundLedger::Reason reason,
+    OnDeltaDataCB&& cb)
 {
     ScopedLockType sl(mtx_);
-    tasks_.emplace_back(task);
-
-    auto reason = task->getTaskParameter().reason;
+    dataReadyCallbacks_.emplace_back(cb);
```
cjcobb23, line 193: Do you want to move from `cb`? If so, I think you want `dataReadyCallbacks_.emplace_back(std::move(cb));`

src/ripple/app/ledger/impl/SkipListAcquire.cpp
```diff
@@ -179,23 +178,19 @@ SkipListAcquire::processData(
...
 }
 
 void
-SkipListAcquire::addTask(std::shared_ptr<LedgerReplayTask>& task)
+SkipListAcquire::addDataCallback(OnSkipListDataCB&& cb)
 {
     ScopedLockType sl(mtx_);
-    tasks_.emplace_back(task);
-    if (failed_)
+    dataReadyCallbacks_.emplace_back(cb);
```
cjcobb23, line 188: If you want to move from `cb`, change this to `dataReadyCallbacks_.emplace_back(std::move(cb));`

src/ripple/app/ledger/impl/SkipListAcquire.h
```diff
@@ -132,18 +146,18 @@ class SkipListAcquire final
...
-    notifyTasks(ScopedLockType& sl);
+    notify(ScopedLockType& sl);
 
     InboundLedgers& inboundLedgers_;
     LedgerReplayer& replayer_;
     std::uint32_t ledgerSeq_ = 0;
     std::unique_ptr<PeerSet> peerSet_;
-    std::vector<ripple::uint256> skipList_;
-    std::list<std::weak_ptr<LedgerReplayTask>> tasks_;
+    std::shared_ptr<std::vector<ripple::uint256> const> skipList_;
```
cjcobb23, line 159: Why did you wrap this in `std::shared_ptr`?

src/ripple/app/ledger/impl/LedgerReplayTask.cpp
```diff
@@ -116,6 +116,23 @@ void
...
     JLOG(journal_.debug()) << "Task start " << hash_;
+
+    std::weak_ptr<LedgerReplayTask> wptr = shared_from_this();
+    skipListAcquirer_->addDataCallback(
+        [wptr](
+            bool good,
+            uint256 const& hash,
+            std::uint32_t seq,
+            std::shared_ptr<std::vector<uint256> const> const& skipList) {
+            if (auto sptr = wptr.lock(); sptr)
```
cjcobb23, line 127: In what situation would this check fail? Mainly, when would the `LedgerReplayTask` get destroyed before the `SkipListAcquire` calls it's callback?

cjcobb23 (commented):
I like the callback interface. I think this is an improvement over the previous solution, with the circular dependencies.

src/ripple/app/ledger/impl/LedgerReplayer.cpp
```diff
@@ -0,0 +1,286 @@
...
+                        << parameter.finishHash;
+
+        auto i = skipLists_.find(parameter.finishHash);
+        if (i != skipLists_.end())
+        {
+            skipList = i->second.lock();
+            if (!skipList)
+            {
+                skipLists_.erase(i);
+            }
```
cjcobb23, line 92: I'm concerned that there is a situation where the skiplist is never erased. It seems to me that the skiplist is only erased in 2 situations. First, if we try to add a new `LedgerReplayTask` that needs the same skiplist. Second, if we get the skiplist from the network, but all of the associated `LedgerReplayTask`s have been destroyed. Is there a possibility that neither of these things will happen and then the skiplist never gets erased from `skipLists_`? Maybe I am missing something.

cjcobb23: You could probably eliminate the `skipLists_` datastructure entirely and just loop through `tasks_`. Something like:
```
for(auto const& t: tasks_)
{
     
     if(task.getTaskParameter().finishHash == parameter.finishHash)
         skiplist = task.getSkiplist();
}
```
Then there's no need for the weak ptr semantics. `tasks_` is a small vector anyways, so looping through the tasks like this is probably as fast or even faster than using the `hash_map`.

src/ripple/app/ledger/impl/LedgerDeltaAcquire.cpp
pwang200: Thanks, changed.

src/ripple/app/ledger/impl/SkipListAcquire.cpp
pwang200: Thanks, changed.

src/ripple/app/ledger/impl/SkipListAcquire.h
pwang200: Good question! I had to think and ended up wrapping more data. :-) Let's see if it makes sense. I wrapped the `ledgerSeq` together with the `skipList` as a const object, and changed how tasks get the `skipList` data. I think the `shared_ptr` is cheaper to copy (a skip list is 8KB), and the meaning of an empty `shared_ptr` is clear (in our case, "data is not ready") so it is easier to use. We currently limit the ledger range that can be replayed to 256, to have a simpler design. One future usage case needs larger range. It could be handled with multiple tasks, or we could improve the design of ledger replay and allow multiple `SkipListAcquire` per task. I prefer the latter case. I think the latest changes make that a bit simpler.

src/ripple/app/ledger/impl/LedgerReplayTask.cpp
pwang200: You are right. Currently no. Being defensive.

src/ripple/app/ledger/impl/LedgerReplayer.cpp
pwang200: The `skipLists_` can be cleaned in `sweep()`.

You are right, we can eliminate the `skipLists_` since `tasks_` is so small. But we have `deltas_` anyway, we could increase the limit of tasks, and a task can have multiple `skipList` if we eventually code the change that allow multiple `skipList` per task...

thejohnfreeman (approved):

thejohnfreeman:
Here are the PlantUML sources for the diagrams I included [in this comment](https://github.com/ripple/rippled/pull/3659#issuecomment-726424834). (I had to change the file extensions to satisfy GitHub.)

[ledger-replay-class-graph.txt](https://github.com/ripple/rippled/files/5673387/ledger-replay-class-graph.txt)
[ledger-replay-sequence.txt](https://github.com/ripple/rippled/files/5673388/ledger-replay-sequence.txt)

src/ripple/app/ledger/impl/LedgerReplayTask.cpp
cjcobb23: Can we add an assert then? Since you don't think it should happen

src/ripple/app/ledger/LedgerReplayTask.h
```diff
@@ -0,0 +1,196 @@
...
+        InboundLedger::Reason reason;
+        uint256 finishHash;
+        std::uint32_t totalLedgers;  // including the start and the finish
+
+        // to be updated
+        std::uint32_t finishSeq = 0;
+        std::vector<uint256> skipList = {};  // including the finishHash
+        uint256 startHash = {};
+        std::uint32_t startSeq = 0;
+        bool full = false;
```
cjcobb23, line 59: All of these member names should end with `_`, even if they are public.

src/ripple/app/ledger/impl/LedgerReplayer.cpp
cjcobb23: I see. I think you can get rid of the `erase` actually, and just create a new skiplist in place:
```
        auto i = skipLists_.find(parameter.finishHash);
        if (i != skipLists_.end())
        {
            skipList = i->second.lock();
            if (!skipList)
            {
                skipList = std::make_shared<SkipListAcquire>(
                  app_,
                  inboundLedgers_,
                  parameter.finishHash,
                  peerSetBuilder_->build());
                newSkipList = true;
            }
      }
```
You can then remove the `if` statement starting on line 95

src/ripple/app/ledger/impl/LedgerReplayer.cpp
```diff
@@ -0,0 +1,284 @@
...
+            {
+                std::lock_guard<std::mutex> lock(mtx_);
+                auto i = deltas_.find(*skipListItem);
+                if (i != deltas_.end())
+                {
+                    delta = i->second.lock();
+                    if (!delta)
+                    {
+                        deltas_.erase(i);
+                    }
```
cjcobb23, line 159: I don't think you have to erase here. I think you can just do:
```
if (!delta)
{
    delta = std::make_shared<LedgerDeltaAcquire>(
        app_,
        inboundLedgers_,
        *skipListItem,
        seq,
        peerSetBuilder_->build());
        newDelta = true;
}
```
And then get rid of the `if` starting on line 162

src/ripple/app/ledger/impl/SkipListAcquire.h
```diff
@@ -0,0 +1,174 @@
...
+     * @param sl  lock. this function must be called with the lock
+     */
+    void
+    notify(ScopedLockType& sl);
+
+    InboundLedgers& inboundLedgers_;
+    std::unique_ptr<PeerSet> peerSet_;
+    std::vector<OnSkipListDataCB> dataReadyCallbacks_;
+    std::shared_ptr<SkipListData const> data_;
+    std::uint32_t noFeaturePeerCount = 0;
```
cjcobb23, line 166: append `_` to this member name

src/ripple/app/ledger/impl/TimeoutCounter.h
```diff
@@ -0,0 +1,142 @@
...
+    QueueJobParameter queueJobParameter_;
+
+private:
+    /** Calls onTimer() if in the right state.
+     *  Only called by queueJob().
+     */
+    void
+    invokeOnTimer();
+
+    boost::asio::basic_waitable_timer<std::chrono::steady_clock> mTimer;
```
cjcobb23, line 137: change the name of this member to `timer_` to be consistent

src/ripple/app/ledger/impl/TimeoutCounter.h
```diff
@@ -0,0 +1,142 @@
...
+   whether to postpone failure and reset the timeout. However, if it can
+   complete all its work in one synchronous step (while it holds the lock), then
+   it can ignore `mProgress`.
+
+   This class is split from the PeerSet class.  // TODO remove after PR
+*/
+class TimeoutCounter
+{
+protected:
+    using ScopedLockType = std::unique_lock<std::recursive_mutex>;
```
cjcobb23, line 71: can we make this a non-recursive mutex?

src/ripple/app/ledger/impl/LedgerReplayTask.cpp
pwang200: I realized that there could be a case during the `LedgerReplayer` `Stoppable` stops. The `LedgerReplayTask` could be destroyed right before the `if` statement. Sorry, I realized that when trying to add the assert...

src/ripple/app/ledger/impl/LedgerReplayer.cpp
pwang200: Right, we can get rid of the `erase`. But we cannot remove the `if` because the `if` checks (1) cannot find the `SkipList` in the table, or (2) found but cannot lock the weak pointer. 
The code is currently unchanged, but it could be:
```
        auto i = skipLists_.find(parameter.finishHash_);
        if (i != skipLists_.end())
            skipList = i->second.lock();

        if (!skipList)  // cannot find, or found but cannot lock
        {
            skipList = std::make_shared<SkipListAcquire>(
                app_,
                inboundLedgers_,
                parameter.finishHash_,
                peerSetBuilder_->build());
            skipLists_[parameter.finishHash_] = skipList;
            newSkipList = true;
        }
```

src/ripple/app/ledger/LedgerReplayTask.h
pwang200: thanks, changed.

src/ripple/app/ledger/impl/LedgerReplayer.cpp
pwang200: same as the `SkipList` case.

src/ripple/app/ledger/impl/SkipListAcquire.h
pwang200: thanks, changed.

src/ripple/app/ledger/impl/TimeoutCounter.h
pwang200: thanks, changed.

src/ripple/app/ledger/impl/TimeoutCounter.h
pwang200: The new derived classes don't need `recursive_mutex`. I looked at the existing ones quickly, i.e. `InboundLedger` and `TransactionAcquire`. I didn't see they needed it. But I don't know enough of those classes and how to test those to make the change.

src/ripple/app/ledger/impl/LedgerReplayer.cpp
pwang200: Changed after tests.

src/ripple/app/ledger/impl/LedgerReplayer.cpp
pwang200: Changed after tests.

pwang200:
@thejohnfreeman I modified your description and diagrams a little and added them as a README in the doc folder.

cjcobb23 (approved):
Code looks good to me. Thanks for addressing all of my comments.

nbougalis (approved):

pwang200:
@thejohnfreeman I refactored `onStop()` as we discussed. 
- check `isStopping()` before creating new tasks/subtasks
- cancel subtasks (in addition to tasks) when `onStop()` is called
- move checks of `isDone()` to the entry points of tasks/subtasks, i.e. task init, timeout handler, queued job start, network message received.

nbougalis:
@pwang200: this needs to be rebased against the tip of **develop** and squashed into its final form. Once that's done, I'd also like @JoelKatz to give it a quick lookover.

thejohnfreeman (changes requested):
I made some changes to the documentation that I uploaded in [this commit](https://github.com/thejohnfreeman/rippled/commit/1b7a208fb8e74c8fa87841cdc750e1913c0c1f71). Hopefully they make it easier to read and understand.

pwang200:
thanks @thejohnfreeman for the README changes, cherry-picked.

thejohnfreeman (approved):

pwang200:
thanks @gregtatcam for the review of peer handshake part of the code.

gregtatcam:
> thanks @gregtatcam for the review of peer handshake part of the code.

sure, no pr.

PR #3661, opened 2020-11-12
Shamap sparse inner

seelabs:

A large percentage of inner nodes only store a small number of children. Memory can be saved by storing the inner node's children in sparse arrays. Measurements show that on average a typical SHAMap's inner nodes can be stored using only 25% of the original space. This results in saving approx 3/4 of a gig of ram on a typical rippled server.

### Context of Change

This is part of an effort to reduce the memory footprint of the rippled server.

### Type of Change

- [X] Memory or Compute Optimization

## Before / After

Before this change, SHAMapInnerNodes always kept space for storing all 16 of its children's `hashes` and `shared_ptrs`. After this change, when a child is not present, its `hash` and `shared_ptr` may not be explicitly stored, saving space.

The optimal sizes of the sparse arrays was calculated by measuring the number of children inner nodes had, and calculating the savings for different number of boundaries, and choosing the best. This jupyter notebook shows that work: https://github.com/seelabs/shamap_compress/blob/master/shamap_compress.ipynb

I also benchmarked this "sparse" version vs. the "normal" dense version. Result is here: https://gist.githubusercontent.com/seelabs/e98d2dfe41d495a8d997730154ff1694/raw/9b5c838354aae1ad031abea1348209825835c702/compressed_shamap.md (best on a wide window).

## Test Plan

There are no externally visible changes from this patch. Existing unit tests should cover this change. I will look at CI after it runs to check for unexercised code.

## Future Tasks

The memory in the allocator pools is never reclaimed. This means memory spikes may leave these pools larger than we'd like. We should consider how to reclaim this memory after memory spikes (or identify likely places for memory spikes and reclaim memory after that. I.e. after shard finalization).

There is also some unnecessary copying and zeroing of buffers when changing the number of children (but not terrible; it is much better than a naive solution). I made an attempt to reduce this to the minimal number of copies, but the code ended up being too complex. I will revisit this issue in the future.

### Notes to reviewers

- Only the commits starting from `[FOLD Above] Move SHAMapInnerNode to its own file:` need to be reviewed. It is based off of Nik's branch because that branch made many changes to SHAMap.
- This started out as a simple patch that used one sparse array size and one dense array size. Much of the complexity in the patch comes from tweaks that give us "medium" space savings (see the jupyter notebook referenced later for the calculation showing the extra space saving we get from multiple spare array sizes). We also don't _need_ to use a TaggedPointer. While I decided the extra code complexity was worth the space savings, if you decide this patch is overly complex I do know how to simplify things (at the cost of some extra space, of course).

### DIFF
--- a/src/ripple/shamap/SHAMapInnerNode.h
+++ b/src/ripple/shamap/SHAMapInnerNode.h
@@ -25,6 +25,7 @@
 #include <ripple/shamap/SHAMapItem.h>
 #include <ripple/shamap/SHAMapNodeID.h>
 #include <ripple/shamap/SHAMapTreeNode.h>
+#include <ripple/shamap/impl/TaggedPointer.h>
 
 #include <bitset>
 #include <cstdint>
@@ -38,15 +39,75 @@ namespace ripple {
 class SHAMapInnerNode final : public SHAMapTreeNode,
                               public CountedObject<SHAMapInnerNode>
 {
-    std::array<SHAMapHash, 16> mHashes;
-    std::shared_ptr<SHAMapTreeNode> mChildren[16];
-    int mIsBranch = 0;
-    std::uint32_t mFullBelowGen = 0;
+public:
+    /** Each inner node has 16 children (the 'radix tree' part of the map) */
+    static inline constexpr unsigned int branchFactor = 16;
+
+private:
+    /** Opaque type that contains the `hashes` array (array of type
+       `SHAMapHash`) and the `children` array (array of type
+       `std::shared_ptr<SHAMapInnerNode>`).
+     */
+    TaggedPointer hashesAndChildren_;
+
+    std::uint32_t fullBelowGen_ = 0;
+    std::uint16_t isBranch_ = 0;
 
     static std::mutex childLock;
 
+    /** Convert arrays stored in `hashesAndChildren_` so they can store the
+        requested number of children.
+
+        @param toAllocate allocate space for at least this number of children
+        (must be <= branchFactor)
+
+        @note the arrays may allocate more than the requested value in
+        `toAllocate`. This is due to the implementation of TagPointer, which
+        only supports allocating arrays of 4 different sizes.
+     */
+    void
+    resizeChildArrays(std::uint8_t toAllocate);
+
+    /** Get the child's index inside the `hashes` or `children` array (stored in
+        `hashesAndChildren_`.
+
+        These arrays may or may not be sparse). The optional will be empty is an
+        empty branch is requested and the arrays are sparse.
+
+        @param i index of the requested child
+     */
+    std::optional<int>
+    getChildIndex(int i) const;
+
+    /** Call the `f` callback for all 16 (branchFactor) branches - even if
+        the branch is empty.
+
+        @param f a one parameter callback function. The parameter is the
+        child's hash.
+    */
+    template <class F>
+    void
+    iterChildren(F&& f) const;
+
+    /** Call the `f` callback for all non-empty branches.
+
+        @param f a two parameter callback function. The first parameter is
+        the branch number, the second parameter is the index into the array.
+        For dense formats these are the same, for sparse they may be
+        different.
+    */
+    template <class F>
+    void
+    iterNonEmptyChildIndexes(F&& f) const;
+
 public:
-    SHAMapInnerNode(std::uint32_t cowid);
+    explicit SHAMapInnerNode(
+        std::uint32_t cowid,
+        std::uint8_t numAllocatedChildren = branchFactor);
+
+    SHAMapInnerNode(SHAMapInnerNode const&) = delete;
+    SHAMapInnerNode&
+    operator=(SHAMapInnerNode const&) = delete;
 
     std::shared_ptr<SHAMapTreeNode>
     clone(std::uint32_t cowid) const override;
@@ -71,27 +132,35 @@ public:
 
     bool
     isEmpty() const;
+
     bool
     isEmptyBranch(int m) const;
+
     int
     getBranchCount() const;
+
     SHAMapHash const&
     getChildHash(int m) const;
 
     void
     setChild(int m, std::shared_ptr<SHAMapTreeNode> const& child);
+
     void
     shareChild(int m, std::shared_ptr<SHAMapTreeNode> const& child);
+
     SHAMapTreeNode*
     getChildPointer(int branch);
+
     std::shared_ptr<SHAMapTreeNode>
     getChild(int branch);
+
     virtual std::shared_ptr<SHAMapTreeNode>
     canonicalizeChild(int branch, std::shared_ptr<SHAMapTreeNode> node);
 
     // sync functions
     bool
     isFullBelow(std::uint32_t generation) const;
+
     void
     setFullBelowGen(std::uint32_t gen);
 
@@ -121,34 +190,22 @@ public:
     makeCompressedInner(Slice data);
 };
 
-inline SHAMapInnerNode::SHAMapInnerNode(std::uint32_t cowid)
-    : SHAMapTreeNode(cowid)
-{
-}
-
 inline bool
 SHAMapInnerNode::isEmptyBranch(int m) const
 {
-    return (mIsBranch & (1 << m)) == 0;
-}
-
-inline SHAMapHash const&
-SHAMapInnerNode::getChildHash(int m) const
-{
-    assert(m >= 0 && m < 16);
-    return mHashes[m];
+    return (isBranch_ & (1 << m)) == 0;
 }
 
 inline bool
 SHAMapInnerNode::isFullBelow(std::uint32_t generation) const
 {
-    return mFullBelowGen == generation;
+    return fullBelowGen_ == generation;
 }
 
 inline void
 SHAMapInnerNode::setFullBelowGen(std::uint32_t gen)
 {
-    mFullBelowGen = gen;
+    fullBelowGen_ = gen;
 }
 
 }  // namespace ripple
--- a/src/ripple/shamap/impl/SHAMapInnerNode.cpp
+++ b/src/ripple/shamap/impl/SHAMapInnerNode.cpp
@@ -28,12 +28,12 @@
 #include <ripple/protocol/HashPrefix.h>
 #include <ripple/protocol/digest.h>
 #include <ripple/shamap/SHAMapTreeNode.h>
+#include <ripple/shamap/impl/TaggedPointer.ipp>
 
 #include <openssl/sha.h>
 
 #include <algorithm>
 #include <array>
-#include <atomic>
 #include <iterator>
 #include <mutex>
 #include <utility>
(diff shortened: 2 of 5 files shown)


src/ripple/shamap/impl/SHAMapInnerNode.cpp
```diff
@@ -33,26 +33,718 @@
...
+                f(i, i);
+            }
+        }
+    }
+    else
+    {
+        // sparse case
+        int curHashI = 0;
+        SHAMapHash zero;
+        zero.zero();
```
nbougalis, line 293: Unnecessary and can be removed.

src/ripple/shamap/impl/SHAMapInnerNode.cpp
```diff
@@ -33,26 +33,718 @@
...
+    {
+        // dense case
+        for (int i = 0; i < SHAMapInnerNode::branchFactor; ++i)
+            f(hashes[i]);
+    }
+    else
+    {
+        // sparse case
+        int curHashI = 0;
+        SHAMapHash zero;
```
nbougalis, line 256: Should be `static const`; in fact, since this is a template you should consider making this a `static const` member of `SHAMapInnerNode`

src/ripple/shamap/impl/SHAMapInnerNode.cpp
```diff
@@ -33,26 +33,718 @@
...
+    std::shared_ptr<SHAMapAbstractNode>*cloneChildren, *thisChildren;
+    // structured bindings can't be captured in c++ 17; use tie instead
+    std::tie(std::ignore, cloneHashes, cloneChildren) =
+        p->hashesAndChildren_.getHashesAndChildren();
+    std::tie(std::ignore, thisHashes, thisChildren) =
+        hashesAndChildren_.getHashesAndChildren();
+
+    if (thisIsSparse)
+    {
+        // sparse
```
nbougalis, line 718: Comment here and below seems unnecessary.

src/ripple/shamap/impl/SHAMapInnerNode.cpp
```diff
@@ -176,42 +872,33 @@ SHAMapInnerNode::serializeWithPrefix(Serializer& s) const
...
-        {
-            ret += "\n";
-            ret += std::to_string(i);
-            ret += " = ";
-            ret += to_string(mHashes[i]);
-        }
-    }
+    auto hashes = hashesAndChildren_.getHashes();
+    iterNonEmptyChildIndexes([&](auto branchNum, auto indexNum) {
+        ret += "\nb";
```
nbougalis, line 897: Is the `b` here intentional?

src/ripple/shamap/impl/SHAMapInnerNode.cpp
```diff
@@ -239,51 +946,72 @@ SHAMapInnerNode::shareChild(
...
+    assert((m >= 0) && (m < branchFactor));
+    if (auto const i = getChildIndex(m))
+        return hashesAndChildren_.getHashes()[*i];
+
+    static SHAMapHash const zero = [] {
+        SHAMapHash zero;
+        zero.zero();
+        return zero;
+    }();
+    return zero;
```
nbougalis, line 990: Per my earlier comment, I think this needs to be a `static const` member of `SHAMapInnerNode`.

nbougalis (commented):
First pass. Looks good but left some comments. Will do a more thorough walk-through the logic now.

src/ripple/shamap/SHAMapInnerNode.h
```diff
@@ -121,34 +308,22 @@ class SHAMapInnerNode final : public SHAMapAbstractNode,
...
 {
-    return (mIsBranch & (1 << m)) == 0;
-}
-
-inline SHAMapHash const&
-SHAMapInnerNode::getChildHash(int m) const
-{
-    assert(m >= 0 && m < 16);
-    return mHashes[m];
+    return (isBranch_ & (1 << m)) == 0;
```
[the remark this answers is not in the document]
seelabs: @nbougalis Interesting idea. It is less generated code, but a benchmark shows that's actually _slightly_ slower to lookup the precomputed mask than it is to compute it. The benchmark result I got was:
```
-------------------------------------------------------------------
Benchmark                         Time             CPU   Iterations
-------------------------------------------------------------------
BM_isBranchEmpty_lookup        14.5 ns         14.5 ns     48198924
BM_isBranchEmpty_compute       14.0 ns         14.0 ns     49912717
```

FYI: Here's a gist of the benchmark I used: https://gist.github.com/ee1e6b3bf6176b72c6a8ee6bd3fd8789 (also contains a benchmark for popcount).

src/ripple/shamap/impl/SHAMapInnerNode.cpp
```diff
@@ -33,26 +33,718 @@
...
+    return pmrArrayFuns[boundaryIndex].deallocate(
+        p, arrayChunkSizeBytes[boundaryIndex]);
+}
+#endif
+
+[[nodiscard]] inline int
+popcnt16(std::uint16_t a)
+{
+#if defined(__clang__) || defined(__GNUC__)
+    return __builtin_popcount(a);
```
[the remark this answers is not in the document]
seelabs: @nbougalis Looks like the `__builtin_popcount` is faster than my "fallback" table lookup even when I don't use `-march`. Now I'm interested in what algorithm they used. Interesting that using `-march` actually isn't much faster. This is likely because the constant benchmark overhead is such a significant portion of the timing. Given this, it probably isn't worth jumping through hoops to enable `popcnt` for this case (tho we should still look at what speedups we can get in other places).

Also, just to defend myself, I _did_ know that that it wouldn't use the `popcnt` assembler instruction unless we used `-march` :-)


Popcount with `-march`
```
-------------------------------------------------------------------
Benchmark                         Time             CPU   Iterations
-------------------------------------------------------------------
BM_popcnt_builtin              6.98 ns         6.98 ns    100474566
BM_popcnt_lookup               12.3 ns         12.3 ns     57152228
```
Popcount without `-march`
```
------------------------------------------------------------
Benchmark                  Time             CPU   Iterations
------------------------------------------------------------
BM_popcnt_builtin       7.80 ns         7.80 ns     88919230
BM_popcnt_lookup        12.2 ns         12.2 ns     57612945
```
FYI: Here's a gist of the benchmark I used: https://gist.github.com/seelabs/b22db88a9b6c539a9fe1980cffbcd7e6

src/ripple/shamap/impl/SHAMapInnerNode.cpp
seelabs: Yes. `b` for branch (I assume). So it will print at `b0 = XXX` `b1 = YYY` instead of `0 = XXX`. Note, I did not write that, I just moved the code over. That's been in the code base since 2012 at least.

src/ripple/shamap/SHAMapInnerNode.h
```diff
@@ -38,15 +38,194 @@ namespace ripple {
...
+
+    class TaggedPointer
+    {
+        // Upper bits are the pointer, lowest two bits are the tag
+        // A moved-from object will have a tp_ of zero.
+        std::uintptr_t tp_ = 0;
+        // bit-and with this mask to get the tag bits (lowest two bits)
+        static constexpr std::uintptr_t tagMask = 3;
+        // bit-and with this mask to get the pointer bits (mask out the tag)
+        static constexpr std::uintptr_t ptrMask = ~tagMask;
```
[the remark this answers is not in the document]
seelabs: Good idea. Fixed in `2fa4a8d59e [fold] Doxygen comments and misc cleanups 1:`

src/ripple/shamap/SHAMapInnerNode.h
```diff
@@ -38,15 +38,194 @@ namespace ripple {
...
+    // in array index 1. There are functions to convert between a child's tree
+    // index and the child's index in a sparse array.
+    //
+    // The motivation for this class is saving RAM. A large percentage of inner
+    // nodes only store a small number of children. Memory can be saved by
+    // storing the inner node's children in sparse arrays. Measurements show
+    // that on average a typical SHAMap's inner nodes can be stored using only
+    // 25% of the original space.
+
+    class TaggedPointer
```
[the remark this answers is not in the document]
seelabs: Agreed about the doxygen-style comments, Fixed in `2fa4a8d59e [fold] Doxygen comments and misc cleanups 1:`

Let me think about moving `TaggedPointer` into the cpp file. That class started out as a set of free functions inside the cpp file, but I decided to package together `uintptr_t` and the related functions into that `TaggedPointer` class. If we do decide to move that back into the cpp, I'd probably use free functions again. I'm don't love creating a `TaggedPointer` object just to call member functions like that.

I think I prefer the inner-class implementation, but since you brought it up, let me think about it some more.

src/ripple/shamap/SHAMapInnerNode.h
```diff
@@ -38,15 +38,194 @@ namespace ripple {
...
+        // constructors _must_ be run after this constructor is run. This
+        // constructor is private and only used in places where the
+        // hashes/children constructor are subsequently run.
+        explicit TaggedPointer(RawAllocateTag, std::uint8_t numChildren);
+
+    public:
+        TaggedPointer() = delete;
+        explicit TaggedPointer(std::uint8_t numChildren);
+
+        // Constructor is used change the number of allocated children. Existing
```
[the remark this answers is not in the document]
seelabs: Fixed in `2fa4a8d59e [fold] Doxygen comments and misc cleanups 1:`

src/ripple/shamap/SHAMapInnerNode.h
```diff
@@ -38,15 +38,194 @@ namespace ripple {
...
+
+        ~TaggedPointer();
+
+        // Return the tag and pointer
+        [[nodiscard]] std::pair<std::uint8_t, void*>
+        decode() const;
+
+        // Return the number of elements allocated for each array
+        [[nodiscard]] std::uint8_t
+        numAllocated() const;
```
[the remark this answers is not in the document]
seelabs: Agreed. Fixed in `2fa4a8d59e [fold] Doxygen comments and misc cleanups 1:`

src/ripple/shamap/impl/SHAMapInnerNode.cpp
seelabs: Fixed in `2fa4a8d59e [fold] Doxygen comments and misc cleanups 1:`

src/ripple/shamap/impl/SHAMapInnerNode.cpp
seelabs: Good idea - although I just made it a file-level variable instead of a static member of `SHAMapInnerNode`. Fixed in `2fa4a8d59e [fold] Doxygen comments and misc cleanups 1:`

src/ripple/shamap/impl/SHAMapInnerNode.cpp
seelabs: Fixed in `2fa4a8d59e [fold] Doxygen comments and misc cleanups 1:`

src/ripple/shamap/impl/SHAMapInnerNode.cpp
seelabs: Fixed in `2fa4a8d59e [fold] Doxygen comments and misc cleanups 1:`

src/ripple/shamap/impl/SHAMapInnerNode.cpp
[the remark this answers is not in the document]
seelabs: I peeked at what gcc is using as a fallback if `popcnt` isn't available. It's a series of "logical shift right" followed by "add with carry". Since C++ doesn't have access to the "add with carry" instruction, we can't use this algorithm without using intrinsics (or assembler).

I also searched for other `popcount` implementations. Here's one that's almost as fast as the gcc fallback:
```
[[nodiscard]] inline int
popcount(std::uint32_t v)
{
    v = v - ((v >> 1) & 0x55555555); 
    v = (v & 0x33333333) + ((v >> 2) & 0x33333333);
    return ((v + (v >> 4) & 0xF0F0F0F) * 0x1010101) >> 24;
}
```
When I benchmark that I get:
```
---------------------------------------------------------------
Benchmark                     Time             CPU   Iterations
---------------------------------------------------------------
BM_popcnt_builtin          7.81 ns         7.80 ns     90264669
BM_popcnt_lookup           12.1 ns         12.1 ns     57758234
BM_popcnt_bittwiddle       8.00 ns         8.00 ns     87404423
```

However, that's a little too magic for my tastes. I think the table lookup is a good enough fallback.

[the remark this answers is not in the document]
nbougalis: Yeah, [Bit Twiddling Hacks](https://graphics.stanford.edu/~seander/bithacks.html) are awesome, but I don't know that I'd like to see them in the `rippled` codebase in particular. Again, awesome, but it's the sort of esoteric wizard code whose performance is highly-dependent on the CPU it runs on and it's written so that that nobody understands it, except a guy who used to program the PDP/11 that was in the basement and is now on this 3rd retirement.

Also, the performance of the `__builtin_popcount` fall-back implementation is only marginally slower than hardware operation. It would take a 1.25 billion invocations to "shave" a second off and that's if all we ever did was count bits.

src/ripple/shamap/SHAMapInnerNode.h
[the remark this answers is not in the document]
nbougalis: Interesting and, to be honest, surprising to me. I didn't expect the "naive" version to be as fast as the lookup-based version. I had to test for myself and my results are basically the same; I got **7.18 ns** for the lookup and **7.32 ns** for the compare. Inverted, but either way, the difference seems to basically be noise.

I'm fine with whichever way you choose to go.

src/ripple/shamap/SHAMapInnerNode.h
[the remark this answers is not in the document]
nbougalis: I'm fine either way; I just think that having a `class` in the **.cpp** abstracts away the details of the class, since its interface is never intended to be public and only pollutes here.

Having free functions works, but, at the same time, I think having the class emphasizes that there's some kind of structured type being operated on.

At your discretion, really.

src/ripple/shamap/impl/SHAMapInnerNode.cpp
nbougalis: Huh, you're right. In the diff, it matched up against something else, so I didn't spot the 'b'. I'm fine with it.

nbougalis (approved):

src/ripple/shamap/SHAMapInnerNode.h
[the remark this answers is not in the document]
seelabs: @nbougalis I decided to move the `TaggedPointer` class and split it into a header file included into SHAMapInnerNode.h and a `.ipp` implementation file included into SHAMapInnerNode.cpp. (see the patch `713b7f7612 [FOLD] Move TaggedPoint to its own file`.

I didn't like having the class in the cpp because we'd have to create the class from the opaque `uintptr`, which makes the code a little awkward. I also prefer the class to free functions. I think this .h/.ipp split is a reasonable solution.

src/ripple/shamap/impl/TaggedPointer.ipp
```diff
@@ -0,0 +1,647 @@
...
+{
+    if (this == &other)
+        return *this;
+    destroyHashesAndChildren();
+    tp_ = other.tp_;
+    other.tp_ = 0;
+    return *this;
+}
+
+TaggedPointer::~TaggedPointer()
```
HowardHinnant, line 598: Missing `inline`.

seelabs: Thanks! Fixed in `c338b8280c [fold] Add missing inline`

HowardHinnant:
Using a non-unity build:
```
[100%] Linking CXX executable rippled
Undefined symbols for architecture x86_64:
  "ripple::TaggedPointer::~TaggedPointer()", referenced from:
      std::__1::__shared_ptr_emplace<ripple::SHAMapInnerNode, std::__1::allocator<ripple::SHAMapInnerNode> >::~__shared_ptr_emplace() in SHAMap.cpp.o
      std::__1::__shared_ptr_emplace<ripple::SHAMapInnerNode, std::__1::allocator<ripple::SHAMapInnerNode> >::~__shared_ptr_emplace() in SHAMap.cpp.o
      std::__1::__shared_ptr_emplace<ripple::SHAMapInnerNode, std::__1::allocator<ripple::SHAMapInnerNode> >::__on_zero_shared() in SHAMap.cpp.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

The easiest fix is to move `~TaggedPointer()` from TaggedPointer.ipp to TaggedPointer.h.

However consider creating TaggedPointer.cpp and moving all non-templated sizeable code there, especially those two huge constructors.

seelabs:
@HowardHinnant Fixed the non-unity linker error in `3550ff996 [fold] Fix nonunity linker error` I may re-organize the .h/.ipp into a .h/.ipp/.cpp, but there are tradeoffs for that and I wanted to get a fix out.

HowardHinnant (approved):

seelabs:
Squashed and rebased onto the rollup PR

seelabs:
Rebased onto b5

PR #3668, opened 2020-11-18
Shard size tuning doc

cdy20:
# Shard Size Tuning

The purpose of this document is to compare the sizes of shards containing
varying amounts of ledgers.

## Methodology

One can see visually from a block explorer that typical mainnet ledger
consists of about 30 offer transactions issued by about 8 different accounts,
and several transactions of other types. To simulate this situation and
similar situations we had constructed deterministic shards of differenet sizes,
with different number of offer transactions per ledger and different number
of accounts issuing these offers.

Number of ledgers per shard starts from 256 to 16K with doubling the size on
each step. We considered the following numbers of offers per ledger:
0, 1, 5, 10 and 30. Also we considered case of 1 and 8 accounts issuing offers.
For each constructed deterministic shard we counted its size. Finally we
compared doubled size of the shard with N ledgers and the size of a shard with
2*N ledgers where othere parameters such as number of offers and accounts are
the same. This comparison is sufficient to determine which number or ledgers
per shard leads to less storage size on the disk. 

Note that we minimize total storage size on the disk, but not the size of each
shard because data below shows that size of typical shard is not large than
10G, but sizes of modern disks, even SSD, start from 250G. So there is
no problem to fit a single shard to a disk, even small.

## Raw results table

All sizes of constructed shards are shown in the following table.
Rows corresponds to shard sizes (S) counted in ledgers, columns corresponds
to numbers of offers (O) per ledger. In each cell there are two numbers:
first number corresponds to the case of 1 account issuing offers, the second
number corrspoonds to 8 accounts. Each number is a size of the shard with
given parameters measured in megabytes.

|S\O|0|1|5|10|30|
|---|---|---|---|---|---|
|256|2.2/2.2|3.4/3.3|5.3/7.3|7.7/10.9|17.1/21.9|
|512|4.4/4.5|7.0/7.0|11.2/15.6|16.4/23.7|36.9/47.9|
|1K|8.9/9.0|14.7/14.6|23.7/33.6|35.0/51.0|78.2/ 102.9|
|2K|17.8/18.0|30.5/30.7|50.4/72.2|74.3/ 111.9|166.2/ 221.0|
|4K|35.5/35.9|63.6/64.2|106.2/ 154.8|156.1/ 238.7|354.7/ 476.0|
|8K|71.1/71.9|133.4/ 134.5|222.2/ 328.1|329.1/ 511.2|754.6/ 1021.0|
|16K|142.3/ 143.9|279/9 280.8|465.7/ 698.1|696.4/ 1094.2|1590.5/ 2166.6|

## Preliminary conclusion

If one compares a doubled size of shard with N ledgers and a size of shard
with 2*N ledgers anywhere in the above table than the conlusion will be that
the second number is greater. For example, the following table shows the
percentage by which the second number is greater for the most interesting case
of 30 offers per ledger. The first row corresponds to the case of 1 account
issuing offers, and the second row corresponds to the case of 8 issuing
accounts.

|A\N|256|512|1K|2K|4K|8K|
|---|---|---|---|---|---|---|
|1|8%|6%|6%|6%|7%|6%|5%|
|8|9%|7%|7%|8%|6%|7%|6%|

The common conclusion in this model is that if one doubled the number of 
the ledgers in a shard then the total disk space utilized will raise by 5-9%.

## Adding accounts into consideration

Previous model does not take into account that there are large number of
XRP accounts in the mainnet, and each shard should contain information
about each of these accounts. As of January 2020, there were about 1.9 millions
XRP accounts, and stored information about each of them is not less than 133
bytes. The constant 133 was obtained from debug print of rippled program when
it saves account object to the database. So the actual size of each shard from
raw table should be increased by at least 1.9M * 133 = 252.7M. Thus we obtained
the following table of shard sizes for the most interesting case (30 offers
per ledger and 8 issuing accounts) where S is shard size in ledgers and M is
shard size in megabytes

|S|256|512|1K|2K|4K|8K|16K|
|---|---|---|---|---|---|---|---|
|M|274.6|300.6|355.6|473.7|728.7|1273.7|2419.3|

Now we can see from the last table that even considering minimum assumption
about number of accounts and corresponding additional size of a shard,
doubled size of shard with N ledgers is larger than size of a shard with
2*N ledgers. If number of accounts increase then this inequality will be
even stronger.

## Using mainnet data

Next idea to improve model is to count real shard sizes from mainnet.
We used real 16k-ledgers shards with indexes from 2600 to 3600 with step 100,
and corresponding real 8k-ledgers shards. Each 16k-ledgers shard consists
of two 8k-ledgers shards which are called "corresponding". For example,
16k-ledgers shard with index 2600 consists of two 8k-ledgers shards with
indexes 5200 and 5201.

In the following table we compare size of a 16k-ledgers shard with sum of sizes
of two corresponding 8k-ledgers shards. There we only count size of nudb.dat
file, sizes are in GB. Ratio is the size of two 8k-ledgers shards divided
to the size of 16k-ledgers shard.

|Index|16k-ledgers|8k-ledgers sum|Ratio|
|---|---|---|---|
|2600|2.39|1.49 + 1.63 = 3.12|1.31|
|2700|2.95|1.77 + 1.94 = 3.71|1.26|
|2800|2.53|1.54 + 1.75 = 3.29|1.30|
|2900|3.83|2.26 + 2.35 = 4.61|1.20|
|3000|4.49|2.70 + 2.59 = 5.29|1.18|
|3100|3.79|2.35 + 2.25 = 4.60|1.21|
|3200|4.15|2.54 + 2.43 = 4.97|1.20|
|3300|5.19|3.23 + 2.80 = 6.03|1.16|
|3400|4.18|2.53 + 2.51 = 5.04|1.21|
|3500|5.06|2.90 + 3.04 = 5.94|1.17|
|3600|4.18|2.56 + 2.51 = 5.07|1.21|
|Average|3.89|2.35 + 2.35 = 4.70|1.21|

Note that shard on the disk consists of 4 files each of which can be large too.
These files are nudb.dat, nudb.key, ledger.db, transaction.db. Next table is
similar to previout with the following exception: each number is total size
of these 2 files: nudb.dat and nudb.key. We decided not to count sizes of
ledger.db and transaction.db since these sizes are not permanent instead of
sizes of nudb.* which are permanent for deterministic shards.

|Index|16k-ledgers|8k-ledgers sum|Ratio|
|---|---|---|---|
|2600|2.76|1.73 + 1.89 = 3.62|1.31|
|2700|3.40|2.05 + 2.25 = 4.30|1.26|
|2800|2.91|1.79 + 2.02 = 3.81|1.31|
|2900|4.40|2.62 + 2.71 = 5.33|1.21|
|3000|5.09|3.09 + 2.96 = 6.05|1.19|
|3100|4.29|2.69 + 2.57 = 5.26|1.23|
|3200|4.69|2.90 + 2.78 = 5.68|1.21|
|3300|5.92|3.72 + 3.21 = 6.93|1.17|
|3400|4.77|2.91 + 2.89 = 5.80|1.22|
|3500|5.73|3.31 + 3.47 = 6.78|1.18|
|3600|4.77|2.95 + 2.90 = 5.85|1.23|
|Average|4.43|2.70 + 2.70 = 5.40|1.22|

We can see that in all tables ratio is greater then 1, so using shards with
16 ledgers is preferred.

## Compare 16K shards and 32K shards

To claim that shards with 16K ledgers are the best choice, we also assembled
shards with 32k ledgers per shard with indexes from 1300 to 1800 with step 50
and corresponding shards with 16k ledgers per shard. For example, 32k-ledgers
shard 1800 correnspond to 16k-ledgers shards with indexes 3600 and 3601 etc.

Here are result tables for these shards similar to tables from previous part.
In the first table we only take into consideration sizes of nudb.dat files.

|Index|32k-ledgers|16k-ledgers sum|Ratio|
|---|---|---|---|
|1300|4.00|2.39 + 2.32 = 4.71|1.18|
|1350|5.23|2.95 + 3.02 = 5.97|1.14|
|1400|4.37|2.53 + 2.59 = 5.12|1.17|
|1450|7.02|3.83 + 3.98 = 7.81|1.11|
|1500|7.53|4.49 + 3.86 = 8.35|1.11|
|1550|6.85|3.79 + 3.89 = 7.68|1.12|
|1600|7.28|4.15 + 3.99 = 8.14|1.12|
|1650|8.10|5.19 + 3.76 = 8.95|1.10|
|1700|7.58|4.18 + 4.27 = 8.45|1.11|
|1750|8.95|5.06 + 4.77 = 9.83|1.10|
|1800|7.29|4.18 + 4.02 = 8.20|1.12|
|Average|6.75|3.88 + 3.68 = 7.56|1.12|

In the second table we take into consideration total sizes of files nudb.dat
and nudb.key.

|Index|32k-ledgers|16k-ledgers sum|Ratio|
|---|---|---|---|
|1300|4.59|2.76 + 2.68 = 5.44|1.19|
|1350|5.98|3.40 + 3.47 = 6.87|1.15|
|1400|4.99|2.91 + 2.98 = 5.89|1.18|
|1450|8.02|4.40 + 4.56 = 8.96|1.12|
|1500|8.51|5.09 + 4.39 = 9.48|1.11|
|1550|7.73|4.29 + 4.42 = 8.71|1.13|
|1600|8.20|4.69 + 4.52 = 9.21|1.12|
|1650|9.20|5.92 + 4.29 = 10.21|1.11|
|1700|8.61|4.77 + 4.87 = 9.64|1.12|
|1750|10.09|5.73 + 5.41 = 11.14|1.10|
|1800|8.27|4.77 + 4.59 = 9.36|1.13|
|Average|7.69|4.43 + 4.20 = 8.63|1.12|

## Conclusion

We showed that using shards with 8k ledgers leads to raising required disk size
by 22% in comparison with using shards with 16k ledgers. In the same way,
using shards with 16k ledgers leads to raising required disk space by 12%
in comparison with using shards with 16k ledgers. Note that increase ratio 12%
is much less than 22% so using 32k-ledgers shards will bring us not so much
economy in disk space.

At the same time, size is one thing to compare but there are other aspects.
Smaller shards have an advantage that they take less time to acquire and
finalize. They also make for smaller archived shards which take less time to
download and import. Having more/smaller shards might also lead to better
database concurrency/performance.

It is hard to maintain both size and time parameters by a single optimization
formulae because different chooses for weights of size and time may lead to
different results. But using "common sense" arguments we can compare
16k-ledgers shards and 32k-ledgers as follows: using 32k-ledgers shards give us
12% advantage in size, and about 44% disadvantage in time, because average size
of 16k-ledgers shard is about 56% of average 32k-ledgers shard. At the same,
if we compare 16k-ledgers shards with 8k-ledgers, then the first has 22%
advantage in size and 39% disadvantage in time. So the balance of
advantages/disadvantages is better when we use 16k-ledgers shards.

Thus we recommend use shards with 16K ledgers.

### DIFF
--- a/src/ripple/nodestore/ShardSizeTuning.md
+++ b/src/ripple/nodestore/ShardSizeTuning.md
@@ -1,0 +1,213 @@
+# Shard Size Tuning
+
+The purpose of this document is to compare the sizes of shards containing
+varying amounts of ledgers.
+
+## Methodology
+
+One can see visually from a block explorer that a typical mainnet ledger
+consists of about 30 offer transactions issued by about 8 different accounts,
+and several transactions of other types. To simulate this situation and
+similar situations we have constructed deterministic shards of differenet
+sizes, with varying amounts of offers per ledger and varying amounts of
+accounts issuing these offers.
+
+In the following results table, the number of ledgers per shard ranges from 256
+to 16K with the size doubling the size at each step. We considered the
+following numbers of offers per ledger: 0, 1, 5, 10 and 30. Also we considered
+case of 1 and 8 accounts issuing offers. For each constructed deterministic
+shard we counted its size. Finally we compared doubled size of the shard with
+N ledgers and the size of a shard with 2*N ledgers where othere parameters such
+as number of offers and accounts are the same. This comparison is sufficient to
+determine which number of ledgers per shard leads to less storage size on the
+disk.
+
+Note that we minimize total storage size on the disk, but not the size of each
+shard because data below shows that the size of a typical shard is not larger
+than 10G, but sizes of modern disks, even SSDs, start from 250G. So there is
+no problem to fit a single shard to a disk, even small.
+
+
+## Raw results table
+
+All sizes of constructed shards are shown in the following table.
+Rows corresponds to shard sizes (S) counted in ledgers, columns corresponds
+to numbers of offers (O) per ledger. In each cell there are two numbers:
+first number corresponds to the case of 1 account issuing offers, the second
+number corresponds to 8 accounts. Each number is a size of the shard with
+given parameters measured in megabytes.
+
+|S\O|0|1|5|10|30|
+|---|---|---|---|---|---|
+|256|2.2/2.2|3.4/3.3|5.3/7.3|7.7/10.9|17.1/21.9|
+|512|4.4/4.5|7.0/7.0|11.2/15.6|16.4/23.7|36.9/47.9|
+|1K|8.9/9.0|14.7/14.6|23.7/33.6|35.0/51.0|78.2/ 102.9|
+|2K|17.8/18.0|30.5/30.7|50.4/72.2|74.3/ 111.9|166.2/ 221.0|
+|4K|35.5/35.9|63.6/64.2|106.2/ 154.8|156.1/ 238.7|354.7/ 476.0|
+|8K|71.1/71.9|133.4/ 134.5|222.2/ 328.1|329.1/ 511.2|754.6/ 1021.0|
+|16K|142.3/ 143.9|279/9 280.8|465.7/ 698.1|696.4/ 1094.2|1590.5/ 2166.6|
+
+## Preliminary conclusion
+
+If one compares a doubled size of shard with N ledgers and a size of shard
+with 2*N ledgers anywhere in the above table than the conlusion will be that
+the second number is greater. For example, the following table shows the
+percentage by which the second number is greater for the most interesting case
+of 30 offers per ledger. The first row corresponds to the case of 1 account
+issuing offers, and the second row corresponds to the case of 8 issuing
+accounts.
+
+|A\N|256|512|1K|2K|4K|8K|
+|---|---|---|---|---|---|---|
+|1|8%|6%|6%|6%|7%|6%|5%|
+|8|9%|7%|7%|8%|6%|7%|6%|
+
+The common conclusion in this model is that if one doubled the number of 
+the ledgers in a shard then the total disk space utilized will raise by 5-9%.
+
+## Adding accounts into consideration
+
+Previous model does not take into account that there are large number of
+XRP accounts in the mainnet, and each shard should contain information
+about each of these accounts. As of January 2020, there were about 1.9 million
+XRP accounts, and stored information about each of them is not less than 133
+bytes. The constant 133 was obtained from debug print of rippled program when
+it saves account object to the database. So the actual size of each shard from
+raw table should be increased by at least 1.9M * 133 = 252.7M. Thus we obtained
+the following table of shard sizes for the most interesting case (30 offers
+per ledger and 8 issuing accounts) where S is shard size in ledgers and M is
+shard size in megabytes
+
+|S|256|512|1K|2K|4K|8K|16K|
+|---|---|---|---|---|---|---|---|
+|M|274.6|300.6|355.6|473.7|728.7|1273.7|2419.3|
+
+Now we can see from the last table that even considering minimum assumption
+about number of accounts and corresponding additional size of a shard,
+doubled size of shard with N ledgers is larger than size of a shard with
+2*N ledgers. If number of accounts increase then this inequality will be
+even stronger.
+
+## Using mainnet data
+
+Next idea to improve model is to count real shard sizes from mainnet.
+We used real 16k-ledgers shards with indexes from 2600 to 3600 with step 100,
+and corresponding real 8k-ledgers shards. Each 16k-ledgers shard consists
+of two 8k-ledgers shards which are called "corresponding". For example,
+16k-ledgers shard with index 2600 consists of two 8k-ledgers shards with
+indexes 5200 and 5201.
+
+In the following table we compare size of a 16k-ledgers shard with sum of sizes
+of two corresponding 8k-ledgers shards. There we only count size of nudb.dat
+file, sizes are in GB. Ratio is the size of two 8k-ledgers shards divided
+to the size of 16k-ledgers shard.
+
+|Index|16k-ledgers|8k-ledgers sum|Ratio|
+|---|---|---|---|
+|2600|2.39|1.49 + 1.63 = 3.12|1.31|
+|2700|2.95|1.77 + 1.94 = 3.71|1.26|
+|2800|2.53|1.54 + 1.75 = 3.29|1.30|
+|2900|3.83|2.26 + 2.35 = 4.61|1.20|
+|3000|4.49|2.70 + 2.59 = 5.29|1.18|
+|3100|3.79|2.35 + 2.25 = 4.60|1.21|
+|3200|4.15|2.54 + 2.43 = 4.97|1.20|
+|3300|5.19|3.23 + 2.80 = 6.03|1.16|
+|3400|4.18|2.53 + 2.51 = 5.04|1.21|
+|3500|5.06|2.90 + 3.04 = 5.94|1.17|
+|3600|4.18|2.56 + 2.51 = 5.07|1.21|
+|Average|3.89|2.35 + 2.35 = 4.70|1.21|
+
+Note that shard on the disk consists of 4 files each of which can be large too.
+These files are nudb.dat, nudb.key, ledger.db, transaction.db. Next table is
+similar to previous with the following exception: each number is total size
+of these 2 files: nudb.dat and nudb.key. We decided not to count sizes of
+ledger.db and transaction.db since these sizes are not permanent instead of
+sizes of nudb.* which are permanent for deterministic shards.
+
+|Index|16k-ledgers|8k-ledgers sum|Ratio|
+|---|---|---|---|
+|2600|2.76|1.73 + 1.89 = 3.62|1.31|
+|2700|3.40|2.05 + 2.25 = 4.30|1.26|
+|2800|2.91|1.79 + 2.02 = 3.81|1.31|
+|2900|4.40|2.62 + 2.71 = 5.33|1.21|
+|3000|5.09|3.09 + 2.96 = 6.05|1.19|
+|3100|4.29|2.69 + 2.57 = 5.26|1.23|
+|3200|4.69|2.90 + 2.78 = 5.68|1.21|
+|3300|5.92|3.72 + 3.21 = 6.93|1.17|
+|3400|4.77|2.91 + 2.89 = 5.80|1.22|
+|3500|5.73|3.31 + 3.47 = 6.78|1.18|
+|3600|4.77|2.95 + 2.90 = 5.85|1.23|
+|Average|4.43|2.70 + 2.70 = 5.40|1.22|
+
+We can see that in all tables ratio is greater then 1, so using shards with
+16 ledgers is preferred.
+
+## Compare 16K shards and 32K shards
+
+To claim that shards with 16K ledgers are the best choice, we also assembled
+shards with 32k ledgers per shard with indexes from 1300 to 1800 with step 50
+and corresponding shards with 16k ledgers per shard. For example, 32k-ledgers
+shard 1800 correnspond to 16k-ledgers shards with indexes 3600 and 3601 etc.
+
+Here are result tables for these shards similar to tables from previous part.
+In the first table we only take into consideration sizes of nudb.dat files.
+
+|Index|32k-ledgers|16k-ledgers sum|Ratio|
+|---|---|---|---|
+|1300|4.00|2.39 + 2.32 = 4.71|1.18|
+|1350|5.23|2.95 + 3.02 = 5.97|1.14|
+|1400|4.37|2.53 + 2.59 = 5.12|1.17|
+|1450|7.02|3.83 + 3.98 = 7.81|1.11|
+|1500|7.53|4.49 + 3.86 = 8.35|1.11|
+|1550|6.85|3.79 + 3.89 = 7.68|1.12|
+|1600|7.28|4.15 + 3.99 = 8.14|1.12|
+|1650|8.10|5.19 + 3.76 = 8.95|1.10|
+|1700|7.58|4.18 + 4.27 = 8.45|1.11|
+|1750|8.95|5.06 + 4.77 = 9.83|1.10|
+|1800|7.29|4.18 + 4.02 = 8.20|1.12|
+|Average|6.75|3.88 + 3.68 = 7.56|1.12|
+
+In the second table we take into consideration total sizes of files nudb.dat
+and nudb.key.
+
+|Index|32k-ledgers|16k-ledgers sum|Ratio|
+|---|---|---|---|
+|1300|4.59|2.76 + 2.68 = 5.44|1.19|
+|1350|5.98|3.40 + 3.47 = 6.87|1.15|
+|1400|4.99|2.91 + 2.98 = 5.89|1.18|
+|1450|8.02|4.40 + 4.56 = 8.96|1.12|
+|1500|8.51|5.09 + 4.39 = 9.48|1.11|
+|1550|7.73|4.29 + 4.42 = 8.71|1.13|
+|1600|8.20|4.69 + 4.52 = 9.21|1.12|
+|1650|9.20|5.92 + 4.29 = 10.21|1.11|
+|1700|8.61|4.77 + 4.87 = 9.64|1.12|
+|1750|10.09|5.73 + 5.41 = 11.14|1.10|
+|1800|8.27|4.77 + 4.59 = 9.36|1.13|
+|Average|7.69|4.43 + 4.20 = 8.63|1.12|
+
+## Conclusion
+
+We showed that using shards with 8k ledgers leads to raising required disk size
+by 22% in comparison with using shards with 16k ledgers. In the same way,
+using shards with 16k ledgers leads to raising required disk space by 12%
+in comparison with using shards with 32k ledgers. Note that increase ratio 12%
+is much less than 22% so using 32k-ledgers shards will bring us not so much
+economy in disk space.
+
+At the same time, size is one thing to compare but there are other aspects.
+Smaller shards have an advantage that they take less time to acquire and
+finalize. They also make for smaller archived shards which take less time to
+download and import. Having more/smaller shards might also lead to better
+database concurrency/performance.
+
+It is hard to maintain both size and time parameters by a single optimization
+formulae because different choices for weights of size and time may lead to
+different results. But using "common sense" arguments we can compare
+16k-ledgers shards and 32k-ledgers as follows: using 32k-ledgers shards give us
+12% advantage in size, and about 44% disadvantage in time, because average size
+of 16k-ledgers shard is about 56% of average 32k-ledgers shard. At the same,
+if we compare 16k-ledgers shards with 8k-ledgers, then the first has 22%
+advantage in size and 39% disadvantage in time. So the balance of
+advantages/disadvantages is better when we use 16k-ledgers shards.
+
+Thus we recommend use shards with 16K ledgers.


MarkusTeufelberger:
Did you also run this analysis on actual Mainnet ledgers? There will always be some overhead by duplicating nodes between shards that didn't change. I would have expected this overhead to go mostly linearly from near 100% (every single ledger is its own shard) to 0% (Shards are unlimited in size, there is only one single "shard" - the situation right now).

Let's say there's a ledger range of 64k(binary, so 65536) ledgers. The smallest way to stuff the data for these ledgers in a shard should be a single shard and the largest to create 65536 single shards that all contain a single full ledger. While being a GB+ in size those tiny shards would only differ from their successor in a few kB of transactions and state.

Why does the disk space used grow if there's more duplicate data to be stored (unchanged nodes)? In your results it seems like there's some benefit to storing the _same_ data just split more often, even including duplication of data on top. How did you generate these tests? Is this some weird overhead effect from NuDB?

cdy20:
> Did you also run this analysis on actual Mainnet ledgers? There will always be some overhead by duplicating nodes between shards that didn't change. I would have expected this overhead to go mostly linearly from near 100% (every single ledger is its own shard) to 0% (Shards are unlimited in size, there is only one single "shard" - the situation right now).
> 
> Let's say there's a ledger range of 64k(binary, so 65536) ledgers. The smallest way to stuff the data for these ledgers in a shard should be a single shard and the largest to create 65536 single shards that all contain a single full ledger. While being a GB+ in size those tiny shards would only differ from their successor in a few kB of transactions and state.
> 
> Why does the disk space used grow if there's more duplicate data to be stored (unchanged nodes)? In your results it seems like there's some benefit to storing the _same_ data just split more often, even including duplication of data on top. How did you generate these tests? Is this some weird overhead effect from NuDB?

I corrected the model to tightly correspond with mainnet.

carlhua (commented):
I think this write-up's approach on how to tune the size of the shard is quantitative and presents good rationale on why 16k should be chosen. 

Can we consider adding a section that extrapolates when we will need to halve the number of ledgers in a shard due to the number of accounts increasing?

MarkusTeufelberger:
> If number of accounts increase then this inequality will be even stronger.

This means if the number of accounts increases, 32k shards should be considered, not 8k shards. It would be interesting to see the numbers for 32k, 64k and 128k shards by the way.

carlhua:
> > If number of accounts increase then this inequality will be even stronger.
> 
> This means if the number of accounts increases, 32k shards should be considered, not 8k shards. It would be interesting to see the numbers for 32k, 64k and 128k shards by the way.

I am primarily concerned with the size of the shards. If we were to increase the number of ledgers when the number of accounts increases, then the shard size would also increase. What is your take on that?

MarkusTeufelberger:
Yeah, that can of course be a problem and currently is already a problem (as I said, you can view the current situation as a single huge shard) for example on nuDB where it requires you to have a file that's quite a few TB large on a single volume.

There is far more state to consider than just accounts (I can create gigabytes of ledger state with just a handful of accounts as long as I have the funds to pay for that), so these numbers would always just be estimates. Hence my initial question about verifying these estimates with actual shards from mainnet, because I don't believe that they are just ~2.5 GiB in size.

Personally I would rather state some target size and hard upper limit for shard sizes, e.g. "We aim to have shards that are about 30 GiB each and will definitely reduce the number of ledgers per shards once mainnet consistently reaches shard sizes upwards of 80 GiB". The larger the amount of ledgers per shard, the smaller the overhead, but the larger the increment of storage required. This will only get (MUCH!) worse in the future, with amendments like "Hooks" looming on the horizon which can massively inflate the static state size.

Unfortunately there's also no easy way to make this very dynamic or automatic - a fixed shard size at least means you just need some ledger height and can directly know which shard it belongs to without having to consider how much the ledger state has changed until that ledger height was reached. If you close a shard once a certain amount of storage is used, you then need a place to store that information again (which is implementation specific, yet should likely be common across the network of the `rippled` implementation of the `XRPL` protocol)

To change the size either a cutoff height could be used ("shards contain 64k ledgers until ledger xxxx, 32k ledgers until ledger yyyy and 16k ledger after") or old shards could be split/merged ("all shards now contain 8k ledgers, there will be a background task that splits your old shards"). In the latter case it would make sense to also add the number of ledgers per shard to the shard name, so it is clear that `shard_3_16k` will be split into `shard_6_8k` and `shard_7_8k` which are then a bit larger in sum, but smaller individually.

My take would be: Spin up a full history node, let it generate shards (the current default is 16k ledgers per shard) and plot the shard sizes since 2013, that should give you some better insight. To learn more about actual overheads you can also modify https://github.com/ripple/rippled/blob/76a6956138c4ecd156c5c408f136ed3d6ab7d0c1/src/ripple/nodestore/DatabaseShard.h#L179 recompile and re-build the shard database(s). Then compare disk usage of shard stores between each node. This will give you actionable real world data at the cost of a few AWS instances + storage and some waiting time while the servers work. You can then do all kinds of analysis (extrapolation, outlier detection...) if you treat that data as a time series.

miguelportilla:
@MarkusTeufelberger, thank you for your comments, much appreciated.

I agree we should be using real-world data from the main-net. I don't know if we need to create full history shards, a sampling of every hundredth shard or so may suffice. I also think it is a good idea to compare shards that hold more than 16k ledgers.

MarkusTeufelberger:
Sure, but to create these shards you still need to have nearly full history anyways. The optimal solution might be probably binary search (create the "middle" shard at the moment, then start creating the "middle" shard from each remaining non-sharded part of the database until you have enough samples), but either creating every 100th shard or something more fancy is something that `rippled` doesn't support to my knowledge. It is likely cheaper to just create all shards over a weekend or so than to actually implement that sampling.

miguelportilla:
@MarkusTeufelberger rippled doesn't support the sampling, but the modification to do so is simple. I've been gathering shards since last week for this purpose and will post the data as soon as it is complete.

src/ripple/nodestore/ShardSizeTuning.md
```diff
@@ -0,0 +1,83 @@
+# Shard Size Tuning
+
+The purpose of this document is to compare shard size of 8K and 16K ledgers.
```
undertome, line 3: This sentence sounds ambiguous. Consider: _"... is to compare the sizes of shards containing varying amounts of ledgers."_

src/ripple/nodestore/ShardSizeTuning.md
```diff
@@ -0,0 +1,83 @@
...
+consists of about 30 offer transactions issued by about 8 different accounts,
+and several transactions of other types. To simulate this situation and
+similar situations we had constructed deterministic shards of differenet sizes,
+with different number of offer transactions per ledger and different number
+of accounts issuing these offers.
+
+Number of ledgers per shard starts from 256 to 16K with doubling the size on
+each step. We considered the following numbers of offers per ledger:
+0, 1, 5, 10 and 30. Also we considered case of 1 and 8 accounts issuing offers.
+For each constructed deterministic shard we counted its size. Finally we
```
undertome, line 17: At this point in the document it isn't clear why you're comparing the doubled size of a shard containing _n_ ledgers with the size of a shard containing _2n_ ledgers. Please state why this comparison is necessary.

src/ripple/nodestore/ShardSizeTuning.md
```diff
@@ -0,0 +1,83 @@
...
+|4K|35.5/35.9|63.6/64.2|106.2/ 154.8|156.1/ 238.7|354.7/ 476.0|
+|8K|71.1/71.9|133.4/ 134.5|222.2/ 328.1|329.1/ 511.2|754.6/ 1021.0|
+|16K|142.3/ 143.9|279/9 280.8|465.7/ 698.1|696.4/ 1094.2|1590.5/ 2166.6|
+
+## Preliminary conclusion
+
+If one compares a doubled size of shard with N ledgers and a size of shard
+with 2*N ledgers anythere in the above table than the conlusion will be that
+the second number is greater. For example, the following table shows the
+percentage in which the second number is greater for the most interesting case
```
undertome, line 47: Change to _"... percentage by which ..."_

src/ripple/nodestore/ShardSizeTuning.md
```diff
@@ -0,0 +1,83 @@
...
+of 30 offers per ledger. The first row corresponds to the case of 1 account
+issuing offers, and the second row corresponds to the case of 8 issuing
+accounts.
+
+|A\N|256|512|1K|2K|4K|8K|
+|---|---|---|---|---|---|---|
+|1|8%|6%|6%|6%|7%|6%|5%|
+|8|9%|7%|7%|8%|6%|7%|6%|
+
+The common conclusion in this model is that if one doubled shard size then
```
undertome, line 57: Do you mean that if one doubled the number of ledgers in the shard?

src/ripple/nodestore/ShardSizeTuning.md
```diff
@@ -0,0 +1,83 @@
...
+the total disk space utilized will raise by 5-9%.
+
+## Adding accounts into consideration
+
+Previous model does not take into account that there are large number of
+XRP accounts in the mainnet, and each shard should contain information
+about each of these accounts. As of January 2020, there was about 1.9 millions
```
undertome, line 64: Change to _"... there were about 1.9 million ..."_.

src/ripple/nodestore/ShardSizeTuning.md
```diff
@@ -0,0 +1,83 @@
...
+|1K|8.9/9.0|14.7/14.6|23.7/33.6|35.0/51.0|78.2/ 102.9|
+|2K|17.8/18.0|30.5/30.7|50.4/72.2|74.3/ 111.9|166.2/ 221.0|
+|4K|35.5/35.9|63.6/64.2|106.2/ 154.8|156.1/ 238.7|354.7/ 476.0|
+|8K|71.1/71.9|133.4/ 134.5|222.2/ 328.1|329.1/ 511.2|754.6/ 1021.0|
+|16K|142.3/ 143.9|279/9 280.8|465.7/ 698.1|696.4/ 1094.2|1590.5/ 2166.6|
+
+## Preliminary conclusion
+
+If one compares a doubled size of shard with N ledgers and a size of shard
+with 2*N ledgers anythere in the above table than the conlusion will be that
```
undertome, line 45: anywhere*

src/ripple/nodestore/ShardSizeTuning.md
```diff
@@ -0,0 +1,83 @@
...
+
+## Adding accounts into consideration
+
+Previous model does not take into account that there are large number of
+XRP accounts in the mainnet, and each shard should contain information
+about each of these accounts. As of January 2020, there was about 1.9 millions
+XRP accounts, and stored information about each of them is not less than 133
+bytes. The constant 133 was obtained from debug print of rippled program when
+it saves account object to the database. So the actual size of each shard from
+raw table shuld be increased by at least 1.9M * 133 = 252.7M. Thus we ontained
```
undertome, line 68: obtained*

undertome:
What's the motivation for minimizing the size of full history as opposed to minimizing the size of individual shards?

src/ripple/nodestore/ShardSizeTuning.md
cdy20: Done.

src/ripple/nodestore/ShardSizeTuning.md
cdy20: Done.

src/ripple/nodestore/ShardSizeTuning.md
cdy20: Done.

src/ripple/nodestore/ShardSizeTuning.md
cdy20: Done.

src/ripple/nodestore/ShardSizeTuning.md
cdy20: Done.

src/ripple/nodestore/ShardSizeTuning.md
cdy20: Done.

src/ripple/nodestore/ShardSizeTuning.md
cdy20: Done.

cdy20:
> It is likely cheaper to just create all shards over a weekend or so than to actually implement that sampling.

I can acquire about 4 shards per day on my office PC with 32G memory. So acquiring full history is not for weekend, but for years.

cdy20:
> What's the motivation for minimizing the size of full history as opposed to minimizing the size of individual shards?

Answered in the text: Note that we minimize total storage size on the disk, but not the size of each
shard because data below shows that size of typical shard is not large than
10G, but sizes of modern disks, even SSD, start from 250G. So there is
no problem to fit a single shard to a disk, even small.

miguelportilla:
Here is the data for shards @16k and 32k ledgers. Dmitry has the data for 8k ledgers.

Shards with 16384 ledgers

Shard index                 2600
Last ledger                 42614784
nudb.dat size               2393157311
nudb.dat SHA-1 checksum     1b77bb8435b6cf2081e9140f0e28d70d9ead57a0
nudb.key size               368607232
nudb.key SHA-1 checksum     beccc6084e9e470c6345c02f3cb43b0c6bd6599f

Shard index                 2700
Last ledger                 44253184
nudb.dat size               2954326448
nudb.dat SHA-1 checksum     0767d92fe71ba1d285483deb0c9527b1ed2adacc
nudb.key size               442929152
nudb.key SHA-1 checksum     dff0a65dabcf56d66c1a4344d1d2f9cc0f4074d3

Shard index                 2800
Last ledger                 45891584
nudb.dat size               2532112792
nudb.dat SHA-1 checksum     f6f92dc6c4965247d9b21d30fa991544ee47e115
nudb.key size               372936704
nudb.key SHA-1 checksum     fcd8abc93ab5a16dcd127a8c7f8a34cb888b36d9

Shard index                 2900
Last ledger                 47529984
nudb.dat size               3827960048
nudb.dat SHA-1 checksum     fa6e6c16503c3e478379bd1453badf646b4d92f2
nudb.key size               576094208
nudb.key SHA-1 checksum     f2818338c3ddcec1fa7e4e83293eb92aa4f966c4

Shard index                 3000
Last ledger                 49168384
nudb.dat size               4486718052
nudb.dat SHA-1 checksum     3dabbb9e548de1f26a4f119494c7fe1d0ca9c7d8
nudb.key size               602615808
nudb.key SHA-1 checksum     ce8b945108e4cd24f03e3bae978da53707b3bf8c

Shard index                 3100
Last ledger                 50806784
nudb.dat size               3786057058
nudb.dat SHA-1 checksum     509936af5171f68021cfc70921dde32ee58ce42b
nudb.key size               507273216
nudb.key SHA-1 checksum     eefe93c7e3de27872629c1a649c5dafeb2caa422

Shard index                 3200
Last ledger                 52445184
nudb.dat size               4146749369  
nudb.dat SHA-1 checksum     c824d7cb680f7ce7a3b40d616221332f53abdfd3
nudb.key size               543186944
nudb.key SHA-1 checksum     5f87e51ac8ed75133a9039f1b056ae90d0801eb6

Shard index                 3300
Last ledger                 54083584
nudb.dat size               5186218493
nudb.dat SHA-1 checksum     05962f58207d51bc669c2e8f3a8b5105ef499714
nudb.key size               733659136
nudb.key SHA-1 checksum     0053d5d4cca738c1c42a8c192b3b6ed9d2a001d1

Shard index                 3400
Last ledger                 55721984
nudb.dat size               4182037553
nudb.dat SHA-1 checksum     baa3380b5410e2cd7d284d7234de74bbb09cd1f6
nudb.key size               583929856
nudb.key SHA-1 checksum     a78aac0c5bd50fc222826024e82cb048dd214561

Shard index                 3500
Last ledger                 57360384
nudb.dat size               5057026477
nudb.dat SHA-1 checksum     65ccc182618b202fc544997cb83e55cce75a1f09       
nudb.key size               676020224
nudb.key SHA-1 checksum     fcd9728289253ec6b21a854f76d8784df34280fc

Shard index                 3600
Last ledger                 58998784
nudb.dat size               4179657732
nudb.dat SHA-1 checksum     41228788fa4dc52df0b9e1272cbdd4ea48bb2889
nudb.key size               595046400       
nudb.key SHA-1 checksum     47bf7c370bc0e06a7aa5171ea5a6846ee67fa5b3

-------------------------------------------------------------------------------------
Shards with 32768 ledgers

LPS 32768 
Shard index                 1300
Last ledger                 42614784
nudb.dat size               3998000663
nudb.key size               590393344

Shard index                 1350
Last ledger                 44253184
nudb.dat size               5225160554
nudb.key size               754798592

Shard index                 1400
Last ledger                 45891584
nudb.dat size               4370632329
nudb.key size               616435712

Shard index                 1450
Last ledger                 47529984
nudb.dat size               7017311723 
nudb.key size               1003855872

Shard index                 1500
Last ledger                 49168384
nudb.dat size               7531276864
nudb.key size               978538496

Shard index                 1550
Last ledger                 50806784
nudb.dat size               6849170115
nudb.key size               877436928

Shard index                 1600
Last ledger                 52445184
nudb.dat size               7283597938
nudb.key size               914292736 

Shard index                 1650
Last ledger                 54083584 
nudb.dat size               8100305083
nudb.key size               1098620928

Shard index                 1700
Last ledger                 55721984
nudb.dat size               7581115823
nudb.key size               1024409600

Shard index                 1750
Last ledger                 57360384
nudb.dat size               8954887847 
nudb.key size               1139757056

Shard index                 1800
Last ledger                 58998784
nudb.dat size               7286299391
nudb.key size               981643264

src/ripple/nodestore/ShardSizeTuning.md
```diff
@@ -141,6 +141,32 @@ sizes of nudb.* which are permanent for deterministic shards.
...
+4.43 * 2 = 8.86 greather than 7.65 by 16 percents.
+
 ## Conclusion
 
-We recommend use shards with 16K ledgers.
+We showed that using shards with 8k ledgers leads to raising required disk
+size by 20% in comparison with using shards with 16k ledgers. In the same way,
+using shards with 16k ledgers leads to raising required disk space by 15%
+in comparison with using shards with 16k ledgers. So if we only take into
+consideration disk usage, then optimal is using shards with 32k ledgers.
```
MarkusTeufelberger, line 165: ...if that's your only option on the table. I bet that if you try 64k shards or 128k shards, there will be additional savings. ;-)

src/ripple/nodestore/ShardSizeTuning.md
```diff
@@ -141,6 +141,32 @@ sizes of nudb.* which are permanent for deterministic shards.
...
+But we also should take into consideration computational power of nodes.
+For example, my machine with 32G memory is best for acquire shards with 16k
```
MarkusTeufelberger, line 167: Really? A 32k shard fits 4 times into 32GiB of memory. What makes this "best" for half the size?

src/ripple/nodestore/ShardSizeTuning.md
```diff
@@ -141,6 +141,32 @@ sizes of nudb.* which are permanent for deterministic shards.
 We can see that in all tables ratio is greater then 1, so using shards with
 16 ledgers is preferred.
 
+## Compare 16K shards and 32K shards
+
+To claim that shards with 16K ledgers are the best choise, we also assembled
+shards with 32 ledgers per shard. There were 11 such a shards computed with
```
MarkusTeufelberger, line 147:
```suggestion
shards with 32k ledgers per shard. There were 11 such a shards computed with
```

src/ripple/nodestore/ShardSizeTuning.md
```diff
@@ -141,6 +141,32 @@ sizes of nudb.* which are permanent for deterministic shards.
...
+indexes from 1300 to 1800 with step 50. These are the shards which correspond
+to previously calculated shards with 16 ledgers.
```
MarkusTeufelberger, line 149:
```suggestion
to previously calculated shards with 16k ledgers.
```

MarkusTeufelberger:
Thanks @miguelportilla - did you also create the adjacent 16k shards, so you can accurately compare the sizes of the same ledger ranges in shard form instead of just assuming that there's a perfect/rough 1:1 split in size of shards at these times? I guess though that it is already good enough to show how high the overhead can/will become with smaller shard sizes.

miguelportilla:
@MarkusTeufelberger We acquired the adjacent shards to accurately compare 8k vs 16k shards. I still need to do the same to accurately compare 16k vs 32k shards.

miguelportilla:
Shards with 16384 ledgers

Shard index 2599
Last ledger 42598400
nudb.dat size 2408533645 
nudb.key size 372957184

Shard index 2699
Last ledger 44236800
nudb.dat size 2893066099 
nudb.key size 434114560 

Shard index 2799
Last ledger 45875200
nudb.dat size 2507272897
nudb.key size 375164928

Shard index 2899
Last ledger 47513600
nudb.dat size 3589450893
nudb.key size 561233920

Shard index 2999
Last ledger 49152000
nudb.dat size 4395718474 
nudb.key size 592498688 

Shard index 3099
Last ledger 50790400
nudb.dat size 3988022879 
nudb.key size 539291648 

Shard index 3199
Last ledger 52428800
nudb.dat size 4324385887
nudb.key size 559828992

Shard index 3299
Last ledger 54067200
nudb.dat size 6690419588 
nudb.key size 963944448

Shard index 3399
Last ledger 55705600
nudb.dat size 4165614944
nudb.key size 578134016

Shard index 3499
Last ledger 57344000
nudb.dat size 4864081453
nudb.key size 643645440

Shard index 3599
Last ledger 58982400
nudb.dat size 4060358687
nudb.key size 572018688

src/ripple/nodestore/ShardSizeTuning.md
cdy20: Added another argument.

src/ripple/nodestore/ShardSizeTuning.md
cdy20: There are also some internal structures in rippled which occupy the memory.

src/ripple/nodestore/ShardSizeTuning.md
cdy20: Done.

src/ripple/nodestore/ShardSizeTuning.md
cdy20: Done.

src/ripple/nodestore/ShardSizeTuning.md
MarkusTeufelberger: Yes? Still `rippled` uses nowhere near this amount of memory on my node and if it really keeps several copies of the same data in memory when dealing with or constructing shards, this seems like a serious issue.

cdy20: Ok, I will check 32k ledgers shards on my 32G laptop. Generally said, Ripple/XRP is serious issue itself.

src/ripple/nodestore/ShardSizeTuning.md
MarkusTeufelberger: Saving ~10% of storage space on a several TiB data set is not worth the effort according to this argument? I'm not totally convinced by that one to be honest.

src/ripple/nodestore/ShardSizeTuning.md
cdy20: Checked on 32k ledgers shards. First shard successfully acquired, and the program stopped downloading shards at this point, none further shards downloading. May be compound reason: lack of resources + software bug.

src/ripple/nodestore/ShardSizeTuning.md
cdy20: This is secondary reason, the main is problems with acquiring large shards.

src/ripple/nodestore/ShardSizeTuning.md
```diff
@@ -144,29 +144,59 @@ We can see that in all tables ratio is greater then 1, so using shards with
 ## Compare 16K shards and 32K shards
 
 To claim that shards with 16K ledgers are the best choise, we also assembled
```
seelabs, line 146: Spelling: choise -> choice

cdy20: Done.

src/ripple/nodestore/ShardSizeTuning.md
```diff
@@ -186,17 +186,27 @@ and nudb.key.
...
 in comparison with using shards with 16k ledgers. Note that increase ratio 12%
 is much less than 22% so using 32k-ledgers shards will bring us not so much
-economy in disk space. At the same time, we also may take into consideration
-computational power of nodes. For example, my machine with 32G memory is best
-for acquire shards with 16k ledgers, but acquiring shards with 32k ledgers
-is probably problematic for such a machine. If we decide to use 32k ledgers
-per shard, then we should consider 64G minimum memory capacity for a node,
-and economy of disk size will be not so much in comparison with 16k ledgers
-shards.
+economy in disk space.
```
miguelportilla, line 194: ```using shards with 16k ledgers leads to raising required disk space by 12% in comparison with using shards with 16k ledgers.```

I believe you meant

``````using shards with 16k ledgers leads to raising required disk space by 12% in comparison with using shards with 32k ledgers.```

**Review by miguelportilla — APPROVED** [review]
Left a comment for a minor correction but the document looks good. 👍

File: src/ripple/nodestore/ShardSizeTuning.md (hunk 13 above)
Comment (cdy20) [reply src/ripple/nodestore/ShardSizeTuning.md:194]: Done.

File: src/ripple/nodestore/ShardSizeTuning.md [hunk 14]
```diff
@@ -0,0 +1,212 @@
+# Shard Size Tuning
+
+The purpose of this document is to compare the sizes of shards containing
+varying amounts of ledgers.
+
+## Methodology
+
+One can see visually from a block explorer that typical mainnet ledger
```
undertome, line 8: Should say "that **a** typical mainnet ledger"

src/ripple/nodestore/ShardSizeTuning.md
```diff
@@ -0,0 +1,212 @@
...
+consists of about 30 offer transactions issued by about 8 different accounts,
+and several transactions of other types. To simulate this situation and
+similar situations we had constructed deterministic shards of differenet sizes,
```
undertome, line 11: "we **have** constructed"

src/ripple/nodestore/ShardSizeTuning.md
```diff
@@ -0,0 +1,212 @@
...
+with different number of offer transactions per ledger and different number
```
undertome, line 12: Consider: "with varying amounts of offers per ledger and varying amounts of accounts issuing..."

src/ripple/nodestore/ShardSizeTuning.md
```diff
@@ -0,0 +1,212 @@
...
+of accounts issuing these offers.
+
+Number of ledgers per shard starts from 256 to 16K with doubling the size on
```
undertome, line 15: Consider: "In the following results table, the number of ledgers per shard ranges from 256 to 16K with the size doubling the size at each step"

src/ripple/nodestore/ShardSizeTuning.md
```diff
@@ -0,0 +1,212 @@
...
+each step. We considered the following numbers of offers per ledger:
+0, 1, 5, 10 and 30. Also we considered case of 1 and 8 accounts issuing offers.
+For each constructed deterministic shard we counted its size. Finally we
+compared doubled size of the shard with N ledgers and the size of a shard with
+2*N ledgers where othere parameters such as number of offers and accounts are
+the same. This comparison is sufficient to determine which number or ledgers
```
undertome, line 21: "which number **of** ledgers"

src/ripple/nodestore/ShardSizeTuning.md
```diff
@@ -0,0 +1,212 @@
...
+per shard leads to less storage size on the disk. 
+
+Note that we minimize total storage size on the disk, but not the size of each
+shard because data below shows that size of typical shard is not large than
```
undertome, line 25: "shows that **the** size of a typical shard is not **larger**"

src/ripple/nodestore/ShardSizeTuning.md
```diff
@@ -0,0 +1,212 @@
...
+10G, but sizes of modern disks, even SSD, start from 250G. So there is
```
undertome, line 26: Consider: "SSDs"

src/ripple/nodestore/ShardSizeTuning.md
```diff
@@ -0,0 +1,212 @@
...
+no problem to fit a single shard to a disk, even small.
+
+
+## Raw results table
+
+All sizes of constructed shards are shown in the following table.
+Rows corresponds to shard sizes (S) counted in ledgers, columns corresponds
+to numbers of offers (O) per ledger. In each cell there are two numbers:
+first number corresponds to the case of 1 account issuing offers, the second
+number corrspoonds to 8 accounts. Each number is a size of the shard with
```
undertome, line 36: corresponds*

src/ripple/nodestore/ShardSizeTuning.md
```diff
@@ -0,0 +1,212 @@
...
+|8|9%|7%|7%|8%|6%|7%|6%|
+
+The common conclusion in this model is that if one doubled the number of 
+the ledgers in a shard then the total disk space utilized will raise by 5-9%.
+
+## Adding accounts into consideration
+
+Previous model does not take into account that there are large number of
+XRP accounts in the mainnet, and each shard should contain information
+about each of these accounts. As of January 2020, there were about 1.9 millions
```
undertome, line 71: million*

src/ripple/nodestore/ShardSizeTuning.md
```diff
@@ -0,0 +1,212 @@
...
+|3200|4.15|2.54 + 2.43 = 4.97|1.20|
+|3300|5.19|3.23 + 2.80 = 6.03|1.16|
+|3400|4.18|2.53 + 2.51 = 5.04|1.21|
+|3500|5.06|2.90 + 3.04 = 5.94|1.17|
+|3600|4.18|2.56 + 2.51 = 5.07|1.21|
+|Average|3.89|2.35 + 2.35 = 4.70|1.21|
+
+Note that shard on the disk consists of 4 files each of which can be large too.
+These files are nudb.dat, nudb.key, ledger.db, transaction.db. Next table is
+similar to previout with the following exception: each number is total size
```
undertome, line 121: previous*

src/ripple/nodestore/ShardSizeTuning.md
```diff
@@ -0,0 +1,212 @@
...
+economy in disk space.
+
+At the same time, size is one thing to compare but there are other aspects.
+Smaller shards have an advantage that they take less time to acquire and
+finalize. They also make for smaller archived shards which take less time to
+download and import. Having more/smaller shards might also lead to better
+database concurrency/performance.
+
+It is hard to maintain both size and time parameters by a single optimization
+formulae because different chooses for weights of size and time may lead to
```
undertome, line 203: choices*

undertome:
Added a few minor comments to improve readability and spelling. Overall looks good!

src/ripple/nodestore/ShardSizeTuning.md
cdy20: Done.

src/ripple/nodestore/ShardSizeTuning.md
cdy20: Done.

src/ripple/nodestore/ShardSizeTuning.md
cdy20: Done.

src/ripple/nodestore/ShardSizeTuning.md
cdy20: Done.

src/ripple/nodestore/ShardSizeTuning.md
cdy20: Done.

src/ripple/nodestore/ShardSizeTuning.md
cdy20: Done.

src/ripple/nodestore/ShardSizeTuning.md
cdy20: Done.

src/ripple/nodestore/ShardSizeTuning.md
cdy20: Done.

src/ripple/nodestore/ShardSizeTuning.md
cdy20: Done.

src/ripple/nodestore/ShardSizeTuning.md
cdy20: Done.

src/ripple/nodestore/ShardSizeTuning.md
cdy20: Done.

PR #3675, opened 2020-11-23
Rollup (3603, 3646 and 3672)

nbougalis:
This PR incorporates and supersedes three separate PRs. If merged, it will close #3603, close #3646 and close #3672.


seelabs (approved):
:+1: My approval from #3646 can carry over to this PR.

.travis.yml
```diff
@@ -323,7 +311,7 @@ matrix:
         - export TEST_EXTRA_ARGS="--unittest-ipv6"
     - <<: *macos
       osx_image: xcode11.2
-      name: xcode11, debug
+      name: xcode11.2, debug
```
cjcobb23, line 314: There are two builds named "xcode11.2, debug" now (here and on line 266), and I think they are identical. We only need one of these right?

HowardHinnant (approved):

gregtatcam (commented):
#3672 looks fine. I still have #3646 on my plate though.

.travis.yml
nbougalis: I believe on runs IPv6-specific unit tests. I don't know that really gets us much and we could remove it.

gregtatcam (approved):
:+1: for #3646, #3672

scottschurr (commented):
https://github.com/ripple/rippled/pull/3603 still looks good with the exception of concerns about the API change.  Here's where the API change was first noted in 3606: https://github.com/ripple/rippled/pull/3603#discussion_r487348481

Here's the most recent comment regarding the API change, which includes an (untested) patch that puts the API change under API versioning: https://github.com/ripple/rippled/pull/3603#issuecomment-733140844

I'm okay if:
- We decide that the API change is non-breaking, or
- Decide the API change _is_ breaking but that breakage is justified and we'll just document it, or
- We'll modify this pull request to back out the API change, or
- We'll modify this pull request to version the API change, or
- We'll have a follow-on pull request to version the API change before the next release candidate.

But I'm uncomfortable with just moving ahead and not stating explicitly how we intend to address the API change.  Please, let's have an explicit call on how the API change will be handled.

Thanks.

src/ripple/rpc/handlers/Peers.cpp
```diff
@@ -32,17 +32,31 @@ doPeers(RPC::JsonContext& context)
...
 
+    jvResult[jss::peers] = context.app.overlay().json();
+
+    // Legacy support
+    if (context.apiVersion == 1)
     {
-        jvResult[jss::peers] = context.app.overlay().json();
+        for (auto& p : jvResult[jss::peers])
+        {
+            if (auto s = p[jss::track].asString(); !s.empty())
```
scottschurr, line 42: This code has a subtle bug courtesy of our JSON library.  By _examining_ `jss::track` in this way you have injected `jss::track` into object `s` with a value of `null`.  You can avoid the bug with brute force by doing something like this:
```
        for (auto& p : jvResult[jss::peers])
        {
            if (p.isMember(jss::track)) // Important test
            {
                if (auto s = p[jss::track].asString(); !s.empty())
                {
                    if (s == "diverged")
                        p["sanity"] = "insane";
                    if (s == "unknown")
                        p["sanity"] = "unknown";
                }
            }
        }
```
Or perhaps you have a more clever suggestion.

scottschurr (changes requested):
I think there's a bug in the API patch.  Take a look and see if you agree.  I think this is really close...

src/ripple/rpc/handlers/Peers.cpp
nbougalis: Ooof, you're right. I totally missed that, thanks. I'll adopt your approach.

nbougalis: I've fixed this thusly: 

https://github.com/ripple/rippled/blob/c52a3db8db49214385ab9f3f1d2944a2412e4222/src/ripple/rpc/handlers/Peers.cpp#L38-L52

scottschurr (approved):
:+1:

PR #3676, opened 2020-11-24
Remove DinD Service from container build template

northrup:
Refactor how our Docker calls are invoked on the GitLab runners using `services` methods and moving away from Docker-in-Docker.

### DIFF
--- a/Builds/containers/gitlab-ci/pkgbuild.yml
+++ b/Builds/containers/gitlab-ci/pkgbuild.yml
@@ -45,11 +45,7 @@ stages:
   variables:
     docker_driver: overlay2
   image:
-    name: docker:latest
-  services:
-    # workaround for TLS issues - consider going back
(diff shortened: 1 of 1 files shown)


MarkusTeufelberger:
If you want to get rid of `docker build` alltogether, I can recommend taking a look at `kaniko` by the way.

manojsdoshi (approved):
LGTM

nbougalis:
Kaniko looks interesting, @MarkusTeufelberger. Thanks for the pointer. This isn't really my field of expertise, so I'll let others decide whether this should be done.

northrup:
@MarkusTeufelberger that is a good point, and we have `kaniko` elsewhere, for the image building that we do that's a good idea and I'll look to refactor those, we also use docker here to spawn build jobs within a CI run.

PR #3677, opened 2020-11-24
Rework asynchronous fetches from SHAMap::getMissingNodes to Database::threadEntry

JoelKatz:

This change reworks the way the node database handles asynchronous fetches, eliminating the positive and negative caches and allowing the reads to be processed as they complete rather than waiting for the last one to finish before processing the first one.

### Context of Change

The read generation logic was implemented some time ago and has always been a bit kludgy. It had important edge cases it mishandled, such as the case where an async read got pushed out of memory before it was retrieved. This removes that logic and replaces it with a callback scheme.

## Before / After

Prior to this change, asynchronous fetches were batched into generations with the results being placed into the database layer's positive and negative caches. The get missing nodes logic then retrieved the results from the caches. After this change, asynchronous fetches include a callback function that is called as soon as the operation completes. There is no need for the caches. The get missing nodes logic processes the reads as they complete.

## Future Tasks

Pushing the callback logic all the way into the NuDB layer and eliminating the Database layer's read threads would be a huge win. Every operation system we support has some form of async I/O and this would allow a larger I/O queue depth, improving performance on modern SSDs and also reducing context switches.

## Review notes

Some of the types are annoyingly verbose. It may be considered an improvement to replace them with `using` declarations or other trickery (like `::mapped_type` or something like that). Improvements welcomed.


seelabs:
We started formatting the code with clang-format. Here's a patch to cherry pick that's be formatted: https://github.com/seelabs/rippled/commit/796a45f16caf6dbbd44fed62affac091bad06dc7

src/ripple/shamap/impl/SHAMap.cpp
```diff
@@ -147,33 +147,40 @@ SHAMap::findKey(uint256 const& id) const
...
+            makeSlice(object->getData()), hash);
+        if (node)
+            canonicalize(hash, node);
+    }
+    catch (std::exception const&)
+    {
+        JLOG(journal_.warn()) << "Invalid DB node " << hash;
+        return std::shared_ptr<SHAMapTreeNode>();
+    }
     return node;
```
nbougalis, line 184: Micro-nit. Recommend:

```
    try
    {
        auto node = SHAMapAbstractNode::makeFromPrefix(
            makeSlice(object->getData()), hash);
        if (node)
            canonicalize(hash, node);
        return node;
    }
    catch (std::exception const&)
    {
        JLOG(journal_.warn()) << "Invalid DB node " << hash;
        return {};
    }
```

src/ripple/shamap/impl/SHAMapSync.cpp
```diff
@@ -327,7 +323,7 @@ SHAMap::getMissingNodes(int max, SHAMapSyncFilter* filter)
     MissingNodes mn(
         max,
         filter,
-        f_.db().getDesiredAsyncReadCount(ledgerSeq_),
+        4096,   // number of async reads per pass
```
nbougalis, line 326: Should this number be tunable? Is it possible that 4096 will be too much (or too little)?

src/ripple/nodestore/impl/DatabaseNodeImp.cpp
```diff
@@ -77,8 +61,8 @@ DatabaseNodeImp::fetchNodeObject(
     FetchReport& fetchReport)
 {
     // See if the node object exists in the cache
```
nbougalis, line 63: This comment is no longer applicable and should be removed.

nbougalis (commented):
Looks good. Left a few comments, mostly stylistic. Happy to do do those and squash this down into one commit for you.

src/ripple/nodestore/impl/Database.cpp
```diff
@@ -312,19 +290,20 @@ Database::threadEntry()
...
             readLastHash_ = lastHash;
         }
 
-        // Perform the read
-        fetchNodeObject(lastHash, lastSeq, FetchType::async);
+        for (auto const& req : entry)
+        {
+            auto obj = fetchNodeObject(lastHash, req.first, FetchType::async);
+            req.second (obj);
+        }
```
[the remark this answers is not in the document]
JoelKatz: I think that won't work since you can't bind a non-const reference to a temporary.

src/ripple/shamap/impl/SHAMapSync.cpp
JoelKatz: There's really no basis on which to tune it. It controls how quickly getMissingNodes switches between looking for more async reads to do and waiting for async reads to complete. There's no reason to think more is better, you just need a high enough number that the cost of waiting for reads to complete is lost in the noise. You just want to stay in the "discovering reads we need to do" phase long enough that a couple of the reads you've pended have completed to minimize the time you're doing nothing.

src/ripple/nodestore/impl/Database.cpp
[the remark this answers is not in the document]
nbougalis: Yes, but I think that the callback can (and, given how it's used _should_) be changed to take `std::shared_ptr<NodeObject> const&` or, perhaps better still, simply `NodeObject*`. Sorry I wasn't more clear.

src/ripple/shamap/impl/SHAMapSync.cpp
nbougalis: :+1:

src/ripple/nodestore/Database.h
```diff
@@ -135,20 +135,23 @@ class Database : public Stoppable
...
+        @return true if the operation completed, false if pending
     */
     virtual bool
     asyncFetch(
         uint256 const& hash,
         std::uint32_t ledgerSeq,
-        std::shared_ptr<NodeObject>& nodeObject) = 0;
+        std::shared_ptr<NodeObject>& nodeObject,
+        std::function<void(
+            std::shared_ptr<NodeObject>&)>&& callback) = 0;
```
seelabs, line 154: I'd like to refactor this interface. `nodeObject` is never used, we always get the `nodeObject` through the callback. All implementations always post a read, _except_ `shards`, which first checks that the ledger is part  of its collection. If it's not, it doesn't post a read, and will not retrieve the object. Given that, I'd like to get rid of the `nodeObject` parameter, and make the return value `true` if pending (i.e. "success") and `false` if the object will never be found (failure).

Here's a patch for that refactor: https://github.com/seelabs/rippled/commit/f272b74741a8bc9d0e51f279fa978f9ed4a8ecf9 

Two notes: 
* this patch is based on top of a clang-reformat patch. You need the reformat-patch before applying this patch).
* this patch contains a fix for `DatabaseShardImp::asyncFetch` that I describe in another comment

src/ripple/nodestore/impl/DatabaseShardImp.cpp
```diff
@@ -1027,16 +1028,13 @@ DatabaseShardImp::asyncFetch(
 
         auto const it{shards_.find(acquireIndex_)};
         if (it == shards_.end())
-            return false;
+            return true;
         shard = it->second;
     }
```
seelabs, line 1033: This function doesn't look correct. We check if the `acquireIndex_` is in shards and we don't fetch if it's not? That doesn't make sense to me. I spoke with @miguelportilla and he agrees that this looks buggy. He suggested the following fix:
```
bool
DatabaseShardImp::asyncFetch(
    uint256 const& hash,
    std::uint32_t ledgerSeq,
    std::function<void(std::shared_ptr<NodeObject>&)>&& callback)
{
    auto const shardIndex{seqToShardIndex(ledgerSeq)};
    {
        std::lock_guard lock(mutex_);
        assert(init_);
        if (shards_.find(shardIndex) == shards_.end())
            return false;
    }

    // Otherwise post a read
    return Database::asyncFetch(hash, ledgerSeq, std::move(callback));
}
```
Notes:
* this is part of the patch here: https://github.com/seelabs/rippled/commit/f272b74741a8bc9d0e51f279fa978f9ed4a8ecf9
* this uses the new return value semantics (true means pending, false mean it's never coming).

seelabs (commented):
Leaving a partial review for refactoring the `asyncFetch` interface. This patch has the suggestions: https://github.com/seelabs/rippled/commit/f272b74741a8bc9d0e51f279fa978f9ed4a8ecf9

Edit: Here's a second patch that builds on the first: https://github.com/ripple/rippled/commit/6147739d1a71b522597b632988fb055170c707dc

src/ripple/nodestore/impl/DatabaseShardImp.cpp
seelabs: Actually, looking closer, `DatabaseShardImp::fetchNodeObject(` already does this check. This means we can get rid of all the overides. I'll make a patch to do so.

src/ripple/nodestore/Database.h
seelabs: It looks like `DatabaseShardImp::asyncFetch` can be reomved completely. In that case, this function can now return `void`. Here's a second patch that builds on the first: https://github.com/ripple/rippled/commit/6147739d1a71b522597b632988fb055170c707dc

src/ripple/nodestore/impl/DatabaseShardImp.cpp
seelabs: Here's a second patch that builds on the first: https://github.com/ripple/rippled/commit/6147739d1a71b522597b632988fb055170c707dc @miguelportilla Please dummy check that it is OK to remove `DatabaseShardImp::asyncFetch` and use `fetchNodeObject` to do this check for us.

src/ripple/nodestore/impl/Database.cpp
```diff
@@ -312,19 +290,20 @@ Database::threadEntry()
...
+            entry = std::move(it->second);
             read_.erase(it);
             readLastHash_ = lastHash;
         }
 
-        // Perform the read
-        fetchNodeObject(lastHash, lastSeq, FetchType::async);
+        for (auto const& req : entry)
+        {
+            auto obj = fetchNodeObject(lastHash, req.first, FetchType::async);
```
seelabs, line 304: How often do we expect to have multiple callbacks in the `entry` collection with the same ledgerSeq? When that happens we make redundant calls to `fetchNodeObject` that we don't need.

Two suggestions to address this:
* Sort the `entry` vector by the first element in the pair, and keep track of the last ledgerIndex used in `fetchNodeObject` and only call that function when `req.first` is different from the last call
* Instead of entry being a vector of pairs, it could be a `unordered_map<std::uint32_t,std::vector<std::function<std::shared_ptr<NodeObject>&)>>>>`

Of course, if we think this is a rare case it might not be worth handling it.

src/ripple/shamap/SHAMap.h
```diff
@@ -478,6 +491,12 @@ class SHAMap
     gmn_ProcessNodes(MissingNodes&, MissingNodes::StackEntry& node);
     void
     gmn_ProcessDeferredReads(MissingNodes&);
+
+    // fetch from DB helper function
+    std::shared_ptr<SHAMapAbstractNode>
+    finishFetch(
+        SHAMapHash const& hash,
+        std::shared_ptr<NodeObject>& object) const;
```
seelabs, line 499: `object` could be `const&`

src/ripple/shamap/impl/SHAMap.cpp
```diff
@@ -377,19 +385,29 @@ SHAMap::descendAsync(
 
         if (!ptr && backed_)
         {
-            std::shared_ptr<NodeObject> obj;
-            if (!f_.db().asyncFetch(hash.as_uint256(), ledgerSeq_, obj))
+            std::shared_ptr<NodeObject> object;
+            if (f_.db().asyncFetch(hash.as_uint256(), ledgerSeq_, object,
```
seelabs, line 389: I'll note this is the only call to `asyncFetch`, which means all the calls use the same callback. We could consider removing the callback parameters and hard coding the callback - this makes the code easier to understand.

On the other hand, keeping the callbacks makes for a better interface to `Database` and keeps the components separate.

I'm in favour of keeping things as they are, but wanted to point this out in case other people had strong feelings of wanting to change this.

src/ripple/shamap/impl/SHAMapSync.cpp
```diff
@@ -202,25 +202,35 @@ SHAMap::gmn_ProcessNodes(MissingNodes& mn, MissingNodes::StackEntry& se)
...
-
-            if (!d)
+            auto d = descendAsync(node, branch, mn.filter_, pending,
+            [node, nodeID, branch, &mn](
+                std::shared_ptr<SHAMapAbstractNode> found,
+                SHAMapHash const&)
             {
-                fullBelow = false;  // for now, not known full below
+                // a read completed asynchronously
+                std::unique_lock<std::mutex> lock{mn.deferLock_};
```
seelabs, line 212: C++-17 can infer the template parameter. So this can be:
```
std::unique_lock lock{mn.deferLock_};
```

src/ripple/shamap/impl/SHAMapSync.cpp
```diff
@@ -259,30 +269,28 @@ SHAMap::gmn_ProcessNodes(MissingNodes& mn, MissingNodes::StackEntry& se)
...
     // Process all deferred reads
-    int hits = 0;
-    for (auto const& deferredNode : mn.deferredReads_)
+    int complete = 0;
+    while (complete != mn.deferred_)
     {
+        std::tuple<SHAMapInnerNode*, SHAMapNodeID, int,
+            std::shared_ptr<SHAMapAbstractNode>> deferredNode;
+        {
+            std::unique_lock<std::mutex> lock{mn.deferLock_};
```
seelabs, line 279: C++-17 can infer the template parameter. This can be just:
```
std::unique_lock lock{mn.deferLock_};
```

seelabs (commented):
*love* these changes. It simplifies existing code, uses less memory, and gives a performance boost. What's not to love?

Left a few more comments for you to consider.

src/ripple/nodestore/impl/Database.cpp
```diff
@@ -296,7 +301,9 @@ Database::threadEntry()
     while (true)
     {
         uint256 lastHash;
-        std::uint32_t lastSeq;
+        std::vector< std::pair <std::uint32_t,
+            std::function<void(std::shared_ptr<NodeObject>&)>>> entry;
+
```
miguelportilla, line 278: I believe we can eliminate `lastHash` and move the `entry` declaration outside the loop.
eg.
```
    std::vector<std::pair<
        std::uint32_t,
        std::function<void(std::shared_ptr<NodeObject>&)>>>
        entry;
    while (true)
    {
        {
            std::unique_lock<std::mutex> lock(readLock_);
            while (!readShut_ && read_.empty())
            {
                // All work is done
                readCondVar_.wait(lock);
            }
            if (readShut_)
                break;

            // Read in key order to make the back end more efficient
            auto it = read_.lower_bound(readLastHash_);
            if (it == read_.end())
            {
                // start over from the beginning
                it = read_.begin();
            }
            readLastHash_ = it->first;
            entry = std::move(it->second);
            read_.erase(it);
        }

        for (auto const& req : entry)
        {
            auto obj =
                fetchNodeObject(readLastHash_, req.first, FetchType::async);
```

src/ripple/nodestore/impl/Database.cpp
miguelportilla: @seelabs It is redundant only when the ledger sequence and the node object hash are the same. My guess is that's unlikely but it should be examined closer.

@JoelKatz You are using the same node object hash for every fetch. You may need to make `entry` the same type as `read_` to retain the hash.

src/ripple/nodestore/impl/DatabaseShardImp.cpp
miguelportilla: The patches look good to me. The drawback is that we can now post a read knowing that the ledger sequence does not correspond to any shard. But it won't make it to IO as `DatabaseShardImp::fetchNodeObject` will return early (as you mentioned). This cleanup's benefit far outweighs that.

miguelportilla (changes requested):

src/ripple/nodestore/impl/Database.cpp
JoelKatz: I think this is sufficiently unlikely to happen and the consequences of it happening (an extra read) so mild that I left the code as simple as possible.

JoelKatz: Suppose the same node is in two different ledgers, each in a different shard. We may be searching for both nodes simultaneously. It is critical that we neither tell one query that the node is not in the shard if it is and also critical that we not tell one query that the node is in the shard if it isn't. Otherwise, we may build incomplete shards and fail to retrieve from complete ones!

src/ripple/nodestore/impl/Database.cpp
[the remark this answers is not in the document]
JoelKatz: I made this change, but it forces an extra copy. You can move both to and from a parameter passed by value. When you pass a parameter by reference, the recipient cannot move from it and must make a deep copy.

miguelportilla:
`src/ripple/nodestore/impl/Tuning.h` can be removed from the project.

nbougalis:
I squashed and rebased this set of commits (02a3269..35b86f0) against the latest beta, which made changes to the `SHAMap` code. You can find the result here: [a66e39be134788dee9d216b09f12475c3e9f789f](https://github.com/nbougalis/rippled/commit/a66e39be134788dee9d216b09f12475c3e9f789f) if you want to use it.

PR #3686, opened 2020-12-04
Fix double counting unit test failures:

seelabs:

Unit tests are counting test failures, process crashes, and process exit code
failures in the count. Since a failing tests causes the process exit code to
return failure, we get extra counts. This patch removes process exit code
failures from the count.

### Type of Change


seelabs:
Fixes https://github.com/ripple/rippled/issues/3685

thejohnfreeman (approved):

ximinez (approved):

PR #3691, opened 2020-12-07
Correct handling of corner-cases involving custom iterators:

nbougalis:

This commit, if merged, corrects the handling of corner-cases involving custom iterators:

- Under some conditions, comparing `ReadViewFwdRange::iterators` for equality could result in derefencing an empty `std::unique_ptr` which will result in a crash.
- The `OpenView::txs_iter_impl::equal` function could throw an `std::bad_cast` exception which was unexpected.

### Type of Change

### DIFF
--- a/src/ripple/app/ledger/Ledger.cpp
+++ b/src/ripple/app/ledger/Ledger.cpp
@@ -100,8 +100,9 @@ public:
     bool
     equal(base_type const& impl) const override
     {
-        auto const& other = dynamic_cast<sles_iter_impl const&>(impl);
-        return iter_ == other.iter_;
+        if (auto const p = dynamic_cast<sles_iter_impl const*>(&impl))
+            return iter_ == p->iter_;
+        return false;
     }
 
     void
@@ -148,8 +149,9 @@ public:
     bool
     equal(base_type const& impl) const override
     {
-        auto const& other = dynamic_cast<txs_iter_impl const&>(impl);
-        return iter_ == other.iter_;
+        if (auto const p = dynamic_cast<txs_iter_impl const*>(&impl))
+            return iter_ == p->iter_;
+        return false;
     }
 
     void
--- a/src/ripple/ledger/impl/OpenView.cpp
+++ b/src/ripple/ledger/impl/OpenView.cpp
@@ -45,8 +45,9 @@ public:
     bool
     equal(base_type const& impl) const override
     {
-        auto const& other = dynamic_cast<txs_iter_impl const&>(impl);
-        return iter_ == other.iter_;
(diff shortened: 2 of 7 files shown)


src/ripple/ledger/impl/OpenView.cpp
```diff
@@ -45,8 +45,9 @@ class OpenView::txs_iter_impl : public txs_type::iter_base
     bool
     equal(base_type const& impl) const override
     {
-        auto const& other = dynamic_cast<txs_iter_impl const&>(impl);
-        return iter_ == other.iter_;
+        if (auto p = dynamic_cast<txs_iter_impl const*>(&impl))
```
scottschurr, line 48: Consider `auto const p`.  Not that it makes any difference right now...

scottschurr (approved):
:+1: Yup, these changes look right to me...

src/ripple/ledger/impl/OpenView.cpp
nbougalis: I fixed this and did the same fix in `Ledger::txs_iter_impl`, which I'd previously missed.

scottschurr (approved):
:+1:  Yup, this looks good to me.

cjcobb23 (approved):
LGTM

src/ripple/app/ledger/Ledger.cpp
```diff
@@ -100,8 +100,9 @@ class Ledger::sles_iter_impl : public sles_type::iter_base
     bool
     equal(base_type const& impl) const override
     {
-        auto const& other = dynamic_cast<sles_iter_impl const&>(impl);
-        return iter_ == other.iter_;
+        if (auto const p = dynamic_cast<sles_iter_impl const*>(&impl))
+            return iter_ == p->iter_;
+        return false;
```
HowardHinnant, line 105: Do we want to check for the case `p == nullptr`?  If we consider this case a logic error then a crash is appropriate (which is what it does with this code).

nbougalis: I'm confused? The `if` should never be entered if `p == nullptr` or am I misunderstanding?

```
if (auto const p = dynamic_cast<sles_iter_impl const*>(&impl))
    return iter_ == p->iter_;
return false;
```

HowardHinnant: My bad.  I completely misread it.  Thanks for the correction.

HowardHinnant (approved):

nbougalis:
Merged as deea16e14fae35166ee1b9e39d333672e7b30761

PR #3695, opened 2020-12-09
Stoppable housecleaning

thejohnfreeman:
Starting off with removing duplicate overrides and shrinking public interfaces. These are changes that I'm not expecting to be controversial, that I made to ease investigation of the `Stoppable` hierarchy.

These changes pass all of the (non-manual) tests on my machine, and I've replicated Scott's start-sync-stop-loop test to repeatedly test shutdowns in the wild, and it has successfully shutdown 108 times (after taking 9 minutes to sync, on average, and less than 11 seconds to shutdown).


src/ripple/basics/impl/PerfLogImp.cpp
```diff
@@ -477,14 +477,12 @@ PerfLogImp::onStop()
...
         thread_.join();
     }
-    if (areChildrenStopped())
-        stopped();
 }
 
 void
 PerfLogImp::onChildrenStopped()
 {
-    onStop();
```
thejohnfreeman, line 487: These changes should exhibit exactly the same behavior, but I think are easier to read. When `onChildrenStopped()` calls `onStop()`, `thread_.joinable()` should return `false` because it was joined in the earlier call to `onStop` (which is called before `onChildrenStopped` in the shutdown sequence). Thus, the first conditional branch is skipped. The second conditional tests `areChildrenStopped()` which returns the value of `m_childrenStopped`, which starts as `false`, and is only set to `true` once, immediately before calling `onChildrenStopped`, which occurs some time after the call to `onStop`. So in `onStop`, `areChildrenStopped()` should always return `false`, and in `onChildrenStopped`, `areChildrenStopped()` should always return `true`.

src/ripple/core/Stoppable.h
```diff
@@ -360,17 +360,9 @@ class RootStoppable : public Stoppable
     bool
     isStopping() const;
 
-    /** Prepare all contained Stoppable objects.
```
thejohnfreeman, line 363: `prepare` is only ever called right before `start`. There's no need for it to be a separate function, and I think the separation is confusing additional complexity.

src/ripple/core/impl/Stoppable.cpp
```diff
@@ -177,22 +177,16 @@ RootStoppable::isStopping() const
...
     // Courtesy call to prepare.
-    if (m_prepared.exchange(true) == false)
-        prepareRecursive();
+    if (m_prepared.exchange(true))
+        return;
 
-    if (m_started.exchange(true) == false)
-        startRecursive();
+    prepareRecursive();
+    m_started.exchange(true);
```
thejohnfreeman, line 188: This doesn't need to be behind a second conditional because it is guarded by the one above. In the old code, one could imagine a scenario where one thread calls `start()`, which enters `prepareRecursive()`, and while it's in there, another thread calls `start()`, which skips the call to `prepareRecursive()` and dives into `startRecursive()` before `prepareRecursive()` has finished in the other thread, potentially violating the assumptions of `Stoppable`s on the call sequence of their `onPrepare` and `onStart` callbacks. That doesn't happen in the code as it was written, but this change makes it impossible.

src/ripple/core/impl/Stoppable.cpp
```diff
@@ -201,13 +195,6 @@ RootStoppable::stop(beast::Journal j)
     // Must have a prior call to start()
     assert(m_started);
 
-    if (stopAsync(j))
-        stopRecursive(j);
-}
```
thejohnfreeman, line 206: `stopAsync` returns true on the first time it is called, so this is just a guard to call `stopRecursive` the first time as well. I think it's clearer to just fold `stopAsync` into this method.

HowardHinnant (approved):
Very clean.  Thanks for the small modular commits.

src/ripple/core/Stoppable.h
```diff
@@ -411,17 +403,6 @@ class RootStoppable : public Stoppable
...
-
-        Returns true on the first call to this method, false otherwise.
-
-        Thread safety:
-            Safe to call from any thread at any time.
-    */
-    bool
-    stopAsync(beast::Journal j);
-
     std::atomic<bool> m_prepared{false};
```
nbougalis, line 406: Random comment: _once_ we migrate to C++20, these should be changed to `[std::atomic_flag](https://en.cppreference.com/w/cpp/atomic/atomic_flag)`. Can you add a `TODO` comment tagged with C++20 to that effect please?

nbougalis (commented):
Looks good :+1:

scottschurr (approved):
:+1: Nicely organized pull request.  It was easy to follow the motivation for the changes.  Nice simplifications.

I left a couple of suggestions for you to consider but they are optional.

Just for extra confidence I ran the start-sync-stop test on this branch over the weekend.  It successfully ran over 750 iterations with no crashes.  That looks good.

src/ripple/core/impl/Stoppable.cpp
```diff
@@ -177,22 +177,16 @@ RootStoppable::isStopping() const
...
-RootStoppable::prepare()
-{
-    if (m_prepared.exchange(true) == false)
-        prepareRecursive();
-}
-
 void
 RootStoppable::start()
 {
     // Courtesy call to prepare.
```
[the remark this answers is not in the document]
thejohnfreeman: I'm just removing this comment entirely and fixing the comment on the declaration.

scottschurr:
@thejohnfreeman, is it your intention that this pull request is ready for review?  I thought it was ready.  However a commit that came in after I finished my review (this commit: https://github.com/ripple/rippled/pull/3695/commits/9f688b4b8da0ca9ccf3713de57629fcc31a38d5c) looks like it requires careful examination.   And, as far as I can tell, that change was not requested by any of the reviewers.

Please let me know when the pull request is stable and ready for review.

thejohnfreeman:
@scottschurr It's stable now. For 9f688b4, most of the changes are whitespace, so hiding that will make it easier to review. The only significant change is releasing the lock before calling `std::condition_variable::notify_one()`, but I made that change based on the second paragraph of the notes [here](https://en.cppreference.com/w/cpp/thread/condition_variable/notify_one):

> The notifying thread does not need to hold the lock on the same mutex as the one held by the waiting thread(s); in fact doing so is a pessimization, since the notified thread would immediately block again, waiting for the notifying thread to release the lock.

scottschurr (approved):
@thejohnfreeman, sorry for getting excited about https://github.com/ripple/rippled/commit/9f688b4b8da0ca9ccf3713de57629fcc31a38d5c  When I saw the diff on Github it looked like there were really significant changes.  I panicked.  But you're right, the changes were mostly whitespace.  Sorry for the noise.

At any rate, this all looks good to me.  :+1:

src/ripple/app/misc/SHAMapStore.h
```diff
@@ -21,23 +21,23 @@
...
 /**
  * class to create database, launch online delete thread, and
  * related SQLite database
  */
-class SHAMapStore : public Stoppable
+class SHAMapStore
 {
 public:
-    SHAMapStore(Stoppable& parent) : Stoppable("SHAMapStore", parent)
+    virtual ~SHAMapStore()
```
seelabs, line 40: Could be `= default`

seelabs (approved):
:+1: Very nice cleanups!

thejohnfreeman:
Forgive me, I found one more duplicate override to squash. I used the opportunity to take @seelabs's suggestion. Hopefully it's just a 30-second review of the [last two commits](https://github.com/ripple/rippled/pull/3695/files/ddbfa4cd5dcc196de547149e3afe6ea1c6ae252d..ccb4ab0665f09a60096065a985b491f183c74efe).

scottschurr (approved):
:+1: LGTM.

seelabs (approved):
Still :+1: from me.

pwang200 (approved):
:+1: Thanks for the discussion about the removal of `SHAMapStoreImp::onChildrenStopped()`

PR #3696, opened 2020-12-10
Levelization analysis and documentation

ximinez:

This PR adds documentation for our levelization process and intentions, as well as a script and Github Action that enforces the existing dependency tree. The script did find problems (cyclic dependencies) in the current code. No effort was made to address those at this time. (See "Future Tasks".)

There are *no* changes to rippled code.

### Context of Change

The PR #3590 introduced a bunch of changes that broke my personal understanding of levelization, but I was unable to explain what that should mean. This PR arose out of efforts to provide documentation for that PR. I realized the results of the analysis would be useful in general for rippled.

### Type of Change

- [X ] Documentation Updates
- [X ] Release (because the Github Action will prevent release of changes that break levelization)

## Test Plan

So long as the new automatic Github Action doesn't fail, there should be nothing to test.

## Future Tasks

This work found and documented several cyclical dependencies in rippled (loops.txt). Future work can be done to unwind those cycles so the dependency arrow only points in one direction. 

---
This change is [Reviewable](https://reviewable.io/reviews/ripple/rippled/3696)


Builds/levelization/levelization.sh
```diff
@@ -0,0 +1,132 @@
+#!/bin/bash
+
```
thejohnfreeman, line 2: What is the interface of this script? Does it assume a particular working directory? Is it customizable through any environment variables or command line arguments or input? If so, can we document that interface at the top in a comment?

Builds/levelization/README.md
```diff
@@ -0,0 +1,74 @@
...
+| 13           | test/net test/protocol test/ledger test/consensus test/core test/server test/nodestore
+| 14           | test/rpc test/app
+
+(Note that `test` levelization is *much* less important and *much* less
+strictly enforced than `ripple` levelization, other than the requirement
+that `test` code should *never* be included in `ripple` code.)
+
+## Validation
+
+The [levelization.sh](levelization.sh) script can be run at any time on
```
thejohnfreeman, line 45: If we add usage instructions to the script, maybe give them an explicit callout here.

thejohnfreeman (changes requested):

Builds/levelization/README.md
```diff
@@ -0,0 +1,74 @@
+# Levelization
+
+Levelization is the term used to describe efforts to prevent rippled from
+having or creating cyclic dependencies.
+
+rippled code is organized into directories under `src/rippled` (and
+`src/test`) representing modules. The modules are intended to be
+organized into "tiers" or "levels" such that a module from one level can
+only include code from the same or lower levels, though including code
```
thejohnfreeman, line 9: Permitting edges at the same level permits the possibility of a cycle. If there are no cycles in practice, then it is possible to introduce new levels as needed such that no module depends on another module at the same level. Should we just change the permission and introduce those levels?

cdy20:
Ha-ha, it seems to be working. At least it prohibits publishing of SQL interface PR with the following log:

```
diff --git a/Builds/levelization/results/loops.txt b/Builds/levelization/results/loops.txt
index a203e4a..be2c62f 100644
--- a/Builds/levelization/results/loops.txt
+++ b/Builds/levelization/results/loops.txt
@@ -8,11 +8,14 @@ Loop: ripple.app ripple.net
   ripple.app > ripple.net
 
 Loop: ripple.app ripple.nodestore
-  ripple.app > ripple.nodestore
+  ripple.nodestore ~= ripple.app
 
 Loop: ripple.app ripple.overlay
   ripple.overlay > ripple.app
 
+Loop: ripple.app ripple.peerfinder
+  ripple.peerfinder ~= ripple.app
+
 Loop: ripple.app ripple.rpc
   ripple.rpc > ripple.app
```
A question is: what means signs `>`, `~=`? Also, would it prohibit publishing of the PR if there is no any sign `+` in diff, but present some `-`?

Builds/levelization/levelization.sh
ximinez: Good question. It has no options, and can be run from anywhere, as long as the script is in it's correct place in the repo tree. That's what line 3: `pushd $( dirname $0 )` does.

Builds/levelization/README.md
ximinez: Yes, you're right. I'll reword this.

ximinez:
> Ha-ha, it seems to be working. At least it prohibits publishing of SQL interface PR with the following log:

Woo hoo!

> A question is: what means signs `>`, `~=`?

I'll add this to the README, too. The signs are basically related to the number of times each module is included in the other.
* `A > B` means that A is included in B significantly more than B is included in A, and thus A should probably be at a higher level. Incidentally, this is why I ended up putting `ripple/rpc` above `ripple/app`.
* `A ~= B` means that A and B are included in each other a different number of times, but the values are so close that the script can't definitively say that one should be above the other.
* `A = B` means that A and B include each other the same number of times, so the script has no clue which should be higher. Some of these may currently be documented at the same level (though I'm not sure now), so as @thejohnfreeman said, we'll need more levels once we tease them apart.

>  Also, would it prohibit publishing of the PR if there is no any sign `+` in diff, but present some `-`?

Yeah, any change will prevent the push. If the change is beneficial, the idea is you can run the script on your end, and commit the changes. It takes less than 30 seconds on Linux. Reviewers would be expected to verify that the change is beneficial.

cdy20:
May be it makes a sense to print the number of includes in illegal direction for each loop? This would help to determine if any new illegal include added.

Builds/levelization/levelization.sh
ximinez: "Usage" added to the script.

Builds/levelization/README.md
ximinez: Done

ximinez:
> May be it makes a sense to print the number of includes in illegal direction for each loop? This would help to determine if any new illegal include added.

I had that kind of output in an earlier revision of the script, but I decided to simplify it so that the Github Action wouldn't trigger on smaller changes, even if they're for the worse. I didn't want it to get so annoying that it would get ignored or removed. I also didn't want to require a lot of effort spent merging the result files.

Fortunately, it's easy enough to get the numbers yourself (I'll put this in the README):
1. Run `levelization.sh`
2. Grep the modules in `paths.txt`. For example, to see how `A` relates to `B`, `grep -w A Builds/levelization/results/paths.txt | grep -w B`

Builds/levelization/levelization.sh
```diff
@@ -0,0 +1,137 @@
...
+grep -r '#include.*/.*\.h' src/ripple/ src/test/ | \
+    grep -v boost | tee ${includes}
+popd
+pushd results
+
+oldifs=${IFS}
+IFS=:
+mkdir includes
+mkdir includedby
+# i=1000
```
thejohnfreeman, line 23: Is it ok to remove this commented code? There are a few other blocks of commented code. Can they be removed too?

thejohnfreeman (changes requested):

Builds/levelization/README.md
```diff
@@ -0,0 +1,95 @@
...
+in one level should never include code in an `impl` folder of any level
+other than it's own.
+
+Unfortunately, over time, enforcement of levelization has been
+inconsistent, so the current state of the code doesn't necessarily
+reflect these rules. Whenever possible, developers should refactor any
+levelization violations they find (by moving files or individual
+classes). At the very least, don't make things worse.
+
+The table below summarizes the _desired_ division of modules.
```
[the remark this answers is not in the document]
ximinez: Good idea. Done.

Builds/levelization/README.md
```diff
@@ -0,0 +1,95 @@
...
+  changed.  Unfortunately, if changes are detected, it can't tell if
+  they are improvements or not, so if you have resolved any issues or
+  done anything else to improve levelization, run `levelization.sh`,
+  and commit the updated results.
+
+The  `loops.txt` and `ordering.txt` files relate the modules
+using comparison signs, which indicate the number of times each
+module is included in the other.
+
+* `A > B` means that A is included in B significantly more than B is included in A, and thus A should probably be at a higher level. `ordering.txt` will only include these types of results, because by definition `B` doesn't include `A` at all, otherwise it would be in `loops.txt`.
```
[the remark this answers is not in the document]
ximinez: Math is hard! :grinning: Fixed.

Builds/levelization/levelization.sh
ximinez: Done. I rephrased some of the comments, for accuracy, too.

thejohnfreeman (approved):

PR #3697, opened 2020-12-10
Remove ReadViewFwdRange `end` iterator cache:

seelabs:

ReadViewFwdRange was storing a cached `end_` iterator that was lazily
created in an iterators `end()` function. When the cache is empty, and
the range it iterated from multiple threads, this creates a race
condition.

This change has performance consequences for "old style" for loops.
For example:

```
 // don't do this
for(auto i = tx_range.begin(); i != tx_range.end(); ++i)

```

Can call the now expensive `end()` function more often than needed.
Range-based for loop (I.e. `for(auto const& t : tx_range)`) should be
used instead.

### Type of Change

## Future Tasks

We may want consider a `for_each` style interface with a lamba in the future, to prevent people for writing old-style for loops.

### DIFF
--- a/src/ripple/ledger/impl/ReadView.cpp
+++ b/src/ripple/ledger/impl/ReadView.cpp
@@ -135,11 +135,9 @@ ReadView::sles_type::begin() const -> iterator
 }
 
 auto
-ReadView::sles_type::end() const -> iterator const&
+ReadView::sles_type::end() const -> iterator
 {
-    if (!end_)
-        end_ = iterator(view_, view_->slesEnd());
-    return *end_;
+    return iterator(view_, view_->slesEnd());
 }
 
 auto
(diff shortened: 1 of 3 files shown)


miguelportilla (approved):
👍

src/ripple/ledger/impl/ReadView.cpp
```diff
@@ -165,11 +163,9 @@ ReadView::txs_type::begin() const -> iterator
 }
 
 auto
-ReadView::txs_type::end() const -> iterator const&
+ReadView::txs_type::end() const -> iterator
 {
-    if (!end_)
-        end_ = iterator(view_, view_->txsEnd());
-    return *end_;
+    return iterator(view_, view_->txsEnd());
```
cjcobb23, line 168: why not keep the caching/lazy evaluation and just use a mutex to avoid the race condition?

nbougalis (approved):

nbougalis:
Merged as 183dcd08d94e2fd087fed40f00ef088458808e04

PR #3698, opened 2020-12-10
Fix nested locks in ValidatorList:

ximinez:

* Found several functions called under lock that take a lock. Refactor
  to require a lock as a parameter instead.
* Found several functions called under lock that don't take a lock, but
  should. Refactored those as well to require a lock as a parameter.

### Context of Change

Due to a change in 1.7.0-b5 that changed mutex type, so UB behavior went from being harmless to causing deadlocks.

### Type of Change

## Test Plan

Run rippled with a loop making frequent requests to the `crawl/` endpoint.

---
This change is [Reviewable](https://reviewable.io/reviews/ripple/rippled/3698)

### DIFF
--- a/src/ripple/app/misc/impl/ValidatorList.cpp
+++ b/src/ripple/app/misc/impl/ValidatorList.cpp
@@ -195,20 +195,23 @@ ValidatorList::load(
 }
 
 boost::filesystem::path
-ValidatorList::GetCacheFileName(PublicKey const& pubKey)
+ValidatorList::GetCacheFileName(
+    ValidatorList::unique_lock const&,
+    PublicKey const& pubKey)
 {
     return dataPath_ / (filePrefix_ + strHex(pubKey));
 }
 
 void
 ValidatorList::CacheValidatorFile(
+    ValidatorList::unique_lock const& lock,
     PublicKey const& pubKey,
     PublisherList const& publisher)
 {
     if (dataPath_.empty())
         return;
 
-    boost::filesystem::path const filename = GetCacheFileName(pubKey);
+    boost::filesystem::path const filename = GetCacheFileName(lock, pubKey);
 
     boost::system::error_code ec;
 
@@ -310,7 +313,7 @@ ValidatorList::applyList(
 
     Json::Value list;
     PublicKey pubKey;
-    auto const result = verify(list, pubKey, manifest, blob, signature);
+    auto const result = verify(lock, list, pubKey, manifest, blob, signature);
     if (result != ListDisposition::accepted)
     {
         if (result == ListDisposition::same_sequence &&
@@ -429,7 +432,7 @@ ValidatorList::applyList(
     }
 
     // Cache the validator list in a file
-    CacheValidatorFile(pubKey, publisher);
+    CacheValidatorFile(lock, pubKey, publisher);
 
     return applyResult;
 }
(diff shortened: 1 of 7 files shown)


src/ripple/app/misc/impl/ValidatorList.cpp
```diff
@@ -997,7 +1030,7 @@ ValidatorList::getNegativeUNL() const
 void
 ValidatorList::setNegativeUNL(hash_set<PublicKey> const& negUnl)
 {
-    std::lock_guard lock{mutex_};
+    std::unique_lock lock{mutex_};
```
HowardHinnant, line 1033: Super nitpicky:  This could be `unique_lock` (unqualified).  Only difference is readability.  I found myself wondering why the local typedefs weren't used.  Ditto 2 places above with `std::shared_lock`.

HowardHinnant (approved):

src/ripple/app/misc/impl/ValidatorList.cpp
ximinez: My intention was to only use the typedefs where they were required, because the template params couldn't be determined by the compiler. I don't have a strong opinion against changing it, but I'm fine leaving it as-is.

HowardHinnant: Fine with me either way.  Approved.  Nice detective work on this one!

ximinez: Thanks!

ximinez:
@HowardHinnant Thanks for the speedy first review! Can I trouble you for another pass to look at the second commit? The first one is unchanged.

src/ripple/shamap/impl/SHAMapInnerNode.cpp
```diff
@@ -49,6 +49,10 @@ SHAMapInnerNode::SHAMapInnerNode(
 {
 }
 
+SHAMapInnerNode::~SHAMapInnerNode()
+{
+}
+
```
HowardHinnant, line 55: Consider:  `SHAMapInnerNode::~SHAMapInnerNode() = default;`
(yes it actually works fine in a source)
Regardless, approved.

HowardHinnant (approved):

src/ripple/shamap/impl/SHAMapInnerNode.cpp
ximinez: I like that better, assuming it doesn't cause any problems.

src/ripple/app/misc/impl/ValidatorList.cpp
```diff
@@ -195,20 +195,23 @@ ValidatorList::load(
 }
 
 boost::filesystem::path
-ValidatorList::GetCacheFileName(PublicKey const& pubKey)
+ValidatorList::GetCacheFileName(
+    ValidatorList::unique_lock const&,
+    PublicKey const& pubKey)
```
pwang200, line 200: Super nitpicky: const?

pwang200 (approved):
:+1: Looks great.

nbougalis:
Merged as 6eb93e9f8adc91c19859dcb71eaa7f3f4ed4487d

src/ripple/app/misc/impl/ValidatorList.cpp
ximinez: Looks like I already made that change in #3619, and this is already merged, so I'll let it go from there.

PR #3703, opened 2020-12-15
Make TicketBatch a SupportedAmendment so it can be enabled.

scottschurr:

`TicketBatch` has been in the code base for a while, but it has not been part of the `SupportedAmendments`.  This was to prevent accidental enablement.

Since we're thinking that it may make sense to enable `TicketBatch` after the release of 1.7.0 this commit adds it to `SupportedAmendments`.  The commit also removes a number of workarounds in the unit tests.  The workarounds are no longer needed now that `TicketBatch` is in `SupportedAmendments`.

This commit does not to be reflected in the release notes, unless it is to note that `TicketBatch` is a `SupportedAmendment` with all of the attendant risks and benefits :wink:

### Type of Change

- [x] Allow feature to be enabled on the network.

## Future Tasks

Shepherd the `TicketBatch` amendment once it attains 80% support on the network.

### DIFF
--- a/src/ripple/protocol/impl/Feature.cpp
+++ b/src/ripple/protocol/impl/Feature.cpp
@@ -132,7 +132,7 @@ detail::supportedAmendments()
         "HardenedValidations",
         "fixAmendmentMajorityCalc",
         //"NegativeUNL",      // Commented out to prevent automatic enablement
-        //"TicketBatch",      // Commented out to prevent automatic enablement
+        "TicketBatch",
     };
     return supported;
 }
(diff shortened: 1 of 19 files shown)


intelliot (approved):
I do think this commit should be reflected in the release notes, even if just to note that TicketBatch is now a SupportedAmendment. That is important to know!

scottschurr:
Rebased to 1.7.0-b9 with no conflicts.

pwang200 (approved):
:+1: 
Sorry for the late review.

PR #3706, opened 2020-12-16
Refactor PerfLog

thejohnfreeman:
Fixed some problems with `PerfLog` that I noticed while studying the code, and tried simplifying portions where I could. I tried making small, independently-reviewable commits to ease review. There's some other minor improvements to test code lumped in.


src/ripple/basics/impl/PerfLogImp.cpp
```diff
@@ -79,89 +80,82 @@ PerfLogImp::Counters::countersJson() const
     Rpc totalRpc;
     for (auto const& proc : rpc_)
     {
-        Json::Value p(Json::objectValue);
+        Rpc const* value;
```
seelabs, line 83: Since we're no longer making a copy, I think the new code now access the member vars outside the lock.

One possibility is to make the copy before checking the `continue` condition. This is less efficient, but the it isn't _that_ bad (sometimes unnecessarily copying 32 bytes) and the code works out nicely. Of course, we could go back the lambda/optional solution too. I probably prefer that, but I'm OK with either.

src/ripple/basics/impl/PerfLogImp.cpp
```diff
@@ -79,89 +80,82 @@ PerfLogImp::Counters::countersJson() const
...
     }
 
     Json::Value jqobj(Json::objectValue);
     // totalJq represents all jobs. All enqueued, started, finished, etc.
-    Jq totalJq("total");
+    Jq totalJq;
     for (auto const& proc : jq_)
     {
-        Json::Value j(Json::objectValue);
+        Jq const* value;
```
seelabs, line 122: See above: we're now accessing member vars outside of a lock.

src/ripple/basics/impl/PerfLogImp.cpp
```diff
@@ -273,6 +260,7 @@ PerfLogImp::run()
     {
         {
             std::unique_lock<std::mutex> lock(mutex_);
+            cond_.wait_until(lock, lastLog_ + setup_.logInterval);
```
seelabs, line 263: Except for the first iteration, this looks like the same logic to me. I don't object to the change, but I'm not sure I see the reason for the change.

src/ripple/basics/impl/PerfLogImp.cpp
```diff
@@ -302,10 +289,13 @@ PerfLogImp::report()
 
     Json::Value report(Json::objectValue);
     report[jss::time] = to_string(date::floor<microseconds>(present));
-    report[jss::workers] = counters_.workers_;
+    {
+        std::lock_guard lock{counters_.jobsMutex_};
```
seelabs, line 293: Nice find on this!

seelabs (changes requested):
Nice set of cleanups - and thank you for organizing the commits so well for easy review. I spotted one potential issue with locks, but other than that this looks good.

src/ripple/basics/impl/PerfLogImp.cpp
seelabs: It's a bit of a shame we now have to take a lock here. We could consider making `workers_` `atomic` and losing the lock here, but your call.

src/ripple/basics/impl/PerfLogImp.cpp
thejohnfreeman: It doesn't report to the old file one more time after a rotate command has been received, and it shuts down immediately instead of trying to make one last report. It feels "more right" to me.

src/ripple/basics/impl/PerfLogImp.cpp
```diff
@@ -24,6 +24,7 @@
 #include <ripple/json/json_writer.h>
 #include <ripple/json/to_string.h>
 #include <boost/optional.hpp>
+#include "core/JobTypes.h"
```
mtrippled, line 27: brackets, not quotes

src/ripple/basics/impl/PerfLogImp.cpp
seelabs: That makes sense. Thanks for the clarification.

src/ripple/basics/impl/PerfLogImp.cpp
mtrippled: Is the number of worker threads guaranteed to be the size of the jobs_ vector? Or is the jobs_ vector the list of currently executing jobs? If it's the latter, then the meaning of the jss:workers field changes. Namely, it's there to describe the total number of worker threads. That way, the user can compare the number of running jobs against this value to see if the job queue is saturated. In normal operation, workers_ stays the same after the process sets the job queue size.

Please put workers_ back the way it was. Regarding whether to make it atomic, that's probably the most sensible thing.

src/ripple/basics/impl/PerfLogImp.h
```diff
@@ -110,7 +110,6 @@ class PerfLogImp : public PerfLog, Stoppable
         std::unordered_map<std::string, Locked<Rpc>> rpc_;
         std::unordered_map<JobType, Locked<Jq>> jq_;
         std::vector<std::pair<JobType, steady_time_point>> jobs_;
-        int workers_{0};
```
mtrippled, line 122: See above, put this back but make atomic.

src/ripple/basics/impl/PerfLogImp.cpp
```diff
@@ -80,26 +80,26 @@ PerfLogImp::Counters::countersJson() const
...
-        p[jss::duration_us] = std::to_string(value->duration.count());
-        totalRpc.duration += value->duration;
+        p[jss::started] = std::to_string(value.started);
+        totalRpc.started += value.started;
+        p[jss::finished] = std::to_string(value.finished);
+        totalRpc.finished += value.finished;
+        p[jss::errored] = std::to_string(value.errored);
+        totalRpc.errored += value.errored;
+        p[jss::duration_us] = std::to_string(value.duration.count());
+        totalRpc.duration += value.duration;
```
mtrippled, line 102: This looks good--I made my previous comments while you were working on this fix.

mtrippled (changes requested):

src/ripple/basics/impl/PerfLogImp.cpp
```diff
@@ -79,89 +80,82 @@ PerfLogImp::Counters::countersJson() const
...
-            totalRpc.sync.started += sync->started;
-            p[jss::finished] = std::to_string(sync->finished);
-            totalRpc.sync.finished += sync->finished;
-            p[jss::errored] = std::to_string(sync->errored);
-            totalRpc.sync.errored += sync->errored;
-            p[jss::duration_us] = std::to_string(sync->duration.count());
-            totalRpc.sync.duration += sync->duration;
+            }
+            value = proc.second.value;
         }
```
seelabs, line 92: What do you think about something like this?
```
        Rpc const value = [&] {
            std::lock_guard lock(proc.second.mutex);
            return proc.second.value;
        }();

        if (!value.started && !value.finished && !value.errored)
        {
            continue;
        }
```
This will sometimes save a default construction, makes the var `const`, and makes it clearer  that `value` is just a copy of `proc.second.second`. OTOH, it will sometimes copy in the vaues of `proc.second.value` when it doesn't need to. I'm fine with it as-is if you prefer. Your call.

seelabs (approved):
:+1: Looks good. I one minor suggestion, but it's good to go whichever way to decide to resolve that.

src/ripple/basics/impl/PerfLogImp.cpp
seelabs: @mtrippled The three options I see are:
1. Make `value` and optional, check for continue, assign into the optional. This does the minimal amount of copies.
2. Default construct `value`, check for continue, assign into the value. This does an extra default construction.
3. Unconditionally copy the value (under a lock), then check for continue. This does an extra copy sometimes.

I wouldn't expect copying the values to be much more expensive than default constructing - zero out 32 bytes vs. copy 32 bytes. And I think the code is a little nicer - make the variable `const` and get rid of the `continue` check that interrupts the initialization. I prefer (2) myself, but all of these solutions are fine.

I had something like this in mind:
```
Rpc const value = [&] {
            std::lock_guard lock(proc.second.mutex);
            return proc.second.value;
        }();

        if (!value.started && !value.finished && !value.errored)
        {
            continue;
        }
```

I'll also trot out my one line copy under a lock troll-code:  
```
Rpc const value = (std::lock_guard{proc.second.mutex}, proc.second.value);
```

src/ripple/basics/impl/PerfLogImp.cpp
thejohnfreeman: The number of worker threads is guaranteed to be the size of the `jobs_` vector. Notice:

- An item in `jobs_` has job type `INVALID` and default-constructed "start time" whenever it [finishes](https://github.com/ripple/rippled/pull/3706/files#diff-64246f9206962a784344c0dbcbff598c1052de9290342f27faaafe566bfa5d1fL438) or when it is [constructed](https://github.com/ripple/rippled/pull/3706/files#diff-64246f9206962a784344c0dbcbff598c1052de9290342f27faaafe566bfa5d1fL447).
- `workers_` was assigned only when the size of `jobs_` changed during [resize](https://github.com/ripple/rippled/pull/3706/files#diff-64246f9206962a784344c0dbcbff598c1052de9290342f27faaafe566bfa5d1fL445), not when jobs started or finished.

Reporting already acquires a bunch of locks to copy out the counter for every used job type and RPC method type. I don't think the difference of one more lock is substantial enough to make it worth introducing the complexity of maintaining a second job counter.

src/ripple/basics/impl/PerfLogImp.cpp
```diff
@@ -183,12 +183,6 @@ PerfLogImp::Counters::currentJson() const
         if (j.first == jtINVALID)
             continue;
         Json::Value jobj(Json::objectValue);
-        auto const e = jq_.find(j.first);
-        if (e == jq_.end())
-        {
-            assert(false);
-            continue;
-        }
```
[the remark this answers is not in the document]
thejohnfreeman: Wouldn't that be the same check as is performed in `jobStart`?

src/ripple/basics/impl/PerfLogImp.h
mtrippled: @thejohnfreeman Can you address/respond to this please? ^

mtrippled (approved):
:+1: LGTM

PR #3712, opened 2020-12-23
Fix a couple of unusual startup and connection issues

ximinez:

1. Ensure that the `PeerImp::doAccept` incoming connection handler call to `async_write` has it's callback run on the strand.
2. Move all peer post-connection messages to `doProtocolStart`.
3. Also, only send validator list on incoming connection instead of both incoming and outgoing.
4. Move the `ShardArchiveHandler` and `ValidatorSites` startup calls to the end of `setup`.

### Context of Change

For number 1, it's possible, though unlikely for a `async_write` to be called while the `async_write` for the response is in progress. This asserts in beast. 2 and 3 are just cleanups that made sense with this change.

Number 4, fixes an unusual startup race condition. With a stopped rippled still open in gdb, attempting to start another rippled fails because the ports are still locked up, but the failed rippled doesn't cleanly exit. This can be a separate commit once the other commits are squashed.

### Type of Change

## Test Plan

I'll be running this code on my dogfood machine, which accepts incoming connections, indefinitely to see if it doesn't assert.

Additionally, a test network consisting of only nodes which include this change should continue to function normally.

---
This change is [Reviewable](https://reviewable.io/reviews/ripple/rippled/3712)

### DIFF
--- a/src/ripple/overlay/impl/PeerImp.cpp
+++ b/src/ripple/overlay/impl/PeerImp.cpp
@@ -553,22 +553,6 @@ PeerImp::fail(std::string const& name, error_code ec)
     close();
 }
 
-void
-PeerImp::failinstrand(std::string const& name, error_code ec)
-{
-    if (!strand_.running_in_this_thread())
-        return post(
-            strand_,
-            std::bind(
-                (void (Peer::*)(std::string const&, error_code)) &
-                    PeerImp::fail,
-                shared_from_this(),
-                name,
-                ec));
-    else
-        return fail(name, ec);
-}
-
 boost::optional<RangeSet<std::uint32_t>>
 PeerImp::getShardIndexes() const
 {
@@ -785,29 +769,26 @@ PeerImp::doAccept()
         return buf;
     }();
 
-    // Put a placeholder in the send queue so no other messages will be
-    // sent until this one is complete.
-    send_queue_.push({});
     // Write the whole buffer and only start protocol when that's done.
     boost::asio::async_write(
         stream_,
         write_buffer->data(),
         boost::asio::transfer_all(),
-        [this, write_buffer, self = shared_from_this()](
-            error_code ec, std::size_t bytes_transferred) {
-            // Remove the placeholder from the send queue.
-            assert(!send_queue_.front());
-            send_queue_.pop();
-            if (!socket_.is_open())
-                return;
-            if (ec == boost::asio::error::operation_aborted)
-                return;
-            if (ec)
-                return failinstrand("onWriteResponse", ec);
-            if (write_buffer->size() == bytes_transferred)
-                return doProtocolStart();
-            return fail("Failed to write header");
-        });
+        bind_executor(
+            strand_,
+            [this, write_buffer, self = shared_from_this()](
+                error_code ec, std::size_t bytes_transferred) {
+                assert(strand_.running_in_this_thread());
+                if (!socket_.is_open())
+                    return;
+                if (ec == boost::asio::error::operation_aborted)
+                    return;
+                if (ec)
+                    return fail("onWriteResponse", ec);
+                if (write_buffer->size() == bytes_transferred)
+                    return doProtocolStart();
+                return fail("Failed to write header");
+            }));
 }
 
 std::string
--- a/src/ripple/overlay/impl/PeerImp.cpp
+++ b/src/ripple/overlay/impl/PeerImp.cpp
@@ -182,12 +182,8 @@ PeerImp::run()
     else
         doProtocolStart();
 
-    // Request shard info from peer
-    protocol::TMGetPeerShardInfo tmGPS;
-    tmGPS.set_hops(0);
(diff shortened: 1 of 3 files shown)


src/ripple/overlay/impl/PeerImp.cpp
```diff
@@ -773,6 +772,16 @@ PeerImp::doAccept()
...
 
+    {
+        // Put a pseudo-message into the queue to hold any messages until we get
+        // the response. This message should never be sent, but create a ping
+        // just in case it "escapes".
+        protocol::TMPing message;
+        message.set_type(protocol::TMPing::ptPING);
+        message.set_seq(0);
+
+        send_queue_.push(std::make_shared<Message>(message, protocol::mtPING));
```
gregtatcam, line 783: This is going to block sending permanently. You need to add `onWriteMessage()` (via strand) to the call-back handler in `async_write()` below. I think we simply should put all initial sends into `doProtocolStart()`. After all that's what the meaning of `doProtocolStart()` is - a peer is ready to send and receive messages. Something like this: https://gist.github.com/gregtatcam/37d5a8c9ae1418828d0f0cf112c83159

gregtatcam (changes requested):
I think this can be done without the ping message. See my comments.

src/ripple/overlay/impl/PeerImp.cpp
nbougalis: I agree with @gregtatcam. I have a slightly different implementation, which is along the lines of what Greg has above: https://github.com/nbougalis/rippled/commit/e284d0a18e50b2d486db9841c286e01a51997c89

Generally, I think the protocol _should_ make it very clear than after the onus is on the peer that **established** the connection (i.e. the peer that made the connection which will have `inbound_ == false`): it sends the HTTP upgrade request, and after it receives the response, it must "kick" things by sending further data.

nbougalis (changes requested):
Not quite happy with this. I agree with @gregtatcam.

src/ripple/overlay/impl/PeerImp.cpp
ximinez: @gregtatcam Yep. You're right. My local node ran ok since I started testing this, but any incoming connections lock up until they time out.

@nbougalis So it looks like you moved the `setTimer` and shard request to `doProtocolStart`, which makes sense, and inlined `doAccept` with no significant changes, which seems unnecessary, and IMHO makes `run` harder to read. I pushed a commit which I think works well.

k1jason2082 (approved):

gregtatcam (commented):
:+1: LGTM

src/ripple/overlay/impl/PeerImp.cpp
```diff
@@ -784,13 +785,19 @@ PeerImp::doAccept()
         return buf;
     }();
 
+    // Put a placeholder in the send queue so no other messages will be
+    // sent until this one is complete.
+    send_queue_.push({});
```
gregtatcam, line 790: This is going to block the queue permanently. If there is another `send()` before the handshake response callback is called then the message from this send is added to the send queue since the queue is not empty. The queue now has the dummy message and the valid message. When the queue is popped in the callback and the dummy message is removed, the valid message will permanently block the queue. A way to do this is to call `onWriteMessage({},0)` in the callback to kick-start the queue processing. This has to be done on the strand to synchronize access to the send queue. How about just adding a flag `protocolStarted_` initially set to `false`. The flag is set to `true` in `doProtocolStart()` before any message is sent. Then we can assert in the `send()` if the flag is false. This is a simpler solution and in general we should not send any message before the handshake is complete. Also the dummy message doesn't address a case, albeit hypothetical, when a message is sent before the dummy message is added to the queue.

src/ripple/overlay/impl/PeerImp.cpp
```diff
@@ -945,6 +952,7 @@ PeerImp::onWriteMessage(error_code ec, std::size_t bytes_transferred)
     if (!send_queue_.empty())
     {
         // Timeout on writes only
+        assert(send_queue_.front());
```
gregtatcam, line 955: Don't need this if the dummy message is not going to be used.

gregtatcam (changes requested):

src/ripple/overlay/impl/PeerImp.cpp
ximinez: Removed the unnecessary asserts.

ximinez:
Note to self, and reviewers: I'm going to rewrite this PR and the commit message to focus more on "fix unusual app and connection start up problems". I'll also plan to squash the existing commits sooner than later, since most of them are unrealized experiments. If anyone is reviewing before I do, I'd suggest looking at the full changeset instead of going commit by commit.

gregtatcam (approved):
:+1: LGTM

ximinez:
Updated the PR description to more accurately reflect the changes.

As described by the test plan, so far: I have been running rippled for more than 3 days, 18 hours without incident, where previously, it would hit this assert within 12 hours. At this moment, it has 23 incoming connections.

PR #3716, opened 2020-12-28
Add start-sync-stop Python script

thejohnfreeman:
Added a Python script for start-sync-stop tests based on Bash scripts from @scottschurr. I'd appreciate some Windows testers, please. I may need to think of an alternative implementation for the log rotation.


k1jason2082 (approved):

bin/start_sync_stop.py
```diff
@@ -0,0 +1,137 @@
...
+    """A context manager for a rippled process."""
+    # Start the server.
+    process = await asyncio.create_subprocess_exec(
+        path,
+        '--conf',
+        config,
+        stdout=subprocess.DEVNULL,
+        stderr=subprocess.DEVNULL,
+    )
+    logging.info(f'rippled started with pid {process.pid}')
```
scottschurr, line 34: _You_ know the `<num>` in `debug.<num>.log` indicates the number of iterations.  But that doesn't make it obvious to the rest of us.  Consider passing the iteration as an argument to `rippled()`, then print it here.  Perhaps something like...
```
logging.info(f'iteration {iteration} started. rippled pid: {process.pid}')
```

bin/start_sync_stop.py
```diff
@@ -0,0 +1,137 @@
...
+        yield process
+    finally:
+        # Ask it to stop.
+        logging.info(f'asking rippled ({process.pid}) to stop')
+        start = time.time()
+        process.terminate()
+
+        # Wait nicely.
+        try:
+            await asyncio.wait_for(process.wait(), 60)
```
scottschurr, line 45: Running a debug version of the server locally, I find that a polite shutdown can take on the order of 90 seconds.  I suggest you crank the `await` timeout to `120`.

bin/start_sync_stop.py
```diff
@@ -0,0 +1,137 @@
...
+            logging.warning(f'killing rippled ({process.pid})')
+            try:
+                process.kill()
+            except ProcessLookupError:
+                pass
+
+        code = await process.wait()
+        end = time.time()
+        logging.info(
+            f'rippled stopped after {end - start} seconds with code {code}'
```
scottschurr, line 57: This echos a _lot_ of significant digits.  Consider reducing the visual noise with something like...
```
        logging.info(
            'rippled stopped after {0:.2f} seconds with code {1}'.format(
                end - start, code)
```

bin/start_sync_stop.py
```diff
@@ -0,0 +1,137 @@
...
+                continue
+
+        try:
+            state = body['result']['state']['server_state']
+        except KeyError as cause:
+            logging.warning(f'server_state response missing key: {cause.key}')
+            stopwatch = 0
+            continue
+        logging.info(f'server_state: {state}')
+        if state == 'full':
```
scottschurr, line 89: There are three different state values that all correspond to `full`: `full`, `validating`, and `proposing`.  You should allow all three.

I also suggest that you require that the server _stay_ in sync for the full duration.  What you currently have allows a server to fall out of sync repeatedly.  I think you should require the fully synced time to be contiguous.

Given that, I suggest you change the code to something like this:
```
        full_states = {'full', 'validating', 'proposing'}
        if state in full_states:
            stopwatch += interval
        else:
            # rippled must stay in sync for the full duration.  Start over if
            # it loses sync.
            stopwatch = 0
```

bin/start_sync_stop.py
```diff
@@ -0,0 +1,137 @@
...
+            if done == {exited}:
+                code = exited.result()
+                logging.warning(
+                    f'server halted for unknown reason with code {code}'
+                )
+            else:
+                assert done == {synced}
+                assert synced.exception() is None
+            end = time.time()
+            logging.info(f'{end - start} seconds to sync')
```
scottschurr, line 128: Whatever you do to improve the float formatting up above, you should do the same here as well.

scottschurr (changes requested):
Nice!  Much tidier and easier to maintain than the scripts I cobbled together.  I'm not much of a Python person, so I may have left suggestions that are not "pythonic".  Please fix (or ignore) them as you see fit.

That said, I think the change to the
```
if state == 'full':
```
condition is pretty important.  If you choose not to address that you'll need a pretty good motivation.

I'm open to being wrong on any of these comments.  Let me know what you think.

bin/start_sync_stop.py
```diff
@@ -0,0 +1,137 @@
+#!/usr/bin/env python
+"""A script to test rippled in an infinite loop of start-sync-stop.
+
+- Expects 'rippled' to be on the `PATH` and 'rippled.cfg' to be in the current
+  directory.
```
seelabs, line 5: I keep many exe and config files around. It would be nice if the executable and config file could be set with command line arguments.

bin/start_sync_stop.py
```diff
@@ -134,4 +151,45 @@ async def loop(path='rippled', config='rippled.cfg', port=5005):
...
+    help='Number of seconds to wait between polls of state.',
+)
+args = parser.parse_args()
+
+
+def test():
+    return sync(args.port, duration=args.duration, interval=args.interval)
+
+
+asyncio.run(loop(test, exe=args.rippled, config=args.conf))
```
scottschurr, line 195: Specifying an "interesting" path on the command line causes the program to error out.  Here's the command line I specified:
```
python3 start_sync_stop.py ../build/rippled --conf ../../../.config/ripple/rippled.cfg
```
Here's the traceback:
```
Traceback (most recent call last):
  File "start_sync_stop.py", line 195, in <module>
    asyncio.run(loop(test, exe=args.rippled, config=args.conf))
  File "/usr/local/Cellar/python/3.7.7/Frameworks/Python.framework/Versions/3.7/lib/python3.7/asyncio/runners.py", line 43, in run
    return loop.run_until_complete(main)
  File "/usr/local/Cellar/python/3.7.7/Frameworks/Python.framework/Versions/3.7/lib/python3.7/asyncio/base_events.py", line 587, in run_until_complete
    return future.result()
  File "start_sync_stop.py", line 128, in loop
    async with rippled(exe, config) as process:
  File "/usr/local/Cellar/python/3.7.7/Frameworks/Python.framework/Versions/3.7/lib/python3.7/contextlib.py", line 170, in __aenter__
    return await self.gen.__anext__()
  File "start_sync_stop.py", line 42, in rippled
    stderr=subprocess.DEVNULL,
  File "/usr/local/Cellar/python/3.7.7/Frameworks/Python.framework/Versions/3.7/lib/python3.7/asyncio/subprocess.py", line 217, in create_subprocess_exec
    stderr=stderr, **kwds)
  File "/usr/local/Cellar/python/3.7.7/Frameworks/Python.framework/Versions/3.7/lib/python3.7/asyncio/base_events.py", line 1533, in subprocess_exec
    f"program arguments must be a bytes or text string, "
TypeError: program arguments must be a bytes or text string, not PosixPath
```

scottschurr (changes requested):
Most of the changes look great, but interesting paths on the command line (or hacked into the source code) don't appear to work.  That should probably be fixed.

bin/start_sync_stop.py
thejohnfreeman: You're using Python 3.7 which doesn't automatically stringify `Path` arguments to `subprocess` functions. I can easily make some changes to accommodate that though, and add an assertion for the minimum required version (right now it's just a comment).

ximinez:
Naive Windows test:
```
$ ./bin/start_sync_stop.py --conf ~/.config/ripple/rippled.cfg --port 5015 ./build/cmake/msvc.ON/Debug/rippled.exe
Traceback (most recent call last):
  File "./bin/start_sync_stop.py", line 199, in <module>
    asyncio.run(loop(test, exe=args.rippled, config=args.conf))
  File "C:\Program Files (x86)\Python37-32\lib\asyncio\runners.py", line 43, in run
    return loop.run_until_complete(main)
  File "C:\Program Files (x86)\Python37-32\lib\asyncio\base_events.py", line 568, in run_until_complete
    return future.result()
  File "./bin/start_sync_stop.py", line 130, in loop
    os.symlink(target, link)
OSError: symbolic link privilege not held
```

scottschurr (approved):
:+1: Works great for me with the latest changes.  I approve assuming @ximinez's issue is addressed.

ximinez:
Unfortunately, the symlink permission isn't the only problem. When I comment it out, the result is
```
$ ./bin/start_sync_stop.py --conf ~/.config/ripple/rippled.cfg --port 5015 ./build/cmake/msvc.ON/Debug/rippled.exe
2021-01-07 12:47:40 INFO     iteration 0 logging to debug.0.log
Traceback (most recent call last):
  File "./bin/start_sync_stop.py", line 199, in <module>
    asyncio.run(loop(test, exe=args.rippled, config=args.conf))
  File "C:\Program Files (x86)\Python37-32\lib\asyncio\runners.py", line 43, in run
    return loop.run_until_complete(main)
  File "C:\Program Files (x86)\Python37-32\lib\asyncio\base_events.py", line 568, in run_until_complete
    return future.result()
  File "./bin/start_sync_stop.py", line 132, in loop
    async with rippled(exe, config) as process:
  File "C:\Program Files (x86)\Python37-32\lib\contextlib.py", line 170, in __aenter__
    return await self.gen.__anext__()
  File "./bin/start_sync_stop.py", line 46, in rippled
    stderr=subprocess.DEVNULL,
  File "C:\Program Files (x86)\Python37-32\lib\asyncio\subprocess.py", line 217, in create_subprocess_exec
    stderr=stderr, **kwds)
  File "C:\Program Files (x86)\Python37-32\lib\asyncio\base_events.py", line 1516, in subprocess_exec
    bufsize, **kwargs)
  File "C:\Program Files (x86)\Python37-32\lib\asyncio\base_events.py", line 444, in _make_subprocess_transport
    raise NotImplementedError
NotImplementedError
```
That looks a bit more significant.

ximinez (changes requested):
Other than the Windows compatibility issues, which have prevented me from testing yet, this looks pretty good. I just have a few questions.

bin/start_sync_stop.py
```diff
@@ -0,0 +1,199 @@
...
+            data=json.dumps({
+                'method': 'server_state'
+            }).encode(),
+            headers={'Content-Type': 'application/json'},
+        )
+        with urllib.request.urlopen(request) as response:
+            try:
+                body = json.loads(response.read())
+            except urllib.error.HTTPError as cause:
+                logging.warning(f'server_state returned s.o.t. JSON: {cause}')
```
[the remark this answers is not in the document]
thejohnfreeman: "something other than". I wrote these messages on the fly in a different state of mind. Happy to change them to something more comprehensible.

thejohnfreeman:
@ximinez It seems you're running Python 3.7 too, and in that version, the default event loop for Windows is the `SelectorEventLoop` which does not support subprocesses. In 3.8, that default was [changed](https://docs.python.org/3/library/asyncio-platforms.html#windows) to the `ProactorEventLoop` which does support subprocesses. I can add a version and platform check to always [use](https://docs.python.org/3.7/library/asyncio-platforms.html#subprocess-support-on-windows) `ProactorEventLoop` on Windows.

For the symbolic links, I'll just switch to moving the log out of the way after each iteration.

bin/start_sync_stop.py
```diff
@@ -0,0 +1,199 @@
...
+
+        try:
+            state = body['result']['state']['server_state']
+        except KeyError as cause:
+            logging.warning(f'server_state response missing key: {cause.key}')
+            stopwatch = 0
+            continue
+        logging.info(f'server_state: {state}')
+        if state in SYNC_STATES:
+            stopwatch += interval
```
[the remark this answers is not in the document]
thejohnfreeman: :+1: I've switched to elapsed time.

ximinez (changes requested):
The Windows run got a lot farther! It handled rippled just fine. Now the problem is with the log file, it looks like.

bin/start_sync_stop.py
```diff
@@ -0,0 +1,196 @@
...
+            logging.info(f'synced after {end - start:.0f} seconds')
+        os.replace('debug.log', f'debug.{id}.log')
+        id += 1
+
+
+logging.basicConfig(
+    format='%(asctime)s %(levelname)-8s %(message)s',
+    level=logging.INFO,
+    datefmt='%Y-%m-%d %H:%M:%S',
+)
```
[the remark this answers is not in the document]
ximinez: I just got this same basic error on Linux using python 3.8.6.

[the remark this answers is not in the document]
thejohnfreeman: This is the trouble with rippled using a configuration file but missing a command-line override for every setting. In my local configuration file, I have set `[debug_logfile]` to `debug.log`, but the default is `/var/log/rippled/debug.log`. Setting `[debug_logfile]` multiple times ends up with no logging, so trying to copy the file and override the setting at the beginning or end doesn't work. I guess I can try to make this script read the configuration file and find the setting, but does anyone else find that annoying? Configuration files should be a convenient method of saving and repeating command-line settings; they should not be the only way to change a setting. /rant

bin/start_sync_stop.py
```diff
@@ -0,0 +1,230 @@
...
+DEFAULT_POLL_INTERVAL = 5
+SYNC_STATES = ('full', 'validating', 'proposing')
+
+
+def read_config(config_file):
+    # strict = False: Allow duplicate keys, e.g. [rpc_startup].
+    # all_no_value = True: Allow keys with no values. Generally, these
+    # instances use the "key" as the value, and the section name is the key,
+    # e.g. [debug_logfile].
+    config = configparser.ConfigParser(strict=False, allow_no_value=True)
```
ximinez, line 48: Windows paths can contain the `:` character, which is a default delimiter for `configParser`, but adding `delimiters=('=')` will let the line be read without problem. Well, assuming nobody builds a path using `=`.

bin/start_sync_stop.py
```diff
@@ -40,14 +41,31 @@
 SYNC_STATES = ('full', 'validating', 'proposing')
 
 
+def find_log_file(config_file):
+    """Try to figure out what log file the user has chosen. Raises all kinds
+    of exceptions if there is any possibility of ambiguity."""
+    # strict = False: Allow duplicate keys, e.g. [rpc_startup].
+    # all_no_value = True: Allow keys with no values. Generally, these
```
ximinez, line 45: Typo in comment: "# allow_no_value = True"

ximinez (changes requested):
With the two code changes I describe below, this works beautifully in Windows.

One question: When I stop the script with a Ctrl-C, it stops `rippled` nicely, but then dumps a stack trace for the KeyboardInterrupt exception. Is that expected?
```
2021-01-11 18:30:43 INFO     asking rippled (pid: 4520) to stop
2021-01-11 18:30:43 INFO     rippled stopped after 0.3 seconds with code 1
Traceback (most recent call last):
  File "./bin/start_sync_stop.py", line 230, in <module>
    asyncio.run(loop(test, exe=args.rippled, config_file=args.conf))
  File "C:\Program Files (x86)\Python37-32\lib\asyncio\runners.py", line 43, in run
    return loop.run_until_complete(main)
  File "C:\Program Files (x86)\Python37-32\lib\asyncio\base_events.py", line 555, in run_until_complete
    self.run_forever()
  File "C:\Program Files (x86)\Python37-32\lib\asyncio\base_events.py", line 523, in run_forever
    self._run_once()
  File "C:\Program Files (x86)\Python37-32\lib\asyncio\base_events.py", line 1722, in _run_once
    event_list = self._selector.select(timeout)
  File "C:\Program Files (x86)\Python37-32\lib\asyncio\windows_events.py", line 405, in select
    self._poll(timeout)
  File "C:\Program Files (x86)\Python37-32\lib\asyncio\windows_events.py", line 703, in _poll
    status = _overlapped.GetQueuedCompletionStatus(self._iocp, ms)
KeyboardInterrupt
```

thejohnfreeman:
Yes, the stack trace is expected, but I can squelch it.

bin/start_sync_stop.py
```diff
@@ -41,23 +40,40 @@
...
+            f'too many [debug_logfile] in configuration file: {config_file}')
     return values[0]
 
 
+def find_http_port(config_file):
+    config = read_config(config_file)
+    names = list(config['server'].keys())
+    for name in names:
+        server = config[name]
+        if server.get('protocol', None) == 'http':
```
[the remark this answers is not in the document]
thejohnfreeman: How would a list be written? Comma-separated? Spaces allowed? The `ConfigParser` will return a string, not a list, so we will have to parse the value separately.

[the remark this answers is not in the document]
ximinez: Here's what my config looks like
```
[port_rpc1]
port = 5015
ip = ::
admin = ::ffff:127.0.0.1, ::1
protocol = https, http
permessage_deflate = 1
```
So comma-separated, and spaces allowed. I realize my suggestion won't work if the port is https-only, so either parse or look for word boundaries on "http". BTW, the example I gave _does_ work with my config. I assume the `in` operator finds substrings.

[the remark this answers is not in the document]
thejohnfreeman: `b in a` calls `a.__contains__(b)`. For built-in sequences, that method uses `==` to compare. Substrings will not match.

scottschurr (approved):
:+1: This script is currently working great for me on my Mac.  Looks good!

ximinez (approved):
💯

PR #3722, opened 2021-01-13
Improve manifest relaying:

nbougalis:
The manifest relay code would only ever relay manifests from validators on a server's UNL which means that the manifests of validators that are not broadly trusted can fail to propagate across the network, which can make it difficult to detect and track such validators.
    
This commit, if merged, propagates all manifests on a best-effort basis resulting in broader availability of manifests on the network and avoid the need to introduce on-ledger manifest storage or to establish one or more manifest repositories.

### DIFF
--- a/src/ripple/overlay/impl/OverlayImpl.cpp
+++ b/src/ripple/overlay/impl/OverlayImpl.cpp
@@ -658,12 +658,10 @@ OverlayImpl::onManifests(
     std::shared_ptr<protocol::TMManifests> const& m,
     std::shared_ptr<PeerImp> const& from)
 {
-    auto& hashRouter = app_.getHashRouter();
     auto const n = m->list_size();
     auto const& journal = from->pjournal();
 
-    JLOG(journal.debug()) << "TMManifest, " << n
-                          << (n == 1 ? " item" : " items");
+    protocol::TMManifests relay;
 
     for (std::size_t i = 0; i < n; ++i)
     {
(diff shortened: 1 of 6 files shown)


src/ripple/overlay/impl/PeerImp.cpp
```diff
@@ -991,6 +991,17 @@ PeerImp::onMessageEnd(
...
 {
+    auto const s = m->list_size();
+
+    if (s == 0)
+    {
+        fee_ = Resource::feeUnwantedData;
+        return;
+    }
+
+    if (s > 100)
```
carlhua, line 1002: I think 100 might be too big to begin with?

src/ripple/overlay/impl/OverlayImpl.cpp
```diff
@@ -662,73 +662,57 @@ OverlayImpl::onManifests(
...
+    protocol::TMManifests relay;
 
     for (std::size_t i = 0; i < n; ++i)
     {
         auto& s = m->list().Get(i).stobject();
 
         if (auto mo = deserializeManifest(s))
         {
-            uint256 const hash = mo->hash();
-            if (!hashRouter.addSuppressionPeer(hash, from->id()))
```
HowardHinnant, line 675: OverlayImpl.cpp:661:11: warning: unused variable 'hashRouter' [-Wunused-variable]
```
    auto& hashRouter = app_.getHashRouter();
          ^
```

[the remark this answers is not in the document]
nbougalis: We will not: manifests are only forwarded if they are **successfully** applied—that is, if we haven't seen them before.

HowardHinnant (approved):

gregtatcam (approved):
Looks good assuming the hashRouter warning pointed out by @HowardHinnant  is fixed.

thejohnfreeman (approved):

PR #3725, opened 2021-01-14
[TRIVIAL] Add instructions to use 64-bit Windows build tools

ximinez:

I finally, accidentally figured out why my Windows command line builds (using msbuild) will fail every once in a while to link because a file memory mapping fails, or a plain old out of memory. By default, the 32-bit versions of these tools are used, so the system has ample memory available, the process does not. This change adds instructions to use the `CMAKE_GENERATOR_TOOLSET` to use the 64-bit versions of the tools. The [cmake docs](https://cmake.org/cmake/help/latest/variable/CMAKE_GENERATOR_TOOLSET.html) are clear that "The value of this variable should never be modified by project code.".

Note that this has never happened to me in the IDE, presumably because the default settings use the 64-bit tools.

I also updated Travis to use this option, though I'll only know if it works when that finishes. I'll remove the second commit if not.

### Type of Change

- [X ] Documentation Updates

## Test Plan

No code changes to test. If CI works, it works, and the other change is to documentation.

---
This change is [Reviewable](https://reviewable.io/reviews/ripple/rippled/3725)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -400,6 +400,7 @@ matrix:
       name: windows, visual studio, debug
       script:
         - mkdir -p build.ms && cd build.ms
+        - export CMAKE_EXTRA_ARGS="${CMAKE_EXTRA_ARGS} -DCMAKE_GENERATOR_TOOLSET=host=x64"
         - cmake -G "Visual Studio 15 2017 Win64" ${CMAKE_EXTRA_ARGS} ..
         - export DESTDIR=${PWD}/_installed_
         - travis_wait ${MAX_TIME_MIN} cmake --build . --parallel --verbose --config ${BLD_CONFIG} --target install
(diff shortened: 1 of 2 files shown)


nbougalis (approved):

nbougalis (changes requested):
Seems to fail CI with an error.

ximinez:
@nbougalis CI issue fixed.

nbougalis (approved):

nbougalis:
Looks good. We should get a signoff from @miguelportilla or someone else with Windows though, just in case. If we can do that, this should get into **b12**, @manojsdoshi.

ximinez:
Rebased on to 1.7.0-b11, and squashed the two Travis-only commits into one. The commits will all be squashed to one once I get final approval.

miguelportilla (approved):
👍

ximinez:
Squashed and rewrote the commit message.

PR #3732, opened 2021-01-19
Fix telBAD_PUBLIC_KEY message:

mDuo13:
The previous error description was focused on keys that are too long,
but this error can occur if the key is too short or does not contain
the correct prefix.

Changes the error message for `telBAD_PUBLIC_KEY` to be more accurate.

### Context of Change

See https://github.com/ripple/xrpl-dev-portal/pull/952 for the related documentation update.

### Type of Change

- [x] Documentation Updates

Transaction error descriptions are reported to users in the API but are not part of the consensus process, and the documentation [explicitly calls out that these messages are subject to change without notice](https://xrpl.org/transaction-results.html#immediate-response), so this is a non-breaking, non-amendment change.


scottschurr (approved):
:+1:  Thanks @mDuo13!

nbougalis:
@manojsdoshi we should try to include this in the proposed **b12**.

nbougalis (approved):

intelliot (approved):

PR #3741, opened 2021-01-23
Redesign stoppable objects

thejohnfreeman:
This change implements a new model for "stoppable" objects, heretofore represented by the `Stoppable` interface. The benefits include:

- An easier-to-understand intuition
- A simpler prescription for implementations
- Less code that is more readable (around 1000 fewer lines in non-test source)
- Less work and less synchronization at runtime

In this change, stoppable objects are modeled after [`std::thread`][1], which is already very well documented and widely understood.

## `std::thread`

`std::thread` is a simple state machine with these states:

1. **Empty**: A default-constructed `std::thread` starts in this state. An active `std::thread` that is joined, detached, or move-assigned to another `std::thread` transitions to this state.
2. **Active**: A non-default-constructed `std::thread` starts in this state. An empty `std::thread` that is move-assigned from an active `std::thread` transitions to this state.
3. **Destroyed**: An empty `std::thread` whose destructor is called transitions to this state.

An active `std::thread` that is destroyed or overwritten calls `std::terminate`, and an empty `std::thread` that is joined or detached throws an exception, meaning those are not legal transitions. To prevent trying them, callers can check the return value of [`std::thread::joinable`][4] to see which state the object is in: `true` for active, and `false` for empty.

Once a `std::thread` is joined, it never "restarts". Instead, you can move an active `std::thread` started elsewhere into an empty `std::thread`.

Methods that initiate a state transition (move assignment, `join`, and `detach`) return only after the state transition is complete. State transitions are atomic and synchronous (i.e. blocking).

## Stoppable

The new stoppable objects in rippled no longer implement a common interface, but instead follow a common pattern:

`std::thread` concept | "stoppable" equivalent
---|---
empty | stopped
active | started
non-default-constructor | constructor or `void start()` [^1]
`void join()` | `void stop()`

[^1]: Or `bool start()` in the pesky case of `NodeStore::DatabaseShard`.

Unlike `Stoppable::onStop` and `Stoppable::onChildrenStopped`, the new `stop` methods are guaranteed to block until the state transition is complete, just like [`std::thread::join`][5].

## Caveats

I would like to start every stoppable in its constructor, like `std::thread`, but some stoppable objects are constructed but never started in some tests (e.g. `ApplicationImp` in `TxQ1_test::testMaximum`). To accommodate them, we give them a single synchronous `start` method whose pre-condition is that the object is stopped and whose post-condition is that the object is started.

I would like to give `stop` the pre-condition that the object is started, matching the behavior of `std::thread::join`, but some stoppables are "stopped" twice, e.g. `LoadManager` in the test `Subscribe_test::testServer`.

I would like to give their destructors the pre-condition that the object is stopped, like `std::thread`, but some stoppables are never stopped, e.g. `JobQueue` (which technically "starts" in its constructor) in `TxQ1_test::testMaximum`.

Stoppables often exhibit overlapping but unnested lifetimes. `JobQueue` starts in its constructor, called by `ApplicationImp`'s constructor, but `ApplicationImp` starts in its own `start` method, or in its `setup` method, depending on how you look at it, both of which are called after its constructor, but `JobQueue` is stopped early in the `stop` method of `ApplicationImp`. I would prefer that they stop in the reverse order that they started.

## RAII

`std::thread` expects that each non-default-construction matches with exactly one call to `std::thread::join`. C++ already has an excellent technique for guaranteeing matching function calls: constructors and destructors, and the [RAII pattern][3]. `std::thread` does not follow that pattern, which led frustrated developers to create [`std::jthread`][2] as an alternative. Trying to make the stoppable objects in rippled follow the RAII pattern requires significant changes that I consider just one bridge too far for this changeset. I want to stop here and build support for these changes before pushing further.

## Commits

I've tried to make each commit small, separately reviewable, build, and pass all tests, to make it easy to review step-by-step, if that's your preference, but it means there are many commits.

[1]: https://en.cppreference.com/w/cpp/thread/thread
[2]: https://en.cppreference.com/w/cpp/thread/jthread
[3]: https://en.cppreference.com/w/cpp/language/raii
[4]: https://en.cppreference.com/w/cpp/thread/thread/joinable
[5]: https://en.cppreference.com/w/cpp/thread/thread/join


src/ripple/app/main/Application.cpp
```diff
@@ -1110,7 +1064,30 @@ class ApplicationImp : public Application, public RootStoppable, public BasicApp
...
+            reportingETL_->stop();
+        }
+        m_inboundLedgers->stop();
+        m_inboundTransactions->stop();
+        overlay_->stop();
+        perfLog_->stop();
+        grpcServer_->stop();
+        m_networkOPs->stop();
+        m_ledgerMaster->stop();
+        m_loadManager->stop();
```
cjcobb23, line 1090: `The order of these stop calls is delicate`. I am concerned about this. I feel that this situation is exactly the problem that `Stoppable` was meant to solve (managing the order in which things are stopped); with this PR, now we have to manually maintain this order.

cjcobb23 (changes requested):
This PR changes the order of stopping and starting of various components. Some of this I know to be incorrect (`JobQueue` needs to be stopped after `GRPCServer` for instance). Others I _think_ are incorrect based on the diagram that used to be in `core/Stoppable.h`, which showed the dependencies between the different components. If that diagram was actually incorrect, and you think this new order is safe, I will have to do a deeper dive and review whether the new order is actually safe.

Also, I think we need a similar diagram to what used to be in `core/Stoppable.h`, to clearly represent the dependencies between the different components, so that way developers know which order to maintain when changing the code.

src/ripple/app/main/Application.cpp
thejohnfreeman: We had to manually maintain the order before. Now the order is visible. That's an improvement.

src/ripple/app/main/Application.cpp
```diff
@@ -1110,7 +1064,30 @@ class ApplicationImp : public Application, public RootStoppable, public BasicApp
                 return validators().trustedPublisher(pubKey);
             });
 
-        stopped();
+        // The order of these stop calls is delicate.
+        // Re-ordering them risks undefined behavior.
+        m_shaMapStore->stop();
+        m_jobQueue->stop();
```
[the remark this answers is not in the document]
thejohnfreeman: You're mistaken. This is the consequence of `Stoppable` being so misleading. `Stoppable` gave the impression that children depend on their parents, parents would stop after their children, and because `JobQueue` was so high in the hierarchy, it would stop late in the sequence.

In fact, `JobQueue` was the FIRST `Stoppable` stopped, and it was [stopped](https://github.com/ripple/rippled/blob/c11037fd2764f6fea8e0d184b0b7aa0daa81136f/src/ripple/core/impl/Stoppable.cpp#L214) in `RootStoppable::stop`. Once the `jobCounter` (which is a `ClosureCounter`) is joined, no more jobs will be [created](https://github.com/ripple/rippled/blob/develop/src/ripple/core/JobQueue.h#L168-L173) (all "wrappers" will be [empty](https://github.com/ripple/rippled/blob/c11037fd2764f6fea8e0d184b0b7aa0daa81136f/src/ripple/core/ClosureCounter.h#L178-L187)), and the call to `ClosureCounter::join` [blocks](https://github.com/ripple/rippled/blob/c11037fd2764f6fea8e0d184b0b7aa0daa81136f/src/ripple/core/ClosureCounter.h#L156-L166) until they have all finished executing. There are no jobs and there will not be any more jobs before the first `Stoppable::onStop` is even entered.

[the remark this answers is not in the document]
thejohnfreeman: Luckily, it appears that `GRPCServer` is prepared to [handle](https://github.com/ripple/rippled/blob/c11037fd2764f6fea8e0d184b0b7aa0daa81136f/src/ripple/app/main/GRPCServer.cpp#L130-L135) the case where `JobQueue` stops before it.

src/ripple/app/main/Application.cpp
```diff
@@ -262,35 +258,42 @@ class ApplicationImp : public Application, public RootStoppable, public BasicApp
...
+              config_->section(SECTION_INSIGHT),
+              logs_->journal("Collector")))
+
+        , m_jobQueue(std::make_unique<JobQueue>(
+              m_collectorManager->group("jobq"),
+              logs_->journal("JobQueue"),
+              *logs_,
+              *perfLog_))
+
+        , m_nodeStoreScheduler(*m_jobQueue)
```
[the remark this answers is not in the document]
thejohnfreeman: The biggest changes to `NodeStoreScheduler` were made in [this commit](https://github.com/ripple/rippled/pull/3741/commits/50a99674bd0d96aba6553832033c4feca5f29d29), which is not very large. It makes it easy to see that `NodeStoreScheduler` never needed to be a `Stoppable`. It used `JobQueue`, but was not responsible for stopping `JobQueue`, and it had no resources (e.g. `std::thread`s) that needed to be stopped, thus it was not responsible for stopping anything, thus it did not need to be a `Stoppable`, even in the old framework where `Stoppable` existed. It was one of the first and easiest classes I could remove from the `Stoppable` hierarchy.

If you read my description for this PR, you'll see I'm trying to move to RAII, where starting and stopping is not just "like" pushing and popping from a stack, but enforced by the language. We agree on that. Right now, `Stoppable` is a mess. In `develop` right now, `JobQueue` starts in its constructor, but is stopped before anything else, and "stopped" again (a no-op) in its destructor (this second stop is to cover the case when the `ApplicationImp` is never stopped and thus doesn't stop the `JobQueue`). I'm trying to get it to be less of a mess, but I stopped short of perfect. I stopped at this point to build support for this direction. I want to reach RAII eventually, but getting there will require even more radical changes that will be easier to swallow if this changeset is the starting point, instead of the current state of `develop`.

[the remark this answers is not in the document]
thejohnfreeman: You seemed to have picked up on another mislead of `Stoppable`: that children depend on their parents. `NodeStoreScheduler` and `JobQueue` were a strange contradiction of that that I cannot explain. This changeset removes that contradiction.

src/ripple/overlay/impl/OverlayImpl.cpp
```diff
@@ -566,29 +537,22 @@ OverlayImpl::onPrepare()
...
+OverlayImpl::stop()
 {
-    std::lock_guard lock(mutex_);
-    checkStopped();
+    strand_.dispatch(std::bind(&OverlayImpl::stopChildren, this));
+    {
+        std::unique_lock<decltype(mutex_)> lock(mutex_);
+        cond_.wait(lock, [this] { return list_.empty(); });
+    }
+    m_peerFinder->stop();
```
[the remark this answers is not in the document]
thejohnfreeman: You have to ignore the diagram and documentation of `Stoppable`. This changeset is based on the code as it was, not on how it was documented.

src/ripple/nodestore/impl/DatabaseShardImp.cpp
```diff
@@ -696,21 +689,17 @@ DatabaseShardImp::getCompleteShards()
...
-    for (auto const& e : shards_)
-        e.second->stop();
-}
+    Database::stop();
+    {
+        std::lock_guard lock(mutex_);
+        for (auto const& [_, shard] : shards_)
+            shard->stop();
+    }
+    taskQueue_.stop();
```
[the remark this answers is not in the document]
thejohnfreeman: In the `Stoppable` hierarchy, `onStop` is called top-down, parent-before-children. This change preserves that order. Whether it is the "correct" order I would like to settle in a future changeset, but consider out of scope for this one.

src/ripple/app/main/Application.cpp
```diff
@@ -1110,7 +1064,30 @@ class ApplicationImp : public Application, public RootStoppable, public BasicApp
...
+            shardArchiveHandler_->stop();
+        m_nodeStore->stop();
+        if (shardStore_)
+            shardStore_->stop();
+        if (config_->reporting())
+        {
+#ifdef RIPPLED_REPORTING
+            pgPool_->stop();
+#endif
+            reportingETL_->stop();
```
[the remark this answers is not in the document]
thejohnfreeman: I will trust your judgment on it and make that change. Because it is so difficult for me to test with reporting enabled, can I get you to do that for me?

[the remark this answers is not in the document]
cjcobb23: Yeah I'll test and report back

src/ripple/app/main/Application.cpp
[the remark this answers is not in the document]
cjcobb23: You're right, I let myself get mislead. `onStop()` is called parent down, though `m_stopped` seems to be set to `true` child up.

src/ripple/nodestore/impl/DatabaseShardImp.cpp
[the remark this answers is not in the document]
cjcobb23: You are right, `onStop()` is called top-down, my mistake. While I do think that the correct order would be child up, I agree that would be better addressed in a future PR.

src/ripple/overlay/impl/OverlayImpl.cpp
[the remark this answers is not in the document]
cjcobb23: Got it, thanks.

src/ripple/app/main/Application.cpp
nbougalis: Should we maintain a stack in `ApplicationImpl`, that we can simply `pop` from to get things in the right order?

src/ripple/app/main/Application.cpp
```diff
@@ -1110,7 +1064,30 @@ class ApplicationImp : public Application, public RootStoppable, public BasicApp
                 return validators().trustedPublisher(pubKey);
             });
 
-        stopped();
+        // The order of these stop calls is delicate.
+        // Re-ordering them risks undefined behavior.
```
nbougalis, line 1068: I share @cjcoob23's concern:

>I feel that this situation is exactly the problem that Stoppable was meant to solve (managing the order in which things are stopped); with this PR, now we have to manually maintain this order.

This commit reinforces that. I'm not saying that this cleanup and the explicit documentation of the ordering (instead of the implicit and opaque emergent ordering imposed by `Stoppable`) isn't an improvement.

But this makes it hard to figure out where something new should go, or to even be sure that we properly stop everything.

Nothing actionable; food for thought.

src/ripple/app/main/LoadManager.cpp
```diff
@@ -109,19 +102,22 @@ LoadManager::run()
     beast::setCurrentThreadName("LoadManager");
 
     using namespace std::chrono_literals;
-    using clock_type = std::chrono::system_clock;
+    using clock_type = std::chrono::steady_clock;
```
nbougalis, line 105: I'd like to get @HowardHinnant's sign-off on this particular change and the changes it required further down in the function.

src/ripple/overlay/impl/OverlayImpl.cpp
```diff
@@ -103,25 +105,21 @@ OverlayImpl::Timer::on_timer(error_code ec)
     if ((overlay_.timer_count_ % Tuning::checkIdlePeers) == 0)
         overlay_.deleteIdlePeers();
 
-    timer_.expires_from_now(std::chrono::seconds(1));
-    timer_.async_wait(overlay_.strand_.wrap(std::bind(
-        &Timer::on_timer, shared_from_this(), std::placeholders::_1)));
+    start();
```
nbougalis, line 108: That's a little weird. I'd expect the `start` function to only be called once.

src/ripple/app/main/LoadManager.cpp
HowardHinnant: This looks good to me.  It simplifies things.  The change to `steady_clock` enables the removal of the "time jump" check below which was almost certainly never used anyway.

One nitpicky change that could be made in this function is to drop the `<std::mutex>` from `std::unique_lock` on line 113.  I thought we did a sweep to eliminate these earlier, but I guess we missed this one.

HowardHinnant (approved):

src/ripple/app/ledger/impl/LedgerCleaner.cpp
```diff
@@ -51,8 +51,8 @@ class LedgerCleanerImp : public LedgerCleaner
 
     std::thread thread_;
 
-    enum class State : char { readyToClean = 0, startCleaning, cleaning };
-    State state_ = State::readyToClean;
+    enum class State : char { notCleaning = 0, cleaning };
+    State state_ = State::notCleaning;
```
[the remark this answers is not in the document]
thejohnfreeman: `std::atomic_flag` seems to be missing the predicated `wait` overload, and because the wakeup condition is complex (stopped _or_ cleaning), I don't think we win anything by switching to `std::atomic_flag`.

src/ripple/app/main/Application.cpp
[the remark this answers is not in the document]
cjcobb23: Actually @thejohnfreeman what was the stopping order of these before this PR? To stay in the spirit of the rest of the PR, I'd prefer to preserve the stopping order actually.

src/ripple/app/main/Application.cpp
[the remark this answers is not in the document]
cjcobb23: Got it, thanks for the explanation. I was mistaken about the stopping order of the various `Stoppable`s. I do think RAII would be best, and that what you have here is clearer than what's currently in `develop`.

src/ripple/app/main/Application.cpp
thejohnfreeman: Here's my idea:

The new way is RAII. (This is actually the 30 years old way, but it sadly hasn't seemed to have caught on yet.) Start your object in its constructor, and stop it in its destructor. Or if you insist on having `start` and `stop` methods, either (1) have `start` return an object that calls `stop` in its destructor or (2) use a separate object, a la `std::lock_guard`, that calls `start` in its constructor and `stop` in its destructor, and never call `start` or `stop` yourself. If you need to dynamically end a lifetime early, own the RAII object via `std::unique_ptr` or `std::optional` (depending on how you want to manage storage) and clear it when you want to stop.

Most objects depend on objects that do not depend on it (no circular dependencies). They should be started after their dependencies and stopped before them. This matches perfectly with the behavior of RAII.

I'd like to make these changes in our codebase, but I expect they will be very large changes. I'd like to start with this changeset as the new baseline.

If it comforts you to know, the order isn't as delicate as the comment makes it seem; I just don't want people touching it. Add your new stoppable objects around it. Start yours after everything else in `ApplicationImp::start`, and stop it before everything else in `ApplicationImp::stop`. If we assume that this order properly stops everything, then changes like that will preserve that invariant.

To prove that this order is correct, I need to demarcate the dependency edges between the components. Part of why this is so difficult is that everyone likes to pull in the kitchen sink (`Application`) even if they only need one or two dependencies. Another part is that we like to build huge monolithic classes with 15 responsibilities and 10 states, with some dependencies active in only a subset of those responsibilities and states. I'd like to get to the point where we know the exact dependencies of each component. That will make it easy to know the correct start and stop sequences (among other benefits).

src/ripple/overlay/impl/OverlayImpl.cpp
thejohnfreeman: Think of it this way and let me know if that changes your expectation:

`Timer` is a state machine. `start` is a transition that enters the "waiting" state. It doesn't care what state it's coming from. It can be called from any state, including the "just constructed" state and the "notified" state.

src/ripple/app/main/Application.cpp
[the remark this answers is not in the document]
thejohnfreeman: That's difficult to answer because (a) do you want to consider the order objects call `stopped()` (which can be non-deterministic) or (b) do you want to consider the order that `RootStoppable::stop` blocked on those calls or (c) do you want to consider the order that `Stoppable`s first learned that they needed to stop (which can be non-deterministic)?

pwang200:
The commit "[fold] LedgerReplayer is not a Stoppable e8a8d3e" itself is good. I will review other parts especially the ordering of stopping different components later.

src/ripple/app/main/Application.cpp
```diff
@@ -1714,17 +1689,33 @@ ApplicationImp::setup()
...
+    {
+        setSweepTimer();
+        setEntropyTimer();
+    }
+
+    m_io_latency_sampler.start();
+    m_resolver->start();
+    m_loadManager->start();
+    m_shaMapStore->start();
+    overlay_->start();
```
cjcobb23, line 1715: `overlay_` should not be started when running in reporting mode.

cjcobb23 (changes requested):

src/ripple/app/main/Application.cpp
cjcobb23: `overlay_` is actually never even instantiated and is just `nullptr` when running in reporting mode.

scottschurr (commented):
Nice work!  Untwisting the contents of `Stoppable` must have been quite a task.  I think what you have left us with is much easier to reason about.

There are a few places in this pull request where you and I disagree on implementation details.  I'm happy to discuss those points with you.

src/ripple/app/main/Application.cpp
```diff
@@ -46,8 +46,8 @@
 #include <ripple/app/reporting/ReportingETL.h>
 #include <ripple/app/tx/apply.h>
 #include <ripple/basics/ByteUtilities.h>
-#include <ripple/basics/PerfLog.h>
 #include <ripple/basics/ResolverAsio.h>
+#include <ripple/basics/impl/PerfLogImp.h>
```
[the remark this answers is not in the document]
thejohnfreeman: To keep the `make_PerfLog` factory function which existed before this change, I'd have to move `start` and `stop` into the interface of `PerfLog`. I'd rather handle start and stop business in constructors and destructors, which are part of every class interface, but doing that for `PerfLog` would cascade to the construction of `JobQueue`, and on down the line. Those are changes I would like to make, but not in this changeset. I struck the balance of referencing the implementation as a member, but I can settle for switching back to an abstract interface and just adding the `start` and `stop` methods to it.

src/ripple/app/main/Application.cpp
```diff
@@ -270,13 +270,12 @@ class ApplicationImp : public Application, public RootStoppable, public BasicApp
...
         // PerfLog must be started before any other threads are launched.
-        , perfLog_(perf::make_PerfLog(
+        , perfLog_(perf::make_PerfLogImp(
               perf::setup_PerfLog(
                   config_->section("perf"),
                   config_->CONFIG_DIR),
-              *this,
               logs_->journal("PerfLog"),
-              [this]() { signalStop(); }))
+              [&] { signalStop(); }))
```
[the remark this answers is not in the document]
thejohnfreeman: I was under the impression that `[&]` was the only way to enable the "frame pointer" optimization for closures, but that may be incorrect, and it isn't safe anyway in this case (like it is in the `std::condition_variable::wait` cases). I prefer explicit capture as well; I didn't make these changes for my writing convenience. I've changed them back to `[this]`.

src/ripple/app/main/Application.cpp
```diff
@@ -1714,7 +1713,12 @@ ApplicationImp::run()
     JLOG(m_journal.info()) << "Received shutdown request";
     stop(m_journal);
     if (config_->reporting())
```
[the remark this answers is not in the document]
thejohnfreeman: I want to make the calls to start and stop symmetrical so that the change to RAII will be straightforward and easy to review. I want to give `stop` methods the pre-condition that the object is started, just like `std::thread::join` assumes the thread is active.

src/ripple/nodestore/impl/Database.cpp
```diff
@@ -57,20 +54,21 @@ Database::~Database()
     // crash during shutdown when its members are accessed by one of
     // these threads after the derived class is destroyed but before
     // this base class is destroyed.
-    stopReadThreads();
+    stop();
```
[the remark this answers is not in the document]
thejohnfreeman: You're right. I think I just made a mistake with this.

src/ripple/nodestore/impl/Database.cpp
```diff
@@ -57,20 +54,21 @@ Database::~Database()
     // crash during shutdown when its members are accessed by one of
```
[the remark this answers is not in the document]
scottschurr: Sorry, I didn't leave a very clear comment last time.  The changes you made in response to the comment were good.  But all I really intended to point out is that the method named `stopReadThreads()` no longer exists, but there is a comment that references it.

I suggest you do a grep for `stopReadThreads()` in your branch and fix up any references so they just say `Database::stop()`.  I think you'll only find one reference, and that's on line 52 of Database.cpp.

scottschurr (approved):
Looks great to me.  Thanks for taking this on!  I spotted one place where a comment should be patched up, but I won't hold up approval based on that.  Nice work, and thanks again!

src/ripple/app/main/Application.cpp
thejohnfreeman: I dug through the remaining stoppable objects (everything in this block that has its `stop` method called) to try to figure out the dependency graph.

1. Each object is a singleton of its type, so we can represent them by their types. The exceptions are both `NodeStore::Database`, but the second database is always a `NodeStore::DatabaseShard`, so we can distinguish them with those types.
2. Objects tend to depend on each other by going through a getter method in `Application`.
3. Two classes, `NetworkOPs` and `LedgerMaster` are special. They depend on many objects that depend on them, creating cycles, but they are treated like low-level abstractions. The separation between `Application` and `ApplicationImp` is what lets us tie the recursive knot.
4. `JobQueue` is very low level in the hierarchy, but it stops early because some other objects only stop when they cannot schedule any more jobs. Objects that depend on `JobQueue` will not break if it stops before they do.

In the hierarchy below, with the exception of `JobQueue` and `NetworkOPs`, no object depends on another object at the same or higher level... to my best belief. Some stoppables depend on other stoppables indirectly, and while I tried to chase all those connections down, there are many classes and LOC involved here; I may have missed something.

Level 0:
- `PerfLog`
- `NetworkOPs`
- `LedgerMaster`
- `NodeStore::Database`
- `PgPool`

Level 1:
- `JobQueue`
- `SHAMapStore`
- `Overlay`

Level 2:
- `LoadManager`
- `GRPCServer`
- `InboundTransactions`
- `NodeStore::DatabaseShard`

Level 3:
- `RPC::ShardArchiveHandler`
- `InboundLedgers`

Level 4:
- `LedgerReplayer`
- `LedgerCleaner`

thejohnfreeman: Further, `LedgerMaster` is only a "stoppable" because it owns `LedgerCleaner`. The connection between them is very weak, and I plan to move `LedgerCleaner` to be a member of `Application` directly. No stoppable depends on `LedgerCleaner`.

src/ripple/rpc/impl/ServerHandlerImp.cpp
```diff
@@ -138,9 +137,13 @@ ServerHandlerImp::setup(Setup const& setup, beast::Journal journal)
 //------------------------------------------------------------------------------
 
 void
-ServerHandlerImp::onStop()
+ServerHandlerImp::stop()
```
carlhua, line 140: Is this function ever invoked in the source base? If not, is it intentional?

thejohnfreeman: `Stoppable::onStop` was the first of two stop callbacks in the call tree rooted at `RootStoppable::stop`. It was always called at shutdown.

src/ripple/app/main/LoadManager.cpp
```diff
@@ -89,13 +88,15 @@ LoadManager::onStart()
 void
 LoadManager::onStop()
 {
+    {
+        std::lock_guard<std::mutex> sl(mutex_);
```
miguelportilla, line 85: nit: consider Class template argument deduction. eg. ```std::lock_guard sl(mutex_);```

src/ripple/app/ledger/impl/LedgerCleaner.cpp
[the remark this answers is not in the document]
miguelportilla: I like the simpler ```bool cleaning_{false};```

src/ripple/app/ledger/impl/LedgerCleaner.cpp
```diff
@@ -108,6 +108,7 @@ class LedgerCleanerImp : public LedgerCleaner
             wakeup_.notify_one();
         }
         thread_.join();
+        stopped();
```
miguelportilla: Consider
```
    if (thread_.joinable())
        thread_.join();
```

src/ripple/app/main/Application.cpp
```diff
@@ -288,7 +289,22 @@ class ApplicationImp : public Application, public RootStoppable, public BasicApp
                                    : nullptr)
 #endif
 
-        , m_nodeStoreScheduler(*this)
+        , m_collectorManager(CollectorManager::New(
```
miguelportilla, line 287: Nit: consider renaming ```CollectorManager::New``` to ```make_CollectorManager```

src/ripple/app/main/Application.cpp
```diff
@@ -476,9 +479,6 @@ class ApplicationImp : public Application, public RootStoppable, public BasicApp
         //  started in this constructor.
         //
 
-        // VFALCO HACK
-        m_nodeStoreScheduler.setJobQueue(*m_jobQueue);
-
```
miguelportilla, line 490: Nice!

src/ripple/app/main/NodeStoreScheduler.h
```diff
@@ -52,11 +40,7 @@ class NodeStoreScheduler : public NodeStore::Scheduler, public Stoppable
     onBatchWrite(NodeStore::BatchWriteReport const& report) override;
 
 private:
-    void
-    doTask(NodeStore::Task& task);
-
-    JobQueue* m_jobQueue{nullptr};
-    std::atomic<int> m_taskCount{0};
+    JobQueue* m_jobQueue;
```
miguelportilla, line 43: Consider making ```m_jobQueue``` a reference.

src/ripple/rpc/impl/ServerHandlerImp.cpp
```diff
@@ -141,6 +142,11 @@ void
 ServerHandlerImp::onStop()
 {
     m_server->close();
+    {
+        std::unique_lock<std::mutex> lock(mutex_);
+        condition_.wait(lock, [this]() { return stopped_; });
```
miguelportilla: Nit: I'll suggest using similar naming across classes on variables relating to stopping. ```cv_, wakeup_, condition_``` could be changed to ```_condition```.  ```isTimeToStop, stop_, stopping_``` could be changed to ```stop_``` etc...

src/ripple/rpc/impl/ServerHandlerImp.cpp
```diff
@@ -150,11 +156,14 @@ ServerHandlerImp::onAccept(
...
+    auto const& port = session.port();
 
-    auto const c = ++count_[session.port()];
+    auto const c = [this, &port]() {
+        std::lock_guard lock(mutex_);
+        return ++count_[port];
+    }();
 
-    if (session.port().limit && c >= session.port().limit)
+    if (port.limit && c >= port.limit)
```
miguelportilla, line 163: Log message can be changed from session.port().name to port.name()

src/ripple/nodestore/impl/DatabaseRotatingImp.h
```diff
@@ -34,21 +34,13 @@ class DatabaseRotatingImp : public DatabaseRotating
...
-        Stoppable& parent,
         std::shared_ptr<Backend> writableBackend,
         std::shared_ptr<Backend> archiveBackend,
         Section const& config,
         beast::Journal j);
 
-    ~DatabaseRotatingImp() override
-    {
-        // Stop read threads in base before data members are destroyed
-        stopReadThreads();
```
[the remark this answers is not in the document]
miguelportilla: @scottschurr  I believe with the new top-down approach, since the Application object owns the nodestore, it must call the nodestore stop() after its stop() is called.

src/ripple/nodestore/impl/DatabaseShardImp.cpp
```diff
@@ -693,32 +693,25 @@ DatabaseShardImp::stop()
...
+    for (auto const& wptr : shards)
     {
-        if (!e.expired())
+        if (auto const shard{wptr.lock()})
         {
-            std::string shardIndex;
-            if (auto const shard{e.lock()}; shard)
-                shardIndex = std::to_string(shard->index());
-
+            std::string shardIndex = std::to_string(shard->index());
```
miguelportilla, line 714: Consider...
```
        if (auto const shard{wptr.lock()})
        {
            JLOG(j_.warn()) << " shard " << shard->index() << " unexpired";
        }
```

carlhua:
One thing that I would like to see in this PR is to instrument the code prior to the changes to show/determine the ordering of starting and stopping of the new code is correct. We should verify the call sequence is correct - though I understand there are some tricky conditions (non-deterministic of the ordering of when services are actually stopped) - we should still review the call ordering and make sure the **NEW** ordering is correct.

thejohnfreeman:
@carlhua What does correct mean? Please see [this comment](https://github.com/ripple/rippled/pull/3741#discussion_r564778658) and [this comment](https://github.com/ripple/rippled/pull/3741#discussion_r568760856). I've tested many times with the start-sync-stop test with no failures, but it is certainly difficult to be sure I've tested all possible configurations. I don't think the previous stop sequence is going to tell us much about what must be stopped after what. That will only be possible through careful review of the code.

carlhua:
> @carlhua What does correct mean? Please see [this comment](https://github.com/ripple/rippled/pull/3741#discussion_r564778658) and [this comment](https://github.com/ripple/rippled/pull/3741#discussion_r568760856). I've tested many times with the start-sync-stop test with no failures, but it is certainly difficult to be sure I've tested all possible configurations. I don't think the previous stop sequence is going to tell us much about what must be stopped after what. That will only be possible through careful review of the code.

Consider the following:
1. Instrument the code to log to file when a service is started/stopped.
2. Run through the sequence of start and stop (potentially multiple times)
3. Inspect the log with what you have provided to make sure there is no surprise.

This is more of a sanity check to make sure what you have dug up is indeed correct.

thejohnfreeman:
Here was the stop sequence with `Stoppable`:


Order | Class | Callback
---|---|---
1 | ApplicationImp | [onStop][80]
2 | LoadManager | [onStop][88]
3 | SHAMapStoreImp | [onStop][102]
4 | RPC::ShardArchiveHandler | [onStop][108]
5 | OverlayImpl | [onStop][85]
6 | PeerFinder::ManagerImp | [onStop][82]
7 | NodeStore::DatabaseShardImp | [onStop][96]
8 | NodeStore::TaskQueue | [onStop][93]
9 | GRPCServer | [onStop][100]
10 | NetworkOPsImp | [onStop][104]
11 | ServerHandlerImp | [onStop][108]
12 | InboundTransactionsImp | [onStop][89]
13 | InboundLedgersImp | [onStop][90]
14 | LedgerCleanerImp | [onStop][92]
15 | NodeStore::DatabaseRotatingImp | [onStop][94]
16 | PerfLogImp | [onStop][106]
17 | OverlayImpl | [onChildrenStopped][86]
18 | NodeStore::DatabaseShardImp | [onChildrenStopped][97]
19 | NodeStore::DatabaseRotatingImp | [onChildrenStopped][95]
20 | JobQueue | [onChildrenStopped][98]
21 | NodeStoreScheduler | [onChildrenStopped][101]
22 | PerfLogImp | [onChlidrenStopped][107]


[80]: https://github.com/ripple/rippled/blob/ccb4ab0665f09a60096065a985b491f183c74efe/src/ripple/app/main/Application.cpp#L999-L1064
[81]: https://github.com/ripple/rippled/blob/ccb4ab0665f09a60096065a985b491f183c74efe/src/ripple/peerfinder/impl/PeerfinderManager.cpp#L223-L228
[82]: https://github.com/ripple/rippled/blob/ccb4ab0665f09a60096065a985b491f183c74efe/src/ripple/peerfinder/impl/PeerfinderManager.cpp#L230-L235
[83]: https://github.com/ripple/rippled/blob/ccb4ab0665f09a60096065a985b491f183c74efe/src/ripple/overlay/impl/OverlayImpl.cpp#L496-L569
[84]: https://github.com/ripple/rippled/blob/ccb4ab0665f09a60096065a985b491f183c74efe/src/ripple/overlay/impl/OverlayImpl.cpp#L571-L579
[85]: https://github.com/ripple/rippled/blob/ccb4ab0665f09a60096065a985b491f183c74efe/src/ripple/overlay/impl/OverlayImpl.cpp#L581-L585
[86]: https://github.com/ripple/rippled/blob/ccb4ab0665f09a60096065a985b491f183c74efe/src/ripple/overlay/impl/OverlayImpl.cpp#L587-L592
[87]: https://github.com/ripple/rippled/blob/ccb4ab0665f09a60096065a985b491f183c74efe/src/ripple/app/main/LoadManager.cpp#L80-L87
[88]: https://github.com/ripple/rippled/blob/ccb4ab0665f09a60096065a985b491f183c74efe/src/ripple/app/main/LoadManager.cpp#L89-L102
[89]: https://github.com/ripple/rippled/blob/ccb4ab0665f09a60096065a985b491f183c74efe/src/ripple/app/ledger/impl/InboundTransactions.cpp#L241-L249
[90]: https://github.com/ripple/rippled/blob/ccb4ab0665f09a60096065a985b491f183c74efe/src/ripple/app/ledger/impl/InboundLedgers.cpp#L384-L393
[91]: https://github.com/ripple/rippled/blob/ccb4ab0665f09a60096065a985b491f183c74efe/src/ripple/app/ledger/impl/LedgerCleaner.cpp#L95-L99
[92]: https://github.com/ripple/rippled/blob/ccb4ab0665f09a60096065a985b491f183c74efe/src/ripple/app/ledger/impl/LedgerCleaner.cpp#L101-L111
[93]: https://github.com/ripple/rippled/blob/ccb4ab0665f09a60096065a985b491f183c74efe/src/ripple/nodestore/impl/TaskQueue.cpp#L33-L38
[94]: https://github.com/ripple/rippled/blob/ccb4ab0665f09a60096065a985b491f183c74efe/src/ripple/nodestore/impl/Database.cpp#L60-L66
[95]: https://github.com/ripple/rippled/blob/ccb4ab0665f09a60096065a985b491f183c74efe/src/ripple/nodestore/impl/Database.cpp#L68-L72
[96]: https://github.com/ripple/rippled/blob/ccb4ab0665f09a60096065a985b491f183c74efe/src/ripple/nodestore/impl/DatabaseShardImp.cpp#L686-L697
[97]: https://github.com/ripple/rippled/blob/ccb4ab0665f09a60096065a985b491f183c74efe/src/ripple/nodestore/impl/DatabaseShardImp.cpp#L699-L726
[98]: https://github.com/ripple/rippled/blob/ccb4ab0665f09a60096065a985b491f183c74efe/src/ripple/core/impl/JobQueue.cpp#L460-L465
[99]: https://github.com/ripple/rippled/blob/ccb4ab0665f09a60096065a985b491f183c74efe/src/ripple/app/main/GRPCServer.cpp#L441-L453
[100]: https://github.com/ripple/rippled/blob/ccb4ab0665f09a60096065a985b491f183c74efe/src/ripple/app/main/GRPCServer.cpp#L455-L466
[101]: https://github.com/ripple/rippled/blob/ccb4ab0665f09a60096065a985b491f183c74efe/src/ripple/app/main/NodeStoreScheduler.cpp#L41-L46
[102]: https://github.com/ripple/rippled/blob/ccb4ab0665f09a60096065a985b491f183c74efe/src/ripple/app/misc/SHAMapStoreImp.cpp#L731-L750
[103]: https://github.com/ripple/rippled/blob/ccb4ab0665f09a60096065a985b491f183c74efe/src/ripple/app/misc/SHAMapStoreImp.h#L254-L259
[104]: https://github.com/ripple/rippled/blob/ccb4ab0665f09a60096065a985b491f183c74efe/src/ripple/app/misc/NetworkOPs.cpp#L640-L667
[105]: https://github.com/ripple/rippled/blob/ccb4ab0665f09a60096065a985b491f183c74efe/src/ripple/basics/impl/PerfLogImp.cpp#L461-L466
[106]: https://github.com/ripple/rippled/blob/ccb4ab0665f09a60096065a985b491f183c74efe/src/ripple/basics/impl/PerfLogImp.cpp#L468-L480
[107]: https://github.com/ripple/rippled/blob/ccb4ab0665f09a60096065a985b491f183c74efe/src/ripple/basics/impl/PerfLogImp.cpp#L482-L486
[108]: https://github.com/ripple/rippled/blob/ccb4ab0665f09a60096065a985b491f183c74efe/src/ripple/rpc/impl/ServerHandlerImp.cpp#L140-L144

thejohnfreeman:
I've just pushed a revised stop sequence. It copies the order that `Stoppable::onStop` was called, as laid out in [my previous comment](https://github.com/ripple/rippled/pull/3741#issuecomment-773532851). Note:

- `NodeStoreScheduler` is no longer a stoppable because, I argue, it never should have been.
- `LedgerCleanerImp` is owned by `LedgerMaster`, which makes `LedgerMaster` responsible for stopping it. Simillarly, `TaskQueue` is owned by `NodeStore::DatabaseShardImp`, and `PeerFinder::ManagerImp` is owned by `OverlayImpl`.

I believe this order with the new design closely reflects the old order because:

- An object should be ready to stop when its `onStop` is called.
- Most objects stopped before returning from `onStop` as indicated by their call of `stopped()`.
- Most objects that did not called `stopped()` from `onStop` could still possibly stop in `onStop`. Any object B whose `onStop` is called after that of A cannot assume that A has not stopped.

In my testing, most of this order is actually quite flexible, except that `JobQueue` cannot be stopped before `SHAMapStoreImp` (it reliably ends in a segfault), because `SHAMapStoreImp::run` indirectly calls `JobQueue::addLoadEvents`.

Just to keep everyone in the loop, I'm still planning to incorporate suggestions from @miguelportilla and to move `LedgerCleanerImp` out of `LedgerMaster`.

src/ripple/app/ledger/impl/LedgerCleaner.cpp
thejohnfreeman: We're assuming that `LedgerCleaner::stop` is called exactly once. I want `stop` methods to be able to assume that. It will make the transition to RAII much easier.

src/ripple/rpc/impl/ServerHandlerImp.cpp
thejohnfreeman: I want that too, eventually, but I want to do it separately so that the two changesets are easy to review: every change in this one is meaningful, there is no clutter of noisy renames, while every change in that one will be just a rename and easy to skim through.

cjcobb23 (approved):
I tested these changes with reporting mode and everything seems to be working fine. Tested both ETL and read-only mode.

src/ripple/app/main/Application.cpp
pwang200: We have a new component `LedgerReplayer`. So I have been thinking the rules of inserting a new component `x` to the list. I'd like to confirm that `x->stop();` can be a blocking call and can call `join()` if `xyz` has threads. And when `x->stop();` returns, if `x` can guarantee that (1) threads are joined if any, (2) will treat calls from other components as no-ops (including callbacks such as a timeout), and to be safe (3) will not call into other components, then `x`'s position in the list does not matter. Right?

thejohnfreeman: After `stop` has returned, the object must be in a "stopped" state. A stopped object can be safely destroyed without waiting for anything else. (Short answer: yes it can block, and yes it can join threads.)

A stopped object is not required to support any operations other than destruction. It may choose to allow other operations on a case-by-case basis, and it may choose to implement them as no-ops, but neither is required. Other objects that want to interact with a stopped object must ensure that the stopped object allows those operations while stopped. Let us call these operations "safe operations", and operations that may only be called while the object is in the started state "unsafe operations".

If stopped object B depends on stopped object A, then B must not call any unsafe operations on A after B has stopped.

To answer your last question, we need to add one more requirement: because `x` may call unsafe operations _before_ `x->stop()` has returned, `x->stop()` must be called _before_ any of its dependencies with unsafe operations are stopped. `x` can always be safely placed first in the stop sequence, but cannot be placed after any of its dependencies on whom it calls unsafe operations.

pwang200: Thanks for the explanation. I agree. My concern is that sometimes circular dependencies and safe operations vs unsafe operations are that obvious. Taking a step back, `LedgerMaster` uses `LedgerReplayer` (for the publish use case), and `LedgerReplayer` uses `LedgerMaster` (to check if some data is available locally). `LedgerReplayer` uses `Peer` to send requests, and `Peer` uses `LedgerReplayer` to process the responses. Are those circular dependencies? It was not very clear to me. So we made all the calls and callbacks to `LedgerReplayer` "stop-safe".
I think the developer of a new component should make an effort to make all the interaction with other components (the service it provides and the callbacks from the services it has requested) "stop-safe". The few exceptions left will sort the stop list.

thejohnfreeman: `LedgerMaster` should never have been a `Stoppable`, and it is no longer a stoppable object after this changeset. `PeerImp` has never been a stoppable object. Thus, in the two examples you gave, there are no circular stoppable dependencies.

If there ever does arise a circular dependency, only one object in the cycle needs to have safe operations to permit a definitively safe stop sequence. I think we'll find in practice that most stoppable objects don't need to worry about it.

miguelportilla (approved):
👍

thejohnfreeman:
Please, @cjcobb23 and @cdy20 can you review the [merge with 1.8.0-b2](https://github.com/ripple/rippled/pull/3741/files/d3c1c1ff79eb8821ec77f7c0821a25955d767be3..145f5a177ff4bf5f08070ec53f55dedb6b056603)? It involves the `RelationalDBInterfacePostgres` and `PgPool` and reporting mode conditional compilation.

cdy20 (approved):

src/ripple/app/main/Application.cpp
```diff
@@ -1033,7 +1021,9 @@ class ApplicationImp : public Application, public BasicApp
         {
             reportingETL_->stop();
 #ifdef RIPPLED_REPORTING
```
cjcobb23, line 1023: I don't think you actually need the `ifdef` here. `RelationalDBInterfacePostgres` is not conditionally compiled. It's just `PgPool` that is conditionally compiled (along with the other classes in Pg.h).

cjcobb23 (approved):
Left one comment about the merge but I'm good with the code as is.

src/ripple/app/misc/SHAMapStoreImp.h
```diff
@@ -135,12 +135,6 @@ class SHAMapStoreImp : public Stoppable, public SHAMapStore
         NodeStore::Scheduler& scheduler,
         beast::Journal journal);
 
-    ~SHAMapStoreImp()
-    {
-        if (thread_.joinable())
-            thread_.join();
-    }
```
ximinez, line 135: If I understand correctly, if `onStop` isn't called, then the `thread_` destructor will throw when `SHAMapStoreImp` is destroyed. I could see where that is intentional if you want to ensure that `onStop` is always called, but I wanted to confirm it's intentional.

ximinez (changes requested):

src/ripple/app/misc/SHAMapStoreImp.h
thejohnfreeman: Yes, it's intentional. `std::thread` is the model for the new design. This guarantees that the stop sequence for `ApplicationImp` is the same every time, because `ApplicationImp::stop` _must_ be called.

src/ripple/app/ledger/LedgerMaster.h
```diff
@@ -31,6 +31,7 @@
 #include <ripple/app/misc/CanonicalTXSet.h>
 #include <ripple/basics/RangeSet.h>
 #include <ripple/basics/StringUtilities.h>
+#include <ripple/basics/UptimeClock.h>
```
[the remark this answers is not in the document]
thejohnfreeman: `UptimeClock` is referenced on line 287. It may be included transitively, but not by any other headers in the same "module" (`basics`).

ximinez (changes requested):
Another incremental review. Nothing really significant.

ximinez (changes requested):

src/ripple/app/main/Application.cpp
```diff
@@ -1035,7 +999,33 @@ class ApplicationImp : public Application, public RootStoppable, public BasicApp
...
+        m_inboundTransactions->stop();
+        m_inboundLedgers->stop();
+        ledgerCleaner_->stop();
+        if (config_->reporting())
+        {
+            reportingETL_->stop();
+            dynamic_cast<RelationalDBInterfacePostgres*>(
+                &*mRelationalDBInterface)
+                ->stop();
+        }
```
ximinez, line 1026: I've been thinking about some of the earlier discussion about this `config->reporting()` check vs. the other places where the ptr is just checked for null. I think it can work and be clear with some simple changes. Here's a commit: https://github.com/ximinez/rippled/commit/8d963bad88f2af783a82f75aa36d428ea407f828

ximinez (changes requested):
I'm finally done going through all of the commits.

ximinez (approved):
Nice work!

PR #3747, opened 2021-01-31
Simplify SHAMapItem construction

nbougalis:
### High Level Overview of Change

The existing class offered several constructors which were mostly unnecessary. This commit eliminates all existing constructors and introduces a single new one, taking a `Slice`.

The commit also changes the internal data buffer from `std::vector` to `Buffer`, saving 8 bytes per instance in the process without any loss of functionality.

### Type of Change

- [X] Improvement (non-breaking change which does not add functionality)


src/ripple/basics/Buffer.h
```diff
@@ -165,53 +173,59 @@ class Buffer
     void
     clear() noexcept
     {
-        p_.reset();
+        delete[] p_;
```
seelabs, line 176: Need to set `p_ = nullptr` here or we can double delete if `clear` is called twice.

seelabs (changes requested):
I just started a review and noticed one of the unit tests was failing. If we fix `Buffer::clear` (see comment below) the unit tests pass.

seelabs (commented):
I like the clean-up to this interface. 

The advanages are:

* Save 8 bytes per `SHAMapItem` (commit message says 16, but I think it's 8)
* Simplifies construction of `SHAMapItem`

The disadvantage is:
* We removed `SHAMapItem` constructors that used to move data with constructors that always allocate and copy data. And there are several places in the code that took advantage of these moves.

I think our options are:

1. replace std::vector in serializer with something we can steal the buffer from
2. replace Buffer with std::vector so we can move other vectors in, and live with the extra 8 bytes per `SHAMapItem`  (what the old code did)
3. live with the extra allocations and copies. (what the new code does)

I think I also weakly prefer `std::unique_ptr` to the raw pointer in `Buffer`. There's no space savings, and I prefer code like:
```
    p_ = std::move(other.p_);
```
to
```
    p_ = other.p_;
    other.p_ = nullptr;
```
It make ownership rules clearer. But given that this class is a `Buffer`, not a `Slice` that's not a strong argument, so I only _weakly_ prefer it. Fine as-is too.

src/ripple/basics/Buffer.h
nbougalis: Oops!

nbougalis:
> I think our options are:
> 
>     1. replace std::vector in serializer with something we can steal the buffer from
> 
>     2. replace Buffer with std::vector so we can move other vectors in, and live with the extra 8 bytes per `SHAMapItem`  (what the old code did)
> 
>     3. live with the extra allocations and copies. (what the new code does)

I definitely think that exploring our options here is a good idea; obviously, the fewer memory allocations we do, the better of we are. However, I think that _before_ we can come up with a clean and comprehensive solution, we are better off simplifying the code. Having to contend with and reason about 4 constructors, two kinds of serializer / deserializers, opaque type aliases (`using Blob = std::vector<...>`) and more is a recipe for lost efficiencies.

I believe this commit does simplify things and more clearly exposes the issue, so that we can tackle it.



> I think I also weakly prefer `std::unique_ptr` to the raw pointer in `Buffer`. There's no space savings, and I prefer code like:
> 
> ```
>     p_ = std::move(other.p_);
> ```
> 
> to
> 
> ```
>     p_ = other.p_;
>     other.p_ = nullptr;
> ```

Agreed. I don't really know what I was thinking with that commit. It was ill-advised and I've removed it from this PR.

HowardHinnant (approved):
I like this simpler code.  We may or may not further optimize this, but I think this is a solid improvement.

seelabs (approved):
:+1: The cleaned-up interface is much nicer. I wish we had a benchmark that showed the performance hit from the new memory allocations. Without that it's hard to see what the trade-off is between the new cleaner interface and the old interface that moves memory. However, without such a benchmark, my gut says it's not too bad. And re-allocating and moving the memory will have some memory savings (it won't keep the over-allocated vector memory around). So I think we should get this in.

I made an issue for looking into re-enabling the memory moves, so we don't lose track: https://github.com/ripple/rippled/issues/3761

cjcobb23 (approved):
Nice cleanup. I am concerned about the extra copying introduced here. But we save an extra 8 bytes, which is nice. We should run some basic performance tests to see the tradeoffs. It's probably not that bad, but we should confirm that we don't see a big regression with this change. I'm ok merging this as is, and running those performance tests later, sometime before this goes into a major release. Given that we are cutting 1.7 right now, we have quite some time to performance test this.

PR #3751, opened 2021-02-03
Use libsecp256k1 instead of OpenSSL for key derivation:

nbougalis:
### High Level Overview of Change

While most of the code associated with secp256k1 operations had been migrated to [libsecp256k1](https://github.com/bitcoin-core/secp256k1), the deterministic key derivation code was still using calls to OpenSSL.

If merged, this commit replaces the OpenSSL-based routines with new libsecp256k1-based implementations. No functional change is expected and the change should be transparent.

This commit also removes several support classes and utility functions that wrapped or adapted various OpenSSL types that
are no longer needed.

A tip of the hat to the original author of this truly superb library, Dr. Pieter Wuille (@sipa), and to all other contributors.

#### Type of Change
- [X] Non-breaking change (fix or feature that would cause existing functionality to not work as expected)


nbougalis:
(The levelization checks are expected to fail: a [FOLD] commit will be added to address this)

src/ripple/protocol/impl/SecretKey.cpp
```diff
@@ -54,45 +53,160 @@ SecretKey::to_string() const
...
-        std::memcpy(ui.data(), seed.data(), seed.size());
-        auto gsk = generatePrivateDeterministicKey(gen_, ui, ordinal);
-        auto gpk = generatePublicDeterministicKey(gen_, ordinal);
-        SecretKey const sk(Slice{gsk.data(), gsk.size()});
-        PublicKey const pk(Slice{gpk.data(), gpk.size()});
-        secure_erase(ui.data(), ui.size());
-        secure_erase(gsk.data(), gsk.size());
-        return {pk, sk};
+        // Generates Nth secret key:
+        auto gsk = [this, tweak = calculateTweak(ordinal)]() {
```
seelabs, line 191: The old code recalcuated this tweak twice. Nice saving here.

src/ripple/protocol/impl/SecretKey.cpp
```diff
@@ -54,45 +53,160 @@ SecretKey::to_string() const
...
+            auto rpk = root_;
+
+            if (secp256k1_ec_privkey_tweak_add(
+                    secp256k1Context(), rpk.data(), tweak.data()) == 1)
```
seelabs, line 195: The old code did not check for zero, and this new code does. Of course, this is an *extremely* unlikely case to hit. I'm just pointing out a functional difference between the old code and the new.

src/ripple/protocol/impl/SecretKey.cpp
```diff
@@ -54,45 +53,160 @@ SecretKey::to_string() const
...
+            {
+                SecretKey sk{Slice{rpk.data(), rpk.size()}};
+                secure_erase(rpk.data(), rpk.size());
+                return sk;
+            }
+
+            LogicError("Unable to add a tweak!");
+        }();
+
```
seelabs, line 204: The old code that the ability to generate public keys without the private key. I believe this was never used, so we may as well lose the functionality. (but again, just pointing out a difference between the old and new implementation).

seelabs (approved):
:+1: Replaced a bunch of code with a much simpler implementation. Love it. I left a couple comments pointing out the difference between old and new code, but nothing that needs to be addressed.

I also wrote a bash script that spun of two rippleds - one with this fix and one without - and generated 10000 keys with each implementation and made sure both implementations would regenerate the same wallet.

src/ripple/protocol/impl/SecretKey.cpp
nbougalis: You're right: this is a difference between the old and new code. Sharp eyes. The odds of this happening (as you point out) are, practically, 0 and I don't think it's particularly helpful to clutter the code up checking for this, but if the consensus is that this is important enough, I'm happy to make the change.

src/ripple/protocol/impl/SecretKey.cpp
nbougalis: Correct: the functionality was never used. It may be worth implementing that mechanism too for reference, but I hesitate to write code (especially cryptographic code) that isn't used. And if I'm being honest, I'd rather remove as much of this code from `rippled` as possible.

This kind of thing should be handled by a separate tool that can generate/derive keys.

PR #3752, opened 2021-02-03
Add instructions for resolving validation failures:

ximinez:

The non-build ~CI jobs~ Github Actions are not always clear about what is going on or how to fix their issues when they fail. These changes, if merged, will put instructions at the site of the failure, which should ease resolving those issues.

* Add instructions to the workflow at the point where the failure would
  occur, which is where someone experiencing a failure is most likely to
  look. To keep things simple, the instructions are always printed. The
  assumption is that if the job succeeds, nobody is likely to look
  anyway.
* Provides the diff of the failure as an artifact, so the user can apply
  it directly to their repo.
* Also update the levelization/README.md to clarify the levels a little
  bit.

### Type of Change

- [X] Documentation Updates

## Test Plan

These changes can be verified by viewing the Github Actions results for this PR (levelization and clang-format). Drill down into the job, then to the "Check for differences" step. The instructions will be displayed.

To test that they show as expected for failures, clone this branch to your own repo, make some changes to the levelization files (`Builds/levelization/results/*.txt`) and/or make changes to a source code file that messes up the formatting. Then commit and push to your own repo. The relevant jobs should fail, and clicking through the errors should lead to the instructions.

---
This change is [Reviewable](https://reviewable.io/reviews/ripple/rippled/3752)

### DIFF
--- a/.github/workflows/clang-format.yml
+++ b/.github/workflows/clang-format.yml
@@ -23,31 +23,38 @@ jobs:
     - name: Format src/test
       run: find src/test -type f \( -name '*.cpp' -o -name '*.h' -o -name '*.ipp' \) -print0 | xargs -0 clang-format-${CLANG_VERSION} -i
     - name: Check for differences
-      shell: bash
+      id: assert
       run: |
         set -o pipefail
-        echo "If you are reading this, you are probably looking at a failed"
-        echo "Github Actions job. That means you pushed a file that did not"
-        echo "conform to the formatting specified in .clang-format. That may"
-        echo "be because you neglected to run 'git clang-format' or "
-        echo "'clang-format' before committing, or that your version of"
-        echo "clang-format has an incompatibility with the one on this"
-        echo "machine, which is:"
-        clang-format-${CLANG_VERSION} --version
-        echo
-        echo "To fix it, you can do one of two things:"
-        echo "1. (The hard way): Download and apply the patch generated as an"
-        echo "   artifact of this job to your repo, commit, and push."
-        echo "2. (The easy way): Run "
-        echo "   'git-clang-format --extensions c,cpp,h,cxx,ipp develop'"
-        echo "   in your repo, commit, and push."
-        git diff --exit-code | tee "${{ github.sha }}.patch"
-    - name: Publish patch
-      if: failure()
+        git diff --exit-code | tee "clang-format.patch"
+    - name: Upload patch
+      if: failure() && steps.assert.outcome == 'failure'
       uses: actions/upload-artifact@v2
       continue-on-error: true
       with:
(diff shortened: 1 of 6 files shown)


.github/workflows/clang-format.yml
```diff
@@ -23,5 +23,31 @@ jobs:
...
+        echo "   in your repo, commit, and push."
+        git diff --exit-code | tee "${{ github.sha }}.patch"
+    - name: Publish patch
+      if: failure()
+      uses: actions/upload-artifact@v2
+      continue-on-error: true
+      with:
+        name: ${{ github.sha }}.patch
+        if-no-files-found: ignore
+        path: ${{ github.sha }}.patch
```
thejohnfreeman, line 52: Here's a slightly different technique that:

- Only prints the message and uploads a patch when the step that checks `git diff --exit-code` fails, instead of when any prior step fails.
- Sticks multi-line strings in environment variables to make them easier to read and edit.

```yaml
- name: Check for differences
  id: assert
  run: git diff --exit-code | tee "${{ github.sha }}.patch"
- name: Print message
  if: steps.assert.outcome == 'failure'
  env:
    PREAMBLE: |
      If you are reading this, you are looking at a failed Github Actions job.
      That means you pushed a file that did not conform to the formatting
      specified in .clang-format. That may be because you neglected to run
      'git clang-format' or 'clang-format' before committing, or that your
      version of clang-format has an incompatibility with the one on this
      machine, which is:
    SUGGESTION: |
      To fix it, you can do one of two things:
      1. (The hard way): Download and apply the patch generated as an
         artifact of this job to your repo, commit, and push.
      2. (The easy way): Run
         'git-clang-format --extensions c,cpp,h,cxx,ipp develop'
         in your repo, commit, and push.
  run:
    echo ${PREAMBLE}
    clang-format-${CLANG_VERSION} --version
    echo ${SUGGESTION}
- name: Upload patch
  if: steps.assert.outcome == 'failure'
  uses: actions/upload-artifact@v2
  continue-on-error: true
  with:
    name: ${{ github.sha }}.patch
    if-no-files-found: ignore
    path: ${{ github.sha }}.patch
```

thejohnfreeman (changes requested):

.github/workflows/clang-format.yml
nbougalis: That's sleek!

ximinez: My initial intent was to keep it "simple", but that's NIIIIIICE. Yeah, I'm totally going to steal it.

thejohnfreeman (approved):

thejohnfreeman (approved):

PR #3753, opened 2021-02-03
[Trivial] Remove dead code

thejohnfreeman:
What it says on the tin. Unused methods, undefined declarations, unused members, etc.


src/ripple/overlay/impl/OverlayImpl.cpp
```diff
@@ -99,9 +99,7 @@ OverlayImpl::Timer::on_timer(error_code ec)
     overlay_.m_peerFinder->once_per_second();
     overlay_.sendEndpoints();
     overlay_.autoConnect();
-
-    if ((overlay_.timer_count_ % Tuning::checkIdlePeers) == 0)
```
gregtatcam, line 103: Why is this deleted? This was added so that we don't check for idle peers too frequently.

thejohnfreeman: `timer_count_` is [initialized](https://github.com/ripple/rippled/pull/3753/files#diff-3f3119d499efbf4f0ec803cf22831e82d48df5a396e321bb343961c920454b8eL141) to 0 and never updated. It is always 0, and this condition is always true.

gregtatcam: oh, right! it's a bug then! it should be incremented.

gregtatcam: i see what happened. it used to be this:

    if ((++overlay_.timer_count_ % Tuning::checkSeconds) == 0)
        overlay_.check();

    if ((overlay_.timer_count_ % Tuning::checkIdlePeers) == 0)
        overlay_.deleteIdlePeers();

but then two top lines were deleted in one of the unrelated commits.

src/ripple/nodestore/impl/Shard.cpp
```diff
@@ -583,17 +583,6 @@ Shard::finalize(
...
-        /*
-        TODO MP
-        A lock is required when calling the NuDB verify function. Because
-        this can be a time consuming process, the server may desync.
-        Until this function is modified to work on an open database, we
-        are unable to use it from rippled.
-
-        // Verify backend integrity
-        backend_->verify();
-        */
```
seelabs, line 595: Let's check with @miguelportilla before removing this and the `verify` implementation.

src/ripple/overlay/impl/OverlayImpl.cpp
seelabs: This doesn't seem like the right PR to fix that bug, but I wouldn't remove the code either. Let's just keep the code for now and open an issue so we don't lose track of this. @gregtatcam can you open an issue?

seelabs (commented):
This looks good - thanks for the cleanup. Left two minor comments.

src/ripple/overlay/impl/OverlayImpl.cpp
gregtatcam: I opened the issue #3754.

src/ripple/nodestore/impl/Shard.cpp
miguelportilla: I am okay removing this commented out code. Although the backend verify function isn't used, it could be useful at some point. For instance, the above situation can be addressed by calling verify on the new deterministic backend. I could also see a server startup switch to have the backend be verified at startup.

cryptobrad (commented):
:100: 

There is some other functions or things that are dead too

`AcceptedLedger::getTxn(int)` at https://github.com/ripple/rippled/blob/develop/src/ripple/app/ledger/AcceptedLedger.h#L72-L73

`AutoSocket::{setSSLOnly|setPlayOnly}` at https://github.com/ripple/rippled/blob/develop/src/ripple/net/AutoSocket.h#L80-L89

`LedgerReplayTaks::getCountedObjectName` at https://github.com/ripple/rippled/blob/develop/src/ripple/app/ledger/LedgerReplayTask.h#L132-L136

`LedgerMaster::getBuildingLedger` at https://github.com/ripple/rippled/blob/develop/src/ripple/app/ledger/impl/LedgerMaster.cpp#L572-L577

`LedgerMaster::NO_VALIDATED_LEDGER_AGE` at https://github.com/ripple/rippled/blob/develop/src/ripple/app/ledger/LedgerMaster.h#L76-L77

`NetworkOps::{getTxMap|hasTxSet}` at https://github.com/ripple/rippled/blob/develop/src/ripple/app/misc/NetworkOPs.cpp#L366-L372

and also these: 

https://github.com/ripple/rippled/blob/develop/src/ripple/app/misc/NetworkOPs.cpp#L144-L145

`NetworksOPs::getConsensusLCL` at https://github.com/ripple/rippled/blob/develop/src/ripple/app/misc/NetworkOPs.h#L244-L245

`LedgerDeltaAcquire::getCountedObjectName` at https://github.com/ripple/rippled/blob/develop/src/ripple/app/ledger/impl/LedgerDeltaAcquire.h#L114-L118

`LedgerReplayTaks::getCountedObjectName` at https://github.com/ripple/rippled/blob/develop/src/ripple/app/ledger/LedgerReplayTask.h#L132-L136

pwang200:
As @cryptobrad said `LedgerDeltaAcquire::getCountedObjectName` and `
LedgerReplayTaks::getCountedObjectName` should be removed. They are not required by `CountedObject` any more.   `SkipListAcquire::getCountedObjectName`  https://github.com/ripple/rippled/blob/80bd107e570b9112de3b2ef52937960986ad5347/src/ripple/app/ledger/impl/SkipListAcquire.h#L112-L116 should also be removed.

seelabs (approved):
:+1: Nice cleanups

thejohnfreeman:
Thanks for the suggestions! I'll work to include them, and another I identified today, so I'll put the PR in draft mode temporarily.

gregtatcam (commented):
Also unused OverlayImpl::serverHandler() and PeerImp::networkID()

nbougalis (approved):

PR #3755, opened 2021-02-04
Add database backend counters

miguelportilla:
This change removes the public Database getBackend function and adds a new function that returns backend read and write stats.

### DIFF
--- a/src/ripple/nodestore/Database.h
+++ b/src/ripple/nodestore/Database.h
@@ -20,8 +20,6 @@
 #ifndef RIPPLE_NODESTORE_DATABASE_H_INCLUDED
 #define RIPPLE_NODESTORE_DATABASE_H_INCLUDED
 
-#include <ripple/basics/KeyCache.h>
-#include <ripple/basics/TaggedCache.h>
 #include <ripple/core/Stoppable.h>
 #include <ripple/nodestore/Backend.h>
 #include <ripple/nodestore/NodeObject.h>
@@ -93,6 +91,11 @@ public:
     */
     virtual std::int32_t
     getWriteLoad() const = 0;
+
+    /** Retrieve backend read and write stats. The return pointer may be null.
+    */
+    virtual Backend::Counters const*
+    getCounters() const = 0;
 
     /** Store the object.
 
@@ -177,9 +180,6 @@ public:
     /** Remove expired entries from the positive and negative caches. */
     virtual void
     sweep() = 0;
-
-    virtual Backend&
-    getBackend() = 0;
 
     /** Gather statistics pertaining to read and write activities.
      *
--- a/src/ripple/nodestore/Database.h
+++ b/src/ripple/nodestore/Database.h
@@ -314,10 +314,15 @@ private:
     for_each(std::function<void(std::shared_ptr<NodeObject>)> f) = 0;
 
     /** Retrieve backend read and write stats.
-        The return pointer is treated like an std::optional.
+
+        @note The Counters struct is specific to and only used
+              by CassandraBackend.
     */
-    virtual Backend::Counters const*
-    getCounters() const = 0;
+    virtual std::optional<Backend::Counters<std::uint64_t>>
+    getCounters() const
+    {
+        return std::nullopt;
+    }
 
     void
     threadEntry();
--- a/src/ripple/nodestore/Database.h
+++ b/src/ripple/nodestore/Database.h
@@ -92,11 +92,6 @@ public:
     virtual std::int32_t
     getWriteLoad() const = 0;
 
-    /** Retrieve backend read and write stats. The return pointer may be null.
-    */
-    virtual Backend::Counters const*
-    getCounters() const = 0;
-
     /** Store the object.
 
         The caller's Blob parameter is overwritten.
@@ -258,10 +253,6 @@ protected:
         storeSz_ += sz;
     }
 
-    // Called by the public asyncFetch function
-    void
-    asyncFetch(uint256 const& hash, std::uint32_t ledgerSeq);
-
     // Called by the public import function
     void
     importInternal(Backend& dstBackend, Database& srcDB);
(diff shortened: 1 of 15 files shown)


src/ripple/nodestore/Database.h
```diff
@@ -94,6 +92,11 @@ class Database : public Stoppable
     virtual std::int32_t
     getWriteLoad() const = 0;
 
+    /** Retrieve backend read and write stats. The return pointer may be null.
+    */
+    virtual Backend::Counters const*
```
seelabs, line 97: We're using the pointer like an `optional<Counters&>`, which I like, but we should probably add a comment saying so.

seelabs (commented):
Very nice changes. Getting rid of `getBackend` is a win. Left one nit.

I would have given a thumbs up, but  `src/ripple/nodestore/Database.h` `src/ripple/nodestore/impl/DatabaseNodeImp.h` need to be reformatted. :+1: after we take care of that.

miguelportilla:
Clang format seems to be case-sensitively comparing include filenames. It wants to place #include <ripple/basics/TaggedCache.h> before #include <ripple/basics/chrono.h>  There may be a way to configure it to be case-insensitive on include filenames. Clang 12 supports SI_CaseInsensitive.

Clang format also wants to indent a single line comment end by a single space. This is avoided if it's a multi-line comment. eg.
```
/**
 */```

File: src/ripple/nodestore/Database.h [hunk 2]
```diff
@@ -322,6 +313,12 @@ class Database : public Stoppable
     virtual void
     for_each(std::function<void(std::shared_ptr<NodeObject>)> f) = 0;
 
+    /** Retrieve backend read and write stats.
+        The return pointer is treated like an std::optional.
```
seelabs, line 317: I'd explicitly say it's treated like a "std::optional to a reference". The reason we're using a pointer is because of the reference, otherwise we would just use a regular `optional`.

seelabs (approved):
:+1: Looks good, but left one nit.

src/ripple/nodestore/impl/DatabaseShardImp.cpp
```diff
@@ -1893,6 +1887,23 @@ DatabaseShardImp::checkHistoricalPaths() const
...
+        std::lock_guard lock(mutex_);
+        assert(init_);
+
+        auto const it{shards_.find(acquireIndex_)};
+        if (it == shards_.end())
+            return nullptr;
+        shard = it->second;
+    }
+
+    return shard->getCounters();
```
scottschurr, line 1904: I am really unfamiliar with the lifetime characteristics of the contents of `shards_`, but this doesn't look safe to me.  The `shard` local keeps the returned pointer valid only for the duration of this function.  That means the returned pointer is not guaranteed to be valid in the caller's context.

Maybe there are other guarantees keeping the pointer valid after the return?  But if so I think it makes sense to leave a comment describing what those guarantees are.

If I've spotted a real issue, one solution would be to not return `Backend::Counters` at all.  Return a `weak_ptr` to the correct `Shard` and let the final destination that needs `Counters` do the access.  That way the scope that actually needs to access `Counters` can verify that the requested `Shard` is still valid and lock it in place.

Another choice would be to return `Backend::Counters` by value.  That would take a bit more work because `Backend::Counters` contains a bunch of `atomic`s.  But if `Backend::Counters` were a template that could accept either `std::atomic<std::uint64_t>` or `std::uint64_t` then you could return the non-`atomic` form from this  method.

Just my take.  I could easily be way off base.

nbougalis: Why even return a pointer instead of a copy?

seelabs: @scottschurr It's safe because `counters` always has a static lifetime (for everything bug the reporting mode casandra backend, but that's doesn't come into play here). Since it's static, the return value is valid even if `shard` goes away.

seelabs: @nbougalis returning a copy (`optional<Counters>`) seems reasonable. The struct is pretty small.

miguelportilla: The ```Backend::Counters``` returned is static but the issue you describe is possible should that change. The Shard class is a private implementation not meant to be visible beyond DatabaseShard. Since Counters was added solely for and used only by ```CassandraBackend```, I propose we have a default base implementation that returns nullptr.

miguelportilla: Pushed a commit with @scottschurr's template suggestion.  @scottschurr Thanks for the idea and help.

seelabs (approved):
Still :+1: after 85862b098237d1c799cc99367696524d526e30d1 (nice cleanup there, btw).

scottschurr (approved):
:+1: Looks great to me!

nbougalis (approved):

thejohnfreeman:
I wanted to pitch an alternative technique for doing this, modeled after `std::ostream`: pass a reference to the output "stream" that the method writes into, instead of a method that returns a named tuple used in only one place and only to be rendered to JSON.

Specifically, let `getCounters` (or maybe rename it to `writeMetrics`) take a `Json::Value&`, assign into it, and return either the same reference (for chaining) or `void`. Document the form of the written metrics in the docstring (i.e. field names and types). Remove the `Counters` class.

Advantages:

- Avoids copying the counts and skips straight to writing strings. In the future, you might imagine an optimized `JsonStream` that just holds a string buffer instead of a map, formatting each field as it is added instead of waiting until the end, minimizing allocations.
- Relieves us from adding and maintaining `Counters` classes everywhere we want to record metrics, and from properly handling null pointers when they are absent.
- Lets us juxtapose the read of the metrics with their JSONification in the same method, instead of in separate, distant methods.
- Ensures we use consistent field names everywhere a metric is extracted.

PR #3756, opened 2021-02-04
Fix idle peer deletion timer

### LINKED ISSUE #3754: increment timer count in the OverlayImpl::on_timer()
## Issue Description

timer_count_ should be incremented in OverlayImpl::on_timer(). Currently time_count_ is initialized to 0 and is not incremented. Consequently deleteIdlePeers() is called every second instead of once in four seconds.

gregtatcam:

Increments OverlayImpl::timer_count_, so that idle peer deletion is checked once every 4 seconds.

### Context of Change

This PR fixes #3754, which was introduced in 1.7.0-b5.

### Type of Change


seelabs (approved):
:+1: LGTM

miguelportilla (approved):
LGTM 👍

PR #3759, opened 2021-02-05
Replace most uses of `boost::optional` with `std::optional

scottschurr:
The rippled code base makes heavy use of `optional`.  It started with `boost::optional`.  That's what was available at the time.  Since the code base now requires C++17, we have `std::optional` available.  This pull request converts most uses of `boost::optional` to `std::optional`.

There are two situations that could not be converted:

- If the `optional` is passed as a parameter to SOCI, then the parameter must be a `boost::optional`.
- If the `optional` is passed to as a parameter to `boost::beast`, then the parameter must be a `boost::optional`.

In all of those places comments were added about why a `boost::optional` was being used.

### Context of Change
Using two types of `optional` indiscriminately in the code base was messy and confusing.  Preferring `std::optional` when it is viable seems like a reasonable approach to reducing the confusion.

### Type of Change

The change was addressed piecemeal.  One comparatively small area was changed and all of the fallout breakage was fixed.  That set of changes was committed.  Then another small set of changes was attempted.  Each commit builds and passes unit tests on its own.

That approach is preserved in the pull request commits.  So there are a ton of commits.  If the pull request passes review I'll squash all of the commits down to a single commit.

If I were reviewing this pull request I'd probably do it commit-by-commit.  But reviewers should feel free to take their favorite approach.

## Future Tasks
There may be other `boost` library types that we wish to convert over to `std`.  Possibly `string_view`?  `filesystem`?

## Release Notes
This change has no user impact, so no release notes are required.

### DIFF
--- a/src/ripple/app/consensus/RCLConsensus.cpp
+++ b/src/ripple/app/consensus/RCLConsensus.cpp
@@ -101,7 +101,7 @@ RCLConsensus::Adaptor::Adaptor(
                     << ", Cookie: " << valCookie_ << ")";
 }
 
-boost::optional<RCLCxLedger>
+std::optional<RCLCxLedger>
 RCLConsensus::Adaptor::acquireLedger(LedgerHash const& hash)
 {
     // we need to switch the ledger we're working from
@@ -124,7 +124,7 @@ RCLConsensus::Adaptor::acquireLedger(LedgerHash const& hash)
                         id, 0, InboundLedger::Reason::CONSENSUS);
                 });
         }
-        return boost::none;
+        return std::nullopt;
     }
 
     assert(!built->open() && built->isImmutable());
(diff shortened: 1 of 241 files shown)


gregtatcam (commented):
There are some build errors in CI: reporting, clang-8, debug

scottschurr:
Thanks @gregtatcam.  Looks like there are some problems in  CassandraFactory.cpp.  I'll see what I can figure out.

cjcobb23:
> Thanks @gregtatcam. Looks like there are some problems in CassandraFactory.cpp. I'll see what I can figure out.

I think these are the same errors currently present on `develop`. There is an open PR to fix this. It's a one line change that you can cherry pick: https://github.com/thejohnfreeman/rippled/commit/bb042312115363f39ce84499d4997f235f5c7367

gregtatcam (approved):
:+1: LGTM. Nice update! There is an issue of the reporting mode build error but I don't think it holds this PR.

miguelportilla (approved):
Nice cleanup of the optional. It's unfortunate that soci has forced us to cling unto boost optional.

LGTM 👍

scottschurr:
Since I've rebased to -rc3 the CI all passes 🎉.  Is everybody good for me to squash this all down to a single commit?  Thanks for the review work!

gregtatcam:
> Since I've rebased to -rc3 the CI all passes . Is everybody good for me to squash this all down to a single commit? Thanks for the review work!

Sure, fine with me.

miguelportilla:
@scottschurr I am fine with you squashing it all down.

scottschurr:
Squashed.  I'll mark as passed once it survives CI.

PR #3760, opened 2021-02-05
Add constexpr constructor for `base_uint`

scottschurr:
Adds the capability of building a `base_uint`, specifying the specific value, at compile time.

### Context of Change
There are a few places in the code where `base_unit` values are known at compile time, but the values could not be built until runtime.  Adding this constructor allows those values to be built at compile time.  Benefits include:

- There are no race conditions regarding the value's construction.
- The compiler does not emit any executable code to produce the values.
- Code executed during `constexpr` evaluation is guaranteed by the compiler to not exhibit undefined behavior.
- It's kinda fun to see what you can get the compiler to do at compile time

There are not a lot of these compile-time values, so this is not an earth shattering change.

### Type of Change
- [ x] Tests (tests were added for this feature)

## Before / After
Before:
```
    static uint256 const nextq = []() {
        uint256 x;
        (void)x.parseHex(		
            "0000000000000000000000000000000000000000000000010000000000000000");		
        return x;		
    }();
```
After:
```
    static constexpr uint256 nextq(
        "0000000000000000000000000000000000000000000000010000000000000000");
```

## Future Tasks
We can consider `constexpr` enabling more stuff.

## Release notes
This change has no user impact, so no release notes are required.

### DIFF
--- a/src/ripple/basics/base_uint.h
+++ b/src/ripple/basics/base_uint.h
@@ -25,6 +25,7 @@
 #ifndef RIPPLE_BASICS_BASE_UINT_H_INCLUDED
 #define RIPPLE_BASICS_BASE_UINT_H_INCLUDED
 
+#include <ripple/basics/contract.h>
 #include <ripple/basics/hardened_hash.h>
 #include <ripple/basics/strHex.h>
 #include <ripple/beast/utility/Zero.h>
@@ -177,20 +178,107 @@ private:
         memcpy(data_.data(), data, bytes);
     }
 
-public:
-    base_uint()
+    // Helper function to initialize a base_uint from a std::string_view.
+    enum class ParseResult {
+        okay,
+        badLength,
+        badChar,
+    };
+
+    constexpr std::pair<ParseResult, decltype(data_)>
+    parseFromStringView(std::string_view sv) noexcept
     {
-        *this = beast::zero;
+        // Local lambda that converts a single hex char to four bits and
+        // ORs those bits into a uint32_t.
+        auto hexCharToUInt = [](char c,
+                                std::uint32_t shift,
+                                std::uint32_t& accum) -> ParseResult {
+            std::uint32_t nibble = 0xFFu;
+            if (c < '0' || c > 'f')
+                return ParseResult::badChar;
+
+            if (c >= 'a')
+                nibble = static_cast<std::uint32_t>(c - 'a' + 0xA);
+            else if (c >= 'A')
+                nibble = static_cast<std::uint32_t>(c - 'A' + 0xA);
+            else if (c <= '9')
+                nibble = static_cast<std::uint32_t>(c - '0');
+
+            if (nibble > 0xFu)
+                return ParseResult::badChar;
+
+            accum |= (nibble << shift);
+
+            return ParseResult::okay;
+        };
+
+        std::pair<ParseResult, decltype(data_)> ret{ParseResult::okay, {}};
+
+        if (sv == "0")
+        {
+            return ret;
+        }
+
+        if (sv.size() != size() * 2)
+        {
+            ret.first = ParseResult::badLength;
+            return ret;
+        }
+
+        auto out = ret.second.data();
+
+        auto in = sv.begin();
+
+        while (in != sv.end())
+        {
+            std::uint32_t accum = {};
+            for (std::uint32_t shift : {4u, 0u, 12u, 8u, 20u, 16u, 28u, 24u})
+            {
+                if (auto const result = hexCharToUInt(*in++, shift, accum);
+                    result != ParseResult::okay)
+                {
+                    ret.first = result;
+                    return ret;
+                }
+            }
+            *out++ = accum;
+        }
+        return ret;
     }
 
-    base_uint(beast::Zero)
+    constexpr decltype(data_)
+    parseFromStringViewThrows(std::string_view sv) noexcept(false)
     {
-        *this = beast::zero;
+        auto const result = parseFromStringView(sv);
+        if (result.first == ParseResult::badLength)
+            Throw<std::invalid_argument>("invalid length for hex string");
+
+        if (result.first == ParseResult::badChar)
+            Throw<std::range_error>("invalid hex character");
+
+        return result.second;
+    }
+
+public:
+    constexpr base_uint() : data_{}
+    {
+    }
+
+    constexpr base_uint(beast::Zero) : data_{}
+    {
     }
 
     explicit base_uint(std::uint64_t b)
     {
         *this = b;
+    }
+
+    // This constructor is intended to be used at compile time since it might
+    // throw at runtime.  Consider declaring this constructor consteval once
+    // we get to C++23.
+    explicit constexpr base_uint(std::string_view sv) noexcept(false)
+        : data_(parseFromStringViewThrows(sv))
+    {
     }
 
     template <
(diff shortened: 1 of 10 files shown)


src/ripple/basics/base_uint.h
```diff
@@ -177,22 +178,82 @@ class base_uint
...
+            accum |= hexCharToUInt(*in++) << 0;
+            accum |= hexCharToUInt(*in++) << 12;
+            accum |= hexCharToUInt(*in++) << 8;
+            accum |= hexCharToUInt(*in++) << 20;
+            accum |= hexCharToUInt(*in++) << 16;
+            accum |= hexCharToUInt(*in++) << 28;
+            accum |= hexCharToUInt(*in++) << 24;
+            *out++ = accum;
+        }
+        return ret;
```
thejohnfreeman, line 232: Is it possible to factor out some common function between this and `parseHex`?

nbougalis: Good question.

I think that instead of factoring out common functionality, I'd simply make `parseHex(std::string_view)` be `constexpr`. Doing this will require making [`charUnHex`](https://github.com/ripple/rippled/blob/develop/src/ripple/basics/impl/strHex.cpp#L25-L47) be `constexpr` but that seems simple and, honestly, desirable in other contexts.

src/ripple/basics/base_uint.h
```diff
@@ -177,22 +178,82 @@ class base_uint
...
-        *this = beast::zero;
     }
 
-    base_uint(beast::Zero)
+    constexpr base_uint(beast::Zero) : data_{}
     {
-        *this = beast::zero;
     }
 
     explicit base_uint(std::uint64_t b)
```
nbougalis, line 244: This is probably easy to also turn into a `constexpr`, but I'm not sure if I'd prefer to do that or eliminate this constructor altogether.

nbougalis (approved):
Looks good, thanks! The more things we can do at complile time, the better!

At some point we should look at `base_uint` more closely. It was inherited from Bitcoin way back in 2011, and it's been modified a lot since then, so it's a bit of a hot mess.

src/ripple/basics/base_uint.h
thejohnfreeman: I thought that wasn't possible because `constexpr` implies `const`, but it looks like that changed in C++14. In that case, I would prefer `constexpr parseHex` too.

scottschurr: I can make `parseHex()` `constexpr`.  In the process most of the code between `parseHex()` and the constructor from `string_view` will be the same.  A `constexpr` `parseHex()` is not likely to be especially useful since it mutates data members.  This means `parseHex()` can be used `constexpr` as long as that mutation stays inside of a single `constexpr` context.  Once that context exits then any `base_uint` produced by that context becomes immutable.

There are some downsides to the shared code approach.  An important distinction between `parseHex()` and the constructor is that `parseHex()` does not throw on errors; the constructor does.  This means the code must be structured a bit differently.

It's also important to note that `base_uint`'s `iterator` and `data()` interfaces can never be used in a `constexpr` context because they require `reinterpret_cast`.  That's why the `constexpr` methods in `base_uint` operate on `std::uint32_t`, which is the actual data type inside the `std::array`.

I've pushed a commit showing what the `parseHex()` changes look like.  If you like it I can keep the commit.  Otherwise it's easy to back out.

src/ripple/basics/base_uint.h
scottschurr: This constructor is not easy to turn `constexpr` since the implementation currently calls `boost::endian::native_to_big()` which is not `constexpr` until boost 1.72.0 for gcc and clang.  It also appears that `native_to_big()` is not `constexpr` for MSVC in any boost release.

HowardHinnant (approved):
Nice change.  I prefer the shared parseHex, but am also fine with this PR without it.

src/ripple/basics/base_uint.h
thejohnfreeman: I was thinking of a different approach to `constexpr parseHex`. The member `data_` is a `std::array`, so we don't really gain anything by constructing it in an initializer instead of default constructing it and assigning into it from the body of `base_uint`'s constructor. I was picturing both `parseHex` and the constructor would call a third function that takes both the `std::string_view` input parameter and a `std::array<...>&` output parameter, returns some error type (`bool`, `ParseResult`, `std::exception_ptr`, whatever), and assigns into the array. `parseHex` would just return the same value, and the constructor would branch on that value to throw the right exception. That "third" function could actually just be `parseHex`, though I wouldn't mind it being a separate function.

scottschurr: @thejohnfreeman, I'm not very clear on the distinction between what you're describing and what I've coded up.  May I suggest you code up the alternative you have in mind and we'll take a look?

In terms of initializing `base_uint::data_` in the constructor, all variables must always be in a known state in a constexpr context.  So if we allow `data_` to default initialize, it will initially be filled with all zeros.  Then we would overwrite those initial values in the constructor body.

To me it seems a bit more to the point to simply initialize `data_` with the right value in the first place.  But your milage may vary.

thejohnfreeman: [This](https://github.com/thejohnfreeman/rippled/commit/73ebc5333981305586d331f3fccc7fa277e117b2) is what I was thinking. It can still be changed to give a better error message, and to distinguish between errors of "too few digits" and "non-digit found", but I think it captures the idea of sharing the main body without too much wrangling over return code vs exception.

Most of the time, I too insist on initializing members in their initializer, but in the case of a `constexpr` constructor, I'm fine relaxing that rule to get code that I think is easier to read. The compiler [produces](https://godbolt.org/z/978T4n) the same thing in the end.

scottschurr: @thejohnfreeman, thanks for the example.  Certainly both approaches do the job.

Personally, I like having the constructor distinguish between bad lengths and invalid characters in the error.  If we sacrifice that usability improvement then what you've coded up works well.  We could add the length/character distinction into the return value of `parseHex()`, but then we'd need to edit all of the callers of `parseHex()`.  I don't think that's worth it.

I'm also not keen on moving `hexCharToUint(char c, T& accum, T shift = 0)` out to be a library function and making it a template.  As a general function it should be receiving more extensive testing.  It also needs some guards on the various kinds of template parameters it can take.  I just don't suspect it will get much use out there.  I like to keep simple specialized functions, like the original version, in a local scope.  But that's just my personal preference.

But if you find your implementation is truly preferable then, cool.  Add appropriate template parameter guards to `hexCharToUint()` and add some unit tests and we can call it good.  I'm open to losing the distinction in error types.

thejohnfreeman (approved):

scottschurr:
Since I've rebased to -rc3 the CI all passes 🎉. Is everybody good for me to squash this all down to a single commit? Thanks for the review work!

scottschurr:
Rebased and squashed.  I'll mark as passed once CI is (presumably) successful.

scottschurr:
Since the rebase to 1.7.0 I'm seeing a CI failure of windows vc2119.

@miguelportilla helped to discover that the failure is caused by vc2019 version 19.28.29910 running on the tip of our develop.  The failure does not seem to be related to any of the changes introduced by this branch.  So I'm going to mark the pull request as passed.

Thanks a ton for the help, @miguelportilla!

PR #3758, opened 2021-02-05
Fix declaration of CassandraBackend::counters_

thejohnfreeman:

PR #3763, opened 2021-02-09
Fix "sign" & "submit" Sequence field autofill with Tickets

scottschurr:

@manojsdoshi found a bug in the handling of `Sequence` autofill by the "sign" and "submit" RPC commands when a `TicketSequence` is part of the transaction.

As it stands, if the `Sequence` field is missing, then the next usable sequence number associated with the account is filled in.  This leads to unexpected behavior.  The non-zero `Sequence` field takes precedent over the `TicketSequence`, so the associated `Ticket` is not consumed.

The fix is for "sign" and "submit" to check the transaction for a `TicketSequence`.  If a `TicketSequence` is present and a `Sequence` field is not, then the commands fill in `'Sequence': 0` which is what is required for the `TicketSequence` to be used.

### Type of Change
- [x] Bug fix (non-breaking change which fixes an issue).
- [x] Test (added unit test to verify fix).

## Test Plan
Using the "sign" or "submit" RPC commands, submit a transaction with a `TicketSequence` but no `Sequence` field.

Before the fix you should observe the transaction returned by the command has a non-zero `Sequence` field filled in.  If the transaction is submitted then the `TicketSequence` will not be consumed.

After the fix you should observe the transaction returned by the command has a `'Sequence': 0` field filled in.  If the transaction is submitted then the `TicketSequence` is consumed if the transaction returns a `tec` or `tesSUCCESS` code.

### DIFF
--- a/src/ripple/app/tx/impl/Transactor.cpp
+++ b/src/ripple/app/tx/impl/Transactor.cpp
@@ -57,7 +57,7 @@ preflight1(PreflightContext const& ctx)
 {
     // This is inappropriate in preflight0, because only Change transactions
     // skip this function, and those do not allow an sfTicketSequence field.
-    if (ctx.tx.getSeqProxy().isTicket() &&
+    if (ctx.tx.isFieldPresent(sfTicketSequence) &&
         !ctx.rules.enabled(featureTicketBatch))
     {
         return temMALFORMED;
@@ -254,18 +254,29 @@ Transactor::checkSeqProxy(
     SeqProxy const t_seqProx = tx.getSeqProxy();
     SeqProxy const a_seq = SeqProxy::sequence((*sle)[sfSequence]);
 
-    if (t_seqProx.isSeq() && t_seqProx != a_seq)
+    if (t_seqProx.isSeq())
     {
-        if (a_seq < t_seqProx)
+        if (tx.isFieldPresent(sfTicketSequence) &&
+            view.rules().enabled(featureTicketBatch))
         {
-            JLOG(j.trace()) << "applyTransaction: has future sequence number "
-                            << "a_seq=" << a_seq << " t_seq=" << t_seqProx;
-            return terPRE_SEQ;
+            JLOG(j.trace()) << "applyTransaction: has both a TicketSequence "
+                               "and a non-zero Sequence number";
+            return temSEQ_AND_TICKET;
         }
-        // It's an already-used sequence number.
-        JLOG(j.trace()) << "applyTransaction: has past sequence number "
-                        << "a_seq=" << a_seq << " t_seq=" << t_seqProx;
-        return tefPAST_SEQ;
+        if (t_seqProx != a_seq)
+        {
+            if (a_seq < t_seqProx)
+            {
+                JLOG(j.trace())
+                    << "applyTransaction: has future sequence number "
+                    << "a_seq=" << a_seq << " t_seq=" << t_seqProx;
+                return terPRE_SEQ;
+            }
+            // It's an already-used sequence number.
+            JLOG(j.trace()) << "applyTransaction: has past sequence number "
+                            << "a_seq=" << a_seq << " t_seq=" << t_seqProx;
+            return tefPAST_SEQ;
+        }
     }
     else if (t_seqProx.isTicket())
     {
--- a/src/ripple/protocol/impl/TER.cpp
+++ b/src/ripple/protocol/impl/TER.cpp
@@ -152,6 +152,7 @@ transResults()
         MAKE_ERROR(temINVALID_ACCOUNT_ID,     "Malformed: A field contains an invalid account ID."),
         MAKE_ERROR(temCANNOT_PREAUTH_SELF,    "Malformed: An account may not preauthorize itself."),
         MAKE_ERROR(temINVALID_COUNT,          "Malformed: Count field outside valid range."),
+        MAKE_ERROR(temSEQ_AND_TICKET,         "Transaction contains a TicketSequence and a non-zero Sequence."),
 
         MAKE_ERROR(terRETRY,                  "Retry transaction."),
         MAKE_ERROR(terFUNDS_SPENT,            "DEPRECATED."),
--- a/src/test/app/Ticket_test.cpp
+++ b/src/test/app/Ticket_test.cpp
@@ -390,6 +390,10 @@ class Ticket_test : public beast::unit_test::suite
         env.require(owners(env.master, 0), tickets(env.master, 0));
 
         env(noop(env.master), ticket::use(1), ter(temMALFORMED));
+        env(noop(env.master),
+            ticket::use(1),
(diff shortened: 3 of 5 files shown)


ximinez:
Should a transaction be considered malformed if it has a non-zero `Sequence` and a `TicketSequence`?

rasloun:
Unsubscribe me you geeks

carlhua:
> Unsubscribe me you geeks

You actually control the subscription from Github notifications. Please follow the steps here to unsubscribe - https://docs.github.com/en/github/managing-subscriptions-and-notifications-on-github/managing-your-subscriptions

I would like to ask you to stop using offensive languages.

Thanks.

scottschurr:
@ximinez, that's a really reasonable suggestion.  It would be really easy to add that validation into `Transactor::checkSeqProxy`.  But it would take an amendment.  The currently proposed change is outside of transaction processing so it doesn't require an amendment.

I'll code that up a commit making that change and add it to this pull request.  Once we see what it looks like we can decide whether it's a change we'd like to make.

cjcobb23 (approved):
Looks good to me.

scottschurr:
@ximinez, I added a commit that includes the suggested amendment.  Let me know what you think.  I do think that the amendment will help manage confusion resulting from errors in usage.

mDuo13:
It doesn't need an amendment if it gets into v1.7.0 because TicketBatch has not been in any stable releases yet. Changes to unreleased amendments are explicitly _not_ considered to be breaking/transaction processing changes.

scottschurr:
@carlhua, what do you think?  Shall we put this in the next release candidate?

scottschurr:
If we were to incorporate the bug fix into 1.7.0, then no additional amendment would be required.  The most recent `[FOLD]` commit shows wha that would look like.

src/ripple/protocol/impl/TER.cpp
```diff
@@ -101,6 +101,7 @@ transResults()
         MAKE_ERROR(tefINVARIANT_FAILED,       "Fee claim violated invariants for the transaction."),
         MAKE_ERROR(tefTOO_BIG,                "Transaction affects too many items."),
         MAKE_ERROR(tefNO_TICKET,              "Ticket is not in ledger."),
+        MAKE_ERROR(tefSEQ_AND_TICKET,         "Transaction contains a TicketSequence and a non-zero Sequence."),
```
ximinez, line 104: Should this be a `tem`? `tef` usually means that the problem is in an interaction with the ledger, whereas this happens because a tx should never have both `TicketSequence` and `Sequence`

src/test/app/Ticket_test.cpp
```diff
@@ -826,6 +830,161 @@ class Ticket_test : public beast::unit_test::suite
...
+        }
+    }
+
+    void
+    testFixBothSeqAndTicket()
+    {
+        // The sign and the submit RPC commands automatically fill in the
+        // Sequence field of a transaction if none is provided.  If a
+        // TicketSequence is provided in the transaction, then the
+        // auto-filled Sequence should be zero.
```
ximinez, line 930: This is the same comment that was in `testSignWithTicketSequence`

ximinez (changes requested):
The Reporting build failed, but that's because of the issue fixed in #3758

cjcobb23 (approved):

src/ripple/app/tx/impl/Transactor.cpp
```diff
@@ -57,7 +57,7 @@ preflight1(PreflightContext const& ctx)
 {
     // This is inappropriate in preflight0, because only Change transactions
     // skip this function, and those do not allow an sfTicketSequence field.
-    if (ctx.tx.getSeqProxy().isTicket() &&
+    if (ctx.tx.isFieldPresent(sfTicketSequence) &&
```
seelabs, line 60: Isn't this transaction breaking? The old check looked at both the field and the type, The new code only looks at the field. I.e. if there was a `sfTicketSequence` field with a string, the old code let it through and the new code flags it as malformed.

src/ripple/app/tx/impl/Transactor.cpp
```diff
@@ -254,18 +254,29 @@ Transactor::checkSeqProxy(
...
-    if (t_seqProx.isSeq() && t_seqProx != a_seq)
+    if (t_seqProx.isSeq())
     {
-        if (a_seq < t_seqProx)
+        if (tx.isFieldPresent(sfTicketSequence) &&
+            view.rules().enabled(featureTicketBatch))
+        {
+            JLOG(j.trace()) << "applyTransaction: has both a TicketSequence "
+                               "and a non-zero Sequence number";
+            return temSEQ_AND_TICKET;
```
seelabs, line 264: This looks transaction breaking. If the old code had a ticket and a sequence it let it through while the new code flags it as an error.

seelabs (commented):
Let two questions about transaction breakage.

src/ripple/app/tx/impl/Transactor.cpp
seelabs: Answering my own question. I see that `TicketBatch` has not been released yet.

src/ripple/app/tx/impl/Transactor.cpp
seelabs: Answering my own question. I see that `TicketBatch` has not been released yet.

src/ripple/app/tx/impl/Transactor.cpp
```diff
@@ -254,18 +254,29 @@ Transactor::checkSeqProxy(
     SeqProxy const t_seqProx = tx.getSeqProxy();
```
seelabs, line 254: Should we add a check for a `sfTicketSequence` field with the wrong type (i.e. a string?).

seelabs (approved):
:+1: LGTM, tho do see the comment about checking for `sfTicketSequence` fields with the wrong type.

src/ripple/app/tx/impl/Transactor.cpp
scottschurr: Great question!

I had forgotten the details myself, so I did some poking.  The `sfTicketSequence` is an `SField` of type `uint32`.  Any time someone tries to serialize something other than a `uint32` into a `sfTicketSequence` then the serialization machinery generates an error.  By the time an `sfTicketSequence` has arrived at the transaction machinery the type of the data is already validated.

So, long story short, this kind of problem is being handled by our serialization machinery.  We don't need to do any validation here.

FWIW, I tried to add a quick unit test for this.  That doesn't work because the unit test framework itself uses the serialization machinery.  I can't even (easily) push this check through our unit tests.

scottschurr:
I'm removing the `Amendment` and `API Change` labels because this change is going into the first release where `TicketBatch` is available.  Since the API change only applies to transactions with tickets there's not change, only an improvement :wink:.

PR #3766, opened 2021-02-12
S log

seelabs:

This is a draft PR to test the water for adding more structure to our logs. 

### Context of Change
My motivation for this stems from a simple log analyzer I wrote to help monitor my servers. Adding more structure would simplify that program and also allow more interesting analyses.

The log format is the usual header, with a timestamp, module, and log level, followed by a message and json data. As an example for how messages would be changed, consider this `LoadMonitor` message:

```
2021-Feb-05 23:26:39.248228946 UTC LoadMonitor:WRN Job: recvValidation->checkValidation run: 1001ms wait: 0ms
```

in the new format, it would be written as:

```
2021-Feb-05 23:26:39.248228946 UTC LoadMonitor:WRN Job Latency {"Job": "recvValidation->checkValidation", "run(ms)": 1001, "wait(ms)": 0, "jlogId": 115}
```

The way to write out these entries to the log is with the new `JLOGV` macro. It takes a stream, a messaage, and name/value pairs. The name/value pairs will be written as a json object at the end of the message.

Here's an example of how the `JLOGV` macro is used:
```c++
    JLOGV(
        j.debug(),
        "RippleCalc Result>",
        jv("actualIn", flowOut.actualAmountIn),
        jv("actualOut", flowOut.actualAmountOut),
        jv("result", flowOut.result()),
        jv("dstAmtReq", saDstAmountReq),
        jv("sendMax", saMaxAmountReq));
```

Note the use of the `jv` function to define name/value pairs. It would be nice to use braces to define these (i.e. `{"actualIn", flowOut.actualAmountIn}`), but I don't believe it's possible to make that syntax work (tho I'd love to be proven wrong).

There are several disadvantages to this change:
1. Messages can become very long. 
2. While the information is the same, it's in an unfamiliar form. When urgently diagnosing issues, people will be looking at unfamiliar log entries.
3. Breaks existing scripts.
4. Messages can become more cryptic.

The biggest advantage is it's much easier to write scripts to analyze and reformat logs.

The purpose of this PR is to test the waters for this change. I don't believe it's necessary to convert the entire code base to this new format before there are benefits. This PR is a start of that conversion to give people a concrete feel for what the code and the log files would look like.

### Type of Change


seelabs:
I wrote a simple project that demonstrates the usefulness of this PR. See this readme: https://github.com/seelabs/rs-log-analyzer/blob/master/README.md

In particular, the latency analyzer is interesting:
```
Job: processLedgerData: Max Run: 13050 Max Wait: 12982 Ave Run: 1878.38 Ave Wait: 3633.30
Run histogram:
     0 : 7      ********************
     1 : 0      
     2 : 1      **
     4 : 0      
     8 : 0      
    16 : 1      **
    32 : 3      ********
    64 : 2      *****
   128 : 3      ********
   256 : 6      *****************
   512 : 7      ********************
  1024 : 9      **************************
  2048 : 11     ********************************
  4096 : 9      **************************
  8192 : 1      **

Wait histogram:
     0 : 4      *****
     1 : 0      
     2 : 0      
     4 : 0      
     8 : 0      
    16 : 0      
    32 : 0      
    64 : 0      
   128 : 1      *
   256 : 0      
   512 : 3      ****
  1024 : 14     *******************
  2048 : 12     ****************
  4096 : 23     ********************************
  8192 : 3      ****
```

MarkusTeufelberger:
See also https://github.com/ripple/rippled/issues/2963

Personally I think that mixing JSON with structured non-JSON data makes it unnecessarily hard to consume this and still requires parsing these "structured" logs (to get the timestamp, module log level and message data).

seelabs:
@MarkusTeufelberger I agree with your point that it's harder to consume than pure json. I was thinking we'd use a tool to convert to pure json - see the readme here: https://github.com/seelabs/rs-log-analyzer/blob/master/README.md for an example.

The reason I'm not advocating for a pure json log are:
1) It harder for humans to read.
2) There has been resistance in the past for introducing structured logs. This seems like a reasonable compromise.

MarkusTeufelberger:
Well, I'd personally use something like https://vector.dev/, fluentd/fluentbit, logstash or promtail to look at logs instead of writing a custom program to make sense of them. Maybe a full JSON version might be at least optionally on the table, for people that really do not care about human readability of log lines and instead care about extracting their information content?

intelliot:
Just a minor anecdote to mention: when using Google, the first hit that I get for `structured logging` is https://stackify.com/what-is-structured-logging-and-why-developers-need-it/ which uses the same "mixing" of human readable text + JSON.

ximinez:
I'm trying to interpret the histograms. It looks like the average times haven't changed much, but the individual times have gotten more extreme. Is that what you're seeing, too?

Second, what do you think about delimiting the message, or requiring it to not have any spaces? I'm thinking of more primitive parsers like unix `cut` which I think would have an easier time of it. I think would effectively change the message into more of a "category", and any more detailed message would have to be a name/value pair.

seelabs:
@ximinez 

> I'm trying to interpret the histograms. 

We can't compare the histograms - they are for two different job types: "processLedgerData" and "InboundLedger" (if I understood your question correctly).

> what do you think about delimiting the message, or requiring it to not have any spaces?

One thing I'm concerned about is people looking at the formatting of structured logs and having a visceral "yuck!" reaction. I'd prefer to aim for "as human readable as possible while allowing for structured data". So I don't think disallowing spaces would work well. Delimiting the message might be OK, but that does hurt human readability and since messages are free-form we have to make sure not existing of future messages use the delimiter (of course, with this proposal, messages would not be allowed to have a `{` character or it could be confused with json data). If we went with a delimiter `:` would be my first choice, tho right now I probably prefer no delimiter.

scottschurr:
Just for comparison's sake, what do we think fully structured logging would look like?  Perhaps each log statement is something like this?
```
{"date":"2021-Feb-05", "time(UTC)":"23:26:39.248228946", "source": "LoadMonitor:WRN", "JobLatency": {"Job": "recvValidation->checkValidation", "run(ms)": 1001, "wait(ms)": 0, "jlogId": 115}}
```

seelabs:
@scottschurr I'd expect a full json log to look like this:
```
{"timestamp": "2021-Feb-13 22:15:16.114602249 UTC", "level": "Warning", "module": "LoadMonitor", "msg": "Job latency", "data": {"jlogId": 115, "job": "processLedgerData", "run(ms)": 1633, "wait(ms)": 0},   }
```
The difference between that and what you posted is there would still be a `msg` field for a free form message, and there would be a `data` field (instead of a `JobLatency` field). The problem with a `JobLatency` field is it's hard to write code to process that. We'd have to determine what kind of message this is by looking for a field that is not one of the standard fields (i.e. `JobLatency` is not `timestamp`, `jlogId`, ect). I'd rather keep a `msg` field that (somewhat) determines the message and a generic `data` field with the json data.

We could encode that information in the `jlogId`, but I'd rather let those ids change easily and I'd rather keep a flexible format.

scottschurr:
Thanks, @seelabs.  Nice.  I have a few thoughts.

I find our logs to be nearly impenetrable as they are.  There's just too much stuff.  Since that's the case I only go to the log as a last resort.  So my opinion on the topic of logs matters less than the opinion of someone who makes heavy use of the logs.

As a human I find reading the structured form only slightly more difficult than the un-JSONed text.  So, from my perspective, going with JSON is a fair trade off to improve machine readability.

I believe there's nothing saying that our JSON objects must be written out in alphabetical order.  So as long as the standard fields in the JSON always appear in every line of the file at the front of the line and in the same order, and the more fluid (or omissible) fields are at the back, that should help with readability by humans.

I'd be inclined to go with the full JSON log.  I think the benefits of full machine readability outweigh the (in my opinion) slightly increased difficulty with human comprehensibility.

But what matters more is to get opinions from folks who make heavy use of logs.  My opinion really shouldn't carry very much weight.

seelabs:
Thanks @scottschurr If the consensus if for full json I'm fine with that.

ximinez:
@scottschurr makes good points about full JSON logs. We could make it even more "compact" and human readable by using very short names for those standard fields. E.g.
```
{"t": "2021-Feb-13 22:15:16.114602249 UTC", "lvl": "Warning", "mod": "LoadMonitor", "msg": "Job latency", "data": {"jlogId": 115, "job": "processLedgerData", "run(ms)": 1633, "wait(ms)": 0} }
```
or
```
{"t": "2021-Feb-13 22:15:16.114602249 UTC", "lvl": "Warning", "mod": "LoadMonitor", "msg": "Job latency", "jlogId": 115, "job": "processLedgerData", "run(ms)": 1633, "wait(ms)": 0 }
```
The second option is even more terse, but we'd have to validate that none of the data items have conflicting names.

scottschurr:
@ximinez, those look workable to me.  My inclination would be to keep the `"data":` field with its own braces.  It's not that many more characters, increases flexibility, and reduces error checking/recovery.

PR #3773, opened 2021-02-18
Don't try to delete from tx table if it's not in use.

mtrippled:
### Type of Change

- [ ] Documentation Updates
- [ ] Release

### DIFF
--- a/src/ripple/app/ledger/impl/LedgerMaster.cpp
+++ b/src/ripple/app/ledger/impl/LedgerMaster.cpp
@@ -2344,8 +2344,14 @@ LedgerMaster::getFetchPackCacheSize() const
 boost::optional<LedgerIndex>
 LedgerMaster::minSqlSeq()
 {
+    if (!app_.config().reporting())
+    {
+        boost::optional<LedgerIndex> seq;
+        auto db = app_.getLedgerDB().checkoutDb();
+        *db << "SELECT MIN(LedgerSeq) FROM Ledgers", soci::into(seq);
+        return seq;
+    }
 #ifdef RIPPLED_REPORTING
-    if (app_.config().reporting())
     {
         auto seq = PgQuery(app_.getPgPool())("SELECT min_ledger()");
         if (!seq)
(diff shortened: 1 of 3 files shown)


src/ripple/app/ledger/impl/LedgerMaster.cpp
```diff
@@ -2344,9 +2344,9 @@ LedgerMaster::getFetchPackCacheSize() const
 boost::optional<LedgerIndex>
 LedgerMaster::minSqlSeq()
 {
-#ifdef RIPPLED_REPORTING
     if (app_.config().reporting())
     {
+#ifdef RIPPLED_REPORTING
```
miguelportilla, line 2349: Do we report or log an error if reporting is enabled through configuration and the build does not support it?

cjcobb23: It actually only throws an error in this situation if you are also using Cassandra as a backend. Should probably change that. The best place to detect this and throw would be in either the `ReportingETL` constructor or `ReportingETL::run()`

cjcobb23 (approved):
Nice catch!

src/ripple/app/misc/SHAMapStoreImp.cpp
```diff
@@ -695,6 +696,9 @@ SHAMapStoreImp::clearPrior(LedgerIndex lastRotated)
     if (health())
         return;
 
+    if (! app_.config().useTxTables())
```
miguelportilla, line 699: Can also be
```if (!transactionDb_)``` or ```if (transactionDb == nullptr)```

miguelportilla (approved):
Left some minor nits but not holding back as is. 👍

cjcobb23:
@mtrippled Here is a commit that you can cherry pick to fix the compilation issue: https://github.com/cjcobb23/rippled/commit/f111f61e898dabc97024b3f1cc8933ba42c793b6

It also adds some additional checks. It checks that the code as built with `-Dreporting=1` if trying to run in reporting mode, and also checks that `use_tx_tables=1` (or is not specified) when running in reporting mode, since reporting mode always writes to the transaction tables.

ximinez:
@mtrippled Another compilation issue in certain MSVC builds is fixed by this commit: https://github.com/ximinez/rippled/commit/07fc0425a713f77e0644dc89564e39f28c885cd6

PR #3778, opened 2021-02-23
Improve the Byzantine validator detector

nbougalis:
This commit expands the detection capabilities of the Byzantine validation detector originally introduced with commit 381606aba2efcd179517f2497dc6e8c2770c679e. Prior to this change only validators that were on a server's UNL were monitored. Now, all the validations that a server receives are passed through the detector.

### DIFF
--- a/src/ripple/app/consensus/RCLConsensus.h
+++ b/src/ripple/app/consensus/RCLConsensus.h
@@ -63,9 +63,8 @@ class RCLConsensus
         InboundTransactions& inboundTransactions_;
         beast::Journal const j_;
 
-        NodeID const nodeID_;
-        PublicKey const valPublic_;
-        SecretKey const valSecret_;
+        // If the server is validating, the necessary keying information:
+        ValidatorKeys const& validatorKeys_;
 
         // A randomly selected non-zero value used to tag our validations
         std::uint64_t const valCookie_;
@@ -386,7 +385,7 @@ class RCLConsensus
             @param failedTxs Populate with transactions that we could not
                              successfully apply.
             @return The newly built ledger
-      */
+        */
         RCLCxLedger
         buildLCL(
             RCLCxLedger const& previousLedger,
--- a/src/ripple/app/consensus/RCLValidations.cpp
+++ b/src/ripple/app/consensus/RCLValidations.cpp
@@ -154,11 +154,13 @@ handleNewValidation(
     std::shared_ptr<STValidation> const& val,
     std::string const& source)
 {
-    PublicKey const& signingKey = val->getSignerPublic();
-    uint256 const& hash = val->getLedgerHash();
+    auto const& signingKey = val->getSignerPublic();
+    auto const& hash = val->getLedgerHash();
+    auto const seq = val->getFieldU32(sfLedgerSequence);
 
     // Ensure validation is marked as trusted if signer currently trusted
     auto masterKey = app.validators().getTrustedKey(signingKey);
+
     if (!val->isTrusted() && masterKey)
         val->setTrusted();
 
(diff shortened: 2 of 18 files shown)


seelabs:
@nbougalis A couple of the validation tests and manifest tests are failing:
```
21> #14 failed: Manifest_test.cpp(990)
21> #15 failed: Manifest_test.cpp(993)
21> #15 failed: Manifest_test.cpp(481)
8> #6 failed: Validations_test.cpp(306)
8> #7 failed: Validations_test.cpp(309)
```

I looked at the first one. It's now returning `badEphemeralKey`, which is correct. Looks like we just need to update the test. I didn't look at the others.

src/ripple/app/misc/NetworkOPs.cpp
```diff
@@ -2506,9 +2507,6 @@ NetworkOPsImp::recvValidation(
     std::shared_ptr<STValidation> const& val,
     std::string const& source)
 {
-    JLOG(m_journal.debug())
-        << "recvValidation " << val->getLedgerHash() << " from " << source;
```
gregtatcam, line 2510: I think this is useful for debugging, especially with validation/proposal reduce-relay feature enabled. Maybe just change it to `trace` to reduce the noise?

src/ripple/app/consensus/RCLConsensus.h
```diff
@@ -63,9 +63,8 @@ class RCLConsensus
         InboundTransactions& inboundTransactions_;
         beast::Journal const j_;
 
-        NodeID const nodeID_;
-        PublicKey const valPublic_;
-        SecretKey const valSecret_;
+        // If the server is validating, the necessary keying information:
+        ValidatorKeys const& validatorKeys_;
```
seelabs, line 67: The advantage of the old code is we don't have to worry about lifetime issues. I code is correct - the value that is passed in lives in the application object, which should outlive this. Still, it makes the code a little more fragile.

I probably weakly prefer the old code, but fine as-is as well.

src/ripple/overlay/impl/PeerImp.cpp
```diff
@@ -2240,17 +2240,17 @@ PeerImp::onMessage(
...
     try
     {
+        auto const closeTime = app_.timeKeeper().closeTime();
+
+        if (m->validation().size() < 50)
+        {
+            JLOG(p_journal_.warn()) << "Validation: Too small";
+            fee_ = Resource::feeInvalidRequest;
+            return;
+        }
```
seelabs, line 2252: Is there a reason this was moved inside the try block? I don't believe `closeTime` or the other functions can throw.

I don't object, but in this same file we move some code outside of a `try` block. I'm not sure the reason this was was moved in while the other one was moved out.

src/ripple/overlay/impl/PeerImp.cpp
```diff
@@ -2683,16 +2683,16 @@ PeerImp::checkValidation(
...
-        if (!cluster() && !val->isValid())
-        {
-            JLOG(p_journal_.warn()) << "Validation is invalid";
-            charge(Resource::feeInvalidRequest);
-            return;
-        }
+        JLOG(p_journal_.debug()) << "Validation forwarded by peer is invalid";
+        charge(Resource::feeInvalidRequest);
+        return;
+    }
```
seelabs, line 2691: See comment above: I'm not sure why this was moved outside the `try` block while the other was moved in.

seelabs (commented):
Looks good. Left some minor comments. Will wait until the unit tests pass before I give a thumbs up.

src/ripple/app/consensus/RCLValidations.cpp
```diff
@@ -154,68 +154,66 @@ handleNewValidation(
...
-        s << (val->isTrusted() ? "trusted" : "untrusted") << " "
-          << (val->isFull() ? "full" : "partial") << " validation: " << hash
-          << " from " << id << " via " << source << ": " << msg << "\n"
-          << " [" << val->getSerializer().slice() << "]";
-    };
+    auto& validations = app.getValidations();
 
     // masterKey is seated only if validator is trusted or listed
-    if (masterKey)
+    auto const outcome =
```
bachase, line 174: It's been a while since I've looked at this, can you clarify this comment being unchanged?

> masterKey is seated only if validator is trusted or listed

Your comment https://github.com/ripple/rippled/pull/3778#issue-578184333 suggested we only previously checked trusted validations.

The

gregtatcam (commented):
It looks good to me. I'll wait for the unittest fixes to approve.

src/ripple/app/misc/NetworkOPs.cpp
nbougalis: :+1:

src/ripple/app/consensus/RCLConsensus.h
nbougalis: The old code makes it more difficult to 're-key' a server (e.g. by allowing it to switch its manifest) although that's not something currently supported. It also dumps keying material in whatever random memory location the `this` exists in which makes it more difficult to "secure" the memory for sensitive keying material (e.g. by marking it as "do not include in core dump" and by other similar, high-security techniques, including, but not limited do storing the key itself inside an HSM).

I understand the concern here, but the code is structured so that the object being passed in a singleton. I guess we could _reinforce_ that by passing a `std::shared_ptr` or somesuch.

src/ripple/overlay/impl/PeerImp.cpp
nbougalis: No, it can be moved out.

src/ripple/app/consensus/RCLValidations.cpp
nbougalis: This is somewhat confusing, and I understand why you'd call this out. The comment is still accurate (sorta-kinda): we have a validation signed with some key K which _may_ be an ephemeral key or it may be a "direct" key (i.e. set via `[validation_seed]`). If it's signed by a manifest **AND** we know the manifest, we want to retrieve the master key and use that, otherwise, we use the ephemeral key.

Operating with "direct" keys (i.e. without a manifest) should be deprecated and, eventually, not allowed. And, if we receive a validation for an ephemeral key whose manifest the server does not have, we ought to just ignore it, especially post 1.7 that fixed manifest propagation.

bachase (approved):
LGTM (assuming other comments on unit tests are addressed).

seelabs (approved):
:+1:

PR #3780, opened 2021-02-24
Update to use Ripple Artifactory for docker cache

northrup:
This updates the build process to use the local Artifactory server as a docker image cache to avoid being rate limited by docker hub during the build process.


MarkusTeufelberger:
Just as a hint: You can configure Artifactory and Docker such that your Gitlab runner will pull from Artifactory instead of dockerhub using the original name (e.g. `centos:7`) while Artifactory acts as a cache in the middle.

This leaks the existence of your internal systems to the public and makes it harder for others to use this repository on their own Gitlab instance for example.

manojsdoshi (approved):
LGTM

PR #3785, opened 2021-03-05
Remove legacy protocol version (1.2)

alloynetworks:
This pull request removes support for the legacy protocol. It has been observed that nodes running versions as low as 0.60 occupy peer slots. From the perspective of hubs, not accepting connections from legacy protocol peers, will free up both slots and bandwidth consumption.

### DIFF
--- a/src/ripple/overlay/README.md
+++ b/src/ripple/overlay/README.md
@@ -152,8 +152,7 @@ appropriate HTTP error code (e.g. by sending an HTTP 400 "Bad Request" response)
 
 Protocol versions are string of the form `XRPL/` followed by a dotted major
 and minor protocol version number, where the major number is greater than or
-equal to 2 and the minor is greater than or equal to 0. The legacy version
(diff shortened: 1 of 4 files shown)


MarkusTeufelberger:
I suspect that there's more to remove than just these identifying strings if the protocol actually is to be deprecated?

nbougalis:
Since 1.5.0 the code has supported `XRPL/2.0` and treated `RTXP/1.2` as a legacy shortcut for that.

The removal of the legacy handling means that pre-1.5.0 servers won't be able to connect, but such servers would be amendment blocked anyways, so it's fine.

This looks good to me 👍

nbougalis (approved):

nbougalis (approved):

troylougheed:
Unsubscribe￼

nbougalis:
@alloynetworks, thanks for this submission. I fixed a couple of minor issues caused by a PR that was merged, migrating us from `boost::optional` to `std::optional`, and squashed my fixes into yours.

PR #3786, opened 2021-03-05
Create comparator wrapper classes for MSVC 2019 build issues:

ximinez:

Travis CI builds started failing for Visual Studio 2019 builds. [Example 1](https://travis-ci.com/github/ripple/rippled/builds/219024792), [example 2](https://travis-ci.com/github/ripple/rippled/builds/219024792).

@scottschurr , @miguelportilla , and I tracked the problem down to a recent update to VS 2019. The unaltered current develop branch fails with this version, so it is definitely not due to changes in rippled code.

The underlying cause is that the `std::less` and `std::equal_to` comparators have added the `[[nodiscard]]` attribute, which conflicts with `boost::bimap` type validations, which call the functions without capturing the result.

This fix adds classes to exactly wrap the comparators, removing `[[nodiscard]]` attributes.

### Type of Change

## Future Tasks

The new wrapper classes can be removed when either:
* A Boost update fixes this issue.
* A VS 2019 update fixes this issue.

---
This change is [Reviewable](https://reviewable.io/reviews/ripple/rippled/3786)

### DIFF
--- a/src/ripple/basics/comparators.h
+++ b/src/ripple/basics/comparators.h
@@ -1,0 +1,76 @@
+//------------------------------------------------------------------------------
+/*
+    This file is part of rippled: https://github.com/ripple/rippled
+    Copyright (c) 2021 Ripple Labs Inc.
+
+    Permission to use, copy, modify, and/or distribute this software for any
+    purpose  with  or without fee is hereby granted, provided that the above
+    copyright notice and this permission notice appear in all copies.
+
+    THE  SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+    WITH  REGARD  TO  THIS  SOFTWARE  INCLUDING  ALL  IMPLIED  WARRANTIES  OF
+    MERCHANTABILITY  AND  FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+    ANY  SPECIAL ,  DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+    WHATSOEVER  RESULTING  FROM  LOSS  OF USE, DATA OR PROFITS, WHETHER IN AN
+    ACTION  OF  CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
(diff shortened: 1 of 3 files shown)


src/test/csf/ledgers.h
```diff
@@ -21,6 +21,7 @@
 
 #include <ripple/basics/UnorderedContainers.h>
 #include <ripple/basics/chrono.h>
+#include <ripple/basics/comparitors.h>
```
scottschurr, line 24: I think this is misspelled?  "comparators.h"?

scottschurr:
Also, the commit message misspells "comparators" I think.  That would be good to fix.

ximinez:
> Also, the commit message misspells "comparators" I think. That would be good to fix.

:man_facepalming: Will fix.

ximinez:
I fixed the typo in a `FOLD` commit, and rewrote history to fix the commit message, but didn't change any of the code in that commit.

scottschurr:
:+1: Looks great!

scottschurr (approved):

src/ripple/basics/comparators.h
```diff
@@ -0,0 +1,64 @@
+//------------------------------------------------------------------------------
+/*
+    This file is part of rippled: https://github.com/ripple/rippled
+    Copyright (c) 2012, 2013 Ripple Labs Inc.
```
miguelportilla, line 4: Copyright (c) 2021 Ripple Labs Inc.

src/ripple/basics/comparators.h
```diff
@@ -0,0 +1,64 @@
...
+    constexpr bool
+    operator()(const T& left, const T& right) const
+    {
+        return std::equal_to<T>()(left, right);
+    }
+};
+
+}  // namespace ripple
+
+#endif
```
miguelportilla: Missing carriage return

src/ripple/basics/comparators.h
ximinez: :facepalm: Fixed.

src/ripple/basics/comparators.h
ximinez: Fixed

miguelportilla (approved):
👍 LGTM

ximinez:
Squashed and rebased onto 1.8.0-b1

PR #3788, opened 2021-03-10
Discard non-validator TMSquelch message

gregtatcam:

Discard TMSquelch message with a non-validator public key

### Context of Change

Introduced in 1.7.0.

### Type of Change

### DIFF
--- a/src/ripple/overlay/impl/PeerImp.cpp
+++ b/src/ripple/overlay/impl/PeerImp.cpp
@@ -2489,6 +2489,16 @@ PeerImp::onMessage(std::shared_ptr<protocol::TMSquelch> const& m)
     }
     PublicKey key(slice);
 
+    // Ignore non-validator squelch
+    if (!app_.validators().listed(key))
+    {
+        charge(Resource::feeBadData);
+        JLOG(p_journal_.debug())
+            << "onMessage: TMSquelch discarding non-validator squelch "
+            << slice;
+        return;
+    }
+
     // Ignore the squelch for validator's own messages.
     if (key == app_.getValidationPublicKey())
     {


src/ripple/overlay/impl/PeerImp.cpp
```diff
@@ -2489,6 +2489,16 @@ PeerImp::onMessage(std::shared_ptr<protocol::TMSquelch> const& m)
     }
     PublicKey key(slice);
 
+    // Ignore non-validator squelch
+    if (!app_.validators().listed(key))
```
pwang200, line 2493: If the key was listed but recently removed from manifests, one peer may have the key removed already and the other may still have the key, for a short time. So a legit unsquelch could be filtered out, but that is OK because the squelch will expire eventually. Right?
How about the case that a key is recently added, and a legit squelch message is filtered out. That should also be fine, right? Because the correctness of the algorithm should not depend on connected peers' behavior. It should only affect performance. (Sorry, the algorithm is a bit fuggy to me to work out exactly what happens.) 
There are other similar cases due to not completely overlapping sets of listed keys. But the same argument applies, right?

gregtatcam: Ultimately the validators list is going to be synced up between a node and its peers. In the meantime, if a peer gets the squelch message for an unlisted validator it'll ignore the squelch and relay messages for this validator anyways. But the node will ignore these messages until the squelch is expired. So the only drawback is that the messages are relayed until the list is synced up but I don't think it'll happen too often.

pwang200: Agree.

pwang200 (approved):
:+1: LGTM

Hope PeerImp.cpp could be unit tested.

PR #3791, opened 2021-03-10
Serialize metadata as blob in tx RPC when binary flag is true

cjcobb23:
In the `tx` command, transaction metadata is always returned in expanded JSON form, even if `binary` is set to true in the request. The transaction itself is encoded as a blob if `binary` is set to true, just not the metadata. This PR fixes that.

---
This change is [Reviewable](https://reviewable.io/reviews/ripple/rippled/3791)

### DIFF
--- a/src/ripple/rpc/handlers/Tx.cpp
+++ b/src/ripple/rpc/handlers/Tx.cpp
@@ -236,7 +236,14 @@ doTxHelp(RPC::Context& context, TxArgs const& args)
 
     if (ledger && meta)
     {
-        result.meta = meta;
+        if (args.binary)
+        {
+            result.meta = meta->getAsObject().getSerializer().getData();
+        }
+        else
+        {
+            result.meta = meta;
+        }
         result.validated = isValidated(
             context.ledgerMaster, ledger->info().seq, ledger->info().hash);
     }
--- a/src/test/rpc/Transaction_test.cpp
+++ b/src/test/rpc/Transaction_test.cpp
@@ -49,18 +49,23 @@ class Transaction_test : public beast::unit_test::suite
         env.close();
 
         std::vector<std::shared_ptr<STTx const>> txns;
+        std::vector<std::shared_ptr<STObject const>> metas;
         auto const startLegSeq = env.current()->info().seq;
         for (int i = 0; i < 750; ++i)
         {
             env(noop(alice));
             txns.emplace_back(env.tx());
             env.close();
+            metas.emplace_back(
+                env.closed()->txRead(env.tx()->getTransactionID()).second);
         }
         auto const endLegSeq = env.closed()->info().seq;
 
         // Find the existing transactions
-        for (auto&& tx : txns)
+        for (size_t i = 0; i < txns.size(); ++i)
         {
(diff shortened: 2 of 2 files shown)


mDuo13:
Just confirmed that this is a regression in 1.7.0. The command works as expected in 1.6.0.

mDuo13:
This PR should also add unit tests for this so it doesn't happen again.

scottschurr:
I spent some time looking at history.  If I'm reading the history correctly, the `binary` flag caused the metadata to be returned in binary until 1.5.0-b6.  So, since then, the metadata has been returned as JSON regardless of the state of the `binary` flag.

I'm of two minds.  When the bug was released to the wild with 1.5.0 nobody noticed.  So apparently no one was using the feature a year ago.  A year has passed, and _probably_ no one is relying on the changed behavior?  But we don't know for sure until we change the behavior back and potentially break someone.

If we put this change in now then the bug will be resident in released software versions for approximately 1.5 years before we switch it back.

The best path forward is not clear to me.  There are advantages and disadvantages to leaving the bug in place or fixing it.

scottschurr:
Oh! Thanks @mDuo13.  A test is better than me mis-reading the code.

 Since the behavior was correct in 1.6.0, I think the best choice is to release the fix in a 1.7.1.  That way we don't have 6 months of users getting the wrong interface and having it switched back at the end of it.

And, yes, I agree with @mDuo, the final version of this commit wants a unit test to verify the behavior.

cjcobb23:
I'll add some unit tests, hopefully tomorrow. If we had those unit tests before, we could have caught this during review. So we definitely need the unit tests.

natenichols:
Looks like this bug is on me, here's where the regression occurred (#3564).

Will approve after unit tests are added, and I defer to y'all about the best path forward.

cjcobb23:
> Looks like this bug is my fault, here's where the regression occurred (#3564).
> 
> Will approve after unit tests are added, and I defer to y'all about the best path forward.

I reviewed your original PR and didn't catch this, so also my fault.

thejohnfreeman (approved):

scottschurr (approved):
Looks great.  Just to build confidence, I back ported just the test to 1.6.0.  Your changed test passed with flying colors on 1.6.0.  So, yeah, I think you've nailed it.  :+1:

natenichols (approved):

intelliot (approved):

ximinez:
@natenichols @cjcobb23 Post-mortems are valuable to determine ways to prevent a problem from recurring; blame less so. As a group, we all share responsibility for what is released, from the coders to the formal reviewers to the informal reviewers to the testers to the mergers to the bosses to those who didn't look at all ("Eh, it's got two reviews. I'm sure it's fine."). The important things are that the problem got caught and fixed before it caused too many problems, and that unit tests are being added to catch similar mistakes in the future.

sublimator:
haha, I read that as " bosses who didn't look at all " somehow

>

ximinez:
>  to those who didn't look at all

I was specifically thinking about myself with this one, because there are _a lot_ of PRs I don't pay attention to. :grinning:

PR #3793, opened 2021-03-12
Remove obsolete URLs and references to Ripple

cryptobrad:

Remove the old comments about Hebbian learning and out dated information and broken links to old Ripple wiki. Its sad that this is needed since the name was taken by the Ripple company but we will clean it up
 

### Type of Change

- [X] Documentation Updates
- [ ] Release


nbougalis (approved):
Looks good. Some of the markdown files were outdated or not particularly helpful and it's ok that they're deleted but they could _probably_ have been salvaged.

In a perfect world, almost every folder would contain a markdown document with at least a small description.

src/ripple/protocol/impl/PublicKey.cpp
```diff
@@ -106,7 +106,7 @@ sliceToHex(Slice const& slice)
     @param strict_param whether to enforce strictly canonical semantics
 
     @note For more details please see:
-    https://ripple.com/wiki/Transaction_Malleability
+    https://xrpl.org/transaction-malleability.html#transaction-malleability
```
mDuo13, line 109: This link will be a little more stable if you don't include the anchor.

```suggestion
    https://xrpl.org/transaction-malleability.html
```

src/ripple/app/tx/impl/Escrow.cpp
```diff
@@ -58,22 +58,22 @@ namespace ripple {
     expiration time, the XRP can only be returned to the sender.
 
     For more details on escrow, including examples, diagrams and more please
-    visit https://ripple.com/build/escrow/#escrow
+    visit https://xrpl.org/escrow.html#escrow
```
mDuo13, line 61: This link will be more stable without the anchor.

```suggestion
    visit https://xrpl.org/escrow.html
```

src/ripple/app/paths/README.md
```diff
@@ -1,137 +0,0 @@
-Theory of Pathfinding
```
mDuo13, line 1: This document still has some kind-of useful information in it? You're right that it's extremely dated though.

src/ripple/app/paths/Pathfinder.cpp
```diff
@@ -444,8 +444,7 @@ Pathfinder::computePathRanks(int maxPaths)
 static bool
 isDefaultPath(STPath const& path)
 {
-    // TODO(tom): default paths can consist of more than just an account:
-    // https://forum.ripple.com/viewtopic.php?f=2&t=8206&start=10#p57713
+    // FIXME: default paths can consist of more than just an account:
```
mDuo13, line 447: For future reference, xrpl.org has some discussion of [default paths](https://xrpl.org/paths.html#default-paths) which may be useful for fixing this up.

No action needed here at this time though.

mDuo13 (approved):
Thank you for doing this. I left a couple suggestions (mostly to make the links more likely to continue to work if any changes happen on the docs side) but I think you can consider all of those optional nice-to-haves.

src/ripple/app/paths/README.md
nbougalis: I went through this and honestly, there's very little captured in that doc that is helpful. I think it's better off removed for now.

nbougalis:
@cryptobrad, I think @mDuo13's URL tweaks help, so I just used GitHub's web-based UI to commit suggestions to the PR. They should be squashed down when this commit is merged.

PR #3805, opened 2021-03-30
Adjust OpenSSL defaults and mitigate CVE-2021-3499:

nbougalis:

OpenSSL 1.1.1k was recently released and included a fix for **CVE-2021-3499** which allows a malicious client to crash a server using earlier versions of OpenSSL by initiating a renegotiation while in TLS 1.2 mode. For details, see the OpenSSL security advisory published at https://www.openssl.org/news/secadv/20210325.txt.

While it is always recommended to build and link against the most recent version of OpenSSL, this commit goes above and beyond to mitigate this issue and, potentially, other issues, by disabling renegotiation support entirely. Furthermore, is now requires the use of TLS 1.2 or later, by disabling legacy TLS 1.0 and TLS 1.1 support. Lastly, this commit also changes the default list of ciphers that the server offers, limiting it only to ciphers that are part of TLS 1.2.

## Before / After
TLS 1.0 and 1.1 connections were supported; attempts use them will now fail. Generally, both were already avoided by any relatively modern TLS implementation and are, at this point, effectively deprecated.
 
## Test Plan
Ensure that server-server connectivity continues to work and that clients are able to connect to WSS ports.

### DIFF
--- a/src/ripple/basics/impl/make_SSLContext.cpp
+++ b/src/ripple/basics/impl/make_SSLContext.cpp
@@ -29,13 +29,21 @@ namespace ripple {
 namespace openssl {
 namespace detail {
 
-// We limit the ciphers we request and allow to ensure that weak
-// ciphers aren't used. While this isn't strictly necessary for
-// the rippled server-server use case, where we only need MITM
-// detection/prevention, we also have websocket and rpc scenarios
-// and want to ensure weak ciphers can't be used.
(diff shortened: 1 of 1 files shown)


JoelKatz (approved):
LGTM. I don't see any compatibility issues.

aanchal4:
LGTM.

aanchal4 (approved):

PR #3813, opened 2021-04-07
Fix compilation of krb5 dependency with GCC 10 and Clang 10

thejohnfreeman:
[This is the problem](https://github.com/cockroachdb/cockroach/issues/49734#issuecomment-758389602) and [this is the fix](https://github.com/cockroachdb/cockroach/pull/58895/files).


cjcobb23 (approved):
Approving now, though we should make sure CI passes.

nbougalis:
This change has already been included in the codebase, as part of a2a37a928a3caf3e0e475294c6ed3e4e3dcc29c0, by @thejohnfreeman.

PR #3815, opened 2021-04-09
Fix crash in SQL interface in the case use_tx_tables = 0

cdy20:

This is small fix of crash in SQL interface in the case use_tx_tables = 0. At the initial version of fix we return default (empty/zero) values from SQL interface in the case if Tx tables are not used. May be in some cases we need to log error or throw an exception instead of that.

### Type of Change

- [ ] Documentation Updates
- [ ] Release


src/ripple/app/rdb/backend/RelationalDBInterfaceSqlite.cpp
```diff
@@ -173,6 +175,7 @@ class RelationalDBInterfaceSqliteImp : public RelationalDBInterfaceSqlite
 
 private:
     Application& app_;
+    bool useTxTables_;
```
seelabs, line 178: The issue is that the code assumes that at least one of the transaction db or the shard db will be present, and recent changes means that both databases may be absent.

Instead of a `useTxTables_` flag, what about structuring the code so the calls look like:
```
if (existsTransaction()){
   return withTxDb();
}
else if (existsShard()){
   return withShardsDb();
}
return default; // std::nullopt or somesuch
```

src/ripple/app/rdb/backend/RelationalDBInterfaceSqlite.cpp
```diff
@@ -417,6 +420,10 @@ RelationalDBInterfaceSqliteImp::getMinLedgerSeq()
 std::optional<LedgerIndex>
 RelationalDBInterfaceSqliteImp::getTransactionsMinLedgerSeq()
 {
+    /* if Tx tables not used, return empty value */
+    if (!useTxTables_)
+        return {};
```
seelabs, line 425: Doesn't the `getMinLedgerSeq` on line 402 need this check too? Same with `getMaxLedgerSeq`?

seelabs (commented):
Asked two quick questions before I do a complete review.

src/ripple/app/rdb/backend/RelationalDBInterfaceSqlite.cpp
cdy20: I don't know what could be implementation of existsShard(). Each request to shards as a rule will touch several shards databases. Also need to take into account what means option use_tx_tables = 0. Is it prohibit use of node transaction databases only, or overall transactions databases which includes shard transaction databases too. In the last case my implementation looks more correct.

src/ripple/app/rdb/backend/RelationalDBInterfaceSqlite.cpp
cdy20: No. On line 402 and getMaxLedgerSeq() we look for an answer in ledger databases, but in line 423 and similar we look it in transaction databases.

src/ripple/app/rdb/backend/RelationalDBInterfaceSqlite.cpp
seelabs: I thought `existsShard()` was just `app_.getShardStore() != nullptr`. But you're right, we need to take into account what `use_tx_tables = 0` means, and I agree your implementation looks more correct. Let's keep yours.

src/ripple/app/rdb/backend/RelationalDBInterfaceSqlite.cpp
seelabs: Yes, you're right. I got confused with the ledger database and the transaction database.

src/ripple/app/rdb/backend/RelationalDBInterfaceSqlite.cpp
```diff
@@ -1315,6 +1386,10 @@ RelationalDBInterfaceSqliteImp::getTransaction(
     std::optional<ClosedInterval<uint32_t>> const& range,
     error_code_i& ec)
 {
+    /* if Tx tables not used, return empty value */
```
seelabs, line 1389: Comment isn't quite right. It returns "unknown", not an empty value.

seelabs (approved):
:+1: LGTM

src/ripple/app/rdb/backend/RelationalDBInterfaceSqlite.cpp
mtrippled: This is what use_tx_tables = 0 means (from cfg/rippled-example.cfg):
#      use_tx_tables        Valid values: 1, 0
#                           The default is 1 (true). Determines whether to use
#                           the SQLite transaction database. If set to 0,
#                           rippled will not write to the transaction database,
#                           and will reject tx, account_tx and tx_history RPCs.
#                           In Reporting Mode, this setting is ignored.

It doesn't stop the use of the shard transaction databases though.

seelabs: I'll defer to @mtrippled and @cjcobb23 as to what the correct behaviour should be for `use_tx_tables = 0` for this class, since they are more familiar with how that switch is mean to be used.

cjcobb23: The correct behavior should be to not write to the transaction databases at all if use_tx_tables is 0. This includes the node transaction database and the shard transaction database.

cdy20: In second patch of changes I based on the last point of view. But for alternative point it need to change one line in the code.

src/ripple/app/rdb/backend/RelationalDBInterfaceSqlite.cpp
cdy20: Ok

src/ripple/app/rdb/backend/RelationalDBInterfaceSqlite.cpp
cdy20: Done

src/ripple/app/rdb/backend/RelationalDBInterfaceSqlite.cpp
```diff
@@ -884,38 +1035,44 @@ RelationalDBInterfaceSqliteImp::getOldestAccountTxs(
...
+                    limit_used += total;
                     opt.offset = 0;
+                }
                 else
-                    opt.offset -= total;
-            }
-            return true;
-        });
+                {
+                    total = ~total;
```
seelabs, line 1062: Bitwise invert? That looks like a bug. If it's not a bug it definitely needs a comment (here and the other places doing `total = ~total`.

src/ripple/app/rdb/backend/RelationalDBInterfaceSqlite.cpp
```diff
@@ -884,38 +1035,44 @@ RelationalDBInterfaceSqliteImp::getOldestAccountTxs(
...
+                        opt.offset -= total;
+                }
+                return true;
+            });
+
+        return ret;
+    }
 
-    return ret;
+    /* else return empty value */
```
seelabs, line 1074: No change needed, but IMO comments like this actually detract from the code. It's easy to see that it returns an empty value, there's no need to point it out with a comment. The problem with comments like this is it adds "noise" to the code. We want comments to have a purpose.

I normally wouldn't point out a single instance of a comment like this, but there are many such comments in this file. I'm OK leaving them in (it's a minor nit), but it's something to keep in mind for future code.

seelabs (commented):
Left a question about bitwise invert, otherwise looks good.

cjcobb23 (approved):
LGTM

src/ripple/app/rdb/backend/RelationalDBInterfaceSqlite.cpp
cdy20: Added comments on it.

src/ripple/app/rdb/backend/RelationalDBInterfaceSqlite.cpp
cdy20: Did not removed such a comments because they finished if-else comment constructions:
/* if A do this */
...
/* else, if B, do that */
...
/* else return default */

src/ripple/app/rdb/backend/RelationalDBInterfaceSqlite.cpp
```diff
@@ -243,6 +246,26 @@ class RelationalDBInterfaceSqliteImp : public RelationalDBInterfaceSqlite
         return !!txdb_;
```
mtrippled, line 246: Nit: this idiom is not necessary hurts readability. The unque_ptr bool operator returns whether there's a managed object or not: http://www.cplusplus.com/reference/memory/unique_ptr/operator%20bool/

mtrippled (approved):
LGTM :+1:
I left a nit to be optionally fixed.

src/ripple/app/rdb/backend/RelationalDBInterfaceSqlite.cpp
seelabs: Thanks for adding the comment, but that still doesn't clarify why it was done like this.I would have expected `-number` not `~number`.

It looks in all these function, the code looks like:
```
        if (!total && limit_used)
        {
             // ....
             total = ~total
        }
```
If that's true, then if we used negative instead of bitwise invert, then `0` means none processed and none skipped, positive means num processed, and negative means num skipped. Is there a reason `~` was chosen over `-`?

The only reason I could think of, is if we needed to distinguish between 0 processed and 0 skipped, but that doesn't appear to be the case. 

This isn't part of this PR, so no change needed, tho I'd still like to understand why `~` was chosen.

src/ripple/app/rdb/backend/RelationalDBInterfaceSqlite.cpp
seelabs: Right, the code is structured like this:
```
    /* if databases exists, use it */
    if (existsTransaction())
    /* else use shard databases, if available */
    if (existsShardTransaction())
    /* else return empty value */
    return {};
```
Those comments don't really serve a purpose. Like I said in the previous comment, I don't mind leaving them, but please keep it in mind for future code.

seelabs (approved):
I'm approving this, but please see my question on using bitwise inverse.

src/ripple/app/rdb/backend/RelationalDBInterfaceSqlite.cpp
cdy20: Done.

cdy20:
> I'm approving this, but please see my question on using bitwise inverse.

Изменено ~ на -. Также была исправлена ошибка для случая, когда какой-то шард не содержал транзакции для данной учетной записи. В предыдущей версии мы завершали поиск, но теперь продолжаем поиск со следующим шардом. Однако часть интерфейса SQL, связанная с шардом, может содержать много других ошибок, поскольку она не была протестирована.

mtrippled:
> > I'm approving this, but please see my question on using bitwise inverse.
> 
> Changed ~ to -. Also a bug was fixed for the case where some shard do not contain any transaction for given account. In the previous version we finish search, but now we continue search with next shard. However the shard-related part of SQL interface may contain many another bugs because it was not tested.

If this is the case, then I believe that this PR should not be merged. Each implementer and reviewer must be confident that we're not introducing bugs. Therefore, I suggest that either:
1) Reduce the scope so that the original issue is the only thing being addressed: the need to support no transaction tables in non-shard mode, or
2) Test the current PR to cover the changes that are potentially buggy.

mtrippled (approved):
:+1: LGTM

Though I approved already, oh well.

nbougalis:
This PR has been superseded by #3888

PR #3817, opened 2021-04-09
Add load_factor to server_info in reporting mode

cjcobb23:
load_factor was missing from server_info when the server was running in reporting mode. This commit fixes that by querying the p2p node for the load_factor, and propagating that info back to the client. If the p2p node is unavailable or returns an error, load_factor is set to 1.

---
This change is [Reviewable](https://reviewable.io/reviews/ripple/rippled/3817)

### DIFF
--- a/src/ripple/rpc/handlers/ServerInfo.cpp
+++ b/src/ripple/rpc/handlers/ServerInfo.cpp
@@ -17,7 +17,9 @@
 */
 //==============================================================================
 
+#include <ripple/app/main/Application.h>
 #include <ripple/app/misc/NetworkOPs.h>
+#include <ripple/app/reporting/P2pProxy.h>
 #include <ripple/json/json_value.h>
 #include <ripple/net/RPCErr.h>
 #include <ripple/protocol/jss.h>
@@ -38,6 +40,15 @@ doServerInfo(RPC::JsonContext& context)
         context.params.isMember(jss::counters) &&
             context.params[jss::counters].asBool());
 
+    if (context.app.config().reporting())
+    {
+        Json::Value proxied = forwardToP2p(context);
+        if (proxied.isMember(jss::result) &&
+            proxied[jss::result].isMember(jss::info) &&
+            proxied[jss::result][jss::info].isMember(jss::load_factor))
+            ret[jss::info][jss::load_factor] =
+                proxied[jss::result][jss::info][jss::load_factor];
+    }
     return ret;
 }
 
--- a/src/ripple/rpc/handlers/ServerInfo.cpp
+++ b/src/ripple/rpc/handlers/ServerInfo.cpp
@@ -42,14 +42,9 @@ doServerInfo(RPC::JsonContext& context)
 
     if (context.app.config().reporting())
     {
-        Json::Value proxied = forwardToP2p(context);
-        if (proxied.isMember(jss::result) &&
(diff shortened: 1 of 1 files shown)


cjcobb23:
@ximinez Do you think we should set `load_factor` to 1 if the p2p node is unavailable?

ximinez:
> Do you think we should set `load_factor` to 1 if the p2p node is unavailable?

Yeah, I think that would be smart, and better than an error in this case.

src/ripple/rpc/handlers/ServerInfo.cpp
```diff
@@ -38,6 +40,17 @@ doServerInfo(RPC::JsonContext& context)
         context.params.isMember(jss::counters) &&
             context.params[jss::counters].asBool());
 
+    if (context.app.config().reporting())
```
mtrippled, line 43: I think this can be simplified into a try ... catch such as:
try:
    ret[field] = proxied[result][jssinfo][load_factor]
catch:
  ret[field] = 1

mtrippled (approved):
LGTM :+1:
Left an optional suggestion if it makes sense to you.

src/ripple/rpc/handlers/ServerInfo.cpp
```diff
@@ -38,6 +40,17 @@ doServerInfo(RPC::JsonContext& context)
...
+    {
+        Json::Value proxied = forwardToP2p(context);
+        if (proxied.isMember(jss::result) &&
+            proxied[jss::result].isMember(jss::info) &&
+            proxied[jss::result][jss::info].isMember(jss::load_factor))
+            ret[jss::info][jss::load_factor] =
+                proxied[jss::result][jss::info][jss::load_factor];
+        else
+            ret[jss::info][jss::load_factor] = 1;
```
ximinez, line 52: Is this testable? Can you add or update a unit test for it?

ximinez (changes requested):

src/ripple/rpc/handlers/ServerInfo.cpp
cjcobb23: No this is not until testable. Proxying requires running two servers and should be tested in integration/QA.

mtrippled: Not really testable without adding some type of virtual proxy. This functionality (reporting) will be split into its own code base, so I think this stuff will be removed from rippled itself in that project.

cjcobb23:
@ximinez What changes are you requesting on this? I see your suggestion about the try-catch and the unit test comment. I don't think this change is unit testable, and I interpreted the try-catch comment as just a suggestion. Let me know what you actually want me to change here.

ximinez:
> What changes are you requesting on this?

I try to mark significant suggestions as "changes requested" so I know I need to follow-up after a change is made, or there's a discussion to not do it.

cjcobb23:
> > What changes are you requesting on this?
> 
> I try to mark significant suggestions as "changes requested" so I know I need to follow-up after a change is made, or there's a discussion to not do it.
> 

Got it, thanks for clarifying

natenichols (approved):
LGTM 👍

ximinez (approved):
Looks good!

MarkusTeufelberger:
I don't use reporting mode, but I assume this is already in `server_status` (the machine-readable version of `server_info`)?

PR #3819, opened 2021-04-11
Support boost 1.76:

seelabs:
* remove unused `std::hash` specialization
* Use `std::chrono::floor` instead of `floor`

This patch was tested against boost 1.76-rc1. We should wait to merge it until boost 1.76 is officially released.


HowardHinnant:
Waiting for release of 1.76 to test.

HowardHinnant (changes requested):
Looks good, but needs a `#include <vector>` here: src/test/unit_test/multi_runner.cpp

seelabs:
Thanks @HowardHinnant. Added missing include in the patch `c1c5fb24ea Add missing include`

seelabs:
Squashed.

PR #3820, opened 2021-04-12
Automatically determine the node size:

nbougalis:
Allows the `node_size` parameter to be automatically adjusted based on the hardware that the server is running on.

### Context of Change

The `[node_size]` configuration parameter is used to tune various parameters based on the hardware that the code is running on. The parameter can take five distinct values: `tiny`, `small`, `medium`, `large` and `huge`.

The default value in the code is `tiny` but the default configuration file sets the value to `medium`. This commit attempts to detect the amount of RAM on the system and adjusts the node size default value based on the amount of RAM and the number of hardware execution threads on the system.

The decision matrix currently used is:

|         |   1  | 2 or 3 |   ≥ 4  |
|:-------:|:----:|:------:|:------:|
|  < ~8GB | tiny |   tiny |   tiny |
| < ~12GB | tiny |  small |  small |
| < ~16GB | tiny |  small | medium |
| < ~24GB | tiny |  small |  large |
| < ~32GB | tiny |  small |   huge |

The node size will now be reported in the `server_info` API when invoked in 'admin' mode in the `node_size` field.

### Type of Change
- [X] Documentation Updates
- [ ] Release

### DIFF
--- a/cfg/rippled-example.cfg
+++ b/cfg/rippled-example.cfg
@@ -1426,8 +1426,20 @@
 #   Tunes the servers based on the expected load and available memory. Legal
 #   sizes are "tiny", "small", "medium", "large", and "huge". We recommend
 #   you start at the default and raise the setting if you have extra memory.
-#   If no value is specified, the code assumes the proper size is "tiny". The
-#   default configuration file explicitly specifies "medium" as the size.
+#
+#   The code attempts to automatically determine the appropriate size for
+#   this parameter based on the amount of RAM and the number of execution
+#   cores availabe to the server. The current decision matrix is:
+#
+#   |         |         Cores          |
+#   |---------|------------------------|
+#   |   RAM   |   1  | 2 or 3 |   ≥ 4  |
+#   |---------|------|--------|--------|
+#   |  < ~8GB | tiny |   tiny |   tiny |
+#   | < ~12GB | tiny |  small |  small |
+#   | < ~16GB | tiny |  small | medium |
+#   | < ~24GB | tiny |  small |  large |
+#   | < ~32GB | tiny |  small |   huge |
 #
 # [signing_support]
 #
@@ -1588,9 +1600,6 @@ protocol = ws
 #protocol = wss
 
 #-------------------------------------------------------------------------------
-
-[node_size]
-medium
 
 # This is primary persistent datastore for rippled.  This includes transaction
 # metadata, account states, and ledger headers.  Helpful information can be
--- a/src/ripple/app/main/Main.cpp
+++ b/src/ripple/app/main/Main.cpp
@@ -745,8 +745,6 @@ run(int argc, char** argv)
 
 }  // namespace ripple
 
-// Must be outside the namespace for obvious reasons
-//
 int
 main(int argc, char** argv)
 {
@@ -770,7 +768,5 @@ main(int argc, char** argv)
 
     atexit(&google::protobuf::ShutdownProtobufLibrary);
 
-    auto const result(ripple::run(argc, argv));
-
-    return result;
+    return ripple::run(argc, argv);
 }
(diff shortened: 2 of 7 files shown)


MarkusTeufelberger:
How does this work with external limits? E.g. I have a Kubernetes host with 128 GB RAM and 32 cores, but `rippled` runs in a container that's limited to 16 GB RAM and 3 CPU cores? Seems like it would scale up to `huge` and run straight against a wall... (this is a common issue when running "smart" Java software in containers btw. and happens more often than one might think).

I would prefer to have the individual knobs that these "sizes" modify made available and documented individually instead of having even more automatisms in there that can cause problems if they detect/interpret something wrong.

ximinez:
@MarkusTeufelberger I quickly skimmed the code changes, and it doesn't look like @nbougalis removed the ability to override the value in the config, though he did remove the value from the _example_ config file. This code looks like it's only determining the default value if you _don't_ specify it in the config.

MarkusTeufelberger:
It also removes the node_size from the default config.

nbougalis:
### TL;DR:
I do appreciate your feedback, @MarkusTeufelberger and if I could have tagged you as a reviewer, I would have. But I was confident you'd chime in. 

I understand you prefer more manual knobs instead of more automation, and if you feel strongly enough to veto this from being merged, then I'll withdraw it.

### Details
> It also removes the node_size from the default config.

Correct: it does. What's the point of having auto-detection of an appropriate<sup>†</sup> value, if you explicitly override it with a static, less-appropriate one? Beyond that, the _actual_ default value (`tiny`) was different from the value specified in the default config, which only added to the confusion.

Now, to reference the <sup>†</sup> above:

> How does this work with external limits? E.g. I have a Kubernetes host with 128 GB RAM and 32 cores, but `rippled` runs in a container that's limited to 16 GB RAM and 3 CPU cores? Seems like it would scale up to `huge` and run straight against a wall... (this is a common issue when running "smart" Java software in containers btw. and happens more often than one might think).

The auto-detection code is limited and can only make decisions based on the data that is provided to it by the environment it's running on. I'd argue that _if_ a container limits the RAM or CPU cores available to a containerized process, it should be the container's responsibility to provide those updated limits to the process, when asked, instead of requiring the process to use a dowsing rod.
 
I absolutely agree with you that this is a concern, and hence the option to _explicitly_ override the default value.

> I would prefer to have the individual knobs that these "sizes" modify made available and documented individually instead of having even more automatisms in there that can cause problems if they detect/interpret something wrong.

Me too: I think that splitting some of these values out and making them separate tunable options is a good idea… **for expert users** (like yourself!) but I think that having some kind of automatic tuning makes sense too.

You can argue that the automatic tuning doesn't do a good enough job. And I agree, it's not perfect, not by any stretch of the imagination. But it's probably better than a simple hard-coded default.

### Oh one more thing...
I know that in the past you'd manually adjusted some of those parameters, resulting in significant improvement in performance; I tried to find the relevant issue but couldn't conjure up the right combination of search terms. If you remember what your tweaks were, I'd love to take a look again.

MarkusTeufelberger:
> I'd argue that if a container limits the RAM or CPU cores available to a containerized process, it should be the container's responsibility to provide those updated limits to the process, when asked, instead of requiring the process to use a dowsing rod.

That's unfortunately(?) not how it works, depending on which APIs you call (didn't look too deeply into this patch which ones get used in the end).

I'm definitely not strictly against auto tuning, it just might make it harder to debug some cases than one might expect. For debugging it might be helpful to have these numbers (not just the node_size) in `server_info` and `server_state`(!).

I had to tune down `SizedItem::ledgerFetch` since it tended to overload my DSL line at home when fetching large packs in parallel. Everything else about this server is sized large enough for a `huge` node. This was especially an issue when syncing history, once it was mostly in sync it didn't matter that much any more. Currently my node is offline (storage ran low and I don't want to overload my local network segment during home office times) so I am not sure if that patch is still needed. In my experience this setting is far more sensitive to network throughput than anything else and that's a much harder target to hit (and measure) anyways. Installing 64 GB RAM doesn't unclog my connection or make my provider's modem faster unfortunately.

HowardHinnant (approved):

PR #3821, opened 2021-04-14
[Trivial] Future proof rippled by changing variable name 'requires' to 'require'

r0bertz:
'requires' is a keyword in C++20.

https://en.cppreference.com/w/cpp/keyword/requires

This is the least intrusive change. Another option is to rename the require class (and the associated variable name) to something else. If the latter is desired, I will make the change. Please let me know what you think. Thanks!

### DIFF
--- a/src/test/jtx/JTx.h
+++ b/src/test/jtx/JTx.h
@@ -42,7 +42,7 @@ class Env;
 struct JTx
 {
     Json::Value jv;
-    requires_t requires;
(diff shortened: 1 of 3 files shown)


ximinez (approved):
Thank you for noticing this issue, and your submission to resolve it. These changes look good to me.

scottschurr (approved):
:+1:

PR #3823, opened 2021-04-15
CheckCashMakesTrustLine amendment

scottschurr:

With this amendment, the `CheckCash` transaction creates a trust line if needed.  The change is modeled after `Offer` crossing.  And, similar to `Offer` crossing, cashing a `Check` allows an account to exceed its trust line limit.

### Context of Change

It has been determined that the user experience of needing to create a trust line prior to cashing a `Check` was both inconvenient and not necessary.  Similar to an `Offer`, anyone who cashes a `Check` clearly wants to accept the funds involved.  Automatically creating the trust line, if necessary, would be convenient and has precedent with `Offer` crossing.

Therefore this amendment introduces two changes to the behavior of cashing a `Check`:

1. If the account cashing the `Check` does not already have a trust line for the currency being received, then that trust line is created if possible.  It usually is possible.
 
    Reasons it might not be possible to create the trust line include:

    - The account cashing the check does not have sufficient XRP to meet the reserve required to create the trust line.
    - Rippling would be required to transfer the funds and the issuing account has not set the `lsfDefaultRipple` flag.
    - The account issuing the currency has set the `lsfGlobalFreeze` flag.
    - The account issuing the currency has set the `lsfRequireAuth` flag, so an automatically created trust line would not be authorized.

2. The currency being brought into the account by the `Check` is allowed to exceed any limit specified the in trust line.  There are three reasons for this:

    - If the account is explicitly cashing the `Check` then they must want the currency in question.
    - `Offer`s have the same behavior.  It makes sense to limit the number of unique behaviors on the ledger.
    - An automatically created trust line has limits of zero.  If we didn't have this rule then the automatically created trust line would be useless.

If cashing the `Check` fails for any reason, then the trust line is not added to the account.

### Type of Change

- [x] Breaking change (a feature that would cause existing functionality to not work as expected)
- [x] Tests (added unit tests)
- [x] Documentation updates will be required

If this change is merged then documentation will need to be updated to cover the behavior changes.  This change will also need to be mentioned in the release notes.

## Test Plan

Unit tests were added to verify that prior to the introduction of the amendment cashing a `Check` without a pre-existing trust line would fail.  After the introduction of the amendment cashing a `Check` without a pre-existing trust line succeeds by adding the necessary trust line to the ledger.

A number of different scenarios were tested including:

- Cashing a `Check` that was created by the issuer of the currency in question.
- Cashing a `Check` written by an account other than the issuer.
- A variety of `AccountRoot` flags were tested which either allowed or prevented the automatic trust line creation.
- Trust lines automatically created by `Offer` crossing were compared to trust lines automatically created by `Check` cashing and found to be very similar (as intended).

## Requests of Reviewers

Please consider situations I might have forgotten which might affect creation of a trust line.  I tested the account root flags I _thought_ would matter.  Did I miss important account root flags?  Are there conditions other than account root flags that need to be tested?  Thanks for the help.

## Future Tasks

Once a trust line is added to the ledger, it can be challenging for a ledger user to remove the trust line.  It may be useful to explore ways for non-issuer accounts to more easily remove trust lines.

### DIFF
--- a/src/ripple/app/tx/impl/CashCheck.cpp
+++ b/src/ripple/app/tx/impl/CashCheck.cpp
@@ -196,31 +196,18 @@ CashCheck::preclaim(PreclaimContext const& ctx)
         // An issuer can always accept their own currency.
         if (!value.native() && (value.getIssuer() != dstId))
         {
-            auto const sleIssuer = ctx.view.read(keylet::account(issuerId));
             auto const sleTrustLine =
                 ctx.view.read(keylet::line(dstId, issuerId, currency));
 
-            if (!sleTrustLine)
+            if (!sleTrustLine &&
+                !ctx.view.rules().enabled(featureCheckCashMakesTrustLine))
             {
-                if (!ctx.view.rules().enabled(featureCheckCashMakesTrustLine))
-                {
-                    JLOG(ctx.j.warn())
-                        << "Cannot cash check for IOU without trustline.";
-                    return tecNO_LINE;
-                }
-                else if (sleIssuer)
-                {
-                    // We can only create a trust line if the issuer does not
-                    // have requireAuth set.
-                    std::uint32_t const issuerFlags = {sleIssuer->at(sfFlags)};
-                    if (issuerFlags & lsfRequireAuth)
-                        return tecNO_AUTH;
-                }
+                JLOG(ctx.j.warn())
+                    << "Cannot cash check for IOU without trustline.";
+                return tecNO_LINE;
             }
 
-            // It would be nice to check this earlier, but moving it earlier
-            // in the code would be transaction changing (returning different
-            // tec codes).  So we leave it where it is.
+            auto const sleIssuer = ctx.view.read(keylet::account(issuerId));
             if (!sleIssuer)
             {
                 JLOG(ctx.j.warn())
@@ -231,15 +218,21 @@ CashCheck::preclaim(PreclaimContext const& ctx)
 
             if (sleIssuer->at(sfFlags) & lsfRequireAuth)
             {
+                if (!sleTrustLine)
+                {
+                    // We can only create a trust line if the issuer does not
+                    // have requireAuth set.
+                    return tecNO_AUTH;
+                }
+
                 // Entries have a canonical representation, determined by a
                 // lexicographical "greater than" comparison employing strict
                 // weak ordering. Determine which entry we need to access.
                 bool const canonical_gt(dstId > issuerId);
 
                 bool const is_authorized(
-                    sleTrustLine &&
-                    (sleTrustLine->at(sfFlags) &
-                     (canonical_gt ? lsfLowAuth : lsfHighAuth)));
+                    sleTrustLine->at(sfFlags) &
+                    (canonical_gt ? lsfLowAuth : lsfHighAuth));
 
                 if (!is_authorized)
                 {
--- a/src/ripple/app/tx/impl/CashCheck.cpp
+++ b/src/ripple/app/tx/impl/CashCheck.cpp
@@ -21,6 +21,7 @@
 #include <ripple/app/paths/Flow.h>
 #include <ripple/app/tx/impl/CashCheck.h>
 #include <ripple/basics/Log.h>
+#include <ripple/basics/scope.h>
 #include <ripple/protocol/Feature.h>
 #include <ripple/protocol/Indexes.h>
 #include <ripple/protocol/STAccount.h>
@@ -89,13 +90,13 @@ CashCheck::preclaim(PreclaimContext const& ctx)
     }
 
     // Only cash a check with this account as the destination.
-    AccountID const dstId{(*sleCheck)[sfDestination]};
+    AccountID const dstId = sleCheck->at(sfDestination);
     if (ctx.tx[sfAccount] != dstId)
     {
         JLOG(ctx.j.warn()) << "Cashing a check with wrong Destination.";
         return tecNO_PERMISSION;
     }
-    AccountID const srcId{(*sleCheck)[sfAccount]};
+    AccountID const srcId = sleCheck->at(sfAccount);
     if (srcId == dstId)
     {
         // They wrote a check to themselves.  This should be caught when
@@ -127,7 +128,7 @@ CashCheck::preclaim(PreclaimContext const& ctx)
     {
         using duration = NetClock::duration;
         using timepoint = NetClock::time_point;
-        auto const optExpiry = (*sleCheck)[~sfExpiration];
+        auto const optExpiry = sleCheck->at(~sfExpiration);
 
         // Expiration is defined in terms of the close time of the parent
         // ledger, because we definitively know the time that it closed but
@@ -148,7 +149,7 @@ CashCheck::preclaim(PreclaimContext const& ctx)
             return optAmount ? *optAmount : tx[sfDeliverMin];
         }(ctx.tx)};
 
-        STAmount const sendMax{(*sleCheck)[sfSendMax]};
+        STAmount const sendMax = sleCheck->at(sfSendMax);
         Currency const currency{value.getCurrency()};
         if (currency != sendMax.getCurrency())
         {
@@ -172,7 +173,7 @@ CashCheck::preclaim(PreclaimContext const& ctx)
         {
             STAmount availableFunds{accountFunds(
                 ctx.view,
-                (*sleCheck)[sfAccount],
+                sleCheck->at(sfAccount),
                 value,
                 fhZERO_IF_FROZEN,
                 ctx.j)};
@@ -211,7 +212,7 @@ CashCheck::preclaim(PreclaimContext const& ctx)
                 {
                     // We can only create a trust line if the issuer does not
                     // have requireAuth set.
-                    std::uint32_t const issuerFlags = {(*sleIssuer)[sfFlags]};
+                    std::uint32_t const issuerFlags = {sleIssuer->at(sfFlags)};
                     if (issuerFlags & lsfRequireAuth)
                         return tecNO_AUTH;
                 }
@@ -228,7 +229,7 @@ CashCheck::preclaim(PreclaimContext const& ctx)
                 return tecNO_ISSUER;
             }
 
-            if ((*sleIssuer)[sfFlags] & lsfRequireAuth)
+            if (sleIssuer->at(sfFlags) & lsfRequireAuth)
             {
                 // Entries have a canonical representation, determined by a
                 // lexicographical "greater than" comparison employing strict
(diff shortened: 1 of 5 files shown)


src/ripple/app/tx/impl/CashCheck.cpp
```diff
@@ -376,10 +508,14 @@ CashCheck::doApply()
                         << "flow did not produce DeliverMin.";
                     return tecPATH_PARTIAL;
                 }
-                if (doFix1623)
+                if (doFix1623 && !checkCashMakesTrustLine)
                     // Set the delivered_amount metadata.
                     ctx_.deliver(result.actualAmountOut);
             }
+            // Set the delivered amount metadata in all cases, not just
```
intelliot, line 515: Prior to `checkCashMakesTrustLine`, why was the delivered amount metadata set only if DeliverMin was used?

scottschurr: Hi @intelliot, I believe that was a bug.  As part of this pull request I added some unit tests.  I expected that in all cases the result of a successful `CheckCash` transactions would include a `delivered_amount` field.  I found that `delivered_amount` was not always set.  I surmised `delivered_amount` was not set because `DeliveredAmount` was not always set due to the conditional above.

When I changed the code so setting `DeliveredAmount` was only conditional on the amendment then the unit tests returned a `delivery_amount` field in all of the tested cases.

src/test/app/Check_test.cpp
```diff
@@ -1848,22 +1883,654 @@ class Check_test : public beast::unit_test::suite
...
+            IOU const OF5 = gw2["OF5"];
+            env(offer(gw2, XRP(92), OF5(92)));
+            env.close();
+            BEAST_EXPECT(
+                env.le(keylet::line(gw2, alice, OF5.currency)) == nullptr);
+            env(offer(alice, OF5(92), XRP(92)), ter(tecNO_LINE));
+            env.close();
+
+            // No one's owner count should have changed.
+            BEAST_EXPECT(ownerCount(env, gw1) == 0);
```
ximinez, line 2436: This should be `gw2`. It might be worth moving `gw1` into a new separate scope for the tests above to ensure this code can't see it.

src/test/app/Check_test.cpp
```diff
@@ -1848,22 +1883,654 @@ class Check_test : public beast::unit_test::suite
...
+            BEAST_EXPECT(ownerCount(env, alice) == 6);
+            BEAST_EXPECT(ownerCount(env, bob) == 5);
+
+            // Because gw2 has set lsfRequireAuth, neither trust line
+            // is created.
+            BEAST_EXPECT(
+                env.le(keylet::line(gw2, bob, OF5.currency)) == nullptr);
+            BEAST_EXPECT(
+                env.le(keylet::line(gw2, bob, CK5.currency)) == nullptr);
+        }
```
ximinez, line 2505: I think this could use another set of cases where the authorized trust lines are created. Obviously, there won't be any trustlines created automatically, but it can still test going over the trust line limit, and verify things still work with the `asfRequireAuth` flag.

src/test/app/Check_test.cpp
```diff
@@ -1848,22 +1883,654 @@ class Check_test : public beast::unit_test::suite
...
+            verifyDeliveredAmount(env, CK1(98));
+            env.close();
+
+            // gw1's check should be consumed.
+            // Since gw1's check was consumed and the trust line was not
+            // created by gw1, gw1's owner count should be 0.
+            BEAST_EXPECT(ownerCount(env, gw1) == 0);
+
+            // alice's automatically created trust line bumps her owner count.
+            BEAST_EXPECT(ownerCount(env, alice) == 2);
```
ximinez, line 2095: Overall, these tests are _fantastic_.
One thing I think could help with readability, would be to use a set of integers to keep track of what each account's owner count _should_ be. Then increment (or decrement) that value whenever a change is expected. This would allow the reader to look at each test case separately without needing to flip up to the previous test to check the prior value. Down the road, it would also allow reordering the test cases or inserting new test cases between the existing ones.
e.g.
```
        Account const gw1{"gw1"};
        std::size_t gw1Count = 0;
        Account const alice{"alice"};
        std::size_t aliceCount = 0;
        Account const bob{"bob"};
        std::size_t bobCount = 0;
[...]
            // gw1's check should be consumed.
            // Since gw1's check was consumed and the trust line was not
            // created by gw1, gw1's owner count should be 0.
            BEAST_EXPECT(ownerCount(env, gw1) == gw1Count);

            // alice's automatically created trust line bumps her owner count.
            BEAST_EXPECT(ownerCount(env, alice) == ++aliceCount);
```

src/test/jtx/impl/Env.cpp
```diff
@@ -342,8 +342,9 @@ Env::postconditions(JTx const& jt, TER ter, bool didApply)
     if (jt.ter &&
         !test.expect(
             ter == *jt.ter,
-            "apply: " + transToken(ter) + " (" + transHuman(ter) + ") != " +
-                transToken(*jt.ter) + " (" + transHuman(*jt.ter) + ")"))
+            "apply: Got " + transToken(ter) + " (" + transHuman(ter) +
+                "); Expected " + transToken(*jt.ter) + " (" +
+                transHuman(*jt.ter) + ")"))
```
ximinez, line 347: Heh. This has been a long time coming.

ximinez (changes requested):

scottschurr (commented):
Thanks for the great review @ximinez!  I've pushed a commit that I believe addresses your comments.  Please let me know if I missed anything.

I've also rebased on top of https://github.com/ripple/rippled/pull/3840 since it has passed and provides functionality that I'm using.  I anticipate that https://github.com/ripple/rippled/pull/3840 will be merged before this pull request is.

Any future reviewers will want to skip the commits below the one named `Introduce CheckCashMakesTrustLine amendment`.

src/test/app/Check_test.cpp
```diff
@@ -1848,22 +1883,654 @@ class Check_test : public beast::unit_test::suite
...
+                auto cmpOptField = [this, offerLine, checkLine](
+                                       auto const& sfield) {
+                    // Expect both fields to either be present or absent.
+                    BEAST_EXPECT(
+                        offerLine->isFieldPresent(sfield) ==
+                        checkLine->isFieldPresent(sfield));
+
+                    if (!offerLine->isFieldPresent(sfield) ||
+                        !checkLine->isFieldPresent(sfield))
+                        return;
```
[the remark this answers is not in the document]
ximinez: Oh shoot. I missed that!

ximinez (changes requested):

scottschurr:
@ximinez, thanks for the suggestion regarding the macro.  I followed your lead but tweaked it a bit.  At your convenience take a look and see what you think.

scottschurr:
Rebased to 1.8.0-b3.

ximinez (approved):
Sorry for the delay on this followup. The latest changes look good!

scottschurr:
@nbougalis is really busy right now, so I'm swapping in @thejohnfreeman as a technical reviewer.  Thanks for the help, @thejohnfreeman!

src/ripple/app/tx/impl/CashCheck.cpp
```diff
@@ -195,16 +196,31 @@ CashCheck::preclaim(PreclaimContext const& ctx)
...
+            auto const sleIssuer = ctx.view.read(keylet::account(issuerId));
             auto const sleTrustLine =
                 ctx.view.read(keylet::line(dstId, issuerId, currency));
+
             if (!sleTrustLine)
             {
-                JLOG(ctx.j.warn())
-                    << "Cannot cash check for IOU without trustline.";
-                return tecNO_LINE;
+                if (!ctx.view.rules().enabled(featureCheckCashMakesTrustLine))
```
thejohnfreeman, line 205: One way to do the changes from line 200 (in the original code) to line 225 that requires fewer actual code changes is to just add `&& !ctx.view.rules().enabled(featureCheckCashMakesTrustLine)` to the condition on line 200, and insert this condition on line 218:

```cpp
if (!sleTrustLine) {
    return tecNO_AUTH;
}
```

src/test/app/Check_test.cpp
```diff
@@ -1848,22 +1883,741 @@ class Check_test : public beast::unit_test::suite
...
+            env(check::cash(zoe, chkId, CK9(100)),
+                ter(features[featureCheckCashMakesTrustLine]
+                        ? TER(tesSUCCESS)
+                        : TER(tecNO_LINE)));
+            env.close();
+        }
+
+        // All remaining tests require featureCheckCashMakesTrustLine.
+        if (!features[featureCheckCashMakesTrustLine])
+            return;
```
thejohnfreeman, line 1974: With CheckCashMakesTrustLine disabled, this function is only the test above this return, which duplicates a test elsewhere. I personally find it easier to understand if this function is only run with CheckCashMakesTrustLine enabled, without lines 1943 - 1974 (the sanity check and early return), but maybe with an assertion that the feature is enabled.

thejohnfreeman (commented):
Can we add a test that a second check for the same currency can be cashed after one that created a trust line (i.e. that the limits are ignored even when a trust line already existed)?

src/ripple/app/tx/impl/CashCheck.cpp
thejohnfreeman: Here it is as a commit: https://github.com/thejohnfreeman/rippled/commit/d57c0e801dff73946ee69faeb641d66117c32312

scottschurr:
Thanks for the great review @thejohnfreeman.  I've pushed a couple of commits that I believe address your comments.  Please let me know if there's anything else you'd like addressed.

ximinez (approved):
Still looks good. I like the changes.

thejohnfreeman (approved):

thejohnfreeman:
Looks great! Thank you!

scottschurr:
@mDuo13 and @intelliot, does this change look reasonable to you?  I want to make sure before I mark it passed.  Thanks.

mDuo13:
I certainly appreciate the idea, and the semantics as laid out in the description sound perfect to me. I haven't tested out the changes or looked through the code to confirm that it works as described or that there aren't any relevant edge cases I haven't thought of.

To be clear, when cashing a Check for a "flexible" amount, is the amount it delivers is no longer bounded by the Check recipient's trust limits (even if there's an existing line and trust limits are set to nonzero values)? That's sensible to me, though it may be surprising in at least a couple cases, so it's something to make sure the docs clarify. I guess in that case, the only things that may limit the redemption to less than the Check's full amount are:

(a) the amount the check sender holds, if they aren't the issuer
(b) the transfer fee, if the check sender is not the issuer

Right?

scottschurr:
@mDuo13, that sounds correct.

PR #3828, opened 2021-04-20
Improve unit test coverage of AccountDelete transactor

scottschurr:

Adds unit tests for the `AccountDelete` transactor to improve coverage.

### Context of Change

While reading through the `AccountDelete` transactor source code for other reasons, I thought I spotted a potential bug.  And I saw, when running code coverage locally, that the case I was wondering about was not exercised by the unit tests.

So I added a unit test to cover the case I was wondering about (it ended up the there was no bug).  Then I added more unit tests to try and make the `AccountDelete` code coverage as complete as is reasonable.

### Type of Change

- [x] Tests (added tests for code that already exists)

This change only adds unit tests, so there's no reason to mention the changes in the release notes.

### DIFF
--- a/src/ripple/app/tx/impl/DeleteAccount.cpp
+++ b/src/ripple/app/tx/impl/DeleteAccount.cpp
@@ -204,7 +204,8 @@ DeleteAccount::preclaim(PreclaimContext const& ctx)
             uDirEntry,
             dirEntry,
             ctx.j))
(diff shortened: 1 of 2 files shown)


sgramkumar (approved):

intelliot (approved):

pwang200 (approved):
:+1: looks good!

PR #3829, opened 2021-04-20
[Trivial] Convert warn-level log to trace level.  Closes #3153

### LINKED ISSUE #3153: Error occurs after leger 256 is stand-alone mode Amendments
What is Amendment and how to use it in rippled stand-alone mode?

When I reached up to 256 ledger, then it send one transaction automatically called `EnableAmendment` & shows error like `Transaction in ledger 257 affects no accounts`.

So, how can I use this in rippled stand-alone mode?

scottschurr:

There are a couple of places in the code that log at warning level about transactions that affect no accounts.  However there are pseudo transactions that can occur every flag ledger, and these pseudo transactions affect no accounts (directly).  Seeing the warning generated has been shown to cause concern for at least one user.

This pull request leaves the logging messages in place, but reduces their threshold to trace level, so they are less likely to be seen by the average user.

### Context of Change

The messages of concern were first reported in https://github.com/ripple/rippled/issues/3153.  This change will close that issue.

### Type of Change

Since the change only affects logging levels, there's really no reason to include this change in the release notes.

### DIFF
--- a/src/ripple/app/rdb/impl/RelationalDBInterface_nodes.cpp
+++ b/src/ripple/app/rdb/impl/RelationalDBInterface_nodes.cpp
@@ -317,12 +317,14 @@ saveValidatedLedger(
                     JLOG(j.trace()) << "ActTx: " << sql;
                     *db << sql;
                 }
-                else
+                else if (auto const sleTxn = acceptedLedgerTx->getTxn();
+                         !isPseudoTx(*sleTxn))
                 {
+                    // It's okay for pseudo transactions to not affect any
+                    // accounts.  But otherwise...
                     JLOG(j.warn()) << "Transaction in ledger " << seq
                                    << " affects no accounts";
-                    JLOG(j.warn()) << acceptedLedgerTx->getTxn()->getJson(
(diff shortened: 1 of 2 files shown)


cjcobb23 (commented):
Sorry, for taking so long to review this. What do you think of setting this to debug instead of trace?

scottschurr:
I'd be okay with `debug`, but I'm not sure the benefit?  You can see transactions like these on any flag ledger.  Since that was the case `trace` seemed obvious.  But I don't rely on these log statements and you do.  So your call.

cjcobb23:
I think the best solution would be to actually distinguish between pseudo transactions and normal transactions. Then we could keep the log at the warning level for normal transactions, and just not log at all for pseudo transactions. I'm not sure how hard it is to distinguish though. What do you think @scottschurr ?

scottschurr:
Hi @cjcobb23 , I pushed a commit that implements the earlier suggestion.  Let me know what you think at your convenience.  Thanks.

cjcobb23 (approved):
LGTM

PR #3830, opened 2021-04-22
Replace dirAdd() calls with dirAppend() or dirInsert()

scottschurr:

Just tidying up some unfinished business here: https://github.com/ripple/rippled/issues/2516

Calls to `dirDelete()` had been removed in a previous commit.  Now that the `SortedDirectories` amendment is unconditionally enabled, it's okay to remove the calls to `dirAdd()`.  Once all `dirAdd()` calls are removed, then `dirAdd()` itself can be removed.  That's what we do here.

### Type of Change

This change is purely a refactor and should have no effect on functionality.  So the change should not need any mention in the release notes.

### DIFF
--- a/src/ripple/app/tx/impl/CreateCheck.cpp
+++ b/src/ripple/app/tx/impl/CreateCheck.cpp
@@ -183,7 +183,7 @@ CreateCheck::doApply()
             return tecINSUFFICIENT_RESERVE;
     }
 
-    // Note that we we use the value from the sequence or ticket as the
+    // Note that we use the value from the sequence or ticket as the
     // Check sequence.  For more explanation see comments in SeqProxy.h.
     std::uint32_t const seq = ctx_.tx.getSeqProxy().value();
     Keylet const checkKeylet = keylet::check(account_, seq);
@@ -216,7 +216,7 @@ CreateCheck::doApply()
             describeOwnerDir(dstAccountId));
 
         JLOG(j_.trace()) << "Adding Check to destination directory "
-                         << to_string(sleCheck->key()) << ": "
+                         << to_string(checkKeylet.key) << ": "
                          << (page ? "success" : "failure");
 
         if (!page)
@@ -232,7 +232,7 @@ CreateCheck::doApply()
             describeOwnerDir(account_));
 
         JLOG(j_.trace()) << "Adding Check to owner directory "
-                         << to_string(sleCheck->key()) << ": "
+                         << to_string(checkKeylet.key) << ": "
                          << (page ? "success" : "failure");
 
         if (!page)
--- a/src/ripple/app/tx/impl/CreateCheck.cpp
+++ b/src/ripple/app/tx/impl/CreateCheck.cpp
@@ -186,7 +186,8 @@ CreateCheck::doApply()
     // Note that we we use the value from the sequence or ticket as the
     // Check sequence.  For more explanation see comments in SeqProxy.h.
     std::uint32_t const seq = ctx_.tx.getSeqProxy().value();
-    auto sleCheck = std::make_shared<SLE>(keylet::check(account_, seq));
+    Keylet const checkKeylet = keylet::check(account_, seq);
+    auto sleCheck = std::make_shared<SLE>(checkKeylet);
 
     sleCheck->setAccountID(sfAccount, account_);
     AccountID const dstAccountId = ctx_.tx[sfDestination];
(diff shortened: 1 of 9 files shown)


src/ripple/ledger/ApplyView.h
```diff
@@ -273,22 +273,19 @@ class ApplyView : public ReadView
...
         std::function<void(std::shared_ptr<SLE> const&)> const& describe)
     {
-        return dirAppend(directory, key.key, describe);
+        if (key.type != ltOFFER)
+        {
+            assert(!"Only Offers are appended to book directories.  "
+                "Call dirInsert() instead.");
+            return std::nullopt;
+        }
+        return dirAdd(true, directory, key.key, describe);
```
gregtatcam, line 288: The "High Level Overview of Change" in the PR through me off at first since it says "remove the calls to dirAdd()". Perhaps it should clarify what "dirAdd()" should be deleted?

gregtatcam (approved):
:+1: LGTM

src/ripple/app/tx/impl/CreateCheck.cpp
```diff
@@ -186,7 +186,8 @@ CreateCheck::doApply()
     // Note that we we use the value from the sequence or ticket as the
     // Check sequence.  For more explanation see comments in SeqProxy.h.
     std::uint32_t const seq = ctx_.tx.getSeqProxy().value();
-    auto sleCheck = std::make_shared<SLE>(keylet::check(account_, seq));
+    Keylet const checkKeylet = keylet::check(account_, seq);
```
miguelportilla, line 189: Since you are in the neighborhood, please correct the comment by removing the extra 'we'.

src/ripple/app/tx/impl/CreateCheck.cpp
```diff
@@ -209,13 +210,10 @@ CreateCheck::doApply()
...
-        auto const page = dirAdd(
-            view(),
+        auto const page = view().dirInsert(
             keylet::ownerDir(dstAccountId),
-            sleCheck->key(),
-            false,
-            describeOwnerDir(dstAccountId),
-            viewJ);
+            checkKeylet,
+            describeOwnerDir(dstAccountId));
```
miguelportilla, line 216: Nit: it may be worth replacing `sleCheck->key()` in the log calls below with the new local `checkKeylet.key`.

src/ripple/app/tx/impl/CreateTicket.cpp
```diff
@@ -117,21 +117,17 @@ CreateTicket::doApply()
...
-        auto const page = dirAdd(
-            view(),
+        auto const page = view().dirInsert(
             keylet::ownerDir(account_),
-            sleTicket->key(),
-            false,
-            describeOwnerDir(account_),
-            viewJ);
+            ticketKeylet,
+            describeOwnerDir(account_));
```
miguelportilla, line 130: Nit: it may be worth replacing sleTicket->key() in the log call below with the new local ticketKeylet.key.

miguelportilla (approved):
LGTM 👍

scottschurr:
@miguelportilla, thanks for the careful review.  I made your suggested changes in this commit: https://github.com/ripple/rippled/pull/3830/commits/c2a3b9f09fb94bba91ccc8c2ca5cefafb7b405f2  Let me know if you're good with those changes.  If you are, and the branch passes CI, then I'll squash and mark the pull request as passed.  Thanks.

src/ripple/ledger/ApplyView.h
nbougalis: I appreciate the `assert` to try to keep us from doing anything weird, but I'm not sure it's a good idea. This function is intended to append to the particular directory, instead of trying to fit it into the first page with an empty slot. It's true that this operation is currently only used by order books (so as to maintain temporal ordering of offers), but there's nothing that prevents this from being used differently, so I think the `assert` my communicate the wrong information.

For example, imagine if we added a new directory for, say, NFT offers.

nbougalis (approved):

src/ripple/ledger/ApplyView.h
scottschurr: At @nbougalis, I think the `assert` is a good idea because it helps keep folks less familiar with the code on the straight and narrow.  To someone who is not intimate with this code, it may not be obvious whether a particular object should be added or appended to the directory.  As it stands there's currently only one situation where append is appropriate: offers in book directories.  All other cases use add.

In the future there may, indeed, be other situations where appending is the right thing to do.  But the source code is malleable, so the assert can be removed or changed when the new situation arises.

Just my take...

miguelportilla:
@scottschurr Changes look great, thank you.

PR #3831, opened 2021-04-23
Fix CMake check for Windows platform

thejohnfreeman:
Unlike [previous Visual Studio generators](https://cmake.org/cmake/help/git-stage/generator/Visual%20Studio%2015%202017.html), the [`Visual Studio 16 2019` generator for CMake](https://cmake.org/cmake/help/git-stage/generator/Visual%20Studio%2016%202019.html) does not have a `Win64` variant. Instead, users are expected to pass the platform argument through the `-A` option.

### DIFF
--- a/Builds/CMake/RippledSanity.cmake
+++ b/Builds/CMake/RippledSanity.cmake
@@ -73,9 +73,9 @@ if ("${CMAKE_CURRENT_SOURCE_DIR}" STREQUAL "${CMAKE_BINARY_DIR}")
 endif ()
 
 if ("${CMAKE_GENERATOR}" MATCHES "Visual Studio" AND
-    NOT ("${CMAKE_GENERATOR}" MATCHES .*Win64.*))
+    NOT ("${CMAKE_GENERATOR_PLATFORM}" MATCHES "x64"))
   message (FATAL_ERROR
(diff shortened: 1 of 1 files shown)


thejohnfreeman:
Travis failed on the only Windows build that uses a Visual Studio generator (the others use Ninja). The log doesn't show the error, possibly hiding it in `C:/Users/travis/build/ripple/rippled/build.ms/CMakeFiles/CMakeOutput.log`. I have a suspicion that the generator doesn't set the `CMAKE_GENERATOR_PLATFORM` variable, if the platform is specified in the generator name instead of with the `-A` option. I'll have to test that locally to confirm. I'm going to leave this up for now because I have another lead: perhaps it sets the variable [`CMAKE_VS_PLATFORM_NAME`](https://cmake.org/cmake/help/latest/variable/CMAKE_VS_PLATFORM_NAME.html#variable:CMAKE_VS_PLATFORM_NAME).

ximinez:
It's not quite as nice, but https://github.com/ripple/rippled/commit/6696663f899999933b54cdf9e48b28ce8e856264#diff-98e23f6822b062c4c2067887578953ee7d81f8fb28920ddc8a917842cddc1edeR74 handles the different cases correctly. Feel free to steal, or just wait a day or so for me to open a PR on that branch.

miguelportilla (approved):
Tested using the Ninja and Visual Studio 2019 generators, both worked. 👍

PR #3833, opened 2021-04-29
Add higher bound to validations to keep

pwang200:
Change `toKeep_` in the validations container to a range of `seq` with lower and higher bounds instead of just the lower bound. With the added higher bound of the range, validations with crazy high `seq` outside of the range can expire.

## Test Plan
- Unit tests
- A network test, connected a node (a validator but was not on the UNL) to the Devnet for about a day. The number of validations kept by the node (from `get_counts` reply) did not grow overtime.

### DIFF
--- a/src/ripple/consensus/Validations.h
+++ b/src/ripple/consensus/Validations.h
@@ -743,7 +743,7 @@ public:
                     if (!validationMap.empty())
                     {
                         auto const seq = validationMap.begin()->second.seq();
-                        if (seq >= toKeep_->low_ && seq < toKeep_->high_)
+                        if (toKeep_->low_ <= seq && seq < toKeep_->high_)
                         {
                             byLedger_.touch(i);
                         }
@@ -752,7 +752,7 @@ public:
 
                 for (auto i = bySequence_.begin(); i != bySequence_.end(); ++i)
                 {
-                    if (i->first >= toKeep_->low_ && i->first < toKeep_->high_)
+                    if (toKeep_->low_ <= i->first && i->first < toKeep_->high_)
                     {
                         bySequence_.touch(i);
                     }
--- a/src/ripple/consensus/Validations.h
+++ b/src/ripple/consensus/Validations.h
@@ -324,8 +324,13 @@ class Validations
         beast::uhash<>>
         bySequence_;
 
-    // Sequence of the earliest validation to keep from expire
-    std::optional<Seq> toKeep_;
+    // A range [low_, high_) of validations to keep from expire
+    struct KeepRange
+    {
+        Seq low_;
+        Seq high_;
+    };
+    std::optional<KeepRange> toKeep_;
 
     // Represents the ancestry of validated ledgers
     LedgerTrie<Ledger> trie_;
@@ -698,14 +703,17 @@ public:
     }
 
     /**
-     * Set the smallest sequence number of validations to keep from expire
-     * @param s the sequence number
+     * Set the range [low, high) of validations to keep from expire
+     * @param low the lower sequence number
+     * @param high the higher sequence number
+     * @note high must be greater than low
      */
     void
-    setSeqToKeep(Seq const& s)
+    setSeqToKeep(Seq const& low, Seq const& high)
     {
         std::lock_guard lock{mutex_};
-        toKeep_ = s;
+        assert(low < high);
+        toKeep_ = {low, high};
     }
 
     /** Expire old validation sets
@@ -723,7 +731,8 @@ public:
             {
                 auto const& validationMap = i->second;
                 if (!validationMap.empty() &&
-                    validationMap.begin()->second.seq() >= toKeep_)
+                    validationMap.begin()->second.seq() >= toKeep_->low_ &&
+                    validationMap.begin()->second.seq() < toKeep_->high_)
                 {
                     byLedger_.touch(i);
                 }
@@ -731,7 +740,7 @@ public:
 
             for (auto i = bySequence_.begin(); i != bySequence_.end(); ++i)
             {
-                if (i->first >= toKeep_)
+                if (i->first >= toKeep_->low_ && i->first < toKeep_->high_)
                 {
                     bySequence_.touch(i);
                 }
--- a/src/ripple/consensus/Validations.h
+++ b/src/ripple/consensus/Validations.h
@@ -727,22 +727,35 @@ public:
         std::lock_guard lock{mutex_};
         if (toKeep_)
         {
-            for (auto i = byLedger_.begin(); i != byLedger_.end(); ++i)
+            // We only need to refresh the keep range when it's just about to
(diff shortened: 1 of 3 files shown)


scottschurr (approved):
Looks great, and a pleasure to code review!  The NegativeUNLVote.cpp file has really nice paragraph comments that describe the intent of the code.  That made it very easy to determine why the changes were being made.  What you have looks like a good fix for the identified problem.  I poked at it quite a bit and didn't really see anything in the pull request that I would want to change.

As I was nosing around the code I noticed some preexisting inefficiencies in Validations.h.  I noted a way to patch up one of those inefficiencies in this code review.  You might consider that change or something like it.  But that's really just a patch.

I think the real root of the problem is the pair of `beast::aged_unordered_map`s storing the cached validations.  It would be interesting to explore (in a different branch) replacing the `beast::aged_unordered_map`s in Validations.h with `boost::multi_index_container`.  That could provide a lot more flexibility in the indexing schemes and how expiration would be handled.

src/ripple/consensus/Validations.h
```diff
@@ -723,15 +731,16 @@ class Validations
             {
```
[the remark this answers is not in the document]
pwang200: Thanks, that is a good idea. Changed and added some unit tests. 
BTW, I think `Validations::expire()` is called every minute, instead of every ledger close.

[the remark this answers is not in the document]
scottschurr: You're partly right about how often `Validations::expire()` is called.  I put a print at the top of `Validations::expire()`. On my system I see that print pop out about every 10 or 11 seconds, not every four.  So it's certainly not happening with every ledger close.  Maybe `Validations::expire()` is called from more places than you realize?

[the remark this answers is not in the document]
pwang200: Scott and I had some discussion. To summarize, `Validations::expire()` is called in `Application::doSweep()`. Scott found out that the call frequency depended on node size. My node size is `medium`, so `doSweep()` happens every minute, and Scott's is every 10 seconds because his node size is `tiny`. 

Note that the time guard Scott proposed still works in all cases.

src/ripple/consensus/Validations.h
```diff
@@ -719,21 +727,35 @@ class Validations
...
+                    parms_.validationFRESHNESS;
+
+                for (auto i = byLedger_.begin(); i != byLedger_.end(); ++i)
                 {
-                    byLedger_.touch(i);
+                    auto const& validationMap = i->second;
+                    if (!validationMap.empty())
+                    {
+                        auto const seq = validationMap.begin()->second.seq();
+                        if (seq >= toKeep_->low_ && seq < toKeep_->high_)
```
ximinez, line 746: Style nit: When there are multiple comparisons like this, I like to have the comparison operators all "pointing" in the same direction, and in cases like this where you're checking a range, I like to have the range boundaries on the "outside". i.e.
```
if (toKeep_->low_ <= seq && seq < toKeep_->high_)
```
I find that more readable, and easier to reason about what's being checked.

src/ripple/consensus/Validations.h
```diff
@@ -719,21 +727,35 @@ class Validations
...
                 }
-            }
 
-            for (auto i = bySequence_.begin(); i != bySequence_.end(); ++i)
-            {
-                if (i->first >= toKeep_)
+                for (auto i = bySequence_.begin(); i != bySequence_.end(); ++i)
                 {
-                    bySequence_.touch(i);
+                    if (i->first >= toKeep_->low_ && i->first < toKeep_->high_)
```
ximinez, line 755: Same style suggestion here re: the symmetry of the operators.

src/test/consensus/Validations_test.cpp
```diff
@@ -707,21 +707,45 @@ class Validations_test : public beast::unit_test::suite
...
+        harness.clock().advance(harness.parms().validationSET_EXPIRES);
+        harness.vals().expire();
+        BEAST_EXPECT(harness.vals().numTrustedForLedger(ledgerA.id()) == 0);
 
-        // Keep the validation from expire
+        // use setSeqToKeep to keep the validation from expire
+        Ledger ledgerB = h["ab"];
+        BEAST_EXPECT(ValStatus::current == harness.add(a.validate(ledgerB)));
+        BEAST_EXPECT(harness.vals().numTrustedForLedger(ledgerB.id()) == 1);
+        harness.vals().setSeqToKeep(ledgerB.seq(), ++ledgerB.seq());
```
ximinez, line 724: This reads really weird for me. When I see `++`, I'm expecting to be incrementing something that will persist, like a variable. I get that it's legal, but it's confusing. I'd much rather see this written as `ledgerB.seq() + Ledger::Seq(1)`.

And while I'm thinking about it, I think most of these objects should be declared as `const` so it's clear that nothing is changing in them. `a`, `ledgerA`, `ledgerB`, `loops`, `ledgerC`.

src/test/consensus/Validations_test.cpp
```diff
@@ -707,21 +707,45 @@ class Validations_test : public beast::unit_test::suite
...
         harness.clock().advance(harness.parms().validationSET_EXPIRES);
-        harness.vals().setSeqToKeep(ledgerA.seq());
         harness.vals().expire();
-        BEAST_EXPECT(harness.vals().numTrustedForLedger(ledgerA.id()));
+        BEAST_EXPECT(harness.vals().numTrustedForLedger(ledgerB.id()) == 1);
+        // change toKeep
+        harness.vals().setSeqToKeep(++ledgerB.seq(), ++(++ledgerB.seq()));
```
ximinez, line 729: Similar to above, I'd rather see this as `setSeqToKeep(ledgerB.seq() + Ledger::Seq(1), ledgerB.seq() + Ledger::Seq(2))`

I'm realizing that the explicit `Ledger::Seq` can get unwieldy, so how about defining
```
constexpr Ledger::Seq one(1);
constexpr Ledger::Seq two(2);
```
Then you can do `ledgerB.seq() + one`, etc.

ximinez (changes requested):

src/test/consensus/Validations_test.cpp
pwang200: Good idea, changed, thanks.

src/test/consensus/Validations_test.cpp
pwang200: Thanks, changed.

src/ripple/consensus/Validations.h
pwang200: Thanks, changed.

src/ripple/consensus/Validations.h
pwang200: Thanks, changed. It is more readable. I spent less time staring at the line after the change.

ximinez (approved):
Looks good!

Issue #2945, opened 2019-05-18
minLedgerVersion transaction parameter (like Bitcoin "LockTime")


tim-dorcey:
It would be useful if a transaction could specify the earliest ledger version after which the transaction may appear. This could be used to recover funds if a private key is lost.

When transaction N is created, the account would also sign and save transaction N+1 that transfers the remaining balance to a recovery account, but only in some future ledger, say 30 days away.

The private key for the recovery account could be kept in cold storage, since it is never needed unless a recovery transaction is executed. Furthermore, it is of no value unless the parent account submits no new transactions for 30 days, since each new transaction N+1 automatically cancels the previous recovery transaction by consuming its sequence number.

MarkusTeufelberger:
How would that be different from the existing Escrow feature?

tim-dorcey:
The existing Escrow cannot be canceled before the time lock expires.  This is automatically canceled by any subsequent transaction.

JoelKatz:
Great example use case! I like this idea a lot.

smlu:
Just for consideration, it'd be interesting to have on chain delayed/deferred transactions in combination with authorization.
In such system, you could define a transaction wait time as one of the multi-signing weight. Such configuration could then recover an account if some of minimum threshold keys are lost, depending on the weight of wait condition.
This could also be useful in case of "single key - multi sig" configuration where you'd set one key and a wait time to define the min authorization threshold. User could then monitor his/her account and if deferred transaction is published to the network which was not approved by the user, he/she can cancel it thus preventing losing funds. User also get an information that his/her keys are compromised and should change them.

For reference see the sophisticated [permissions system](https://medium.com/@eoscafeblock/how-permissions-work-in-eos-82e55eb1d896) in [EOS](https://github.com/EOSIO/eos) blockchain.

mDuo13:
This has also [been suggested](https://github.com/ripple/rippled/issues/2456#issuecomment-375849740) under the name `FirstLedgerSequence`.

intelliot:
This is a nice idea, but I haven't recently heard from anyone that this is something they would actually use. Setting `Low Priority` as this issue hasn't seen activity in over a year. @tim-dorcey if this is a feature you'd like to see, please confirm that you would use it (and if possible, point to others who would use it too!)

timethernom:
[...]

PR #3837, opened 2021-05-05
`TemporalTxValidity`: Implement temporal transaction validity rules:

nbougalis:

This commit introduces the `TemporalTxValidity` amendment which, if enabled, makes it possible for a transaction to control the time period during which the transaction should be considered valid.
    
This feature can be especially useful for transactions that use tickets instead of sequence numbers.
    
Two new optional fields are introduced that can be used with any submitted transactions:
    
1. The `NotValidBefore` field specifies the point in time before which the transaction should be rejected.
2. The `NotValidAfter` field specifies the point in time after which the transaction should be rejected.
    
Both fields specify the time as [seconds after the "Ripple epoch"](https://xrpl.org/basic-data-types.html#specifying-time) and one or both may be present, with the restriction that if both are used then the `NotValidAfter` time must be strictly greater than the `NotValidBefore` time.

### Type of Change

- [X] Amendment
- [X] Documentation Updates
- [ ] Release

## Test Plan

Unit tests are included with this code, but it would be nice to add integration tests that exercise the code more.

### DIFF
--- a/src/ripple/app/tx/impl/Transactor.cpp
+++ b/src/ripple/app/tx/impl/Transactor.cpp
@@ -48,6 +48,23 @@ preflight0(PreflightContext const& ctx)
         return temINVALID;
     }
 
+    if (!ctx.rules.enabled(featureTemporalTxValidity))
+    {
+        if (ctx.tx.isFieldPresent(sfNotValidAfter))
+            return temMALFORMED;
+
+        if (ctx.tx.isFieldPresent(sfNotValidBefore))
+            return temMALFORMED;
+    }
+    else
+    {
+        auto const notBefore = ctx.tx[~sfNotValidBefore];
+        auto const notAfter = ctx.tx[~sfNotValidAfter];
+
+        if (notBefore && notAfter && (notBefore >= notAfter))
+            return temBAD_TEMPORAL_VALIDITY;
+    }
+
     return tesSUCCESS;
 }
 
@@ -446,6 +463,31 @@ Transactor::apply()
     }
 
     return doApply();
+}
+
+NotTEC
+Transactor::checkTemporalTxValidity(PreclaimContext const& ctx)
+{
+    using d = NetClock::duration;
+    using tp = NetClock::time_point;
+
+    if (ctx.view.rules().enabled(featureTemporalTxValidity))
+    {
+        // We check time against the close time of the parent ledger,
+        // because that is the time that the network, as a whole,
+        // agrees on.
+        auto const pct = ctx.view.parentCloseTime();
+
+        if (auto const cutoff = ctx.tx[~sfNotValidBefore];
+            cutoff && (pct < tp{d{*cutoff}}))
+            return tefTOO_EARLY;
+
+        if (auto const cutoff = ctx.tx[~sfNotValidAfter];
+            cutoff && (pct > tp{d{*cutoff}}))
+            return tefTOO_LATE;
+    }
+
+    return tesSUCCESS;
 }
 
 NotTEC
--- a/src/ripple/protocol/impl/TxFormats.cpp
+++ b/src/ripple/protocol/impl/TxFormats.cpp
@@ -55,6 +55,8 @@ TxFormats::TxFormats()
             {sfClearFlag, soeOPTIONAL},
             {sfTickSize, soeOPTIONAL},
             {sfTicketSequence, soeOPTIONAL},
+            {sfNotValidBefore, soeOPTIONAL},
+            {sfNotValidAfter, soeOPTIONAL},
         },
         commonFields);
 
@@ -65,6 +67,8 @@ TxFormats::TxFormats()
             {sfQualityIn, soeOPTIONAL},
             {sfQualityOut, soeOPTIONAL},
             {sfTicketSequence, soeOPTIONAL},
+            {sfNotValidBefore, soeOPTIONAL},
+            {sfNotValidAfter, soeOPTIONAL},
         },
         commonFields);
 
@@ -76,6 +80,8 @@ TxFormats::TxFormats()
             {sfExpiration, soeOPTIONAL},
             {sfOfferSequence, soeOPTIONAL},
             {sfTicketSequence, soeOPTIONAL},
+            {sfNotValidBefore, soeOPTIONAL},
+            {sfNotValidAfter, soeOPTIONAL},
         },
         commonFields);
 
@@ -84,6 +90,8 @@ TxFormats::TxFormats()
         {
             {sfOfferSequence, soeREQUIRED},
             {sfTicketSequence, soeOPTIONAL},
+            {sfNotValidBefore, soeOPTIONAL},
+            {sfNotValidAfter, soeOPTIONAL},
         },
         commonFields);
 
@@ -92,6 +100,8 @@ TxFormats::TxFormats()
         {
             {sfRegularKey, soeOPTIONAL},
             {sfTicketSequence, soeOPTIONAL},
+            {sfNotValidBefore, soeOPTIONAL},
+            {sfNotValidAfter, soeOPTIONAL},
         },
         commonFields);
 
@@ -106,6 +116,8 @@ TxFormats::TxFormats()
             {sfDestinationTag, soeOPTIONAL},
             {sfDeliverMin, soeOPTIONAL},
             {sfTicketSequence, soeOPTIONAL},
+            {sfNotValidBefore, soeOPTIONAL},
+            {sfNotValidAfter, soeOPTIONAL},
         },
         commonFields);
 
@@ -119,6 +131,8 @@ TxFormats::TxFormats()
             {sfFinishAfter, soeOPTIONAL},
             {sfDestinationTag, soeOPTIONAL},
             {sfTicketSequence, soeOPTIONAL},
+            {sfNotValidBefore, soeOPTIONAL},
+            {sfNotValidAfter, soeOPTIONAL},
         },
         commonFields);
 
@@ -130,6 +144,8 @@ TxFormats::TxFormats()
             {sfFulfillment, soeOPTIONAL},
             {sfCondition, soeOPTIONAL},
             {sfTicketSequence, soeOPTIONAL},
+            {sfNotValidBefore, soeOPTIONAL},
+            {sfNotValidAfter, soeOPTIONAL},
         },
         commonFields);
 
@@ -139,6 +155,8 @@ TxFormats::TxFormats()
             {sfOwner, soeREQUIRED},
             {sfOfferSequence, soeREQUIRED},
             {sfTicketSequence, soeOPTIONAL},
+            {sfNotValidBefore, soeOPTIONAL},
+            {sfNotValidAfter, soeOPTIONAL},
         },
         commonFields);
 
@@ -175,6 +193,8 @@ TxFormats::TxFormats()
         {
             {sfTicketCount, soeREQUIRED},
             {sfTicketSequence, soeOPTIONAL},
+            {sfNotValidBefore, soeOPTIONAL},
+            {sfNotValidAfter, soeOPTIONAL},
         },
         commonFields);
 
@@ -186,6 +206,8 @@ TxFormats::TxFormats()
             {sfSignerQuorum, soeREQUIRED},
             {sfSignerEntries, soeOPTIONAL},
             {sfTicketSequence, soeOPTIONAL},
+            {sfNotValidBefore, soeOPTIONAL},
+            {sfNotValidAfter, soeOPTIONAL},
         },
         commonFields);
 
@@ -199,6 +221,8 @@ TxFormats::TxFormats()
             {sfCancelAfter, soeOPTIONAL},
             {sfDestinationTag, soeOPTIONAL},
             {sfTicketSequence, soeOPTIONAL},
+            {sfNotValidBefore, soeOPTIONAL},
+            {sfNotValidAfter, soeOPTIONAL},
         },
         commonFields);
 
@@ -209,6 +233,8 @@ TxFormats::TxFormats()
             {sfAmount, soeREQUIRED},
             {sfExpiration, soeOPTIONAL},
             {sfTicketSequence, soeOPTIONAL},
+            {sfNotValidBefore, soeOPTIONAL},
+            {sfNotValidAfter, soeOPTIONAL},
         },
         commonFields);
 
@@ -221,6 +247,8 @@ TxFormats::TxFormats()
             {sfSignature, soeOPTIONAL},
             {sfPublicKey, soeOPTIONAL},
             {sfTicketSequence, soeOPTIONAL},
+            {sfNotValidBefore, soeOPTIONAL},
+            {sfNotValidAfter, soeOPTIONAL},
         },
         commonFields);
 
@@ -233,6 +261,8 @@ TxFormats::TxFormats()
             {sfDestinationTag, soeOPTIONAL},
             {sfInvoiceID, soeOPTIONAL},
             {sfTicketSequence, soeOPTIONAL},
+            {sfNotValidBefore, soeOPTIONAL},
+            {sfNotValidAfter, soeOPTIONAL},
         },
         commonFields);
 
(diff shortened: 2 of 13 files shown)


scottschurr:
@ximinez, any thoughts on how this will interact with the `TxQ`?

intelliot:
A previously-proposed idea was to add a `FirstLedgerSequence` field that provides a lower bound on a transaction's valid period. A reasonable auto-fill value would be the current ledger index. This makes it easier to handle an edge case around getting transaction's final outcome. This is because the range of ledgers you need to check for a transaction's final outcome becomes self-contained in the signed transaction instructions.

@mDuo13 Would `NotValidBefore` be better than `FirstLedgerSequence`?

I suppose we'd need a way to map from (Ripple epoch) timestamps to ledger index numbers. I'm not sure how difficult that is to do.

nbougalis:
I think that the two mechanisms are slightly different.

Having a time-based option is probably better suited for human operations, including things like contracts.

At any rate, if ledger sequence restrictions enjoy more support, I'm fine with that.

src/ripple/protocol/jss.h
```diff
@@ -597,6 +597,9 @@ JSS(warnings);                // out: server_info, server_state
 JSS(workers);
 JSS(write_load);   // out: GetCounts
 JSS(NegativeUNL);  // out: ValidatorList; ledger type
+JSS(NotValidAfter);
+JSS(NotValidBefore);
```
ximinez, line 601: This file has historically ordered the JSS items alphabetically, with first letter upper case items up top. All three of these should be moved to fit in with that order.

ximinez:
> any thoughts on how this will interact with the `TxQ`?

Because all of the new results are "not likely to claim a fee", and they're all checked under `preclaim`, the straightforward interactions with the `TxQ` should be clean.
* If the transaction is new (`TxQ::apply`), and one of the temporal-related results is returned, it won't get queued.
* If the transaction is attempting to be removed from the queue an processed into the open ledger (`TxQ::accept`), `preclaim` is called again, so those tests are executed against the updated time stamps. If the temporal check fails, that's a `tef`, which results in the transaction being dropped entirely.
* The only other relevant place is `TxQ::tryClearAccountQueueUpThruTx`, and if `preclaim` fails there, the function exits and falls back to "the normal process".

One possible negative interaction I can think of is that if the transaction is constructed such that the `NotValidAfter` time is really close to the current time, then the transaction could get queued and broadcast only to become invalid before there's any chance of succeeding.  This is parallel to the `LastLedgerSequence`; the `TxQ` checks that it has a sufficiently large buffer here: https://github.com/ripple/rippled/blob/develop/src/ripple/app/misc/impl/TxQ.cpp#L387-L395 A similar check should be added there for `NotValidAfter`. `minimumLastLedgerBuffer` is 2, so the time buffer should be roughly equivalent to the number of seconds for 2 ledgers, preferably "rounded up". i.e. 10 seconds.

mDuo13:
Time-based and ledger-index-based validity limits each have their advantages. As @nbougalis says, the time-based ones are probably better for human purposes, since the pace of ledger advancement is unpredictable and human activities are mostly scheduled based on real-world time.

`FirstLedgerSequence` (#2945) would be simpler for setting a self-contained range of ledgers so that you can definitively say whether a transaction succeeded or failed without needing to know external information like how early the transaction could've possibly confirmed. (Or full history.)

src/ripple/app/tx/impl/Transactor.cpp
```diff
@@ -448,6 +465,31 @@ Transactor::apply()
...
+{
+    using d = NetClock::duration;
+    using tp = NetClock::time_point;
+
+    if (ctx.view.rules().enabled(featureTemporalTxValidity))
+    {
+        // We check time against the close time of the parent ledger,
+        // because that is the time that the network, as a whole,
+        // agrees on.
+        auto const pct = ctx.view.parentCloseTime();
```
pwang200, line 479: Using `parentCloseTime` makes sense. 
Many people probably already know, but I was curious about what the `parentCloseTime` looked like, so I did some measurement. I connected my local machine to the mainnet overnight, and recorded my local time when ledgers were closed and the network agreed close time. 
The network agreed time follows the pattern, 0, 1, 2, 10, 11, 12, 20, 21, 30... There are the same number of `parentCloseTime` ending with 0 and ending with 1. There are less `parentCloseTime` ending with 2, (about 22%). 
The result of ```network agreed close time - my local time of closing``` are shown in the histogram below.  
Figure_1
It seems the differences are in the range [-3, 6]. (I had one outlier at -12 when my machine lost sync.)  

When clients assign `NotValidBefore` or `NotValidAfter`, all the practical values (in tens of seconds) should be fine to avoid false positive. (Clients should also consider the time buffer (10 seconds to `NotValidAfter`) that @ximinez suggested.

src/ripple/app/tx/impl/Transactor.cpp
```diff
@@ -448,6 +465,31 @@ Transactor::apply()
...
+
+        if (auto const cutoff = ctx.tx[~sfNotValidBefore];
+            cutoff && (pct < tp{d{*cutoff}}))
+            return tefTOO_EARLY;
```
pwang200, line 483: I feel `tefTOO_EARLY` and `tefTOO_LATE` are slightly different since a tx that is too early could be retried without modification and a tx that is too late cannot be retried. So could `tefTOO_EARLY` be `terTOO_EARLY`?
(Sorry I am not sure the consequences if we change, such as its behavior in `TxQ`. I could run some tests later if needed.)

src/ripple/protocol/impl/TxFormats.cpp
```diff
@@ -55,6 +55,8 @@ TxFormats::TxFormats()
             {sfClearFlag, soeOPTIONAL},
             {sfTickSize, soeOPTIONAL},
             {sfTicketSequence, soeOPTIONAL},
+            {sfNotValidBefore, soeOPTIONAL},
+            {sfNotValidAfter, soeOPTIONAL},
```
pwang200, line 59: The new fields `sfNotValidBefore` and `sfNotValidAfter` are comparable to the existing field `sfLastLedgerSequence` inside `commonFields`. So could the new fields be promoted to `commonFields` since they are also `soeOPTIONAL`?

pwang200 (commented):
I think this could be very useful (especially in one of my projects), since the client can assign the values using its local clock (with a practice time buffer). To use `LastLedgerSequence`, however, the client must have a good idea of the current ledger sequence.

src/ripple/app/tx/impl/Transactor.cpp
mDuo13: Added a section in the docs to clarify this: https://github.com/ripple/xrpl-dev-portal/pull/1073

intelliot:
> `FirstLedgerSequence` (#2945) would be simpler for setting a self-contained range of ledgers so that you can definitively say whether a transaction succeeded or failed without needing to know external information like how early the transaction could've possibly confirmed. (Or full history.)

@mDuo13 would `NotValidBefore` successfully satisfy the same envisioned use cases that `FirstLedgerSequence` would satisfy?

i.e. would `NotValidBefore` allow for easily calculating a self-contained range of ledgers without external information or full history?

If so, then we should push for `NotValidBefore` and close #2945; but if not, then we should consider adding support for `FirstLedgerSequence` to this amendment.

ximinez:
> @mDuo13 would `NotValidBefore` successfully satisfy the same envisioned use cases that `FirstLedgerSequence` would satisfy?

IMHO, I think there's room and not-entirely-overlapping use cases for both options. Off the top of my head, `NotValidBefore` can be used in cases where you're creating a transaction offline or intended to be applied in the relatively far future. `FirstLedgerSequence` can be used when you're live with the ledger, want something in the next few ledgers, and want to be certain of which ledgers you need to search to find the transaction.

In terms of application, `NotValidBefore` would probably be more useful to make a revocable promise of sorts that something will happen after a specific time in the future, while `FirstLedgerSequence` might be more useful for reliable transaction submission. ("Revocable" in this case means the account could submit a second, unrestricted, transaction with the same sequence before the `NotValidBefore` time of the first transaction, thus rendering it defunct.)

> but if not, then we should consider adding support for FirstLedgerSequence to this amendment.

That's a pretty good idea.

scottschurr:
Automated unit tests for the win! :tada:

This branch has a bunch of failures in the KnownFormatsToGRPC unit test.  That's because gRPC has not yet been taught about the newly added fields.  If we educate gRPC, then the unit tests pass without even touching the unit test!

Here's a commit you could cherry-pick that will get gRPC caught up on the changes: https://github.com/scottschurr/rippled/commit/9204fea649635e27bf8d3a181a87f58a569d9e84  The changes may not be perfect.  You should check with @cjcobb23 to see if he wants any changes.

scottschurr (commented):
Implementation seems really reasonable.  Unit test coverage is good.  I left a few comments for your consideration.  As @ximinez noted, this could probably use a tweak or two in the `TxQ`.  However the required tweaks may not be too bad if we set a reasonable rule for the minimum valid window.

I don't personally have an opinion about whether `FirstLedgerSequence` should be added as well.  I'll leave that decision to the user facing folks like @mDuo13 and @intelliot.  But `FirstLedgerSequence` should be easy to add to this pull request if it is desirable.  It too would need a minimum valid window.

mDuo13:
> @mDuo13 would `NotValidBefore` successfully satisfy the same envisioned use cases that `FirstLedgerSequence` would satisfy?
> 
> i.e. would `NotValidBefore` allow for easily calculating a self-contained range of ledgers without external information or full history?

Short answer, you can do it, but it's more steps.

Long answer: OK, so, first off, proving whether or not a transaction has executed is a devil's proof: if you find affirmative evidence that it has, then you're done; but the only way to prove that a transaction has _not_ executed is to define a specific, finite, and complete set of circumstances in which the transaction could have executed, then exhaustively examine them all to confirm that it did not, in fact, execute in any of those circumstances.

The generally accepted approach is to first define a range of ledgers that the transaction could have executed in, then confirm that you are in possession of every ledger in that range (and they're all final/validated), then finally confirm that the transaction is in none of the ledgers in question. With the `min_ledger` and `max_ledger` fields of the `tx` command, you can do the second and third steps in a single atomic command.

So let's talk about the first step, defining the range of ledgers.

The `LastLedgerSequence` parameter very neatly puts an upper limit on the range of ledgers. It's a signed part of the transaction _and_ it uniquely identifies a validated ledger in a way that you can look it up directly. (Aside: I am using "ledger index" to mean the same as ledger sequence number.)

At present, the only way to put a lower limit on the range of ledgers (other than full history) is to know additional information about which ledgers were validated when the transaction was submitted, and track that alongside the transaction itself. Furthermore, there's no inherent proof that this information is accurate; if the transaction was first submitted earlier than you thought it was, that invalidates any conclusions you drew about the transaction's non-execution. And, since this information not a signed part of the transaction, you cannot prove that it's correct. `FirstLedgerSequence` would solve this problem trivially in the exact same way `LastLedgerSequence` does for the upper bound.

So, finally, let's talk about `NotValidBefore` and `NotValidAfter` in this context. As real-world times, these values are related to ledger indexes indirectly. The rate and exact time of ledger closes is non-deterministic, but it is at least true that ledger close times are monotonically increasing, so it _is_ possible to map close times to ledgers so that you can be certain you know the most recently closed ledger at any point in time. So, you _can_ use these fields to set a range on the ledgers in which a transaction could have executed and thereby prove that the transaction did not execute. However, a challenge presents itself in that ledgers are not typically indexed for lookup by close time. (Not that you _couldn't_; the Data API used to do this.) But, realistically, you have to crawl through ledgers until you find the one with the last close time that is before the `NotValidBefore` value, and note that ledger's index. Same for `NotValidAfter` if you're using that instead of `LastLedgerSequence`. Then you still need more information to ensure there are no gaps, because you cannot tell from looking at two close times alone whether or not there was a ledger in between. To conclusively recognize that you have "the first ledger that could possibly include transactions with a `NotValidBefore` value of X" you need _two_ ledgers and at least 4 pieces of information: 

1. A close time that is after `NotValidBefore` 
2. A close time that is _before_ the `NotValidBefore`.
3. Either (a) the `parent_hash` or the (b) `ledger_index` or the of the "after" ledger
4. Either the `ledger_hash` of the "before" ledger, which must match (a); or the `ledger_index` of the "before" ledger, which must be equal to (b) minus 1.

Determining the upper end of the range using `NotValidAfter` follows the same pattern. And, of course, you need to ensure that all the ledgers in question are validated.

With either the time-based or index-based fields, the goal is to be able to conclude definitively, "This transaction has not executed," using the _signed_ (therefore, immutable) transaction instructions and the server's database(s) of ledgers and transactions. But the `tx` command, as implemented, does not take time-based arguments similar to `min_ledger` and `max_ledger`. There is no single command that can look up the latest validated ledger at a given point in time. And, as described above, proving that a given ledger represents the edge of a range requires looking up the headers and probably the ledger indexes of multiple ledgers. By contrast, looking up whether a ledger index is the edge of the range is trivial and inherent to the idea of ledger indexes.

tl;dr please implement `FirstLedgerSequence`.

src/ripple/app/tx/impl/Transactor.cpp
mDuo13: > @scottschurr wrote:
> There may be some value in supplying both a terTOO_EARLY and a tefTOO_EARLY. The tefTOO_EARLY code is the more general one. If the transaction is a day, an hour, or even 5 minutes early, then everyone is better off if the transaction simply fails. The number of retries that the server is willing perform will not make a transaction that is 5 minutes early succeed.
>
> However, if the transaction is 10 seconds early or less, then it makes sense for the server to retry the transaction a few times. This is the case where terTOO_EARLY makes sense.
>
> The question we have to ask is whether it's worth adding code to support the "just a little bit early" case. I could go either way, but I have a slight bias that the added complexity does not bring enough usability value.

Rhetorical question: does the distinction between "ter" and "tef" classes exist for the sake of the user, or for the sake of the server? Is the purpose to make it easier for the user to know how to process the outcome of the transaction? Or is the purpose to direct the server in how to handle the failure?

If the answer is for the sake of the user, then this should be a `ter` code, to signal to the user, "This transaction, unmodified, could become valid later."

If the answer is for the sake of the server, then making it a `tef` code so that the server does not need to retry it fruitlessly makes sense.

scottschurr: It turns out that `tem` and `ter` codes are handled somewhat differently inside the server.  So the codes are intended for both parties.  But which one you select has consequences for how long the server holds onto the transaction before giving up.  The server will work harder at pushing a `ter` into the ledger, or at least that was true the last time I looked.  I have to admit, these things evolve over time...

ximinez:
> But `FirstLedgerSequence` should be easy to add to this pull request if it is desirable. It too would need a minimum valid window.

I don't think `FirstLedgerSequence` needs any kind of range checking, at least not in the `TxQ`. A transaction must return a `tes` or `tec` from `preclaim` to be considered for queueing. If the transaction is too early, it'll return a `ter` or `tef` (doesn't matter with in this case), which makes it ineligible for being broadcast and queued.

RichardAH:
Adding a few thoughts here:
`FirstLedgerSequence` is used by Emitted Transactions in Hooks (as an absolute bar to an emitted txn being included in the same ledger it was emitted). It is very easy to implement.

Transaction logic based around human times are best implemented for the specific case in a specific Hook, that then emits the required txn at the right time. As soon as you want any sort of slightly complex logic like `NotValidBeforeUnlessCondition` adding fields to transactors becomes infeasible.

intelliot:
(notes: needs unit tests; needs to address open comments; this change may not be needed at all)

intelliot:
@nbougalis still interested in seeing this through?

I have no opinion either way, but it will experience bitrot if left alone. We should decide to get it in, or not do it (yet) - and if we decide on the latter, we can close the PR.

intelliot:
This seems like a useful feature that is likely to be added at some point, but I've seen minimal real user demand. Were there any users (external or even internal) who would certainly use this feature, if deployed and activated?

If we don't have certainty that it will be used, it might not be worth doing right now. I also see that there are merge conflicts, so at minimum, those would need to be resolved - in addition to code reviews and testing. @nbougalis any thoughts on whether this makes sense to move forward right now?

We could close this PR until it makes sense to resurrect the idea.

intelliot:
Closing due to inactivity; however, I expect this feature to be revisited at some point in the future. At that time, feel free to re-open this PR, or open a new one.

PR #3842, opened 2021-05-07
Clean up Section in BasicConfig.h

scottschurr:

During other work I noticed that `Section`, defined in BasicConfig.h, was dependent on `beast::unit_test::detail::const_container` and the dependency was really unnecessary.  It was also the only file in the code base with such a dependency.  While removing the dependency I noticed a few other things that could be cleaned up in `Section`...

### Context of Change

The following changes were made:
- Removed dependency on `const_container` template defined in `beast` detail namespace.
- Removed `Section::find()` method which had an obsolete interface.
- Made `Section::get<>()` easier to use for the common case of retrieving a `std::string`.  The revised `get()` method replaces old calls to `Section::find()`.
- Provided a default template parameter to free function `get<>(Section config, std::string name)` so it stays similar to `Section::get<>()`.

Then the rest of the code was adapted to these changes.

- Calls to `Section::find()` were replaced with calls to `Section::get()`.
- Unnecessary `get<std::string>()` template arguments were reduced to `get()`.

These changes dug up an interesting artifact in the SHAMap unit tests.  I'm not sure why the tests were working before, but there was a character case problem in a `Section` key.  The unit test is fixed.

### Type of Change

This change should have no effect on behavior, so it doesn't need to be mentioned in the release notes.

### DIFF
--- a/src/ripple/app/reporting/ReportingETL.cpp
+++ b/src/ripple/app/reporting/ReportingETL.cpp
@@ -27,6 +27,8 @@
 #include <boost/asio/ip/tcp.hpp>
 #include <boost/beast/core.hpp>
 #include <boost/beast/websocket.hpp>
+#include <cctype>
+#include <charconv>
 #include <cstdlib>
 #include <iostream>
 #include <string>
@@ -855,29 +857,26 @@ ReportingETL::ReportingETL(Application& app)
             JLOG(journal_.debug()) << "val is " << v;
             Section source = app_.config().section(v);
 
-            std::pair<std::string, bool> ipPair = source.find("source_ip");
-            if (!ipPair.second)
+            auto optIp = source.get("source_ip");
+            if (!optIp)
                 continue;
 
-            std::pair<std::string, bool> wsPortPair =
-                source.find("source_ws_port");
-            if (!wsPortPair.second)
+            auto optWsPort = source.get("source_ws_port");
+            if (!optWsPort)
                 continue;
 
-            std::pair<std::string, bool> grpcPortPair =
-                source.find("source_grpc_port");
-            if (!grpcPortPair.second)
+            auto optGrpcPort = source.get("source_grpc_port");
+            if (!optGrpcPort)
             {
                 // add source without grpc port
                 // used in read-only mode to detect when new ledgers have
                 // been validated. Used for publishing
                 if (app_.config().reportingReadOnly())
-                    loadBalancer_.add(ipPair.first, wsPortPair.first);
+                    loadBalancer_.add(*optIp, *optWsPort);
                 continue;
             }
 
-            loadBalancer_.add(
-                ipPair.first, wsPortPair.first, grpcPortPair.first);
+            loadBalancer_.add(*optIp, *optWsPort, *optGrpcPort);
         }
 
         // this is true iff --reportingReadOnly was passed via command line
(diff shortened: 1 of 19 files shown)


src/ripple/app/reporting/ReportingETL.cpp
```diff
@@ -904,21 +901,20 @@ ReportingETL::ReportingETL(Application& app, Stoppable& parent)
         // if not passed via command line, check config for start sequence
         if (!startSequence_)
         {
-            std::pair<std::string, bool> start = section.find("start_sequence");
-            if (start.second)
+            auto const optStartSeq = section.get("start_sequence");
+            if (optStartSeq)
             {
-                startSequence_ = std::stoi(start.first);
+                startSequence_ = std::stoi(*optStartSeq);
```
undertome, line 907: If the `start_sequence` value supplied in the configuration is invalid, `std::stoi` will cause the program to crash without providing helpful output as to the cause. The same applies for `flushInterval_` and `numMarkers_`.

src/ripple/basics/BasicConfig.h
```diff
@@ -40,15 +39,18 @@ using IniFileSections = std::map<std::string, std::vector<std::string>>;
 /** Holds a collection of configuration values.
     A configuration file contains zero or more sections.
 */
-class Section : public beast::unit_test::detail::const_container<
-                    std::map<std::string, std::string, boost::beast::iless>>
+class Section
 {
 private:
     std::string name_;
+    std::map<std::string, std::string> lookup_;
```
undertome, line 46: Why not `std::unordered_map` instead? We'd get better performance on inserts and lookups. The drawback is that we could get inconsistent results when using the stream operator. If that's a deal breaker then we could add a sorting feature to the stream operator, which won't be terrible since performance isn't critical for that operation.

src/ripple/app/reporting/ReportingETL.cpp
scottschurr: @undertome, thanks for pointing that out.  @HowardHinnant, do you see any downside to replacing all uses of  `std::stoi()` in the code base with calls to `std::from_chars()`?  Or can you suggest a better replacement?

HowardHinnant: `from_chars` sounds good to me as long as we always check the return for error, and if we have something sane to do on error.  If we don't have a sane error recovery strategy, then letting `stoi` throw doesn't sound unreasonable.

src/ripple/basics/BasicConfig.h
scottschurr: I used `std::map` because we were previously inheriting from `beast::unit_test::detail::const_container<std::map...>`.  Leaving it as `std::map` was the smallest change that accomplished the goal of removing this dependency on `beast::unit_test::detail`.

That said, I think keeping `lookup_` as a `std::map` is a good conservative choice.  The `Section`'s container type is exposed to users through the `begin()` and `end()` accessors.  If we change `lookup_` to a `std::unordered_map`, then we'll need to audit all users of those iterators to make sure they are not depending on `lookup_` being pre-sorted.

That's a long way of saying that leaving it a `std::map` is the conservative change.  I'm not convinced the speed improvement from changing to a `std::unordered_map` justifies the additional work and risk.

src/ripple/basics/BasicConfig.h
```diff
@@ -40,15 +39,18 @@ using IniFileSections = std::map<std::string, std::vector<std::string>>;
...
     std::vector<std::string> lines_;
     std::vector<std::string> values_;
     bool had_trailing_comments_ = false;
 
+    using const_iterator =
+        typename std::map<std::string, std::string>::const_iterator;
```
undertome, line 52: Nit: The `typename` keyword isn't needed for this declaration. Also consider `decltype(lookup_)` for concision and less redundancy.

src/ripple/nodestore/backend/CassandraFactory.cpp
```diff
@@ -241,16 +239,14 @@ class CassandraBackend : public Backend
             Throw<std::runtime_error>(ss.str());
         }
 
-        std::string username = get<std::string>(config_, "username");
+        std::string username = get(config_, "username");
         if (username.size())
         {
-            std::cout << "user = " << username.c_str() << " password = "
-                      << get<std::string>(config_, "password").c_str()
+            std::cout << "user = " << username.c_str()
```
undertome, line 245: Is there a benefit to invoking `c_str()`? Also applies to lines 246 and 249.

undertome:
Overall looks good - just left a few comments.

src/ripple/nodestore/backend/CassandraFactory.cpp
scottschurr: Good eyes, @undertome.  The `.c_str()` calls on lines 245 and 246 serve no useful purpose.

The `.c_str()` calls on line 249 I left alone, however.  I believe the database folks are invoking a function that expects a pointer to a null terminated C-style string.  Things _compile_ if I remove the calls to `c_str()`.  But I'm not 100% confident that still provides the desired behavior.  So I'm inclined to leave it alone.

src/ripple/app/reporting/ReportingETL.cpp
scottschurr: @undertome, with @HowardHinnant's guidance in mind I looked harder at your question.  Everything gets more difficult when you take (consistent) error handling into account.

If we look higher in `ReportingETL`'s constructor we see that `throw`ing without logging is the expected error handling mechanism.  That makes a certain amount of sense.  The log file may not have been set up this early while assembling the server.

As you point out, `std::stoi()` does throw, but the error that it throws does not provide enough context for a user to figure out what's going on.  

I came up with an approach that may improve the error handling in `ReportingETL`'s constructor.  You can see it here:

https://github.com/scottschurr/rippled/commit/5a36c6b271aa12ad87edbf188a133ab509728224

Unfortunately there is zero unit test coverage in that portion of the code.  So I lack confidence in the proposed change.  If @cjcobb23 looks at it and thinks it's okay I'm good with including the commit.  Otherwise I think the change is too risky.

scottschurr:
@undertome, thanks for the review.  I've added a `[FOLD]` commit to address your comments.

src/ripple/app/reporting/ReportingETL.cpp
cjcobb23: This patch looks good @scottschurr . Crashing with a useful error message makes sense.

src/ripple/nodestore/backend/CassandraFactory.cpp
cjcobb23: `cass_cluster_set_credentials()` expects the params to be `const char *`. The other two uses are fine to change. This whole file is only compiled if the `RIPPLED_REPORTING` macro is defined.

cjcobb23 (approved):
LGTM. Thanks for cleaning this up.

src/ripple/app/reporting/ReportingETL.cpp
scottschurr: Thanks @cjcobb23, I've added that patch to this branch.

undertome (approved):
Looks good - thanks for addressing the comments!

scottschurr:
Thanks for the reviews @undertome and @cjcobb23.  Rebased and squashed.  Once CI passes I'll mark the pull request as passed.

PR #3846, opened 2021-05-10
Add vl.xrplf.org to validators.txt

bharathchari:
Add the list provider vl.xrplf.org (XRP Ledger Foundation) to validators.txt and include the keys.


manojsdoshi (approved):
LGTM

PR #3853, opened 2021-05-24
Maintenance of Offer and PayStrand unit tests

scottschurr:

While doing other work I came across two unit tests that needed maintenance.

### Context of Change

1. Sections of Offer_test.cpp had been made unreadable by clang-format.  For those only those unreadable sections I turned off clang-format and returned the formatting to what was there before clang-format was applied.  These changes are formatting only.  No tests were added, removed, or modified.
2. I discovered an early return in PayStrand_test.cpp which caused a number of the test to be skipped.  I removed the early return and then, with help from @seelabs, repaired the tests that had not been executing and were now failing.

### Type of Change

- [x] Bug fix
- [x] Tests (You added tests for code that already exists)

Only unit tests were affected by this change, so it does not need to be mentioned in the release notes.


seelabs (changes requested):
One test needs to be fixed (it's my fault it was broken, not yours). I provided a patch. :+1: after that patch is applied.

seelabs (approved):
:+1: Thanks for noticing and fixing these!

thejohnfreeman (approved):

PR #3854, opened 2021-05-24
[Trivial] Fix clang warnings about copies vs references:

scottschurr:

Fix warnings identified by a recent version of clang.

### Context of Change

A recent version of clang warns regarding several ranged for loops where the code base was either:

- making unnecessary copies or
- was extending life times of temporaries referred to by `const` lvalue references.

This fixes the places that clang identified.

The warning are raised by `Apple clang version 12.0.0 (clang-1200.0.32.21)`.  I can't tell you which trunk version of clang it is, since Apple has their own clang versioning scheme.

### Type of Change

This change does not affect functionality, so there is no need for it to be mentioned in the release notes.

But it will allow me to build using clang while warnings-as-errors is set, so I hope it merges to develop soon... 😄

### DIFF
--- a/src/test/csf/BasicNetwork_test.cpp
+++ b/src/test/csf/BasicNetwork_test.cpp
@@ -49,7 +49,7 @@ public:
             auto t = scheduler.in(1s, [&] { set.insert(0); });
             if (id == 0)
             {
-                for (auto const& link : net.links(this))
+                for (auto const link : net.links(this))
                     net.send(this, link.target, [&, to = link.target] {
                         to->receive(net, this, 1);
                     });
@@ -68,7 +68,7 @@ public:
             ++m;
             if (m < 5)
             {
-                for (auto const& link : net.links(this))
+                for (auto const link : net.links(this))
                     net.send(this, link.target, [&, mm = m, to = link.target] {
                         to->receive(net, this, mm);
                     });
--- a/src/test/rpc/NoRipple_test.cpp
+++ b/src/test/rpc/NoRipple_test.cpp
@@ -84,7 +84,7 @@ public:
 
         // fix1578 changes the return code.  Verify expected behavior
         // without and with fix1578.
-        for (auto const tweakedFeatures :
+        for (auto const& tweakedFeatures :
              {features - fix1578, features | fix1578})
         {
             Env env(*this, tweakedFeatures);
--- a/src/ripple/app/misc/impl/ValidatorList.cpp
+++ b/src/ripple/app/misc/impl/ValidatorList.cpp
@@ -99,7 +99,7 @@ void
 ValidatorList::PublisherListStats::mergeDispositions(
     PublisherListStats const& src)
 {
-    for (auto const [disp, count] : src.dispositions)
+    for (auto const& [disp, count] : src.dispositions)
     {
         dispositions[disp] += count;
     }
--- a/src/ripple/app/misc/impl/ValidatorSite.cpp
+++ b/src/ripple/app/misc/impl/ValidatorSite.cpp
@@ -408,7 +408,7 @@ ValidatorSite::parseJsonResponse(
     sites_[siteIdx].lastRefreshStatus.emplace(
         Site::Status{clock_type::now(), applyResult.bestDisposition(), ""});
 
-    for (auto const [disp, count] : applyResult.dispositions)
+    for (auto const& [disp, count] : applyResult.dispositions)
     {
         switch (disp)
         {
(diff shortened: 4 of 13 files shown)


src/test/rpc/NoRipple_test.cpp
```diff
@@ -84,7 +84,7 @@ class NoRipple_test : public beast::unit_test::suite
 
         // fix1578 changes the return code.  Verify expected behavior
         // without and with fix1578.
-        for (auto const tweakedFeatures :
+        for (auto const& tweakedFeatures :
```
ximinez, line 87: Interesting. I found this one, but only this one, with VS 2019 over at https://github.com/ripple/rippled/pull/3851/commits/a51cfddc0aea40387d7ab578e68b6ebb7beee927#diff-9bec26b172c29e97751e1152a678d71aadb13f48bb3f62fe6962113dbdca7910. I doubt it'll cause any merge problems, but it's interesting.

HowardHinnant (approved):

src/test/csf/BasicNetwork_test.cpp
```diff
@@ -49,7 +49,7 @@ class BasicNetwork_test : public beast::unit_test::suite
             auto t = scheduler.in(1s, [&] { set.insert(0); });
             if (id == 0)
             {
-                for (auto const& link : net.links(this))
```
thejohnfreeman, line 52: I'm guessing this was flagged as extending the lifetime of the temporary? What exactly is wrong with that in this case?

thejohnfreeman (commented):
I don't understand the concern it has over lifetime extension. Could you explain it to me? I picked out one example, but the specific example is not important. I would like to learn the rule.

src/test/csf/BasicNetwork_test.cpp
scottschurr: There's actually nothing "wrong".  C++ does the lifetime extension, so there's no actual risk there.  But, on the other hand, reading the code is misleading.  The source code _looks_ like you're taking a reference, but you've actually made a copy.  So it's more expensive than it looks.

Warnings as errors is a little tricky.  A compiler is welcome to warn about anything that it wants to, even valid C++.  That's what is happening in this case.

thejohnfreeman: So before, this would create a temporary from the expression `net.links(this)`, then construct two iterators from that temporary with `std::begin($temporary)` and `std::end(temporary)`, then iterate with that iterator, binding a reference to the return value of `*$iterator`, which may itself be a temporary and not a reference? Is that right?

And now it does all the same thing, except it copies the return value of `*$iterator` into a named variable? And if `*$iterator` returns a pr-value that is not a reference, this is the same behavior? But if it returns a reference, then this is an additional copy? But the compiler is telling us it's not a reference so we can safely leave off the `&` in our binding? Am I getting all that right?

From my current understanding, it seems like no performance benefit to drop the `&` in this case, but that if the iterator constructed by `net.links(this)` ever changes to return a reference, it will. And I understand it is unlikely we ever change the implementation of that iterator, but binding a reference is a simple rule to follow that is always safe and optimal performance, that doesn't require me to know the exact return type of the iterator dereference. Is that right?

scottschurr: Your understanding is correct.  There is no performance benefit to removing the `&` here.  And it's certainly possible to argue that if the container's content is not modified during the iteration, and copies of the contents are not desired, then we should always use `const &`.  C++ does provide lifetime extension of temporaries bound to `const &`.

You are also correct that if the iterator were to change in the future, so it returns by reference rather than by value, then this loop (without the `&`) would start making copies.

This second case would not happen, however, as long as clang continues to warn about unnecessary copies in ranged `for` loops.  I personally believe we can expect this warning to be generated (as long as we don't explicitly turn it off).

So my personal take is that:

- We should rely on clang generating the unnecessary copies warning in ranged `for` loops, and
- With assistance from those warnings, have the source code reflect what's actually happening, rather than have hidden copies that are managed by lifetime extension.

But that's not the only viable position to take.  Another approach would be for us to turn off this warning from clang.  Or to turn off warnings as errors.  Both of those changes feel like a step backward to me.  But your milage may vary.

thejohnfreeman: I like warnings as errors, and my personal preference would be to turn off this warning, but I don't feel strongly enough to contest it. Just wanted to make sure I understood what was going on. Thank you for the answer.

thejohnfreeman (approved):

PR #3856, opened 2021-06-01
[TRIVIAL] Do not use __func__ inside lambdas

seelabs:

Using `__func__` inside a lambda with output `operator()`. The intended functionality is to output the enclosing function name. To do this, we capture `__func__` outside the lambda.
### Type of Change

## Before / After

Before
```
foo(){
[]{
    std::cout << __func__ << '\n'; //prints "operator()"
}
}
```

After
```
foo(){
[fname = __func__]{
    std::cout << fname << '\n'; //prints "foo"
}
}
```


cjcobb23 (approved):
Nice fix, LGTM.

intelliot (approved):

PR #3859, opened 2021-06-03
Fix nested locks in ValidatorSite

ximinez:

Fixes a deadlock caused by an attempt to take the same lock recursively. Modifies unit tests to exercise the issue.

### Context of Change

https://github.com/ripple/rippled/commit/2c71802e389a59118024ea0152123144c084b31c introduced a change to ValidatorSite downloads. If a download succeeds, the VL will be written to a cache file in the data directory, but that cache file is not normally read. If the download fails, the site manager will add any existing cache files to the list of sites. Apparently that particular case never got tested because the call sequence is:
* `onSiteFetch` - takes the `sites_mutex_` lock
* `onError` lambda
* `missingSite`
* `load` - also takes the `sites_mutex_` lock
* Frozen thread

This will prevent rippled from shutting down.

I included two commits in this PR, which should be squashed before merging. The first is the fix, the second is a unit test which reproduces the issue. If the test is run without the fix, the `ValidatorSite` test suite will hang at the first `redirect_forever` test.

### Type of Change

## Before / After

I added a private `ValidatorSite::load` overload to take a reference to the lock when called internally. The lock now gets passed from `onSiteFetch` to `onError` to `missingSite` to the new `load` overload.

I also changed some of the existing functions which take locks to take a `const&` instead of just a `&`.

## Test Plan

1. Set up a node with only the vl.ripple.com validator list site, and run the node long enough for a cache file to be created in the `database_path` folder. Stop the node
   * Or use a preexisting node that already has a cache file.
2. Add another validator list site containing an validly formed URL that does not lead to a real page, or that will have some other connection problem.
3. Start the node up.
4. Try to stop the node after a few minutes.
5. Old behavior: Node will hang and be unable to stop. New behavior: Node will shut down correctly.

### DIFF
--- a/src/ripple/app/misc/ValidatorSite.h
+++ b/src/ripple/app/misc/ValidatorSite.h
@@ -194,10 +194,16 @@ public:
     getJson() const;
 
 private:
+    /// Load configured site URIs.
+    bool
+    load(
+        std::vector<std::string> const& siteURIs,
+        std::lock_guard<std::mutex> const&);
+
     /// Queue next site to be fetched
     /// lock over state_mutex_ required
     void
-    setTimer(std::lock_guard<std::mutex>&);
+    setTimer(std::lock_guard<std::mutex> const&);
 
     /// request took too long
     void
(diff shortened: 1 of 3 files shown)


seelabs (approved):
:+1: Thanks for taking this on!

undertome (approved):
Looks good to me. Separating the commits made it easy to observe the deadlock and confirm that the fix resolves it.

ximinez:
Squashed and rebased onto 1.8.0-b3.

PR #3860, opened 2021-06-06
Simplify processing and verifying signer lists:

a-noni-mousse:
I am trying this PR again.

The commit reduces the scope of the signer list parsing code to the
Transactor and simplifies the SetSignerList transactor. It introduces
an amendment because it requries the signer list to be ordered when
submitting transactions that modify the list.

### Type of Change

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)


src/ripple/app/tx/impl/Transactor.h
```diff
@@ -82,8 +83,20 @@ struct PreclaimContext
 class TxConsequences;
 struct PreflightResult;
 
+/** Represents the entry in a singer list */
+using SignerEntry = std::pair<AccountID, std::uint16_t>;
```
thejohnfreeman, line 87: A middle ground, if you would like to keep the meaningful field names, is to just remove the constructor from the old `struct`, treat it as an aggregate, and define `operator<` for it.

src/ripple/app/tx/impl/SetSignerList.cpp
```diff
@@ -78,31 +88,77 @@ SetSignerList::determineOperation(
...
-        // Validate our settings.
-        auto const account = ctx.tx.getAccountID(sfAccount);
-        NotTEC const ter = validateQuorumAndSignerEntries(
-            std::get<1>(result), std::get<2>(result), account, ctx.j);
-        if (ter != tesSUCCESS)
+        auto const result = determineOperation(ctx.tx);
+        if (std::get<0>(result) != tesSUCCESS)
+            return std::get<0>(result);
+
+        if (std::get<3>(result) == unknown)
```
thejohnfreeman, line 144: How about creating some named constants for these fields of the tuple, e.g. `TEC_` and `OPERATION_`?

src/ripple/app/tx/impl/SetSignerList.cpp
```diff
@@ -78,31 +88,77 @@ SetSignerList::determineOperation(
...
-    if (!isTesSuccess(ret))
+    if (auto const ret = preflight1(ctx); !isTesSuccess(ret))
         return ret;
 
-    auto const result = determineOperation(ctx.tx, ctx.flags, ctx.j);
-    if (std::get<0>(result) != tesSUCCESS)
-        return std::get<0>(result);
-
-    if (std::get<3>(result) == unknown)
+    if (ctx.rules.enabled(featureSimplifiedSetSignerList) &&
```
thejohnfreeman, line 94: The "simplified" code path starting here largely duplicates most of the existing code path. Is it possible to de-duplicate that code to highlight the difference? What is the difference exactly?

thejohnfreeman (commented):
Can we do a cost-benefit analysis on introducing an amendment for this? Amendments have significant costs: maintaining and testing multiple code paths, forcing validator operators to update, building community support for ratification, planning eventual [retirement](https://github.com/XRPLF/XRPL-Standards/discussions/19). What are the benefits that _require_ an amendment? It seems to me that the only benefit that _requires_ an amendment is the removal of one call to `std::sort` on a `std::vector` of no more than 8 22-byte objects. If that's the case, then it doesn't seem worth it to me to introduce an amendment, but I would like to hear the argument for it.

cryptobrad:
I think that the requirement of sorting the list helps ensure a canonical repreesntation of the transaction but it would require the change in all librries so yes you are right not a good idea to do it. I can change that but many of my changes get stuck and nothing happens with them which is annoying thing.

thejohnfreeman:
I had reviewed this 4 days after it was submitted, and none of the questions were addressed in over a year. Why is it stuck?

intelliot:
(notes: not ready to merge; waiting for response from author)

If we don't hear from @a-noni-mousse in 1 month, we'll close this.

intelliot:
Closing due to inactivity. @a-noni-mousse if you'd like to return to this PR, feel free to re-open or open a new PR.

PR #3861, opened 2021-06-07
Some code cleanups tagged by static analysis

seelabs:

Misc code cleanups tagged by static analysis. This include reducing copies, removing unneeded moves, and adding explicit.

### Type of Change

### DIFF
--- a/src/ripple/app/rdb/backend/RelationalDBInterfaceSqlite.cpp
+++ b/src/ripple/app/rdb/backend/RelationalDBInterfaceSqlite.cpp
@@ -174,13 +174,13 @@ public:
     bool
     transactionDbHasSpace(Config const& config) override;
 
-    int
+    std::uint64_t
     getKBUsedAll() override;
 
-    int
+    std::uint64_t
     getKBUsedLedger() override;
 
-    int
+    std::uint64_t
     getKBUsedTransaction() override;
 
     void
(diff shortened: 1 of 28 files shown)


src/ripple/app/rdb/backend/RelationalDBInterfaceSqlite.cpp
```diff
@@ -174,13 +174,13 @@ class RelationalDBInterfaceSqliteImp : public RelationalDBInterfaceSqlite
...
+    std::uint64_t
     getKBUsedAll() override;
 
-    int
+    std::uint64_t
     getKBUsedLedger() override;
 
-    int
+    std::uint64_t
     getKBUsedTransaction() override;
```
HowardHinnant, line 185: The use of these three functions in src/ripple/rpc/handlers/GetCounts.cpp needs updating to reflect the change from `int` to `std::uint64_t`.  Also in src/ripple/core/SociDB.h we're using `size_t` for these returns.  We should make these the same type alias.

HowardHinnant (changes requested):
Looks good.  Just one minor comment that needs a small fix.

src/ripple/app/rdb/backend/RelationalDBInterfaceSqlite.cpp
seelabs: Thanks! I changed them all to `std::uint32_t`. Fixed in `975aebe35c [fold] Measure db memory usage with a uint32 var
`

HowardHinnant (approved):

intelliot (approved):
I'm curious whether we expect this to have any impact on performance test results?

seelabs:
@intelliot While these changes do reduce allocations, I doubt it's measurable in anything but a micro-benchmark.

PR #3863, opened 2021-06-09
Don't build gRPC benchmarks

cjcobb23:
The rippled build was failing on linux with gcc-11. The issue was related to gRPC, specifically when gRPC is not installed system wide but rather is being built from source via `ExternalProject_Add`. The error only occurs while trying to build the gRPC benchmark tests, which rippled does not need. This change skips building the benchmark tests.

Test Plan:
Built rippled on Ubuntu 20.04 with gcc-11. Without this change, the build failed. With this change, the build succeeded.


seelabs (approved):
:+1: Thanks for resolving this, it will make my life a little easier.

thejohnfreeman (approved):

ximinez (approved):
Builds fine on Windows (non-reporting mode)

PR #3864, opened 2021-06-09
Disable false-alarm gcc warning

seelabs:

gcc 11 gives a false warning about overrunning array bounds here. This patch disables the warning.


HowardHinnant (approved):

cjcobb23 (approved):
LGTM

PR #3867, opened 2021-06-12
Redistribute unit test priorities based on current measurements

scottschurr:

Unit test priorities were introduced so parallel unit tests would execute faster.  Unit tests with longer running times have higher priorities than unit tests with shorter running times.

As tests have changed, the amount of time that unit tests take to execute has changed.  Some unit tests that currently have long running times did not have priorities.  Some unit tests that used to have long running times run faster than they used to.

This pull request adds priorities to tests that currently have long running times but did not have priorities assigned.  Some unit tests that had a priority have their priority removed, since they do not run for very long.

Comparative unit test execution times were measured using my Mac mini.  Other platforms may have different results.

### Type of Change

Since this pull request only affects the order in which unit tests execute it does not need to be mentioned in the release notes.

I've requested review from @ximinez and @thejohnfreeman, since they seem to be paying the most attention to continuous integration these days.

### DIFF
--- a/src/beast/extras/beast/unit_test/suite.hpp
+++ b/src/beast/extras/beast/unit_test/suite.hpp
@@ -670,6 +670,11 @@ run(runner& r)
     @endcode
 
     The macro invocation must appear in the same namespace as the test class.
+
+    Unit test priorities were introduced so parallel unit_test::suites would
+    execute faster. Suites with longer running times have higher priorities
+    than unit tests with shorter running times.  Suites with no priorities
+    are assumed to run most quickly, so they run last.
 */
 
 #if BEAST_NO_UNIT_TEST_INLINE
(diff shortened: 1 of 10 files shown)


ximinez (approved):
I get a slightly different order in my local tests, but IMHO the important thing is that the prioritized tests are running before anything else, regardless of the priority.

thejohnfreeman:
The summary of this PR (at least the first paragraph) is a great bit of history and context that would help document the code. Is it written anywhere in the codebase? Is there a good place it can go? The `beast/extras/beast/unit_test` module does not presently have a README, but maybe it can go in `suite_info.hpp`.

thejohnfreeman (approved):

scottschurr:
Excellent suggestion @thejohnfreeman.  I appended a commit that adds the comment you suggested.  Let me know how that works or if you'd like something different.  Thanks.

PR #3879, opened 2021-07-05
make building tests optional

RichardAH:
Small change to build system and preprocessor macros to exclude all unittest related code from builds unless explicitly enabled with `cmake -Dtests=ON`

### Context of Change
Build time is significantly increased by the inclusion of unit tests. However in many cases there is no need to build these tests (e.g. rapid prototyping builds, etc.) Making the building of tests optional will improve developer efficiency.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

NB: Nik asked me to create a PR for this a while ago. So here it is :)


ximinez (changes requested):
I can see this being valuable in some situations. A few changes, though, before I'll consider approving:
* Make "ON" the default
* Fix the indentation to be consistent in the cmake files
* If possible, make CI skip the tests when they're not built. I'm not sure how you'd do that, but it's probably doable.

cjcobb23:
> I can see this being valuable in some situations. A few changes, though, before I'll consider approving:
> 
>     * Make "ON" the default
> 
>     * Fix the indentation to be consistent in the cmake files
> 
>     * If possible, make CI skip the tests when they're not built. I'm not sure how you'd do that, but it's probably doable.

I agree that "ON" should be the default. 

In what situation would CI not build the tests?

ximinez:
> In what situation would CI not build the tests?

Woops. I flubbed that one. I meant to suggest adding a job where CI doesn't build the tests, and correctly skips them. I was thinking of it in terms of catching references to the test code that are added without gating, because I expect that a lot of developers are going to forget that this option exists.

> Edit: Actually, I think we should have at least one CI job that builds rippled without the tests. We can probably do that similar to how we do the coverage build:
> 
> https://github.com/ripple/rippled/blob/7bd5d51e4e4e76a5547051d30b330739618eddb0/.travis.yml#L115

Yep, like that. :grinning:

RichardAH:
Not very familiar with Travis but I added an entry that looks approximately right?
(And corrected the other things as requested)

.travis.yml
```diff
@@ -123,6 +123,27 @@ matrix:
...
+      name: coverage, gcc-8
+      env:
+        - MATRIX_EVAL="CC=gcc-8 && CXX=g++-8"
+        - BUILD_TYPE=Debug
+        - CMAKE_ADD="-Dtests=OFF"
+        - TARGET=coverage_report
+        - SKIP_TESTS=true
+    - <<: *linux
+      if: commit_message !~ /travis_run_/ OR commit_message =~ /travis_run_linux/ OR commit_message =~ /travis_run_cov/
+      compiler: clang-8
```
cjcobb23, line 139: Do we really need a clang and a gcc build for the no tests situation? Travis takes forever to run, so I would rather be conservative about adding more CI jobs. I think the main thing to test here is that the includes are correct, and so I don't think clang and gcc are going to behave differently in that regard (I'm not a build expert by any means, so I could be wrong). 

However, I am wondering if we should have a no unity build for this. @ximinez what do you think?

cjcobb23 (changes requested):

RichardAH:
Updated travis again according to @cjcobb23's comments. I made one of the two build without tests as `clang+no-unity` and the other builds without tests as `gcc+unity`. Between these two I think you would know in 99.99% of cases if the tests=OFF option was causing any sort of issue with the build.

Builds/CMake/RippledSettings.cmake
```diff
@@ -6,6 +6,11 @@ option (assert "Enables asserts, even in release builds" OFF)
 
 option (reporting "Build rippled with reporting mode enabled" OFF)
 
+option (tests "Build tests" ON)
+if (tests)
+  add_compile_definitions(ENABLE_TESTS)
```
cjcobb23, line 11: `add_compile_definitions` is not supported in our minimum supported cmake version. We usually do this with `target_compile_definitions`, like so:
https://github.com/ripple/rippled/blob/7bd5d51e4e4e76a5547051d30b330739618eddb0/Builds/CMake/RippledCore.cmake#L979

cjcobb23 (commented):
We need to run clang-format on the code. There is an example here: https://github.com/ripple/rippled/blob/7bd5d51e4e4e76a5547051d30b330739618eddb0/.github/workflows/clang-format.yml#L24

Run it on the src and test folders. You have to replace `CLANG_VERSION` with your actual clang version.

RichardAH:
I think? all requests are satisfied now. Let me know if there's something else that needs to be done

ximinez (changes requested):
I'm not sure why, but if I run `rippled --unittest` on a build without tests, it gets picked up as the hidden `unittest-child` option on line 453. Because that option is not checked independently, it falls through to run rippled in p2p mode, as if there were no params. This should be an error condition.
```
    hidden.add_options()(
        "parameters",
        po::value<vector<string>>(),
        "Specify rpc command and parameters. This option must be repeated "
        "for each command/param. Positional parameters also serve this "
        "purpose, "
        "so this option is not needed for users")
#ifdef ENABLE_TESTS
        ("unittest-child",
         "For internal use only when spawning child unit test processes.")
#endif  // ENABLE_TESTS
            ("fg", "Deprecated: server always in foreground mode.");
```

ximinez (changes requested):

src/ripple/app/main/Main.cpp
```diff
@@ -433,6 +438,7 @@ run(int argc, char** argv)
         "unittest-jobs",
         po::value<std::size_t>(),
         "Number of unittest jobs to run in parallel (child processes).");
+#endif  // ENABLE_TESTS
 
     // These are hidden options, not intended to be shown in the usage/help
     // message
```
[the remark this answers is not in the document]
RichardAH: hi @ximinez I've guarded those options now on the no-tests build and explicitly exit if they're used

ximinez (approved):
Looks great, and my `--unittest` issue has been nicely resolved.

cjcobb23 (approved):
LGTM

manojsdoshi:
Hi, @RichardAH your PR has merge conflicts can you rebase them against the develop branch so I can include them in the next beta. Thanks

ximinez:
@manojsdoshi I went ahead and did the rebase, resolved the conflicts, squashed the commits, and rewrote the commit message a little bit. https://github.com/ripple/rippled/compare/develop...ximinez:RAH-optional-tests

PR #3883, opened 2021-07-15
PeerFinder enhancements

gregtatcam:

Set default max outbound and total peers to 5/11.
Remove storing livecache endpoints by hops.
Randomize livecache endpoints for broadcast.
Add peer eviction when inbound slots cap out in hubs
with more than 100 configured inbound peers.

### Context of Change

All enhancements relate to how the PeerFinder manages and selects the endpoints.
Changing the default max out/total peers removes excessive peer redundancy.
Eliminating the store and distribution by hops and randomization of endpoints independently
of hops makes the network more resilient to Byzantine behavior.
Lastly, the peer eviction enables fairer usage of hubs resources among peers and results
in overall better connected network.

### Type of Change

## Before / After

- The Livecache is a simple aged set of endpoints as opposed to an array of lists of endpoints,
   where the array is indexed by hops. Selection of endpoints for auto-connect, redirect, and broadcast
   is done by randomly selecting required number of endpoints from the set.
- The eviction logic is configurable and is disabled by default. The logic kicks in when a number of
   inbound points in a hub (a server with >= 100 inbound points) caps out. The logic first selects 
   the candidates for removal (inbound, not fixed, and not reserved) and then eliminates from the list 8 peers 
   with the best latency, 4 peers which recently received validation or proposal messages, 4 peers based 
   on the endpoint network group assignment, and 50% of peers with the longest connection time. Then 5% 
   of the remaining peers are stopped. 

## Test Plan

A PeerFinder unit test is added to test condition that kicks in the eviction logic. Other unit tests are updated
to account for new logic and default configuration values.


intelliot:
(notes: large restructuring change; worth considering; changes will require additional testing, e.g. changing inbound/outbound peers, affecting overlay. could run simulations of the network to see the change on the graph with a change from 10/11 to 5/11. could significantly change the network graph.)

intelliot:
@gregtatcam if you think this may still be worth doing, please resolve conflicts and then contact reviewers

gregtatcam:
> @gregtatcam if you think this may still be worth doing, please resolve conflicts and then contact reviewers

It's a high risk change, which may substantially impact the network connectivity. I don't think it's worth doing.

intelliot:
Makes sense to me. Can you close this pull request?

PR #3888, opened 2021-07-22
Tx tables fix

undertome:
This is just an extension for a previously approved PR - #3815 which is from a deleted repository.

I added some changes in an attempt to improve logic and readbility. Fortunately these changes are included in a separate commit to make things easier.

### DIFF
--- a/src/ripple/app/rdb/RelationalDBInterface_nodes.h
+++ b/src/ripple/app/rdb/RelationalDBInterface_nodes.h
@@ -277,7 +277,11 @@ getTxHistory(
  * @param j Journal.
  * @return Vector of pairs of found transactions and their metadata
  *         sorted in ascending order by account sequence.
- *         Also number of transactions processed.
+ *         Also number of transactions processed or skpiied.
+ *         If this number is >= 0, then it means number of transactions
+ *         processed, if it is < 0, then ~number means number of transactions
+ *         skipped. We need to skip some quantity of transactions if option
+ *         offset is > 0 in the options structure.
  */
 std::pair<RelationalDBInterface::AccountTxs, int>
 getOldestAccountTxs(
@@ -304,7 +308,11 @@ getOldestAccountTxs(
  * @param j Journal.
  * @return Vector of pairs of found transactions and their metadata
  *         sorted in descending order by account sequence.
- *         Also number of transactions processed.
+ *         Also number of transactions processed or skipped.
+ *         If this number is >= 0, then it means number of transactions
+ *         processed, if it is < 0, then ~number means number of transactions
+ *         skipped. We need to skip some quantity of transactions if option
+ *         offset is > 0 in the options structure.
  */
 std::pair<RelationalDBInterface::AccountTxs, int>
 getNewestAccountTxs(
@@ -331,7 +339,11 @@ getNewestAccountTxs(
  * @param j Journal.
  * @return Vector of tuples of found transactions, their metadata and
  *         account sequences sorted in ascending order by account
- *         sequence. Also number of transactions processed.
+ *         sequence. Also number of transactions processed or skpiied.
+ *         If this number is >= 0, then it means number of transactions
+ *         processed, if it is < 0, then ~number means number of transactions
+ *         skipped. We need to skip some quantity of transactions if option
+ *         offset is > 0 in the options structure.
  */
 std::pair<std::vector<RelationalDBInterface::txnMetaLedgerType>, int>
 getOldestAccountTxsB(
@@ -357,7 +369,11 @@ getOldestAccountTxsB(
  * @param j Journal.
  * @return Vector of tuples of found transactions, their metadata and
  *         account sequences sorted in descending order by account
- *         sequence. Also number of transactions processed.
+ *         sequence. Also number of transactions processed or skpiied.
+ *         If this number is >= 0, then it means number of transactions
+ *         processed, if it is < 0, then ~number means number of transactions
+ *         skipped. We need to skip some quantity of transactions if option
+ *         offset is > 0 in the options structure.
  */
 std::pair<std::vector<RelationalDBInterface::txnMetaLedgerType>, int>
 getNewestAccountTxsB(
--- a/src/ripple/app/rdb/RelationalDBInterface_nodes.h
+++ b/src/ripple/app/rdb/RelationalDBInterface_nodes.h
@@ -277,9 +277,9 @@ getTxHistory(
  * @param j Journal.
  * @return Vector of pairs of found transactions and their metadata
  *         sorted in ascending order by account sequence.
- *         Also number of transactions processed or skpiied.
+ *         Also number of transactions processed or skipped.
  *         If this number is >= 0, then it means number of transactions
- *         processed, if it is < 0, then ~number means number of transactions
+ *         processed, if it is < 0, then -number means number of transactions
  *         skipped. We need to skip some quantity of transactions if option
  *         offset is > 0 in the options structure.
  */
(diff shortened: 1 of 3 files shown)


src/ripple/app/rdb/RelationalDBInterface_nodes.h
```diff
@@ -277,7 +277,11 @@ getTxHistory(
  * @param j Journal.
  * @return Vector of pairs of found transactions and their metadata
  *         sorted in ascending order by account sequence.
- *         Also number of transactions processed.
+ *         Also number of transactions processed or skpiied.
```
cjcobb23, line 280: typo on `skpiied`

src/ripple/app/rdb/backend/RelationalDBInterfaceSqlite.cpp
```diff
@@ -1175,32 +1351,38 @@ RelationalDBInterfaceSqliteImp::oldestAccountTxPage(
...
-                limit_used,
-                page_length);
-            opt.marker = marker;
-            if (total < 0)
-                return false;
-            limit_used += total;
-            return true;
-        });
+    /* else use shard databases, if available */
+    if (shardStoreExists() && useTxTables_)
```
cjcobb23, line 1355: I think all of these functions can just return early if `useTxTables_` is false. There's no need to do anything else. So at the top of the function:

```
if(!useTxTables_)
    return {};
```

cjcobb23 (commented):
There are a lot of places where we have to check the value of `useTxTables_`. It's seems like this might be a little error prone. What do you think of creating two separate classes, one for when `useTxTables_` is true, and one for when it's false? The class used in the false case would just return an empty optional for all of the functions that query the tx tables.

src/ripple/app/rdb/RelationalDBInterface_nodes.h
```diff
@@ -277,7 +277,11 @@ getTxHistory(
...
+ *         If this number is >= 0, then it means number of transactions
+ *         processed, if it is < 0, then ~number means number of transactions
+ *         skipped. We need to skip some quantity of transactions if option
```
seelabs, line 283: Returning `~number` for the number of transactions skipped is a very strange interface. I'd suggest returning `-number` instead - unless there's some strange reason why `~number` is better, in which case we should document that reason (here and other functions as well).

src/ripple/app/rdb/backend/RelationalDBInterfaceSqlite.cpp
```diff
@@ -457,14 +470,20 @@ RelationalDBInterfaceSqliteImp::getMinLedgerSeq()
...
 
-    /* else use shard databases */
-    std::optional<LedgerIndex> res;
-    iterateLedgerForward(
-        {}, [&](soci::session& session, std::uint32_t shardIndex) {
-            res = ripple::getMinLedgerSeq(session, TableType::Ledgers);
-            return !res;
-        });
-    return res;
+    /* else use shard databases, if available */
```
seelabs, line 473: The code is better off without comments that repeat what the next line of code does (here and in many other places in this file). You don't need to change that now (it's a minor nit), but new code should not include such comments.

undertome: Noted. I'll consider removing these as I dig deeper into refactoring.

src/ripple/app/rdb/RelationalDBInterface_nodes.h
undertome: Done

src/ripple/app/rdb/RelationalDBInterface_nodes.h
undertome: Done

src/ripple/app/rdb/backend/RelationalDBInterfaceSqlite.cpp
```diff
@@ -257,23 +260,33 @@ class RelationalDBInterfaceSqliteImp : public RelationalDBInterfaceSqlite
...
     /**
-     * @brief checkoutLedger Checks if node ledger DB exists.
+     * @brief existsLedger Checks if node ledger DB exists.
      * @return True if node ledger DB exists.
      */
     bool
     existsLedger()
     {
-        return !!lgrdb_;
+        return (bool)lgrdb_;
```
mtrippled, line 269: the bool() operator automatically evaluates whether a unique_ptr contains something, so casting is redundant.

Also, the way to cast this if you really needed to would be by using something like static_cast<bool>(anotherpodtype);

src/ripple/app/rdb/backend/RelationalDBInterfaceSqlite.cpp
```diff
@@ -257,23 +260,33 @@ class RelationalDBInterfaceSqliteImp : public RelationalDBInterfaceSqlite
...
     /**
-     * @brief checkoutTransaction Checks if node transaction DB exists.
+     * @brief existsTransaction Checks if node transaction DB exists.
      * @return True if node transaction DB exists.
      */
     bool
     existsTransaction()
     {
-        return !!txdb_;
+        return (bool)txdb_;
```
mtrippled, line 279: See casting comment above.

mtrippled (changes requested):
I tested that this works with use_tx_tables = 0, so it fixes the original problem.

Looks good to me besides the casting I mentioned.

seelabs (approved):
:+1: Thanks for changing the interface to use `-number` instead of `~number`.

src/ripple/app/rdb/backend/RelationalDBInterfaceSqlite.cpp
undertome: The compiler actually doesn't actually invoke the `bool` operator without the cast. Changed to `static_cast` instead.

src/ripple/app/rdb/backend/RelationalDBInterfaceSqlite.cpp
undertome: Changed to `static_cast`

src/ripple/app/rdb/backend/RelationalDBInterfaceSqlite.cpp
undertome: Done

undertome:
> There are a lot of places where we have to check the value of `useTxTables_`. It's seems like this might be a little error prone. What do you think of creating two separate classes, one for when `useTxTables_` is true, and one for when it's false? The class used in the false case would just return an empty optional for all of the functions that query the tx tables.

I like that approach in certain scenarios. For instance I used it with the `ShardArchiveHandler` where a derived type is instantiated when the handler is starting in recovery mode. But I think its a bit heavy handed to create a separate class with separate functions just for this single `bool`.

src/ripple/app/rdb/backend/RelationalDBInterfaceSqlite.cpp
undertome: Ended up removing them when addressing CJ's comment about returning early when transaction tables are disabled. Done

src/ripple/app/rdb/backend/RelationalDBInterfaceSqlite.cpp
```diff
@@ -266,7 +266,7 @@ class RelationalDBInterfaceSqliteImp : public RelationalDBInterfaceSqlite
     bool
     existsLedger()
     {
-        return (bool)lgrdb_;
+        return static_cast<bool>(lgrdb_);
```
mtrippled, line 269: https://www.cplusplus.com/reference/memory/unique_ptr/operator%20bool/
The bool() operator in that object is called that does exactly what you want.

mtrippled (approved):
LGTM :+1:

undertome:
> LGTM +1

Ran the test again - it checks out.

cjcobb23 (approved):
LGTM

PR #3892, opened 2021-08-02
Minor cleanups in offer processing code

nbougalis:
Remove the comparison code between the newer `Flow` and older `Taker` code, as well as the feature that can be used to enable or disable the comparison at runtime.

Also, adjust some log levels to reduce the possibility of confusion when examining logs.

### DIFF
--- a/src/ripple/app/ledger/OpenLedger.h
+++ b/src/ripple/app/ledger/OpenLedger.h
@@ -222,8 +222,7 @@ OpenLedger::apply(
     {
         try
         {
-            // Dereferencing the iterator can
-            // throw since it may be transformed.
+            // Dereferencing the iterator can throw since it may be transformed.
             auto const tx = *iter;
             auto const txId = tx->getTransactionID();
             if (check.txExists(txId))
@@ -233,9 +232,10 @@ OpenLedger::apply(
             if (result == Result::retry)
                 retries.insert(tx);
         }
-        catch (std::exception const&)
+        catch (std::exception const& e)
         {
-            JLOG(j.error()) << "Caught exception";
+            JLOG(j.error())
+                << "OpenLedger::apply: Caught exception: " << e.what();
         }
     }
     bool retry = true;
--- a/src/ripple/app/ledger/impl/LedgerMaster.cpp
+++ b/src/ripple/app/ledger/impl/LedgerMaster.cpp
@@ -638,11 +638,11 @@ LedgerMaster::getValidatedRange(std::uint32_t& minVal, std::uint32_t& maxVal)
             }
             return true;
         }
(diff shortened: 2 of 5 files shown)


cjcobb23:
Why can the comparison be removed?

scottschurr:
@cjcobb23, good question.  The comparison was introduced during the development of `FlowCross`.  It helped to identify what was happening in those few circumstances when the results from `Taker` (the old offer crossing code) differed from the results of `FlowCross` (the new offer crossing code).

The `FlowCross` amendment was enabled August 4th, 2020.  That means we've been doing offer crossing with `FlowCross` on the main network for about a year.  So it is becoming less important to be able to characterize differences between `FlowCross` and `Taker` behavior.

This particular cleanup had been on my radar.  I had intended to do this cleanup at the same time we unconditionalize `FlowCross` and remove the `Taker` code, after `FlowCross` has been enabled on the network for 2 years.  But I, personally, have no real concerns about removing the comparison code now.

scottschurr (commented):
All of the changes here look reasonable.  I am wondering, however, whether it makes sense to wait on this cleanup until we unconditionalize the `FlowCross` amendment.  That would be a delay of at least a year, since we've been waiting until amendments have been live for two years before unconditionalizing them.  `FlowCross` was enabled on August 4, 2020: https://xrpl.org/known-amendments.html

If we decide to proceed with this cleanup, then there's more that can be done.  The `CashDiff` class is only used to help determine the differences which are no longer being computed.  So the `CashDiff` class, and anything associated with it, can be removed.  Here's a commit that would remove `CashDiff`: https://github.com/scottschurr/rippled/commit/2b7b1b77970d11c98fc7406ba0133134c1518c62

I can go with either waiting or proceeding.  Your call.

cjcobb23 (approved):
Thanks for the explanation Scott. 

Code looks good to me. I like the log level adjustment.

nbougalis:
Generally I'd fine with delaying the removal until `FlowCross` is unconditionalized.

However, I think the code is already cluttered enough and since this isn't a transaction-processing change but rather a "check" I feel pretty good about removing it.

If anyone objects, I'm fine with putting this on the backburner.

And yeah, I think I'll pick up the `CashDiff` removal patch as well.

scottschurr:
@nbougalis, still waiting on incorporating the `cashdiff` removal patch...

nbougalis:
Your commit 2b7b1b77970d11c98fc7406ba0133134c1518c62 was cherry-picked and squashed into mine, as is included in the version of commit that's integrated into the proposed **1.8.0-b5**.

seelabs (approved):
:+1:

PR #3894, opened 2021-08-05
check protobuf field size before calling uint256::fromVoid

cjcobb23:

There were numerous places in the code where we were calling `uint256::fromVoid()` to convert a protobuf byte array to a `uint256` without first checking the size of the byte array. If the byte array is less than 32 bytes, `fromVoid()` will read past the end of the array, which is undefined behavior. This PR changes the code to check the size of the byte array before calling `fromVoid()`, and returns an appropriate error if the byte array is malformed.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### DIFF
--- a/src/ripple/rpc/handlers/LedgerData.cpp
+++ b/src/ripple/rpc/handlers/LedgerData.cpp
@@ -153,35 +153,33 @@ doLedgerDataGrpc(
         return {response, errorStatus};
     }
 
-    ReadView::key_type key = ReadView::key_type();
-    if (request.marker().size() != 0)
+    uint256 startKey;
+    if (auto key = uint256::fromVoidChecked(request.marker()))
     {
-        if (request.marker().size() != uint256::size())
-        {
-            grpc::Status errorStatus{
-                grpc::StatusCode::INVALID_ARGUMENT, "marker malformed"};
-            return {response, errorStatus};
-        }
-        key = uint256::fromVoid(request.marker().data());
+        startKey = *key;
+    }
+    else if (request.marker().size() != 0)
+    {
+        grpc::Status errorStatus{
+            grpc::StatusCode::INVALID_ARGUMENT, "marker malformed"};
+        return {response, errorStatus};
     }
 
     auto e = ledger->sles.end();
-    ReadView::key_type stopKey = ReadView::key_type();
-    if (request.end_marker().size() != 0)
+    if (auto key = uint256::fromVoidChecked(request.end_marker()))
     {
-        if (request.end_marker().size() != uint256::size())
-        {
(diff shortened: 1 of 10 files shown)


thejohnfreeman (commented):
Could there be a benefit to having one function that validates and unpacks `GetLedgerDataResponse`s? Or that reads a uint256 from a protobuf?

src/ripple/rpc/handlers/LedgerData.cpp
```diff
@@ -137,36 +137,45 @@ doLedgerDataGrpc(
...
+        }
+        else
+        {
+            errorStatus =
+                grpc::Status(grpc::StatusCode::NOT_FOUND, status.message());
+        }
         return {response, errorStatus};
     }
 
     ReadView::key_type key = ReadView::key_type();
```
natenichols, line 156: Question for my own understanding: Why are you using `ReadView::key_type` here, but using `uint256::fromVoid` later in the function?

They [are the same thing](https://github.com/ripple/rippled/blob/develop/src/ripple/ledger/ReadView.h/#L198). Why not just declare `key` and `stopKey` as `ripple::uint256`'s? Can you call `ReadView::key_type::fromVoid`?

cjcobb23: No good reason. Probably something that started as a copy paste from somewhere else. I'll switch this to `ripple::uint256`, since I think that is clearer.

cjcobb23:
> Could there be a benefit to having one function that validates and unpacks `GetLedgerDataResponse`s? Or that reads a uint256 from a protobuf?

There is definitely a benefit to have a helper function that reads a uint256 from protobuf. I'll code that up. 

As to the first suggestion, `GetLedgerDataResponse` is only processed in `ETLSource.cpp`. `ReportingETL.cpp` is processing a `GetLedgerResponse`. The first is downloading a ledger in full, while the second is processing a diff, and some data may represent a deleted object. There's some common code though, so I'll play around with writing a helper function to reuse some bits of logic.

thejohnfreeman (approved):
This isn't how I though the protobuf parse would play out. Ended up adding twice as many lines as it removed. I guess I prefer the earlier version, but don't feel strongly enough to ask for one or the other. I will leave it up to you.

natenichols (approved):

cjcobb23:
> This isn't how I though the protobuf parse would play out. Ended up adding twice as many lines as it removed. I guess I prefer the earlier version, but don't feel strongly enough to ask for one or the other. I will leave it up to you.

Yeah it didn't really reduce the number of lines, though I think it made the code more robust. It's harder to forget to check the size now.

PR #3901, opened 2021-08-10
Don't restart rippled during apt upgrade

legleux:
Keep outgoing version of rippled running during upgrade. Rippled must explicitly be restarted.

### Context of Change
The debian packages generated now automatically restart rippled when upgrading. Installing rippled after this change will require a manual restart but subsequent upgrades will keep the old version running, i.e. it will take *2* upgrades to exhibit desired behavior.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Test Plan
[What it looks like in practice](https://gist.github.com/legleux/8e9439fcc32fd92377eccd36fea7de3f)

## Future Tasks
Eventually all this package generation should be handled by native CMake features rather than CMake calling scripts that call CMake that call deb-helper programs.

### DIFF
--- a/Builds/containers/packaging/dpkg/debian/rules
+++ b/Builds/containers/packaging/dpkg/debian/rules
@@ -10,6 +10,7 @@ export CXXFLAGS:=$(subst -Werror=format-security,,$(CXXFLAGS))
 
 %:
 	dh $@ --with systemd
+override_dh_systemd_start:
 
 override_dh_auto_configure:
 	env
@@ -38,5 +39,3 @@ override_dh_auto_install:
 	rm -rf debian/tmp/opt/ripple/lib64/cmake/date
 	rm -rf bld
 	rm -rf bld_vl
-
-


thejohnfreeman (approved):

thejohnfreeman:
Great context, thank you for that. Is it true that this will also prevent the service from starting automatically when the machine boots? @manojsdoshi, do you think that will be a problem for anyone?

legleux:
> Great context, thank you for that. Is it true that this will also prevent the service from starting automatically when the machine boots? @manojsdoshi, do you think that will be a problem for anyone?

The rippled.service unit file packaged still dictates that it starts on boot. This is just some in-between uninstall/install magic.

ximinez (changes requested):
A brief summary of our offline conversation:
1. A lot of operators are configured to do automatic rippled upgrades. Until they manually intervene to restart rippled, they could have a pretty long downtime, which could be bad for them, and for the network. I would very much prefer if you could find a solution that checks after install for a running rippled, and starts it if not running (which I think we can assume indicates the old version was stopped by the old .deb config).
2. Please rewrite the commit message to be more informative and link back to the issue. e.g.
```
Don't restart rippled during apt upgrade:

Resolves #3294
```

legleux:
I could have sworn the obvious `dh_systemd_start --no-restart-on-upgrade` was one of the first things I tried and got an error like `--no-restart-on-upgrade parameter unrecognized` which led me to think it wasn't available in the set of deb_helper programs on the version of ubuntu we generate pkgs  with.
Apparently I was wrong and this seems to work and is much cleaner than my implementation of this behavior.

manojsdoshi (approved):
LGTM

ximinez (approved):
Looks good

PR #3911, opened 2021-08-25
[TRIVIAL] Remove the `ltINVALID` ledger entry type

nbougalis:
The `ltINVALID` ledger type doesn't really have a raison d'être anymore. In the few places where it's used, `ltANY` seems like a better match. So remove it and simplify the codebase by a smidge.


HowardHinnant (approved):

nbougalis:
This PR has been incorporated into #3920

PR #3914, opened 2021-08-26
Health check endpoint ignores fee escalation:

ximinez:

Changed the health check endpoint (`https://[host]:[peerport]/health`) to only consider and consume the portion of the `load_factor` that is limited to the server. This ignores any changes related to fee escalation. Fee escalation describes how busy the ledger is, not the server.

Also refactored that function to use `jss` labels instead of strings for JSON fields.

### Context of Change

I was looking at the health of my server and was annoyed when it went into warning status solely with a load factor in the 500's because fee escalation kicked in.

It is my opinion that this is not correct or desired behavior. I may be wrong about that, so that discussion would be appropriate here.

### Type of Change

## Before / After

Before: If the server's `load_factor` goes over 100 because of fee escalation, the `/health` endpoint will return a 503 result code and include the `load_factor` in the result JSON.

After: If the server's `load_factor` goes over 100 because of fee escalation, but the server's internal load is under 100, the `/health` endpoint will return a 200, and not include `load_factor` in the JSON. If the internal load goes over 100 because the server itself is busy, it will continue to be flagged by `/health` as before.

### DIFF
--- a/src/ripple/overlay/impl/OverlayImpl.cpp
+++ b/src/ripple/overlay/impl/OverlayImpl.cpp
@@ -1045,15 +1045,16 @@ OverlayImpl::processHealth(http_request_type const& req, Handoff& handoff)
     auto info = getServerInfo();
 
     int last_validated_ledger_age = -1;
-    if (info.isMember("validated_ledger"))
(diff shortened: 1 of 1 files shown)


manojsdoshi (approved):
LGTM

PR #3915, opened 2021-08-26
Fix cmake soci patch (duplicate)

ximinez:

This is a copy of https://github.com/ripple/rippled/pull/3886 with some changes suggested in review, but that @guidovranken has not yet addressed. When merging *do not* squash their commit so that their work can be acknowledged.

*Original from #3886*
This prevents `soci` getting patched more than once over the course of multiple builds within the same CMake build environment.

### Context of Change

*Original from #3886*
The various `unity_*_cxx.cxx.o` kept getting built anew after they've been built already.
The reason for this is that each time the `make` command is issued, `soci` is patched anew. Because `soci` is a dependency of the `unity_*_cxx.cxx.o` files, the build system then reasons that the `unity_*_cxx.cxx.o` files must be built anew because `soci` has a more recent timestamp.

Patching `soci` only needs to be done once; from that point on it can be reused any number of times.

This PR makes a change to the script responsible for patching, and only proceeds to patch if it finds that there is no back-up of original (unpatched) `soci`. If a back-up already exists, this indicates that `soci` has already been patched, and the patch procedure is skipped.

*Update:*
This PR makes a change to the script responsible for patching. It patches the file, then compares the patched file to the original. If they are identical, it deletes the patched file and exits. If not, it backs up the original, then replaces it with the patch.

### Type of Change

## Before / After

Before: Repeated builds in the same CMake environment with no changes will unnecessarily build many of the files. In unity builds, they will be `unity_*_.cxx`. In non-unity builds, they will be individual source files.

After: Repeated builds in the same CMake environment does not build any source, allowing the build to take far less time. On my workstation, time for various build configs was reduced from a range of 11s - 42s to a range of 2s - 7s. Unfortunately, CMake still has some overhead dealing with external projects and such, which accounts for some of the remaining time.

## Future Tasks

Find any other unnecessary file replacements and fix those.

### DIFF
--- a/Builds/CMake/deps/Rocksdb.cmake
+++ b/Builds/CMake/deps/Rocksdb.cmake
@@ -63,7 +63,7 @@ if (local_rocksdb)
     GIT_TAG v6.7.3
     PATCH_COMMAND
       # only used by windows build
-      ${CMAKE_COMMAND} -E copy
+      ${CMAKE_COMMAND} -E copy_if_different
       ${CMAKE_CURRENT_SOURCE_DIR}/Builds/CMake/rocks_thirdparty.inc
       <SOURCE_DIR>/thirdparty.inc
     COMMAND
--- a/Builds/CMake/soci_patch.cmake
+++ b/Builds/CMake/soci_patch.cmake
@@ -1,13 +1,30 @@
 # This patches unsigned-types.h in the soci official sources
 # so as to remove type range check exceptions that cause
 # us trouble when using boost::optional to select int values
+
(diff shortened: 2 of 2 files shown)


ximinez:
I took a quick look at the other patch commands and noticed that Rocksdb was unnecessarily copying a file repeatedly. This fix was nearly trivial: Change `copy` to `copy_if_different`. Since it's so simple, I'm pushing it here. LMK if there are any objections.

ximinez:
Oh, and I also cleaned up the commit messages a little bit.

ximinez:
Squashed and marked passed

PR #3916, opened 2021-08-26
Make a stand-in for `std::expected` available in the code base

scottschurr:
## Make a Stand-in For `std::expected` Available in the Code Base

`std::expected` is a proposed type for the standard library.  The current paper for `std::expected` is: http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2021/p0323r10.html

The interface of `std::expected` looks great to me.  It's like a `std::optional` that tells you _why_ something came back empty.  I'd like to be able to use it in the code base without waiting for us to adopt C++23.

### Context of Change

Boost does not directly provide an interface like `std::expected`.  However Boost has something similar called `boost::outcome_v2::result`.  I've coded a fairly thin wrapper around `boost::outcome_v2::result` that provides the characteristics of `std::expected` that I particularly like.  I've named that wrapper `ripple::Expected`.

That wrapper, and unit tests for it, are in the first commit.

My initial impression was that `Expected` would be immediately usable in significant portions of the code base.  So I integrated `Expected` into our code base.  To my surprise I found relatively few places where I could directly drop in `Expected` as a replacement for returning a `std::pair` where one of the pair members was the reason for the failure.  But I did find a few.  You can find that integration in the second commit.

Reviewers should consider two things:
- Given that `Expected` is not essential to the code base, should we still incorporate it?  I think we should.  But you may not agree
- If it is worth incorporating `Expected` into the code base, do you see any coding errors or places for improvement?

For what it's worth, the `std::expected` paper is on track to go into C++23 and has passed the Library Evolution working group.  But there's some risk because it has not yet survived the Library (wording) working group.

### Type of Change

This is an internals-only change.  It does not need to be mentioned in the release notes.

## Before / After

From an interface perspective, an implementation returning an `Expected` changes something sorta like this ...
```
std::pair<std::optional<AcctRootRd>, NotTEC >
makeAcctRoot(std::shared_ptr<STLedgerEntry const> slePtr)
{
    if (NotTEC const ter = validateAcctRootSle(slePtr); !isTesSuccess(ter))
        return {nullopt, ter}

    return {AcctRootRd(std::move(slePtr), tesSUCCESS};
}
```
... to this ...
```
Expected<AcctRootRd, NotTEC>
makeAcctRoot(std::shared_ptr<STLedgerEntry const> slePtr)
{
    if (NotTEC const ter = validateAcctRootSle(slePtr); !isTesSuccess(ter))
        return Unexpected(ter);

    return AcctRootRd(std::move(slePtr));
}
```
At the call site this ...
```
    auto const acctRoot = makeAcctRoot(ctx.view.read(keylet::account(id)));
    if (!isTesSuccess(acctRoot.second))
        return acctRoot.second;

    std::uint32_t const uFlagsIn = {acctRoot.first->flags()};
```
... becomes this ...
```
    auto const acctRoot = makeAcctRoot(ctx.view.read(keylet::account(id)));
    if (!acctRoot)
        return acctRoot.error();

    std::uint32_t const uFlagsIn = {acctRoot->flags()};
```
The examples have not been run through a compiler, so there may be bugs.

## Test Plan

Unit tests for `Expected` are included.  The integration also builds confidence that the `Expected` wrapper is behaving as expected.

## Future Tasks

Presuming `std::expected` makes it into C++23 it will make sense to remove `ripple::Expected` from the code base and replace its usage with `std::expected`.  There _shouldn't_ be significant usage changes to `std::expected`, since it has passed through Library Evolution.  So the replacement should be straight forward at that time.

### DIFF
--- a/src/ripple/basics/Expected.h
+++ b/src/ripple/basics/Expected.h
@@ -1,0 +1,243 @@
+//------------------------------------------------------------------------------
+/*
+    This file is part of rippled: https://github.com/ripple/rippled
+    Copyright (c) 2021 Ripple Labs Inc.
+
+    Permission to use, copy, modify, and/or distribute this software for any
+    purpose  with  or without fee is hereby granted, provided that the above
+    copyright notice and this permission notice appear in all copies.
+
+    THE  SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+    WITH  REGARD  TO  THIS  SOFTWARE  INCLUDING  ALL  IMPLIED  WARRANTIES  OF
+    MERCHANTABILITY  AND  FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+    ANY  SPECIAL ,  DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+    WHATSOEVER  RESULTING  FROM  LOSS  OF USE, DATA OR PROFITS, WHETHER IN AN
+    ACTION  OF  CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+    OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+//==============================================================================
+
+#ifndef RIPPLE_BASICS_EXPECTED_H_INCLUDED
+#define RIPPLE_BASICS_EXPECTED_H_INCLUDED
+
+#include <ripple/basics/contract.h>
+#include <boost/outcome.hpp>
+#include <stdexcept>
+#include <type_traits>
+
+namespace ripple {
+
+/** Expected is an approximation of std::expected (hoped for in C++23)
+
+    See: http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2021/p0323r10.html
+
+    The implementation is entirely based on boost::outcome_v2::result.
+*/
+
+// Exception thrown by an invalid access to Expected.
+struct bad_expected_access : public std::runtime_error
+{
+    bad_expected_access() : runtime_error("bad expected access")
+    {
+    }
+};
+
+namespace detail {
+
+// Custom policy for Expected.  Always throw on an invalid access.
+struct throw_policy : public boost::outcome_v2::policy::base
+{
+    template <class Impl>
+    static constexpr void
+    wide_value_check(Impl&& self)
+    {
+        if (!base::_has_value(std::forward<Impl>(self)))
+            Throw<bad_expected_access>();
+    }
+
+    template <class Impl>
+    static constexpr void
+    wide_error_check(Impl&& self)
+    {
+        if (!base::_has_error(std::forward<Impl>(self)))
+            Throw<bad_expected_access>();
+    }
+
+    template <class Impl>
+    static constexpr void
+    wide_exception_check(Impl&& self)
+    {
+        if (!base::_has_exception(std::forward<Impl>(self)))
+            Throw<bad_expected_access>();
+    }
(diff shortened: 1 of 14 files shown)


scottschurr:
I pushed a commit with a partial specialization for `Expected<void, E>`.  To me it looks like it should serve the purpose.  Let me know what you think.  Thanks.

src/ripple/basics/Expected.h
```diff
@@ -0,0 +1,244 @@
...
+class [[nodiscard]] Expected<void, E>
+    : private boost::outcome_v2::result<void, E, detail::throw_policy>
+{
+    using Base = boost::outcome_v2::result<void, E, detail::throw_policy>;
+
+public:
+    // We'd like to avoid a default constructor.  That means someone needs
+    // explicitly say they intent to return success.  Seems like
+    // std::true_type is a good way to do that...
+    constexpr Expected(std::true_type) : Base(boost::outcome_v2::success())
```
seelabs, line 215: It looks like the proposed standard `expected` has a default constructor for a `void` return (ref: http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2019/p0323r9.html#expected.object.ctor). I think a default constructor makes sense in this case, and give the standard has it, I'd like to remove this constructor and replace it with a default constructor.

seelabs (commented):
Thanks for coding up the `void` specialization. I had one comment about removing the `true_type` constructor and replacing it with a default constructor. Other than that it looks great.

src/ripple/basics/Expected.h
scottschurr: Thanks for catching the default constructor for `Expected<void, E>`.  I looked for something like `boost::outcome_v2::success()` in the paper and didn't find it, so I punted.  Matching the currently proposed version of the paper seems like the best approach.  I've included an additional `[FOLD]` to make this change.

src/ripple/basics/base_uint.h
```diff
@@ -236,27 +233,26 @@ class base_uint
...
         auto const result = parseFromStringView(sv);
-        if (result.first == ParseResult::badLength)
-            Throw<std::invalid_argument>("invalid length for hex string");
+        if (!result)
+        {
+            if (result.error() == ParseResult::badLength)
+                Throw<std::invalid_argument>("invalid length for hex string");
 
-        if (result.first == ParseResult::badChar)
+            assert(result.error() == ParseResult::badChar);
```
seelabs, line 252: Why does `badChar` assert but `badLength` doesn't assert? Consider losing the `assert`. It's changing the behavior of debug builds, and the change doesn't involve the major focus of this PR - which is `expected`. I think it's better if PRs focus on one thing only.

Having said that, I only weakly object. If you feel strongly, then it's your call.

seelabs (approved):
:+1: Great job on this!

I'm approving this now, but please see my comment about adding the `assert`. I'm approving now because however you want to resolve that is fine with me (including keeping it as-is).

src/ripple/basics/base_uint.h
scottschurr: The assert was intended to be a way to guarantee that all of the non-`okay` `ParseResult` values are covered.  We can see that by inspection now.  But in case another error code was added in the future the `assert` would let us know that the `Throw` had the wrong text.  But I don't feel strongly about it.  I'm fine with removing the `assert`.

seelabs (approved):
Still :+1: from me

thejohnfreeman (approved):

scottschurr:
I squashed and pushed and expected CI to run.  Apparently Travis needs a payment before CI will run: https://app.travis-ci.com/github/ripple/rippled/requests

I'll wait until CI has passed before marking the pull request as passed.

scottschurr:
Re-pushed to kick Travis into gear.

Issue #3881, opened 2021-07-08
libpq.so.5 missing in reporting mode (Version: 1.7.2)


N3TC4T:
## Issue Description
When rippled compiled with `-Dreporting=ON` and PostgreSQL is not installed on the server you will get below error message when running rippled:

`error while loading shared libraries: libpq.so.5: cannot open shared object file: No such file or directory`

the solution would be to install required dependency by:

`sudo apt-get install libpq-dev`

I guess we can add this somewhere in the documentation. 

## Steps to Reproduce
Compile rippled with `-Dreporting=ON` in `Ubuntu 20` without PostgreSQL installed and run executable `rippled`

```
cmake -DCMAKE_BUILD_TYPE=Release -Dreporting=ON -DCMAKE_INSTALL_PREFIX=/opt/ripple ..
cmake --build . --target install -- -j8
```

## Expected Result
Should run rippled without problem

## Environment
Ubuntu 20.04.2 LTS with Boost 1.70

cjcobb23:
We actually try to build libpq ourselves if it's not installed: https://github.com/ripple/rippled/blob/7bd5d51e4e4e76a5547051d30b330739618eddb0/Builds/CMake/deps/Postgres.cmake#L25

Installing it of course fixes the problem. This passes in CI without installing libpq-dev, though that's running on Ubuntu 18.04. What compiler are you using? Also, can you check if there are any other errors in the build?

Also, can you go into the `.nih` folder (it's hidden at the root level, once you actually try to build), and execute `find . -name "libpq.*"`, and post the output here?

N3TC4T:
There is no error, I can see that the cmake cannot find any postgres library in the system.

```
Could NOT find PostgreSQL (missing: PostgreSQL_LIBRARY PostgreSQL_INCLUDE_DIR PostgreSQL_TYPE_INCLUDE_DIR) 
find_package did not find postgres
No system installed Postgres found. Will build
```

also looks like in the build process the lib is already compiled

```
./unix_makefiles/GNU_9.3.0/Release/src/postgres_src/doc/src/sgml/libpq.sgml
./unix_makefiles/GNU_9.3.0/Release/src/postgres_src/src/include/libpq/libpq.h
./unix_makefiles/GNU_9.3.0/Release/src/postgres_src/src/interfaces/libpq/libpq.so
./unix_makefiles/GNU_9.3.0/Release/src/postgres_src/src/interfaces/libpq/libpq.so.5
[...]

PR #3918, opened 2021-08-26
Link Postgres statically 

legleux:
Postgres was linked dynamically now it's static.
Had to tack on a few more libraries.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### DIFF
--- a/Builds/CMake/deps/Postgres.cmake
+++ b/Builds/CMake/deps/Postgres.cmake
@@ -1,21 +1,15 @@
-
 if(reporting)
-
     find_package(PostgreSQL)
-
     if(NOT PostgreSQL_FOUND)
-            message("find_package did not find postgres")
+        message("find_package did not find postgres")
+        find_library(postgres NAMES pq libpq libpq-dev pq-dev postgresql-devel)
+        find_path(libpq-fe NAMES libpq-fe.h PATH_SUFFIXES postgresql pgsql include)
 
-
-            find_library(postgres NAMES pq libpq libpq-dev pq-dev postgresql-devel)
-            find_path(libpq-fe NAMES libpq-fe.h PATH_SUFFIXES postgresql pgsql include)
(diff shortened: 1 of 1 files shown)


cjcobb23 (approved):

thejohnfreeman (approved):

thejohnfreeman:
Curious: is it possible to just install the Postgres shared libraries when they are built as an External Project?

cjcobb23:
> Curious: is it possible to just install the Postgres shared libraries when they are built as an External Project?

Install them system wide? You can just install all of this via the package manager actually.

thejohnfreeman:
Hmmm, #3881 wasn't entirely clear. Let me spell out my understanding of the issue:

On a system without Postgres installed, if you build rippled (with CMake), it will download and build Postgres as an External Project. Then if you install rippled (with CMake), it will installed rippled and supporting libraries, but not Postgres, such that if you run rippled from the installed location (`/opt/ripple/rippled` in the example in #3881), then it will complain that Postgres shared libraries cannot be found.

Is that a fair assessment?

cjcobb23:
> Hmmm, #3881 wasn't entirely clear. Let me spell out my understanding of the issue:
> 
> On a system without Postgres installed, if you build rippled (with CMake), it will download and build Postgres as an External Project. Then if you install rippled (with CMake), it will installed rippled and supporting libraries, but not Postgres, such that if you run rippled from the installed location (`/opt/ripple/rippled` in the example in #3881), then it will complain that Postgres shared libraries cannot be found.
> 
> Is that a fair assessment?

Pretty much. Though I don't know for sure that all supporting libraries are also installed. We link supporting libraries statically by default, in which case they don't need to be installed. In the linked issue, the libraries were being linked statically. Postgres was being built as a shared library always though.

manojsdoshi (approved):
LGTM

PR #3920, opened 2021-08-30
Improve documentation for the LedgerEntryType enum

nbougalis:
The existing documentation was all over the place. This PR improves the documentation (although further improvements are needed) and highlights some important caveats associated with  making changes to ledger entry type identifiers.
    
The commit also adds a safety check to the `KnownFormats<>` class, that will catch the the accidental reuse of format identifiers. Ideally, this should be done at compile time but C++ does not allow for the sort of introspection that would enable this as of C++17.

This PR supersedes PR #3911; the commit from that PR is included here.

### Type of Change
- [X] Documentation Updates

### DIFF
--- a/src/ripple/protocol/KnownFormats.h
+++ b/src/ripple/protocol/KnownFormats.h
@@ -24,6 +24,7 @@
 #include <ripple/protocol/SOTemplate.h>
 #include <boost/container/flat_map.hpp>
 #include <algorithm>
+#include <beast/type_name.h>
 #include <forward_list>
 
 namespace ripple {
@@ -35,7 +36,7 @@ namespace ripple {
 
     @tparam KeyType The type of key identifying the format.
 */
-template <class KeyType>
+template <class KeyType, class Derived>
 class KnownFormats
 {
 public:
@@ -90,7 +91,9 @@ public:
 
         Derived classes will load the object with all the known formats.
     */
-    KnownFormats() = default;
+    KnownFormats() : name_(beast::type_name<Derived>())
+    {
+    }
 
     /** Destroy the known formats object.
 
@@ -111,12 +114,11 @@ public:
     KeyType
     findTypeByName(std::string const& name) const
     {
-        Item const* const result = findByName(name);
-
-        if (result != nullptr)
+        if (auto const result = findByName(name))
             return result->getType();
-        Throw<std::runtime_error>("Unknown format name");
-        return {};  // Silence compiler warning.
+        Throw<std::runtime_error>(
+            name_ + ": Unknown format name '" +
+            name.substr(0, std::min(name.size(), std::size_t(32))) + "'");
     }
 
     /** Retrieve a format based on its type.
@@ -170,6 +172,13 @@ protected:
         std::initializer_list<SOElement> uniqueFields,
         std::initializer_list<SOElement> commonFields = {})
     {
+        if (auto const item = findByType(type))
+        {
+            LogicError(
+                std::string("Duplicate key for item '") + name +
+                "': already maps to " + item->getName());
+        }
+
         formats_.emplace_front(name, type, uniqueFields, commonFields);
         Item const& item{formats_.front()};
 
@@ -180,6 +189,8 @@ protected:
     }
 
 private:
+    std::string name_;
+
     // One of the situations where a std::forward_list is useful.  We want to
     // store each Item in a place where its address won't change.  So a node-
     // based container is appropriate.  But we don't need searchability.
--- a/src/ripple/protocol/LedgerFormats.h
+++ b/src/ripple/protocol/LedgerFormats.h
@@ -24,76 +24,184 @@
 
 namespace ripple {
 
-/** Ledger entry types.
+/** Identifiers for on-ledger objects.
 
-    These are stored in serialized data.
+    Each ledger object requires a unique type identifier, which is stored
+    within the object itself; this makes it possible to iterate the entire
+    ledger and determine each object's type and verify that the object you
+    retrieved from a given hash matches the expected type.
 
-    @note Changing these values results in a hard fork.
+    @warning Since these values are stored inside objects stored on the ledger
+             they are part of the protocol. **Changing them should be avoided
+             because without special handling, this will result in a hard
+   fork.**
+
+    @note Values outside this range may be used internally by the code for
+          various purposes, but attempting to use such values to identify
+          on-ledger objects will results in an invariant failure.
+
+    @note When retiring types, the specific values should not be removed but
+          should be marked as [[deprecated]]. This is to avoid accidental
+          reuse of identifiers.
+
+    @todo The C++ language does not enable checking for duplicate values
+          here. If it becomes possible then we should do this.
 
     @ingroup protocol
 */
-// Used as the type of a transaction or the type of a ledger entry.
-enum LedgerEntryType {
-    /** Special type, anything
-        This is used when the type in the Keylet is unknown,
-        such as when building metadata.
-    */
-    ltANY = -3,
+// clang-format off
+enum LedgerEntryType : std::uint16_t
+{
+    /** A ledger object which describes an account.
 
-    /** Special type, anything not a directory
-        This is used when the type in the Keylet is unknown,
-        such as when iterating
-    */
-    ltCHILD = -2,
+        \sa keylet::account
+     */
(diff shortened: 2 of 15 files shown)


src/ripple/protocol/LedgerFormats.h
```diff
@@ -24,76 +24,177 @@
...
+    within the object itself; this makes it possible to iterate the entire
+    ledger and determine each object's type and verify that the object you
+    retrieved from a given hash matches the expected type.
 
-    @note Changing these values results in a hard fork.
+    @warning The range of valid on-ledger identifiers is between 1 and 32767
+             inclusive. Since these values are stored inside objects stored
+             on the ledger they are part of the protocol. **Changing them
+             should be avoided because without special handling, because
+             it will result in a hard fork.**
```
gregtatcam, line 38: Typo. Second `because` should be removed.

src/ripple/protocol/LedgerFormats.h
```diff
@@ -24,76 +24,177 @@
...
+        unknown or unimportant. One such use case is when building metadata.
+
+        \sa keylet::unchecked
+    */
+    ltANY = 0xF001,
+
+    /** A special type, matching any ledger entry type that is not a directory.
+
+        This type is used when the desired object must not be a directory entry
+        but otherwise the object's specific type is either unimportant.
```
gregtatcam, line 166: Typo. Should be `either unknown or unimportant`.

src/ripple/protocol/LedgerFormats.h
```diff
@@ -163,7 +163,7 @@ enum LedgerEntryType : std::uint16_t
     /** A special type, matching any ledger entry type that is not a directory.
 
         This type is used when the desired object must not be a directory entry
-        but otherwise the object's specific type is either unimportant.
+        but otherwise the object's specific type is either unkown or unimportant.
```
HowardHinnant, line 166: One more nitpicky type-o fix:  unknown

src/ripple/protocol/LedgerFormats.h
```diff
@@ -24,76 +24,177 @@
...
+/** Identifiers for on-ledger objects.
 
-    These are stored in serialized data.
+    Each ledger object requires a unique type identifier, which is stored
+    within the object itself; this makes it possible to iterate the entire
+    ledger and determine each object's type and verify that the object you
+    retrieved from a given hash matches the expected type.
 
-    @note Changing these values results in a hard fork.
+    @warning The range of valid on-ledger identifiers is between 1 and 32767
```
thejohnfreeman, line 34: How about adding a parenthetical "(all positive integers representable with 15 bits)"?

src/ripple/protocol/LedgerFormats.h
```diff
@@ -24,76 +24,177 @@
...
+    /** A ledger object which describes an account.
 
-    /** Special type, anything not a directory
-        This is used when the type in the Keylet is unknown,
-        such as when iterating
-    */
-    ltCHILD = -2,
+        \sa keylet::account
+     */
+    ltACCOUNT_ROOT = 0x0061,
```
thejohnfreeman, line 60: How about adding a comment for these converted ASCII values pointing out that they were chosen based on an ASCII character? e.g. a trailing comment `// 'a'`?

src/ripple/protocol/LedgerFormats.h
```diff
@@ -24,76 +24,177 @@
...
+     */
+    ltCHECK = 0x0043,
+
+    /** A ledger object which describes a deposit preauthorization.
+
+        \sa keylet::depositPreauth
+     */
+    ltDEPOSIT_PREAUTH = 0x0070,
+
+    /** The ledger object is the negative UNL object.
```
thejohnfreeman, line 145: Nit: "object that is the"

src/ripple/protocol/impl/STParsedJSON.cpp
```diff
@@ -312,14 +312,7 @@ parseLeaf(
...
 
-                            if (!(0u <= type &&
-                                  type <=
-                                      std::min<unsigned>(
-                                          std::numeric_limits<
-                                              std::uint16_t>::max(),
-                                          std::numeric_limits<
-                                              std::underlying_type_t<
-                                                  LedgerEntryType>>::max())))
+                            if (type < 1 || type >= 32768)
```
thejohnfreeman, line 315: How about `std::numeric_limits<std::underlying_type_t<LedgerEntryType>>::max() - 1`? or with `std::uint16_t` instead of `std::underlying_type_t<LedgerEntryType>`?

thejohnfreeman (commented):
Great stuff!

src/ripple/protocol/LedgerFormats.h
nbougalis: Hmm... I'll have to think about that because I intentionally wanted to move away from the  "ascii characters" thing. In an ideal world, we wouldn't have to deal with any of these, as these values would have had used normal `enum` semantics, so that we could avoid having to explicitly choose & specify the value of each member, which is error prone and unnecessarily verbose.

src/ripple/protocol/impl/STParsedJSON.cpp
nbougalis: The problem is that anything over 2<sup>15</sup> is not valid in this context, which is why I changed this code. I hear what you're saying thought; these are just magic numbers, pulled out of thin air, as far as the reader is concerned.

I wondered whether this is an improvement:

```
inline bool isValidLedgerEntryType(std::uint16_t value)
{
    return (value != 0) && (value < 32768);
}

inline bool isValidLedgerEntryType(LedgerEntryType value)
{
    return isValidLedgerEntryType<std::underlying_type_t<LedgerEntryType>>(value);
}
```
This check also be added to the `LedgerEntryTypesMatch` invariant check; it has an explicit check for all "known" values, but this check would be cheap and add an additional layer of safety.

But, after typing this, it occurs to me that this may be overkill: since these are "protocol defined" minima and maxima we could simply add two constants to **`Protocol.h`**:

```
/** The smallest valid type identifier for a ledger entry. */
std::uint16_t constexpr minLedgerEntryType = 1;

/** The largest valid type identifier for a ledger entry. */
std::uint16_t constexpr maxLedgerEntryType = 32767;
```

Thoughts?

src/ripple/protocol/LedgerFormats.h
nbougalis: I think that the existing comment is confusing enough and (imo) adding that increase confusion. How about using _2^15-1_ instead, with an additional reference to the `minLedgerEntryType` and `maxLedgerEntryType` constants, described below?

nbougalis:
I force-pushed this branch to add several improvements. @thejohnfreeman, I gave some thought to some of your comments, revisited the changes and I think that the end result is an improvement; really curious to see what you think.

HowardHinnant (approved):

src/ripple/protocol/KnownFormats.h
```diff
@@ -35,7 +36,7 @@ namespace ripple {
 
     @tparam KeyType The type of key identifying the format.
 */
-template <class KeyType>
+template <class KeyType, class Derived>
```
seelabs, line 39: The only reason to use the CRTP is to initialize the `name_` data member.  CRTP seems like overkill for such a use case. I'd prefer that we just pass the name to the constructor. This will always be done from the constructor from the derived class, so it's still easy for callers to use.
As a side note, if you did want to use the CRTP, you don't need to store the member.

src/ripple/protocol/impl/Keylet.cpp
```diff
@@ -23,17 +23,16 @@
...
+    assert(sle.getType() != ltANY || sle.getType() != ltCHILD);
+
     if (type == ltANY)
         return true;
-    if (type == ltINVALID)
-        return false;
+
     if (type == ltCHILD)
-    {
-        assert(sle.getType() != ltDIR_NODE);
```
seelabs, line 34: What are the cases where we expect `check` to fail? I think they are: 1) logic errors, 2) corrupt data, 3) hash collisions. If that's true, maybe we should assert of any `check` failure? At any rate, I'm OK removing the assert, it does seem like it should either always assert on any `check` failure or never assert.

seelabs (commented):
Looks good, but before I give a thumbs up consider the comment about replacing the CRTP implementation with passing a string to the constructor. I'll give a thumbs up after we talk about that.

PR #3924, opened 2021-09-03
Eliminate code duplication and improve documentation:

nbougalis:
The legacy functions `cdirFirst` and `dirFirst` were mostly identical; the differences were only type-related. The same situation existed with `cdirNext` and `dirNext`.

This commit removes the duplicated code by introducing new template functions that abstract away the differences that are present between each pair of functions.

This commit also improves the naming of function arguments, helping to elucidate their purpose & use and to make the code self-documenting.

- [ ] Documentation Updates
- [ ] Release

### DIFF
--- a/src/ripple/ledger/View.h
+++ b/src/ripple/ledger/View.h
@@ -125,30 +125,6 @@ transferRate(ReadView const& view, AccountID const& issuer);
 [[nodiscard]] bool
 dirIsEmpty(ReadView const& view, Keylet const& k);
 
-// Return the first entry and advance uDirEntry.
-// <-- true, if had a next entry.
-// VFALCO Fix these clumsy routines with an iterator
-bool
-cdirFirst(
-    ReadView const& view,
-    uint256 const& uRootIndex,            // --> Root of directory.
-    std::shared_ptr<SLE const>& sleNode,  // <-> current node
-    unsigned int& uDirEntry,              // <-- next entry
-    uint256& uEntryIndex,  // <-- The entry, if available. Otherwise, zero.
-    beast::Journal j);
-
-// Return the current entry and advance uDirEntry.
-// <-- true, if had a next entry.
-// VFALCO Fix these clumsy routines with an iterator
-bool
-cdirNext(
-    ReadView const& view,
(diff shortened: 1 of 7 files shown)


scottschurr (approved):
:+1:  Looks like a good change to me.  Should be all set to go once clang-format is happy.

src/ripple/ledger/View.h
```diff
@@ -222,29 +198,69 @@ adjustOwnerCount(
...
 
-// Return the first entry and advance uDirEntry.
-// <-- true, if had a next entry.
-// VFALCO Fix these clumsy routines with an iterator
+/** @{ */
+/** Returns the first entry in the directory, advancing the index
+
+    @deprecated These are legacy function that are considered deprecated
+                and will soon be replaced with an iterator-based model
+                that is easier to use. You should not use them in new code.
```
seelabs, line 206: I'm not sure we can deprecate something until the replacement it available.

src/ripple/ledger/impl/View.cpp
```diff
@@ -34,6 +34,126 @@
...
 namespace ripple {
 
+namespace detail {
+
+template <
+    class V,
+    class N,
+    class = std::enable_if_t<
+        std::is_same_v<std::remove_cv_t<N>, SLE> &&
+        std::is_base_of_v<ReadView, V>>>
```
seelabs, line 44: Consider removing the `enable_if`. Rational: we aren't using `enable_if` to resolve an overload, we're using it as a type-check and as documentation. I'd probably leave it off entirely and add a comment. This is a `detail` function only called from this file (it's not on a public interface where clients can call it with weird parameter types and wonder what's wrong). If the wrong types are passed in this file, we'll still get a compile error and quickly figure out why. 

If you don't like the comment, as an alternative, consider adding a `static_assert`. I'd prefer we didn't increase our use of `enable_if` to cases like this. When we see `enable_if` I want to think "this is part of an overload set".

seelabs (commented):
This looks good, but I'd like to discuss the `enable_if` before giving a thumbs up.

PR #3929, opened 2021-09-11
Improve the MITM detection during session establishment:

### LINKED ISSUE #2413: peer handshake violates TLS transparency
Current design of the peer protocol handshake limits alternative implementations in languages other than C/C++ as described in [this article][2]. Indeed, `src/ripple/overlay/README.md` refers to the current implementation rather than describing the protocol:
```
* `Session-Signature`

    For more details see the corresponding source code.
```
apparently referring to [lines 31-93][1] in `src/ripple/overlay/impl/TMHello.cpp`.

OpenSSL routines `SSL_get_finished` and `SSL_get_peer_finished` are being used to access `Finished` messages sent over the socket which violates TLS socket transparency. These low-lever routines are only available in C/C++ for a reason.

Requesting protocol upgrade to RTXP/1.3 with `Session-Signature` (HTTP header) and `nodeproof` (the corresponding field of the `Hello` message in `protobuf` protocol) replaced by a more portable mechanism which would not be language-specific and could be implemented in other programming languages, for example using Node.js.

[1]: https://github.com/ripple/rippled/blob/develop/src/ripple/overlay/impl/TMHello.cpp#L31-L93
[2]: https://medium.com/@alexberegszaszi/why-is-it-unlikely-to-have-a-complete-alternative-implementation-of-ripple-513c8f258238

nbougalis:
Even for SSL/TLS encrypted connections, it is possible for a determined attacker to mount certain types of relatively inexpensive MITM attacks that might make it possible to intelligently tamper with messages exchanged between endpoints.

This risk can be mitigated if each side has a certificate from CA that the other side trusts. But, in the context of a decentralized and permissionless network, requiring certificates is neither reasonable nor desirable.

Ultimately, the goal is to ensure that two endpoints A and B know that they are talking directly to each other over a single end-to-end TLS session instead of two separate TLS sessions, with the attacker acting as a proxy.

The XRP Ledger protocol prevents this by leveraging the fact that the two servers each have a node identity, in the form of a **`secp256k1`** keypair.

By deriving a secure fingerprint associated with the TLS session and never sent over the wire, which both sides sign with their respective node private key and send to the other, we can strongly bind the TLS session to the node identities of each of the servers at the end of the TLS session.

This commit changes the way the fingerprint is derived, moving away from using `SSL_get_finished` and `SSL_get_peer_finished`, in favor of using `SSL_export_keying_material`.

The change is backwards compatible and servers with this commit will still generate and verify old-style fingerprints, but they also introduce support for new-style fingerprints, generated using `SSL_export_keying_material`.

For a fuller discussion on this topic, please see:
    https://github.com/openssl/openssl/issues/5509

This commit fixes #2413, which was already closed as a 'WONTFIX'.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### DIFF
--- a/src/ripple/basics/base_uint.h
+++ b/src/ripple/basics/base_uint.h
@@ -524,6 +524,7 @@ public:
 using uint128 = base_uint<128>;
 using uint160 = base_uint<160>;
 using uint256 = base_uint<256>;
+using uint512 = base_uint<512>;
 
 template <std::size_t Bits, class Tag>
 inline int
--- a/src/ripple/basics/impl/make_SSLContext.cpp
+++ b/src/ripple/basics/impl/make_SSLContext.cpp
@@ -20,8 +20,6 @@
 #include <ripple/basics/chrono.h>
 #include <ripple/basics/contract.h>
 #include <ripple/basics/make_SSLContext.h>
-#include <ripple/beast/container/aged_unordered_set.h>
-#include <cstdint>
 #include <sstream>
 #include <stdexcept>
 
@@ -106,7 +104,6 @@ using rsa_ptr = custom_delete_unique_ptr<RSA>;
 static rsa_ptr
 rsa_generate_key(int n_bits)
 {
-#if OPENSSL_VERSION_NUMBER >= 0x00908000L
     BIGNUM* bn = BN_new();
     BN_set_word(bn, RSA_F4);
 
@@ -118,9 +115,6 @@ rsa_generate_key(int n_bits)
     }
 
     BN_free(bn);
-#else
-    RSA* rsa = RSA_generate_key(n_bits, RSA_F4, nullptr, nullptr);
-#endif
 
     if (rsa == nullptr)
         LogicError("RSA_generate_key failed");
--- a/src/ripple/overlay/README.md
+++ b/src/ripple/overlay/README.md
@@ -65,9 +65,10 @@ Network-ID: 1
 Network-Time: 619234489
 Public-Key: n94MvLTiHQJjByfGZzvQewTxQP2qjF6shQcuHwCjh5WoiozBrdpX
 Session-Signature: MEUCIQCOO8tHOh/tgCSRNe6WwOwmIF6urZ5uSB8l9aAf5q7iRAIgA4aONKBZhpP5RuOuhJP2dP+2UIRioEJcfU4/m4gZdYo=
+Session-EKM-Signature: 4F73F4DF23453250BDDF81AFBB5CBF0FD32B92AAF548331DB0ECF1DB988BE2F3D7264621FC1E93404B77612F235AC6FFD4F1B7B4D420E8071384273D9A60201D
 Remote-IP: 192.0.2.79
-Closed-Ledger: llRZSKqvNieGpPqbFGnm358pmF1aW96SDIUQcnMh6HI=
-Previous-Ledger: q4aKbP7sd5wv+EXArwCmQiWZhq9AwBl2p/hCtpGJNsc=
+Closed-Ledger: 4F73F45E95343F7446F91B5F70E3AE4E155CCAA2B1CF2F267C99FD39C1C24178
+Previous-Ledger: F9D75AAB7D975BADC6D008C4AE94D9B7B6AA323EEE4B133F5B75CE92445C88ED
 ```
 
 ##### Example HTTP Upgrade Response (Success)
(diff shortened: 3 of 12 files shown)


donovanhide:
That took a while :-) I remember pointing out the problems with `SSL_get_finished` and `SSL_get_peer_finished` way back in a security report in February 2015. Never did get that bounty :-)

nbougalis:
> That took a while :-) I remember pointing out the problems with `SSL_get_finished` and `SSL_get_peer_finished` way back in a security report in February 2015. Never did get that bounty :-)

It did take a while but better late than never, I guess. 😅

As for the bounty, I'm sure I have your original report somewhere, can you please send it to me via email or (in the interest of full transparency) paste it here?

If it qualifies as a security issue I'll look into having a bounty awarded.

With that said, I am skeptical; the technique used is secure, and there are no bugs—at least that I'm aware of—in the implementation. Yes, it's awkward and reaches way too deeply into the TLS session and I know it frustrated experts like @richsalz, but, mess or not, it does work and work correctly.

And since I mentioned @richsalz, I'm curious whether he feels that is a better that using ``SSL_get_finished` and `SSL_get_peer_finished`

Still a fix is now available and, hopefully, this version passes muster. 
 
So send over the report and I'll review it and, if appropriate, recommend that you receive an appropriate bounty.

donovanhide:
Here's the original report from February 2015:

https://docs.google.com/document/d/1109KT3AJUe68XjnRLUkidzQrRhFs5pKc-2A_hXsuvEI/edit?usp=sharing

Seeing as session resumption is still not disabled in rippled and you can still use `SSL_get_finished` and `SSL_get_peer_finished`, maybe it is still possible...

https://github.com/ripple/rippled/blob/7bd5d51e4e4e76a5547051d30b330739618eddb0/src/ripple/basics/impl/make_SSLContext.cpp#L390

nbougalis:
> Here's the original report from February 2015:
> 
> https://docs.google.com/document/d/1109KT3AJUe68XjnRLUkidzQrRhFs5pKc-2A_hXsuvEI/edit?usp=sharing
> 
> Seeing as session resumption is still not disabled in rippled and you can still use `SSL_get_finished` and `SSL_get_peer_finished`, maybe it is still possible...
> 
> https://github.com/ripple/rippled/blob/7bd5d51e4e4e76a5547051d30b330739618eddb0/src/ripple/basics/impl/make_SSLContext.cpp#L390

Thanks! I'll review this closely and follow up.

nbougalis:
3SHAKE aside, it's really tricky to use exporters correctly and I'm no longer convinced that this technique is as effective as we think. At a minimum, a mechanism is needed to allow both endpoints to collaboratively generate a random "cookie" that then gets passed into `SSL_export_keying_material`, so that the resulting fingerprint is not under the control of either party.

I'm going to close this PR and think this through more carefully.

richsalz:
@nbougalis: yes, using exporters is not as easy as it perhaps should be.  I'm glad to see you're taking another look. (FWIW, I don't do OpenSSL any more, but happy to talk about TLS in generic terms.)

MarkusTeufelberger:
https://github.com/libp2p/specs/blob/master/tls/tls.md was already mentioned in the issue a few years ago, maybe it might be better to implement this in https://github.com/libp2p/cpp-libp2p and get the other features from there than trying to make the current construction secure? Would be a larger undertaking though.

nbougalis:
> @nbougalis: yes, using exporters is not as easy as it perhaps should be.  I'm glad to see you're taking another look. (FWIW, I don't do OpenSSL any more, but happy to talk about TLS in generic terms.)
> 
> 

Thanks for following up. It's just hard to tightly bind a higher-level protocol to a TLS session, when with something like EKM. 

In reading @MarkusTeufelberger's comment below, about encoding the **`secp256k1** key in an X.509 structure, I kinda wish that it was possible to use such keys with TLS and have the master secret derived using **ECIES**.

By the way maybe you can answer a question: why does the EKM spec _only_ use the master secret and the two client random values instead of, say, the finished messages themselves? They contain, inter alia, things like any exchanged certs, which seem relevant.

nbougalis:
> https://github.com/libp2p/specs/blob/master/tls/tls.md was already mentioned in the issue a few years ago, maybe it might be better to implement this in https://github.com/libp2p/cpp-libp2p and get the other features from there than trying to make the current construction secure? Would be a larger undertaking though.

I think that migrating the P2P layer to a library is a little ambitious; it's unclear to me if the benefits would outweigh the costs, but I think it's reasonable to have a conversation about it. Would you like to open an issue to discuss this topic separately?

The idea of encoding the node's public key into a certificate is interesting. It's certainly possible for the self-signed certificates most nodes have, but it could be a little annoying (or maybe not even possible) for operators that use certificates from trusted CAs (like my employer) so that's something to consider.

MarkusTeufelberger:
Inter-node communication can maybe use different certificates than pure API endpoints. I doubt that you'll get a certificate with custom fields from a widely trusted CA, but I haven't looked into this - maybe to differentiate themselves from Let's Encrypt this is easier to get from them nowadays? Though "[Honest Achmed's Used Cars and Certificates](https://bugzilla.mozilla.org/show_bug.cgi?id=647959)" surely will be able to sell you whatever you want!

Joking aside, it might at least be a bit easier to implement in a non-OpenSSL way.

richsalz:
> why does the EKM spec only use the master secret and the two client random values instead of, say, the finished messages themselves? They contain, inter alia, things like any exchanged certs, which seem relevant.

Because you don't know when the EKM will happen and you don't want to force everyone to keep the messages around. The TLS connection is authenticated because the digest includes all the messages anyway.

PR #3930, opened 2021-09-15
Build packages with new LICENSE.md file.

mtrippled:
## High Level Overview of Change
The LICENSE file is copied as part of package building. This file name was recently changed to LICENSE.md, which broke package building.

### Type of Change

- [x ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Refactor (non-breaking change that only restructures code)
- [ ] Tests (You added tests for code that already exists, or your new feature included in this PR)
- [ ] Documentation Updates
- [ ] Release

Package building was broken in b5 by the following commit: 0d17dd8228c4a5b40710d476a1ebf8c300624554

PR #3931, opened 2021-09-16
A non markdown LICENSE file

legleux:
### Context of Change

LICENSE.md broke packaging

### Type of Change


legleux:
It's only visible if it's in the develop branch 😑

MarkusTeufelberger:
Might also be useful to tag file with https://spdx.dev/ - but that's a bigger change.

LICENSE
```diff
@@ -1,5 +1,3 @@
-ISC License
```
nbougalis, line 1: This will cause license autodetection to fail in GitHub. It needs to be added back.

nbougalis (changes requested):

LICENSE
legleux: Are you just referring to this when you go to the license or is there something else to it?
image
It wasn't visible because I removed the commit from my develop branch.

nbougalis: Actually, do we want this or #3930?

I'm fine with either, but if we go with this, the comment I left needs to be addressed.

PR #3933, opened 2021-09-17
Implement tx stream with history

pwang200:
## Subscribe to an account's transaction stream for both historical and future transactions.

- An option `account_history_tx_stream` is added to the `subscribe` RPC, e.g.
```
{
  "command": "subscribe",
  "account_history_tx_stream": {
    "account": "r4rbNiTmYdSfvDkaKULBcRdYy1oTvcTpQB"
  }
}
```
- The transactions are streamed according to the order they are applied to the ledgers.
- Both future and historical transactions will be streamed. Future transactions will have  `account_history_tx_index` fields starting from `0`. Historical ones will have negative `account_history_tx_index` fields starting from `-1`.
- The historical part of the stream can be stopped by setting an optional field `stop_history_tx_only` as true in a `unsubscribe` RPC, e.g.
```
{
  "command": "unsubscribe",
  "account_history_tx_stream": {
    "account": "rHb9CJAWyB4rj91VRWn96DkukG4bwdtyTh",
    "stop_history_tx_only": true
  }
}
```
The future transactions will still be streamed in this case. To stop both directions, send
```
{
  "command": "unsubscribe",
  "account_history_tx_stream": {
    "account": "rHb9CJAWyB4rj91VRWn96DkukG4bwdtyTh"
  }
}
```
- The first transaction of the account will have an optional field `account_history_tx_first` returned as true.
 
### Type of Change

## Test Plan
- unit tests for XRP payments and IOU payments
- manual tests for some other transaction types


src/ripple/net/InfoSub.h
```diff
@@ -168,6 +197,12 @@ class InfoSub : public CountedObject<InfoSub>
     void
     deleteSubAccountInfo(AccountID const& account, bool rt);
 
+    bool
```
miguelportilla, line 200: It would be nice to document what the return value means.

src/ripple/rpc/handlers/Subscribe.cpp
```diff
@@ -201,6 +201,24 @@ doSubscribe(RPC::JsonContext& context)
...
+        if (!req.isMember(jss::account) || !req[jss::account].isString())
+            return rpcError(rpcINVALID_PARAMS);
+
+        auto const id = parseBase58<AccountID>(req[jss::account].asString());
+        if (!id)
+            return rpcError(rpcINVALID_PARAMS);
+
+        if (auto result = context.netOps.subAccountHistory(ispSub, *id);
+            result != rpcSUCCESS)
+            return rpcError(result);
```
miguelportilla, line 216: Nit: consider using curly brackets here.

cjcobb23:
I am hesitant about merging this PR.

1. This functionality can already be achieved with a combination of `account_tx` and subscribing to the `accounts` stream. The functionality added here is basically just merging those two RPCs together in a way that can already be done client side. This PR adds additional code and complexity to rippled and the API that is not necessary.
2. The SQLite database that account_tx uses will be removed as part of migrating large parts of the RPC subsystem to project clio. When that happens, the historical part of this command is not going to work, because the historical part of this command uses the same database as account_tx. Migrating the RPC subsystem to project clio is a very large and difficult task, and I don't want to add another blocker to that effort, especially when we can just achieve this functionality with the existing API.
3. I am concerned about load on the server. account_tx is one of the most expensive RPC calls and imposes the most load on the server of probably any RPC call. The functionality added here basically allows a client to force the server to continuously call account_tx. A single client can use this new functionality to subscribe to many different accounts, and force the server to basically continuously call account_tx for many accounts. The difference between this new functionality and account_tx though is that the server can reject calls to account_tx when under high load. With this new functionality though, is there a way for the server to determine that it is under too high of a load and thus terminate some of these long running `account_history_tx_stream` subscriptions?

All in all, I am opposed to merging this. However, I understand this is for sidechains, and we are under some deadline pressure there, so I am ok with merging this (after review) as long as a few things are done:
1. We mark this as experimental, absolutely DO NOT DOCUMENT this and add a warning to the response saying something like "This functionality can be removed without warning at any time".
2. We commit to removing this and implementing the required functionality client side in side chains within about 6 months time.

Let me know if this makes sense.

faves4less (approved):

faves4less (approved):

src/ripple/app/misc/NetworkOPs.cpp
```diff
@@ -3117,6 +3286,296 @@ NetworkOPsImp::unsubAccountInternal(
...
+                auto [txns, marker] =
+                    dynamic_cast<RelationalDBInterfaceSqlite*>(
+                        &app_.getRelationalDBInterface())
+                        ->newestAccountTxPage(options);
+
+                for (auto const& [tx, meta] : txns)
+                {
+                    if (!tx || !meta)
+                    {
+                        send(rpcError(rpcINTERNAL), true);
```
cjcobb23, line 3407: I think we should log and assert here.

src/ripple/app/misc/NetworkOPs.cpp
```diff
@@ -3117,6 +3286,296 @@ NetworkOPsImp::unsubAccountInternal(
...
+                        return;
+                    }
+                    std::shared_ptr<STTx const> stTxn = tx->getSTransaction();
+                    if (!stTxn)
+                    {
+                        send(rpcError(rpcINTERNAL), true);
+                        return;
+                    }
+                    Json::Value jvTx = transJson(
+                        *stTxn, meta->getResultTER(), true, curTxLedger);
```
cjcobb23, line 3424: I think you can just call `stTxn->getJson()`, and then you don't need to pass in the ledger. Then you can avoid looking up the ledger from `LedgerMaster` for every transaction

src/ripple/app/misc/NetworkOPs.cpp
```diff
@@ -3117,6 +3286,296 @@ NetworkOPsImp::unsubAccountInternal(
...
+    {
+        simIterator->second.emplace(isrListener->getSeq(), ahi);
+    }
+
+    auto const ledger = app_.getLedgerMaster().getValidatedLedger();
+    if (ledger)
+        subAccountHistoryStart(ledger, ahi);
+    else
+        JLOG(m_journal.debug())
+            << "subAccountHistory, no validated ledger yet, delay start";
```
cjcobb23, line 3532: What actually happens in this case? Is the subscription successful? Maybe add a comment explaining what happens in this case.

cjcobb23 (changes requested):
Did a first pass. Mostly small things, but we need to make this work with reporting mode, because right now I think it's going to crash.

src/ripple/app/misc/NetworkOPs.cpp
```diff
@@ -3117,6 +3286,296 @@ NetworkOPsImp::unsubAccountInternal(
...
+                    return;
+                }
+
+                RelationalDBInterface::AccountTxPageOptions options = {
+                    accountId, startLedgerSeq, lastLedgerSeq, {}, 0, true};
+
+                auto [txns, marker] =
+                    dynamic_cast<RelationalDBInterfaceSqlite*>(
+                        &app_.getRelationalDBInterface())
+                        ->newestAccountTxPage(options);
```
cjcobb23, line 3401: I think we should charge the consumer for each of these calls. These calls are heavy weight, and impose a lot of load on the server. Here is an example of how to do that: https://github.com/ripple/rippled/blob/8a25f3282426f382e1933242ef875d5aaf3e31e9/src/ripple/rpc/impl/ServerHandlerImp.cpp#L425
We usually consider account_tx calls as `feeMediumBurdenRPC`

src/ripple/rpc/handlers/Subscribe.cpp
```diff
@@ -201,6 +201,24 @@ doSubscribe(RPC::JsonContext& context)
         JLOG(context.j.debug()) << "doSubscribe: accounts: " << ids.size();
     }
 
+    if (context.params.isMember(jss::account_history_tx_stream))
+    {
```
cjcobb23, line 205: I think we should charge consumers for this call. We can do that by just adding:
```
context.loadType = Resource::feeMediumBurdenRPC;
```

cjcobb23 (changes requested):
Two comments about charging consumers for calling this RPC. This RPC can put a lot of load on the server, since it's basically calling account_tx in a loop (one of the most burdensome RPCs that we support), and I want to prevent clients from subscribing to this stream a bunch of times and thus putting a ton of load on the server.

nbougalis:
I share @cjcobb23's concerns here. Why do we need an entirely new command for this? It seems that `account_tx` is sufficient. If necessary, we could implement one minor improvement, by allowing the command to accept a transaction ID and beginning to walk back from there.

@pwang200, @seelabs: can you explain why this is needed and why using `account_tx` is insufficient?

pwang200:
@nbougalis,

@cjcobb23 @seelabs and I had discussions, and here is a recap of the conversations: 
The RPC grew to the current state and it does the following currently with one stream, 
-- wait for a validated ledger, as one design to make sure no gap nor overlap on the two directions of the stream.
-- go forward and stream the txns similar to `accounts` stream
-- go backward and stream the txns similar to `account_tx`
-- misc conveniences for the federator, such as set a flag when streaming the first historical tx and wait for historical txns if the server is backfilling its missing ledgers, provide continuous indexes of both backward and forward txns, etc.

I agree that the RPC's logic can be implemented in the client side (the federator), with minor improvement to the existing RPCs. The current federator, however, relies on the conveniences. The relevant parts of the federator can be redesigned but not trivially. Also when we started, we did not know what we know now. The historical direction of the RPC did not use `account_tx` logic, it used `PreviousTxnID` of `AccountRoot`.

Creating the PR and hopefully merging it has some benefits. The code will be reviewed, and if merged it will be run on devnet, testnet, and maybe mainnet. This allows developers to have an accessible "mainchain" with existing documentation of how to connect etc, so that they can focus on their sidechains. The downside is that we will take the RPC out of the codebase when federator is refactored to move the logic of the RPC over to the client side.

src/ripple/app/misc/NetworkOPs.cpp
pwang200: In `transJson()` it calls `getJson()` and adds more fields. Some of the fields are needed.

src/ripple/app/misc/NetworkOPs.cpp
pwang200: Thanks, good idea, added.

src/ripple/rpc/handlers/Subscribe.cpp
pwang200: Thanks, good idea, added.

src/ripple/app/misc/NetworkOPs.cpp
pwang200: Thanks, comment added.

src/ripple/app/misc/NetworkOPs.cpp
```diff
@@ -3117,6 +3286,296 @@ NetworkOPsImp::unsubAccountInternal(
...
+                        *stTxn, meta->getResultTER(), true, curTxLedger);
+                    jvTx[jss::meta] = meta->getJson(JsonOptions::none);
+                    jvTx[jss::account_history_tx_index] = txHistoryIndex--;
+                    RPC::insertDeliveredAmount(
+                        jvTx[jss::meta], *curTxLedger, stTxn, *meta);
+                    if (isFirstTx(tx, meta))
+                        jvTx[jss::account_history_tx_first] = true;
+
+                    send(jvTx, false);
+                    if (jvTx.isMember(jss::account_history_tx_first))
```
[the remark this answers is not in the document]
pwang200: right, changed

src/ripple/app/misc/NetworkOPs.cpp
pwang200: Do `tx` and `meta` guaranteed to be not null? I cannot tell from the `getRelationalDBInterface()` function document, and too unfamiliar with database code.

src/ripple/app/misc/NetworkOPs.cpp
```diff
@@ -3117,6 +3286,296 @@ NetworkOPsImp::unsubAccountInternal(
...
+                    setAccountHistoryJobTimer(subInfo);
+                    return;
+                }
+
+                RelationalDBInterface::AccountTxPageOptions options = {
+                    accountId, startLedgerSeq, lastLedgerSeq, {}, 0, true};
+
+                auto [txns, marker] =
+                    dynamic_cast<RelationalDBInterfaceSqlite*>(
+                        &app_.getRelationalDBInterface())
```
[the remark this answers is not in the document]
pwang200: right, checking dynamic cast result now.

src/ripple/app/misc/NetworkOPs.cpp
```diff
@@ -3117,6 +3286,296 @@ NetworkOPsImp::unsubAccountInternal(
...
+                            bool unsubscribe) -> bool {
+                if (auto sptr = subInfo.sinkWptr_.lock(); sptr)
+                {
+                    sptr->send(jvObj, true);
+                    if (unsubscribe)
+                        unsubAccountHistory(sptr, accountId, false);
+                    return true;
+                }
+                else
+                {
```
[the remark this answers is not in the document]
pwang200: right, changed

src/ripple/app/misc/NetworkOPs.cpp
```diff
@@ -3117,6 +3286,296 @@ NetworkOPsImp::unsubAccountInternal(
...
+                }
+            };
+
+            /*
+             * search backward until the genesis ledger or asked to stop
+             */
+            while (lastLedgerSeq >= 2 && !subInfo.index_->stopHistorical_)
+            {
+                auto startLedgerSeq =
+                    (lastLedgerSeq > 256 + 2 ? lastLedgerSeq - 256 : 2);
```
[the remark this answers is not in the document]
pwang200: yeah, added

src/ripple/app/misc/NetworkOPs.cpp
```diff
@@ -3117,6 +3286,296 @@ NetworkOPsImp::unsubAccountInternal(
...
+                                inner)
+                            {
+                                if (inner->getAccountID(sfAccount) == accountId)
+                                {
+                                    return true;
+                                }
+                            }
+                        }
+                    }
+                    catch (std::exception const&)
```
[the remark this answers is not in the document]
pwang200: right, removed.

src/ripple/app/misc/NetworkOPs.cpp
```diff
@@ -3117,6 +3286,296 @@ NetworkOPsImp::unsubAccountInternal(
...
+                    {
+                        continue;
```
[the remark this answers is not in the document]
pwang200: right, removed.

src/ripple/app/misc/NetworkOPs.cpp
```diff
@@ -3117,6 +3286,296 @@ NetworkOPsImp::unsubAccountInternal(
     }
 }
 
+void
+NetworkOPsImp::addAccountHistoryJob(SubAccountHistoryInfoWeak subInfo)
+{
+    app_.getJobQueue().addJob(
+        jtCLIENT, "AccountHistory", [this, subInfo](Job&) {
```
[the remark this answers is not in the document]
pwang200: ok, changed.

src/ripple/app/misc/NetworkOPs.cpp
```diff
@@ -3006,12 +3129,48 @@ NetworkOPsImp::pubAccountTransaction(
...
+                            ++it;
+                        }
+                        else
+                        {
+                            it = subs.erase(it);
+                        }
+                    }
+                    if (subs.empty())
+                        mSubAccountHistory.erase(histoIt);
+                }
```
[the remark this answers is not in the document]
pwang200: right, changed

src/ripple/app/misc/NetworkOPs.cpp
```diff
@@ -605,6 +638,55 @@ class NetworkOPsImp final : public NetworkOPs
     using SubInfoMapType = hash_map<AccountID, SubMapType>;
     using subRpcMapType = hash_map<std::string, InfoSub::pointer>;
 
+    struct SubAccountHistoryIndex
```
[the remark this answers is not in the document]
pwang200: right, added

src/ripple/rpc/handlers/Subscribe.cpp
pwang200: thanks, changed.

src/ripple/net/InfoSub.h
pwang200: thanks, added.

src/ripple/app/misc/NetworkOPs.cpp
cjcobb23: Well actually, in the regular `account_tx` handler, we do check for null too, so I think it's fine to leave this as is.

src/ripple/app/misc/NetworkOPs.cpp
```diff
@@ -3117,6 +3290,337 @@ NetworkOPsImp::unsubAccountInternal(
...
+            auto db = dynamic_cast<RelationalDBInterfaceSqlite*>(
+                &app_.getRelationalDBInterface());
+            if (!db)
+            {
+                JLOG(m_journal.debug())
+                    << "AccountHistory job for account " << toBase58(accountId)
+                    << " DB interface type wrong";
+                send(rpcError(rpcINTERNAL), true);
+                return;
+            }
```
cjcobb23, line 3365: We need to handle this case, and not return an error. Reporting mode can do what you want here. The Relational DB Interface was just poorly designed and thus you need to call a different function in this case.

cjcobb23 (changes requested):

src/ripple/app/misc/NetworkOPs.cpp
pwang200: Thanks for pointing out the reporting mode. I read the code, `RelationalDBInterfacePostgres` uses different functions (different interface) as you said, and the guarantee of the interface is not that clear to me from the docs. Of course it seems doable by following the example of `accountTx` in reporting mode. And more testing will be needed. 

But before making the change, I am wondering if it is needed, because as I understand, we don't plan to add this to reporting mode so we won't have a `RelationalDBInterfacePostgres` database backend.

cjcobb23: We have to add this to reporting mode, because our public client handlers on testnet and mainnet will be running in reporting mode. It will make it harder for people to use this if we don't add it to reporting mode.

pwang200: Right, added

PR #3935, opened 2021-09-27
Set OpenSSL 1.1.1 as the minimum version, recommend 3.0.0 for Windows

ximinez:

CMake config defined OpenSSL 1.0.2 as the minimum, even though building rippled would fail with that version for about 6 months. Update it to the newest commonly available version, 1.1.1.

3.0.0 was released a few weeks ago, but I'm not making that the minimum until it has had more time to be used and "shake out" issues, and to be included in more linux distros. However, I updated the Windows build instructions to recommend 3.0.0, which seems to work well there.

### Context of Change

Builds fail with older versions of OpenSSL. Because OpenSSL is security sensitive, we should keep up our requirements to match the latest version.

OpenSSL is referenced in two places in the CMake config. I don't know why that is, and am leaving that question as outside the scope of this PR.


seelabs (approved):
:+1: I don't have a window's system to tests this with, but the code changes look good. Thanks for doing this.

ximinez:
@seelabs Sorry, I was premature. My quick checks with 3.0.0 looked good, but the longer checks dug up some deprecated functions in 3.0.0 that prevent rippled from building, so I changed the VS README back to 1.1.1L.  I'm double-checking with that version now, but since I had previously been using 1.1.1K, I don't anticipate any problems.

legleux (approved):
🚢

manojsdoshi:
"Merged as part of #3948"

PR #3936, opened 2021-09-28
Remove old boost workaround

legleux:
Remove guard for [boost asio](https://github.com/boostorg/asio/commit/43874d5) issue on older apple clang.
Relevant for xcode 10.2 and boost < 1.69 which rippled requires higher of both now.

Remove spaces in CMake functions

### Type of Change

### DIFF
--- a/Builds/CMake/deps/Boost.cmake
+++ b/Builds/CMake/deps/Boost.cmake
@@ -2,49 +2,49 @@
    NIH dep: boost
 #]===================================================================]
 
-if ((NOT DEFINED BOOST_ROOT) AND (DEFINED ENV{BOOST_ROOT}))
-  set (BOOST_ROOT $ENV{BOOST_ROOT})
-endif ()
-file (TO_CMAKE_PATH "${BOOST_ROOT}" BOOST_ROOT)
-if (WIN32 OR CYGWIN)
+if((NOT DEFINED BOOST_ROOT) AND(DEFINED ENV{BOOST_ROOT}))
+  set(BOOST_ROOT $ENV{BOOST_ROOT})
+endif()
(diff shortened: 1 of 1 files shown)


Builds/CMake/deps/Boost.cmake
```diff
@@ -2,49 +2,49 @@
    NIH dep: boost
 #]===================================================================]
 
-if ((NOT DEFINED BOOST_ROOT) AND (DEFINED ENV{BOOST_ROOT}))
-  set (BOOST_ROOT $ENV{BOOST_ROOT})
-endif ()
-file (TO_CMAKE_PATH "${BOOST_ROOT}" BOOST_ROOT)
-if (WIN32 OR CYGWIN)
+if((NOT DEFINED BOOST_ROOT) AND(DEFINED ENV{BOOST_ROOT}))
```
seelabs, line 5: I prefer to keep the space after `if`, `else`, `elseif`, and `endif`

seelabs (commented):
The patch to remove the workaround looks good. I'd prefer we didn't make some of the formatting changes, tho I don't feel that strongly about it.

Builds/CMake/deps/Boost.cmake
legleux: Well, it's inconsistent in many places (even in the same file sometime) and I decided to go with what I've always seen in CMake docs/other peoples code and slowly make changes as I come across them.
Doesn't really matter at the end of the day, just wanted it consistent but spaces in CMake look weird to me.

seelabs: Looking at the official cmake documentation, it looks like they don't use a space after the `if` (although looking at the book "mastering cmake they do put a space after the `if`). Since I don't do much cmake coding anymore, I'm fine going with whatever your preference it. If you'd rather remove the space then let's do that.

PR #3937, opened 2021-09-29
Move TxMeta to xrpl_core

cjcobb23:
* add peekNodes() to TxMeta for const access to nodes

`TxMeta` is a simple class that contains useful abstractions and functions that other non-rippled C++ applications can use to build on the XRPL. rippled builds a library called xrpl_core that contains most of the XRPL core datastructures. This library can be used by other C++ applications. Specifically, Project Clio makes heavy use of xrpl_core. `TxMeta` is needed by Project Clio. Without the `TxMeta` abstraction, applications would be forced to use `STObject` to interact with the metadata, which can be cumbersome, since common operations like figuring out the delivered amount, traversing the nodes and getting affected accounts would need to be reimplemented. Transaction metadata is a core concept for XRPL, and providing easy ways to view the metadata is necessary for applications that build on XRPL.

Also, added a method for getting a const reference to the affected nodes.

### Context of Change

Needed for Project Clio, and useful for any C++ project building on XRPL.

### Type of Change

- [ ] Documentation Updates
- [ ] Release


src/ripple/protocol/TxMeta.h
```diff
@@ -101,6 +101,11 @@ class TxMeta
     {
         return (mNodes);
     }
+    STArray const&
+    peekNodes() const
```
thejohnfreeman, line 105: It wouldn't hurt to just call this `getNodes` too.

thejohnfreeman (approved):

ximinez (approved):
I agree with @thejohnfreeman's suggested name change. Otherwise, this is great. :+1:

src/ripple/protocol/TxMeta.h
cjcobb23: oh yeah, for some reason I thought this wouldn't compile. Let me change that.

natenichols (approved):

PR #3939, opened 2021-10-01
Disable Boost's deque forward declaration

ximinez:

Does what it says on the tin: Defines a build variable that tells Boost to include `deque` directly instead of trying to predeclare it. That variable is called `BOOST_CONTAINER_FWD_BAD_DEQUE`, 

### Context of Change

After upgrading Boost from 1.70 to 1.75, building in Visual Studio 2017 would fail, giving errors about `std::deque` missing template parameters. The `boost/detail/container_fwd.hpp` tries to forward declare the `std::deque` template without specifying default paramters. Defining `BOOST_CONTAINER_FWD_BAD_DEQUE` simply includes the `deque` header and removes the forward declaration.

I realized at some point that I had already "fixed" this issue in #3851 with the commit labelled "Add missing "deque" headers in a bunch of places". This fix is simpler and therefore better. It's also now causing problems across branches, so I'm making a separate PR for it.

### Type of Change

## Test Plan

This is only a build issue. There is nothing to test.

### DIFF
--- a/Builds/CMake/RippledCore.cmake
+++ b/Builds/CMake/RippledCore.cmake
@@ -128,6 +128,7 @@ target_include_directories (xrpl_core
 target_compile_definitions(xrpl_core
   PUBLIC
     BOOST_ASIO_USE_TS_EXECUTOR_AS_DEFAULT
+    BOOST_CONTAINER_FWD_BAD_DEQUE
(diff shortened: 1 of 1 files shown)


cjcobb23 (approved):
Thanks for fixing this. LGTM.

seelabs (approved):
:+1: LGTM

PR #3943, opened 2021-10-05
[TRIVIAL] Remove unused variable

seelabs:
Clang 13 warns about `nbytes` being unused. This patch removes this unused variable.


ximinez (approved):
:+1:, assuming CI builds pass. They're having some trouble with docker images at the moment.
(It did pass my local tests, though.)

PR #3951, opened 2021-10-11
Update account and owner reserve

alloynetworks:
Update defaults from 20/5 to 10/2 for account and owner reserves respectively. The network has been voting for 10/2 for an adequate period of time and a new server installation should default to the new reserves.


MarkusTeufelberger:
Remove the default _vote_, leave in the default _value_ and add a value representing the current fee structure on XRPL to the default config file since running `rippled` as a server for XRPL is likely the most common use case.

PR #3952, opened 2021-10-13
Move packaging code to another repo

legleux:
Created a new repo with the contents of `Build/containers` which runs packaging when the repo `rippled` is pushed to or receives a PR via a GitHub workflow.

This PR requires a few more changes.
1.  New repository created like [legleux/rippled-packages](https://github.com/legleux/rippled-packages)
2. Access token created for rippled repo and added as a secret so above repo can be triggered to build packages.

### Context of Change
This decouples the building/packaging pipeline from the source code of rippled. 

Another repo must be created in github.com/ripple (or whomever will build packages.) Currently this fork triggers packages to be built in [legleux/rippled-packages](https://github.com/legleux/rippled-packages)
The rippled repository needs an access token created and added as a secret so the workflow can use the `repository_dispatch` to trigger the build in the `rippled-packages` repo.

The base docker images used for packaging reside in `legleux/rippled-{dpkg,rpm}-builder` and are the images in artifactory. These images should be updated and moved to ripple's dockerhub account or authentication with our artifactory used in the `rippled-packages` repository.

### Type of Change


intelliot:
@legleux what's the status of this?

(if we are moving packaging code out of rippled, we should check with XRPLF, as they may want to take over the creation of packages. @RichardAH @WietseWind )

legleux:
> @legleux what's the status of this?
> 
> (if we are moving packaging code out of rippled, we should check with XRPLF, as they may want to take over the creation of packages. @RichardAH @WietseWind )

This PR was made a long time ago with the intent of building packages in gitlab.
If I were to do this today, I'd do it like clio does it (a github action job clones the clio packaging repo and builds them from clio repo.)
I'd like to unify the package building more and perhaps use [holy build box](https://github.com/phusion/holy-build-box) to generate the binary to be packaged or some more determinisic package building method

intelliot:
@legleux anything here still worth doing? 

If so, let me know what your next steps are. If not, feel free to close this PR.

PR #3956, opened 2021-10-18
batch test

pwang200:
Added a batch test that tries to accumulate cross chain payment quickly, so that the number of queued payment request is more than the threshold of triggering the door close event. When all the payment requests are processed, a door open event should happen. During the door close/open event loop, the ticket based unordered event handling protocol runs two times.


README.md
```diff
@@ -1,3 +1,97 @@
...
+
+The latest production code for the XRP ledger is on the "master" branch.
+
+Until this branch is merged with the mainline branch, it will periodically be
+rebased on that branch and force pushed to github.
+
+## What are side chains?
+
+Side chains are independent ledgers. They can have their own transaction types,
+their own UNL list (or even their own consensus algorithm), and their own set of
```
MarkusTeufelberger, line 19:
```suggestion
their own UNL (or even their own consensus algorithm), and their own set of
```

unique node list list is redundant.

PR #3964, opened 2021-10-22
Prefer using a local Sandbox over using the Context's view

scottschurr:

If a transactor has access to a `Sandbox`, using that `Sandbox` is preferable to directly using the `View` underlying the `Sandbox`.  This branch fixes some places in the `CreateOffer` transactor that were not following this guideline.

### Context of Change

These are errors that have crept in over time.  There have been a couple of recent bugs that brought attention to this kind of error, which is how they came to my attention.  The changed uses of `View` should not have caused any bugs, but were counter to current guidance.

### Type of Change

This change should not be user-visible.  The commit does not need to be mentioned in the release notes.  Standard unit tests and integration tests should be sufficient for test coverage.

I consider the change to be trivial, so I'm only requesting one reviewer.  I'm open to being persuaded that it needs a more serious review.

### DIFF
--- a/src/ripple/app/tx/impl/CreateOffer.cpp
+++ b/src/ripple/app/tx/impl/CreateOffer.cpp
@@ -637,7 +637,7 @@ CreateOffer::takerCross(
     Sandbox& sbCancel,
     Amounts const& takerAmount)
 {
-    NetClock::time_point const when = ctx_.view().parentCloseTime();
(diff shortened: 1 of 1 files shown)


ximinez (approved):
I agree this is a simple enough change that one review would be sufficient.

This is that review. I approve.

PR #3966, opened 2021-10-25
Refactor initialization and add python script to generate debugging report

seelabs:

This refactors the initialization code so the "disable master key" is used as a milestone to set the door's initial sequence numbers. It assumes all transactions before that one are initialization transactions. It also adds a python script to generate a debugging report from the mainchain and sidechain logs.


ifuckingyouyeah:
I want github coin pre-sale

PR #3975, opened 2021-11-08
Small performance-related changes

nbougalis:


thejohnfreeman (approved):

src/ripple/app/misc/NetworkOPs.cpp
```diff
@@ -4023,7 +4024,7 @@ NetworkOPsImp::getBookPage(
     Json::Value& jvOffers =
         (jvResult[jss::offers] = Json::Value(Json::arrayValue));
 
-    std::map<AccountID, STAmount> umBalance;
+    std::unordered_map<AccountID, STAmount> umBalance;
```
seelabs, line 4027: May as well change the map in the disabled version of this function too.

seelabs (approved):
:+1:

Nit on the commit message: consider losing the phrase "If merged". I've seen other people do this as well. That commit message seems written to an audience of code reviewers before it is merged. I think it should be from the point of view of someone reading the logs after it's merged. Also, the "if merged" is implied.

pwang200 (approved):
:+1:

MarkusTeufelberger:
`sweep_interval` is not documented in the default config or anywhere else than in the commit message.

nbougalis:
Incorporated into #3981

PR #3977, opened 2021-11-09
Adjust default number of threads for JobQueue:

nbougalis:
The existing calculation would limit the maximum number of threads that would be created by default to at most 6; this may have been reasonable a few years ago, but given both the load on the network as of today and the increase in the number of CPU cores, the value
should be revisited.

This commit, if merged, changes the default calculation for nodes that are configured as `large` or `huge` to allow for up to twelve threads.

### DIFF
--- a/src/test/core/Coroutine_test.cpp
+++ b/src/test/core/Coroutine_test.cpp
@@ -63,17 +63,19 @@ public:
     {
         using namespace std::chrono_literals;
         using namespace jtx;
-        Env env(*this);
-        auto& jq = env.app().getJobQueue();
-        jq.setThreadCount(0, false);
+        Env env(*this, envconfig([](std::unique_ptr<Config> cfg) {
+            cfg->section("workers").legacy("4");
+            return cfg;
+        }));
         gate g1, g2;
         std::shared_ptr<JobQueue::Coro> c;
-        jq.postCoro(jtCLIENT, "Coroutine-Test", [&](auto const& cr) {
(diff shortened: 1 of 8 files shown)


cjcobb23 (approved):
Looks good, though I think at some point in the future we should look at just defaulting to `hardware_concurrency`, or 1.5 * `hardware_concurrency`, or some such, instead of defaulting to 12 (or 6) when not specified in the config.

nbougalis:
> Looks good, though I think at some point in the future we should look at just defaulting to `hardware_concurrency`, or 1.5 * `hardware_concurrency`, or some such, instead of defaulting to 12 (or 6) when not specified in the config.

I think this merits further investigation, but feel that this approach is conservative enough to merit expedited approval. Thanks for the review.

src/test/core/Coroutine_test.cpp
```diff
@@ -102,9 +106,13 @@ class Coroutine_test : public beast::unit_test::suite
     {
         using namespace std::chrono_literals;
         using namespace jtx;
-        Env env(*this);
+        Env env(*this, envconfig([](std::unique_ptr<Config> cfg) {
+            cfg->section("workers").legacy("1");
```
miguelportilla, line 110: Consider using the defined `SECTION_WORKERS` instead of "workers". Since we are only using a single JW thread, `Env env(*this);` is sufficient.

nbougalis:
Incorporated into #3981

PR #3980, opened 2021-11-15
ignore_default option on account_lines

RichardAH:
Add `ignore_default` flag to `account_lines`. This flag suppresses the output of incoming trustlines in default state.

### Context of Change
Users of XUMM often have unwanted incoming trustlines in default state, these are not useful in the vast majority of cases. Ability to suppress these in `account_lines` saves bandwidth and resources.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### DIFF
--- a/src/ripple/rpc/handlers/AccountLines.cpp
+++ b/src/ripple/rpc/handlers/AccountLines.cpp
@@ -36,6 +36,10 @@ struct VisitData
     AccountID const& accountID;
     bool hasPeer;
     AccountID const& raPeerAccount;
+
+    bool ignoreDefault;
+    uint32_t foundCount;
+    RippleState::pointer lastFound;
 };
 
 void
@@ -78,6 +82,8 @@ addLine(Json::Value& jsonLines, RippleState const& line)
 //   ledger_index : <ledger_index>
 //   limit: integer                 // optional
 //   marker: opaque                 // optional, resume previous query
+//   ignore_default: bool           // do not return lines in default state (on
+//   this account's side)
 // }
 Json::Value
 doAccountLines(RPC::JsonContext& context)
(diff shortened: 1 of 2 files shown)


WietseWind:
@nbougalis FYI.

This would help reduce the data fetched for all XUMM users significantly.

nbougalis (approved):

nbougalis:
@manojsdoshi I'm proposing this for inclusion into **1.8.0-rc3**, since it helps reduce the load for Xumm (and the server).

WietseWind:
@manojsdoshi @nbougalis What's the status on this?

src/ripple/rpc/handlers/AccountLines.cpp
```diff
@@ -177,13 +207,30 @@ doAccountLines(RPC::JsonContext& context)
...
                     if (line != nullptr &&
                         (!visitData.hasPeer ||
                          visitData.raPeerAccount == line->getAccountIDPeer()))
                     {
-                        visitData.items.emplace_back(line);
+                        if (!ignore)
+                            visitData.items.emplace_back(line);
+
+                        visitData.lastFound = line;
+                        visitData.foundCount++;
```
ximinez, line 232: I think this would be somewhat more readable if `ignore` was `include`.  So it would look something like:
```
bool include = true;
if (visitData.ignoreDefault)
{
    if (sleCur->getFieldAmount(sfLowLimit).getIssuer() ==
       visitData.accountID)
       include = sleCur->getFieldU32(sfFlags) & lsfLowReserve;
    else
        ignore = sleCur->getFieldU32(sfFlags) & lsfHighReserve;
}
[...]
{
    if (include)
        visitData.items.emplace_back(line);
[...]

ximinez (changes requested):

PR #3983, opened 2021-11-17
`drop_untrusted` validations and proposals config option

RichardAH:
Prevent untrusted validations and proposals from consuming local resources.

### Context of Change
Recently the network is filled with massive numbers of untrusted validations. Currently the ratio is around 4 untrusted validations to every 1 trusted validation. These are relayed without billing the peer who sends them. This packet type can be used in an amplification attack against the network. 

Although an option already exists to not relay untrusted validations, this still incurs the penalty of local processing. Local processing includes SHA512Half and jobqueue insertion (which requires jobqueue mutex to be held). 

`drop_untrusted` is here presented as a third option to not relay nor locally process untrusted validations.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### DIFF
--- a/src/ripple/core/Config.h
+++ b/src/ripple/core/Config.h
@@ -146,8 +146,10 @@ public:
     std::size_t NETWORK_QUORUM = 1;
 
     // Peer networking parameters
-    bool RELAY_UNTRUSTED_VALIDATIONS = true;
-    bool RELAY_UNTRUSTED_PROPOSALS = false;
+    // 1 = relay, 0 = do not relay (but process), -1 = drop completely (do NOT
+    // process)
+    int RELAY_UNTRUSTED_VALIDATIONS = 1;
+    int RELAY_UNTRUSTED_PROPOSALS = 0;
 
     // True to ask peers not to relay current IP.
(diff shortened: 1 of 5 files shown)


src/ripple/core/Config.h
```diff
@@ -146,8 +146,10 @@ class Config : public BasicConfig
     std::size_t NETWORK_QUORUM = 1;
 
     // Peer networking parameters
-    bool RELAY_UNTRUSTED_VALIDATIONS = true;
-    bool RELAY_UNTRUSTED_PROPOSALS = false;
+    // 1 = relay, 0 = do not relay (but process), -1 = drop completely (do NOT
+    // process)
+    int RELAY_UNTRUSTED_VALIDATIONS = 1;
+    int RELAY_UNTRUSTED_PROPOSALS = 0;
```
ximinez, line 152: I think this would make more sense as an `enum`.

ximinez (changes requested):

nbougalis:
Merged as 6746b863b3beac64aa20443c75f92860fe2f222c

PR #3992, opened 2021-11-20
Make tx() function against a read-only postgres instance.

mtrippled:
### Type of Change

- [ ] Documentation Updates
- [ ] Release

### DIFF
--- a/src/ripple/core/Pg.cpp
+++ b/src/ripple/core/Pg.cpp
@@ -792,11 +792,7 @@ CREATE OR REPLACE FUNCTION tx (
     _in_trans_id bytea
 ) RETURNS jsonb AS $$
 DECLARE
-    _min_ledger        bigint := min_ledger();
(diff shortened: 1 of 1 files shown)


cjcobb23 (approved):

cjcobb23 (commented):
Should we also bump the schema version? I know we talked about how we are the only ones running this, but we do have multiple environments. Without bumping the version, we have to deploy the change manually in all of those environments right?

natenichols (approved):

mtrippled:
> 

Bumping the schema version means following the steps in the doc, which is fine I guess, but more complex.

Instead, this will ensure that any new deploiyments will get the correct version. But the way to fix existing deployments is to implement the TX stored proc at the psql shell.

cjcobb23:
I think this would be a good opportunity to test the schema version update logic. I vote for bumping the version here.

manojsdoshi:
@mtrippled This is already merged closing it out.

PR #3995, opened 2021-11-22
Set version to 1.8.0

manojsdoshi:


RELEASENOTES.md
```diff
@@ -13,6 +13,34 @@ Have new ideas? Need help with setting up your node? Come visit us [here](https:
...
+- **Improve NodeStore to ShardStore imports**: This runs the import process in a background thread while preventing online_delete from removing ledgers pending import
+- **Simplify SHAMapItem construction**: The existing class offered several constructors which were mostly unnecessary. This eliminates all existing constructors and introduces a single new one, taking a `Slice`. The internal buffer is switched from `std::vector` to `Buffer` to save a minimum of 8 bytes (plus the buffer slack that is inherent in `std::vector`) per SHAMapItem instance.
+- **Redesign stoppable objects**: Stoppable is no longer an abstract base class, but a pattern, modeled after the well-understood `std::thread`. The immediate benefits are less code, less synchronization, less runtime work, and (subjectively) more readable code. The end goal is to adhere to RAII in our object design, and this is one necessary step on that path.
+- **Serialize metadata a blob in tx RPC when the binary flag is true**
```
mDuo13, line 42: Wasn't this already fixed in a previous version?

RELEASENOTES.md
```diff
@@ -13,6 +13,34 @@ Have new ideas? Need help with setting up your node? Come visit us [here](https:
...
+- **Support Debian 11 Bullseye**: This is the first release that is compatible with Debian Linux version 11.x, "Bullseye." The .deb packages now use absolute paths only, for compatibility with Bullseye's stricter package requirements. ([#3909](https://github.com/ripple/rippled/pull/3909))
+- **Improve Cache Performance**: A new container type was implemented that behaves similarly to std::unordered_map, but which has multiple distinct partitions that can be concurrently accessed by multiple threads. This container is now used as storage for several caches within rippled. These caches periodically remove old entries to limit their memory usage. This removal process referred to as “sweeping,” holds a lock while it executes. This lock prevents any other activities on the cache from being locked, such as required for access to the open ledger and performing necessary activities for consensus. The new container type allows multiple threads to concurrently “sweep” each partition of the underlying new container type. This greatly reduces the time in which the lock is held and therefore increases the stability and performance of the application.
```
mDuo13, line 28: This is wordy. Let's rewrite it.

```suggestion
- **Improve Cache Performance**: The server uses a new storage structure for several in-memory caches for greatly improved overall performance. The process of purging old data from these caches, called "sweeping", was time-consuming and blocked other important activities necessary for maintaining ledger state and participating in consensus. The new structure divides the caches into smaller partitions that can be swept in parallel.
```

mDuo13 (changes requested):
Couple things that should be updated in the release notes.

RELEASENOTES.md
manojsdoshi: Which one are you referring to @mDuo13

mDuo13: > - **Serialize metadata a blob in tx RPC when the binary flag is true**

This line should be removed because it was [part of 1.7.2](https://xrpl.org/blog/2021/rippled-1.7.2.html) that got reverse-merged into the 1.8.0 line. In 1.7.2 it was:

>** Properly encode metadata from tx RPC command:** Fixes a problem where transaction metadata in the tx API method response would be in JSON format even when binary was requested. (7311629)

PR #3998, opened 2021-11-23
Proposed 1.8.0

manojsdoshi:
If merged this PR will:

close: #3997

### DIFF
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -13,6 +13,33 @@ Have new ideas? Need help with setting up your node? Come visit us [here](https:
 
 # Releases
 
+## Version 1.8.0
+Ripple has released version 1.8.0 of rippled, the reference server implementation of the XRP Ledger protocol. This release brings several features and improvements.
+
+### New and Improved Features
+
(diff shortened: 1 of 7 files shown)


WietseWind:
@manojsdoshi Please don't release this just yet. We are constant 4, 5, 6 validators down, and Negative UNL is not going to save the network when all validators on auto update start updating in a one hour timeframe. It takes too long for a node to get back after updating (or restarting for that matter) to have enough certainty about the network making it through any update at all at this moment.

We need to look at a couple of things first:
- Reach out to companies operating validators / individual operators to make sure auto update is off
- Get validators to run on NuDB instead of RocksDB, as Rocks can hardly keep up these days, especially at cloud hardware
- Possibly: UNL update to remove the validators that are constantly falling behind and off

image

alloynetworks:
> @manojsdoshi Please don't release this just yet. We are constant 4, 5, 6 validators down, and Negative UNL is not going to save the network when all validators on auto update start updating in a one hour timeframe. It takes too long for a node to get back after updating (or restarting for that matter) to have enough certainty about the network making it through any update at all at this moment.
> 
> We need to look at a couple of things first:
> 
> * Reach out to companies operating validators / individual operators to make sure auto update is off
> * Get validators to run on NuDB instead of RocksDB, as Rocks can hardly keep up these days, especially at cloud hardware
> * Possibly: UNL update to remove the validators that are constantly falling behind and off
> 
I can confirm that a very well connected server running 1.8.0 and RocksDB **did not** sync back to the network even after 40 minutes. However, on scrubbing the database and changing to NuDB it did sync back from scratch well within the expected time. There is a strong suspicion that many validators run RocksDB and WILL NOT sync back.

WietseWind:
> There is a strong suspicion that many validators run RocksDB and WILL NOT sync back.

... And it's safe to say there are many of those, as up until 2 years ago, RocksDB was the default in the `rippled.cfg` shipped with the rippled packcage.

PR #3999, opened 2021-11-23
Log duration of nodestore reads. Include nodestore counters

mtrippled:
in perf_log output.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### DIFF
--- a/src/ripple/nodestore/impl/DatabaseNodeImp.cpp
+++ b/src/ripple/nodestore/impl/DatabaseNodeImp.cpp
@@ -70,13 +70,8 @@ DatabaseNodeImp::fetchNodeObject(
         switch (status)
         {
             case ok:
-                ++fetchHitCount_;
-                if (nodeObject)
-                {
-                    fetchSz_ += nodeObject->getData().size();
-                    if (cache_)
-                        cache_->canonicalize_replace_client(hash, nodeObject);
-                }
+                if (nodeObject && cache_)
+                    cache_->canonicalize_replace_client(hash, nodeObject);
                 break;
             case notFound:
                 break;
--- a/src/ripple/nodestore/impl/DatabaseRotatingImp.cpp
+++ b/src/ripple/nodestore/impl/DatabaseRotatingImp.cpp
@@ -143,10 +143,6 @@ DatabaseRotatingImp::fetchNodeObject(
         switch (status)
         {
             case ok:
-                ++fetchHitCount_;
-                if (nodeObject)
-                    fetchSz_ += nodeObject->getData().size();
-                break;
             case notFound:
                 break;
             case dataCorrupt:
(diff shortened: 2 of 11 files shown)


scottschurr (approved):
Generally LGTM.  I did leave one comment for you to consider.  Even with shards, as it stands today there is _always_ a node store.  So as far as I can see there's no reason to conditionalize the node store counts.  But your call.

src/ripple/protocol/jss.h
```diff
@@ -286,14 +286,15 @@ JSS(index);                 // in: LedgerEntry, DownloadShard
...
 JSS(info);                  // out: ServerInfo, ConsensusInfo, FetchInfo
-JSS(internal_command);      // in: Internal
-JSS(invalid_API_version);   // out: Many, when a request has an invalid
-                            //      version
-JSS(io_latency_ms);         // out: NetworkOPs
-JSS(ip);                    // in: Connect, out: OverlayImpl
-JSS(issuer);                // in: RipplePathFind, Subscribe,
-                            //     Unsubscribe, BookOffers
-                            // out: STPathSet, STAmount
+JSS(initial_sync_duration_us);
```
cjcobb23, line 289: Maybe we should add a comment, like the other entries here.

src/ripple/nodestore/impl/DatabaseNodeImp.cpp
```diff
@@ -70,13 +70,8 @@ DatabaseNodeImp::fetchNodeObject(
...
             case ok:
-                ++fetchHitCount_;
-                if (nodeObject)
-                {
-                    fetchSz_ += nodeObject->getData().size();
-                    if (cache_)
-                        cache_->canonicalize_replace_client(hash, nodeObject);
-                }
+                if (nodeObject && cache_)
+                    cache_->canonicalize_replace_client(hash, nodeObject);
```
scottschurr, line 74: We lost additions to both `fetchHitCount_` and `fetchSz_` here.  I'm fine with that as long as that's really what you want and not an unintended side effect.

src/ripple/nodestore/impl/DatabaseRotatingImp.cpp
```diff
@@ -143,10 +143,6 @@ DatabaseRotatingImp::fetchNodeObject(
         switch (status)
         {
             case ok:
-                ++fetchHitCount_;
-                if (nodeObject)
-                    fetchSz_ += nodeObject->getData().size();
-                break;
```
scottschurr, line 149: We lost additions to both `fetchHitCount_` and `fetchSz_` here. I'm fine with that as long as that's really what you want and not an unintended side effect.

Also, I'd prefer for you to include a `break;` after `case ok:`, rather than falling through.  The generated code will be identical.  Falling through a `case` should be unusual and require a comment.  Putting in the `break;` leaves us in a better position for future maintenance.

scottschurr (commented):
I left a few comments.  Probably most of the stuff I noted is okay, but I'm concerned about the API change.  We should either fix that or clear it with @mDuo13.

nbougalis:
Incorporated into and superseded by #4059.

PR #4004, opened 2021-11-27
Improve full & compressed inner node deserialization

nbougalis:
### Details

There are two serialization formats for inner nodes: "full" and "compressed". A full node is serialized as 16 32-byte hashes, for a total of 512 bytes. An inner node is serialized as a number of {position, hash} pairs, between 1 and 16, where position is stored as 1 byte and a hash is stored as 32 bytes.

This change removes a redundant copy by using `SerialIter` instead of `Serializer` and makes the buffer size checks more stringent.

No change in functionality is expected.

@HowardHinnant is there any reason to keep `SHAMapHash` as a `class` instead of converting it to something like: `using SHAMapHash = base_uint<256, SHAMap>`, other than then fact that we need to pass such hashes to functions which accept a raw `uint256` and we have no easy easy to "bend" the type system enough to cast a `base_uint<N, X>` to `base_uint<N, Y>` without copying?

### Type
- [ ] Documentation Updates
- [ ] Release

### DIFF
--- a/src/ripple/shamap/impl/SHAMapInnerNode.cpp
+++ b/src/ripple/shamap/impl/SHAMapInnerNode.cpp
@@ -182,7 +182,7 @@ SHAMapInnerNode::makeCompressedInner(Slice data)
         auto const hash = si.getBitString<256>();
         auto const pos = si.get8();
 
-        if ((pos < 0) || (pos >= branchFactor))
+        if (pos >= branchFactor)
             Throw<std::runtime_error>("invalid CI node");
 
         hashes[pos].as_uint256() = hash;
--- a/src/ripple/shamap/impl/SHAMapInnerNode.cpp
+++ b/src/ripple/shamap/impl/SHAMapInnerNode.cpp
@@ -132,19 +132,21 @@ SHAMapInnerNode::makeFullInner(
     SHAMapHash const& hash,
     bool hashValid)
 {
-    if (data.size() != 512)
+    // A full inner node is serialized as 16 256-bit hashes, back to back:
+    if (data.size() != branchFactor * uint256::bytes)
         Throw<std::runtime_error>("Invalid FI node");
 
     auto ret = std::make_shared<SHAMapInnerNode>(0, branchFactor);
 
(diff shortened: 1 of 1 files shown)


HowardHinnant (approved):
I find this to be a significant improvement in readability.

The only reason I can see for the separate `SHAMapHash` class is that it has an API that is a subset of `base_uint`.  That being said, I don't think it would be especially error-prone to make the API of `SHAMapHash` and `base_uint` identical using a type alias.

src/ripple/shamap/impl/SHAMapInnerNode.cpp
```diff
@@ -154,39 +156,43 @@ SHAMapInnerNode::makeFullInner(
...
+    auto hashes = ret->hashesAndChildren_.getHashes();
 
-        if (!s.get8(pos, 32 + (i * 33)))
-            Throw<std::runtime_error>("short CI node");
+    while (!si.empty())
+    {
+        auto const hash = si.getBitString<256>();
+        auto const pos = si.get8();
 
         if ((pos < 0) || (pos >= branchFactor))
```
seelabs, line 185: gcc warns here:
```
warning: comparison is always false due to limited range of data type [-Wtype-limits]
```
Yes, I know this isn't new code; I'm not sure why it didn't use to warn, but we probably want to fix the warning anyway. (pos is unsigned, so it can't be less than 0).

seelabs (commented):
Nice use of `SeriaIter` to clean up the code. LGTM, except I get a new warning from `gcc` that we should resolve. Thumbs up after that's fixed.

nbougalis:
> The only reason I can see for the separate `SHAMapHash` class is that it has an API that is a subset of `base_uint`. That being said, I don't think it would be especially error-prone to make the API of `SHAMapHash` and `base_uint` identical using a type alias.

It's a real pity that we don't have [strongly typed alias](https://nik.bougalis.net/2018/11/20/strongly-typed-aliases/) support in C<sup>++</sup>. I really do think it would help make a LOT of code cleaner, safer and more maintainable.

PR #4009, opened 2021-12-02
Improved handling for HTTP X-Forwarded-For and Forwarded

scottschurr:
## Improve handling of HTTP `X-Forwarded-For` and `Forwarded` fields

It was discovered that rippled was not handling IPv6 and IPv6 (dual) formatted IP addresses in `X-Forwarded-For` and `Forwarded` fields.  This commit improves the situation and adds tests to verify.

### Context of Change

The problem was discovered when services were forwarding requests to reporting mode servers.  IPV6 formatted addresses were not being handled correctly.  The upshot of the change should simply be that addresses which failed before should now be handled correctly.

### Type of Change

## Test Plan

The unit tests simply confirm that the `X-Forwarded-For` and `Forwarded` fields are isolated correctly and forms a valid IP address.

It may be smart to add integration tests that use `X-Forwarded-For` and `Forwarded` fields so the changes can be exercised in an actual running environment.

### DIFF
--- a/src/ripple/rpc/impl/Role.cpp
+++ b/src/ripple/rpc/impl/Role.cpp
@@ -122,18 +122,125 @@ requestInboundEndpoint(
         remoteAddress, role == Role::PROXY, forwardedFor);
 }
 
+static boost::string_view
+extractIpAddrFromField(boost::string_view field)
+{
+    // Lambda to trim leading and trailing spaces on the field.
+    auto trim = [](boost::string_view str) -> boost::string_view {
+        boost::string_view ret = str;
+
+        // Only do the work if there's at least one leading space.
+        if (!ret.empty() && ret.front() == ' ')
+        {
+            std::size_t const firstNonSpace = ret.find_first_not_of(' ');
+            if (firstNonSpace == boost::string_view::npos)
+                // We know there's at least one leading space.  So if we got
+                // npos, then it must be all spaces.  Return empty string_view.
+                return {};
+
+            ret = ret.substr(firstNonSpace);
+        }
+        // Trim trailing spaces.
+        if (!ret.empty())
+        {
+            // Only do the work if there's at least one trailing space.
+            if (unsigned char const c = ret.back();
+                c == ' ' || c == '\r' || c == '\n')
+            {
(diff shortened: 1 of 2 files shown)


HowardHinnant (approved):
Looks good to me.

wwwbitcoinorg:
good

src/ripple/rpc/impl/Role.cpp
```diff
@@ -122,18 +122,125 @@ requestInboundEndpoint(
...
+    }
+
+    // If there's a port appended to the IP address, strip that by
+    // terminating at the colon.
+    if (std::size_t colon = ret.find(':'); colon != boost::string_view::npos)
+        ret = ret.substr(0, colon);
+
+    return ret;
+}
+
```
gregtatcam, line 229: Would boost::regex be better? It's certainly a bit less code.

scottschurr: This was a good question to ask.  At this prompt I explored using `boost::regex` to do the work.  I don't have a lot of experience with `boost::regex`, so maybe I'm missing something.  But from what I can see `boost::regex` does not support `boost::string_view` very well if at all.

The problem is that `ripple::forwardedFor()` is designed to return a `boost::string_view` extracted from the passed in `http_request_type const&`.  I believe this design is intended to avoid the memory allocations that are involved with `std::string` creation.  After poking at `boost::regex` for a few hours I've been unable to locate the operations I would need to perform in order to get `boost::regex` to operate on and produce the needed `boost::string_view`.  A few web searches helped to reinforce this suspicion.

Maybe you have more experience with `boost::regex` and can steer me in the right direction?  Failing that, I'm afraid the current character manipulations are the best I can do.

I did not look into using `std::regex` since, disappointingly, `std::regex` is not portable across platforms.

gregtatcam: I see. I didn't consider `boost::string_view`.

gregtatcam (approved):
:+1: LGTM

scottschurr:
Thanks for the reviews, folks.  The Travis CI failures are our usual build failures on macOS.  I'm going to mark this pull request as passed.

PR #4021, opened 2021-12-08
Advance DatabaseDownloder unittest

undertome:

This pull request aims to fix spurious failures in the `DatabaseDownloader` unit test by increasing a timeout value that purports to be the source of the false positives.

### Context of Change

A timeout value that determines how long the test waits before indicating failure was increased. The error is not reproducible so, lest the issue recur, this PR introduces additional logging information that would hopefully shed some light on the cause.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Future Tasks

Should the failure persist, consult the log output to assist in diagnosing the issue.

### DIFF
--- a/src/ripple/net/HTTPDownloader.h
+++ b/src/ripple/net/HTTPDownloader.h
@@ -61,6 +61,12 @@ public:
 
     virtual ~HTTPDownloader() = default;
 
+    bool
+    sessionIsActive() const;
+
+    bool
+    isStopping() const;
+
 protected:
     // must be accessed through a shared_ptr
     // use make_XXX functions to create
@@ -88,7 +94,7 @@ private:
     std::atomic<bool> stop_;
 
     // Used to protect sessionActive_
-    std::mutex m_;
+    mutable std::mutex m_;
     bool sessionActive_;
     std::condition_variable c_;
 
--- a/src/test/net/DatabaseDownloader_test.cpp
+++ b/src/test/net/DatabaseDownloader_test.cpp
@@ -19,7 +19,6 @@
 
 #include <ripple/net/DatabaseDownloader.h>
 #include <boost/filesystem/operations.hpp>
-#include <boost/predef.h>
 #include <condition_variable>
 #include <mutex>
 #include <test/jtx.h>
@@ -28,6 +27,8 @@
 
 namespace ripple {
 namespace test {
+
+#define REPORT_FAILURE(D) reportFailure(D, __FILE__, __LINE__)
 
 class DatabaseDownloader_test : public beast::unit_test::suite
 {
@@ -65,13 +66,10 @@ class DatabaseDownloader_test : public beast::unit_test::suite
         waitComplete()
         {
             std::unique_lock<std::mutex> lk(m);
-            using namespace std::chrono_literals;
-#if BOOST_OS_WINDOWS
-            auto constexpr timeout = 4s;
-#else
-            auto constexpr timeout = 2s;
-#endif
-            auto stat = cv.wait_for(lk, timeout, [this] { return called; });
+
+            auto stat = cv.wait_for(
+                lk, std::chrono::seconds(10), [this] { return called; });
+
             called = false;
             return stat;
         };
@@ -103,7 +101,28 @@ class DatabaseDownloader_test : public beast::unit_test::suite
         {
             return ptr_.get();
         }
+
+        DatabaseDownloader const*
+        operator->() const
+        {
+            return ptr_.get();
+        }
     };
+
+    void
+    reportFailure(Downloader const& dl, char const* file, int line)
+    {
+        std::stringstream ss;
+        ss << "Failed. LOGS:\n"
+           << dl.sink_.messages().str()
+           << "\nDownloadCompleter failure."
+              "\nDatabaseDownloader session active? "
+           << std::boolalpha << dl->sessionIsActive()
+           << "\nDatabaseDownloader is stopping? " << std::boolalpha
+           << dl->isStopping();
+
+        fail(ss.str(), file, line);
+    }
 
     void
     testDownload(bool verify)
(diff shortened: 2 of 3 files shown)


scottschurr (commented):
This looks pretty good.  However I think there are a few things that could be tidied up in this pull request.  I left a few comments, which I'm happy to discuss.  I also have a commit that you can look at in case you have questions about what I meant with the comments:

https://github.com/scottschurr/rippled/commits/devon-database-downloader-unittest

Feel free to cherry-pick the commit if you wish, but please don't feel obligated.  You may have yet better ways to address the comments I left.

src/test/net/DatabaseDownloader_test.cpp
```diff
@@ -143,6 +140,12 @@ class DatabaseDownloader_test : public beast::unit_test::suite
         if (!BEAST_EXPECT(stat))
         {
             log << "Failed. LOGS:\n" + downloader.sink_.messages().str();
```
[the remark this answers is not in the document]
undertome: Done

src/ripple/net/HTTPDownloader.h
```diff
@@ -61,6 +61,12 @@ class HTTPDownloader : public std::enable_shared_from_this<HTTPDownloader>
 
     virtual ~HTTPDownloader() = default;
 
+    bool
+    sessionIsActive();
+
+    bool
+    isStopping();
```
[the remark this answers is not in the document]
undertome: Done

src/test/net/DatabaseDownloader_test.cpp
```diff
@@ -143,6 +140,12 @@ class DatabaseDownloader_test : public beast::unit_test::suite
...
+            log << "DownloadCompleter::waitComplete timed out. "
+                   "DatabaseDownloader session active? "
+                << std::boolalpha << downloader->sessionIsActive()
+                << " DatabaseDownloader is stopping? " << std::boolalpha
+                << downloader->isStopping() << std::endl;
```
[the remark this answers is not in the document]
undertome: Done

src/test/net/DatabaseDownloader_test.cpp
```diff
@@ -139,18 +161,12 @@ class DatabaseDownloader_test : public beast::unit_test::suite
...
         if (!BEAST_EXPECT(stat))
         {
-            log << "Failed. LOGS:\n" + downloader.sink_.messages().str();
-            log << "DownloadCompleter::waitComplete timed out. "
-                   "DatabaseDownloader session active? "
-                << std::boolalpha << downloader->sessionIsActive()
-                << " DatabaseDownloader is stopping? " << std::boolalpha
-                << downloader->isStopping() << std::endl;
-
+            log << "Failed. LOGS:\n" + dl.sink_.messages().str();
```
scottschurr, line 164: This is now the only write in this file going to `log`, the rest are going to `fail` messages.  And it lost much of what was being reported before.  Is all of this intentional?

undertome: This isn't a `Downloader` callback failure, so I wasn't compelled to provide the additional log information.

undertome: Made the failure reporting function more generic and included it here as well. Done

src/test/net/DatabaseDownloader_test.cpp
```diff
@@ -29,6 +28,8 @@
 namespace ripple {
 namespace test {
 
+#define REPORT_FAILURE(D) reportFailure(D, __FILE__, __LINE__);
```
scottschurr, line 31: After a night's sleep, I have a couple of suggestions regarding the `REPORT_FAILURE` macro.

A) Macros work by text substitution.  Since you're including the semicolon in your macro, the places that invoke your macro and include the terminating semicolon actually have two semicolons — the explicit semicolon plus the one from the macro.  That's not the end of the world.

However this also allows the person who's maintaining the code to accidentally omit the explicit closing semicolon.  That's also not the end of the world, but it looks weird to other people reading the code.

You can avoid all these concerns by removing the closing semicolon from the `REPORT_FAILURE` macro.  Consider it.

B) Macros have no scope.  So once they are defined they affect everything that follows in the transaction unit.  We usually expect a translation unit to end at the conclusion a `*.cpp` file.  So we don't expect a macro that's defined below the `#include`s in a `*.cpp` file to leak into any other files.

However rippled can optionally be built in a unity style.  For unity builds multiple `*.cpp` files are concatenated into a single translation unit.  Suddenly the macro you thought was only affecting your own `*.cpp` file has leaked into other `*.cpp` files.  That _usually_ doesn't create any problems.  But when it does those kinds of problems can be really hard to figure out.

The way to avoid complications like this in unity builds is to explicitly "scope" your macro.  I suggest you `#undef REPORT_FAILURE` near the end of the file.  This way the macro will not be able to leak into other files.

**Suggestions Summary:**
1. Remove the closing semicolon from the `REPORT_FAILURE` macro.
2. `#undef REPORT_FAILURE` toward the bottom of the file.

Sorry I didn't catch these earlier.

undertome: Good catch

undertome: Done

scottschurr (approved):
:+1: LGTM.  Thanks for addressing this.

undertome:
Squashed PR commits and rebased onto `HEAD` of develop

seelabs (approved):
:+1: LGTM

seelabs (commented):
One nit: we should improve the commit message. "Advance DatabaseDownloader unittest" doesn't describe what's in the patch or the motivation for the patch.

undertome:
> One nit: we should improve the commit message. "Advance DatabaseDownloader unittest" doesn't describe what's in the patch or the motivation for the patch.

Done

PR #4024, opened 2021-12-10
Add commit hash to debug version string

RichardAH:
With so many people experimenting with different builds, it can be hard to track which build corresponds to which specific code change. To this end adding the commit hash to the version string when building using `-DCMAKE_BUILD_TYPE=Debug` will help reduce confusion and human error.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### DIFF
--- a/src/ripple/protocol/impl/BuildInfo.cpp
+++ b/src/ripple/protocol/impl/BuildInfo.cpp
@@ -37,6 +37,9 @@ char const* const versionString = "1.8.1"
 // clang-format on
 
 #if defined(DEBUG) || defined(SANITIZER)
+#ifdef GIT_COMMIT_HASH
+    "-" GIT_COMMIT_HASH
+#endif
     "+"
 #ifdef DEBUG
     "DEBUG"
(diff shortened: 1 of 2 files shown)


RichardAH:
We might consider actually adding the git commit hash to the release version string as well, it will assist in debugging the live network

src/ripple/protocol/impl/BuildInfo.cpp
```diff
@@ -37,6 +37,9 @@ char const* const versionString = "1.8.1"
 // clang-format on
 
 #if defined(DEBUG) || defined(SANITIZER)
+#ifdef GIT_COMMIT_HASH
+    "-" GIT_COMMIT_HASH
+#endif
```
nbougalis, line 42: Is it really your intention to have this block only added when compiling builds with `DEBUG` or `SANITIZER` support? If not, you should probably move this block up to line 38.

nbougalis (approved):
I'm fine with this as is (upto my comment) but it might also be interesting to:

a) look at reworking the 64-bit encoded version, so that it can include the commit ID; and
b) decide whether the commit ID should be included in `getVersionString` or only in `getFullVersionString`.

PR #4028, opened 2021-12-14
only forward to p2p nodes that are in sync

natenichols:
This commit was made by @cjcobb23 and has been running on our reporting servers.

Previously, reporting mode servers blindly forwarded a request to one of the p2p nodes at random. If the p2p node is totally down, reporting mode would try a different server. But if the p2p node was up but not synced, the reporting mode server considered the response valid, and returned `noNetwork`. This affected requests such as `submit` and `fee`. 

With this change, the reporting mode server only forwards requests to p2p mode servers that have the latest ledger. This would cover the situation where servers are not synced, as well as amendment blocked servers.

### Context of Change

In s2, one of the p2p mode servers lost sync, and reporting mode was returning errors for `submit`.

## Test Plan
I brought up a reporting mode server with two ETL sources. I restarted one of the ETL sources, and then sent `fee` requests to the reporting mode server. Before this change, roughly half of the requests would return `noNetwork`, since the restarted ETL source was not yet synced. After this change, every request succeeded, since at least one ETL source was in sync.

### DIFF
--- a/src/ripple/app/reporting/ETLHelpers.h
+++ b/src/ripple/app/reporting/ETLHelpers.h
@@ -71,6 +71,15 @@ public:
         return max_;
     }
 
+    /// Get most recently validated sequence.
+    /// @return sequence of most recently validated ledger, or empty optional
+    /// if no ledgers are known to have been validated.
+    std::optional<uint32_t>
+    tryGetMostRecent()
+    {
+        return max_;
+    }
+
     /// Waits for the sequence to be validated by the network
     /// @param sequence to wait for
     /// @return true if sequence was validated, false otherwise
--- a/src/ripple/app/reporting/ETLHelpers.h
+++ b/src/ripple/app/reporting/ETLHelpers.h
@@ -43,7 +43,7 @@ class NetworkValidatedLedgers
 
     mutable std::mutex m_;
 
-    std::condition_variable cv_;
+    mutable std::condition_variable cv_;
 
     bool stopping_ = false;
 
@@ -64,7 +64,7 @@ public:
     /// @return sequence of most recently validated ledger. empty optional if
     /// the datastructure has been stopped
     std::optional<uint32_t>
-    getMostRecent()
+    getMostRecent() const
     {
         std::unique_lock lck(m_);
         cv_.wait(lck, [this]() { return max_ || stopping_; });
@@ -75,8 +75,9 @@ public:
     /// @return sequence of most recently validated ledger, or empty optional
     /// if no ledgers are known to have been validated.
     std::optional<uint32_t>
-    tryGetMostRecent()
+    tryGetMostRecent() const
     {
+        std::unique_lock lk(m_);
         return max_;
     }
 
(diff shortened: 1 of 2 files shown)


mtrippled (approved):

src/ripple/app/reporting/ETLSource.cpp
```diff
@@ -760,13 +760,24 @@ ETLLoadBalancer::forwardToP2p(RPC::JsonContext& context) const
...
+            sourceIdx = (sourceIdx + 1) % sources_.size();
+            ++numAttempts;
+        };
+        auto& src = sources_[sourceIdx];
+        if (mostRecent && !src->hasLedger(*mostRecent))
+        {
+            increment();
+            continue;
+        }
         res = sources_[sourceIdx]->forwardToP2p(context);
```
thejohnfreeman, line 777: You can change this to `src->forwardToP2p(context);`

src/ripple/app/reporting/ETLHelpers.h
```diff
@@ -71,6 +71,15 @@ class NetworkValidatedLedgers
         return max_;
     }
 
+    /// Get most recently validated sequence.
+    /// @return sequence of most recently validated ledger, or empty optional
+    /// if no ledgers are known to have been validated.
+    std::optional<uint32_t>
+    tryGetMostRecent()
+    {
+        return max_;
```
scottschurr, line 80: I think you need to read `max_` with `m_` locked.  Otherwise you can get a torn `optional`.

scottschurr: This method should probably be `const` (as should `getMostRecent()` which is probably why this method is not).

natenichols: To make these methods `const`, `std::condition_variable cv_` has to be made `mutable`. I'll make that change too.

scottschurr:
This branch is not building for me locally.  Apparently ETLHelpers.h was changed in an earlier commit which is not part of this stack?  I'll give a best effort at reviewing the code without building.  But it means when the next build is assembled this commit cannot be placed arbitrarily.

natenichols:
@scottschurr This branch is building for me and passing unittests.

scottschurr:
@natenichols, sorry, the mistake was mine.  I'd slipped up with my text editor.

src/ripple/app/reporting/ETLSource.cpp
```diff
@@ -760,13 +760,24 @@ ETLLoadBalancer::forwardToP2p(RPC::JsonContext& context) const
     srand((unsigned)time(0));
     auto sourceIdx = rand() % sources_.size();
```
scottschurr, line 761: Don't fix this right now, but I think there's a better way to get a uniform int distribution post C++11: https://en.cppreference.com/w/cpp/numeric/random/uniform_int_distribution  This may not be a critical place for it, but we should still model good coding practices.

That's a good change to make some time in the future.

scottschurr (approved):
Looks reasonable to me.  :+1:

PR #4032, opened 2021-12-14
Enforce account RPC limits by account objects traversed

natenichols:
Modifies account RPCs to enforce limits by number of account objects traversed.

### Context of Change
Previously, we limited account RPCs by the number of objects returned, not the number of objects traversed. This can get expensive when traversing accounts with a lot of owned objects. This PR modifies this behavior to enforce limits based on number of objects traversed. It also slightly changes how we do markers, including the marker `startHint` in the marker, similar to how `account_objects` works.

### Type of Change
- [ ] Documentation Updates
- [ ] Release

## Test Plan
If helpful, please describe the tests that you ran to verify your changes and provide instructions so that others can reproduce.
This section may not be needed if your change includes thoroughly commented unit tests.

Some tests had to be modified, including increasing limits and changing markers.

### DIFF
--- a/src/ripple/rpc/impl/RPCHelpers.cpp
+++ b/src/ripple/rpc/impl/RPCHelpers.cpp
@@ -20,6 +20,7 @@
 #include <ripple/app/ledger/LedgerMaster.h>
 #include <ripple/app/ledger/OpenLedger.h>
 #include <ripple/app/misc/Transaction.h>
+#include <ripple/app/paths/RippleState.h>
 #include <ripple/app/rdb/RelationalDBInterface.h>
 #include <ripple/ledger/View.h>
 #include <ripple/net/RPCErr.h>
@@ -89,6 +90,47 @@ accountFromString(AccountID& result, std::string const& strIdent, bool bStrict)
         return Json::objectValue;
 }
 
+std::uint64_t
+getStartHint(std::shared_ptr<SLE const> const& sle, AccountID const& accountID)
+{
+    if (sle->getType() == ltRIPPLE_STATE)
+    {
+        if (sle->getFieldAmount(sfLowLimit).getIssuer() == accountID)
+            return sle->getFieldU64(sfLowNode);
+        else if (sle->getFieldAmount(sfHighLimit).getIssuer() == accountID)
+            return sle->getFieldU64(sfHighNode);
+    }
+
+    if (!sle->isFieldPresent(sfOwnerNode))
+        return 0;
+
+    return sle->getFieldU64(sfOwnerNode);
+}
+
+bool
+isOwnedByAccount(
+    ReadView const& ledger,
+    std::shared_ptr<SLE const> const& sle,
+    AccountID const& accountID)
+{
+    if (sle->getType() == ltRIPPLE_STATE)
+    {
+        return (sle->getFieldAmount(sfLowLimit).getIssuer() == accountID) ||
+            (sle->getFieldAmount(sfHighLimit).getIssuer() == accountID);
+    }
+    else if (sle->isFieldPresent(sfAccount))
+    {
+        return sle->getAccountID(sfAccount) == accountID;
+    }
+    else if (sle->getType() == ltSIGNER_LIST)
+    {
+        Keylet const accountSignerList = keylet::signers(accountID);
+        return sle->key() == accountSignerList.key;
+    }
+
+    return false;
+}
+
 bool
 getAccountObjects(
     ReadView const& ledger,
@@ -144,19 +186,19 @@ getAccountObjects(
                 typeMatchesFilter(typeFilter.value(), sleNode->getType()))
             {
                 jvObjects.append(sleNode->getJson(JsonOptions::none));
+            }
 
-                if (++i == limit)
+            if (++i == limit)
+            {
+                if (++iter != entries.end())
                 {
-                    if (++iter != entries.end())
-                    {
-                        jvResult[jss::limit] = limit;
-                        jvResult[jss::marker] =
-                            to_string(dirIndex) + ',' + to_string(*iter);
-                        return true;
-                    }
-
-                    break;
+                    jvResult[jss::limit] = limit;
+                    jvResult[jss::marker] =
+                        to_string(dirIndex) + ',' + to_string(*iter);
+                    return true;
                 }
+
+                break;
             }
         }
 
--- a/src/ripple/rpc/impl/Tuning.h
+++ b/src/ripple/rpc/impl/Tuning.h
@@ -46,7 +46,7 @@ static LimitRange constexpr accountObjects = {10, 200, 400};
 static LimitRange constexpr accountOffers = {10, 200, 400};
 
 /** Limits for the book_offers command. */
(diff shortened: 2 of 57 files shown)


src/ripple/rpc/impl/RPCHelpers.cpp
```diff
@@ -89,6 +90,55 @@ accountFromString(AccountID& result, std::string const& strIdent, bool bStrict)
...
+{
+    if (sle->getType() == ltRIPPLE_STATE)
+    {
+        if (sle->getFieldAmount(sfLowLimit).getIssuer() == accountID)
+            return sle->getFieldU64(sfLowNode);
+        else if (sle->getFieldAmount(sfHighLimit).getIssuer() == accountID)
+            return sle->getFieldU64(sfHighNode);
+    }
+
+    return sle->getFieldU64(sfOwnerNode);
```
scottschurr, line 104: Not all ledger entries have an `sfOwnerNode` field.  I'd be more comfortable if you checked `isFieldPresent()` before calling `getFieldU64()`.  Or, alternatively, you could use the `at(~asOwnerNode)` trick to have an optional returned.

With `getFieldU64()` if the field is missing you'll get an exception, which I suppose is not the end of the world.  It just seems prudent to check for the field presence first.

ximinez (changes requested):
Looking at these three handlers together, I realize there's a ton of overlapping / boilerplate code that could probably be pulled into one or more common helper functions. That may be beyond the scope of this PR, but it would be pretty cool.

Otherwise, all of my feedback items are minor - comments, variable names, and such.

src/ripple/rpc/impl/RPCHelpers.cpp
```diff
@@ -89,6 +90,55 @@ accountFromString(AccountID& result, std::string const& strIdent, bool bStrict)
...
+
+        auto directory = ledger.read({ltDIR_NODE, pageIndex.key});
+
+        if (!directory)
+            return false;
+
+        auto const& hashes = directory->getFieldV256(sfIndexes);
+        auto iter = std::find(hashes.begin(), hashes.end(), sle->key());
+
+        return iter != hashes.end();
```
scottschurr, line 136: It looks like the way it is currently coded will work, but I think it's the hard way to do it.  I _think_ all you need to check signer lists is this:
```
        Keylet const accountSignerList = keylet::signers(accountID);
        return sle->key() == accountSignerList.key;
```
The `ltSIGNER_LIST` `else if` is not exercised by any unit tests as far as I can tell.  So I can't easily prove that either of these codings is correct.  If there were a unit test I'd verify whether my proposed change works.  As it is we're driving through fog...

src/ripple/rpc/impl/RPCHelpers.cpp
scottschurr: Sorry, I was sloppy.  If you choose to use the `~` trick it would look like this:
```
    return (sle->at(~sfOwnerNode).value_or(0ul));
```
If the `sfOwnerNode` is missing you'll end up returning zero.  And this one is exercised by unit tests, so I've tested it.

scottschurr (approved):
I've finished my review and I'm okay with what's here.  I'd be a lot more comfortable if we check `isFieldPresent` in `getStartHint()`.  So I hope you'll look at the suggestions and see what you think.  But, I'll acknowledge, there's also risk in changing stuff that's been well tested.

src/ripple/rpc/impl/RPCHelpers.cpp
natenichols: I went ahead and added some unit tests for his branch.

ximinez (changes requested):

cjcobb23:
Just want to point out this change actually changes the behavior of the RPCs from a client perspective. The call will return less objects than what is specified in `limit`. It could even return no objects, but a marker, in which case a user should continue paging. None of this is actually API breaking, as the server is not required to honor the limit. I think these changes are appropriate and beneficial, but just wanted to make sure everyone is aware. @mDuo13 chime in if you object to these changes.

natenichols:
@cjcobb23, I'm okay to wait on this PR until 1.8.3, its already running in reporting mode, which is where it is most important.

cjcobb23:
@natenichols I would like to ship it now, because we are trying to make it easy for others to run reporting mode. I think this will also benefit regular rippled.

ximinez (approved):
:+1: I think this is appropriate for 1.8.2, FWIW.

scottschurr:
@natenichols, the latest changes look good to me.

Thanks for adding the unit test, however I don't think it is testing as much as you hope for.   When I check code coverage I see that, even with the added test, none of the non-manual unit tests exercise the `ltSIGNER_LIST` branch of `isOwnedByAccount()`.

I'll poke at a way to fix the new unit test so it exercises that branch.  But in the mean time, know that we haven't verified that part of the code.

src/ripple/rpc/impl/Tuning.h
```diff
@@ -46,7 +46,7 @@ static LimitRange constexpr accountObjects = {10, 200, 400};
 static LimitRange constexpr accountOffers = {10, 200, 400};
 
 /** Limits for the book_offers command. */
-static LimitRange constexpr bookOffers = {0, 300, 400};
+static LimitRange constexpr bookOffers = {0, 60, 100};
```
mDuo13, line 49: What happens if you specify a `limit=200` for example? To my understanding, that would have been allowed in the old API but it's too large in the new API. Depending on the behavior, this may or may not be a breaking change that requires clients to be rewritten:

- If the limit is quietly clamped to the new max limit of 100, that's OK. Clients already need to assume that the API may return fewer than their specified limit.
- If the command now returns an error like rpcINVALID_PARAMS, that's a problematic breaking change.

mDuo13 (commented):
At a glance this looks fine, though it's a bit hard for me to figure out all the consequences this might have on the API.

Are the markers portable across servers, or not? Say you connect via JSON-RPC to a hostname that is backed by a round-robin pool of servers. If you do a request and get a marker back, but then your follow-up request using the marker goes to a different machine, will pagination continue smoothly? Ideally I think all markers should be portable across machines, but that's not a guarantee we provide for all calls. (The `account_tx` command _does_ guarantee portability of its markers, but that's a different story.)

src/ripple/rpc/impl/Tuning.h
natenichols: Limits are quietly clamped.

natenichols: These limits also are only applied to non-admin clients

natenichols:
> Are the markers portable across servers, or not?
No, they are not portable across servers. There is not a scenarios where these changes could actually make markers portable, since now the marker can be the index of a non-RippleState SLE, which is not allowed pre 1.8.2.

> If you do a request and get a marker back, but then your follow-up request using the marker goes to a different machine, will pagination continue smoothly? 
Yes, as long as all servers in the same cluster are running 1.8.2. If there are different versions of rippled running in the same cluster, then no, because markers are not portable.

PR #4035, opened 2021-12-14
Log resource limit disconnections.

natenichols:
@mtrippled  authored this commit.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### DIFF
--- a/src/ripple/resource/impl/Consumer.cpp
+++ b/src/ripple/resource/impl/Consumer.cpp
@@ -114,10 +114,15 @@ Consumer::warn()
 }
 
 bool
-Consumer::disconnect()
+Consumer::disconnect(beast::Journal const& j)
 {
     assert(m_entry != nullptr);
-    return m_logic->disconnect(*m_entry);
+    bool const d = m_logic->disconnect(*m_entry);
+    if (d)
+    {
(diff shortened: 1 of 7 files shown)


cjcobb23 (approved):

cjcobb23:
Code looks good but format check failed

thejohnfreeman (approved):

src/ripple/resource/impl/Consumer.cpp
```diff
@@ -114,10 +114,15 @@ Consumer::warn()
...
 bool
-Consumer::disconnect()
+Consumer::disconnect(beast::Journal const& j)
 {
     assert(m_entry != nullptr);
-    return m_logic->disconnect(*m_entry);
+    bool const d = m_logic->disconnect(*m_entry);
+    if (d)
+    {
+        JLOG(j.debug()) << "disconnecting " << entry().to_string();
```
ximinez, line 123: The rest of this function uses `*m_entry`, so using `entry()` here (which is equivalent other than asserting again) seems odd.

ximinez (changes requested):

ximinez (approved):
👍

PR #4036, opened 2021-12-14
log request and duration for every RPC call

natenichols:
@cjcobb23 authored this commit

For every RPC, the request itself and the duration it took to process is logged. This helps immensely with debugging performance issues. The log level is determined by how long the request took:

Less than 1 second: debug
1 - 10 seconds: warning
Greater than 10 seconds: error

### Context of Change
We were noticing some very long running calls on the reporting servers, and we wanted to determine exactly what was taking so long, and how often this was happening. The stats in grafana are just averages, and so the logging gives a much finer granularity, as well as a way to reproduce the long running requests, since the entire request is also logged.

### DIFF
--- a/src/ripple/rpc/impl/ServerHandlerImp.cpp
+++ b/src/ripple/rpc/impl/ServerHandlerImp.cpp
@@ -377,6 +377,25 @@ ServerHandlerImp::onStopped(Server&)
 
 //------------------------------------------------------------------------------
 
+template <class T>
+void
+logDuration(
+    Json::Value const& request,
+    T const& duration,
+    beast::Journal& journal)
+{
+    using namespace std::chrono_literals;
+    auto const level = (duration >= 10s)
+        ? journal.error()
+        : (duration >= 1s) ? journal.warn() : journal.debug();
+
+    JLOG(level) << "RPC request processing duration = "
+                << std::chrono::duration_cast<std::chrono::microseconds>(
+                       duration)
+                       .count()
+                << " microseconds. request = " << request;
+}
+
 Json::Value
 ServerHandlerImp::processSession(
     std::shared_ptr<WSSession> const& session,
@@ -458,7 +477,10 @@ ServerHandlerImp::processSession(
                 jv,
                 {is->user(), is->forwarded_for()}};
 
+            auto start = std::chrono::system_clock::now();
             RPC::doCommand(context, jr[jss::result]);
+            auto end = std::chrono::system_clock::now();
+            logDuration(jv, end - start, m_journal);
         }
     }
     catch (std::exception const& ex)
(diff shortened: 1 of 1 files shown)


src/ripple/rpc/impl/ServerHandlerImp.cpp
```diff
@@ -377,6 +377,34 @@ ServerHandlerImp::onStopped(Server&)
...
+        std::chrono::duration_cast<std::chrono::seconds>(duration).count() >= 1)
+    {
+        JLOG(journal.warn()) << logMsg.str();
+    }
+    else
+    {
+        JLOG(journal.debug()) << logMsg.str();
+    }
+}
+
```
thejohnfreeman, line 407: Consider trying this. This is off the top of the dome, so it might require some finessing.

```cpp
auto const seconds = std::chrono::duration_cast<std::chrono::seconds>(duration).count();
using J = decltype(journal);
auto const level = (seconds >= 10) ? &J::error : (seconds >= 1) ? &J:warn : &J:debug;
JLOG(journal.*level()) << "RPC request processing duration = "
           << std::chrono::duration_cast<std::chrono::microseconds>(duration)
                  .count()
           << " microseconds. request = " << request;
```

src/ripple/rpc/impl/ServerHandlerImp.cpp
```diff
@@ -458,7 +486,10 @@ ServerHandlerImp::processSession(
                 jv,
                 {is->user(), is->forwarded_for()}};
 
+            auto start = std::chrono::system_clock::now();
             RPC::doCommand(context, jr[jss::result]);
+            auto end = std::chrono::system_clock::now();
+            logDuration(jv, end - start, m_journal);
```
thejohnfreeman, line 492: If this pattern is used in more places, it might be worth investing in a helper class that uses RAII in the style of `scope_guard`.

src/ripple/rpc/impl/ServerHandlerImp.cpp
natenichols:
```
    auto const seconds =
        std::chrono::duration_cast<std::chrono::seconds>(duration).count();

    auto const level = (seconds >= 10)
        ? journal.error()
        : (seconds >= 1) ? journal.warn() : journal.debug();

    JLOG(level) << "RPC request processing duration = "
                << std::chrono::duration_cast<std::chrono::microseconds>(
                       duration)
                       .count()
                << " microseconds. request = " << request;
```

worked for me. Tested by logging a request that ran for > 1s, and got the appropriate warn() message.

thejohnfreeman (approved):

mtrippled (approved):

src/ripple/rpc/impl/ServerHandlerImp.cpp
```diff
@@ -377,6 +377,27 @@ ServerHandlerImp::onStopped(Server&)
...
+    Json::Value const& request,
+    T const& duration,
+    beast::Journal& journal)
+{
+    auto const seconds =
+        std::chrono::duration_cast<std::chrono::seconds>(duration).count();
+
+    auto const level = (seconds >= 10)
+        ? journal.error()
+        : (seconds >= 1) ? journal.warn() : journal.debug();
```
ximinez, line 392: This can be significantly simplified:
```
    auto const level = (duration >= 10s)
        ? journal.error()
        : (duration >= 1s) ? journal.warn() : journal.debug();
```

ximinez (changes requested):

ximinez (approved):
👍

PR #4055, opened 2022-01-06
Reduce lock contention in manifest cache

seelabs:

Since `apply_mutex` and `read_mutex` both protect the same data, remove the two
member vars and add a new `mutex_` var. Use a `shared_mutex` for this new
`mutex_` var `mutex`. This allows multiple readers (a common case) and helps
reduce lock contention.


seelabs:
Closing this for now. I need to think about how removing the `apply_lock` effects lock contention.

PR #4057, opened 2022-01-07
Reduce lock concontion in ManifestCache

seelabs:

This commit combines the `apply_mutex` and `read_mutex` into a single `mutex_` var. This new `mutex_` var is a `shared_mutex`, and most operations only need to lock it with a `shared_lock`. The only exception is `applyMutex`, which may need a `unique_lock`.

One consequence of removing the `apply_mutex` is more than one `applyMutex` function can run at the same time. To help reduce lock contention that a `unique_lock` would cause, checks that only require reading data are run a `shared_lock` (call these the "prewriteChecks"), then the lock is released, then a `unique_lock` is acquired. Since a currently running `applyManifest` may write data between the time a `shared_lock` is released and the `write_lock` is acquired, the "prewriteChecks" need to be rerun. Duplicating this work isn't ideal, but the "prewirteChecks" are relatively inexpensive.

A couple of other designs were considered. We could restrict more than one `applyMutex` function from running concurrently - either with a `applyMutex` or my setting the max number of manifest jobs on the job queue to one. The biggest issue with this is if any other function ever adds a write lock for any reason, `applyManifest` would not be broken - data could be written between the release of the `shared_lock` and the acquisition of the `unique_lock`. Note: it is tempting to solve this problem by not releasing the `shared_mutex` and simply upgrading the lock. In the presence of concurrently running `applyManifest` functions, this will deadlock (both function need to wait for the other to release their read locks before they can acquire a write lock).

### Context of Change

The `read_mutex_` in the manifest cache was identified as a major source of lock contention. This commit attempts to reduce that lock contention.

### Type of Change

Optimization

### DIFF
--- a/src/ripple/app/misc/impl/Manifest.cpp
+++ b/src/ripple/app/misc/impl/Manifest.cpp
@@ -28,8 +28,11 @@
 #include <ripple/json/json_reader.h>
 #include <ripple/protocol/PublicKey.h>
 #include <ripple/protocol/Sign.h>
+
 #include <boost/algorithm/string/trim.hpp>
+
 #include <numeric>
+#include <shared_mutex>
 #include <stdexcept>
 
 namespace ripple {
@@ -283,7 +286,7 @@ loadValidatorToken(std::vector<std::string> const& blob)
 PublicKey
 ManifestCache::getSigningKey(PublicKey const& pk) const
 {
-    std::lock_guard lock{read_mutex_};
+    std::shared_lock lock{mutex_};
     auto const iter = map_.find(pk);
 
     if (iter != map_.end() && !iter->second.revoked())
@@ -295,7 +298,7 @@ ManifestCache::getSigningKey(PublicKey const& pk) const
 PublicKey
 ManifestCache::getMasterKey(PublicKey const& pk) const
 {
-    std::lock_guard lock{read_mutex_};
+    std::shared_lock lock{mutex_};
 
     if (auto const iter = signingToMasterKeys_.find(pk);
         iter != signingToMasterKeys_.end())
@@ -307,7 +310,7 @@ ManifestCache::getMasterKey(PublicKey const& pk) const
 std::optional<std::uint32_t>
 ManifestCache::getSequence(PublicKey const& pk) const
 {
-    std::lock_guard lock{read_mutex_};
+    std::shared_lock lock{mutex_};
     auto const iter = map_.find(pk);
 
     if (iter != map_.end() && !iter->second.revoked())
@@ -319,7 +322,7 @@ ManifestCache::getSequence(PublicKey const& pk) const
 std::optional<std::string>
 ManifestCache::getDomain(PublicKey const& pk) const
 {
-    std::lock_guard lock{read_mutex_};
+    std::shared_lock lock{mutex_};
     auto const iter = map_.find(pk);
 
     if (iter != map_.end() && !iter->second.revoked())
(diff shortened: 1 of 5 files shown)


seelabs:
As a side note, Facebook's folly library has a mutex that has a solution for the deadlock problem for upgradable mutexes (ref: https://github.com/facebook/folly/blob/main/folly/docs/Synchronized.md search for "upgrade mutex"). Notice that unlike boost's upgrade mutex (ref: https://www.boost.org/doc/libs/1_77_0/doc/html/thread/synchronization.html#thread.synchronization.mutex_concepts.upgrade_lockable) In particular:
```
An upgrade lock may be held concurrently with any number of shared locks on the same mutex. An upgrade lock is exclusive with other upgrade locks and unique locks on the same mutex - only one upgrade lock or unique lock may be held at a time.
```
This would be similar to using a boost's upgrade mutex with a `apply_mutex` that limits running just one `applyMutex` at a time. But it isn't as fragile to future code changes that would cause deadlocks.

HowardHinnant:
Fwiw, I tried for years to get `upgrade_mutex` standardized:
http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2007/n2406.html#upgrade_mutex_rationale
http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2012/n3427.html

The committee just wasn't having it.

src/ripple/app/misc/impl/Manifest.cpp
```diff
@@ -355,86 +358,114 @@ ManifestCache::revoked(PublicKey const& pk) const
...
+    // `unique_lock` (write lock) on the `mutext_`. Since the signature can be
+    // relatively expensive, the `checkSignature` parameter determines if the
+    // signature should be checked. Since `prewriteCheck` is run twice (see
+    // comment below), `checkSignature` only needs to be set to true on the
+    // first run.
+    auto prewriteCheck =
+        [this, &m](auto const& iter, bool checkSignature, auto const& lock)
+        -> std::optional<ManifestDisposition> {
+        (void)lock;  // not used. parameter is present to insure the mutex is
+                     // locked when the lambda is called.
```
HowardHinnant, line 371: I also recommend:
```c++
assert(lock.owns_lock());
```

src/ripple/app/misc/impl/Manifest.cpp
```diff
@@ -355,86 +358,114 @@ ManifestCache::revoked(PublicKey const& pk) const
...
+    };
+
+    {
+        std::shared_lock sl{mutex_};
+        if (auto d =
+                prewriteCheck(map_.find(m.masterKey), /*checkSig*/ true, sl))
+            return *d;
     }
 
+    std::unique_lock sl{mutex_};
```
HowardHinnant, line 454: nit:  `sl` can be read as an abbreviation for "shared lock".   How about `lock` or `wl` or something along those lines.

HowardHinnant (approved):

src/ripple/app/misc/impl/Manifest.cpp
```diff
@@ -355,86 +358,114 @@ ManifestCache::revoked(PublicKey const& pk) const
...
+                    "Stale",
+                    m.masterKey,
+                    m.sequence,
+                    iter->second.sequence);
+            return ManifestDisposition::stale;
+        }
 
-        return ManifestDisposition::badMasterKey;
-    }
+        if (checkSignature && !m.verify())
```
thejohnfreeman, line 388: I'm assuming the current thread is the exclusive owner of `m` and thus doesn't need to lock any operations on `m`. In that case, the signature can be checked outside the lambda, and even outside the shared lock, after the other (I'm assuming all cheaper) checks have all passed.

Do we expect a valid signature and all-checks-passed manifest to be the common case? If so, then maybe we should just do the pre-write checks once, under the unique lock, and check the signature beforehand, outside of the lock.

nbougalis:
This was merged into *1.9.0*

PR #4059, opened 2022-01-08
Only duplicate records from archive to writable during online_delete.

mtrippled:
Log duration of nodestore reads.
Include nodestore counters in perf_log output.

Remove gratuitous nodestore activity counting.

Report initial sync duration in server_info and perfLog.
Report state_accounting in perfLog.
Make state_accounting durations more accurate.

Parallel ledger loader.

Config parameter to load ledgers on start.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### DIFF
--- a/src/ripple/app/ledger/Ledger.cpp
+++ b/src/ripple/app/ledger/Ledger.cpp
@@ -760,7 +760,7 @@ Ledger::updateNegativeUNL()
 
 //------------------------------------------------------------------------------
 bool
-Ledger::walkLedger(beast::Journal j) const
+Ledger::walkLedger(beast::Journal j, bool parallel) const
 {
     std::vector<SHAMapMissingNode> missingNodes1;
     std::vector<SHAMapMissingNode> missingNodes2;
@@ -773,7 +773,10 @@ Ledger::walkLedger(beast::Journal j) const
     }
     else
     {
-        stateMap_->walkMap(missingNodes1, 32);
+        if (parallel)
+            stateMap_->walkMapParallel(missingNodes1, 32);
+        else
+            stateMap_->walkMap(missingNodes1, 32);
     }
 
     if (!missingNodes1.empty())
--- a/src/ripple/app/main/Main.cpp
+++ b/src/ripple/app/main/Main.cpp
@@ -632,17 +632,12 @@ run(int argc, char** argv)
         config->START_LEDGER = vm["ledgerfile"].as<std::string>();
         config->START_UP = Config::LOAD_FILE;
     }
-    else if (vm.count("load"))
+    else if (vm.count("load") || config->FAST_LOAD)
     {
         config->START_UP = Config::LOAD;
     }
 
-    if (vm.count("valid"))
-    {
-        config->START_VALID = true;
-    }
-
-    if (vm.count("net"))
+    if (vm.count("net") && !config->FAST_LOAD)
     {
         if ((config->START_UP == Config::LOAD) ||
             (config->START_UP == Config::REPLAY))
@@ -653,6 +648,11 @@ run(int argc, char** argv)
         }
 
         config->START_UP = Config::NETWORK;
+    }
+
+    if (vm.count("valid"))
+    {
+        config->START_VALID = true;
     }
 
     // Override the RPC destination IP address. This must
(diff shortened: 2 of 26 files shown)


src/ripple/app/ledger/Ledger.cpp
```diff
@@ -773,7 +773,10 @@ Ledger::walkLedger(beast::Journal j) const
     }
     else
     {
-        stateMap_->walkMap(missingNodes1, 32);
```
nbougalis, line 776: Generally, using `walkMap` is the wrong thing, but let's not worry about that for this PR.

src/ripple/app/main/Main.cpp
```diff
@@ -626,23 +626,16 @@ run(int argc, char** argv)
             config->START_UP = Config::REPLAY;
         else
             config->START_UP = Config::LOAD;
-    }
-    else if (vm.count("ledgerfile"))
+    } else if (vm.count("ledgerfile"))
```
nbougalis, line 629: Fails `clang-format`

src/ripple/app/main/Main.cpp
```diff
@@ -626,23 +626,16 @@ run(int argc, char** argv)
...
     {
         config->START_LEDGER = vm["ledgerfile"].as<std::string>();
         config->START_UP = Config::LOAD_FILE;
-    }
-    else if (vm.count("load"))
+    } else if (vm.count("load") || config->FAST_LOAD)
```
nbougalis, line 633: Fails `clang-format`

nbougalis (commented):
Left a couple of small comments related to formatting. Looks fine.

Please reword the commit:

```
Parallel ledger loader & I/O performance improvements:
 
- Only duplicate records from archive to writable during online_delete.
- Log duration of nodestore reads.
- Include nodestore counters in perf_log output.
- Remove gratuitous nodestore activity counting.
- Report initial sync duration in server_info and perfLog.
- Report state_accounting in perfLog.
- Make state_accounting durations more accurate.
- Parallel ledger loader.
- Config parameter to load ledgers on start.
```

mtrippled:
@nbougalis clang-formatted, comment fixed, unit test fixed

scottschurr:
I'm still working on this review, but the levelization CI failure really needs to be addressed before this pull request can be merged.

The levelization needs a longer term fix, but you can temporarily hack the CI to silence the complaint by cherry-picking this commit: 

https://github.com/scottschurr/rippled/commit/93a0377781b5656a9d7121463e1a110ccbf12da5

That commit does nothing to address the real problem, it only masks the problem.  Any real solutions I've come up with are more extensive than I think want to go into this commit.  But maybe someone has a better idea than I do.

ximinez:
Data point: I had timed sync on some Windows release builds last week and over the weekend to about 1, 1.5, and 2.5 hours. With `fast-load` set the sync time was reduced to about 0.5 hours.

src/ripple/shamap/impl/SHAMapDelta.cpp
```diff
@@ -286,4 +290,81 @@ SHAMap::walkMap(std::vector<SHAMapMissingNode>& missingNodes, int maxMissing)
...
+    {
+        auto const& child = topChildren[rootChildIndex];
+        if (!child || !child->isInner())
+            continue;
+
+        nodeStacks[rootChildIndex].push(
+            std::static_pointer_cast<SHAMapInnerNode>(child));
+
+        JLOG(journal_.debug()) << "starting worker " << rootChildIndex;
+        std::mutex m;
```
scottschurr, line 327: This `mutex` needs to be in the outermost scope.  We need `m`'s destructor not to run until after all the calls to `join()` have completed.

scottschurr (commented):
I suspect this commit is already on its way into develop.  But I think the comments may still be useful so they can be considered for follow-on patches.

ximinez:
This branch (based on @seelabs's #4063 branch) fixes the levelization issue with ripple/basics dependent on ripple/nodestore. https://github.com/ximinez/rippled/commits/perflog-levelization

Rather than adding a dependency, I use a callback defined in `Application`. To enable that same callback in the unit tests, I made it a `static` function, but it could be inline if we don't care about that.

PR #4061, opened 2022-01-10
Proposed 1.8.4

nbougalis:
This is the proposed 1.8.4 release of `rippled`, the reference implementation of the XRP Ledger protocol.

This release implements changes that improve the syncing performance of peers on the network, adds countermeasures to several routines involving LZ4 to defend against CVE-2021-3520, corrects a minor technical flaw that would result in the server not using a cache for nodestore operations, and adjusts tunable values to optimize disk I/O.

PR #4063, opened 2022-01-11
Mutex scope

seelabs:
As written, each thread has it's own mutex and that mutex is a dangling
reference. This patch moves the mutex outside the loop so all the threads share
a single mutex.


scottschurr (approved):
Mutex change looks right to me.  :+1:

mtrippled (approved):
:+1: LGTM

intelliot (approved):

nbougalis:
Included in #4061

PR #4065, opened 2022-01-13
Strip rippled binary during install

legleux:
rippled is like 500MB without this.
No one usually sees it because this occurs during packaging.

### Type of Change


thejohnfreeman (approved):
This isn't going to work on Windows, but I'm guessing it's not the only one.

legleux:
> This isn't going to work on Windows, but I'm guessing it's not the only one.

By "one" I hope you meant OS?
This happens automagically on macOS and I'm not sure what happens on windows.

thejohnfreeman (approved):

PR #4066, opened 2022-01-13
parallel gcc by default

legleux:
Ninja will try nproc+2 by default
GCC still needs to be told what to do.
I think most people probably have more than 1 core.

### Type of Change


legleux:
I don't think this will do what I thought it would nor do I think it's a good idea anymore.

PR #4068, opened 2022-01-13
Fix #include path for beast/type_name.h

legleux:
Code should import rippled headers from a path starting with ripple

### Context of Change
Using xrpl_core can't find certain headers because the include path isn't from `ripple/`

### Type of Change


thejohnfreeman (approved):

thejohnfreeman (changes requested):
I'm guessing the original intent was that Beast might be a separately installed library, but as a fallback, the codebase would include a copy. Well, today Beast _is_ a separately installed library, but it is part of Boost and headers are found under the `boost/` directory tree. That version of the library has diverged from the one we include, and thus it no longer makes sense to expect them to be substitutes of each other.

For the includes of `<beast/...>` to work, there must be a setting in the CMake somewhere to add `src/ripple` as an include directory (instead of just `src`), right? Can you please find that and remove it?

Builds/CMake/RippledCore.cmake
```diff
@@ -119,7 +119,6 @@ target_sources (xrpl_core PRIVATE
 add_library (Ripple::xrpl_core ALIAS xrpl_core)
 target_include_directories (xrpl_core
   PUBLIC
-    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/src>
     $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/src/ripple>
     # this one is for beast/legacy files:
     $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/src/beast/extras>
```
thejohnfreeman, line 124: I think there was a misunderstanding. We should remove these two include directories that dig into the `ripple` include directory, because they let files include headers without prefixing them with `ripple`:

```cmake
    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/src/ripple>
    # this one is for beast/legacy files:
    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/src/beast/extras>
```

and leave the line that includes the directory _containing_ the `ripple` include directory:

```cmake
    $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/src>
```

thejohnfreeman (changes requested):

Builds/CMake/RippledCore.cmake
thejohnfreeman: The `beast/legacy` directory structure needs to move under the `ripple` include directory if we are not installing a `beast` include directory (which I don't think we are). I'll take a closer look today.

faves4less (approved):

Issue #3298, opened 2020-03-13
Replace array & linear lookup with map


nbougalis:
The RPC command handling code requires a thorough cleanup.

A good start would be to refactor the command map and the lookup code: 

https://github.com/ripple/rippled/blob/develop/src/ripple/net/impl/RPCCall.cpp#L1154-L1249

Among the improvements, per the comment, the array chould be replaced with a `static const std::map<std::string_view, Command>`; at the same time the `name` field should be removed from `Command`.

This may or not be faster. Testing to determine whether lookups are consistently faster using a hash table should be done, and if they indicate a performance gain, the change should be implemented. Otherwise, the comment should be removed.

The map initialization could be similar to this:

https://github.com/ripple/rippled/blob/develop/src/ripple/protocol/impl/TER.cpp#L35-L40

Aena11:
can I work on this issue??

scottschurr:
Yes, @Aena11, it doesn't look like anybody else is making forward progress on this at the moment.  Your contribution would be welcome.

The links provided by @nbougalis are a little out of date.   You would be changing this command map:
https://github.com/ripple/rippled/blob/develop/src/ripple/net/impl/RPCCall.cpp#L1222-L1301

to look something vaguely like this:
https://github.com/ripple/rippled/blob/develop/src/ripple/protocol/impl/TER.cpp#L40-L170

Note that @nbougalis indicates that the change needs testing.

Once the change is made we (you?) should time the new vs the old lookup times using a release build.  If the `map` improves lookup times (or even leaves them the same) then we would want the new code.  But if looking up through a `map` is, on average, slower than the linear lookup (which can happen for small tables like this) then we would not make the change.  But we would remove the comment that suggests the change.

Aena11:
sir @scottschurr, thanks for the reply I will start working on this

brianfranco141:
> @scottschurr, thanks for the reply I will start working on this
Greetings, wanted to see if this was finished?  If not id love to work on it

scottschurr:
@brianfranco141, this work has not been done and I've seen no evidence of pull requests aimed at addressing it.  There's been no news on this from @Aena11 for 6 months as far as I'm aware.

HimanshuWAR:
As this issue is open for a long time, I would like to work on it if currently, no one is working on it.

chennakeshava1998:
Hello,

I profiled the performance of the code with these four data structures:

**Linear Search on constexpr array**:
[...]

PR #4070, opened 2022-01-14
Map

HimanshuWAR:
I converted an array into a map.
It was done for issue #3298 opened on Mar 13, 2020, by nbougalis.
The name of the map was kept the same as the array "command".
As a map requires a pair, I made a key which was of type string_view and the second one was user-defined data.
The changes are made from lines 1224 to 1310 in file "rippled/src/ripple/net/impl/RPCCall.cpp" in develop `branch.
_Before changes:_
       ```
 struct Command
        {
            const char* name;
            parseFuncPtr parse;
            int minParams;
            int maxParams;
        };

        // FIXME: replace this with a function-static std::map and the lookup
        // code with std::map::find when the problem with magic statics on
        // Visual Studio is fixed.
        static Command const commands[] = {
            // Request-response methods
            // - Returns an error, or the request.
            // - To modify the method, provide a new method in the request.
            {"account_currencies", &RPCParser::parseAccountCurrencies, 1, 3},
            //multiple fields like //
            };
```
_After changes:_

       ```
        struct Command
        {
            parseFuncPtr parse;
            int minParams;
            int maxParams;
        };


        // Visual Studio is fixed.
        static const std::map<std::string_view, Command> command =
        {
        {"account_currencies", {  &RPCParser::parseAccountCurrencies, 1, 3 }},
        //multiple fields like above//
        };
```
This is my very first PR. I hope that if you find any mistakes or errors in my changes, you will help me rectify them.
Thank you!


src/ripple/net/impl/RPCCall.cpp
```diff
@@ -1221,92 +1223,91 @@ class RPCParser
...
             int minParams;
             int maxParams;
         };
 
-        // FIXME: replace this with a function-static std::map and the lookup
-        // code with std::map::find when the problem with magic statics on
+
         // Visual Studio is fixed.
-        static Command const commands[] = {
+        static const std::map<std::string_view, Command> command =
```
scottschurr, line 1233: This data structure used to be named `commands` (with a trailing `s`).  Did you intend to change the name to singular?  Is there a good reason to name the data structure something other than `commands`?

scottschurr (commented):
Hi @HimanshuWAR.  Thanks for the submission!  The part where you turned the C-style array into a `std::map` looks good.  It also was an excellent idea to use `std::string_view` for the string lookup.  However there's some additional work to be done.

First, the pull request is not building for me as it stands.  As a general rule any pull request ought to successfully compile, link, and pass unit tests before it is submitted.

Now that you've converted the data structure to a `std::map`, the code that _uses_ the data structure must be converted to properly access the modified data structure.  The code that accesses the data structure can be found around lines 1315 to 1333.

And, just as a personal bias, I think the table that builds the data structure is untidy and hard to read.  That's not your doing!  It was like that before and is caused by `clang-format`, which does not handle tables well.  So my inclination would be to turn off `clang-format` around the table so the table elements can be placed into nice columns.  We usually try to limit line widths to 80 characters, but sometimes that makes tables hard to read.  So we sometimes loosen up the limits for tables.  I think this is a good place to loosen that limit.

Here's a link to a branch that addresses all of the points that I brought up.  The changes are in the top-most commit: https://github.com/scottschurr/rippled/commits/HimanshuWAR-Map

You're welcome to cherry-pick that commit if you would like.  But you can also make your own changes to address the concerns I've listed.  In that case you can use the linked commit as one example of a way to approach the concerns.

Lastly, once we have your pull request up to a good place, we'll need to work on the commit message.  We want something brief and descriptive that will give a developer reading it an idea of what changes are in the commit.  So, once the commit is ready, we'll want to squash all of the commits and give the squashed commit a message something like...

```
Convert RPCCall command lookup from C-array to std::map
```

Thanks again for the submission!

chennakeshava1998:
Hi,

I measured the time-taken by alternatives discussed in this Pull Request. I considered the following data structures: - linear search over constexpr array, binary_search over sorted array, search in `std::map` and search operation in `std::unordered_map`.

The commands array is primarily used for search operation only within the parseCommand function.

Here are the results for these data structures. The ranges refer to bucket-frequency information for ease of understanding. The time is measured in nano-seconds.

*Linear Search on constexpr array:*
avg ns: 2927; longest ns: 793916
ranges -- 0-9ns: 0; 10-99ns: 389; 100-999ns: 44156; 1000-9999ns: 79586; 10000-99999ns: 2827; 100000ns and over: 18

*Binary Search on a constexpr sorted array:*
avg ns: 3949; longest ns: 6460959
ranges -- 0-9ns: 0; 10-99ns: 83; 100-999ns: 26735; 1000-9999ns: 98521; 10000-99999ns: 1810; 100000ns and over: 150

*`std::map` data structure:*
avg ns: 3998; longest ns: 2221792 ranges -- 0-9ns: 0; 10-99ns: 123; 100-999ns: 21839; 1000-9999ns: 103047; 10000-99999ns: 2119; 100000ns and over: 171

*`std::unordered_map` data structure:*
avg ns: 2666; longest ns: 2667042   ranges -- 0-9ns: 94; 10-99ns: 1679; 100-999ns: 43767; 1000-9999ns: 81085; 10000-99999ns: 653; 100000ns and over: 21

Both `constexpr array` and `std::unordered_map` have similar time performance.

I feel the improvements are marginal so its pragmatic to stick with a constexpr array instead of switching to unordered_map. the map does not have the advantages of compile-time static-assert checks. If the array were to get a lot bigger the unordered_map might be justified.  

[These observations are as of July 2022]

Many many thanks for Scott Schurr for providing the profiling code and also helping in debugging and brainstorming these observations.

greg7mdp:
Just a thought. `std::unordered_map`  is not very efficient, `absl::flat_hash_map` would much likely be faster.

scottschurr:
@greg7mdp, good point.  However we don't (yet) have  abseil as a dependency.  If that happens we should revisit a number of the containers used in the code base.

greg7mdp:
Oh, my bad, thank you @scottschurr , I thought we already did have an abseil dependency!

scottschurr:
I suggest that this pull request be closed.  We've not heard from the contributor for quite a while, and the changes suggested in this pull request have been superseded by https://github.com/XRPLF/rippled/commit/9aaa0dff5fd422e5f6880df8e20a1fd5ad3b4424 which is now in the develop branch.

nbougalis:
Per @scottschurr's comment, this PR has been superseded and most of the functionality has already been implemented.

PR #4071, opened 2022-01-14
Make gateway_balances admin only for reporting mode.

mtrippled:
### Type of Change

- [ ] Documentation Updates
- [ ] Release


cjcobb23 (approved):

natenichols (approved):

cjcobb23:
Actually, I realized this makes gateway_balances admin only if we compile in reporting mode, irrespective of whether we are running in reporting mode. We could compile in reporting mode but run in p2p mode, so I'm not totally sure this is what we want.

mtrippled:
This is so we can support reporting mode from a single source tree. Getting the RPC roles to support config options would take more time than this change, though that would be the next step probably if this ever becomes a problem.

PR #4072, opened 2022-01-16
rippled-dev requires protobuf-compiler and libprotoc-dev

legleux:
Check that all required components of Protobuf are present during build.
CMake's `FindProtobuf` will provide targets for the compiler and libraries if found but only _requires_ that `libprotobuf` is found and will fail when we check for the compiler's imported targets.
Specifies exactly which part is missing because this was annoying.

Also requires the `libprotoc-dev` and `protobuf-compiler` packages for `rippled-dev` packages

Fixes https://github.com/XRPLF/xrpl-dev-portal/issues/1283

### Context of Change
Ubuntu 21.04's Protobuf is version 3.12 and is now above what is specified in CMake (3.8) and the required protoc targets will not be built because `rippled-dev` packages only specify `libprotobuf-dev`.

### Type of Change


manojsdoshi (approved):
LGTM

fluksurasak:
Yes

nbougalis:
Merged as b7e0306d0a69427055d33fe8b75a982b24f7f2f4

thejohnfreeman:
Sorry that I'm getting to this late. `CMAKE_BUILD_TYPE` will not be defined for multi-configuration generators, so we may need to amend this in the future. For now, I don't expect it will be a problem because we only use single-configuration generators (Makefiles, Ninja) on platforms where we install Protobuf externally (Linux), and we only install Protobuf internally (via `ExternalProject_Add`) on platforms where we use multi-configuration generators (Visual Studio).

legleux:
> Sorry that I'm getting to this late. `CMAKE_BUILD_TYPE` will not be defined for multi-configuration generators, so we may need to amend this in the future. For now, I don't expect it will be a problem because we only use single-configuration generators (Makefiles, Ninja) on platforms where we install Protobuf externally (Linux), and we only install Protobuf internally (via `ExternalProject_Add`) on platforms where we use multi-configuration generators (Visual Studio).

Confusing, but this is not the commit that was merged, Nik's reply mentions the correct code that was committed.

thejohnfreeman:
I'm surprised that code works, because `CMAKE_BUILD_TYPE` is almost guaranteed to be empty if `is_multiconfig` is true.

thejohnfreeman:
Realized my mistake now: I missed the `else()` on line 14.

PR #4073, opened 2022-01-17
Added ResourceReport message.

bramv72:

Implemented according to:
https://gist.github.com/RichardAH/0ee2c651093d2ca2fa8a0484292990ec

This change allows peers to send, receive, and log resource usage statistics of other peers.

### Context of Change

Diagnostics and troubleshooting improvement.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Before / After
After this change, the ripple configuration can be adjusted to enable the resource_report:

```
[resource_report]
1
```

Once enabled, debug log files may contain:

`2022-Jan-16 13:44:14.740510620 UTC Protocol:DBG [001] Received ResourceReport for [016] : Idle_perc: 78.2841797; FDSize_rippled: 64; LoadAvg_5min: 1.11000001; SwapFree_kb: 33518332; SwapTotal_kb: 33520636; LoadAvg_15min: 0.850000024; Uptime_h: 483.333344; MemAvailable_kb: 64439300; Uptime_rippled_h: 0.0166666675; VmSwap_rippled_kb: 0; LoadAvg_1min: 3.21000004; Cpu_rippled_perc: 91.6722488; VmSize_rippled_kb: 2206480; num_threads_rippled: 24; MemFree_kb: 38513104; MemTotal_kb: 65638480;
`

## Test Plan

- Run two instances of rippled and make sure these nodes are connected to each other by specifying the [ips_fixed] in the rippled configuration file. Inspect the logfile to ensure that the resource report is logged (see above).
- Execute peermon (https://github.com/RichardAH/xrpl-peermon) and verify that messages with type mtResourceReport are received.

## Future Tasks

Potentially add additional useful resource metrics?

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1203452398484741


Builds/CMake/RippledCore.cmake
```diff
@@ -142,7 +142,8 @@ target_link_libraries (xrpl_core
     NIH::secp256k1
     NIH::ed25519-donna
     date::date
-    Ripple::opts)
+    Ripple::opts
+    gmp)
```
bramv72, line 146: gmp was required on ubuntu 20.04.3 with boost 1.70, even without the new pfs dependency.

Builds/CMake/deps/pfs.cmake
```diff
@@ -0,0 +1,35 @@
...
+    find_library(pfs NAMES pfs)
+    if(NOT pfs)
+        message("pfs not found. will build")
+        add_library(pfs_lib STATIC IMPORTED GLOBAL)
+        ExternalProject_Add(pfs_src
+            PREFIX ${nih_cache_path}
+            GIT_REPOSITORY https://github.com/dtrugman/pfs.git
+            GIT_TAG v0.4.5
+            CMAKE_ARGS
+            -DBUILD_SHARED_LIBS=OFF
```
bramv72, line 11: This should probably be replaced with the value of the -Dstatic option.

src/ripple/overlay/impl/PeerImp.cpp
```diff
@@ -1008,6 +1031,7 @@ void
 PeerImp::onMessageUnknown(std::uint16_t type)
 {
     // TODO
+    JLOG(journal_.debug()) << "onMessageUnknown: " << type;
```
nbougalis, line 1034: This `JLOG` can be used to flood logs and provided little useful information. I'd remove it.

src/ripple/overlay/impl/PeerImp.cpp
```diff
@@ -896,6 +909,16 @@ PeerImp::doProtocolStart()
 // Called repeatedly with protocol message data
 void
 PeerImp::onReadMessage(error_code ec, std::size_t bytes_transferred)
+{
+    this->onReadMessageInternal(ec, bytes_transferred);
```
nbougalis, line 913: The `this->` is almost certainly redundant. It's also unclear why this needs to be a separate function?

Is the intention to _always_ prevent an error message if `ec` is set? I'm not sure I think this level of logging is needed, buf if that's the intention why not simply define `onReadMessageInternal` as a directly-invoked lambda?

nbougalis (commented):
Nicely done but I have some concerns. I also don't know that it makes sense to _blindly_ send this information over the peer protocol.

src/ripple/overlay/impl/PeerImp.cpp
bramv72: Done, removed.

src/ripple/overlay/impl/PeerImp.cpp
bramv72: Done, removed.
(This was a remnant of debugging when I wanted to be sure there weren't any unreported errors. The only reason for the "Internal" method was to be able to log any error codes.)

bramv72:
> 

Understood. Under which conditions would it be acceptable to send the information?

bramv72:
In the most recent commit, I've moved the construction and ownership of the ResourceUsageCollector from PeerImp to OverlayImpl, because this ensures that there is only one instance of the ResourceUsageCollector, which can be reused by each PeerImp instance. Also, the number of files where compiler directives are used is reduced to 1, which (hopefully) improves maintainability.

PR #4080, opened 2022-01-24
Reduce memory requirements for `RippleState`

seelabs:

The motivation for these patches is saving space in `RippleState`. During path finding many millions of `RippleState` objects may be created and reducing the size of `RippleState` can help reduce the memory needed for path finding.

There are two patches. The first patch replaces many uses of `shared_ptr<RippleState>` with `RippleState`.
 
`RippleState` was normally handled as a `shared_ptr`. It appears this was done to avoid exceptions. The `make` function returned a `shared_ptr` and used a `nullptr` to signal an error.

While there is runtime overhead in allocating on the heap, this patch is motivated by the space overhead of using a `shared_ptr` as well as the space overhead of allocating on the heap. The overhead of malloc is typically 8-16 bytes. The overhead of a
shared_ptr is typically around 48 bytes.

The second patch removes the `sle` that was stored in the `RippleState` object. Storing the sle means storing some fields twice or keeping references to the data. If the sle is able to be reclaimed, this reduces the memory required to store
the RippleState.

There was also no reason for this class to have a vtable. The virtual destructor was removed and the class was marked final.

### Type of Change

- [ x] Memory Optimization (non-breaking change)

### DIFF
--- a/src/ripple/app/paths/PathFindTrustLine.h
+++ b/src/ripple/app/paths/PathFindTrustLine.h
@@ -37,18 +37,21 @@ namespace ripple {
     "low" account and a "high" account. This wraps the
     SLE and expresses its data from the perspective of
     a chosen account on the line.
+
+    This wrapper is primarily used in the path finder and there can easily be
+    tens of millions of instances of this class. When modifying this class think
+    carefully about the memory implications.
 */
-// VFALCO TODO Rename to TrustLine
-class RippleState final : public CountedObject<RippleState>
+class PathFindTrustLine final : public CountedObject<PathFindTrustLine>
 {
 public:
-    RippleState() = delete;
+    PathFindTrustLine() = delete;
 
-    static std::optional<RippleState>
+    static std::optional<PathFindTrustLine>
     makeItem(AccountID const& accountID, std::shared_ptr<SLE const> const& sle);
 
     // Must be public, for make_shared
-    RippleState(
+    PathFindTrustLine(
         std::shared_ptr<SLE const> const& sle,
         AccountID const& viewAccount);
 
@@ -169,7 +172,7 @@ private:
     bool mViewLowest;
 };
 
-std::vector<RippleState>
+std::vector<PathFindTrustLine>
 getRippleStateItems(AccountID const& accountID, ReadView const& view);
 
 }  // namespace ripple
--- a/src/ripple/app/paths/RippleState.cpp
+++ b/src/ripple/app/paths/RippleState.cpp
@@ -25,17 +25,6 @@
 
 namespace ripple {
 
-std::optional<RippleState>
-RippleState::makeItem(
-    AccountID const& accountID,
-    std::shared_ptr<SLE const> const& sle)
-{
-    // VFALCO Does this ever happen in practice?
-    if (!sle || sle->getType() != ltRIPPLE_STATE)
-        return {};
-    return std::optional{RippleState{sle, accountID}};
-}
-
 RippleState::RippleState(
     std::shared_ptr<SLE const> const& sle,
     AccountID const& viewAccount)
@@ -52,6 +41,17 @@ RippleState::RippleState(
 {
     if (!mViewLowest)
         mBalance.negate();
+}
+
+std::optional<RippleState>
+RippleState::makeItem(
+    AccountID const& accountID,
+    std::shared_ptr<SLE const> const& sle)
+{
+    // VFALCO Does this ever happen in practice?
+    if (!sle || sle->getType() != ltRIPPLE_STATE)
+        return {};
+    return std::optional{RippleState{sle, accountID}};
 }
 
 Json::Value
--- a/src/ripple/app/paths/RippleState.cpp
+++ b/src/ripple/app/paths/RippleState.cpp
@@ -42,15 +42,13 @@ RippleState::RippleState(
     : key_(sle->key())
     , mLowLimit(sle->getFieldAmount(sfLowLimit))
     , mHighLimit(sle->getFieldAmount(sfHighLimit))
-    , mLowID(mLowLimit.getIssuer())
-    , mHighID(mHighLimit.getIssuer())
+    , mBalance(sle->getFieldAmount(sfBalance))
     , lowQualityIn_(sle->getFieldU32(sfLowQualityIn))
     , lowQualityOut_(sle->getFieldU32(sfLowQualityOut))
     , highQualityIn_(sle->getFieldU32(sfHighQualityIn))
     , highQualityOut_(sle->getFieldU32(sfHighQualityOut))
-    , mBalance(sle->getFieldAmount(sfBalance))
     , mFlags(sle->getFieldU32(sfFlags))
-    , mViewLowest(mLowID == viewAccount)
+    , mViewLowest(mLowLimit.getIssuer() == viewAccount)
 {
     if (!mViewLowest)
         mBalance.negate();
@@ -60,8 +58,8 @@ Json::Value
 RippleState::getJson(int)
 {
     Json::Value ret(Json::objectValue);
-    ret["low_id"] = to_string(mLowID);
-    ret["high_id"] = to_string(mHighID);
+    ret["low_id"] = to_string(mLowLimit.getIssuer());
+    ret["high_id"] = to_string(mHighLimit.getIssuer());
     return ret;
 }
 
--- a/src/ripple/app/paths/RippleState.cpp
+++ b/src/ripple/app/paths/RippleState.cpp
@@ -25,7 +25,7 @@
 
 namespace ripple {
 
-RippleState::pointer
+std::optional<RippleState>
 RippleState::makeItem(
     AccountID const& accountID,
     std::shared_ptr<SLE const> sle)
@@ -33,7 +33,7 @@ RippleState::makeItem(
     // VFALCO Does this ever happen in practice?
     if (!sle || sle->getType() != ltRIPPLE_STATE)
         return {};
-    return std::make_shared<RippleState>(std::move(sle), accountID);
+    return std::optional{RippleState{std::move(sle), accountID}};
 }
 
 RippleState::RippleState(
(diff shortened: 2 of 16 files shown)


src/ripple/app/paths/RippleState.h
```diff
@@ -153,27 +151,27 @@ class RippleState
...
-
-    STAmount const& mLowLimit;
-    STAmount const& mHighLimit;
+    STAmount const mLowLimit;
+    STAmount const mHighLimit;
 
-    AccountID const& mLowID;
-    AccountID const& mHighID;
+    AccountID const mLowID;
+    AccountID const mHighID;
```
nbougalis, line 160: You actually don't _need_ the `mLowID` and `mHighID` variables: you can always access those via `m[Low|High]Limit.getIssuer()` without incurring any significant runtime cost. This saves 40 more bytes.

src/ripple/app/paths/RippleState.cpp
```diff
@@ -66,17 +65,17 @@ RippleState::getJson(int)
...
+std::vector<RippleState>
 getRippleStateItems(AccountID const& accountID, ReadView const& view)
 {
-    std::vector<RippleState::pointer> items;
+    std::vector<RippleState> items;
     forEachItem(
         view,
         accountID,
         [&items, &accountID](std::shared_ptr<SLE const> const& sleCur) {
             auto ret = RippleState::makeItem(accountID, sleCur);
```
nbougalis, line 76: I'd remove `makeItem` entirely; `forEachItem` will never pass in a `sleCur` that's `nullptr` and you can just inline the check for `ltRIPPLE_STATE`:

    [&items, &accountID](std::shared_ptr<SLE const> const& sleCur) {
        if (sleCur->getType() == ltRIPPLE_STATE)
            item.emplace_back(sleCur, accountID);
    }

You'll need to run this check in the other place where `makeItem` is invoked, but overall it's cleaner and simpler.

src/ripple/app/paths/RippleState.h
```diff
@@ -153,27 +151,27 @@ class RippleState
...
 
     Rate lowQualityIn_;
     Rate lowQualityOut_;
     Rate highQualityIn_;
     Rate highQualityOut_;
 
     STAmount mBalance;
+
```
nbougalis, line 168: Note: investigate whether re-ordering all these variables can result in more effective packing for the structure.

nbougalis (changes requested):

src/ripple/app/paths/RippleState.cpp
seelabs:
```
You'll need to run this check in the other place where makeItem is invoked, but overall it's cleaner and simpler.
```
We should talk about this. Since there's no way for a constructor to signal an error (other than an exception, which we disallow), the our options on error are:
 1) it's a logic error - terminate the program; 
2) an `bool OK()` function that needs to be checked after construction. 
(2) Doesn't work well in practice. It's too easy to forget to call.
(1) is a reasonable option if the constructor is only called in a very limited context. Then it's easy to audit all the callers to make sure the preconditions are met, and we don't expect new calls to be added.

I think this is class is too generally useful to expect no new calls to be added. I think a `make` function is the least bad option we have.

I'm very open to discussion on this. But for now, I vote we keep the `make` function. We should also talk about relaxing our restrictions on throwing.

src/ripple/app/paths/RippleState.h
seelabs: That's the reason I moved the `flags` and `viewLowest` members to the end. Right now there should be no unneeded padding, but we could move the `mbalance` before the `quality` since its larger.

src/ripple/app/paths/RippleState.h
seelabs: That's a good idea. Will do.

In the longer term, we should replace the `STAmount` variables with `IOUAmount` variables and keep the currency and issuers separate. The reason I don't want to do that now is we'd have to carefully audit those two classes to make sure all the corner cases are handed exactly the same.

src/ripple/app/paths/RippleState.h
seelabs: Fixed in `7643b21204 Remove lowID and highID from RippleState:`

ximinez (approved):
I gave this a quick once over, not a deep review, so I could use it in my work. All the changes look clean and sensible to me.

src/ripple/app/paths/RippleState.h
```diff
@@ -153,27 +151,24 @@ class RippleState
...
-    STAmount const& mHighLimit;
-
-    AccountID const& mLowID;
-    AccountID const& mHighID;
+    STAmount mBalance;
 
     Rate lowQualityIn_;
     Rate lowQualityOut_;
     Rate highQualityIn_;
     Rate highQualityOut_;
```
nbougalis, line 164: Since you're in here, only two of these four values are used by this class (either the `low` or the `high` versions, depending on `mViewLowest`, so we can save 8 bytes if we remove the other two.

We _could_ actually save 16 bytes by removing all four values entirely; they only place where they're referenced from is in `AccountLines.cpp` at which point, you might as well reload the `SLE` and grab these values directly.

src/ripple/app/paths/RippleState.cpp
```diff
@@ -25,58 +25,55 @@
...
+        mBalance.negate();
+}
+
+std::optional<RippleState>
 RippleState::makeItem(
     AccountID const& accountID,
-    std::shared_ptr<SLE const> sle)
+    std::shared_ptr<SLE const> const& sle)
 {
     // VFALCO Does this ever happen in practice?
```
nbougalis, line 51: Please remove this comment, or at least update it:

It's possible that `sle == nullptr` but that would indicate an orphaned entry in an owner directory, which should, IMO, be caught elsewhere.

It's common to have `sle->getType() != ltRIPPLE_STATE` because we call this function as we iterate owner directories, for every entry in that directory, and entries of different types are expected.

src/ripple/app/paths/RippleState.h
```diff
@@ -24,8 +24,10 @@
 #include <ripple/protocol/Rate.h>
 #include <ripple/protocol/STAmount.h>
 #include <ripple/protocol/STLedgerEntry.h>
+
 #include <cstdint>
-#include <memory>  // <memory>
+#include <memory>
```
nbougalis, line 29: Is this _still_ needed now that we eliminated the `std::make_shared`?

nbougalis (approved):

src/ripple/app/paths/RippleState.cpp
seelabs: fixed in `0a3e192530 [fold] Rm old comments`

src/ripple/app/paths/RippleState.h
seelabs: fixed in `7a5b4d3665 [fold] Rm rate from trust line cache for the path finder`

I did some refactoring to do this. We know have a base class `TrustLineBase` and two classes that inherit from it `PathFindTrustLine` that doesn't have the qualities, so we save space on the tens of millions of those. We also have an `RPCTrustLine` that has the rates that is used for the `AccountLines` command.

src/ripple/app/paths/RippleState.h
seelabs: fixed in `875f1e28c8 [fold] Rm unused include file`

src/ripple/app/paths/PathFindTrustLine.h
```diff
@@ -42,19 +42,18 @@ namespace ripple {
...
-    PathFindTrustLine() = delete;
-
-    static std::optional<PathFindTrustLine>
-    makeItem(AccountID const& accountID, std::shared_ptr<SLE const> const& sle);
-
-    // Must be public, for make_shared
-    PathFindTrustLine(
+protected:
+    // This class should not be instantiated directly. Use one of the derived
+    // classes.
```
HowardHinnant: This comment can be more strongly enforced by making the special members protected:
```c++
    ~TrustLineBase() = default;
    TrustLineBase(TrustLineBase const&) = default;
    TrustLineBase& operator=(TrustLineBase const&) = delete;
    TrustLineBase(TrustLineBase &&) = default;
```
Plus this keeps clients from slicing to a `TrustLineBase&`.  I omitted the deleted move assignment operator as that seemed redundant to me.  But add it in if you think it reads better.  And I added the deleted copy assignment operator (redundantly) otherwise the reader has to inspect the data members to discover that it is deleted.

seelabs: Thanks! Fixed in `bde04717c4 [fold] Additional protection agains instantiating TrustLineBase`

src/ripple/app/paths/PathFindTrustLine.h
```diff
@@ -42,19 +42,18 @@ namespace ripple {
...
+    TrustLineBase(
         std::shared_ptr<SLE const> const& sle,
         AccountID const& viewAccount);
 
+public:
+    TrustLineBase() = delete;
```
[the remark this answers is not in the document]
seelabs: Fixed in `bde04717c4 [fold] Additional protection agains instantiating TrustLineBase`

src/ripple/rpc/handlers/AccountLines.cpp
```diff
@@ -218,6 +206,21 @@ doAccountLines(RPC::JsonContext& context)
                             RPC::getStartHint(sleCur, visitData.accountID);
                     }
 
+                    if (sleCur->getType() != ltRIPPLE_STATE)
+                        return true;
```
thejohnfreeman, line 210: Is this a sanity check? Is there ever a case this early return would be hit?

thejohnfreeman (approved):

intelliot:
@mvadari @mtrippled it would be fascinating to run a couple tests to see what effect these patches have on the max stable capacity of a test network

seelabs:
@intelliot This patch only effects `path_find`. If the benchmark isn't running path_find I wouldn't expect this to have any effect.

nbougalis:
This was merged into *1.9.0*

PR #4082, opened 2022-01-25
Compile rippled with "-pie"

legleux:
Add gcc position independent executable flag for added binary hardening.

### Type of Change


legleux:
Looks like this may need `-pie` added to the linker invocation so this isn't quite done.
[Slew of CMake issues discussing PIE](https://gitlab.kitware.com/cmake/cmake/-/issues/14983)
[Can manually do it](https://mropert.github.io/2018/02/02/pic_pie_sanitizers/#:~:text=final%20executable%20to%20add%20%2Dpie%20by%20hand%20using%20CMAKE_EXE_LINKER_FLAGS)

thejohnfreeman (changes requested):
Leaving it to policy CMP0083 requires CMake 3.14, but we only require at least 3.9. I don't think we want to bump that requirement just for this feature so two options:

- If everything will still link successfully, but perhaps position dependently, with CMake <3.14, then I can approve as-is.
- If this will not link successfully because of the missing flag when using CMake <3.14, then we need to either add the flag manually (my preference) or revert these changes.

legleux:
Hmm, rippled doesn't build with CMake 3.9 anymore because of the updated rocksdb version.
Let's just bump the minimum as high as we date.

Builds/CMake/RippledCore.cmake
```diff
@@ -3,6 +3,16 @@
...
+if (CMAKE_VERSION VERSION_LESS 3.14)
+  string (APPEND CMAKE_CXX_FLAGS " -pie")
+else()
+  if(POLICY CMP0083)
+    cmake_policy(SET CMP0083 NEW)
+    include(CheckPIESupported)
+    check_pie_supported()
+  endif()
+endif()
+
```
thejohnfreeman, line 15: If #4109 goes in, this block should be reduced to just lines 11 and 12.

thejohnfreeman (changes requested):

thejohnfreeman (approved):

intelliot:
(notes: would be good to merge)

@RichardAH does this interfere with any work you're doing with Hooks or any other project?

legleux:
Almost none of these files even exist anymore

PR #4087, opened 2022-01-28
lower_bound for clio ETL

cjcobb23:

 clio keeps a table of object successors, which given an object, allows one to look up the next object in a given ledger. clio computes this using an in memory cache, but on startup, the cache is empty. So on startup, clio needs to ask rippled for this information, until clio is able to fill its cache and use the cache. To achieve this, a new function was added to `SHAMap`, called `lower_bound`, which returns the first object prior to a given object ID. To implement this function, a new function was added to `SHAMap` called `lastBelow`, which returns the last item at or below a given node.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,6 +5,7 @@ if (POLICY CMP0074)
 endif ()
 
 project (rippled)
+set (CMAKE_CXX_STANDARD 17)
 
 # make GIT_COMMIT_HASH define available to all sources
 find_package(Git)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,7 +5,9 @@ if (POLICY CMP0074)
 endif ()
 
 project (rippled)
-set (CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_EXTENSIONS OFF)
+set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD_REQUIRED ON)
 
 # make GIT_COMMIT_HASH define available to all sources
 find_package(Git)
--- a/src/ripple/proto/org/xrpl/rpc/v1/get_ledger.proto
+++ b/src/ripple/proto/org/xrpl/rpc/v1/get_ledger.proto
@@ -70,6 +70,9 @@ message GetLedgerResponse
     // map diff
     bool object_neighbors_included = 9;
 
+
+    // Successor information for book directories modified as part of this
+    // ledger
     repeated BookSuccessor book_successors = 10;
 }
 
--- a/src/ripple/proto/org/xrpl/rpc/v1/get_ledger.proto
+++ b/src/ripple/proto/org/xrpl/rpc/v1/get_ledger.proto
@@ -32,6 +32,10 @@ message GetLedgerRequest
     // coming from a secure_gateway host, then the client is not subject to
     // resource controls
     string user = 6;
+
+    // For every object in the diff, get the object's predecessor and successor
+    // in the state map. Only used if get_objects is also true.
+    bool get_object_neighbors = 7;
 }
 
 message GetLedgerResponse
@@ -58,6 +62,15 @@ message GetLedgerResponse
 
     // True if request was exempt from resource controls
     bool is_unlimited = 7;
+
+    // True if the response contains the state map diff
+    bool objects_included = 8;
+
+    // True if the response contains key of objects adjacent to objects in state
+    // map diff
+    bool object_neighbors_included = 9;
+
+    repeated BookSuccessor book_successors = 10;
 }
 
 message TransactionHashList
--- a/src/ripple/proto/org/xrpl/rpc/v1/ledger.proto
+++ b/src/ripple/proto/org/xrpl/rpc/v1/ledger.proto
@@ -59,12 +59,6 @@ message RawLedgerObject
     bytes predecessor = 4;
 
     bytes successor = 5;
-
-    bool is_first_book_dir = 6;
-
-    bool was_first_book_dir = 7;
-
-    bytes new_first_book_dir = 8;
 }
 
 message RawLedgerObjects
--- a/src/ripple/proto/org/xrpl/rpc/v1/ledger.proto
+++ b/src/ripple/proto/org/xrpl/rpc/v1/ledger.proto
@@ -54,10 +54,13 @@ message RawLedgerObject
         DELETED = 3;
     }
 
+    // Whether the object was created, modified or deleted
     ModificationType mod_type = 3;
 
+    // Key of the object preceding this object in the desired ledger
     bytes predecessor = 4;
 
+    // Key of the object succeeding this object in the desired ledger
     bytes successor = 5;
 }
 
@@ -66,11 +69,16 @@ message RawLedgerObjects
     repeated RawLedgerObject objects = 1;
 }
 
+// Successor information for book directories. The book base is (usually) not
+// an actual object, yet we need to be able to ask for the successor to the
+// book base.
 message BookSuccessor {
 
+    // Base of the book in question
     bytes book_base = 1;
 
-    // empty value here means the entire book is deleted
+    // First book directory in the book. An empty value here means the entire
+    // book is deleted
     bytes first_book = 2;
 
 };
--- a/src/ripple/proto/org/xrpl/rpc/v1/ledger.proto
+++ b/src/ripple/proto/org/xrpl/rpc/v1/ledger.proto
@@ -55,10 +55,29 @@ message RawLedgerObject
     }
 
     ModificationType mod_type = 3;
+
+    bytes predecessor = 4;
+
+    bytes successor = 5;
+
+    bool is_first_book_dir = 6;
+
+    bool was_first_book_dir = 7;
+
+    bytes new_first_book_dir = 8;
 }
 
 message RawLedgerObjects
 {
     repeated RawLedgerObject objects = 1;
 }
+
+message BookSuccessor {
+
+    bytes book_base = 1;
+
+    // empty value here means the entire book is deleted
+    bytes first_book = 2;
+
+};
 
--- a/src/ripple/rpc/handlers/LedgerHandler.cpp
+++ b/src/ripple/rpc/handlers/LedgerHandler.cpp
@@ -105,6 +105,7 @@ LedgerHandler::check()
 std::pair<org::xrpl::rpc::v1::GetLedgerResponse, grpc::Status>
 doLedgerGrpc(RPC::GRPCContext<org::xrpl::rpc::v1::GetLedgerRequest>& context)
 {
+    auto begin = std::chrono::system_clock::now();
     org::xrpl::rpc::v1::GetLedgerRequest& request = context.params;
     org::xrpl::rpc::v1::GetLedgerResponse response;
     grpc::Status status = grpc::Status::OK;
@@ -212,12 +213,104 @@ doLedgerGrpc(RPC::GRPCContext<org::xrpl::rpc::v1::GetLedgerRequest>& context)
                 obj->set_mod_type(org::xrpl::rpc::v1::RawLedgerObject::DELETED);
             else
                 obj->set_mod_type(org::xrpl::rpc::v1::RawLedgerObject::CREATED);
+            auto blob = inDesired ? inDesired->slice() : inBase->slice();
+            LedgerEntryType objectType =
+                LedgerEntryType{(uint16_t)(blob[1] << 8 | blob[2])};
+
+            if (request.get_object_neighbors())
+            {
+                if (!(inBase && inDesired))
+                {
+                    auto lb = desired->stateMap().lower_bound(k);
+                    auto ub = desired->stateMap().upper_bound(k);
+                    if (lb != desired->stateMap().end())
+                        obj->set_predecessor(
+                            lb->key().data(), lb->key().size());
+                    if (ub != desired->stateMap().end())
+                        obj->set_successor(ub->key().data(), ub->key().size());
+                    if (objectType == ltDIR_NODE)
+                    {
+                        auto sle = std::make_shared<SLE>(SerialIter{blob}, k);
+                        if (!sle->isFieldPresent(sfOwner))
+                        {
+                            auto bookBase = keylet::quality({ltDIR_NODE, k}, 0);
+                            if (!inBase && inDesired)
+                            {
+                                auto firstBook =
+                                    desired->stateMap().upper_bound(
+                                        bookBase.key);
+                                if (firstBook != desired->stateMap().end() &&
+                                    firstBook->key() <
+                                        getQualityNext(bookBase.key) &&
+                                    firstBook->key() == k)
+                                {
+                                    auto succ = response.add_book_successors();
+                                    succ->set_book_base(
+                                        bookBase.key.data(),
+                                        bookBase.key.size());
+                                    succ->set_first_book(
+                                        firstBook->key().data(),
+                                        firstBook->key().size());
+                                }
+                            }
+                            if (inBase && !inDesired)
+                            {
+                                auto oldFirstBook =
+                                    base->stateMap().upper_bound(bookBase.key);
+                                if (oldFirstBook != base->stateMap().end() &&
+                                    oldFirstBook->key() <
+                                        getQualityNext(bookBase.key) &&
+                                    oldFirstBook->key() == k)
+                                {
+                                    obj->set_was_first_book_dir(true);
+                                    auto succ = response.add_book_successors();
+                                    succ->set_book_base(
+                                        bookBase.key.data(),
+                                        bookBase.key.size());
+                                    auto newFirstBook =
+                                        desired->stateMap().upper_bound(
+                                            bookBase.key);
+
+                                    if (newFirstBook !=
+                                            desired->stateMap().end() &&
+                                        newFirstBook->key() <
+                                            getQualityNext(bookBase.key))
+                                    {
+                                        succ->set_first_book(
+                                            newFirstBook->key().data(),
+                                            newFirstBook->key().size());
+                                        obj->set_new_first_book_dir(
+                                            newFirstBook->key().data(),
+                                            newFirstBook->key().size());
+                                    }
+                                }
+                            }
+                        }
+                    }
+                }
+            }
         }
+        response.set_objects_included(true);
+        response.set_object_neighbors_included(request.get_object_neighbors());
         response.set_skiplist_included(true);
     }
 
     response.set_validated(
         RPC::isValidated(context.ledgerMaster, *ledger, context.app));
+
+    auto end = std::chrono::system_clock::now();
+    auto duration =
+        std::chrono::duration_cast<std::chrono::milliseconds>(end - begin)
+            .count() *
+        1.0;
+    JLOG(context.j.warn())
+        << __func__ << " - Extract time = " << duration
+        << " - num objects = " << response.ledger_objects().objects_size()
+        << " - num txns = " << response.transactions_list().transactions_size()
+        << " - ms per obj "
+        << duration / response.ledger_objects().objects_size()
+        << " - ms per txn "
+        << duration / response.transactions_list().transactions_size();
 
     return {response, status};
 }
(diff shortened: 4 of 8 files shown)


src/ripple/proto/org/xrpl/rpc/v1/get_ledger.proto
```diff
@@ -58,6 +62,15 @@ message GetLedgerResponse
...
     bool is_unlimited = 7;
+
+    // True if the response contains the state map diff
+    bool objects_included = 8;
+
+    // True if the response contains key of objects adjacent to objects in state
+    // map diff
+    bool object_neighbors_included = 9;
+
+    repeated BookSuccessor book_successors = 10;
```
natenichols, line 73: For `book_successors`, are you getting individual `Offer` objects successor and predecessors, or `DirectoryNode` predecessors and successors?

If it's the former, there's really no need to have that information in Clio. You can simply traverse `DirectoryNodes` similar to `ledger_data`, and then batch fetch all the offers, instead of traversing the offers like a linked list.

src/ripple/shamap/impl/SHAMap.cpp
```diff
@@ -475,6 +475,45 @@ SHAMap::firstBelow(
...
+        auto n = std::static_pointer_cast<SHAMapLeafNode>(node);
+        stack.push({node, {leafDepth, n->peekItem()->key()}});
+        return n.get();
+    }
+    auto inner = std::static_pointer_cast<SHAMapInnerNode>(node);
+    if (stack.empty())
+        stack.push({inner, SHAMapNodeID{}});
+    else
+        stack.push({inner, stack.top().second.getChildNodeID(branch)});
+    for (int i = branchFactor; i >= 0;)
```
seelabs, line 496: This looks like an off by one error. `branchFactor` isn't a valid index. I think we want to start at `branchFactor-1`

src/ripple/shamap/impl/SHAMap.cpp
```diff
@@ -475,6 +475,45 @@ SHAMap::firstBelow(
...
+            assert(!stack.empty());
+            if (node->isLeaf())
+            {
+                auto n = std::static_pointer_cast<SHAMapLeafNode>(node);
+                stack.push({n, {leafDepth, n->peekItem()->key()}});
+                return n.get();
+            }
+            inner = std::static_pointer_cast<SHAMapInnerNode>(node);
+            stack.push({inner, stack.top().second.getChildNodeID(branch)});
+            i = branchFactor;  // scan all 16 branches of this new node
```
seelabs, line 510: Off by one error here too. Should be `branchFactor-1`

src/ripple/shamap/impl/SHAMap.cpp
```diff
@@ -475,6 +475,45 @@ SHAMap::firstBelow(
...
+        }
+        else
+            --i;  // scan next branch
+    }
+    return nullptr;
+}
```
seelabs, line 516: This new `lastBelow` function is almost identical to `firstBelow`. We could use a single implementation. Something like (untested, but you get the idea):
```
namespace detail {
struct Forward
{
    static int
    start()
    {
        return 0;
    }
    static int
    end()
    {
        return SHAMap::branchFactor;
    }
    static int
    next(int i)
    {
        return i + 1;
    }
};
struct Reverse
{
    static int
    start()
    {
        return SHAMap::branchFactor - 1;
    }
    static int
    end()
    {
        return -1;
    }
    static int
    next(int i)
    {
        return i - 1;
    }
};
}  // namespace detail

template <class I>
SHAMapLeafNode*
SHAMap::belowHelper(
    std::shared_ptr<SHAMapTreeNode> node,
    SharedPtrNodeStack& stack,
    int branch) const
{
    // Return the first item at or below this node
    if (node->isLeaf())
    {
        auto n = std::static_pointer_cast<SHAMapLeafNode>(node);
        stack.push({node, {leafDepth, n->peekItem()->key()}});
        return n.get();
    }
    auto inner = std::static_pointer_cast<SHAMapInnerNode>(node);
    if (stack.empty())
        stack.push({inner, SHAMapNodeID{}});
    else
        stack.push({inner, stack.top().second.getChildNodeID(branch)});
    for (int i = I::start(); i != I::end();)
    {
        if (!inner->isEmptyBranch(i))
        {
            node = descendThrow(inner, i);
            assert(!stack.empty());
            if (node->isLeaf())
            {
                auto n = std::static_pointer_cast<SHAMapLeafNode>(node);
                stack.push({n, {leafDepth, n->peekItem()->key()}});
                return n.get();
            }
            inner = std::static_pointer_cast<SHAMapInnerNode>(node);
            stack.push({inner, stack.top().second.getChildNodeID(branch)});
            i = I::start();  // scan all 16 branches of this new node
        }
        else
            i = I::next(i);  // scan next branch
    }
    return nullptr;
}

SHAMapLeafNode*
SHAMap::firstBelow(
    std::shared_ptr<SHAMapTreeNode> node,
    SharedPtrNodeStack& stack,
    int branch) const
{
    return belowHelper<detail::Forward>(std::move(node), stack, branch);
}
SHAMapLeafNode*
SHAMap::lastBelow(
    std::shared_ptr<SHAMapTreeNode> node,
    SharedPtrNodeStack& stack,
    int branch) const
{
    return belowHelper<detail::Reverse>(std::move(node), stack, branch);
}

```

BTW, I don't object to two functions either. Sometimes more generic is harder to understand and debug. If you prefer using two function I'm fine with that. The above is just a thought.

src/ripple/shamap/impl/SHAMap.cpp
```diff
@@ -619,6 +658,43 @@ SHAMap::upper_bound(uint256 const& id) const
...
+                    return const_iterator(
+                        this, leaf->peekItem().get(), std::move(stack));
+                }
+            }
+        }
+        stack.pop();
+    }
+    // TODO: what to return here?
+    return end();
+}
```
seelabs, line 697: This function is almost identical to `upper_bound`. We could try to refactor it into a single function. But like I said in the other comment, I don't object to two functions here either.

src/ripple/shamap/SHAMap.h
```diff
@@ -213,6 +213,10 @@ class SHAMap
     const_iterator
     upper_bound(uint256 const& id) const;
 
+    // traverse functions
+    const_iterator
+    lower_bound(uint256 const& id) const;
```
seelabs, line 218: We should document what these functions do. I know there's an existing almost identical function here that isn't documented. Still, I think new functions should include a a brief description of what it does.

Ditto for the new `lastBelow` function.

src/ripple/rpc/handlers/LedgerHandler.cpp
```diff
@@ -212,13 +213,105 @@ doLedgerGrpc(RPC::GRPCContext<org::xrpl::rpc::v1::GetLedgerRequest>& context)
                 obj->set_mod_type(org::xrpl::rpc::v1::RawLedgerObject::DELETED);
             else
                 obj->set_mod_type(org::xrpl::rpc::v1::RawLedgerObject::CREATED);
+            auto blob = inDesired ? inDesired->slice() : inBase->slice();
```
seelabs, line 216: Can we add unit tests that excercises this new functionality?

seelabs (commented):
Did a pass; Looks good, but I have a couple suggestions:
1) Can we add unit tests that exercise the new functionality?
2) Can we add some brief comments in the header explaining what the functions do?
3) I left some other minor suggestions in the code. I'm honestly not sure if I'd refactor the almost identical functions into common implementations. There are tradeoffs either way, but I showed one possible refactoring to consider.
4) Take a look at the off-by-one bug I flagged.

src/ripple/rpc/handlers/LedgerHandler.cpp
```diff
@@ -212,13 +213,105 @@ doLedgerGrpc(RPC::GRPCContext<org::xrpl::rpc::v1::GetLedgerRequest>& context)
...
+            LedgerEntryType objectType =
+                LedgerEntryType{(uint16_t)(blob[1] << 8 | blob[2])};
```
seelabs, line 218: I tend to avoid c-style casts. Can't we just cast directly to the `LedgerEntryType`? I.e.
```
            auto const objectType =
                static_cast<LedgerEntryType>(blob[1] << 8 | blob[2]);
```
Also, I think we can make this `const`.

src/ripple/shamap/impl/SHAMap.cpp
cjcobb23: good catch

src/ripple/rpc/handlers/LedgerHandler.cpp
cjcobb23: Yeah I'll add some tests

cjcobb23:
@seelabs this is ready for another look when you get a chance

src/ripple/shamap/SHAMap.h
```diff
@@ -210,10 +210,14 @@ class SHAMap
     peekItem(uint256 const& id, SHAMapHash& hash) const;
 
     // traverse functions
+
+    // finds the object in the tree with the smallest object id greater than the
+    // input id
```
seelabs, line 215: Nit: Probably should be javadoc style comments. On the function below too.

src/ripple/rpc/handlers/LedgerHandler.cpp
seelabs: Nit: could be `const`

src/ripple/shamap/impl/SHAMap.cpp
```diff
@@ -468,14 +470,37 @@ SHAMap::firstBelow(
...
+SHAMap::lastBelow(
+    std::shared_ptr<SHAMapTreeNode> node,
+    SharedPtrNodeStack& stack,
+    int branch) const
+{
+    return below(
+        node,
+        stack,
+        branch,
+        {branchFactor - 1, [](int i) { return i >= 0; }, [](int& i) { --i; }});
```
natenichols, line 490: This one is bikeshedding, but IMO it would be easier to read if we did something like:
```cpp
auto init = branchFactor - 1;
auto cmp = [](int i) { return i >= 0; };
auto inc = [](int& i) { --i; };

return below(node, stack, branch, {init, cmp, inc});
```
versus just passing in a bunch of lambdas in as a tuple.

Same for firstBelow, totally up to you.

src/ripple/proto/org/xrpl/rpc/v1/ledger.proto
```diff
@@ -55,10 +55,29 @@ message RawLedgerObject
     }
 
     ModificationType mod_type = 3;
+
+    bytes predecessor = 4;
```
natenichols, line 59: Would you add comments to these protobuf definitions, similar to `get_ledger.proto`?

natenichols (approved):
Nice work, and the tests seem to cover the additions as far as I can tell. Left a couple of nitpicks, but nothing that should hold up approval.

LGTM 👍

src/test/ledger/View_test.cpp
```diff
@@ -384,6 +384,279 @@ class View_test : public beast::unit_test::suite
...
+            BEAST_EXPECT(
+                ledger->stateMap().upper_bound(uint256(85))->key() ==
+                uint256(100));
+            BEAST_EXPECT(
+                ledger->stateMap().upper_bound(uint256(98))->key() ==
+                uint256(100));
+            BEAST_EXPECT(ledger->stateMap().upper_bound(uint256(100)) == e);
+            BEAST_EXPECT(ledger->stateMap().upper_bound(uint256(155)) == e);
+        }
+    }
```
thejohnfreeman, line 658: I appreciate that there are tests, but I'd like to try to be a little more judicious with the code (less to read, less to compile, less to run). The tests here seem to span a few categories:

- When the ID is in vs not in the SHAMap.
- When the request is for lower vs upper bound.
- When the requested ID is in vs not in the SHAMap.
- When the requested ID has the same parent inner node vs sibling parent inner node vs distant parent inner node.

So that's 24 tests that can be satisfied with one example SHAMap. Would be nice to put the test strategy in a comment, if not going so far as to name them separately.

cjcobb23:
> Can we get an explanation of why clio keeps the table of successors, and how it uses this response to construct that table?

Clio keeps a table of successors to implement book_offers and ledger_data. Both of these RPCs use the upper_bound function of SHAMap, so clio needs to be able to do this as well. Since clio's data model is flattened, and not a tree, doing upper_bound without the successor table would require a table scan, which is quite slow, and becomes slower as history grows.

src/ripple/shamap/impl/SHAMap.cpp
```diff
@@ -437,12 +437,14 @@ SHAMap::unshareNode(std::shared_ptr<Node> node, SHAMapNodeID const& nodeID)
 }
 
 SHAMapLeafNode*
-SHAMap::firstBelow(
+SHAMap::below(
     std::shared_ptr<SHAMapTreeNode> node,
     SharedPtrNodeStack& stack,
-    int branch) const
+    int branch,
+    std::tuple<int, std::function<bool(int)>, std::function<void(int&)>> const&
```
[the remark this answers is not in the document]
cjcobb23: I don't think packing all of the parameters together makes sense. This parameter pack is the loopParams, and so I think it makes sense to pack them together, since they are intimately related to each other. I originally wasn't going to pack them at all, but @seelabs suggested I pack them.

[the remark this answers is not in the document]
seelabs: I suggested we put them into a struct, but I also said I was OK with whatever solution cj decided on (and I still am).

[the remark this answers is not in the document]
thejohnfreeman: His suggestion packed them into a type to make one template parameter pack for all the template parameters. Here, they're not template parameters (despite being constants).

src/test/ledger/View_test.cpp
```diff
@@ -384,6 +384,279 @@ class View_test : public beast::unit_test::suite
...
+            auto b = ledger->stateMap().begin();
+            auto e = ledger->stateMap().end();
+            BEAST_EXPECT(ledger->stateMap().lower_bound(uint256(0)) == e);
+            BEAST_EXPECT(ledger->stateMap().lower_bound(uint256(1)) == b);
+            BEAST_EXPECT(
+                ledger->stateMap().lower_bound(uint256(5))->key() ==
+                uint256(4));
+            BEAST_EXPECT(
+                ledger->stateMap().lower_bound(uint256(5))->key() ==
+                uint256(4));
```
thejohnfreeman, line 562: These lines are duplicates of the above.

src/test/ledger/View_test.cpp
```diff
@@ -384,6 +384,279 @@ class View_test : public beast::unit_test::suite
...
+                uint256(16));
+            BEAST_EXPECT(
+                ledger->stateMap().lower_bound(uint256(24))->key() ==
+                uint256(20));
+            BEAST_EXPECT(
+                ledger->stateMap().lower_bound(uint256(31))->key() ==
+                uint256(30));
+            BEAST_EXPECT(
+                ledger->stateMap().lower_bound(uint256(31))->key() ==
+                uint256(30));
```
thejohnfreeman, line 583: These lines are duplicates of the above.

src/test/ledger/View_test.cpp
thejohnfreeman: Not all of the assertions in the last test are necessary either, so that's not the solution I'm asking for. My guiding principle is this: what does this assertion tell us that no other assertion tells us? If an assertion provides no new information, then it is a duplicate and should be eliminated. Writing a test that is just like another test but bigger is not helpful. It adds more code to read, compile, run, and maintain without adding any value.

Here's what I would write, including the comment:

```cpp
// The tests here try to span a few categories:
// - When the given ID is in vs not in the SHAMap.
// - When the request is for lower vs upper bound.
// - When the requested ID is in vs not in the SHAMap.
// - When the given ID and requested ID are have the same parent inner node vs sibling parent inner nodes vs distant parent inner nodes.

setup({
    0,  1,  2,  3,  4,  5,  6,  7,  8,  9,  10, 11, 12, 13, 14, 15,
    16,
    32,
    256});
auto const& state = ledger->stateMap();
auto const& end = state.end();

BEAST_EXPECT(state.lower_bound(uint256(0)) == end);
BEAST_EXPECT(state.lower_bound(uint256(1)) == uint256(0));
BEAST_EXPECT(state.lower_bound(uint256(16)) == uint256(15));
BEAST_EXPECT(state.lower_bound(uint256(32)) == uint256(16));
BEAST_EXPECT(state.lower_bound(uint256(64)) == uint256(32));
BEAST_EXPECT(state.lower_bound(uint256(512)) == uint256(256));

BEAST_EXPECT(state.upper_bound(uint256(512)) == end);
BEAST_EXPECT(state.upper_bound(uint256(256)) == end);
BEAST_EXPECT(state.upper_bound(uint256(64)) == uint256(256));
BEAST_EXPECT(state.upper_bound(uint256(32)) == uint256(256));
BEAST_EXPECT(state.upper_bound(uint256(16)) == uint256(32));
BEAST_EXPECT(state.upper_bound(uint256(15)) == uint256(16));
BEAST_EXPECT(state.upper_bound(uint256(0)) == uint256(1));
```

src/test/ledger/View_test.cpp
thejohnfreeman, line 583: These lines are duplicates.

src/ripple/shamap/impl/SHAMap.cpp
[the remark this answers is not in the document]
cjcobb23: I feel like we are straying into bike shedding territory here. What is the tangible issue with the way the code is right now? Readability, maintainability, something else?

src/test/ledger/View_test.cpp
thejohnfreeman, line 562: These lines are duplicates.

src/ripple/shamap/impl/SHAMap.cpp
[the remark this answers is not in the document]
thejohnfreeman: I think and hope the team wants to converge on a single code style for readability and maintenance, so "bikeshedding" is a necessary part of that process. Am I alone in that?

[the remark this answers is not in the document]
cjcobb23: So what is the single code style that should be used here then?

src/test/ledger/View_test.cpp
cjcobb23: I would rather have too many tests than too little. If there is a case that we are missing, let me know, but I'm not really comfortable removing these tests that appear unnecessary. The SHAMap is not a trivial data structure. A bigger tree with more branches is different than a smaller tree with less, and I would prefer to test both cases, even if it seems like that's not necessary.

CMakeLists.txt
```diff
@@ -5,6 +5,7 @@ if (POLICY CMP0074)
 endif ()
 
 project (rippled)
+set (CMAKE_CXX_STANDARD 17)
```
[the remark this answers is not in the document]
legleux: This is what's in clio at the moment, but I have to modify it. Either way I want this to be set in as few places as possible preferably with the same syntax. 
```
set(CMAKE_CXX_EXTENSIONS OFF)
set(CMAKE_CXX_STANDARD 17)
set(CMAKE_CXX_STANDARD_REQUIRED ON)
```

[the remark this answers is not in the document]
cjcobb23: Added these

src/test/ledger/View_test.cpp
cjcobb23: removed

src/test/ledger/View_test.cpp
cjcobb23: removed

src/ripple/shamap/SHAMap.h
```diff
@@ -400,12 +408,31 @@ class SHAMap
...
         int branch = 0) const;
 
+    // returns the last item at or below this node
+    SHAMapLeafNode*
+    lastBelow(
+        std::shared_ptr<SHAMapTreeNode> node,
+        SharedPtrNodeStack& stack,
+        int branch = branchFactor) const;
+
+    // helper function for the above
```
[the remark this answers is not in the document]
cjcobb23: changed

PR #4092, opened 2022-02-02
fix bug in TaggedPointer move constructor

seelabs:
The !inSrc, !inDst case was not handled correctly. This case is when a caller
explicitly removes a child that's not part of the node. Fortunately, existing
callers of this code never used this case.


nbougalis (approved):

nbougalis:
Merged as 61389a8befd79104534f32275394ffb953196741

PR #4093, opened 2022-02-03
Don't dereference empty optional in ReportingETL constructor

cjcobb23:
An empty optional was being dereferenced in the ReportingETL constructor in certain configurations. Assign a value to it first so we can then dereference and assign a different value.


nbougalis:
Merged as 5b085a75fd4f2a28804f15f6559a5ecd4d862d81

PR #4097, opened 2022-02-10
Amendment: ExpandedSignerList

RichardAH:
An amendment __ExpandedSignerList__, which if accepted, would:
1. Expand the maximum signer list to 32 entries
2. Allow the optional specification of 32 byte tags with each signer list entry

### Context of Change
The motivation for this amendment is two fold:
1. Large organisations may have extensive signer lists which require more than 8 keys and may benefit from tags (for example to identify who is in possession of each key.)
2. Layer two smart contract solutions such as Evernode-hotpocket, codius and others gain a security benefit for a common use-case: 
- Often it is useful for layer 2 contracts to control layer 1 accounts, from which they send and receive tokens and currency. To do this, the servers that collectively run the layer 2 contract must hold private keys. The more keys required to sign a layer 1 transaction the harder it is for a third party to compromise the layer 1 account. 

# Remarks
This amendment PR includes a small but important refactor:
The _Rules_ class has been extracted from ReadView.h/cpp and placed in its own files (Rules.h/cpp). A Rules reference may now be more easily passed in the codebase, including for example to signature verification routines (which makes sense because signature verification can be altered by amendments).

At @nbougalis's suggestion serialised field `sfWalletLocator` was reused as the uint256 for the optional tag in each SignerListEntry.

The ability to add tags to SignerListEntries required a change to the InnerObjectFormats definition. This object definition is used both in *ltSIGNER_LIST* and in *ttSIGNER_LIST_SET*. Thus care must be taken at the SetSignerList transactor to ensure tags cannot be set before amendment activation.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Test Plan

MultSign_test.cpp has been updated with a new test for the tags and additional signers (up to 32).

In addition I have tested using xrpl.js to sign transactions that:
* Set signer list with and without tags
* Set signer list of 8 (success) and 9 (fail) entries when amendment is not active
* Set signer list of 8 (success), 9 (success), 32 (success) and 33 (fail) entries when amendment is active.

I have not included these tests anywhere, I encourage code reviewer to write own tests to avoid reliance on one set of tests.

For reference here is the JSON format for Signer Entries with tags:
```json
{                                                                                                      
    "SignerEntry": {                                                                                   
        "Account": "rsA2LpzuawewSBQXkiju3YQTMzW13pAAdW",                                               
        "SignerWeight": 2,                                                                             
        "WalletLocator": "CAFECAFECAFECAFECAFECAFECAFECAFECAFECAFECAFECAFECAFECAFECAFECAFE"            
    }                                                                                                  
}, 
```

### DIFF
--- a/src/ripple/app/tx/impl/SignerEntries.cpp
+++ b/src/ripple/app/tx/impl/SignerEntries.cpp
@@ -22,6 +22,7 @@
 #include <ripple/protocol/STArray.h>
 #include <ripple/protocol/STObject.h>
 #include <cstdint>
+#include <optional>
 
 namespace ripple {
 
@@ -41,7 +42,7 @@ SignerEntries::deserialize(
     }
 
     std::vector<SignerEntry> accountVec;
-    accountVec.reserve(STTx::maxMultiSigners);
+    accountVec.reserve(STTx::maxMultiSigners());
 
     STArray const& sEntries(obj.getFieldArray(sfSignerEntries));
     for (STObject const& sEntry : sEntries)
@@ -57,7 +58,9 @@ SignerEntries::deserialize(
         // Extract SignerEntry fields.
         AccountID const account = sEntry.getAccountID(sfAccount);
         std::uint16_t const weight = sEntry.getFieldU16(sfSignerWeight);
-        accountVec.emplace_back(account, weight);
+        std::optional<uint256> const tag = sEntry.at(~sfWalletLocator);
+
+        accountVec.emplace_back(account, weight, tag);
     }
     return accountVec;
 }
--- a/src/ripple/protocol/STTx.h
+++ b/src/ripple/protocol/STTx.h
@@ -21,7 +21,9 @@
 #define RIPPLE_PROTOCOL_STTX_H_INCLUDED
 
 #include <ripple/basics/Expected.h>
+#include <ripple/protocol/Feature.h>
 #include <ripple/protocol/PublicKey.h>
+#include <ripple/protocol/Rules.h>
 #include <ripple/protocol/STObject.h>
 #include <ripple/protocol/SecretKey.h>
 #include <ripple/protocol/SeqProxy.h>
@@ -47,7 +49,16 @@ class STTx final : public STObject, public CountedObject<STTx>
 
 public:
     static std::size_t const minMultiSigners = 1;
-    static std::size_t const maxMultiSigners = 8;
+
+    // if rules are not supplied then the largest possible value is returned
+    static std::size_t
+    maxMultiSigners(Rules const* rules = 0)
+    {
+        if (rules && !rules->enabled(featureExpandedSignerList))
+            return 8;
+
+        return 32;
+    }
 
     STTx() = delete;
     STTx(STTx const& other) = default;
@@ -108,7 +119,8 @@ public:
     */
     enum class RequireFullyCanonicalSig : bool { no, yes };
     Expected<void, std::string>
-    checkSign(RequireFullyCanonicalSig requireCanonicalSig) const;
+    checkSign(RequireFullyCanonicalSig requireCanonicalSig, Rules const& rules)
+        const;
 
     // SQL Functions with metadata.
     static std::string const&
@@ -130,7 +142,9 @@ private:
     checkSingleSign(RequireFullyCanonicalSig requireCanonicalSig) const;
 
     Expected<void, std::string>
-    checkMultiSign(RequireFullyCanonicalSig requireCanonicalSig) const;
+    checkMultiSign(
+        RequireFullyCanonicalSig requireCanonicalSig,
+        Rules const& rules) const;
 
     STBase*
     copy(std::size_t n, void* buf) const override;
--- a/src/ripple/protocol/impl/InnerObjectFormats.cpp
+++ b/src/ripple/protocol/impl/InnerObjectFormats.cpp
@@ -28,6 +28,7 @@ InnerObjectFormats::InnerObjectFormats()
         {
             {sfAccount, soeREQUIRED},
             {sfSignerWeight, soeREQUIRED},
+            {sfWalletLocator, soeOPTIONAL},
         });
 
     add(sfSigner.jsonName.c_str(),
--- a/src/test/jtx/impl/multisign.cpp
+++ b/src/test/jtx/impl/multisign.cpp
@@ -22,6 +22,8 @@
 #include <ripple/protocol/Sign.h>
 #include <ripple/protocol/UintTypes.h>
 #include <ripple/protocol/jss.h>
+#include <optional>
+#include <sstream>
 #include <test/jtx/multisign.h>
 #include <test/jtx/utility.h>
 
@@ -46,6 +48,8 @@ signers(
         auto& je = ja[i][sfSignerEntry.getJsonName()];
         je[jss::Account] = e.account.human();
         je[sfSignerWeight.getJsonName()] = e.weight;
+        if (e.tag)
+            je[sfWalletLocator.getJsonName()] = to_string(*e.tag);
     }
     return jv;
 }
--- a/Builds/CMake/RippledCore.cmake
+++ b/Builds/CMake/RippledCore.cmake
@@ -82,6 +82,7 @@ target_sources (xrpl_core PRIVATE
   src/ripple/protocol/impl/PublicKey.cpp
   src/ripple/protocol/impl/Quality.cpp
   src/ripple/protocol/impl/Rate2.cpp
+  src/ripple/protocol/impl/Rules.cpp
   src/ripple/protocol/impl/SField.cpp
   src/ripple/protocol/impl/SOTemplate.cpp
   src/ripple/protocol/impl/STAccount.cpp
@@ -209,6 +210,7 @@ install (
     src/ripple/protocol/PublicKey.h
     src/ripple/protocol/Quality.h
     src/ripple/protocol/Rate.h
+    src/ripple/protocol/Rules.h
     src/ripple/protocol/SField.h
     src/ripple/protocol/SOTemplate.h
     src/ripple/protocol/STAccount.h
--- a/src/ripple/app/consensus/RCLConsensus.cpp
+++ b/src/ripple/app/consensus/RCLConsensus.cpp
@@ -632,7 +632,7 @@ RCLConsensus::Adaptor::doAccept(
         auto const lastVal = ledgerMaster_.getValidatedLedger();
         std::optional<Rules> rules;
         if (lastVal)
-            rules.emplace(*lastVal, app_.config().features);
+            rules = makeRulesGivenLedger(*lastVal, app_.config().features);
         else
             rules.emplace(app_.config().features);
         app_.openLedger().accept(
--- a/src/ripple/app/ledger/Ledger.cpp
+++ b/src/ripple/app/ledger/Ledger.cpp
@@ -626,7 +626,7 @@ Ledger::setup(Config const& config)
 
     try
     {
-        rules_ = Rules(*this, config.features);
+        rules_ = makeRulesGivenLedger(*this, config.features);
     }
     catch (SHAMapMissingNode const&)
     {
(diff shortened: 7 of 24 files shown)


RichardAH:
updated: ran levelization and patched inconsistency between clang-format-10 (CI) and 13 (my version)

src/ripple/ledger/impl/Rules.cpp
```diff
@@ -0,0 +1,127 @@
...
+    ACTION  OF  CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+    OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+//==============================================================================
+
+#include <ripple/ledger/Rules.h>
+
+namespace ripple {
+
+class Rules::Impl
```
nbougalis, line 24: I know you just moved this here and it was already present, but we should think about migrating away from the PIMPL idiom. It adds nothing, in this case.

src/ripple/protocol/impl/InnerObjectFormats.cpp
```diff
@@ -28,6 +28,7 @@ InnerObjectFormats::InnerObjectFormats()
         {
             {sfAccount, soeREQUIRED},
             {sfSignerWeight, soeREQUIRED},
+            {sfWalletLocator, soeOPTIONAL},
```
nbougalis, line 31: I understand why you want this field and concede that it's useful, but it may be slightly controversial in terms of size.

So is a 256-bit value _really_ needed? Could we get away with something smaller? I think your original suggestion was to use a 64-bit tag. I suspect that would be fine, but maybe a little too constrained for what you want to do. 128 bits is probably sufficient, especially if you use it as an [RFC-4122 UUID](https://datatracker.ietf.org/doc/html/rfc4122) (likely V1, V4 or V5, depending on the usecase).

I am not _overly_ concerned and I think it's fine as is, but I do think it's worth at least having a discussion about this to ensure the field is "right-sized".

nbougalis (approved):
Looks fine.

I don't think that increasing the number of signers is likely to have a significant impact on the ledger.

Left a comment about the `sfWalletLocator` field, and would like to see others chime in on this too.

trueIoV:
Hi RichardAH, why not increase the number of signers to 100 in this case?
Especially since KAVA once stated that it could not integrate XRPL precisely because they need 100 signatures and no less.
Thanks.

RichardAH:
@trueIoV 128 is possible but there should be some debate about the "right" number. Too many is an attack vector as follows:
Suppose we allow 65535 signatures and someone continually submits txns with 65534 correct signatures and 1 incorrect signature at the end. The nodes they send these txns to have to check all 65535 signatures and then discard the txn without billing any account on-ledger. This vector could be mitigated in several ways, one of which is to allow a fee claim failure to occur on any multisig transaction with more than X correct signatures, where X is an absolute number not relative to the total.

src/ripple/protocol/impl/InnerObjectFormats.cpp
RichardAH: Reusing an existing (essentially unused) serialized field is great because it makes it backwards compatible with old codecs. So if someone proposes to change the size of the tag they should also propose an existing sfField that matches their proposed size.

src/ripple/ledger/impl/Rules.cpp
RichardAH: Agree, but changing that was out of scope for this amendment (and in fact doesn't require an amendment).

nbougalis:
> Hi RichardAH, why not increase the number of signers to 100 in this case? Especially since KAVA once stated that it could not integrate XRPL precisely because they need 100 signatures and no less. Thanks.

I don't know what Kava's rationale for "100 signatures and no less" is, but it smells a little like snake-oil; absent an _actual_, _**demonstrable**_ need for such a high number of signatures and a _rational_ explanation for why alternative solutions (e.g. Ed25519 threshold signatures) aren't sufficient, I don't think we should expand the signer list by that much.

Engineering is about trade-offs and especially in the case of a system like the XRPL I think we should be frugal in the 'cost' that participants are forced to bear in terms of computing resources, in the absence of a compelling reason for a feature. If that means some entity can't integrate... well, that's too bad.

ximinez (changes requested):
I took a quick look at these changes, and I don't have any feedback about the logic, but I did leave some comments with changes that will improve code safety, clarity, and readability.

src/ripple/app/tx/impl/SignerEntries.cpp
```diff
@@ -57,7 +59,12 @@ SignerEntries::deserialize(
         // Extract SignerEntry fields.
         AccountID const account = sEntry.getAccountID(sfAccount);
         std::uint16_t const weight = sEntry.getFieldU16(sfSignerWeight);
-        accountVec.emplace_back(account, weight);
+        std::optional<uint256> tag;
+
+        if (sEntry.isFieldPresent(sfWalletLocator))
+            tag = sEntry.getFieldH256(sfWalletLocator);
```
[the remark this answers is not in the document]
RichardAH: Noted thanks

src/ripple/protocol/STTx.h
```diff
@@ -44,7 +46,16 @@ class STTx final : public STObject, public CountedObject<STTx>
...
-    static std::size_t const maxMultiSigners = 8;
+
+    // if rules are not supplied then the largest possible value is returned
+    static std::size_t
+    maxMultiSigners(Rules const* rules = 0)
+    {
+        if (rules && !rules->enabled(featureExpandedSignerList))
+            return 8;
+
+        return 32;
```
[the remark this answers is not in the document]
RichardAH: It’s written in this way to make it more explicit that different sets of rules can produce different results. Ternary operator is less clear especially if someone adds a third possibility later. I will leave it how I have it for clarity

src/test/jtx/impl/multisign.cpp
```diff
@@ -46,6 +48,12 @@ signers(
         auto& je = ja[i][sfSignerEntry.getJsonName()];
         je[jss::Account] = e.account.human();
         je[sfSignerWeight.getJsonName()] = e.weight;
+        if (e.tag)
+        {
+            std::stringstream ss;
+            ss << *(e.tag);
+            je[sfWalletLocator.getJsonName()] = ss.str();
```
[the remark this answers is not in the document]
RichardAH: Will change this thanks!

RichardAH:
I’ll make these changes today and push an update

nbougalis:
@RichardAH: I would suggest rethinking the 256-bit wallet locator value. Again, I'm not opposed to it, but can we get away with a 128-bit value? If you want to use it as a ledger locator, then the answer is no, but then `WalletLocator` isn't a good fit either. `UserData` or somesuch is probably a better fit.

I also think that some exposition, informing the motivation or possible uses behind this field might be helpful. I know you have specific use cases in mind, but others may not know what they are and that may make it more difficult.

RichardAH:
> @RichardAH: I would suggest rethinking the 256-bit wallet locator value. Again, I'm not opposed to it, but can we get away with a 128-bit value? If you want to use it as a ledger locator, then the answer is no, but then `WalletLocator` isn't a good fit either. `UserData` or somesuch is probably a better fit.
> 
> I also think that some exposition, informing the motivation or possible uses behind this field might be helpful. I know you have specific use cases in mind, but others may not know what they are and that may make it more difficult.

I have been thinking about whether or not 32 bytes of key identifier is the “right amount” and let me make the case why it is now:

In a large organisation such as a university or government department, assets including HSMs are typically assigned a 128 bit uuid which is printed on the asset. Good, but that only tells you which asset not where to find the asset. Leaving some extra bytes free to describe a place and building plus the asset number would make sense.

In the case of a layer two network which is trying to use the tag to find which servers in its cluster can sign: anything less than 32 bytes basically precludes using ipv6.  It is the same problem as above. You could fit exactly one ipv6 address into a 128 bit field with nothing left over for extra information (like for example a port or protocol).

For these reasons I think the we should not reduce the size of the tag. If you are concerned about the data storage then we should instead increase ownercount per tag

wojake:
I have a question related to sidechains, since door accounts can now have ~32 federators participating after this amendment is passed, would it be appropriate to create a new type of account for door accounts, perhaps we could create a new setting for this? 

Rationale: Users are able to confidently indicate if an account is a door account, this would avoid any confusion or doubts as it won't look like a 'normal' account.

RichardAH:
@wojake if anyone can create and mark an account in such a way how does it help users verify anything?

wojake:
Yikes, I didn't think about that (my apologies). Currently, there's no way to verify if an account is a door account, so this is pointless if everyone can mark their account as a door account.
Unless there's a way to verify if an account is a door account on-chain, I don't think my proposal makes sense.

src/ripple/protocol/impl/InnerObjectFormats.cpp
scottschurr: In my opinion reusing a pre-existing field is, in this case, probably a false economy.  This commit will be bound together with many others in a release.  At least one of the other commits in the release will likely change the codec.  So having this particular commit avoid changing the codec does not prevent the codec from changing in a release.

Most programmers agree that names are hard.  Giving a field the wrong name or a misleading name will only make the XRP Ledger even harder to use correctly.

My personal opinion is that you should pick the right sized `SField` type and select a truly meaningful name for the field.  Proceed with that.  If you are truly sold that `WalletLocator` is the best name you can come up with then I'd love to hear that explanation.  Thanks.

scottschurr (changes requested):
There are some problems in this pull request that should be addressed before it is accepted.

1. The `Rules` changes damage levelization.  That needs to be fixed.  Here is a commit that shows a possible solution: https://github.com/scottschurr/rippled/commit/edf3aa5dec316c477223cb67fd5dc4fdb51accdc. Feel free to cherry-pick that commit if you would like to use it.

2. This commit causes the KnownFormatsToGRPC unit test to fail.  The right way to fix that is to update the GRPC serialization code.  Here is a commit that fixes the problem and makes the unit test pass: https://github.com/scottschurr/rippled/commit/6fd72f6f05160c80ff5bd1ff3a1981fd72e242c8  Feel free to cherry-pick that commit if you would like.

3. With this pull request the MultiSign unit test no longer tests the pre-ExtendedSignerList functionality.  The MultiSign unit test should be modified so it tests the old functionality at least until the ExpandedSignerList amendment has been enabled on the network.  Here's a commit showing one way to achieve that change: https://github.com/scottschurr/rippled/commit/b176359e658b8840405b7cd3ca80c12d2fe75887  Feel free to cherry-pick that commit if you would like.

I have an additional concern that is, however, not a blocker.  My suspicion is that reusing the `WalletLocator` locator `SField` is a false economy.  I think you are better off if you carefully pick a name that represents your intended use and then introduce that as a new `SField` with the _right_ name.  If you can argue that `WalletLocator` is really a good name for the intended use then I can go with that.

RichardAH:
Thanks for the review. I went ahead and squashed your commits into mine and added you as a co-author (seemed like the right approach to avoid too many commits for such a small amendment)

RichardAH:
> I have an additional concern that is, however, not a blocker. My suspicion is that reusing the `WalletLocator` locator `SField` is a false economy. I think you are better off if you carefully pick a name that represents your intended use and then introduce that as a new `SField` with the _right_ name. If you can argue that `WalletLocator` is really a good name for the intended use then I can go with that.

I somewhat agree with you on this. The main advantage to its reuse is that old codecs can serialize and unserialize it. Which means legacy multisign tools can be trivially updated to use tags. I also do think `WalletLocator` is not an entirely inaccurate description. `Tag` is much less descriptive, so probably we would end up with something like `KeyDescription` or `KeyLocation`. At the end of the day it's a toss-up between nice backwards compatibility with existing codecs and a slightly incongruous name. I would lean toward keeping WalletLocator. I'd also add this is a niche feature, and almost no one will use or see it (relative to general use of the XRPL) so being pretty isn't important.

scottschurr (approved):
@RichardAH, thanks for your explanation for why you prefer `WalletLocator` for the added `SField`.  I can go with that motivation.  BTW,  it was nice of you to flag me as a coauthor, but really not necessary.  I consider my contribution to simply be part of the code review process.

Changes look good.  The CI failures that I see are the result of ongoing CI problems, not a result of this pull request.

ximinez (approved):
Looks good!

RichardAH:
(Rebased on most recent develop)

PR #4099, opened 2022-02-12
Incremental improvements to path finding memory usage

ximinez:

As the number of trust lines on the ledger has exploded, so has the amount of memory used by servers for path finding.

Some particularly gnarly requests have been observed to load over 20 million trust lines into memory. These changes build on #4080, which reduces the amount of memory used per trust line, so reviewers only need to review the commit(s) starting at "Incremental improvements to path finding memory usage".

Background path finding jobs can continue after a websocket client has created another request, or disconnected entirely. This can cause a lot of memory to be used unnecessarily while other requests are using even more memory. The main changes in this PR free up the relevant memory more quickly when the request is no longer relevant. 

### Type of Change

## Test Plan

Run an expensive `path_find` request. Wait until it returns a `full_reply: true` result, indicating that the background job has started. Invalidate the request using one of the following methods:
* Disconnect the client.
* Issue a `{ command: path_find, subcommand: close }` to close the request.
* Issue a new `path_find` request on the same connection.

With version of code before this PR, the path request can be observed to continue running by tracking process memory, or using the `get_counts` command to see the number of `PathFindTrustLine` continue to grow. (Note that the counter for `PathFindTrustLine` was only added in #4080.) With these changes, the memory usage will drop within about a few seconds to a minute, and the `PathFindTrustLine` count will drop to 0 for all cases but the last.

## Future Tasks

Now that the trust line memory is getting freed up more quickly, I'm working on preventing as many trust lines and other objects from being loaded in the first place.

### DIFF
--- a/src/ripple/protocol/impl/BuildInfo.cpp
+++ b/src/ripple/protocol/impl/BuildInfo.cpp
@@ -33,7 +33,7 @@ namespace BuildInfo {
 //  and follow the format described at http://semver.org/
 //------------------------------------------------------------------------------
 // clang-format off
-char const* const versionString = "1.9.0-b1"
+char const* const versionString = "1.9.0-b2"
 // clang-format on
 
 #if defined(DEBUG) || defined(SANITIZER)


src/ripple/app/paths/PathRequests.cpp
```diff
@@ -40,18 +40,28 @@ PathRequests::getLineCache(
...
         (authoritative &&
          ((lgrSeq + 8) < lineSeq)) ||  // we jumped way back for some reason
         (lgrSeq > (lineSeq + 8)))      // we jumped way forward for some reason
     {
-        mLineCache = std::make_shared<RippleLineCache>(ledger);
+        JLOG(mJournal.debug())
+            << "getLineCache creating new cache for " << lgrSeq;
+        // Assign to the local before the member, because the member is a
+        // weak_ptr, and will immediately discard it if there are no other
+        // references.
```
seelabs, line 60: Nice comment. Thanks for that.

src/ripple/app/paths/PathRequests.cpp
```diff
@@ -80,15 +90,28 @@ PathRequests::updateAll(
 
     int processed = 0, removed = 0;
 
+    auto getSubscriber = [](PathRequest::pointer& request) -> InfoSub::pointer {
```
seelabs, line 93: Param could be `const&`

src/ripple/app/paths/PathRequests.cpp
```diff
@@ -97,17 +120,26 @@ PathRequests::updateAll(
...
-                            Json::Value update =
-                                request->doUpdate(cache, false);
+                            // Release the shared ptr to the subscriber so that
+                            // it can be freed if the client disconnects, and
+                            // thus fail to lock later.
+                            ipSub.reset();
+                            Json::Value update = request->doUpdate(
+                                cache, false, [&getSubscriber, &request]() {
+                                    return (bool)getSubscriber(request);
+                                });
```
seelabs, line 134: It's not obvious what that lambda is doing. Maybe make a local variable for the callback just for the purpose of nameing it? I.e.
```
auto conintueCallback = [&getSubscriber, &request]()...
```

src/ripple/app/paths/Pathfinder.cpp
```diff
@@ -157,6 +157,20 @@ smallestUsefulAmount(STAmount const& amount, int maxPaths)
...
+{
+    auto iter = v.begin();
+    if (iter != v.end())
+    {
+        s << *iter;
+        for (++iter; iter != v.end(); ++iter)
+            s << ", " << *iter;
+    }
+    return s;
+}
```
seelabs, line 172: This kind of iterator comes up all the time - I wish someone would add it to boost or std. This is the best I could find:
https://en.cppreference.com/w/cpp/experimental/ostream_joiner
but that's experimental, I'd rather not use that.

seelabs (commented):
Looks good. Left some minor comments, but nothing big.

src/ripple/app/paths/PathRequest.h
```diff
@@ -91,13 +89,18 @@ class PathRequest : public std::enable_shared_from_this<PathRequest>,
...
+    void
+    doAborting() const;
 
     // update jvStatus
     Json::Value
-    doUpdate(std::shared_ptr<RippleLineCache> const&, bool fast);
+    doUpdate(
+        std::shared_ptr<RippleLineCache> const&,
+        bool fast,
+        std::function<bool(void)> continueCallback = {});
```
seelabs, line 103: Consider passing `function` by `const&`. Here's a patch to do that if you want: https://github.com/seelabs/rippled/commit/97ede1ef2530d6dd4fba87bb6a5ef43cc02cf5ac

src/ripple/app/paths/PathRequests.cpp
ximinez: Fixed

src/ripple/app/paths/PathRequests.cpp
ximinez: Done.

src/ripple/app/paths/PathRequest.h
ximinez: Will do.

src/ripple/app/paths/Pathfinder.cpp
ximinez: I couldn't help myself, so I wrote something to do it.

mtrippled (approved):
:+1: LGTM

faves4less (approved):

src/ripple/basics/join.h
```diff
@@ -0,0 +1,109 @@
...
+        s << delimiter << *iter;
+    return s;
+}
+
+template <class Collection, class Str>
+class CollectionAndDelimiter
+{
+public:
+    Collection const& collection;
+    Str const& delimiter;
```
seelabs, line 44: The lifetime management of `delimiter` gives me pause. In the common case - passing a statically allocated c string - we're fine. But this is a very general class and I could see `delimiter` pointing to some temporary that went out of scope. Since `delimiters` are going to be small (usually one or two characters) and `std::string` has a small-object optimization (no allocations for small strings), what do you think about using a `std::string` for the delimiter instead?

FWIW, I'm _not_ concerned about the lifetime management of `collection`. The caller should expect this to keep a reference to the object. 

I saw in your tests that you were using integers and uint256 as crazy delimiters. But I think just supporting strings is enough (if you have a crazy delimiter, convert it to a string first). Here's a patch if you want to see what this looks like: https://github.com/seelabs/rippled/commit/2bf76479b11f714086726be7d6c639d3f827cc92

seelabs (commented):
I think this looks good. And thanks for coding up the `join` utility. That's awesome!

I did have a concern about lifetime management of the delimiter. Take a look and see what you think.

src/ripple/basics/join.h
```diff
@@ -44,7 +42,7 @@ class CollectionAndDelimiter
     Collection const& collection;
     std::string const delimiter;
 
-    explicit CollectionAndDelimiter(Collection const& c, std::string delim)
+    explicit CollectionAndDelimiter(Collection const& c, std::string&& delim)
```
seelabs, line 45: I would probably leave `delim` as a value, not an rvalue. The common use case of passing a cstring will act the same. It will act differently if someone already has a `delim` as a string that they want to pass in. Making this a rvalue might have a performance benefit - it forces the caller to decide if they need to copy the delimiter or if they can just move it. But it also make the interface a little harder to use. For a class like this I'd choose to make the interface easier to use. 

I'm fine with it as-is too. But thought I'd make the note.

seelabs (approved):
:+1: I'm approving this now, but I did leave a comment on forcing the `delim` to be an rvalue ref. I'd leave that as a value parameter to keep the interface easier to use.

src/ripple/basics/join.h
ximinez: Yeah, that makes sense. I'll change it back.

ximinez:
Marked "passed" and squashed.

PR #4101, opened 2022-02-15
Introduce NFT support (XLS-20)

scottschurr:

Support for XLS-20 NFTs on the XRP Ledger.

### Context of Change

If integrated, this branch provides native NFT support on the XRP Ledger.

Please see https://github.com/XRPLF/XRPL-Standards/discussions/46 for the spec and further details.

### Type of Change

- [x] New feature on amendment.  So enabling the amendment will amendment block older versions of the software.
- [x] Tests for NFTs are added.


RichardAH:
Renaming global types (in this case the serialized field types) inside an amendment commit is inadvisable. I realise this amendment introduces new types but renaming existing types inside it breaks any number of other backlogged PRs (including probably mine) which all need to be individually updated to be merge-compatible again.

Instead change the code to continue with the existing naming convention and do the rename in a separate non-amendment PR after backlog has been cleared. That way we can update type names once, "one and done" so to speak, without needlessly multiplying work.

I'm speaking about `SF_HASH256`-> `SF_UINT256` and so on. I'm _not_ saying this rename shouldn't be done, but I am saying it shouldn't be done _here_.

I think as a rule amendment PR's should not change anything that isn't explicitly and exactly needed for that PR to function. There should be a clear delineation between writing new functionality into the codebase and refactoring existing functionality for the reasons above (multiplication of work)

src/ripple/app/tx/impl/details/NFTokenUtils.cpp
```diff
@@ -0,0 +1,498 @@
...
+
+    createCallback(view, owner);
+
+    return (first.key <= np->key()) ? np : cp;
+}
+
+static bool
+compareTokens(uint256 const& a, uint256 const& b)
+{
+    return (a & nft::pageMask) < (b & nft::pageMask);
```
scottschurr, line 180: I think there's a latent bug here.  We need every server to produce identically sorted `NFTPage`s.  But this comparison only sorts on the low 96-bits.  If there are NFTs with identical low 96-bits then the sort order becomes nondeterministic.

So I think this function needs to look more like this:
```
    // The sort of NFTokens needs to be fully deterministic, but the sort
    // is weird because we sort on the low 96-bits first.  But if the low
    // 96-bits are identical we still need a fully deterministic sort.
    // So we sort on the low 96-bits first.  If those are equal we sort on
    // the whole thing.
    int const lowBitsCmp = compare(a & nft::pageMask, b & nft::pageMask);
    if (lowBitsCmp != 0)
        return lowBitsCmp < 0;

    return a < b;
```

nbougalis:
> Renaming global types (in this case the serialized field types) inside an amendment commit is inadvisable. I realise this amendment introduces new types but renaming existing types inside it breaks any number of other backlogged PRs (including probably mine) which all need to be individually updated to be merge-compatible again.

I agree with you, and in a perfect world this wouldn't have happened. My original set of commits had a separate commit for this, but somewhere along the way I squashed it down into a single commit.

> Instead change the code to continue with the existing naming convention and do the rename in a separate non-amendment PR after backlog has been cleared. That way we can update type names once, "one and done" so to speak, without needlessly multiplying work.
> 
> I'm speaking about `SF_HASH256`-> `SF_UINT256` and so on. I'm _not_ saying this rename shouldn't be done, but I am saying it shouldn't be done _here_.

I _could_ unsquash it but I don't know if it's worth the trouble to do this.  Rationale: the renaming here doesn't alter the binary protocol; it's a purely mechanical change and any conflicts with other PRs are likely to be very simple to resolve.

> I think as a rule amendment PR's should not change anything that isn't explicitly and exactly needed for that amendment to function. There should be a clear delineation between writing new functionality into the codebase and refactoring existing functionality for the reasons above (multiplication of work)

I think that going forward, it's a good rule for contributors and maintainers to adopt: changes to global types should be a standalone commit. I also agree with your rationale about some mechanism to avoid conflict between different PRs that introduce new fields and/or types.

While I don't think we need to do this for this commit, if you feel strongly enough that this should be done here, I'll put in the work to do it. Let me know.

src/ripple/protocol/Protocol.h
```diff
@@ -53,6 +57,35 @@ std::size_t constexpr dirNodeMaxEntries = 32;
...
+    a basis point; that is a value of 1000 represents a transfer fee of 1% and
+    a value of 10000 represents a transfer fee of 10%.
+
+    Note that for extremely low transfer fees values, it is possible that the
+    calculated fee will be 0.
+ */
+std::uint16_t constexpr maxTransferFee = 50000;
+
+/** The maximum length of a URI inside an NFT */
+std::size_t constexpr maxTokenURILength = 512;
```
scottschurr, line 84: This `512` limit does not match the spec or the documentation, which sets the limit at `256`.  This value should be changed to `256`.

src/ripple/ledger/View.h
```diff
@@ -110,12 +134,41 @@ forEachItem(
...
     std::function<bool(std::shared_ptr<SLE const> const&)> f);
 
+/** Iterate all items in an account's owner directory. */
+inline void
+forEachItem(
+    ReadView const& view,
+    AccountID const& id,
+    std::function<void(std::shared_ptr<SLE const> const&)> f)
+{
+    return forEachItem(view, keylet::ownerDir(id), std::move(f));
```
seelabs, line 150: No need to move once we change the parameter type to `const&` in the `forEach` functions.

src/ripple/ledger/View.h
```diff
@@ -110,12 +134,41 @@ forEachItem(
...
+forEachItemAfter(
+    ReadView const& view,
+    AccountID const& id,
+    uint256 const& after,
+    std::uint64_t const hint,
+    unsigned int limit,
+    std::function<bool(std::shared_ptr<SLE const> const&)> f)
+{
+    return forEachItemAfter(
+        view, keylet::ownerDir(id), after, hint, limit, std::move(f));
```
seelabs, line 169: No need to move once we change the parameter type to `const&`

src/ripple/protocol/SField.h
```diff
@@ -150,7 +153,8 @@ class SField
         const char* fn,
         int meta = sMD_Default,
         IsSigning signing = IsSigning::yes);
-    explicit SField(private_access_tag_t, int fc);
+    explicit SField(private_access_tag_t, int fc) = delete;
```
seelabs, line 156: Remove this constructor rather than marking it `delete`.

src/ripple/protocol/TER.h
```diff
@@ -114,11 +114,14 @@ enum TEMcodes : TERUnderlyingType {
     temINVALID_ACCOUNT_ID,
     temCANNOT_PREAUTH_SELF,
     temINVALID_COUNT,
+    temBAD_TRANSFER_FEE,
+    temTOKEN_IS_NONFLAMMABLE,
```
seelabs, line 118: Nice name

src/ripple/protocol/TER.h
```diff
@@ -278,7 +282,17 @@ enum TECcodes : TERUnderlyingType {
...
     tecTOO_SOON = 152,
-    tecHOOK_ERROR [[maybe_unused]] = 153
+    tecHOOK_ERROR [[maybe_unused]] = 153,
+    tecMAX_SEQUENCE_REACHED = 154,
+    tecNO_SUITABLE_PAGE = 155,
+    tecBUY_SELL_MISMATCH = 156,
+    tecOFFER_TYPE_MISMATCH = 157,
+    tecCANT_ACCEPT_OWN_OFFER = 158,
+    tecINSUFFICIENT_FUNDS = 159,
+    tecOBJECT_NOT_FOUND = 160,
```
seelabs, line 292: This could be `tecNO_NFT_OFFER` rather than the more generic `OBJECT_NOT_FOUND`. But I don't object, I'm fine either way. I'm making a note because we already have errors like `tecNO_LINE` and `tecNO_DST` that are specific to the missing objects.

src/ripple/protocol/impl/Indexes.cpp
```diff
@@ -20,9 +20,11 @@
 #include <ripple/protocol/Indexes.h>
 #include <ripple/protocol/SeqProxy.h>
 #include <ripple/protocol/digest.h>
+#include <ripple/protocol/nftPageMask.h>
 #include <boost/endian/conversion.hpp>
 #include <algorithm>
 #include <cassert>
+#include <string_view>
```
seelabs, line 27: `string_view` is not needed

src/ripple/protocol/impl/Rate2.cpp
```diff
@@ -34,6 +34,12 @@ as_amount(Rate const& rate)
 
 }  // namespace detail
 
+Rate
+transferFeeAsRate(std::uint16_t fee)
```
seelabs, line 38: This is too easily confused with IOU transfer fees. I'd rename this `nftTransferFeeAsRate` or put it in an nft namespace.

src/ripple/protocol/impl/SField.cpp
```diff
@@ -93,6 +93,7 @@ CONSTRUCT_TYPED_SFIELD(sfUNLModifyDisabling,    "UNLModifyDisabling",   UINT8,
 CONSTRUCT_TYPED_SFIELD(sfLedgerEntryType,       "LedgerEntryType",      UINT16,     1, SField::sMD_Never);
 CONSTRUCT_TYPED_SFIELD(sfTransactionType,       "TransactionType",      UINT16,     2);
 CONSTRUCT_TYPED_SFIELD(sfSignerWeight,          "SignerWeight",         UINT16,     3);
+CONSTRUCT_TYPED_SFIELD(sfTransferFee,           "TransferFee",          UINT16,     4);
```
seelabs, line 96: rename to NFTTransferFee

src/ripple/protocol/impl/SField.cpp
```diff
@@ -139,6 +140,9 @@ CONSTRUCT_TYPED_SFIELD(sfSignerListID,          "SignerListID",         UINT32,
 CONSTRUCT_TYPED_SFIELD(sfSettleDelay,           "SettleDelay",          UINT32,    39);
 CONSTRUCT_TYPED_SFIELD(sfTicketCount,           "TicketCount",          UINT32,    40);
 CONSTRUCT_TYPED_SFIELD(sfTicketSequence,        "TicketSequence",       UINT32,    41);
+CONSTRUCT_TYPED_SFIELD(sfTokenTaxon,            "TokenTaxon",           UINT32,    42);
```
seelabs, line 143: I wouldn't use "token" for nft. That's too generic a term for what NFTs are. I'd rename this `NFTTaxon`.

We should have consistent nomenclature for the new NFT functionality. I'd use `nft` everywhere. It's short and won't be confused with other functionality. Consider renaming all the `NFToken` files and functions to `nft` and all the `token` functions that are nft specific to `nft`.

Consider renaming the other nft functions - like `minter` and `OfferNode` so they either live in an `nft` namespace or have an `nft` prefix. `minter` is less confusing than `OfferNode` or `transferFee`, but I'd change them all.

src/ripple/protocol/impl/SField.cpp
```diff
@@ -305,7 +320,19 @@ SField::SField(
...
+//    : fieldCode(fc)
+//    , fieldType(STI_UNKNOWN)
+//    , fieldValue(0)
+//    , fieldMeta(sMD_Never)
+//    , fieldNum(++num)
+//    , signingField(IsSigning::yes)
+//    , jsonName(fieldName.c_str())
+//{
+//    knownCodeToField[fieldCode] = this;
+//}
```
seelabs, line 333: rm dead code

src/ripple/protocol/impl/STParsedJSON.cpp
```diff
@@ -860,8 +860,9 @@ parseObject(
 
         return data;
     }
-    catch (STObject::FieldErr const&)
+    catch (STObject::FieldErr const& xxx)
```
seelabs, line 863: rename `xxx` -> `e`

src/ripple/protocol/impl/TER.cpp
```diff
@@ -80,27 +80,39 @@ transResults()
...
-        MAKE_ERROR(tefWRONG_PRIOR,            "This previous transaction does not match."),
-        MAKE_ERROR(tefBAD_AUTH_MASTER,        "Auth for unclaimed account needs correct master key."),
-        MAKE_ERROR(tefINVARIANT_FAILED,       "Fee claim violated invariants for the transaction."),
-        MAKE_ERROR(tefTOO_BIG,                "Transaction affects too many items."),
-        MAKE_ERROR(tefNO_TICKET,              "Ticket is not in ledger."),
+        MAKE_ERROR(tecMAX_SEQUENCE_REACHED,   "The maximum sequence number was reached."),
+        MAKE_ERROR(tecNO_SUITABLE_PAGE,       "A suitable page could not be located."),
+        MAKE_ERROR(tecBUY_SELL_MISMATCH,      "The 'Buy' and 'Sell' offers are mismatched."),
+        MAKE_ERROR(tecOFFER_TYPE_MISMATCH,    "The type of offer is incorrect."),
+        MAKE_ERROR(tecCANT_ACCEPT_OWN_OFFER,  "An offer cannot be claimed by its owner."),
```
seelabs, line 87: Preface with `nft` or too easily confused with payment engine offers (previous three values at least)

src/ripple/protocol/jss.h
```diff
@@ -551,6 +564,7 @@ JSS(time);
 JSS(timeouts);                // out: InboundLedger
 JSS(track);                   // out: PeerImp
 JSS(traffic);                 // out: Overlay
+JSS(tokenid);                 // in: nft_sell_offers, nft_buy_offers
```
seelabs, line 567: rename to `nftid`

src/ripple/protocol/nftPageMask.h
```diff
@@ -0,0 +1,37 @@
+//------------------------------------------------------------------------------
```
seelabs, line 1: Style nit: Rename file to `NFTPageMask.h`

src/ripple/rpc/impl/GRPCHelpers.cpp
```diff
@@ -770,6 +770,14 @@ populateIndexes(T& to, STObject const& from)
     populateProtoVec256([&to]() { return to.add_indexes(); }, from, sfIndexes);
 }
 
+template <class T>
+void
+populateTokenOffers(T& to, STObject const& from)
```
seelabs, line 775: Rename to `populateNFTOffers`. OK, I'll stop adding comment for functions to rename. There's a general comment of name nft things "nft" and don't use "token" or make the function too generic sounding.

src/test/app/NFToken_test.cpp
```diff
@@ -0,0 +1,4765 @@
...
+    void
+    run() override
+    {
+        using namespace test::jtx;
+        auto const sa = supported_amendments();
+        testWithFeats(sa);
+    }
+};
+
+BEAST_DEFINE_TESTSUITE_PRIO(NFToken, tx, ripple, 3);
```
seelabs, line 4763: A 5000 like file is pretty big. We might want to split this up.

src/test/jtx/impl/Account.cpp
```diff
@@ -46,14 +46,25 @@ Account::Account(
 }
 
 Account
-Account::fromCache(std::string name, KeyType type)
+Account::fromCache(AcctStringType stringType, std::string name, KeyType type)
 {
     auto p = std::make_pair(name, type);  // non-const so it can be moved from
```
seelabs, line 51: I'll note that this allocates, which isn't great for a function that's meant to be a cache. But not new code, so no change needed. But we may want to clean this up in another PR.

src/test/jtx/token.h
```diff
@@ -0,0 +1,231 @@
...
+mint(jtx::Account const& account, std::uint32_t tokenTaxon = 0);
+
+/** Sets the optional TransferFee on an NFTokenMint. */
+class xferFee
+{
+private:
+    std::uint16_t xferFee_;
+
+public:
+    xferFee(std::uint16_t fee) : xferFee_(fee)
```
seelabs, line 48: `explicit` (the other constructors in this file should be `explicit` as well)

src/test/jtx/token.h
```diff
@@ -0,0 +1,231 @@
...
+
+    void
+    operator()(Env&, JTx& jtx) const;
+};
+
+/** Sets the optional Issuer on an NFTokenMint. */
+class issuer
+{
+private:
+    std::string issuer_;
```
seelabs, line 60: I see we're storing string representation of the wrapped object rather then the object itself (here and in the other classes). I'd probably store the object - they are 20-byte object, there's not much cost to copying them and storing them. But it's also fine as-is if you prefer it.

src/test/jtx/token.h
```diff
@@ -0,0 +1,231 @@
...
+
+public:
+    issuer(jtx::Account const& issue) : issuer_(issue.human())
```
seelabs, line 63: `explicit`

seelabs (commented):
Great job on this! Lots of tests and the code is really well done. I left some comments. The biggest things I'd like to tackle are:

1) Naming. I'd like to have consistent nomenclature for this new functionality. I prefer `nft`, not `token` or `NFToken`. I'd also prefer to rename the nft functionality that overlaps with existing functionality - like transfer fees and offers.

2) IOU transfer fees should probably be paid by the seller, not the buyer.

3) One "typo" bug in "invariants.

4) The rest is the usual collection of minor issues.

src/ripple/protocol/impl/SField.cpp
scottschurr: First, @seelabs, I really appreciate having your eyes on the pull request.  You've spotted a bunch of good things to change.  Thanks.

I'm fine with doing the renames, but I need confirmation from @nikb that he's okay with it before I proceed.

I'm thinking that I'll split the update into two commits.  One where I fix the bugs and non-naming improvements that are identified.  Then I'd leave all the renames for a second commit.  That will both improve the readability of the commits and allow me to get going on the non-rename portion of the changes right away.  It may also take a few iterations before we get the renames all settled to everyone's satisfaction.  Does that work for you?

src/ripple/app/tx/impl/NFTokenCreateOffer.cpp
```diff
@@ -0,0 +1,234 @@
...
+
+    {
+        auto const amount = ctx.tx[sfAmount];
+
+        if (!isXRP(amount))
+        {
+            if (nftFlags & nft::flagOnlyXRP)
+                return temBAD_AMOUNT;
+
+            if (!amount)
```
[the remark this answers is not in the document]
scottschurr: Yeah.  As it currently works you can give an NFT away by specifying `XRP(0)` for the `Amount`.  We don't allow specifying zero `IOU` values to give away an NFT.  I think the intention is to reduce errors.  We could change the rules, but that's how it currently works.

src/ripple/protocol/impl/ErrorCodes.cpp
```diff
@@ -26,98 +26,77 @@ namespace RPC {
...
+    {rpcSRC_ACT_NOT_FOUND,       "srcActNotFound",         "Source account not found."},
+    {rpcSRC_CUR_MALFORMED,       "srcCurMalformed",        "Source currency is malformed."},
+    {rpcSRC_ISR_MALFORMED,       "srcIsrMalformed",        "Source issuer is malformed."},
+    {rpcSTREAM_MALFORMED,        "malformedStream",        "Stream malformed."},
+    {rpcTOO_BUSY,                "tooBusy",                "The server is too busy to help you now."},
+    {rpcTXN_NOT_FOUND,           "txnNotFound",            "Transaction not found."},
+    {rpcUNKNOWN_COMMAND,         "unknownCmd",             "Unknown method."},
+    {rpcSENDMAX_MALFORMED,       "sendMaxMalformed",       "SendMax amount malformed."},
+    {rpcOBJECT_NOT_FOUND,        "objectNotFound",         "The requested object was not found."}};
+// clang-format off
```
scottschurr, line 99: Should be `clang-format on`.

src/ripple/app/tx/impl/NFTokenBurn.cpp
```diff
@@ -0,0 +1,130 @@
...
+
+    std::size_t totalOffers = 0;
+
+    {
+        Dir buys(ctx.view, keylet::nft_buys(id));
+        totalOffers += std::distance(buys.begin(), buys.end());
+    }
+
+    if (totalOffers > maxDeletableTokenOfferEntries)
+        return tefTOO_BIG;
```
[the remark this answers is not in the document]
scottschurr: Yeah.  That's an unfortunate consequence of having the act of burning an NFT automatically remove any offers for the burned NFT.  If we don't set a limit then we can get excessive metadata (from all the cancelled offers) and the transaction fails anyway.

The only way out that I see is to always leave the stale offers on the ledger.  I think that would probably be a worse choice than what we have here.

src/ripple/app/tx/impl/InvariantCheck.cpp
```diff
@@ -485,4 +488,192 @@ ValidNewAccountRoot::finalize(
...
+void
+ValidNFTokenPage::visitEntry(
+    bool,
+    std::shared_ptr<SLE const> const& before,
+    std::shared_ptr<SLE const> const& after)
+{
+    using namespace std::string_view_literals;
+
+    static uint256 constexpr accountBits(
+        "ffffffffffffffffffffffffffffffffffffffff000000000000000000000000"sv);
```
scottschurr, line 502: Use `nft::pageMask` instead of local definition.

src/ripple/app/tx/impl/NFTokenAcceptOffer.cpp
```diff
@@ -0,0 +1,366 @@
...
+NFTokenAcceptOffer::pay(
+    AccountID const& from,
+    AccountID const& to,
+    STAmount amount)
+{
+    // This should never happen, but it's easy and quick to check.
+    if (amount < beast::zero)
+        return tecINTERNAL;
+
+    return accountSend(view(), from, to, amount, j_);
```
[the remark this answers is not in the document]
scottschurr: I _think_ this is okay as written, but maybe you'll convince me otherwise.  Let me walk through the steps.

1. The total paid for the NFT always equals or exceeds all fees associated with the transfer -- both the transfer fee and the broker fee.
2. If there is a broker, the broker is paid their fee first.  That is subtracted from the total paid.
3. If the seller specified a selling price then the remainder, after the broker is paid, must equal or exceed the specified selling price.
4. The issuer's portion is determined as a percentage of this remainder.  We now pay that portion to the issuer.  The amount paid to the issuer is then subtracted from the total paid.
5. At this point, the last bit of the remainder may be _below_ the amount requested by the seller.  We consider that to be okay because the seller is paying the transfer fee to the issuer.
6. These last remaining funds are then transferred to the seller.

So the effect is that the seller is paying the transfer fee, even though the funds did not actually transfer from buyer to seller to issuer.  The final effect is that the seller is paying the transfer fee to the issuer.

I'm open to further discussion if this doesn't seem right to you.

src/ripple/app/tx/impl/NFTokenAcceptOffer.cpp
```diff
@@ -0,0 +1,366 @@
...
+            return r;
+    }
+
+    // Now transfer the NFT:
+    auto nft = nft::findToken(view(), seller, tokenID);
+
+    if (!nft)
+        return tecINTERNAL;
+
+    if (auto const ret = nft::removeToken(view(), seller, tokenID);
```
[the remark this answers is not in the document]
scottschurr: As it stands right now, `nft::findToken()` can't do this work, because it takes a `ReadView` parameter instead of an `ApplyView`.  What I'll try is adding a new free function, `nft::findTokenAndPage`, that will take an `ApplyView` and return both the token object and a `std::shared_ptr<SLE>` to the page.  That may help.  And it may be useful elsewhere.  We'll see.

src/ripple/protocol/TER.h
scottschurr: @nbougalis I have a more general concern about the additional `tec` codes.  The only codes that need to be captured in the metadata are `tes` and `tec`.  From years ago I remember hearing that there is some trick for how `tec` codes are serialized that does not allow us to currently use, say, a full 256 values for possible `tec` codes.  As I recall, vaguely, that limit was around 159.

Is that recollection correct?  If so, how close does this bring us to whatever the current limit is?  Is it time to put an escape hatch in place so we can support new `tec` codes with yet higher numbers?

scottschurr: @nbougalis I have a more general concern about the additional `tec` codes.  The only codes that need to be captured in the metadata are `tes` and `tec`.  From years ago I remember hearing that there is some trick for how `tec` codes are serialized that does not allow us to currently use, say, a full 256 values for possible `tec` codes.  As I recall, vaguely, that limit was around 159.

Is that recollection correct?  If so, how close does this bring us to whatever the current limit is?  Is it time to put an escape hatch in place so we can support new `tec` codes with yet higher numbers?

src/ripple/protocol/SField.h
```diff
@@ -150,7 +153,8 @@ class SField
...
+    explicit SField(private_access_tag_t, int fc, bool);
```
scottschurr, line 157: The `bool` parameter in this signature is unused.  Remove the `bool`, which takes this signature back to the original signature.

scottschurr, line 157: The `bool` parameter in this signature is unused.  Remove the `bool`, which takes this signature back to the original signature.

src/test/jtx/token.h
scottschurr: Storing it as a string makes it a little easier to use, since we're just about to use it to build JSON.  I'll leave as is.  But thanks for the comment.

scottschurr: Storing it as a string makes it a little easier to use, since we're just about to use it to build JSON.  I'll leave as is.  But thanks for the comment.

src/ripple/protocol/TER.h
```diff
@@ -114,11 +114,14 @@ enum TEMcodes : TERUnderlyingType {
...
     temCANNOT_PREAUTH_SELF,
     temINVALID_COUNT,
+    temBAD_TRANSFER_FEE,
+    temTOKEN_IS_NONFLAMMABLE,
 
     temUNCERTAIN,  // An internal intermediate result; should never be returned.
     temUNKNOWN,    // An internal intermediate result; should never be returned.
 
     temSEQ_AND_TICKET,
+    tefNOT_POSSIBLE,
```
scottschurr, line 124: `tefNOT_POSSIBLE` was added to the wrong `enum`!

src/ripple/protocol/TER.h
scottschurr: I did a little more digging into this question.  Only `tesSUCCESS` ('0x0`) and `tec` codes (`0x64` through `0xFF`) are stored in metadata.  They are captured in an `sfTransactionResult`, which is a `UINT8`.  So without any complicated encoding we can capture a total of 155 `tec` codes.  With this commit we will have used 64 of those codes, so we'll have 91 left -- more than half.

From there we could start using the range from `0x1` through `0x63` for `tec` codes.  After we've used those we can't create more `tec` codes without a more complicated encoding.

So it looks like the need is not urgent.  But it wouldn't hurt to have a plan for `tec` codes beyond 255.

scottschurr:
@seelabs, thank you very much for the thorough review.  I've just added a commit that I believe addresses all of your comments that did not require renames.

Today or tomorrow I'll submit a comment here that can be a starting place for the rename discussion.  Renames will be complicated by the fact that we should probably not change API-facing names that are in the spec.  That means a number of the names are cast in stone.  Non-API facing names should harmonize with the names we can't change.  More thoughts to follow.

I'm assuming I should hold off on rebasing until the review is complete.

src/ripple/protocol/TER.h
scottschurr: Addressing the naming question of `tecOBJECT_NOT_FOUND` vs `tecNO_NFT_OFFER`, my personal bias is toward `tecNO_NFT_OFFER`.  I find failing transactions really difficult to debug because the error codes are often quite unspecific.  We have way too many cases of `temMALFORMED`.

On the flip side, I also see the value minimizing the number of error codes -- especially `tec` codes -- in the code base.  See the previous discussion.

So I'm inclined to leave this code alone to make it more reusable.  Unless other folks weigh in on the side of `tecNO_NFT_OFFER`, in which case I'm fine with changing the name.

src/ripple/protocol/impl/Rate2.cpp
scottschurr: I'm good with putting this in the `nft` namespace.

src/ripple/protocol/impl/SField.cpp
scottschurr: Given the recent work in the documentation to change the term `IOU` to `Token` I have agree with you.  Unqualified `token` is a poor synonym for `NFToken`.  Thanks very much for pointing that out.

However I think there's too much inertia in the spelling `NFToken` in the API to switch that to just `NFT`.  `NFToken` is everywhere throughout the spec and folks have been implementing against the transactions named `TFToken*`.

But I agree with you that the `SField`s that use just a `Token` designation are too vague given the current state of the documentation.  I'm going to suggest that we rename the `SField`s that use just `Token` to use `NFToken`.  So we'd get...

- `sfNFTokenTaxon`
- `sfMintedNFTokens`
- `sfBurnedNFTokens`

and the like.

There are also a couple of `SField`s that fully spell out `NonFungibleToken`.  My inclination is to rename those to `NFToken` as well.  That, also, will increase the regularity of the naming.

Thoughts?

src/ripple/protocol/TER.h
scottschurr: I looked more closely at the proposed `TER` codes.  There were a few that were defined but unused.  There were a few more that I decided to make more consistent with the API by adding `NFTOKEN` to their strings.  Here's the commit with the changes I'm proposing: https://github.com/ripple/rippled/pull/4101/commits/043afbdd3b22f4c35c87e3566283960e3e35b26b

I think this is an improvement over the current situation, but it also is a change to the API.  I don't believe any of these `TER` codes are part of the spec.  So it's a relatively minor change, as API changes go.  I believe this should have minimal impact on people developing new XLS20 code.

But I've also isolated the commit, so if someone says that we shouldn't make this change it's relatively easy to undo.

Let me know what you think.

src/ripple/protocol/jss.h
scottschurr: Renamed to `nft_id` to match similar fields.  This is an API change, so we'll need to clear it with the authorities.  I made the change in a single commit so it will be easy to back out if necessary: https://github.com/ripple/rippled/pull/4101/commits/1a932bf835574686ca7f058640d498c509627899

src/ripple/rpc/impl/GRPCHelpers.cpp
scottschurr: This rename I'm not so confident about.  The problem is that `populateTokenOffers()` is supposed to echo the name of the `SField`, which is `TokenOffers`.  I'd prefer not to rename this function unless we also rename the `SField`.  Consistent naming is interconnected...

scottschurr:
I have been looking for a way to address the naming concerns raised by @seelabs.  I agree with @seelabs that we have some serious naming inconsistencies and opportunities for developer confusion.  I've pushed a couple of commits that address non-`SField`s, but still have some minor impact on the API.  Those commits are:

- https://github.com/ripple/rippled/pull/4101/commits/043afbdd3b22f4c35c87e3566283960e3e35b26b (`[FOLD] Improve names of and remove unused TER codes`) and
- https://github.com/ripple/rippled/pull/4101/commits/1a932bf835574686ca7f058640d498c509627899 (`[FOLD] Rename jss::tokenid to jss::nft_id for RPC field consistency`)

If someone has concerns with those two commits they should be easy to remove.

That leaves changes to `SField`s to consider.   This is the last chance we have to change these names.  I think changing these names is the right call.  But changing the names now will require editing the spec.  It will also cause some heartache for those diligent developers who have been coding against the NFT testnet.  We have to weight future self-consistency against current-day pain.

Since I'm less confident that there is appetite for changing the `SField` names I've pushed those commits in a different branch.  If we choose to pick up the changes they will be very easy to cherry-pick.  The commits that change `SField` names are these:

- https://github.com/scottschurr/rippled/commit/5c6e8a985b0cc47f0c98cb3e7140690a516711ec (`[FOLD] Rename SFields to NFTokenTaxon, MintedNFTokens, and BurnedNFTokens`)
- https://github.com/scottschurr/rippled/commit/8f9167fa3b7d01824e3b5de54dd1c7b2ad6c720d (`[FOLD] Rename SFields to NFTokenOffer and NFTokenOffers`)
- https://github.com/scottschurr/rippled/commit/7b5ecf3de9e10bba9431c92c11e84855d0bf1928 (`[FOLD] Rename SFields to NFTokenMinter and NFTokenBrokerFee`)
- https://github.com/scottschurr/rippled/commit/af7c54e80fe5fe5bd698f13297cb05a31aa929f4 (`[FOLD] Rename SFields to NFToken and NFTokens`) and
- https://github.com/scottschurr/rippled/commit/8a0fce1b41b1d34be52f99c93f03eae85adb88be (`[FOLD] Rename SFields to NFTokenOfferNode, NFTokenBuyOffer, and NFTokenSellOffer`)

I'd appreciate reviewers weighing in on whether or not we should change the `SField` names now.  If not now, then we should assume it will be never.

Thanks.

scottschurr:
All name changes are incorporated.  Rebased to 1.9.0-b2 and squashed.

nbougalis (approved):

PR #4109, opened 2022-02-26
rippled doesn't build with gcc-7 or clang-7

legleux:
rippled has moved along and cannot be built with gcc-7 anymore.
RocksDB has been updated and cannot be built with CMake 3.9 anymore.
Ubuntu 20.04 provides CMake 3.16.3


legleux:
This would also allow https://github.com/ripple/rippled/pull/4082 to be committed as-is

thejohnfreeman:
I don't know what trips up GCC 7, but GCC 8 was released in 2018, so I don't see any reason to continue supporting GCC 7. I'll let someone else advocate for it if there is one.

thejohnfreeman (approved):

legleux:
Already committed in https://github.com/ripple/rippled/commit/eb576790859d6b66b4ddd792f9bcb5aa671d6c4d

PR #4113, opened 2022-03-04
[Trivial] Use 3.16 for minimum CMake supported that Travis tests

legleux:
### Context of Change

Travis tests the minimum CMake version supported. 
It is currently 3.9 and rocksdb cannot be built with this version which necessitates https://github.com/ripple/rippled/pull/4109 increasing that to CMake 3.16.


manojsdoshi (approved):
LGTM

PR #4123, opened 2022-03-16
Add AMM Instance Create.

gregtatcam:

Adds AMM core functionality Instance Create, which creates AMM pool with two tokens.

### Context of Change

Adds AMM Root Account, trustline for each IOU AMM token, trustline to track Liquidity Provider Tokens (LPT),
and directory entry to track all AMM's with different weights. Note that only equal weights are supported
in the first release. A new RPC method `amm_info` is added to fetch the pool and LPT balances.

### Type of Change

- [ x] New feature (non-breaking change which adds functionality)

## Test Plan

AMM unit test is added to test AMM Creation.

## Future Tasks

Additional core functionality is AMM Deposit, Withdrawl, and Swap.

### DIFF
--- a/Builds/CMake/RippledCore.cmake
+++ b/Builds/CMake/RippledCore.cmake
@@ -375,6 +375,8 @@ target_sources (rippled PRIVATE
   src/ripple/app/reporting/ReportingETL.cpp
   src/ripple/app/reporting/ETLSource.cpp
   src/ripple/app/reporting/P2pProxy.cpp
+  src/ripple/app/misc/impl/AMM.cpp
+  src/ripple/app/misc/impl/AMM_formulae.cpp
   src/ripple/app/misc/CanonicalTXSet.cpp
   src/ripple/app/misc/FeeVoteImpl.cpp
   src/ripple/app/misc/HashRouter.cpp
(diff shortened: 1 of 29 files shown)


RichardAH:
Hi! Interesting addition to the codebase. A couple preliminary comments:
1. Anything that adds to or alters the behaviour of on-ledger objects must be guarded using an amendment. Otherwise some nodes will update and execute the new code and some will not, resulting in a disagreement about what the canonical state of the ledger is, causing a fork.
2. Glancing at your repo there are more parts to this amendment, and these should all be squashed together into as few commits as is reasonable, again with all behaviour-altering code guarded with, in this case, `if (view.rules().enabled(featureAMM))`, and submitted as a single PR (IMHO).

gregtatcam:
> Hi! Interesting addition to the codebase. A couple preliminary comments:
> 
> 1. Anything that adds to or alters the behaviour of on-ledger objects must be guarded using an amendment. Otherwise some nodes will update and execute the new code and some will not, resulting in a disagreement about what the canonical state of the ledger is, causing a fork.
> 2. Glancing at your repo there are more parts to this amendment, and these should all be squashed together into as few commits as is reasonable, again with all behaviour-altering code guarded with, in this case, `if (view.rules().enabled(featureAMM))`, and submitted as a single PR (IMHO).

Thanks for the feedback @RichardAH. Adding the amendment is on the the todo list. This is the core feature branch so we can do our sprints. It'll be squashed into one PR when it's pushed into the `develop`.

PR #4124, opened 2022-03-24
[Trivial] Remove checks for CMake < 3.16

legleux:
Removes  all conditional code for  CMake < 3.16 that will never get hit.

### Context of Change
CMake 3.16 or higher is required now so all this is cruft.

### Type of Change


legleux:
Should I add this to `.git-blame-ignore-revs`?

thejohnfreeman (commented):
This is more than a formatting change, so please do not add it to the ignored commits. Do we have a pending PR for the Visual Studio 64-bit check? If not, can we add it to this?

manojsdoshi (approved):
LGTM

PR #4130, opened 2022-04-05
Fix Clio build

legleux:

Fix `endif()` position so nudb library builds for projects that include rippled.
### Context of Change
Other projects which depend on rippled would fail because the wrong `endif()` was left in place. Rippled built fine despite the incorrect placement since this guard is for _other_ projects using rippled.

The wrong `endif()` statement was removed leaving nudb unbuilt in projects outside rippled (Clio)

### Type of Change


manojsdoshi (approved):
LGTM

PR #4132, opened 2022-04-05
Do not install core-utils on rocky linux

legleux:
A core-utils-single package already exists on the image and prevents the install and thus tests from succeeding.

### Type of Change


manojsdoshi (approved):
LGTM

PR #4133, opened 2022-04-06
AMM Core Functionality

gregtatcam:

Adds AMM core functionality Instance Create, Deposit, Withdraw, Governance, and Auctioning.

### Context of Change

Adds AMM Root Account, trustline for each IOU AMM token, trustline to track Liquidity Provider Tokens (LPT),
and directory entry to track all AMM's with different weights. Note that only equal weights are supported
in the first release. A new RPC method `amm_info` is added to fetch the pool and LPT balances.

### Type of Change

- [ x] New feature (non-breaking change which adds functionality)

## Test Plan

AMM unit test is added to test AMM Create, Deposit, Withdraw, Swap.

## Future Tasks

Add Number library to facilitate AMM equations coding and Payment Engine Integration.


HowardHinnant:
I've reviewed the integration with `Number` and it looks good to me.

gregtatcam:
> I've reviewed the integration with `Number` and it looks good to me.

Thanks for reviewing!

src/ripple/proto/org/xrpl/rpc/v1/transaction.proto
```diff
@@ -125,6 +147,66 @@ message AccountDelete
...
+
+    FeeVal fee_val = 2;
+}
+
+// Next field: 5
+message AMMBid
+{
+    AMMHash ammhash = 1;
+
+    MinSlotPrice min_slot_price = 2;
```
injaelee, line 203: Just an FYI on the new AMM RPC calls. (Clio: @cjcobb23 @natenichols)

src/ripple/rpc/handlers/AMMInfo.cpp
```diff
@@ -0,0 +1,314 @@
...
+        }
+    }
+    if (!params.isMember(jss::AMMHash))
+        result[jss::AMMHash] = to_string(ammHash);
+
+    return result;
+}
+
+std::pair<org::xrpl::rpc::v1::GetAmmInfoResponse, grpc::Status>
+doAmmInfoGrpc(RPC::GRPCContext<org::xrpl::rpc::v1::GetAmmInfoRequest>& context)
```
cjcobb23, line 167: This is not needed

src/ripple/rpc/GRPCHandlers.h
```diff
@@ -71,6 +71,9 @@ std::pair<org::xrpl::rpc::v1::GetLedgerDiffResponse, grpc::Status>
 doLedgerDiffGrpc(
     RPC::GRPCContext<org::xrpl::rpc::v1::GetLedgerDiffRequest>& context);
 
+std::pair<org::xrpl::rpc::v1::GetAmmInfoResponse, grpc::Status>
+doAmmInfoGrpc(RPC::GRPCContext<org::xrpl::rpc::v1::GetAmmInfoRequest>& context);
+
```
cjcobb23, line 76: This is not needed. gRPC methods like this have been deprecated, no need to add more

src/ripple/proto/org/xrpl/rpc/v1/transaction.proto
```diff
@@ -27,6 +27,16 @@ message Transaction
...
 
+        AMMInstanceCreate amminstance_create = 41;
+
+        AMMDeposit ammdeposit = 42;
+
+        AMMWithdraw ammwithdraw = 43;
+
+        AMMVote ammvote = 44;
+
+        AMMBid ammbid = 45;
```
cjcobb23, line 38: This whole part of the code has been deprecated. I would prefer to not add these, but some of the unit tests might fail.

src/ripple/proto/org/xrpl/rpc/v1/xrp_ledger.proto
```diff
@@ -51,4 +52,7 @@ service XRPLedgerAPIService {
   // ledgers. Note, this method has no JSON equivalent.
   rpc GetLedgerDiff(GetLedgerDiffRequest) returns (GetLedgerDiffResponse);
 
+  // Get account info for an account on the XRP Ledger.
+  rpc GetAmmInfo (GetAmmInfoRequest) returns (GetAmmInfoResponse);
+
```
cjcobb23, line 57: Let's not add this. These gRPC methods were for the xpring sdk, which has been abandoned

cjcobb23 (changes requested):

src/ripple/proto/org/xrpl/rpc/v1/xrp_ledger.proto
gregtatcam: @cjcobb23 I'm working on another AMM PR. I'll incorporate your feedback in this new PR. To sum it up - AMM transactions and GetAmmInfo() should be removed from gRPC.

src/ripple/proto/org/xrpl/rpc/v1/transaction.proto
scottschurr: @cjcobb23

> This whole part of the code has been deprecated. I would prefer to not add these, but some of the unit tests might fail.

Precisely.  This dead code has been in the server for a while now.  Can you please work with someone to remove all of the gRPC serialization code (not just the unit tests) from the rippled code base correctly and without harming Clio?  It sounds like Clio may be in a stable enough place that this problem can finally be addressed.

natenichols: Personally, I'd like to remove gRPC in its entirety and extract data over https or websocket. It seems much easier to remove in its entirety versus having gRPC as a dependency while whittling down gRPC functionality.

This wouldn't be a huge challenge, removes something like 15k Lines of Code from rippled, and can reuse a ton of http/s server code in rippled. 

This is not a technically hard task, it just touches a lot of code and requires familiarity with Clio. I think writing this branch took me 20ish hours of work https://github.com/natenichols/rippled/tree/remove-grpc, but that's with no tests, no polish, and without rewriting Clio. 

It's also challenging because `reporting mode` extracts data over gRPC. We'd either have to rewrite `ETLSource` in rippled, or deprecate reporting mode before we could entirely remove gRPC.

scottschurr: Thanks for your research, @natenichols.

I hear you about the preferred path being to remove gRPC all together.  And I hope we get there eventually.  I'm under the impression that once Clio is fully battle tested then deprecating reporting mode should be possible.  But, personally, I'd prefer not to wait an indefinite time for a perfect solution when we have a problem right now: maintaining the gRPC serializer.

What I would hope for is that, with @cjcobb23's guidance, the serializer potion of the rippled gRPC implementation could be removed very soon.  My understanding has been that reporting mode has no dependency on the serializer.  But I could be wrong on that.  I look to you and @cjcobb23 for the accuracy of that.

Once the serializer code is gone I anticipate that it becomes much more obvious what work needs to be done when new transactions and ledger types are added to the code base.

But someone needs to be tasked with that work and given guidance from @cjcobb23.

gregtatcam:
Closed, new PR is based on amm-core-functionality branch.

PR #4135, opened 2022-04-06
Comment out Intel specific instruction `_mm_pause` on Apple arm platforms, which may cause a slight performance hit.

greg7mdp:
### Type of Change

Add #ifdef in order to use x86 intrinsics only when not on arm platform. 

The use of intrinsic instruction (`_mm_pause()`) is an optimization for yielding the processor and its absence on arm builds should not be an issue.

### DIFF
--- a/src/ripple/shamap/impl/SHAMapInnerNode.cpp
+++ b/src/ripple/shamap/impl/SHAMapInnerNode.cpp
@@ -34,8 +34,10 @@
 #include <iterator>
 #include <utility>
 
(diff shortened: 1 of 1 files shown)


seelabs (commented):
This looks good. One minor nit: can we update the commit message to give a little more detail what this change is doing. Something about `_mm_pause` being intel specific intrinsic instruction and this patch disables that instruction when building on arm platforms. This will result is slightly poorer performance on arm platforms.

PR #4142, opened 2022-04-11
Periodically reshare federator sigs

seelabs:

This PR will periodically reshare federator signatures and clean out old signatures.  It will only remove signatures for txns the federator is unaware of. It will reshare mainchain and sidechain signatures on a heartbeat timer. It will wwitch between sharing sidechain and mainchain signatures on each timeout, in an attempt to reduce network traffic.

### DIFF
--- a/src/ripple/app/sidechain/Federator.cpp
+++ b/src/ripple/app/sidechain/Federator.cpp
@@ -45,7 +45,6 @@
 
 #include <boost/algorithm/string/classification.hpp>
 #include <boost/algorithm/string/split.hpp>
-#include <boost/asio/deadline_timer.hpp>
 #include <boost/asio/error.hpp>
 #include <boost/system/detail/errc.hpp>
 
--- a/src/ripple/app/sidechain/Federator.cpp
+++ b/src/ripple/app/sidechain/Federator.cpp
@@ -41,14 +41,18 @@
 #include <ripple/rpc/Role.h>
 #include <ripple/rpc/impl/RPCHelpers.h>
 #include <ripple/rpc/impl/TransactionSign.h>
-#include <mutex>
 #include <ripple.pb.h>
 
 #include <boost/algorithm/string/classification.hpp>
 #include <boost/algorithm/string/split.hpp>
+#include <boost/asio/deadline_timer.hpp>
+#include <boost/asio/error.hpp>
+#include <boost/system/detail/errc.hpp>
 
 #include <chrono>
 #include <cmath>
+#include <memory>
+#include <mutex>
 #include <sstream>
 
 namespace ripple {
@@ -815,6 +819,13 @@ Federator::init(
     ticketRunner_.setRpcChannel(false, sidechainListener_);
     mainDoorKeeper_.setRpcChannel(mainchainListener_);
     sideDoorKeeper_.setRpcChannel(sidechainListener_);
+
+    heartbeatTimer =
+        std::make_unique<boost::asio::deadline_timer>(ios, heartbeatInterval);
+    heartbeatTimer->async_wait(
+        [self = shared_from_this()](boost::system::error_code const& ec) {
+            self->heartbeatTimerHandler(ec);
+        });
 }
 
 Federator::~Federator()
@@ -851,6 +862,25 @@ Federator::stop()
         running_ = false;
     }
     mainchainListener_->shutdown();
+}
+
+void
+Federator::heartbeatTimerHandler(const boost::system::error_code& ec)
+{
+    if (ec == boost::asio::error::operation_aborted)
+        return;
+
+    if (ec == boost::system::errc::success)
+    {
+        onEvent(event::HeartbeatTimer{});
+    }
+
+    heartbeatTimer->expires_at(
+        heartbeatTimer->expires_at() + heartbeatInterval);
+    heartbeatTimer->async_wait(
+        [self = shared_from_this()](boost::system::error_code const& ec) {
+            self->heartbeatTimerHandler(ec);
+        });
 }
 
 void
@@ -1354,6 +1384,13 @@ Federator::onEvent(event::HeartbeatTimer const& e)
         "Federator::onEvent",
         jv("eventtype", "HeartbeatTimer"),
         jv("event", e.toJson()));
+
+    static bool shareMain = false;
+    shareMain = !shareMain;
+    if (shareMain)
+        mainSigCollector_.reshareSigs();
+    else
+        sideSigCollector_.reshareSigs();
 }
 
 void
--- a/src/ripple/app/sidechain/impl/SignatureCollector.cpp
+++ b/src/ripple/app/sidechain/impl/SignatureCollector.cpp
@@ -139,21 +139,17 @@ void
 SignatureCollector::expire()
 {
     std::lock_guard lock(mtx_);
-    // Never expire collections with this servers sig or submitted txns
+    // Never expire collections with this server's sig or submitted txns
     for (auto i = messages_.begin(), e = messages_.end(); i != e; ++i)
     {
         auto const& multiSigMsg = i->second;
-        if (multiSigMsg.submitted_)
+        if (multiSigMsg.submitted_ ||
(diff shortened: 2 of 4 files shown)


seelabs:
@greg7mdp Please review.

greg7mdp:
Thanks @seelabs , I'm looking at it.

src/ripple/app/sidechain/Federator.cpp
```diff
@@ -815,6 +819,13 @@ Federator::init(
     ticketRunner_.setRpcChannel(false, sidechainListener_);
     mainDoorKeeper_.setRpcChannel(mainchainListener_);
     sideDoorKeeper_.setRpcChannel(sidechainListener_);
+
+    heartbeatTimer =
+        std::make_unique<boost::asio::deadline_timer>(ios, heartbeatInterval);
+    heartbeatTimer->async_wait(
+        [self = shared_from_this()](boost::system::error_code const& ec) {
```
greg7mdp, line 826: `this` should work here, right, since `heartbeatTimer` will be destructed before its containing object?

greg7mdp:
Just was looking at `class PublicKey`. Could we make the `size_` member `std::byte` instead of `std::size_t`?

src/ripple/app/sidechain/impl/SignatureCollector.cpp
```diff
@@ -139,9 +139,41 @@ void
...
+        if (multiSigMsg.submitted_)
+        {
+            messages_.touch(i);
+            continue;
+        }
+        for (auto const& [pk, sig] : multiSigMsg.sigMaps_)
+        {
+            if (pk == myPubKey_)
+            {
+                messages_.touch(i);
```
greg7mdp, line 155: add `break;` after the `touch(i)`?

greg7mdp: or replace the whole second loop with something like:
```
if (std::any_of(multiSigMsg.sigMaps_.begin(), multiSigMsg.sigMaps_.end(),
                [&](const auto& p) { return p.first == myPubKey_; })) 
{
    messages_.touch(i);
}
```

greg7mdp (changes requested):
Hey Scott, I think I roughly understand what you are doing, but not in detail, so not 100% sure of my comments.

src/ripple/app/sidechain/impl/SignatureCollector.cpp
```diff
@@ -139,9 +139,41 @@ void
...
+SignatureCollector::reshareSigs()
+{
+    std::lock_guard lock(mtx_);
+    for (auto const& [messageId, multiSigMsg] : messages_)
+    {
+        if (multiSigMsg.submitted_)
+            continue;
+        for (auto const& [pk, sig] : multiSigMsg.sigMaps_)
+        {
+            shareSig(messageId, sig);
```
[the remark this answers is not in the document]
seelabs: `submitted_` is set to true when the transaction is submitted in the `submit` function. If the txn is already submitted there's no reason to rebroadcast the signature, as the federator already has enough signatures to submit the transaction itself.

src/ripple/app/sidechain/impl/SignatureCollector.cpp
seelabs: Breaking after the touch is a good idea. Will do. I prefer a raw loop over `any_of` here, but that's just a personal preference; both are equally easy to read.

src/ripple/app/sidechain/Federator.cpp
seelabs: If the heartbeat timer gets destroyed is the callback also destroyed? I'm used to using `shared_from_this` in asio because the lifetime of the executor can outlive the lifetime of the captured object (especially during shutdown). You might be right that a timer destroys the call back, but we'd have to check. Let's leave it as-is for now.

greg7mdp: > if the heartbeat timer gets destroyed is the callback also destroyed?

yes I believe that is the case, see [this](https://www.boost.org/doc/libs/1_75_0/doc/html/boost_asio/reference/basic_deadline_timer/_basic_deadline_timer.html)

src/ripple/app/sidechain/impl/SignatureCollector.cpp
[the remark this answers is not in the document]
greg7mdp: That makes sense, thanks Scott!

src/ripple/app/sidechain/Federator.cpp
seelabs: That cancels the timer, which means the callback is still exists and may be called. If the callback is running when the `heartbeatTimer` is deleted then `this` would be become dangling. Even if it isn't in the middle on a call, it will still call the callback with `ec` set to `operation_aborted` and a dangling `this`. In the current implementation (and most sane implementations) we check `ec` before accessing `self`. Still, I'd rather not have the pointer dangle at all. Yeah, I think we need the `shared_from_this` here.

src/ripple/app/sidechain/impl/SignatureCollector.cpp
seelabs: Yeah, the `continue` in one and a `break` in another is a good argument. I'll make the change.

src/ripple/app/sidechain/Federator.cpp
greg7mdp: >  which means the callback is still exists and may be called. 

I don't think so. It says: This function destroys the timer, **cancelling any outstanding asynchronous wait operations associated with the timer as if by calling cancel**.
I'm pretty sure the cancel will happen in the deadline_timer destructor, which means the callback will be called then, when the class owning the deadline_timer is still present. I don't see how `this` can be dangling.
Just discussing to make sure we both agree on what would happen, I have no issue with using `shared_from_this`.

greg7mdp (approved):
Looks good!

src/ripple/app/sidechain/Federator.cpp
seelabs: The captured variable `dangling` points to a dangling variable in the code below. The destructor will cause the operation to be cancelled, but that doesn't mean the callback is called before the destructor is finished running.

Even if cancellation isn't an issue, destroying the object in the middle of a callback would still be an issue.

```
#include <boost/asio.hpp>
#include <boost/asio/deadline_timer.hpp>
#include <iostream>
#include <limits>
#include <memory>
#include <vector>

int
main ()
{
    boost::asio::io_context ios;
    {
        boost::posix_time::seconds const heartbeatInterval{5};
        boost::asio::deadline_timer t (ios, heartbeatInterval);
        int dangling = 0;
        t.async_wait ([&dangling] (boost::system::error_code const& ec) {
            std::cout << "ec: " << ec << '\n';
            std::cout << "dangling: " << dangling << '\n';
        });
    }
    ios.run ();

    return 0;
}
```

greg7mdp:
```
 #include <boost/asio.hpp>
#include <boost/asio/deadline_timer.hpp>
#include <iostream>
#include <limits>
#include <memory>
#include <vector>

struct DestructorProbe
{
    int x{2};
    boost::asio::deadline_timer t;

    DestructorProbe(boost::asio::io_context &ios) :
        t(ios, boost::posix_time::seconds{5})
    {
        t.async_wait ([this] (boost::system::error_code const& ec) {
            std::cout << "ec: " << ec << '\n';
            std::cout << "dangling: " << x << '\n';
        });
    }
        
    ~DestructorProbe ()
    {
        x = 42;
    }
};

int
main ()
{
    boost::asio::io_context ios;
    {
        DestructorProbe dangling(ios);
    }
    ios.run ();

    return 0;
}
```

seelabs: Just leaving a note: we talked in person; the `shared_pointer` is needed. resolving this conversation.

seelabs:
Merged into the sidechain branch

Issue #4005, opened 2021-11-28
RPC 'tooBusy' response should have 50x (503?) status code


WietseWind:
Currently, if rippled (over RPC) returns a tooBusy response, like below, it does so with a HTTP 200 status code. This doesn't deserve a 200 status code, all is not OK, and one is definitely not getting the expected response.

I propose returning a HTTP 503 status code along with the tooBusy error message. Any error for that matter should have a non-200 HTTP status code.

By not assigning a non-200 status code:
- Clients are mislead on the validity of the response
- Reverse proxies / load balancers can't auto failover to another node as they get a 200 response

```
{
  "result": {
    "error": "tooBusy",
    "error_code": 9,
    "error_message": "The server is too busy to help you now.",
    "request": {
      "account": "rTooLkitCksh5mQa67eaa2JaWHDBnHkpy",
      "command": "account_tx",
      "ledger_index_max": 61974273,
      "limit": 20
    },
    "status": "error"
  }
}
```

cjcobb23:
I agree that `tooBusy` should have a 50x status code. I don't know that every error should though. I don't think an account not found or transaction not found error means anything is actually wrong with the server, without knowing whether that data should have been found.

intelliot:
If we want to follow the HTTP status codes:
- 200 OK means the request was successfully received, understood, and accepted. The response includes the requested resource
- 503 Service Unavailable is correct for a server that is too busy
- 404 Not Found may be used for account not found or transaction not found
- 204 No Content may be used for e.g. account_tx with no txs in the requested range of ledgers

WietseWind:
> I agree that `tooBusy` should have a 50x status code. I don't know that every error should though. I don't think an account not found or transaction not found error means anything is actually wrong with the server, without knowing whether that data should have been found.

Just had the same thought. There's a distinction between "unexpected", infra-like error (must have an error code) and 'logic' error (200, but with an error).

@intelliot I think the 404 for e.g. account not found is a bad idea as that would trigger auto-retry/failover load balancers to retry with another uplink.

@intelliot @cjcobb23 I think this could be a good thing to keep in mind deciding if something needs a non-200 status code: **"would the same call on another node potentially yield different (better, expected) results?"**? If so: non-200 (error, so next uplink can be queried). If not (e.g. account doesn't exist, address encoding wrong, ledger sequence in the future, name it...): 200 status code but potentially with an error explaining what was wrong.

syeduali93:
[...]

PR #4143, opened 2022-04-11
RPC tooBusy response now has 503 HTTP status code:

scottschurr:

Fixes [#4005](https://github.com/ripple/rippled/issues/4005)

Makes it possible for internal RPC Error Codes to associate themselves with a non-OK (200) HTTP status code.  This commit also adds non-OK HTTP responses to a few additional RPC error responses.

### Context of Change

Issue [#4005](https://github.com/ripple/rippled/issues/4005) notes that returning an HTTP 200 (OK) status code when the server is overloaded is misleading for load balancers.  This proposed change allows each RPC error response to be associated with its own HTTP status code.

If an explicit HTTP status code is not specified, then 200 (OK) is assumed.  This preserves much of the status quo since prior to this change a 200 (OK) status code was returned in all cases.

The pull request includes a best guess at other cases where a non-200 status code should be returned.  Reviewers are requested to look those over carefully and recommend changes.  Thanks.

### Type of Change

- [x] Potentially breaking change (fix or feature that could cause existing functionality to not work as expected).

The change is potentially breaking in cases where servers are expecting to get a 200 (OK) status code back since they will no longer get that status code in certain error cases.  It's unknown whether that problem would occur in practice.

There is no amendment associated with the change, since it does not change transaction results.  However the change does affect the  **API**.  So care is justified.

## Before / After

I hacked the `server_info` command to return an `rpc_TOO_BUSY` error, then I used the following `curl` command:
```
#!/bin/bash
# Use curl to make a server_info request
curl -i -d '{ "method": "server_info", "params": [ {"ripplerpc": "1.0"} ] }' -H 'Content-Type: application/json' -X POST http://127.0.0.1:5005/
```

Without the change I get the following response:
```
HTTP/1.1 200 OK
Date: Mon, 11 Apr 2022 17:53:55 +0000
Connection: Keep-AliveContent-Length: 195
Content-Type: application/json; charset=UTF-8
Server: ripple-json-rpc/rippled-1.9.0+7c66747d27869f9f3c96617bd4227038f1fa92b8.DEBUG

{"result":{"error":"tooBusy","error_code":9,"error_message":"The server is too busy to help you now.","request":{"command":"server_info","ripplerpc":"1.0"},"status":"error"},"ripplerpc":"1.0"}
```

However with the change I get the following response (note the changed status code):
```
HTTP/1.1 503 Server is overloaded
Date: Mon, 11 Apr 2022 18:03:41 +0000
Connection: Keep-AliveContent-Length: 195
Content-Type: application/json; charset=UTF-8
Server: ripple-json-rpc/rippled-1.9.0+7c66747d27869f9f3c96617bd4227038f1fa92b8.DEBUG

{"result":{"error":"tooBusy","error_code":9,"error_message":"The server is too busy to help you now.","request":{"command":"server_info","ripplerpc":"1.0"},"status":"error"},"ripplerpc":"1.0"}
```

## Test Plan
The rippled testing infrastructure currently ignores the returned HTTP status code.  I could figure out how to fix that.  But I'd like to suggest that we content ourselves with integration tests.  Potentially the integration tests could use `curl` commands similar to what I used for manual testing.  However we'd need to actually overload the server, rather than hacking a command to return an inaccurate result.

More thoughts are appreciated regarding testing.

## Reviewers
I'd appreciate review and comments from @WietseWind who wrote the initial issue.

### DIFF
--- a/src/ripple/protocol/impl/ErrorCodes.cpp
+++ b/src/ripple/protocol/impl/ErrorCodes.cpp
@@ -18,6 +18,7 @@
 //==============================================================================
 
 #include <ripple/protocol/ErrorCodes.h>
+#include <array>
 #include <cassert>
 #include <stdexcept>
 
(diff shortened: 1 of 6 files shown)


WietseWind:
Thank you @scottschurr! Great solution, being able to configure alternative error codes per error. 

I see that you already added error codes for the most important errors/codes. Going through the list I feel there are many more eligible for (mostly) 400 / 401 / 403 / 500 HTTP status codes (e.g. `rpcFORBIDDEN`, `rpcNOT_READY`, `rpcNO_NETWORK`, `rpcJSON_RPC`, `rpcNO_PERMISSION`).

Question is where to draw the line between something the client asked for (response still OK, 200, to be expected) and where it's a response that should explicitly inform the client that something is off.

intelliot (approved):
I believe this change should be accompanied by an addition to the [release notes](https://github.com/ripple/rippled/blob/develop/RELEASENOTES.md) and [docs](https://github.com/XRPLF/xrpl-dev-portal)

scottschurr:
@WietseWind, I'm hoping you'll have time to indicate where you'd like additional HTTP status codes added.  Is there anything you need from me to push that forward?  Thanks.

scottschurr:
Hey @WietseWind, on April 11th you noted:
>  If you agree I'll go through the list and add some more 4xx/5xx status codes.

That's great and would be appreciated.  But you've had five weeks now to do that.  I know you're really busy.  If you can get back to me within the next week to at least let me know when you can get around to identifying those additional status codes I'd really appreciate it.

If I don't hear back from you within the week I'll assume you're too busy to make the assessment right now and we can commit the HTTP status code changes that are currently in the pull request.  If that should happen, then you can make a later pull request to put in the additional status codes.

Thanks for the help.

WietseWind:
Hi @scottschurr, sorry for that. Will do right now, lost track of this thread. Thanks for the reminder.

I propose the codes below, where a non-standard HTTP code is only returned if one can expect (e.g. when submitting against a pool) another node to respond differently (first call: error, second call, other machine: OK) (except for calls needing auth).

https://gist.github.com/WietseWind/661776d0b4c849bfa5f237101104e02a

scottschurr:
Thanks for the help @WietseWind.  I've added a commit with the extra status codes and a bit more code that was required to make the new status codes work.  I think the status codes are in their final state for this pull request.  It would be great if the reviewers would look it over and let me know if there are other things that need fixing.  Thanks.

WietseWind:
@scottschurr Thanks, appreciate it :D

scottschurr:
Ping @cjcobb23 and @mDuo13.  How do you folks feel about this pull request in its current state?  Do you want any changes?  Thanks.

mDuo13 (changes requested):
I'm supportive of the general idea, but I think this change is fairly likely to break clients, so we should not do it in a patch release and ideally it would be on a version switch or something.

There are languages & libraries that raise an error (potentially crashing) if they get a non-200 error code. This change means that some clients that previously correctly handled error messages may crash instead.

Also, there are a decent number of error codes that are more or less specific cases of `invalidParams` but that code uses 400 Bad Request and several of the other codes are still on 200 OK. That's…less than ideal from an API design standpoint. It wouldn't be as bad if it was limited to just 503 Service Unavailable when the server isn't fully operational, but returning 400 Bad Request based on user input is definitely a big risk in terms of breaking people's existing code.

#3418 is an example of a PR where we made a smaller change but put it on an API v2 switch. We should consider doing that with at least some of these changes.

WietseWind:
Makes sense @mDuo13, thanks for chiming in. 

Thoughts regarding the version flag: if that same flag would also change the behaviour of other version related options, one would never be able to use the previous (current, default) version of the API but *with* the response codes that allows for failover in reverse proxy environments (for example). Not ideal from a consistency standpoint, but could it be worth triggering this behaviour based on a HTTP request header?

src/ripple/protocol/impl/ErrorCodes.cpp
```diff
@@ -26,105 +27,96 @@ namespace RPC {
...
-};
+// clang-format off
+constexpr static ErrorInfo unorderedErrorInfos[]{
+    {rpcACT_MALFORMED,          "actMalformed",         "Account malformed."},
+    {rpcACT_NOT_FOUND,          "actNotFound",          "Account not found."},
+    {rpcALREADY_MULTISIG,       "alreadyMultisig",      "Already multisigned."},
+    {rpcALREADY_SINGLE_SIG,     "alreadySingleSig",     "Already single-signed."},
+    {rpcAMENDMENT_BLOCKED,      "amendmentBlocked",     "Amendment blocked, need upgrade.", 503},
+    {rpcEXPIRED_VALIDATOR_LIST, "unlBlocked",           "Validator list expired.", 503},
+    {rpcATX_DEPRECATED,         "deprecated",           "Use the new API or specify a ledger range."},
```
mDuo13, line 54: Suggested code: 400

src/ripple/protocol/impl/ErrorCodes.cpp
```diff
@@ -26,105 +27,96 @@ namespace RPC {
...
+    {rpcBAD_KEY_TYPE,           "badKeyType",           "Bad key type."},
```
mDuo13, line 55: Suggested code: 400

src/ripple/protocol/impl/ErrorCodes.cpp
```diff
@@ -26,105 +27,96 @@ namespace RPC {
...
+    {rpcBAD_FEATURE,            "badFeature",           "Feature unknown or invalid.", 500},
+    {rpcBAD_ISSUER,             "badIssuer",            "Issuer account malformed."},
```
mDuo13, line 57: Suggested code: 400

src/ripple/protocol/impl/ErrorCodes.cpp
```diff
@@ -26,105 +27,96 @@ namespace RPC {
...
+    {rpcBAD_MARKET,             "badMarket",            "No such market."},
```
mDuo13, line 58: Either 400 or 404. Might need to dig more into when this comes up

src/ripple/protocol/impl/ErrorCodes.cpp
```diff
@@ -26,105 +27,96 @@ namespace RPC {
...
+    {rpcBAD_SECRET,             "badSecret",            "Secret does not match account."},
```
mDuo13, line 59: 403 Forbidden

src/ripple/protocol/impl/ErrorCodes.cpp
```diff
@@ -26,105 +27,96 @@ namespace RPC {
...
+    {rpcBAD_SEED,               "badSeed",              "Disallowed seed."},
```
mDuo13, line 60: 403

src/ripple/protocol/impl/ErrorCodes.cpp
```diff
@@ -26,105 +27,96 @@ namespace RPC {
...
+    {rpcBAD_SYNTAX,             "badSyntax",            "Syntax error.", 400},
+    {rpcCHANNEL_MALFORMED,      "channelMalformed",     "Payment channel is malformed."},
```
mDuo13, line 62: 400

src/ripple/protocol/impl/ErrorCodes.cpp
```diff
@@ -26,105 +27,96 @@ namespace RPC {
...
+    {rpcCHANNEL_AMT_MALFORMED,  "channelAmtMalformed",  "Payment channel amount is malformed."},
```
mDuo13, line 63: 400

src/ripple/protocol/impl/ErrorCodes.cpp
```diff
@@ -26,105 +27,96 @@ namespace RPC {
...
+    {rpcCOMMAND_MISSING,        "commandMissing",       "Missing command entry."},
```
mDuo13, line 64: 400

src/ripple/protocol/impl/ErrorCodes.cpp
```diff
@@ -26,105 +27,96 @@ namespace RPC {
...
+    {rpcDB_DESERIALIZATION,     "dbDeserialization",    "Database deserialization error."},
```
mDuo13, line 65: 502 Bad Gateway or 503 Service Unavailable

src/ripple/protocol/impl/ErrorCodes.cpp
```diff
@@ -26,105 +27,96 @@ namespace RPC {
...
+    {rpcDST_ACT_MALFORMED,      "dstActMalformed",      "Destination account is malformed."},
+    {rpcDST_ACT_MISSING,        "dstActMissing",        "Destination account not provided."},
```
mDuo13, line 67: 400 for both

src/ripple/protocol/impl/ErrorCodes.cpp
```diff
@@ -26,105 +27,96 @@ namespace RPC {
...
+    {rpcDST_ACT_NOT_FOUND,      "dstActNotFound",       "Destination account not found."},
```
mDuo13, line 68: 404

src/ripple/protocol/impl/ErrorCodes.cpp
```diff
@@ -26,105 +27,96 @@ namespace RPC {
...
+    {rpcDST_AMT_MALFORMED,      "dstAmtMalformed",      "Destination amount/currency/issuer is malformed."},
+    {rpcDST_AMT_MISSING,        "dstAmtMissing",        "Destination amount/currency/issuer is missing."},
+    {rpcDST_ISR_MALFORMED,      "dstIsrMalformed",      "Destination issuer is malformed."},
+    {rpcEXCESSIVE_LGR_RANGE,    "excessiveLgrRange",    "Ledger range exceeds 1000."},
```
mDuo13, line 72: 400 for all four

src/ripple/protocol/impl/ErrorCodes.cpp
```diff
@@ -26,105 +27,96 @@ namespace RPC {
...
+    {rpcFORBIDDEN,              "forbidden",            "Bad credentials.", 403},
+    {rpcFAILED_TO_FORWARD,      "failedToForward",      "Failed to forward request to p2p node", 503},
+    {rpcHIGH_FEE,               "highFee",              "Current transaction fee exceeds your limit."},
```
mDuo13, line 75: 402 Payment Required? :thinking:

src/ripple/protocol/impl/ErrorCodes.cpp
```diff
@@ -26,105 +27,96 @@ namespace RPC {
...
+    {rpcINTERNAL,               "internal",             "Internal error.", 500},
+    {rpcINVALID_LGR_RANGE,      "invalidLgrRange",      "Ledger range is invalid."},
```
mDuo13, line 77: 400, probably? Not sure if this can be caused by situations other than invalid user input

src/ripple/protocol/impl/ErrorCodes.cpp
```diff
@@ -26,105 +27,96 @@ namespace RPC {
...
+    {rpcINVALID_PARAMS,         "invalidParams",        "Invalid parameters.", 400},
+    {rpcJSON_RPC,               "json_rpc",             "JSON-RPC transport error.", 500},
+    {rpcLGR_IDXS_INVALID,       "lgrIdxsInvalid",       "Ledger indexes invalid."},
+    {rpcLGR_IDX_MALFORMED,      "lgrIdxMalformed",      "Ledger index malformed."},
```
mDuo13, line 81: both 400

src/ripple/protocol/impl/ErrorCodes.cpp
```diff
@@ -26,105 +27,96 @@ namespace RPC {
...
+    {rpcLGR_NOT_FOUND,          "lgrNotFound",          "Ledger not found."},
```
mDuo13, line 82: 404

src/ripple/protocol/impl/ErrorCodes.cpp
```diff
@@ -26,105 +27,96 @@ namespace RPC {
...
+    {rpcLGR_NOT_VALIDATED,      "lgrNotValidated",      "Ledger not validated.", 202},
```
mDuo13, line 83: Are we sure 202 is correct here? That's a "successful" class code so I wouldn't expect an error

src/ripple/protocol/impl/ErrorCodes.cpp
```diff
@@ -26,105 +27,96 @@ namespace RPC {
...
+    {rpcMASTER_DISABLED,        "masterDisabled",       "Master key is disabled."},
```
mDuo13, line 84: 403 Forbidden

src/ripple/protocol/impl/ErrorCodes.cpp
```diff
@@ -26,105 +27,96 @@ namespace RPC {
...
+    {rpcNOT_ENABLED,            "notEnabled",           "Not enabled in configuration.", 501},
+    {rpcNOT_IMPL,               "notImpl",              "Not implemented.", 501},
+    {rpcNOT_READY,              "notReady",             "Not ready to handle this request.", 202},
```
mDuo13, line 87: Sounds more like a 503 to me

src/ripple/protocol/impl/ErrorCodes.cpp
```diff
@@ -26,105 +27,96 @@ namespace RPC {
...
+    {rpcNOT_SUPPORTED,          "notSupported",         "Operation not supported.", 501},
+    {rpcNO_CLOSED,              "noClosed",             "Closed ledger is unavailable.", 503},
+    {rpcNO_CURRENT,             "noCurrent",            "Current ledger is unavailable.", 503},
+    {rpcNOT_SYNCED,             "notSynced",            "Not synced to the network.", 503},
+    {rpcNO_EVENTS,              "noEvents",             "Current transport does not support events."},
```
mDuo13, line 92: 405 Method Not Allowed

src/ripple/protocol/impl/ErrorCodes.cpp
```diff
@@ -26,105 +27,96 @@ namespace RPC {
...
+    {rpcNO_NETWORK,             "noNetwork",            "Not synced to the network.", 503},
+    {rpcNO_PERMISSION,          "noPermission",         "You don't have permission for this command.", 401},
+    {rpcNO_PF_REQUEST,          "noPathRequest",        "No pathfinding request in progress."},
```
mDuo13, line 95: 404 Not Found

src/ripple/protocol/impl/ErrorCodes.cpp
```diff
@@ -26,105 +27,96 @@ namespace RPC {
...
+    {rpcOBJECT_NOT_FOUND,       "objectNotFound",       "The requested object was not found."},
```
mDuo13, line 96: 404

src/ripple/protocol/impl/ErrorCodes.cpp
```diff
@@ -26,105 +27,96 @@ namespace RPC {
...
+    {rpcPUBLIC_MALFORMED,       "publicMalformed",      "Public key is malformed."},
```
mDuo13, line 97: 400

src/ripple/protocol/impl/ErrorCodes.cpp
```diff
@@ -26,105 +27,96 @@ namespace RPC {
...
+    {rpcREPORTING_UNSUPPORTED,  "reportingUnsupported", "Requested operation not supported by reporting mode server"},
```
mDuo13, line 98: 405 Method Not Allowed

src/ripple/protocol/impl/ErrorCodes.cpp
```diff
@@ -26,105 +27,96 @@ namespace RPC {
...
+    {rpcSENDMAX_MALFORMED,      "sendMaxMalformed",     "SendMax amount malformed."},
```
mDuo13, line 99: 400

src/ripple/protocol/impl/ErrorCodes.cpp
```diff
@@ -26,105 +27,96 @@ namespace RPC {
...
+    {rpcSIGNING_MALFORMED,      "signingMalformed",     "Signing of transaction is malformed."},
```
mDuo13, line 100: 400

src/ripple/protocol/impl/ErrorCodes.cpp
```diff
@@ -26,105 +27,96 @@ namespace RPC {
...
+    {rpcSLOW_DOWN,              "slowDown",             "You are placing too much load on the server.", 429},
+    {rpcSRC_ACT_MALFORMED,      "srcActMalformed",      "Source account is malformed."},
+    {rpcSRC_ACT_MISSING,        "srcActMissing",        "Source account not provided."},
```
mDuo13, line 103: 400

src/ripple/protocol/impl/ErrorCodes.cpp
```diff
@@ -26,105 +27,96 @@ namespace RPC {
...
+    {rpcSRC_ACT_NOT_FOUND,      "srcActNotFound",       "Source account not found."},
```
mDuo13, line 104: 404

src/ripple/protocol/impl/ErrorCodes.cpp
```diff
@@ -26,105 +27,96 @@ namespace RPC {
...
+    {rpcSRC_CUR_MALFORMED,      "srcCurMalformed",      "Source currency is malformed."},
+    {rpcSRC_ISR_MALFORMED,      "srcIsrMalformed",      "Source issuer is malformed."},
+    {rpcSTREAM_MALFORMED,       "malformedStream",      "Stream malformed."},
```
mDuo13, line 107: 400 for all three

src/ripple/protocol/impl/ErrorCodes.cpp
```diff
@@ -26,105 +27,96 @@ namespace RPC {
...
+    {rpcNO_NETWORK,             "noNetwork",            "Not synced to the network.", 503},
+    {rpcNO_PERMISSION,          "noPermission",         "You don't have permission for this command.", 401},
+    {rpcNO_PF_REQUEST,          "noPathRequest",        "No pathfinding request in progress."},
+    {rpcOBJECT_NOT_FOUND,       "objectNotFound",       "The requested object was not found."},
+    {rpcPUBLIC_MALFORMED,       "publicMalformed",      "Public key is malformed."},
+    {rpcREPORTING_UNSUPPORTED,  "reportingUnsupported", "Requested operation not supported by reporting mode server"},
+    {rpcSENDMAX_MALFORMED,      "sendMaxMalformed",     "SendMax amount malformed."},
+    {rpcSIGNING_MALFORMED,      "signingMalformed",     "Signing of transaction is malformed."},
+    {rpcSLOW_DOWN,              "slowDown",             "You are placing too much load on the server.", 429},
+    {rpcSRC_ACT_MALFORMED,      "srcActMalformed",      "Source account is malformed."},
```
mDuo13, line 102: 400

src/ripple/protocol/impl/ErrorCodes.cpp
```diff
@@ -26,105 +27,96 @@ namespace RPC {
...
+    {rpcSRC_ACT_MISSING,        "srcActMissing",        "Source account not provided."},
+    {rpcSRC_ACT_NOT_FOUND,      "srcActNotFound",       "Source account not found."},
+    {rpcSRC_CUR_MALFORMED,      "srcCurMalformed",      "Source currency is malformed."},
+    {rpcSRC_ISR_MALFORMED,      "srcIsrMalformed",      "Source issuer is malformed."},
+    {rpcSTREAM_MALFORMED,       "malformedStream",      "Stream malformed."},
+    {rpcTOO_BUSY,               "tooBusy",              "The server is too busy to help you now.", 503},
+    {rpcTXN_NOT_FOUND,          "txnNotFound",          "Transaction not found."},
```
mDuo13, line 109: 404

mDuo13 (commented):
I suggested appropriate HTTP status codes for all the other error codes, but I still think these breaking changes should not be implemented in the v1 API and we probably want to consolidate a lot of these errors into fewer for API v2 anyway

intelliot:
> Thoughts regarding the version flag: if that same flag would also change the behaviour of other version related options, one would never be able to use the previous (current, default) version of the API but _with_ the response codes that allows for failover in reverse proxy environments (for example). Not ideal from a consistency standpoint, but could it be worth triggering this behaviour based on a HTTP request header?

I recommend having HTTP status code changes be the the **only** change in `"api_version": 3`.

Using a special HTTP request header isn't worth the hassle in code, documentation, or increased number of special cases. And [version numbers are free*](https://tom.preston-werner.com/2022/05/23/major-version-numbers-are-not-sacred.html) and we will not run out. Let's use them.

I have updated [XLS-22d API Versioning](https://github.com/XRPLF/XRPL-Standards/discussions/54) to note that changing status codes is a breaking change.

*(Thanks to @thejohnfreeman for this link.)

intelliot:
Note, I'm not saying that this is a perfect solution or avoids all potential problems. We are just trying to make the best trade-off given the goals we have in mind. Using `"api_version": 3` will require clients to adapt to the change that was made for `"api_version": 2`:

https://github.com/XRPLF/rippled/pull/3770

I think clients will have to adapt anyway, because clients may have existing code that expects certain status codes. When they are ready to gain the benefits of automatic failover at the reverse proxy layer, then they should audit / test their code to ensure that they comply with the latest API, including both the new status codes and the signer_list location.

intelliot (commented):
I agree with @mDuo13 's request except I think it should be `"api_version": 3` or later

MarkusTeufelberger:
I'm not too sure if HTTP return codes should reflect the underlying JSON-RPC calls, but the spec there is surprisingly sparse...

scottschurr (commented):
I looked into using the `"api_version"` to gate this feature, but I ran into some awkwardness.  The `"api_version"` is not readily available where I need it.

How would y'all feel about using the `"ripplerpc"` version to gate the feature?  I have easy access to that field where I need it.  `"ripplerpc`" currently supports versions "1.0" and "2.0".  I could bump that to "2.1" or "3.0".  I wrote a quick test case using `"ripplerpc"` as the gate and it worked.

Thoughts?

scottschurr:
I've pushed a commit that adds the HTTP status codes suggested by @mDuo13.  It also gates returning this status codes based on the `"ripplerpc"` field of the original request being `"3.0"` or greater.

Let me know what y'all think.

intelliot:
I think it's fine to have this use `"ripplerpc": "3.0"`. In fact, that might actually be preferable; I hadn't thought of it as an option before. `ripplerpc` affects HTTP, but not WebSockets. And that's good since HTTP status codes don't apply to WebSockets anyway.

intelliot:
I would wait for confirmation from @WietseWind before merging, though :)

intelliot:
Also, it's important that this gets documented on xrpl.org. We should also document the difference between ripplerpc 1.0 and 2.0, if we haven't already.

scottschurr:
Rebased to 1.9.2.  @mDuo13 and @WietseWind, are you good with this version of the code?  Thanks.

WietseWind:
> Rebased to 1.9.2. @mDuo13 and @WietseWind, are you good with this version of the code? Thanks.

Very happy with it :) Thank you!

intelliot (approved):

mDuo13 (commented):
I am on board with applying these changes in the case that users supply "3.0", yes.

I also still need to research and fully understand the differences in "2.0" so I can document those.

scottschurr:
Thanks @mDuo13.  If you think these changes are ready to be committed, could you please mark the pull request as approved?  But, since you want more time to research the "2.0" case, then you may want to withhold approval until that research is complete.  That's fine also.

@cjcobb23, would you like to weigh in on these changes as well?  They may have some consequences in Clio.

mDuo13 (approved):

scottschurr:
Thanks @mDuo13!  @cjcobb23, it's now your call.

cjcobb23 (approved):
LGTM

scottschurr:
Thanks for the review @cjcobb23!  I've squashed and rebased the commit and marked it as passed.

PR #4149, opened 2022-04-21
Fix test failures with newer msvc compilers (related to string aliasing)

greg7mdp:
### Context of Change

When building rippled on Windows with the latest msvc compilers & toolset (newer that Visual Studio 2017), there are string aliasing issues (demonstrated by this [repo](https://github.com/ximinez/staticstring)), where assigning a `constexpr` `const char *` to a `static const char *` actually stores a pointer to a new string, so the two pointers don't compare equal (strcmp still works of course).

This issue is seen only with specific flag combinations. I have seen it with the [`/ZI`](https://docs.microsoft.com/en-us/cpp/build/reference/z7-zi-zi-debug-information-format?view=msvc-170) flag and the [`/Gy-`](https://docs.microsoft.com/en-us/cpp/build/reference/gy-enable-function-level-linking?view=msvc-170) flag which [disables Function-Level Linking](https://docs.microsoft.com/en-us/cpp/build/reference/gy-enable-function-level-linking?view=msvc-170).

In order to avoid the issue altogether regardless if the compiler aliases strings, this change updates the code we don't do string pointer comparisons, but consistently use strcmp.

My suspicion is that the pointer comparisons should work, but we'd have to see where it is specified in the standard, otherwise we are just relying on UB. Anyways I have file a [bug report](https://developercommunity.visualstudio.com/t/assigning-constexpr-char--to-static-cha/10021357?port=1025&fsid=d8283102-25c7-4356-b049-032b8433fba9&entry=problem) with Microsoft, we'll see what they say. I also asked the [question](https://stackoverflow.com/questions/71960681/c-assigning-constexpr-char-to-static-char-copies-the-string-is-it-a) on stack overflow.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### DIFF
--- a/src/ripple/rpc/impl/RPCHelpers.cpp
+++ b/src/ripple/rpc/impl/RPCHelpers.cpp
@@ -696,19 +696,31 @@ parseRippleLibSeed(Json::Value const& value)
 std::optional<Seed>
 getSeedFromRPC(Json::Value const& params, Json::Value& error)
 {
-    // The array should be constexpr, but that makes Visual Studio unhappy.
-    static char const* const seedTypes[]{
-        jss::passphrase.c_str(), jss::seed.c_str(), jss::seed_hex.c_str()};
+    using string_to_seed_t =
+        std::function<std::optional<Seed>(const std::string&)>;
+    using seed_match_t = std::pair<const char*, string_to_seed_t>;
+
+    static seed_match_t const seedTypes[]{
+        {jss::passphrase.c_str(),
+         [](const std::string& s) { return parseGenericSeed(s); }},
+        {jss::seed.c_str(),
+         [](const std::string& s) { return parseBase58<Seed>(s); }},
+        {jss::seed_hex.c_str(), [](const std::string& s) {
+             uint128 i;
+             if (i.parseHex(s))
+                 return std::optional<Seed>(Slice(i.data(), i.size()));
+             return std::optional<Seed>{};
+         }}};
 
     // Identify which seed type is in use.
-    char const* seedType = nullptr;
+    const seed_match_t* seedType = nullptr;
     int count = 0;
-    for (auto t : seedTypes)
+    for (const auto& t : seedTypes)
     {
-        if (params.isMember(t))
+        if (params.isMember(t.first))
         {
             ++count;
-            seedType = t;
+            seedType = &t;
         }
     }
 
@@ -722,28 +734,17 @@ getSeedFromRPC(Json::Value const& params, Json::Value& error)
     }
 
     // Make sure a string is present
-    if (!params[seedType].isString())
+    const auto& param = params[seedType->first];
+    if (!param.isString())
     {
-        error = RPC::expected_field_error(seedType, "string");
+        error = RPC::expected_field_error(seedType->first, "string");
         return std::nullopt;
     }
 
-    auto const fieldContents = params[seedType].asString();
+    auto const fieldContents = param.asString();
 
     // Convert string to seed.
-    std::optional<Seed> seed;
-
-    if (seedType == jss::seed.c_str())
-        seed = parseBase58<Seed>(fieldContents);
-    else if (seedType == jss::passphrase.c_str())
-        seed = parseGenericSeed(fieldContents);
-    else if (seedType == jss::seed_hex.c_str())
-    {
-        uint128 s;
-
-        if (s.parseHex(fieldContents))
-            seed.emplace(Slice(s.data(), s.size()));
-    }
+    std::optional<Seed> seed = seedType->second(fieldContents);
 
     if (!seed)
         error = rpcError(rpcBAD_SEED);
@@ -757,7 +758,6 @@ keypairForSignature(Json::Value const& params, Json::Value& error)
     bool const has_key_type = params.isMember(jss::key_type);
 
     // All of the secret types we allow, but only one at a time.
-    // The array should be constexpr, but that makes Visual Studio unhappy.
     static char const* const secretTypes[]{
         jss::passphrase.c_str(),
         jss::secret.c_str(),
@@ -811,7 +811,8 @@ keypairForSignature(Json::Value const& params, Json::Value& error)
             return {};
         }
 
-        if (secretType == jss::secret.c_str())
+        if (strcmp(secretType, jss::secret.c_str()) ==
+            0)  // don't assume that pointers are the same
         {
             error = RPC::make_param_error(
                 "The secret field is not allowed if " +
@@ -823,7 +824,8 @@ keypairForSignature(Json::Value const& params, Json::Value& error)
     // ripple-lib encodes seed used to generate an Ed25519 wallet in a
     // non-standard way. While we never encode seeds that way, we try
     // to detect such keys to avoid user confusion.
-    if (secretType != jss::seed_hex.c_str())
+    if (strcmp(secretType, jss::seed_hex.c_str()) !=
+        0)  // don't assume that pointers are the same
     {
         seed = RPC::parseRippleLibSeed(params[secretType]);
 
--- a/src/ripple/rpc/impl/RPCHelpers.cpp
+++ b/src/ripple/rpc/impl/RPCHelpers.cpp
@@ -811,8 +811,8 @@ keypairForSignature(Json::Value const& params, Json::Value& error)
             return {};
         }
 
-        if (strcmp(secretType, jss::secret.c_str()) ==
-            0)  // don't assume that pointers are the same
+        // using strcmp as pointers may not match (see  https://bit.ly/3OGvf0E)
+        if (strcmp(secretType, jss::secret.c_str()) == 0)
         {
             error = RPC::make_param_error(
                 "The secret field is not allowed if " +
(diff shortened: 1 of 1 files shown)


src/ripple/rpc/impl/RPCHelpers.cpp
```diff
@@ -696,19 +696,28 @@ parseRippleLibSeed(Json::Value const& value)
...
+                return std::optional<Seed>(Slice(i.data(), i.size()));
+            return std::optional<Seed>{}; }}
+        };
 
     // Identify which seed type is in use.
-    char const* seedType = nullptr;
+    const seed_match_t* seedType = nullptr;
     int count = 0;
-    for (auto t : seedTypes)
+    for (const auto &t : seedTypes)
```
thejohnfreeman, line 715: I think we generally use east const style, but I may be wrong.

src/ripple/rpc/impl/RPCHelpers.cpp
```diff
@@ -811,7 +808,7 @@ keypairForSignature(Json::Value const& params, Json::Value& error)
             return {};
         }
 
-        if (secretType == jss::secret.c_str())
+        if (strcmp(secretType, jss::secret.c_str()) == 0) // don't assume that pointers are the same
```
thejohnfreeman, line 811: Please add a link in a comment somewhere to the bug report you filed.

thejohnfreeman (changes requested):

src/ripple/rpc/impl/RPCHelpers.cpp
greg7mdp: Thanks @thejohnfreeman will do (and also set clang-format to run automatically).

src/ripple/rpc/impl/RPCHelpers.cpp
greg7mdp: Thanks @thejohnfreeman, done.

src/ripple/rpc/impl/RPCHelpers.cpp
greg7mdp: done!

src/ripple/rpc/impl/RPCHelpers.cpp
thejohnfreeman: I meant in the code. When someone comes along and wonders, "why can't I assume the pointers are the same? that doesn't make any sense", it will be helpful to have an explanation right there.

src/ripple/rpc/impl/RPCHelpers.cpp
thejohnfreeman: I'm not sure any changes were committed. They don't seem to show up in this PR. If I may request, please don't force push your branches. Just add another commit and prefix the commit message subject with `[fold]` (team practice). That way we can switch between reviewing just the new updates in isolation or all of the changes together. Every `[fold]`-tagged commit will be squashed when the PR is merged.

greg7mdp: Thanks @thejohnfreeman. Yes the changes are committed. I didn't know about the `[fold]` tag and I will use this from now on. I think if you look at the current diffs you'll see your requested change was made.

src/ripple/rpc/impl/RPCHelpers.cpp
greg7mdp: @thejohnfreeman added link in the code.

src/ripple/rpc/impl/RPCHelpers.cpp
thejohnfreeman: Ok, I'm not sure what changed then. Here is the change I was requesting:

```cpp
    for (auto const& t : seedTypes)
```

Not `const auto& t` or `const auto &t`, but `auto const& t`. I only flagged this one instance, but there are several more in the changeset. The old code was not entirely consistent itself, but it was mostly in this style known as "[east const](https://hackingcpp.com/cpp/design/east_vs_west_const.html)". Don't need to fix the whole file, but any code that is touched should be fixed. clang-format won't fix or flag any of this because it is not a whitespace change, which is all clang-format does.

greg7mdp: @thejohnfreeman thanks for explaining. Is the preference you state documented anywhere?

greg7mdp: east const it is.

thejohnfreeman: Yes, it's here: https://github.com/ripple/rippled/blob/develop/docs/CodingStyle.md

src/ripple/rpc/impl/RPCHelpers.cpp
```diff
@@ -696,19 +696,31 @@ parseRippleLibSeed(Json::Value const& value)
...
+        {jss::seed_hex.c_str(), [](const std::string& s) {
+             uint128 i;
+             if (i.parseHex(s))
+                 return std::optional<Seed>(Slice(i.data(), i.size()));
+             return std::optional<Seed>{};
+         }}};
 
     // Identify which seed type is in use.
-    char const* seedType = nullptr;
+    const seed_match_t* seedType = nullptr;
```
thejohnfreeman, line 716: east const

src/ripple/rpc/impl/RPCHelpers.cpp
```diff
@@ -722,28 +734,17 @@ getSeedFromRPC(Json::Value const& params, Json::Value& error)
     }
 
     // Make sure a string is present
-    if (!params[seedType].isString())
+    const auto& param = params[seedType->first];
```
thejohnfreeman, line 737: east const

src/ripple/rpc/impl/RPCHelpers.cpp
```diff
@@ -811,7 +811,8 @@ keypairForSignature(Json::Value const& params, Json::Value& error)
             return {};
         }
 
-        if (secretType == jss::secret.c_str())
+        // using strcmp as pointers may not match (see  https://bit.ly/3OGvf0E)
```
thejohnfreeman, line 814: In general, I recommend using the long link for long-lasting documentation. I treat short links as temporarily-convenient throwaways.

src/ripple/rpc/impl/RPCHelpers.cpp
```diff
@@ -696,19 +696,31 @@ parseRippleLibSeed(Json::Value const& value)
 std::optional<Seed>
 getSeedFromRPC(Json::Value const& params, Json::Value& error)
 {
-    // The array should be constexpr, but that makes Visual Studio unhappy.
-    static char const* const seedTypes[]{
-        jss::passphrase.c_str(), jss::seed.c_str(), jss::seed_hex.c_str()};
+    using string_to_seed_t =
+        std::function<std::optional<Seed>(const std::string&)>;
```
thejohnfreeman, line 700: east const

src/ripple/rpc/impl/RPCHelpers.cpp
```diff
@@ -696,19 +696,31 @@ parseRippleLibSeed(Json::Value const& value)
...
+    using seed_match_t = std::pair<const char*, string_to_seed_t>;
```
thejohnfreeman, line 701: east const

src/ripple/rpc/impl/RPCHelpers.cpp
```diff
@@ -696,19 +696,31 @@ parseRippleLibSeed(Json::Value const& value)
...
+
+    static seed_match_t const seedTypes[]{
+        {jss::passphrase.c_str(),
+         [](const std::string& s) { return parseGenericSeed(s); }},
+        {jss::seed.c_str(),
+         [](const std::string& s) { return parseBase58<Seed>(s); }},
+        {jss::seed_hex.c_str(), [](const std::string& s) {
```
thejohnfreeman, line 708: east const on each lambda parameter

thejohnfreeman (changes requested):

src/ripple/rpc/impl/RPCHelpers.cpp
greg7mdp: done

src/ripple/rpc/impl/RPCHelpers.cpp
greg7mdp: done

src/ripple/rpc/impl/RPCHelpers.cpp
greg7mdp: done

src/ripple/rpc/impl/RPCHelpers.cpp
greg7mdp: done

src/ripple/rpc/impl/RPCHelpers.cpp
greg7mdp: done

src/ripple/rpc/impl/RPCHelpers.cpp
greg7mdp: done

thejohnfreeman (approved):

PR #4150, opened 2022-04-21
Check triggers cmake error unnecessarily

greg7mdp:
### Context of Change

When using cmake-gui to generate solution files on windows, the default platform is stated to be x64, so it is natural to leave the field unfilled, but even among the proposed choices, Win64 is nowhere to be found, we see only x64.

So whether we leave the field blank, or select x64 in the dropdown, cmake generation fails because it checks for the presence of `Win64`.

This change makes the generation of solution files more user friendly on windows, by not causing an error unless we find `Win32`. 

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### DIFF
--- a/Builds/CMake/RippledSanity.cmake
+++ b/Builds/CMake/RippledSanity.cmake
@@ -72,10 +72,8 @@ if ("${CMAKE_CURRENT_SOURCE_DIR}" STREQUAL "${CMAKE_BINARY_DIR}")
     "directory from ${CMAKE_CURRENT_SOURCE_DIR} and try building in a separate directory.")
 endif ()
 
-if ("${CMAKE_GENERATOR}" MATCHES "Visual Studio" AND
-   ("${CMAKE_GENERATOR_PLATFORM}" MATCHES .*Win32.*))
-  message (FATAL_ERROR
-    "Visual Studio 32-bit build is not supported. Use -G\"${CMAKE_GENERATOR} Win64\"")
+if (MSVC AND CMAKE_GENERATOR_PLATFORM STREQUAL "Win32")
+  message (FATAL_ERROR "Visual Studio 32-bit build is not supported.")
 endif ()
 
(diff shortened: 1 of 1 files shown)


greg7mdp:
image

legleux:
For Visual Studio 2017 and prior, the default wasn't x64 so arch couldn't leave it blank.
I vote we don't even bother with 2017, but only use 2019 and 2022.

greg7mdp:
>  I vote we don't even bother with 2017, but only use 2019 and 2022.

sounds good to me!

legleux (approved):

Builds/CMake/RippledSanity.cmake
```diff
@@ -73,7 +73,7 @@ if ("${CMAKE_CURRENT_SOURCE_DIR}" STREQUAL "${CMAKE_BINARY_DIR}")
 endif ()
 
 if ("${CMAKE_GENERATOR}" MATCHES "Visual Studio" AND
-    NOT ("${CMAKE_GENERATOR}" MATCHES .*Win64.*))
+   ("${CMAKE_GENERATOR_PLATFORM}" MATCHES .*Win32.*))
   message (FATAL_ERROR
     "Visual Studio 32-bit build is not supported. Use -G\"${CMAKE_GENERATOR} Win64\"")
```
thejohnfreeman, line 78: How about this instead?

```cmake
if (MSVC AND CMAKE_GENERATOR_PLATFORM STREQUAL "Win32")
  message (FATAL_ERROR
    "Visual Studio 32-bit build is not supported. Use -A x64")
```

greg7mdp: Yes, looks better. I'll do a sanity check and update it.

greg7mdp: @thejohnfreeman I did the change, removing the `-A x64` as it does not seem to work.

PR #4152, opened 2022-04-23
Address performance issues reported with 1.9.0

nbougalis:


thejohnfreeman (approved):

greg7mdp:
I think I see a potential deadlock. Suppose one of the threads is interrupted right after checking `isStopping()` (see image below) and then another thread calls `Database::stop()`, sets `readStopping_` to true, takes the lock and calls `readCondVar_.notify_all();`.

When the thread resumes, it will wait forever on `readCondVar_.wait(lock);`, right?

image

I suggest adding the two highlighted lines below:


image

src/ripple/nodestore/impl/Database.cpp
```diff
@@ -68,14 +74,20 @@ Database::Database(
                         std::unique_lock<std::mutex> lock(readLock_);
 
                         if (read_.empty())
+                        {
+                            runningThreads_--;
                             readCondVar_.wait(lock);
+                            runningThreads_++;
+                        }
```
greg7mdp, line 81: I think I see a potential deadlock. Suppose one of the threads is interrupted right after checking `while (!isStopping())`  and then another thread calls `Database::stop()`, sets `readStopping_` to true, takes the lock and calls `readCondVar_.notify_all();.`

When the thread resumes, it will wait forever on `readCondVar_.wait(lock)`, I think?

greg7mdp: I suggest adding :
```
if (isStopping())
    continue;
```

just after taking the lock at line 74

src/ripple/nodestore/impl/Database.cpp
```diff
@@ -68,14 +74,20 @@ Database::Database(
                         std::unique_lock<std::mutex> lock(readLock_);
```
greg7mdp, line 74:
```suggestion
                        std::unique_lock<std::mutex> lock(readLock_);
                        if (isStopping())
                            continue;                        
```

PR #4157, opened 2022-05-05
Add featureImmediateOfferKilled for tfImmediateOrCancel offers:

scottschurr:

Fixes https://github.com/ripple/rippled/issues/4115

### Context of Change

@mDuo13 noted:
> It's confusing and unintuitive to see an "immediate or cancel" transaction treated as "successful" when it did nothing because the "or cancel" part applied. A different result code makes it apparent at a glance what happened.

The approach taken to this fix reuses the `tecKILLED` code.  I hesitated on this because the descriptive text associated with `tecKILLED` needs to change when applied in this new context.  And that description change could not easily be protected by the amendment.  However @mDuo13 pointed out:

> ... no amendment is necessary for the description change. Per https://xrpl.org/transaction-results.html#immediate-response

So this gave us the leash to make make the change minimally invasive.  Even though, admittedly, it requires an amendment.

For me, the strongest motivation for this change is as follows (also from @mDuo13):

> A major benefit of this change is that you could write code to assume that, if the code is tesSUCCESS (and the transaction is post-amendment) then there will be balance changes of some kind in the metadata, and if the code is tecKILLED then you don't have to look at the metadata to know that no balance change occurred (other than burning the tx cost of course). 

### Type of Change

- [x] Amendment (Functionality waits for amendment to pass. Once passed may cause amendment blocking.)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected).
- [x] Tests (You added tests for code that already exists, or your new feature included in this PR).

## Before / After

**Before**: an `OfferCreate` transaction with the `tfImmediateOrCancel` flag set that causes no funds to move returns `tesSUCCESS`.

**After:** an `OfferCreate` transaction with the `tfImmediateOrCancel` flag set that causes no funds to move returns `tecKILLED`.

### DIFF
--- a/src/ripple/app/tx/impl/CreateOffer.cpp
+++ b/src/ripple/app/tx/impl/CreateOffer.cpp
@@ -1107,6 +1107,12 @@ CreateOffer::applyGuts(Sandbox& sb, Sandbox& sbCancel)
     if (bImmediateOrCancel)
     {
         JLOG(j_.trace()) << "Immediate or cancel: offer canceled";
+        if (!crossed && sb.rules().enabled(featureImmediateOfferKilled))
+            // If the ImmediateOfferKilled amendment is enabled, any
+            // ImmediateOrCancel offer that transfers absolutely no funds
+            // returns tecKILLED rather than tesSUCCESS.  Motivation for the
+            // change is here: https://github.com/ripple/rippled/issues/4115
+            return {tecKILLED, false};
         return {tesSUCCESS, true};
     }
 
--- a/src/ripple/protocol/Feature.h
+++ b/src/ripple/protocol/Feature.h
@@ -74,7 +74,7 @@ namespace detail {
 // Feature.cpp. Because it's only used to reserve storage, and determine how
 // large to make the FeatureBitset, it MAY be larger. It MUST NOT be less than
 // the actual number of amendments. A LogicError on startup will verify this.
-static constexpr std::size_t numFeatures = 51;
+static constexpr std::size_t numFeatures = 52;
 
 /** Amendments that this server supports and the default voting behavior.
    Whether they are enabled depends on the Rules defined in the validated
@@ -338,6 +338,7 @@ extern uint256 const fixNFTokenDirV1;
 extern uint256 const fixNFTokenNegOffer;
 extern uint256 const featureNonFungibleTokensV1_1;
 extern uint256 const fixTrustLinesToSelf;
+extern uint256 const featureImmediateOfferKilled;
 
 }  // namespace ripple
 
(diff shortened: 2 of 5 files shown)


mDuo13 (approved):

mDuo13:
25 lines, including tests? Now that's impressive work. I know that keeping it simple is harder than it looks, and usually involves unseen work that's discarded on the way. So, thanks for the elegant solution.

seelabs (approved):
Nice job on this. LGTM :+1:

src/ripple/app/tx/impl/CreateOffer.cpp
```diff
@@ -1107,6 +1107,12 @@ CreateOffer::applyGuts(Sandbox& sb, Sandbox& sbCancel)
     if (bImmediateOrCancel)
     {
         JLOG(j_.trace()) << "Immediate or cancel: offer canceled";
+        if (!crossed && sb.rules().enabled(featureImmediateOfferKilled))
+            // If the ImmediateOfferKilled amendment is enabled, any
+            // ImmediateOrCancel offer that transfers absolutely no funds
+            // returns tecKILLED rather than tesSUCCESS.  Motivation for the
+            // change is here: https://github.com/ripple/rippled/issues/4115
+            return {tecKILLED, false};
```
nbougalis, line 1115: This is broken: it basically changes the semantics of **`Immediate or Cancel`** to be those of **`Fill or Kill`**.

A "Fill or Kill" order means: "if the offer can't be entirely filled by crossing against existing offers, abort. Never place an order on the order book."

An "Immediate or Cancel" order, on the other hand, means: "fill as much as you can by crossing against existing offers. If the offer isn't fully filled, that's fine, cancel the rest and never put an offer on the book."

So returning `false` from here is a bug. It should return `true` because you **want** to apply the executed crossings, but you don't want to go any further (i.e. to place the offer on the book). 

Given the difference between FoK and IoC, I think that `tecKILLED` is the wrong error code here; it should be `tecPARTIALLY_FILLED` or somesuch.

nbougalis (changes requested):
Veto. @scottschurr: this breaks the "Immediate or Cancel" semantics. See comment below.

src/ripple/app/tx/impl/CreateOffer.cpp
scottschurr: @nbougais, are you _sure_ this is broken?  Note the check for `!crossed` in the conditional.  So we should `return {tecKILLED, false}` only in the case where _no_ funds are transferred.  If _any_ funds were transferred, then we continue to `return {tesSUCCESS, true}` as before.  At least that's what I read `crossed` as representing.

What am I missing?

scottschurr:
Rebased to 1.9.2.

PR #4134, opened 2022-04-06
update boost version requirement to 1.71

greg7mdp:
…ilding on M1 mac

### Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Refactor (non-breaking change that only restructures code)
- [ ] Tests (You added tests for code that already exists, or your new feature included in this PR)
- [x] Documentation Updates
- [ ] Release

PR #4158, opened 2022-05-10
Propose 1.9.1-b1 

manojsdoshi:
If merged this PR will:

close #4155
close #4152
close #4150
close #4149
close #4139 
close #4134 
close #4111
close #4097
close #4021
close #3965


MarkusTeufelberger:
Please don't introduce amendments in patch versions...

https://semver.org/

MarkusTeufelberger:
I still changes the API in a backwards-compatible way. Just call it 1.10.0, numbers are cheap and plentiful. :-)

MarkusTeufelberger:
I see this rather pragmatically, a release is a release and unfortunately there is usually far less focus on making sure version numbers are meaningful in this space in general to rely on some real long term support or compatibility (imagine e.g. `rippled` 1.9.0 making it into Ubuntu LTS 22.04 - I'm skeptical that it will still be able to sync after the 5-9 years of support run out (https://endoflife.date/ubuntu). I'ts not like as if there are several branches from the past that are still maintained.

It was just an observation, no "omg, the sky is falling" kinda scenario. :-)

PR #4160, opened 2022-05-10
Ensure that `rngfill` generates the requested amount of randomness:

nbougalis:
One of the two versions of the `rngfill` function accepts a pointer to a buffer and a size (in bytes). The function aims to fill the provided `buffer` with `size` random bytes. It does this in chunks of 8 bytes, for long as possible, and then fills any left-over gap one byte at a time.

To avoid an annoying and incorrect warning about a potential buffer overflow in the "trailing write", commit 78bc2727f707485d59364d7497dec1f464c2d356 used a `#pragma` to instruct the compiler to not generate the incorrect diagnostic. Unfortunately, this change _also_ eliminated the trailing write code, which means that, under some cases, the `rngfill` function would generate between 1 and 7 fewer random bytes than requested.

This problem would only manifest on builds that do not define `__GNUC__` which, as of this writing, means MSVC.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### DIFF
--- a/src/ripple/beast/utility/rngfill.h
+++ b/src/ripple/beast/utility/rngfill.h
@@ -32,6 +32,7 @@ void
 rngfill(void* buffer, std::size_t bytes, Generator& g)
 {
     using result_type = typename Generator::result_type;
+
     while (bytes >= sizeof(result_type))
     {
         auto const v = g();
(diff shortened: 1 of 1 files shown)


greg7mdp (approved):
Looks good to me. That whole business with the second part could be avoided with a loop `while (bytes > 0)` and `bytes -= std::min(bytes, sizeof(v))`

HowardHinnant (approved):

greg7mdp:
How come my approval doesn't seem to count? Only 1 approval shown?

HowardHinnant:
Not sure.  I just added you to https://github.com/orgs/ripple/teams/rippled_pr_reminder/members.  That might make a difference on future approvals.

greg7mdp:
> Not sure. I just added you to https://github.com/orgs/ripple/teams/rippled_pr_reminder/members. That might make a difference on future approvals.

Hey @HowardHinnant, I'd be surprised if this helped, but thanks anyways!

nbougalis:
> Looks good to me. That whole business with the second part could be avoided with a loop `while (bytes > 0)` and `bytes -= std::min(bytes, sizeof(v))`

:thinking: maybe something to consider when this code gets refactored. It would be really nice C++ had better randomness facilities.

nbougalis:
> How come my approval doesn't seem to count? Only 1 approval shown?

I think it's because you're not in the right "group". I can look into adding you, Greg, but as the repo is going to soon be migrating over to the XRP Ledger Foundation, I don't think it'll make much difference.

nbougalis:
Merged as 7ca1f78446e1ec940dd4af85be3a1ab61719d636

PR #4167, opened 2022-05-16
Rename SHAMapStoreImp::stopping() to SHAMapStoreImp::handleHealth()

scottschurr:
### Context of Change

A recent change modified the name of a member function from `SHAMapStoreImp::health()` to `SHAMapStoreImp::stopping()`.  Personally, I find the name change to `stopping()` misleads me when I read the source code.  The function `stopping()` performs the job of looking at the health of rippled and _sleeping the thread_ if rippled is not sufficiently healthy.

I would _never_ expect a function named `stopping()` that returns a `bool` to sleep the thread.

This pull request renames that function to `SHAMapStoreImp::handleHealth()`.  To assist with readability, `handleHealth()` returns an `enum` of either `stopping` or `keepGoing`.  If `handleHealth()` returns `stopping`, then rippled is attempting to shut down and appropriate actions should be taken.

### Type of Change

### DIFF
--- a/src/ripple/app/misc/SHAMapStoreImp.cpp
+++ b/src/ripple/app/misc/SHAMapStoreImp.cpp
@@ -269,7 +269,7 @@ SHAMapStoreImp::copyNode(std::uint64_t& nodeCount, SHAMapTreeNode const& node)
         true);
     if (!(++nodeCount % checkHealthInterval_))
     {
-        if (stopping())
+        if (healthWait() == stopping)
             return false;
     }
 
@@ -327,7 +327,7 @@ SHAMapStoreImp::run()
 
         bool const readyToRotate =
             validatedSeq >= lastRotated + deleteInterval_ &&
-            canDelete_ >= lastRotated - 1 && !stopping();
+            canDelete_ >= lastRotated - 1 && healthWait() == keepGoing;
 
         // Make sure we don't delete ledgers currently being
         // imported into the ShardStore
@@ -359,7 +359,7 @@ SHAMapStoreImp::run()
                 << ledgerMaster_->getValidatedLedgerAge().count() << 's';
 
             clearPrior(lastRotated);
-            if (stopping())
+            if (healthWait() == stopping)
                 return;
 
             JLOG(journal_.debug()) << "copying ledger " << validatedSeq;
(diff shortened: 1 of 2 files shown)


nbougalis (changes requested):
I agree that `stopping` is a bad name for this. However, I'm not convinced that `handleHealth` (and the corresponding `enum`) is much better. Even `healthy()` or `isHealthy()` is a problem, since it doesn't describe what the function really does.

The function's primary purpose is to determine whether we should stop (as signaled by an external object) _and_, as a side-efffect, updates some internal state (also doing some ugly stuff with the lock at the same time, which I'm not sure is needed).

Can we take this opportunity to determine whether we even need to grab the `mutex_ here`? For example, why isn't `stop_` a `std::atomic<bool>` (and, once we get C++20, `std::atomic_flag`, which has a number of additional goodies)?

scottschurr:
@nbougalis, I disagree that the function's primary purpose is to determine whether we should stop.  I _believe_ the function's primary purpose is to sleep the thread if rippled is currently unhealthy.  The thread may be awakened because the server is stopping, but that is not the primary point of the function.  Expand the diff to show the internals of the function so you can see what's going on.

I'm not arguing that `handleHealth` is the best name.  But neither `healthy` nor `stopping` represent the intent behind the function as I currently understand it.  As it stands the return value from the function is a side effect of the primary purpose of the function.

greg7mdp:
> @scottschurr I'd love to have the word `wait` in the name somewhere - or something that implies that this function can sleep.

or `yield()`?

scottschurr:
Maybe `waitForHealth()`?  If that has too many characters, possibly `healthWait()`?

seelabs:
@scottschurr I'm fine with either of those. It's hard to name this function because it does two things - it waits for a condition and it returns if we're stopping. I understand why it does that, but it makes it hard to put a name to it. Anyway, either of those names sounds good to me. Thanks!

scottschurr:
Rebased to 1.9.2.

gregtatcam (approved):
👍 LGTM

mtrippled (approved):

scottschurr:
@nbougalis, this pull request how has two (non-authorized) approvals, and I've squashed the commits.  So I think this pull request is ready to go.  I'll have to leave it to you to mark the pull request as passed, since I no longer have those permissions.  Thanks.

scottschurr:
Rebased to 1.10.0-b1.  @nbougalis, I believe this pull request is ready to be marked as passed.  Please let me know if it needs any changes.  Thanks.

PR #4163, opened 2022-05-12
include `assert.h` in rngfill.h

natenichols:
## High Level Overview of Change
xrpl_core was not building correctly in Clio. Including `assert.h` fixes this.

### Context of Change
As of rippled version `1.9.1-rc1`, I would see the following error messages when I tried to build clio. 

```
In file included from /space/build/clio/src/rpc/handlers/Random.cpp:1:
/space/build/clio/build/unix_makefiles/GNU_9.3.0/Debug/rippled-src/src/ripple/beast/utility/rngfill.h: In function 'void beast::rngfill(void*, std::size_t, Generator&)':
/space/build/clio/build/unix_makefiles/GNU_9.3.0/Debug/rippled-src/src/ripple/beast/utility/rngfill.h:44:5: error: there are no arguments to 'assert' that depend on a template parameter, so a declaration of 'assert' must be available [-fpermissive]
   44 |     assert(bytes < sizeof(result_type));
      |     ^~~~~~
/space/build/clio/build/unix_makefiles/GNU_9.3.0/Debug/rippled-src/src/ripple/beast/utility/rngfill.h:44:5: note: (if you use '-fpermissive', G++ will accept your code, but allowing the use of an undeclared name is deprecated)
/space/build/clio/build/unix_makefiles/GNU_9.3.0/Debug/rippled-src/src/ripple/beast/utility/rngfill.h: In instantiation of 'void beast::rngfill(void*, std::size_t, Generator&) [with Generator = ripple::csprng_engine; std::size_t = long unsigned int]':
/space/build/clio/src/rpc/handlers/Random.cpp:11:68:   required from here
/space/build/clio/build/unix_makefiles/GNU_9.3.0/Debug/rippled-src/src/ripple/beast/utility/rngfill.h:44:11: error: 'assert' was not declared in this scope; did you mean 'mpl_::assert'?
   44 |     assert(bytes < sizeof(result_type));
      |     ~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |     mpl_::assert
```

### Type of Change
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Refactor (non-breaking change that only restructures code)
- [ ] Tests (You added tests for code that already exists, or your new feature included in this PR)
- [ ] Documentation Updates
- [ ] Release

PR #4169, opened 2022-05-18
Proposed 1.9.1-rc2

manojsdoshi:
If merged will 

close #4163


nbougalis (changes requested):
The commit message needs to be fixed: there's a "FOLD" commit under the fold. Just reword to:

>Add missing <cassert> include directive in beast/rngfill.h

Other than that, looks good. :+1: from me.

ximinez (approved):
Windows build looks good. Synced in about 800 seconds.

nbougalis (approved):

PR #4173, opened 2022-05-23
Set cluster timer only when in a cluster

nbougalis:
Avoid doing cluster-related work if the server is not configured to be part of a cluster.

### Type of Change

- [ ] Documentation Updates
- [ ] Release


pwang200 (approved):
Looks good!

PR #4174, opened 2022-05-23
Improve wrapper around OpenSSL RAND

nbougalis:
### Type of Change

- [ ] Documentation Updates
- [ ] Release

### DIFF
--- a/src/ripple/crypto/csprng.h
+++ b/src/ripple/crypto/csprng.h
@@ -39,9 +39,6 @@ class csprng_engine
 private:
     std::mutex mutex_;
 
-    void
-    mix(void* buffer, std::size_t count, double bitsPerByte);
-
 public:
     using result_type = std::uint64_t;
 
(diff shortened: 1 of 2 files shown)


intelliot:
Reopening because it looks like this was merged with zero reviews. @ximinez can you confirm that you reviewed this change as part of https://github.com/XRPLF/rippled/pull/4270? 

If so, you can put an approval on this PR. 

I know it has already been merged, but we'd like to confirm approval since this change (if not reverted) will be included in the 1.10 release.

drlongle:
LGTM!

drlongle (approved):

scottschurr (approved):
I'm not an expert with openSSL RAND, but these changes look reasonable to me.

intelliot:
- There used to be a lock around OpenSSL code. OpenSSL internally implements a lock for this, so the other lock is no longer needed.
- Super minor performance improvement when OpenSSL randomness is used, which is not often.

PR #4181, opened 2022-05-27
Improve AccountID string conversion caching

nbougalis:
Caching the base58check encoded version of an `AccountID` has performance advantages, because because of the computationally heavy cost associated with the conversion, which requires the application of SHA-256 twice.

This commit makes the cache significantly more efficient in terms of memory used: it maps an `AccountID` to `char*` instead of an `std::string` (eliminating the need for dynamic memory allocation and the fixed overhead of `std::string`) and effectively eliminates runtime allocation and deallocation of memory by grabbing a single, large block of memory at initialization.

The eviction policy is also improved and ensures that the cache, once full, will always remain full. Although not currently coded the structure of the cache is such that it can be used to easily evict the least-recently-used entry without additional overhead.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### DIFF
--- a/src/ripple/app/paths/PathRequest.cpp
+++ b/src/ripple/app/paths/PathRequest.cpp
@@ -552,9 +552,16 @@ PathRequest::findPaths(
             continueCallback);
         mContext[issue] = ps;
 
-        auto& sourceAccount = !isXRP(issue.account)
-            ? issue.account
-            : isXRP(issue.currency) ? xrpAccount() : *raSrcAccount;
+        auto const& sourceAccount = [&] {
+            if (!isXRP(issue.account))
+                return issue.account;
+
+            if (isXRP(issue.currency))
+                return xrpAccount();
+
+            return *raSrcAccount;
+        }();
+
         STAmount saMaxAmount = saSendMax.value_or(
             STAmount({issue.currency, sourceAccount}, 1u, 0, true));
 
@@ -675,10 +682,8 @@ PathRequest::doUpdate(
             destCurrencies.append(to_string(c));
     }
 
-    newStatus[jss::source_account] =
-        app_.accountIDCache().toBase58(*raSrcAccount);
-    newStatus[jss::destination_account] =
-        app_.accountIDCache().toBase58(*raDstAccount);
+    newStatus[jss::source_account] = toBase58(*raSrcAccount);
+    newStatus[jss::destination_account] = toBase58(*raDstAccount);
     newStatus[jss::destination_amount] = saDstAmount.getJson(JsonOptions::none);
     newStatus[jss::full_reply] = !fast;
 
--- a/src/ripple/core/impl/Config.cpp
+++ b/src/ripple/core/impl/Config.cpp
@@ -108,26 +108,27 @@ namespace ripple {
 
 // clang-format off
 // The configurable node sizes are "tiny", "small", "medium", "large", "huge"
-inline constexpr std::array<std::pair<SizedItem, std::array<int, 5>>, 12>
+inline constexpr std::array<std::pair<SizedItem, std::array<int, 5>>, 13>
 sizedItems
 {{
     // FIXME: We should document each of these items, explaining exactly
     //        what they control and whether there exists an explicit
     //        config option that can be used to override the default.
 
-    //                                tiny    small   medium    large     huge
-    {SizedItem::sweepInterval,   {{     10,      30,      60,      90,     120 }}},
-    {SizedItem::treeCacheSize,   {{ 262144,  524288, 2097152, 4194304, 8388608 }}},
-    {SizedItem::treeCacheAge,    {{     30,      60,      90,     120,     900 }}},
-    {SizedItem::ledgerSize,      {{     32,      32,      64,     256,     384 }}},
-    {SizedItem::ledgerAge,       {{     30,      60,     180,     300,     600 }}},
-    {SizedItem::ledgerFetch,     {{      2,       3,       4,       5,       8 }}},
-    {SizedItem::hashNodeDBCache, {{      4,      12,      24,      64,     128 }}},
-    {SizedItem::txnDBCache,      {{      4,      12,      24,      64,     128 }}},
-    {SizedItem::lgrDBCache,      {{      4,       8,      16,      32,     128 }}},
-    {SizedItem::openFinalLimit,  {{      8,      16,      32,      64,     128 }}},
-    {SizedItem::burstSize,       {{      4,       8,      16,      32,      48 }}},
-    {SizedItem::ramSizeGB,       {{      8,      12,      16,      24,      32 }}},
+    //                                   tiny    small   medium    large     huge
+    {SizedItem::sweepInterval,      {{     10,      30,      60,      90,     120 }}},
+    {SizedItem::treeCacheSize,      {{ 262144,  524288, 2097152, 4194304, 8388608 }}},
+    {SizedItem::treeCacheAge,       {{     30,      60,      90,     120,     900 }}},
+    {SizedItem::ledgerSize,         {{     32,      32,      64,     256,     384 }}},
+    {SizedItem::ledgerAge,          {{     30,      60,     180,     300,     600 }}},
+    {SizedItem::ledgerFetch,        {{      2,       3,       4,       5,       8 }}},
+    {SizedItem::hashNodeDBCache,    {{      4,      12,      24,      64,     128 }}},
+    {SizedItem::txnDBCache,         {{      4,      12,      24,      64,     128 }}},
+    {SizedItem::lgrDBCache,         {{      4,       8,      16,      32,     128 }}},
+    {SizedItem::openFinalLimit,     {{      8,      16,      32,      64,     128 }}},
+    {SizedItem::burstSize,          {{      4,       8,      16,      32,      48 }}},
+    {SizedItem::ramSizeGB,          {{      8,      12,      16,      24,      32 }}},
+    {SizedItem::accountIdCacheSize, {{  20047,   50053,   77081,  150061,  300007 }}}
 }};
 
 // Ensure that the order of entries in the table corresponds to the
--- a/src/ripple/protocol/impl/AccountID.cpp
+++ b/src/ripple/protocol/impl/AccountID.cpp
@@ -17,17 +17,95 @@
 */
 //==============================================================================
 
+#include <ripple/basics/hardened_hash.h>
+#include <ripple/basics/spinlock.h>
 #include <ripple/protocol/AccountID.h>
 #include <ripple/protocol/PublicKey.h>
 #include <ripple/protocol/digest.h>
 #include <ripple/protocol/tokens.h>
+#include <array>
 #include <cstring>
+#include <mutex>
 
 namespace ripple {
+
+namespace detail {
+
+/** Caches the base58 representations of AccountIDs */
+class AccountIdCache
+{
+private:
+    struct CachedAccountID
+    {
+        AccountID id;
+        char encoding[40] = {0};
+    };
+
+    // The actual cache
+    std::vector<CachedAccountID> cache_;
+
+    // We use a hash function designed to resist algorithmic complexity attacks
+    hardened_hash<> hasher_;
+
+    // 64 spinlocks, packed into a single 64-bit value
+    std::atomic<std::uint64_t> locks_ = 0;
+
+public:
+    AccountIdCache(std::size_t count) : cache_(count)
+    {
+        // This is non-binding, but we try to avoid wasting memory that
+        // is caused by overallocation.
+        cache_.shrink_to_fit();
+    }
+
+    std::string
+    toBase58(AccountID const& id)
+    {
+        auto const index = hasher_(id) % cache_.size();
+
+        packed_spinlock sl(locks_, index % 64);
+
+        {
+            std::lock_guard lock(sl);
+
+            // The check against the first character of the encoding ensures
+            // that we don't mishandle the case of the all-zero account:
+            if (cache_[index].encoding[0] != 0 && cache_[index].id == id)
+                return cache_[index].encoding;
+        }
+
+        auto ret =
+            encodeBase58Token(TokenType::AccountID, id.data(), id.size());
+
+        assert(ret.size() <= 38);
+
+        {
+            std::lock_guard lock(sl);
+            cache_[index].id = id;
+            std::strcpy(cache_[index].encoding, ret.c_str());
+        }
+
+        return ret;
+    }
+};
+
+}  // namespace detail
+
+static std::unique_ptr<detail::AccountIdCache> accountIdCache;
+
+void
+initAccountIdCache(std::size_t count)
+{
+    if (!accountIdCache && count != 0)
+        accountIdCache = std::make_unique<detail::AccountIdCache>(count);
+}
 
 std::string
 toBase58(AccountID const& v)
 {
+    if (accountIdCache)
+        return accountIdCache->toBase58(v);
+
     return encodeBase58Token(TokenType::AccountID, v.data(), v.size());
 }
 
@@ -110,54 +188,6 @@ to_issuer(AccountID& issuer, std::string const& s)
         return false;
     issuer = *account;
     return true;
-}
-
-//------------------------------------------------------------------------------
-
-/*  VFALCO NOTE
-    An alternate implementation could use a pair of insert-only
-    hash maps that each use a single large memory allocation
-    to store a fixed size hash table and all of the AccountID/string
-    pairs laid out in memory (wouldn't use std::string here just a
-    length prefixed or zero terminated array). Possibly using
-    boost::intrusive as the basis for the unordered container.
-    This would cut down to one allocate/free cycle per swap of
-    the map.
-*/
-
-AccountIDCache::AccountIDCache(std::size_t capacity) : capacity_(capacity)
-{
-    m1_.reserve(capacity_);
-}
-
-std::string
-AccountIDCache::toBase58(AccountID const& id) const
-{
-    std::lock_guard lock(mutex_);
-    auto iter = m1_.find(id);
-    if (iter != m1_.end())
-        return iter->second;
-    iter = m0_.find(id);
-    std::string result;
-    if (iter != m0_.end())
-    {
-        result = iter->second;
-        // Can use insert-only hash maps if
-        // we didn't erase from here.
-        m0_.erase(iter);
-    }
-    else
-    {
-        result = ripple::toBase58(id);
-    }
-    if (m1_.size() >= capacity_)
-    {
-        m0_ = std::move(m1_);
-        m1_.clear();
-        m1_.reserve(capacity_);
-    }
-    m1_.emplace(id, result);
-    return result;
 }
 
 }  // namespace ripple
--- a/src/ripple/app/ledger/AcceptedLedger.cpp
+++ b/src/ripple/app/ledger/AcceptedLedger.cpp
@@ -31,11 +31,9 @@ AcceptedLedger::AcceptedLedger(
     transactions_.reserve(256);
 
     auto insertAll = [&](auto const& txns) {
-        auto const& idcache = app.accountIDCache();
-
         for (auto const& item : txns)
             transactions_.emplace_back(std::make_unique<AcceptedLedgerTx>(
-                ledger, item.first, item.second, idcache));
+                ledger, item.first, item.second));
     };
 
     if (app.config().reporting())
--- a/src/ripple/app/ledger/AcceptedLedgerTx.cpp
+++ b/src/ripple/app/ledger/AcceptedLedgerTx.cpp
@@ -28,8 +28,7 @@ namespace ripple {
 AcceptedLedgerTx::AcceptedLedgerTx(
     std::shared_ptr<ReadView const> const& ledger,
     std::shared_ptr<STTx const> const& txn,
-    std::shared_ptr<STObject const> const& met,
-    AccountIDCache const& accountCache)
+    std::shared_ptr<STObject const> const& met)
     : mTxn(txn)
     , mMeta(txn->getTransactionID(), ledger->seq(), *met)
     , mAffected(mMeta.getAffectedAccounts())
@@ -52,7 +51,7 @@ AcceptedLedgerTx::AcceptedLedgerTx(
     {
         Json::Value& affected = (mJson[jss::affected] = Json::arrayValue);
         for (auto const& account : mAffected)
-            affected.append(accountCache.toBase58(account));
+            affected.append(toBase58(account));
     }
 
     if (mTxn->getTxnType() == ttOFFER_CREATE)
--- a/src/ripple/app/ledger/AcceptedLedgerTx.h
+++ b/src/ripple/app/ledger/AcceptedLedgerTx.h
@@ -46,8 +46,7 @@ public:
     AcceptedLedgerTx(
         std::shared_ptr<ReadView const> const& ledger,
         std::shared_ptr<STTx const> const&,
-        std::shared_ptr<STObject const> const&,
-        AccountIDCache const&);
+        std::shared_ptr<STObject const> const&);
 
     std::shared_ptr<STTx const> const&
     getTxn() const
--- a/src/ripple/app/main/Application.cpp
+++ b/src/ripple/app/main/Application.cpp
@@ -177,7 +177,6 @@ public:
     NodeStoreScheduler m_nodeStoreScheduler;
     std::unique_ptr<SHAMapStore> m_shaMapStore;
     PendingSaves pendingSaves_;
-    AccountIDCache accountIDCache_;
     std::optional<OpenLedger> openLedger_;
 
     NodeCache m_tempNodeCache;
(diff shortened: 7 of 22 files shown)


greg7mdp (changes requested):

src/ripple/protocol/impl/AccountID.cpp
```diff
@@ -25,9 +25,135 @@
...
 namespace ripple {
 
+#ifndef RIPPLED_ACCOUNTID_CACHE_SIZE
+#define RIPPLED_ACCOUNTID_CACHE_SIZE 1000000
+#endif
+
+namespace detail {
+
+/** Caches the base58 representations of AccountIDs */
+class AccountIdCache
```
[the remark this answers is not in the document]
greg7mdp: Please ignore the fact that I "requested changes". I can't find a way to undo the request.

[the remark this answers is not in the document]
greg7mdp: I believe that we still have the `std::string` new/delete overhead, because `encodeBase58Token` returns a std::string (which is destroyed immediately).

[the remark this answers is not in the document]
nbougalis: Thanks for the review!

>If, instead of storing a char * as the mapped_type, we store the character buffer itself (for ex std::array<char, size_>), we wouldn't need to do the manual memory management of the memory block, and the number of allocations would be the same (with the benefit that we wouldn't have to preallocate a giant memory block at the beginning).

I agree that using `std::array` would be nice to avoid the need for the memory allocation/deallocation but I don't think we can since the size of the buffer is not known at compile time. Unless I am missing something?

I could have used a `std::vector` really, but given that we're only doing the allocation once at start-up and (effectively) never deallocating (the `std::unique_ptr` will go out of scope, if the process shuts down cleanly, but by then we're well into shutting down). I think the "risk" and the programmer overhead of managing the lifetime is basically zero.


>When we don't have a cache hit, we still have the std::string new/delete overhead, because encodeBase58Token returns a std::string (which is destroyed immediately). It would be nice if we had a version that used a provided buffer.

Agreed, hence the [comment](https://github.com/ripple/rippled/pull/4181/files#diff-e5fa0ea64bdeab9635beb8f04eeb2fc076d0fa6df2e133a01209ab3ff33905eaR126-R127) I left to that effect. I think it's worth adding such an API, but I didn't want to do this as part of this commit.


>Now the issue is what should we do when the hash_map is full. Ideally we'd want to replace the element in the array that is the first probe of the hash value. I don't think there is a way to do this with the standard unordered_map

I'm pretty sure that that's what this commit tries to do: it finds the bucket where the new item belongs and replaces the first item with it; if there's no such item (i.e. that bucket is empty) it finds the first item in the next bucket, wrapping around if necessary. It's not _perfect_, in that it still does a `remove` and an `insert`, but I think that's close enough to the functionality you're suggesting, no?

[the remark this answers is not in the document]
greg7mdp: Thanks Nik!

> I agree that using std::array would be nice to avoid the need for the memory allocation/deallocation but I don't think we can since the size of the buffer is not known at compile time. Unless I am missing something?

Maybe I'm misunderstanding something. I thought that each `char *` inserted in the map points to a constant size buffer of 40 characters (inside the big block allocated at the start).

Since this map entry is dynamically allocated anyways, I was suggesting to replace the `char *` in the `unordered_map` with the 40 character buffer itself. It would not change the number of allocations, but be simpler and allocate memory as needed instead of all at the beginning.

> It's not perfect, in that it still does a remove and an insert, but I think that's close enough to the functionality you're suggesting, no?

Yes, that part is equivalent. The main thing I was suggesting was that, if we really want to reduce memory allocations to the minimum, we could use a hash map that doesn't allocate a bucket per entry, but one that stores all entries in a big array.

src/ripple/protocol/impl/AccountID.cpp
```diff
@@ -25,9 +25,135 @@
...
+    std::size_t requests_ = 0;
+
+    // The number of times were able to satisfy the request from cache
+    std::size_t hits_ = 0;
+
+public:
+    AccountIdCache(std::size_t count)
+        : count_(count)
+        , chars_(std::make_unique<char[]>(count_ * size_))
+        , map_(count_ / 971)
```
greg7mdp, line 66: why `count_ / 971`?

src/ripple/protocol/impl/AccountID.cpp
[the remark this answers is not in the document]
greg7mdp: This is what I was suggesting (untested)

```
class AccountIdCache
{
private:
    // The total number of items the cache buffer can contain
    std::size_t const count_;

    // The maximum size (in bytes) of an item, along with a null terminator
    // and aligned at 8 bytes
    static constexpr std::size_t const size_ = 40;

    // The offset of the next available block of `size_` contiguous bytes
    std::size_t index_ = 0;

    struct  base58_str
    {
        explicit base58_str(std::string const &val) {
            assert(val.size() < size_);
            std::strcpy(str.data(), val.c_str());
        }
        
        std::array<char, size_> str;
    };

    // The map
    hardened_hash_map<AccountID, base58_str> map_;

    std::mutex mutable mutex_;

    // The total number of requests to format an AccountID
    std::size_t requests_ = 0;

    // The number of times were able to satisfy the request from cache
    std::size_t hits_ = 0;

public:
    AccountIdCache(std::size_t count)
        : count_(count)
        , map_(count_ / 971)
    {
    }

    std::string
    toBase58(AccountID const& id)
    {
        std::lock_guard lock(mutex_);

        ++requests_;

        // We use the map's bucket interface:
        auto const b = map_.bucket(id);

        auto iter = std::find_if(map_.begin(b), map_.end(b), [&id](auto& me) {
            return me.first == id;
        });

        if (iter != map_.end(b))
        {
            ++hits_;
            return iter->second.str.data();
        }

        assert (index_ <= count_);

        if (index_ == count_) {
            // Remove the first item from the first non-empty bucket,
            // starting with this one:
            for (std::size_t i = 0; i != map_.bucket_count();  ++i)
            {
                if (map_.bucket_size((b + i) % map_.bucket_count()) != 0)
                {
                    iter = map_.begin((b + i) % map_.bucket_count());

                    assert(iter != map_.end((b + i) % map_.bucket_count()));

                    map_.erase(iter->first);
                }
            }
        }

        // This copy could be optimized away if we had a "dangerous"
        // interface that accepted the output buffer as an argument.
        auto base58_str =  encodeBase58Token(TokenType::AccountID, id.data(), id.size());
        map_.emplace(id, base58_str);

        return std::move(base58_str);
    }
};
```

src/ripple/protocol/impl/AccountID.cpp
nbougalis: I wanted to give the map "reasonable" number of buckets; 971 is a prime number and assuming a reasonable hash function (not an entirely unreasonable assumption, I think), this would mean that, once full, the map would be expected to have _about_ 1,000 items per bucket.

src/ripple/protocol/impl/AccountID.cpp
```diff
@@ -25,9 +25,135 @@
...
+        // This copy could be optimized away if we had a "dangerous"
+        // interface that accepted the output buffer as an argument.
+        std::strcpy(
+            buf,
+            encodeBase58Token(TokenType::AccountID, id.data(), id.size())
+                .c_str());
+        assert(buf[size_ - 2] == 0 && buf[size_ - 1] == 0);
+        map_.emplace(id, buf);
+
+        return buf;
```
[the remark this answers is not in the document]
nbougalis: Good point.

src/ripple/protocol/impl/AccountID.cpp
greg7mdp: My understanding is that the bucket_count provided is just an initial count, and that the bucket_count will increase as the table fills up to prevent the linked list in each bucket to grow too large (lists of 1000 entries / bucket would be very inefficient). I think maybe an initialization with `971` (instead of `count_ / 971`) would be my choice.

src/ripple/protocol/impl/AccountID.cpp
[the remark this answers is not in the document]
greg7mdp: When `map_.emplace(id, base58_str);` is called, it dynamically allocates a new entry containing the map::value_type and the linked list pointers. If that new entry contains the 40 bytes buffer (instead of a char *), we piggyback on the call to `new` that the hash map does internally, we don't allocate a separate buffer.

src/ripple/protocol/impl/AccountID.cpp
```diff
@@ -25,9 +25,135 @@
...
+    {
+        std::memset(chars_.get(), 0, count_ * size_);
+    }
+
+    std::string
+    toBase58(AccountID const& id)
+    {
+        assert(chars_.get() != nullptr);
+
+        std::lock_guard lock(mutex_);
```
seelabs, line 76: My biggest concern with this cache is lock contention. We know that base58 encoding is expensive. On a cache miss this change does that encoding under a lock. 

What do you think about using a boost `UpgradeLock` here? We could check out a read lock, and on a cache miss, do the base58 encoding _before_ we upgrade the lock and write to the cache. That might help with lock contention. It might be a good idea to write some simple benchmarks to experiment with such changes.

seelabs (commented):
This looks good. I left a suggesting that might help with lock contention.

scottschurr:
Having thought about this for a little while, I'm inclined toward adopting @seelabs' change rather than this one.  @seelabs' change will improve both encoding and decoding of all base58 values, not just `AccountID`s.  The change in this pull request only improves `to_string` conversions for `AccountID`s.  So, at least for an initial take, the other change provides more utility.  But, of course, @seelabs' code has not yet passed code review.  We should wait to see how that goes.

nbougalis:
> Having thought about this for a little while, I'm inclined toward adopting @seelabs' change rather than this one. @seelabs' change will improve both encoding and decoding of all base58 values, not just `AccountID`s. The change in this pull request only improves `to_string` conversions for `AccountID`s. So, at least for an initial take, the other change provides more utility. But, of course, @seelabs' code has not yet passed code review. We should wait to see how that goes.

I hear what you're saying about @seelabs's change and agree: it's a better, faster encoder that improves all base58 encodings.

_HOWEVER_, converting `AccountID` to its corresponding `Base58Check` format is expensive and something that servers that service _clients_ do a lot; think how many times it would be invoked in servicing a single `account_lines` for Bitstamp or GateHub, for example. That's the scenario that the cache helps.

I decided to tweak this a bit further to eliminate the need for the `hardened_hash_map` and (imo) reduce complexity but I'm sure it will be controversial (I'm not 100% sold myself).

src/ripple/protocol/impl/AccountID.cpp
nbougalis: The previous cache did the same thing, but I agree that avoiding doing expensive work (like encoding) under the lock makes sense. My latest changes should address this in two ways: (a) not holding a lock while encoding; and (b) carefully allowing concurrent access to the cache.

I'm marking this as resolved, but feel free to undo this if you think it's still an issue.

src/ripple/protocol/impl/AccountID.cpp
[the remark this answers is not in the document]
nbougalis: I learned something new today; thanks! I think your approach would be an improvement over the _existing_ cache and perhaps even the original version of my PR, however, after thinking about your comments more broadly, I decided to adopt a different approach and, effectively, eliminate the `hardened_hash_map` aka `std::unordered_map` altogether. Curious what you think of the changes!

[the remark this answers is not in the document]
greg7mdp: If removing the ordered_map  doesn't affect performance significantly, then I fully agree with removing it. In my opinion any added complexity should have significant benefits, otherwise it is not worth it. After adding seelabs new base64 implementation, we can always revisit the cache idea and see if we can come up with something which is a clear improvement.

src/ripple/protocol/impl/AccountID.cpp
```diff
@@ -17,17 +17,101 @@
...
+        assert(
+            buf != nullptr && buf >= chars_.get() &&
+            buf < (chars_.get() + (count_ * size_)));
+
+        {
+            std::lock_guard lock(mutex_[index % mutex_.size()]);
+
+            if (buf[offset_] != 0 &&
+                std::memcmp(buf, id.data(), id.size()) == 0)
+                return buf;
```
greg7mdp, line 80: I think it should be: `return buf + offset;`

src/ripple/protocol/impl/AccountID.cpp
```diff
@@ -17,17 +17,101 @@
...
+private:
+    // The total number of items the cache buffer can contain
+    std::size_t const count_;
+
+    // The maximum size (in bytes) of an item, along with a null terminator
+    // and aligned at 8 bytes
+    std::size_t const size_ = 64;
+
+    // The offset of the string representation of an AccountID
+    std::size_t const offset_ = 24;
```
greg7mdp, line 45: Shouldn't it be 20 and not 24?

greg7mdp (changes requested):
Love the new design! Indeed it is better than using an unordered_map.

src/ripple/core/impl/Config.cpp
```diff
@@ -108,26 +108,27 @@ namespace ripple {
...
+    {SizedItem::ledgerSize,         {{     32,      32,      64,     256,     384 }}},
+    {SizedItem::ledgerAge,          {{     30,      60,     180,     300,     600 }}},
+    {SizedItem::ledgerFetch,        {{      2,       3,       4,       5,       8 }}},
+    {SizedItem::hashNodeDBCache,    {{      4,      12,      24,      64,     128 }}},
+    {SizedItem::txnDBCache,         {{      4,      12,      24,      64,     128 }}},
+    {SizedItem::lgrDBCache,         {{      4,       8,      16,      32,     128 }}},
+    {SizedItem::openFinalLimit,     {{      8,      16,      32,      64,     128 }}},
+    {SizedItem::burstSize,          {{      4,       8,      16,      32,      48 }}},
+    {SizedItem::ramSizeGB,          {{      8,      12,      16,      24,      32 }}},
+    {SizedItem::accountIdCacheSize, {{  10000,   25000,  100000,  150000,  300000 }}}
```
greg7mdp, line 131: I believe the accountIDCacheSize would benefit from being prime numbers, for better distribution after the modulo operation.

src/ripple/protocol/impl/AccountID.cpp
```diff
@@ -17,17 +17,101 @@
...
+    {
+        std::memset(chars_.get(), 0, count_ * size_);
+    }
+
+    std::string
+    toBase58(AccountID const& id)
+    {
+        assert(chars_.get() != nullptr);
+
+        auto const index = hasher_(id.data()) % count_;
```
greg7mdp, line 67: Do we even need to hash the id? Why not use the low order uint64_t?

nbougalis: We don't _really_ need to hash the whole thing, no. The cost to do it is likely not that big, but there are some things to watch out for: first, `hasher_` is a hardened hash (i.e. it resists algorithmic complexity attacks, by using tagged hashing) and we need that functionality.

Something like this might be sufficient, assuming `cookie_` is the `std::uint64_t` random tag:

    auto const index = (*reinterpret_cast<std::uint64_t const*>(id.data()) ^ cookie_) % count_;

greg7mdp: Yes, I think this would be sufficient for sure. The hardened hash is useful when attackers craft ID to cause collisions in hash maps and slow lookups because of excessive probing. Here we don't really use a hash map with probing, but overwrite the first entry if it doesn't match what we look for, so the worse case scenario would be that the cache is not effective, which I don't think would be crippling.
So I think the xor with a randomly generated value should be plenty safe!

nbougalis:
@greg7mdp, @scottschurr, @seelabs: I took your feedback to hard and updated the code. I think it's dramatically simpler and cleaner now. I would appreciate your feedback. This PR now is on top of #4201, to leverage the new `packed_spinlock` code, which should (in theory) help avoid contention for the cache lock under most scenarios.

src/ripple/core/impl/Config.cpp
nbougalis: Fixed.

src/ripple/protocol/impl/AccountID.cpp
nbougalis: I fixed it by removing it.

src/ripple/protocol/impl/AccountID.cpp
nbougalis: Fixed (code no longer there)

greg7mdp:
Looks great and very simple indeed. I don't know enough about spinlocks to comment on using it instead of std::mutex.

greg7mdp (approved):

greg7mdp:
>  the new `packed_spinlock` code, which should (in theory) help avoid contention for the cache lock under most scenarios.

Just a thought. Maybe we never need to wait? We could just try to acquire the mutex (try_lock), and if it fails just call `encodeBase58Token`. In any case the function is not that long, and the cases where that happens should be quite rare.

scottschurr:
I really like the changes that were made to this pull request.  I especially like replacing the hash table with an array.  Very clever.

Given the changes I decided that it made sense to redo the timings I did previously.  I used the same technique, but I included the fix so actual nanoseconds are reported, not `steady_clock` ticks...  Then I looked at the results and was startled by how similar the results were between the approaches.

So I decided there must be something wrong with my measurement technique.  Ah ha!  The optimizer must be moving where `steady_clock::now()` is being read.  So I added an `atomic` access just after the constructor's read of `now()` and just before the destructor's read of `now()`.  The atomic accesses act as fences to prevent code execution from being moved to either side.  That will make sure the two reads of `now()` bracket the code of interest.  But it also added two atomic accesses that would now be included in the time measurement.

What I learned from this is that accessing `atomic` variables is expensive.  Adding those two `atomic` accesses added about 10 ns to the average result on all three techniques.  But adding the fences did not change the outcome very much beyond adding about 10 ns.  Here are the results I got with real nanoseconds and with the atomic accesses in place acting as fences:

Timing of develop (1.9.1) with the fences in place:
```
AccountID to string -- conversions: 1003520; avg ns: 53; longest ns: 65882
  ranges -- 0-9ns:      0
          10-99ns: 968040
        100-999ns:  35332
      1000-9999ns:     73
    10000-99999ns:     75 
100000ns and over:      0
```
Timing of Nik's latest code with fences in place:
```
AccountID to string -- conversions: 1003520; avg ns: 49; longest ns: 31047
  ranges -- 0-9ns:      0
          10-99ns: 993192
        100-999ns:  10065
      1000-9999ns:    151
    10000-99999ns:    112
100000ns and over:      0
```
Timing of Scott D's base58 work with the fences in place:
```
AccountID to string -- conversions: 1003520; avg ns: 51; longest ns: 47268
  ranges -- 0-9ns:      0
          10-99ns: 974942
        100-999ns:  28359
      1000-9999ns:    125
    10000-99999ns:     94
100000ns and over:      0
```
Meh.  The cache does give the best result.  But it's still only saving 4 ns over the non-cached code in 1.9.1.  If we subtract out the (approximately) 10 ns added by the fences, then we'd be going from an average of 43 ns to an average of 39 ns.  About a 10% improvement.  I'm not convinced the cache is worth it for a reduction that small.

It's certainly possible there's a problem with my measurement technique.  Good measurements are hard.  But I can't figure out what that problem might be.

Thoughts?

greg7mdp:
Thanks @scottschurr, very useful. Some random thoughts I have.
1. strange that "longest ns" is longer for Scott D's branch than default. I would assume these occur when we don't have a cache hit, right?
2. I wonder if the cache is widely too large. How many different AccountIDs does the test use? Are they randomly accessed?  I'd suggest trying with a small cache size (maybe 100 to 500 entries) so the cache can reside in the L3 processor cache. Otherwise the whole benchmark is probably limited by the speed of memory accesses, leading to similar results.

nbougalis:
I appreciate you redoing the timings, @scottschurr. 

There's a couple of things I'd like to point out: first, I suspect that we're already reaping benefits from the existing cache. An interesting approach would be to disable it entirely and run measurements then.

Second, even if this change is, roughly, as fast the existing mechanism, I still think it's a net gain: the code _is_ simpler and avoids doing work at runtime and that can only be a good thing. And I  don't think that this is such a huge hot spot that it's worth squeezing every last drop of perfomance out of it. 

With that said, why is the performance not _significantly_ better? A couple of thoughts there:

First, despite optimizing the cache itself, we can't just hand a raw pointer in the cache without pinning the memory, so we still end up having to allocate a `std::string` to return. Perhaps we could use `boost::static_string<40>`? Also, we could simplify the hashing; the existing code uses the full 160 bits but, as @greg7mdp pointed out, perhaps that's overkill, and using the first 64 bits is sufficient.

Second, I suspect the test scenario you're running is probably not very representative, despite being run against real data. I think that the biggest impact will likely be felt on servers that handle a large volume of RPC requests and have to do these conversions  hundreds of times per one RPC call (imagine, for example, responding to an `account_lines`) times the number of RPC requests handled per second. The same `AccountID` might be needed tens, or maybe even hundreds of times per call, and so sven with a moderate hit ratio the cache may actually add up.

Apropos of this, I think it's worth examining the size of the cache. Greg has suggested that the defined sizes may be too large, and perhaps that's true. If a cache that's half the size is, say, 90% as efficient, why go larger?

So to sum up my thoughts on this: I feel this commit actually cleans up the code a lot and that alone may be sufficient to get it merged. We could make the cache a runtime configurable option so operators can choose whether to use it or not but that may be overkill. Lastly, I'm not opposed to just removing the cache entirely but I think that before we do that we ought to be very sure that performance doesn't suffer on the servers which actually stand to benefit the most from the cache; I think it's worth running this branch on *s1* or *s2* and capturing performance over a large amount of queries and comparing results.

nbougalis:
>That's a worthwhile point. Can we demonstrate the benefit? Could we, for example, install an instrumented 1.9.1 rippled and an instrumented version of the code-with-caching in two of our client handlers?

I think this should be doable. Can you link me to your code please?

scottschurr:
@nbougalis wrote...
> I think this should be doable. Can you link me to your code please?

Sure thing.  Also feel free to adjust the code in case you think there might be problems with the way it is doing the timing.

Here's a branch that does timing for 1.9.1: https://github.com/scottschurr/rippled/tree/to-base-58-timing-1-9-1

Here's a branch that does identical timing for the most recent version of this `aidc` branch: https://github.com/scottschurr/rippled/tree/to-base-58-timing-nik-aidc

Be aware that the timing results are written to standard out, not to the log.  The log is not available at the point we are measuring the timing.  So you may need to redirect the server's output to a file or the like.

Thanks for doing the timing.  I'm curious to see what the results tell us.

nbougalis:
So, I've been running this branch against a server that's been serving public RPC/WebSocket queries for a little over 2 weeks. Over that time, the cache has a hit ratio that's just shy of 89%, which is _significant_. About 25% of the calls were ≤ 50ns; over 96% of the rest were ≤ 100 ns.

scottschurr:
@nbougalis, thanks for making the measurements you did.  A cache hit ratio of 89% is significant.  What I've not yet seen from your timings is evidence of a speed up.  For that evidence we'd need to see timings without the cache on the same machine and a comparable workload.  If the timings with the cache are significantly faster than the timings without the cache then the complexity introduced by the cache is justified.

Make sense?

intelliot:
Opening for review since this may be included in 1.10. https://github.com/XRPLF/rippled/commit/e2eed966b0ecb6445027e6a023b48d702c5f4832 ([The contributing process](https://github.com/XRPLF/rippled/blob/develop/CONTRIBUTING.md#pull-requests) requires approvals from at least two reviewers)

intelliot:
This PR will not be merged, so ignore the merge conflicts mentioned by GitHub below.

scottschurr (approved):
Although I'm not convinced (by measurements taken) that this approach provides significant speedups, I don't think it does any harm.  :+1: from me.

src/ripple/protocol/impl/AccountID.cpp
```diff
@@ -25,9 +25,135 @@
...
+/** Caches the base58 representations of AccountIDs */
+class AccountIdCache
+{
+private:
+    // The total number of items the cache buffer can contain
+    std::size_t const count_;
+
+    // The maximum size (in bytes) of an item, along with a null terminator
+    // and aligned at 8 bytes
+    std::size_t const size_ = 40;
```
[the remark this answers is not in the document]
intelliot: @scottschurr given I'll be closing this PR soon (and it might not get looked at again after that), would it be worth [opening an issue](https://github.com/XRPLF/rippled/issues/new/choose) for this?

If not, we can leave as-is.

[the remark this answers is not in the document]
scottschurr: It's not worth an issue.  We can leave it as is.

src/ripple/app/paths/PathRequest.cpp
```diff
@@ -552,9 +552,16 @@ PathRequest::findPaths(
...
+        auto const& sourceAccount = [&] {
+            if (!isXRP(issue.account))
+                return issue.account;
+
+            if (isXRP(issue.currency))
+                return xrpAccount();
+
+            return *raSrcAccount;
+        }();
+
```
drlongle, line 564: Nitpick. I am not sure what the beauty of a lambda in this case.

drlongle (approved):

intelliot:
Thank you both for the reviews!

PR #4182, opened 2022-05-27
Catch missing node error when rotating database

seelabs:
While there should never be a missing node when copying the SHAMap,
rippled should not terminate when there's an error rotating the
database. This patch aborts the database rotation rather than aborting rippled.

### DIFF
--- a/src/ripple/app/misc/SHAMapStoreImp.cpp
+++ b/src/ripple/app/misc/SHAMapStoreImp.cpp
@@ -17,17 +17,18 @@
 */
 //==============================================================================
 
+#include <ripple/app/misc/SHAMapStoreImp.h>
+
 #include <ripple/app/ledger/TransactionMaster.h>
 #include <ripple/app/misc/NetworkOPs.h>
-#include <ripple/app/misc/SHAMapStoreImp.h>
 #include <ripple/app/rdb/State.h>
 #include <ripple/app/rdb/backend/SQLiteDatabase.h>
 #include <ripple/beast/core/CurrentThreadName.h>
 #include <ripple/core/ConfigSections.h>
 #include <ripple/core/Pg.h>
-#include <ripple/nodestore/impl/DatabaseRotatingImp.h>
-
 #include <ripple/nodestore/Scheduler.h>
+#include <ripple/nodestore/impl/DatabaseRotatingImp.h>
+#include <ripple/shamap/SHAMapMissingNode.h>
 
 #include <boost/algorithm/string/predicate.hpp>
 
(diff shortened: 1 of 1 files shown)


src/ripple/app/misc/SHAMapStoreImp.cpp
```diff
@@ -363,11 +364,24 @@ SHAMapStoreImp::run()
...
+                    std::bind(
+                        &SHAMapStoreImp::copyNode,
+                        this,
+                        std::ref(nodeCount),
+                        std::placeholders::_1));
+            }
+            catch (SHAMapMissingNode const& e)
+            {
+                JLOG(journal_.error())
+                    << "Missing node while copying ledger before roate: "
```
mtrippled, line 380: spell check "rotate"

mtrippled (approved):
LGTM :+1: but fix the spelling of "rotate"

src/ripple/app/misc/SHAMapStoreImp.cpp
seelabs: Fixed.

src/ripple/app/misc/SHAMapStoreImp.cpp
```diff
@@ -30,6 +30,7 @@
 #include <ripple/nodestore/Scheduler.h>
 
 #include <boost/algorithm/string/predicate.hpp>
+#include "ripple/shamap/SHAMapMissingNode.h"
```
scottschurr, line 33: Could you use chevrons to enclose this `#include`?  And move it up with the `ripple` includes please?  Thanks.

scottschurr (commented):
Changes look good to me once the new `#include` is fixed up.  Thanks.

I'm going to let this code burn in for a while on the same machine that saw the crash yesterday.  There's no guarantee that I'll see the same problem, but I may as well give it a try.  If I don't see any crashes after it burns in for a few hours I'll give it the thumbs up.

src/ripple/app/misc/SHAMapStoreImp.cpp
seelabs: @scottschurr Fixed - also an explanation: I changed my editor configuration and it started "helpfully" adding include files for me. I need to figure out how to turn that off. I missed that is added that for me. Thanks for spotting it!

haihai423: 你们和雷达币有关吗

scottschurr (approved):
I did a burn in with this code over the weekend.  Unfortunately, and to no one's surprise, the burning in code neither crashed nor left a message in the log to verify that we really fixed things up with this change.  But I think there's a high likelihood that this is the right change.

Let's do this.  Thanks for the fix.  :+1:

src/ripple/app/misc/SHAMapStoreImp.cpp
nbougalis: @haihai423: Short answer: no. Longer answer: I believe that Radar Coin simply cloned this code (along with several libraries and even entire websites), only changing the name.

seelabs:
Squashed and marked passed

Issue #4007, opened 2021-11-29
Incorrect version in self signed certificate


donovanhide:
https://github.com/ripple/rippled/blob/9d89d4c1885e13e3c0aa7cdb4a389c6fbfd3790a/src/ripple/basics/impl/make_SSLContext.cpp#L159

X.509 certificates can have a version number between 1 and 3 inclusive. The constant `NID_X509` equals 12. The correct constants are one of `X509_VERSION_1`,`X509_VERSION_2` or `X509_VERSION_3`. Maybe this is an ancient copy paste error? Anyway, it stops Go's `crypto\tls` library parsing the certificate, reporting an `invalid version`.

donovanhide:
@nbougalis Another one for you?

donovanhide:
Looks like this might have been the copy paste source :-)

https://stackoverflow.com/questions/2883164/openssl-certificate-lacks-key-identifiers

donovanhide:
Simplest fix for this is to simply delete the erroneous line above and accept the default version (`X509_VERSION_1`):

https://www.openssl.org/docs/man3.0/man3/X509_set_version.html

Perhaps this is a good lesson in checking return codes from C functions :-) Want a PR?

MarkusTeufelberger:
Practically every x509 certificate out there has version 3, please use that if you fix this.

donovanhide:
https://datatracker.ietf.org/doc/html/rfc5280#section-4.1.2.1
 

>    This field describes the version of the encoded certificate.  When
>    extensions are used, as expected in this profile, version MUST be 3
>    (value is 2).  If no extensions are present, but a UniqueIdentifier
>    is present, the version SHOULD be 2 (value is 1); however, the
>    version MAY be 3.  If only basic fields are present, the version
>    SHOULD be 1 (the value is omitted from the certificate as the default
>    value); however, the version MAY be 2 or 3.

To me, that reads like it should be `X509_VERSION_1`.

donovanhide:
Bear in mind that this is only for the self-signed certificate. not the one that can be specified in the configuration file. 

https://github.com/ripple/rippled/blob/9d89d4c1885e13e3c0aa7cdb4a389c6fbfd3790a/src/ripple/basics/impl/make_SSLContext.cpp#L268-L283

It is notable that a lot of return codes are ignored in this code section as well.

PR #4185, opened 2022-05-31
Improve self-signed certificate generation:

nbougalis:
When starting, the code generates a new ephemeral private key and a corresponding certificate to go along with it. This process can take time and, while this is unlikely to matter for normal server operations, it can have a significant impact for unit testing and development. Profiling data suggests that ~20% of the time needed for a unit test run can be attributed to this.

This commit does several things:

1. It restructures the code so that a new self-signed certificate and its corresponding private key are only initialized once at startup; this has minimal impact on the operation of a regular server.
2. It provides new default DH parameters. This doesn't impact the security of the connection, but those who compile from scratch can generate new parameters if they so choose.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### DIFF
--- a/src/ripple/basics/impl/make_SSLContext.cpp
+++ b/src/ripple/basics/impl/make_SSLContext.cpp
@@ -17,17 +17,54 @@
 */
 //==============================================================================
 
-#include <ripple/basics/chrono.h>
 #include <ripple/basics/contract.h>
 #include <ripple/basics/make_SSLContext.h>
-#include <ripple/beast/container/aged_unordered_set.h>
-#include <cstdint>
-#include <sstream>
+#include <ctime>
 #include <stdexcept>
 
 namespace ripple {
 namespace openssl {
 namespace detail {
+
+/** The default strength of self-signed RSA certifices.
+
+    Per NIST Special Publication 800-57 Part 3, 2048-bit RSA is still
+    considered acceptably secure. Generally, we would want to go above
+    and beyond such recommendations (e.g. by using 3072 or 4096 bits)
+    but there is a computational cost associated with that may not
+    be worth paying, considering that:
+
+    - We regenerate a new ephemeral certificate and a securely generated
+      random private key every time the server is started; and
+    - There should not be any truly secure information (e.g. seeds or private
+      keys) that gets relayed to the server anyways over these RPCs.
+
+      @note If you increase the number of bits you need to generate new
+            default DH parameters and update defaultDH  accordingly.
+ * */
+int defaultRSAKeyBits = 2048;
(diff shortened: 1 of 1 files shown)


nbougalis:
Generally, I think it might be a good idea to remove this automatic certificate generation from within the code. It's a convenient option to be sure, but it's also limited; for example, it will only generate RSA certificates. 

with options like LetsEncrypt for those whose servers have an associated domain, it is now trivially easy to get a "real" certificate from a well-known registrar, and it would be better if server operators did that instead.

Those who can't get a certificate from a CA could still generate a self-signed one. Information on how to do it on the web is plentiful but if necessary, we could include instructions on how to do it and it might even be something that could happen during initial package installation by invoking the right incantation.

src/ripple/basics/impl/make_SSLContext.cpp
```diff
@@ -45,195 +82,88 @@ namespace detail {
...
-static x509_ptr
-x509_new()
-{
-    X509* x509 = X509_new();
+        if (x509 == nullptr)
+            LogicError("X509_new failed");
 
-    if (x509 == nullptr)
-        LogicError("X509_new failed");
+        X509_set_version(x509, NID_X509);
```
donovanhide, line 131: Still the wrong version:

https://github.com/ripple/rippled/issues/4007

nbougalis: Thanks for reminding me of this. I'll fix this.

greg7mdp:
Maybe it is a naive question, because I didn't look at the code, but assuming most unit tests don't need the certificate, would it be possible to generate it the first time it is needed?

ximinez (approved):
I don't have a lot of SSL expertise, but these changes don't appear to be changing anything in that realm. The code is doing the same operations in the same order, only saving the self-signed certificate in a `static` context for easy reuse. 👍

nbougalis:
> Maybe it is a naive question, because I didn't look at the code, but assuming most unit tests don't need the certificate, would it be possible to generate it the first time it is needed?

We would need to (a) restructure the startup sequence; and (b) make it possible to avoid instantiating the `Overlay` inside `ApplicationImp`. Both of these are good ideas and worth doing even if this is accepted & merged, but they are both much more complicated than this simple fix.

I folded a commit into this that addresses the concern by @donovanhide and fixes #4007. Apologies, I hadn't realized you had already reviewed, @ximinez, @greg7mdp.

nbougalis:
Merged as 0ecfc7cb1a958b731e5f184876ea89ae2d4214ee

PR #4188, opened 2022-06-01
Fix race conditions in shard

seelabs:

This PR introduces two changes:
1) It fixes race conditions in Shard.{h,cpp}
2) It introduces clang `ThreadSafetyAnalysis`

To compile with `ThreadSafetyAnalysis`, compile with clang and use `-Dthread_safety_analysis=On` with cmake. Note: this will compile with clang's standard library and may not link. This is OK, we are running this for the compile warnings, not for the executable.

This would be a good PR to discuss objections to the annotations required by `ThreadSafetyAnalysis`

### Type of Change

### DIFF
--- a/src/ripple/nodestore/impl/Shard.h
+++ b/src/ripple/nodestore/impl/Shard.h
@@ -324,7 +324,7 @@ private:
     GUARDED_BY(mutex_) std::uint32_t fdRequired_{0};
 
     // NuDB key/value store for node objects
-    GUARDED_BY(mutex_) std::unique_ptr<Backend> backend_;
+    std::unique_ptr<Backend> backend_ GUARDED_BY(mutex_);
 
     std::atomic<std::uint32_t> backendCount_{0};
 
--- a/src/ripple/nodestore/impl/Shard.h
+++ b/src/ripple/nodestore/impl/Shard.h
@@ -26,6 +26,7 @@
 #include <ripple/basics/KeyCache.h>
 #include <ripple/basics/MathUtilities.h>
 #include <ripple/basics/RangeSet.h>
+#include <ripple/basics/ThreadSafetyAnalysis.h>
 #include <ripple/core/DatabaseCon.h>
 #include <ripple/nodestore/NodeObject.h>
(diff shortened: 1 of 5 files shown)


scottschurr (approved):
:+1: I ran the failing unit test in a loop 10525 times with no seg faults.  I think this particular race condition is fixed.  Thanks for taking care of this!

src/ripple/nodestore/impl/Shard.h
```diff
@@ -316,29 +318,30 @@ class Shard final
     boost::filesystem::path const dir_;
 
     // Storage space utilized by the shard
-    std::uint64_t fileSz_{0};
+    GUARDED_BY(mutex_) std::uint64_t fileSz_{0};
```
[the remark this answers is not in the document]
seelabs: As a side note: one down side to these annotations being bolted-on and not part of the language is we usually put the annotations at the end of a declaration, but when the variable is initialized, we need to put it at the beginning. That's unfortunate, but not a killer.

src/ripple/nodestore/impl/Shard.h
```diff
@@ -316,29 +318,30 @@ class Shard final
...
 
     // Number of file descriptors required by the shard
-    std::uint32_t fdRequired_{0};
+    GUARDED_BY(mutex_) std::uint32_t fdRequired_{0};
```
undertome, line 324: If you move the initialization to the constructor, can you place the annotation at the end of this declaration?

undertome (changes requested):

src/ripple/nodestore/impl/Shard.h
seelabs: Yes - but while I'm not happy about putting the annotation at the beginning, I'd rather do that than disallowing initializing in the declaration

scottschurr:
@undertome, has @seelabs addressed your concerns?

undertome:
> @undertome, has @seelabs addressed your concerns?

yes

undertome (approved):

PR #4199, opened 2022-06-07
Add PeerScheduler

thejohnfreeman:

This is a first version of a peer scheduler. There are some comments in the code explaining parts of my thinking, but I'm working on a longer Markdown document to go with it. Please take a look, ask questions (it will help me know what to include in the document and how to explain it), and poke holes (it will improve the design and implementation).

### Context of Change

We need to treat connections to peers and their capacity for messages as a limited resource. Each message to a peer essentially represents a remote job executed on that peer. We already recognize local threads are a limited resource. We throttle the number of simultaneously executing jobs to avoid overloading the underlying machine's capacity for simultaneously executing threads, and we schedule jobs according to priority during periods of high contention. We should do the same for remote jobs.

### Type of Change


thejohnfreeman:
Closed in favor of #4217.

PR #4202, opened 2022-06-09
Limit how often endpoint messages can be processed:

nbougalis:
The peer discovery protocol depends on peers exchanging messages listing IP addresses for other peers.

Under normal circumstances, these messages should not be sent frequently; the existing code would track the earliest time a new message should be processed, but did not actually enforce that limit.

### Type of Change

- [ ] Documentation Updates
- [ ] Release


seelabs (approved):
:+1: LGTM

thejohnfreeman (approved):

PR #4204, opened 2022-06-12
Improve STVector256 deserialization

nbougalis:
The existing `STVector256` is not very optimal: it requires a lot of unnecessary copying. This commit fixes that _and_ improves detection improperly serialized values: now such values will result in an exception being raised (exceptions from that codepath are expected _and_ properly handled).

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### DIFF
--- a/src/ripple/basics/Slice.h
+++ b/src/ripple/basics/Slice.h
@@ -48,6 +48,7 @@ private:
     std::size_t size_ = 0;
 
 public:
+    using value_type = std::uint8_t;
     using const_iterator = std::uint8_t const*;
 
     /** Default constructed Slice has length 0. */
--- a/src/ripple/basics/Slice.h
+++ b/src/ripple/basics/Slice.h
@@ -49,7 +49,7 @@ private:
 
 public:
     using value_type = std::uint8_t;
-    using const_iterator = std::uint8_t const*;
+    using const_iterator = value_type const*;
 
     /** Default constructed Slice has length 0. */
     Slice() noexcept = default;
@@ -76,13 +76,13 @@ public:
         This may be zero for an empty range.
     */
     /** @{ */
-    std::size_t
+    [[nodiscard]] std::size_t
     size() const noexcept
     {
         return size_;
     }
 
-    std::size_t
+    [[nodiscard]] std::size_t
     length() const noexcept
     {
         return size_;
--- a/src/ripple/protocol/impl/STVector256.cpp
+++ b/src/ripple/protocol/impl/STVector256.cpp
@@ -26,19 +26,19 @@ namespace ripple {
 
 STVector256::STVector256(SerialIter& sit, SField const& name) : STBase(name)
 {
-    Blob data = sit.getVL();
-    auto const count = data.size() / (256 / 8);
-    mValue.reserve(count);
-    Blob::iterator begin = data.begin();
-    unsigned int uStart = 0;
-    for (unsigned int i = 0; i != count; i++)
-    {
-        unsigned int uEnd = uStart + (256 / 8);
-        // This next line could be optimized to construct a default
-        // uint256 in the vector and then copy into it
-        mValue.push_back(uint256(Blob(begin + uStart, begin + uEnd)));
-        uStart = uEnd;
-    }
+    auto const slice = sit.getSlice(sit.getVLDataLength());
+
+    if (slice.size() % uint256::size() != 0)
+        Throw<std::runtime_error>(
+            "Bad serialization for STVector256: " +
+            std::to_string(slice.size()));
+
+    auto const cnt = slice.size() / uint256::size();
+
+    mValue.reserve(cnt);
+
+    for (std::size_t i = 0; i != cnt; ++i)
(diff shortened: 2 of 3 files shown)


src/ripple/basics/Slice.h
```diff
@@ -48,6 +48,7 @@ class Slice
     std::size_t size_ = 0;
 
 public:
+    using value_type = std::uint8_t;
     using const_iterator = std::uint8_t const*;
```
greg7mdp, line 52: maybe `using const_iterator = value_type const*;

src/ripple/protocol/impl/STVector256.cpp
```diff
@@ -26,19 +26,19 @@ namespace ripple {
...
-        // uint256 in the vector and then copy into it
-        mValue.push_back(uint256(Blob(begin + uStart, begin + uEnd)));
-        uStart = uEnd;
-    }
+    auto const slice = sit.getSlice(sit.getVLDataLength());
+
+    if (slice.size() % uint256::size() != 0)
+        Throw<std::runtime_error>(
+            "Bad serialization for STVector256: " +
+            std::to_string(slice.size()));
```
seelabs, line 34: The `NFTokenCancelOffer` uses a `STVector256` field. The old code may have accepted an incorrectly serialized `sfNFTTokenOffers` field where as the new code throws. I'd be surprised if an incorrectly serialized transaction could ever make into the ledger, but I'm not sure I can prove it (code like this would assert in debug mode: https://github.com/nbougalis/rippled/blob/c497a62fb95c98a1e29f0eb0557ca36adcb89c7d/src/ripple/app/tx/impl/Transactor.cpp#L785-L802)

I think it's fine as-is, but I did want to bring this up.

nbougalis: Thanks for bringing this up. So two comments:

1. This cannot occur on the main network since the `featureNonFungibleTokensV1` amendment isn't active, and so the `NFTokenCancelOffer` transaction cannot be executed.
2. In addition to the code you showed, the code will call [`sterilize`](https://github.com/ripple/rippled/blob/25474343a98922ee5b44019f45657e381080954d/src/ripple/protocol/impl/STTx.cpp#L550-L557) on submitted transactions, which _should_ detect this in _some_ circumstances. 

Given than we're planning to introduce another fix amendment to correct a minor issue identified with the XLS-20 implementation, which (for obvious reasons!) should be activated _before_ XLS-20 implementation, I think my suggestion would be to include this code with this release. It doesn't need to be directly gated by the amendment, but if the fix amendment activates before `featureNonFungibleTokensV1` (and again, it should), then this issue will never be able to trigger.

Apropos of this, I wonder what the performance implications of _always_ executing the serialization/deserialization check that's currently only done if compiled with `-DDEBUG` would be. Also, I wonder whether this check more properly belongs in either `preflight` or `preclaim` (we ought to claim a fee if someone submits a broken transaction and entomb the transation in the ledger!).

There are good reasons to have it in `operator()` (i.e. during "application") too, but, as I said above, if this check fails, it would be better to `tec` out of there, instead of simply calling `assert`.

HowardHinnant (approved):

greg7mdp:
The STVector256 class duplicates a lot of the std::vector interface. I think it would be simpler to declare it as:

```
class STVector256 : public STBase, public std::vector<uint256> { ...};
```

This would allow to remove all the duplicated vector APIs (and also get the newer ones like emplace :-))

HowardHinnant:
I disagree with publicly inheriting from `std::vector<uint256>`.  This would introduce an unwanted _implicit_ conversion to `std::vector<uint256>`.  The current API has an _explicit_ conversion to `std::vector<uint256>`.  It would also introduce implicit conversions between `STVector256*` and `std::vector<uint256>*`

greg7mdp:
@HowardHinnant , why do you find the implicit conversion objectionable? IMHO, the (incomplete) duplication of the vector interface is worse.

HowardHinnant:
If we want an implicit conversion to vector, that is one thing.  But the existing API has an explicit conversion.  The distinction between implicit and explicit conversions is not small.  It can have a very large impact on client code that goes well beyond just syntax.

We should not change the implementation to something more convenient without a full accounting of the desired API of the class.  The priority of correctness is more important than the priority of ease of writing.

greg7mdp:
I'm not quite convinced that doing what I suggest is as dangerous as you make it to be, and I think there is also potential danger having extra code duplicating an API unnecessarily. But you seem to feel strongly about it, so let's leave it as is.

nbougalis:
I'll be honest, I've thought about this before, @greg7mdp; eliminating some of the unnecessary wrapper functions appeals to me. _However_ there are issues: _not_ all `std::vector` operations are appropriate in an `STVector256`, and I do share some of @HowardHinnant's concerns.

I think it's worth thinking about this in the broader context of classes like `STVector` and `STBlob` but it's not something that should be done without careful consideration.

nbougalis:
@greg7mdp, I addressed the comment you left. If you wouldn't mind approving this PR, thanks.

greg7mdp (approved):
Thanks for addressing my comment Nik!

PR #4206, opened 2022-06-13
Enable c++-20

seelabs:

This PR enables C++-20 support in rippled. It also ups the minimum supported compilers to gcc 11. We need to determine what the minimum supported clang and visual studio should be as well and update the docs and build scripts.

@legleux Can I ask you to patch the docs that build scripts that need to be updated so we can add them to this PR? Thanks!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,7 +6,7 @@ endif ()
 
 project (rippled)
 set(CMAKE_CXX_EXTENSIONS OFF)
-set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD 20)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 
 # make GIT_COMMIT_HASH define available to all sources
--- a/src/ripple/beast/utility/PropertyStream.h
+++ b/src/ripple/beast/utility/PropertyStream.h
@@ -78,12 +78,6 @@ protected:
     virtual void
     add(std::string const& key, unsigned char value);
     virtual void
-    add(std::string const& key, wchar_t value);
-#if 0
-    virtual void add (std::string const& key, char16_t value);
-    virtual void add (std::string const& key, char32_t value);
-#endif
-    virtual void
     add(std::string const& key, short value);
     virtual void
     add(std::string const& key, unsigned short value);
@@ -139,12 +133,6 @@ protected:
     add(signed char value);
     virtual void
     add(unsigned char value);
-    virtual void
-    add(wchar_t value);
-#if 0
-    virtual void add (char16_t value);
-    virtual void add (char32_t value);
-#endif
     virtual void
     add(short value);
     virtual void
--- a/src/ripple/app/consensus/RCLConsensus.cpp
+++ b/src/ripple/app/consensus/RCLConsensus.cpp
@@ -421,7 +421,9 @@ RCLConsensus::Adaptor::onAccept(
     Json::Value&& consensusJson)
 {
     app_.getJobQueue().addJob(
-        jtACCEPT, "acceptLedger", [=, cj = std::move(consensusJson)]() mutable {
+        jtACCEPT,
+        "acceptLedger",
+        [=, this, cj = std::move(consensusJson)]() mutable {
             // Note that no lock is held or acquired during this job.
             // This is because generic Consensus guarantees that once a ledger
             // is accepted, the consensus results and capture by reference state
--- a/src/ripple/app/paths/impl/BookStep.cpp
+++ b/src/ripple/app/paths/impl/BookStep.cpp
@@ -1125,10 +1125,10 @@ bookStepEqual(Step const& step, ripple::Book const& book)
     bool const inXRP = isXRP(book.in.currency);
     bool const outXRP = isXRP(book.out.currency);
     if (inXRP && outXRP)
-        return equalHelper<
-            XRPAmount,
-            XRPAmount,
-            BookPaymentStep<XRPAmount, XRPAmount>>(step, book);
+    {
+        assert(0);
+        return false;  // no such thing as xrp/xrp book step
+    }
     if (inXRP && !outXRP)
         return equalHelper<
             XRPAmount,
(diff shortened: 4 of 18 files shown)


nbougalis (approved):
Looks reasonable. Please reword commit: "Switch from C++17 to C++20"

CMakeLists.txt
```diff
@@ -21,6 +21,8 @@ if(Git_FOUND)
     endif()
 endif() #git
 
+set(CMAKE_CXX_STANDARD 20)
```
greg7mdp, line 24: should remove `set(CMAKE_CXX_STANDARD 17)` at line 9

scottschurr (approved):
I'm good with this once you make the change @greg7mdp asked for in CMakeLists.txt (good eyes @greg7mdp!)

Here are a couple of additional changes you could make to comments, but they are not required:

```
--- a/src/ripple/ledger/OpenView.h
+++ b/src/ripple/ledger/OpenView.h
@@ -77,7 +77,7 @@ private:
     };
 
     // List of tx, key order
-    // Use the boost pmr functionality instead of the c++-17 standard pmr
+    // Use boost::pmr functionality instead of std::pmr
     // functions b/c clang does not support pmr yet (as-of 9/2020)
     using txs_map = std::map<
         key_type,
```
and
```
--- a/src/ripple/ledger/detail/RawStateTable.h
+++ b/src/ripple/ledger/detail/RawStateTable.h
@@ -119,7 +119,7 @@ private:
         }
     };
 
-    // Use the boost pmr functionality instead of the c++-17 standard pmr
+    // Use boost::pmr functionality instead of std::pmr
     // functions b/c clang does not support pmr yet (as-of 9/2020)
     using items_t = std::map<
         key_type,
```

src/ripple/beast/utility/PropertyStream.h
```diff
@@ -77,9 +77,9 @@ class PropertyStream
     add(std::string const& key, signed char value);
     virtual void
     add(std::string const& key, unsigned char value);
+#if 0
     virtual void
     add(std::string const& key, wchar_t value);
-#if 0
     virtual void add (std::string const& key, char16_t value);
     virtual void add (std::string const& key, char32_t value);
 #endif
```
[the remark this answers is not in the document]
seelabs: Fixed in: 598dd3412e Remove unused code

seelabs:
@scottschurr Fixed comments in: 06ea306092 [fold] Reword some comments

src/ripple/beast/utility/PropertyStream.h
```diff
@@ -139,9 +133,9 @@ class PropertyStream
     add(signed char value);
     virtual void
     add(unsigned char value);
+#if 0
```
scottschurr, line 136: Hmmm.  I'd read @nbougalis's suggesting as suggesting the `#if`d out code be removed as well.  But maybe I misunderstood.

scottschurr (approved):
:+1: TGTM, although I did have one question about some compiled-out code.

src/ripple/beast/utility/PropertyStream.h
seelabs: I thought I removed them, but I guess I missed this. Will fix.

seelabs: Fixed in fcf5f2cf6e [fold] Rm more dead code

ximinez (commented):
Bad news. Visual Studio 19 and 21 both choke on building this. https://gist.github.com/ximinez/df7f9a43d1572f43d04ad0026c4b2fa9

I'm building with Boost 1.75. Since it's involved, I'll try a more updated version next.

ximinez:
With Boost 1.77, it's a different set of errors. https://gist.github.com/ximinez/ab7783cee57da1b9aa2b2507adcc3925
It looks like `unique()` just needs to be replaced with `use_count() == 1`, and I think `result_of_t` can be replaced with `invoke_result_t`.

seelabs:
@ximinez I just pushed a patch that might fix the boost 1.77 errors, can you try this ( [[fold] Fixes for visual studio: shared_ptr unique and type_traits](https://github.com/ripple/rippled/pull/4206/commits/a4db876306105a3a7599276d6673f7cd2288808c)

ximinez:
The `unique()` error is fixed, but `result_of_t` is still broken. It looks like that was removed entirely from the C++-20 spec. https://en.cppreference.com/w/cpp/types/result_of

ximinez:
Quick note that it builds now. Tests are running.

scottschurr:
@ximinez, if I'm recalling correctly , the explicit `this` captures that you removed to make VS17 happy were added to prevent warnings from clang in C++20 mode.  Should we consider abandoning VS17?  Or do we need a more complicated solution?

ximinez:
> if I'm recalling correctly , the explicit `this` captures that you removed to make VS17 happy were added to prevent warnings from clang in C++20 mode. Should we consider abandoning VS17? Or do we need a more complicated solution?

I have no objection to abandoning VS17. We've fixed the issues with VS19 (and thus 21), and I've completely migrated my workflow. I don't know of any other devs that are using 17 either, but I could be wrong. The only significant downside is that our only [VS documentation is for 17](https://github.com/ripple/rippled/blob/develop/Builds/VisualStudio2017/README.md) (AFAIK), so that'll have to be updated.

ximinez:
To clarify, the clang failures I'm seeing are on your unmodified branch. I've tried both boost 1.75 and 1.77.

scottschurr:
Unfortunately it will be hard to compare versions of clang.  Clang on Mac uses its own versioning scheme.  Here's what I get:
```
Apple clang version 11.0.0 (clang-1100.0.33.17)
```

greg7mdp (approved):

ximinez (changes requested):
See comment https://github.com/XRPLF/rippled/pull/4206#issuecomment-1168748963

Adding enums together is still a problem when building with clang 14.

seelabs:
@ximinez Added a patch that should fix the enum issue. Also rebased onto the latest develop. Can you retest when you get a chance?

ximinez (approved):
Looks good to go so far.

seelabs:
Squashing and marking this as passed

nbougalis:
This was merged as 92d35e54c7de6bbe44ff6c7c52cc0765b3f78258

PR #4213, opened 2022-06-17
Properly handle incorrect port numbers in parseURL (fixes #4200)

### LINKED ISSUE #4200: Potential Incorrect Acceptance Handling in parseUrl() for Port Parameter
The custom `parseUrl()` implementation contains a potential edge case issue regarding the parsing of ports.

Consider the following group in the regular expression that determines basic acceptance of the URL: [StringUtilities.cpp#L61-L62](https://github.com/ripple/rippled/blob/9d89d4c1885e13e3c0aa7cdb4a389c6fbfd3790a/src/ripple/basics/impl/StringUtilities.cpp#L61-L62)

```
 // optional port
"(?::([[:digit:]]+))?"
```

This expression will accept an arbitrary digit sequence with length >=1, including numbers which are larger than the `uint16_t` that are used for port numbers.

On a match with port number content, [StringUtilities.cpp#L90-L93](https://github.com/ripple/rippled/blob/9d89d4c1885e13e3c0aa7cdb4a389c6fbfd3790a/src/ripple/basics/impl/StringUtilities.cpp#L90-L93) will use the custom `LexicalCast.h` based implementation to unconditionally convert the port:

```
 if (!port.empty())
{
    pUrl.port = beast::lexicalCast<std::uint16_t>(port);
}
````

Although benign, the code appears to incorrectly accept ports that are out of range and should have been rejected as an invalid URL. Instead it handles them as port 0 by falling back to the default `uint16_t` value. Additionally, port 0 is also accepted when used in the URL directly.

chennakeshava1998:
This Pull Request fixes issue 4200 (https://github.com/ripple/rippled/issues/4200). 

Previously, inputs for very large port numbers (more than the 16 bit capacity of the field) was handled incorrectly i.e. lexicalCast returns a zero for such inputs.

The code in this PR returns a boolean value of false if the port number is zero.
This code also handles the case of an input of zero (0) to the port number field.

### DIFF
--- a/src/ripple/basics/impl/StringUtilities.cpp
+++ b/src/ripple/basics/impl/StringUtilities.cpp
@@ -90,6 +90,12 @@ parseUrl(parsedURL& pUrl, std::string const& strUrl)
     if (!port.empty())
     {
         pUrl.port = beast::lexicalCast<std::uint16_t>(port);
+
+        // For inputs larger than 2^32-1 (65535), lexicalCast returns 0. 
+        // parseUrl returns false for such inputs.
+        if (pUrl.port == 0) {
+            return false;
+        }
     }
     pUrl.path = smMatch[6];
 
--- a/src/test/basics/StringUtilities_test.cpp
+++ b/src/test/basics/StringUtilities_test.cpp
@@ -289,6 +289,12 @@ public:
             BEAST_EXPECT(!parseUrl(pUrl, "nonsense"));
             BEAST_EXPECT(!parseUrl(pUrl, "://"));
             BEAST_EXPECT(!parseUrl(pUrl, ":///"));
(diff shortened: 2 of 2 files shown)


chennakeshava1998:
With regards to the failed Github Actions task, clang-format is not a recognised git command. Do I need to use an extension/tool (something like this? https://github.com/llvm-mirror/clang/blob/master/tools/clang-format/git-clang-format) to format my code?

scottschurr:
I think you can use homebrew to install `$ brew install clang-format`.  I always just try to remember to invoke `clang-format` before I push to my GitHub repo.  There are other folks who have done something or other that is more automated.

chennakeshava1998:
Okay, thanks for the suggestion Scott! I'll look into that.

thejohnfreeman (approved):

PR #4215, opened 2022-06-23
Introduce fixNFTokenNegOffer amendment

scottschurr:

- Fix a bug that allowed an offer for an NFToken to hold a negative amount.
- Allow a `Destination` field on an NFToken buy offer.

### Context of Change

Testing revealed a bug in the NFTokenCreateOffer implementation.  The implementation allows a negative `Amount` to be specified in an offer for an NFToken.  Since the implementation already explicitly disallows offers with an amount of 0 for IOUs, allowing negative amounts makes no sense.  This amendment fixes the bug.

Additionally, there was a request by folks who want to use the broker feature.  Currently the implementation does not allow a buy offer to specify a `Destination`.  This prevents a buy offer from specifying a desired broker for that offer.  So this amendment adds the capability for a buy offer to specify a `Destination`.

If an NFToken buy offer includes a `Destination` field, the account ID in that field may only be one of two accounts:

1. The account that owns the NFToken.  The consequence is that the offer may only be accepted directly by the NFToken owner.  This will prevent the offer from ever being completed by a broker.
2. An account that will act as the broker for that offer.  In this case the owner of the NFToken cannot directly accept the buy offer; the offer can only be completed by the specified broker.

This second capability was not considered important enough to justify its own amendment.  So it is piggybacking on the amendment that fixes the negative amount problem.

### Type of Change

- [x] Tests (unit tests were added to verify the new code)
- [x] Documentation (docs will need to change to describe the new behaviors) 

Preventing an NFToken offer from containing a negative amount is a bug fix.

Allowing an NFToken buy offer to provide a `Destination` is a feature add.

The champion for `Destination` for a buy offer has been @ledhed2222.  I'm hoping to get a review from them as well.  Also heads up to @DennisDawson that this change is in the winds.

### DIFF
--- a/src/ripple/protocol/impl/Feature.cpp
+++ b/src/ripple/protocol/impl/Feature.cpp
@@ -431,7 +431,7 @@ REGISTER_FEATURE(RequireFullyCanonicalSig,      Supported::yes, DefaultVote::yes
 REGISTER_FIX    (fix1781,                       Supported::yes, DefaultVote::yes);
 REGISTER_FEATURE(HardenedValidations,           Supported::yes, DefaultVote::yes);
 REGISTER_FIX    (fixAmendmentMajorityCalc,      Supported::yes, DefaultVote::yes);
-REGISTER_FEATURE(NegativeUNL,                   Supported::yes, DefaultVote::no);
+REGISTER_FEATURE(NegativeUNL,                   Supported::yes, DefaultVote::yes);
 REGISTER_FEATURE(TicketBatch,                   Supported::yes, DefaultVote::yes);
 REGISTER_FEATURE(FlowSortStrands,               Supported::yes, DefaultVote::yes);
 REGISTER_FIX    (fixSTAmountCanonicalize,       Supported::yes, DefaultVote::yes);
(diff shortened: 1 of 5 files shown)


src/ripple/protocol/impl/Feature.cpp
```diff
@@ -431,7 +431,7 @@ REGISTER_FEATURE(RequireFullyCanonicalSig,      Supported::yes, DefaultVote::yes
 REGISTER_FIX    (fix1781,                       Supported::yes, DefaultVote::yes);
 REGISTER_FEATURE(HardenedValidations,           Supported::yes, DefaultVote::yes);
 REGISTER_FIX    (fixAmendmentMajorityCalc,      Supported::yes, DefaultVote::yes);
-REGISTER_FEATURE(NegativeUNL,                   Supported::yes, DefaultVote::no);
+REGISTER_FEATURE(NegativeUNL,                   Supported::yes, DefaultVote::yes);
```
ledhed2222, line 434: when is it safe to change the default vote like this?

PR #4218, opened 2022-06-29
Introduce a slab-based memory allocator and optimize SHAMapItem

nbougalis:
#### Introduce support for a slabbed allocator:

When instantiating a large amount of fixed-sized objects on the heap the overhead that dynamic memory allocation APIs impose will quickly become significant.

In some cases, allocating a large amount of memory at once and using a slabbing allocator to carve the large block into fixed-sized units that are used to service requests for memory out will help to reduce memory fragmentation significantly and, potentially, improve overall performance.

This commit introduces a new SlabAllocator<> class that wraps around the complexity and exposes an API that is similar to the C++ concept of an Allocator.

This class is not a general-purpose allocator and should not be used unless profiling and analysis of specific memory allocation patterns indicates that the additional complexity introduced will improve the performance of the system overall and subsequenting profiling proves that.

#### Optimize SHAMapItem and leverage new slab allocator:

The `SHAMapItem` class contains a variable-sized buffer that holds the serialized data associated with a particular item inside a `SHAMap`.

Prior to this commit, the buffer for the serialized data was allocated separately. Coupled with the fact that most instances of `SHAMapItem` were wrapped around a `std::shared_ptr` meant that an instantiation might result in up to three separate memory allocations.

This commit switches away from `std::shared_ptr` for `SHAMapItem` and uses `boost::intrusive_ptr` instead, allowing the reference count for an instance to live inside the instance itself. Coupled
with using a slab-based allocator to optimize memory allocation for the most commonly sized buffers, the net result is significant memory savings. In testing, the reduction in memory usage hovers between 400MB and 650MB. Other scenarios might result in larger savings.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### DIFF
--- a/src/ripple/basics/SlabAllocator.h
+++ b/src/ripple/basics/SlabAllocator.h
@@ -1,0 +1,421 @@
+//------------------------------------------------------------------------------
+/*
+    This file is part of rippled: https://github.com/ripple/rippled
+    Copyright 2022, Nikolaos D. Bougalis <nikb@bougalis.net>
+
+    Permission to use, copy, modify, and/or distribute this software for any
+    purpose  with  or without fee is hereby granted, provided that the above
+    copyright notice and this permission notice appear in all copies.
+
+    THE  SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+    WITH  REGARD  TO  THIS  SOFTWARE  INCLUDING  ALL  IMPLIED  WARRANTIES  OF
+    MERCHANTABILITY  AND  FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+    ANY  SPECIAL ,  DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+    WHATSOEVER  RESULTING  FROM  LOSS  OF USE, DATA OR PROFITS, WHETHER IN AN
+    ACTION  OF  CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+    OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+//==============================================================================
+
+#ifndef RIPPLE_BASICS_SLABALLOCATOR_H_INCLUDED
+#define RIPPLE_BASICS_SLABALLOCATOR_H_INCLUDED
+
+#include <ripple/beast/type_name.h>
+#include <algorithm>
+#include <atomic>
+#include <cassert>
+#include <cstdint>
+#include <mutex>
+
+#include <boost/align.hpp>
+#include <boost/container/static_vector.hpp>
+#include <boost/predef.h>
+
+#if BOOST_OS_LINUX
+#include <sys/mman.h>
+#endif
+
+namespace ripple {
+
+template <typename Type>
+class SlabAllocator
+{
+    static_assert(
+        sizeof(Type) >= sizeof(std::uint8_t*),
+        "SlabAllocator: the requested object must be larger than a pointer.");
+
+    static_assert(alignof(Type) == 8 || alignof(Type) == 4);
+
+    /** A block of memory that is owned by a slab allocator */
+    struct SlabBlock
+    {
+        // A mutex to protect the freelist for this block:
+        std::mutex m_;
+
+        // A linked list of appropriately sized free buffers:
+        std::uint8_t* l_ = nullptr;
+
+        // The next memory block
+        SlabBlock* next_;
+
+        // The underlying memory block:
+        std::uint8_t const* const p_ = nullptr;
+
+        // The extent of the underlying memory block:
+        std::size_t const size_;
+
+        SlabBlock(
+            SlabBlock* next,
+            std::uint8_t* data,
+            std::size_t size,
+            std::size_t item)
+            : next_(next), p_(data), size_(size)
+        {
+            // We don't need to grab the mutex here, since we're the only
+            // ones with access at this moment.
+
+            while (data + item <= p_ + size_)
+            {
+                *reinterpret_cast<std::uint8_t**>(data) = l_;
+                l_ = data;
+                data += item;
+            }
+        }
+
+        ~SlabBlock()
+        {
+            // Calling this destructor will release the allocated memory but
+            // will not properly destroy any objects that are constructed in
+            // the block itself.
+        }
+
+        SlabBlock(SlabBlock const& other) = delete;
+        SlabBlock&
+        operator=(SlabBlock const& other) = delete;
+
+        SlabBlock(SlabBlock&& other) = delete;
+        SlabBlock&
+        operator=(SlabBlock&& other) = delete;
+
+        /** Determines whether the given pointer belongs to this allocator */
+        bool
+        own(std::uint8_t const* p) const noexcept
+        {
+            return (p >= p_) && (p < p_ + size_);
+        }
+
+        std::uint8_t*
+        allocate() noexcept
+        {
+            std::uint8_t* ret;
+
+            {
+                std::lock_guard l(m_);
+
+                ret = l_;
+
+                if (ret)
+                    l_ = *reinterpret_cast<std::uint8_t**>(ret);
+            }
+
+            return ret;
+        }
+
+        /** Return an item to this allocator's freelist.
+
+            @param ptr The pointer to the chunk of memory being deallocated.
+
+            @note This is a dangerous, private interface; the item being
+                  returned should belong to this allocator. Debug builds
+                  will check and assert if this is not the case. Release
+                  builds will not.
+         */
+        void
+        deallocate(std::uint8_t* ptr) noexcept
+        {
+            assert(own(ptr));
+
+            std::lock_guard l(m_);
+            *reinterpret_cast<std::uint8_t**>(ptr) = l_;
+            l_ = ptr;
+        }
+    };
+
+private:
+    // A linked list of slabs
+    std::atomic<SlabBlock*> slabs_ = nullptr;
+
+    // The alignment requirements of the item we're allocating:
+    std::size_t const itemAlignment_;
+
+    // The size of an item, including the extra bytes requested and
+    // any padding needed for alignment purposes:
+    std::size_t const itemSize_;
+
+    // The size of each individual slab:
+    std::size_t const slabSize_;
+
+public:
+    /** Constructs a slab allocator able to allocate objects of a fixed size
+
+        @param count the number of items the slab allocator can allocate; note
+                     that a count of 0 is valid and means that the allocator
+                     is, effectively, disabled. This can be very useful in some
+                     contexts (e.g. when mimimal memory usage is needed) and
+                     allows for graceful failure.
+     */
+    constexpr explicit SlabAllocator(
+        std::size_t extra,
+        std::size_t alloc = 0,
+        std::size_t align = 0)
+        : itemAlignment_(align ? align : alignof(Type))
+        , itemSize_(
+              boost::alignment::align_up(sizeof(Type) + extra, itemAlignment_))
+        , slabSize_(alloc)
+    {
+        assert((itemAlignment_ & (itemAlignment_ - 1)) == 0);
+    }
+
+    SlabAllocator(SlabAllocator const& other) = delete;
+    SlabAllocator&
+    operator=(SlabAllocator const& other) = delete;
+
+    SlabAllocator(SlabAllocator&& other) = delete;
+    SlabAllocator&
+    operator=(SlabAllocator&& other) = delete;
+
+    ~SlabAllocator()
+    {
+        // FIXME: We can't destroy the memory blocks we've allocated, because
+        //        we can't be sure that they are not being used. Cleaning the
+        //        shutdown process up could make this possible.
+    }
+
+    /** Returns the size of the memory block this allocator returns. */
+    constexpr std::size_t
+    size() const noexcept
+    {
+        return itemSize_;
+    }
+
+    /** Returns a suitably aligned pointer, if one is available.
+
+        @return a pointer to a block of memory from the allocator, or
+                nullptr if the allocator can't satisfy this request.
+     */
+    std::uint8_t*
+    allocate() noexcept
+    {
+        auto slab = slabs_.load();
+
+        while (slab != nullptr)
+        {
+            if (auto ret = slab->allocate())
+                return ret;
+
+            slab = slab->next_;
+        }
+
+        // No slab can satisfy our request, so we attempt to allocate a new
+        // one here:
+        std::size_t size = slabSize_;
+
+        // We want to allocate the memory at a 2 MiB boundary, to make it
+        // possible to use hugepage mappings on Linux:
+        auto buf =
+            boost::alignment::aligned_alloc(megabytes(std::size_t(2)), size);
+
+        // clang-format off
+        if (!buf) [[unlikely]]
+            return nullptr;
+        // clang-format on
+
+#if BOOST_OS_LINUX
+        // When allocating large blocks, attempt to leverage Linux's
+        // transparent hugepage support. It is unclear and difficult
+        // to accurately determine if doing this impacts performance
+        // enough to justify using platform-specific tricks.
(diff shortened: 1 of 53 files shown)


src/ripple/shamap/SHAMapItem.h
```diff
@@ -20,52 +20,189 @@
...
+    }
+}
+
+inline boost::intrusive_ptr<SHAMapItem>
+make_shamapitem(uint256 const& tag, Slice data)
+{
+    assert(data.size() <= megabytes<std::size_t>(64));
+
+    std::uint8_t* raw = [&data]() -> std::uint8_t* {
+        if (data.size() < 128)
```
mtrippled, line 164: Should these be <= instead of < ? If not, what's the extra byte in the slab for?

mtrippled (approved):
In perftesting with NFTs, this commit reduces memory size by about 15% sustained over long duration.
LGTM
:+1:

scottschurr (changes requested):
Just a couple of notes:

1. The clang macOS library does not support `std::aligned_alloc` yet.  As you noted, that can be patched by `#include <stdlib.h>` and using `aligned_alloc`.  That patch is required in two files: SlabAllocator.h and SHAMapItem.h
2. Once that patch is made, you can remove the `#include <cstdlib>` in SlabAllocator.h, at least with clang on macOS.
3. Clang warns me that the `std::move` on line 101 of SHAMapSync_test.cpp is a pessimization.  I removed that `std::move`, since I build with warnings as errors.
4. Once those changes are made, Immediately upon invoking the unit tests I get an assert:
```
Assertion failed: (p_ != nullptr), function SlabAllocator, file /Users/scott/Development/rippled/src/ripple/basics/SlabAllocator.h, line 126.
Abort trap: 6
```
That assert suggests using non-`std` `aligned_alloc` may not work as well as hoped.

That's all I have for now.  I'll look more once those comments are addressed.

src/ripple/basics/SlabAllocator.h
```diff
@@ -0,0 +1,235 @@
...
+    std::atomic<std::uint64_t> dealloc_fast_count_ = 0;
+
+    // Scrubs the memory range [ptr, ptr+Size)
+    std::uint8_t*
+    scrub(std::uint8_t* ptr, [[maybe_unused]] std::uint8_t value)
+    {
+        assert(ptr != nullptr);
+
+#ifdef SLAB_SCRUB_MEMORY
+        std::memset(ptr, value, Size);
```
HowardHinnant, line 90: Where is `Size` defined?  Did you mean `size`?

src/ripple/basics/SlabAllocator.h
```diff
@@ -0,0 +1,235 @@
...
+            dealloc_fast_count_ = 0;
+        }
+    }
+
+    ~SlabAllocator()
+    {
+        if (p_ != nullptr)
+            std::free(p_);
+    }
+
```
HowardHinnant, line 145: The `SlabAllocator` copy constructor and copy assignment operators don't look right to me.  I believe they can result in a double-free.

src/ripple/basics/SlabAllocator.h
```diff
@@ -0,0 +1,235 @@
...
+    /** Constructs a slab allocator able to allocate up to count items
+
+        @param count the number of items the slab allocator can allocate; note
+                     that a count of 0 is valid and means that the allocator
+                     is, effectively, disabled. This can be very useful in some
+                     contexts (e.g. when mimimal memory usage is needed) and
+                     allows for graceful failure.
+     */
+    SlabAllocator(
```
HowardHinnant, line 116: Should be marked `explicit` to prevent implicit conversion from `size_t` to `SlabAllocator<T>`.

src/ripple/basics/SlabAllocator.h
greg7mdp: @nbougalis I fixed that in my branch

src/ripple/basics/SlabAllocator.h
```diff
@@ -0,0 +1,235 @@
...
+    count() const
+    {
+        return count_;
+    }
+
+    /** Determines whether the given pointer belongs to this allocator */
+    bool
+    own(std::uint8_t const* p) const
+    {
+        return (p >= p_) && (p < p_ + (size * count_));
```
HowardHinnant, line 157: Technically this is undefined behavior if `!own` because you can't compare pointers to different objects.  However from a practical standpoint, this works, and I know of no other way to do it.  I've written code like this myself and used it widely (in libc++).  I note this just so that we go into it with eyes wide open.

src/ripple/nodestore/impl/DecodedBlob.cpp
```diff
@@ -38,7 +38,7 @@ DecodedBlob::DecodedBlob(void const* key, void const* value, int valueBytes)
 
     m_success = false;
     m_key = key;
-    // VFALCO NOTE Ledger indexes should have started at 1
+    // VFALCO NOTE Ledger indexes should hav    e started at 1
```
HowardHinnant, line 41: Is this an errant modification?

src/test/shamap/SHAMapSync_test.cpp
```diff
@@ -98,7 +98,7 @@ class SHAMapSync_test : public beast::unit_test::suite
         for (int i = 0; i < items; ++i)
         {
             source.addItem(
-                SHAMapNodeType::tnACCOUNT_STATE, std::move(*makeRandomAS()));
+                SHAMapNodeType::tnACCOUNT_STATE, std::move(makeRandomAS()));
```
HowardHinnant, line 101: Remove `std::move` to allow the temporary `intrusive_ptr` to bind directly to the `addItem` argument.

src/ripple/basics/SlabAllocator.h
```diff
@@ -0,0 +1,235 @@
...
+    SlabAllocator(
+        std::size_t count,
+        std::string name = "Slab: " + beast::type_name<Type>())
+        : name_(name), count_(count)
+    {
+        if (count != 0)
+        {
+            p_ = reinterpret_cast<std::uint8_t*>(
+                std::aligned_alloc(alignof(Type), size * count_));
+            assert(p_ != nullptr);
```
HowardHinnant, line 125: Major snag:  `aligned_alloc` always returns `nullptr` using very latest macOS and tools.

HowardHinnant: I replaced `alignof(Type)` with `(alignof(Type) >= 8 ? alignof(Type) : 8)` (everywhere) and that got `aligned_alloc` to return non-null pointers.  But I then got tons of unittest failures.

HowardHinnant (changes requested):
Not yet working on macOS, see comments.

scottschurr:
I thought I saw a fairly recent push to this pull request.  So I tried building it.  The previously observed build failures on macOS are still present.

greg7mdp:
@scottschurr this is Nik's branch with my my additional changes which should build https://github.com/greg7mdp/rippled/tree/slab_alloc

mtrippled:
@greg7mdp I think you should open a new PR with all of the changes, and @nbougalis should close this one. What do you think, @HowardHinnant & @scottschurr ?

HowardHinnant:
I have a slight preference for keeping this one open so I can remember what I said 19 days ago.  But I won't object if everyone wants to move to a new PR.

greg7mdp:
@mtrippled I also think it would make sense to create a new PR. I suggested it a few days back, but @nbougalis said he would take care of it, probably he'd like to review the branch first.
@scottschurr I just verified that [this branch](https://github.com/greg7mdp/rippled/tree/slab_alloc) builds fine on mac (with some unrelated warnings).

scottschurr:
@greg7mdp, I'm fine either way, either updating this pull request or creating a new one.

scottschurr:
And, @greg7mdp, I tried building your branch locally.  Yup, your version works way better on macOS than the previous version did.  Thanks!  I think I will wait until your version is in an official pull request, however, before I start a full fledged review of the revised code.

greg7mdp:
Thanks @scottschurr for trying the mac version. I did build my branch on all 3 platforms (linux, windows, mac). 
@nbougalis , we should either update that PR or start a new one, so the review can proceed.

nbougalis:
I've addressed the concerns and issues highlighted by reviewers. I have also incorporated some of the ideas by @greg7mdp into `SlabAllocator`, and refined the memory allocation engine.

I _believe_ that this is ready for another review and, provided it's signed-off, a good candidate for merging.

I expect a reduction in overall memory usage of at least 15% on mainnet servers, but it depends on a number of factors. On my test bed, the reduction is closer to 22%. Performance should also improve.

Thanks.

src/ripple/basics/SlabAllocator.h
```diff
@@ -0,0 +1,421 @@
...
+public:
+    /** Constructs a slab allocator able to allocate objects of a fixed size
+
+        @param count the number of items the slab allocator can allocate; note
+                     that a count of 0 is valid and means that the allocator
+                     is, effectively, disabled. This can be very useful in some
+                     contexts (e.g. when mimimal memory usage is needed) and
+                     allows for graceful failure.
+     */
+    constexpr SlabAllocator(
```
HowardHinnant, line 167: Constructor should be marked explicit to prevent implicit conversion from `size_t` to `SlabAllocator<T>`.

src/ripple/nodestore/impl/DecodedBlob.cpp
HowardHinnant: Ping.

HowardHinnant:
```
rippled/src/test/shamap/SHAMapSync_test.cpp:101:50: warning: moving a temporary object prevents copy elision [-Wpessimizing-move]
                SHAMapNodeType::tnACCOUNT_STATE, std::move(makeRandomAS()));
                                                 ^
/Users/howardhinnant/Development/rippled/src/test/shamap/SHAMapSync_test.cpp:101:50: note: remove std::move call here
                SHAMapNodeType::tnACCOUNT_STATE, std::move(makeRandomAS()));
                                                 ^~~~~~~~~~              ~
1 warning generated.
```

HowardHinnant (changes requested):

intelliot:
@nbougalis please review Howard's suggestions above, at your convenience.

HowardHinnant (approved):

intelliot:
@scottschurr can you confirm whether your concerns have been resolved?

scottschurr:
@intelliot, no, I can't.  I have another higher priority code review I'm working on.  I can't provide an ETA for when I'll get to this one.

intelliot:
I checked with @scottschurr and we agreed that he has no particular need to review this PR, and we think it would be best to assign a different reviewer in order to move this forward. scottschurr is currently focused on other reviews that have higher priority.

I'll find another reviewer shortly. If that other reviewer approves, then this PR can be merged.

mtrippled:
> 

Both @HowardHinnant and I have approved this already. How many reviewers does this need?

intelliot:
@drlongle just FYI since this was discussed with Mark today. This PR isn't blocking on your review, but please feel free to have a look and share any questions/comments.

nbougalis:
This has a unit test failure that’s unrelated to these changes and likely traces back to the recent “simplified fee” proposal by @ximinez.

src/ripple/basics/SlabAllocator.h
HowardHinnant: Resolved.

src/ripple/basics/SlabAllocator.h
HowardHinnant: Resolved.

src/ripple/basics/SlabAllocator.h
HowardHinnant: Resolved.

src/ripple/nodestore/impl/DecodedBlob.cpp
HowardHinnant: Resolved.

src/test/shamap/SHAMapSync_test.cpp
HowardHinnant: Resolved.

HowardHinnant (approved):
All of my comments are resolved.

This builds and unittests for me using the latest macOS tools.

intelliot:
I merged `develop` and applied clang-format.patch. If acceptable to you, @nbougalis , this PR will be squashed when it is merged, so it's OK to leave the merge commit for the moment. Here is the commit message I suggest; @nbougalis please confirm that this looks good (or suggest an alternative).

```
Optimize SHAMapItem with slabbed allocator: (#4218)

* Introduce support for a slab-based memory allocator:

When instantiating a large amount of fixed-sized objects on the heap the
overhead that dynamic memory allocation APIs impose will quickly become
significant.

In some cases, allocating a large amount of memory at once and using a
slabbing allocator to carve the large block into fixed-sized units that
are used to service requests for memory out will help to reduce memory
fragmentation significantly and, potentially, improve overall
performance.

The new `SlabAllocator<>` class exposes an API that is _similar_ to the
C++ concept of an `Allocator` but it is not meant to be a
general-purpose allocator.

It should not be used unless profiling and analysis of specific memory
allocation patterns indicates that the additional complexity introduced
will improve the performance of the system overall, and subsequent
profiling proves it.

A helper class, `SlabAllocatorSet<>` simplifies handling of variably
sized objects that benefit from slab allocations.

This commit incorporates improvements suggested by Greg Popovitch
(@greg7mdp).

* Optimize `SHAMapItem` and leverage new slab allocator:

The `SHAMapItem` class contains a variable-sized buffer that holds the
serialized data associated with a particular item inside a `SHAMap`.

Previously, the buffer for the serialized data was allocated separately.
The fact that most instances of `SHAMapItem` were wrapped around a
`std::shared_ptr` meant that an instantiation might result in up to
three separate memory allocations.

This commit switches away from `std::shared_ptr` for `SHAMapItem` and
uses `boost::intrusive_ptr` instead, allowing the reference count for an
instance to live inside the instance itself. Coupled with using a
slab-based allocator to optimize memory allocation for the most commonly
sized buffers, the net result is significant memory savings. In testing,
the reduction in memory usage hovers between 400MB and 650MB. Other
scenarios might result in larger savings.

In performance testing with NFTs, this commit reduces memory size by
about 15% sustained over long duration.

* Avoid using std::shared_ptr when not necessary:

The `Ledger` class contains two `SHAMap` instances: the state and
transaction maps. Previously, the maps were dynamically allocated using
`std::make_shared` despite the fact that they did not require lifetime
management separate from the lifetime of the `Ledger` instance to which
they belong.

The two `SHAMap` instances are now regular member variables. Some smart
pointers and dynamic memory allocation was avoided by using stack-based
alternatives.
```

nbougalis:
Squashing these three together feels a bit like going mixing the appetizer, the main course and the dessert together on one plate.

I think that the three commits are functionally separate, and each is self-contained. IMO, keeping them that way is cleaner, more professional and makes it easier to review and, if needed, run `git bisect` in the future.

intelliot:
We had a comparable situation with https://github.com/XRPLF/rippled/pull/4192: the PR author recommended keeping the commits as-is, and so we did. Later, we learned there was some discontent with that, and that we should have squashed all the commits together. 

I don't personally have any opinion either way, but if we're going to recommend merging this as three separate commits, then we should at least confirm that each commit stands on its own and passes unit tests & clang-format. It may make sense to open 2-3 new PRs for this purpose.

HowardHinnant:
What we decided in #4192 is that if reviewers feel that a commit should be merged, and the commit message does not begin with `[FOLD]`, then the reviewer should recommend a change to the commit message.  And vice-versa.

intelliot:
blocked: let's get some agreement on whether this PR should end up as 3 commits, or 1.

If it's 3 commits, I think it would be best to open a couple of separate PRs the individual parts, so that we can confirm they are mergable on their own (they make sense and pass unit tests & clang-format).

intelliot:
@seelabs confirmed 3 commits for this is fine. I am rebasing the commits individually to ensure they pass CI on GitHub. If they pass, I will merge them no earlier than 2023-04-11 at 8:00 AM PT. If you have any objections, please comment here on this PR.

intelliot:
will merge after github checks pass, if I don't hear any objections

intelliot:
@manojsdoshi to try to corroborate the 15-22% memory use reduction

PR #4183, opened 2022-05-28
[TRIVIAL] Fix bitwise or on boolean operands warning / error

ximinez:
## High Level Overview of Change

I compile rippled with `-werr`. Clang 14 generated the following error.
```
src/ripple/app/tx/impl/CreateOffer.cpp:731:13: error: use of bitwise '&' with boolean operands [-Werror,-Wbitwise-instead-of-logical]
        if (!takerAmount.in.native() & !takerAmount.out.native())
            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
                                     &&
src/ripple/app/tx/impl/CreateOffer.cpp:731:13: note: cast one or both operands to int to silence this warning
1 error generated.
```

### Context of Change

I updated to Ubuntu 22.04, which ships with clang 14. Building rippled on the new system failed because of this issue.

### Type of Change

- [X ] Refactor (non-breaking change that only restructures code)

PR #4180, opened 2022-05-25
[TRIVIAL] Lower the message level for missing optional doc components

ximinez:
## High Level Overview of Change

If certain doc-related components are not available at build time, cmake will output warnings, which are noisy and annoying and completely unnecessary if one doesn't intend to build docs. I've observed this mostly on Windows where the components are less likely to already be available by default. This change lowers the message from a `WARNING` to a `NOTICE`, which is much less disruptive, but still informative.

### Type of Change

- [ X] Documentation Updates

 ## Before / After

### Before

```
[...]
1> [CMake] -- Pausing to download NuDB...
1> [CMake] -- Could NOT find Protobuf (missing: Protobuf_LIBRARIES Protobuf_INCLUDE_DIR) (Required is at least version "3.8")
1> [CMake] -- using local protobuf build.
1> [CMake] -- Couldn't find protobuf_protoc_lib
1> [CMake] CMake Warning at Builds/CMake/RippledDocs.cmake:34 (message):
1> [CMake]   could not find plantuml.jar
1> [CMake] Call Stack (most recent call first):
1> [CMake]   Builds/CMake/RippledDocs.cmake:40 (verbose_find_path)
1> [CMake]   CMakeLists.txt:77 (include)
1> [CMake] 
1> [CMake] 
1> [CMake] CMake Warning at Builds/CMake/RippledDocs.cmake:34 (message):
1> [CMake]   could not find dot
1> [CMake] Call Stack (most recent call first):
1> [CMake]   Builds/CMake/RippledDocs.cmake:41 (verbose_find_path)
1> [CMake]   CMakeLists.txt:77 (include)
1> [CMake] 
1> [CMake] 
1> [CMake] -- Configuring done
[...]
```

### After

```
[...]
1> [CMake] -- Pausing to download NuDB...
1> [CMake] -- Could NOT find Protobuf (missing: Protobuf_LIBRARIES Protobuf_INCLUDE_DIR) (Required is at least version "3.8")
1> [CMake] -- using local protobuf build.
1> [CMake] -- Couldn't find protobuf_protoc_lib
1> [CMake] could not find plantuml.jar
1> [CMake] could not find dot
1> [CMake] -- Configuring done
[...]
```

PR #4166, opened 2022-05-16
Reduce console noise coming from unit tests:

scottschurr:
### Context of Change

A few unit tests have historically generated noise to the console caused by log writes.  This noise was not useful and made it harder to locate actual test failures.

By changing the log level of these tests from

- `severities::kError` to
- `severities::kDisabled`

it was possible to remove that noise coming from the logs.

Only those unit tests currently producing log noise were changed.  This is so, in the future, we can see if pre-existing unit tests start producing unexpected log noise.

### Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)

PR #4219, opened 2022-06-30
Proposed 1.9.2-rc1

nbougalis:
Roll-up PR for **1.9.2-rc**, the first release candidate for **1.9.2**, which introduces a fix for the amendment voting persistence issue described in #4220, as well as a fix for "negative offers" for NFTs.

Primarily, this release will introduce a new amendment, **`NonFungibleTokensV1_1`** which is a combination of three amendments: **`NonFungibleTokensV1`**, **`fixNFTokenDirV1`** and **`fixNFTokenNegOffer`**. The maintainers of the reference implementation of the protocol recommend that operators that wish to vote for activation of the XLS-20 NFT enhancements vote in favor of **`NonFungibleTokensV1_1`** instead of **`NonFungibleTokensV1`**.

If merged, this release will:

- close #4166
- close #4173 
- close #4178 
- close #4180 
- close #4182
- close #4183 
- close #4188 
- close #4202
- close #4204 
- close #4213
- close #4215 
- close #4221 
- close #4222

### DIFF
--- a/src/ripple/protocol/Feature.h
+++ b/src/ripple/protocol/Feature.h
@@ -74,7 +74,7 @@ namespace detail {
 // Feature.cpp. Because it's only used to reserve storage, and determine how
 // large to make the FeatureBitset, it MAY be larger. It MUST NOT be less than
 // the actual number of amendments. A LogicError on startup will verify this.
-static constexpr std::size_t numFeatures = 49;
+static constexpr std::size_t numFeatures = 50;
 
 /** Amendments that this server supports and the default voting behavior.
    Whether they are enabled depends on the Rules defined in the validated
@@ -337,6 +337,7 @@ extern uint256 const featureNonFungibleTokensV1;
 extern uint256 const featureExpandedSignerList;
 extern uint256 const fixNFTokenDirV1;
 extern uint256 const fixNFTokenNegOffer;
+extern uint256 const featureNonFungibleTokensV1_1;
 
 }  // namespace ripple
 
--- a/src/ripple/protocol/Feature.h
+++ b/src/ripple/protocol/Feature.h
@@ -74,7 +74,7 @@ namespace detail {
 // Feature.cpp. Because it's only used to reserve storage, and determine how
 // large to make the FeatureBitset, it MAY be larger. It MUST NOT be less than
 // the actual number of amendments. A LogicError on startup will verify this.
-static constexpr std::size_t numFeatures = 48;
+static constexpr std::size_t numFeatures = 49;
 
 /** Amendments that this server supports and the default voting behavior.
    Whether they are enabled depends on the Rules defined in the validated
@@ -336,6 +336,7 @@ extern uint256 const featureCheckCashMakesTrustLine;
 extern uint256 const featureNonFungibleTokensV1;
 extern uint256 const featureExpandedSignerList;
 extern uint256 const fixNFTokenDirV1;
+extern uint256 const fixNFTokenNegOffer;
 
 }  // namespace ripple
 
--- a/src/ripple/protocol/impl/Rules.cpp
+++ b/src/ripple/protocol/impl/Rules.cpp
@@ -17,9 +17,8 @@
 */
 //==============================================================================
 
+#include <ripple/protocol/Feature.h>
 #include <ripple/protocol/Rules.h>
-
-#include <ripple/protocol/Indexes.h>
 
 namespace ripple {
 
@@ -40,9 +39,8 @@ public:
         std::unordered_set<uint256, beast::uhash<>> const& presets,
         std::optional<uint256> const& digest,
         STVector256 const& amendments)
-        : presets_(presets)
+        : presets_(presets), digest_(digest)
     {
-        digest_ = digest;
         set_.reserve(amendments.size());
         set_.insert(amendments.begin(), amendments.end());
     }
@@ -83,6 +81,18 @@ bool
 Rules::enabled(uint256 const& feature) const
 {
     assert(impl_);
+
+    // The functionality of the "NonFungibleTokensV1_1" amendment is
+    // precisely the functionality of the following three amendments
+    // so if their status is ever queried individually, we inject an
+    // extra check here to simplify the checking elsewhere.
+    if (feature == featureNonFungibleTokensV1 ||
+        feature == fixNFTokenNegOffer || feature == fixNFTokenDirV1)
+    {
+        if (impl_->enabled(featureNonFungibleTokensV1_1))
+            return true;
+    }
+
     return impl_->enabled(feature);
 }
 
--- a/src/ripple/protocol/impl/Rules.cpp
+++ b/src/ripple/protocol/impl/Rules.cpp
@@ -39,7 +39,7 @@ public:
         std::unordered_set<uint256, beast::uhash<>> const& presets,
         std::optional<uint256> const& digest,
         STVector256 const& amendments)
-        : presets_(presets), digest_(digest)
+        : digest_(digest), presets_(presets)
     {
         set_.reserve(amendments.size());
         set_.insert(amendments.begin(), amendments.end());
--- a/Builds/CMake/RippledCore.cmake
+++ b/Builds/CMake/RippledCore.cmake
@@ -156,6 +156,7 @@ install (
     src/ripple/basics/MathUtilities.h
     src/ripple/basics/safe_cast.h
     src/ripple/basics/Slice.h
+    src/ripple/basics/spinlock.h
     src/ripple/basics/StringUtilities.h
     src/ripple/basics/ThreadSafetyAnalysis.h
     src/ripple/basics/ToString.h
--- a/Builds/CMake/RippledCore.cmake
+++ b/Builds/CMake/RippledCore.cmake
@@ -157,6 +157,7 @@ install (
     src/ripple/basics/safe_cast.h
     src/ripple/basics/Slice.h
     src/ripple/basics/StringUtilities.h
+    src/ripple/basics/ThreadSafetyAnalysis.h
     src/ripple/basics/ToString.h
     src/ripple/basics/UnorderedContainers.h
     src/ripple/basics/XRPAmount.h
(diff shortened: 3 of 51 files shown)


nbougalis:
**Important Note for Reviewers, Maintainers and other interested parties**:

This PR includes commit c4d707aa656eaeb2541f615a0649a58a27ef1ced, which **changes the default vote** for two amendments: the previously introduced `fixNFTokenDirV1` and `fixNFTokenNegOffer` (introduced with PR #4215).

Prior to this commit, the two fix amendments had a default vote of **no** and this changes the vote to **yes**.

Considering that both of these amendments **correct bugs** (as opposed to introducing new features) and that, _even if activated_,  their code is dormant and will remain gated under the `featureNonFungibleTokensV1` amendment (which defaulted and continues to default to a **no**) the impact of this change should be minimal. The primary rationale for this change is to add an additional layer of protection, so as to avoid the accidental activation of the `featureNonFungibleTokensV1` without these important fixes.

Note that, while not presently used outside the inactive NFT code, the fixes introduced in the `fixNFTokenDirV1` amendment may prove useful independently of whether NFT support is added, as the new-style page structures can be leveraged to build a more compact and efficient method for dealing with owner directories.

nbougalis:
> While I'm in favour of this, a default YES vote would initiate the timer once sufficient are updated and would cause an early amendment block if sufficient didn't change their vote to no. This may be a problem.

I agree but that's the same as with _any_ amendment. I think it would be a good to rethink the way amendments are proposed how they work, especially for amendments that are dependent on other amendment.

rippleitinnz:
Ultimately, the resolution would be to have trusted validators who were timely, proactive in discussing and engaging with the voting process on our UNL lists.

nbougalis:
So, obviously there's some controversy about setting the two fix amendments to "default yet". I've noticed some vigorous conversations on Twitter and the Mattermost channels, perhaps in the Discord too.

I think this is important to discuss publicly and transparently.

As I wrote earlier, _"[c]onsidering that both of these amendments correct bugs (as opposed to introducing new features) and that, even if activated, their code is dormant and will remain gated under the featureNonFungibleTokensV1 amendment (which defaulted and continues to default to a no) the impact of this change should be minimal._"

I think my analysis is reasonable but my conclusion is not: the impact is **not** minimal, in the sense that if these amendments activate, anyone running **1.8.x**, **1.9.0** or **1.9.1** will become amendment blocked _for no real reason_: the amendments don't do anything.

Broadly, these raises some questions about the flexibility of the amendment code and whether it should be improved to allow, inter alia, the possibility of an amendment that activates but doesn't amendment block anyone, because it's dependent on another amendment. But I don't want to detract from the issue at hand: changing the default vote for the fixes.

Again, I believe this is a change that I can defend rationally, but I understand that it may be controversial, especially given the broader context.

I hope people will chime in here and we can have a productive and meaningful conversation that results in consensus on the way forward, regardless of what that consensus is.

Thanks!

Mwni:
Why don't we add the option for an amendment to default on the setting of another amendment?
So the example would be, the two fixes would have the `default = nonFungibleTokensV2.enabled`. That way, unconscientious validator operators would be opted out of all NFT related amendments, while those who care still have full control.

nbougalis:
> Why don't we add the option for an amendment to default on the setting of another amendment? So the example would be, the two fixes would have the `default = nonFungibleTokensV2.enabled`. That way, unconscientious validator operators would be opted out of all NFT related amendments, while those who care still have full control.

Interesting suggestion. We'd discussed something like this before. I think having "conditional amendments" is a good idea: "A requires B" or "C implies D, E and F" but we don't have that and coding it would, itself, require an amendment. So chicken and egg for _this_ case. But certainly worth considering.

MarkusTeufelberger:
Well, servers could just not propose Amendment A until they see also Amendment B supported on the network by default if A would be broken without B.

whywefight:
While I am somehow in favour of a default yes, if dependencies exist, we shouldn't do it this time. Our main target, for now, should be getting as many nodes updated as possible, as soon as possible. From a dUNL standpoint we should be fine, its the rest of the network we need to worry about. As mentioned before, we aren't even sure if outdated nodes will have an impact but the risk is too high from my point of view. 

> Broadly, these raises some questions about the flexibility of the amendment code and whether it should be improved to allow, inter alia, the possibility of an amendment that activates but doesn't amendment block anyone, because it's dependent on another amendment. But I don't want to detract from the issue at hand: changing the default vote for the fixes.

Moving forward that's something that needs to be looked into to avoid similar sitations.

Silkjaer:
> It makes sense to merge all NFToken related code+fixes into a single votable amendment.

I agree, as I believe it’s a lot easier for people to follow one voting process (instead of waiting for two fixes and main amendment). And then it sets a precedent for how to deal with similar situations in the future, without making changes to support dependency amendments or similar.

Silkjaer:
Option 3

crypticrabbit:
I agree with @gnurag and @Silkjaer that deprecating the current amendments in favor of `featureNonFungibleTokensV1.1`, and I agree that the amendment should default to `no`.

whywefight:
I am in favour of Option 3

Mwni:
Option 3

rippleitinnz:
Option 3 is the most logical. Having said that, a voter who has voted to ACCEPT, who becomes unaware of any fix and the subsequent removal of the amendment by it being replaced with a newer version, will have their vote recorded as REJECT despite them thinking they had voted to ENABLE. 

Currently there is no way to know that a vote you have cast has been changed or removed without checking your voting status.

alloynetworks:
I would go with option 3.

jonaagenilsen:
Option #3 😊

nixer89:
Option 3 makes the most sense for me 👌

wojake:
Option 3 is best IMO since it compiles all the *current* NFT-related amendments into a single amendment providing a better decision making process for all rippled node operators.

Digifin:
Option 3 sounds the best.

rswarthout:
Option 3

lathanbritz:
im good with option 3

calvincs:
Option three looks like a solid choice to me.

src/ripple/protocol/impl/Rules.cpp
```diff
@@ -83,6 +82,18 @@ bool
...
+    // precisely the functionality of the following three amendments
+    // so if their status is ever queried individually, we inject an
+    // extra check here to simplify the checking elsewhere.
+    if (feature == featureNonFungibleTokensV1 ||
+        feature == fixNFTokenNegOffer || feature == fixNFTokenDirV1)
+    {
+        if (impl_->enabled(featureNonFungibleTokensV1_1))
+            return true;
+    }
+
```
ximinez, line 96: The only downside to doing it this way is that this check will need to be made on every call to `enabled`. It's a tiny downside, since it's a simple check. *But* what do you think of, instead of this check, you modify `Change::applyAmendment`. If the enabled feature is `featureNonFungibleTokensV1_1`, you add the other three to the amendments list, too?
https://github.com/XRPLF/rippled/blob/83aa1c1818e2854e98eebce040b9dc6c72de640e/src/ripple/app/tx/impl/Change.cpp#L195-L197

src/ripple/protocol/Feature.h
```diff
@@ -337,6 +337,7 @@ extern uint256 const featureNonFungibleTokensV1;
 extern uint256 const featureExpandedSignerList;
 extern uint256 const fixNFTokenDirV1;
 extern uint256 const fixNFTokenNegOffer;
+extern uint256 const featureNonFungibleTokensV1_1;
```
ximinez, line 340: You need to increment `numFeatures`. https://github.com/XRPLF/rippled/blob/83aa1c1818e2854e98eebce040b9dc6c72de640e/src/ripple/protocol/Feature.h#L77

src/ripple/protocol/impl/Rules.cpp
```diff
@@ -17,10 +17,10 @@
 */
 //==============================================================================
 
+#include <ripple/protocol/Feature.h>
+//#include <ripple/protocol/Indexes.h>
```
ximinez, line 21: May as well remove the commented line.

ximinez (changes requested):

ximinez:
It also turns out that the `NonFungibleTokensV1_1` breaks several unit tests. I pushed a commit that fixes them, as well as `numFeatures`. https://github.com/ximinez/rippled/commit/ab5c80988edb5af2f1366b036beebd24aafa2ea6

nbougalis:
Per the discussion here and in the mattermost chat, I reverted the commit that changed the default votes for the two fix amendments and have, instead, introduced a new amendment, **`NonFungibleTokensV1_1`** which is a combination of three amendments: **`NonFungibleTokensV1`**, **`fixNFTokenDirV1`** and **`fixNFTokenNegOffer`**.

Assuming that this is merged, it should be interpreted as a recommendation by the maintainers of the reference implementation of the protocol that any operators who wish to vote for activation of the XLS-20 NFT enhancements vote in favor of **`NonFungibleTokensV1_1`** instead of **`NonFungibleTokensV1`**.

src/ripple/protocol/impl/Rules.cpp
```diff
@@ -40,9 +39,8 @@ class Rules::Impl
         std::unordered_set<uint256, beast::uhash<>> const& presets,
         std::optional<uint256> const& digest,
         STVector256 const& amendments)
-        : presets_(presets)
+        : presets_(presets), digest_(digest)
```
scottschurr, line 42: You need to swap the positions of these two member variables.  The `Rules::Impl` member variables are declared...
```
private:
    std::unordered_set<uint256, hardened_hash<>> set_;
    std::optional<uint256> digest_;
    std::unordered_set<uint256, beast::uhash<>> const& presets_;
```

src/ripple/protocol/impl/Rules.cpp
nbougalis: I thought about this (and prepared a branch to do just that, actually) but a couple of people that I floated that idea by felt that it was weird, because it could result in an amendment that doesn't, on its own, have majority appearing to get arbitrarily enabled.

I do think that's a better option in the long term and will likely introduce a PR to make that possible, but I don't think it's the right option here.

nixer89 (approved):
looking good

PR #4226, opened 2022-07-10
Building RocksDB is optional

legleux:

RocksDB is not a hard requirement for rippled and there's been requests for this as many people have no need.

### DIFF
--- a/Builds/CMake/deps/Rocksdb.cmake
+++ b/Builds/CMake/deps/Rocksdb.cmake
@@ -2,7 +2,7 @@
    NIH dep: rocksdb
 #]===================================================================]
 
-option(use_rocksdb "build rocksdb" OFF)
+option(use_rocksdb "build rocksdb" ON)
 
 if(NOT use_rocksdb)
   return()
--- a/Builds/CMake/deps/Rocksdb.cmake
+++ b/Builds/CMake/deps/Rocksdb.cmake
@@ -2,6 +2,12 @@
    NIH dep: rocksdb
 #]===================================================================]
 
+option(use_rocksdb "build rocksdb" OFF)
+
+if(NOT use_rocksdb)
+  return()
+endif()
+
 add_library (rocksdb_lib UNKNOWN IMPORTED GLOBAL)
 set_target_properties (rocksdb_lib
   PROPERTIES INTERFACE_COMPILE_DEFINITIONS RIPPLE_ROCKSDB_AVAILABLE=1)


legleux:
We may not need this as the release this would be merged will hopefully contain the Conan dependencies instead.

RichardAH:
Thanks for this PR!
The default option should mirror the current build regime as to not make a breaking change to existing build environments / build engines.
`option(use_rocksdb "build rocksdb" ON)`

RichardAH:
Can you squash/rebase into a single commit?

@thejohnfreeman Until the new build engine is PRed and merged, all changes to the existing build engine should progress as if the new engine did not exist. This is not to dismiss the new engine, but rather to ensure the existing engine remains serviceable in the meantime, and to ensure there is not undue pressure to deliver the new system.

RichardAH:
> @RichardAH not sure why you're addressing me; I didn't comment in this thread. I don't understand the rush to get this merged instead of addressing the open Conan PR that is ready to be merged, waiting on reviewers, while I lack permissions to request any reviewers through GitHub.
> 
> I don't care if this is merged or not. I will have to address the conflict with my `conan` branch, but it's not a big deal. I do care that the Conan PR seems to get far less attention. Is it because it is more complicated than this PR? Is that why you are avoiding it?

I’d prefer to let your colleagues at Ripple review it, if it’s all the same to you. That might take a while so in the meantime I support this small functional change to the existing build system.

For the record I think the build system does need an overhaul and I am generally in support of your efforts. I only mentioned you because you tagged this PR in your commit.

Ripple seem to have some closed book review procedures so I figure let them use those for their own commits. In different circumstances I would be happy to review your PR

legleux:
> Can you squash/rebase into a single commit?

I could, but the merge button should have an option to do it. In settings under general, I'm not sure what we have as I've never seen the settings.

image

RichardAH:
@legleux when I build your branch using the new `use_rocksdb=OFF` option, but supply a rippled.cfg that specifies RocksDB as the node_db type, I get a very unhelpful error message:
```
Loading: "/root/rocksoff/off/rippled.cfg"
2022-Jul-13 06:53:12.683127443 UTC JobQueue:NFO Using 14  threads
terminate called after throwing an instance of 'std::runtime_error'
  what():  Your rippled.cfg is missing a [node_db] entry, please see the rippled-example.cfg file!
Aborted (core dumped)
```

The exact same config when `use_rocksdb=ON` runs fine. Suggest adding a helpful error message if trying to use an incompatible config with the rocks-free build.

ximinez (approved):
Conan changes aside, this works.

intelliot:
@legleux does this still make sense with Conan?

Feel free to reach out to @thejohnfreeman if you'd like assistance figuring this out

legleux:
> @legleux does this still make sense with Conan?
> 
> Feel free to reach out to @thejohnfreeman if you'd like assistance figuring this out
Option is provided by conan now.
The error pointed out will still remain though but I don't think it's within scope of this feature, other syntax errors in `rippled.cfg` will give just as unhelpful results.

PR #4233, opened 2022-07-13
Wrap doCommand in a try block

seelabs:


seelabs:
Reviewers: @cjcobb23 @nbougalis

src/ripple/rpc/impl/ServerHandlerImp.cpp
```diff
@@ -874,7 +874,17 @@ ServerHandlerImp::processRequest(
...
+        try
+        {
+            RPC::doCommand(context, result);
+        }
+        catch (std::exception const& ex)
+        {
+            result[jss::result] = RPC::make_error(rpcINTERNAL);
+            JLOG(m_journal.error())
+                << "Exception while processing WS: " << ex.what() << "\n"
+                << "Input JSON: " << Json::Compact{Json::Value{result}};
```
nbougalis, line 886: Are we sure that attempting to print the (potentially) offending JSON won't, itself, generate an exception?

Also, I think that `result[jss::result] = RPC::make_error(rpcINTERNAL);` may be insufficient in some cases. See the code around line 896.

src/ripple/rpc/impl/ServerHandlerImp.cpp
```diff
@@ -874,7 +874,17 @@ ServerHandlerImp::processRequest(
...
         Json::Value result;
         auto start = std::chrono::system_clock::now();
-        RPC::doCommand(context, result);
+        try
+        {
+            RPC::doCommand(context, result);
+        }
+        catch (std::exception const& ex)
+        {
+            result[jss::result] = RPC::make_error(rpcINTERNAL);
```
cjcobb23, line 883: Is this nesting correct? Or are we supposed to just assign to `result` here? @natenichols

nbougalis (changes requested):
I think this is insufficient. Someone better versed in the RPC code should triple-check.

cjcobb23 (commented):
You could expand the scope of the try block if you want, to be more consistent with how we do it for websockets (https://github.com/XRPLF/rippled/blob/4e724794c57cddcd8b2f0754a4b32592b1ef3342/src/ripple/rpc/impl/ServerHandlerImp.cpp#L418)

src/ripple/rpc/impl/ServerHandlerImp.cpp
cjcobb23: I think we want to print the request here, not `result`

src/ripple/rpc/impl/ServerHandlerImp.cpp
natenichols: I _think_ we just assign to `result` here. 

The code at Line 901/902 and just uses `result[jss::error_code]` and `result[jss::error]`.

It looks like that code path would get run here, and those values would be `null`.

PR #4235, opened 2022-07-14
Correctly use the configured network_id parameter

chennakeshava1998:
The existing code would correctly parse the network_id parameter from the configuration
file, but it would not properly store it inside the overlay, which means that the server would
ignore the configured value for some operations. This commit corrects this flaw.


nbougalis (changes requested):
This should also remove the declaration of the `networkID_` member variable from `OverlayImpl.h` around line 128.

PR #4236, opened 2022-07-14
include network_id in server_info command - Issue 3383

### LINKED ISSUE #3383: Return network_id upon initial websocket connection and/or in server_info
## Summary
Aside from the live XRP Ledger network (known as Mainnet), there are also a few test networks that are used for testing and development purposes. Ripple operates two such networks, called Testnet and Devnet. When using a client library, users can make the mistake of connecting to a Mainnet server when they intended to use a server on a test network, or vice versa. Also, users on wallets or explorers may mix up which network they are on. Since there is only one XRP Ledger network that involves real money and transfers real value (Mainnet), it is the only one that is important to distinguish. It is useful to know, and confirm, whether you are currently using the Mainnet versus any other test network.

Building on the X-address format (which uses a different prefix depending on whether the address is intended for use on Mainnet versus a test network), and https://github.com/ripple/rippled/pull/3049 * (which adds a `network_id` configuration option), this feature request proposes that the rippled server return the `network_id` via websockets when a client first connects, and/or includes the field in `server_info` responses.

A `network_id` of `0` or `main` indicates Mainnet. Any other value indicates a test network.

\* Released in [1.5.0-b1](https://github.com/ripple/rippled/pull/3167) and 1.5.0 ([commit](https://github.com/crypticrabbit/rippled/commit/f6916bfd429ce654e017ae9686cb023d9e05408b)).

## Motivation
There are a few places in ripple-lib where we encode an X-address. If we don't know whether we are on Mainnet, then there is no way to know which prefix to use. Currently, the library assumes Mainnet and uses the 'X' prefix when parsing transactions. This will be 'fixed' by allowing the user to pass a parameter (indicating whether they are using Mainnet) when initializing a client, similar to [what the SDK already does](https://github.com/xpring-eng/Xpring-JS#xrpclient). However, users could still set the parameter to indicate Mainnet and still connect to a test network, since there is no way to check that the server is on the expected network.

While a malicious rippled server could always lie, having this feature would help users by eliminating accidental mixups.

Some developers figure out if they're on the wrong network by just observing whether their transaction fails, and then switching over. This could be dangerous when dealing with real money. If clients knew when they were connected to Mainnet, then a transaction involving a 'T' address could be easily and quickly rejected client-side, since it is certainly intended for a test network.

(Mainnet and test networks use the same transaction ID format, and if the same keys are used, it is theoretically possible for test transactions to run on Mainnet and vice versa.)

## Solution
Return the network_id to clients when they first connect via websockets.

Alternatively, or additionally, the `network_id` field can be included in `server_info` responses.

This tells the client which network the server is configured to follow.

## Paths Not Taken
We could argue this isn't useful because servers can lie. But that clearly isn't the case, since the same applies to everything that rippled returns. If the server lies, it's of no more consequence than it misbehaving in other ways. The goal here is to improve the user / developer experience, and this feature would. Admittedly, it's not a big deal as it just helps to catch mistakes. So, I suggest that this is a low-priority request.

With this feature, clients could theoretically "auto-detect" which network they are on. I do not recommend this. Rather, this feature should be used as an additional check; it allows clients to detect a mismatch, and report an error to the user.

ckeshava:
This PR implements the second solution proposed in the Issue 3383 (https://github.com/XRPLF/rippled/issues/3383#issue-609222880)

The existing code includes server_info only in the "reporting" mode, hence it is redundant.

Fix #3383


src/ripple/app/misc/NetworkOPs.cpp
```diff
@@ -2323,9 +2323,8 @@ NetworkOPsImp::getServerInfo(bool human, bool admin, bool counters)
     if (!app_.config().SERVER_DOMAIN.empty())
         info[jss::server_domain] = app_.config().SERVER_DOMAIN;
 
-    if (!app_.config().reporting())
-        if (auto const netid = app_.overlay().networkID())
-            info[jss::network_id] = static_cast<Json::UInt>(*netid);
+    if (auto const netid = app_.overlay().networkID(); netid.has_value())
```
nbougalis, line 2326: I don't know if removing the check for reporting mode is correct. Maybe @cjcobb23 can chime in?

Second, the `if` is unnecessarily complicated. It should be reverted to its previous form:

```
if (auto const netid = app_.overlay().networkID())
    info[jss::network_id] = static_cast<Json::UInt>(*netid);
```

nbougalis (changes requested):
Like with #4235, this should remove the declaration of the `networkID_` member variable around line 128 of `OverlayImpl.h`.

src/ripple/app/misc/NetworkOPs.cpp
chennakeshava1998: Okay, I didn't want to run into undefined behavior, so I did explicit checks. I'll revert it 👍

Issue #3318, opened 2020-03-25
CLI usage statement doesn't include manifest and validator_info (Version: 1.5.0-rc1)


mDuo13:
## Issue Description
The usage statement does list the `manifest` and `validator_info` commands, added by #3197.

## Steps to Reproduce
Print the usage statement:

```sh
rippled -h
```

## Expected Result
The result should list all supported commands.

## Actual Result
The usage statement does not list either `manifest` or `validator_info`.

## Environment
Arch Linux, x86_64, self-compiled v1.5.0-rc1.

## Supporting Files

The commands portion of the usage statement is in [printHelp() of Main.cpp](https://github.com/ripple/rippled/blob/0c6d380780ae368a2236a2e8e3e42efa4a1d2b46/src/ripple/app/main/Main.cpp#L125).

It might be possible to make the usage statement dynamic. Much of the necessary info is in [Handler.cpp](https://github.com/ripple/rippled/blob/develop/src/ripple/rpc/impl/Handler.cpp) although some of the parameter parsing isn't.

brianfranco141:
was this issue fixed? if not id gladly like to work on it

PR #4237, opened 2022-07-15
Fix Issue 3318 

chennakeshava1998:
Fixes Issue 3318 (https://github.com/XRPLF/rippled/issues/3318)
If merged, this commit will:
- Include manifest, validator_info information in the output of `rippled -h` command. 

Note: I found that printHelp() does not have all the commands listed in a lexicographically sorted order. If this is a requirement I can make that change as well.

For example: `validators` appears before `validator_list_sites`, but a computer would arrange these strings in the opposite order.


nbougalis (changes requested):
Looks good. Please reword the commit message:

Update command-line usage help message (fixes #3318)

PR #4239, opened 2022-07-15
Change enum values to be powers of two (fix #3417)

### LINKED ISSUE #3417: Software does not distinguish between different Conditions (Version: 1.5)
I discovered some code that just doesn't smell right. I don't know the intended behavior, but I suspect the intent is different than what is actually happening.

The code that checks whether the condition that is required for a given RPC is met does not actually consider what condition is required. All of the conditions except `NO_CONDITION` are treated exactly the same. All of the following lines (in the same function) evaluate to the exact same value (false if `required_condition` is `NO_CONDITION` and true otherwise).
https://github.com/ripple/rippled/blob/3d86b49dae8173344b39deb75e53170a9b6c5284/src/ripple/rpc/impl/Handler.h#L80
https://github.com/ripple/rippled/blob/3d86b49dae8173344b39deb75e53170a9b6c5284/src/ripple/rpc/impl/Handler.h#L90
https://github.com/ripple/rippled/blob/3d86b49dae8173344b39deb75e53170a9b6c5284/src/ripple/rpc/impl/Handler.h#L91
https://github.com/ripple/rippled/blob/3d86b49dae8173344b39deb75e53170a9b6c5284/src/ripple/rpc/impl/Handler.h#L97
https://github.com/ripple/rippled/blob/3d86b49dae8173344b39deb75e53170a9b6c5284/src/ripple/rpc/impl/Handler.h#L117
The reason these lines all evaluate to the exact same value is because of the way these conditions are defined: all except `NO_CONDITION` have the least significant bit set to 1. So, when we bitwise and any two conditions together, the result is false if either are `NO_CONDITION` and true otherwise. I don't think this was the intended behavior of the code, since all of these lines could be replaced with `condition_required != NO_CONDITION`. Maybe someone who understands the intention of this piece of code can chime in here.

ckeshava:
If merged, this code will:

- Implement the simplified condition evaluation
- removes the complex bitwise and(&) operator
- Implement the second proposed solution in Nik Bougalis's comment - https://github.com/XRPLF/rippled/issues/3417#issuecomment-635116233
- I have tested this code change by performing RPC calls with the commands `server_info`, `server_state`, `peers` and `validation_info`. These commands worked as expected.


chennakeshava1998:
@nbougalis  I have a question about the choice of integers inside the Condition enum. Does it necessarily have to be powers of 2? With respect to the operations in this file, it suffices to choose different integer values right? 

I understand that the choice of even/odd values is important because we do bit-wise and operator.

greg7mdp:
They have to be powers of two if you want to be able to set and test the bits individually.

For example if you define:

```
enum { A=1, B=2, C=3};
```

and you do:
```
uint32_t x = C;
```

the value of x will be 0x3 (the two lower bits set), so you would have both `(x & A) != 0` and `(x & B) != 0`

if you use powers of two the bits can be set and tested individually. I used to use enums like:

```
enum {
   A = 1 << 0,
   B = 1 << 1,
   C = 1 << 2,
   ...
   };
```

chennakeshava1998:
> They have to be powers of two if you want to be able to set and test the bits individually.
> 
> For example if you define:
> 
> ```
> enum { A=1, B=2, C=3};
> ```
> 
> and you do:
> 
> ```
> uint32_t x = C;
> ```
> 
> the value of x will be 0x3 (the two lower bits set), so you would have both `(x & A) != 0` and `(x & B) != 0`
> 
> if you use powers of two the bits can be set and tested individually. I used to use enums like:
> 
> ```
> enum {
>    A = 1 << 0,
>    B = 1 << 1,
>    C = 1 << 2,
>    ...
>    };
> ```

okay got it, thanks for the explanation! I would individually check all the conditions and that is quite verbose, this is concise 👍

greg7mdp:
You can use a sequence 1,2,3,4,5,... if all the conditions are exclusive from one another. But if they are flags that can be set independently, then it has to be separate bits to be workable.

chennakeshava1998:
> You can use a sequence 1,2,3,4,5,... if all the conditions are exclusive from one another. But if they are flags that can be set independently, then it has to be separate bits to be workable.

Okay, got it 👍

src/ripple/rpc/impl/Handler.h
```diff
@@ -39,8 +39,8 @@ namespace RPC {
 enum Condition {
     NO_CONDITION = 0,
     NEEDS_NETWORK_CONNECTION = 1,
-    NEEDS_CURRENT_LEDGER = 2 + NEEDS_NETWORK_CONNECTION,
-    NEEDS_CLOSED_LEDGER = 4 + NEEDS_NETWORK_CONNECTION,
+    NEEDS_CURRENT_LEDGER = 2 * NEEDS_NETWORK_CONNECTION,
+    NEEDS_CLOSED_LEDGER = 4 * NEEDS_NETWORK_CONNECTION,
```
HowardHinnant, line 43: Consider this purely stylistic change:
```c++
enum Condition {
    NO_CONDITION = 0,
    NEEDS_NETWORK_CONNECTION = 1,
    NEEDS_CURRENT_LEDGER = 2,
    NEEDS_CLOSED_LEDGER = 4,
};
```

HowardHinnant (approved):

intelliot:
@ckeshava would you like to pick this up?

ckeshava:
Hello,
I have completed the code changes. The unit tests are passing locally, I'm not sure why the CI tests are failing on the ubuntu machine.

src/ripple/rpc/impl/Handler.h
```diff
@@ -39,8 +39,8 @@ namespace RPC {
 enum Condition {
     NO_CONDITION = 0,
     NEEDS_NETWORK_CONNECTION = 1,
-    NEEDS_CURRENT_LEDGER = 2 + NEEDS_NETWORK_CONNECTION,
-    NEEDS_CLOSED_LEDGER = 4 + NEEDS_NETWORK_CONNECTION,
+    NEEDS_CURRENT_LEDGER = 1 << 1,
```
greg7mdp, line 42: Is `NEEDS_NETWORK_CONNECTION` not required anymore?

src/ripple/rpc/impl/Handler.h
```diff
@@ -94,20 +94,18 @@ conditionMet(Condition condition_required, T& context)
     }
 
     if (context.app.getOPs().isAmendmentBlocked() &&
-        (condition_required & NEEDS_CURRENT_LEDGER ||
-         condition_required & NEEDS_CLOSED_LEDGER))
+        (condition_required != NO_CONDITION))
```
greg7mdp, line 97: This is not the same test as before. Is the change on purpose?

ckeshava: yes, I agree that the conditions have changed. I made this change approximately one year ago, so I am spacing out on the details. Let me think/check and get back to you

ximinez: Based on the old values, I think the conditions are equivalent. Between `NEEDS_CURRENT_LEDGER` and `NEEDS_CLOSED_LEDGER`, having any of the three low bits set in `condition_required` would result in the expression evaluating to true, so that covers all of the `!= NO_CONDITION` conditions. So functionally, they are equivalent.

Now whether that was the _intent_ is a different question.

It turns out there are only two operations which have `NEEDS_NETWORK_CONNECTION`: `ledger_cleaner` and `tx`. I think it's fine to disallow both of them if the server is amendment blocked.

ximinez (approved):
I think this looks good, and doesn't appear to actually change any existing behavior.

intelliot:
(Holding this for 1.12 unless someone requests otherwise)

intelliot:
This has CI failures which need to be resolved (or explained). @ckeshava or @thejohnfreeman - could you have a look?

thejohnfreeman:
Please merge `develop`.

ckeshava:
OK, I've rebased to the `develop` branch, thanks for the tip!

intelliot:
@HowardHinnant because there have been changes pushed after your earlier approval, could you do a quick re-review?

HowardHinnant (approved):

PR #4240, opened 2022-07-15
Fix for issue 3298

chennakeshava1998:
If merged, this commit will:
- Provide comparative analysis of Data Structures
- time/profile performance of parseCommand function
- declare commands array as constexpr

### DIFF
--- a/src/ripple/net/impl/RPCCall.cpp
+++ b/src/ripple/net/impl/RPCCall.cpp
@@ -1227,10 +1227,17 @@ public:
             int maxParams;
         };
 
-        // FIXME: replace this with a function-static std::map and the lookup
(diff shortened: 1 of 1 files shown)


chennakeshava1998:
@scottschurr  @nbougalis  let me know if this looks okay.

nbougalis (changes requested):
Thank you for doing this, but I think that the inline comment is not necessary.

Instead, what I recommend, is that you change this commit to simply remove the `FIXME` comment and provide a description of the tests you performed in the commit (i.e. what you have as a comment) as part of the commit message.

chennakeshava1998:
> 

Okay, I have done the suggested changes 👍

intelliot (approved):

PR #4242, opened 2022-07-18
TaggedCache: Fix swept object memory release which can occur within the …

greg7mdp:
…lock

When a `std::shared_ptr` is destructed, even if there is no other `shared_ptr` referencing the same object, the object memory is not necessarily released. This is because `std::weak_ptr ` referencing the same object still need to reference the control block, and the control block is deleted only when the last `std::weak_ptr` is deleted.

When the object is allocated with `std::make_shared`, both the control block and the object are allocated within the same malloc block, and that malloc block is released only when no `std::shared_ptr` **or** `std::weak_ptr` reference it. Many commonly used objects are allocated with std::make_shared, including the SHAMap tree nodes.

As a result, the release of the memory used by many objects is still occurring within the lock in the `TaggedCache` (because it occurs when the `std::weak_ptr` is destructed).

> Note: even prior to this change, the destructor for the shared objects is indeed called outside the lock, since the object destructor is called when the last strong reference (by `shared_ptr`) is destructed. It is only the actual memory release which occurred within the lock when the `std::weak_ptr` is destroyed.

### Context of Change

This is a performance issue. It has very likely been present for a very long time.
I think this change should increase rippled's performance, as the TaggedCache lock will be held for a shorter time.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

This change gathers both pointer types ( `std::shared_ptr` and `std::weak_ptr`) in vectors which are destroyed outside of the lock, ensuring that the memory deallocation occurs outside the lock, for both object memory and control block, regardless of whether they are allocated together or not.

I ran the unittests (no failure), and verified that rippled syncs correctly. In my opinion the change is simple and straightforward enough that it is unlikely to cause any problem.

### DIFF
--- a/src/ripple/basics/TaggedCache.h
+++ b/src/ripple/basics/TaggedCache.h
@@ -727,7 +727,7 @@ private:
         clock_type::time_point const& when_expire,
         clock_type::time_point const& now,
         typename KeyOnlyCacheType::map_type& partition,
-        SweptPointersVector& stuffToSweep,
+        SweptPointersVector&,
         std::atomic<int>& allRemovals)
     {
         return std::thread([&, this]() {
--- a/src/ripple/basics/TaggedCache.h
+++ b/src/ripple/basics/TaggedCache.h
@@ -247,7 +247,8 @@ public:
                     now,
                     m_cache.map()[p],
                     allStuffToSweep[p],
-                    allRemovals));
+                    allRemovals,
+                    lock));
             }
             for (std::thread& worker : workers)
                 worker.join();
@@ -655,7 +656,8 @@ private:
         [[maybe_unused]] clock_type::time_point const& now,
         typename KeyValueCacheType::map_type& partition,
         SweptPointersVector& stuffToSweep,
-        std::atomic<int>& allRemovals)
+        std::atomic<int>& allRemovals,
+        std::lock_guard<std::recursive_mutex> const&)
     {
         return std::thread([&, this]() {
             int cacheRemovals = 0;
@@ -728,7 +730,8 @@ private:
         clock_type::time_point const& now,
         typename KeyOnlyCacheType::map_type& partition,
         SweptPointersVector&,
-        std::atomic<int>& allRemovals)
+        std::atomic<int>& allRemovals,
+        std::lock_guard<std::recursive_mutex> const&)
     {
         return std::thread([&, this]() {
             int cacheRemovals = 0;
--- a/src/ripple/basics/TaggedCache.h
+++ b/src/ripple/basics/TaggedCache.h
@@ -196,14 +196,17 @@ public:
         return true;
     }
 
+    using SweptPointersVector = std::pair<
+        std::vector<std::shared_ptr<mapped_type>>,
+        std::vector<std::weak_ptr<mapped_type>>>;
+
     void
     sweep()
     {
         // Keep references to all the stuff we sweep
         // For performance, each worker thread should exit before the swept data
         // is destroyed but still within the main cache lock.
-        std::vector<std::vector<std::shared_ptr<mapped_type>>> allStuffToSweep(
-            m_cache.partitions());
+        std::vector<SweptPointersVector> allStuffToSweep(m_cache.partitions());
 
         clock_type::time_point const now(m_clock.now());
         clock_type::time_point when_expire;
@@ -244,8 +247,7 @@ public:
                     now,
                     m_cache.map()[p],
                     allStuffToSweep[p],
-                    allRemovals,
-                    lock));
+                    allRemovals));
             }
             for (std::thread& worker : workers)
                 worker.join();
@@ -652,9 +654,8 @@ private:
         clock_type::time_point const& when_expire,
         [[maybe_unused]] clock_type::time_point const& now,
         typename KeyValueCacheType::map_type& partition,
-        std::vector<std::shared_ptr<mapped_type>>& stuffToSweep,
-        std::atomic<int>& allRemovals,
-        std::lock_guard<std::recursive_mutex> const& lock)
+        SweptPointersVector& stuffToSweep,
+        std::atomic<int>& allRemovals)
     {
         return std::thread([&, this]() {
             int cacheRemovals = 0;
@@ -662,7 +663,8 @@ private:
 
             // Keep references to all the stuff we sweep
             // so that we can destroy them outside the lock.
-            stuffToSweep.reserve(partition.size());
+            stuffToSweep.first.reserve(partition.size());
+            stuffToSweep.second.reserve(partition.size());
             {
                 auto cit = partition.begin();
                 while (cit != partition.end())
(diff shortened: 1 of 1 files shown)


greg7mdp:
Yes @nbougalis you are 100% correct, thanks for clarifying, I have updated the PR description.

nbougalis (approved):
Looks good to me.

We'll need to fold these two commits into one and adjust the commit message during merge. You can do it now if you prefer, @greg7mdp.

greg7mdp:
Thanks @nbougalis I merged the commits and updated the message.

mtrippled (approved):
:+1: LGTM

src/ripple/basics/TaggedCache.h
```diff
@@ -722,17 +727,15 @@ class TaggedCache
         clock_type::time_point const& when_expire,
         clock_type::time_point const& now,
         typename KeyOnlyCacheType::map_type& partition,
-        std::vector<std::shared_ptr<mapped_type>>& stuffToSweep,
-        std::atomic<int>& allRemovals,
-        std::lock_guard<std::recursive_mutex> const& lock)
+        SweptPointersVector& stuffToSweep,
```
HowardHinnant, line 730: Consider marking this parameter with `[[maybe_unused]]`.

HowardHinnant (approved):

src/ripple/basics/TaggedCache.h
greg7mdp: Do you mean `stuffToSweep`? Why are you suggesting this? I think it is likely to be used.

greg7mdp: Thanks. Is adding the `[[maybe_unused]]` attribute preferred to just removing the variable name, which is what I usually do?

HowardHinnant: I usually just remove the variable name too.  I noticed `[[maybe_unused]]` was used in the other `sweepHelper` overload, so I reached for it.  But I'm not going to loose a lot of sleep over either choice. :-)

HowardHinnant: Hmm... making guidelines up on the fly, which is never a good idea.  But perhaps remove the variable name if it is definitely never used, and use `[[maybe_unused]]` if it _might_ or _might_not_ be used, such as in an `assert` or under a `ifdef`.

mtrippled:
Is [[maybe_unused]] preferable to boost::ignore_unused()?

HowardHinnant:
Yes, in C++17 and later, because it is in the C++ language.

HowardHinnant:
pre-C++17 I would actually prefer:
```c++
(void)foo;
```
over:
```c++
boost::ignore_unused(foo);
```

The `void` solution is a decades old pattern that is well understood and doesn't depend on a library.  I don't know what problem `boost::ignore_unused` solves.

src/ripple/basics/TaggedCache.h
greg7mdp: That makes sense to me, remove the variable name if there is no reference to it at all, and `[[maybe_unused]]` if it could be under some condition. Thanks Howard!

ximinez (changes requested):
In addition to the suggestions below, there's a typo in the first commit message. "objet" should be "object". I would also suggest rewriting it as "Release TaggedCache object memory outside the lock". (Per https://cbea.ms/git-commit/, leave the period off the end, too.)

src/ripple/basics/TaggedCache.h
```diff
@@ -196,14 +196,17 @@ class TaggedCache
...
+
     void
     sweep()
     {
         // Keep references to all the stuff we sweep
         // For performance, each worker thread should exit before the swept data
         // is destroyed but still within the main cache lock.
-        std::vector<std::vector<std::shared_ptr<mapped_type>>> allStuffToSweep(
-            m_cache.partitions());
+        std::vector<SweptPointersVector> allStuffToSweep(m_cache.partitions());
```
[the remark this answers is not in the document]
greg7mdp: I don't think it would work, because partitions are processed concurrently in different threads, so then need different vectors. I don't believe the `reserve()` in `sweepHelper` is done under a lock that protects from concurrent access from the workers.

src/ripple/basics/TaggedCache.h
```diff
@@ -652,17 +654,17 @@ class TaggedCache
         clock_type::time_point const& when_expire,
         [[maybe_unused]] clock_type::time_point const& now,
         typename KeyValueCacheType::map_type& partition,
-        std::vector<std::shared_ptr<mapped_type>>& stuffToSweep,
-        std::atomic<int>& allRemovals,
-        std::lock_guard<std::recursive_mutex> const& lock)
+        SweptPointersVector& stuffToSweep,
+        std::atomic<int>& allRemovals)
```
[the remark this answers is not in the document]
greg7mdp: Thanks for noticing, @ximinez, I did not realize this was a common pattern. I should put it back unless we decide to abandon the PR (see comment below).

greg7mdp:
@HowardHinnant @nbougalis @ximinez @mtrippled  I think I would like to abandon this PR.
Because of the effort that was made to release the `SHAMAPTreeNodes` outside the lock, I was under the impression that holding the cache lock for the minimum time possible was of the utmost importance, and that is why I proposed this change. It seemed like a good idea, but I did not test whether it the change had any impact on performance. 

However, after submitting the PR, I experimented with replacing the `hardened_partitioned_hash_map` with another hash map of mine which doesn't hold a global lock, but a lock for each partition (basically this [commit]( https://github.com/greg7mdp/rippled/commit/a0a5a511f476f46b4fd33e6215e43360355bf94e) ), and this should have eliminated most locking, but the time for rippled to sync increased slightly from 4 minutes to almost 5 minutes (maybe this is not the right benchmark), so I am not sure anymore that reducing slightly the locking duration is critical.

Please let me know what you guys think. Thanks!

mtrippled:
@greg7mdp I think that this improvement is beneficial, but probably not measurably so. It can't hurt--I always assumed that an expired weak_ptr had already been destroyed, but I guess that's not necessarily the case. I don't see a problem with keeping this, and I've already approved it.

mtrippled:
Also, @greg7mdp you can measure the impact of your change with normal operation by grepping for "sweep lock duration" from the debug log. That measures the duration between the first partition lock being acquired and the last partition lock being freed. There are several caches like this, but the worst one is the "tree node cache". Any decrease in lock duration is a good thing, and this is a way to easily measure any changes.

src/ripple/basics/TaggedCache.h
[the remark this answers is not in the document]
ximinez: Oh *duh*. Of course this would be a terrible idea with multiple threads. Nevermind! 😃

ximinez:
I agree with @mtrippled. I don't think this should be abandoned unless the sweep messages indicate degraded performance.

greg7mdp:
OK, fair enough and thanks @mtrippled and @ximinez , I'll make the requested changes.

src/ripple/basics/TaggedCache.h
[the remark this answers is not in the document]
greg7mdp: @ximinez - I restored the pattern, thanks!

ximinez (approved):

Issue #3214, opened 2020-01-08
Remove deprecated `ledger` fields


mDuo13:
The JSON format for ledger objects (as returned by the [`ledger` method](https://xrpl.org/ledger.html) among others) could probably stand to be better-unified [a la transactions](https://github.com/ripple/rippled/issues/2599) but for starters, we could at least remove some of the redundant fields that have been deprecated for years:

> The following fields are deprecated and may be removed without further notice: `accepted`, `hash` (use `ledger_hash` instead), `seqNum` (use `ledger_index` instead), `totalCoins` (use `total_coins` instead).

Given that we've had that statement up since probably 2015 if not earlier, I think we'd be justified in removing the mentioned fields even without having full "API versioning" in place. Anyone who's still using them at this point is almost certainly doing something highly inadvisable if whatever they're doing even still works with all the amendments to the protocol since then.

chennakeshava1998:
Hello, 
I'm a beginner to XRPL and I have a question. After removal of the deprecated fields, this seems to be the output of the command `./rippled ledger validated` :
```
Loading: "/Users/keshava/.config/ripple/rippled.cfg"
2022-Jul-20 17:39:49.765444 UTC HTTPClient:NFO Connecting to 127.0.0.1:5005

{
   "result" : {
      "ledger" : {
         "account_hash" : "725DB4D108D882CEC03FCCFABDEB969A6DB9626E3F2CF1425E61F394A95C631E",
         "close_flags" : 0,
         "close_time" : 711653980,
         "close_time_human" : "2022-Jul-20 17:39:40.000000 UTC",
         "close_time_resolution" : 10,
         "closed" : true,
         "ledger_hash" : "5C0A53406F1AC9D479ACF8114072AF47ECBA2747E8CB409126379117BF8CDA6D",
         "ledger_index" : "73137147",
         "parent_close_time" : 711653972,
         "parent_hash" : "199689B8B6545D92CB19F8A92386C9E23D7009733A1B7906E08D9411969AE8C1",
         "total_coins" : "99989392350108618",
         "transaction_hash" : "470B3CBAA9E3E88C1F98E85C97C736CC530929B62393D1C2E7B38F63F841EA12"
      },
      "ledger_hash" : "5C0A53406F1AC9D479ACF8114072AF47ECBA2747E8CB409126379117BF8CDA6D",
      "ledger_index" : 73137147,
      "status" : "success",
      "validated" : true
   }
}
```

[...]

PR #4244, opened 2022-07-20
remove deprecated fields in ledger command, fix for #3214

ckeshava:
I removed the following fields from the ledger command - `accepted`, `seq`, `hash` and `totalCoins`.

Some of these variable names (`totalCoins`, `seq`) are used in managing the database but I don't think there is any redundancy in the creation/updation/deletion of the tables.

### DIFF
--- a/src/ripple/app/ledger/impl/LedgerToJson.cpp
+++ b/src/ripple/app/ledger/impl/LedgerToJson.cpp
@@ -53,7 +53,6 @@ fillJson(Object& json, bool closed, LedgerInfo const& info, bool bFull)
 {
     json[jss::parent_hash] = to_string(info.parentHash);
     json[jss::ledger_index] = to_string(info.seq);
-    json[jss::seqNum] = to_string(info.seq);  // DEPRECATED
 
     if (closed)
     {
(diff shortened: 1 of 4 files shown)


nbougalis:
I don't know if we can just remove the deprecated field; someone might still be relying on it. Do we know how long it's been marked DEPRECATED for? @mDuo13, any objections?

mDuo13:
Per #3214, the fields have been marked as deprecated dating back to at least 2015. I think we can just remove them.

intelliot (approved):
I'm fine with just removing these, but let's get at least one more review.

gregtatcam (commented):
It looks ok but `jss::hash` is listed in the `ledger_request` documentation: https://xrpl.org/ledger_request.html.

intelliot:
@mDuo13 any thoughts on `jss::hash`?

mDuo13:
The `hash` field in ledger_request should stay (for now). That command has its own problems, but I don't think it uses the shared ledger deserialization that this ticket is about.

ledhed2222 (approved):

drlongle (approved):

intelliot:
This is slated for inclusion in 1.12 (not 1.11) unless anyone comments here to request otherwise.

PR #4245, opened 2022-07-20
remove deprecated AccountTxOld.cpp - fix #2926

### LINKED ISSUE #2926: Remove long-deprecated account_tx switch
_(Porting this issue from Ripple's internal Jira ticket RIPD-1569)_

## Description

The account_tx method has an "old" mode that's been deprecated since at least 2013, with a ["temporary switch"](https://github.com/ripple/rippled/blob/develop/src/ripple/rpc/handlers/AccountTxSwitch.cpp) to use the old mode if people provide certain parameters (offset, count, descending, ledger_max, or ledger_min) in the request, and the new mode otherwise.

I'd say it's about time to remove the old mode.

## Blocker

Per @scottschurr:

> I made a run at this fix, but there's a complication. The command-line version of the account_tx command relies on the deprecated form of the RPC command. Until we come up with a new version of the command line (and figure out a transition process between the old and new forms) we're stuck with the deprecated version of the RPC command.
> 
> From a development perspective, the simplest possible change to the command line form would be to simply remove it and tell folks they must use the json command-line command to do account_tx queries. This approach requires no transition process and requires no new code.
> 
> I'm unsure just how user hostile that approach would be, however.
> 
> I welcome other suggestions.

chennakeshava1998:
If merged, this PR will:
- delete the deprecated file `AccountTxOld.cpp`
- Update the help message in `Main.cpp` to remove deprecated fields.
- The unit tests are running correctly. Manual RPC calls to `account_tx` command are working as expected.


chennakeshava1998:
I have a question. This file `src/ripple/protocol/jss.h` has certain references to `AccountTxOld` and `offset` in its comments. Shall I remove them as well?

PR #4246, opened 2022-07-21
Work around soci build issues

ximinez:

Soci's CMake setup leaves flags in place that will cause warnings to
be treated as errors, but some compiler versions throw "new" warnings
that then cause the build to fail. Simplify that until soci fixes
those warnings.

Note that this PR will probably become obsolete if / when #4223 is merged.

### Context of Change

This only affects the build process for the soci dependency.

### Type of Change

## Before / After

Before this change, the soci cmake file would need to be patched manually before being able to successfully build with an affected compiler (clang 14, maybe others). After, the build should proceed seamlessly.

### DIFF
--- a/Builds/CMake/SociConfig.cmake.patched
+++ b/Builds/CMake/SociConfig.cmake.patched
@@ -1,0 +1,106 @@
+################################################################################
+# SociConfig.cmake - CMake build configuration of SOCI library
+################################################################################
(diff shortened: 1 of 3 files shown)


ximinez:
Requesting a review from @seelabs, who I can't assign, because he came up with the original patch.

thejohnfreeman (approved):

PR #4247, opened 2022-07-21
`XRPFees`: Fee setting and handling improvements

### LINKED ISSUE #3932: Simplify fee settings & related calculations
The fee code is unnecessarily complex, involving arcane concepts like "fee units" and "reference transactions", some of which are adjustable and some of which are not. These values are represented as integers of various widths and a lot of effort is expended to keep all that straight and to help developers avoid doing the wrong thing.

Since the fee units are hard-coded to 10 and the "reference transaction" fee is also 10 and dividing the two gives the identity, there's an opportunity to simply, quickly and easily remove all that complexity, eliminate the whole concept of "fee units" and move to something _sane_: `XRPAmount` for all fees.

All scaling/adjusting of fees would operate directly on `XRPAmount` objects (which are scalars) instead.

ximinez:

This branch resolves #3932. The concept of "fee units" has been (almost) completely removed. It remains in a few places to preserve backward compatibility. Additionally, a new amendment, `XRPFees` has been introduced which, if enabled by the validators, will convert all representations of fees in the code, ledger, and protocol that are still raw numbers into `XRPAmount`s.

Additionally, it fixes an issue where a sidechain / test network is able to confirm a 0 drop base fee, but is unable to accept submissions of transactions with 0 drop fees. The transaction engine appears to be perfectly fine with such fees, but fee escalation logic had an unstated assumption that the base fee would never be 0. This fix works around that assumption while still enabling fee escalation to raise required fees when the network gets busy - it effectively treats the base fee as 1 drop for purposes of getting into the open ledger *ONLY* when fees are escalated. No amendment is required for this fix because it only affects transaction submission, not processing. I am reasonably sure this solution is complete, but I would encourage anybody dealing with the issue to test this branch to ensure I haven't missed any edge cases.

### Type of Change

### DIFF
--- a/src/ripple/app/ledger/Ledger.cpp
+++ b/src/ripple/app/ledger/Ledger.cpp
@@ -591,29 +591,9 @@ Ledger::setup(Config const& config)
 {
     bool ret = true;
 
-    fees_.base = config.FEE_DEFAULT;
-    fees_.units = config.TRANSACTION_FEE_BASE;
-    fees_.reserve = config.FEE_ACCOUNT_RESERVE;
-    fees_.increment = config.FEE_OWNER_RESERVE;
-
     try
     {
-        if (auto const sle = read(keylet::fees()))
-        {
-            // VFALCO NOTE Why getFieldIndex and not isFieldPresent?
-
-            if (sle->getFieldIndex(sfBaseFee) != -1)
-                fees_.base = sle->getFieldU64(sfBaseFee);
-
-            if (sle->getFieldIndex(sfReferenceFeeUnits) != -1)
-                fees_.units = sle->getFieldU32(sfReferenceFeeUnits);
-
-            if (sle->getFieldIndex(sfReserveBase) != -1)
-                fees_.reserve = sle->getFieldU32(sfReserveBase);
-
-            if (sle->getFieldIndex(sfReserveIncrement) != -1)
-                fees_.increment = sle->getFieldU32(sfReserveIncrement);
-        }
+        rules_ = makeRulesGivenLedger(*this, config.features);
     }
     catch (SHAMapMissingNode const&)
     {
@@ -624,9 +604,56 @@ Ledger::setup(Config const& config)
         Rethrow();
     }
 
+    fees_.base = config.FEE_DEFAULT;
+    fees_.reserve = config.FEE_ACCOUNT_RESERVE;
+    fees_.increment = config.FEE_OWNER_RESERVE;
+
     try
     {
-        rules_ = makeRulesGivenLedger(*this, config.features);
+        if (auto const sle = read(keylet::fees()))
+        {
+            bool oldFees = false;
+            bool newFees = false;
+            {
+                auto const baseFee = sle->at(~sfBaseFee);
+                auto const reserveBase = sle->at(~sfReserveBase);
+                auto const reserveIncrement = sle->at(~sfReserveIncrement);
+                if (baseFee)
+                    fees_.base = *baseFee;
+                if (reserveBase)
+                    fees_.reserve = *reserveBase;
+                if (reserveIncrement)
+                    fees_.increment = *reserveIncrement;
+                oldFees = baseFee || reserveBase || reserveIncrement;
+            }
+            {
+                auto const baseFeeXRP = sle->at(~sfBaseFeeXRP);
+                auto const reserveBaseXRP = sle->at(~sfReserveBaseXRP);
+                auto const reserveIncrementXRP =
+                    sle->at(~sfReserveIncrementXRP);
+                auto assign = [&ret](
+                                  XRPAmount& dest,
+                                  std::optional<STAmount> const& src) {
+                    if (src)
+                    {
+                        if (src->native())
+                            dest = src->xrp();
+                        else
+                            ret = false;
+                    }
+                };
+                assign(fees_.base, baseFeeXRP);
+                assign(fees_.reserve, reserveBaseXRP);
+                assign(fees_.increment, reserveIncrementXRP);
+                newFees = baseFeeXRP || reserveBaseXRP || reserveIncrementXRP;
+            }
+            if (oldFees && newFees)
+                // Should be all of one or the other, but not both
+                ret = false;
+            if (!rules_.enabled(featureXRPFees) && newFees)
+                // Can't populate the new fees before the amendment is enabled
+                ret = false;
+        }
     }
     catch (SHAMapMissingNode const&)
     {
--- a/src/ripple/app/ledger/Ledger.cpp
+++ b/src/ripple/app/ledger/Ledger.cpp
@@ -627,10 +627,10 @@ Ledger::setup(Config const& config)
                 oldFees = baseFee || reserveBase || reserveIncrement;
             }
             {
-                auto const baseFeeXRP = sle->at(~sfBaseFeeXRP);
-                auto const reserveBaseXRP = sle->at(~sfReserveBaseXRP);
+                auto const baseFeeXRP = sle->at(~sfBaseFeeDrops);
+                auto const reserveBaseXRP = sle->at(~sfReserveBaseDrops);
                 auto const reserveIncrementXRP =
-                    sle->at(~sfReserveIncrementXRP);
+                    sle->at(~sfReserveIncrementDrops);
                 auto assign = [&ret](
                                   XRPAmount& dest,
                                   std::optional<STAmount> const& src) {
--- a/src/ripple/app/misc/FeeVote.h
+++ b/src/ripple/app/misc/FeeVote.h
@@ -42,9 +42,6 @@ public:
         /** The cost of a reference transaction in drops. */
         XRPAmount reference_fee{10};
 
-        /** The cost of a reference transaction in fee units. */
-        static constexpr FeeUnit32 reference_fee_units{10};
-
         /** The account reserve requirement in drops. */
         XRPAmount account_reserve{10 * DROPS_PER_XRP};
 
@@ -60,7 +57,10 @@ public:
         @param baseValidation
     */
     virtual void
-    doValidation(Fees const& lastFees, STValidation& val) = 0;
+    doValidation(
+        Fees const& lastFees,
+        Rules const& rules,
+        STValidation& val) = 0;
 
     /** Cast our local vote on the fee.
 
--- a/src/ripple/app/misc/FeeVoteImpl.cpp
+++ b/src/ripple/app/misc/FeeVoteImpl.cpp
@@ -33,44 +33,70 @@ class VotableValue
 {
 private:
     using value_type = XRPAmount;
-    value_type const mCurrent;  // The current setting
-    value_type const mTarget;   // The setting we want
-    std::map<value_type, int> mVoteMap;
+    value_type const current_;  // The current setting
+    value_type const target_;   // The setting we want
+    std::map<value_type, int> voteMap_;
+    std::optional<value_type> vote_;
 
 public:
     VotableValue(value_type current, value_type target)
-        : mCurrent(current), mTarget(target)
+        : current_(current), target_(target)
     {
         // Add our vote
-        ++mVoteMap[mTarget];
+        ++voteMap_[target_];
     }
 
     void
     addVote(value_type vote)
     {
-        ++mVoteMap[vote];
+        ++voteMap_[vote];
     }
 
     void
     noVote()
     {
-        addVote(mCurrent);
+        addVote(current_);
     }
+
+    void
+    setVotes();
 
     value_type
     getVotes() const;
+
+    template <class Dest>
+    Dest
+    getVotesAs() const;
+
+    bool
+    voteChange() const
+    {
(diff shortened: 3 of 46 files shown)


src/ripple/protocol/SystemParameters.h
```diff
@@ -46,7 +46,7 @@ constexpr XRPAmount INITIAL_XRP{100'000'000'000 * DROPS_PER_XRP};
 inline bool
 isLegalAmount(XRPAmount const& amount)
 {
-    return amount <= INITIAL_XRP;
+    return amount >= -INITIAL_XRP && amount <= INITIAL_XRP;
```
nbougalis, line 49: Why? This may have implications in other parts of the code.

src/ripple/protocol/impl/STValidation.cpp
```diff
@@ -58,9 +58,14 @@ STValidation::validationFormat()
         {sfSigningPubKey,       soeREQUIRED},
         {sfSignature,           soeREQUIRED},
         {sfConsensusHash,       soeOPTIONAL},
+        // featureHardenedValidations
         {sfCookie,              soeDEFAULT},
         {sfValidatedHash,       soeOPTIONAL},
         {sfServerVersion,       soeOPTIONAL},
+        // featureXRPFees
+        {sfBaseFeeXRP,          soeOPTIONAL},
```
nbougalis, line 66: Should these fields be, more accurately, named `sf[Whatever]Drops` instead? I don't mind it, as is, but I am curious what others think.

ximinez: I'd be fine with that. I just wanted to distinguish from the existing `sfBaseFee`, etc.

My thinking was that the data type of the field is `STAmount`, which has an `xrp()` function to return an `XRPAmount`, so for consistency "XRP" in the `SField` name makes sense.

src/ripple/protocol/SystemParameters.h
ximinez: Yeah, on second thought, this may be the wrong way / place to implement this. The reason is that somewhere in the development process, I ran into something where a negative `XRPAmount` was possible, and could have caused trouble. In other words, I can't imagine negative one-gazillion XRP would be any better than positive one-gazillion XRP. But I'll reexamine this.

src/ripple/app/misc/FeeVoteImpl.cpp
```diff
@@ -121,26 +158,33 @@ FeeVoteImpl::doValidation(Fees const& lastFees, STValidation& v)
         JLOG(journal_.info())
             << "Voting for base fee of " << target_.reference_fee;
 
-        if (auto const f = target_.reference_fee.dropsAs<std::uint64_t>())
```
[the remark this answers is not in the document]
ximinez: I was trying to avoid code duplication, but yes, it's probably clearer. I can change this.

src/ripple/app/misc/FeeVoteImpl.cpp
```diff
@@ -33,44 +33,70 @@ class VotableValue
...
+        return getVotes() != current_;
+    }
 };
 
+void
+VotableValue::setVotes()
+{
+    // Need to explicitly remove any value from vote_, because it will be
+    // returned from getVotes() if set.
+    vote_.reset();
```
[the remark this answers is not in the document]
ximinez: The idea was that I wanted to cache the result of the voting so I wouldn't have to duplicate the work of `getVotes()` repeatedly and to add the `voteChange()` function. However, I just got a much better and simpler idea - I'll return a `pair<XRPAccount, bool>`, where the `bool` indicates whether the value changed, from `getVotes()`, and remove all the new functions.

src/ripple/app/misc/FeeVote.h
```diff
@@ -42,9 +42,6 @@ class FeeVote
         /** The cost of a reference transaction in drops. */
         XRPAmount reference_fee{10};
 
-        /** The cost of a reference transaction in fee units. */
-        static constexpr FeeUnit32 reference_fee_units{10};
-
         /** The account reserve requirement in drops. */
         XRPAmount account_reserve{20 * DROPS_PER_XRP};
```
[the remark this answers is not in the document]
ximinez: I'll try it and see what happens. I think the `Env` setup specifies fees to avoid this exact problem.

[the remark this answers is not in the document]
ximinez: Done, and updated a couple of tests to be future-proof.

src/ripple/app/misc/FeeVoteImpl.cpp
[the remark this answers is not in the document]
ximinez: Fixed.

src/ripple/app/misc/FeeVoteImpl.cpp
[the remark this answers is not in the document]
ximinez: Done.

src/ripple/protocol/SystemParameters.h
ximinez: Fixed.

src/ripple/protocol/impl/STValidation.cpp
ximinez: Done, but in a separate commit in case we decide that `sf[Whatever]XRP` is better.

src/ripple/app/ledger/Ledger.cpp
```diff
@@ -591,29 +591,9 @@ Ledger::setup(Config const& config)
 {
```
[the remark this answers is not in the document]
ximinez: See #4319

mDuo13:
Does this also fix the fact that the server (in some places?) serializes `BaseFee` in hexadecimal for some weird reason?

ximinez:
> Does this also fix the fact that the server (in some places?) serializes `BaseFee` in hexadecimal for some weird reason?

I'm not sure. Can you point me to some places where you see it?

intelliot:
(@mDuo13 whenever you have a chance - there's an open question above)

mDuo13:
> > Does this also fix the fact that the server (in some places?) serializes `BaseFee` in hexadecimal for some weird reason?
> 
> I'm not sure. Can you point me to some places where you see it?

#3165

[ledger_entry for the FeeSettings singleton object](https://xrpl.org/websocket-api-tool.html?server=wss%3A%2F%2Fs1.ripple.com%2F&req=%7B%22command%22%3A%22ledger_entry%22%2C%22index%22%3A%224BC50C9B0D8515D3EAAE1E74B29A95804346C491EE1A95BF25E4AAB854A6A651%22%2C%22ledger_index%22%3A%22validated%22%7D)

ximinez:
```
$ rippled -q feature XRPFees

{
   "result" : {
      "93E516234E35E08CA689FA33A6D38E103881F8DCB53023F728C307AA89D515A7" : {
         "enabled" : true,
         "name" : "XRPFees",
         "supported" : true,
         "vetoed" : false
      },
      "status" : "success"
   }
}

$ rippled -q ledger_entry 4BC50C9B0D8515D3EAAE1E74B29A95804346C491EE1A95BF25E4AAB854A6A651

{
   "result" : {
      "index" : "4BC50C9B0D8515D3EAAE1E74B29A95804346C491EE1A95BF25E4AAB854A6A651",
      "ledger_current_index" : 3,
      "node" : {
         "BaseFeeDrops" : "10",
         "Flags" : 0,
         "LedgerEntryType" : "FeeSettings",
         "ReserveBaseDrops" : "10000000",
         "ReserveIncrementDrops" : "2000000",
         "index" : "4BC50C9B0D8515D3EAAE1E74B29A95804346C491EE1A95BF25E4AAB854A6A651"
      },
      "status" : "success",
      "validated" : false
   }
}
```
Note that I ran this test using my `ledgerinit` branch, which is based off of this one, but creates the fee object on genesis, and just made it simpler for me.

mDuo13 (approved):

intelliot:
@ximinez when you have a chance, could you resolve the merge conflicts? no need to force-push; we'll squash this PR to one commit when merging.

ximinez:
> @ximinez when you have a chance, could you resolve the merge conflicts?

Done.

ximinez:
@intelliot I updated the new `SField` values to avoid conflicts with the pending Hooks PR #4225 .

intelliot:
This needs review. @xzhaous @nbougalis @RichardAH ?

intelliot:
@thejohnfreeman @HowardHinnant are reviewing and will update here when they have feedback/questions

HowardHinnant (approved):

src/ripple/app/misc/FeeVoteImpl.cpp
```diff
@@ -111,36 +118,78 @@ FeeVoteImpl::FeeVoteImpl(Setup const& setup, beast::Journal journal)
...
+        auto to32 = [](XRPAmount target) {
+            return target.dropsAs<std::uint32_t>();
+        };
+        auto to64 = [](XRPAmount target) {
+            return target.dropsAs<std::uint64_t>();
+        };
+        auto vote = [&v, this](
+                        auto const current,
+                        XRPAmount target,
+                        auto convertCallback,
```
drlongle, line 167: Nitpick. Should we pass by reference to avoid a copy?

drlongle (approved):

intelliot:
@thejohnfreeman any last thoughts on this PR?

thejohnfreeman (commented):
Should we remove the section "Fee level" in `src/ripple/app/misc/FeeEscalation.md` and amend the rest of the documentation?

ximinez:
> Should we remove the section "Fee level" in `src/ripple/app/misc/FeeEscalation.md` and amend the rest of the documentation?

@thejohnfreeman I don't think so. Fee levels are not the same as fee units. I removed fee units here. Fee levels are still there and necessary for the transaction queue to work fairly. I'm hoping that after this is merged, there will be less confusion about the whole thing.

As for the rest of the documentation, I don't think "fee units" are really documented anywhere because the were a transparent implementation detail, but if there are, then yeah, that would need to be updated.

src/ripple/app/tx/impl/Change.cpp
```diff
@@ -90,8 +90,42 @@ Change::preclaim(PreclaimContext const& ctx)
...
+                    ctx.tx.isFieldPresent(sfReferenceFeeUnits) ||
+                    ctx.tx.isFieldPresent(sfReserveBase) ||
+                    ctx.tx.isFieldPresent(sfReserveIncrement))
+                    return temMALFORMED;
+            }
+            else
+            {
+                // The transaction format formerly enforced these fields as
+                // required. With featureXRPFees, those fields were made
+                // optional with the expectation that they won't be used once
```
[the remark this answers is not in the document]
ximinez: I may need to clarify the comment, but "optional" refers to the transaction format, changed here: https://github.com/XRPLF/rippled/pull/4247/files#diff-3e7bd959490f33226dff70ce2bfc3e957f2126a35476b2f5048a0e0328f2e77bL158-R162

At the object level, all of the fields are defined as optional, because that's the only real way I know of to have backward compatibility.

So, yes, they are forbidden in the branch above, if the amendment is enabled. In this block, they're actually required, so the code needs to check that they're all there.

src/ripple/app/tx/impl/Change.cpp
```diff
@@ -93,12 +93,16 @@ Change::preclaim(PreclaimContext const& ctx)
         case ttFEE:
             if (ctx.view.rules().enabled(featureXRPFees))
             {
+                // The ttFEE transaction format defines these fields as
+                // optional, but once the XRPFees feature is enabled, they are
+                // forbidden.
```
thejohnfreeman, line 98: These are the required ones, right? Not forbidden? The code returns malformed if they are missing.

src/ripple/app/tx/impl/Change.cpp
```diff
@@ -93,12 +93,16 @@ Change::preclaim(PreclaimContext const& ctx)
...
                 if (!ctx.tx.isFieldPresent(sfBaseFeeDrops) ||
                     !ctx.tx.isFieldPresent(sfReserveBaseDrops) ||
                     !ctx.tx.isFieldPresent(sfReserveIncrementDrops))
                     return temMALFORMED;
-                // The transaction should only have one set of fields or the
-                // other.
+                // The ttFEE transaction format defines these fields as
+                // optional, but once the XRPFees feature is enabled, they are
+                // required.
```
thejohnfreeman, line 105: Similarly, these are the forbidden ones.

thejohnfreeman (changes requested):

src/ripple/app/tx/impl/Change.cpp
ximinez: Yeah, woops. Got those backwards.

src/ripple/app/tx/impl/Change.cpp
ximinez: Yeah, woops. Got those backwards.

thejohnfreeman (approved):

PR #4254, opened 2022-07-26
Properly handle self-assignment of PublicKey

a-noni-mousse:
The copy assignment function of the `PublicKey` behaves incorrectly if it does something `pk = pk;`


greg7mdp (approved):
Nice catch!

ximinez (approved):
Looks good!

PR #4259, opened 2022-07-29
Implement fee scaling based on the size of the memos

a-noni-mousse:
A revision of #3007 to  scale the fees for transactions with many memos which is csauing a spamming problem. It may be better to just change the coe to change by the size of the transaction instead which works for other expensive things too likea lot of paths.

### DIFF
--- a/src/ripple/protocol/impl/Feature.cpp
+++ b/src/ripple/protocol/impl/Feature.cpp
@@ -442,6 +442,7 @@ REGISTER_FEATURE(ExpandedSignerList,            Supported::yes, DefaultVote::no)
 REGISTER_FIX    (fixNFTokenDirV1,               Supported::yes, DefaultVote::no);
 REGISTER_FIX    (fixNFTokenNegOffer,            Supported::yes, DefaultVote::no);
 REGISTER_FEATURE(NonFungibleTokensV1_1,         Supported::yes, DefaultVote::no);
+REGISTER_FEATURE(MemoFeeScaling,                Supported::yes, DefaultVote::yes);
 
 // The following amendments have been active for at least two years. Their
 // pre-amendment code has been removed and the identifiers are deprecated.
--- a/src/ripple/app/tx/impl/Transactor.cpp
+++ b/src/ripple/app/tx/impl/Transactor.cpp
@@ -152,7 +152,17 @@ Transactor::calculateBaseFee(ReadView const& view, STTx const& tx)
     std::size_t const signerCount =
         tx.isFieldPresent(sfSigners) ? tx.getFieldArray(sfSigners).size() : 0;
 
-    return baseFee + (signerCount * baseFee);
+    std::uint32_t memoCount = 0;
+
+    if (tx.isFieldPresent(sfMemos) &&
+        view.rules().enabled(featureMemoFeeScaling))
+    {
+        Serializer s(2048);
+        tx.getFieldArray(sfMemos).add(s);
+        memoCount = 1 + static_cast<std::uint32_t>(s.size() / 32);
+    }
+
+    return baseFee * (1 + signerCount + memoCount);
 }
 
 XRPAmount
--- a/src/ripple/protocol/Feature.h
+++ b/src/ripple/protocol/Feature.h
@@ -74,7 +74,7 @@ namespace detail {
 // Feature.cpp. Because it's only used to reserve storage, and determine how
 // large to make the FeatureBitset, it MAY be larger. It MUST NOT be less than
 // the actual number of amendments. A LogicError on startup will verify this.
-static constexpr std::size_t numFeatures = 50;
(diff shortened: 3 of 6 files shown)


src/ripple/protocol/impl/Feature.cpp
```diff
@@ -442,6 +442,7 @@ REGISTER_FEATURE(ExpandedSignerList,            Supported::yes, DefaultVote::no)
 REGISTER_FIX    (fixNFTokenDirV1,               Supported::yes, DefaultVote::no);
 REGISTER_FIX    (fixNFTokenNegOffer,            Supported::yes, DefaultVote::no);
 REGISTER_FEATURE(NonFungibleTokensV1_1,         Supported::yes, DefaultVote::no);
+REGISTER_FEATURE(MemoFeeScaling,                Supported::yes, DefaultVote::yes);
```
wojake, line 445: Why do you think this amendment's defaultVote should be set as `Yes`?

wojake: I don't think the defaultVote value should be set as `Yes` due to the economical changes that the amendment introduces to performing & implementing business models on-chain (that requires the use of the memos field). With the introduction of the Hooks amendment, federated sidechains and L2 apps (codius/evernode), this would incur additional cost to participants as these concepts are utilized at scale.

nbougalis: I agree that the amendment should be "default no" and I think that that's probably the direction for almost all new feature amendments. Certainly the fact that adoption of this amendment would require libraries etc to be updated suggests that "no" is the safer option.

But I don't think that the argument that there are things that use memos is an argument _against_ increasing the fees for transactions with memos. The fact is that memos impose a load on the network: they increase the size of the transactions and increase the amount of storage that full history servers have to carry. It's certainly not unreasonable to charge a fee for that increased load.

lathanbritz:
I would like to see some sort of Flag, or something that gets added to the response above and beyond simply bumping the fee. How does a legitimate dev know they may have tripped up on this? 

There needs to be at least something say hey, you'r causing the fee's to rise.

src/ripple/protocol/impl/Feature.cpp
wojake: I'm not against the amendment: In fact, I want it to pass voting, considering the load that the XRPL will experience in the future. This change will affect on-chain businesses/apps that actively use memos, this should be manually voted by validators. 💯 Agree with your statement.

nbougalis:
> I would like to see some sort of Flag, or something that gets added to the response above and beyond simply bumping the fee. How does a legitimate dev know they may have tripped up on this?
> 
> There needs to be at least something say hey, you'r causing the fee's to rise.

I'm not quite sure I follow? The fee change here isn't dynamic based on load: that is, it increases the amount of the fee purely based on the size of the memo, so a developer would simply check if the corresponding amendment is activated and, if so, calculate the fee using the algorithm implemented in this pull requests.

lathanbritz:
> I'm not quite sure I follow? The fee change here isn't dynamic based on load: that is, it increases the amount of the fee purely based on the size of the memo, so a developer would simply check if the corresponding amendment is activated and, if so, calculate the fee using the algorithm implemented in this pull requests.

Apologies, I took for granted that there was a point at which it kicked in (exceeds limit). Fee is just dynamic end of story 👍🏻

MarkusTeufelberger:
I'm not certain this requires an amendment? It would make coordination/communication easier, but in the end transactions with higher fees are still valid to every node that doesn't run this code and if enough validators run this code, transactions with lower fees than required by the code would just get rejected.

nbougalis:
> I'm not certain this requires an amendment? It would make coordination/communication easier, but in the end transactions with higher fees are still valid to every node that doesn't run this code and if enough validators run this code, transactions with lower fees than required by the code would just get rejected.

As implemented, it does: it changes the code that calculates the _minimum_ fee.

A non-amendment based implementation is possible, where servers could require higher local fees, but would accept transactions with lower fees if the network reached consensus about including such a transaction.

Personally, I'm fine with the amendment-based approach, because it makes it easier for everyone to calculate the minimum required fee, although it's hardly an "easy" thing to do, especially when faced with queue congestion/levels.

intelliot:
@a-noni-mousse what are your latest thoughts on this change?

intelliot:
@a-noni-mousse can you resolve the merge conflicts?

Otherwise, this PR may be closed due to inactivity. You can always open a new PR in the future, of course.

intelliot:
@a-noni-mousse if merge conflicts are not resolved, this PR will be closed. But don't lose heart - you can always re-open it or open a new PR with the changes in the future :)

intelliot:
Closing due to inactivity. Feel free to open a new PR with this change (rebased on the head of `develop`).

PR #4261, opened 2022-08-01
Avoid unnecessary copying and dynamic memory allocations

nbougalis:
This commit implements an optimization that @chennakeshava1998 and I identified; this commit reduces unnecessary dynamic memory allocations when processing proposals.

### DIFF
--- a/src/ripple/app/consensus/RCLCxPeerPos.h
+++ b/src/ripple/app/consensus/RCLCxPeerPos.h
@@ -28,6 +28,7 @@
 #include <ripple/protocol/HashPrefix.h>
 #include <ripple/protocol/PublicKey.h>
 #include <ripple/protocol/SecretKey.h>
+#include <boost/container/static_vector.hpp>
 #include <chrono>
 #include <cstdint>
 #include <string>
@@ -61,10 +62,6 @@ public:
         uint256 const& suppress,
         Proposal&& proposal);
 
-    //! Create the signing hash for the proposal
-    uint256
-    signingHash() const;
-
     //! Verify the signing hash of the proposal
     bool
     checkSign() const;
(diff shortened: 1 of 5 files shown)


src/ripple/protocol/impl/PublicKey.cpp
```diff
@@ -221,6 +221,8 @@ verifyDigest(
     Slice const& sig,
     bool mustBeFullyCanonical) noexcept
 {
+    if (sig.size() == 0)
+        return false;
```
greg7mdp, line 225: I don't think this added check is needed as this is taken care of by `ecdsaCanonicality` below.

greg7mdp (changes requested):

src/ripple/app/consensus/RCLCxPeerPos.h
```diff
@@ -73,49 +70,38 @@ class RCLCxPeerPos
...
-            Slice const& signature,
-            uint256 const& suppress,
-            Proposal&& proposal);
-    };
-
-    std::shared_ptr<Data> data_;
+    PublicKey publicKey_;
+    uint256 suppression_;
+    Proposal proposal_;
+    boost::container::static_vector<std::uint8_t, 72> signature_;
```
chennakeshava1998, line 104: I have a hypothetical question. If the size of the `signature_` variable was a multiple of 32 bytes, then
can we use a `ripple::base_uint<33>` for storing the `signature_` data? Will that be slower than `boost::static_vector`?

Is one of them faster than the other?

greg7mdp: `ripple::base_uint<33>` would not work as it specifies 33 bits. You need 72 bytes or 72 x 8 bits, and it is designed to store uint32_t not an array of bytes. Also `base_uint` doesn't manage a size, so you'd have to keep track separately of how long the signature is. `static_vector` is the right choice here!

chennakeshava1998: I see okay, thanks for the clarification.

src/ripple/protocol/impl/PublicKey.cpp
nbougalis: Agreed, I'll remove it.

HowardHinnant (approved):
Nice job.

PR #4267, opened 2022-08-04
refactor NegativeUNL class - use PublicKey instead of NodeID

ckeshava:
Summary of changes:

- refactor NegativeUNL class to use PublicKey instead of NodeID 
- modify the constructors of STValidation to include masterPubKey_ 
- major changes to unit tests of NegativeUNL and STValidation

- [ x] Tests (You added tests for code that already exists, or your new feature included in this PR)

### DIFF
--- a/src/ripple/app/misc/NegativeUNLVote.cpp
+++ b/src/ripple/app/misc/NegativeUNLVote.cpp
@@ -23,8 +23,8 @@
 
 namespace ripple {
 
-NegativeUNLVote::NegativeUNLVote(NodeID const& myId, beast::Journal j)
-    : myId_(myId), j_(j)
+NegativeUNLVote::NegativeUNLVote(PublicKey const& myValPubKey, beast::Journal j)
+    : myValPubKey_(myValPubKey), j_(j)
 {
 }
 
@@ -41,20 +41,9 @@ NegativeUNLVote::doVoting(
     // -- pick one to disable and one to re-enable if any
     // -- if found candidates, add ttUNL_MODIFY Tx
 
-    // Build NodeID set for internal use.
-    // Build NodeID to PublicKey map for lookup before creating ttUNL_MODIFY Tx.
-    hash_set<NodeID> unlNodeIDs;
-    hash_map<NodeID, PublicKey> nidToKeyMap;
-    for (auto const& k : unlKeys)
-    {
-        auto nid = calcNodeID(k);
-        nidToKeyMap.emplace(nid, k);
-        unlNodeIDs.emplace(nid);
-    }
-
     // Build a reliability score table of validators
-    if (std::optional<hash_map<NodeID, std::uint32_t>> scoreTable =
-            buildScoreTable(prevLedger, unlNodeIDs, validations))
+    if (std::optional<hash_map<PublicKey, std::uint32_t>> scoreTable =
+            buildScoreTable(prevLedger, unlKeys, validations))
     {
         // build next negUnl
         auto negUnlKeys = prevLedger->negativeUNL();
@@ -65,39 +54,28 @@ NegativeUNLVote::doVoting(
         if (negUnlToReEnable)
             negUnlKeys.erase(*negUnlToReEnable);
 
-        hash_set<NodeID> negUnlNodeIDs;
-        for (auto const& k : negUnlKeys)
-        {
-            auto nid = calcNodeID(k);
-            negUnlNodeIDs.emplace(nid);
-            if (!nidToKeyMap.count(nid))
-            {
-                nidToKeyMap.emplace(nid, k);
-            }
-        }
-
         auto const seq = prevLedger->info().seq + 1;
         purgeNewValidators(seq);
 
         // Process the table and find all candidates to disable or to re-enable
         auto const candidates =
-            findAllCandidates(unlNodeIDs, negUnlNodeIDs, *scoreTable);
+            findAllCandidates(unlKeys, negUnlKeys, *scoreTable);
 
         // Pick one to disable and one to re-enable if any, add ttUNL_MODIFY Tx
         if (!candidates.toDisableCandidates.empty())
         {
             auto n =
                 choose(prevLedger->info().hash, candidates.toDisableCandidates);
-            assert(nidToKeyMap.count(n));
-            addTx(seq, nidToKeyMap[n], ToDisable, initialSet);
+            assert(unlKeys.count(n));
+            addTx(seq, n, ToDisable, initialSet);
         }
 
         if (!candidates.toReEnableCandidates.empty())
         {
             auto n = choose(
                 prevLedger->info().hash, candidates.toReEnableCandidates);
-            assert(nidToKeyMap.count(n));
-            addTx(seq, nidToKeyMap[n], ToReEnable, initialSet);
+            assert(negUnlKeys.count(n));
+            addTx(seq, n, ToReEnable, initialSet);
         }
     }
 }
@@ -114,7 +92,6 @@ NegativeUNLVote::addTx(
         obj.setFieldU32(sfLedgerSequence, seq);
         obj.setFieldVL(sfUNLModifyValidator, vp.slice());
     });
-
     uint256 txID = negUnlTx.getTransactionID();
     Serializer s;
     negUnlTx.add(s);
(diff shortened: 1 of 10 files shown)


src/ripple/app/misc/NegativeUNLVote.cpp
```diff
@@ -135,29 +112,28 @@ NegativeUNLVote::addTx(
...
 {
     assert(!candidates.empty());
-    static_assert(NodeID::bytes <= uint256::bytes);
-    NodeID randomPad = NodeID::fromVoid(randomPadData.data());
-    NodeID txNodeID = candidates[0];
+    PublicKey chosenValPubKey = candidates[0];
     for (int j = 1; j < candidates.size(); ++j)
     {
-        if ((candidates[j] ^ randomPad) < (txNodeID ^ randomPad))
+        if ((*candidates[j].data() ^ *randomPadData.data()) <
```
chennakeshava1998, line 124: @nbougalis  Is this a good way to compare the candidate validators? I'm not sure if `*candidates[j].data()` is the best way to extract the data in the public key.

src/ripple/protocol/PublicKey.h
```diff
@@ -129,12 +129,9 @@ class PublicKey
     {
         return slice();
     }
-};
 
-/** Print the public key to a stream.
- */
-std::ostream&
-operator<<(std::ostream& os, PublicKey const& pk);
+    friend std::string to_string(PublicKey const&);
```
chennakeshava1998, line 133: @nbougalis  I added this additional method. Is this okay?

chennakeshava1998:
I executed this command `git-clang-format --extensions c,cpp,h,cxx,ipp develop` but it did not modify any files. Any idea on why the CI checks are failing?

wojake:
What's the benefit of this?

nbougalis:
> What's the benefit of this?

This is a WIP and, once complete, will improve the code is several ways, including by making sure the implementation of the `NegativeUNL` feature more faithfully aligns with the specification, eliminating the need for converting public keys to node IDs, simplifying the code that calculate the set of changes to the negative UNL and more, and pave the way to simplifying the `SecretKey` and `PublicKey` classes, by eliminating their default constructors, which will help simplify other parts of the code.

It will require an amendment, when it's ready.

src/ripple/app/misc/NegativeUNLVote.cpp
cryptobrad, line 124: This is very wrong. You XOR the pointers so this will be different only every mahcine. Also if you change the NodeID to PublicKey will require amendment because it changes the selection logic for the candidate.

chennakeshava1998: 1. I'm dereferencing the pointers, so I'm using the data contained within these pointers. I'm not using the actual pointer themselves.
2. Yes, this code change will require an amendment.

intelliot:
@ckeshava do you still think this change is worth making?

ckeshava:
Hello,
Yes, this change would reduce the memory usage of the rippled servers. It's been quite a while since I looked at the codebase, so some refactoring might be required.

ckeshava:
thanks for the comments @drlongle  It's been a while since I made these changes, so I have lost context. I'll take a look at your suggestions and make the necessary changes.

intelliot:
note: this simplifies parts of the NegativeUNL calculation and may add value to the code base

scottschurr:
I have not looked closely at this pull request.  That said, I'm not convinced that the proposed change brings enough value to justify...

1. The risk associated with the change, or
2. The costs associated with the change (review time spent, testing costs, and an additional amendment).

It looks like this change was initially intended to be part of a larger sweep of changes that @nbougalis had in mind.  Without Nik's hand on the rudder I don't foresee that larger sweep coming to fruition.  So I need to look at this proposed change in isolation.

For my eyes, at least, the risk and costs associated with this proposed change out weigh any benefits I perceive.

I'm open to being convinced otherwise.

src/ripple/app/misc/NegativeUNLVote.cpp
ckeshava: I believe it requires an amendment because it changes the rules for Negative UNL selection. At the moment, Node IDs are used in the selection, but the specification and the documentation for NegativeUNL indicates that PublicKey is used for the same.

ckeshava:
Hello,
1. **Correctness**
The documentation states that we have to use the public keys to choose a candidate for the negative UNL. But we are using the `NodeID` in such calculations [here](https://github.com/XRPLF/rippled/blob/1151fba415a5949adfa1bca8bebf135355a0619c/src/ripple/app/misc/NegativeUNLVote.cpp#L149)
image

This would entail changes to the consensus portion of the code. But in my perspective, this fixes a mistake in the code, it is not a new feature addition. Hence, an amendment should not be required, as much as an extensive code review.

I'm not familiar with the amendment process. Were amendments required for bug fixes in the past?


2. **Efficiency**
At the moment, we are [inter-converting](https://github.com/XRPLF/rippled/blob/1151fba415a5949adfa1bca8bebf135355a0619c/src/ripple/app/misc/NegativeUNLVote.cpp#L48) between `NodeID` and `PublicKey` at every voting ledger (255th ledger). This conversion is not necessary and could be eliminated.


3. **Accuracy of documentation**
I'm not convinced that the code adheres to the documentation apropos the reliability calculation.
image
The code checks if a peer has sent a validation [here](https://github.com/XRPLF/rippled/blob/1151fba415a5949adfa1bca8bebf135355a0619c/src/ripple/app/misc/NegativeUNLVote.cpp#L203). If so, the score is incremented by 1. While this ensures good network connectivity with the peer, it does not check the *behavior/non-maliciousness* of the peer. If the peer sends incorrect validations about a ledger, I believe the code would still increment the count by 1.

I feel its a good change for all these reasons.

@pwang200 You worked on this piece of code, I'd love to hear your opinions too

scottschurr:
@ckeshava, thanks for that documentation.  That helps.  And I too would appreciate hearing @pwang200's opinion regarding this change.

Regarding the three points you bring up:

1. Correctness: you are citing part of the XRPL.org documentation: https://xrpl.org/negative-unl.html#choosing-from-multiple-candidate-validators.  It seems like the easier path would be to fix that documentation.  Adding an amendment so that (working) code matches the documentation seems backwards to me.
2. Efficiency: you have a point here, that the interconversion is unnecessary.  However, does improving this particular efficiency measurably impact the efficiency of the server?  The efficiency argument should come with measurements.  There are lots of places where the server is inefficient.  We have to select places to change so that we make measurable improvements.
3. Reliability Measurement: I see you arguing that the current code does not check the...

> the *behavior/non-maliciousness* of the peer.

But it looks like this pull request does nothing to change that situation.  I don't see any changes in this pull request in `NegativeUNLVote::buildScoreTable()` to check the maliciousness of the peer more than what is currently done.  That portion of the pull request looks like it's simply switching from checking `NodeID`s to `PublicKey`s.

So to my eyes there are two good paths forward with this pull request:

1. Show numbers demonstrating that the improved efficiency from the pull request justifies the change, or
2. Close the pull request and fix the documentation so it describes the actual (current) behavior of the code.

Thoughts?  I'm still open to being convinced that this pull request has _sufficient_ merit to justify it going forward.  But I'm not convinced yet.

ckeshava:
@scottschurr  thanks for looking into it.
I agree with all of your points. Since this code runs only during the flag ledgers, the marginal improvements to efficiency might not amount to much.

I noticed the point-3 a few minutes ago and I thought I'll discuss it here. But, I don't have the foggiest idea on how to tackle it, I'll explore it 👍

pwang200:
I agree with what Scott said. 

Regarding point-1, the goal is that validators can agree on the candidate if there are more than 1, assuming the validators have the same set of candidates. So what we need is a deterministic way to sort them. It doesn't matter if public key or the nodeID is used. Although inconsistent with the doc, but the code does what it should do. It is easier to change the doc. 

This will need an amendment. Because if some nodes use nodeID to sort and some nodes use public key to sort, they will pick different candidates. An amendment ensures that all nodes use the same data to sort. 

Regarding point-2, if we can avoid the inter-converting between NodeID and PublicKey, (it was not obvious to me how to avoid inter-converting 3.5 years ago, but if we can now...) that is good. But I am not sure if the performance improvement is measurable.

ckeshava:
@pwang200 thanks for weighing in your opinions. I agree with your points 👍

ckeshava:
Hello,
Considering the pros and cons of this PR, I'm choosing to close it. As rightly pointed out by Scott Schurr and Peng Wang, the efficiency benefits brought about by this PR is eclipsed by the risk and amendment requirements.

Thanks for looking into this PR!

PR #4268, opened 2022-08-09
Remove charUnHex

a-noni-mousse:
### Type of Change

- [ ] Documentation Updates
- [ ] Release


src/ripple/basics/StringUtilities.h
```diff
@@ -48,6 +48,24 @@ template <class Iterator>
 std::optional<Blob>
 strUnHex(std::size_t strSize, Iterator begin, Iterator end)
 {
+    static constexpr std::array<int, 256> const unxtab = []() {
```
scottschurr, line 51: This file now needs a `#include <array>`.  Please add that near the top of the file like this...
```
...
#include <boost/utility/string_view.hpp>
#include <array>
#include <optional>
...
```

src/ripple/basics/StringUtilities.h
```diff
@@ -56,25 +74,23 @@ strUnHex(std::size_t strSize, Iterator begin, Iterator end)
...
 
         out.push_back(c);
-        ++iter;
     }
 
     while (iter != end)
     {
-        int cHigh = charUnHex(*iter);
+        int cHigh = unxtab[*iter++];
         ++iter;
```
scottschurr, line 88: By accident you left in a double increment.  Fortunately it causes massive unit test failures.  Notes:
1. You'll need to remove the `++iter;` from this line.
2. Consider running the unit tests before submitting pull requests in the future.  It's easy.  Build rippled, then invoke...
```
$ <path to rippled> --unittest
```
Then wait for the results.  It takes a while, but if it says `0 failures` at the end then you're less likely to have introduced a bug.

scottschurr (changes requested):
Thanks for the submission.  This pull request looks pretty close to right, but needs a couple of fixes.  Once those fixes are in place I think this will be good to go.

cryptobrad:
Thanks for looking closely
I fixed the bug you found

scottschurr (approved):
:+1: Latest version looks good to me!  Thanks.

nbougalis:
Merged as 83ac141f656b1a95b5661853951ebd95b3ffba99

PR #4251, opened 2022-07-25
fix #4250 - Update broken link, point to XRPLF docs

chennakeshava1998:
Two other files contain references to the old Github repository - `Builds/linux/README.md` and `Builds/VisualStudio2017/README.md`. But these links automatically point to the new XRPLF repository, so I don't know if they need to be changed.

PR #4195, opened 2022-06-04
Improve detection & handling of duplicate Node ID:

nbougalis:
Each node on the network is supposed to have a unique cryptographic identity. Typically, this identity is generated randomly at startup and stored for later reuse in the (poorly named) file `wallet.db`.

If the file is copied, it is possible for two nodes to share the same node identity. This is generally not desirable and existing servers will detect and reject connections to other servers that have the same key.

This commit achives three things:

1. It improves the detection code to pinpoint instances where two distinct servers with the same key connect with each other. In that case, servers will log an appropriate error and shut down pending intervention by the server's operator.
2. It makes it possible for server administrators to securely and easily generate new cryptographic identities for servers using the new `--newnodeid` command line arguments. When a server is started using this command, it will generate and save a random secure identity.
3. It makes it possible to configure the identity using a command line option, which makes it possible to derive it from data or parameters associated with the container or hardware where the instance is running by passing the `--nodeid` option, followed by a single argument identifying the infomation from which the node's identity is derived. For example, the following command will result in nodes with different hostnames having different node identities: `rippled --nodeid $HOSTNAME`

The last option is particularly useful for automated cloud-based deployments that minimize the need for storing state.

**Important note for server operators:**
Depending on variables outside of the the control of this code, such as operating system version or configuration, permissions, and more, it may be possible for other users or programs to be able to access the command line arguments of other processes on the system.

If you are operating in a shared environment, you should avoid using this option, preferring instead to use the [node_seed] option in the configuration file.

The commit also updates the minimum supported server protocol version to `XRPL/2.1`, which has been supported since version 1.5.0 and eliminates support for `XPRL/2.0`.

### Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Refactor (non-breaking change that only restructures code)
- [ ] Tests (You added tests for code that already exists, or your new feature included in this PR)
- [ ] Documentation Updates
- [ ] Release

PR #4270, opened 2022-08-09
Proposed 1.10.0-b1

### LINKED ISSUE #4250: Fix broken Doxygen link in README
After porting the codebase to XRPLF Github repository, the link to Doxygen documentation is broken. Previously, it pointed to `https://rippled.github.io/rippled/`. It needs to be reconfigured for `https://xrplf.github.io/rippled/` 

I'm referring to the README file at the root of the repository - `README.md`

nbougalis:
If merged this PR will:

- close #4105 (with fix for return value added!)
- close #4174
- close #4181 
- close #4195 
- close #4237 
- close #4240 
- close #4245 
- close #4249 
- close #4250 
- close #4251
- close #4254 
- close #4256 
- close #4261


src/ripple/protocol/impl/AccountID.cpp
```diff
@@ -17,17 +17,95 @@
...
+
+        auto ret =
+            encodeBase58Token(TokenType::AccountID, id.data(), id.size());
+
+        assert(ret.size() <= 38);
+
+        {
+            std::lock_guard lock(sl);
+            cache_[index].id = id;
+            std::strcpy(cache_[index].encoding, ret.c_str());
```
ximinez, line 85: I know there's an `assert` a couple of lines up, but `strcpy` always raises my hackles. Would it hurt to use `strncpy`?

ximinez (approved):
Tested in Windows with VS2019.

src/ripple/app/main/NodeIdentity.cpp
```diff
@@ -20,34 +20,49 @@
...
         auto secretKey = generateSecretKey(KeyType::secp256k1, *seed);
         auto publicKey = derivePublicKey(KeyType::secp256k1, secretKey);
 
         return {publicKey, secretKey};
     }
 
     auto db = app.getWalletDB().checkoutDb();
+
+    if (cmdline.count("newnodeid") != 0)
+        clearNodeIdentity(*db);
```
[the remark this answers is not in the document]
nbougalis: I don't _think_ there's a problem with doing that. `--newnodeid` clears any node identities stored in the database, but it may be worth notifying. We ought to fix this separately, IMO.

src/ripple/protocol/impl/AccountID.cpp
nbougalis: It wouldn't hurt. I'll make a follow-up fix for this, if you don't mind merging this as is.

PR #4272, opened 2022-08-11
Remove gRPC code previously used for the xpring SDK

cjcobb23:
There was a lot of gRPC code in rippled that has long been deprecated. It was used for the xpring sdk, which has been abandoned. Some of the gRPC code is used for reporting mode and clio however. This PR removes the code that is no longer needed, and only retains the gRPC code needed for clio and reporting mode.


nbougalis:
HALLELUJAH!

cfg/rippled-example.cfg
```diff
@@ -1629,10 +1629,10 @@ ip = 127.0.0.1
 admin = 127.0.0.1
 protocol = ws
 
-#[port_grpc]
-#port = 50051
-#ip = 0.0.0.0
-#secure_gateway = 127.0.0.1
+[port_grpc]
```
natenichols, line 1632: Why are we uncommenting this in the `rippled-example.cfg`. Do we want gRPC to be enabled by default?

src/ripple/proto/org/xrpl/rpc/v1/ledger.proto
```diff
@@ -25,16 +25,6 @@ message LedgerSpecifier
     }
 }
 
-// Next field: 3
-message LedgerRange
```
natenichols, line 29: Do we use this in Clio or reporting mode, or only get the LedgerRange from the `ledger` subscription stream?

src/ripple/proto/org/xrpl/rpc/v1/xrp_ledger.proto
```diff
@@ -4,39 +4,18 @@ package org.xrpl.rpc.v1;
...
-
-  // Get all validated transactions associated with a given account
-  rpc GetAccountTransactionHistory(GetAccountTransactionHistoryRequest) returns (GetAccountTransactionHistoryResponse);
-
-  /////////////////////////////////////////////////////////////////////////////
-  // The below methods do not mimic the JSON API exactly, and are mostly binary
-
   // Get a specific ledger, optionally including transactions and any modified,
   // added or deleted ledger objects
   rpc GetLedger(GetLedgerRequest) returns (GetLedgerResponse);
```
natenichols, line 21: Can we also remove `GetLedgerDiff()` and `GetLedgerEntry`? AFAIK, Clio only uses `GetLedgerRequest` and `GetLedgerDataRequest`

src/ripple/rpc/handlers/Tx.cpp
```diff
@@ -251,101 +250,6 @@ doTxHelp(RPC::Context& context, TxArgs const& args)
...
-                }
-            }
-        }
-        response.set_validated(result.validated);
-    }
-    return {response, status};
-}
-
 Json::Value
 populateJsonResponse(
```
natenichols, line 254: I don't actually know what this looked like before we added these gRPC methods, so take what I say next with a grain of salt.

Should we leave `doTxHelp` and `populateJsonResponse` as their own functions, or move it into `doTx()`. It occurs to me that we may have split this out to enable code reuse between the gRPC and JSON-RPC endpoints. 

Is it worth leaving `doTxHelp` as a helper function?

src/test/rpc/ReportingETL_test.cpp
```diff
@@ -773,25 +773,6 @@ class ReportingETL_test : public beast::unit_test::suite
     testNeedCurrentOrClosed()
     {
         testcase("NeedCurrentOrClosed");
-        {
-            org::xrpl::rpc::v1::GetAccountInfoRequest request;
```
natenichols, line 777: Why are we removing this test? Flagging this because it's in `ReportingETL_test.cpp`

natenichols (changes requested):
After a first pass I have a couple of questions

cfg/rippled-example.cfg
cjcobb23: We talked about this in the huddle. I asked if we could uncomment this by default, and the agreement was yes if the xpring sdk code was removed.

src/ripple/proto/org/xrpl/rpc/v1/ledger.proto
cjcobb23: We don't use this in clio or reporting mode

src/ripple/proto/org/xrpl/rpc/v1/xrp_ledger.proto
cjcobb23: Those RPCs are not part of the xpring sdk, and I actually think they could be useful. If we want to remove them, we should discuss and do it as part of a separate PR.

src/test/rpc/ReportingETL_test.cpp
cjcobb23: `GetAccountInfoRequest` is removed. This won't compile if we leave it.

scottschurr (approved):
Thanks so much for taking care to this @cjcobb23!  It's great to see so much of this code gone.  I left a couple of comments that may be relevant.  But other than those this looks to me like it is good to go.  :+1:

natenichols (approved):

cfg/rippled-example.cfg
```diff
@@ -1629,10 +1629,10 @@ ip = 127.0.0.1
...
+port = 50051
+ip = 0.0.0.0
+secure_gateway = 127.0.0.1
```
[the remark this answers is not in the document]
cjcobb23: This is just to make it easier for people to run clio, since clio uses this port to talk to rippled. Though now that I'm thinking about it, it might be best to change the IP to `127.0.0.1`, so that way the port will only listen locally by default. If someone is running clio remotely, they'll need to modify their config anyways.

scottschurr (changes requested):
Thanks for the update @cjcobb23.  I think there are a couple more things that need to be tuned.

1. The RippledCore.cmake file needs to have the KnownFormatToGRPC_test.cpp file removed.
2. The ordering.txt file needs to be tweaked so the levelization CI test passes.

If it's helpful, here's a commit that makes those changes: https://github.com/scottschurr/rippled/commit/f5d9f050f859accca5368f6dcc5440ad172c10a9

Once those changes are made I think this pull request is good to go.

scottschurr (approved):
:+1: Looks great!  Thanks.

PR #4275, opened 2022-08-18
refactor unit tests to use static_asserts - Account and PublicKey

ckeshava:
use type_traits and static asserts instead of unit tests

- [ X ] Refactor (non-breaking change that only restructures code)
- [ X ] Tests (You added tests for code that already exists, or your new feature included in this PR)


src/ripple/protocol/PublicKey.h
```diff
@@ -131,6 +131,12 @@ class PublicKey
     }
 };
 
+static_assert(std::is_default_constructible_v<PublicKey>);
+static_assert(std::is_copy_constructible_v<PublicKey>);
+static_assert(std::is_move_constructible_v<PublicKey>);
```
chennakeshava1998, line 136: Is it a good practise to check for move constructibility? PublicKey class only has the compiler provided default move constructor.

scottschurr: Speaking only for myself, I think it's often good practice to check at compile time whether a given class has the expected interfaces, regardless of whether they are coded explicitly or provided implicitly by the compiler.  If a user defines copy, move, and/or assignment this can have unexpected consequences unless you (and the readers of your code) know C++ backwards and forwards.  `static_asserts` like these provide an easy compile-time way to see if you have the interface you expect.  They also provide a flag in case someone unintentionally changes that interface in the future.

chennakeshava1998: ok, thanks for the clarification!!

src/test/protocol/PublicKey_test.cpp
```diff
@@ -433,32 +433,11 @@ class PublicKey_test : public beast::unit_test::suite
...
-                KeyType::secp256k1, generateSeed("masterpassphrase")));
-
-        PublicKey pk2(pk1);
-        BEAST_EXPECT(pk1 == pk2);
-        BEAST_EXPECT(pk2 == pk1);
-
-        PublicKey pk3;
-        pk3 = pk2;
-        BEAST_EXPECT(pk3 == pk2);
-        BEAST_EXPECT(pk1 == pk3);
```
nbougalis, line 453: I like these tests and I don't think we should remove them. They verify several things:
- the copy constructor and assignment operators; and
- the equality operator.

If anything, I think they ought to be expanded to cover:
- things like self-assignment and self-move-assignment; and
- other comparison operators.

src/test/jtx/Env_test.cpp
```diff
@@ -49,13 +49,6 @@ class Env_test : public beast::unit_test::suite
...
-        {
-            Account a;
-            Account b(a);
-            a = b;
-            a = std::move(b);
-            Account c(std::move(a));
-        }
         Account("alice");
         Account("alice", KeyType::secp256k1);
         Account("alice", KeyType::ed25519);
```
nbougalis, line 54: I'm not sure what these things test here... that these constructors are callable? If so, then these can be converted to `static_assert`.

src/test/protocol/PublicKey_test.cpp
chennakeshava1998: I agree with your third and fourth bullet point about self assignment.
But do we need these tests even after having static_asserts for checking copy-construction, copy-assignment etc?

src/test/jtx/Env_test.cpp
chennakeshava1998: Yes, I think the portion of the code I have removed tests copy and move constructors' callability.

src/test/protocol/PublicKey_test.cpp
ximinez: > I agree with your third and fourth bullet point about self assignment. But do we need these tests even after having static_asserts for checking copy-construction, copy-assignment etc?

I was thinking about this same issue. The static asserts, AFAIK, only check that the functions _exist_. Actually doing the copy, assignment, etc. allows us to verify that the functions _work correctly_, which is way more important.

chennakeshava1998: oh interesting. I was under the assumption that static_asserts also check for the correctness of the function. 

In that case, we will need these unit tests even after having static_asserts.

I'll make the necessary changes, thanks.

intelliot:
@ckeshava what are your latest thoughts on this change?

ckeshava:
I believe using static_asserts is a good programming practice, it helps improve the readability of the code too.

scottschurr (changes requested):
There were tests removed from PublicKey_test.cpp.  The test method was called `testMiscOperations()`.  Reviewers requested that those tests not be removed.  It looks like the pull request has not yet re-added those tests.  Let's get `testMiscOperations()` re-added.

ckeshava:
Hi Scott,
thanks for the review. I'm doing my final exams now, is it okay if I take a look at this after a few weeks? Thanks!

scottschurr:
Hi @ckeshava, sure.  No hurry on my account.  I left the comment because I was nudged by @intelliot.

I hope your exams go well.  Don't let this distract you.  It will wait.

ckeshava:
thank you scott!

ckeshava:
I have addressed the comments on the PR. Please let me know if this looks ok. I can merge/collapse the commits into one single commit. @intelliot  @scottschurr

intelliot:
> I can merge/collapse the commits into one single commit.

There is no need for that, as the PR will be squashed when it is merged.

Can you suggest a final commit message for the squashed commit?

src/test/jtx/Account.h
```diff
@@ -145,6 +146,12 @@ class Account
     std::string human_;  // base58 public key string
 };
 
+static_assert(std::is_default_constructible_v<Account>);
+static_assert(std::is_copy_constructible_v<Account>);
+static_assert(std::is_move_constructible_v<Account>);
+static_assert(std::is_copy_assignable_v<Account>);
+static_assert(std::is_move_assignable_v<Account>);
```
seelabs, line 153: I don't see a lot of value is adding static asserts like this. For example, in the rare cases when you make a class that should be default constructable not-default constructable you'll find that out when you try to use it. I don't object too strongly if you really want to add these although I do prefer the code without it.

ckeshava:
Ok. The final commit message could be: "If merged, this commit includes
additional static_assert statements. Certain old tests are eliminated to
reduce redundancy in unit tests."
Regards,
Keshava

src/test/jtx/Account.h
ckeshava: yeah, that's a reasonable critique. i forgot the context with which I began making this change 😅

seelabs:
@ckeshava The commit message shouldn't include the phrase "If merged...". If we're reading this in the git log, it's been merged, and if we include that phrase every commit message would contain the same starting phrase.

ckeshava:
Got it, okay.
Regards,
Keshava

src/test/jtx/Env_test.cpp
```diff
@@ -49,13 +49,6 @@ class Env_test : public beast::unit_test::suite
     testAccount()
     {
         using namespace jtx;
-        {
-            Account a;
-            Account b(a);
-            a = b;
-            a = std::move(b);
-            Account c(std::move(a));
-        }
```
scottschurr, line 58: I'm not a huge fan of the `static_asserts` being sprinkled in PublicKey.h and jtx/Account.h.  I've actually done similar stuff when I was designing new classes and I wanted to be confident I had the right semantics.  But it was usually during initial design.  I've also sometimes added similar `static_assert`s if I was modifying a class that I suspected I didn't understand well.

However both `PublicKey` and `jtx::Account` are well exercised and wrung out classes.  And we're not modifying them.  Adding the `static_assert`s there does not really improve our confidence that the classes are doing the right thing.  We already have pretty good confidence in that.

If we move our focus to the (above) code in the braces, that code is just verifying that all of the usual constructors and assignment operators compile and don't crash if they execute.  So we could hypothetically replace them with `static_assert`s like this:
```
        static_assert(std::is_default_constructible_v<Account>);
        static_assert(std::is_copy_constructible_v<Account>);
        static_assert(std::is_move_constructible_v<Account>);
        static_assert(std::is_copy_assignable_v<Account>);
        static_assert(std::is_move_assignable_v<Account>);
```
But there's a (small) problem here as well.  The `static_assert`s only show that the compiler is fine with calling those special member functions.  What the `static_assert`s _don't_ demonstrate is whether calling those special member functions crashes.  So by switching to the `static_assert`s we are actually reducing our test coverage.

Now, in truth, at this point in time we're pretty well convinced that all these special member functions:

1. Are callable,
2. Don't crash,  and
3. Actually do what they are expected to do (which, incidentally, the code in braces does not verify).

So, while the changes in this pull request are interesting, I don't know that they actually bring a lot of value to the code base.

I'm open to discussion if you have questions or disagree.

scottschurr (commented):
This is an interesting pull request.  But I'm not convinced that it brings a lot of value to the code base.  Thanks for the submission, but at the moment I'm inclined to close the pull request without merging.  Details for why are in the comment.

Don't be discouraged.  You're learning new stuff.  Thanks for the effort.

ckeshava:
No worries, I understand, thank you for the explanation :+1.
Regards,
Keshava

ckeshava:
I agree with the other reviewers of the PR. I believe the existing tests already cover and test the construct, move and copy-able ness of the `PublicKey` class. I don't think these changes are necessary. Hence, closing the PR. Thanks for reviewing it!

PR #4277, opened 2022-08-19
Proposed 1.9.3

nbougalis:
This release corrects a minor technical flaw that could cause servers to incorrectly process a request to vote in support for an amendment during server startup. A minor technical fix in the copy constructor of `PublicKey` is also included.

If merged, this will:

- close #4254
- close #4256
- close #4263

### DIFF
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -7,6 +7,38 @@ This document contains the release notes for `rippled`, the reference server imp
  
 Have new ideas? Need help with setting up your node? Come visit us [here](https://github.com/xrplf/rippled/issues/new/choose)
 
+# Introducing XRP Ledger version 1.9.3
+
+Version 1.9.3 of `rippled`, the reference server implementation of the XRP Ledger protocol is now available. This release corrects minor technical flaws with the code that loads configured amendment votes after a startup and the copy constructor of `PublicKey`.
+
+## Install / Upgrade
+
+On supported platforms, see the [instructions on installing or updating `rippled`](https://xrpl.org/install-rippled.html).
+
+## Changelog
+
+## Contributions
+
+This releases contains the following bug fixes:
+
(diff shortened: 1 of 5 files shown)


RELEASENOTES.md
```diff
@@ -7,6 +7,38 @@ This document contains the release notes for `rippled`, the reference server imp
...
+
+On supported platforms, see the [instructions on installing or updating `rippled`](https://xrpl.org/install-rippled.html).
+
+## Changelog
+
+## Contributions
+
+This releases contains the following bug fixes:
+
+- **Change by-value to by-reference to persist vote**: A minor technical flaw, caused by use of a copy instead of a reference, resulted in operator-configured "yes" votes to appear to not persist after a restart. ([#4256](https://github.com/XRPLF/rippled/pull/4256))
```
mDuo13, line 24: The opening paragraph (line 12) says that the server would "incorrectly process" the request to vote. Here it says "to appear to not persist" (so they _do_ persist but they don't seem to?) That sounds like a contradiction to me.

We should clarify whether the votes actually persist and if the only "incorrect" processing is that they don't persist we should be clear in both places.

mDuo13 (changes requested):

RELEASENOTES.md
alloynetworks: The correct vote is present in the database and is retrieved correctly. The `copy instead of reference` however makes the processing of the same incorrect. AFAIK :)

nbougalis: So, the intro says "incorrectly process a request to vote in support for an amendment _during server startup_" (emphasis added). In my mind, this accurately describes the problem, but I can understand your confusion, @mDuo13.

We could change the intro to read "incorrectly processes the stored amendment voting configuration"

ximinez (approved):
Tested in Windows with VS2019.

nbougalis:
@mDuo13, I've addressed your concern by rewording `RELEASENOTES.md`. I would appreciate an review by another maintainer, so that we can proceed to merge this and make a release.

mDuo13 (approved):
The revised wording looks good. Thanks!

PR #4280, opened 2022-08-22
Use strongly-typed AcctRoot class to internally represent account root ledger objects

scottschurr:

Up to this point the server's internal code has presented every ledger object type to programmers as a

- `std::shared_ptr<STLedgerEntry>` or as a
- `std::shared_ptr<STLedgerEntry const>`.

The choice of representation was based on whether or not the returned object was mutable.

That's fine, and it has gotten us a long way.  But it has some limitations.

The different types of objects in the ledger are in fact, different.  We use the same type, a `std::shared_ptr<STLedgerEntry>`, to represent an account root, a directory node, and a trust line (RippleState).  These things are very different, and programmers think about them them differently.  But we're treating all three of these types (and many more) as identical with the C++ type system.

It doesn't have to be like that, and this pull request is a step along that journey.

### Context of Change

We've been considering ways to have ledger objects take up less space in the ledger.  But many the changes that these space reductions would entail are hard to manage given the ad hoc way that we handle ledger entry objects.  It would be nice to put a uniform face on, say, a `TrustLine`.  Then any funny stuff we do to save space can be hidden behind that uniform face.  We don't have to pollute every individual piece of code that operates on `TrustLine` with the funny stuff.  We can hide any funny stuff behind the uniform face.

So that's the motivation for the change.  But it's not the only benefit.  If an account root has the type `AcctRoot`, and we pass a reference to it, then the receiving end knows at _compile time_ that they got an `AcctRoot`.  As it stands today the receiving end gets a `std::shared_ptr<STLedgerEntry>` which _could_ be an account root or a signer list or a ticket or an escrow or something else altogether.

#### Techniques

To achieve this end we'll use the following techniques:

1. We'll make our `Keylet`s strongly typed.  So each kind of `Keylet` is a distinct type that knows exactly which type it is supposed to retrieve from the ledger at compile time.
2. These strongly typed `Keylet`s will allow the `AppyView` or `ReadView` to know which ledger object type needs to be returned.
3. We'll use templates to reduce the boiler plate involved in making strongly typed `Keylet`s and ledger object types.

#### Steps

This pull request contains early stages toward making all ledger objects strongly typed.  But it does not finish the job.  This pull request has three separate commits, each of which takes us one step along the way:

1. The first commit introduces a base class to `Keylet`.  This gives us a place to put common behavior that is shared between all of the strongly typed `Keylet`s.

2. The second commit renames the `ReadView::read()` method to `ReadView::readSLE()` and `ApplyView::peek()` method to `ApplyView::peekSLE()`.  This rename is propagated to the derived classes as well.  The renamed methods will continue to return serialized ledger entries.  This is in anticipation of having future `read()` and `peek()` methods that return strongly typed objects.

3. The third commit introduces the first strongly typed `Keylet` and its corresponding strongly typed ledger object `AcctRoot`.  The account root was chosen as the first type because it is used a lot and should provide a good test ground for the methodology.

The first two commits allow the transition to using strongly-typed ledger objects to be incremental.  The third commit is just the first of many.  Future commits will convert more ledger objects to be strongly typed.

### Type of Change

## Before / After

Before:
```
std::shared_ptr<SLE const> sleAcct = view.read(keylet::account(acctID));
if (!sleAcct)
    return terNO_ACCOUNT;

SeqProxy const acctSeqProx = SeqProxy::sequence((*sleAccount)[sfSequence]);
```

After:
```
std::optional<AcctRoot> acctRoot = view.read(keylet::account(acctID));
if (!acctRoot)
    return terNO_ACCOUNT;

SeqProxy const acctSeqProx = SeqProxy::sequence(acctRoot->sequence());
```

## Test Plan

I looked at the code coverage of these changes based on existing unit tests.  I was surprised and pleased at how thoroughly these refactoring changes were exercised by pre-existing unit tests.  As a result I felt no inclination to introduce additional tests.  Reviewers may feel differently.

## Future Tasks

These are only the first few steps of this project.  In order to have a coherent code base it will be important to complete the conversion of all other ledger types to strongly typed.  I anticipate that each newly introduced strongly typed ledger object will arrive in a separate pull request.

### DIFF
--- a/src/ripple/app/ledger/impl/LocalTxs.cpp
+++ b/src/ripple/app/ledger/impl/LocalTxs.cpp
@@ -156,7 +156,7 @@ public:
                 return true;
 
             AccountID const acctID = txn.getAccount();
-            auto const sleAcct = view.read(keylet::account(acctID));
+            auto const sleAcct = view.readSLE(keylet::account(acctID));
 
             if (!sleAcct)
                 return false;
--- a/src/ripple/app/ledger/impl/LocalTxs.cpp
+++ b/src/ripple/app/ledger/impl/LocalTxs.cpp
@@ -156,13 +156,12 @@ public:
                 return true;
 
             AccountID const acctID = txn.getAccount();
-            auto const sleAcct = view.readSLE(keylet::account(acctID));
+            auto const acctRoot = view.read(keylet::account(acctID));
 
-            if (!sleAcct)
+            if (!acctRoot)
                 return false;
 
-            SeqProxy const acctSeq =
-                SeqProxy::sequence(sleAcct->getFieldU32(sfSequence));
+            SeqProxy const acctSeq = SeqProxy::sequence(acctRoot->sequence());
             SeqProxy const seqProx = txn.getSeqProxy();
 
             if (seqProx.isSeq())
--- a/src/ripple/app/misc/TxQ.h
+++ b/src/ripple/app/misc/TxQ.h
@@ -306,7 +306,7 @@ public:
 
     /** Return the next sequence that would go in the TxQ for an account. */
     SeqProxy
-    nextQueuableSeq(std::shared_ptr<SLE const> const& sleAccount) const;
+    nextQueuableSeq(std::optional<AcctRootRd const> const& acctRoot) const;
 
     /** Returns fee metrics in reference fee level units.
      */
@@ -363,7 +363,7 @@ private:
     // Implementation for nextQueuableSeq().  The passed lock must be held.
     SeqProxy
     nextQueuableSeqImpl(
-        std::shared_ptr<SLE const> const& sleAccount,
+        std::optional<AcctRootRd const> const& acctRoot,
         std::lock_guard<std::mutex> const&) const;
 
     /**
@@ -812,7 +812,7 @@ private:
         STTx const&,
         ApplyFlags const,
         OpenView const&,
-        std::shared_ptr<SLE const> const& sleAccount,
+        std::optional<AcctRootRd const> const& acctRoot,
         AccountMap::iterator const&,
         std::optional<TxQAccount::TxMap::iterator> const&,
         std::lock_guard<std::mutex> const& lock);
--- a/src/ripple/app/misc/impl/TxQ.cpp
+++ b/src/ripple/app/misc/impl/TxQ.cpp
@@ -742,7 +742,7 @@ TxQ::apply(
     // If the account is not currently in the ledger, don't queue its tx.
     auto const account = (*tx)[sfAccount];
     Keylet const accountKey{keylet::account(account)};
-    auto const sleAccount = view.read(accountKey);
+    auto const sleAccount = view.readSLE(accountKey);
     if (!sleAccount)
         return {terNO_ACCOUNT, false};
 
@@ -1109,7 +1109,7 @@ TxQ::apply(
             // Create the test view from the current view.
             multiTxn.emplace(view, flags);
 
-            auto const sleBump = multiTxn->applyView.peek(accountKey);
+            auto const sleBump = multiTxn->applyView.peekSLE(accountKey);
             if (!sleBump)
                 return {tefINTERNAL, false};
 
@@ -1661,7 +1661,7 @@ TxQ::tryDirectApply(
     beast::Journal j)
 {
     auto const account = (*tx)[sfAccount];
-    auto const sleAccount = view.read(keylet::account(account));
+    auto const sleAccount = view.readSLE(keylet::account(account));
 
     // Don't attempt to direct apply if the account is not in the ledger.
     if (!sleAccount)
@@ -1778,7 +1778,7 @@ TxQ::getTxRequiredFeeAndSeq(
     auto const baseFee = calculateBaseFee(view, *tx);
     auto const fee = FeeMetrics::scaleFeeLevel(snapshot, view);
 
-    auto const sle = view.read(keylet::account(account));
+    auto const sle = view.readSLE(keylet::account(account));
 
     std::uint32_t const accountSeq = sle ? (*sle)[sfSequence] : 0;
     std::uint32_t const availableSeq = nextQueuableSeqImpl(sle, lock).value();
--- a/src/ripple/app/misc/impl/TxQ.cpp
+++ b/src/ripple/app/misc/impl/TxQ.cpp
@@ -374,7 +374,7 @@ TxQ::canBeHeld(
     STTx const& tx,
     ApplyFlags const flags,
     OpenView const& view,
-    std::shared_ptr<SLE const> const& sleAccount,
+    std::optional<AcctRootRd const> const& acctRoot,
     AccountMap::iterator const& accountIter,
     std::optional<TxQAccount::TxMap::iterator> const& replacementIter,
     std::lock_guard<std::mutex> const& lock)
@@ -419,7 +419,7 @@ TxQ::canBeHeld(
         return telCAN_NOT_QUEUE_FULL;
 
     // This is the next queuable sequence-based SeqProxy for the account.
-    SeqProxy const nextQueuable = nextQueuableSeqImpl(sleAccount, lock);
+    SeqProxy const nextQueuable = nextQueuableSeqImpl(acctRoot, lock);
     if (txSeqProx != nextQueuable)
         // The provided transaction does not fill the next open sequence gap.
         return telCAN_NOT_QUEUE_FULL;
@@ -741,13 +741,13 @@ TxQ::apply(
 
     // If the account is not currently in the ledger, don't queue its tx.
     auto const account = (*tx)[sfAccount];
-    Keylet const accountKey{keylet::account(account)};
-    auto const sleAccount = view.readSLE(accountKey);
-    if (!sleAccount)
+    AccountRootKeylet const accountKey{keylet::account(account)};
+    auto const acctRoot = view.read(accountKey);
+    if (!acctRoot)
         return {terNO_ACCOUNT, false};
 
     // If the transaction needs a Ticket is that Ticket in the ledger?
-    SeqProxy const acctSeqProx = SeqProxy::sequence((*sleAccount)[sfSequence]);
+    SeqProxy const acctSeqProx = SeqProxy::sequence(acctRoot->sequence());
     SeqProxy const txSeqProx = tx->getSeqProxy();
     if (txSeqProx.isTicket() &&
         !view.exists(keylet::ticket(account, txSeqProx)))
@@ -961,13 +961,7 @@ TxQ::apply(
             // If the transaction is queueable, create the multiTxn
             // object to hold the info we need to adjust for prior txns.
             TER const ter{canBeHeld(
-                *tx,
-                flags,
-                view,
-                sleAccount,
-                accountIter,
-                replacedTxIter,
-                lock)};
+                *tx, flags, view, acctRoot, accountIter, replacedTxIter, lock)};
             if (!isTesSuccess(ter))
                 return {ter, false};
 
@@ -1018,7 +1012,7 @@ TxQ::apply(
                 // transaction fits in proper sequence order with the
                 // previous transaction or is a ticket.
                 if (txSeqProx.isSeq() &&
-                    nextQueuableSeqImpl(sleAccount, lock) != txSeqProx)
+                    nextQueuableSeqImpl(acctRoot, lock) != txSeqProx)
                     return {telCAN_NOT_QUEUE, false};
             }
 
@@ -1077,7 +1071,7 @@ TxQ::apply(
                 Transactions stuck in the queue are mitigated by
                 LastLedgerSeq and MaybeTx::retriesRemaining.
             */
-            auto const balance = (*sleAccount)[sfBalance].xrp();
+            auto const balance = acctRoot->balance().xrp();
             /* Get the minimum possible account reserve. If it
                is at least 10 * the base fee, and fees exceed
                this amount, the transaction can't be queued.
@@ -1109,8 +1103,8 @@ TxQ::apply(
             // Create the test view from the current view.
             multiTxn.emplace(view, flags);
 
-            auto const sleBump = multiTxn->applyView.peekSLE(accountKey);
-            if (!sleBump)
+            auto acctRootBump = multiTxn->applyView.peek(accountKey);
+            if (!acctRootBump)
                 return {tefINTERNAL, false};
 
             // Subtract the fees and XRP spend from all of the other
@@ -1122,15 +1116,16 @@ TxQ::apply(
                 potentialTotalSpend > XRPAmount{0} ||
                 (potentialTotalSpend == XRPAmount{0} &&
                  multiTxn->applyView.fees().base == 0));
-            sleBump->setFieldAmount(sfBalance, balance - potentialTotalSpend);
+            acctRootBump->setBalance(balance - potentialTotalSpend);
             // The transaction's sequence/ticket will be valid when the other
             // transactions in the queue have been processed. If the tx has a
             // sequence, set the account to match it. If it has a ticket, use
             // the next queueable sequence, which is the closest approximation
             // to the most successful case.
-            sleBump->at(sfSequence) = txSeqProx.isSeq()
-                ? txSeqProx.value()
-                : nextQueuableSeqImpl(sleAccount, lock).value();
+            acctRootBump->setSequence(
+                txSeqProx.isSeq()
+                    ? txSeqProx.value()
+                    : nextQueuableSeqImpl(acctRoot, lock).value());
         }
     }
 
@@ -1208,7 +1203,7 @@ TxQ::apply(
     if (!multiTxn)
     {
         TER const ter{canBeHeld(
-            *tx, flags, view, sleAccount, accountIter, replacedTxIter, lock)};
+            *tx, flags, view, acctRoot, accountIter, replacedTxIter, lock)};
         if (!isTesSuccess(ter))
         {
             // Bail, transaction cannot be held
@@ -1583,10 +1578,10 @@ TxQ::accept(Application& app, OpenView& view)
 //
 // Acquires a lock and calls the implementation.
 SeqProxy
-TxQ::nextQueuableSeq(std::shared_ptr<SLE const> const& sleAccount) const
+TxQ::nextQueuableSeq(std::optional<AcctRootRd const> const& acctRoot) const
 {
     std::lock_guard<std::mutex> lock(mutex_);
-    return nextQueuableSeqImpl(sleAccount, lock);
+    return nextQueuableSeqImpl(acctRoot, lock);
 }
 
 // The goal is to return a SeqProxy for a sequence that will fill the next
@@ -1597,18 +1592,18 @@ TxQ::nextQueuableSeq(std::shared_ptr<SLE const> const& sleAccount) const
 // be found and returned.
 SeqProxy
 TxQ::nextQueuableSeqImpl(
-    std::shared_ptr<SLE const> const& sleAccount,
+    std::optional<AcctRootRd const> const& acctRoot,
     std::lock_guard<std::mutex> const&) const
 {
     // If the account is not in the ledger or a non-account was passed
     // then return zero.  We have no idea.
-    if (!sleAccount || sleAccount->getType() != ltACCOUNT_ROOT)
+    if (!acctRoot)
         return SeqProxy::sequence(0);
 
-    SeqProxy const acctSeqProx = SeqProxy::sequence((*sleAccount)[sfSequence]);
+    SeqProxy const acctSeqProx = SeqProxy::sequence(acctRoot->sequence());
 
     // If the account is not in the queue then acctSeqProx is good enough.
-    auto const accountIter = byAccount_.find((*sleAccount)[sfAccount]);
+    auto const accountIter = byAccount_.find(acctRoot->accountID());
     if (accountIter == byAccount_.end() ||
         accountIter->second.transactions.empty())
         return acctSeqProx;
@@ -1661,13 +1656,13 @@ TxQ::tryDirectApply(
     beast::Journal j)
 {
     auto const account = (*tx)[sfAccount];
-    auto const sleAccount = view.readSLE(keylet::account(account));
+    auto const acctRoot = view.read(keylet::account(account));
 
     // Don't attempt to direct apply if the account is not in the ledger.
-    if (!sleAccount)
+    if (!acctRoot)
         return {};
 
-    SeqProxy const acctSeqProx = SeqProxy::sequence((*sleAccount)[sfSequence]);
+    SeqProxy const acctSeqProx = SeqProxy::sequence(acctRoot->sequence());
     SeqProxy const txSeqProx = tx->getSeqProxy();
 
     // Can only directly apply if the transaction sequence matches the account
@@ -1778,10 +1773,11 @@ TxQ::getTxRequiredFeeAndSeq(
     auto const baseFee = calculateBaseFee(view, *tx);
     auto const fee = FeeMetrics::scaleFeeLevel(snapshot, view);
 
-    auto const sle = view.readSLE(keylet::account(account));
+    auto const acctRoot = view.read(keylet::account(account));
 
-    std::uint32_t const accountSeq = sle ? (*sle)[sfSequence] : 0;
-    std::uint32_t const availableSeq = nextQueuableSeqImpl(sle, lock).value();
+    std::uint32_t const accountSeq = acctRoot ? acctRoot->sequence() : 0;
+    std::uint32_t const availableSeq =
+        nextQueuableSeqImpl(acctRoot, lock).value();
 
     return {mulDiv(fee, baseFee, baseLevel).second, accountSeq, availableSeq};
 }
--- a/src/ripple/app/tx/impl/Payment.cpp
+++ b/src/ripple/app/tx/impl/Payment.cpp
@@ -216,9 +216,9 @@ Payment::preclaim(PreclaimContext const& ctx)
     STAmount const saDstAmount(ctx.tx[sfAmount]);
 
     auto const k = keylet::account(uDstAccountID);
-    auto const sleDst = ctx.view.readSLE(k);
+    auto const acctRootDst = ctx.view.read(k);
 
-    if (!sleDst)
+    if (!acctRootDst)
     {
         // Destination account does not exist.
         if (!saDstAmount.native())
@@ -256,7 +256,7 @@ Payment::preclaim(PreclaimContext const& ctx)
         }
     }
     else if (
-        (sleDst->getFlags() & lsfRequireDestTag) &&
+        acctRootDst->isFlag(lsfRequireDestTag) &&
         !ctx.tx.isFieldPresent(sfDestinationTag))
     {
         // The tag is basically account-specific information we don't
@@ -329,31 +329,35 @@ Payment::doApply()
 
     // Open a ledger for editing.
     auto const k = keylet::account(uDstAccountID);
-    SLE::pointer sleDst = view().peekSLE(k);
+    auto acctRootDst = view().peek(k);
 
-    if (!sleDst)
+    if (!acctRootDst)
     {
         std::uint32_t const seqno{
             view().rules().enabled(featureDeletableAccounts) ? view().seq()
                                                              : 1};
 
         // Create the account.
-        sleDst = std::make_shared<SLE>(k);
+        SLE::pointer sleDst = std::make_shared<SLE>(k);
         sleDst->setAccountID(sfAccount, uDstAccountID);
         sleDst->setFieldU32(sfSequence, seqno);
 
         view().insert(sleDst);
+        acctRootDst = view().peek(k);
+        if (!acctRootDst)
+            // Should never happen.
+            return tefINTERNAL;
     }
     else
     {
         // Tell the engine that we are intending to change the destination
         // account.  The source account gets always charged a fee so it's always
         // marked as modified.
-        view().update(sleDst);
+        view().update(acctRootDst);
     }
 
     // Determine whether the destination requires deposit authorization.
-    bool const reqDepositAuth = sleDst->getFlags() & lsfDepositAuth &&
+    bool const reqDepositAuth = acctRootDst->isFlag(lsfDepositAuth) &&
         view().rules().enabled(featureDepositAuth);
 
     bool const depositPreauth = view().rules().enabled(featureDepositPreauth);
@@ -438,13 +442,13 @@ Payment::doApply()
 
     // Direct XRP payment.
 
-    auto const sleSrc = view().peekSLE(keylet::account(account_));
-    if (!sleSrc)
+    auto acctRootSrc = view().peek(keylet::account(account_));
+    if (!acctRootSrc)
         return tefINTERNAL;
 
     // uOwnerCount is the number of entries in this ledger for this
     // account that require a reserve.
-    auto const uOwnerCount = sleSrc->getFieldU32(sfOwnerCount);
+    auto const uOwnerCount = acctRootSrc->ownerCount();
 
     // This is the total reserve in drops.
     auto const reserve = view().fees().accountReserve(uOwnerCount);
@@ -496,20 +500,19 @@ Payment::doApply()
                 XRPAmount const dstReserve{view().fees().accountReserve(0)};
 
                 if (saDstAmount > dstReserve ||
-                    sleDst->getFieldAmount(sfBalance) > dstReserve)
+                    acctRootDst->balance() > dstReserve)
                     return tecNO_PERMISSION;
             }
         }
     }
 
     // Do the arithmetic for the transfer and make the ledger change.
-    sleSrc->setFieldAmount(sfBalance, mSourceBalance - saDstAmount);
-    sleDst->setFieldAmount(
-        sfBalance, sleDst->getFieldAmount(sfBalance) + saDstAmount);
+    acctRootSrc->setBalance(mSourceBalance - saDstAmount);
+    acctRootDst->setBalance(acctRootDst->balance() + saDstAmount);
 
     // Re-arm the password change fee if we can and need to.
-    if ((sleDst->getFlags() & lsfPasswordSpent))
-        sleDst->clearFlag(lsfPasswordSpent);
+    if (acctRootDst->isFlag(lsfPasswordSpent))
+        acctRootDst->clearFlag(lsfPasswordSpent);
 
     return tesSUCCESS;
 }
--- a/src/ripple/app/tx/impl/Payment.cpp
+++ b/src/ripple/app/tx/impl/Payment.cpp
@@ -353,7 +353,7 @@ Payment::doApply()
         // Tell the engine that we are intending to change the destination
         // account.  The source account gets always charged a fee so it's always
         // marked as modified.
-        view().update(acctRootDst);
+        view().update(*acctRootDst);
     }
 
     // Determine whether the destination requires deposit authorization.
--- a/src/ripple/app/tx/impl/Payment.cpp
+++ b/src/ripple/app/tx/impl/Payment.cpp
@@ -216,7 +216,7 @@ Payment::preclaim(PreclaimContext const& ctx)
     STAmount const saDstAmount(ctx.tx[sfAmount]);
 
     auto const k = keylet::account(uDstAccountID);
-    auto const sleDst = ctx.view.read(k);
+    auto const sleDst = ctx.view.readSLE(k);
 
     if (!sleDst)
     {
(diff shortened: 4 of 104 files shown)


greg7mdp:
I had a quick look at the changes, and I fully agree with the direction they are taking us towards.
One thing that I found immediately surprising was `view.read()` returning `std::optional<AcctRoot>`, which looks like a value, so I wondered if there was any copying?
There is not, because `AcctRoot` is a pointer in disguise, being a `LedgerEntryWrapper` which itself encapsulates a `std::shared_ptr`.
I think developers will have to be aware of that, so I think it would be more straightfworward to have `view.read()` return the `AcctRoot` (instead of `std::optional<AcctRoot>`), and `AcctRoot` could have an operator bool() which would just return `LedgerEntryWrapper::wrapped_::operator bool()`.

src/ripple/protocol/LedgerEntryWrapper.h
```diff
@@ -0,0 +1,194 @@
...
+            return wrapped_;
+    }
+
+    [[nodiscard]] std::shared_ptr<STLedgerEntry> const&
+    slePtr()
+    {
+        static_assert(
+            Writable, "Cannot access non-const SLE of const ledger entry.");
+        return wrapped_;
+    }
```
greg7mdp, line 156: I think it could be problematic to return a `shared_ptr` reference, even `const`, as the pointed to memory could be freed at any time.

scottschurr: This interface is provided to provide compatibility with the pre-existing interfaces that are passed `std::shared_ptr<STLedgerEntry> const&`, for example:

- https://github.com/XRPLF/rippled/pull/4280/files#diff-fd246da4e6e9e21f97f1b825c5144977282f5730efad3856f20e11bf4355d1bbR254 and
- https://github.com/XRPLF/rippled/pull/4280/files#diff-fd246da4e6e9e21f97f1b825c5144977282f5730efad3856f20e11bf4355d1bbR201

So unless we we're motivated to do some messing with the existing machinery, I thing we need to return a `std::shared_ptr<STLedgerEntry> const`.

But I do hear your concerns bout returning a reference to a `std::shard_ptr`.  The lifetime of the reference will only be as long as the `LedgerEntryWrapper` that returns the reference.  If we're sufficiently worried about the safely of this interface we could return a copy of the `std::shared_ptr`.  That would provide the lifetime extension properties of a `shared_ptr`, but also incur the cost of copying the `shared_ptr`.

Do you think we should return a copy of the `shared_ptr`?  Is the added safety worth the cost of the atomic increment and decrement in the control block?  Or do you see another alternative that I missed?

greg7mdp: I honestly don't see why we would pass `std::shared_ptr<STLedgerEntry> const&` to an interface. Passing a raw pointer `STLedgerEntry *` is functionally equivalent, and more versatile in case we happen to have only a raw pointer available.

I also don't like the idea of returning copies of the `std::shared_ptr` because of the cost of doing the memory access to the pointed value to increment (and decrement later) the count, so I would vote for leaving the code as you have it if we can't just use raw pointers when not managing the lifetime of the pointed objects.

scottschurr:
Rebased to 1.10.0-b1.

src/ripple/app/misc/TxQ.h
```diff
@@ -306,7 +306,7 @@ class TxQ
 
     /** Return the next sequence that would go in the TxQ for an account. */
     SeqProxy
-    nextQueuableSeq(std::shared_ptr<SLE const> const& sleAccount) const;
+    nextQueuableSeq(std::optional<AcctRootRd const> const& acctRoot) const;
```
seelabs, line 309: Can we replace the `std::optional` with a `AcctRootRd const&`? I can't imagine it ever makes sense to ask for a queuable seq for a non-existant account root. I understand the old interface took a pointer type - and it probably should not have.

There are other functions that are taking a `optional<AcctRoot...>`, can we audit these as well to see if we can remove the `optional`?

seelabs, line 309: This interface also shows a problem with the implementation of the wrappers. `AcctRootRd` is _really_ just an `AcctRoot` that is `const`, but instead of the standard C++ `AcctRoot const` we need to use a special `AcctRootRd` class, and I guess we may as well mark that `const` too.

I *know* this is a hard problem - I'm not criticizing. I'm fine with the current implementation, although I wish we could solve this in a reasonable way (we did explore a `const_cast` solution at one point, but I understand why we didn't pursue that - that has it's own set of trade-offs).

src/ripple/app/tx/impl/SetTrust.cpp
```diff
@@ -233,8 +233,8 @@ SetTrust::doApply()
         std::uint32_t uHighQualityOut;
         auto const& uLowAccountID = !bHigh ? account_ : uDstAccountID;
         auto const& uHighAccountID = bHigh ? account_ : uDstAccountID;
-        SLE::ref sleLowAccount = !bHigh ? sle : sleDst;
-        SLE::ref sleHighAccount = bHigh ? sle : sleDst;
+        std::optional<AcctRoot>& lowAcctRoot = !bHigh ? acctRoot : destAcctRoot;
+        std::optional<AcctRoot>& highAcctRoot = bHigh ? acctRoot : destAcctRoot;
```
seelabs, line 237: It looks like these can be juse `AccountRoot&`, no `optional` needed.

src/ripple/ledger/View.h
```diff
@@ -246,7 +246,7 @@ areCompatible(
 void
 adjustOwnerCount(
     ApplyView& view,
-    std::shared_ptr<SLE> const& sle,
+    std::optional<AcctRoot>& acctRoot,
```
seelabs, line 249: Can we lose the `optional`?

src/ripple/ledger/View.h
```diff
@@ -330,15 +330,15 @@ trustCreate(
...
-    SLE::ref sleAccount,        // --> the account being set.
-    const bool bAuth,           // --> authorize account.
-    const bool bNoRipple,       // --> others cannot ripple through
-    const bool bFreeze,         // --> funds cannot leave
-    STAmount const& saBalance,  // --> balance of account being set.
-                                // Issuer should be noAccount()
-    STAmount const& saLimit,    // --> limit for account being set.
-                                // Issuer should be the account being set.
+    uint256 const& uIndex,              // --> ripple state entry
+    std::optional<AcctRoot>& acctRoot,  // --> the account being set.
```
seelabs, line 334: Can we lose the `optional`?

src/ripple/protocol/AcctRoot.h
```diff
@@ -0,0 +1,377 @@
...
+    }
+
+    [[nodiscard]] std::optional<AccountID>
+    regularKey() const
+    {
+        return wrapped_->at(~sfRegularKey);
+    }
+
+    void
+    setRegularKey(AccountID const& newRegKey)
```
seelabs, line 176: Can we add a `requires Writable` and lose the `static_assert` (here and all the other non-const functions)? I think this is cleaner:
```
    void
    setRegularKey(AccountID const& newRegKey)
        requires Writable
    {
        this->setOptional(sfRegularKey, newRegKey);
    }
```

src/ripple/protocol/AcctRoot.h
```diff
@@ -0,0 +1,377 @@
...
+        this->setOptional(sfBurnedNFTokens, newBurnedCount);
+    }
+};
+
+using AcctRootRd = AcctRootImpl<false>;
+using AcctRoot = AcctRootImpl<true>;
+
+// clang-format off
+#ifndef __INTELLISENSE__
+static_assert(not std::is_default_constructible_v<AcctRootRd>);
```
seelabs, line 359: Use `!` instead of `not`

src/ripple/protocol/AcctRoot.h
```diff
@@ -0,0 +1,377 @@
...
+static_assert(    std::is_copy_constructible_v<AcctRootRd>);
+static_assert(    std::is_move_constructible_v<AcctRootRd>);
+static_assert(    std::is_copy_assignable_v<AcctRootRd>);
+static_assert(    std::is_move_assignable_v<AcctRootRd>);
+static_assert(    std::is_nothrow_destructible_v<AcctRootRd>);
+
+static_assert(not std::is_default_constructible_v<AcctRoot>);
```
seelabs, line 366: Use `!` instead of `not`

src/ripple/protocol/Keylet.h
```diff
@@ -29,26 +30,95 @@ class STLedgerEntry;
...
+    ~KeyletBase() = default;
+
+public:
     /** Returns true if the SLE matches the type */
     bool
     check(STLedgerEntry const&) const;
 };
 
+#ifndef __INTELLISENSE__
+static_assert(not std::is_default_constructible_v<KeyletBase>);
```
seelabs, line 72: Use `!` instead of `not` (here and below)

src/ripple/protocol/Keylet.h
```diff
@@ -29,26 +30,95 @@ class STLedgerEntry;
...
+
+    using KeyletBase::check;
+
+    AccountRootKeylet(uint256 const& key) : KeyletBase(ltACCOUNT_ROOT, key)
+    {
+    }
+};
+
+#ifndef __INTELLISENSE__
+static_assert(not std::is_default_constructible_v<AccountRootKeylet>);
```
seelabs, line 114: Use `!` instead of `not`

seelabs (commented):
Nice job on this! 

Did a pass. I'll come back and do another pass later, but I wanted to get the current set of comments in now.

src/ripple/protocol/AcctRoot.h
seelabs: I just noticed the "this". I don't really object, but we usually don't put "this" in front of member functions. Is there a reason it's done here?

scottschurr (commented):
Current top-most commit addresses comments from @seelabs.  Thanks for the review!

src/ripple/app/tx/impl/Payment.cpp
```diff
@@ -329,31 +329,35 @@ Payment::doApply()
...
                                                              : 1};
 
         // Create the account.
-        sleDst = std::make_shared<SLE>(k);
+        SLE::pointer sleDst = std::make_shared<SLE>(k);
         sleDst->setAccountID(sfAccount, uDstAccountID);
         sleDst->setFieldU32(sfSequence, seqno);
 
         view().insert(sleDst);
+        acctRootDst = view().peek(k);
```
gregtatcam, line 346: Why is this check added/required here?

src/ripple/protocol/LedgerEntryWrapper.h
```diff
@@ -0,0 +1,194 @@
...
+    // There is no virtual table, so the most derived class must be destroyed.
+    ~LedgerEntryWrapper() = default;
+
+    //--------------------------------------------------------------------------
+    // Helper functions that are useful to some derived classes.
+    template <typename SF, typename T>
+    void
+    setOptional(SF const& field, T const& value)
+    {
+        static_assert(Writable, "Cannot set member of const ledger entry.");
```
gregtatcam, line 71: Perhaps a good place to use `requires` here and below?

gregtatcam (approved):
Nice job! I left some suggestions.

src/ripple/protocol/LedgerEntryWrapper.h
scottschurr: Good point.  I switched to using `requires` for consistency with AcctRoot.h.

src/ripple/protocol/LedgerEntryWrapper.h
```diff
@@ -0,0 +1,194 @@
...
+        static_assert(
```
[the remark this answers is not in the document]
scottschurr: I did change this to use `requires`.  But it shows that our `clang-format` rules don't handle `requires` clauses well.  The `clang-format` rules don't need to be fixed immediately, but they should be fixed.

scottschurr (commented):
The most recent commit addresses review comments from @gregtatcam.  Thanks for the review!

intelliot:
@seelabs : 

> Did a pass. I'll come back and do another pass later, but I wanted to get the current set of comments in now.

Are you still planning to come back and do another pass later?

scottschurr:
@intelliot, In addition to code reviews, there's an additional concern regarding this pull request.

I think this pull request starts us down a good road in terms of compile-time coding safety for transactors.  It also provides some scaffolding that is likely to be useful if we work on reducing the in-ledger storage size of particular node types.  However this pull request is only a start down the road.

In order to be truly effective a similar conversion to strong types needs to happen for most, if not all, in-ledger object types.  This includes at least:

- Offers,
- RippleStates (trust lines),
- Escrows,
- Tickets,
- SignerLists,
- PayChannels,
- Checks, and
- NFTokenOffers.

It may also include directory-like things, like

- DirectoryNodes and
- NFTokenPages.

In my opinion if we start down this road, then we need to commit to finishing the work.  See https://youtu.be/B2XtqVZcSdM?t=609  (that whole video is worth watching, by the way).

Let's not start down this path unless we're committed to finishing the refactor to include all of the appropriate ledger objects.  As David Sankel says in the linked video, "A partial change is destruction of value."

This beginning refactor shows that the approach can be successful.  I chose to start with AccountRoots because they are so ubiquitous and used in so many ways.  We now know that this approach works.  But if we start the refactor and don't finish it, then we are just adding a new mess to the code base.

So accepting this change into the code base means future resources _must_ be committed to completing the refactor.  If that commitment can't be made, then this branch should not be merged.

greg7mdp:
Very important point, @scottschurr ! And I really enjoyed the video too :-).

greg7mdp:
> a similar conversion to strong types needs to happen for most, if not all, in-ledger object types.

Also the ones not merged yet like the sidechain ones.

ximinez:
> So accepting this change into the code base means future resources _must_ be committed to completing the refactor. If that commitment can't be made, then this branch should not be merged.

How about merging to a feature branch similar to the one that was just created for nft-fixes? There will be maintenance, and a risk of bit rot in that feature branch, but it'll make it easier to forget the whole thing if we _don't_ get the follow through.

scottschurr:
@ximinez, working with a feature branch would be possible.  However I have concerns:

- Converting all of the ledger types will touch a _lot_ of stuff.  Review of the final branch will be close to impossible because so much will be touched.  Managing the bit rot will also be extremely challenging.
- Suppose we started on that branch, got 90% of the way through, and then gave up.  That's a huge amount of effort to just throw away.  Would we have the will to throw it out?  Or would we submit the mostly-completed work anyway?

All I'm saying is that if we don't have the will to complete this work then we should not merge just this first part.  I don't think that working on a feature branch avoids that decision.  That same decision is needed whether we merge this first part or continue the work using a feature branch.

intelliot:
@ximinez @scottschurr What's your opinion? Perhaps, after taking a step back and considering all the effort required (including reviews), we've learned that this refactor is not worthwhile at this time?

scottschurr:
From @intelliot:
> What's your opinion? Perhaps, after taking a step back and considering all the effort required (including reviews), we've learned that this refactor is not worthwhile at this time?

Personally, I think the effort is quite worthwhile.  If we can't do this kind of code maintenance then the code base will fall into disarray over time.  I think this is a worthwhile change.

However I'm not in charge of where people spend development time.  If big features and code reviews will soak up all the time then we should not embark on significant maintenance efforts like this one.

ximinez:
> @ximinez @scottschurr What's your opinion? Perhaps, after taking a step back and considering all the effort required (including reviews), we've learned that this refactor is not worthwhile at this time?

I concur with @scottschurr. Even though I haven't done a deep dive into this particular PR, the concept is good, and this is the kind of investment that will pay off with higher code quality for years to come.

intelliot:
1. need estimate of how much time this would take
2. need recommended prioritization vs OKRs and goals

intelliot:
- want the interface across all objects to be uniform. This PR only changes AcctRoot, but there are many other types of ledger objects.
- to have a consistent interface throughout the codebase, every type of ledger object would need to be changed.
  - 2 years ago, this was considered viable, but things have changed since then.
  - this is purely a refactor / maintenance with no effect on users - and is fairly expensive.
  - would be counterproductive to merge this PR without following through to update all other object types.
- potential end-user benefit: enabling future reduction in trust line size

@seelabs @ximinez to consider whether it makes sense to keep this alive - balancing the value of this work with other priorities

src/ripple/app/ledger/impl/LocalTxs.cpp
```diff
@@ -156,13 +156,12 @@ class LocalTxsImp : public LocalTxs
                 return true;
 
             AccountID const acctID = txn.getAccount();
-            auto const sleAcct = view.read(keylet::account(acctID));
+            auto const acctRoot = view.read(keylet::account(acctID));
```
ckeshava, line 159: As far as the organization of the code is concerned, isn't it cleaner to use a constructor from the `AccountRootKeylet` class instead of the existing `keylet` namespace? I feel all the `keylet::<<lt_construction_functions>>` could be moved to their respective Strongly typed classes

src/ripple/protocol/AcctRoot.h
```diff
@@ -0,0 +1,354 @@
...
+class AcctRootImpl final : public LedgerEntryWrapper<Writable>
+{
+private:
+    using Base = LedgerEntryWrapper<Writable>;
+    using SleT = typename Base::SleT;
+    using Base::wrapped_;
+
+    // This constructor is private so only the factory functions can
+    // construct an AcctRootImpl.
+    AcctRootImpl(std::shared_ptr<SleT>&& w) : Base(std::move(w))
```
ckeshava, line 39: when we construct a new `AccountRoot` ledger object, are we not validating the fields using `KnownFormat` ? 

I know that this is unlikely, but what if a developer incorrectly passes a `Offers` shared pointer to this constructor, the exception throwing mechanism is identical to the old code right?

src/ripple/protocol/AcctRoot.h
```diff
@@ -0,0 +1,354 @@
...
+        return *this;
+    }
+
+    ~AcctRootImpl() = default;
+
+    // Conversion operator from AcctRootImpl<true> to AcctRootImpl<false>.
+    operator AcctRootImpl<true>() const
+    {
+        return AcctRootImpl<false>(
+            std::const_pointer_cast<std::shared_ptr<STLedgerEntry const>>(
```
ckeshava, line 74: why isn't the type of the shared_ptr `AcctRootImpl` ? Wouldn't that provide a stronger type guarentee?

scottschurr:
As far as I can tell there are no active code reviews running.  So I've rebased this pull request in order to allow it to be built by conan.

src/ripple/app/misc/impl/TxQ.cpp
```diff
@@ -1097,8 +1091,8 @@ TxQ::apply(
             // Create the test view from the current view.
             multiTxn.emplace(view, flags);
 
-            auto const sleBump = multiTxn->applyView.peek(accountKey);
-            if (!sleBump)
+            auto acctRootBump = multiTxn->applyView.peek(accountKey);
```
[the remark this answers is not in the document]
ckeshava: ok, thanks for the clarification. 

In the existing code, `auto const sleBump` would prevent us from updating the control block of the shared_ptr, since it is marked as const, am I right? But that goes against the functionality of a shared pointer (incrementing and decrementing the number of references to the SLE object)

src/ripple/protocol/AcctRoot.h
ckeshava: okay, my bad. got it 👍

src/ripple/app/ledger/impl/LocalTxs.cpp
ckeshava: yes, thank you

src/ripple/protocol/Keylet.h
```diff
@@ -29,26 +30,95 @@ class STLedgerEntry;
...
+
+    // You should be able to assign derived classes.
+    KeyletBase&
+    operator=(KeyletBase const&) = default;
+    KeyletBase&
+    operator=(KeyletBase&&) = default;
+
+    // The destructor is protected so derived classes cannot be destroyed
+    // through a pointer to the base.  It also allows us to do derivation
+    // without needing a virtual table.
```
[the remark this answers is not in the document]
ckeshava: thank you for the explanation! i did not know that.

What do you mean by aggregation here?

src/ripple/app/misc/impl/TxQ.cpp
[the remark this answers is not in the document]
scottschurr: I'm pretty sure that with `auto const sleBump` the control block is still mutable.  It's the local representation of the `std::shared_ptr` that is `const`.  So you can't, for example, assign to it.  All you can do to the local representation is destroy it, which would have the effect of decrementing the reference count in the control block.

src/ripple/protocol/Keylet.h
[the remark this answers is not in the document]
scottschurr: By "aggregation" I mean making something a member (as opposed to inheriting from it).  The terms "aggregation" and "composition" are kinda interchangeable in this context.

Do a Google search of inheritance vs aggregation or inheritance vs composition to get a sense of the benefits and disadvantages of each approach.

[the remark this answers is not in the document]
ckeshava: okay, I am familiar with the term composition 👍

src/ripple/protocol/AcctRoot.h
```diff
@@ -0,0 +1,367 @@
...
+    // as a friend unless the class declaration is visible at this point.
+    friend class ReadView;
+    friend class ApplyView;
+
+public:
+    AcctRootImpl(AcctRootImpl const&) = default;
+    AcctRootImpl(AcctRootImpl&&) = default;
+
+    AcctRootImpl&
+    operator=(AcctRootImpl const& rhs)
```
ckeshava, line 55: Why is the behavior of copy constructor different from that of copy-assignment operator? If we had resorted to the compiler generated default for copy-assignment, wouldn't it do exactly the same thing?

scottschurr: Good spotting!  Yes, I think you're right that the copy and move assignment operators should use `= default` in place of the explicit implementation.  Good catch.

scottschurr: Excellent point, @greg7mdp.  The most recent commit applies the Rule of 0 to the `AcctRoot` wrapper.  That's particularly important because that wrapper class will be used as an example for other ledger entry wrapper implementations.  It will reduce boiler plate in all of the wrappers.

scottschurr: @ckeshava, as a heads up, you should be aware that I've modified the `AcctRoot` implementation as @greg7mdp suggested.  And I'll explicitly say Thank You to both you and @greg7mdp for the improvements.

scottschurr:
@ckeshava identified some places that the `AcctRoot` was not being used but could be.  The most recent commit incorporates @ckeshava's suggested changes.

scottschurr:
The most recent commit is a change suggested by @ckeshava.

thejohnfreeman:
I've got an [alternative design](https://github.com/thejohnfreeman/rippled/tree/acctroot) in branch `acctroot` in my fork. Here's what I like about it:

`ReadView::read`, `ApplyView::peek`, and their callers are dealing directly in `std::shared_ptr`s again. No `optional<AcctRootRd>`. No `Writable` template parameter to approximate constness. Just familiar const and non-const. `AcctRootImpl` inherits directly from `STLedgerEntry`. No `LedgerEntryWrapper`. Helper methods are moved to `STLedgerEntry`. I believe that modulo renamed variables, this design has fewer differences with the starting code, meaning fewer changes are needed to adapt it.

I don't like that construction of `STLedgerEntry`s is handled through a virtual method of `KeyletBase`. That is not required, though, just the fastest way I could get to this proof-of-concept. I'd like to see the keylet class hierarchy rearranged:

- Rename `KeyletBase` back to `Keylet`. Rename `Keylet` to something like `GenericKeylet`.
- Overload `ReadView::read` and `ApplyView::peek` for `GenericKeylet` vs what will soon be a growing set of strongly-typed keylets, including `AccounttRootKeylet`[^1].
- Make `read(GenericKeylet)` return a `std::shared_ptr<SLE>` while all the other overloads can return `std::shared_ptr`s to derived types.

Builds and all tests pass on my machine. CI is still in-progress as I write this comment.

[^1]: Why is it named `AccountRootKeylet` but `AcctRootImpl`, `AcctRoot`, and `AcctRootRd`? I would like "Account" spelled out in every name.

scottschurr:
I looked over @TheJohnFreeman's alternative design.  I spotted what I consider to be a big flaw and a smaller issue.  

1. Each wrapper _is_ an `STLedgerEntry`.  So the wrapper no longer makes it difficult to use the `STObject` interface.  Each wrapper still leaves that interface fully available.  So a good portion of the error prevention that the wrappers provide evaporates.  The safer interface provided by the wrappers is available, but it is not encouraged.  This makes the wrappers kind of pointless from my perspective.

2. `Ledger::readSLE()` and `Ledger::peekSLE()` return a `shared_ptr<SLE>`, and `SLE` is the base class.  This forces ReadView::read() and ApplyView::peek() to do downcasts.  Not the end of the world, but not desirable.

So the next question is what can be done about these objections.

- Objection 1 could be addressed by using private inheritance.  But that removes the availability of the public interfaces that were moved from `LedgerEntryWrapper` into `STLedgerEntry`.  Also (I think) using private inheritance makes it harder to return an `STLedgerEntry` from one of the wrappers.  This is a back door that is necessary some times.

- Objection 1 could also be addressed by going back to using aggregation, rather than inheritance.  This would be much like the old wrappers, but rather than wrapping a `std::shared_ptr<SLE>` the wrapper would just wrap a `SLE`.  But then `Ledger::readSLE()` and `Ledger::peekSLE()` would need to return different types depending on the `Keylet`.  That problem might be solvable by adding to `Ledger`...
```
    template <class K, typename T = typename K::sle_type>
    auto
    typedPeek(K const& keylet) -> std::shared_ptr<T>
```
and similarly for `typedRead()`.  I'm not sure.  I haven't tried it.  And you can't call them just `peek()`/`read()`, because that would hide names.  But providing `Ledger::typedPeek` and `Ledger::typedRead`  would also address Objection 2 (assuming it works).

Or folks could say that Objection 1 is not worth worrying about.  Personally, I think it's a very serious drawback, but mine is not the only voice.

If folks agree that Objection 1 is worth addressing then I could make a run at implementing the second of the two approaches described above.  If that doesn't pan out then I'd be leaning toward the original proposal, `template<bool>` and all.  Or if that's too abhorrent we could just drop the whole idea.

Regarding the `Keylet` renames, I don't really see the point?  A `Keylet` _is_ a `GenericKeylet`.  We would have to go through and rename all of the current uses of `Keylet` to `GenericKeylet`.  I don't feel like that really clarifies anything.  Yes, the number of uses of `Keylet` will reduce over time.  If the number of uses gets small enough in the future then we could consider the rename at that point?  But it's not an over-my-dead-body kind of a thing.  If that renaming makes more sense to you (and others) then I'm okay with it.

Regarding the `AcctRoot` name for the wrapper, I have no problems with fully spelling out `Account`.  However something I've learned from watching @ckeshava implement more wrappers is that some of the wrappers are hard to name because the code base has already taken the good name.  So I'm thinking that all of the wrappers might want to be named things like `AccountLedgerObj`, `TicketLedgerObj`, `EscrowLedgerObj` and the like.  That will give the various typed objects a consistent naming.  Or, I suppose, another approach would be to put them all in their own name space.  So it would be, for example, `ledgerObj::Account`.

Thoughts?  Opinions?

thejohnfreeman:
Objection 1 does not require private inheritance. It can be addressed by making specific methods private in the derived classes with using-declarations. Or you can have an intermediate base class that makes them protected. Or make them protected in `STObject`. Personally, I don't share this objection. I consider the primary benefit to be the convenience of the added strongly-typed specifically-named methods, an attractive carrot that sells itself, not the hiding of the generic methods.

For objection 2, the choices are either (a) downcast or (b) construct a wrapper. Both yield the same code when optimized (a no-op), but (a) doesn't have the same risk of calling a copy constructor when unoptimized, nor does it require an extra class that brings several problems (e.g. special methods, hiding methods by default, the maintenance cost of more code). I much prefer (a) to (b).

If renaming `Keylet` to `GenericKeylet` is not worthwhile, then why was `Keylet` renamed to `KeyletBase`? That was the original unworthwhile sin in my opinion. Now `KeyletBase` is littered in many more places in the code than `Keylet`, and after everything is migrated, we'll be stuck with `KeyletBase` and no `Keylet`. Why? It's ugly in my opinion.

`Obj` suffix or namespace prefix are both means of disambiguating names by making them longer. In these scenarios, I prefer namespaces because they can be optionally removed or abbreviated with using-declarations.

scottschurr:
> Objection 1 does not require private inheritance. It can be addressed by making specific methods private in the derived classes with using-declarations. Or you can have an intermediate base class that makes them protected. Or make them protected in `STObject`.

I suspect I'm not following your meaning.  I'm reading this as though you are suggesting that things that are `public` in a base class should be hidden (through whatever means) in a publicly derived class?  This seems particularly unwise to me.  It's easy to subvert the hiding by simply upcasting.  It means that the public inheritance does not model is-a.  The ISO C++ FAQ also thinks this is a very bad idea: https://isocpp.org/wiki/faq/proper-inheritance#hiding-inherited-public.  But I'm probably misunderstanding your meaning.

Another alternative you suggest is making some currently public methods in `STObject` protected.  I don't think the wrappers are worth anywhere close to the cost of dealing with that breakage.  I prefer to leave `STObject` alone if at all possible, at least with regard to this particular change.

> Personally, I don't share this objection. I consider the primary benefit to be the convenience of the added strongly-typed specifically-named methods, an attractive carrot that sells itself, not the hiding of the generic methods.

I hear this.  You and I simply have different experiences with human nature.  The `SField` interfaces in `STObject` have been in use for over 10 years now.  I don't think that simply providing a specifically-named method will overcome the habit of reaching immediately for an `SField`.  In my opinion the wrappers need to both...

1. Provide a good interface and
2. Actively discourage using what we think is the less good interface.

Otherwise we'll see mixed usage and be even worse off than we were before the wrappers.

Just my opinion.

> For objection 2, the choices are either (a) downcast or (b) construct a wrapper. Both yield the same code when optimized (a no-op), but (a) doesn't have the same risk of calling a copy constructor when unoptimized, nor does it require an extra class that brings several problems (e.g. special methods, hiding methods by default, the maintenance cost of more code). I much prefer (a) to (b).

My personal view is that publicly deriving the wrappers from `STLedgerEntry` will result in a mess.  My reasons for that opinion are given above.

> If renaming `Keylet` to `GenericKeylet` is not worthwhile, then why was `Keylet` renamed to `KeyletBase`?

`Keylet` was not renamed to `KeyletBase`.  There used to be one name: `Keylet`.  We added one typed keylet: `AccountRootKeylet`.  So we now have two names: `Keylet` (the generic) and `AccountRootKeylet` (specific).  There are some characteristics that the generic `Keylet` and all of the typed keylet(s) will share.  It makes sense to push those shared characteristics into a base class: `KeyletBase`.

> That was the original unworthwhile sin in my opinion. Now `KeyletBase` is littered in many more places in the code than `Keylet`, and after everything is migrated, we'll be stuck with `KeyletBase` and no `Keylet`. Why? It's ugly in my opinion.

Did you count them?  In the original pull request there are 59 instances of `KeyletBase`.  There are 13 instances of `AccountRootKeylet`.  There are 207 instances of just `Keylet` (without "Base" or "AccountRoot").  It sounds like you are suggesting that it would be advantageous to rename 207 instances of `Keylet` to `GenericKeylet` so we can replace the 59 instances of `KeyletBase` with `Keylet`.

For now I would prefer the leave the 207 instances of `Keylet` and 59 instances of `KeyletBase`.  As more wrappers are introduced it may eventually get to the point where occurrences of `KeyletBase` outnumber occurrences of `Keylet`.  I suggest we wait and consider the rename you are suggesting if that time comes.

I clearly don't evaluate "sin", "litter", and "ugly" the same way you do.  I am not convinced those words encourage a productive conversation.  But you do you.

> `Obj` suffix or namespace prefix are both means of disambiguating names by making them longer. In these scenarios, I prefer namespaces because they can be optionally removed or abbreviated with using-declarations.

Sure.  I'm fine with using namespaces to corral the wrappers.  But it seems like that is the least of our disagreements.

thejohnfreeman:
You are correctly understanding my meaning. I don't treat the ISO C++ FAQ as dogma, but I partly agree with the sentiment, in that (as I said before) I don't see primary value in hiding the generic public `STObject` methods. That is your desire, not mine, and I was offering a means to accommodate it that you had not mentioned. Upcasting to circumvent access controls sounds like a very clear expression of intent to accept the risks and responsibilities of calling the generic methods.

Access controls are a game we choose to play with the compiler. With the wrapper implementation, "it's easy to subvert the hiding by simply" declaring a friend. It is as simple as upcasting. If you feel like the one is not worth worrying about, for whatever reason, then why not both?

I prefer to leave `STObject` alone as well. It is still an option worth stating.

There are very few people with the "habit of reaching immediately for an `SField`". I think we can name them all, and I am betting they will all prefer the specifically-named methods.

It sounds like you really want protected inheritance then. I don't know why you are worried "that removes the availability of the public interfaces that were moved from `LedgerEntryWrapper` into `STLedgerEntry`". They can just be moved into another intermediate class that protected inherits from `STLedgerEntry`.

Your other concern was it "makes it harder to return an `STLedgerEntry` from one of the wrappers". How? You can do it with a member method in the exact same way. Here is a Godbolt demonstrating both [workarounds](https://godbolt.org/z/bcEz46fzq):

```cpp
class STLedgerEntry {
public:
    int get();
};
using SLE = STLedgerEntry;

class TypedEntry : protected STLedgerEntry {
public:
    SLE* sle() {
        return this;
    }
};

class AccountRoot : public TypedEntry {
public:
    int id() {
        return get();
    }
};

int main() {
    AccountRoot a;
    //a.get(); // error: inaccessible
    a.id();
    //SLE* s1 = &a; // error: inaccessible
    SLE* s2 = a.sle();
    s2->get();
    return 0;
}
```

ckeshava:
I have a few questions with respect to @thejohnfreeman 's proposed idea.
1. Since you are not using aggregation to store a `std::shared_ptr<SLE>`, does your code have significant performance benefits? I'm not sure how prevalently these ledger objects are passed around in the codebase.

2. I share @scottschurr's Objection-1. If we return a `std::shared_ptr<SLE>` from the `read` and `peek` functions, what is the benefit of this Ledger Entry Wrapper? 
"I consider the primary benefit to be the convenience of the added strongly-typed specifically-named methods, an attractive carrot that sells itself, not the hiding of the generic methods." -> The named methods are identical in `STObject` and also the proposed Strongly-typed wrapper class. What do you mean by "specifically-named" methods?

The end goal of strongly-typed ledger objects appears to be points 1-3 (Source: https://gist.github.com/scottschurr/c637b5bd04ac09157763bfb6e74d28ac)
image
If we return a `std::shared_ptr` from `read` and `peek` functions, I'm not sure how we can work towards these stated goals. I'm aware that the proposed design by Scott Schurr and Scott Determan doesn't yet meet all these goals, but I can see a path going forward. But I couldn't quite get that from your proposed approach.

3. I feel `GenericKeylet` is just as informative as `KeyletBase`. I don't have a strong preference for either pair of proposed names.

4. I didn't understand @scottschurr's Objection-2. What do you mean by upcasting or downcasting? With respect to the later comments, I think it's a bad design to hide/privatise functions in child classes when `STObject` itself publicly exposes many generic functions. I'd instead prefer to change access controls in `STObject` itself in a separate set of changes.

I like that you are deriving the different ledger objects from `STLedgerEntry` instead of handling the different ledger entry types at the `Keylet` level.

scottschurr:
At this point I think it's appropriate to back up and reflect on where this pull request came from.  At the time this code was initially developed, about June of 2021, the work was considered a priority.  I felt like @seelabs and I had a good design.  Then NFToken development took precedent, so this work was put aside.

In August of 2022 I resurrected the work, even though it was no longer a priority.  I felt like the design was in a good place and it would be a shame to lose the effort.

Here we are in June of 2023.  The change being introduced is still not a high priority.  The design is being called into question.  I think that, with effort, the design differences could be settled out.  But, given that this work is not high priority, I think it's better to put that effort elsewhere, rather than into the redesign of this change.

So, given all that, I think the best thing to do is close this pull request.  @intelliot, what do you think?

ckeshava:
I concur with @scottschurr . 

We still haven't factored in the design effort required for the directory node ledger objects. We need to adapt the design slightly in order to accomodate those objects.

intelliot:
Given the benefits to these changes aren't crystal clear, I'm happy to go with @scottschurr and @ckeshava 's recommendation. If you feel that we've learned enough for now and no code changes to `rippled` are necessary, then go ahead and close this PR.

scottschurr:
Closing due to a mismatch between the apparent value of the change and the cost of resolving design disagreements.

PR #4278, opened 2022-08-19
Pin reporting mode dependency versions

legleux:
### Type of Change
Pins reporting mode's dependencies to fixed versions
- [x] Bug fix (non-breaking change which fixes an issue)

PR #4281, opened 2022-08-22
Build packages on release candidate and beta branches

legleux:

Build rippled packages on release candidate and beta branches, all pull-requests and manually to make sure the packaging is not broken.

Runs `rippled --unit-test` after.

~This build will fail as https://github.com/XRPLF/rippled/pull/4278 is required for `rippled-reporting` package builds to succeed.~ Update: #4278 has been merged and released in 1.9.4.

[Successful run](https://github.com/legleux/rippled/actions/runs/2905883005)

### Type of Change


.github/workflows/build_packages.yml
```diff
@@ -0,0 +1,72 @@
...
+              script: dpkg
+            - suffix: rpm
+              image: rippleci/rippled-rpm-builder:2020-02-10
+              script: rpm
+      container:
+        image: ${{ matrix.type.image }}
+
+      steps:
+        - name: Checkout
+          uses: actions/checkout@v3
```
thejohnfreeman, line 27: This step is unnecessary because it is already in the `package` action, right?

.github/actions/package/action.yml
```diff
@@ -0,0 +1,32 @@
+name: 'Package rippled'
```
thejohnfreeman, line 1: If this action is only used in one place, is it simpler to just inline it?

thejohnfreeman (changes requested):

.github/actions/package/action.yml
legleux: Having this be an action is overkill but I was just trying it out and anticipating removing this code from the repo.

.github/workflows/build_packages.yml
legleux: Checkout can be removed from the action since it's in the workflow.

intelliot:
@legleux what's the status of this? Is this ready to merge?

legleux:
> @legleux what's the status of this? Is this ready to merge?

This will need to be modified with the imminent ubuntu update so might as well wait for that

intelliot:
I think this PR is no longer needed. @legleux can you confirm?

legleux:
> I think this PR is no longer needed. @legleux can you confirm?

This will have to be implemented some other way since our github actions have changed a bit since this was approved.

PR #4295, opened 2022-09-07
Eliminate the use of the empty struct Coro_create_t

ckeshava:
This commit seeks to eliminate the use of the empty struct Coro_create_t.

To my knowledge, empty structs are used for tag dispatch or method coloring. Both of these use cases are irrelevant here because there is only one constructor in the Coro class. Furthermore, no access-restriction (private or protected) was placed on Coro_create_t and hence it was publicly usable.

- [ x ] Refactor (non-breaking change that only restructures code)

### DIFF
--- a/src/ripple/core/JobQueue.h
+++ b/src/ripple/core/JobQueue.h
@@ -37,10 +37,6 @@ class PerfLog;
 }
 
 class Logs;
-struct Coro_create_t
-{
-    explicit Coro_create_t() = default;
-};
 
 /** A pool of threads to perform work.
 
@@ -76,7 +72,7 @@ public:
     public:
         // Private: Used in the implementation
         template <class F>
-        Coro(Coro_create_t, JobQueue&, JobType, std::string const&, F&&);
+        Coro(JobQueue&, JobType, std::string const&, F&&);
 
         // Not copy-constructible or assignable
         Coro(Coro const&) = delete;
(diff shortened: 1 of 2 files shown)


nbougalis (commented):
This code was meant, more than anything, the accidental instantiation of `Coro` for what was, primarily, intended to be an internal `JobQueue` interface.

One way to achieve that would be to make `JobQueue` a `friend` of `Coro` but the problem is that it's not `JobQueue` that constructs the `Coro` but, rather, `std::make_shared` on behalf of `JobQueue`.

Another way to achieve this might be to mark the `Coro` constructor as `protected` (and remove the `Coro_create_t` part) and mark its destructor as `virtual`. Then, in `JobQueue::postCoro` we could do this:
 
```
    struct ConstructibleCoro : public Coro
    {
        using Coro::Coro;
    };

    std::shared_ptr<Coro> coro = std::make_shared<ConstructibleCoro>(
        *this, t, name, std::forward<F>(f));

    // ... the rest of the code is the same
```

Not sure that it's worth the extra bit of complexity; curious to hear what @scottschurr and @HowardHinnant think.

Also, please fix the comment there https://github.com/XRPLF/rippled/blob/develop/src/ripple/core/JobQueue.h#L413-L416:

```
JobQueue::postCoro(JobType t, std::string const& name, F&& f)
{
    /*  First param is a detail type to make construction private.
        Last param is the function the coroutine runs. Signature of
        void(std::shared_ptr<Coro>).
    */
```

src/ripple/core/JobQueue.h
```diff
@@ -76,7 +72,7 @@ class JobQueue : private Workers::Callback
     public:
         // Private: Used in the implementation
         template <class F>
-        Coro(Coro_create_t, JobQueue&, JobType, std::string const&, F&&);
+        Coro(JobQueue&, JobType, std::string const&, F&&);
```
nbougalis, line 75: Recommend making this `private` and adding `friend class JobQueue` on line 57:

```
    class Coro : public std::enable_shared_from_this<Coro>
    {
        friend class JobQueue;
    
    private:
        ... variables ...

        Coro(JobQueue&, JobType, std::string const&, F&&);
```

intelliot:
@ckeshava does this change still make sense to apply? If not, feel free to close it for now.

nbougalis:
An improved version of this commit has been included in #4172. This can be closed.

PR #4298, opened 2022-09-09
[TRIVIAL] Allow gcc 12 compilation

seelabs:
Compiling with gcc 12 on manjaro (arch variant) had compilation errors without adding an additional include file.


nbougalis (approved):

intelliot (approved):

intelliot:
Already merged in 1.9.4 https://github.com/XRPLF/rippled/pull/4302

PR #4299, opened 2022-09-09
Remove const_cast usage in BookChanges

godexsoft:

Removing an unnecessary `const_cast`.

### Context of Change

While porting `book_changes` to `clio` a couple of `const_cast`s were discovered. 
It's entirely possible to do the same `downcast` without `const_cast` so here we are.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### DIFF
--- a/src/ripple/rpc/BookChanges.h
+++ b/src/ripple/rpc/BookChanges.h
@@ -88,13 +88,10 @@ computeBookChanges(std::shared_ptr<L const> const& lpAccepted)
                 !node.isFieldPresent(sfPreviousFields))
                 continue;
 
-            STObject& finalFields = (const_cast<STObject&>(node))
-                                        .getField(sfFinalFields)
-                                        .downcast<STObject>();
-
-            STObject& previousFields = (const_cast<STObject&>(node))
-                                           .getField(sfPreviousFields)
-                                           .downcast<STObject>();
+            auto const& ffBase = node.peekAtField(sfFinalFields);
+            auto const& finalFields = ffBase.template downcast<STObject>();
+            auto const& pfBase = node.peekAtField(sfPreviousFields);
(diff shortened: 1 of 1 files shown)


src/ripple/rpc/BookChanges.h
```diff
@@ -88,13 +88,10 @@ computeBookChanges(std::shared_ptr<L const> const& lpAccepted)
...
                 continue;
 
-            STObject& finalFields = (const_cast<STObject&>(node))
-                                        .getField(sfFinalFields)
-                                        .downcast<STObject>();
-
-            STObject& previousFields = (const_cast<STObject&>(node))
-                                           .getField(sfPreviousFields)
-                                           .downcast<STObject>();
+            auto const& ffBase = node.peekAtField(sfFinalFields);
```
seelabs, line 91: I don't really care, but I probably would have left off the intermediate `ffBase` and `pfBase` and just done something like:
```
            auto const& finalFields =
                node.peekAtField(sfFinalFields).template downcast<STObject>();
```
It's fine as-is too.

seelabs (approved):
Thanks for getting rid of the const cast. :+1: LGTM

src/ripple/rpc/BookChanges.h
godexsoft: That was my initial code and `clang-format` was producing this:
```
            auto const& finalFields =
                node.peekAtField(sfFinalFields).template downcast<STObject>();
            auto const& previousFields = node.peekAtField(sfPreviousFields)
                                             .template downcast<STObject>();
``` 
While this is slightly shorter - it looks so bad that i would probably not sleep tonight if i did this :smiley: 
That's how it ended up being what it is. As the compiler generally does a great job optimizing the intermediate away (generated assembly is the same) I thought it's better be readable than short - I believe there is no harm done :+1:

Issue #4300, opened 2022-09-11
Trustline flag set on NFTokenMint TX's Can Max a Minters Reserve and Create Infinite Issued Currencies with No Reserve (Version: 1.9.2)


x-Tokenize:
---
## About: 
NFTs Utilizing Flag 12/13 on NFTokenMint Tx's with a Transfer Fee >0 are susceptible to spam trust lines and an increase in their reserve requirement while also allowing a currency issuer to create infinite currencies for no cost or reserve requirement.
## labels: 
NFTokenMint, Trustline Spam, neglibile reserve requirement

---

## Issue Description
NFTs minted with the trustline & transferable flag enabled with a TransferFee > 0 are susceptible to attack by a malicious user which allows the attacker to create an infinite number of currencies on their attacking issuing account and maxes out the victims reserve requirement on the  NFT account. At conception, this attack was very expensive to execute as an attacker but after optimization in the attacking method, it became a negligible cost and allowed the Currency issuer to issue assets without affecting their reserve requirement.

### To reproduce, there are 4 accounts involved in the attack:
Victim Account: Issues an NFT, that is transferable, auto-creates the trustline and has a transfer fee > 0\
Attacking Account A: Will be setting trustlines, creating and accepting NFT offers and sending back token to issuer.
Attacking Account B: Same as Account A
Attacking Currency Issuer (Account C): Issues Currencies to Account A and Account B

Attack sequence:
1) Account A acquires a vulnerable NFT
2) Account B creates a trustline to Account C for random currency 'xyz'
3) Account C sends 1 'xyz' to Account B
4) Account A creates a sell offer for NFT with destination set to Account B for 1 'xyz'
5) Account B Accepts the sell offer from Account A (Account A Gets (1 'xyz'- (transferFee*1 'xyz')) and Victim gets (transferFee * 1 'xyz')) 
6) Account B removes the trustline for 'xyz'
7) Account A Sends the received amount back to the issuer (this results in clearing reserve requirement for Account C)
8) Repeat the process starting at step 2 with reversing the roles of Account A and Account B

## Expected Result
The Attacker would be able to clog up the reserve requirement of the NFT issuer coming with a high cost to the attacker.

## Actual Result
With the reserve requirement also becoming negligible, it allowed Account C to issue an infinite amount of currencies and max out the vulnerable NFT Minters reserve with a negligible cost to the attacker (only transaction fees).

## Environment
Tested on rippled 1.9.2

## Supporting
[...]

PR #4301, opened 2022-09-13
Introduce fixRemoveNFTokenAutoTrustLine amendment

scottschurr:

It turns out that the feature enabled by the `tfTrustLine` flag on an `NFTokenMint` transaction could be used as a means to attack the NFToken issuer.  Details are in https://github.com/XRPLF/rippled/issues/4300

The `fixRemoveNFTokenAutoTrustLine` amendment in this pull request removes the ability to set the `tfTrustLine` flag on an `NFTokenMint` transaction.

Closes https://github.com/XRPLF/rippled/issues/4300

### Context of Change

There are two commits in this pull request.  They are both needed, but they should be kept separate (not squashed).

- The first commit, by @HowardHinnant, is required to allow the code to be compiled without warnings by the latest version of Xcode (Version 13.4.1.  This commit has already been accepted into the 1.10.0-b1 develop branch.  This first commit does not require review.

- The second commit disables the `tfTrustLine` flag on an NFTokenMint transaction.  This commit needs review.

### Type of Change

- [x] Amendment (amendment going live will cause amendment blocking of out-of-date servers)
- [x] Tests (tests were added for the new amendment)
- [x] Requires documentation updates

### DIFF
--- a/src/ripple/app/tx/impl/NFTokenMint.cpp
+++ b/src/ripple/app/tx/impl/NFTokenMint.cpp
@@ -53,10 +53,10 @@ NFTokenMint::preflight(PreflightContext const& ctx)
     // The fixRemoveNFTokenAutoTrustLine amendment disables minting with the
     // tfTrustLine flag as a way to prevent the attack.  But until the
     // amendment passes we still need to keep the old behavior available.
-    std::uint32_t const NFTokeMintMask =
+    std::uint32_t const NFTokenMintMask =
         ctx.rules.enabled(fixRemoveNFTokenAutoTrustLine) ? tfNFTokenMintMask
                                                          : tfNFTokenMintOldMask;
-    if (ctx.tx.getFlags() & NFTokeMintMask)
+    if (ctx.tx.getFlags() & NFTokenMintMask)
         return temINVALID_FLAG;
 
     if (auto const f = ctx.tx[~sfTransferFee])
--- a/src/ripple/app/tx/impl/NFTokenMint.cpp
+++ b/src/ripple/app/tx/impl/NFTokenMint.cpp
@@ -40,7 +40,23 @@ NFTokenMint::preflight(PreflightContext const& ctx)
     if (auto const ret = preflight1(ctx); !isTesSuccess(ret))
         return ret;
 
-    if (ctx.tx.getFlags() & tfNFTokenMintMask)
+    // Prior to fixRemoveNFTokenAutoTrustLine, transfer of an NFToken between
+    // accounts allowed a TrustLine to be added to the issuer of that token
+    // without explicit permission from that issuer.  This was enabled by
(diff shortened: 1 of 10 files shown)


src/ripple/app/tx/impl/NFTokenMint.cpp
```diff
@@ -40,7 +40,23 @@ NFTokenMint::preflight(PreflightContext const& ctx)
...
+    // building up any number of TrustLines on the issuer, increasing the
+    // issuer's reserve without bound.
+    //
+    // The fixRemoveNFTokenAutoTrustLine amendment disables minting with the
+    // tfTrustLine flag as a way to prevent the attack.  But until the
+    // amendment passes we still need to keep the old behavior available.
+    std::uint32_t const NFTokeMintMask =
+        ctx.rules.enabled(fixRemoveNFTokenAutoTrustLine) ? tfNFTokenMintMask
+                                                         : tfNFTokenMintOldMask;
+    if (ctx.tx.getFlags() & NFTokeMintMask)
```
ximinez, line 59: Nit: Should this variable be spelled `NFTokenMintMask`?

ximinez (changes requested):

src/ripple/app/tx/impl/NFTokenMint.cpp
```diff
@@ -40,7 +40,23 @@ NFTokenMint::preflight(PreflightContext const& ctx)
...
+    //
+    // That capability could be used to attack the NFToken issuer.  It
+    // would be possible for two accounts to trade the NFToken back and forth
+    // building up any number of TrustLines on the issuer, increasing the
+    // issuer's reserve without bound.
+    //
+    // The fixRemoveNFTokenAutoTrustLine amendment disables minting with the
+    // tfTrustLine flag as a way to prevent the attack.  But until the
+    // amendment passes we still need to keep the old behavior available.
+    std::uint32_t const NFTokeMintMask =
```
nbougalis, line 56: Minor typo here `NFTokeMintMask` is missing an `n`.

src/ripple/app/tx/impl/NFTokenMint.cpp
scottschurr: 😀 🚬 Good catch.  Yes, that's quite the misspelling.

nbougalis (approved):
Looks good, just a small typo.

src/ripple/app/tx/impl/NFTokenMint.cpp
gadget78: its also missing on line 59 too ...

scottschurr:
Typo fixed.  Thanks for catching that.

intelliot:
[Released in version 1.9.4](https://xrpl.org/blog/2022/rippled-1.9.4.html)

PR #4304, opened 2022-09-16
Simplify the Transaction class

a-noni-mousse:

It simplifies the Transaction class and removes unnecessary exception code and variables

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### DIFF
--- a/src/ripple/rpc/handlers/AccountTx.cpp
+++ b/src/ripple/rpc/handlers/AccountTx.cpp
@@ -335,7 +335,7 @@ populateJsonResponse(
                         insertDeliveredAmount(
                             jvObj[jss::meta], context, txn, *txnMeta);
                         insertNFTSyntheticInJson(
-                            jvObj, context, txn->getSTransaction(), *txnMeta);
+                            jvObj, context, txn->getSerializedTx(), *txnMeta);
                     }
                 }
             }


src/ripple/app/misc/Transaction.h
```diff
@@ -66,79 +73,63 @@ class Transaction : public std::enable_shared_from_this<Transaction>,
...
-    // boost::optional (not std::optional) parameters.
-    static TransStatus
-    sqlTransactionStatus(boost::optional<std::string> const& status);
+        std::shared_ptr<STTx const> tx,
+        TransStatus status = NEW,
+        std::uint32_t ledgerSeq = 0) noexcept;
 
     std::shared_ptr<STTx const> const&
-    getSTransaction()
+    getSerializedTx() const noexcept
```
ximinez, line 81: Since this returns an `STTx` ptr, why not call the function `getSTTx()` or just `getTx()`?

src/ripple/app/rdb/backend/detail/impl/Node.cpp
```diff
@@ -617,6 +617,25 @@ getHashesByIndex(
     return res;
 }
 
+// The boost::optional parameter is used because the SOCI API requires it.
+static Transaction::pointer
+transactionFromSQL(
+    boost::optional<std::uint64_t> const& ledgerSeq,
+    boost::optional<std::string> const& status,
+    Blob const& rawTxn,
+    Application& app)
```
ximinez, line 626: You can safely remove the `app` parameter.

src/ripple/rpc/handlers/Submit.cpp
```diff
@@ -117,12 +117,11 @@ doSubmit(RPC::JsonContext& context)
...
-    auto tpTrans = std::make_shared<Transaction>(stpTrans, reason, context.app);
+    auto tpTrans = std::make_shared<Transaction>(stpTrans);
     if (tpTrans->getStatus() != NEW)
     {
         jvResult[jss::error] = "invalidTransaction";
-        jvResult[jss::error_exception] = "fails local checks: " + reason;
+        jvResult[jss::error_exception] = "fails local checks";
 
         return jvResult;
     }
```
ximinez, line 127: This block should be impossible to reach now. It can probably be safely replaced with `assert(tpTrans->getStatus() == NEW);`

src/ripple/rpc/handlers/Submit.cpp
```diff
@@ -246,13 +246,11 @@ doSubmitGrpc(
...
-    auto tpTrans = std::make_shared<Transaction>(stpTrans, reason, context.app);
+    auto tpTrans = std::make_shared<Transaction>(stpTrans);
     if (tpTrans->getStatus() != NEW)
     {
         grpc::Status errorStatus{
-            grpc::StatusCode::INVALID_ARGUMENT,
-            "invalid transaction: " + reason};
+            grpc::StatusCode::INVALID_ARGUMENT, "invalid transaction"};
         return {result, errorStatus};
     }
```
ximinez, line 255: This block should be impossible to reach now. It can probably be safely replaced with `assert(tpTrans->getStatus() == NEW);`

src/ripple/rpc/impl/TransactionSign.cpp
```diff
@@ -561,12 +561,11 @@ transactionConstructImpl(
...
-        tpTrans = std::make_shared<Transaction>(stpTrans, reason, app);
+        tpTrans = std::make_shared<Transaction>(stpTrans);
         if (tpTrans->getStatus() != NEW)
         {
-            ret.first = RPC::make_error(
-                rpcINTERNAL, "Unable to construct transaction: " + reason);
+            ret.first =
+                RPC::make_error(rpcINTERNAL, "Unable to construct transaction");
             return ret;
         }
```
ximinez, line 570: This block should be impossible to reach now. It can probably be safely replaced with `assert(tpTrans->getStatus() == NEW);`

gregtatcam (changes requested):

src/ripple/app/rdb/backend/detail/impl/Node.cpp
```diff
@@ -656,8 +674,7 @@ getTxHistory(
             else
                 rawTxn.clear();
 
-            if (auto trans = Transaction::transactionFromSQL(
-                    ledgerSeq, status, rawTxn, app))
+            if (auto trans = transactionFromSQL(ledgerSeq, status, rawTxn, app))
```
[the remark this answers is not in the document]
HowardHinnant: Ditto:
```
src/ripple/app/rdb/backend/detail/impl/Node.cpp:677:30: error: no matching function for call to 'transactionFromSQL'
            if (auto trans = transactionFromSQL(ledgerSeq, status, rawTxn, app))
                             ^~~~~~~~~~~~~~~~~~
/Users/howardhinnant/Development/rippled/src/ripple/app/rdb/backend/detail/impl/Node.cpp:622:1: note: candidate function not viable: requires 3 arguments, but 4 were provided
transactionFromSQL(
^
```

HowardHinnant (changes requested):

nbougalis:
I like this. If possible, consider whether the `TransStatus` and `TxnSql` enumerations can be combined. I _think_ that `TransStatus` is only "internal" to the running binary, so it should be possible to just combine the two, keeping the values from `TxnSql`, and eliminating the various completely unused status codes or at least marking them as `[[deprecated, maybe_unused]]`.

HowardHinnant (approved):

src/ripple/rpc/handlers/Submit.cpp
```diff
@@ -116,15 +116,8 @@ doSubmit(RPC::JsonContext& context)
...
-    auto tpTrans = std::make_shared<Transaction>(stpTrans, reason, context.app);
-    if (tpTrans->getStatus() != NEW)
-    {
-        jvResult[jss::error] = "invalidTransaction";
-        jvResult[jss::error_exception] = "fails local checks: " + reason;
-
-        return jvResult;
-    }
+    auto tpTrans = std::make_shared<Transaction>(stpTrans);
+    assert(tpTrans);
```
ximinez, line 120: If `std::make_shared` fails, it'll throw, so if the execution reaches the `assert`, then `tpTrans` will be set. I think it makes more sense to `assert(tpTrans->getStatus() == NEW);`, or since `NEW` is the default ctor parameter, remove the assert entirely.

src/ripple/rpc/impl/TransactionSign.cpp
```diff
@@ -559,17 +559,9 @@ transactionConstructImpl(
...
-        tpTrans = std::make_shared<Transaction>(stpTrans, reason, app);
-        if (tpTrans->getStatus() != NEW)
-        {
-            ret.first = RPC::make_error(
-                rpcINTERNAL, "Unable to construct transaction: " + reason);
-            return ret;
-        }
-    }
+    auto tpTrans = std::make_shared<Transaction>(stpTrans);
+    assert(tpTrans);
```
ximinez, line 563: As above, if `std::make_shared` fails, it'll throw, so if the execution reaches the `assert`, then `tpTrans` will be set. I think it makes more sense to `assert(tpTrans->getStatus() == NEW);`, or since `NEW` is the default ctor parameter, remove the assert entirely.

ximinez (changes requested):

ximinez:
> I like this. If possible, consider whether the `TransStatus` and `TxnSql` enumerations can be combined. I _think_ that `TransStatus` is only "internal" to the running binary, so it should be possible to just combine the two, keeping the values from `TxnSql`, and eliminating the various completely unused status codes or at least marking them as `[[deprecated, maybe_unused]]`.

@nbougalis This is a really cool idea. I did a quick check, and found that of the 9 `TransStatus` values, the first 6 map to `TxnSql` values, and of the other 3, only `OBSOLETE` is used. However, from what I can tell, even though it can be set to several different values, when the `TransStatus` is read, it's only compared against `INCLUDED` and `INVALID`. In other words, we could probably roll `OBSOLETE` into one of the other values, and get rid of `TransStatus`, but it would take some work that might be beyond the scope of this PR.

ximinez (changes requested):
The good news is that it looks like all my concerns have been addressed. However, it doesn't build now.

This is the build error:
```
C:\Dev\rippled\review1\src\ripple\rpc\handlers\Tx.cpp(299,52): error C2039: 'getSTransaction': is not a member of 'ripple::Transaction' [C:\Dev\rippled\review1\build\cmake\msvc19.OFF\rippled.vcxproj]
```
I fixed the error and took @gregtatcam's idea for the ctor and ran with it. I think it's a lot cleaner this way. Here's the commit if you want to cherry pick it onto your branch. https://github.com/XRPLF/rippled/commit/4def06e026d7d1faab6769cfecc087d5b95fe5f0

Then, since I was on a roll, I implemented @nbougalis's [idea to collapse the `TxnSql` and `TransStatus` enums](https://github.com/XRPLF/rippled/pull/4304#issuecomment-1566277360). Here's that commit if you want to cherry pick it. https://github.com/XRPLF/rippled/commit/78dcdc0aee991ddf64391e8d3e6541b4d81cdd85

*Edit:* On second thought, I like this solution for `TransStatus` better https://github.com/XRPLF/rippled/commit/9201b263dafd8394a0602943d9d53bd670d2cf07.

intelliot:
@a-noni-mousse - will you be able to consider (and possibly merge or cherry-pick) the commits that ximinez offered above?

PR #4305, opened 2022-09-18
Support for boost 1.80.0

seelabs:
Boost intrusive renamed the `comp()` function -> `get_comp()`


ximinez (approved):
Builds and passes tests in Windows with both Boost 1.77 and 1.80.

PR #4323, opened 2022-10-17
Catch transaction deserialization error in doLedgerGrpc

cjcobb23:
* Allow clio to extract ledgers with transactions that can no longer be deserialized. The problem transactions will be skipped.

Some old transactions cannot be deserialized by current rippled code. Clio can still extract the ledger headers and state objects, just not the transactions.

### DIFF
--- a/src/ripple/rpc/handlers/LedgerHandler.cpp
+++ b/src/ripple/rpc/handlers/LedgerHandler.cpp
@@ -164,7 +164,8 @@ doLedgerGrpc(RPC::GRPCContext<org::xrpl::rpc::v1::GetLedgerRequest>& context)
             JLOG(context.j.error())
                 << __func__ << " - Error deserializing transaction in ledger "
                 << ledger->info().seq
-                << " . skipping transaction. you should look into this further";
+                << " . skipping transaction and following transactions. You "
+                   "should look into this further";
         }
     }
 
--- a/src/ripple/rpc/handlers/LedgerHandler.cpp
+++ b/src/ripple/rpc/handlers/LedgerHandler.cpp
@@ -134,27 +134,37 @@ doLedgerGrpc(RPC::GRPCContext<org::xrpl::rpc::v1::GetLedgerRequest>& context)
 
     if (request.transactions())
     {
-        for (auto& i : ledger->txs)
+        try
         {
-            assert(i.first);
-            if (request.expand())
+            for (auto& i : ledger->txs)
             {
-                auto txn =
-                    response.mutable_transactions_list()->add_transactions();
-                Serializer sTxn = i.first->getSerializer();
-                txn->set_transaction_blob(sTxn.data(), sTxn.getLength());
-                if (i.second)
+                assert(i.first);
+                if (request.expand())
                 {
-                    Serializer sMeta = i.second->getSerializer();
-                    txn->set_metadata_blob(sMeta.data(), sMeta.getLength());
+                    auto txn = response.mutable_transactions_list()
+                                   ->add_transactions();
+                    Serializer sTxn = i.first->getSerializer();
+                    txn->set_transaction_blob(sTxn.data(), sTxn.getLength());
+                    if (i.second)
+                    {
+                        Serializer sMeta = i.second->getSerializer();
+                        txn->set_metadata_blob(sMeta.data(), sMeta.getLength());
+                    }
+                }
+                else
+                {
+                    auto const& hash = i.first->getTransactionID();
+                    response.mutable_hashes_list()->add_hashes(
+                        hash.data(), hash.size());
                 }
             }
-            else
-            {
-                auto const& hash = i.first->getTransactionID();
-                response.mutable_hashes_list()->add_hashes(
-                    hash.data(), hash.size());
-            }
+        }
+        catch (std::exception const& e)
+        {
+            JLOG(context.j.error())
+                << __func__ << " - Error deserializing transaction in ledger "
+                << ledger->info().seq
+                << " . skipping transaction. you should look into this further";
         }
     }
 


mDuo13:
Does this error only apply to gRPC deserialization, or do these transactions also fail to deserialize to JSON in the rippled APIs directly? Do you have an example tx hash to look up? (If there's no way to deserialize these transactions from their binary form, that's very undesirable; I don't like having "black box" transactions in history; the 32570 missing ledgers are bad enough.)

Note, I am not saying this PR should fix the problem of deserializing things, but the answers to my question will inform whether or not we open a ticket to fix the issue more broadly, and the scope thereof.

cjcobb23:
Here's one:
```
hash: "C7CE7D48C50D6ABEBC8DA01447C8BDA21324094B58C0738AED30D1FCD399FAD6"
ledger sequence: 7266393
actual transaction:
            {
               "Account" : "rEX2Q18vQziVAJoiLwoVHN5u87CXm29LCQ",
               "Fee" : "10",
               "Flags" : 2147483648,
               "Memos" : [
                  {
                     "TransactionMetaData" : {
                        "TransactionMetaData" : {
                           "TransactionMetaData" : {
                              "TransactionMetaData" : {
                                 "TransactionMetaData" : {
                                    "TransactionMetaData" : {
                                       "TransactionMetaData" : {
                                          "TransactionMetaData" : {
                                             "TransactionMetaData" : {
                                                "TransactionMetaData" : {
                                                   "TransactionMetaData" : {
                                                      "TransactionMetaData" : {
                                                         "TransactionMetaData" : {
                                                            "TransactionMetaData" : {
                                                               "TransactionMetaData" : {}
                                                            }
                                                         }
                                                      }
                                                   }
                                                }
                                             }
                                          }
                                       }
                                    }
                                 }
                              }
                           }
                        }
                     }
                  }
               ],
               "Sequence" : 51,
               "SigningPubKey" : "033D098A1F0664D894D0D542006541CDDC997E1B5A5F1A6E63BE5B83E12AAF00D2",
               "TxnSignature" : "304402204DF684C465B2D110F9DF39100E205084C972FEF885AF2A9036DBB58866E224AE022055C4DF3345F8111552C4D01B896D3E22CBFAD450EF020737DD9F2C3907294D3E",
               "hash" : "C7CE7D48C50D6ABEBC8DA01447C8BDA21324094B58C0738AED30D1FCD399FAD6",
               "metaData" : {
                  "AffectedNodes" : [
                     {
                        "ModifiedNode" : {
                           "FinalFields" : {
                              "Account" : "rEX2Q18vQziVAJoiLwoVHN5u87CXm29LCQ",
                              "Balance" : "32005425",
                              "Flags" : 0,
                              "OwnerCount" : 2,
                              "Sequence" : 52
                           },
                           "LedgerEntryType" : "AccountRoot",
                           "LedgerIndex" : "BCCEA6B346A29656AFD37B8AB7F573DA705AB7A86D9FE197AC6CB1E9D5772EDB",
                           "PreviousFields" : {
                              "Balance" : "32005435",
                              "Sequence" : 51
                           },
                           "PreviousTxnID" : "82FC7296AE62216C6A58DFF6707CD441C57ECF835DD784449C8C0CA78A9ED143",
                           "PreviousTxnLgrSeq" : 7266389
                        }
                     }
                  ],
                  "TransactionIndex" : 6,
                  "TransactionResult" : "tesSUCCESS"
               }
            },
```

cjcobb23:
The error applies to both gRPC deserialization as well as JSON deserialization. The error actually just lies at transaction deserialization itself, where rippled cannot properly deserialize the transaction at all anymore.

scottschurr:
Okay.  I now understand the previous comments.  The identified _problem_ applies both to clio and, more generally, to any time rippled tries to deserialize these transactions.  The supplied _fix_ only works for clio.  I'm good with that, but that understanding gives me context for what I'm reviewing.

src/ripple/rpc/handlers/LedgerHandler.cpp
```diff
@@ -134,27 +134,37 @@ doLedgerGrpc(RPC::GRPCContext<org::xrpl::rpc::v1::GetLedgerRequest>& context)
...
-                response.mutable_hashes_list()->add_hashes(
-                    hash.data(), hash.size());
-            }
+        }
+        catch (std::exception const& e)
+        {
+            JLOG(context.j.error())
+                << __func__ << " - Error deserializing transaction in ledger "
+                << ledger->info().seq
+                << " . skipping transaction. you should look into this further";
```
scottschurr, line 167: Since the try/catch surrounds the entire `for` loop, I don't think that what is written to the log is accurate.  It's not just one transaction that is being skipped.  It's also any transactions in the ledger that might follow that transaction.  I'm okay with that implementation if it works for clio, but maybe the message should be accurate.

src/ripple/rpc/handlers/LedgerHandler.cpp
```diff
@@ -134,27 +134,37 @@ doLedgerGrpc(RPC::GRPCContext<org::xrpl::rpc::v1::GetLedgerRequest>& context)
...
-                    txn->set_metadata_blob(sMeta.data(), sMeta.getLength());
+                    auto txn = response.mutable_transactions_list()
+                                   ->add_transactions();
+                    Serializer sTxn = i.first->getSerializer();
+                    txn->set_transaction_blob(sTxn.data(), sTxn.getLength());
+                    if (i.second)
+                    {
+                        Serializer sMeta = i.second->getSerializer();
+                        txn->set_metadata_blob(sMeta.data(), sMeta.getLength());
+                    }
```
scottschurr, line 152: I _think_ the things that may `throw` here are the calls to `getSerializer()`.  If I got that wrong, then the rest of this comment may not apply.

I see two possible issues with the way this code is structured.

1. If one transaction in a ledger throws, then you will not return any transactions later in that ledger.

2. It looks like it may be possible to return an empty or partially filled entry in a `transactions_list` if an exception does occur.

I think these problems can be fixed by restructuring the code a bit.  Consider the following approach which moves the `try/catch`(s) inside of the `for` loop.
```
    if (request.transactions())
    {
        for (auto& i : ledger->txs)
        {
            assert(i.first);
            if (request.expand())
            {
                try
                {
                    Serializer const sTxn = i.first->getSerializer();
                    std::optional<Serializer> const sMeta(
                        i.second ? std::optional<Serializer>(
                                       i.second->getSerializer())
                                 : std::nullopt);

                    auto txn = response.mutable_transactions_list()
                                   ->add_transactions();

                    txn->set_transaction_blob(sTxn.data(), sTxn.getLength());
                    if (i.second)
                        txn->set_metadata_blob(
                            sMeta->data(), sMeta->getLength());
                }
                catch (std::exception const&)
                {
                    JLOG(context.j.error())
                        << __func__
                        << " - Error deserializing transaction in ledger "
                        << ledger->info().seq
                        << ". skipping transaction. "
                           "you should look into this further";
                }
            }
            else
            {
                try
                {
                    auto const& hash = i.first->getTransactionID();
                    response.mutable_hashes_list()->add_hashes(
                        hash.data(), hash.size());
                }
                catch (std::exception const&)
                {
                    JLOG(context.j.error())
                        << __func__
                        << " - Error deserializing transaction hash in ledger "
                        << ledger->info().seq
                        << ". skipping transaction. "
                           "you should look into this further";
                }
            }
        }
    }
```
With this restructure we do the work that might `throw` before we create a response entry.  I think this will help keep any response that is provided from being empty or partial.  We also attempt to return any transactions that might be in the ledger following the transaction with the serialization issue.

scottschurr (commented):
I have a couple of thoughts that may be improvements.  But let me know what you think.  I'm open to being wrong.

cjcobb23:
I actually just copied the way the JSON handler does this: https://github.com/XRPLF/rippled/blob/ebbf4b64a57d2673d54a6b25def7726601d99bf9/src/ripple/app/ledger/impl/LedgerToJson.cpp#L176

I agree it would be better to continue on failure, but maybe that should be a separate PR? I just wanted to get things back up and running for clio, because right now the entire call returns an error (unlike the JSON version of the call).

I unfortunately don't have the patch that actually deserialized this transaction, it was so long ago.

scottschurr:
@cjcobb23, I'm not sure I understand the motivation for putting the suggested change in a separate pull request.  The suggested code changes are not extensive and they are localized.  I'm not guaranteeing the suggested change is bug free; it needs testing.  But the implementation in the pull request _has_ bugs that have been pointed out.

You haven't yet convinced me that it's a better idea to stay with the implementation in the pull request when we know that the pull request has fixable flaws.  But I'm listening.

cjcobb23:
The implementation in the pull request mirrors the JSON implementation that already exists. The exception is actually thrown as part of the loop iteration, not in the body of the loop. So there won't be any incomplete data. We will skip later transactions, as does the JSON implementation. However, I'd rather solve that issue in a later pull request.

scottschurr (approved):
I now understand what @cjcobb23 means by...

>  The exception is actually thrown as part of the loop iteration, not in the body of the loop.

It means the exception is thrown here:
```
for (auto& i : ledger->txs)
```
So the restructure I suggested will not actually catch the exception.

Thanks for the explanation.  I withdraw my suggestion.  I think this pull request is good to go.

intelliot:
Mentioning @manojsdoshi - please sync with @cjcobb23 to see how this should be tested.

PR #4325, opened 2022-10-19
Fix clang 15 warnings

seelabs:
Clang warns about the code removed in this patch with the following warning:
```
warning: out-of-line definition of constexpr static data member is redundant in C++17 and is deprecated [-Wdeprecated]
```

### DIFF
--- a/src/ripple/app/misc/impl/TxQ.cpp
+++ b/src/ripple/app/misc/impl/TxQ.cpp
@@ -78,8 +78,6 @@ increase(FeeLevel64 level, std::uint32_t increasePercent)
 
 //////////////////////////////////////////////////////////////////////////
 
-constexpr FeeLevel64 TxQ::baseLevel;
-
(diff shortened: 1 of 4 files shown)


scottschurr:
Nit: the commit message contains a very long line.  Consider wrapping the lines at 70 characters.

scottschurr (approved):
:+1: Changes look good.  I'd prefer that the lines in the commit message were limited to 70 characters in length, but I won't hold up the pull request just for that.

seelabs:
@scottschurr Thanks for catching the long lines in the commit message. Fixed and force pushed. The only change in the force push was the shorting the lines in the commit messages.

HowardHinnant (approved):

ximinez (approved):
I didn't realize C++17 had made these redundant. But I'm glad, because they were really annoying. Thanks for the cleanup.

PR #4328, opened 2022-10-24
Allow port numbers be be specified using a either a colon or a space

RichardAH:
An annoying *gotcha* for node operators is accidentally specifying IP:PORT combinations using a colon. Rippled expects a space, not a colon, and for whatever reason does not provide good feedback when this operator error is made.

This small PR simply allows this mistake (colon) to be replaced inline with a space, preserving the intention of the operator.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Test Plan
* `testColons()` has been added to the config unit tests. This tests the functionality when specifying IP:PORT in rippled.cfg.
* The RPCCall test regime is not specific enough to test this functionality, but I have tested it by hand (and you should too).

### DIFF
--- a/src/ripple/net/impl/RPCCall.cpp
+++ b/src/ripple/net/impl/RPCCall.cpp
@@ -482,17 +482,31 @@ private:
         return jvRequest;
     }
 
-    // connect <ip> [port]
+    // connect <ip[:port]> [port]
     Json::Value
     parseConnect(Json::Value const& jvParams)
     {
         Json::Value jvRequest(Json::objectValue);
-
-        jvRequest[jss::ip] = jvParams[0u].asString();
-
+        std::string ip = jvParams[0u].asString();
         if (jvParams.size() == 2)
+        {
+            jvRequest[jss::ip] = ip;
             jvRequest[jss::port] = jvParams[1u].asUInt();
+            return jvRequest;
+        }
 
+        // handle case where there is one argument of the form ip:port
+        if (std::count(ip.begin(), ip.end(), ':') == 1)
+        {
+            std::size_t colon = ip.find_last_of(":");
+            jvRequest[jss::ip] = std::string{ip, 0, colon};
+            jvRequest[jss::port] =
+                Json::Value{std::string{ip, colon + 1}}.asUInt();
+            return jvRequest;
+        }
+
+        // default case, no port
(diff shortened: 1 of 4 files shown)


src/ripple/net/impl/RPCCall.cpp
```diff
@@ -482,17 +482,31 @@ class RPCParser
...
+
+        // handle case where there is one argument of the form ip:port
+        std::size_t colon = ip.find_last_of(":");
+        if (colon != std::string::npos)
+        {
+            jvRequest[jss::ip] = std::string{ip, 0, colon};
+            jvRequest[jss::port] =
+                Json::Value{std::string{ip, colon + 1}}.asUInt();
+            return jvRequest;
+        }
```
nbougalis, line 506: This logic breaks if you specify an IPv6 address, whether you use brackets or not, and that can be a moderately annoying problem.

Perhaps considering checking if `ip` can be successfully parsed as either an IPv6 addresses using, for example,
[make_address_v6](https://www.boost.org/doc/libs/1_80_0/doc/html/boost_asio/reference/ip__address_v6/make_address_v6.html) and only doing this is if it isn't.

nbougalis (changes requested):

RichardAH:
Thanks for the review @nbougalis. After some consideration I decided simply detecting the presence of more than one colon should be sufficient to filter out IPv6 cases. Since IPv6 can be ambiguous if a colon is used for port (in particular if square braces are also not used), IPv6 ports must remain specified via space character.

nbougalis (approved):
Looks good with the change to detect IPv6 addresses.

I have an idea for a more comprehensive fix (improve Boost's parsers, and use those :laughing:) but this is fine for now.

ximinez (changes requested):

intelliot:
@RichardAH when you have a chance, please respond to ximinez's comments above. Thanks!

ximinez (approved):
I wouldn't mind seeing `develop` merged in to the branch just so CI can get a shot at it, but I don't see a problem if not. Sorry for the delay!

intelliot:
@RichardAH I merged `develop` into the branch. If you feel this PR is ready to merge, please put the `Passed` label on it.

(do not merge it yet)

PR #4331, opened 2022-10-27
fixes issue where brokers cannot perform NFTokenAcceptOffer in certain scenarios

ledhed2222:

In the follow scenario, an account cannot perform NFTokenAcceptOffer even though it should be allowed to:

0) BROKER has < `S`
1) ALICE offers to sell token for `S`
2) BOB offers to buy token for > `S`
3) BROKER tries to bridge the two offers

This will result in a tecINSUFFICIENT_FUNDS, but should not because BROKER is not spending any funds in this transaction, beyond the transaction fee.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

The fix here is to only validate that the tx submitter has enough funds to cover the sell offer when NOT bridging two offers. This fixes the issue because in the case of bridging two offers, the submitter is the broker, not the buyer.

This fix is OK and doesn't need to happen at all in bridged mode, because there is already a check in preclaim to ensure that the buyer has sufficient funds if there is a BuyOffer in play. Here, we appropriately are checking the funds of the owner of the NFTokenOffer object, not the tx submitter.

Tested manually with some scripting and a standalone node. Also added integration test scenario for this and verified that it fails with change backed out, but passes with change included.

### DIFF
--- a/src/ripple/app/tx/impl/NFTokenAcceptOffer.cpp
+++ b/src/ripple/app/tx/impl/NFTokenAcceptOffer.cpp
@@ -199,19 +199,18 @@ NFTokenAcceptOffer::preclaim(PreclaimContext const& ctx)
             if (auto const dest = so->at(~sfDestination);
                 dest.has_value() && *dest != ctx.tx[sfAccount])
                 return tecNO_PERMISSION;
+
+            // The account offering to buy must have funds:
+            auto const needed = so->at(sfAmount);
+            if (accountHolds(
+                    ctx.view,
+                    ctx.tx[sfAccount],
+                    needed.getCurrency(),
+                    needed.getIssuer(),
+                    fhZERO_IF_FROZEN,
+                    ctx.j) < needed)
+                return tecINSUFFICIENT_FUNDS;
         }
-
-        // The account offering to buy must have funds:
-        auto const needed = so->at(sfAmount);
-
-        if (accountHolds(
-                ctx.view,
-                ctx.tx[sfAccount],
-                needed.getCurrency(),
-                needed.getIssuer(),
-                fhZERO_IF_FROZEN,
-                ctx.j) < needed)
-            return tecINSUFFICIENT_FUNDS;
     }
 
     return tesSUCCESS;
--- a/src/test/app/NFToken_test.cpp
+++ b/src/test/app/NFToken_test.cpp
@@ -3927,31 +3927,33 @@ class NFToken_test : public beast::unit_test::suite
             env.close();
         }
 
-        // Lambda to set the balance of all passed in accounts to gwXAU(1000).
-        auto setXAUBalance_1000 =
+        // Lambda to set the balance of all passed in accounts to
+        // gwXAU(amount)
+        auto setXAUBalance =
             [this, &gw, &gwXAU, &env](
                 std::initializer_list<std::reference_wrapper<Account const>>
                     accounts,
-                int line) {
+                int line,
+                int amount = 1000) {
                 for (Account const& acct : accounts)
                 {
-                    static const auto xau1000 = gwXAU(1000);
+                    auto const xauAmt = gwXAU(amount);
                     auto const balance = env.balance(acct, gwXAU);
-                    if (balance < xau1000)
+                    if (balance < xauAmt)
                     {
-                        env(pay(gw, acct, xau1000 - balance));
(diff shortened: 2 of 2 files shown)


src/test/app/NFToken_test.cpp
```diff
@@ -3927,31 +3927,33 @@ class NFToken_test : public beast::unit_test::suite
...
-        auto setXAUBalance_1000 =
+        // Lambda to set the balance of all passed in accounts to
+        // gwXAU(amount)
+        auto setXAUBalance =
             [this, &gw, &gwXAU, &env](
                 std::initializer_list<std::reference_wrapper<Account const>>
                     accounts,
-                int line) {
+                int line,
+                int amount = 1000) {
```
scottschurr, line 3937: I'm not wild about the default argument.  If I read the test's code, `setXAUBalance` no longer tells me what the balance is being set to.  And it makes a difference in each of the tests.  But since this is unit test code this is not a hill I'm ready to die on.

There are two approaches that would be better in my opinion.

1. You could leave the code as it was and have `broker` pay `gwXAU(500)` to `gw` shortly after calling `setXAUBalance_1000`.  Then `broker` would have the necessary lower balance and the old interface would be left alone.

2. You could make the `amount` mandatory. This would be more editing because you'd need to add `1000` to almost every call to `setXAUBalance`.  But it's probably the cleaner solution.  Also, if you follow this approach, you should swap the order of the `line` and `amount` arguments.  The `line` argument is just, uh, line noise (:smile:) when reading the code.  It should go last.

But, again, I'll not block the pull request for this particular change.

src/test/app/NFToken_test.cpp
```diff
@@ -4166,6 +4168,48 @@ class NFToken_test : public beast::unit_test::suite
             BEAST_EXPECT(env.balance(buyer, gwXAU) == gwXAU(0));
             BEAST_EXPECT(env.balance(broker, gwXAU) == gwXAU(1050));
 
+            // Burn the NFT so the next test starts with a clean state.
+            env(token::burn(buyer, nftID));
+            env.close();
+        }
+        // broker has a balance less than the seller offer
+        {
+            checkOwnerCountIsOne({issuer, minter, buyer, broker}, __LINE__);
```
scottschurr, line 4177: I really like that you added this unit test.  Thank you very much!

That said, I think we need to have the test run both with and without the amendment.  It's not hard to do, but it can be a bit messy.  I have an example for one possible approach to the change in the top-most commit here:

https://github.com/scottschurr/rippled/commits/gregw-brokered-mode-nft-patch

But don't feel that you need to follow that example.  There may be better ways to do it.

src/ripple/app/tx/impl/NFTokenAcceptOffer.cpp
```diff
@@ -199,19 +199,18 @@ NFTokenAcceptOffer::preclaim(PreclaimContext const& ctx)
...
 
-        if (accountHolds(
-                ctx.view,
-                ctx.tx[sfAccount],
-                needed.getCurrency(),
-                needed.getIssuer(),
-                fhZERO_IF_FROZEN,
-                ctx.j) < needed)
-            return tecINSUFFICIENT_FUNDS;
+            // The account offering to buy must have funds:
```
scottschurr, line 203: I think this fix looks essentially right.  Good job.

However, even though you're only moving braces, this change affects transaction results.  That's kinda the point.  So the change needs to go on an amendment.  I have an example of how that might be done in the top-most commit here:

https://github.com/scottschurr/rippled/commits/gregw-brokered-mode-nft-patch

There may be better ways to do it, so just use that example as a starting place.

scottschurr (changes requested):
The fix looks basically right.  Also, thanks very much for including the unit test!

Since the bug fix is transaction changing, I think we need to put it on an amendment.  Also I'd like the unit test to exercise the case both before and after the change.  That shows how your test exercises the failing case and how the amendment fixes the problem.

Thanks for working on this!

ledhed2222:
thanks @scottschurr ! yes agree this is an amendment. could use your help understanding the ramifications of that WRT release and anything development-wise here.

intelliot:
note: this may get bundled with other NFT bug fixes and may be ready around mid-January

ledhed2222:
Closing - this fix is combined with another in PR 4380

PR #4332, opened 2022-10-31
Work around gdb bug by changing a template parameter:

seelabs:
There's a bug in gdb where unsigned template parameters cause issues with RTTI. This patch changes a template parameter from `size_t` to `int` to workaround this gdb bug.

### DIFF
--- a/src/ripple/protocol/STBitString.h
+++ b/src/ripple/protocol/STBitString.h
@@ -25,9 +25,15 @@
 
 namespace ripple {
 
-template <std::size_t Bits>
+// The template parameter could be an unsigned type, however there's a bug in
+// gdb (last checked in gdb 12.1) that prevents gdb from finding the RTTI
+// information of a template parameterized by an unsigned type. This RTTI
+// information is needed to write gdb pretty printers.
+template <int Bits>
 class STBitString final : public STBase
 {
+    static_assert(Bits > 0, "Number of bits must be positive");
+
 public:
     using value_type = base_uint<Bits>;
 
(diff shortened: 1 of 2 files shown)


src/ripple/protocol/STBitString.h
```diff
@@ -25,7 +25,7 @@
 
 namespace ripple {
 
-template <std::size_t Bits>
+template <int Bits>
```
ximinez, line 28: I suggest adding a comment to the class declaration explaining the reason that this template parameter is `int` so that nobody comes along later and changes it back thinking they're doing the right thing.

ximinez (changes requested):

src/ripple/protocol/STBitString.h
```diff
@@ -25,7 +25,11 @@
 
 namespace ripple {
 
-template <std::size_t Bits>
+// The template parameter could be an unsigned type, however there's a bug in
+// gdb (last checked in gdb 12.1) that prevents gdb from finding the RTTI
+// information of a template parameterized by an unsigned type. This RTTI
+// information is needed to write gdb pretty printers.
+template <int Bits>
```
scottschurr, line 32: There's a bit of an issue here since `Bits` is used as a parameter to the `base_uint` template which takes a `std::size_t` argument.  If someone passes a negative `Bits` then C++ conversion rules will turn that into a very big `std::size_t` for `base_uint`.  Yeah that will collapse early at run time. But I'd prefer to catch the problem at build time.

The simplest fix would be to add
```
static_assert(Bits > 0);
```
to the body of `STBitString`.  Probably best to place that `static_assert` before the `using value_type...`.

seelabs: Yep, good suggestion. Fixed in https://github.com/XRPLF/rippled/pull/4332/commits/ffef6fd66b4ebbd9e2f32d10b19c2b5218853626

scottschurr:
:+1: LGTM.  Happy pretty printing.

scottschurr (approved):

ximinez (approved):
Looks good.

seelabs:
Thanks both for the quick review. Squashed and put a "passed" label on this.

ximinez (changes requested):
I just noticed that "template" is misspelled in the first line of the commit message.

seelabs:
Fixed commit message. (@ximinez )

ximinez (approved):

PR #4346, opened 2022-11-16
`fixUnburnableNFToken`: Allow NFT to be burned when number of offers is greater than 500

shawnxie999:
Ready for Review
(disable whitespaces to see file comparisons better)

This PR introduces an amendment called `fixUnburnableNFToken` which allows NFTs with more than 500 offers to be burned. This is enabled by deleting 500 offers but leaving the others untouched.

### Context of Change
Current NFT cannot be burned when it has over 500 offers. We want to remove this restriction. So we delete exactly 500 offers, and will leave out other offers untouched if there are more.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1203706294512476

### DIFF
--- a/src/ripple/app/tx/impl/NFTokenBurn.cpp
+++ b/src/ripple/app/tx/impl/NFTokenBurn.cpp
@@ -111,7 +111,7 @@ NFTokenBurn::doApply()
 
     if (ctx_.view().rules().enabled(fixUnburnableNFToken))
     {
-        // Delete up to 500 offers. 
+        // Delete up to 500 offers.
         // Delete buy offers first to minimize a buyer's effort
         // to manually delete leftover offers.
         int const deletedBuyOffers = nft::removeTokenOffersWithLimit(
--- a/src/ripple/app/tx/impl/NFTokenBurn.cpp
+++ b/src/ripple/app/tx/impl/NFTokenBurn.cpp
@@ -114,7 +114,7 @@ NFTokenBurn::doApply()
         // Delete up to 500 offers in total.
         // Because the number of sell offers is likely to be less than
         // the number of buy offers, we prioritize the deletion of sell
-        // offers in order to clean up sell offer directory 
+        // offers in order to clean up sell offer directory
         int const deletedSellOffers = nft::removeTokenOffersWithLimit(
             view(),
             keylet::nft_sells(ctx_.tx[sfNFTokenID]),
--- a/src/ripple/app/tx/impl/NFTokenBurn.cpp
+++ b/src/ripple/app/tx/impl/NFTokenBurn.cpp
@@ -77,9 +77,14 @@ NFTokenBurn::preclaim(PreclaimContext const& ctx)
         }
     }
 
-    // If there are too many offers, then burning the token would produce too
-    // much metadata.  Disallow burning a token with too many offers.
-    return nft::notTooManyOffers(ctx.view, ctx.tx[sfNFTokenID]);
+    if (!ctx.view.rules().enabled(fixUnburnableNFToken))
+    {
+        // If there are too many offers, then burning the token would produce
+        // too much metadata.  Disallow burning a token with too many offers.
+        return nft::notTooManyOffers(ctx.view, ctx.tx[sfNFTokenID]);
+    }
+
+    return tesSUCCESS;
 }
 
 TER
@@ -104,9 +109,30 @@ NFTokenBurn::doApply()
         view().update(issuer);
     }
 
-    // Optimized deletion of all offers.
-    nft::removeAllTokenOffers(view(), keylet::nft_sells(ctx_.tx[sfNFTokenID]));
-    nft::removeAllTokenOffers(view(), keylet::nft_buys(ctx_.tx[sfNFTokenID]));
+    if (ctx_.view().rules().enabled(fixUnburnableNFToken))
+    {
+        // Delete up to 500 offers, but prefers buy offers first
+        auto const deletedBuyOffers = nft::removeTokenOffersWithLimit(
+            view(),
+            keylet::nft_buys(ctx_.tx[sfNFTokenID]),
+            maxDeletableTokenOfferEntries);
+
+        if (maxDeletableTokenOfferEntries - deletedBuyOffers > 0)
+        {
+            nft::removeTokenOffersWithLimit(
+                view(),
+                keylet::nft_sells(ctx_.tx[sfNFTokenID]),
+                maxDeletableTokenOfferEntries - deletedBuyOffers);
+        }
+    }
+    else
+    {
+        // Optimized deletion of all offers.
+        nft::removeAllTokenOffers(
+            view(), keylet::nft_sells(ctx_.tx[sfNFTokenID]));
+        nft::removeAllTokenOffers(
+            view(), keylet::nft_buys(ctx_.tx[sfNFTokenID]));
+    }
 
     return tesSUCCESS;
 }
--- a/src/ripple/app/tx/impl/NFTokenBurn.cpp
+++ b/src/ripple/app/tx/impl/NFTokenBurn.cpp
@@ -129,14 +129,14 @@ NFTokenBurn::doApply()
     {
         // Deletion of all offers.
         nft::removeTokenOffersWithLimit(
-                view(),
-                keylet::nft_sells(ctx_.tx[sfNFTokenID]),
-                std::numeric_limits<int>::max());
+            view(),
+            keylet::nft_sells(ctx_.tx[sfNFTokenID]),
+            std::numeric_limits<int>::max());
 
         nft::removeTokenOffersWithLimit(
-                view(),
-                keylet::nft_buys(ctx_.tx[sfNFTokenID]),
-                std::numeric_limits<int>::max());
+            view(),
+            keylet::nft_buys(ctx_.tx[sfNFTokenID]),
+            std::numeric_limits<int>::max());
     }
 
     return tesSUCCESS;
--- a/src/ripple/app/tx/impl/NFTokenBurn.cpp
+++ b/src/ripple/app/tx/impl/NFTokenBurn.cpp
@@ -115,12 +115,12 @@ NFTokenBurn::doApply()
         // Because the number of sell offers is likely to be less than
         // the number of buy offers, we prioritize the deletion of sell
         // offers in order to clean up sell offer directory
-        int const deletedSellOffers = nft::removeTokenOffersWithLimit(
+        std::size_t const deletedSellOffers = nft::removeTokenOffersWithLimit(
             view(),
             keylet::nft_sells(ctx_.tx[sfNFTokenID]),
             maxDeletableTokenOfferEntries);
 
-        if (maxDeletableTokenOfferEntries - deletedSellOffers > 0)
+        if (maxDeletableTokenOfferEntries > deletedSellOffers)
         {
             nft::removeTokenOffersWithLimit(
                 view(),
--- a/src/ripple/app/tx/impl/NFTokenBurn.cpp
+++ b/src/ripple/app/tx/impl/NFTokenBurn.cpp
@@ -111,7 +111,9 @@ NFTokenBurn::doApply()
 
     if (ctx_.view().rules().enabled(fixUnburnableNFToken))
     {
-        // Delete up to 500 offers, but prefers buy offers first
+        // Delete up to 500 offers. 
+        // Delete buy offers first to minimize a buyer's effort
+        // to manually delete leftover offers.
         int const deletedBuyOffers = nft::removeTokenOffersWithLimit(
             view(),
             keylet::nft_buys(ctx_.tx[sfNFTokenID]),
--- a/src/ripple/app/tx/impl/NFTokenBurn.cpp
+++ b/src/ripple/app/tx/impl/NFTokenBurn.cpp
@@ -112,7 +112,7 @@ NFTokenBurn::doApply()
     if (ctx_.view().rules().enabled(fixUnburnableNFToken))
     {
         // Delete up to 500 offers, but prefers buy offers first
-        auto const deletedBuyOffers = nft::removeTokenOffersWithLimit(
+        int const deletedBuyOffers = nft::removeTokenOffersWithLimit(
             view(),
             keylet::nft_buys(ctx_.tx[sfNFTokenID]),
             maxDeletableTokenOfferEntries);
@@ -127,11 +127,16 @@ NFTokenBurn::doApply()
     }
     else
     {
-        // Optimized deletion of all offers.
-        nft::removeAllTokenOffers(
-            view(), keylet::nft_sells(ctx_.tx[sfNFTokenID]));
-        nft::removeAllTokenOffers(
-            view(), keylet::nft_buys(ctx_.tx[sfNFTokenID]));
+        // Deletion of all offers.
+        nft::removeTokenOffersWithLimit(
+                view(),
+                keylet::nft_sells(ctx_.tx[sfNFTokenID]),
+                std::numeric_limits<int>::max());
+
+        nft::removeTokenOffersWithLimit(
+                view(),
+                keylet::nft_buys(ctx_.tx[sfNFTokenID]),
+                std::numeric_limits<int>::max());
     }
 
     return tesSUCCESS;
--- a/src/ripple/app/tx/impl/NFTokenBurn.cpp
+++ b/src/ripple/app/tx/impl/NFTokenBurn.cpp
@@ -111,20 +111,21 @@ NFTokenBurn::doApply()
 
     if (ctx_.view().rules().enabled(fixUnburnableNFToken))
     {
-        // Delete up to 500 offers.
-        // Delete buy offers first to minimize a buyer's effort
-        // to manually delete leftover offers.
-        int const deletedBuyOffers = nft::removeTokenOffersWithLimit(
+        // Delete up to 500 offers in total.
+        // Because the number of sell offers is likely to be less than
+        // the number of buy offers, we prioritize the deletion of sell
+        // offers in order to clean up sell offer directory 
+        int const deletedSellOffers = nft::removeTokenOffersWithLimit(
             view(),
-            keylet::nft_buys(ctx_.tx[sfNFTokenID]),
+            keylet::nft_sells(ctx_.tx[sfNFTokenID]),
             maxDeletableTokenOfferEntries);
 
-        if (maxDeletableTokenOfferEntries - deletedBuyOffers > 0)
+        if (maxDeletableTokenOfferEntries - deletedSellOffers > 0)
         {
             nft::removeTokenOffersWithLimit(
                 view(),
-                keylet::nft_sells(ctx_.tx[sfNFTokenID]),
-                maxDeletableTokenOfferEntries - deletedBuyOffers);
+                keylet::nft_buys(ctx_.tx[sfNFTokenID]),
+                maxDeletableTokenOfferEntries - deletedSellOffers);
         }
     }
     else
--- a/src/ripple/app/tx/impl/details/NFTokenUtils.cpp
+++ b/src/ripple/app/tx/impl/details/NFTokenUtils.cpp
@@ -550,6 +550,39 @@ removeAllTokenOffers(ApplyView& view, Keylet const& directory)
     });
 }
 
+std::size_t
+removeTokenOffersWithLimit(
+    ApplyView& view,
+    Keylet const& directory,
+    std::size_t maxDeletableOffers)
+{
+    std::optional<std::uint64_t> pi{0};
+    std::vector<uint256> offers;
+    offers.reserve(maxDeletableOffers);
+
+    do
+    {
+        auto const page = view.peek(keylet::page(directory, *pi));
+        if (!page)
+            break;
+
+        for (auto const& id : page->getFieldV256(sfIndexes))
+        {
+            offers.push_back(id);
+            if (maxDeletableOffers == offers.size())
+                break;
+        }
+        pi = (*page)[~sfIndexNext];
+    } while (pi.value_or(0) && maxDeletableOffers != offers.size());
+
+    for (auto const& id : offers)
+    {
+        if (auto const offer = view.peek(keylet::nftoffer(id)))
+            deleteTokenOffer(view, offer);
+    }
+    return offers.size();
+}
+
 TER
 notTooManyOffers(ReadView const& view, uint256 const& nftokenID)
 {
--- a/src/ripple/app/tx/impl/details/NFTokenUtils.cpp
+++ b/src/ripple/app/tx/impl/details/NFTokenUtils.cpp
@@ -539,6 +539,10 @@ removeTokenOffersWithLimit(
         if (!page)
             break;
 
+        // We get the index of the next page in case the current
+        // page is deleted after all of its entries have been removed
+        pageIndex = (*page)[~sfIndexNext];
+
         auto offerIndexes = page->getFieldV256(sfIndexes);
 
         // We reverse-iterate the offer directory page to delete all entries.
@@ -562,7 +566,6 @@ removeTokenOffersWithLimit(
             if (maxDeletableOffers == deletedOffersCount)
                 break;
         }
-        pageIndex = (*page)[~sfIndexNext];
     } while (pageIndex.value_or(0) && maxDeletableOffers != deletedOffersCount);
 
     return deletedOffersCount;
--- a/src/ripple/app/tx/impl/details/NFTokenUtils.cpp
+++ b/src/ripple/app/tx/impl/details/NFTokenUtils.cpp
@@ -527,7 +527,7 @@ removeTokenOffersWithLimit(
     Keylet const& directory,
     int maxDeletableOffers)
 {
-    if(maxDeletableOffers == 0)
+    if (maxDeletableOffers == 0)
         return 0;
 
     std::optional<std::uint64_t> pageIndex{0};
@@ -542,18 +542,21 @@ removeTokenOffersWithLimit(
         auto offerIndexes = page->getFieldV256(sfIndexes);
 
         // We reverse-iterate the offer directory page to delete all entries.
-        // Deleting an entry in a NFTokenOffer directory page won't cause entries 
-        // from other pages to move to the current, so, it is safe to delete entries
-        // one by one in the page.
-        // It is required to iterate backwards to handle iterator invalidation for vector, 
-        // as we are deleting during iteration.
-        for (int i = offerIndexes.size() - 1; i >= 0; --i) {
-            if (auto const offer = view.peek(keylet::nftoffer(offerIndexes[i]))){
-                if(deleteTokenOffer(view, offer))
+        // Deleting an entry in a NFTokenOffer directory page won't cause
+        // entries from other pages to move to the current, so, it is safe to
+        // delete entries one by one in the page. It is required to iterate
+        // backwards to handle iterator invalidation for vector, as we are
+        // deleting during iteration.
+        for (int i = offerIndexes.size() - 1; i >= 0; --i)
+        {
+            if (auto const offer = view.peek(keylet::nftoffer(offerIndexes[i])))
+            {
+                if (deleteTokenOffer(view, offer))
                     ++deletedOffersCount;
                 else
                     Throw<std::runtime_error>(
-                        "Offer " + to_string(offerIndexes[i]) + " cannot be deleted!");
+                        "Offer " + to_string(offerIndexes[i]) +
+                        " cannot be deleted!");
             }
 
             if (maxDeletableOffers == deletedOffersCount)
--- a/src/ripple/app/tx/impl/details/NFTokenUtils.cpp
+++ b/src/ripple/app/tx/impl/details/NFTokenUtils.cpp
@@ -556,13 +556,13 @@ removeTokenOffersWithLimit(
     Keylet const& directory,
     std::size_t maxDeletableOffers)
 {
-    std::optional<std::uint64_t> pi{0};
+    std::optional<std::uint64_t> pageIndex{0};
     std::vector<uint256> offers;
     offers.reserve(maxDeletableOffers);
 
     do
     {
-        auto const page = view.peek(keylet::page(directory, *pi));
+        auto const page = view.peek(keylet::page(directory, *pageIndex));
         if (!page)
             break;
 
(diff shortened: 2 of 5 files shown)


RichardAH:
Hi @shawnxie999 
You need to encapsulate your proposed changes inside a ledger amendment, otherwise changes to transactors can/will cause a ledger fork (where nodes who are yet to upgrade to the new code disagree with nodes that have upgraded.)

Have a look at https://github.com/XRPLF/rippled/pull/4336/files for an example of how to do this.

nbougalis (changes requested):
Please address the issues identified (the biggest one is the looping bug over the offer directory pages) and ping me again for a follow-up review.

nbougalis (approved):
Looks good, except for some `clang-format` issues.

scottschurr:
I'd like to make an observation based only on the description of the pull request.  I've not looked at the code yet.

The observation is a bit long-winded, and starts with how the ledger prevents duplicate NFTokens from appearing in the ledger.

Each NFToken ID incorporates the 160-bit ID of the issuing account.  That account ID plus the `MintedNFTokens` number associated with that account are sufficient to prevent one NFT from duplicating any other.  That is, unless the issuing account is deleted.

Once the issuing account is quiescent for 20 minutes or so, it can deleted.  Once an account is deleted it can be recreated, but a recreated account has its account root in an initial state.  So a recreated account has a `MintedNFTokens` value of zero.  That is, if the account is recreated it can issue NFTs with _identical_ IDs to the ones that were issued by the account that was deleted.

_Specifically_ in order to prevent duplicate NFTokens in the ledger there is a rule that prevents an account from being deleted if it has ever issued an NFToken that has not also been burned.  If an account issues no NFTokens it automatically meets that requirement.  If a different account has issued 1700 NFTokens and all 1700 of those NFTokens have been burned, that account also meets the requirements.

This is where canceling all of the offers for a burned NFToken comes in.  Let's suppose the behavior proposed by this pull request comes into play.  Then we could see this sequence of events play out:

1. Alice creates an account.
2. Alice issues one NFToken.
3. Folks (including Alice) create 500+M offers for Alice's lone NFToken.
4. Alice burns her lone NFToken.  That leaves M offers for that NFToken in the ledger, even though that NFToken no longer exists.
5. Alice waits 20 minutes and then deletes her account.
6. Alice re-creates her account.
7. Alice issues one NFToken with an ID that is identical to the one in step 2.
8. There are now M offers in the ledger for Alice's NFT, _even though Alice burned the original NFT_.

Also note that this scenario is only possible if the behavior proposed by this pull request comes into play.  If we don't accept this pull request then there are two possible outcomes:

1. Alice is unable to burn her lone NFToken because there are too many offers for it, or
2. Alice starts out by canceling some of the offers for her lone NFToken.  Once she gets the number of offers below 500, then she can burn her NFToken.

Alice can cancel as many as 500 offers for her NFToken in a single transaction.  So it would take a particularly concerted effort by attackers to prevent Alice from burning her NFToken.  The attacker needs to generates NFToken create offer transactions at 500 times the rate that Alice issues cancel transactions.  It seems improbable that this attack would be worth it.

I don't know what usability issue prompted this pull request.  But I think that usability issue needs to be balanced against the possibility of there being NFTokenOffers in the ledger for an NFToken that is not the same as the originally intended token.

The scenario is a bit convoluted.  I'm happy to work through the details more closely with anyone who's curious.

Thanks for reading.

src/ripple/app/tx/impl/details/NFTokenUtils.cpp
```diff
@@ -550,6 +550,39 @@ removeAllTokenOffers(ApplyView& view, Keylet const& directory)
...
+            break;
+
+        for (auto const& id : page->getFieldV256(sfIndexes))
+        {
+            offers.push_back(id);
+            if (maxDeletableOffers == offers.size())
+                break;
+        }
+        pi = (*page)[~sfIndexNext];
+    } while (pi.value_or(0) && maxDeletableOffers != offers.size());
```
ledhed2222, line 576: this can be refactored to remove at least one of the breaks (the first one). also i don't love the variable name `pi`, which i assume stands for `pageIndex`. The latter is more descriptive.

shawnxie999:
@scottschurr 
So let's say if this PR is merged in.  What are the worst consequences that it would cause? 

From my understanding, one ramification would be that it would cause confusion when a NFT is _re-minted_, such that left-over offers from the NFT object of the past ledger are carried over to the new one. Just curious in your perspective, how serious of a problem would this scenario be?

scottschurr:
> From my understanding, one ramification would be that it would cause confusion when a NFT is re-minted, left-over offers from the burnt NFT object are carried over to the new one. Just curious in your perspective, how serious of a problem would this scenario be?

@shawnxie999, my opinion on the severity of the problem is irrelevant.  Different observers will have different opinions on how serious the matter is. However consider this:

1. Alice mints an NFToken with a specific URL.
2. Bill (amongst over 500 others) makes an offer to buy Alice's NFToken.
3. Alice burns her NFToken, but Bill's offer is left in the ledger.
4. Alice delete's her account.
5. Alice recreate's her account.
6. Alice re-mints her NFToken, giving it the same ID but a different URL.
7. Alice accepts Bill's offer for the NFToken.

Note that the NFToken that Alice re-mints has exactly the same ID as the original one, but the associated URL is different.  So, from a legitimate perspective, even though the ID is the same, the NFToken itself is different.

How does Bill feel about having paid for an NFT that is not the same as the one he created the offer for?  That depends on Bill.  Given the folks who populate the crypto world I'd guess that there's a pretty good likelihood that Bill would be royally torqued once he found out.

So my opinion doesn't matter here.  The question is how a user of the ledger would feel.  That's above my pay grade.

ximinez:
> So my opinion doesn't matter here. The question is how a user of the ledger would feel. That's above my pay grade.

One of the guiding principles of the XRPL is that users don't get things that they didn't consent to (e.g. trust lines, `NFTAcceptOffer` transactions). Allowing this scenario, however "harmless" it might be violates that principal.

shawnxie999:
This PR is ready for review.

HowardHinnant (approved):

scottschurr:
Okay, I've reviewed the code.  I'll submit that review soon.

That said, the approach this pull request takes is not my favorite way to solve the problem.  I believe the problem that this pull request is trying to address is...

### The Problem Being Addressed

1. Alice mints an NFToken with the `lsfBurnable` flag set.  That means Alice can burn the NFToken at any point in the future, no matter who owns the token at the time.
2. Bob creates an offer for Alice's NFToken.  Alice accepts the offer.  Bob now own's Alice's NFT.
3. One or more accounts create offers for Alice's NFToken (which Bob holds), until the number of offers exceeds 500.
4. Alice attempts to burn the token that Bob holds.  The burn fails with a `tefTOO_BIG` because there are too many offers that would need to be canceled.

If Alice owned the NFToken, then Alice could cancel bunches of the offers for the token, until the number of offers fell below 500, and then successfully burn the token.  But since Alice does not own the token, Alice cannot cancel any of the offers.  Alice doesn't have that recourse.  Only Bob, who owns the NFToken, and the offer creators have that recourse.

### This Pull Request's Solution to the Problem

This pull request aims to solve the problem by letting Alice burn the token regardless of the number of offers against the token.  The process of burning the token will cancel up to 500 offers.  Any number of offers beyond that will be left in the ledger for others to deal with.  The leftover offers will just be litter in the ledger.  They will refer to an NFToken that no longer exists.

Personally, I find that approach to be uncharacteristic of the XRP Ledger.  There certainly are situations where the XRP Ledger can't complete an operation because it entails too much work.  Here are some I'm aware of:

1. Any transaction that produces too much metadata fails with a `tecOVERSIZE`.  The transaction does not succeed.
2. Executing a non-partial payment that consumes more than 1500 offers in the order books fails.
3. Deleting an account with too many deletable entries in the directory fails with `tefTOO_BIG`.

So I'm accustomed to the ledger being requested to do something and either doing the full job or failing.  No half measures.

This pull request, in contrast, specifically allows half measures.  It cleans up some of the mess, but if it can't clean up the whole thing it leaves some of the mess behind.

Just to be clear, we should look at what kind of mess is left behind.  Here is what gets left in the ledger:

- Every un-canceled NFTokenOffer for the burned NFToken is a stand-alone left-behind ledger object.
- For every left-behind NFTokenOffer there is one 256-bit entry in the directory of the offer's owner
- As long as there is at least one sell offer for the burned NFToken, then there is an NFToken sell offer directory that probably consists of at least two nodes -- the directory root node and one or more nodes containing the remaining sell offer(s).
- As long as there is at least one buy offer for the burned NFToken, then there is an NFToken buy offer directory.  The situation is just like the one for sell offers.

To be fair, every left behind NFTokenOffer uses an owner reserve for the owning account.  So the owner is motivated to eventually cancel the obsolete NFTokenOffer.  But they need not get around to it until they want the XRP that's trapped in the reserve.  There's no reason for them to hurry.

### Another Approach

But we still think it is important for Alice to be able to burn the NFT that she minted and Bob now owns.  If I don't like the solution proposed by this pull request I should propose another solution .

We noted earlier that if Alice owned the NFToken this would be a solvable problem.  Alice could cancel offers for the NFToken in batches until the number of offers fell below 500.  Then the NFToken becomes burnable.

So I propose we allow Alice to cancel offers for NFTokens that Alice issued with the `lsfBurnable` flag set.

The implementation is trivial.  We'd add something like the following lines to NFTokenCancelOffer.cpp:
```
            // The issuer can cancel if the NFToken has the `lsfBurnable`
            // flag set.
            if (auto const tokenID = (*offer)[~sfNFTokenID];
                tokenID &&
                (nft::getFlags(tokenID.value()) & nft::flagBurnable) &&
                nft::getIssuer(tokenID.value()) == account)
                return !ctx.view.rules().enabled(fixUnburnableNFToken);
```
There's a more efficient implementation, but this gets the idea across concisely.  The actual change is not quite this simple, and it would need unit tests.

It feels weird to let Alice cancel an offer that is not Alice's.  But, in reality, the weird thing is that Alice can burn an NFToken that Alice issued but does not own.  That's weird, but that's water under the bridge.  It's the world we live in.

Let's say Alice really does not want account Creep to own the NFToken that Alice issued.  In the world where this alternative is implemented, Alice could watch and carefully cancel any offer that Creep made for the NFToken (which Bob now owns).  It would end up being a race between Alice, Bob, and Creep.  Maybe Creep could get the offer into the ledger and Bob could accept it before Alice detected the offer and canceled it.  Maybe not.  If Creep and Bob tried often enough, they would eventually get the NFToken transferred over to Creep.

So suppose Creep has successfully acquired the NFToken from Bob.  This really steams Alice.  But Alice has recourse since Alice both issued the NFToken _and_ set the `lsfBurnable` flag.  Alice can burn the NFToken that is now in Creep's possession.  That has been true for as long as XLS-20 has been on the network.

### Pros and Cons

The two approaches being discussed have advantages and disadvantages for different users of the ledger.  Here's my view of those tradeoffs.

**Issuers:** They probably prefer the approach in this pull request.  They don't have to deal with the special case that there are too many offers for an NFToken they want to burn.

**NFToken Offer Makers:** They probably prefer the alternative approach.  These folks will not have their offers hanging around adding to their reserve after an NFToken is burned.

**Rippled Node Runners:** They probably prefer the alternative approach.  The alternative approach leaves no junk behind in the ledger after the NFToken is burned.  Less stuff in the ledger means fewer resources required to run a node.

So the two approaches benefit different XRP Ledger users differently.

Thoughts?

scottschurr (commented):
First off, nice tests!  I appreciate that you're carefully testing the proposed change.

I left a number of comments, most of which I think should be addressed.  But I'm very open to being wrong.  Feel free to comment on any of the points I raised and we can discuss the best path forward.

src/ripple/app/tx/impl/NFTokenBurn.cpp
```diff
@@ -104,9 +109,35 @@ NFTokenBurn::doApply()
...
+        // Deletion of all offers.
+        nft::removeTokenOffersWithLimit(
+            view(),
+            keylet::nft_sells(ctx_.tx[sfNFTokenID]),
+            std::numeric_limits<int>::max());
+
+        nft::removeTokenOffersWithLimit(
+            view(),
+            keylet::nft_buys(ctx_.tx[sfNFTokenID]),
+            std::numeric_limits<int>::max());
```
scottschurr, line 139: Just FYI, if we were not removing _all_ of the NFT offers while we're inside this `else`, then this would be a transaction breaking change.  However since we're removing every last one of the NFT offers, it doesn't matter what order we delete them in.  The order of deletion does not affect the resulting metadata.

There's nothing that needs to be changed here.  I'm leaving a note in case you work on other transactors in the future.

scottschurr (changes requested):
There are a couple more things to tidy up. The important one is to move where local variable `pageIndex` is being updated.  A possibly less important suggestion is that you delete sell offers before deleting buy offers.

Once those are fixed I think this pull request meets its goals.

intelliot:
@shawnxie999 says he has addressed Scott’s comments. This is ready for re-review.

src/ripple/app/tx/impl/NFTokenBurn.cpp
```diff
@@ -104,9 +109,35 @@ NFTokenBurn::doApply()
         view().update(issuer);
     }
 
-    // Optimized deletion of all offers.
-    nft::removeAllTokenOffers(view(), keylet::nft_sells(ctx_.tx[sfNFTokenID]));
-    nft::removeAllTokenOffers(view(), keylet::nft_buys(ctx_.tx[sfNFTokenID]));
+    if (ctx_.view().rules().enabled(fixUnburnableNFToken))
+    {
+        // Delete up to 500 offers, but prefers buy offers first
```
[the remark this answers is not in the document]
intelliot: team would like to discuss with @ledhed2222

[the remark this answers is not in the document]
shawnxie999: test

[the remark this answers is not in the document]
shawnxie999: I'm not sure what how severe it is to having the extra directory on the ledger. Down the line, we are looking to implement (from Scott D.)
 
 _When anyone deletes one of their orphaned offers, we should delete as many orphaned offers as we can in that transaction (like as do with expired offers in the payment engine)._ 

So the sell directory is unlikely to hang on the ledger forever, as soon as someone submits a transaction to delete one of their orphaned offers, all the orphaned offers of the burnt deleted will be removed. So I don't think it would be a big problem then.

shawnxie999:
@scottschurr I think if we do implement the following, then having left-over offer wouldn't be as bad
>When anyone deletes one of their orphaned offers, we should delete as many orphaned offers as we can in that transaction (like as do with expired offers in the payment engine).

Eventually all the junk offer will be deleted, if someone remembers to delete one of their offers. And it is highly likely that someone will come back and delete the orphaned offers to reclaim their reserve. I understand it won't solve the entire problem immediately, but it does slightly mitigate the concern regarding left-over ledger mess. With this feature, even though the offer junk will still remain on the ledger for some time, they will eventually be deleted. 

As for allowing allowing issuer to delete offers, from my perspective, it might be too much control in the hands of the issuer, and other users may feel that too.

>But we still think it is important for Alice to be able to burn the NFT that she minted and Bob now owns. If I don't like the solution proposed by this pull request I should propose another solution .

For the approach above, I think don't think it solves the root issue that this PR is trying to address - offers be spammed faster than deleted. This is an unlikely scenario, but possible. So regardless if we give the the permission to the issuer to delete offers, buy offers can still be maliciously spammed and make the NFT unburnable.

src/ripple/app/tx/impl/details/NFTokenUtils.cpp
```diff
@@ -520,34 +520,52 @@ findTokenAndPage(
...
 
-        adjustOwnerCount(
-            view,
-            view.peek(keylet::account(owner)),
-            -1,
-            beast::Journal{beast::Journal::getNullSink()});
+            if (maxDeletableOffers == deletedOffersCount)
+                break;
+        }
+        pageIndex = (*page)[~sfIndexNext];
```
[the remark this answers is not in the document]
scottschurr: Yes, you didn't see a problem, but that still doesn't mean it was a good idea.

> Just to clarify, this situation is a problem because the `view.peek()` returns a shared_ptr right?

Yes, but it's better to take a larger view on this.  For every C++ value you can either have a copy or a reference.  The `shared_ptr` is a kind of reference.  Whenever you have a reference you can't really control when the thing referred to will go away.  In contrast, if you have a (local) copy, then you control all of the interactions with that value, including who has references to it.

That's a generalization.  There are specifics to lifetime rules using `shared_ptr` to `const` or non-`const` values.  But we don't need to explore those specifics here.

As it turns out, the `ApplyView` implementation saves up all of the deletes for the end of the transaction.  That's why you were still able to read `IndexNext` from the page node after the call to `erase`.  But it still was not a good idea.  After the call to `erase` on the node, what you had in your hands is called a "stale reference".  The value you want may still be there, but the implementation has every right to remove it due to some completely different software change.  What you had was a latent crash.  Thankfully that's fixed now.

src/ripple/app/tx/impl/NFTokenBurn.cpp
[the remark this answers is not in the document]
ledhed2222: i'm curious - if i have an NFT with 600 buy offers, which I burn, there are now 100 orphaned offers. what parties are able to cancel the orphaned offers at that point, other than the person who owns the offer (and if they are expired, anyone)? presumably I should still be able to cancel them since I am listed as the "owner" on the offer.

but generally, I don't think i have a huge POV/opinion on which offers get cancelled with higher priority. what scott is saying makes sense. i might go with that.

however, one thing i'm wondering is whether we can included a `warning` if the token burn did not successfully cancel all offers. something like `Not all offers for this NFT could be cancelled. You should immediately cancel any remaining offers`.

scottschurr:
@shawnxie999 
> I don't think it solves the root issue that this PR is trying to address - offers be spammed faster than deleted. This is an unlikely scenario, but possible. So even if we give the permission to the issuer to delete offers, buy offers can still be maliciously spammed and cause the NFT to be unburnable.

This is simply not correct.

Please look at the documentation for `NFTokenCancelOffer`: https://xrpl.org/nftokencanceloffer.html

The `NFTokenCancelOffer` transaction carries an _array_ of offers to cancel.  That array is allowed to contain as many as _500_ entries: https://github.com/XRPLF/rippled/blob/develop/src/ripple/protocol/Protocol.h#L67

In contrast, the `NFTokenCreateOffer` transaction creates _one_ offer per transaction.  Even if someone cobbled together 100 accounts to create NFTokenOffers in parallel, a single account deleting offers can out run that 5:1.

So while, admittedly, the spamming makes it very inconvenient for the issuer of the NFT to burn their NFT, the spam will not prevent them from doing the burn as long as the issuer is allowed to cancel the offers in the first place.

src/ripple/app/tx/impl/NFTokenBurn.cpp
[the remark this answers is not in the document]
scottschurr: Currently there are two parties that have permission to cancel NFTokenOffers, whether those offers are orphaned or not:

- The `Owner` of the offer can always cancel the offer.
-  _If_ there's a `Destination`, then the `Destination` can cancel the offer.

That's assuming that the offer has not expired, in which case the offer can be canceled by anybody.

Here's the code: https://github.com/XRPLF/rippled/blob/develop/src/ripple/app/tx/impl/NFTokenCancelOffer.cpp#L77-L87

Regarding a `warning`, that's not really something that the ledger internals can provide.  There are three types of responses:

- `tesSUCCESS` is what you get if the transaction succeeded.  If the `NFTokenBurn` succeeds, whether it leaves orphan offers or not, the appropriate response is `tesSUCCESS`.
- `tec*` is the range of responses where the ledger did not do what you asked, but burned your fee anyway.  That's not what is happening here, since the burn was successful.
- The are other error types (`ter`, `ter`, and so forth) to indicate other kinds of failures.  But they all mean that your transaction did not change the ledger in any way.

At any rate, such a warning would be returned to the wrong person.

The person who is leaving the orphaned offers is the issuer burning the token.  They really don't care whether there are left over offers in the ledger.  The only folks who care are the offer owners.  And they only care once they want to recover the reserve on their offers.  So they may not be motivated to cancel their offers for a long time.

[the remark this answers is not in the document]
scottschurr: @seelabs' suggestion may help, but I have doubts about its effectiveness in a larger scale.  Here's why.

Offers on the DEX are encountered whenever someone wants to trade a currency pair.  There are some currency pairs that are very popular, so their order books get a lot of traffic.  There are other currency pairs that see a lot less traffic.  For those low-traffic currencies, dead offers can sit around for a long time before the DEX stumbles across them and removes them.

Compared to currency pairs, there are _lots_ of NFTs on the ledger.  And there are up to two order books for _every individual_ NFT in the ledger.  I would be astonished if _any_ NFTs are traded frequently.  The bulk of NFTs, and offers for them, will just sit on the ledger.  Any single NFToken will probably be traded relatively infrequently.

So I doubt that suggestion will help much.

But at a more meta level, what I hear you saying is that the patch you are providing for this problem is insufficient.  We need yet another patch to clean up what is left behind by this patch.  Amendments are not free.  We should get everything that we think is needed into one amendment.

shawnxie999:
@scottschurr This is ready for re-review. Updated to delete sell offers, as well the test cases.

src/ripple/app/tx/impl/NFTokenBurn.cpp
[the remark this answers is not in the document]
shawnxie999: @scottschurr (cc. @seelabs )
>But at a more meta level, what I hear you saying is that the patch you are providing for this problem is insufficient. We need yet another patch to clean up what is left behind by this patch. Amendments are not free. We should get everything that we think is needed into one amendment.

Our team's idea is try to get this change into the next release(1.10), so that we can get voting and hear community's voice as early as possible. The voting may or may not be in our favor, but we want to get it into next release to get a feel what the community likes, and perhaps making more adjustments from there. 

While I do agree that it feels more robust to include the patch of cleaning orphaned offers into the existing one, but unfortunately this what not considered for scoping, and including it would push back the timeline even further.

scottschurr (approved):
Looks like this achieves its intended purpose.  Also, nice tests.

I still disagree with this approach to solving the problem.  And I've not seen any convincing arguments why this approach is better than my preferred solution.  But I'm hoping that the problem will occur seldom enough that this solution not will induce heavy trash accumulation on the ledger.  I'm giving this pull request a hesitant :+1:

shawnxie999:
> Looks like this achieves its intended purpose. Also, nice tests.
> 
> I still disagree with this approach to solving the problem. And I've not seen any convincing arguments why this approach is better than my preferred solution. But I'm hoping that the problem will occur seldom enough that this solution not will induce heavy trash accumulation on the ledger. I'm giving this pull request a hesitant 👍

Thanks for the heads up. Our team want to get this change to the community and get some feedbacks first. With the upcoming plan for fixing NFT remint and orphaned offers, I think we might have enough mitigations for a rare scenario.

src/ripple/app/tx/impl/details/NFTokenUtils.cpp
```diff
@@ -520,34 +520,55 @@ findTokenAndPage(
...
-void
-removeAllTokenOffers(ApplyView& view, Keylet const& directory)
-{
-    view.dirDelete(directory, [&view](uint256 const& id) {
-        auto offer = view.peek(Keylet{ltNFTOKEN_OFFER, id});
 
-        if (!offer)
-            Throw<std::runtime_error>(
-                "Offer " + to_string(id) + " not found in ledger!");
+int
```
drlongle, line 524: Nitpick. ```int```'s size is implementation-dependent. Ideally, we want to control the data size (e.g. ```uint64_t```).

src/ripple/app/tx/impl/details/NFTokenUtils.cpp
```diff
@@ -520,34 +520,55 @@ findTokenAndPage(
...
+
+        auto offerIndexes = page->getFieldV256(sfIndexes);
+
+        // We reverse-iterate the offer directory page to delete all entries.
+        // Deleting an entry in a NFTokenOffer directory page won't cause
+        // entries from other pages to move to the current, so, it is safe to
+        // delete entries one by one in the page. It is required to iterate
+        // backwards to handle iterator invalidation for vector, as we are
+        // deleting during iteration.
+        for (int i = offerIndexes.size() - 1; i >= 0; --i)
```
drlongle, line 554: Nitpick. Depending on the scenario, we could have an underflow here.

Also, it seems that we can use a range-based loop here.

src/ripple/protocol/Protocol.h
```diff
@@ -67,7 +67,7 @@ std::size_t constexpr maxDeletableDirEntries = 1000;
 std::size_t constexpr maxTokenOfferCancelCount = 500;
 
 /** The maximum number of offers in an offer directory for NFT to be burnable */
-std::size_t constexpr maxDeletableTokenOfferEntries = 500;
+int constexpr maxDeletableTokenOfferEntries = 500;
```
drlongle, line 70: Nitpick. Preferrably, we want an unsigned data type here so that the compiler automatically spots a mistake like
`size_t constexpr maxDeletableTokenOfferEntries = -500;`

src/ripple/app/tx/impl/details/NFTokenUtils.cpp
shawnxie999: Looping by index backwards was chosen on purpose to delete the entries from `STVector256`

src/ripple/protocol/Protocol.h
shawnxie999: https://github.com/XRPLF/rippled/pull/4346#discussion_r1072849839
It was discussed here that to allow negative numbers. Because otherwise, the line 
` if (maxDeletableTokenOfferEntries - deletedBuyOffers > 0)`
does not really make sense

drlongle: For unsigned type, we can write
```if (maxDeletableTokenOfferEntries > deletedBuyOffers)```

It seems strange to use a signed type to store a non-negative value.

File: src/ripple/app/tx/impl/details/NFTokenUtils.cpp (hunk 6 above)
Comment (drlongle) [reply src/ripple/app/tx/impl/details/NFTokenUtils.cpp:554]: Since we are using C++20, we could use ```std::ranges::view``` for looping backward

`for (auto const& offerIndex :  offerIndexes | std::views::reverse) {...}`

Depending on whether the underlying data type is primitive or complex, we can use `auto const` or `auto const&`.

File: src/ripple/protocol/Protocol.h (hunk 7 above)
Comment (shawnxie999) [reply src/ripple/protocol/Protocol.h]: cool will change

Comment (scottschurr) [reply src/ripple/protocol/Protocol.h]: FWIW, the change to `int` was prompted by a bug in the original code that involved:

1. unsigned numbers,
2. a subtraction and
3. a comparison for `<=`

The original code ended up, unintentionally, being the equivalent of a `while (true)`.  The bug was a prime illustration of the hazards of unsigned integers in the hands of the unwary.   This hazard is further aggravated by the presence of `auto`.  A very straight forward fix was to convert from using unsigned to signed arithmetic. 

There are sporadic debates within the standards committee where certain influential voices shout that we should be using signed arithmetic as many places as possible.  They champion signed arithmetic precisely because of bugs like the one described above.  Other voices argue that we should consistently use unsigned where unsigned is already being used.

I'm not gonna stick my nose any further into this discussion, other than to provide that background.  I'm sure that @drlongle's changes are a fine way to approach the problem.

File: src/ripple/app/tx/impl/details/NFTokenUtils.cpp (hunk 5 above)
Comment (shawnxie999) [reply src/ripple/app/tx/impl/details/NFTokenUtils.cpp]: would `size_t` work?

File: src/ripple/app/tx/impl/details/NFTokenUtils.cpp (hunk 6 above)
Comment (shawnxie999) [reply src/ripple/app/tx/impl/details/NFTokenUtils.cpp:554]: hmm...it says that `views` is is not in namespace `std`. I think it is because apple clang does not support views, and im on mac

Comment (scottschurr) [reply src/ripple/app/tx/impl/details/NFTokenUtils.cpp:554]: You may need to `#include <ranges>`.  It's also certainly possible that the clang library does not have all of the C++20 features yet.  We've had that problem with other C++20 features.

Comment (shawnxie999) [reply src/ripple/app/tx/impl/details/NFTokenUtils.cpp:554]: Tried that and still won't work. And ranges isn't supported by apple clang either (red box)
image

Comment (shawnxie999) [reply src/ripple/app/tx/impl/details/NFTokenUtils.cpp:554]: Saw on [stackoverflow](https://stackoverflow.com/questions/73628848/is-the-stdviews-namespace-not-available-in-xcodes-c?rq=1)  that `views` isn't in apple clang

Comment (shawnxie999) [reply src/ripple/app/tx/impl/details/NFTokenUtils.cpp:554]: But I think i can just keep index looping as it is, shouldnt be a problem

File: src/ripple/app/tx/impl/details/NFTokenUtils.cpp (hunk 5 above)
Comment (drlongle) [reply src/ripple/app/tx/impl/details/NFTokenUtils.cpp]: Yes. `size_t` is implementation-dependent but it can store the maximum size of any type (including array).

An STL container's functions `size()` returns a `size_t` result.

https://en.cppreference.com/w/cpp/types/size_t

**Comment by intelliot** [comment]
Trivial: Let's consider naming this amendment `NonFungibleTokensV1_2` because additional fixes will be included in it. It's not semver or anything, but at least it makes sense. 

This does not block merging or reviewing - please proceed as before. The name will be finalized later, before `feature/nft-fixes` gets merged into `develop`.

**Review by scottschurr — APPROVED** [review]
I left one comment that requires no action.  I think this code still meets its intended purpose.

File: src/ripple/app/tx/impl/NFTokenBurn.cpp [hunk 8]
```diff
@@ -115,12 +115,12 @@ NFTokenBurn::doApply()
...
         // the number of buy offers, we prioritize the deletion of sell
         // offers in order to clean up sell offer directory
-        int const deletedSellOffers = nft::removeTokenOffersWithLimit(
+        std::size_t const deletedSellOffers = nft::removeTokenOffersWithLimit(
             view(),
             keylet::nft_sells(ctx_.tx[sfNFTokenID]),
             maxDeletableTokenOfferEntries);
 
-        if (maxDeletableTokenOfferEntries - deletedSellOffers > 0)
+        if (maxDeletableTokenOfferEntries > deletedSellOffers)
```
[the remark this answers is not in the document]
shawnxie999: Good point! Didn't know about this jargon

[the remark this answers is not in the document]
intelliot: Confirmed with Shawn that this can be left as-is

Issue #19, opened 2013-03-11
JS: Automatic unsubscribe in browser not working


justmoon:
In several places, ripple-lib subscribes to events automatically when an event listener is added.

To unsubscribe, it listens to the "removeListener" event. However, this is not officially supported: http://nodejs.org/api/events.html

Consequently, it's not implemented in webpack's port of the EventEmitter class, so it doesn't work in the browser version of ripple-lib.

The correct solution depends on whether Node.js developers wish to officially support this feature in the future or not. I asked them: https://github.com/joyent/node/issues/4977

In the meantime, the only practical effect of this bug is that the client stays subscribed to order books after the user has navigated to a different one. (It will not subscribe multiple times to the same order book.) I think this should be ok, since most people will only access a small number of order books over the course of a session. But if it does cause server load issues, please let me know.

Issue #18, opened 2013-03-11
RippleCalc should round up for XRP input


ahbritto:
When delivering USD in small amounts, there is no room for rounding down after going in reverse. Resulting in a path dry.

Issue #17, opened 2013-03-08
Client sync loss


nelisky:
From a couple of days ago the web client appears to lose the websocket connection but not state it is offline. Instead it shows no balance on any currency and no history as if it had been opened with a new wallet.
The problem also shows up in a slightly different way, where balances and history are correct, except no XRP and that makes the client show the initial 'Welcome to Ripple' screen.

I've been seeing this on Chrome 25.0.1364.155, OSX 10.6.8

Issue #16, opened 2013-03-06
path bug from forum


jedmccaleb:
Everything is failing just the same, I'm afraid.

So the workflow that is failing is, in a nutshell:

walletA trusts bitstamp for 1 BTC
walletB trusts bitstamp for 1 BTC
wallet A gets a 0.1 BTC transfer from bitstamp
wallet A cannot send any BTC to wallet B - Paths are calculated but when sending one of partial path or fee insufficient will be thrown.

Are you able to run the workflow described yourself?

[edit] - above is the simplified workflow that is failing now, the original issue was slightly more complicated:

walletA trusts bitstamp for 1 BTC
walletB trusts walletA for 1 BTC
walletB trusts bitstamp for 1 BTC
walletA gets a 0.1 BTC transfer from bitstamp
walletA sends 0.05 BTC to walletB - it appears as walletA BTC IOU
walletB cannot send any BTC to bitstamp - Paths are calculated but when sending one of partial path or fee insufficient will be thrown.

PR #4349, opened 2022-11-20
Add true partitioning support to TaggedCache:

nbougalis:

The primary change introduced in this commit is partitioning of the `TaggedCache`, with each partition being independent of each other, making it possible to potentially perform multiple cache operations in parallel. In particular, the `sweep` operation is now parallelized by default on systems with at least four cores present.

The `TaggedCache` could also be instantiated in 'key-only' mode which complicated the interface significantly but was only used by a single consumer (the `FullBelowCache`). This commit splits the 'key-only' functionality of `TaggedCache`, and incorporates directly into `FullBelowCache`, resulting in simple and cleaner interfaces for both `TaggedCache` and `FullBelowCache` but at a cost: some code duplication.

Lastly, this commit includes a medley of changes, including the restructuring of `Transaction`, reducing its size by 48 bytes.

### Type of Change


scottschurr:
I just tried building this under MacOS.  The news is not good.  Here's the first compilation error I get:
```
/Users/scott/Development/rippled/src/ripple/basics/TaggedCache.h:574:26: error: no member named 'execution' in namespace 'std'; did you mean 'exception'?
                    std::execution::sequenced_policy>)
                    ~~~~~^~~~~~~~~
                         exception
```
TaggedCache.h does. properly include `<execution>`; that must not be the problem.  So I took a look inside the `<execution>` header on my machine.  In total it looks like this:
```
// -*- C++ -*-
//===----------------------------------------------------------------------===//
//
// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
// See https://llvm.org/LICENSE.txt for license information.
// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
//
//===----------------------------------------------------------------------===//

#ifndef _LIBCPP_EXECUTION
#define _LIBCPP_EXECUTION

#include <__config>
#include <version>

#if defined(_LIBCPP_HAS_PARALLEL_ALGORITHMS) && _LIBCPP_STD_VER >= 17
#   include <__pstl_execution>
#endif

#if !defined(_LIBCPP_HAS_NO_PRAGMA_SYSTEM_HEADER)
#pragma GCC system_header
#endif

#endif // _LIBCPP_EXECUTION
```
I looked for a file named __pstl_execution on my machine and found nothing.

So It looks to me like, at the moment, the clang library has no support for the contents of the `<execution>` header.

nbougalis:
Thanks. I guess I can just remove the dependency on `<execution>` altogether since I'm not using the parallel support of the standard library anyways to avoid the library doing weird things (like running `std::execution::par` as if it was really `std::execution::seq`!) under me.

scottschurr:
More build issues coming from clang.  I see two this time:
```
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX12.3.sdk/usr/include/c++/v1/atomic:963:37:     : public __atomic_base<_Tp>error: 
             ^_Atomic cannot be applied to type 'std::shared_ptr<ripple::detail::FullBelowCacheImpl<ripple::base_uint<256>>>' which is not trivially copyable

  _LIBCPP_DISABLE_EXTENSION_WARNING _Atomic(_Tp) __a_value;
                                    ^/Users/scott/Development/rippled/src/ripple/shamap/FullBelowCache.h
```
and this one...
```
/Users/scott/Development/rippled/src/ripple/shamap/FullBelowCache.h:523:44: error: private field 'cache_' is not used [-Werror,-Wunused-private-field]
    std::atomic<std::shared_ptr<KeyCache>> cache_;
                                           ^
```
Maybe @HowardHinnant can provide some insight on the `std::atomic<std::shared_ptr<>>` issue?

scottschurr:
Yes, my understanding is that it's a standard C++20 feature.  I've heard folks muttering that the clang standard library is sadly trailing the standard.  But @HowardHinnant is probably more informed on that topic than I am.

HowardHinnant:
According to this page: https://libcxx.llvm.org/Status/Cxx20.html, work on atomic shared_ptr has not yet begun in libc++.

scottschurr:
Hmmmm.  This table suggests that clang has not implemented `atomic<shared_ptr>` yet: https://en.cppreference.com/w/cpp/compiler_support#C.2B.2B20_library_features

nbougalis:
Well that kinda sucks. We can either roll back to things like `std::atomic_load` and `std::atomic_store`, but those are _deprecated_ in C++20 in favor of `std::atomic<std::shared_ptr>>`.

OSX isn't an "officially" supported platform (whatever _that_ means) but several people do their day-to-day on OSX, so breaking that build doesn't seem like a good idea. What would be your preference?

scottschurr:
Yes, it's not a great situation.  I don't have any great answers.  Unless @HowardHinnant feels like he could roll an `atomic<shared_ptr>` for us, I think the only choice will be using things like `std::atomic_load` and `std::atomic_store`.  They may be officially deprecated, but we probably won't get deprecation warnings for a while.

HowardHinnant:
I haven't looked into this, but perhaps it might be easy to roll a non-optimized `atomic<shared_ptr>` using an internal `mutex`, and `#ifdef` its use just for those platforms where we don't have the `std` one yet.

HowardHinnant:
Slaps forehead:  this should be easy.  On those platforms where `atomic<shared_ptr>` isn't implemented yet, it is likely that `atomic_load` (et al) on `shared_ptr` hasn't yet been marked as deprecated.  So on those platforms we can just make `atomic<shared_ptr>` a wrapper around the (not yet deprecated on that platform) `atomic_load` API.

intelliot:
(notes: changes many things, large - potentially risky. Needs a lot of testing before including it in a stable release)

intelliot:
questions:
- what's the purpose of this change?
- is someone asking for this change?


note: @mtrippled will share his findings regarding this change

mtrippled:
This degrades throughput in testing. The results from this branch don't look too good--previously I've tested 1.10.0-b1 to do about 2200/s (similar to tests from May 2022), but this only had peak sustained of 1984/s.

nbougalis:
> questions:
> 
> * what's the purpose of this change?
> * is someone asking for this change?
> 
> note: @mtrippled will share his findings regarding this change

Increased parallelization and a proper implementation of the parallel tagged cache support originally implemented by Mark T.

It’s possible the benchmarks show a regression and we’ll need to understand why, but I’m confident this is a net gain, module any small hiccups.

sophiax851 (commented):
To improve performance, we'd normally first get to know our bottleneck(s), then test to see the effect of the change. And as bottleneck goes, it always the most imminent one restricts the overall performance. So for this change to be effectively improve the performance, the concurrent access of the TaggedCache has to be one of the current bottleneck(s). I do see one of the operation that was often blocked during the Accept phase that maybe related to the TaggedCache lookup, if that's the case, then this change is promising. Here is that code block and the call trace:
void
LedgerMaster::checkAccept(uint256 const& hash, std::uint32_t seq)
{
    std::size_t valCount = 0;

    if (seq != 0)
    {
        // Ledger is too old
        if (seq < mValidLedgerSeq)
            return;

        auto validations = app_.validators().negativeUNLFilter(
            app_.getValidations().getTrustedForLedger(hash));
        valCount = validations.size();
        if (valCount >= app_.validators().quorum())
        {
            std::lock_guard ml(m_mutex);        <---------- This lock
            if (seq > mLastValidLedger.second)
                mLastValidLedger = std::make_pair(hash, seq);
        }

Call stack:

Thread 14 (Thread 0x7f90a27fc700 (LWP 1400574)):
#0  __lll_lock_wait (futex=futex@entry=0x55a85d56e0f8, private=0) at lowlevellock.c:52
#1  0x00007f90c4e74131 in __GI___pthread_mutex_lock (mutex=0x55a85d56e0f8) at ../nptl/pthread_mutex_lock.c:115
#2  0x000055a85b8a8fe7 in __gthread_mutex_lock (__mutex=<optimized out>) at /usr/include/x86_64-linux-gnu/c++/9/bits/gthr-default.h:749
#3  __gthread_recursive_mutex_lock (__mutex=<optimized out>) at /usr/include/x86_64-linux-gnu/c++/9/bits/gthr-default.h:811
#4  std::recursive_mutex::lock (this=<optimized out>) at /usr/include/c++/9/mutex:106
#5  0x000055a85bd024ac in std::lock_guard<std::recursive_mutex>::lock_guard (__m=..., this=<synthetic pointer>) at /usr/include/c++/9/bits/std_mutex.h:158
#6  ripple::LedgerMaster::checkAccept (this=0x55a85d56e0e0, hash=..., seq=seq@entry=75833166) at /space/tools/newAMM/rippled/src/ripple/app/ledger/impl/LedgerMaster.cpp:994
#7  0x000055a85bd3aaf6 in ripple::handleNewValidation (app=..., val=..., source=...) at /space/tools/newAMM/rippled/src/ripple/app/consensus/RCLValidations.cpp:180
#8  0x000055a85bb98a3e in ripple::NetworkOPsImp::recvValidation (this=this@entry=0x55a85d56f820, val=..., source=...) at /space/tools/newAMM/rippled/src/ripple/app/misc/NetworkOPs.cpp:2265
#9  0x000055a85b8ff021 in ripple::PeerImp::checkValidation (this=0x7f90bc022950, val=..., key=..., packet=...) at /usr/include/c++/9/bits/basic_string.h:6555
#10 0x000055a85b8ff202 in ripple::PeerImp::<lambda()>::operator() (__closure=<optimized out>) at /space/tools/newAMM/rippled/src/ripple/overlay/impl/PeerImp.cpp:2615
#11 ripple::ClosureCounter<void>::Substitute<ripple::PeerImp::onMessage(const std::shared_ptr<protocol::TMValidation>&)::<lambda()> >::operator() (this=<optimized out>) at /space/tools/newAMM/rippled/src/ripple/core/ClosureCounter.h:140
#12 std::_Function_handler<void(), ripple::ClosureCounter<void>::Substitute<ripple::PeerImp::onMessage(const std::shared_ptr<protocol::TMValidation>&)::<lambda()> > >::_M_invoke(const std::_Any_data &) (__functor=...) at /usr/include/c++/9/bits/std_function.h:300
#13 0x000055a85ba6b71a in std::function<void ()>::operator()() const (this=0x7f90a27fb710) at /usr/include/c++/9/bits/std_function.h:683
#14 ripple::Job::doJob (this=this@entry=0x7f90a27fb700) at /space/tools/newAMM/rippled/src/ripple/core/impl/Job.cpp:68
#15 0x000055a85ba54399 in ripple::JobQueue::processTask (this=0x55a85d52fa90, instance=9) at /space/tools/newAMM/rippled/src/ripple/core/impl/JobQueue.cpp:374
#16 0x000055a85ba53517 in ripple::Workers::Worker::run (this=0x55a85d5311a0) at /space/tools/newAMM/rippled/src/ripple/core/impl/Workers.cpp:241
#17 0x000055a85c8a6654 in execute_native_thread_routine ()
#18 0x00007f90c4e71609 in start_thread (arg=<optimized out>) at pthread_create.c:477
#19 0x00007f90c4d94133 in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95

Regardless, if this is not urgent, I can take it in my lab to do a full evaluation after we are done with other projects.

src/ripple/shamap/FullBelowCache.h
```diff
@@ -114,38 +457,74 @@ class BasicFullBelowCache
...
+    beast::Journal journal_;
+    beast::insight::Collector::ptr collector_;
+    std::size_t const targetSize_;
+    std::chrono::seconds const expiration_;
 
-}  // namespace detail
-
-using FullBelowCache = detail::BasicFullBelowCache;
+    std::atomic<std::uint32_t> generation_ = 0;
+    std::atomic<std::shared_ptr<KeyCache>> cache_;
```
HowardHinnant, line 523: This: https://github.com/HowardHinnant/rippled/commit/ef666ae4951ba30ca521109defc5b5a59ef172ff must be merged in to get things to compile on the latest macOS tools.  Otherwise:
```
error: _Atomic cannot be applied to type 'std::shared_ptr<ripple::detail::FullBelowCacheImpl<ripple::base_uint<256>>>' which is not trivially copyable
  _LIBCPP_DISABLE_EXTENSION_WARNING _Atomic(_Tp) __a_value;
                                    ^
```

HowardHinnant (changes requested):

intelliot:
@nbougalis what should be the next step for this PR?

nbougalis:
I'll rebase it, and maybe try to figure out if there's a reason for the performance degradation that Mark noticed. After that, and assuming nobody objects, it can be merged.

mtrippled:
> I'll rebase it, and maybe try to figure out if there's a reason for the performance degradation that Mark noticed. After that, and assuming nobody objects, it can be merged.

I don't think it should be merged unless the degradation is resolved. I don't feel I currently have the time to figure it out or keep retesting it. Maybe next quarter though.

nbougalis:
I made some minor changes which _may_ affect performance (the parallelization factor goes from 256-way to 64-way, and the number of threads is fixed at 4 instead of being dynamically adjusted). 

With that said, I am skeptical that this code is responsible for actual performance degradation and it does have measurable performance improvements on the periodic sweeps we execute because it can _now_ truly execute those in parallel.

Here's an example of the improvements without this patch:

```
2023-May-02 03:38:15.042064917 UTC NodeFamily:FTL [TreeNodeCache: NodeFamily] sweep: Iteration over 19475585 items took 5.999 seconds.
2023-May-02 03:38:21.310557087 UTC NodeStore:FTL [DatabaseNodeImp] sweep: Iteration over 19487333 items took 6.0 seconds.
2023-May-02 03:40:28.398042791 UTC NodeFamily:FTL [TreeNodeCache: NodeFamily] sweep: Iteration over 20268742 items took 6.0 seconds.
2023-May-02 03:40:34.909957824 UTC NodeStore:FTL [DatabaseNodeImp] sweep: Iteration over 20271501 items took 6.999 seconds.
2023-May-02 03:42:47.428984934 UTC NodeFamily:FTL [TreeNodeCache: NodeFamily] sweep: Iteration over 11862447 items took 12.0 seconds.
2023-May-02 03:42:53.709914301 UTC NodeStore:FTL [DatabaseNodeImp] sweep: Iteration over 20319469 items took 7.0 seconds.
```

With this patch, the sweep _rarely_ exceeds 2 seconds, which is a massive win IMO, _and_ on top of that the caches can still be used during a sweep, since locking is now, _truly_ per-partition, meaning that only operations on the partition(s) actively being sweep'd are delayed.

Whatever you want to do with this is fine by me, but I don't really have the time (or, to be honest, the interest) in rebasing this on an ongoing basis to avoid bit-rot.

mtrippled:
@nbougalis Thanks, Nik. I'll rebase this when I'm ready to test it again, and then look for problems, if any. I don't know why it didn't test well the last time I tried it, but it needs further analysis. On paper it looks good, and I know it should actually help. Anyway, @intelliot , I'll rebase this in the future when I test it, and if it looks good then, we should merge it. I expect some time Q3.

nbougalis:
Thanks @mtrippled. Sorry if my previous comment came across as overly harsh; it wasn't my intention. Anyways, curious to see what happens when you get to this.

I think I may try a further refinement in the meantime (avoiding `std::weak_ptr` entirely, and only storing strong pointers instead, which will reduce the complexity of the code further and eliminate a lot of "tickling" of the `std::variant` I'm using (implementations of `std::variant` can be flaky).

mtrippled:
I wasn't offended, nor noticed anything harsh. Anyway, keep adding to this
if you think it improves things, or not, as you feel like. I'll give it
some time later on and if I run into trouble I'll let you know.

intelliot:
This is not a priority and there is no evidence of benefit. Feel free to open a new PR (against the head of `develop`) if you still feel this is important to do. Closing due to inactivity.

nbougalis:
I think this changeset was a general improvement in an especially old and crufty part of the codebase, and merited merging even if it provided no measurable performance improvements during synthetic performance test.

No further action needed, but for whatever it's worth, I wouldn't have unilaterally closed this (or any other) PR on grounds of what Ripple considers a "priority", especially considering that priorities are relative.

Issue #4341, opened 2022-11-11
[ripple node is stopped many times] (1.9.4+ba3c0e51455a88d76d90b996f20c0f102ac3f5a0.DEBUG)


past2017:
2022-Nov-11 08:16:22.234032320 UTC LoadMonitor:WRN Job: processLedgerData run: 1986ms wait: 0ms
2022-Nov-11 08:16:23.513537418 UTC LoadMonitor:WRN Job: processLedgerData run: 1153ms wait: 0ms
2022-Nov-11 08:16:41.381873448 UTC LoadMonitor:WRN Job: gotFetchPack run: 2492ms wait: 0ms
2022-Nov-11 08:16:41.575358272 UTC LoadMonitor:WRN Job: processLedgerData run: 1150ms wait: 0ms
2022-Nov-11 08:16:52.974906477 UTC LoadMonitor:WRN Job: processLedgerData run: 1008ms wait: 0ms
2022-Nov-11 08:16:53.051118374 UTC LoadMonitor:WRN Job: processLedgerData run: 1289ms wait: 0ms
2022-Nov-11 08:16:53.716542567 UTC LoadMonitor:WRN Job: gotFetchPack run: 741ms wait: 523ms
2022-Nov-11 08:16:56.235694966 UTC LoadMonitor:WRN Job: processLedgerData run: 1139ms wait: 0ms
2022-Nov-11 08:16:56.414271764 UTC LoadMonitor:WRN Job: processLedgerData run: 1173ms wait: 0ms
2022-Nov-11 08:17:00.844353959 UTC LoadMonitor:WRN Job: processLedgerData run: 646ms wait: 473ms
2022-Nov-11 08:17:00.918079269 UTC LoadMonitor:WRN Job: gotFetchPack run: 1352ms wait: 154ms
2022-Nov-11 08:17:11.624420645 UTC LoadMonitor:WRN Job: gotFetchPack run: 2469ms wait: 0ms
2022-Nov-11 08:17:15.716968443 UTC LoadMonitor:WRN Job: processLedgerData run: 1091ms wait: 0ms
2022-Nov-11 08:17:15.893754960 UTC LoadMonitor:WRN Job: processLedgerData run: 1387ms wait: 0ms
2022-Nov-11 08:17:15.894065428 UTC LoadMonitor:WRN Job: processLedgerData run: 0ms wait: 1240ms
[...]

PR #4351, opened 2022-11-21
Don't try to read SLE with key 0 from the ledger:

ximinez:

Debug builds are occasionally hitting an `assert` at Ledger.cpp line 436 (see #4341). The `assert` is at least 8 years old and appears to still be relevant. `Ledger::read` should not be called with an ID of 0.

`NFTokenAcceptOffer` transactions can be written (presumably incorrectly) in such a way that the `NFTokenSellOffer` and `NFTokenBuyOffer` fields can contain a 0 value. The `preclaim` function for this transaction did not validate the value before calling `Ledger::read`.

Note that this issue is **not** harmful in release builds. If an `assert` fails in a debug build, as it did in this case, it will stop the program. However, `assert`s are not included in release builds, so they are skipped. In this case, `Ledger::read` properly handles that case, and returns a `nullptr` if the ID is 0. `NFTokenSellOffer::preclaim` checks that `nullptr` and correctly returns `tecOBJECT_NOT_FOUND`, as it would if the transaction provided a valid ID which was not in the ledger.

This fix preserves that functionality, but checks for 0 in `NFTokenSellOffer::preclaim` before calling `Ledger::read`, and returns the same `tecOBJECT_NOT_FOUND` result. Thus, an amendment is **not** required.

### Type of Change

### DIFF
--- a/src/ripple/app/tx/impl/NFTokenAcceptOffer.cpp
+++ b/src/ripple/app/tx/impl/NFTokenAcceptOffer.cpp
@@ -67,6 +67,9 @@ NFTokenAcceptOffer::preclaim(PreclaimContext const& ctx)
         -> std::pair<std::shared_ptr<const SLE>, TER> {
         if (id)
         {
+            if (id->isZero())
+                return {nullptr, tecOBJECT_NOT_FOUND};
+
             auto offerSLE = ctx.view.read(keylet::nftoffer(*id));
 
             if (!offerSLE)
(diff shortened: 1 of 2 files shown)


scottschurr (commented):
Change looks good, and I think this is a good approach to avoiding an amendment.  Thanks for finding and fixing the problem.

But I do think it would be good to add unit tests.  I don't think the tests need to be extensive.  The topmost commit here I think would be sufficient: https://github.com/scottschurr/rippled/commits/ed-readidzero

ximinez:
Thanks for the tests! Super helpful. I also verified that the `assert` fires without the fix in place.

scottschurr (approved):
:+1: Looks great!  Thanks.

nbougalis (approved):

nbougalis (approved):

intelliot (approved):

PR #4357, opened 2022-11-30
Update Github Actions to ubuntu-latest:

ximinez:

Per actions/runner-images#6002, the ubuntu-18.04 image is being deprecated in Github Actions. They are having scheduled "brownouts" to drive the point home. That could get pretty annoying if the timing is bad.

Instead of picking a version, I changed the jobs to use `ubuntu-latest`. This will reduce maintenance as new versions are introduced, if the jobs keep working. And if they stop working, we'll need to fix them anyway, so this will let us catch it early.

### Type of Change

## Test Plan

No specific testing is needed. If the Github Actions jobs work, that's it.

## Future Tasks

Jobs in pending PRs that affect Github Actions may need to be updated. E.g. #4223 #3851

### DIFF
--- a/.github/workflows/clang-format.yml
+++ b/.github/workflows/clang-format.yml
@@ -4,7 +4,7 @@ on: [push, pull_request]
 
 jobs:
   check:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     env:
       CLANG_VERSION: 10
     steps:
--- a/.github/workflows/clang-format.yml
+++ b/.github/workflows/clang-format.yml
@@ -11,7 +11,7 @@ jobs:
     - uses: actions/checkout@v2
     - name: Install clang-format
       run: |
-        codename=$( lsb_release -cs )
+        codename=$( lsb_release --codename --short )
         sudo tee /etc/apt/sources.list.d/llvm.list >/dev/null <<EOF
         deb http://apt.llvm.org/${codename}/ llvm-toolchain-${codename}-${CLANG_VERSION} main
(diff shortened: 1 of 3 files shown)


.github/workflows/clang-format.yml
```diff
@@ -4,16 +4,17 @@ on: [push, pull_request]
...
   check:
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-latest
     env:
       CLANG_VERSION: 10
     steps:
     - uses: actions/checkout@v2
     - name: Install clang-format
       run: |
+        codename=$( lsb_release -cs )
```
thejohnfreeman, line 14: I generally prefer to use long option names in scripts, but it's not a blocker.

thejohnfreeman (approved):

gregtatcam (approved):
LGTM

.github/workflows/clang-format.yml
ximinez: I think that sounds good. Updated.

ximinez:
I just noticed the `clang-format` job is failing in this PR, but working on my branch. On my branch, `ubuntu-latest` is giving me `ubuntu-20.04` (focal). On the PR, `ubuntu-latest` is getting `ubuntu-22.04` (jammy). And just to make things fun, LLVM for jammy needs a minimum version of 13 (https://apt.llvm.org/jammy/dists/), while focal goes all the way down to 9 (https://apt.llvm.org/focal/dists/).

So, for now, I'm going to force that action back to `ubuntu-20.04`, but we'll need to decide if we want to standardize on a new version before that one gets deprecated.

Issue #4120, opened 2022-03-15
Proposal: Eliminate custom issued token math


mDuo13:
## Summary
Would it be possible to eliminate the custom number format the XRPL uses to represent issued tokens, and migrate to strictly integer math? I think it could be possible and we should really consider it.

## Motivation

[The XRP Ledger's current fungible token math](https://github.com/ripple/rippled/blob/1a8eb5e6e3b1c844ae36bb3b7d363d76269ea450/src/ripple/basics/IOUAmount.h) is entirely custom and covers many edge cases—_too many_ edge cases. The result is more pitfalls for developers than useful features.

Different trust lines can hold the same tokens to operate at different scales depending on the amounts involved. At the extremes, I could hold something in excess of 123×10<sup>80</sup> of some token, and you could hold 123×10<sup>-80</sup> of the same token from the same issuer. Here, let me write those amounts out _without_ scientific notation to really emphasize the insane scale difference between them:

I have: $12,300,000,000,000,000,000,000,000,000,000,000,000,000,000,000,000,000,000,000,000,000,000,000,000,000,000,000
You have: $0.0000000000000000000000000000000000000000000000000000000000000000000000000000000123

There is simply no reasonable way for these amounts to interact. As a result, weird edge cases happen. Like, if I send you $10<sup>20</sup>, my balance doesn't even decrease; new tokens are created via rounding up. User interfaces to the XRPL have the challenge of needing to represent all these amounts to users—admittedly, they probably just resort to scientific notation, but there are reasons people don't typically use scientific notation with money. 

This custom number format means that it's challenging to do math off-ledger, like summing up the balance of a set of trust lines or offers. Not all programming languages have easy access to a specialized, arbitrary-precision decimal library, but even if you do, you will probably end up with slightly different results than the ledger itself uses. To replicate the ledger's math exactly, you would have to write custom low-level math functions that match the C++ implementations.

This power isn't really necessary; heck, it's estimated there are only about 10<sup>80</sup> atoms in the universe, which is less than the amount of tokens you could issue on each of multiple trust lines. Even the most hyperinflated currencies in history didn't really reach those levels. (Fun fact, the largest bank note ever issued was a Hungarian pengo for 10<sup>20</sup>.)

## Goal State

[...]

PR #4359, opened 2022-12-01
Fix edge case with delivered_amount for tiny payments:

RichardAH:
When a Payment Transaction for an IOU is executed and the receiving balance is much larger than the amount sent, then (due to floating point mechanics) no changes occur to the RippleState objects. Despite this the `delivered_amount` field, relied on heavily by exchanges, will report that the full amount was delivered, when in fact it wasn't and no balance changes occurred at all.

Example transaction:
https://hash.xrp.fans/7043CE511E9E0223ABBBDD004D279A1F16C18DEC012FAE856FDA4122D646DC61

```json
{
  "result": {
    "Account": "rMJgdYV2ffVfLj3bQqXSDbuwYcHA1hC5XD",
    "Amount": {
      "currency": "ABC",
      "issuer": "rMJgdYV2ffVfLj3bQqXSDbuwYcHA1hC5XD",
      "value": "1"
    },
    "Destination": "rpPbXmxxwM4EhYNjHrjNWnWycR4BCcz7U5",
    "Fee": "15",
    "LastLedgerSequence": 76131423,
    "Sequence": 141,
    "SigningPubKey": "02F2ACB6FB9A394223D2158ECA447F8A01BFFAF3811957227669BDE939F4E06024",
    "TransactionType": "Payment",
    "TxnSignature": "304402203E9F36E00A5DCBF75A3568D1C05C3A094B9885078D851797F8908EA8C0836F4D0220571B9BE7E64AD947890603B96E22B94E02DA81436B060364D0BA7BECA3DA3085",
    "date": 723217311,
    "hash": "7043CE511E9E0223ABBBDD004D279A1F16C18DEC012FAE856FDA4122D646DC61",
    "inLedger": 76131415,
    "ledger_index": 76131415,
    "meta": {
      "AffectedNodes": [
        {
          "ModifiedNode": {
            "FinalFields": {
              "Account": "rMJgdYV2ffVfLj3bQqXSDbuwYcHA1hC5XD",
              "Balance": "164771284",
              "Flags": 9437184,
              "OwnerCount": 4,
              "RegularKey": "rMCouhZ7x83FDCgXfaew3pfn4CD15uD4sv",
              "Sequence": 142
            },
            "LedgerEntryType": "AccountRoot",
            "LedgerIndex": "130E1F264151AD80B984737FBB18B4F73FBAA7915FE8748F34D70C25FCDEDA40",
            "PreviousFields": {
              "Balance": "164771299",
              "Sequence": 141
            },
            "PreviousTxnID": "F528C4009E2E4B817CEA90738EE0ADFC10E28CBF1F61D8D3CB162A6FF2E2EB6C",
            "PreviousTxnLgrSeq": 76131394
          }
        }
      ],
      "TransactionIndex": 23,
      "TransactionResult": "tesSUCCESS",
      "delivered_amount": {
        "currency": "ABC",
        "issuer": "rMJgdYV2ffVfLj3bQqXSDbuwYcHA1hC5XD",
        "value": "1"
      }
    },
    "validated": true,
    "warnings": [
      {
        "id": 1004,
        "message": "This is a reporting server.  The default behavior of a reporting server is to only return validated data. If you are looking for not yet validated data, include \"ledger_index : current\" in your request, which will cause this server to forward the request to a p2p node. If the forward is successful the response will include \"forwarded\" : \"true\""
      }
    ]
  },
  "status": "success",
  "type": "response"
}
```

This PR fixes this edge case by computing the actual difference between the Previous and Final fields of the relevant RippleState object according to the transaction metadata. If for some reason this can't be calculated then the RPC caller sees: `"delivered_amount": "unavailable"` for this k-v pair.

The same txn running on this PR:
```json
{
  "result": {
    "Account": "rMJgdYV2ffVfLj3bQqXSDbuwYcHA1hC5XD",
    "Amount": {
      "currency": "ABC",
      "issuer": "rMJgdYV2ffVfLj3bQqXSDbuwYcHA1hC5XD",
      "value": "1"
    },
    "Destination": "rpPbXmxxwM4EhYNjHrjNWnWycR4BCcz7U5",
    "Fee": "15",
    "LastLedgerSequence": 76131423,
    "Sequence": 141,
    "SigningPubKey": "02F2ACB6FB9A394223D2158ECA447F8A01BFFAF3811957227669BDE939F4E06024",
    "TransactionType": "Payment",
    "TxnSignature": "304402203E9F36E00A5DCBF75A3568D1C05C3A094B9885078D851797F8908EA8C0836F4D0220571B9BE7E64AD947890603B96E22B94E02DA81436B060364D0BA7BECA3DA3085",
    "date": 723217311,
    "hash": "7043CE511E9E0223ABBBDD004D279A1F16C18DEC012FAE856FDA4122D646DC61",
    "inLedger": 76131415,
    "ledger_index": 76131415,
    "meta": {
      "AffectedNodes": [
        {
          "ModifiedNode": {
            "FinalFields": {
              "Account": "rMJgdYV2ffVfLj3bQqXSDbuwYcHA1hC5XD",
              "Balance": "164771284",
              "Flags": 9437184,
              "OwnerCount": 4,
              "RegularKey": "rMCouhZ7x83FDCgXfaew3pfn4CD15uD4sv",
              "Sequence": 142
            },
            "LedgerEntryType": "AccountRoot",
            "LedgerIndex": "130E1F264151AD80B984737FBB18B4F73FBAA7915FE8748F34D70C25FCDEDA40",
            "PreviousFields": {
              "Balance": "164771299",
              "Sequence": 141
            },
            "PreviousTxnID": "F528C4009E2E4B817CEA90738EE0ADFC10E28CBF1F61D8D3CB162A6FF2E2EB6C",
            "PreviousTxnLgrSeq": 76131394
          }
        }
      ],
      "TransactionIndex": 23,
      "TransactionResult": "tesSUCCESS",
      "delivered_amount": "unavailable"
    },
    "validated": true
  },
  "status": "success",
  "type": "response"
}
```

While this will probably never be an issue for "real" (i.e. fiat) currencies, it should still be fixed. `delivered_amount` needs to be highly reliable to maintain confidence in the ledger.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### DIFF
--- a/src/ripple/rpc/impl/DeliveredAmount.cpp
+++ b/src/ripple/rpc/impl/DeliveredAmount.cpp
@@ -54,27 +54,160 @@ getDeliveredAmount(
     if (!serializedTx)
         return {};
 
-    if (transactionMeta.hasDeliveredAmount())
+    // The field transactionMeta.hasDeliveredAmount value is always correct
+    // for partial payments, but is not 100% guaranteed to be correct for
+    // CheckCash.
+    bool const isCheckCash =
+        serializedTx->getFieldU16(sfTransactionType) == ttCHECK_CASH;
+
+    if (transactionMeta.hasDeliveredAmount() && !isCheckCash)
     {
         return transactionMeta.getDeliveredAmount();
     }
 
-    if (serializedTx->isFieldPresent(sfAmount))
-    {
-        using namespace std::chrono_literals;
+    using namespace std::chrono_literals;
 
-        // Ledger 4594095 is the first ledger in which the DeliveredAmount field
-        // was present when a partial payment was made and its absence indicates
-        // that the amount delivered is listed in the Amount field.
-        //
-        // If the ledger closed long after the DeliveredAmount code was deployed
-        // then its absence indicates that the amount delivered is listed in the
-        // Amount field. DeliveredAmount went live January 24, 2014.
-        // 446000000 is in Feb 2014, well after DeliveredAmount went live
-        if (getLedgerIndex() >= 4594095 ||
-            getCloseTime() > NetClock::time_point{446000000s})
+    // Ledger 4594095 is the first ledger in which the DeliveredAmount field
+    // was present when a partial payment was made and its absence indicates
+    // that the amount delivered is listed in the Amount field.
+    //
+    // If the ledger closed long after the DeliveredAmount code was deployed
+    // then its absence indicates that the amount delivered is listed in the
+    // Amount field. DeliveredAmount went live January 24, 2014.
+    // 446000000 is in Feb 2014, well after DeliveredAmount went live
+    if (isCheckCash || getLedgerIndex() >= 4594095 ||
+        getCloseTime() > NetClock::time_point{446000000s})
+    {
+        bool const hasAmount = serializedTx->isFieldPresent(sfAmount);
+        bool const hasDeliverMin = serializedTx->isFieldPresent(sfDeliverMin);
+
+        if (hasAmount || hasDeliverMin)
         {
-            return serializedTx->getFieldAmount(sfAmount);
+            // This code previously simply returned the Amount field of the
+            // transaction.  However there are corner cases where that fails,
+            // in part because of the floating-point like limitations of
+            // non-XRP STAmounts.
+            //
+            // Consider an account with a very large IOU balance on a trust
+            // line receiving a tiny IOU payment.  If the payment is small
+            // enough, then the very large IOU balance may not change at all.
+            // So, in effect, no money changed hands.
+            //
+            // To accommodate this, we compute the actual difference between
+            // the Previous and Final fields of the relevant RippleState
+            // object according to the transaction metadata. If for some
+            // reason this can't be calculated then an empty optional is
+            // returned and the RPC caller sees: "delivered_amount":
+            // "unavailable".
+
+            STObject const meta = transactionMeta.getAsObject();
+
+            // if there are no modified fields (somehow) then nothing was
+            // delivered
+            if (!meta.isFieldPresent(sfAffectedNodes))
+                return {};
+
+            // CheckCash may specify DeliverMin instead of Amount
+            auto const txAmt = serializedTx->getFieldAmount(
+                hasAmount ? sfAmount : sfDeliverMin);
+
+            // if it's XRP then it was all delivered if the payment was
+            // successful
+            if (isXRP(txAmt))
+                return transactionMeta.hasDeliveredAmount()
+                    ? transactionMeta.getDeliveredAmount()
+                    : txAmt;
+
+            // It's not XRP.  Return the difference between final and initial
+            // fields in the metadata.
+
+            // in a CheckCash txn the destination is the sender.  Otherwise
+            // we need an explicit destination.
+            if (!isCheckCash && !serializedTx->isFieldPresent(sfDestination))
+                return {};
+
+            // get the destination
+            auto const txAcc = serializedTx->getAccountID(
+                isCheckCash ? sfAccount : sfDestination);
+
+            // get the issuer/currency
+            auto const issue = txAmt.issue();
+
+            // place the accounts into the canonical ripple state order
+            auto const& accA = issue.account < txAcc ? issue.account : txAcc;
+            auto const& accB = issue.account < txAcc ? txAcc : issue.account;
+
+            // search the Affected nodes in the metadata to find the receiving
+            // trustline
+            for (auto const& node : meta.getFieldArray(sfAffectedNodes))
+            {
+                uint16_t nodeType = node.getFieldU16(sfLedgerEntryType);
+                if (nodeType != ltRIPPLE_STATE)
+                    continue;
+
+                // if newly created RippleState
+                if (node.isFieldPresent(sfNewFields) &&
+                    !node.isFieldPresent(sfFinalFields) &&
+                    !node.isFieldPresent(sfPreviousFields))
+                {
+                    auto const& nfBase = node.peekAtField(sfNewFields);
+                    auto const& newFields =
+                        nfBase.template downcast<STObject>();
+
+                    if (newFields.getFieldAmount(sfLowLimit).getIssuer() !=
+                            accA ||
+                        newFields.getFieldAmount(sfHighLimit).getIssuer() !=
+                            accB)
+                        continue;
+
+                    STAmount balance = newFields.getFieldAmount(sfBalance);
+                    balance.setIssue(issue);
+
+                    if (balance.negative())
+                        balance.negate();
+
+                    return balance;
+                }
+
+                // if RippleState not modified then keep looking
+                if (!node.isFieldPresent(sfFinalFields) ||
+                    !node.isFieldPresent(sfPreviousFields))
+                    continue;
+
+                auto const& ffBase = node.peekAtField(sfFinalFields);
+                auto const& finalFields = ffBase.template downcast<STObject>();
+                auto const& pfBase = node.peekAtField(sfPreviousFields);
+                auto const& previousFields =
+                    pfBase.template downcast<STObject>();
+
+                if (finalFields.getFieldAmount(sfLowLimit).getIssuer() !=
+                        accA ||
+                    finalFields.getFieldAmount(sfHighLimit).getIssuer() != accB)
+                    continue;
+
+                // execution to here means we are on the correct trustline
+                try
+                {
+                    // compute and return the balance mutation on this trustline
+                    STAmount difference =
+                        finalFields.getFieldAmount(sfBalance) -
+                        previousFields.getFieldAmount(sfBalance);
+
+                    difference.setIssue(issue);
+
+                    if (difference.negative())
+                        difference.negate();
+
+                    return difference;
+                }
+                catch (std::runtime_error& e)
+                {
+                    // overflow computing difference so
+                    // err on the side of caution and
+                    // do not return a delivered amount
+                    return {};
+                }
+            }
         }
     }
 
--- a/src/test/rpc/DeliveredAmount_test.cpp
+++ b/src/test/rpc/DeliveredAmount_test.cpp
@@ -17,9 +17,13 @@
 */
 //==============================================================================
 
+#include <ripple/basics/chrono.h>
 #include <ripple/beast/unit_test.h>
 #include <ripple/protocol/Feature.h>
 #include <ripple/protocol/jss.h>
+#include <ripple/resource/Fees.h>
+#include <ripple/rpc/Context.h>
+#include <ripple/rpc/DeliveredAmount.h>
 #include <test/jtx.h>
 #include <test/jtx/WSClient.h>
 
@@ -266,6 +270,7 @@ class DeliveredAmount_test : public beast::unit_test::suite
             BEAST_EXPECT(checkDeliveredAmount.checkExpectedCounters());
         }
     }
+
     void
     testTxDeliveredAmountRPC()
     {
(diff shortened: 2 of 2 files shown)


intelliot:
note: needs (brief) writeup for release notes and documentation

RichardAH:
Please stand by: Further logic needed to handle `CheckCash` and we are still looking into some strange corner cases with order crossing / pathing. Will push further changes to this branch.

RichardAH:
I've added checkcash support. I *think* comparing the RippleState before and after is the correct approach. It might be possible the amount is delivered somehow into an active offer instead of a RippleState object. If someone has an example / more info on this please share.

scottschurr:
Thanks for the updates @RichardAH.  Could you leave a note here when you think you've covered all the corner case you intend to explore?  Thanks.

RichardAH:
> Thanks for the updates @RichardAH. Could you leave a note here when you think you've covered all the corner case you intend to explore? Thanks.

I think I'm happy with it now, but it needs more eyes on it than just mine :)

scottschurr (commented):
Changes look good, but there are parts of the change that I'd feel more comfortable with if they were better exercised by the unit tests.  Code coverage is showing a number of gaps.

I'd be great if someone else wants to code those unit tests.  I don't personally have time to do that right now.  I'm behind schedule on a different (large) code review.  I could potentially write such a unit test in January 2023, if folks can wait that long.

src/ripple/rpc/impl/DeliveredAmount.cpp
```diff
@@ -74,7 +74,94 @@ getDeliveredAmount(
...
+
+            // place the accounts into the canonical ripple state order
+            auto const& accA = issue.account < txAcc ? issue.account : txAcc;
+            auto const& accB = issue.account < txAcc ? txAcc : issue.account;
+
+            // search the Affected nodes in the metadata to find the receiving
+            // trustline
+            for (auto const& node : meta.getFieldArray(sfAffectedNodes))
+            {
+                SField const& metaType = node.getFName();
```
scottschurr, line 122: The `metaType` field is unused.  This line should be removed.

gregtatcam (approved):
LGTM

intelliot:
per @scottschurr , this needs better unit tests

scottschurr:
I'm at a place where I can work on unit tests for this change.  If someone has already picked up that gauntlet please say so now.  There's no point in duplicating effort.

src/ripple/rpc/impl/DeliveredAmount.cpp
```diff
@@ -74,7 +74,93 @@ getDeliveredAmount(
...
+            auto const meta = transactionMeta.getAsObject();
+
+            // if there are no modified fields (somehow) then nothing was
+            // delivered
+            if (!meta.isFieldPresent(sfAffectedNodes))
+                return {};
+
+            bool const hasDeliverMin =
+                serializedTx->isFieldPresent(sfDeliverMin);
+            bool const hasAmount = serializedTx->isFieldPresent(sfAmount);
```
scottschurr, line 86: There's something funny with the logic here.  We know that `hasAmount` is true because we wouldn't be inside the `if` that opens on line 62 if `Amount` were absent.  So `hasAmount` will always be `true` and the `return` on line 91 can never fire.

Since a `CheckCash` transaction might not have an `Amount` field, I suspect that the check on line 62 is not quite right.  Possibly line 62 should be...
```
if (serializedTx->isFieldPresent(sfAmount) || serializedTx->isFieldPresent(sfDeliverMin)
```

scottschurr:
If it's important to address this issue, then the problem is more widely spread than is being addressed by this code change.

A successful `CheckCash` transaction _always_ leaves a `DeliveredAmount` field in its metadata.  And that `DeliveredAmount` does not take the actual change in the trust line into account (as one might expect since it was coded prior to the change being proposed here).

What this means is that the "completely wrong" `DeliveredAmount` may be permanently captured in the metadata of any `CheckCash` transaction that is moving an IOU (not XRP).

Since those fields are part of history, there's no going back and changing them.

If you want to change the `DeliveredAmount` metadata of future transactions then I think we need to change the way `flow` calculates `path::RippleCalc::Output::actualAmountOut`.  And that change would need to be on an amendment.

In the mean time, this pull request affects `delivered_amount`, not `DeliveredAmount`.  If we change the `delivered_amount` for a `CheckCash` transaction to reflect this new approach, then the `delivered_amount` and `DeliveredAmount` for the same transaction could report different values.  Do we want that?  I could be convinced either way.

In the interests of too much information, here's a quick unit test that demonstrates the issue with a `CheckCash`:
```
        using namespace test::jtx;

        Account const brynn("brynn");
        Account const gw("gateway");
        auto const USD = gw["USD"];

        Env env{*this};
        env.fund(XRP(10000), brynn, gw);
        env.close();

        // trust USD(2e90)
        env.trust(STAmount(USD.issue(), 2, 90), brynn);
        env.close();

        // initial balance USD(1e90)
        env(pay(gw, brynn, STAmount(USD.issue(), 1, 90)));
        env.close();

        // Cash a check so small relative to the current trust line that
        // no funds actually move.
        auto const checkId = keylet::check(gw, env.seq(gw)).key;
        env(check::create(gw, brynn, USD(1)));
        env.close();

        env(check::cash(brynn, checkId, check::DeliverMin(USD(1))));
        std::shared_ptr<STObject const> meta = env.meta();
        env.close();

        if (BEAST_EXPECT(meta))
        {
            std::cout << "Cash Check Meta: "
                      << meta->getJson(JsonOptions::none).toStyledString()
                      << std::endl;
        }
```
If you run this code note two things:

1. The metadata contains a `DeliveredAmount` of `USD(1)`, however
2. There are no modified `RippleState` nodes.  That means no funds were _actually_ delivered.

intelliot:
Given `DeliveredAmount` could be incorrect, could we ensure that `delivered_amount` is always included and is correct, and recommend to API consumers that they only use `delivered_amount`?

In a separate PR, we could hide `DeliveredAmount` and bump the API version. Users can always use the older API version if they really need the raw metadata, but actual use cases for that data are slim.

scottschurr:
@intelliot, we have to be careful about what we mean by "incorrect".

Let's say you have two 64-bit floats with values like

- 1e-200 and
- 1e200

then add them together.  The resulting value is _exactly_ the same as 1e200.  Adding a non-zero value did not result in a change from the larger value.  Many people are initially surprised when they learn this.  People who have worked with floating point for a while are occasionally caught off guard when this happens, but they get why it happens.

The `STAmount` class, which rippled uses to represent IOU values internally, is a floating point representation.  It does not follow IEEE conventions, but it is still floating point.  `STAmount` has exactly the problem I just described.  But when it exhibits that problem it is behaving as it was designed.  If we use XRPL IOUs then we _must_ accept this behavior, which has been part of the XRP Ledger from day one.

So when this pull request refers to "incorrect amount" what it means is that, for example, Alice and Bob have a trust line that holds 1e90 USD.  Then Bob sends Alice 1 USD on that trust line.  Even though the payment is successful, the value on the trust line does not change.  In effect, the 1 USD payment is below the resolution that particular trust line can handle.  For this payment the `DeliveredAmount` would show 1 USD.  But, in fact, no funds actually moved according to the ledger.  That's what "incorrect amount" means in the context of this pull request.

If we think this kind of behavior is truly a problem, then I think we should pause and think hard about all of the corners where it shows up and make a plan for how to address _all_ of the corners we identify.

Patching corners here and there will only lead to a messy and inconsistent bunch of special cases.  This pull request would be the first of several patches.  Maybe the second patch is the one you suggest above.  Possibly the third patch would address the "problem" with `CheckCash`'s `DeliveredAmount` I describe above.  Possibly a fourth patch would be to return a `tec` error on payments where the trust line did not change.

Please, let's do our best to determine the _set_ of desirable changes to address this behavior of the ledger.  Then let's implement that set of changes.  Patching our way along haphazardly is unlikely to lead to happiness for XRP Ledger users.

scottschurr (changes requested):
The code seems to have the desired effect with `Payment` transactions, but none of the changes aimed at `CheckCash` were working.  I reworked the code to make it work with `CheckCash`, and you can find those changes in the topmost commit here: https://github.com/scottschurr/rippled/commits/RichardAH-delivered_amount

That commit also includes a new unit test that exercises some corners of this code.  You are welcome to cherry-pick that entire commit if you wish.  At a minimum I think this pull request should include the new unit tests.

On a more general note, I suspect that this pull request is just the leading edge of several changes that will be desired along these lines.  I'd like to encourage us to make a plan of action identifying all of the desired changes and making sure those changes work in a cohesive manner.  Patching the changes individually would probably lead to worse results.

intelliot:
Awaiting reply from @RichardAH

mDuo13:
This sounds very closely related to #4120. Regardless of the decision(s) in this PR, the discussion and edge cases covered here are probably relevant to the larger-scale discussion happening there.

RichardAH:
Cherry-picked scott's commit (thanks!)

scottschurr (approved):
:+1: LGTM.

intelliot:
@gregtatcam given that this PR has changed since your earlier approval, could you give this a re-review?

intelliot:
> On a more general note, I suspect that this pull request is just the leading edge of several changes that will be desired along these lines. I'd like to encourage us to make a plan of action identifying all of the desired changes and making sure those changes work in a cohesive manner. Patching the changes individually would probably lead to worse results.

I agree that "this pull request is just the leading edge of several changes that will be desired along these lines", but I don't think it is necessarily worthwhile to preemptively create a plan of action identifying all of the desired changes and make sure those changes work in a cohesive manner. While I would be happy to have such a plan, I do not have the ability to create it; nor is it clear that its implementation would generate substantially more value, relative to the time and effort required to create and implement it. In other words, perfect is the enemy of good.

@scottschurr this PR has made a lot of progress, and it looks like you have given it a good amount of consideration. If you still believe that, without a comprehensive plan of action, merging this PR "would probably lead to worse results", then please let us know in a comment here. On the other hand, if you've been convinced that this PR is now an improvement worth including, then I intend to merge it if/when @gregtatcam approves.

intelliot:
@RichardAH - commit message suggestion:
```
Fix edge case with delivered_amount for tiny payments: (#4359)

Previously, delivered_amount could be misleading for certain extremely
small payments made to a recipient with a very large IOU balance.

With this change, the delivered_amount is computed from the `Previous`
and `Final` fields of the relevant Affected nodes (except for partial
payments).

* If delivered_amount cannot be calculated, then
  `"delivered_amount": "unavailable"` is returned.
* This change affects `delivered_amount`, not `DeliveredAmount`.
  * Fields permanently captured in the ledger are in PascalCase, while
    computed (derived) fields are in snake_case.
* Includes support for CheckCash.
  * With `delivered_amount` for a CheckCash transaction changed to
    reflect this new approach, the `delivered_amount` and
    `DeliveredAmount` for the same transaction could report different
    values.
* Includes unit tests.

Context:

When a Payment Transaction for an IOU is executed and the receiving
balance is much larger than the amount sent, then (due to floating point
mechanics) no changes occur to the RippleState objects. Despite this,
the `delivered_amount` field reported that the full amount was
delivered, when in fact no balance changes occurred.

This is unlikely to be an issue for fiat currencies because the edge
case only occurs when there is an extreme discrepancy between the size
of the payment and the size of the balance involved.

---------

Co-authored-by: Scott Schurr <scott@ripple.com>
```

Please approve, or provide an alternative.

src/test/rpc/DeliveredAmount_test.cpp
```diff
@@ -325,12 +330,181 @@ class DeliveredAmount_test : public beast::unit_test::suite
...
+            env(offer(carol, USD(1), BTC(1)));
+            env.close();
+
+            env(pay(alice, brynn, BTC(1)), path(~BTC), sendmax(USD(1)));
+            std::shared_ptr<STTx const> tx = env.tx();
+            std::shared_ptr<STObject const> meta = env.meta();
+            env.close();
+
+            // We can show that carol actually sent BTC(1) to brynn, even
+            // though tbrynn's balance did not change, because carol's
```
gregtatcam, line 483: Nit. typo `tbrynn`.

gregtatcam (approved):
👍 LGTM

scottschurr:
@intelliot, regarding...
> I agree that "this pull request is just the leading edge of several changes that will be desired along these lines", but I don't think it is necessarily worthwhile to preemptively create a plan of action identifying all of the desired changes and make sure those changes work in a cohesive manner. While I would be happy to have such a plan, I do not have the ability to create it; nor is it clear that its implementation would generate substantially more value, relative to the time and effort required to create and implement it. In other words, perfect is the enemy of good.

Yup.  I have to agree.  A comprehensive plan would be better.  But lacking that comprehensive plan, this still seems like a good step forward.  I say we merge it.  Thanks for asking.

The suggested commit message looks good to me.  Thanks for writing it up.

JoelKatz:
This behavior was intentional. If you need the details, you can find them in the metadata. If you don't, it's much clearer to tell the recipient that the payment suceeded than that it did not.

Suppose due to some quirks I wind up owing you less than the granularity of the payment system you and I agreed to use. Telling you that my payments to you failed leave a mess that cannot be cleaned up unless you specifically know how to handle tiny payments. Telling you that may payments succeeded make things "just work" even if you don't have special code to handle the situation.

If you are sending me bills for "$0.00" because you think I owe you a millionth of a penny, and we've agreed to use a payment system that can't send amounts that small, and I try to send you a millionth of a penny, we should tell you the payment _worked_ so that the obligation is cleared.

This is especially true with the actual cases in which this is the most likely to happen. An IOU issuer might have a very huge amount of money in an account. They don't care that it get incremented by 1 part in 10^15. But they do care that my balance gets decremented!

RichardAH:
> To put it another way, if I tried to send you some particular amount and the transaction did exactly what we agreed should happen when I try to send you that particular amount and nothing goes wrong, you should credit me for the amount I tried to send you -- since nothing went wrong. If you insist on accuracy of more than 1 part in 10^15, don't use XRPL issued assets because that's their resolution. The point of the delivered_amount field is to make it easy for the recipient to know how much they agreed to credit the sender. (They can always look at the rest of the metadata for more details if default reasonable behavior is not sufficient for a strange use case.)

The field is called delivered_amount not sent_amount. If nothing was delivered then I find this hard to justify.
Anyway if your position is that this is a feature and not a bug I am happy for this to be closed.

JoelKatz:
I think of it as the amount the the recipient agreed to consider the transaction to have delivered. But if this is an understanding that exists only in my head, there's absolutely not reason other people have to share it.

intelliot:
@oeggert - may be worth looking at whether this edge case is clear from the docs. I know we have some documentation about the rounding behavior, but I'm not sure the implication for delivered_amount is clear.

seelabs:
@RichardAH  I took a quick look at the code and I have a concern: If I have a payment where the last step is an offer, and any of the offers are owned by the recipient, then looking at the difference starting and ending balances on a trust line won't give the correct delivered amount. To be concrete: if I send 1 USD to Bob by spending 2 XRP, and Bob has an existing offer with TP 2XRP/TG 1USD, then the trust line balance doesn't change at all and it looks like this code would report a delivered amount of 0 USD, even though the payment correctly delivered 1 USD.

seelabs (changes requested):
See comment about offers owned by the recipient.

PR #4360, opened 2022-12-02
Update my affiliation

nbougalis:


RichardAH (approved):

wojake (changes requested):
I think it would be more specific to everyone if a contributor with no official affiliation is to be deemed as `(Independent)`

intelliot (approved):
I hear you @wojake, but I think it's OK to leave the affiliation off. (Independent) could be confusing if there is some company or organization that is named `Independent` 😂

PR #4364, opened 2022-12-05
Limit amendment flapping (fixes #4350)

### LINKED ISSUE #4350: Prevent amendment majority "flapping" (Version: 1.9.4)
## Summary
Improve the way amendment majorities are calculated in order to reduce the amount of "flapping" (an amendment repeatedly gaining and losing majority in a short period of time).

## Motivation
An [amendment](https://xrpl.org/amendments.html) can repeatedly gain and lose majority many times in a short period of time. Often, this occurs when an amendment is on the verge of gaining majority, and a validator **not** in favor of the amendment goes offline. The majority-calculating logic  This creates noise and uncertainty about whether the amendment is on track to become activated.

## Solution
A few different potential solutions have been proposed.

Currently, an amendment majority is calculated by dividing the number of Yeas (numerator) by the # of trusted validations received in that round (denominator). With 35 validators, an amendment that has 28 Yeas will not have "majority". But when one of the validators saying "Nay" misses a validation, then the calculation becomes 28/34, and the amendment **gains** majority. When the "Nay" validator resumes issuing validations, the amendment then loses majority. For example, at the time of writing, this is occurring to "CheckCashMakesTrustLine" (98DECF327BF79997AEC178323AD51A830E457BFC6D454DAF3E46E5EC42DC619F).

Instead of using "the # of trusted validations received in that round" as the denominator, the logic could instead use "the # of validators on my UNL".

## Paths Not Taken
There could be a different threshold that must be met before an amendment loses majority (vs gaining majority). For example, amendments could gain majority with "Yeas" from > 80% of validators, and not lose majority until <= 65% of validators are saying "Yea".

a-noni-mousse:
This commit implements a new amendment that follows one of the recommendations in xrplf/rippled#4350 to reduce the flapping of amendments when there is validator outage and support drops below 80% required majority.

The new threshold for deactivation of the countdown timer of an an amendment is to go below 65% support.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### DIFF
--- a/src/ripple/app/misc/impl/AmendmentTable.cpp
+++ b/src/ripple/app/misc/impl/AmendmentTable.cpp
@@ -98,11 +98,17 @@ class AmendmentSet
 private:
     // How many yes votes each amendment received
     hash_map<uint256, int> votes_;
+
     Rules const& rules_;
+
     // number of trusted validations
     int trustedValidations_ = 0;
-    // number of votes needed
-    int threshold_ = 0;
+
+    // minimum number of votes needed to begin activation countdown
+    int activationThreshold_ = 0;
+
+    // minimum number of votes needed to continue activate countdown
+    int deactivationThreshold_ = 0;
 
 public:
     AmendmentSet(
@@ -128,23 +134,25 @@ public:
             }
         }
 
-        threshold_ = !rules_.enabled(fixAmendmentMajorityCalc)
+        activationThreshold_ = !rules_.enabled(fixAmendmentMajorityCalc)
             ? std::max(
-                  1L,
-                  static_cast<long>(
+                  1,
+                  static_cast<int>(
                       (trustedValidations_ *
                        preFixAmendmentMajorityCalcThreshold.num) /
                       preFixAmendmentMajorityCalcThreshold.den))
             : std::max(
-                  1L,
-                  static_cast<long>(
+                  1,
+                  static_cast<int>(
                       (trustedValidations_ *
                        postFixAmendmentMajorityCalcThreshold.num) /
                       postFixAmendmentMajorityCalcThreshold.den));
+
+        deactivationThreshold_ = activationThreshold_;
     }
 
     bool
-    passes(uint256 const& amendment) const
+    passes(uint256 const& amendment, int threshold) const
     {
         auto const& it = votes_.find(amendment);
 
@@ -158,9 +166,9 @@ public:
         // to gain majority.
         if (!rules_.enabled(fixAmendmentMajorityCalc) ||
             trustedValidations_ == 1)
-            return it->second >= threshold_;
+            return it->second >= threshold;
 
-        return it->second > threshold_;
+        return it->second > threshold;
     }
 
     int
@@ -181,9 +189,15 @@ public:
     }
 
     int
-    threshold() const
+    activationThreshold() const
     {
-        return threshold_;
+        return activationThreshold_;
+    }
+
+    int
+    deactivationThreshold() const
+    {
+        return deactivationThreshold_;
     }
 };
 
@@ -619,8 +633,9 @@ AmendmentTableImpl::doVoting(
     auto vote = std::make_unique<AmendmentSet>(rules, valSet);
 
     JLOG(j_.debug()) << "Received " << vote->trustedValidations()
-                     << " trusted validations, threshold is: "
-                     << vote->threshold();
+                     << " trusted validations. Thresholds are: "
+                     << vote->activationThreshold() << " and "
+                     << vote->deactivationThreshold();
 
     // Map of amendments to the action to be taken for each one. The action is
     // the value of the flags in the pseudo-transaction
@@ -633,7 +648,8 @@ AmendmentTableImpl::doVoting(
     {
         NetClock::time_point majorityTime = {};
 
-        bool const hasValMajority = vote->passes(entry.first);
+        bool const hasValMajority =
+            vote->passes(entry.first, vote->activationThreshold());
 
         {
             auto const it = majorityAmendments.find(entry.first);
@@ -653,7 +669,9 @@ AmendmentTableImpl::doVoting(
             JLOG(j_.debug()) << entry.first << ": amendment got majority";
             actions[entry.first] = tfGotMajority;
         }
-        else if (!hasValMajority && (majorityTime != NetClock::time_point{}))
+        else if (
+            !hasValMajority && (majorityTime != NetClock::time_point{}) &&
+            !vote->passes(entry.first, vote->deactivationThreshold()))
         {
             // Ledger says majority, validators say no
             JLOG(j_.debug()) << entry.first << ": amendment lost majority";
@@ -740,7 +758,7 @@ AmendmentTableImpl::injectJson(
     if (!fs.enabled && lastVote_)
     {
         auto const votesTotal = lastVote_->trustedValidations();
-        auto const votesNeeded = lastVote_->threshold();
+        auto const votesNeeded = lastVote_->activationThreshold();
         auto const votesFor = lastVote_->votes(id);
 
         v[jss::count] = votesFor;
--- a/src/ripple/app/tx/impl/Change.cpp
+++ b/src/ripple/app/tx/impl/Change.cpp
@@ -207,7 +207,7 @@ Change::activateTrustLinesToSelfFix()
 TER
 Change::applyAmendment()
 {
-    uint256 amendment(ctx_.tx.getFieldH256(sfAmendment));
+    uint256 const amendment(ctx_.tx.getFieldH256(sfAmendment));
 
     auto const k = keylet::amendments();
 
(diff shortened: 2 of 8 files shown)


nbougalis:
Great, happy to assign this to you and delegate myself to be a reviewer.

src/ripple/app/misc/impl/AmendmentTable.cpp
```diff
@@ -619,8 +640,9 @@ AmendmentTableImpl::doVoting(
     auto vote = std::make_unique<AmendmentSet>(rules, valSet);
 
     JLOG(j_.debug()) << "Received " << vote->trustedValidations()
-                     << " trusted validations, threshold is: "
-                     << vote->threshold();
+                     << " trusted validations. Tresholds are: "
```
nbougalis, line 643: Typo: missing `h`, should be `Thresholds`

src/ripple/app/misc/impl/AmendmentTable.cpp
```diff
@@ -653,7 +676,9 @@ AmendmentTableImpl::doVoting(
             JLOG(j_.debug()) << entry.first << ": amendment got majority";
             actions[entry.first] = tfGotMajority;
         }
-        else if (!hasValMajority && (majorityTime != NetClock::time_point{}))
+        else if (
+            !hasValMajority && (majorityTime != NetClock::time_point{}) &&
+            !vote->passes(entry.first, vote->deactivationThreshold()))
```
nbougalis, line 681: I _think_ this is right and should have the same behavior as the existing code, but someone else should triple-check the logic when the amendment being checked is unknown. Here's the logic: `hasValMajority` will be `false`; before this commit, we'd always enter this block. Now `vote->passes` will return `false`, and the `!` will turn it to `true` putting us in this block, just like before.

src/ripple/protocol/Feature.h
```diff
@@ -74,7 +74,7 @@ namespace detail {
 // Feature.cpp. Because it's only used to reserve storage, and determine how
 // large to make the FeatureBitset, it MAY be larger. It MUST NOT be less than
 // the actual number of amendments. A LogicError on startup will verify this.
-static constexpr std::size_t numFeatures = 53;
+static constexpr std::size_t numFeatures = 54;
```
nbougalis, line 77: We should consider ways to remove this altogether. It's a bit of a workaround.

src/test/app/AmendmentTable_test.cpp
```diff
@@ -522,9 +523,31 @@ class AmendmentTable_test final : public beast::unit_test::suite
...
+    //            {feat},
+    //            table,
+    //            week,
+    //            validators,
+    //            votes,
+    //            ourVotes,
+    //            enabled,
+    //            majority);
+    //    }
+
```
nbougalis, line 547: Should this commented-out block just be removed?

src/test/app/AmendmentTable_test.cpp
```diff
@@ -856,8 +882,10 @@ class AmendmentTable_test final : public beast::unit_test::suite
                 enabled,
                 majority);
 
-            if (i < 4)  // 16 - 3 = 13 => 13/16 = 0.8125 => > 80%
-            {           // 16 - 4 = 12 => 12/16 = 0.75 => < 80%
+            int lim = features.contains(fixAmendmentFlapping) ? 6 : 4;
+
+            if (i < lim)  // 16 - 3 = 13 => 13/16 = 0.8125 => > 80%
+            {             // 16 - 4 = 12 => 12/16 = 0.75 => < 80%
```
nbougalis, line 888: These comments likely need to be updated.

nbougalis (changes requested):
I like the elegant and minimal solution. I think will work just fine to prevent casual flapping in case a validator goes offline at just the right moment.

It does not add any hysteresis and that may be something else to consider adding _on top_ of this, but then again maybe this is sufficient on its own for all but the most extreme flapping cases and there's something to be said about keeping it simple.

nbougalis (approved):
LGTM.

intelliot:
@a-noni-mousse can you resolve the merge conflicts? thanks!

thejohnfreeman:
I guess the downside of this approach is that if exactly one validator permanently withdraws support for an amendment that just barely has supermajority support, then it will not actually lose supermajority support until enough others withdraw their support too. I prefer the idea of tracking votes on-ledger, but I guess the community finds the risk of this approach acceptable.

thejohnfreeman (approved):

src/ripple/protocol/Feature.h
thejohnfreeman: Does this need to be updated to 55? Was another amendment added while this was being reviewed? CI fails on this number being wrong.

thejohnfreeman (changes requested):
Might need to update the feature count again: https://github.com/XRPLF/rippled/pull/4364#discussion_r1060825182

HowardHinnant:
```
$ rippled --unittests
Logic error: More features defined than allocated. Adjust numFeatures in Feature.h.
```

nbougalis:
> Might need to update the feature count again: [#4364 (comment)](https://github.com/XRPLF/rippled/pull/4364#discussion_r1060825182)

OK but that is bad designing in my thinking when the count is easy and always correct for compiler to calculate instead of programmers.

nbougalis:
My bad, with the close-reopen. Apologies.

mDuo13 (changes requested):
I am against this approach.

It puts more demand on the validators to be stably online _all the time_. If you're one of the swing votes against an amendment, it could gain the majority while you're having a brief outage, and sustain the lower threshold even after you come back, which is not how the system is intended to work.

This changes the threshold for activating an amendment to "65% and some luck" rather than "sustained 80%". If you were to describe this as simply lowering the threshold to 65%, I think people would rightly be against that; but this is even worse because it's "kind of 65%" but more complicated.

The correct fix for amendment flapping would be to change the activation threshold calculation so that it requires >80% of _all validators on the node's UNL_ (ignoring Negative UNL) rather than >80% of validators currently participating in consensus. Then amendments would only "flap" when validators change their votes, which is the intended behavior.

thejohnfreeman:
I'm convinced by @mDuo13 too, but I will comment that using 80% of UNL vs seen validations will not fix the issue either. You'll flap on when you see a yea from 80% the UNL and off when one of those yeas misses a validation.

The only way to tell the difference between "validator A is no longer voting yea" and "validator A appears to be offline" is if you see a validation from A with no yea. We're not observing that distinction in the code. If we start to carefully make that distinction, and a yea-voting validator is temporarily offline, should we pretend it is still voting yea? For how long?

scottschurr:
@thejohnfreeman, excellent point and question.

My personal take, without trying any implementation, is that a validator should be assumed to be holding the same vote as it held before until a new vote actually arrives from that validator.  So if the validator is offline it is assumed to not change its vote.  However, that may be hard to implement, I'm not sure.  I'm hesitant to armchair quarterback an approach that has the right balance of voting stability vs additional complexity.

nbougalis:
@mDuo13, @thejohnfreeman: excellent points. I had not thought of that when reviewing the code. I'm thinking that this isn't ready for "prime time" as is.

nbougalis:
You can't track individual votes on ledger; _whose_ votes would you track? You can't say "the UNL's" because there isn't _one_ UNL or a requirement that my UNL completely overlaps with yours, and you'd need to reach consensus on which ones to include (and you can't include all of them!). It's for the same reason that you can't just track a "count" of yays or nays, or use the "size of the UNL" instead of the number of validations received.

src/ripple/app/misc/impl/AmendmentTable.cpp
```diff
@@ -98,11 +98,17 @@ class AmendmentSet
...
+
     // number of trusted validations
     int trustedValidations_ = 0;
-    // number of votes needed
-    int threshold_ = 0;
+
+    // minimum number of votes needed to begin activation countdown
+    int activationThreshold_ = 0;
+
+    // minimum number of votes needed to continue activate countdown
```
thejohnfreeman, line 110:
```suggestion
    // minimum number of votes needed to continue activation countdown
```

src/ripple/app/misc/impl/AmendmentTable.cpp
thejohnfreeman: If the amendment is unknown, then how would `majorityTime` be non-zero? And if it is zero for unknown amendments, then how would this branch have been entered before the change?

src/ripple/app/tx/impl/Change.cpp
```diff
@@ -233,41 +233,67 @@ Change::applyAmendment()
...
+    if (it != majorities.end())
     {
-        const STArray& oldMajorities =
-            amendmentObject->getFieldArray(sfMajorities);
-        for (auto const& majority : oldMajorities)
+        if (gotMajority)
+            return tefALREADY;
+
+        majorities.erase(it);
+    }
```
thejohnfreeman, line 252: The first time an amendment loses majority, the program will reach line 251, remove the amendment from the `sfMajorities` array, and continue, presumably to add it to the `sfFlappingAmendments` array. The second time an amendment loses majority, the program will reach line 244 and return, never continuing to fix the `sfFlappingAmendments` array. Line 251 needs to be conditional on whether the amendment is in the `sfFlappingAmendments` array (which itself is conditional on the amendment).

There does not seem to be a new test for this new behavior. Have you added one?

thejohnfreeman (changes requested):

scottschurr:
@a-noni-mousse, I glanced over this pull request and, although it follows the suggestion by @nbougalis, I'm not completely sold on the approach.  I'm investigating whether what @mDuo13 suggested is feasible: https://github.com/XRPLF/rippled/pull/4364#pullrequestreview-1236628802  If that approach is feasible, then I think that would be the better path.

So I'm not ignoring you.  Your pull request has prompted further investigation.  Thanks for your patience.

nbougalis:
> @a-noni-mousse, I glanced over this pull request and, although it follows the suggestion by @nbougalis, I'm not completely sold on the approach. I'm investigating whether what @mDuo13 suggested is feasible: [#4364 (review)](https://github.com/XRPLF/rippled/pull/4364#pullrequestreview-1236628802) If that approach is feasible, then I think that would be the better path.

I'm not sure I understand how what @mDuo13 is suggesting would work. He writes:

>The correct fix for amendment flapping would be to change the activation threshold calculation so that it requires >80% of all validators on the node's UNL (ignoring Negative UNL) rather than >80% of validators currently participating in consensus. _Then amendments would only "flap" when validators change their votes, which is the intended behavior._

How will this prevent flapping or does it constitute a "correct fix"? Remember, votes aren't persistent which means that an amendment can flap on and off if as few as *1* validator were to miss a *single* vote.

The problem just isn't the negative UNL; flapping was a problem _before_ that. The problem isn't even that the threshold is calculated based on the number of trusted validators that participated in the vote; if the threshold was calculated against the size of a node's UNL, you'd actually have _*more*_ flapping. See [here](https://github.com/XRPLF/rippled/blob/develop/src/ripple/app/misc/impl/AmendmentTable.cpp#L114-L143).

The problem is that there's a very narrow window of voting and a validator needs to be active during that window for it to cast a "yes" vote.

Now, this can be solved in one of several ways:

1. Make validator votes "sticky." That is, if a validator voted "yes" once, then that "yes" persists until the validator votes "no". That has several problems, some of which I've explained before. At a minimum, the votes need to be persisted. You can't persist them on the ledger itself, which means you need to persist them locally. But if you do that, you need a mechanism to check if the vote you've persisted is stale, and if you don't, you're calculating the threshold incorrectly.
2. Add hysteresis, which the latest commit from OP does. This doesn't _entirely_ prevent flapping, but it effectively 'extends' a grace period by "persisting" the majority (as opposed to persisting the vote) until the next ballot.
3. Allow a validator to vote at _any_ time prior to the tally. This is interesting but still won't _entirely_ prevent flapping either.

thejohnfreeman:
Here's an elaboration of my [previous comment](https://github.com/XRPLF/rippled/pull/4364#issuecomment-1372308239) that I wrote in a conversation with @scottschurr:

If we switch to using the full UNL, and ignore the negative UNL (@mDuo13's suggestion), then the problem becomes that a missing validation is interpreted as a "no" vote. Both "yes" and "no" votes are possibilities for the intent of the missing validation, which means whatever assumption we make in its absence carries some risk of being wrong. A corollary is that any "fix" should focus on flapping less, because the goal of reflecting validators' true intentions is impossible to guarantee.

The risk of being wrong seems to be smaller if we assume that the vote is unchanged from the validator's previous vote, which means one fix is to forward fill votes. That requires keeping track of old votes, which I don't think we do currently. (It _could_ be done on-chain if each validator was associated with an account and we introduced the capability of recording votes, but I'm not suggesting that we do that.) These are the "sticky" votes @nbougalis is talking about. We update these votes with every pre-flag validation to keep them fresh.

Another fix, the hysteresis @nbougalis suggested, will work to reduce flapping. It effectively forward fills "yes" votes (but not "no" votes) for one flag ledger and no more. It does not stop the flapping if the cause is a flaky validator voting "no"[^1]. It becomes more likely to diverge from the validators' true intentions, though, if the outage lasts for more than one flag ledger.

[^1]: For example, assume you have 27 of 34 validators voting "yes", and one of the "no" voters disappears. 27 of 33 validators reaches majority. Then the "no" voter returns, and majority is lost two flag ledgers later. Flapping was not stopped in this situation.

scottschurr:
I've investigated an approach to amendment flapping that I think would address the comment from @mDuo13: https://github.com/XRPLF/rippled/pull/4364#pullrequestreview-1236628802

The new approach seems viable, so I've created a pull request for that approach: https://github.com/XRPLF/rippled/pull/4410

In cases like this I usually try to make a change that could be cherry-picked into the existing pull request.  I didn't do so this time because there's very little in common between the two approaches.  The only thing shared by the two pull requests is the name of the `Feature`.  We'll let reviewers choose which approach they prefer.

Thanks for your efforts on this pull request, @a-noni-mousse.  Even though I've submitted an alternative approach I very much appreciate your efforts here.

JoelKatz:
We could treat absent/unknown votes as votes not to change the status quo.

intelliot:
Closing in favor of https://github.com/XRPLF/rippled/pull/4410

PR #4367, opened 2022-12-14
Use the Conan package manager

thejohnfreeman:
This is a signed, squashed commit capturing all of the changes in #4223.


intelliot (approved):

intelliot:
Even though this is the same as #4223, let's get a quick look from one of the other reviewers before we merge this in.

intelliot:
@mDuo13 are there any docs on XRPL.org that should be updated?

ximinez (approved):
Looking forward to using this full time! Adios `.nih_c` folder!

PR #4370, opened 2022-12-19
Add NetworkID field to transactions to help prevent replay attacks on and from side-chains

RichardAH:
Help to prevent replay attacks by allowing side-chains to require their users specify an accurate `NetworkID` field in every transaction to that chain.

This change introduces a new field:
- `CONSTRUCT_TYPED_SFIELD(sfNetworkID,             "NetworkID",            UINT32,     1);`

And three new local error codes:
- `telNETWORK_ID_MAKES_TX_NON_CANONICAL`
- `telREQUIRES_NETWORK_ID`
- `telWRONG_NETWORK`.

To preserve legacy behaviour (and avoid the need for an amendment) all chains with a network ID less than 1025 retain the existing behaviour. This includes main, testnet, devnet and hooks-testnet. If `sfNetworkID` is present in any txn submitted to any of the nodes on this chain then a `telNETWORK_ID_MAKES_TX_NON_CANONICAL` is returned. 

All other chains now *require* their users to include an appropriate `sfNetworkID` in every transaction to that chain.

Local error codes were chosen because a transaction is not necessarily malformed if it is submitted to a node running on the incorrect chain. This is a local error specific to that node and could be corrected by switching to a different node or by changing the `network_id` on that node.

## Test Plan

A new unit test suite has been added:
```
# ./rippled -u ripple.app.NetworkID
ripple.app.NetworkID Require txn NetworkID to be specified (or not) depending on the network ID of the node
501ms, 1 suite, 1 case, 23 tests total, 0 failures
```

Config unit test was also updated to include some network_id tests:
```
# ./rippled -u ripple.core.Config
ripple.core.Config legacy
ripple.core.Config database_path
Loading: "testSetup12/rippled.cfg"
Loading: "testSetup12/rippled.cfg"
Loading: "testSetup13/rippled.cfg"
Loading: "testSetup13/rippled.cfg"
ripple.core.Config validator keys
ripple.core.Config validators_file
ripple.core.Config amendment
ripple.core.Config overlay: unknown time
ripple.core.Config overlay: diverged time
ripple.core.Config network id
101ms, 1 suite, 8 cases, 260 tests total, 0 failure
```

All tests:
```
Longest suite times:
   42.0s ripple.tx.Offer
   35.0s ripple.app.ValidatorSite
   21.7s ripple.rpc.NodeToShardRPC
   20.6s ripple.tx.NFToken
   17.3s ripple.tx.NFTokenBurn
   13.9s ripple.app.Flow
   13.7s ripple.app.ShardArchiveHandler
   13.0s ripple.app.TheoreticalQuality
    8.3s ripple.app.LedgerReplayer
    7.9s ripple.tx.NFTokenDir
337.1s, 205 suites, 1529 cases, 532189 tests total, 0 failures
```


intelliot:
This is worth considering; however, I believe the recommendation is to use different accounts/keys on side-chains. The X-address format recommends using the `X` prefix for Mainnet and `T` for any test/development network. Additional prefixes could be defined for well-known sidechains; there could also be a prefix for side-chains in general.

RichardAH:
> This is worth considering; however, I believe the recommendation is to use different accounts/keys on side-chains. The X-address format recommends using the `X` prefix for Mainnet and `T` for any test/development network. Additional prefixes could be defined for well-known sidechains; there could also be a prefix for side-chains in general.

This wouldn't prevent replay attacks, it's just a client side syntactic sugar over the top of existing AccountIDs and destination tags. The resulting serialized transactions are the same and can be replayed in the same way on other chains. 

Realistically people will absolutely use the same keys on multiple chains. This is an exceedingly common practice on Ethereum chains. There are even legitimate use-cases for it... such as swapping from your own account on one chain to your own account on another over a bridge without attracting a lot of unwanted/unnecessary AML legislation.

intelliot:
Yes, you are correct. It is an issue if people use the same keys on multiple chains. It may be common practice, but I'm loathe to recommend it, partly because that does open up the risk of replay attacks.

Anyway, I'm fine with requiring NetworkID on new chains (as you said - Mainnet would remain unchanged). Let's get community feedback and get this reviewed.

RichardAH:
Ultimately the code doesn’t affect mainnet at all (and thus doesn’t require an amendment). It’s here to automatically apply to future side chains (which I assume there will be quite a few of given the resources going into XChain amendment).

If the PR is rejected we should still ensure UINT32 field code 1 is reserved for NetworkID. As long as everyone agrees on that then dealing with multiple chains in the future will be a lot easier.

intelliot:
Agreed - that's why I didn't put the `Amendment` label on this

src/ripple/app/tx/impl/Transactor.cpp
```diff
@@ -40,6 +40,27 @@ namespace ripple {
 NotTEC
 preflight0(PreflightContext const& ctx)
 {
+    uint32_t nodeNID = ctx.app.config().NETWORK_ID;
+    std::optional<uint32_t> txNID = ctx.tx[~sfNetworkID];
+
+    if (nodeNID <= 1024)
+    {
+        // legacy networks have ids less than 1024, these networks cannot
```
mDuo13, line 48:
```suggestion
        // legacy networks have IDs 1024 and below. These networks cannot
```

src/ripple/core/impl/Config.cpp
```diff
@@ -479,6 +479,18 @@ Config::loadFromString(std::string const& fileContents)
...
+    {
+        if (strTemp == "main")
+            NETWORK_ID = 0;
+        else if (strTemp == "testnet")
+            NETWORK_ID = 1;
+        else if (strTemp == "devnet")
+            NETWORK_ID = 2;
+        else
+            NETWORK_ID = beast::lexicalCastThrow<uint32_t>(strTemp);
+    }
```
mDuo13, line 492: I noticed this section is similar to [the `OverlayImpl.cpp` version](https://github.com/XRPLF/rippled/blob/e32bc674aa2a035ea0f05fe43d2f301b203f1827/src/ripple/overlay/impl/OverlayImpl.cpp#L1615-L1628); is that something that can be updated to be consistent or shared in some way so this switch doesn't have to be in multiple places?

mDuo13 (approved):
Good change. Just curious, where did you get the 1024 cutoff? Personally I think I'd rather make it so that every non-Mainnet network requires a NetworkID, but I guess it _would_ be disruptive to reset all the test networks just for this.

RichardAH:
> Good change. Just curious, where did you get the 1024 cutoff? Personally I think I'd rather make it so that every non-Mainnet network requires a NetworkID, but I guess it _would_ be disruptive to reset all the test networks just for this.

It comes from the restriction on port numbers in Linux. We feel that new networks should pick a peering port that is = to their network ID, and this should always be > 1024. :)

src/ripple/core/impl/Config.cpp
RichardAH: I also saw this but didn't want to touch OverlayImpl. Feel free to suggest an additional commit with appropriate tests to remove this redundant code.

src/ripple/app/tx/impl/Transactor.cpp
```diff
@@ -40,6 +40,27 @@ namespace ripple {
 NotTEC
 preflight0(PreflightContext const& ctx)
 {
+    uint32_t nodeNID = ctx.app.config().NETWORK_ID;
+    std::optional<uint32_t> txNID = ctx.tx[~sfNetworkID];
```
oleks-rip, line 44: Line 43 and 44 - may be const ?

src/test/app/NetworkID_test.cpp
```diff
@@ -0,0 +1,153 @@
...
+    ANY  SPECIAL ,  DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+    WHATSOEVER  RESULTING  FROM  LOSS  OF USE, DATA OR PROFITS, WHETHER IN AN
+    ACTION  OF  CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+    OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+//==============================================================================
+
+#include <ripple/basics/BasicConfig.h>
+#include <ripple/core/ConfigSections.h>
+#include <test/jtx.h>
```
oleks-rip, line 22: #include <ripple/protocol/jss.h>
doesn't build on my Linux without this header

SFsourDoh:
should this be "help prevent replay attacks"  since network IDs might not be unique?

src/ripple/protocol/impl/SField.cpp
```diff
@@ -104,6 +104,7 @@ CONSTRUCT_TYPED_SFIELD(sfHookExecutionIndex,    "HookExecutionIndex",   UINT16,
 CONSTRUCT_TYPED_SFIELD(sfHookApiVersion,        "HookApiVersion",       UINT16,    20);
 
 // 32-bit integers (common)
+CONSTRUCT_TYPED_SFIELD(sfNetworkID,             "NetworkID",            UINT32,     1);
```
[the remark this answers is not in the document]
RichardAH: I’ve never seen UINT32 field code 1 used anywhere. Not even in old old ledgers. Maybe someone can shed light on why it was free?
It seems appropriate to use it for an important task such as preventing replay attacks.

nbougalis (approved):

sublimator:
Old habits die hard, hey Nik :)

intelliot:
@RichardAH if this still looks good to merge (from your perspective), go ahead and put the `Passed` label on it; or just comment saying that this is good to merge.

RichardAH:
lgtm

sublimator:
@intelliot 

> I believe the recommendation is to use different accounts/keys on side-chains

One would hope!!

gregtatcam (approved):
LGTM 👍

intelliot:
Suggested commit message:
```
Prevent replay attacks with NetworkID field: (#4370)

Add a `NetworkID` field to help prevent replay attacks on and from
side-chains.

The new field must be used when the server is using a network id > 1024.

To preserve legacy behavior, all chains with a network ID less than 1025
retain the existing behavior. This includes Mainnet, Testnet, Devnet,
and hooks-testnet. If `sfNetworkID` is present in any transaction
submitted to any of the nodes on one of these chains, then
`telNETWORK_ID_MAKES_TX_NON_CANONICAL` is returned.

Since chains with a network ID less than 1025, including Mainnet, retain
the existing behavior, there is no need for an amendment.

The `NetworkID` helps to prevent replay attacks because users specify a
`NetworkID` field in every transaction for that chain.

This change introduces a new UINT32 field, `sfNetworkID` ("NetworkID").
There are also three new local error codes for transaction results:

- `telNETWORK_ID_MAKES_TX_NON_CANONICAL`
- `telREQUIRES_NETWORK_ID`
- `telWRONG_NETWORK`

To learn about the other transaction result codes, see
https://xrpl.org/transaction-results.html.

Local error codes were chosen because a transaction is not necessarily
malformed if it is submitted to a node running on the incorrect chain.
This is a local error specific to that node and could be corrected by
switching to a different node or by changing the `network_id` on that
node. See
https://xrpl.org/connect-your-rippled-to-the-xrp-test-net.html.

In addition to using `NetworkID`, it is still generally recommended to
use different accounts and keys on side-chains. However, people will
undoubtedly use the same keys on multiple chains; for example, this is
common practice on other blockchain networks. There are also some
legitimate use cases for this.

A `app.NetworkID` test suite has been added, and `core.Config` was
updated to include some network_id tests.
```

src/test/core/Config_test.cpp
```diff
@@ -411,6 +411,71 @@ port_wss_admin
...
+)rippleConfig");
+        }
+        catch (std::runtime_error& e)
+        {
+            error = e.what();
+        }
+
+        BEAST_EXPECT(error == "");
+        BEAST_EXPECT(c.NETWORK_ID == 10000);
+    }
```
pwang200, line 477: Please consider adding a negative case, e.g. 
[network_id]
abcd

pwang200 (approved):
I have a couple minor comments, but looks good!

intelliot:
explanation:

@pwang200 approved the PR. The suggestions could be incorporated (by @pwang200) in a commit (to be cherry picked); or if this PR has been merged, they can be suggested in a new PR.

@RichardAH confirmed this is good to go as-is.

will-yjn:
For mainnet/testnet/devnet, do we have to upgrade js sdk or add networkId when we construct the transaction? I assume no?

intelliot:
This change does not affect mainnet/testnet/devnet.

> To preserve legacy behaviour (and avoid the need for an amendment) all chains with a network ID less than 1025 retain the existing behaviour. This includes main, testnet, devnet and hooks-testnet.

The NetworkID field must **not** be set in transactions on these networks.

Only sidechains and test networks with IDs of 1025 or higher need to have the NetworkID set.

PR #4380, opened 2023-01-05
[Amendment] Fix 3 issues around NFToken offer acceptance

ledhed2222:
(This PR will share an amendment with https://github.com/XRPLF/rippled/pull/4346 [merged to feature branch on 2023-02-02]. We want them to go out in the same release.)

Fixes 3 separate issues:

### Issue 1

In the following scenario, an account cannot perform NFTokenAcceptOffer even though it should be allowed to:

- BROKER has < S
- ALICE offers to sell token for S
- BOB offers to buy token for > S
- BROKER tries to bridge the two offers

This currently results in `tecINSUFFICIENT_FUNDS`, but should not because BROKER is not spending any funds in this transaction, beyond the transaction fee.

### Issue 2

Situation - when trading an NFT using IOUs, and when the issuer of the IOU has any non-zero value set for TransferFee on their account via AccountSet (not a TransferFee on the NFT), and when the sale amount is equal to the total balance of that IOU that the buyer has, the resulting balance for the issuer of the IOU will become positive, which essentially means that the buyer of the NFT was supposed to have caused a certain amount of IOU to be burned. That amount was unable to be burned because the buyer couldn't cover it. This results in the buyer owing this amount back to the issuer. In a real world scenario, this is appropriate and can be settled off-chain.

### Issue 3

Currency issuers could not make offers for NFTs using their own currency, receiving `tecINSUFFICIENT_FUNDS` if they tried to do so.

With this fix, they are now able to buy/sell NFTs using their own currency.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### DIFF
--- a/src/ripple/app/tx/impl/NFTokenAcceptOffer.cpp
+++ b/src/ripple/app/tx/impl/NFTokenAcceptOffer.cpp
@@ -276,11 +276,9 @@ NFTokenAcceptOffer::pay(
         return result;
     if (result != tesSUCCESS)
         return result;
-    if (accountFunds(view(), from, amount, fhZERO_IF_FROZEN, j_) <
-        amount.zeroed())
+    if (accountFunds(view(), from, amount, fhZERO_IF_FROZEN, j_).signum() < 0)
         return tecINSUFFICIENT_FUNDS;
-    if (accountFunds(view(), to, amount, fhZERO_IF_FROZEN, j_) <
-        amount.zeroed())
+    if (accountFunds(view(), to, amount, fhZERO_IF_FROZEN, j_).signum() < 0)
         return tecINSUFFICIENT_FUNDS;
     return tesSUCCESS;
 }
--- a/src/ripple/app/tx/impl/NFTokenAcceptOffer.cpp
+++ b/src/ripple/app/tx/impl/NFTokenAcceptOffer.cpp
@@ -27,33 +27,6 @@
 
 namespace ripple {
 
-STAmount
-NFTokenAcceptOffer::getAmountForValidation(
-    AccountID const& account,
-    STAmount const& amount)
-{
-    if (account != amount.getIssuer())
-        return amount;
-    auto ret = STAmount(amount);
-    ret.negate();
-    return ret;
-}
-
-bool
-NFTokenAcceptOffer::balanceOKAfterPayment(
-    AccountID const& account,
-    STAmount const& amount) const
-{
-    auto const comparingAmount = getAmountForValidation(account, amount);
-    return accountHolds(
-               view(),
-               account,
-               comparingAmount.getCurrency(),
-               comparingAmount.getIssuer(),
-               fhZERO_IF_FROZEN,
-               j_) >= amount.zeroed();
-}
-
 NotTEC
 NFTokenAcceptOffer::preflight(PreflightContext const& ctx)
 {
@@ -200,11 +173,16 @@ NFTokenAcceptOffer::preclaim(PreclaimContext const& ctx)
         //
         // After this amendment, we allow an IOU issuer to buy an NFT with their
         // own currency
-        auto const needed = ctx.view.rules().enabled(fixUnburnableNFToken)
-            ? getAmountForValidation(bo->at(sfOwner), bo->at(sfAmount))
-            : bo->at(sfAmount);
-
-        if (accountHolds(
+        auto const needed = bo->at(sfAmount);
+        if (ctx.view.rules().enabled(fixUnburnableNFToken))
+        {
+            if (accountFunds(
+                    ctx.view, (*bo)[sfOwner], needed, fhZERO_IF_FROZEN, ctx.j) <
+                needed)
+                return tecINSUFFICIENT_FUNDS;
+        }
+        else if (
+            accountHolds(
                 ctx.view,
                 (*bo)[sfOwner],
                 needed.getCurrency(),
@@ -238,9 +216,9 @@ NFTokenAcceptOffer::preclaim(PreclaimContext const& ctx)
         }
 
         // The account offering to buy must have funds:
+        auto const needed = so->at(sfAmount);
         if (!ctx.view.rules().enabled(fixUnburnableNFToken))
         {
-            auto const needed = so->at(sfAmount);
             if (accountHolds(
                     ctx.view,
                     ctx.tx[sfAccount],
@@ -264,13 +242,10 @@ NFTokenAcceptOffer::preclaim(PreclaimContext const& ctx)
             // mode, because then we are confirming that the broker can
             // cover what the buyer will pay, which doesn't make sense, causes
             // an unncessary tec, and is also resolved with this amendment.
-            auto const needed =
-                getAmountForValidation(ctx.tx[sfAccount], so->at(sfAmount));
-            if (accountHolds(
+            if (accountFunds(
                     ctx.view,
                     ctx.tx[sfAccount],
-                    needed.getCurrency(),
-                    needed.getIssuer(),
+                    needed,
                     fhZERO_IF_FROZEN,
                     ctx.j) < needed)
                 return tecINSUFFICIENT_FUNDS;
@@ -301,8 +276,11 @@ NFTokenAcceptOffer::pay(
         return result;
     if (result != tesSUCCESS)
         return result;
-    if (!(balanceOKAfterPayment(from, amount) &&
-          balanceOKAfterPayment(to, amount)))
+    if (accountFunds(view(), from, amount, fhZERO_IF_FROZEN, j_) <
+        amount.zeroed())
+        return tecINSUFFICIENT_FUNDS;
+    if (accountFunds(view(), to, amount, fhZERO_IF_FROZEN, j_) <
+        amount.zeroed())
         return tecINSUFFICIENT_FUNDS;
     return tesSUCCESS;
 }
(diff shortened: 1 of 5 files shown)


nbougalis:
I haven't reviewed this closely and don't think I will have time to, but wanted to point out that you're doing unnecessary work because you're using `accountHolds` instead of `accountFunds`. The latter _properly_ handles the case when you're checking if an account has sufficient funds in a given asset available and the account is the issuer of the asset.

ledhed2222:
@nbougalis how come GitHub doesn't have a 🐐 emoji response yet

src/ripple/app/tx/impl/NFTokenAcceptOffer.cpp
```diff
@@ -27,6 +27,33 @@
...
+NFTokenAcceptOffer::getAmountForValidation(
+    AccountID const& account,
+    STAmount const& amount)
+{
+    if (account != amount.getIssuer())
+        return amount;
+    auto ret = STAmount(amount);
+    ret.negate();
+    return ret;
+}
```
scottschurr, line 40: I want to start by saying that the way you have this coded is just fine.  However, if you want to, you can make `getAmountForValidation()`  a non-member function that is local to just this implementation file.  You would do that because it reduces the visibility of the function just a hair.  You can make this change because...

- The method is `private`, so it's not needed outside of this implementation file and
- It does not use any of the member variables (which you had already noted by making it a `static` member).

If you choose make this change you'd:

- Remove the member function declaration from the header file and
- Change the declaration here to look like...

```
static STAmount
getAmountForValidation(
    AccountID const& account,
    STAmount const& amount)
```

The leading `static` is what makes it local to this file.  The linker will not make a file scope `static` function visible outside of the containing translation unit.

Note that, if you do this, then the implementation of `getAmountForValidation()` must appear in the file ahead of any place where it's called.  That's already true right now.  I just want to give you a heads up in case you do this same kind of thing in the future.

Whether or not you make this change is entirely up to you.

scottschurr (approved):
I'm impressed!  Good code, great comments, very thorough testing.  Thanks for fixing these!  :+1:

intelliot:
I think this is ready for @ximinez to review

src/test/app/NFToken_test.cpp
```diff
@@ -5041,6 +5137,757 @@ class NFToken_test : public beast::unit_test::suite
...
 
+    void
+    testIOUWithTransferFee(FeatureBitset features)
+    {
+        using namespace test::jtx;
+
+        testcase("Payments with IOU transfer fees");
+
+        for (auto const& tweakedFeatures :
+             {features - fixUnburnableNFToken, features | fixUnburnableNFToken})
```
shawnxie999, line 5148: Not too sure about this, haven't looked too deeply. but just curious if we could avoid having the for-loop by adding 
`testWithFeats(all - fixUnburnableNFToken);`
in `run()`

And then replace `tweakedFeatures[fixUnburnableNFToken]` with `features[fixUnburnableNFToken]` to check if amendment is enabled or not. 

I **think** it reduces number of test being ran? That's how I did it with my remint test cases. Just curious if this approach is applicable. 

CC @scottschurr

shawnxie999: Actually it may increase the test runs.

I'm going to include the remint change into the amendment as well. And I added `testWithFeats(all - fixUnburnableNFToken);` into my code.

scottschurr: There is, as always, a tension between number of test runs and completeness.

Sometimes unit tests do what @ledhed2222 has done here.  That's based on the test-writer's assessment that the amendment they are adding will not have far-reaching changes.  So changes in amendment state are tested locally in just a few places.

Other times it's not so obvious just how localized the changes are.  In that case it is common to run a bunch of the unit tests in the file with different amendments enabled.  Here's an example: https://github.com/XRPLF/rippled/blob/develop/src/test/app/Offer_test.cpp#L5165-L5181

What I see going on is the intent to pile a _bunch_ of changes -- multiple pull requests -- into a single proposed amendment.  That suggests to me that the later approach (like Offer_test.cpp) is probably appropriate.  Any single change due to the amendment may be simple.  But there are many changes and they may interact.

Just my opinion.

ledhed2222: i'm going to mark this as resolved for now and not change. i think there is an opportunity to simplify the number of test runs happening in this file, but it was already pretty damn complex before and i didn't want to mess with it here

drlongle (approved):

intelliot:
@shawnxie999 could you re-review?

@ximinez will you be able to review this?

ximinez:
> @ximinez will you be able to review this?

@intelliot Yes, but not until after I'm done with the AMM review

src/ripple/app/tx/impl/NFTokenAcceptOffer.cpp
```diff
@@ -206,15 +217,39 @@ NFTokenAcceptOffer::preclaim(PreclaimContext const& ctx)
 
         // The account offering to buy must have funds:
         auto const needed = so->at(sfAmount);
```
shawnxie999, line 219: Just curious why the comment above says `The account offering to buy must have funds`. Isn't this for sell offer?

src/ripple/app/tx/impl/NFTokenAcceptOffer.cpp
```diff
@@ -206,15 +217,39 @@ NFTokenAcceptOffer::preclaim(PreclaimContext const& ctx)
...
+                    ctx.tx[sfAccount],
+                    needed.getCurrency(),
+                    needed.getIssuer(),
+                    fhZERO_IF_FROZEN,
+                    ctx.j) < needed)
+                return tecINSUFFICIENT_FUNDS;
+        }
+        else if (!bo)
+        {
+            // After this amendment, we allow buyers to buy with their own
```
shawnxie999, line 233: Not exactly understand why the comments say `buy` instead of `sell` here as well. Isn't this for sell offer ?

src/ripple/app/tx/impl/NFTokenAcceptOffer.cpp
shawnxie999: Nevermind, if i understand correctly, it check if the buyer who accepts a sell offer has enough money.

intelliot:
@ledhed2222 as mentioned, this is ready for your perusal

src/ripple/app/tx/impl/NFTokenAcceptOffer.cpp
```diff
@@ -230,7 +265,22 @@ NFTokenAcceptOffer::pay(
...
+    // their own currency, we know that something went wrong. This was
+    // originally found in the context of IOU transfer fees. Since there are
+    // several payouts in this tx, just confirm that the end state is OK.
+    if (!view().rules().enabled(fixUnburnableNFToken))
+        return result;
+    if (result != tesSUCCESS)
+        return result;
+    if (accountFunds(view(), from, amount, fhZERO_IF_FROZEN, j_).signum() < 0)
+        return tecINSUFFICIENT_FUNDS;
+    if (accountFunds(view(), to, amount, fhZERO_IF_FROZEN, j_).signum() < 0)
```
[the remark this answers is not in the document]
ledhed2222: this is what actually solves issue 2. we aren't checking whether they have enough in addition to IOU transfer fees. so as the safest option, rather than checking it there, we attempt to do the sale but fail if it resulted in negative balances

shawnxie999:
A question about issued currency. Does the account that issued a particular currency has unlimited supply of it? When I try to understand function `accountFunds`, it seems the currency issuer can never go into debt with its own currency

src/ripple/app/tx/impl/NFTokenAcceptOffer.cpp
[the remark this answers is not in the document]
shawnxie999: cool. i have few more questions
1. what does the `signum()` here do?
2. it says when something goes wrong, the IOU-issuer has a positive of its own currency. Then what is the "normal" balance for the IOU issuer when everything goes right? Is it just 0? (just a general question regarding to how IOU works)

src/ripple/app/tx/impl/NFTokenAcceptOffer.cpp
```diff
@@ -206,15 +217,39 @@ NFTokenAcceptOffer::preclaim(PreclaimContext const& ctx)
...
+                    ctx.j) < needed)
+                return tecINSUFFICIENT_FUNDS;
+        }
+        else if (!bo)
+        {
+            // After this amendment, we allow buyers to buy with their own
+            // issued currency.
+            //
+            // In the case of brokered mode, this check is essentially
+            // redundant, since we have already confirmed that buy offer is >
```
shawnxie999, line 237: NIT: 
I understand the else-if condition is to prevent checking the broker's balance, addressing the first issue.

But in the comment, the word `redundant` kind of confused me, making me think this is just a _redundant_ check, and would be feasible even  without the `!bo` condition. But in reality, `!bo` is actually a mandatory check.

shawnxie999 (approved):
non-important comments, looks good

intelliot:
@ledhed2222 if you agree that this PR is ready to merge, go ahead and put the `Passed` label on it (or just comment saying that it's ready)

(from my perspective, it is ready. we have confirmed with Scott and Ed that no additional reviews are required)

shawnxie999 (approved):

Issue #3327, opened 2020-03-31
Figure out why Doxygen can't resolve every method name in ripple/app/ledger/impl/InboundLedgers.cpp


thejohnfreeman:
Take a look at [this call graph](https://thejohnfreeman.github.io/rippled/classripple_1_1InboundLedgersImp.html#a4410398a5eaebf565d5355feba32dd9b). Now look at the [source of that function](https://thejohnfreeman.github.io/rippled/InboundLedgers_8cpp_source.html#l00069) and all the calls to methods of `InboundLedger`, through the `shared_ptr<InboundLedger>` named `inbound`. Notice how the call graph mentions none of those methods. Notice how the source doesn't cross-reference hyperlink any of those methods (unlike the methods of `app_`, for example).

[This](https://github.com/thejohnfreeman/rippled/tree/fd9a5ca0bb20be5cfd7d5ea907a5bbec20b78758) is the source tree that was used to build that documentation. [This](https://github.com/thejohnfreeman/rippled/blob/fd9a5ca0bb20be5cfd7d5ea907a5bbec20b78758/docs/Doxyfile) is the Doxyfile in that tree.

jatinsahijwani:
Hi, i would like to work on this issue, can you enlighten me, what exactly i have to do, since, i am a complete beginner.
Happy to help.

intelliot:
@jatinsahijwani or anyone else: feel free to work on this issue. Please ask any specific questions by commenting here.

PR #4388, opened 2023-01-15
Added clang-uml config files and generated diagrams.

bramv72:

Adds generated uml diagrams to the docs/diagrams folder.
See https://github.com/bkryza/clang-uml

### Context of Change

POC to determine whether clang-uml can be used to auto-generate uml diagrams, and keep these diagrams up-to-date.

### Type of Change

- [x] Documentation Updates
- [ ] Release

## Optional Future Tasks

- For a few subfolders, diagram generation fails (for different reasons). Investigate and fix.
- Enhance diagrams (increase/decrease number of classes, detail). Split up big ones, merge smaller ones, create diagrams for other subsets of classes, etc.
- The .clang-uml files currently require specifying the classes and namespaces within the subfolder. It would be great if this can be improved, so that these files contain less information about the contents of the corresponding subfolder (i.e. then there'd be no need to add a new class to the .clang-uml file).
- The followed approach to create the .clang-uml files assumed that filenames always correspond to the class names. This is not always the case. This could be adjusted in the .clang-uml files, but requires going over every file/class to verify.
- Add more sequence diagrams. Which functions are good candidates for this?
- Integrate with doxygen? (investigate how)


intelliot:
Hi @bramv72 - thanks for the contribution!

Auto-generated UML diagrams could make sense to store and save in a separate repository.

When it comes to the code in this repo, many of the developers prefer to look at the code directly.

The most useful UML diagrams tend to come from humans, who have pared down the details to produce something focused. Automatically generated UML tends to have a lot of noise.

Given that these diagrams don't really belong in this repo, this PR is being closed. But if you find the diagrams useful, feel free to host in in a new repository. You can also propose an edit to the README (or some other markdown document) to link to the new diagrams, wherever you choose to share them.

Hope that makes sense!

thejohnfreeman:
See if you can tweak these to solve issues like https://github.com/XRPLF/rippled/issues/3327, and integrate the generation into the [Doxygen workflow](https://github.com/XRPLF/rippled/blob/develop/.github/workflows/doxygen.yml), i.e. the `docs` CMake target. That would be pretty useful. I don't think we really want to have generated diagrams (or generated _anything_) tracked in version control.

PR #4395, opened 2023-01-23
Improve MITM detection during session establishment:

### LINKED ISSUE #2413: peer handshake violates TLS transparency

nbougalis:
Even with TLS encrypted connections, it is possible for a determined attacker to mount certain types of relatively easy man-in-the-middle attacks which, if successful, could allow an attacker to tamper with messages exchanged between endpoints.

The risk can be mitigated if each side has a certificate issued by a CA that the other side trusts. In the context of a decentralized and permissionless network, this is neither reasonable nor desirable.

To prevent this problem all we need is to allow the two endpoints, A and B, to be able to independently verify that they are connected to each other over a single end-to-end TLS session, instead of separate TLS sessions with the attacker bridges.

The protocol level handshake implements this security check by using digital signatures: each endpoint derives a fingerprint from the TLS session, which it signs with the private key associated with its own node identity. This strongly binds the TLS session to the identities of the two endpoints of the session.

This commit introduces a new fingerprint derivation that uses modern and standardized TLS exporter functionality, instead of the existing derivation whch uses OpenSSL APIs that are non-standard.

The change is backwards compatible and servers with this commit will still generate and verify old-style fingerprints, in addition to the new style fingerprints.

For a fuller discussion on this topic, please see:
    https://github.com/openssl/openssl/issues/5509
    https://github.com/ripple/rippled/issues/2413

This commit was previously introduced as #3929, which was closed. If merged, it also fixes #2413 (which had been closed as a 'WONTFIX').

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Test Plan

Running this against servers with and without this change and verifying that peer connections are established (e.g. by using the `peers` command) should be sufficient. Ideally, in addition to Linux, this should also be tested against servers running on the "compiles-on-but-it's-not supported" platforms: MacOS and Windows.

### DIFF
--- a/src/ripple/overlay/impl/Handshake.cpp
+++ b/src/ripple/overlay/impl/Handshake.cpp
@@ -26,11 +26,7 @@
 #include <ripple/overlay/impl/Handshake.h>
 #include <ripple/protocol/digest.h>
 #include <boost/regex.hpp>
-#include <algorithm>
 #include <chrono>
-
-// VFALCO Shouldn't we have to include the OpenSSL
-// headers or something for SSL_get_finished?
 
 namespace ripple {
 
@@ -118,10 +114,12 @@ makeFeaturesResponseHeader(
                - `SSL_get_peer_finished`.
     @return `true` if successful, `false` otherwise.
 
-    @note This construct is non-standard. There are potential "standard"
-          alternatives that should be considered. For a discussion, on
-          this topic, see https://github.com/openssl/openssl/issues/5509 and
-          https://github.com/ripple/rippled/issues/2413.
+    @deprecated This construct is non-standard and is now deprecated in favor
+                of using `SSL_export_keying_material`. Support for this will
+                be removed in a future version of the codebase.
+                For a fuller discussion on this topic, please see:
+                    https://github.com/openssl/openssl/issues/5509
+                    https://github.com/ripple/rippled/issues/2413.
 */
 static std::optional<base_uint<512>>
 hashLastMessage(SSL const* ssl, size_t (*get)(const SSL*, void*, size_t))
@@ -136,9 +134,31 @@ hashLastMessage(SSL const* ssl, size_t (*get)(const SSL*, void*, size_t))
 
     sha512_hasher h;
 
-    base_uint<512> cookie;
+    uint512 cookie;
     SHA512(buf, len, cookie.data());
     return cookie;
+}
+
+std::optional<uint256>
+getSessionEKM(stream_type& ssl, std::uint64_t instance, bool outgoing)
+{
+    std::string const label = std::string("XRPL-EKM-COOKIE:V1:") +
+        std::to_string(instance) + (outgoing ? ":OUT" : ":IN");
+
+    uint256 km;
+
+    if (SSL_export_keying_material(
+            ssl.native_handle(),
+            km.data(),
+            km.size(),
+            label.data(),
+            label.size(),
+            nullptr,
+            0,
+            0) == 1)
+        return km;
+
+    return std::nullopt;
 }
 
 std::optional<uint256>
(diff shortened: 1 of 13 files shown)


gregtatcam (approved):
👍 LGTM

intelliot:
`Testable`: See `Test Plan` in the PR description.

nbougalis:
Updated this branch to introduce directional cookies and to strongly bind the `Instance-Cookie`. It's ready for a review and, if approved, merging.

src/ripple/overlay/impl/Handshake.cpp
```diff
@@ -194,16 +215,21 @@ buildHandshake(
...
-        toBase58(TokenType::NodePublic, app.nodeIdentity().first));
+    h.insert("Public-Key", app.getNodePublicIdentity());
 
     {
         auto const sig = signDigest(
             app.nodeIdentity().first, app.nodeIdentity().second, sharedValue);
         h.insert("Session-Signature", base64_encode(sig.data(), sig.size()));
     }
 
+    h.insert("FOO", to_string(ekm));
```
ximinez, line 226: Did you mean to include the raw `ekm`?

src/ripple/overlay/impl/Handshake.h
```diff
@@ -48,6 +48,23 @@ using http_request_type =
 using http_response_type =
     boost::beast::http::response<boost::beast::http::dynamic_body>;
 
+/** Returns a value shared by the two endpoints of a TLS-secured connection.
+
+    This value is generated in a secure fashion and is never communicated over
+    the wire, even over an encrypted connection. Used properly, it can help to
+    detect and prevent preventing active MITM attacks.
```
ximinez, line 55: Typo: `prevent preventing` -> `prevent`

ximinez (changes requested):
Couple of simple fix suggestions.

src/ripple/overlay/impl/Handshake.cpp
JoelKatz: This looks like some temporary/debug code that survived into the final build.

JoelKatz:
LGTM. There's the reference to a renamed section by the old section name and the line of debug code that got left in, but other than that, no issues except:

This code currently violates RFC 5705:

> Label values beginning with "EXPERIMENTAL" MAY be used for private use without registration.  All other label values MUST be registered via Specification Required as described by RFC 5226

We can register our two labels, use experimental labels, or use existing registered labels whose uses do not conflict with ours.

src/ripple/overlay/impl/Handshake.cpp
ximinez: That was my assumption, too. But it does need to be removed.

intelliot:
@nbougalis at your convenience, please respond to the suggestions by ximinez

Issue #4392, opened 2023-01-19
ledger_data returns null instead of empty list when all entries are filtered out


mDuo13:
## Issue Description
When you specify the `type` field to the `ledger_data` method, it is possible no objects of the specified type are found. (This can even occur if those types _do_ exist in the ledger, but not in the section that the server iterated over while serving your request.) When this happens, the `state` field of the response has the value `null` instead of being an empty array (`[]`).

This is not _terrible_, but preferably the response format should maintain the same data type so that clients can handle it consistently. For example, in Python you should be able to do `for entry in state` which would work correctly if `state` is an empty list but would raise an exception if `state` is `null`(or `None` after parsing the JSON in Python).

## Steps to Reproduce
Make a request such as [this one](https://xrpl.org/websocket-api-tool.html?server=wss%3A%2F%2Fs1.ripple.com%2F&req=%7B%22command%22%3A%22ledger_data%22%2C%22type%22%3A%22amendments%22%2C%22binary%22%3Afalse%7D) on Mainnet or any other network with sufficiently large network:

```
{
  "command": "ledger_data",
  "type": "amendments",
  "binary": false
}
```

Note 1: Pagination currently works correctly here, so if you do it correctly you will eventually find a non-null response with the `Amendments` singleton object.

Note 2: I think it's not possible to get a null response without using `type` to filter, since even the genesis ledger has a nonzero number of ledger entries. Maybe you could also get an empty list by carefully crafting a `marker` that represents the end of the ledger data? Unclear, but it wouldn't happen under normal operation.

## Expected Result
The `state` field of the result should be an empty array, `[]`.

## Actual Result
The result has `"state": null` instead:

```json
{
  "result": {
    "ledger": {
      "accepted": true,
      "account_hash": "BB3084F4BDC4893319F092C4371A567D6542B46A5F3178731AF3F6027162C169",
      "close_flags": 0,
      "close_time": 727401202,
      "close_time_human": "2023-Jan-18 23:53:22.000000000 UTC",
      "close_time_resolution": 10,
      "closed": true,
      "hash": "5450D641CE2F78110E589564330E08927A57F6E3A9121AD0A0DED2D910640B99",
      "ledger_hash": "5450D641CE2F78110E589564330E08927A57F6E3A9121AD0A0DED2D910640B99",
      "ledger_index": "77209704",
      "parent_close_time": 727401201,
[...]

PR #4398, opened 2023-01-24
fix: Ledger data returns an empty list (instead of null) when all entries are filtered out

drlongle:

This PR resolves [issue 4392](https://github.com/XRPLF/rippled/issues/4392). When the ```type``` field to the ```ledger_data``` method is specified, it is possible that no objects of the specified type are found. When this happens, the ```state``` field of the response has the value ```null``` instead of an empty array ([]).

### Context of Change

The PR initializes the ```state``` field of the response to an empty list so that the field is not a ```null```. This can break client code that explicitly checks for ```null```. However, this behavior is consistent with the documentation.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### DIFF
--- a/src/test/rpc/LedgerData_test.cpp
+++ b/src/test/rpc/LedgerData_test.cpp
@@ -314,6 +314,34 @@ public:
         auto const USD = gw["USD"];
         env.fund(XRP(100000), gw);
 
+        auto makeRequest = [&env](Json::StaticString const& type) {
+            Json::Value jvParams;
+            jvParams[jss::ledger_index] = "current";
+            jvParams[jss::type] = type;
+            return env.rpc(
+                "json",
+                "ledger_data",
+                boost::lexical_cast<std::string>(jvParams))[jss::result];
+        };
+
+        // Assert that state is an empty array.
+        for (auto const& type :
+             {jss::amendments,
+              jss::check,
+              jss::directory,
+              jss::fee,
+              jss::offer,
+              jss::signer_list,
+              jss::state,
+              jss::ticket,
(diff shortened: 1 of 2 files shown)


src/test/rpc/LedgerData_test.cpp
```diff
@@ -314,6 +314,34 @@ class LedgerData_test : public beast::unit_test::suite
         auto const USD = gw["USD"];
         env.fund(XRP(100000), gw);
 
+        auto makeRequest = [&env](Json::StaticString const& type) {
```
justinr1234, line 317: I haven't seen this syntax before. I see it's creating a function, but what does `[&env]` do?

justinr1234 (approved):

src/test/rpc/LedgerData_test.cpp
ximinez: https://en.cppreference.com/w/cpp/language/lambda

drlongle: `[&env]` captures the variable `env` by reference.

thejohnfreeman (approved):

intelliot:
@drlongle if you agree that this is ready to merge, please comment with a suggested commit message for this PR

drlongle:
Suggested commit message

When no entries are found, make ledger_data return an empty array instead of null: (#4392)

When the type field for the ledger_data method is specified, it is possible that no objects of the specified type are found. When this happens, the state field of the response has the value null instead of an empty array. According to the documentation, the state field should be an array.

Issue #4373, opened 2023-01-02
[Bug] Buy/Sell NFTokenOffers with Destination can be brokered by accounts not set as Destination


nixer89:
# Description
When trading NFTs, one can set a "Destination" for his NFTokenOffer. This has the purpose that only the Destination account can accept this NFTokenOffer. It allows direkt NFT transfers or brokered mode for NFT trades.

However, there seems to be a bug with the brokered mode for NFTokenOffers when:

1. Sell-Offer has a Destination account set and there is a Buy-Offer without `Destination` from the account mentioned in the `Destination` field of the Sell-Offer.
2. Buy-Offer has a Destination account set and there is a Sell-Offer without `Destination` from the account mentioned in the `Destination` field of the Buy-Offer.

Here is an example:
Transaction:
`E67A0EA8861E597CF2248C449002FC62E55AAF6A375F2B3E616ADD20A32AFAF7`
is a brokered NFTokenAcceptOffer.

It brokers the Sell offer created in the transaction:
`948C4F6BDC8099200252BB24EDE62B8638617635A2331991DC9640812A8D9C26`

which has the account `r4CizHZZYpDgrRbWavMiNecuyRdbTLTPh` set as destination. And 0 XRP as amount:

```
"DeletedNode": {
                    "FinalFields": {
                        "Amount": "0",
                        "Destination": "r4CizHZZYpDgrRbWavMiNecuyRdbTLTPh",
                        "Flags": 1,
                        "NFTokenID": "0008138873C516A5EF9A8E860889E355EC9B93EEAECFEC7B46C9F20F0000026A",
                        "NFTokenOfferNode": "0",
                        "Owner": "rBZ3FrRzTgy46H3C9Uede8CAuH7rTGA5iE",
                        "OwnerNode": "18",
                        "PreviousTxnID": "948C4F6BDC8099200252BB24EDE62B8638617635A2331991DC9640812A8D9C26",
                        "PreviousTxnLgrSeq": 76810377
                    },
                    "LedgerEntryType": "NFTokenOffer",
                    "LedgerIndex": "314B0A42F7B13E9791DBB8C68982334A1E11DD42B59A3A36C28F76F6FDC46998"
                }
```

and the buy offer from account `r4CizHZZYpDgrRbWavMiNecuyRdbTLTPh` created in transaction:
`88ED6197673596F6FF8E7A1BD415B53AD945232B8D1CA87D274A3BF6A4E60429`

[...]

PR #4399, opened 2023-01-25
fixNFTokenBrokerAccept

dangell7:

Currently with NFTs using broker mode if the sell offer contains a destination and that destination is the buyer account, anyone can broker the transaction.

It is also true that if a buy offer contains a destination and that destination is the seller account, anyone can broker the transaction.

Imo this is not ideal and is misleading to everyone.

The comments in the code are;

```
// If the buyer specified a destination, that destination must be
// the seller or the broker.
```

This is misleading because the broker is the account submitting the transaction. Its anyone. Again, if the buyer specified a destination and that destination is the seller, anyone can settle the tx. I don't think that makes sense. Why even have that? If anyone can settle the tx then there is no point in having this arithmetic I think.

I believe the point was; If you set a destination, that destination needs to be the person settling the transaction.

### Context of Change

The changes made now force the broker to be the destination if they want to settle. If the buyer is the destination, then the buyer must accept the sell offer, as you cannot broker your own offers.

If users want their offers open to the public then not setting a destination would be the recommended way. Inversely, if users want to limit who can settle the offers, then they would set a destination.

There is also a discussion issue [here](https://github.com/XRPLF/rippled/issues/4373):

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Test Plan

I also added a test function to show one of the more recent issues. There are 2 things tested. 

1. The broker cannot broker a destination offer to the buyer/seller and the buyer must accept the sell offer. (0 transfer)
2. If the broker is the destination, the broker will take the difference. (broker mode)

### DIFF
--- a/src/ripple/app/tx/impl/NFTokenAcceptOffer.cpp
+++ b/src/ripple/app/tx/impl/NFTokenAcceptOffer.cpp
@@ -112,19 +112,19 @@ NFTokenAcceptOffer::preclaim(PreclaimContext const& ctx)
         if ((*so)[sfAmount] > (*bo)[sfAmount])
             return tecINSUFFICIENT_PAYMENT;
 
-        // If the buyer specified a destination, that destination must be
-        // the seller or the broker.
+        // If the buyer specified a destination
         if (auto const dest = bo->at(~sfDestination))
         {
-            if (*dest != so->at(sfOwner) && *dest != ctx.tx[sfAccount])
+            // that destination must be the tx account
+            if (*dest != ctx.tx[sfAccount])
                 return tecNFTOKEN_BUY_SELL_MISMATCH;
         }
 
-        // If the seller specified a destination, that destination must be
-        // the buyer or the broker.
+        // If the seller specified a destination
         if (auto const dest = so->at(~sfDestination))
         {
-            if (*dest != bo->at(sfOwner) && *dest != ctx.tx[sfAccount])
+            // that destination must be the tx account
+            if (*dest != ctx.tx[sfAccount])
                 return tecNFTOKEN_BUY_SELL_MISMATCH;
         }
 
--- a/src/ripple/protocol/Feature.h
+++ b/src/ripple/protocol/Feature.h
@@ -74,7 +74,7 @@ namespace detail {
 // Feature.cpp. Because it's only used to reserve storage, and determine how
 // large to make the FeatureBitset, it MAY be larger. It MUST NOT be less than
 // the actual number of amendments. A LogicError on startup will verify this.
-static constexpr std::size_t numFeatures = 54;
+static constexpr std::size_t numFeatures = 55;
 
 /** Amendments that this server supports and the default voting behavior.
    Whether they are enabled depends on the Rules defined in the validated
@@ -341,6 +341,7 @@ extern uint256 const fixTrustLinesToSelf;
 extern uint256 const fixRemoveNFTokenAutoTrustLine;
 extern uint256 const featureImmediateOfferKilled;
 extern uint256 const featureDisallowIncoming;
+extern uint256 const fixNFTokenBrokerAccept;
 
 }  // namespace ripple
 
--- a/src/test/app/NFToken_test.cpp
+++ b/src/test/app/NFToken_test.cpp
@@ -2926,20 +2926,19 @@ class NFToken_test : public beast::unit_test::suite
             BEAST_EXPECT(ownerCount(env, minter) == 1);
             BEAST_EXPECT(ownerCount(env, buyer) == 2);
 
-            // Broker is successful when destination is buyer.
-            env(token::brokerOffers(
-                broker, offerMinterToBuyer, offerBuyerToMinter));
+            // Buyer is successful when destination is buyer.
+            env(token::acceptBuyOffer(buyer, offerMinterToBuyer));
             env.close();
             BEAST_EXPECT(ownerCount(env, issuer) == 1);
             BEAST_EXPECT(ownerCount(env, minter) == 1);
-            BEAST_EXPECT(ownerCount(env, buyer) == 0);
+            BEAST_EXPECT(ownerCount(env, buyer) == 1);
 
             // Clean out the unconsumed offer.
             env(token::cancelOffer(issuer, {offerIssuerToBuyer}));
             env.close();
             BEAST_EXPECT(ownerCount(env, issuer) == 0);
             BEAST_EXPECT(ownerCount(env, minter) == 1);
-            BEAST_EXPECT(ownerCount(env, buyer) == 0);
+            BEAST_EXPECT(ownerCount(env, buyer) == 1);
         }
 
         // Show that if a buy and a sell offer both have the same destination,
@@ -2965,7 +2964,7 @@ class NFToken_test : public beast::unit_test::suite
             env.close();
             BEAST_EXPECT(ownerCount(env, issuer) == 0);
             BEAST_EXPECT(ownerCount(env, minter) == 2);
-            BEAST_EXPECT(ownerCount(env, buyer) == 1);
+            BEAST_EXPECT(ownerCount(env, buyer) == 2);
 
             // Broker is successful if they are the destination of both offers.
             env(token::brokerOffers(
@@ -2973,7 +2972,7 @@ class NFToken_test : public beast::unit_test::suite
             env.close();
             BEAST_EXPECT(ownerCount(env, issuer) == 0);
             BEAST_EXPECT(ownerCount(env, minter) == 0);
-            BEAST_EXPECT(ownerCount(env, buyer) == 1);
+            BEAST_EXPECT(ownerCount(env, buyer) == 2);
         }
     }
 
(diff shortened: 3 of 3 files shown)


src/ripple/app/tx/impl/NFTokenAcceptOffer.cpp
```diff
@@ -112,20 +112,44 @@ NFTokenAcceptOffer::preclaim(PreclaimContext const& ctx)
...
+            {
+                return tecNFTOKEN_BUY_SELL_MISMATCH;
+            }
+
+            // fixNFTokenBrokerAccept: Enabled
+            // the destination must be the tx account
+            if (ctx.view.rules().enabled(fixNFTokenBrokerAccept) &&
+                *dest != ctx.tx[sfAccount])
+            {
                 return tecNFTOKEN_BUY_SELL_MISMATCH;
```
nixer89, line 131: I personally would expect a `tecNO_PERMISSION` error code here, suggesting that the transaction submitting account has no permission to accept this offer. Similar when beeing in "non broker mode":

```
// If not in bridged mode...
        if (!so)
        {
            // If the offer has a Destination field, the acceptor must be the
            // Destination.
            if (auto const dest = bo->at(~sfDestination);
                dest.has_value() && *dest != ctx.tx[sfAccount])
                return tecNO_PERMISSION;
        }
```

The error code `tecNFTOKEN_BUY_SELL_MISMATCH` might be misleading.

dangell7: I agree. `tecNO_PERMISSION` is a better error response here. I will make this change.

nixer89:
@intelliot : this bug really hurts people out there. I request that this fix is included in the next planned release.

This PR solves the issue [4373](https://github.com/XRPLF/rippled/issues/4373)

bharathchari:
Pending any issues that come up during reviews, I think this should go into the next release. These are some example transactions that are a result of this issue:
	
https://bithomp.com/explorer/BA4F42227EC87C5447B94E3FFD376D58B562601C413B103B6EB1FEC3621F72D0
https://bithomp.com/explorer/580ED05BD1F866700C7FA2882065D0CDCBDC1B6D98A7D76F281C39AA7B19F5DA
https://bithomp.com/explorer/E1E70B5F407A13F894258699033D26171501F4AC4455CF1E75F215519EF0E41F

intelliot:
@bharathchari Would it be OK for this to be rolled into the same NFT bugfix amendment that is currently in development? i.e. this branch: https://github.com/XRPLF/rippled/tree/feature/nft-fixes - which will likely include https://github.com/XRPLF/rippled/pull/4380 and https://github.com/XRPLF/rippled/pull/4403, if approved.

Or, should this be its own amendment?

src/ripple/app/tx/impl/NFTokenAcceptOffer.cpp
```diff
@@ -112,20 +112,44 @@ NFTokenAcceptOffer::preclaim(PreclaimContext const& ctx)
...
+            }
+
+            // fixNFTokenBrokerAccept: Enabled
+            // the destination must be the tx account
+            if (ctx.view.rules().enabled(fixNFTokenBrokerAccept) &&
+                *dest != ctx.tx[sfAccount])
+            {
+                return tecNO_PERMISSION;
```
shawnxie999, line 151: Curious why would we use `tecNO_PERMISSION` instead of `tecNFTOKEN_BUY_SELL_MISMATCH` in this case

nixer89:
> @bharathchari Would it be OK for this to be rolled into the same NFT bugfix amendment that is currently in development? i.e. this branch: https://github.com/XRPLF/rippled/tree/feature/nft-fixes - which will likely include #4380 and #4403, if approved.
> 
> Or, should this be its own amendment?

I think this would be okay. Maybe we can then rename the amendment, away from `unburnableNFT` and into something more general, since it includes so many different fixes? any thoughts? :)

src/test/app/NFToken_test.cpp
```diff
@@ -2957,15 +3015,32 @@ class NFToken_test : public beast::unit_test::suite
...
+                // Cannot broker offers when the sell destination is not the
+                // buyer
+                env(token::brokerOffers(
+                        issuer, offerBuyerToBroker, offerMinterToBroker),
+                    ter(tecNO_PERMISSION));
+                env.close();
+                BEAST_EXPECT(ownerCount(env, issuer) == 0);
+                BEAST_EXPECT(ownerCount(env, minter) == 2);
+                BEAST_EXPECT(ownerCount(env, buyer) == 1);
+            }
```
scottschurr, line 3043: I noticed that the only difference between the contents of the `if` and `else` branches is the `tec` code. Consider reducing the duplication by replacing the if/else with something like this:
```
            {
                // Cannot broker offers when the sell destination is not the
                // buyer or the broker.
                TER const expectTer = features[fixNFTokenBrokerAccept]
                    ? tecNO_PERMISSION : tecNFTOKEN_BUY_SELL_MISMATCH;
                env(token::brokerOffers(
                        issuer, offerBuyerToBroker, offerMinterToBroker),
                    ter(expectTer));
                env.close();
                BEAST_EXPECT(ownerCount(env, issuer) == 0);
                BEAST_EXPECT(ownerCount(env, minter) == 2);
                BEAST_EXPECT(ownerCount(env, buyer) == 1);
            }
```

src/test/app/NFToken_test.cpp
```diff
@@ -2872,15 +2872,32 @@ class NFToken_test : public beast::unit_test::suite
...
+                // issuer cannot broker the offers, because they are not the
+                // Destination.
+                env(token::brokerOffers(
+                        issuer, offerBuyerToMinter, offerMinterToBroker),
+                    ter(tecNO_PERMISSION));
+                env.close();
+                BEAST_EXPECT(ownerCount(env, issuer) == 0);
+                BEAST_EXPECT(ownerCount(env, minter) == 2);
+                BEAST_EXPECT(ownerCount(env, buyer) == 1);
+            }
```
scottschurr, line 2900: I noticed that the only difference between the contents of the `if` and `else` branches is the `tec` code.  Consider reducing the duplication by replacing the `if/else` with something like this:
```
            {
                // issuer cannot broker the offers, because they are not the
                // Destination.
                TER const expectTer = features[fixNFTokenBrokerAccept]
                    ? tecNO_PERMISSION : tecNFTOKEN_BUY_SELL_MISMATCH;
                env(token::brokerOffers(
                        issuer, offerBuyerToMinter, offerMinterToBroker),
                    ter(expectTer));
                env.close();
                BEAST_EXPECT(ownerCount(env, issuer) == 0);
                BEAST_EXPECT(ownerCount(env, minter) == 2);
                BEAST_EXPECT(ownerCount(env, buyer) == 1);
            }
```

scottschurr (approved):
:+1: Looks like a reasonable solution to the problem to me.  Nice work with the unit tests.

I did leave comments on a few things that I think might be improvements, but they needn't hold up this pull request.  Those changes can be made at your discretion.  Thanks for the contribution!

dangell7:
> 👍 Looks like a reasonable solution to the problem to me. Nice work with the unit tests.
> 
> I did leave comments on a few things that I think might be improvements, but they needn't hold up this pull request. Those changes can be made at your discretion. Thanks for the contribution!

Thank you very much for the review. I went ahead an made the changes. I also used your suggestion and updated the 3 test [here](https://github.com/XRPLF/rippled/pull/4399/files#diff-a6951aa57c7ac4fcd2b2d445c3c40664fc3b80e66acb3b69dbf00b0b7ae3d66dR2950).

scottschurr (approved):
:+1: Thanks!

intelliot:
> > @bharathchari Would it be OK for this to be rolled into the same NFT bugfix amendment that is currently in development? i.e. this branch: https://github.com/XRPLF/rippled/tree/feature/nft-fixes - which will likely include #4380 and #4403, if approved.
> > Or, should this be its own amendment?
> 
> I think this would be okay. Maybe we can then rename the amendment, away from `unburnableNFT` and into something more general, since it includes so many different fixes? any thoughts? :)

Yes, a suggested name is: `NonFungibleTokensV1_2`, though other ideas are still welcome.

alloynetworks:
If all the pull requests are fixes then the name should start with “fix”. Also only if it’s a fix amendment can it be with default Yes.

intelliot:
> If all the pull requests are fixes then the name should start with “fix”.

Concretely, do you suggest `fixNonFungibleTokensV1_2`? The downside of dropping the `V1_2` (or similar) is that there may be additional fixes in the future.

> Also only if it’s a fix amendment can it be with default Yes.

When the amendment activates, any servers running previous versions will be [amendment blocked](https://xrpl.org/amendments.html#amendment-blocked-servers) and essentially cease to function. For this reason, we'll leave the default at No in the release that first understands the amendment, and consider setting it to Yes in this next release. This increases the likelihood that server operators will have more time to review the release and update.

This is just a general practice for any amendment that isn't extremely urgent; a fix for a particularly severe issue could default to Yes.

alloynetworks:
I’m fine with this approach.

RichardAH:
`fixDescriptionOfBug` should be the preferred naming convention imho. It requires one fewer steps to lookup/understand what the amendment is for. Particularly when you are looking through code and see a section guarded by this amendment you will know immediately if the guarded code is in the right place and more or less if it is doing the right thing.

You could prefix with fixNFT if you really want to group them together.

shawnxie999 (approved):
looks good

intelliot:
> `fixDescriptionOfBug` should be the preferred naming convention imho. It requires one fewer steps to lookup/understand what the amendment is for. Particularly when you are looking through code and see a section guarded by this amendment you will know immediately if the guarded code is in the right place and more or less if it is doing the right thing.
> 
> You could prefix with fixNFT if you really want to group them together.

Thank you. Discussed with Scott B this morning and we will likely group the fixes together in `fixNonFungibleTokensV1_2`. It should be thought of as NFTs v1.2, to be activated via this fix amendment.

intelliot:
@dangell7 no need to action this now, but for future reference: there is no need to squash your commits. In fact, I would prefer that you leave the original commit history, so that subsequent changes are easier to follow. We now squash PRs when merging, so it will all become 1 commit anyway.

intelliot:
(per discussion with Shawn, this PR is awaiting review by @ledhed2222 )

nixer89 (commented):
lgtm

src/ripple/protocol/Feature.h
```diff
@@ -74,7 +74,7 @@ namespace detail {
 // Feature.cpp. Because it's only used to reserve storage, and determine how
 // large to make the FeatureBitset, it MAY be larger. It MUST NOT be less than
 // the actual number of amendments. A LogicError on startup will verify this.
-static constexpr std::size_t numFeatures = 54;
+static constexpr std::size_t numFeatures = 55;
```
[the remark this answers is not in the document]
ximinez: It'll generally get resolved during one of the merge processes, usually when the owner of the PR merges `develop` in, but sometimes the other way around.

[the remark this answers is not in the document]
intelliot: Yep, that will be an easy merge - no problem.

intelliot:
(@ledhed2222 indicated that he will review this before Monday)

src/test/app/NFToken_test.cpp
```diff
@@ -5079,13 +5112,15 @@ class NFToken_test : public beast::unit_test::suite
         using namespace test::jtx;
         FeatureBitset const all{supported_amendments()};
         FeatureBitset const fixNFTDir{fixNFTokenDirV1};
+        FeatureBitset const fixNFTBroker{fixUnburnableNFToken};
```
ledhed2222, line 5115: let's get rid of this - it will cause additional test runs for no real benefit, since we are using the same feature flag

src/test/app/NFToken_test.cpp
```diff
@@ -5079,13 +5112,15 @@ class NFToken_test : public beast::unit_test::suite
...
 
         testWithFeats(all - fixNFTDir);
         testWithFeats(all - disallowIncoming);
+        testWithFeats(all - fixNFTBroker);
```
ledhed2222, line 5119: ditto

src/ripple/app/tx/impl/NFTokenAcceptOffer.cpp
```diff
@@ -112,20 +112,44 @@ NFTokenAcceptOffer::preclaim(PreclaimContext const& ctx)
...
                 return tecNFTOKEN_BUY_SELL_MISMATCH;
+            }
+
+            // fixUnburnableNFToken: Enabled
+            // the destination may only be the account brokering the offer
+            if (ctx.view.rules().enabled(fixUnburnableNFToken) &&
+                *dest != ctx.tx[sfAccount])
+            {
+                return tecNO_PERMISSION;
+            }
```
ledhed2222, line 132: it's not a blocker, but you are checking the rule twice when you could check it just once with some conditional re-ordering

src/test/app/NFToken_test.cpp
```diff
@@ -2872,15 +2872,21 @@ class NFToken_test : public beast::unit_test::suite
...
-            // issuer cannot broker the offers, because they are not the
-            // Destination.
-            env(token::brokerOffers(
-                    issuer, offerBuyerToMinter, offerMinterToBroker),
-                ter(tecNFTOKEN_BUY_SELL_MISMATCH));
-            env.close();
-            BEAST_EXPECT(ownerCount(env, issuer) == 0);
-            BEAST_EXPECT(ownerCount(env, minter) == 2);
-            BEAST_EXPECT(ownerCount(env, buyer) == 1);
+            // fixUnburnableNFToken: Disabled
```
ledhed2222, line 2875: this comment is extraneous

src/test/app/NFToken_test.cpp
```diff
@@ -2917,29 +2923,51 @@ class NFToken_test : public beast::unit_test::suite
...
+                BEAST_EXPECT(ownerCount(env, minter) == 1);
+                BEAST_EXPECT(ownerCount(env, buyer) == 2);
 
-            // Clean out the unconsumed offer.
-            env(token::cancelOffer(issuer, {offerIssuerToBuyer}));
-            env.close();
-            BEAST_EXPECT(ownerCount(env, issuer) == 0);
-            BEAST_EXPECT(ownerCount(env, minter) == 1);
-            BEAST_EXPECT(ownerCount(env, buyer) == 0);
+                // Broker is successful when destination is buyer.
```
ledhed2222, line 2941: this comment is only true for the previous state - now the broker is _NOT_ successful when the destination is the buyer

ledhed2222 (approved):
approved - there's some cleanup that can occur here but it's largely cosmetic

src/test/app/NFToken_test.cpp
dangell7: Gone

src/test/app/NFToken_test.cpp
dangell7: ditto

src/ripple/app/tx/impl/NFTokenAcceptOffer.cpp
dangell7: rewrote it please check it out.

src/test/app/NFToken_test.cpp
dangell7: gone

src/test/app/NFToken_test.cpp
dangell7: rewrote this to be more clear.

dangell7:
@ledhed2222 made the suggested changes. Thank you for reviewing

src/ripple/app/tx/impl/NFTokenAcceptOffer.cpp
ledhed2222: it's better looking to me! this is fine for merge, but IMO I would do this to remove a level of indentation:


```c++
if (ctx.view.rules().enabled(fixUnburnableNFToken))
{
  if (*dest != ctx.tx[sfAccount])
    return tecNO_PERMISSION;
}
else if (*dest != so->at(sfOwner) && *dest != ctx.tx[sfAccount])
  return tecNFTOKEN_BUY_SELL_MISMATCH;
```

src/ripple/app/tx/impl/NFTokenAcceptOffer.cpp
```diff
@@ -112,20 +112,40 @@ NFTokenAcceptOffer::preclaim(PreclaimContext const& ctx)
...
+                // the destination may only be the account brokering the offer
+                if (*dest != ctx.tx[sfAccount])
+                    return tecNO_PERMISSION;
+            }
+            else
+            {
+                // the destination must be the buyer or the broker.
+                if (*dest != bo->at(sfOwner) && *dest != ctx.tx[sfAccount])
+                    return tecNFTOKEN_BUY_SELL_MISMATCH;
+            }
```
ledhed2222, line 148: ditto

src/test/app/NFToken_test.cpp
```diff
@@ -5088,4 +5121,4 @@ class NFToken_test : public beast::unit_test::suite
 
 BEAST_DEFINE_TESTSUITE_PRIO(NFToken, tx, ripple, 2);
 
-}  // namespace ripple
+}  // namespace ripple
```
ledhed2222, line 5124: missing a newline but not a blocker

Issue #4337, opened 2022-11-09
API should never transparently interpret "account" as a seed (Version: 1.9.4)


intelliot:
## Issue Description

In the API, seeds can be used in place of accounts. For example, when calling account_info, you can pass `"account": "foo"`. The string "foo" is treated like a seed, so the method returns `actNotFound` (instead of `actMalformed`, as most developers would intuitively expect).

In the early days, this was a convenience to make testing easier. However, it is no longer a good idea, and it allows for poor security practices. It is controlled by the `strict` option on some methods.

Note, Clio currently does not behave this way. This results in a subtle incompatibility between Clio and rippled, even though Clio is intended to support (and eventually replace) the rippled API. The current thinking from the Clio team is that Clio should not behave this way, i.e. it should not "support" the ability to provide a seed in the "account" field of API requests.

## Steps to Reproduce

Call account_info with `"account": "foo"`.

## Expected Result
The method should return `actMalformed` since "foo" is not a valid account.

## Actual Result
The string "foo" is treated like a seed, so the method returns `actNotFound`.

## Notes

This behavior, since it is available on public-facing servers like s2, allows bad actors to use servers to try to brute-force account seeds, for "free". However, this is not a worry, as accounts are not brute-forceable regardless. While we should have spam prevention (e.g. rate limiting), trying many seeds (with no entropy-reducing "hints") is not a viable attack for attempting to compromise an account.

Removing this behavior from the API is a [breaking change](https://xrpl.org/request-formatting.html#breaking-changes). We could argue that it shouldn't be done without bumping the API version, and we are open to that (depending on user demand). However, in this instance, we doubt that anyone is using the API in the "legacy" way, and it is a huge security hole. It allows/encourages users to send secrets to places where they are not needed, where they could end up in logs, error messages, etc. There's no reason to take such a risk with a seed/secret, since only the public address is needed.

We prefer to break the insecure behavior (by making it safer), even if it means that some implementations may experience a breaking change and need to update. (And even that is unlikely.)

jatinsahijwani:
I want to work on this issue, please guide me, i am a beginner

CrypticBantu:
I would like to work on this issue, could you assign this to me please?

intelliot:
@jatinsahijwani welcome! What kind of help are you looking for?

intelliot:
will be fixed by https://github.com/XRPLF/rippled/pull/4404

Issue #3330, opened 2020-04-02
Drop non-strict account parsing


mDuo13:
## Summary
Relating to #3329, I think non-strict account parsing should be removed.

## Motivation
Accepting secret values in the API just encourages poor operational security and makes the API definition more ambiguous.

## Solution
**This is a breaking change requiring an API version bump.**

All API fields that accept account addresses in base58 should _not_ accept passphrases, secret keys, public keys, RFC-1751 phrases, or other variant forms. (X-Address support would be fine though.)

This applies to all fields named `account`, as well as some other cases like `peer`, `taker`, `issuer`, members of `hotwallets`, etc. I haven't extensively tested which of these fields accept secrets/etc. but the answer should be "none of them."

After doing this, the `strict` parameter can be removed from all API methods that have it.

intelliot:
@drlongle could you add a comment here re: your thoughts / questions?

drlongle:
I would like to know why public keys should not be accepted. Is it because we want to simplify the API and argument parsing (i.e. guessing whether a provided argument is an address or a public key).

@mDuo13 Can you clarify? Thanks.

intelliot:
**drlongle:** What is the rationale for not accepting public key to look up account? Do we have a consensus about that?

**mDuo13:** There is no security issue with using public keys, but for simplicity and consistency, it makes the most sense to accept only one data type per field. I still think the ticket is relevant and worth doing

intelliot:
Resolved by https://github.com/XRPLF/rippled/pull/4404, expected to be released in v1.11, target release date ~June 2023.

PR #4404, opened 2023-01-31
API does not accept seed or public key for account

### LINKED ISSUE #3329: account_objects CLI usage statement is wrong (Version: 1.5.0)
## Issue Description
(Discovered while documenting #3196.) The usage statement for the `account_objects` method is wrong. [[Source code]](https://github.com/ripple/rippled/blob/f00f263852c472938bf8e993e26c7f96f435935c/src/ripple/app/main/Main.cpp#L135)

The account_objects method does not support the `strict` parameter, but the commandline implies it does.

## Steps to Reproduce

Start rippled in stand-alone mode, then run this command:

```sh
rippled account_objects masterpassphrase current strict
```

## Expected Result
The result should be an `actMalformed` error, similar to what you get if you do `account_info` with the same arguments.

```json
{
   "result" : {
      "error" : "actMalformed",
      "error_code" : 35,
      "error_message" : "Account malformed.",
      "request" : {
         "account" : "masterpassphrase",
         "api_version" : 1,
         "command" : "account_objects",
         "strict" : 1
      },
      "status" : "error"
   }
}
```

The "easy" fix is to correct the usage statement not to mention the `strict` parameter, which is not supported in any interface for `account_objects`. But my preferred fix is to **add `strict` support to `account_objects`** for consistency with other account_* methods.

## Actual Result
The server returns a successful response for the genesis account:

```json
{
   "result" : {
      "account" : "rHb9CJAWyB4rj91VRWn96DkukG4bwdtyTh",
      "account_objects" : [],
      "ledger_current_index" : 3,
      "status" : "success",
      "validated" : false
   }
}
```

drlongle:

This PR resolves #3329, #3330 and #4337. Currently, `seeds` and `public keys` can be used in place of ```accounts``` at several locations in the API. This increases the API's complexity because the API needs to guess whether the input is a seed, a public key or an address. This PR prevents the API from accepting a seed or a public key as account ID. Since the API does not interpret ```account``` as ```seed```, the option ```strict``` in several API methods is no longer needed and is removed.

### Context of Change

Documentation needs to be updated.

### Type of Change

- [x] Documentation Updates
- [ ] Release


src/ripple/rpc/handlers/AccountChannels.cpp
```diff
@@ -76,25 +76,29 @@ doAccountChannels(RPC::JsonContext& context)
...
+    if (!ledger->exists(keylet::account(*accountID)))
         return rpcError(rpcACT_NOT_FOUND);
 
     std::string strDst;
     if (params.isMember(jss::destination_account))
         strDst = params[jss::destination_account].asString();
     auto hasDst = !strDst.empty();
 
-    AccountID raDstAccount;
+    std::optional<AccountID> raDstAccount{AccountID()};
```
justinr1234, line 94: Can you explain this line to me and why it was changed?

src/ripple/rpc/handlers/AccountLines.cpp
```diff
@@ -96,31 +96,29 @@ doAccountLines(RPC::JsonContext& context)
...
 
-    std::string strIdent(params[jss::account].asString());
-    AccountID accountID;
-
-    if (auto jv = RPC::accountFromString(accountID, strIdent))
+    std::string const strIdent(params[jss::account].asString());
+    auto const accountID = parseBase58<AccountID>(strIdent);
+    if (!accountID)
     {
-        for (auto it = jv.begin(); it != jv.end(); ++it)
```
justinr1234, line 104: What did this old code do and why is it removed?

src/test/app/TxQ_test.cpp
```diff
@@ -3097,11 +3097,13 @@ class TxQ1_test : public beast::unit_test::suite
         env.fund(XRP(1000000), alice);
         env.close();
 
+        auto const aliceAccountIDStr = toBase58(alice.id());
         auto const withQueue =
-            R"({ "account": ")" + alice.human() + R"(", "queue": true })";
```
justinr1234, line 3101: What does removed `human()` do to the return value? Is it equivalent?

drlongle: [Account::human()](https://github.com/XRPLF/rippled/blob/develop/src/test/jtx/Account.h#L112) returns the human readable public key. [Account::id()](https://github.com/XRPLF/rippled/blob/develop/src/test/jtx/Account.h#L106) returns the uint160 hash of the public key. Those changes are actually unnecessary. I revert them.

src/ripple/rpc/handlers/AccountChannels.cpp
drlongle: The old code calls [RPC::accountFromString()](https://github.com/XRPLF/rippled/blob/develop/src/ripple/rpc/impl/RPCHelpers.cpp#L85) with  `bStrict = false` (default value). This function attempts different ways to parse the input string (it checks whether the input is a public key, an account ID or a seed). If one of these attempts works, the function fills the content of the pointer of type `AccountID` and returns an empty Json object. Otherwise, the function returns a Json object that contains an error code and an error message.

Since we don't accept seed as a way to look up an account, I call the function [parseBase58()](https://github.com/XRPLF/rippled/blob/develop/src/ripple/protocol/impl/AccountID.cpp#L114) instead. This function has a different syntax. It returns an `std::optional<AccountID>` object that only contains an `AccountID` when the input string is actually an account ID (a valid hash).

src/ripple/rpc/handlers/AccountLines.cpp
drlongle: The function [RPC::accountFromString()](https://github.com/XRPLF/rippled/blob/develop/src/ripple/rpc/impl/RPCHelpers.cpp#L85) attempts different ways to parse the input string (it checks whether the input is a public key, an account ID or a seed). If  all attempts fail, the function returns a Json object that contains an error code and an error message. The `for` loop traverses the return Json object and add all its fields (error code and error message) to the `Json` result object.

The new code calls the function [inject_error](https://github.com/XRPLF/rippled/blob/develop/src/ripple/protocol/ErrorCodes.h#L212) and achieves the same effect.

seelabs (commented):
This looks good, thanks. I'm not sure I would have removed the public_key methods. That isn't a security concern, but I also understand the motivation - I don't object too strongly. I left a few suggestions. Nice job on this.

src/ripple/rpc/handlers/AccountLines.cpp
```diff
@@ -96,31 +96,29 @@ doAccountLines(RPC::JsonContext& context)
...
+    if (!ledger->exists(keylet::account(*accountID)))
         return rpcError(rpcACT_NOT_FOUND);
 
     std::string strPeer;
     if (params.isMember(jss::peer))
         strPeer = params[jss::peer].asString();
     auto hasPeer = !strPeer.empty();
 
-    AccountID raPeerAccount;
+    std::optional<AccountID> raPeerAccount{AccountID()};
```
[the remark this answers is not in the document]
drlongle: Thanks. Fixed.

src/ripple/rpc/handlers/AccountOffers.cpp
```diff
@@ -137,11 +134,11 @@ doAccountOffers(RPC::JsonContext& context)
     std::uint64_t nextHint = 0;
     if (!forEachItemAfter(
             *ledger,
-            accountID,
+            *accountID,
             startAfter,
             startHint,
             limit + 1,
-            [&offers, &count, &marker, &limit, &nextHint, &accountID](
+            [&offers, &count, &marker, &limit, &nextHint, accountID](
```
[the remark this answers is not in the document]
drlongle: You are right. Fixed.

seelabs (approved):
:+1: nice job on this.

intelliot:
@drlongle it isn't necessary to rebase and force-push (though you can if you prefer - it's just a little trickier for reviewers to follow the changes). Using merge commits is fine if we're going to squash this PR when merging it.

justinr1234 (approved):

intelliot:
(I believe drlongle still needs to respond to gregtatcam's suggestions)

src/ripple/rpc/handlers/AccountCurrenciesHandler.cpp
```diff
@@ -46,19 +46,19 @@ doAccountCurrencies(RPC::JsonContext& context)
...
                                       : params[jss::ident].asString());
 
-    bool const bStrict =
-        params.isMember(jss::strict) && params[jss::strict].asBool();
-
     // Get info on account.
-    AccountID accountID;  // out param
-    if (auto jvAccepted = RPC::accountFromString(accountID, strIdent, bStrict))
-        return jvAccepted;
+    auto const accountID = parseBase58<AccountID>(strIdent);
```
gregtatcam, line 50: `AccountID` is not really optional. Please consider the updated suggested in `AccountChannels.cpp` comment. I will not repeat this comment for other files. All changed files under `rpc/handlers` have a similar code block: AccountInfo.cpp, AccountLines.cpp, AccountObjects.cpp, AccountOffers.cpp, DepositAuthorized.cpp, GatewayBalances.cpp, NoRipplecheck.cpp, OwnerInfo.cpp.

src/ripple/rpc/handlers/AccountChannels.cpp
```diff
@@ -179,9 +177,9 @@ doAccountChannels(RPC::JsonContext& context)
                 }
 
                 if (count <= limit && sleCur->getType() == ltPAYCHAN &&
-                    (*sleCur)[sfAccount] == accountID &&
-                    (!visitData.hasDst ||
-                     visitData.raDstAccount == (*sleCur)[sfDestination]))
+                    (*sleCur)[sfAccount] == *accountID &&
+                    (!visitData.raDstAccount ||
+                     *visitData.raDstAccount == (*sleCur)[sfDestination]))
```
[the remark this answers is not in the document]
gregtatcam: You are right. I'm taking back this suggestion.

src/ripple/rpc/handlers/AccountLines.cpp
```diff
@@ -227,8 +218,8 @@ doAccountLines(RPC::JsonContext& context)
                             RPCTrustLine::makeItem(visitData.accountID, sleCur);
 
                         if (line &&
-                            (!visitData.hasPeer ||
-                             visitData.raPeerAccount ==
+                            (!visitData.raPeerAccount ||
+                             *visitData.raPeerAccount ==
```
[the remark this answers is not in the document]
gregtatcam: You are right. I'm taking back this suggestion.

src/ripple/rpc/handlers/GatewayBalances.cpp
```diff
@@ -90,17 +86,7 @@ doGatewayBalances(RPC::JsonContext& context)
...
-                if (pk)
-                {
-                    hotWallets.insert(calcAccountID(*pk));
-                    return true;
-                }
-
-                auto const id = parseBase58<AccountID>(j.asString());
-
-                if (id)
+                if (auto const id = parseBase58<AccountID>(j.asString()); id)
```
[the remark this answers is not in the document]
gregtatcam: The original code is:
```
auto const id = parseBase58<AccountID>(j.asString());

                if (id)
                {
                    hotWallets.insert(*id);
                    return true;
                }
```
In this case it does need to be checked. In case of `if` with initializer it's checked if there is no other condition.

src/ripple/rpc/handlers/AccountCurrenciesHandler.cpp
drlongle: I have applied your suggested changes across the board. Let me know if you find any further issues.

gregtatcam (approved):
👍 LGTM

PR #4406, opened 2023-02-01
`fixNFTokenRemint`: prevent NFT re-mint:

shawnxie999:
Currently, a NFT ID can be reminted by the following steps:

1. Alice creates an account and mints an NFT
2. Alice burns the NFT (a NFTokenBurn transaction)
3. Alice deletes her account (a AccountDelete transaction)
4. Alice re-creates her account
5. Alice mints a NFT ( a NFTokenMint transaction with params: NFTokenTaxon = 0, Flags = 9). This will re-mint a NFT with the same NFTokenID as Step 1. The params that construct the NFT ID will cause collision if their values are equal before and after the remint

We want to come up with a new sequence number construct to avoid this scenario

### New NFT sequence construct

- We create a new AccountRoot field `FirstNFTSequence`, which stays constant over time.This field is set to the current account sequence when the account issues their first NFT. Otherwise, it is not set. 

- The sequence of a newly minted NFT is computed by `FirstNFTSequence` + `MintedNFTokens` (then `MintedNFTokens` increments by 1).
### New Account Deletion Restriction
An account can only be deleted if `FirstNFTSequence` + `MintedNFTokens` + `256`  is less than the current ledger sequence (256 was chosen as a heuristic restriction for account deletion and already exists in the account deletion constraint). Without this restriction, NFT is still remintable with the following scenario:

1. Alice's account sequence is at 1
2. Bob is Alice's authorized minter
3. Bob mints 500 NFTs for Alice (NFTs will have sequences 1 - 501, since NFT sequence is computed by FirstNFTokenSequence + MintedNFTokens)
4. Alice deletes her account at ledger 257 (enforced by AccountDelete amendment)
5. Alice re-creates her account at ledger 258
6. Alice mints a NFT. FirstNFTokenSequence initialized to her account sequence (258), and MintedNFTokens to 0. This newly minted NFT would have a sequence number of 258, which is a duplicate of what she issued through authorized minting before she deleted her account.

### Type of Change

- [ ] Documentation Updates
- [ ] Release


nbougalis (commented):
Personally I don't see a problem with this kind or re-minting (burn NFT, delete issuing account, re-create issuing account, re-mint NFT) unless the issue you're trying to solve is leftover offers from previous operations. If that's the concern, then the fix is easy: check for the presence of those during minting, and if present, remove them and if you can't fail the mint.

Even if that's not the concern and there is some other issue, I don't think what this PR does is the right approach anyways.

I am not vetoing this. Just offering my 2¢.

shawnxie999:
> Personally I don't see a problem with this kind or re-minting (burn NFT, delete issuing account, re-create issuing account, re-mint NFT) unless the issue you're trying to solve is leftover offers from previous operations. If that's the concern, then the fix is easy: check for the presence of those during minting, and if present, remove them and if you can't fail the mint.
> 
> Even if that's not the concern and there is some other issue, I don't think what this PR does is the right approach anyways.
> 
> I am not vetoing this. Just offering my 2¢.

This doesn't have to do with the the previous change of leftover offers. It was a concern that duplicate NFT's could cause issues in the long run

Issues brought up by @ledhed2222:

> A malicious issuer can mint a token representing, for example, a carbon credit. In the expected workflow, this token is meant to be burned once the carbon has actually been sequestered. As a result, the issuer has set the tfBurnable flag to allow them to burn the token even if they no longer hold it. They then burn said token, indicating that the specific carbon credit is no longer tradable. At this point, the issuer deletes and recreates their account, mints a new token representing the same carbon credit, and sells it again.

shawnxie999:
> Even if that's not the concern and there is some other issue, I don't think what this PR does is the right approach anyways.

Could you please elaborate why this wouldn't be the right approach? The change has minimal effect on the ledger and does not change how the people use NFTs.

intelliot:
(awaiting review by @ledhed2222 )

nbougalis:
> This doesn't directly have to do with the the previous change of leftover offers. It was a concern that duplicate NFT's could cause issues in the long run
> 
> Something brought up by @ledhed2222:
> 
> > A malicious issuer can mint a token representing, for example, a carbon credit. In the expected workflow, this token is meant to be burned once the carbon has actually been sequestered. As a result, the issuer has set the tfBurnable flag to allow them to burn the token even if they no longer hold it. They then burn said token, indicating that the specific carbon credit is no longer tradable. At this point, the issuer deletes and recreates their account, mints a new token representing the same carbon credit, and sells it again.

I don't think that the ability to "remint" an NFT after an account has been deleted and re-created is a huge problem. It just seems like a far-fetched "one-off" attack, because it leaves behind **indelible cryptographic proof of malfeasance**.

The particular approach adopted here doesn't make much sense to me but if you're so inclined of fixing this, then I think that a cleaner approach would be to simply "restructure" the NFT ID composition by changing the taxon to be the 32-bit ledger sequence:

[16 bit flags (as is)][16 bit transfer fee (as is)][160 bit issuer id (as is)][32 bit ledger sequence (replaces taxon)][32 bit sequence (as is)]

All you need is to make the "Taxon" field optional in the mint transaction, along with a flag `tfAutomaticTaxon`. This avoids the need for another field in the account root.

shawnxie999:
> [16 bit flags (as is)][16 bit transfer fee (as is)][160 bit issuer id (as is)][32 bit ledger sequence (replaces taxon)][32 bit sequence (as is)]
> 
> All you need is to make the "Taxon" field optional in the mint transaction, along with a flag `tfAutomaticTaxon`. This avoids the need for another field in the account root.

Just trying to wrap my head around this, how would this prevent the remint scenario? 

Even with the introducing a new flag, I believe it introduces unnecessary complexity for something that user wouldn't care about. And i believe the level of effort of higher because it would requires quit a bit of change in both rippled and clio (_i think_)

nbougalis:
As I said, I don't think this is a real issue so, if it were up to me, I wouldn't fix it at all. With that said...

> Just trying to wrap my head around this, how would this prevent the remint scenario?

The code prevents accounts from being deleted if their sequence number exceeds the sequence number of the current ledger plus some buffer (I think 256?). Since an account that's "recreated" after being deleted has an initial sequence number equal to the ledger in which it was created, the remint becomes impossible, since the taxon of the new NFT would be greater than the taxon of any previously minted NFT.

Of course, this doesn't _entirely_ prevent the remint scenario, unless you _require_ the automatic taxon functionality, which does suck.

> Even with introducing a new flag, I believe it introduces unnecessary complexity for something that user wouldn't care about. And i believe the level of effort is higher because it would require quite a bit of change in both rippled and clio (_i think_)

I'm curious, if this seems like a serious concern then why not simply introduce a basically one-line amendment that checks if `sfMintedNFTokens` is present in the `AccountRoot` and if it is, then simply don't allow the deletion?

That would make accounts which issue NFTs permanent but that seems like a very straightforward restrictions that's easy to document and understand and it's trivial to code and verify as correct.

intelliot:
> I'm curious, if this seems like a serious concern then why not simply introduce a basically one-line amendment that checks if `sfMintedNFTokens` is present in the `AccountRoot` and if it is, then simply don't allow the deletion?

This was considered and discussed;
- It still requires an amendment, so it's not significantly less impactful.
- It prevents users from getting their reserves back via account deletion.

That said, I believe we are still open to the option. However, the alternative approach offered by this PR is:
1. Already implemented, with tests. It only needs review, and integration testing. (perf testing?)
2. Superior to the "simple" solution because users can still delete their accounts.

src/ripple/app/tx/impl/NFTokenMint.cpp
```diff
@@ -160,14 +160,48 @@ NFTokenMint::doApply()
...
-        std::uint32_t const tokenSeq = (*root)[~sfMintedNFTokens].value_or(0);
+        std::uint32_t tokenSeq;
+        if (ctx_.view().rules().enabled(fixUnburnableNFToken))
         {
-            std::uint32_t const nextTokenSeq = tokenSeq + 1;
-            if (nextTokenSeq < tokenSeq)
-                return Unexpected(tecMAX_SEQUENCE_REACHED);
-
-            (*root)[sfMintedNFTokens] = nextTokenSeq;
+            auto accSeq = (*root)[sfSequence];
```
ledhed2222, line 166: should be const

src/ripple/app/tx/impl/NFTokenMint.cpp
```diff
@@ -160,14 +160,48 @@ NFTokenMint::doApply()
...
+                        ? accSeq
+                        : accSeq - 1);
+
+            // Get the unique sequence number of this token by
+            // sfFirstNFTokenSequence + sfMintedNFTokens
+            tokenSeq = firstNFTokenSeq + (*root)[~sfMintedNFTokens].value_or(0);
+            {
+                std::uint32_t const nextTokenSeq = tokenSeq + 1;
+                if (nextTokenSeq < tokenSeq)
+                    return Unexpected(tecMAX_SEQUENCE_REACHED);
```
ledhed2222, line 187: so you're just checking for int overflow based on ledger seq? that seems unnecessary

src/ripple/app/tx/impl/NFTokenMint.cpp
```diff
@@ -160,14 +160,48 @@ NFTokenMint::doApply()
...
+
+                (*root)[sfMintedNFTokens] =
+                    (*root)[~sfMintedNFTokens].value_or(0) + 1;
+                (*root)[sfFirstNFTokenSequence] = firstNFTokenSeq;
+            }
```
ledhed2222, line 192: i don't see why you need a new context with `{}` for this block. can you remove?

src/ripple/protocol/impl/LedgerFormats.cpp
```diff
@@ -55,6 +55,7 @@ LedgerFormats::LedgerFormats()
             {sfNFTokenMinter,        soeOPTIONAL},
             {sfMintedNFTokens,       soeDEFAULT},
             {sfBurnedNFTokens,       soeDEFAULT},
+            {sfFirstNFTokenSequence, soeOPTIONAL},
```
ledhed2222, line 58: note - this will require change to the SDKs I believe

shawnxie999:
Putting this into draft due to conflicts and also some refactoring needed

src/ripple/protocol/impl/SField.cpp
```diff
@@ -150,6 +150,7 @@ CONSTRUCT_TYPED_SFIELD(sfMintedNFTokens,        "MintedNFTokens",       UINT32,
 CONSTRUCT_TYPED_SFIELD(sfBurnedNFTokens,        "BurnedNFTokens",       UINT32,    44);
 CONSTRUCT_TYPED_SFIELD(sfHookStateCount,        "HookStateCount",       UINT32,    45);
 CONSTRUCT_TYPED_SFIELD(sfEmitGeneration,        "EmitGeneration",       UINT32,    46);
+CONSTRUCT_TYPED_SFIELD(sfFirstNFTokenSequence,  "FirstNFTokenSequence", UINT32,    47);
```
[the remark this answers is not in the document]
shawnxie999: It seems three new 32 bit integer SFields are introduced for AMM and Hooks. I will reserve three field values and set mine to 50

shawnxie999:
This is ready for review

shawnxie999:
@nbougalis do you still think this approach would be feasible? I think the option you suggested with adding optional flag is would still be vulnerable to potential misuses. 

One point discussed is that, currently, there has not been a single scenario where someone re-mints their NFT by recreating their account (observed by Clio, @ledhed2222 correct me if I'm wrong). So it's safe to make the assumption that users expect the uniqueness of their NFTokenIDs to be the correct behavior.  

We want to avoid the situation where someone stumbles into the re-minting situation in the future, and learn that they can mint duplicate NFTokenIDs (because I'm sure most people don't know). So our idea is to fix this up on XRPL to align with users' expectations of NFT, which is to make NFTokenIDs truly unique and non-remintable.

src/ripple/app/tx/impl/DeleteAccount.cpp
```diff
@@ -214,6 +214,23 @@ DeleteAccount::preclaim(PreclaimContext const& ctx)
...
+    // authorized minting is involved. Because when the minter mints a NFToken,
+    // the issuer's sequence does not change. So when the issuer re-creates
+    // their account and mints a NFToken, it is possible that the
+    // NFTokenSequence of this NFToken is the same as the one that the
+    // authorized minter minted in a previous ledger.
+    if (ctx.view.rules().enabled(fixNFTokenRemint) &&
+        ((*sleAccount)[~sfFirstNFTokenSequence].value_or(0) +
+             (*sleAccount)[~sfMintedNFTokens].value_or(0) + seqDelta >
+         ctx.view.seq()))
+        return tecTOO_SOON;
```
nbougalis, line 232: Imagine someone issuing a million "Silly Billy" NFTs, with the first minted in ledger 77,777,777.

This check will prevent the issuer from deleting the account until, at the earliest, ledger 78,778,032; that's not gonna happen for around 4 million seconds, or around a month and a half.

src/ripple/protocol/impl/SField.cpp
```diff
@@ -150,6 +150,9 @@ CONSTRUCT_TYPED_SFIELD(sfMintedNFTokens,        "MintedNFTokens",       UINT32,
 CONSTRUCT_TYPED_SFIELD(sfBurnedNFTokens,        "BurnedNFTokens",       UINT32,    44);
 CONSTRUCT_TYPED_SFIELD(sfHookStateCount,        "HookStateCount",       UINT32,    45);
 CONSTRUCT_TYPED_SFIELD(sfEmitGeneration,        "EmitGeneration",       UINT32,    46);
+// Three field values of 47, 48 and 49 are reserved for 
+// LockCount(Hooks), VoteWeight(AMM), DiscountedFee(AMM)
+CONSTRUCT_TYPED_SFIELD(sfFirstNFTokenSequence,  "FirstNFTokenSequence", UINT32,    50);
```
nbougalis, line 155: Ensure that there's no conflict with hooks or other amendments here. Tag @RichardAH. 

Having a registry for this kind of thing could be nice.

nbougalis (approved):
To be clear, I don't really think this "attack" is a problem. It's something only the issuer can do, and only by engaging in what is obvious and detectable behavior: deleting/burning all his existing issued assets, then deleting the account, before recreating the account. And then, he still couldn't remint a particular NFT without first reminting all previous NFTs.

And if he does remint, so what? It's trivially verifiable that it's a _new_ and _distinct_ NFT from any previous ones.

With that said, there is an implicit promise that a given NFT ID is unique and can never be regenerated. Your solution enforces that promise.

I don't like the fact that your solution bloats the `ACCOUNT_ROOT` by 4 bytes for every NFT issuer, but you can't get around that without making "large" client-breaking changes.

I think this PR is fine as is from a technical standpoint.

src/ripple/app/tx/impl/DeleteAccount.cpp
shawnxie999: Yes, unfortunately this check needs to be in. Without it, the scenario below produces duplicate NFTID:
1. Alice's account sequence is at 1 
2. Bob is Alice's authorized minter
3. Bob mints 500 NFTs for Alice (NFTs will have sequences 1 - 501, since NFT sequence is computed by `FirstNFTokenSequence` + `MintedNFTokens`)
4. Alice deletes her account at ledger 257 (enforced by AccountDelete amendment)
5. Alice re-creates her account at ledger 258
6. Alice mints a NFT. `FirstNFTokenSequence` initialized to her account sequence (258), and MintedNFTokens to 0. This newly minted NFT would have a sequence number of 258, which is a duplicate of what she issued through authorized minting.
7.

nbougalis: I understand. My point was that this kind of test means it's hard for someone to know _when_ their account can be deleted and, depending on the number of NFTs they issue. For some accounts (those which don't have `FirstNFTokenSequence` set) the answer is "immediately" and for others it's "well, sometime in the future."

shawnxie999: But I don't think this will be a common issue, since this restriction is only applicable for authorized minting situations. 

If the issuer has never issued an NFT through authorized minting, then, `FirstNFTokenSequence` + `MintedNFTokens` <= `AccountRoot Sequence` will always be true, which means the AccountDelete amendment (ledger > Account Sequence + 256) already contains the new restriction.

src/ripple/protocol/impl/SField.cpp
shawnxie999: @RichardAH @gregtatcam 

 I reserved three values of the 32-bit field for what it looks like three new `SFields` in both AMM and Hook. Just want to make sure the number of reserved field value is correct and would be no further changes

scottschurr (commented):
First, nice tests!  You exercised the cases that you were interested in and showed the behavior changes.  Good work!

The general flow of this pull request looks right.  I left a number of comments.  Some of my notes are just comment nits or naming issues.  A few are structural.  A few are me being cautious around potential undefined behavior.  I'm happy to discuss any of them further if you have concerns.

Looks pretty good!

src/ripple/app/tx/impl/NFTokenMint.cpp
```diff
@@ -160,15 +160,60 @@ NFTokenMint::doApply()
...
+            (*root)[sfFirstNFTokenSequence] =
+                (*root)[~sfNFTokenMinter] == ctx_.tx[sfAccount] ||
+                    ctx_.tx.getSeqProxy().isTicket()
+                ? accSeq
+                : accSeq - 1;
         }
+
+        auto const mintedNftCnt = (*root)[~sfMintedNFTokens].value_or(0);
+
+        (*root)[sfMintedNFTokens] = mintedNftCnt + 1;
```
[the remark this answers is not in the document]
shawnxie999: just curious what benefit would i get from explicitly write it as an unsigned int?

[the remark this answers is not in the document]
scottschurr: C++ does implicit type conversions between integers.  That's something left over fro C.  If we know that `mintedNftCnt` is unsigned, then adding another unsigned number will not cause the result of the addition to become signed.

I don't have the integer conversion rules committed to memory; they are too complicated.  Instead I just try to be very conservative when I'm dealing with integers that might overflow.  If integer overflow might happen then everything possible needs to be done to make sure everything is unsigned.  At least that's the way I approach it.

src/ripple/app/tx/impl/NFTokenMint.cpp
```diff
@@ -160,15 +160,60 @@ NFTokenMint::doApply()
...
 
-            (*root)[sfMintedNFTokens] = nextTokenSeq;
+            (*root)[sfFirstNFTokenSequence] =
+                (*root)[~sfNFTokenMinter] == ctx_.tx[sfAccount] ||
+                    ctx_.tx.getSeqProxy().isTicket()
+                ? accSeq
+                : accSeq - 1;
         }
+
+        auto const mintedNftCnt = (*root)[~sfMintedNFTokens].value_or(0);
```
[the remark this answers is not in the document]
shawnxie999: Thanks! good to learn this!

shawnxie999:
this is ready for re-review. cc @scottschurr

src/ripple/app/tx/impl/NFTokenMint.cpp
```diff
@@ -160,15 +160,60 @@ NFTokenMint::doApply()
...
+        {
+            auto const accSeq = (*root)[sfSequence];
 
-            (*root)[sfMintedNFTokens] = nextTokenSeq;
+            (*root)[sfFirstNFTokenSequence] =
+                (*root)[~sfNFTokenMinter] == ctx_.tx[sfAccount] ||
+                    ctx_.tx.getSeqProxy().isTicket()
+                ? accSeq
+                : accSeq - 1;
         }
```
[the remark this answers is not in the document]
shawnxie999: Thank you!

src/test/app/NFToken_test.cpp
```diff
@@ -6038,6 +6065,505 @@ class NFToken_test : public beast::unit_test::suite
...
+            // Fund alice to re-create her account
+            env.fund(XRP(10000), alice);
+            env.close();
+
+            // alice's account now exists and has minted 0 NFTokens
+            BEAST_EXPECT(env.closed()->exists(aliceAcctKey));
+            BEAST_EXPECT(env.current()->exists(aliceAcctKey));
+            BEAST_EXPECT((*env.le(alice))[sfMintedNFTokens] == 0);
+
+            // alice mints a NFT with same params as prevNFTokenID
```
scottschurr, line 6550: Nit: there is no `prevNFTokenID` in this scope.  Consider something like:
```
                // The new NFT minted will not have the same ID
                // as any of the NFTs authorized minter minted using tickets
```
Sorry I missed that last time.

scottschurr (approved):
Looks great.  Thanks for accommodating my requests.  I noticed one minor comment problem in a test.  Sorry I missed that last time.  But it doesn't need to hold up this pull request.

scottschurr (approved):
:+1: Looks great!

src/ripple/app/tx/impl/NFTokenMint.cpp
```diff
@@ -160,15 +160,66 @@ NFTokenMint::doApply()
...
+        // There are three situations:
+        //  o If the first token is being minted by the issuer and
+        //     * If the transaction consumes a Sequence number, then the
+        //       Sequence has been pre-incremented by the time we get here in
+        //       doApply.  We must decrement the value in the Sequence field.
+        //     * Otherwise the transaction uses a Ticket so the Sequence has
+        //       not been pre-incremented.  We use the Sequence value as is.
+        //  o The first token is being minted by an authorized minter.  In
+        //    this case the issuer's Sequence field has been left untouched.
+        //    We use the issuer's Sequence value as is.
```
ledhed2222, line 195: great comments

src/ripple/app/tx/impl/NFTokenMint.cpp
```diff
@@ -160,15 +160,66 @@ NFTokenMint::doApply()
...
+                    ctx_.tx.getSeqProxy().isTicket()
+                ? acctSeq
+                : acctSeq - 1;
         }
+
+        std::uint32_t const mintedNftCnt =
+            (*root)[~sfMintedNFTokens].value_or(0u);
+
+        (*root)[sfMintedNFTokens] = mintedNftCnt + 1u;
+        if ((*root)[sfMintedNFTokens] == 0u)
```
ledhed2222, line 211: doesn't need to change - but i don't think you need the `u` specifier since you've already specified that the type is unsigned and arithmetic with a mix of signed/unsigned values should leave the value unsigned in this case.

src/ripple/protocol/impl/LedgerFormats.cpp
ledhed2222: would you confirm this? doesn't block this PR but when a TX modifies this value and one of the libraries is trying to parse the metadata from that TX I think it will raise an exception

ledhed2222 (approved):

src/ripple/app/tx/impl/NFTokenMint.cpp
shawnxie999: it is a conservative approach to make sure things are correctly typed proposed by @scottschurr

scottschurr: @ledhed2222, what you describe is not always the case.  There are situations where operations on a combination of unsigned and signed integers produce a signed value.  See "Usual Arithmetic Conversions": https://en.cppreference.com/w/cpp/language/usual_arithmetic_conversions#Integer_conversion_rank

You'll notice that the rules changed with C++11 and again with C++23.

What the rules seem to be _pretty_ good at holding onto is that if both arguments are unsigned then the result will be unsigned.  Since the literals `0` and `1` are implicitly signed, they can influence the type yielded by the operation.  That's why, since overflow is possible in this calculation, using `0u` and `1u` is safer; they are explicitly unsigned.

I did not go to the effort of determining exactly what conversions the compiler would make -- I'm not that smart.  I just applied a heuristic so I'm increasing the likelihood the compiler will do what I want.

scottschurr (approved):
Still LGTM.  :+1:

intelliot:
At merging, this PR will be squashed into 1 commit. The commit needs a commit message. Here is what I suggest. @shawnxie999 or anyone else, please feel free to make suggestions or offer an alternative.

```
`fixNFTokenRemint`: prevent NFT re-mint: (#4406)

Without the protocol amendment introduced by this commit, an NFT ID can
be reminted in this manner:

1. Alice creates an account and mints an NFT.
2. Alice burns the NFT with an `NFTokenBurn` transaction.
3. Alice deletes her account with an `AccountDelete` transaction.
4. Alice re-creates her account.
5. Alice mints an NFT with an `NFTokenMint` transaction with params:
   `NFTokenTaxon` = 0, `Flags` = 9).

This will mint a NFT with the same `NFTokenID` as the one minted in step
1. The params that construct the NFT ID will cause a collision in
`NFTokenID` if their values are equal before and after the remint.

With the `fixNFTokenRemint` amendment, there is a new sequence number
construct which avoids this scenario:

- A new `AccountRoot` field, `FirstNFTSequence`, stays constant over
  time.
  - This field is set to the current account sequence when the account
    issues their first NFT.
  - Otherwise, it is not set.
- The sequence of a newly-minted NFT is computed by: `FirstNFTSequence +
  MintedNFTokens`.
  - `MintedNFTokens` is then incremented by 1 for each mint.

Furthermore, there is a new account deletion restriction:

- An account can only be deleted if `FirstNFTSequence + MintedNFTokens +
  256` is less than the current ledger sequence.
  - 256 was chosen because it already exists in the current account
    deletion constraint.

Without this restriction, an NFT may still be remintable. Example
scenario:

1. Alice's account sequence is at 1.
2. Bob is Alice's authorized minter.
3. Bob mints 500 NFTs for Alice. The NFTs will have sequences 1-501, as
   NFT sequence is computed by `FirstNFTokenSequence + MintedNFTokens`).
4. Alice deletes her account at ledger 257 (as required by the existing
   `AccountDelete` amendment).
5. Alice re-creates her account at ledger 258.
6. Alice mints an NFT. `FirstNFTokenSequence` initializes to her account
   sequence (258), and `MintedNFTokens` initializes as 0. This
   newly-minted NFT would have a sequence number of 258, which is a
   duplicate of what she issued through authorized minting before she
   deleted her account.

---------

Signed-off-by: Shawn Xie <shawnxie920@gmail.com>
Co-authored-by: ledhed2222 <ledhed2222@users.noreply.github.com>
Co-authored-by: Scott Schurr <scott@ripple.com>
Co-authored-by: Denis Angell <dangell@transia.co>
```

This PR will be merged to `develop` on Monday, March 20, 2023 no earlier than 11:30am PT, unless I hear any opposition before then.

shawnxie999:
@intelliot looks good

PR #4409, opened 2023-02-03
Update documented pathfinding configuration defaults:

ximinez:

https://github.com/XRPLF/rippled/pull/4061/commits/417cfc2fb049dada484d196225cbfe49e54ad411 changed the default Path Finding configuration values, but missed updating the values documented in rippled-example.cfg.

This change updates those defaults, and adds recommended values for node which want to support "advanced" pathfinding, which is effectively the old default behavior.

### Type of Change

- [X ] Documentation Updates


intelliot (approved):

cfg/rippled-example.cfg
```diff
@@ -758,7 +758,9 @@
 #   When searching for paths, the default search aggressiveness. This can take
 #   exponentially more resources as the size is increased.
 #
-#   The default is: 7
+#   The recommended value to support advanced pathfinding is: 7
```
ledhed2222, line 761: why though?

cfg/rippled-example.cfg
```diff
@@ -767,12 +769,19 @@
 #   If you do not need pathfinding, you can set path_search_max to zero to
 #   disable it and avoid some expensive bookkeeping.
 #
-#   The default for 'path_search_fast' is 2. The default for 'path_search_max' is 10.
+#   To support advanced pathfinding the recommended value for
+#   'path_search_fast' is 2, and for 'path_search_max' is 10.
```
ledhed2222, line 773: why?

cfg/rippled-example.cfg
```diff
@@ -767,12 +769,19 @@
...
+#
+#   The default for 'path_search_fast' is 2. The default for 'path_search_max' is 3.
 #
 # [path_search_old]
 #
 #   For clients that use the legacy path finding interfaces, the search
-#   aggressiveness to use. The default is 7.
+#   aggressiveness to use.
+#
+#   The recommended value to support advanced pathfinding is: 7.
```
ledhed2222, line 782: why?

ledhed2222 (approved):
approved for speed - but it would be nice to explain _why_ the specified values are recommended.

cfg/rippled-example.cfg
intelliot: I'm not aware of anything special about `7` other than the fact that it worked well for many years, and is a balance between being unable to find an acceptable path vs. resource consumption. @ximinez @JoelKatz any thoughts to share here?

ximinez: Yep. That's pretty much it. I kind of intentionally left it vague because anybody who doesn't specifically want pathfinding will be just fine with the defaults, and anyone who _does_ probably just wants a reasonable recommendation. If they are more advanced than that, then any "tl;dr" summary in the config file is going to be insufficient - those users should dig into the code or other documentation.

PR #4424, opened 2023-02-16
Refactor getTrustedForLedger()

levinwinter:
The `getTrustedForLedger()` method has been refactored to look for validations associated with a specific ledger ID and sequence number.

### Context of Change
From now on, the method can be used to filter for specific sequence numbers which was adapted throughout the codebase.

### Type of Change

### DIFF
--- a/src/ripple/app/consensus/RCLConsensus.cpp
+++ b/src/ripple/app/consensus/RCLConsensus.cpp
@@ -337,7 +337,7 @@ RCLConsensus::Adaptor::onClose(
             // pseudo-transactions
             auto validations = app_.validators().negativeUNLFilter(
                 app_.getValidations().getTrustedForLedger(
-                    prevLedger->info().parentHash));
+                    prevLedger->info().parentHash, prevLedger->seq() - 1));
             if (validations.size() >= app_.validators().quorum())
             {
                 feeVote_->doVoting(prevLedger, validations, initialSet);
--- a/src/ripple/app/ledger/impl/LedgerMaster.cpp
+++ b/src/ripple/app/ledger/impl/LedgerMaster.cpp
@@ -341,7 +341,8 @@ LedgerMaster::setValidLedger(std::shared_ptr<Ledger const> const& l)
     if (!standalone_)
     {
         auto validations = app_.validators().negativeUNLFilter(
-            app_.getValidations().getTrustedForLedger(l->info().hash));
+            app_.getValidations().getTrustedForLedger(
+                l->info().hash, l->info().seq));
         times.reserve(validations.size());
         for (auto const& val : validations)
             times.push_back(val->getSignTime());
@@ -987,7 +988,7 @@ LedgerMaster::checkAccept(uint256 const& hash, std::uint32_t seq)
             return;
 
         auto validations = app_.validators().negativeUNLFilter(
-            app_.getValidations().getTrustedForLedger(hash));
+            app_.getValidations().getTrustedForLedger(hash, seq));
         valCount = validations.size();
         if (valCount >= app_.validators().quorum())
         {
@@ -1053,7 +1054,8 @@ LedgerMaster::checkAccept(std::shared_ptr<Ledger const> const& ledger)
 
     auto const minVal = getNeededValidations();
     auto validations = app_.validators().negativeUNLFilter(
-        app_.getValidations().getTrustedForLedger(ledger->info().hash));
+        app_.getValidations().getTrustedForLedger(
+            ledger->info().hash, ledger->info().seq));
     auto const tvc = validations.size();
     if (tvc < minVal)  // nothing we can do
     {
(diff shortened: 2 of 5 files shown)


seelabs (approved):
:+1: Code is clean and straight-forward. Thanks for the patch!

src/ripple/app/consensus/RCLConsensus.cpp
```diff
@@ -337,7 +337,7 @@ RCLConsensus::Adaptor::onClose(
             // pseudo-transactions
             auto validations = app_.validators().negativeUNLFilter(
                 app_.getValidations().getTrustedForLedger(
-                    prevLedger->info().parentHash));
+                    prevLedger->info().parentHash, prevLedger->seq() - 1));
```
mtrippled, line 340: A unit test should be made for this. Namely, what happens if this is fed a validation with a bogus sequence?

src/ripple/app/ledger/impl/LedgerMaster.cpp
```diff
@@ -341,7 +341,8 @@ LedgerMaster::setValidLedger(std::shared_ptr<Ledger const> const& l)
     if (!standalone_)
     {
         auto validations = app_.validators().negativeUNLFilter(
-            app_.getValidations().getTrustedForLedger(l->info().hash));
+            app_.getValidations().getTrustedForLedger(
```
mtrippled, line 344: Like comment above, a unit test for this is appropriate. Especially since this is the area with the bigger potential problem. What happens if a bogus sequence number ends up in a validation?

I'm not sure how many tests are needed, whether it's 1 per caller of getTrustedForLedger(), or if it's possible to have fewer tests to prove the point.

src/ripple/app/misc/NegativeUNLVote.cpp
```diff
@@ -198,7 +198,7 @@ NegativeUNLVote::buildScoreTable(
     for (int i = 0; i < FLAG_LEDGER_INTERVAL; ++i)
     {
         for (auto const& v : validations.getTrustedForLedger(
-                 ledgerAncestors[numAncestors - 1 - i]))
+                 ledgerAncestors[numAncestors - 1 - i], seq - 2 - i))
```
mtrippled, line 201: There should also be a Negative UNL test to see what it does with a bogus sequence in a validation, independent from the other tests.

mtrippled (changes requested):

mtrippled (approved):
:+1: Looks good to me! The testing looks sufficient.

pwang200 (approved):
Thanks for the patch! Looks great.

levinwinter:
Thanks @mtrippled! I think the existing unit tests already cover quite some ground.

src/ripple/app/ledger/impl/LedgerMaster.cpp
```diff
@@ -987,7 +988,7 @@ LedgerMaster::checkAccept(uint256 const& hash, std::uint32_t seq)
             return;
 
         auto validations = app_.validators().negativeUNLFilter(
```
nbougalis, line 990: Since you're here add a `const`?

src/ripple/app/ledger/impl/LedgerMaster.cpp
```diff
@@ -1053,7 +1054,8 @@ LedgerMaster::checkAccept(std::shared_ptr<Ledger const> const& ledger)
 
     auto const minVal = getNeededValidations();
     auto validations = app_.validators().negativeUNLFilter(
```
nbougalis, line 1056: And one more `const` here?

nbougalis (approved):
Looks fine to me. Left a few small comments, but none of them are blockers; you can address them at your discretion.

src/ripple/consensus/Validations.h
```diff
@@ -1049,10 +1049,11 @@ class Validations
     /**  Get trusted full validations for a specific ledger
 
          @param ledgerID The identifier of ledger of interest
+         @param seq The sequence number of ledger of interest
          @return Trusted validations associated with ledger
     */
     std::vector<WrappedValidationType>
-    getTrustedForLedger(ID const& ledgerID)
+    getTrustedForLedger(ID const& ledgerID, Seq const& seq)
```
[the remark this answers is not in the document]
intelliot: This is not needed, and if nothing needs the optional, then the API is cleaner without the optional. That said; it is fine either way. @levinwinter I'll leave it up to you; whatever you decide is perfectly fine.

intelliot:
@levinwinter regarding the comments above, I'm happy either way; I will keep this PR open for now so that you have a chance to address them if you wish. Please comment here to indicate when you think that this PR is ready to merge, and I'll merge it.

src/ripple/consensus/Validations.h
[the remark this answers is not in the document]
mtrippled: I cannot think of a circumstance where the sequence should not match the hash for data that comes from an untrusted source, so I don't think making this optional is a good idea.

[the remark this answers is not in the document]
levinwinter: I see that there are some concerns, so let's keep it as is for now. Currently, in all of the places `getTrustedForLedger()` is called, it is reasonable to add the sequence number.

levinwinter:
Thank you for the feedback, @nbougalis! Let's keep it as is for now because I think we should see how to refactor this and the surrounding logic more, which could then include your suggestions. As this is too much for this PR, I'd be happy if @intelliot could merge this.

PR #4428, opened 2023-02-22
Remove recipe for RocksDB and add recipe for Snappy

thejohnfreeman:
- Copies the recipe for Snappy from Conan Center, but removes three lines that explicitly link the standard library, which prevents builders from statically linking it.
- Removes the recipe for RocksDB now that an official recipe for version 6.27.3 is in Conan Center.

Developers will likely need to remove cached versions of both RocksDB and Snappy:

```
conan remove -f rocksdb
conan remove -f snappy
```


thejohnfreeman:
CI is not expected to pass until we change the workflows to use Conan <2.0.

intelliot:
note: the aforementioned PR to change the workflows to use Conan <2.0 is https://github.com/XRPLF/rippled/pull/4430

legleux (approved):
It would appear removing that libstdc++ line from the snappy recipe prevents the missing runtime dependency on Ubuntu and CentOS

intelliot:
Merged via https://github.com/XRPLF/rippled/pull/4431

PR #4429, opened 2023-02-22
Undo API changes introduced in XRPFees:

### LINKED ISSUE #4425: Make XRPFees a non-breaking API change
## Issue Description
https://github.com/XRPLF/rippled/pull/4247 introduces an amendment, `featureXRPFees` (`XRPFees` for short), and causes some breaking API changes.

The motivations for the amendment:
- The fee code is unnecessarily complex due to arcane concepts like "fee units" and "reference transactions" - represented as integers of various widths. A lot of effort is expended to keep all that straight and prevent developers from doing the wrong thing.
- Fee units are hard-coded to 10 and the "reference transaction" fee is also 10, so there is an opportunity to reduce complexity: replace "fee units" with `XRPAmount` for all fees. Scaling/adjusting of fees can operate directly on `XRPAmount` objects instead.

`XRPFees` converts all representations of fees in the code, ledger, and protocol into `XRPAmount`s.

Currently, in the [FeeSettings](https://xrpl.org/feesettings.html) singleton object, `BaseFee` is serialized as hexadecimal; `ReferenceFeeUnits`, `ReserveBase`, `ReserveIncrement` are serialized as integers (when they would make more sense as strings because they are XRP amounts).

With `XRPFees`, the replacement fields are named `BaseFeeDrops`, `ReserveBaseDrops`, and `ReserveIncrementDrops`, and they are all XRP amounts serialized as strings.

It appears that there are likely impacts to ~`server_info`,~ `SetFee` (transaction type) txs and [Subscription](https://xrpl.org/subscribe.html) messages for the `ledger` and `validations` stream.

## Expected Result
Ideally, gain the benefits of `XRPFees` without the downside of breaking existing API consumers + client libraries.

## Actual Result
Some field names and types have changed in the APIs, which would break existing integrations and client library versions.

1. Consider changing the API back to undo any breaking changes that #4247 can cause.
2. If there are case(s) where it really makes sense to break the API (i.e. it is impossible or infeasible to support existing clients), clearly document the change and offer a straightforward migration path for existing users.

- @ximinez thinks it may be worth changing the APIs to avoid breaking changes.

## Contexts
- Direct rippled API consumers (WebSocket, HTTP JSON-RPC)
- Client libraries (xrpl.js, xrpl-py, xrpl4j)

## Related
- https://github.com/XRPLF/xrpl.js/issues/2219

ximinez:

#4247 made some unnecessary breaking changes to the API for the validation subscription stream. This PR reverts those changes.

* Original changes: https://github.com/XRPLF/rippled/commit/e4b17d1cf2c43e33128dd70b6036fe10f38b0c0d
* Resolves #4425

### Type of Change

### DIFF
--- a/src/test/rpc/Subscribe_test.cpp
+++ b/src/test/rpc/Subscribe_test.cpp
@@ -412,24 +412,10 @@ public:
                     return false;
 
                 bool xrpFees = env.closed()->rules().enabled(featureXRPFees);
-                if ((!xrpFees &&
-                     jv.isMember(jss::reserve_base) != isFlagLedger) ||
-                    (xrpFees && jv.isMember(jss::reserve_base)))
+                if (jv.isMember(jss::reserve_base) != isFlagLedger)
                     return false;
 
-                if ((!xrpFees &&
-                     jv.isMember(jss::reserve_inc) != isFlagLedger) ||
-                    (xrpFees && jv.isMember(jss::reserve_inc)))
-                    return false;
-
-                if ((xrpFees &&
-                     jv.isMember(jss::reserve_base_drops) != isFlagLedger) ||
-                    (!xrpFees && jv.isMember(jss::reserve_base_drops)))
-                    return false;
-
-                if ((xrpFees &&
-                     jv.isMember(jss::reserve_inc_drops) != isFlagLedger) ||
-                    (!xrpFees && jv.isMember(jss::reserve_inc_drops)))
+                if (jv.isMember(jss::reserve_inc) != isFlagLedger)
                     return false;
 
                 return true;
--- a/src/ripple/app/misc/NetworkOPs.cpp
+++ b/src/ripple/app/misc/NetworkOPs.cpp
@@ -2184,16 +2184,15 @@ NetworkOPsImp::pubValidation(std::shared_ptr<STValidation> const& val)
         //  simplifies later operations)
         if (auto const baseFeeXRP = ~val->at(~sfBaseFeeDrops);
             baseFeeXRP && baseFeeXRP->native())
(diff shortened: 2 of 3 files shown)


ckniffen:
I would prefer keeping the new fields and the old.  We can then transition over to the new fields after it passes and deprecate the old fields.

src/test/rpc/Subscribe_test.cpp
```diff
@@ -412,24 +412,10 @@ class Subscribe_test : public beast::unit_test::suite
                     return false;
 
                 bool xrpFees = env.closed()->rules().enabled(featureXRPFees);
```
ckniffen, line 414: This variable is no longer used.

ckniffen:
It would also be nice to see these new fields added to `server_info`.  I erroneously thought they were in my initial analysis.

ximinez:
> I would prefer keeping the new fields and the old. We can then transition over to the new fields after it passes and deprecate the old fields.

I disagree. I don't think adding the new fields was a good idea in the first place. I honestly can't remember why I did it in the first place. The values should be able to be changed in place without any negative effects (AFAIK).

ckniffen:
I thought they were changed to normalize all fees to drops.  No worries though.

ckniffen (approved):

thejohnfreeman (approved):

HowardHinnant (approved):

intelliot:
@ximinez at your convenience, please confirm when this is ready to merge. You can put the `Passed` label on it.

intelliot:
- note: see https://github.com/XRPLF/rippled/pull/4741 for details about an API change that takes effect if XRPFees activates

PR #4437, opened 2023-02-25
Verify peer's public key earlier:

intelliot:

Fix an issue introduced in 1.10.0-b1.

No stable versions of rippled (including 1.9.4) are affected.

### Context of Change

An issue was introduced in #4195 / 5a15229 (part of 1.10.0-b1) and is fixed here.

### Type of Change

<hr>

```
Acknowledgements:
Aaron Hook for responsibly disclosing this issue.

We welcome reviews of the rippled code and urge researchers to

    bugs@xrpl.org
```

### DIFF
--- a/src/ripple/overlay/impl/Handshake.cpp
+++ b/src/ripple/overlay/impl/Handshake.cpp
@@ -301,6 +301,24 @@ verifyHandshake(
         throw std::runtime_error("Bad node public key");
     }();
 
+    // This check gets two birds with one stone:
+    //
+    // 1) it verifies that the node we are talking to has access to the
+    //    private key corresponding to the public node identity it claims.
+    // 2) it verifies that our SSL session is end-to-end with that node
+    //    and not through a proxy that establishes two separate sessions.
+    {
+        auto const iter = headers.find("Session-Signature");
+
+        if (iter == headers.end())
+            throw std::runtime_error("No session signature specified");
+
+        auto sig = base64_decode(iter->value().to_string());
+
+        if (!verifyDigest(publicKey, sharedValue, makeSlice(sig), false))
+            throw std::runtime_error("Failed to verify session");
+    }
+
     if (publicKey == app.nodeIdentity().first)
     {
         auto const peerInstanceID = [&headers]() {
@@ -329,24 +347,6 @@ verifyHandshake(
         }
 
         throw std::runtime_error("Self connection");
-    }
-
-    // This check gets two birds with one stone:
-    //
-    // 1) it verifies that the node we are talking to has access to the
-    //    private key corresponding to the public node identity it claims.
-    // 2) it verifies that our SSL session is end-to-end with that node
-    //    and not through a proxy that establishes two separate sessions.
-    {
-        auto const iter = headers.find("Session-Signature");
-
-        if (iter == headers.end())
-            throw std::runtime_error("No session signature specified");
-
-        auto sig = base64_decode(iter->value().to_string());
-
-        if (!verifyDigest(publicKey, sharedValue, makeSlice(sig), false))
-            throw std::runtime_error("Failed to verify session");
     }
 
     if (auto const iter = headers.find("Local-IP"); iter != headers.end())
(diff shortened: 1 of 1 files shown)


scottschurr (approved):
:+1: LGTM.

thejohnfreeman (approved):

nbougalis (approved):
Clean and simple. :+1: from me.

nbougalis (changes requested):
Actually, **veto**. This change is actually insufficient: the shared secret which is signed is identical, so an an attacker can just duplicate the `Session-Signature` field.

Under normal circumstances, this makes no sense: you'd just cause the connection to drop. But with the "accidental misconfiguration detector" code introduced with commit 5a15229eeb13b69c8adf1f653b88a8f8b9480546 there's now an incentive: by switching the direction of the attack, it would just cause servers that _connect to the attacker_ to shut down.

I am going to stack a `[FOLD]` commit on top of this; feel free to squash it. It will disable the duplicate detector, eliminating this attack vector and I will enhance PR #4395 to properly handle this case.

intelliot:
Created #4438 to avoid the distraction of the stale approvals above, and also to preserve history in this PR for easier auditability (e.g. you can easily confirm that #4438 is just this PR squashed, with no other edits).

Superseded by #4438.

nbougalis:
Hey @ximinez! Hope all is well. Thanks for the comment.

Unfortunately, the original fix is **NOT** sufficient.

### Some background
The underlying problem here is two-fold: (1) the `Instance-Cookie` isn't mixed into the value being signed; and (2) both endpoints sign the _same_ value.

By way of background, for those who aren't familiar with the details:
- The `Instance-Cookie` is a 64-bit value that every server randomly generates at startup.
- The `Session-Signature` field is produced by a node signing a 256-bit value `V` using its node private key.
- The value `V` which is generated from the SSL session such that **_both_ endpoints produce the same value**.
- Lastly `rippled` uses [RFC 6979 deterministic signatures](https://www.rfc-editor.org/rfc/rfc6979), by fixing the nonce needed to be deterministically derived from the message. That is, given a key `K` and a message `M`, the resulting signature `S` is "fixed".

### Now for the good stuff...
Consider two peers, _Alice_, a normal, well-behaved honest peer, and _Mallory_, a malicious peer.

The original fix prevents _Mallory_ from _initiating_ a connection to _Alice_ and causing her to shutdown because, per the protocol, _Mallory_ needs to send an HTTP Upgrade request with a valid `Session-Signature` and, since _Mallory_ does not possess _Alice_'s key, he cannot pretend to know _Alice_'s identity.  

But that's _NOT_ the only scenario. What if _Alice_ attempts to make an outgoing connection to _Mallory_? Now, per the protocol, _Alice_ sends an HTTP Upgrade request, which includes her `Instance-Cookie` _and_ (importantly) a `Session-Signature`.

And **_now_** _Mallory_ now has all the information needed to trick _Alice_ into thinking her node identity is shared: the `Session-Signature` that _Alice_ already sent. All he needs to do is produce a response with a different `Instance-Cookie` and the `Session-Signature` that _Alice_ has **already** produced.

When _Alice_ processes _Mallory_'s response what she will see a _valid signature_ and an `Instance-Cookie` that is different from her's. And _Alice_ will proceed to shut down. _Oops_. :cry:  

The solution (which I will implement and submit as a fix for #4395) is to mix the `Instance-Cookie` value into the value being signed, along with an extra "direction" indicator and use this new hardened value for the new `Session-EKM-Signature` checks introduced in #4395.

### Wait, isn't this scheme already broken?
"It sounds like the existing `Session-Signature` mechanism is broken already: _Mallory_ can just spoof the `Session-Signature` as described above right now, regardless of this fix." Yes, he can _blindly_ play back a `Session-Signature` field to the server that sent it, but that's fine. If he does so, all he'll achieve is to cause the connection to drop since it's detected as a "self-connection". He's not hurting anyone other than himself.

ximinez:
Hi @nbougalis! That all makes perfect sense, including the future fix. I had completely forgotten how the handshake works. 

That said, instead of rolling the original fix back entirely, what if we just remove the call to stop the server, but keep logging that different message? That can at least alert operators who check their logs that they have a potential problem. Then once we have a stronger handshake protocol, we can start doing the automatic shutdown...

PR #4438, opened 2023-02-25
Disable duplicate detector:

intelliot:
## High Level Overview of Change

Partially revert the functionality introduced with #4195 / 5a15229 (part of 1.10.0-b1).

### Context of Change

Supersedes #4437. This PR simply squashes the two commits in #4437. I elected to do this on a new branch/PR so that we have a clean slate to re-review and look for any other potential risks.

Alternative to this PR: #4439

### Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)

## Test Plan

@manojsdoshi @sgramkumar to prepare a test plan.

<hr>

```
Acknowledgements:
Aaron Hook for responsibly disclosing this issue.

Bug Bounties and Responsible Disclosures:
We welcome reviews of the rippled code and urge researchers to
responsibly disclose any issues they may find.

To report a bug, please send a detailed report to:

    bugs@xrpl.org
```

Note: This PR will be squashed to 1 commit upon merging.

PR #4439, opened 2023-02-25
Revert "Improve detection & handling of duplicate Node ID:"

intelliot:

This reverts commit 5a15229eeb13b69c8adf1f653b88a8f8b9480546, which was introduced in 1.10.0-b1 and is not present in the code that is currently running in production (version 1.9.4).

Notice that this PR, #4439, is a "smaller" change than #4438 from the perspective of what node operators will be getting when they upgrade from 1.9.4.

### Context of Change

Alternative to #4438.

Given that 1.9.4 is stable and not vulnerable, the safest and most conservative change would be no change at all.

Therefore, we should consider the option of simply reverting the problematic commit, which is what this does. While `git` did not report any conflicts, the revert should still be code reviewed to ensure no unintended consequences.

Only one of {self, #4438} should be merged, a decision ideally based on a consensus of the maintainers.

### Type of Change

## Test Plan

@manojsdoshi @sgramkumar to prepare a test plan.

<hr>

```
Acknowledgements:
Aaron Hook for responsibly disclosing this issue.

We welcome reviews of the rippled code and urge researchers to

    bugs@xrpl.org
```

### DIFF
--- a/src/ripple/app/consensus/RCLConsensus.cpp
+++ b/src/ripple/app/consensus/RCLConsensus.cpp
@@ -86,11 +86,9 @@ RCLConsensus::Adaptor::Adaptor(
     , inboundTransactions_{inboundTransactions}
     , j_(journal)
     , validatorKeys_(validatorKeys)
-    , valCookie_(
-          1 +
-          rand_int(
-              crypto_prng(),
-              std::numeric_limits<std::uint64_t>::max() - 1))
+    , valCookie_{rand_int<std::uint64_t>(
+          1,
+          std::numeric_limits<std::uint64_t>::max())}
     , nUnlVote_(validatorKeys_.nodeID, j_)
 {
     assert(valCookie_ != 0);
--- a/src/ripple/app/main/Application.cpp
+++ b/src/ripple/app/main/Application.cpp
@@ -52,12 +52,10 @@
 #include <ripple/basics/ByteUtilities.h>
 #include <ripple/basics/PerfLog.h>
 #include <ripple/basics/ResolverAsio.h>
-#include <ripple/basics/random.h>
 #include <ripple/basics/safe_cast.h>
 #include <ripple/beast/asio/io_latency_probe.h>
 #include <ripple/beast/core/LexicalCast.h>
 #include <ripple/core/DatabaseCon.h>
-#include <ripple/crypto/csprng.h>
 #include <ripple/json/json_reader.h>
 #include <ripple/nodestore/DatabaseShard.h>
 #include <ripple/nodestore/DummyScheduler.h>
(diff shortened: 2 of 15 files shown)


drlongle (approved):

nbougalis (commented):
The _only_ issue here is the duplicate detection, which can be fixed with a one-line change, literally.

This reversion is more involved than the change in #4438, and therefore riskier. I believe that merging that change is the better path forward.

nbougalis (commented):
I think that reverting the change is riskier than just removing the call that forces the server to stop—a one line fix, _literally_.

I, personally, prefer merging #4438 which fixes this issue and avoids changing any code that doesn’t need to be touched.

RichardAH:
@intelliot isn't the correct solution to simply move the code that checks for self connection (including the new disambiguation with duplicate server) below the signature check? why does this code need to be removed?

RichardAH:
Agree with Nik here, why revert already reviewed and committed code unrelated to the actual attack vector just because it was in the same commit? That doesn't strike me as a sensible thing to do.

intelliot:
@RichardAH : 
1. The original reviews were clearly insufficient, as they failed to catch a critical vulnerability.
2. 100% of the network is currently running version 1.9.4, which is the latest stable release of rippled. 5a15229 was added in 1.10.0-b1, and is not present in 1.9.4. Version 1.10.0 will contain [70+ other commits](https://github.com/XRPLF/rippled/compare/1.9.4...develop), many of which are also risky. Reverting 5a15229 would give a result that has "minimal" changes versus what's currently live in production.

Does that make sense?

@nbougalis Why do you think reverting the change is "riskier"? 

I can certainly see an argument that it is "no safer", if you have 100% confidence that the other changes in 5a15229 are beneficial and contain no bugs. 

Ultimately, it really depends on whether there are any changes in the reverted commit that you feel must be included in 1.10.0. (What changes, specifically, are those?)

**From Aaron Hook:** aside from the stop signal, it also switched to using a cryptographically secure prng for the cookie, and allowed hex encoded values in addition to base64. I'd probably lean towards doing the revert and if any of the other changes are important enough, add them back as a new patch. But I don't have the context for the original patch, so I don't feel strongly either way.

intelliot:
> I, personally, prefer merging #4438 which fixes this issue and avoids changing any code that doesn’t need to be touched.

I understand this point of view. However, by the same argument, it is worth nothing that this PR (#4439) fixes this issue and, compared to 1.9.4 (which is live in production) avoids changing any code that doesn't need to be touched.

RichardAH:
> 1. The original reviews were clearly insufficient, as they failed to catch a critical vulnerability.

I think this is not a very strong argument. The same review process applies to the rest of the commits in the repository for literally years. If the review process is untrustworthy then all commits are.

intelliot:
Canceled in favor of #4438

PR #4440, opened 2023-02-28
pull req within src

MorganBergen:

- not relevant to task or issue, changes made to broken link in docs
- https://github.com/XRPLF/rippled/tree/develop/src#rippled-source

### Context of Change

If a bug fix, when was the bug introduced? What was the behavior?
- The bug was caused by incorrect markdown syntax, the broken link found in the README.md file to link to `apenwarr/git-subtree`, the redirection now works upon clicking on the git-subtree name.
-  the behavior is now fixed

### Type of Change

- [x] Documentation Updates

## Before / After
If relevant, use this section for an English description of the change at a technical level.

### Before

Some of these directories come from entire outside repositories brought in using [git-subtree][]. This means that the source files are inserted directly into the rippled repository. They can be edited and committed just as if they were normal files. [git-subtree]: https://github.com/apenwarr/git-subtree

### After 

Some of these directories come from entire outside repositories brought in using [git-subtree](https://github.com/apenwarr/git-subtree) This means that the source files are inserted directly into the `rippled` repository. They can be edited and committed just as if they were normal files. [git-subtree](https://github.com/apenwarr/git-subtree)

## Test Plan
Tested on vscode markdown previewer extension

## Future Tasks
No future tasks

### DIFF
--- a/src/README.md
+++ b/src/README.md
@@ -1,9 +1,9 @@
 # rippled Source
 
 Some of these directories come from entire outside repositories brought in
-using [git-subtree](.https://github.com/apenwarr/git-subtree) This means that the source files are inserted directly
+using [git-subtree](https://github.com/apenwarr/git-subtree) This means that the source files are inserted directly
 into the `rippled` repository. They can be edited and committed just as if they
-were normal files. [git-subtree](https://github.com/apenwarr/git-subtree) `apenwarr/git-subtree`
+were normal files. [git-subtree](https://github.com/apenwarr/git-subtree) 
 
 If you create a commit that contains files both from a subtree, and from the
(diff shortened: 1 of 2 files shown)


MarkusTeufelberger:
Thanks for your contribution! :-)

I think it would be easier for everyone involved if you open individual pull requests for individual changes you'd like to see (e.g. one for better/broken links in the documentation, another one for coding style in `Pathfinder.cpp`)

intelliot:
Also, please update the PR title to be a description/summary of the changes being proposed.

intelliot:
@MorganBergen as mentioned above, please open individual pull requests for individual changes. Thanks!

PR #4442, opened 2023-03-01
Rename NFT to NFToken for lsfDisallowIncomingNFTokenOffer flags

kennyzlei:

- Rename `lsfDisallowIncomingNFTOffer` to `lsfDisallowIncomingNFTokenOffer`
- Rename `asfDisallowIncomingNFTOffer` to `asfDisallowIncomingNFTokenOffer`
- Update spacing on `TXFlags.h` after rename

### Context of Change

`NFToken` is the more standard naming convention in the codebase compared to `NFT`

### Type of Change
- [ ] Documentation Updates
- [ ] Release


src/ripple/protocol/TxFlags.h
```diff
@@ -147,15 +147,15 @@ constexpr std::uint32_t const tfNFTokenMintMask =
     ~(tfUniversal | tfBurnable | tfOnlyXRP | tfTransferable);
 
 // NFTokenCreateOffer flags:
-constexpr std::uint32_t const tfSellNFToken            = 0x00000001;
+constexpr std::uint32_t const tfSellNFToken                = 0x00000001;
 constexpr std::uint32_t const tfNFTokenCreateOfferMask =
```
thejohnfreeman, line 151: It seems the equals signs in this file are aligned with the one on this line because it has the longest prefix. Two things:

1. After changing "NFT" to "NFToken", the new longest prefix is on line 85. The equals sign on that line should be one space apart from that prefix. In this commit, it is four spaces apart.
2. The equals sign on this line, 151, should be aligned with the others in the file.

thejohnfreeman (changes requested):
Not generally an advocate of rename-only changesets, but I'm fine approving work that has already been done.

src/ripple/protocol/TxFlags.h
kennyzlei: happy to go with that spacing guidance, i had assumed it was a nearest every 4 spaces approach since the last one happened to be % 4 in a clean way

kennyzlei: @thejohnfreeman for point 2. I think I have seen a different approach in other lines that fold so I thought it was more consistent to not align when they fold, see lines https://github.com/XRPLF/rippled/blob/develop/src/ripple/protocol/TxFlags.h#L67 for example

intelliot:
> Not generally an advocate of rename-only changesets, but I'm fine approving work that has already been done.

I don't understand this sentence. Naming is important, and it seems clear that this rename follows the principle of least surprise. If we agree that the rename is worth doing, then it is not a problem to do it in a rename-only changeset.

ckniffen (approved):

src/ripple/protocol/TxFlags.h
justinr1234: Out of curiosity, has it ever been considered to have formatting rules controlled by a linting tool automatically?

justinr1234 (approved):

src/ripple/protocol/TxFlags.h
thejohnfreeman: The [file](https://github.com/XRPLF/rippled/blob/db3b4dd396c210a90c9e29e2d1ea7046b339460e/src/ripple/protocol/TxFlags.h) as it looked before clang-format hit it in the Great Reformat appears to have every equals sign aligned, even on lines that wrap. They are all in column 44, and they are all 4 spaces after the longest prefix. Both of these rules match the alignment you have, @kennyzlei, except for the wrapped lines that you are not touching. Let's keep that alignment and just add a comment at the start, before or after the `// clang-format off` line, explaining this rule. Optionally, you can restore the wrapped lines to conform to the rule.

@justinr1234 We do have that. It's controlled by clang-format. Style configuration in [.clang-format](https://github.com/XRPLF/rippled/blob/develop/.clang-format). Enforcement in [.github/workflows/clang-format.yml](https://github.com/XRPLF/rippled/blob/develop/.github/workflows/clang-format.yml). We disable it for sections of special formatting like this. When deciding to incorporate it, the majority wanted to keep the special formatting.

kennyzlei: @thejohnfreeman added a comment to describe the special formatting

thejohnfreeman (approved):

thejohnfreeman:
@kennyzlei Please, in the future, [no more force-pushes](https://github.com/XRPLF/rippled/blob/develop/CONTRIBUTING.md#pull-requests). Not sure what is going on with review assignments either. Generally, assignments are left up even after a review is submitted.

kennyzlei:
😞 sorry I had to force push again here as my initial commits did not come with pgp signature, i thought i had auto-signed set up

intelliot:
No worries; I think the signature changes when the commits get squashed anyway. This looks fine to me: I'll go ahead and merge; if we find any problems, we can fix with a follow-up PR.

PR #4445, opened 2023-03-02
Packaging pipeline needs to catch new release branch name format

legleux:

The release branch being created as `releases/x.y.z` is missed by the regex that causes the production build pipeline to trigger.

The `get_components.sh` change is a little greedy and will trigger on _any_ branch that starts with `release` so maybe that's not desired unless a branch is never named like that.

### Type of Change

- [ ] Documentation Updates
- [ ] Release


legleux:
We are not pushing to a `releases/x.y.z` branch

PR #4446, opened 2023-03-03
Git hash in versionString when building develop branch

legleux:
When building rippled from the `develop` branch, we want the version string to contain the git hash.
This PR makes the `Release` build version string like this:
```
./build/develop/rippled --version
rippled version 1.10.0-rc3+220aa840435ca782db6776a361f3ced641af7e82
```

This PR in it's current state will produce the above, but certain configurations might create a version string that fails the `beast::SemanticVersion` check that happens at runtime and would crash rippled.

I'm not sure what the best way to do this is.

### Context of Change
To identify exactly which code a develop branch build is built from easily.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### DIFF
--- a/Builds/CMake/RippledSettings.cmake
+++ b/Builds/CMake/RippledSettings.cmake
@@ -120,3 +120,8 @@ if (coverage)
   message (STATUS "coverage build requested - forcing Debug build")
   set (CMAKE_BUILD_TYPE Debug CACHE STRING "build type" FORCE)
 endif ()
+
+git_branch (current_branch)
+if (current_branch STREQUAL "develop")
+  add_definitions(-DDEVELOP=TRUE)
+endif ()
--- a/src/ripple/protocol/impl/BuildInfo.cpp
+++ b/src/ripple/protocol/impl/BuildInfo.cpp
@@ -36,11 +36,13 @@ namespace BuildInfo {
 char const* const versionString = "1.10.0-rc3"
 // clang-format on
 
-#if defined(DEBUG) || defined(SANITIZER)
+#if defined(DEBUG) || defined(SANITIZER) || defined(DEVELOP)
     "+"
 #ifdef GIT_COMMIT_HASH
     GIT_COMMIT_HASH
+#if ! defined(DEVELOP)
     "."
+#endif
 #endif
 #ifdef DEBUG
     "DEBUG"
--- a/src/ripple/protocol/impl/BuildInfo.cpp
+++ b/src/ripple/protocol/impl/BuildInfo.cpp
@@ -38,20 +38,17 @@ char const* const versionString = "1.10.0-rc3"
 
 #if defined(DEBUG) || defined(SANITIZER) || defined(DEVELOP)
     "+"
+#define SEPARATOR ""
 #ifdef GIT_COMMIT_HASH
     GIT_COMMIT_HASH
-#if ! defined(DEVELOP)
-    "."
-#endif
+    #define SEPARATOR "."
 #endif
 #ifdef DEBUG
+    SEPARATOR
     "DEBUG"
+    #define SEPARATOR "."
 #ifdef SANITIZER
-    "."
-#endif
-#endif
-
-#ifdef SANITIZER
+    SEPARATOR
     BOOST_PP_STRINGIZE(SANITIZER)
 #endif
 #endif


legleux:
I'm not sure where this should go.

src/ripple/protocol/impl/BuildInfo.cpp
```diff
@@ -36,12 +36,14 @@ namespace BuildInfo {
 char const* const versionString = "1.10.0-rc3"
 // clang-format on
 
-#if defined(DEBUG) || defined(SANITIZER)
+#if defined(DEBUG) || defined(SANITIZER) || defined(DEVELOP)
     "+"
 #ifdef GIT_COMMIT_HASH
     GIT_COMMIT_HASH
+#if ! defined(DEVELOP)
```
legleux, line 43: Hmm, this logic is wrong in at least one way.
`SANITIZER` defined and this will crash rippled at runtime because of the version string not matching that semver version check

thejohnfreeman:
What is this change trying to do? Answer should go in [first comment](https://github.com/XRPLF/rippled/pull/4446#issue-1607851527). You can delete the template if you're not going to use it.

intelliot:
@legleux and I discussed:

- The packages just produced from `develop` will say rc3 though they were produced from code including `Rename NFT to NFToken for lsfDisallowIncomingNFTokenOffer flags` which the “rc3” packages we put in production yesterday do not have.

Therefore, we should either:

1. Not produce packages from `develop`. Wait for code to land in `release` before producing packages.
2. Make packages from `develop` have the commit hash in the version string.

In general, we add commits to develop without bumping the version. So, we need to implement one of the above 2 options.

- The version string does add the git rev if `DEBUG` is defined. We could change the condition in `BuildInfo.cpp`.
- `DEBUG` is not defined unless explicitly built with `-DCMAKE_BUILD_TYPE=Debug` (even though Release is has debug info and there is by default a `RelWithDebInfo` build in CMake).
- Made the changes but not sure if it’s where it should be. Overall, we want the hash to be in the version string when packages are built from `develop`. @ximinez may be able to help.

thejohnfreeman:
I don't understand @legleux's comment. My suspicion is that we want the string to:

- Append any subset of {`GIT_COMMIT_HASH`, `"DEBUG"`, `BOOST_PP_STRINGIZE(SANITIZER)`}, depending on what macros are defined.
- Separate these components from each other with a single dot (`.`).
- Separate this suffix, if non-empty, from the prefix with a single plus (`+`).

Is that correct?

Builds/CMake/RippledSettings.cmake
```diff
@@ -120,3 +120,8 @@ if (coverage)
   message (STATUS "coverage build requested - forcing Debug build")
   set (CMAKE_BUILD_TYPE Debug CACHE STRING "build type" FORCE)
 endif ()
+
+git_branch (current_branch)
+if (current_branch STREQUAL "develop")
+  add_definitions(-DDEVELOP=TRUE)
```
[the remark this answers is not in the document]
legleux: Yes, this is to specify exactly which commit a nightly `develop` build is coming from.

src/ripple/protocol/impl/BuildInfo.cpp
```diff
@@ -36,12 +36,14 @@ namespace BuildInfo {
 char const* const versionString = "1.10.0-rc3"
```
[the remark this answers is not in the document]
intelliot: > //  The build version number. You must edit this for each release
> //  and follow the format described at http://semver.org/. You may
> //  not use the '+' symbol to include for metadata.

Should this read "You may use the..."?

I think it's okay to put the git short hash after a `+`. I think this is [what we have been doing](https://github.com/XRPLF/rippled/pull/4084) to some extent.

thejohnfreeman:
```cpp
#if defined(DEBUG) || defined(SANITIZER) || defined(DEVELOP)
    "+"
#define SEPARATOR ""
#ifdef GIT_COMMIT_HASH
    GIT_COMMIT_HASH
    #define SEPARATOR "."
#endif
#ifdef DEBUG
    SEPARATOR
    "DEBUG"
    #define SEPARATOR "."
#ifdef SANITIZER
    SEPARATOR
    BOOST_PP_STRINGIZE(SANITIZER)
#endif
#endif
```

thejohnfreeman:
The preprocessor is inescapable in this case, because all of these metadata components come from the preprocessor. Even your `constexpr` code uses the preprocessor. I generally agree with avoiding the preprocessor for other cases, like generating code based off a template, but not with avoiding it dogmatically. I prefer a shorter, easier-to-read solution for this problem. I like both of our macro-based solutions.

src/ripple/protocol/impl/BuildInfo.cpp
```diff
@@ -38,20 +38,17 @@ char const* const versionString = "1.10.0-rc3"
 
 #if defined(DEBUG) || defined(SANITIZER) || defined(DEVELOP)
     "+"
+#define SEPARATOR ""
 #ifdef GIT_COMMIT_HASH
     GIT_COMMIT_HASH
-#if ! defined(DEVELOP)
-    "."
-#endif
+    #define SEPARATOR "."
```
nbougalis, line 44: Won't this (and the rest of these) result in warnings about a redefinition of `SEPARATOR`?

Also, my _personal preference_ would be to have `DEBUG` and `DEVELOP` go _before_ the `git commit` hash.

nbougalis (commented):
Left some comments. Unless there are warnings generated, the changes I discuss are at the author's discretion.

src/ripple/protocol/impl/BuildInfo.cpp
```diff
@@ -38,20 +38,17 @@ char const* const versionString = "1.10.0-rc3"
...
 #endif
 #ifdef DEBUG
+    SEPARATOR
     "DEBUG"
+    #define SEPARATOR "."
 #ifdef SANITIZER
```
nbougalis, line 50: Missing `#endif` here?

src/ripple/protocol/impl/BuildInfo.cpp
thejohnfreeman: Warning? Yes. I offered this revision in a chat with @legleux.
```cpp
#define SEPARATOR "+"
#ifdef GIT_COMMIT_HASH
    SEPARATOR
    GIT_COMMIT_HASH
    #undef SEPARATOR
    #define SEPARATOR "."
#endif
#ifdef DEBUG
    SEPARATOR
    "DEBUG"
    #undef SEPARATOR
    #define SEPARATOR "."
#endif
#ifdef SANITIZER
    SEPARATOR
    BOOST_PP_STRINGIZE(SANITIZER)
    #undef SEPARATOR
    #define SEPARATOR "."
#endif
```

legleux:
I'll revisit this when the github action pipeline is more complete.

Issue #4443, opened 2023-03-01
Publish Conan recipe for rippled


thejohnfreeman:
In the past few weeks, I have helped two coworkers write CMake for new projects that depend on the `xrpl_core` library[^2]. There are probably more people outside the company who want to perform this task. Until recently, we had [outdated instructions](https://github.com/XRPLF/rippled/commit/5ec8783d35a379b0811365e45758a863d3593609#diff-c982ca83cfd66ec2c29f53f8376b8fb951768c200f9b5c6ba212b923c317a3edL192-L234) that haven't worked since the move to Conan because the project's CMake no longer downloads and builds its dependencies[^1].

[^1]: I'll be including updated instructions in [new documentation](https://github.com/XRPLF/rippled/discussions/4433#discussioncomment-5104098).
[^2]: We should rename this to `libxrpl` before we invite even more dependents.

My favored solution is to publish a Conan recipe for rippled and let dependents use Conan for their projects. We're already halfway there with the move to Conan for our own dependencies. In fact, fixing the `package_info` [method](https://github.com/XRPLF/rippled/blob/develop/conanfile.py#L144-L149) in our recipe might be the only missing piece. It would be trivial to publish the recipe in this codebase, and I would consider that a success. It will be a separate effort to get the recipe included in [ConanCenter](https://github.com/conan-io/conan-center-index), but that would be an even bigger success.

PR #4449, opened 2023-03-06
Make it easy for projects to depend on libxrpl

thejohnfreeman:
This changeset does a few things:

- Adds an `ALIAS` target named `xrpl::libxrpl` for projects to link.
  - I chose this name for a few reasons. The current library target is named `xrpl_core`. There is no other "non-core" library target against which we need to distinguish the "core" library. We only export one library target, and it should just be named after the project to keep things simple and predictable. (I think underscores in target or library names are generally discouraged anyway.) Every target exported in CMake should be prefixed with the project name. The project should not be named after the company Ripple. I think most of us prefer to just name it after the ledger.
  - By _adding_ an `ALIAS` target, existing consumers who use the `xrpl_core` target will not be affected. In the future, we can start a migration plan to make `xrpl_core` the `ALIAS` target (and `libxrpl` the "real" target, which will affect the filename of the compiled binary), and eventually remove it entirely.
- Fixes the Conan recipe so that consumers using Conan import a target named `xrpl::libxrpl`. This way, every consumer can use the same instructions.
- Documents the two easiest methods to depend on libxrpl. I have tested both. I'll work on adding CI tests for both methods.

See #4443.

### DIFF
--- a/Builds/CMake/RippledCore.cmake
+++ b/Builds/CMake/RippledCore.cmake
@@ -13,6 +13,10 @@ if (unity)
   set_target_properties(xrpl_core PROPERTIES UNITY_BUILD ON)
 endif ()
 
+add_library(libxrpl INTERFACE)
+target_link_libraries(libxrpl INTERFACE xrpl_core)
+add_library(xrpl::libxrpl ALIAS libxrpl)
+
 
 #[===============================[
     beast/legacy FILES:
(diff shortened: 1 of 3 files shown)


seelabs (approved):
Both the build system changes look good and I liked the doc as well. Nice job on this!

Note that I did not create a project to exercise the instructions in the doc. I'll try to find time to do so, but this is a low risk change. I think it's ok to merge it. :+1: from me.

mvadari (commented):
I don't know CMake well enough to review changes so I'm not going to formally approve this, but I'm glad this is happening and this will be really useful.

intelliot:
@thejohnfreeman if you agree this is ready to merge, you can put the `Passed` label on it, or just comment saying that it's good to go

PR #4455, opened 2023-03-09
add result integer to transaction result codes

dangell7:

Added the result code integer to the transaction results.

### Context of Change

This will help sidechains sync with the registry. Currently if a sidechain wanted to add their own amendment, they would need to bring in every tel code as a placeholder. This allows them to specify a specific result code while skipping other amendment result codes.

### Type of Change

This would break everyone working on an amendment. You would get a clash but hopefully you can use the [registry](https://github.com/XRPLF/sFieldRegistry) to resolve and reserve your fields.

Its just a suggestion. No pressure.

### DIFF
--- a/src/ripple/protocol/TER.h
+++ b/src/ripple/protocol/TER.h
@@ -49,18 +49,18 @@ enum TELcodes : TERUnderlyingType {
     // - Not forwarded
     // - No fee check
     telLOCAL_ERROR = -399,
-    telBAD_DOMAIN,
-    telBAD_PATH_COUNT,
-    telBAD_PUBLIC_KEY,
-    telFAILED_PROCESSING,
-    telINSUF_FEE_P,
-    telNO_DST_PARTIAL,
-    telCAN_NOT_QUEUE,
-    telCAN_NOT_QUEUE_BALANCE,
-    telCAN_NOT_QUEUE_BLOCKS,
-    telCAN_NOT_QUEUE_BLOCKED,
-    telCAN_NOT_QUEUE_FEE,
-    telCAN_NOT_QUEUE_FULL,
+    telBAD_DOMAIN = -398,
+    telBAD_PATH_COUNT = -397,
+    telBAD_PUBLIC_KEY = -396,
+    telFAILED_PROCESSING = -395,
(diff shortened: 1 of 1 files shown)


justinr1234:
I don’t have context to understand how this might affect existing or users.

Does this mean you’re expecting consumers to be able to do a greater than / less than compare?

dangell7:
> I don’t have context to understand how this might affect existing or users.
> 
> Does this mean you’re expecting consumers to be able to do a greater than / less than compare?

See [here](https://github.com/XRPLF/rippled/pull/4292/files#diff-570760ccc79d416a7e98bf3e851c2e298ea5cf90f7207018e06aa6b46d09a013R124). See how on Xchain they needed to add the AMM error codes? This is because when generated without the absolute #, the error codes are generated in order. So for Xchain, they needed to add the AMM error codes in order to add their own error codes after. Or else there would have been a difference in error codes between the 2 builds. AMM error would have been 261 and the first Xchain error would have been 261 as well. 

I just thought it would be helpful if they were numbered.

dangell7:
I'm going to move this draft and possibly close. After more thought, I don't know if it adds that much value. If you are writing an amendment or a sidechain then you need to be aware of the error codes anyways. Technically they are added when the amendment is merged so you have to keep up anyways. 

My bad lol

justinr1234:
> I'm going to move this draft and possibly close. After more thought, I don't know if it adds that much value. If you are writing an amendment or a sidechain then you need to be aware of the error codes anyways. Technically they are added when the amendment is merged so you have to keep up anyways.
> 
> My bad lol

Heh wouldn’t say “your bad” but whatever you think is best! Thanks for helping my understand. 🙏🏻

scottschurr:
Yeah, @dangell7, as you noted I don't know that this change helps the situation very much.

The two kinds of codes that need to be stable, `tes` and `tec`, already have explicit values.  It's important that they be stable because those values are stored in ledger history.  And, yeah, those need to be carefully managed as the code base changes.

For the other kinds of codes, `tel`, `tem` and the like, it is less critical that they stay stable.  I think it's okay for those values to accrete, as they have been, when code is merged to develop.

PR #4459, opened 2023-03-13
Parse flags in account_info method

### LINKED ISSUE #2457: Request: Parse flags in account_info method
One fairly simple bit of functionality that's more complicated for users than it needs to be is parsing account flags. We can make things friendlier for users by modifying the account_info command to break the flags up into separate booleans in the response.

Performing bitwise operations on the `Flags` operation is not inherently difficult, but it is an extra step that's easy to get wrong, because there are three kinds of account flags and it's easy to mix them up. The three types of flags are account-set flags (asf*), AccountSet transaction flags (tf*), and AccountRoot ledger-state flags (lsf*). For purposes of account_info, we only need to care about the ledger-state flags.

I propose adding an `account_flags` object to the top level of the [`account_info` response](https://ripple.com/build/rippled-apis/#account-info), alongside `account_data`, `signer_lists`, etc. This object would be a map flags enabled for the account. **I recommend using the names of the fields from RippleAPI** so we can be consistent across interfaces. (The lsf* names are currently used in the code and documentation but never exposed in the rippled APIs.)

The existing convention of the rippled APIs would dictate that the `account_flags` object in the response should contain only flags for which the account's value is not in the default state. Since all flags default to `false`, all fields which appear in the response would have the value `true`. I disagree with this design and believe that **the response should always contain fields for all flags, whether their values are true or false.** (That makes it easier to write simple code to check a flag's value, and more closely matches the design of RippleAPI.)

| RippleAPI Name | Code Name | Description |
|---|---|---|
| defaultRipple | lsfDefaultRipple | If `true`, new trust lines to this address have rippling enabled by default on this account's side. |
| depositAuth | lsfDepositAuth | If `true`, the account can't be the destination of payments, etc. |
| disableMasterKey | lsfDisableMaster | If `true`, the account can't send transactions signed with its master key pair. |
| disallowIncomingXRP | lsfDisallowXRP | If `true`, client applications should reject or prompt before sending XRP to this account. |
| globalFreeze | lsfGlobalFreeze | If `true`, all currencies issued by this account are frozen. |
| noFreeze | lsfNoFreeze | If `true`, this account has given up the power to selectively freeze issued currencies. |
| passwordSpent | lsfPasswordSpent | If `true`, this account has used its free key reset. |
| requireAuthorization | lsfRequireAuth | If `true`, trust lines to this account require approval before they can hold currency issued by this account. |
| requireDestinationTag | lsfRequireDestTag | If `true`, transactions that send money to this account must specify a destination tag. |

drlongle:

This PR resolves #2457. It adds an object `account_flags` at the top level of the [account_info response](https://xrpl.org/account_info.html). This new object contains relevant flags of an account.

### Context of Change

```
./ripped account_info rG1QQv2nh2gr7RCZ1P8YYcBUKCCN633jCn

{
        "result" :
        {
                "account_data" :
                {
                        ...
                        ...
                },
                "account_flags" :
                {
                        "defaultRipple" : true,
                        "depositAuth" : true,
                        "disableMasterKey" : false,
                        "disallowIncomingCheck" : true,
                        "disallowIncomingNFTokenOffer" : true,
                        "disallowIncomingPayChan" : true,
                        "disallowIncomingTrustline" : true,
                        "disallowIncomingXRP" : true,
                        "globalFreeze" : true,
                        "noFreeze" : true,
                        "passwordSpent" : false,
                        "requireAuthorization" : true,
                        "requireDestinationTag" : true
                },
                "ledger_current_index" : 17,
                "status" : "success",
                "validated" : false
        }
}
```

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### DIFF
--- a/src/ripple/rpc/handlers/AccountInfo.cpp
+++ b/src/ripple/rpc/handlers/AccountInfo.cpp
@@ -78,11 +78,12 @@ doAccountInfo(RPC::JsonContext& context)
         return jvAccepted;
 
     static constexpr std::
-        array<std::pair<std::string_view, LedgerSpecificFlags>, 8>
+        array<std::pair<std::string_view, LedgerSpecificFlags>, 9>
             lsFlags{
                 {{"defaultRipple", lsfDefaultRipple},
                  {"depositAuth", lsfDepositAuth},
                  {"disableMasterKey", lsfDisableMaster},
+                 {"disallowIncomingXRP", lsfDisallowXRP},
                  {"globalFreeze", lsfGlobalFreeze},
                  {"noFreeze", lsfNoFreeze},
                  {"passwordSpent", lsfPasswordSpent},
@@ -90,10 +91,9 @@ doAccountInfo(RPC::JsonContext& context)
                  {"requireDestinationTag", lsfRequireDestTag}}};
 
     static constexpr std::
-        array<std::pair<std::string_view, LedgerSpecificFlags>, 5>
+        array<std::pair<std::string_view, LedgerSpecificFlags>, 4>
             disallowIncomingFlags{
-                {{"disallowIncomingXRP", lsfDisallowXRP},
-                 {"disallowIncomingNFTokenOffer",
+                {{"disallowIncomingNFTokenOffer",
                   lsfDisallowIncomingNFTokenOffer},
                  {"disallowIncomingCheck", lsfDisallowIncomingCheck},
                  {"disallowIncomingPayChan", lsfDisallowIncomingPayChan},
--- a/src/ripple/rpc/handlers/AccountInfo.cpp
+++ b/src/ripple/rpc/handlers/AccountInfo.cpp
@@ -78,12 +78,17 @@ doAccountInfo(RPC::JsonContext& context)
         return jvAccepted;
 
     static constexpr std::
-        array<std::pair<std::string_view, LedgerSpecificFlags>, 9>
+        array<std::pair<std::string_view, LedgerSpecificFlags>, 13>
             lsFlags{
                 {{"defaultRipple", lsfDefaultRipple},
                  {"depositAuth", lsfDepositAuth},
                  {"disableMasterKey", lsfDisableMaster},
                  {"disallowIncomingXRP", lsfDisallowXRP},
+                 {"disallowIncomingNFTokenOffer",
+                  lsfDisallowIncomingNFTokenOffer},
+                 {"disallowIncomingCheck", lsfDisallowIncomingCheck},
+                 {"disallowIncomingPayChan", lsfDisallowIncomingPayChan},
+                 {"disallowIncomingTrustline", lsfDisallowIncomingTrustline},
                  {"globalFreeze", lsfGlobalFreeze},
                  {"noFreeze", lsfNoFreeze},
                  {"passwordSpent", lsfPasswordSpent},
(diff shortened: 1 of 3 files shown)


src/ripple/rpc/handlers/AccountInfo.cpp
```diff
@@ -77,6 +77,19 @@ doAccountInfo(RPC::JsonContext& context)
...
+        array<std::pair<std::string_view, LedgerSpecificFlags>, 9>
+            lsFlags{
+                {{"defaultRipple", lsfDefaultRipple},
+                 {"depositAuth", lsfDepositAuth},
+                 {"disableMasterKey", lsfDisableMaster},
+                 {"disallowIncomingXRP", lsfDisallowXRP},
+                 {"globalFreeze", lsfGlobalFreeze},
+                 {"noFreeze", lsfNoFreeze},
+                 {"passwordSpent", lsfPasswordSpent},
+                 {"requireAuthorization", lsfRequireAuth},
```
ckniffen, line 90: This needs to support the new `DisallowIncoming` flags.

justinr1234 (approved):
Only comment would be to make a docs ticket on xrpl-dev-portal repo, otherwise looks great! Nice work!

src/ripple/rpc/handlers/AccountInfo.cpp
drlongle: Sure. Fixed.

intelliot:
@drlongle will check whether there are any other flags to add

drlongle:
> @drlongle will check whether there are any other flags to add

All account flags defined in [LedgerFormats.h](https://github.com/XRPLF/rippled/blob/develop/src/ripple/protocol/LedgerFormats.h#L223-L245) have been added.

src/ripple/rpc/handlers/AccountInfo.cpp
ckniffen: I would expect only to return those if the amendment is enabled.  It gives the false impression it those values are possible to set.

drlongle: Done.

src/ripple/rpc/handlers/AccountInfo.cpp
```diff
@@ -77,6 +77,28 @@ doAccountInfo(RPC::JsonContext& context)
...
+                 {"requireDestinationTag", lsfRequireDestTag}}};
+
+    static constexpr std::
+        array<std::pair<std::string_view, LedgerSpecificFlags>, 5>
+            disallowIncomingFlags{
+                {{"disallowIncomingXRP", lsfDisallowXRP},
```
ckniffen, line 95: disallowIncomingXRP is a current flag and should not be gated by the DisallowIncoming amendment.

seelabs (approved):
:+1: Looks good to me. Nice job on this.

ckniffen (approved):

intelliot:
@drlongle at your convenience, please comment with a suggested commit message for this PR

drlongle:
Add an object containing account flags to the account_info's response: (#2457)

Currently, the object account_data of the account_info's response contains a single field Flags that stores relevant flags of an account. Users need to performing bitwise operations on this field to retrieve relevant account flags. This PR adds a new object containing relevant flags of an account. This makes it easier to write simple code to check a flag's value.

PR #4473, opened 2023-03-23
replace hand-rolled lexicalCast

dangell7:

Replace hand-rolled code with std::from_chars for better performance and maintainability. The only minor change in semantics is the inability to parse the + sign, but that functionality is unused in practice.

### Type of Change

### DIFF
--- a/src/ripple/beast/core/LexicalCast.h
+++ b/src/ripple/beast/core/LexicalCast.h
@@ -23,6 +23,7 @@
 #include <algorithm>
 #include <cassert>
 #include <cerrno>
+#include <charconv>
 #include <cstdlib>
 #include <iostream>
 #include <iterator>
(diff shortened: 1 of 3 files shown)


intelliot:
> better performance

Can you quantify the change in performance in any way?

dangell7:
> > better performance
> 
> Can you quantify the change in performance in any way?

This change does not have a directly measurable impact in performance and synthetic benchmarks are unlikely to be helpful. The C++ std::from_chars is intended to be as efficient and fast as possible and it's unlikely that it will be any slower than the code that it replaces. And given that it also reduces the amount of hand-rolled code it's a net gain.

gregtatcam (commented):
PR looks fine to me but I'm reluctant to remove support for parsing `+` sign.

intelliot:
> I'm reluctant to remove support for parsing `+` sign.

@gregtatcam why?

gregtatcam:
> > I'm reluctant to remove support for parsing `+` sign.
> 
> @gregtatcam why?

I agree that it's probably not used in practice. But are we 100% sure that it's never used? And it's easy to add support for `+` parsing.

intelliot:
> are we 100% sure that it's never used? And it's easy to add support for `+` parsing.

@dangell7 thoughts?

HowardHinnant (changes requested):
I like that this replaces some of our code with std::code.

On the '+' issue, I recommend just checking `in[0] == '+' && isdigit(in[1])` prior to calling `from_chars`, being careful about strings of length 0 and 1 of course.  It likely won't be used, but it is cheap to support this existing outward facing API.

intelliot:
@dangell7 can you address the '+' issue as discussed above?

src/ripple/beast/core/LexicalCast.h
```diff
@@ -84,6 +84,9 @@ struct LexicalCast<Out, std::string>
         auto first = in.data();
         auto last = in.data() + in.size();
 
+        if (first != last && *first == '+')
```
gregtatcam, line 87: Have to check the next one is digit. See @HowardHinnant comment.

nbougalis: I don't think so - the semantics seem correct to me. If it's a plus and the next character is not a digit, then `std::from_chars` will return some kind of failure and the check, which ensures that (a) not failure is returned; and (b) that the whole string was parsed and nothing was left over will succeed. 

Check out this [example](https://godbolt.org/z/TMfYoMvYE). As you can see, the only cases where the semantics *don't* match is when you have a numeric value prefixed with a `+` sign. The "new" code does what we want, and the old one just fails.

nbougalis (approved):

dangell7:
> Optional drive-by fix: `#include <iostream>` is unnecessary and adds a small amount of code bloat. Needs removing from LexicalCast.h.

Thank you for this. Building and testing now. Is there a program you're using to identify these or any recommendations you can make?

HowardHinnant (approved):

intelliot:
> Is there a program you're using to identify these or any recommendations you can make?

No program. `<iostream>` is the only header in the std::lib that can cause a slight run-time cost just for including it. So it's special and Howard tends to look for it.

PR #4485, opened 2023-04-03
Fix package definition for Conan

thejohnfreeman:
Our project doesn't quite install itself correctly.

- It forgets to install (at least) one header. I have not exhaustively searched for more. We can add them as they come up. Longer term I want us to move headers to a separate directory in the project that should mirror the way they are installed so that reviewers don't have to do any mental calculus.
- It installs secp256k1 headers to a path other than how they are included in our code. I changed the includes because the installation path is correct.
- I forgot a few requirements in the Conan package info. This bug only reveals itself when indirectly trying to use one of the functions exported by those requirements.


intelliot:
For context, can you share example(s) of projects that could (or should) use the project package - and therefore use this fixed package definition?

src/secp256k1/src/bench_ecdh.c
```diff
@@ -6,8 +6,8 @@
 
 #include <string.h>
 
-#include "include/secp256k1.h"
```
intelliot, line 9: note: paths work when building within our codebase but not the source as it would be installed on the system

Builds/CMake/RippledCore.cmake
```diff
@@ -153,6 +153,7 @@ install (
     src/ripple/basics/Blob.h
     src/ripple/basics/Buffer.h
     src/ripple/basics/CountedObject.h
+    src/ripple/basics/Expected.h
```
intelliot, line 156: note: we forgot to add this to the cmake headers - doesn't affect us since we build in our codebase, but others depending on libxrpl would run into this.

this makes it possible for others to depend on libxrpl

intelliot:
note: fairly light review - mostly changing include paths. if needed, @thejohnfreeman could walk through what's changing and why

intelliot:
@seelabs to look

seelabs (approved):
:+1:

intelliot:
@legleux would you be able to review this?

legleux:
LGTM :+1:

legleux (approved):

PR #4495, opened 2023-04-10
Fix the fix for std::result_of

thejohnfreeman:


thejohnfreeman:
Re-created as #4496 under a different branch name.

PR #4499, opened 2023-04-13
Update BUILD.md

sappenin:
Add pointer to build boost from source.


thejohnfreeman:
The error mentioned is triggered by building Boost from source. The fix suggested does not address the problem. What was the full error message? Can use a Gist if it is long.

sappenin:
> The error mentioned is triggered by building Boost from source.

Ah, that's my mistake then -- I did build Boost from source at one point as I was trying to get my initial build to work, but I don't know what actually went wrong to generate that error, and am not able to reproduce it anymore, so I'll just close this PR and re-open it again if I encounter it in the future.

PR #4501, opened 2023-04-16
Fix errors for Clang 16

thejohnfreeman:
- `std::{u,bi}nary_function` were removed in C++17. They were empty classes with a few associated types. We already have conditional code to define the types. Just make it unconditional.
- libc++ checks a cast in an unevaluated context to see if a type inherits from a binary function class in the standard library, e.g. `std::equal_to`, and this causes an error when the type privately inherits from such a class. Change these instances to public inheritance.
- We don't need a middle-man for the empty base optimization. Prefer to inherit directly from an empty class than from `beast::detail::empty_base_optimization`.
- Clang warns when all the uses of a variable are removed by conditional compilation of assertions. Add a `[[maybe_unused]]` annotation to suppress it.
- As a drive-by clean-up, remove commented code.

See related work in #4486.

### DIFF
--- a/src/ripple/beast/container/detail/aged_ordered_container.h
+++ b/src/ripple/beast/container/detail/aged_ordered_container.h
@@ -146,7 +146,7 @@ private:
 
     // VFALCO TODO This should only be enabled for maps.
     class pair_value_compare
-        : public beast::detail::empty_base_optimization<Compare>
+        : public Compare
     {
     public:
         using first_argument = value_type;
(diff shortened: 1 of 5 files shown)


HowardHinnant (approved):

thejohnfreeman:
I expect this whole PR to be squashed into one commit, with subject line matching the title and body matching the first comment.

seelabs (approved):
:+1:

PR #4507, opened 2023-04-21
docs(BUILD): make it easier to find environment.md

intelliot:

Make the instructions a bit easier to follow. Users on different platforms can look for their platform name to find relevant information.

### Context of Change

It was a bit difficult to find the helpful instructions in environment.md.

### Type of Change

- [x] Documentation Updates


BUILD.md
```diff
@@ -34,40 +34,55 @@ below:
...
 
-[Boost]: https://www.boost.org/
+### Mac
 
+Many rippled engineers use macOS for development.
 
-## Prerequisites
+Here are [sample instructions for setting up a C++ development environment on macOS](./docs/build/environment.md#macos).
+
+### Windows
```
ckeshava, line 61: this is nice, we have info about all three platforms in this document.

BUILD.md
```diff
@@ -34,40 +34,55 @@ below:
...
+
+## Conan profile
 
 To build this package, you will need Python (>= 3.7),
 [Conan][] (>= 1.55, < 2), and [CMake][] (>= 3.16).
 
 [Conan]: https://conan.io/downloads.html
 [CMake]: https://cmake.org/download/
 
+After you have a [C++ development environment](./docs/build/environment.md) ready with Git, Python, Conan, CMake, and a C++ compiler, you may need to set up your Conan profile.
```
ckeshava, line 80: Can we surface this line up to the top of "Prerequisites" section? Because once I saw the dependencies and their versions, I downloaded it myself manually.
Since there are a few alternative ways to install these dependencies, it could cause discrepancy in the developer environment.

ckeshava: I meant, bubble up the reference to the environment.md file

ckeshava (approved):
LMK your thoughts. I can work on modifying this document.

BUILD.md
intelliot: Sure, although I think it shouldn't matter how you set up your environment. I believe there are multiple ways that should work.

BUILD.md
```diff
@@ -34,40 +34,55 @@ below:
...
+See [System Requirements](https://xrpl.org/system-requirements.html).
+
+rippled generally requires:
+- git
+- Python
+- Conan
+- CMake
+- a C++ compiler
+
+### Linux
```
intelliot, line 48:
```suggestion
## Prerequisites

See [System Requirements](https://xrpl.org/system-requirements.html).

rippled generally requires git, Python, Conan, CMake, and a C++ compiler. Some guidance on setting up such a [C++ development environment can be found here](./docs/build/environment.md).

### Linux
```

BUILD.md
intelliot: Added a link to the environment.md file near the top of "Prerequisites" - https://github.com/XRPLF/rippled/pull/4507/commits/27750c9e5f17b930546aa22386fe1552ed654d29

intelliot:
@ckeshava feel free to propose other changes to this document, as you like.

ckeshava:
It looks good to me. 

I'm curious: Are there any plans to provide a DockerFile for people to spin up the rippled executable? It might not be suitable for development environment, but it might simplify the user experience for only maintaining a validator on the ledger.

intelliot:
@ckeshava re: Dockerfile, does this provide what you're looking for? -
- https://github.com/XRPLF/rippled/blob/develop/docs/Docker.md
- https://github.com/XRPLF/rippled/blob/develop/Builds/containers/ubuntu-builder/Dockerfile

also: in GitHub, you can go to "Files changed" -> "Review changes" to submit a review of this PR

ckeshava (approved):
the changes look good to me 👍

ggprod (approved):
lgtm

BUILD.md
```diff
@@ -100,16 +110,18 @@ architecture:
...
 use.
-This setting will set the correct variables (`CMAKE_<LANG>_COMPILER`) in the
-generated CMake toolchain file:
-
+Update the `conf.tools.build:compiler_executables` setting in order to set the correct variables (`CMAKE_<LANG>_COMPILER`) in the
+generated CMake toolchain file.
+For example, on Ubuntu 20, you may have gcc at `/usr/bin/gcc` and g++ at `/usr/bin/g++`; if that is the case, you can select those compilers with:
 ```
-conan profile update 'conf.tools.build:compiler_executables={"c": "<path>", "cpp": "<path>"}' default
+conan profile update 'conf.tools.build:compiler_executables={"c": "/usr/bin/gcc", "cpp": "/usr/bin/g++"}' default
```
thejohnfreeman, line 120: This is unnecessary because `cc` and `c++` are the defaults that CMake will use (and CMake is the build system used by almost all of our dependencies), and they will resolve to `/usr/bin/gcc` and `/usr/bin/g++` on the default `PATH`. If builders require this level of instruction, then they likely don't understand that `/usr/bin/g++` is the wrong choice. It is more likely to be the default installed compiler and not the newer version they presumably installed to build rippled. This example seems like it's just going to confuse people more. They will fail to build with the defaults, then set these values that match the defaults, it won't fix their problem, and they'll think this instruction is wrong.

What was the misunderstanding that led to this change?

BUILD.md
```diff
@@ -100,16 +110,18 @@ architecture:
...
 ```
 
+Replace `/usr/bin/gcc` and `/usr/bin/g++` with paths to the desired compilers.
+
```
thejohnfreeman, line 124: Why does this example use real paths while the next example uses the `<path>` placeholder? Can we make this consistent?

BUILD.md
```diff
@@ -34,40 +34,50 @@ below:
...
 
 [Conan]: https://conan.io/downloads.html
 [CMake]: https://cmake.org/download/
 
+After you have a [C++ development environment](./docs/build/environment.md) ready with Git, Python, Conan, CMake, and a C++ compiler, you may need to set up your Conan profile.
+
+These instructions assume a basic familiarity with Conan and CMake.
+
+If you are unfamiliar with Conan, then please read [this crash course](./docs/build/conan.md) or the official [Getting Started][3] walkthrough.
+
```
thejohnfreeman, line 80: Why remove the `**Warning**` designation? It [renders](https://github.com/orgs/community/discussions/16925) as a nice attention-grabbing admonition on GitHub.

BUILD.md
```diff
@@ -34,40 +34,50 @@ below:
...
-> If you are unfamiliar with Conan,
-> then please read our [crash course](./docs/build/conan.md)
-> or the official [Getting Started][3] walkthrough.
+- Additionally, 32-bit Windows development is not supported.
+- Visual Studio 2022 is not yet supported.
+  - rippled generally requires [Boost][] 1.77, which Conan cannot build with VS 2022.
+  - Until rippled is updated for compatibility with later versions of Boost, Windows developers may need to use Visual Studio 2019.
+
+[Boost]: https://www.boost.org/
+
```
thejohnfreeman, line 66: These subsections essentially say the same thing as before, except in more words and with "here" links, which are generally considered [bad practice](https://dap.berkeley.edu/news/whats-wrong-using-click-here-links) on the web. What is the purpose?

To be pedantic, rippled does not require this environment. *Building* rippled requires this environment.

thejohnfreeman (changes requested):

BUILD.md
intelliot: > This is unnecessary because `cc` and `c++` are the defaults that CMake will use (and CMake is the build system used by almost all of our dependencies), and they will resolve to `/usr/bin/gcc` and `/usr/bin/g++` on the default `PATH`. If builders require this level of instruction, then they likely don't understand that `/usr/bin/g++` is the wrong choice. It is more likely to be the default installed compiler and not the newer version they presumably installed to build rippled.

How do you know that `/usr/bin/g++` is the wrong choice?

> This example seems like it's just going to confuse people more. They will fail to build with the defaults, then set these values that match the defaults, it won't fix their problem, and they'll think this instruction is wrong.
> 
> What was the misunderstanding that led to this change?

People have failed to build with the instructions as given because they don't provide a path that could ever work:

```
CMake Error at CMakeLists.txt:49 (project):
  The CMAKE_CXX_COMPILER:

    <path>

  is not a full path and was not found in the PATH.

  Tell CMake where to find the compiler by setting either the environment
  variable "CXX" or the CMake cache entry CMAKE_CXX_COMPILER to the full path
  to the compiler, or to the compiler name if it is in the PATH.
```

BUILD.md
intelliot: If people don't understand it one way, there's a nonzero chance they may at least understand it when it's provided in a different way.

BUILD.md
intelliot: The text color is a lighter grey, making it easier to ignore. That said, if you prefer the **Warning**, we can leave it for now.

BUILD.md
```diff
@@ -34,40 +34,50 @@ below:
 | Apple Clang | 13.1.6
 | MSVC        | 19.23
 
-We do not recommend Windows for rippled production use at this time.
-As of January 2023, the Ubuntu platform has received the highest level of
+## Prerequisites
+
+See [System Requirements](https://xrpl.org/system-requirements.html).
+
+rippled generally requires git, Python, Conan, CMake, and a C++ compiler. Some guidance on setting up such a [C++ development environment can be found here](./docs/build/environment.md).
```
intelliot, line 41:
```suggestion
Building rippled generally requires git, Python, Conan, CMake, and a C++ compiler. Some guidance on setting up such a [C++ development environment can be found here](./docs/build/environment.md).
```

BUILD.md
intelliot: The links are not "click here" links; some have the word "here" in them, but it's just one word in a longer, descriptive linked text.

Added _Building_ in 244622b40df7ba6b455e83dccaff37b5ff3ea63b

BUILD.md
thejohnfreeman: It's because `/usr/bin/g++` is the default value (when it exists on a platform). If you have to set these settings, it's because you have multiple compilers installed and the default compiler, the one picked by CMake or b2 by default, is not the one that can build rippled. If you set this setting to the default value, it's the same as not setting it. The only reason to set this setting is to override the default value.

`<path>` is a placeholder. Isn't that obvious? When people see that notation in any other command line, including CMake help strings, they know what it means: it means this is a parameter and you have to substitute the value that makes sense for you. Do we really have readers that don't know this? Genuine question. It starts to feel like this document is held to an impossible standard and prohibited from following common practices in software documentation.

The reason we have to use a placeholder is because we cannot tell what the right value is for any particular user. It might be any of the following on common platforms:

- `/usr/bin/g++-10`
- `/usr/bin/g++-11`
- `/usr/bin/g++-12`
- `/usr/bin/g++-13`
- `/usr/local/bin/g++-10`
- `/usr/local/bin/g++-11`
- `/usr/local/bin/g++-12`
- `/usr/local/bin/g++-13`
- `/opt/local/bin/g++-10`
- `/opt/local/bin/g++-11`
- `/opt/local/bin/g++-12`
- `/opt/local/bin/g++-13`
- `/opt/homebrew/bin/g++`
- `/opt/homebrew/bin/g++-10`
- `/opt/homebrew/bin/g++-11`
- `/opt/homebrew/bin/g++-12`
- `/opt/homebrew/bin/g++-13`
- `/usr/local/bin/clang-14'
- `/usr/local/bin/clang-15'
- `/usr/local/bin/clang-16'

This list is not exhaustive. It could literally be anywhere. Only the builder knows exactly where.

How do we get people to stop blindly copying commands, or to read the error message they get? The error message you pasted makes it immediately clear what the problem is, doesn't it?

These two sets of commands are coupled. They should either both use a placeholder or both use an example path. Though, if readers cannot recognize a placeholder that needs to be replaced, then how are they going to recognize an example that needs to be replaced?

BUILD.md
thejohnfreeman: I think we need to first stop and examine why people do not understand it the way it is written. Or even _whether_ people do not understand it the way it is written. In fact, there _could_ be a zero chance that readers do not understand the different way, and a non-zero chance that _more_ readers are confused by the different way. @ckeshava can you explain which part of this instruction you didn't understand in your reading?

BUILD.md
intelliot: > It's because `/usr/bin/g++` is the default value (when it exists on a platform). If you have to set these settings, it's because you have multiple compilers installed and the default compiler, the one picked by CMake or b2 by default, is not the one that can build rippled or its dependencies. If you set this setting to the default value, it's the same as not setting it. The only reason to set this setting is to override the default value.

That was not clear from the existing doc. I had styled **If you have multiple compilers** in bold, which would likely have helped, but c8f17324f6b4aa5d8e2d8f7994fb7b76a7d4bf49 makes it even more clear by giving it a new sub-heading, "Multiple compilers".

> `<path>` is a placeholder. Isn't that obvious? When people see that notation in any other command line, including CMake help strings, they know what it means: it means this is a parameter and you have to substitute the value that makes sense for you. Do we really have readers that don't know this? Genuine question. It starts to feel like this document is held to an impossible standard and prohibited from following common practices in software documentation.

CMake help strings are a different context than `BUILD.md`. In `BUILD.md`, anything you provide in code blocks (```) is expected to be copyable, and in fact, GitHub makes it easy to do this by providing a one-click "copy" button near the right side of the code block. To see this, open https://github.com/XRPLF/rippled/blob/develop/BUILD.md in a web browser.

> The reason we have to use a placeholder is because we cannot tell what the right value is for any particular user. It might be any of the following on common platforms:
> 
> * `/usr/bin/g++-10`
> * `/usr/bin/g++-11`
> * `/usr/bin/g++-12`
> * `/usr/bin/g++-13`
> * `/usr/local/bin/g++-10`
> * `/usr/local/bin/g++-11`
> * `/usr/local/bin/g++-12`
> * `/usr/local/bin/g++-13`
> * `/opt/local/bin/g++-10`
> * `/opt/local/bin/g++-11`
> * `/opt/local/bin/g++-12`
> * `/opt/local/bin/g++-13`
> * `/opt/homebrew/bin/g++`
> * `/opt/homebrew/bin/g++-10`
> * `/opt/homebrew/bin/g++-11`
> * `/opt/homebrew/bin/g++-12`
> * `/opt/homebrew/bin/g++-13`
> * `/usr/bin/clang++-14`
> * `/usr/bin/clang++-15`
> * `/usr/bin/clang++-16`
> * `/usr/local/bin/clang++-14`
> * `/usr/local/bin/clang++-15`
> * `/usr/local/bin/clang++-16`
> * ...
> 
> This list is not exhaustive. It could literally be anywhere. Only the builder knows exactly where.

Of course; that's why the document says "Replace `/usr/bin/gcc` and `/usr/bin/g++` with paths to the desired compilers."

That said, in most of the cases I've seen, `/usr/bin/gcc` and `/usr/bin/g++` are actually the right values. If they're not, the user can easily change them.

Furthermore, when providing an example command that someone can use to change the default, it is helpful to provide the default values in that example. Not only does it make it obvious what they are changing the setting _from_, but it also makes it easy to change the setting back to the default, if that is ever desired.

> How do we get people to stop blindly copying commands, or to read the error message they get?

You can remove the code block formatting (```). As for error messages, some will read them. The better solution is to prevent the error message from occurring in the first place.

> The error message you pasted makes it immediately clear what the problem is, doesn't it?

Yes, but it does not make clear what the solution is. `conan profile update 'conf.tools.build:compiler_executables={"c": "/usr/bin/gcc", "cpp": "/usr/bin/g++"}' default` was the solution for some users.

> These two sets of commands are coupled. They should either both use a placeholder or both use an example path. Though, if readers cannot recognize a placeholder that needs to be replaced, then how are they going to recognize an example that needs to be replaced?

It's not about "cannot" - it's about what they actually do, in practice, when trying to build the project while expending the minimal amount of time possible. Time is precious. We're all going to die. If we can change the instructions to save developers 10 minutes of time, all else equal, then we should do it.

As stated, it's not actually true that the example "needs to be replaced", since `/usr/bin/gcc` and `/usr/bin/g++` actually do work for users. And if you're saying that those are the defaults and the command should never be used if those are the desired compilers, then let's remove the code block so it doesn't look like this is a setting that users generally need to set.

BUILD.md
intelliot: That suggestion still makes some information look like a quote, and appear in lighter grey. How about removing the `>` so the text looks normal?

Or, if you want the text to be clearly associated with the **WARNING**, put it inside a cell:

| :warning: **WARNING** :warning:
|---|
| These instructions assume you have a C++ development environment ready with Git, Python, Conan, CMake, and a C++ compiler. |

BUILD.md
intelliot: As for "more words": the words make the text easier to quickly parse, so it's an improvement to the doc.

BUILD.md
intelliot: notes:
- the default is also platform specific
- most dependencies use CMake; boost uses b2

with CMake: when you don't specify compiler on command line, it will search. on Linux, it should start its search at /usr/bin/cc. on Ubuntu, it points to /etc/alt/cc which points to /usr/bin/gcc

thejohnfreeman:
I will try to catch up with this by the end of next week.

BUILD.md
godexsoft: I don't mind the changes. I think it is obvious either way to anyone with more than 30 min exposure to cmake.
There is also non zero chance that nothing good would come out of someone with no prior cmake and c++ experience trying to build rippled - they should just install the prebuilt packages 📦

godexsoft (approved):

BUILD.md
intelliot: Added warning in table cell via https://github.com/XRPLF/rippled/pull/4507/commits/deb142b18ac89586dde95f93bc6ddbab277c6f00

ckeshava (approved):

ckeshava:
I think the commits need to be rebased to the latest develop branch

BUILD.md
intelliot: We'll run with this and if it becomes a problem, go ahead and open a PR in the future to change it again.

Issue #3435, opened 2020-06-03
Convert code base from `boost::string_view` to `std::string_view`


carlhua:
## Issue Description

The codebase currently does not make a lot of use of string_view. Mostly it shows up in the asio facing parts of the code. I tried converting those uses of boost::string_view to std::string_view a few months ago. It almost worked.

There’s a problem in boost 1.70 where in Windows the flag that tells beast to use std::string_view (not boost::string_view) does not work correctly. I understand that particular problem is fixed in Boost 1.71. So once we get to a place where the code base requires boost 1.71 or greater, then I think we can take on this task.

Once we have std::string_view in the code base there are a number of optimizations that can be performed. But I believe it should all start by moving to std::string_view first.

Exported from RIPD-1789

intelliot:
@ckeshava could you think about this and let us know your recommendation?

Feel free to consult with @scottschurr as needed (he opened the original issue)

ckeshava:
Hello,
As per my understanding, we are using Boost v1.77. So we should not run into the issue pointed out earlier. I will test for compatibility on Visual Studio 2019 after making these changes.

I believe `std::string_view` is almost always better than `std::string const&`. Do I need to think about other alternatives for `boost::string_view`? What do you think @scottschurr  ?

scottschurr:
Hi @ckeshava.  Thanks for looking into this.

In terms of changes, my feeling is that, wherever we can, we should be replacing `boost::string_view` with `std::string_view`.  The places that had problems before were `boost::asio` and `boost::beast`.  Hopefully fixing those is now possible.  I also had problems with `soci` wanting to use `boost::string_view` instead of `std::string_view`.  It would be interesting to learn the state of the world on that front as well.

In terms of replacing `std::string` with `std::string_view`, that needs to be done very carefully with an eye toward lifetimes.  So if we want to address that (and there are definitely places where that replacement is a win), that kind of change would want to be in one or more separate commits from the one described in the previous paragraph (just changing `string_view` types).

In case it's helpful, here is an outstanding review of what should (and should not) be done with `string_view`: https://www.youtube.com/watch?v=H9gAaNRoon4

I hope that helps.  Let me know if you have further questions.

thejohnfreeman:
[...]

PR #4509, opened 2023-04-24
Update codebase from boost::string_view into std::string_view

ckeshava:
This PR seeks to solve issue #3435 

### Type of Change

- [ ] Documentation Updates
- [ ] Release

Before:
The codebase makes use of `boost:string_view` for cheap manipulation of strings. 

After:
This PR uses `std::string_view` instead of `boost::string_view` to perform similar operations.

Can I avoid making expensive conversions into `std::string` inside `Ping.cpp` for `jss::ip` and `jss::username` response fields?

The `http_request_type` has a dependency on `boost`. Is it okay to occasionally convert from boost's string_view into `std::string_view`? I'm not sure about the performance implications.


src/ripple/rpc/handlers/Ping.cpp
```diff
@@ -39,13 +39,14 @@ doPing(RPC::JsonContext& context)
             break;
         case Role::IDENTIFIED:
             ret[jss::role] = "identified";
-            ret[jss::username] = context.headers.user.to_string();
+            // Can I avoid making expensive copies for std::string?
+            ret[jss::username] = std::string{context.headers.user};
             if (context.headers.forwardedFor.size())
```
ckeshava, line 44: @scottschurr  I couldn't think of a cheaper solution for this piece of code.
`std::string{context.headers.user.data()}` seems to include garbage trailing data

src/ripple/rpc/impl/ServerHandlerImp.cpp
```diff
@@ -847,8 +848,14 @@ ServerHandlerImp::processRequest(
          */
         if (role != Role::IDENTIFIED && role != Role::PROXY)
         {
-            forwardedFor.clear();
-            user.clear();
+            // Simulate the clearing of string_view through a swap. Is this a
```
ckeshava, line 851: I'm not comfortable with this method. Since these strings (`user` and `forwardedFor`) are being modified, they are not suitable candidates for `std::string_view`. `boost::string_view` has a `clear()` member function, until std. Should I retain the original version or change this occurence to `std:string` instead? @scottschurr

src/ripple/rpc/handlers/Ping.cpp
scottschurr: At this level of the source code I'm not seeing `boost::string_view` used here.  Naively, it seems like you can just leave this alone?  Am I missing something?

Remember, there are _lots_ of places that `std::string_view` is not safe to use.  We should not try to replace uses of `std::string` with `std::string_view` unless we're really sure it is safe.  I suggest you not try _any_ replacements of `std::string` with `std::string_view` in this initial pass.

src/ripple/rpc/impl/ServerHandlerImp.cpp
scottschurr: `std::string_view` doesn't have a `clear()` method, but it does have a `remove_suffix()` method: https://en.cppreference.com/w/cpp/string/basic_string_view/remove_suffix

I think you might be able to use use...
```
forwardedFor.remove_suffix(firwardedFor.size())
```
What do you think?

src/ripple/rpc/handlers/Ping.cpp
ckeshava: I understand. `context.headers.user` and `context.headers.forwardedFor` are of type `boost::string_view` in the current version. Source: `src/ripple/rpc/Context.h:60`

I'm trying to port from this dependence on boost to std.

src/ripple/rpc/impl/ServerHandlerImp.cpp
ckeshava: Yes thanks, this looks much cleaner 👍  I'll incorporate this change.

src/ripple/rpc/handlers/Ping.cpp
scottschurr: Thanks for the clarification.  I don't think there's currently any way to make this cheaper than what you have done here.  The `Json::Value::operator[]` interface does not (yet) allow assignment from a `std::string_view`.But it could be made `string_view` aware in a future pull request if we chose to do that.

Alternatively, we could update the `rippled` codebase to use `boost::json` instead of `Json::Value`.  I think that would also get us `std::string_view` support.  But that would be a much bigger lift (with potentially more benefit).

ckeshava: ok, thanks for the suggestions Scott! 
Yeah, `Json::Value` was written a long time, I'm guessing before boost came up with their json library. It appears to have so much of overlap with implementations from boost!

src/ripple/overlay/impl/OverlayImpl.cpp
```diff
@@ -997,9 +997,13 @@ OverlayImpl::processValidatorList(
         return true;
     };
 
-    auto key = req.target().substr(prefix.size());
```
nbougalis, line 1000: I'd personally leave this alone.

ckeshava: If I leave type-deduction to auto, I'm not 100% sure if it is deduced as `std::string_view`. On line 1006, there is a comparison of `key` with `std::string_view::npos`.

My observation is that construction of `std::string_view` without the second length parameter results in garbage values at the end of `key`. I'm not really sure why that's happening.

scottschurr:
It seems like you're running into some problems interfacing with `boost::beast`, since it is inclined to use `boost::string_view` instead of `std:::string_view`.  Boost must support many users who are stuck on C++11, which doesn't have `std::string_view` as part of the library.  So this limitation makes a lot of sense to the Boost community.

I noticed in the Boost documentation that there's a configuration preprocessor macro that may be useful: `BOOST_BEAST_USE_STD_STRING_VIEW`.  That macro appears to be available in Boost 1.77, which looks like the oldest version of Boost we currently support.  Here's the 1.77 documentation: https://www.boost.org/doc/libs/1_77_0/libs/beast/doc/html/beast/config/configuration_preprocessor_defin.html

What I don't know is how many wrenches that change would throw into the Conan build process.  Perhaps @thejohnfreeman can advise us on how much trouble using that configuration preprocessor macro would cause, or if it's even possible.

scottschurr (commented):
These changes look great to me!  I left a few comments for your consideration, but none of them are required.

That said, I'm not approving this pull request yet.  I'd like to hear from @thejohnfreeman about the feasibility of using the `BOOST_BEAST_USE_STD_STRING_VIEW` boost configuration preprocessor macro.  If that's feasible, it may make sense to integrate that change with this pull request.  Or to stack this pull request on top of that change.

src/ripple/overlay/impl/OverlayImpl.cpp
ckeshava: i missed this comment, thanks for the explanation Scott! yes, that makes sense

src/ripple/rpc/impl/ServerHandlerImp.cpp
```diff
@@ -559,8 +559,9 @@ ServerHandlerImp::processSession(
         [&] {
             auto const iter = session->request().find("X-User");
             if (iter != session->request().end())
-                return iter->value();
-            return boost::beast::string_view{};
+                return std::string_view{
+                    iter->value().data(), iter->value().length()};
+            return std::string_view{};
         }());
```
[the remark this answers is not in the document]
ckeshava: thank you very much Scott! this looks nicer 👍

src/ripple/rpc/impl/Role.cpp
```diff
@@ -305,7 +304,10 @@ forwardedFor(http_request_type const& request)
         std::size_t found = it->value().find(',');
         if (found == boost::string_view::npos)
             found = it->value().length();
-        return extractIpAddrFromField(it->value().substr(0, found));
+        // http_request_type has a dependency on boost. Explicitly convert such
+        // boost::string_view into std::string_view
+        return extractIpAddrFromField(
+            std::string_view{it->value().substr(0, found).data(), found});
```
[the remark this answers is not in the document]
ckeshava: ok, this makes sense 👍

ckeshava:
addressed PR comments. tagging @thejohnfreeman  to get his feedback on Boost's macro usage and it's impact on the conan build process.

thejohnfreeman:
Sorry, I missed the earlier pings for my attention. It is not a problem to use the Boost macro. We already have to do something similar for `std::result_of`, which has [instructions](https://github.com/XRPLF/rippled/blob/develop/BUILD.md#no-stdresult_of) in `BUILD.md`.

If you want to enable / require the macro `BOOST_BEAST_USE_STD_STRING_VIEW`, then the instructions will be similar:

- Add the macro to `conf.tools.build:cxxflags` in the Conan profile. This is the future-proof method.
- Add it to `env.CXXFLAGS` too. This is the backwards-compatible method.
- Do not add it to `CFLAGS` if you can build without it.
- Add it to `options.boost:extra_b2_flags` in case it changes the ABI of the non-header-only Boost libraries we link against.

ckeshava:
Thanks for the tip John. I executed the above commands, but I ran into an issue here:

```
➜  cmake-build-debug git:(updateStringView) ✗ cmake -DCMAKE_TOOLCHAIN_FILE:FILEPATH=build/generators/conan_toolchain.cmake -DCMAKE_BUILD_TYPE=Debug ..  
-- Using Conan toolchain: /Users/ckeshavabs/personal-code/rippled/cmake-build-debug/build/generators/conan_toolchain.cmake
-- Conan toolchain: Setting CMAKE_POSITION_INDEPENDENT_CODE=ON (options.fPIC)
-- Conan toolchain: C++ Standard 20 with extensions OFF
-- Conan toolchain: Setting BUILD_SHARED_LIBS = OFF
-- The C compiler identification is AppleClang 14.0.3.14030022
-- The CXX compiler identification is AppleClang 14.0.3.14030022
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- gch:c966149aeb0daa028cd23c481c0215a15cca68f8
-- Using 5 cores for ExternalProject builds.
-- rippled version: 1.10.1
-- docker NOT found -- won't be able to build containers for packaging
-- Conan: Component target declared 'Boost::diagnostic_definitions'
-- Conan: Component target declared 'Boost::disable_autolinking'
-- Conan: Component target declared 'Boost::dynamic_linking'
-- Conan: Component target declared 'Boost::headers'
-- Conan: Component target declared 'Boost::boost'
-- Conan: Component target declared 'boost::_libboost'
-- Conan: Component target declared 'Boost::atomic'
-- Conan: Component target declared 'Boost::container'
-- Conan: Component target declared 'Boost::context'
-- Conan: Component target declared 'Boost::date_time'
-- Conan: Component target declared 'Boost::exception'
-- Conan: Component target declared 'Boost::math'
-- Conan: Component target declared 'Boost::math_c99'
-- Conan: Component target declared 'Boost::math_c99f'
-- Conan: Component target declared 'Boost::math_c99l'
-- Conan: Component target declared 'Boost::math_tr1'
-- Conan: Component target declared 'Boost::math_tr1f'
-- Conan: Component target declared 'Boost::math_tr1l'
-- Conan: Component target declared 'Boost::program_options'
-- Conan: Component target declared 'Boost::regex'
-- Conan: Component target declared 'Boost::serialization'
-- Conan: Component target declared 'Boost::stacktrace'
-- Conan: Component target declared 'Boost::stacktrace_addr2line'
-- Conan: Component target declared 'Boost::stacktrace_backtrace'
-- Conan: Component target declared 'Boost::stacktrace_basic'
-- Conan: Component target declared 'Boost::stacktrace_noop'
-- Conan: Component target declared 'Boost::system'
-- Conan: Component target declared 'Boost::test'
-- Conan: Component target declared 'Boost::test_exec_monitor'
-- Conan: Component target declared 'Boost::wserialization'
-- Conan: Component target declared 'Boost::chrono'
-- Conan: Component target declared 'Boost::filesystem'
-- Conan: Component target declared 'Boost::json'
-- Conan: Component target declared 'Boost::nowide'
-- Conan: Component target declared 'Boost::prg_exec_monitor'
-- Conan: Component target declared 'Boost::random'
-- Conan: Component target declared 'Boost::thread'
-- Conan: Component target declared 'Boost::timer'
-- Conan: Component target declared 'Boost::type_erasure'
-- Conan: Component target declared 'Boost::unit_test_framework'
-- Conan: Component target declared 'Boost::wave'
-- Conan: Component target declared 'Boost::contract'
-- Conan: Component target declared 'Boost::coroutine'
-- Conan: Component target declared 'Boost::fiber'
-- Conan: Component target declared 'Boost::fiber_numa'
-- Conan: Component target declared 'Boost::graph'
-- Conan: Component target declared 'Boost::iostreams'
-- Conan: Component target declared 'Boost::locale'
-- Conan: Component target declared 'Boost::log'
-- Conan: Component target declared 'Boost::log_setup'
-- Conan: Target declared 'boost::boost'
CMake Error at .debug/build/generators/cmakedeps_macros.cmake:39 (message):
  Library 'boost_log_setup' not found in package.  If 'boost_log_setup' is a
  system library, declare it with 'cpp_info.system_libs' property
Call Stack (most recent call first):
  .debug/build/generators/Boost-Target-debug.cmake:24 (conan_package_library_targets)
  .debug/build/generators/BoostTargets.cmake:26 (include)
  .debug/build/generators/BoostConfig.cmake:16 (include)
  Builds/CMake/deps/Boost.cmake:1 (find_package)
  CMakeLists.txt:69 (include)


-- Configuring incomplete, errors occurred!
```

Are we using `CMake.patch_config_paths` anywhere in the conan configuration? I found this comment to be relevant: https://github.com/conan-io/conan/issues/10816#issuecomment-1071247773

thejohnfreeman:
You executed what commands? You've not included enough context to debug your situation. Please copy a full log that reproduces your error into a Gist (not a GitHub comment). `CMake.patch_config_paths` seems like a red herring; not sure why you're asking about it. We've never used it.

ckeshava:
Ok, here is the complete gist: https://gist.github.com/ckeshava/a47fcf02624c95c7d8a9e3824c961cc9

I'm asking about `Cmake. patch_config_paths` because that configuration parameter is flagged as a possible red-herring in the comment thread linked in my previous comment.

To summarize, I updated the conan profile and generated new
```
conan profile update 'conf.tools.build:cxxflags+=["-DBOOST_BEAST_USE_STD_STRING_VIEW"]' default
conan profile update 'env.CXXFLAGS="-DBOOST_BEAST_USE_STD_STRING_VIEW"' default                
conan profile update 'options.boost:extra_b2_flags="define=BOOST_BEAST_USE_STD_STRING_VIEW

conan install .. --output-folder . --build missing --settings build_type=Debug # from within the build directory
```

The above error is generated in the last command

thejohnfreeman:
"Red herring" means distraction, irrelevant. I don't know why you linked that comment, or what it has to do with this issue. I can see it mentions `CMake.patch_config_paths`, but this issue does not, our code does not and has never.

From your log:

```ini
[options]
boost:extra_b2_flags="define=BOOST_BEAST_USE_STD_STRING_VIEW"
[build_requires]
[env]
CFLAGS=-DBOOST_ASIO_HAS_STD_INVOKE_RESULT
CXXFLAGS=-DBOOST_BEAST_USE_STD_STRING_VIEW
[conf]
tools.build:cflags=['-DBOOST_ASIO_HAS_STD_INVOKE_RESULT']
tools.build:cxxflags=['-DBOOST_ASIO_HAS_STD_INVOKE_RESULT', '-DBOOST_BEAST_USE_STD_STRING_VIEW', '-DBOOST_BEAST_USE_STD_STRING_VIEW']
```

Look at what has happened here. You have added (twice) the string `'-DBOOST_BEAST_USE_STD_STRING_VIEW'` to the array `[conf]tools.build:cxxflags`, but you have replaced the strings `[env]CXXFLAGS` and `[options]boost:extra_b2_flags` instead of adding a definition to the lists in those strings. It looks as if you modified the profile without understanding what you were doing or what you were trying to accomplish. You need a profile that looks like this:

```ini
[options]
boost:extra_b2_flags="define=BOOST_ASIO_HAS_STD_INVOKE_RESULT define=BOOST_BEAST_USE_STD_STRING_VIEW"
[build_requires]
[env]
CFLAGS=-DBOOST_ASIO_HAS_STD_INVOKE_RESULT -DBOOST_BEAST_USE_STD_STRING_VIEW
CXXFLAGS=-DBOOST_ASIO_HAS_STD_INVOKE_RESULT -DBOOST_BEAST_USE_STD_STRING_VIEW
[conf]
tools.build:cflags=['-DBOOST_ASIO_HAS_STD_INVOKE_RESULT', '-DBOOST_BEAST_USE_STD_STRING_VIEW']
tools.build:cxxflags=['-DBOOST_ASIO_HAS_STD_INVOKE_RESULT', '-DBOOST_BEAST_USE_STD_STRING_VIEW']
```

That was not the cause of the error you shared, but it would have likely caused an error later in the pipeline. Your error suggests a corruption in the generated CMake modules that makes CMake look for Boost binaries in a cache directory that you deleted. You should delete your bulid directory (`cmake-build-debug`), fix your Conan profile, and start over from the top.

ckeshava:
Okay thanks for the help, I have followed your suggestions. I fixed my conan profile, deleted the `cmake-build-debug` folder.

I deleted the conan cache with `rm -rf ~/.conan/data`
I executed this command from my buidl directory: `conan install .. --output-folder . --build missing --settings build_type=Debug ` -> This completed successfully (I have included the last few lines of this command's output below)

This command fails: `cmake -DCMAKE_TOOLCHAIN_FILE:FILEPATH=build/generators/conan_toolchain.cmake -DCMAKE_BUILD_TYPE=Debug ..`

Snapshot of the logs:

```
soci/4.0.3 package(): Packaged 1 '.txt' file: LICENSE_1_0.txt
soci/4.0.3 package(): Packaged 49 '.h' files
soci/4.0.3 package(): Packaged 2 '.a' files: libsoci_sqlite3.a, libsoci_core.a
soci/4.0.3: Package 'd83c95fc80dd73000c607492f399917b2447b7a7' created
soci/4.0.3: Created package revision f47470056797bc0450e227dc953b72e4
conanfile.py (xrpl/1.10.1): WARN: Using the new toolchains and generators without specifying a build profile (e.g: -pr:b=default) is discouraged and might cause failures and unexpected behavior
conanfile.py (xrpl/1.10.1): Generator 'CMakeDeps' calling 'generate()'
conanfile.py (xrpl/1.10.1): Generator txt created conanbuildinfo.txt
conanfile.py (xrpl/1.10.1): Calling generate()
conanfile.py (xrpl/1.10.1): WARN: Using the new toolchains and generators without specifying a build profile (e.g: -pr:b=default) is discouraged and might cause failures and unexpected behavior
conanfile.py (xrpl/1.10.1): Preset 'debug' added to CMakePresets.json. Invoke it manually using 'cmake --preset debug'
conanfile.py (xrpl/1.10.1): If your CMake version is not compatible with CMakePresets (<3.19) call cmake like: 'cmake <path> -G "Unix Makefiles" -DCMAKE_TOOLCHAIN_FILE=/Users/ckeshavabs/personal-code/rippled/cmake-build-debug/build/generators/conan_toolchain.cmake -DCMAKE_POLICY_DEFAULT_CMP0091=NEW -DCMAKE_BUILD_TYPE=Debug'
conanfile.py (xrpl/1.10.1): Aggregating env generators
conanfile.py (xrpl/1.10.1): Generated conaninfo.txt
conanfile.py (xrpl/1.10.1): Generated graphinfo
➜  cmake-build-debug git:(updateStringView) ✗ cmake -DCMAKE_TOOLCHAIN_FILE:FILEPATH=build/generators/conan_toolchain.cmake -DCMAKE_BUILD_TYPE=Debug ..  
-- Using Conan toolchain: /Users/ckeshavabs/personal-code/rippled/cmake-build-debug/build/generators/conan_toolchain.cmake
-- Conan toolchain: Setting CMAKE_POSITION_INDEPENDENT_CODE=ON (options.fPIC)
-- Conan toolchain: C++ Standard 20 with extensions OFF
-- Conan toolchain: Setting BUILD_SHARED_LIBS = OFF
-- The C compiler identification is AppleClang 14.0.3.14030022
-- The CXX compiler identification is AppleClang 14.0.3.14030022
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- gch:c966149aeb0daa028cd23c481c0215a15cca68f8
-- Using 5 cores for ExternalProject builds.
-- rippled version: 1.10.1
-- docker NOT found -- won't be able to build containers for packaging
-- Conan: Component target declared 'Boost::diagnostic_definitions'
-- Conan: Component target declared 'Boost::disable_autolinking'
-- Conan: Component target declared 'Boost::dynamic_linking'
-- Conan: Component target declared 'Boost::headers'
-- Conan: Component target declared 'Boost::boost'
-- Conan: Component target declared 'boost::_libboost'
-- Conan: Component target declared 'Boost::atomic'
-- Conan: Component target declared 'Boost::container'
-- Conan: Component target declared 'Boost::context'
-- Conan: Component target declared 'Boost::date_time'
-- Conan: Component target declared 'Boost::exception'
-- Conan: Component target declared 'Boost::math'
-- Conan: Component target declared 'Boost::math_c99'
-- Conan: Component target declared 'Boost::math_c99f'
-- Conan: Component target declared 'Boost::math_c99l'
-- Conan: Component target declared 'Boost::math_tr1'
-- Conan: Component target declared 'Boost::math_tr1f'
-- Conan: Component target declared 'Boost::math_tr1l'
-- Conan: Component target declared 'Boost::program_options'
-- Conan: Component target declared 'Boost::regex'
-- Conan: Component target declared 'Boost::serialization'
-- Conan: Component target declared 'Boost::stacktrace'
-- Conan: Component target declared 'Boost::stacktrace_addr2line'
-- Conan: Component target declared 'Boost::stacktrace_backtrace'
-- Conan: Component target declared 'Boost::stacktrace_basic'
-- Conan: Component target declared 'Boost::stacktrace_noop'
-- Conan: Component target declared 'Boost::system'
-- Conan: Component target declared 'Boost::test'
-- Conan: Component target declared 'Boost::test_exec_monitor'
-- Conan: Component target declared 'Boost::wserialization'
-- Conan: Component target declared 'Boost::chrono'
-- Conan: Component target declared 'Boost::filesystem'
-- Conan: Component target declared 'Boost::json'
-- Conan: Component target declared 'Boost::nowide'
-- Conan: Component target declared 'Boost::prg_exec_monitor'
-- Conan: Component target declared 'Boost::random'
-- Conan: Component target declared 'Boost::thread'
-- Conan: Component target declared 'Boost::timer'
-- Conan: Component target declared 'Boost::type_erasure'
-- Conan: Component target declared 'Boost::unit_test_framework'
-- Conan: Component target declared 'Boost::wave'
-- Conan: Component target declared 'Boost::contract'
-- Conan: Component target declared 'Boost::coroutine'
-- Conan: Component target declared 'Boost::fiber'
-- Conan: Component target declared 'Boost::fiber_numa'
-- Conan: Component target declared 'Boost::graph'
-- Conan: Component target declared 'Boost::iostreams'
-- Conan: Component target declared 'Boost::locale'
-- Conan: Component target declared 'Boost::log'
-- Conan: Component target declared 'Boost::log_setup'
-- Conan: Target declared 'boost::boost'
CMake Error at .debug/build/generators/cmakedeps_macros.cmake:39 (message):
  Library 'boost_log_setup' not found in package.  If 'boost_log_setup' is a
  system library, declare it with 'cpp_info.system_libs' property
Call Stack (most recent call first):
  .debug/build/generators/Boost-Target-debug.cmake:24 (conan_package_library_targets)
  .debug/build/generators/BoostTargets.cmake:26 (include)
  .debug/build/generators/BoostConfig.cmake:16 (include)
  Builds/CMake/deps/Boost.cmake:1 (find_package)
  CMakeLists.txt:69 (include)


-- Configuring incomplete, errors occurred!
➜  cmake-build-debug git:(updateStringView) ✗ conan profile show default
Configuration for profile default:

[settings]
os=Macos
os_build=Macos
arch=armv8
arch_build=armv8
compiler=apple-clang
compiler.version=14
compiler.libcxx=libc++
build_type=Debug
compiler.cppstd=20
[options]
boost:extra_b2_flags="define=BOOST_ASIO_HAS_STD_INVOKE_RESULT define=BOOST_BEAST_USE_STD_STRING_VIEW"
[conf]
tools.build:cflags=['-DBOOST_ASIO_HAS_STD_INVOKE_RESULT', '-DBOOST_BEAST_USE_STD_STRING_VIEW']
tools.build:cxxflags=['-DBOOST_ASIO_HAS_STD_INVOKE_RESULT', '-DBOOST_BEAST_USE_STD_STRING_VIEW']
[build_requires]
[env]
CFLAGS=-DBOOST_ASIO_HAS_STD_INVOKE_RESULT -DBOOST_BEAST_USE_STD_STRING_VIEW
CXXFLAGS=-DBOOST_ASIO_HAS_STD_INVOKE_RESULT -DBOOST_BEAST_USE_STD_STRING_VIEW
➜  cmake-build-debug git:(updateStringView) ✗ 
```

intelliot:
@ckeshava - have you made any progress on this?

ckeshava:
@intelliot  Sorry, I forgot about this PR, I'll work on it.

@thejohnfreeman  Based on your suggestions above, should we expect all rippled validators to perform the conan profile updates manually? Can we make it easier for validators to set up this conan profile?

thejohnfreeman:
I don't think we expect validators to build rippled. I think they use our packages. @legleux @manojsdoshi can you confirm?

Regardless, there are other profile changes we already require and document in the build instructions. This would just be one more instruction.

legleux:
> I don't think we expect validators to build rippled. I think they use our packages. @legleux @manojsdoshi can you confirm?

I think it's a fairly safe assumption that most validtators do _not_ build from source and are installing via binary packages.

ckeshava:
I have brought this branch up to speed, sorry about the delay.

I need to update the build instructions, I'll do that just before the merge. I'm concerned that it will unnecessarily add complexity to the build procedure otherwise.

thejohnfreeman (changes requested):

thejohnfreeman:
If you're not going to put the build instructions in this PR right now, can you please share them in a comment?

src/ripple/overlay/impl/PeerImp.cpp
```diff
@@ -176,7 +176,7 @@ PeerImp::run()
     if (auto const iter = headers_.find("Closed-Ledger");
         iter != headers_.end())
     {
-        closed = parseLedgerHash(iter->value());
+        closed = parseLedgerHash(std::string{iter->value()});
```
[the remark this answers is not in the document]
ckeshava: I feel `const std::string&` provides a stronger guarantee of existence (no dangling pointer) of the data, rather than `std::string_view`. What would happen if the input data goes out of scope? I know that `iter->value()` does not face that issue, but I'm concerned about other calls to `base64_decode`.

Reference: Notes section of https://en.cppreference.com/w/cpp/string/basic_string_view

ckeshava:
Yes, regarding the usage of `BOOST_BEAST_USE_STD_STRING_VIEW ` macro in the conan build process: I believe it is required only in `Role.cpp` and `OverlayImpl.cpp`. It does not appear to affect other parts of the codebase. 

The changes compile and work well even without the addition of this macro. I feel the changes to the build process could be put into another PR, what do you guys think?

src/ripple/overlay/impl/PeerImp.cpp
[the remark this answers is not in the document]
thejohnfreeman: Any existing caller of `base64_decode` is already passing a `std::string`. Changing `base64_decode` to accept a `std::string_view` is safe because there is no use of the parameter, or any alias to it, after the function returns.

thejohnfreeman:
The definition of `BOOST_BEAST_USE_STD_STRING_VIEW` will need to be the same for all translation units whether the source file uses it directly or not.

It is expected that the current PR should compile with or without the definition. The difference is that _with_ the definition, we should be able to get rid of manual conversions between `boost::string_view` and `std::string_view` because they should be the same type. Did you try that?

If we decide to take advantage of that difference, then the build instructions for adding that definition should be included in this PR. If we decide otherwise, then we should remain silent on the definition.

src/ripple/overlay/impl/Handshake.cpp
```diff
@@ -46,8 +44,8 @@ getFeatureValue(
         return {};
     boost::smatch match;
     boost::regex rx(feature + "=([^;\\s]+)");
-    std::string const value = header->value();
-    if (boost::regex_search(value, match, rx))
+    const std::string allFeatures = std::string{header->value()};
+    if (boost::regex_search(allFeatures, match, rx))
```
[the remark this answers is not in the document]
ckeshava: interesting observation!

boost provides the following typedefs for the `boost::smatch match;` variable. We use it to store the appropriate portion of the input. I couldn't find the correct type for this variable. Ideally, I'd need `match_results<std::string_view::const_iterator>` to be the type of `match`.

`match_results<const char*>` is incorrect, because some inputs might not be null-terminated.
`match_results<std::string::const_iterator>` is type incompatible with `std:string_view value;`

```
typedef match_results<const char*> cmatch;
typedef match_results<std::string::const_iterator> smatch;
#ifndef BOOST_NO_WREGEX
typedef match_results<const wchar_t*> wcmatch;
typedef match_results<std::wstring::const_iterator> wsmatch;
#endif
```

I need to use this overload for the regex_search function.
```
template <class BidiIterator, class Allocator, class charT, class traits>
bool regex_search(BidiIterator first, BidiIterator last, 
                  match_results<BidiIterator, Allocator>& m, 
                  const basic_regex<charT, traits>& e, 
                  match_flag_type flags,
                  BidiIterator base)
```

How can I work around this?

src/ripple/overlay/impl/PeerImp.cpp
```diff
@@ -372,8 +372,8 @@ std::string
 PeerImp::getVersion() const
 {
     if (inbound_)
-        return headers_["User-Agent"];
-    return headers_["Server"];
+        return std::string{headers_["User-Agent"]};
+    return std::string{headers_["Server"]};
```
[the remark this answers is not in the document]
ckeshava: okay. But I don't think that will provide us with any performance gains, because we need to convert to `std::string` in order to insert them into `Json::Value` objects. So, there is one extra copy performed somewhere along the stack.

src/ripple/overlay/impl/Handshake.cpp
```diff
@@ -288,8 +289,8 @@ verifyHandshake(
     PublicKey const publicKey = [&headers] {
         if (auto const iter = headers.find("Public-Key"); iter != headers.end())
         {
-            auto pk =
-                parseBase58<PublicKey>(TokenType::NodePublic, iter->value());
+            auto pk = parseBase58<PublicKey>(
```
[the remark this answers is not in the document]
ckeshava: This would entail changing multiple functions along the stack until `decodeBase58Token` in `tokens.cpp`. I'm concerned about my earlier comment about the lifetime of the `std::string_view` object, otherwise I'm happy to do this change.

src/ripple/overlay/impl/OverlayImpl.cpp
```diff
@@ -997,9 +997,16 @@ OverlayImpl::processValidatorList(
...
+    // Convert the boost::string_view, returned by
+    // boost::http::request::target(), into a std::string_view.
+    //    std::string_view key = [&req, &prefix]() {
+    //        std::string_view const key = req.target().substr(prefix.size());
+    //        return key;
+    //    }();
```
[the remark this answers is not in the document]
ckeshava: fixed in [9d9f40e](https://github.com/XRPLF/rippled/pull/4509/commits/9d9f40e6cb0f5e1342fe9c5e156258e72b15da6a)

src/ripple/overlay/impl/Handshake.cpp
```diff
@@ -235,15 +233,15 @@ verifyHandshake(
 {
     if (auto const iter = headers.find("Server-Domain"); iter != headers.end())
     {
-        if (!isProperlyFormedTomlDomain(iter->value()))
+        if (!isProperlyFormedTomlDomain(std::string{iter->value()}))
```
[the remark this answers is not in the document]
ckeshava: fixed in [9d9f40e](https://github.com/XRPLF/rippled/pull/4509/commits/9d9f40e6cb0f5e1342fe9c5e156258e72b15da6a)

src/ripple/rpc/impl/ServerHandlerImp.h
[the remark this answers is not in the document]
ckeshava: fixed in [9d9f40e](https://github.com/XRPLF/rippled/pull/4509/commits/9d9f40e6cb0f5e1342fe9c5e156258e72b15da6a)

src/ripple/overlay/impl/PeerImp.cpp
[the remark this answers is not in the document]
ckeshava: fixed in [9d9f40e](https://github.com/XRPLF/rippled/pull/4509/commits/9d9f40e6cb0f5e1342fe9c5e156258e72b15da6a)

src/ripple/overlay/impl/Handshake.cpp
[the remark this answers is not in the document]
ckeshava: fixed in https://github.com/XRPLF/rippled/pull/4509/commits/2cf0eb896c64d078133f108fc1fe55d79c349391
On a related note, if I add a `Json::Value::asStringView()` function overload, would it provide meaningful benefit? It will reduce one copy of the string object

src/ripple/overlay/impl/Handshake.cpp
```diff
@@ -252,11 +250,14 @@ verifyHandshake(
...
-            [str = std::string(iter->value())]() -> TimeKeeper::time_point {
+        auto const netTime = [str = iter->value()]() -> TimeKeeper::time_point {
             TimeKeeper::duration::rep val;
 
-            if (beast::lexicalCastChecked(val, str))
+            // boost::LexicalCast does not have a specialization for <Out,
+            // std::string_view>. It's not obvious if such an instantiation
+            // is technically feasible. This necessitates the conversion
+            // from std::string_view to std::string
+            if (beast::lexicalCastChecked(val, std::string{str}))
```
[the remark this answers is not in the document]
ckeshava: thanks for calling this out :)
fixed in  https://github.com/XRPLF/rippled/pull/4509/commits/2cf0eb896c64d078133f108fc1fe55d79c349391

src/ripple/overlay/impl/PeerImp.cpp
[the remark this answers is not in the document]
ckeshava: fixed in https://github.com/XRPLF/rippled/pull/4509/commits/2cf0eb896c64d078133f108fc1fe55d79c349391

src/ripple/beast/core/LexicalCast.h
```diff
@@ -78,8 +78,12 @@ struct LexicalCast<Out, std::string>
     std::enable_if_t<
         std::is_integral_v<Integral> && !std::is_same_v<Integral, bool>,
         bool>
-    operator()(Integral& out, std::string const& in) const
+    operator()(Integral& out, std::string_view const& in) const
     {
+        // the underlying data of the std::string_view is no longer alive
+        if (in.empty())
+            return false;
```
thejohnfreeman, line 85: That's not what `empty()` tests. `empty()` returns whether `size() == 0`. There is no test for whether the memory pointed by `std::string_view` is still valid, and even if there were, no such test is necessary in this function. All current callers are passing `std::string`, which is guaranteed to point to valid memory. `std::string` is implicitly convertible to a `std::string_view`, which means no existing callers need to be changed. Only _future_ callers need to be aware that the `std::string_view` they are passing to this function must point to valid memory, but why wouldn't they know that? What did they think would happen to the view they pass? Please remove this condition.

src/ripple/beast/core/LexicalCast.h
```diff
@@ -92,20 +96,27 @@ struct LexicalCast<Out, std::string>
     }
 
     bool
-    operator()(bool& out, std::string in) const
+    operator()(bool& out, std::string_view const& in) const
     {
+        // the underlying data of the std::string_view is no longer alive
+        if (in.empty())
+            return false;
```
thejohnfreeman, line 103: Please remove this condition too.

src/ripple/beast/core/LexicalCast.h
```diff
@@ -126,7 +149,8 @@ struct LexicalCast<Out, char const*>
     bool
     operator()(Out& out, char const* in) const
     {
-        return LexicalCast<Out, std::string>()(out, in);
+        assert(in != 0);
```
thejohnfreeman, line 152: I think it would be more conventional to write this as `assert(in != nullptr)` or (my preference) `assert(in)`.

src/ripple/overlay/impl/Cluster.cpp
```diff
@@ -113,7 +113,8 @@ Cluster::load(Section const& nodes)
             return false;
         }
 
-        auto const id = parseBase58<PublicKey>(TokenType::NodePublic, match[1]);
+        auto const id = parseBase58<PublicKey>(
+            TokenType::NodePublic, std::string_view{match[1].str()});
```
thejohnfreeman, line 117: Are you sure this cast is necessary? Doesn't `str()` return a `std::string` which is implicitly convertible to `std::string_view`?

src/ripple/overlay/impl/Handshake.cpp
```diff
@@ -233,15 +233,15 @@ verifyHandshake(
...
+        if (!isProperlyFormedTomlDomain(iter->value()))
             throw std::runtime_error("Invalid server domain");
     }
 
     if (auto const iter = headers.find("Network-ID"); iter != headers.end())
     {
         std::uint32_t nid;
 
-        if (!beast::lexicalCastChecked(nid, std::string{iter->value()}))
+        if (!beast::lexicalCastChecked(nid, std::string_view{iter->value()}))
```
thejohnfreeman, line 244: Are you sure this cast is necessary? Why is it not necessary on line 236?

src/ripple/overlay/impl/Handshake.cpp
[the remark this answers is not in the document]
thejohnfreeman: A `std::string` is already implicitly convertible to a `std::string_view`, so I don't think an `asStringView()` method is valuable. I do think an `operator= (std::string_view const&)` overload would be valuable, however.

src/ripple/overlay/impl/OverlayImpl.cpp
```diff
@@ -829,7 +829,7 @@ OverlayImpl::getOverlayInfo()
             auto version{sp->getVersion()};
             if (!version.empty())
                 // Could move here if Json::value supported moving from strings
-                pv[jss::version] = version;
+                pv[jss::version] = std::string{version};
```
thejohnfreeman, line 832: Good place here for that `operator= (std::string_view const&)` overload I mentioned.

src/ripple/overlay/impl/OverlayImpl.cpp
```diff
@@ -999,12 +999,10 @@ OverlayImpl::processValidatorList(
...
-    //    std::string_view key = [&req, &prefix]() {
-    //        std::string_view const key = req.target().substr(prefix.size());
-    //        return key;
-    //    }();
-
-    std::string_view key = req.target().substr(prefix.size());
+    std::string_view key = [&req, &prefix]() {
+        boost::string_view const key = req.target().substr(prefix.size());
+        return std::string_view(key.data(), key.length());
+    }();
```
thejohnfreeman, line 1005: What was wrong with the old code? Specifically:

```cpp
std::string_view key = req.target().substr(prefix.size());
```

thejohnfreeman (changes requested):

src/ripple/overlay/impl/Handshake.cpp
ckeshava: Without the explicit cast, this is the error message from the compiler:
`In template: implicit instantiation of undefined template 'beast::detail::LexicalCast<unsigned int, boost::core::basic_string_view<char>>'`

It looks like boost::string_view is not implicitly convertible to std::string_view during template overload resolution

src/ripple/overlay/impl/Handshake.cpp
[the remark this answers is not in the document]
ckeshava: `Json::Value::asString()` returns a copy of a string object. If we could have a `asStringView()` function, we will not need to perform such a copy. The strings in an incoming request would not need an extra copy, if we use an `asStringView` function.

Is `operator= (std::string_view const&)` needed? I'm wondering if passing a `std::string const&` or `std::string_view const&` is better, for the function parameter type?

`operator= (std::string_view const&)` takes up 132 lines 
`operator= (std::string const&)` takes up 122 lines 

From a performance point of view, is `std::string_view const&` a lot better than `std::string const&`?

https://godbolt.org/z/MM1rG3fhj

src/ripple/overlay/impl/OverlayImpl.cpp
ckeshava: hmm okay

src/ripple/overlay/impl/OverlayImpl.cpp
ckeshava: the constructor of `std::string_view` reads a `char*` pointer until null-termination. I observed some garbage values at the end of `key` when the second length parameter wasn't specified

ckeshava: Check out this thread: https://github.com/XRPLF/rippled/pull/4509#discussion_r1176760881

src/ripple/beast/core/LexicalCast.h
ckeshava: thanks for the explanation, fixed in 85e001dcb21516ad0d711e60af70fc9141f29162

src/ripple/beast/core/LexicalCast.h
ckeshava: fixed in https://github.com/XRPLF/rippled/commit/85e001dcb21516ad0d711e60af70fc9141f29162

src/ripple/overlay/impl/Cluster.cpp
ckeshava: thanks, fixed in https://github.com/XRPLF/rippled/commit/85e001dcb21516ad0d711e60af70fc9141f29162

src/ripple/overlay/impl/Handshake.cpp
thejohnfreeman: I think it might be implicitly convertible (which is why line 236 works), but this error is a missing specialization of `beast::detail::LexicalCast`. The language won't try other specializations to see if one might work.

src/ripple/overlay/impl/Handshake.cpp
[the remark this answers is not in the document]
thejohnfreeman: In my opinion, `Json::Value::asString()` should return a `std::string_view`. Let callers handle the case where they need to copy the data.

For `operator=`, when callers already have a `std::string`, passing `std::string const&` is best, basically free. Passing `std::string_view const&` requires constructing and destroying a temporary object, but it is small and allocated entirely on the stack, and it can be optimized away if the compiler knows enough about the callee. But when callers only have a `std::string_view`, which many will after this change, they have to construct and destroy a `std::string`, which requires a heap allocation and deallocation, and that is expensive. Further, that construction must be explicit in the code, which is inconvenient for developers to write and read. Thus, both overloads should be available.

src/ripple/overlay/impl/OverlayImpl.cpp
ckeshava: oh you mean: explicitly specify the type `std::string_view` instead of `auto` ? that could work

thejohnfreeman: What does `req.target().substr(...)` return? A `boost::string_view`? Is it implicitly convertible to `char*` or something? Where does the `char*` come in?

ckeshava: Yes, it returns a `boost::string_view`. I don't remember the internal mechanics exactly, I remember coming across the garbage values. So I decided to use the length parameter.

Now, I have used your suggestion as-is. I have updated my conan profile to use the `BOOST_BEAST_STD_STRING` macro. Ever since, the inter-conversion from `boost::string_view` to `std::string_view` has occured without a hitch.

fixed in [88412ea](https://github.com/XRPLF/rippled/pull/4509/commits/88412eaa6e86782489f073d724fc0acf9d6e6321)

src/ripple/overlay/impl/Handshake.cpp
ckeshava: yeah. i thought of adding one more template specialization for `LexicalCast<Out, boost::string_view>`. but it adds one more abstraction, I can as well convert the type at the point of function call instead of passing it through multiple abstractions

src/ripple/overlay/impl/Handshake.cpp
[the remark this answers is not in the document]
ckeshava: that makes sense, yes

ckeshava:
I believe I have addressed all of your comments @thejohnfreeman . I have updated the build instructions too.

I'm surprised that the Github CI passed all the tests. Without the `BOOST_BEAST_USE_STD_STRING_VIEW` macro, my compiler threw a few errors.

src/ripple/overlay/impl/Handshake.cpp
thejohnfreeman: Without going into detail, these should be zero-cost (at runtime) abstractions. I prefer to add the missing specialization in one place than a cast at every call site.

BUILD.md
```diff
@@ -104,6 +104,17 @@ then you will need to choose the `libstdc++11` ABI:
...
+The below commands ensure inter-operability between `boost::string_view` and 
+`std::string_view` types.
+```
+conan profile update 'conf.tools.build:cxxflags+=["-DBOOST_BEAST_USE_STD_STRING_VIEW"]' default
+conan profile update 'env.CXXFLAGS="-DBOOST_BEAST_USE_STD_STRING_VIEW"' default
+conan profile show default # sanity check: displays the default profile
+```
+Note: If you have other flags in the `conf.tools.build` or `env.CXXFLAGS 
+section`, make sure to retain and append the existing flags. 
+
```
thejohnfreeman, line 117: Do not use `<br>` in the text. Just use a new line. None of this text needs to be bolded either. I feel like this section fits better with the other compiler flag fixes in Troubleshooting, but it's fine to leave it here.

---

Ensure inter-operability between `boost::string_view` and `std::string_view` types:

```
conan profile update 'conf.tools.build:cxxflags+=["-DBOOST_BEAST_USE_STD_STRING_VIEW"]' default
conan profile update 'env.CXXFLAGS="-DBOOST_BEAST_USE_STD_STRING_VIEW"' default
```

If you have other flags in the `conf.tools.build` or `env.CXXFLAGS` sections, make sure to retain the existing flags and append the new ones. You can check them with:

```
conan profile show default
```

thejohnfreeman (changes requested):
The code looks good! I requested a change in the build instructions. Please merge the `develop` branch too.

src/ripple/overlay/impl/Handshake.cpp
ckeshava: okay, yeah I didn't think about the runtime costs. fixed in cee5ff89d714229d417e6bbfde6e40864df1d195

thejohnfreeman:
For your consideration: [Three reasons to pass std::string_view by value](https://quuxplusone.github.io/blog/2021/11/09/pass-string-view-by-value/)

thejohnfreeman (approved):

ckeshava:
@thejohnfreeman thanks for the article. I don't agree with the third point in the blog post, but the first two were pretty satisfactory.

The third example seems contrived, it doesn't feel correct when the `size` data member (which should be ideally `private`), is written into by an external function `byRef`

scottschurr:
@ckeshava, while working on this review I noticed that there are still some `std::string_view const&` parameters.  I _think_ that the article from @thejohnfreeman convinced you that passing `std::string_view` by value is the right choice?  If you're not convinced I'd be curious why.  But I would like to get that detail straight before proceeding with the code review.  Thanks.

ckeshava:
@scottschurr  Thanks for pointing it out, I have modified the relevant occurrences to use pass-by-value semantics in [cfa960f](https://github.com/XRPLF/rippled/pull/4509/commits/cfa960f7ef3c84e61bc67155ee0ebe5714e30677)

Do I need to `git add` the files inside the `.install/include/` directory? I'm not sure if they are auto-generated from within the source code or whether they need to be manually edited.

thejohnfreeman:
@ckeshava, no. I'm assuming you used Cupcake to install locally into the `.install` directory (its default choice). That is part of your personal workflow. That directory should not be added to the tracked files. If you prefer to ignore them, I suggest you add `/.install/` to your [global ignore file](https://gist.github.com/subfuzion/db7f57fff2fb6998a16c) (like I have).

scottschurr (commented):
It looks like you and @thejohnfreeman have done good work here.  I left just a few comments.  Let me know what you think.  Thanks.

intelliot:
- awaiting update or reply from @ckeshava

src/ripple/overlay/impl/PeerImp.cpp
```diff
@@ -368,7 +366,7 @@ PeerImp::cluster() const
     return static_cast<bool>(app_.cluster().member(publicKey_));
 }
 
-std::string
+std::string_view
```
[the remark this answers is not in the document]
ckeshava: Thanks Scott, your explanation of the `std::string_view` use cases is very clear, I have a better understanding now.

I'm returning a `std::string_view` of `headers_[<key-name>]`. `headers_` is a private non-static data member of the `PeerImp` class, it remains in-scope for as long as the `PeerImp` object is alive. But I haven't looked into the internal type of `boost::beast::http::fields const& headers_;` I don't know if it uses any heap-allocated memory.

I'm weakly in favor of retaining the `std::string_view`, but this is a tenuous argument.

I'm open to switching it back to a `std::string`

@thejohnfreeman did I miss any point here? what do you think?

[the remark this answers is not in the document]
thejohnfreeman: I think it's fine to return `std::string_view`. It's like a container returning a `const_reference`. It's generally guaranteed to be alive for the duration of the calling scope, and it's up to the caller to handle any copying necessary. But my preference is weak too. I think @scottschurr has stronger conviction than I do.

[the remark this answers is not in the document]
scottschurr: Thanks, @thejohnfreeman!  Yup, I actually do have some conviction on this one.  Let me explain why.

Returning a `string_view` from those places is an optimization that has some future risk associated with it and no measurement showing that the optimization buys us anything right now.  If someone can show me a measurable performance gain from returning `string_view`s in these locations then I'll reconsider.  Lacking that, I think returning `std::string` is the safest approach with no _measurable_ downside.

ckeshava:
@scottschurr 
Okay, I have addressed your comments, thanks for your feedback.

There are two more instances where a `boost::string_view` is returned from two functions: `extractIpAddrFromField` and `forwardedFor` in `Role.cpp`. This has been in the code for the past 3 years. 

Do you want me to change their return types to `std::string` instead? That would be a change in the behavior of the existing codebase (not only a refactor from `boost::string_view` to `std::string_view`)

thejohnfreeman (approved):

scottschurr (approved):
:+1: LGTM.  Thanks for battling through this.

Regarding...
> There are two more instances where a `boost::string_view` is returned from two functions: `extractIpAddrFromField` and `forwardedFor` in Role.cpp. This has been in the code for the past 3 years.
>
> Do you want me to change their return types to `std::string` instead? That would be a change in the behavior of the existing codebase (not only a refactor from `boost::string_view` to `std::string_view`)

Personally, I think this pull request should leave those two functions alone.  `extractIpAddrFromField` is `static`, so it is local to only that file.  It's probably safe, because is used in a very limited scope.

`forwardedFor` is not `static`.  However, since it has been in use for 3 years, changing it to return a `std::string` without a specific motivation would be a pessimization.  I don't see a good reason to introduce that pessimization at this time.

I hope that makes some sense.

ckeshava:
@scottschurr I think it would be better to change the return type of those functions into a `std::string`. I like your pointer safety rationale in the return types (above comment).

But I agree, another time, another PR

intelliot:
@ckeshava could you confirm whether this is ready to merge, from your perspective? Also, please suggest a commit message for the squashed commit.

ckeshava:
It looks good from my end. Suggested commit message:
`Replaces the usage of boost::string_view to std::string_view`

PR #4513, opened 2023-04-27
Adjust local fees when the job queue is overloaded:

nbougalis:
Commit d36024394de079f4613f260b465f9e380a26c54d accidentally broke the logic used to raise local fees in response to a job queue overload.

This commit simply moves a block of code "up" a level, so that it runs with every loop iteration--the original intention of the code--instead of only after the loop has ended.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### DIFF
--- a/src/ripple/app/main/LoadManager.cpp
+++ b/src/ripple/app/main/LoadManager.cpp
@@ -168,26 +168,26 @@ LoadManager::run()
                 LogicError("Deadlock detected");
             }
         }
-    }
 
-    bool change;
+        bool change;
 
-    if (app_.getJobQueue().isOverloaded())
-    {
-        JLOG(journal_.info()) << "Raising local fee (JQ overload): "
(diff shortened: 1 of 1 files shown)


nbougalis:
The intention of the code is to detect when the `JobQueue` is overloaded (an arguably imperfect proxy for the server itself working too hard) and to adjust the local fees; the intention is to use this as a way of "deterring" the submissions of additional transactions which, obviously, induce additional load.

This seemed like a straightforward thing to do, but then you asked your question.

Transaction submission load is only one part of the load that the server is exposed to; there's peer-to-peer stuff, there's non-submission RPC/websocket queries, etc, none of which pay fees. If I were to venture a guess, I would say that those other things impose far greater load that transaction submission.

So _maybe_ it's worth thinking about eliminating this adjustment.

I'd be curious to hear what @JoelKatz has to say.

PR #4516, opened 2023-05-02
Trivial: Rename `ServerHandlerImp` to `ServerHandler`

scottschurr:

There's only one implementation of `ServerHandler`.  Let's call it what it is.

As part of some historical hack, this was part of our code base:
```
namespace ripple {
using ServerHandler = ServerHandlerImp;
...
}
```
There was no other `ServerHandler` definition, despite the existence of a header suggesting that there was.

### Context of Change

I accidentally stumbled across this weirdness in our code base as part of a code review I was doing.  I almost asked for the change to be incorporated into that pull request.  But the problem was not introduced by that pull request, so that seemed an unfair burden.

Therefore I'm submitting this as a (trivial) stand-alone change.

The point of the change is two fold:
1. Let's not give something two names unless it serves a purpose.  Having two names, `ServerHandler` and `ServerHandlerImp`, for the same thing helped no one.
2. We should prefer simpler names over complicated names.  When collapsing to a single name, `ServerHandler` is simpler than `ServerHandlerImp`.  Let's use the simple name.

### Type of Change

Note that, because of the way git does diffs, the changes _look_ worse than they are.  There was an old (small) file called
```
src/ripple/rpc/ServerHandler.h
```
All of the contents of 
```
src/rippled/rpc/impl/ServerHandlerImp.h
```
were merged into `ServerHandler.h`.  So `ServerHandler.h` looks like it has a lot of changes.  But it's just that it absorbed the contents of `ServerHandlerImp.h`, which were the important parts.


drlongle:
I also came across this hack sometimes ago. Thanks for fixing it.

drlongle (approved):

src/ripple/rpc/ServerHandler.h
```diff
@@ -20,21 +20,200 @@
...
+    // A private type used to restrict access to the ServerHandler constructor.
+    struct ServerHandlerCreator
+    {
+        explicit ServerHandlerCreator() = default;
+    };
+
+    // Friend declaration that allows make_ServerHandler to access the
+    // private type that restricts access to the ServerHandler ctor.
+    friend std::unique_ptr<ServerHandler>
+    make_ServerHandler(
```
ckeshava, line 116: Can you please point me to resources about this `make_XYX` design pattern? I've read about the factory design pattern and I'm trying to understand it better.

1. Why not make this function a member of the class, instead of declaring it as a `friend`?
2. Why do we need this function? It appears to be calling a constructor of the class (without any additional preprocessing). Can't we make the said constructor public, thereby reducing the number of indirections.

I'm new to this design pattern, I'd like to learn more about it.

ckeshava (approved):

src/ripple/rpc/ServerHandler.h
scottschurr: > 1. Why not make this function a member of the class, instead of declaring it as a `friend`?

It could be a `static` member of the class.  I'm not sure why a free function was preferred over a `static` member in this case.  It's _probably_ for uniformity.  There are quite a number of other `make_` factory functions around the code base.  They are easier for developers to use and think about if they have similar interfaces and behavior.

> 2. Why do we need this function?

Factory functions can have a number of different purposes.  In this case it's because we want to prevent anyone from constructing a `ServerHandler` that is not held by a `std::unique_ptr<ServerHandler>`.  If we made the `ServerHandler` constructor `public`, then anyone could create one.  So, in this case, we give the factory the permission to allow the construction and trust that the factory will do the right thing with the `unique_ptr`.  And we don't have to trust very hard, since the implementation for `make_ServerHandler()` is in the implementation file for `ServerHandler`.

Factory functions can be used for other purposes as well.  For example, the Meyer's Singleton is (in effect) a factory function: https://en.wikipedia.org/wiki/Singleton_pattern

The (ancient) Gang of Four Design Patterns book has a chapter on Factory Method.

Factory functions really just solve the problem that sometimes a simple constructor can't do quite enough.  The factory function can do the extra work that is required.

src/ripple/rpc/ServerHandler.h
```diff
@@ -20,21 +20,200 @@
...
+    beast::insight::Counter rpc_requests_;
+    beast::insight::Event rpc_size_;
+    beast::insight::Event rpc_time_;
+    std::mutex mutex_;
+    std::condition_variable condition_;
+    bool stopped_{false};
+    std::map<std::reference_wrapper<Port const>, int> count_;
+
+    // A private type used to restrict access to the ServerHandler constructor.
+    struct ServerHandlerCreator
```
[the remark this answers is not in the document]
ckeshava: okay, that makes a lot of sense. Yeah, we can't impersonate types, understood.

src/ripple/rpc/ServerHandler.h
ckeshava: hmm okay. I'll check out these resources 👍

src/ripple/rpc/impl/ServerHandler.cpp
```diff
@@ -17,6 +17,8 @@
 */
 //==============================================================================
 
+#include <ripple/rpc/ServerHandler.h>
+
```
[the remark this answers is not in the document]
ckeshava: this is really helpful, thanks for letting me know about the convention 👍 
To paraphrase what I understood : It is good to include the header file of an implementation at the beginning of the implementation file. This helps in providing a self-contained list of header files for the implementation right?

I understood the dependency ordering of the #include's in the third paragraph from the end 👍

ckeshava (approved):

intelliot (approved):

intelliot:
Note: the current plan is to hold this until after 1.11 stable has been released

Toluphene (approved):

ckeshava (approved):

PR #4519, opened 2023-05-07
Improve checking of the path lengths during Payments

a-noni-mousse:

The code that does the check of the payment path lengths was sometimes executed but without an effect so change it to only  work when it matters and to not make not necessary copies of the path vectors.

### Context of Change

The code checks size of payment path lengths but it is only pertinente when executing again open view and doing the work when not open.
This is only start I think there are also other issue with the paths codes that need to be fixed.

### Type of Change

## Test Plan

Do the unit tests.


src/ripple/app/tx/impl/Payment.cpp
```diff
@@ -270,27 +270,17 @@ Payment::preclaim(PreclaimContext const& ctx)
...
-                if (path.size() > MaxPathLength)
-                {
-                    pathTooBig = true;
-                    break;
-                }
-
-        if (ctx.view.open() && pathTooBig)
+        if (paths.size() > MaxPathSize ||
+            std::any_of(paths.begin(), paths.end(), [](STPath const& path) {
+                return path.size() > MaxPathSize;
```
nbougalis, line 280: Bug: the check against individual paths should compare against `MaxPathLength` here.

nbougalis (changes requested):
One bug needs to be addressed; once it is, this gets a :+1: from me. Nice find.

nbougalis (approved):

gregtatcam (approved):
👍 LGTM

intelliot:
Suggested commit message:
```
refactor: improve checking of path lengths (#4519)

Improve the checking of the path lengths during Payments. Previously,
the code that did the check of the payment path lengths was sometimes
executed, but without any effect. This changes it to only check when it
matters, and to not make unnecessary copies of the path vectors.
```

@a-noni-mousse - please confirm (1) that this PR is ready to merge, in your opinion, and (2) the above commit message looks good (for the squashed commit when this PR is merged). Alternatively, you can suggest a different commit message.

intelliot:
@manojsdoshi to consider for the next beta - please squash and use the commit message [above](https://github.com/XRPLF/rippled/pull/4519#issuecomment-1678058728)

PR #4523, opened 2023-05-17
Forward Ledger Replay version 2 (FLRv2)

thejohnfreeman:
FLRv2 improves [sync time](https://xrpl.org/server-doesnt-sync.html#normal-syncing-behavior). More motivations for this change are elaborated in the documentation of [`MessageScheduler`](https://github.com/thejohnfreeman/rippled/blob/flr/src/ripple/peerclient/MessageScheduler.md).

First draft of FLR2. To test, checkout, build, and run the existing start-sync-stop test:

```
python bin/start_sync_stop.py
```

It expects `rippled` to be on your `PATH` and `rippled.cfg` to be in your current directory. See the options if those defaults do not work for you:

```
python bin/start_sync_stop.py --help
```

On my machine, syncing with mainnet. I can consistently sync after a recent (<10 minute) restart in 75 seconds[^1]. My first sync since yesterday took 120 seconds. A fresh sync takes around 330 seconds.

Note that the floor for these tests is not 0 seconds. It takes the server around 25 seconds to notice that it is behind and start the sync process (the first call to `LedgerGetter::get`).

[^1]: The test script will report this as 135 seconds because it does not subtract the 60 second duration that the server must report `full` before it is considered synced.


thejohnfreeman:
Just force-pushed a re-organization of the commits. Each builds in isolation and represents a separate phase of implementation.

1. A number of small improvements lumped together into one commit. Very small. 166 total line changes.
2. Add a Boolean `acquire` parameter to `LedgerMaster::getLedgerByHash` so that ledgers can be looked up in the in-memory cache and in the local nodestore without triggering a call to `InboundLedgers::acquire`. Tiny. 12 total line changes.
3. Add a type alias for `LedgerInfo` named `LedgerHeader` and concentrate its type declaration and deserializer into a header sharing its name. Very small. 150 total line changes.
4. Add an asynchronous, lock-free promise / future type, and some associated types (a "coroutine" that represents an asynchronous algorithm and a scheduler definition based on `JobQueue`). Moderately sized. 1079 new lines, of which 806 are code.
5. Add `MessageScheduler`. This is the latest edition of #4217. Large. 1500 line changes.
6. Add `PeerClient`. This is the layer on top of `MessageScheduler` demonstrating two different transmission strategies (one for a single object [ledger header], one for a full SHAMap [transaction set]). Returns futures.
7. Add `LedgerGetter`, which uses forward ledger replay and full copying (depending on what data is available locally).

There are a few problems left to solve:

- It doesn't pass CI. There's some levelization and build issues I haven't yet looked into. Not expecting much difficulty here because I can build and run on my machine.
- Syncing works, but at some point after syncing, while the server continues to replay ledgers (it replays every ledger whose digest it witnesses but hasn't built locally), an [assertion](https://github.com/thejohnfreeman/rippled/blob/flr/src/ripple/app/misc/NetworkOPs.cpp#L2888) is tripped in `NetworkOPs::pubLedger`. I have never before heard of the `AcceptedLedgerCache` and will have to investigate.
- `BasicSenderReceiver` should launch jobs in its `onReady` and `onSuccess` callbacks.
- I want to add some more metrics in `CopyLedger` to measure the speed of copying. I want to use that information to tune the parameters (max objects per request; number of channels per peer; number of connected peers; timeouts) affecting its performance.

bin/start_sync_stop.py
```diff
@@ -38,7 +38,8 @@
 DEFAULT_SYNC_DURATION = 60
 # Number of seconds between polls of state.
 DEFAULT_POLL_INTERVAL = 5
-SYNC_STATES = ('full', 'validating', 'proposing')
+# The 'proposing' state does not yet have a copy of the ledger.
```
ckeshava, line 41: By "ledger", are you referring to the ledger agreed upon by the consensus process? the last closed ledger?

src/ripple/basics/algorithm.h
```diff
@@ -115,6 +116,45 @@ remove_if_intersect_or_match(
...
+            ++first2;
+        }
+    }
+    return d_first;
+}
+
+/** Analog to std::move. */
+template <typename T>
+T
+copy(T const& t)
```
ckeshava, line 153: how is this different from the return-by-value performed by the compiler?

Is this copy function being used in a different commit? (I couldn't find any appropriate usages of this `copy` function in this commit.)

thejohnfreeman: It is used in later commits. It copies so that an lvalue reference can be passed to a function expecting an rvalue reference.

bin/start_sync_stop.py
ckeshava: ok

src/ripple/basics/algorithm.h
ckeshava: okay

bin/start_sync_stop.py
```diff
@@ -119,6 +127,7 @@ async def rippled(exe=DEFAULT_EXE, config_file=DEFAULT_CONFIGURATION_FILE):
 
         code = await process.wait()
         end = time.time()
+        logfile.close()
         logging.info(
             f'rippled stopped after {end - start:.1f} seconds with code {code}'
         )
```
ckeshava, line 133: I have a tangential question (unrelated to this PR) about the `loop()` function in this file.

How is the for-loop on line 185 terminating? I understand that we are dumping the logs in uniquely named files, but how will that function ever stop? Is it supposed to be manually terminated through Ctrl+C (triggering the `KeyboardInterrupt` on line 253) ?

src/ripple/overlay/impl/PeerSet.cpp
```diff
@@ -46,17 +44,15 @@ class PeerSetImpl : public PeerSet
     getPeerIds() const override;
 
 private:
-    // Used in this class for access to boost::asio::io_service and
-    // ripple::Overlay.
-    Application& app_;
     beast::Journal journal_;
```
ckeshava, line 47: it looks like `PeerSetImpl::journal_` is not being used

src/ripple/basics/Slice.h
```diff
@@ -64,6 +65,16 @@ class Slice
     {
     }
 
+    template <class T, class Alloc>
+    Slice(
+        std::vector<T, Alloc> const& v,
```
ckeshava, line 70: Why do we need the second template parameter `Alloc` ? I couldn't find instances where we explicitly specify a non-default Allocation parameter. 

In what cases does Class-Template-Argument-Deduction (CTAD) deduce non-default values for `Alloc` ?

src/ripple/protocol/PublicKey.h
```diff
@@ -131,11 +131,6 @@ class PublicKey
     }
 };
 
-/** Print the public key to a stream.
- */
-std::ostream&
-operator<<(std::ostream& os, PublicKey const& pk);
-
```
scottschurr, line 138: Thanks for removing unused stuff!

scottschurr (commented):
Just a review of the first commit.  More to come...

bin/start_sync_stop.py
thejohnfreeman: It is meant to be an infinite loop terminated by keyboard interrupt.

src/ripple/app/ledger/LedgerHistory.h
```diff
@@ -61,7 +61,7 @@ class LedgerHistory
 
     /** Retrieve a ledger given its hash */
     std::shared_ptr<Ledger const>
-    getLedgerByHash(LedgerHash const& ledgerHash);
+    getLedgerByHash(LedgerHash const& ledgerHash, bool acquire = true);
```
scottschurr, line 64: The implications of `bool acquire` are not at all obvious to the casual reader.  You should probably add documentation of this parameter to the doxygen.

src/ripple/app/ledger/LedgerMaster.h
```diff
@@ -194,7 +194,7 @@ class LedgerMaster : public AbstractFetchPackContainer
     getLedgerBySeq(std::uint32_t index);
 
     std::shared_ptr<Ledger const>
-    getLedgerByHash(uint256 const& hash);
+    getLedgerByHash(uint256 const& hash, bool acquire = true);
```
scottschurr, line 197: The implications of this method used to be pretty obvious.  Now, with the added `bool acquire` argument, it's not so obvious.  It would be good to add doxygen here to document the meaning of the `acquire` argument.

scottschurr (commented):
Just an intermediate review of the second commit.

src/ripple/ledger/impl/LedgerHeader.cpp
scottschurr, line 1: You know rippled's build constraints better than I do.  I notice that, with the addition of this file, there are now two distinct files named LedgerHeader.cpp.  There's this one in ledger.  There's another one in rpc/handlers.  I'm guessing that identically named files, in different directories, will not be a problem for any of our build tools.  But it seems prudent to check.

scottschurr (commented):
Review comments on the third commit.

src/ripple/basics/promises.h
```diff
@@ -0,0 +1,958 @@
...
+            {
+                std::rethrow_exception(p->error());
+            }
+            assert(state == FULFILLED);
+            return f(p->value());
+        });
+    }
+
+    decltype(auto)
+    reify() const
```
ckeshava, line 652: is there any specific reason why you chose this name for the function? I feel `resolve_promise` (something to that effect) is easier to understand

intelliot:
@HowardHinnant : please review the promises library added in https://github.com/XRPLF/rippled/pull/4523/commits/ab603db05c2fe9be5295bd9a2ebd12c84371284a

HowardHinnant:
Using 6806526511b26497ec03935515d8264ef554d5c9 on macOS 13.4.1

After removing `.conan/data` and all contents of `rippled/.build` I ran from `rippled/.build`:
```
conan install .. --output-folder . --build  --settings build_type=Release
```
This had no errors.  Then:
```
cmake -DCMAKE_CXX_FLAGS=-DBOOST_ASIO_HAS_STD_INVOKE_RESULT=1 -DCMAKE_TOOLCHAIN_FILE:FILEPATH=build/generators/conan_toolchain.cmake -DCMAKE_BUILD_TYPE=Release -Dassert=ON -Dunity=OFF  ..
```
I get this error:
```
CMake Error at CMakeLists.txt:99 (target_link_libraries):
  The link interface of target "ripple_libs" contains:

    nudb::core

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.
```

`.conan/data/nudub/2.0.8/_/_/` exists and is populated.

Please advise on how to build.

src/ripple/basics/algorithm.h
```diff
@@ -115,6 +116,37 @@ remove_if_intersect_or_match(
...
+        auto c = comp(*first1, *first2);
+        if (c < 0)
+            *d_first++ = *first1++;
+        else
+        {
+            if (c == 0)
+                ++first1;
+            ++first2;
+        }
+    }
```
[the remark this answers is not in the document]
thejohnfreeman: There's no requirement that commits in a PR are separately mergeable. That's merely a courtesy I tried to extend in this PR. The PR will be merged as one commit. These functions will be merged with the code that uses them, in a single commit.

src/ripple/app/ledger/LedgerMaster.h
thejohnfreeman: I don't agree that the implications used to be obvious if the added parameter whose default value preserves existing behavior does not have obvious implications, but I agree that more documentation will be helpful.

src/ripple/basics/Coroutine.h
```diff
@@ -0,0 +1,109 @@
...
+    Scheduler& jscheduler_;
+
+    explicit Coroutine(Scheduler& scheduler)
+        : output_(scheduler.pending<R>()), jscheduler_(scheduler)
+    {
+    }
+
+    virtual ~Coroutine()
+    {
+    }
```
HowardHinnant, line 46: Recommend:
```c++
    virtual ~Coroutine() = default;
    Coroutine(Coroutine const&) = delete;
    Coroutine& operator=(Coroutine const&) = delete;
```

src/ripple/rpc/handlers/Reservations.cpp
```diff
@@ -19,6 +19,7 @@
 
 #include <ripple/json/json_value.h>
 #include <ripple/net/RPCErr.h>
+#include <ripple/overlay/PeerReservationTable.h>
```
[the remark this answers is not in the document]
thejohnfreeman: No, it is needed on lines 80, 81, 111, and 127 because `ripple::PeerReservationTable` is an incomplete type otherwise.

src/ripple/basics/Coroutine.h
```diff
@@ -0,0 +1,109 @@
...
+};
+
+template <typename Coroutine, typename... Args>
+static auto
+start(Args&&... args)
+{
+    auto coroutine = new Coroutine(std::forward<Args>(args)...);
+    auto future = coroutine->start();
+    future->subscribe([coroutine](auto const&) { delete coroutine; });
```
HowardHinnant, line 103: This doesn't look exception safe.  What if `coroutine->start()` throws?

HowardHinnant (changes requested):
Leaving comments for a partial review at this point.

src/ripple/overlay/impl/ProtocolMessage.h
```diff
@@ -25,14 +25,17 @@
 #include <ripple/overlay/Message.h>
 #include <ripple/overlay/impl/ZeroCopyStream.h>
 #include <ripple/protocol/messages.h>
+#include <ripple.pb.h>
+
 #include <boost/asio/buffer.hpp>
 #include <boost/asio/buffers_iterator.hpp>
 #include <boost/system/error_code.hpp>
+#include <google/protobuf/message.h>
```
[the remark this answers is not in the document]
thejohnfreeman: It is "unneeded" only because it is transitively included elsewhere, but its exports are directly used on lines 286 and 321, so I prefer to leave it in.

src/ripple/overlay/PeerSet.h
```diff
@@ -86,6 +85,11 @@ class PeerSetBuilder
     build() = 0;
 };
 
+/**
+ * This constructor requires an Application because sometimes an Overlay is
+ * not built, and even when it is, it is built after the objects that want
+ * PeerSets. It is non-const because it may construct a new Journal.
+ */
 std::unique_ptr<PeerSetBuilder>
 make_PeerSetBuilder(Application& app);
```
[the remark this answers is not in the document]
thejohnfreeman: `PeerSetBuilder` was added as part of [FLR1](https://github.com/XRPLF/rippled/pull/3659). It is used only by `LedgerReplayer`. It could be replaced with calls to a new function that takes an `Application&` and returns a `std::unique_ptr<PeerSet>`, but that is out-of-scope for this PR.

src/ripple/sync/LedgerGetter.h
```diff
@@ -0,0 +1,97 @@
...
+    friend class ReplayLedgers;
+
+public:
+    LedgerGetter(Application& app);
+
+    LedgerFuturePtr
+    get(LedgerDigest const& digest);
+
+    LedgerFuturePtr
+    get(LedgerIdentifier const& id)
```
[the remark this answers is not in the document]
thejohnfreeman: This overload is not called right now.

src/ripple/basics/promises.h
thejohnfreeman: I think [reification](https://en.wikipedia.org/wiki/Reification_(computer_science)) is a well-known term in computer science. I do not like "resolve" here because it is too easily confused with "fulfill" and suggests transitioning the promise from an idle state to a terminal state, which is not what this method does. This method takes whatever value or exception existed in the fulfilling context and reconstitutes it in the current context, by either returning a value or throwing an exception. I'm open to calling it `reconstitute` or `deliver` or something, but not `resolve`.

src/ripple/basics/promises.h
```diff
@@ -0,0 +1,958 @@
...
+struct Value<void>
+{
+    void
+    get() const
+    {
+    }
+};
+
+template <typename T, typename C, typename V>
+union Storage
```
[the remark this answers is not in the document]
thejohnfreeman: `std::variant` cannot hold `void`. `Value` here is specialized for `void`.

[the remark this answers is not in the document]
HowardHinnant: The `variant` wouldn't hold `void`, it would hold `Value<void>`.  So I guess I'm not completely understanding.  Or perhaps my question was ambiguous.  I'm thinking about something along the lines of:
```c++
template <typename T, typename C, typename V>
class Storage
{
public:
    using callback_type = C;
    using value_type = V;
    using error_type = std::exception_ptr;
private:
   std::variant<
         std::vector<callback_type>
       , std::shared_ptr<T>
       , Value<value_type>
       , error_type
   > data_;
...
```

Or if the API of `std::variant` was sufficient then maybe expose the `variant` directly instead of wrap it in a class:

```c++
template <typename T, typename C, typename V>
using Storage = std::variant<
                               std::vector<C>
                             , std::shared_ptr<T>
                             , Value<V>
                             , std::exception_ptr
                            >;
```

Disclaimer:  I haven't prototyped either of these suggestions.  I just know that `union`s  are tricky to work with directly, and that was the motivation for introducing `variant` into the `std::lib`.

src/ripple/basics/utility.h
```diff
@@ -0,0 +1,35 @@
...
+
+namespace ripple {
+
+/** Analog to std::move. */
+template <typename T>
+T
+copy(T const& t)
+{
+    return t;
+}
```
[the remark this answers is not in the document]
thejohnfreeman: I've [responded](https://github.com/XRPLF/rippled/pull/4523#discussion_r1273974271) elsewhere that this function will be merged in a commit with the code that uses it, but I'd like to take this opportunity to have a separate discussion.

When constructing types, how might we minimize copying, in the general case? In other words, what rules or guidelines can we follow for writing optimal constructors by default?

Consider an example in the C language (approximately):

```c
struct C { ... };
void C_construct(C* this, ...);
struct B { ... };
void B_construct(B* this, ...);
struct A { B b_; C c_; }

A a;
B_construct(&a.b_, ...);
C_construct(&a.c_, ...);
```

This example directly constructs the members of `A`, in the scope that constructs `A`, without constructing temporaries of type `B` or `C`. Can this be accomplished in C++? I think the short answer is "not really". Let's see why. Let's start with the skeleton for C++:

```cpp
struct C { C(...) {...} };
struct B { B(...) {...} };
struct A {
  B b_;
  C c_;
  A(...) {...}
};
```

Assuming we have optimal constructors for `B` and `C`, how do we write the optimal constructor for `A`? (The answer should teach us how to write the constructors for `B` and `C`, fulfilling the assumption.) We must certainly start by constructing the members with initializers:

```cpp
A(...) : b_(...), c_(...) {}
```

What arguments do we pass to the initializers? If we pass `B const&` and `C const&`, then we are just copying the members, requiring us to construct throwaway temporaries of their types. Not what we want.

We might compose the parameter list for the optimal constructor of `A` by concatenating the parameter lists of the optimal constructors of `B` and `C`, but what if they have multiple optimal constructors with different parameter lists? It is easy to manifest if they have any default arguments. If they each have two constructors, then `A` must have four. We don't want a combinatoric explosion in constructors either.

If `A` is an aggregate type, we get direct construction of members for free:

```cpp
A a{ .b_{...}, .c_{...} };
```

Making `A` an aggregate type places severe [restrictions](https://en.cppreference.com/w/cpp/language/aggregate_initialization#Definitions) on `A` that are hard or impossible or undesirable to meet. For example, `A` cannot have any private non-static data members. We don't _have_ to write any constructors for `A`... but we don't _get_ to write any constructors for `A` either.

So while it is technically possible to directly construct the members of `A`, it is impractical. Now the question becomes: how close can we get? Well, we can call the move constructors for the members. A move still requires us to construct a temporary, but at least it is better than a copy. The rule becomes: **take a value for each member by rvalue-reference, and move-construct each member in its initializer.**

```cpp
A(B&& b, C&& c) : b_(std::move(b)), c_(std::move(c)) {}
```

This rule percolates. In a function `f` that constructs a value of type `A`, passing a `B&&` derived from a parameter `b`, then `b` should have type `B&&` too. It percolates until you get to some point in the call graph that you can't easily pass `B&&`. Maybe you have a `B&` that you need to live past the call, or maybe you reach an API boundary and don't want to disturb its parameter list. In those cases you essentially have a `B const&` but need to pass a `B&&`. Then it'd be handy to have a function like `B copy(B const&)` that constructs a prvalue to pass down the call graph.

Does this make sense? Thoughts?

HowardHinnant (approved):

intelliot:
- attempting to replicate potential race condition in promises
  - this feature is ready after that is resolved
  - Folly has a promises implementation which has nearly the same API
    - may be difficult to get Folly to compile with Conan
    - John could write a Conan recipe for it
    - we like the idea of adding Folly but it's a big deal so should be discussed broadly
    - Abseil promises are not non-blocking

intelliot:
- @pwang200 intends to review within the next ~3 weeks
- has some questions about where the protocol contributes to the speed up

intelliot:
note: @thejohnfreeman expects this will be ready for review & merge around the end of this month

src/ripple/sync/ObjectRequester.cpp
```diff
@@ -53,11 +53,11 @@ ObjectRequester::deserialize(ObjectDigest const& digest, Slice const& slice)
     else if (prefix == HashPrefix::innerNode)
     {
         assert(slice.size() == 4 + SHAMapInnerNode::branchFactor * 32);
-        if (isFullBelow(digest))
-        {
-            return;
-        }
-        auto checkpoint = requested_;
+        // There's an opportunity here to exit early if the tree rooted at
```
mtrippled, line 56: When I said this had to do with online_delete, I meant the "generation" variable had to do with online_delete.

The cache needs to stay or performance will likely degrade drastically, unless you track this separately.

mtrippled: Then again, it looks like you're not removing it outright--just from this new code. So it should be OK to do this.

intelliot:
This is ready for testing. @legleux will test. 

MessageScheduler docs will be updated soon.

src/ripple/basics/promises.h
```diff
@@ -33,8 +34,10 @@
 
 namespace ripple {
 
+#define TO_STRING0(x) #x
+#define TO_STRING(x) TO_STRING0(x)
 #define ASSERT_OP(op, x, y) \
-    if (x op y) { std::cerr << #x " == " << x << " " #op " " << y << " == " #y; }
+    if (x op y) { std::cerr << __FILE__ ":" TO_STRING(__LINE__) ": " #x " == " << x << " " #op " " << y << " == " #y << std::endl; std::abort(); }
 #define ASSERT_EQ(x, y) ASSERT_OP(!=, x, y)
 #define ASSERT_NE(x, y) ASSERT_OP(==, x, y)
```
ximinez, line 42: Having `ASSERT_EQ` use the `!=` operator is probably going to cause some confusion and maintainability issues down the line. Maybe you've already tried this and it didn't work, but if not, consider:
```
#define TO_STRING0(x) #x
#define TO_STRING(x) TO_STRING0(x)
#define ASSERT_OP(op, x, y) \
    if (!(x op y)) { std::cerr << __FILE__ ":" TO_STRING(__LINE__) ": " #x " == " << x << " " #op " " << y << " == " #y << std::endl; std::abort(); }
#define ASSERT_EQ(x, y) ASSERT_OP(==, x, y)
#define ASSERT_NE(x, y) ASSERT_OP(!=, x, y)
```

intelliot:
Internal tracker: RPFC-81

legleux:
There's been many changes since this branch was crashing and/or hitting asserts.
I will test again soon.

intelliot:
@thejohnfreeman please comment here with any updates / ideas for moving this forward

godexsoft:
We recently merged a refactor to `develop` that enables clang-tidy's `readability-identifier-naming`. Your branch now has heavy conflicts that are largely mechanical. Below is a workflow that aligns your branch's naming with develop _before_ merging, which should minimize the merge conflicts.

## One-time setup

If you don't already have clang-tidy working in your env, on macOS:

```bash
brew install llvm@21
# Follow brew's hint to put $(brew --prefix llvm@21)/bin on PATH so run-clang-tidy is found.
```

## Workflow on your branch (before merging develop)

**1. Grab the new `.clang-tidy` from develop without pulling anything else.** Sync your fork on GitHub first, then:

```bash
git remote -v   # should show 'upstream' among others; if not:
# git remote set-url upstream git@github.com:XRPLF/rippled.git
git fetch upstream
git checkout upstream/develop -- .clang-tidy
```

**2. Reconfigure conan/cmake so `compile_commands.json` is fresh.**

**3. Apply renames for the files modified in your PR:**

```bash
git diff --name-only $(git merge-base HEAD upstream/develop) HEAD \
  | grep -E '\.(cpp|h|hpp|ipp)$' \
  | xargs run-clang-tidy -p build -fix -allow-no-checks
# or -p .build, or whatever your build dir is called
```

**4. Build + test, then commit as a single dedicated commit:**

```bash
cmake --build build -j8
git commit -am "refactor: Align identifier naming with develop"
```

**5. Now merge develop:**

```bash
git merge upstream/develop
```

## Extra

Run clang-tidy once more after the merge to catch any stragglers introduced from develop's side:

```bash
run-clang-tidy -p build -fix -allow-no-checks src tests
# or -p .build, or whatever your build dir is called
```

PR #4526, opened 2023-05-18
Order of init fix

seelabs:
Note: Only the top commit needs to be reviewed. After review the patch should be cherry picked onto the AMM PR. This PR should not be merged directly (it is marked as "draft" for that reason)

Global variables in different TUs are initialized in an undefined order.
At least one global variable was accessing a global switchover variable.
This caused the switchover variable to be initialized in an incorrect
state.

Since the switchover is always explicitly set before transaction
processing, this bug can not effect transaction processing, but could
effect unit tests (and potentially the value of some global variables).
Note: at the time of this patch the offending bug is not yet in production.

### Type of Change


seelabs:
Closing this. It can be based on top of develop. Will open a non-draft one shortly.

PR #4527, opened 2023-05-18
Ensure that switchover vars are initialized before use:

seelabs:
Global variables in different TUs are initialized in an undefined order. At least one global variable was accessing a global switchover variable. This caused the switchover variable to be initialized in an incorrect state.

Since the switchover is always explicitly set before transaction processing, this bug can not effect transaction processing, but could effect unit tests (and potentially the value of some global variables). Note: at the time of this patch the offending bug is not yet in production.

### Type of Change

### DIFF
--- a/src/ripple/basics/IOUAmount.h
+++ b/src/ripple/basics/IOUAmount.h
@@ -186,7 +186,14 @@ mulRatio(
     std::uint32_t den,
     bool roundUp);
 
-extern LocalValue<bool> stNumberSwitchover;
+// Since many uses of the number class do not have access to a ledger,
+// getSTNumberSwitchover needs to be globally accessible.
+
+bool
+getSTNumberSwitchover();
+
+void
+setSTNumberSwitchover(bool v);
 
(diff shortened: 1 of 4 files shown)


HowardHinnant (approved):
Nice job!

gregtatcam (approved):
:+1: LGTM

HowardHinnant:
I think using `thread_local` is something we should strongly consider, but definitely not in this PR.

We use `LocalValue` in a few places and we should consider a PR which moves all of them to `thread_local` (and does nothing else).  This would be a good move to reduce our technical debt.

But #4527 should remain just about fixing the bug without introducing further complications which would risk introducing a bug.

seelabs:
@a-noni-mousse @HowardHinnant LocalValue is meant to work with coroutines and thread_local isn't going to play well with co-routines. Having said that, I'm not sure any co-routines are actually using `LocalValue`. If we audit the code and it's not used we _may_ be able to replace it, but even then we'd have to make sure it was never mistakenly used with a co-routine.

PR #4533, opened 2023-05-22
Move faulty assert

ximinez:

Does what it says on the tin. One of the asserts introduced in #4319 (66627b26cfae8e1c902546f0778dd9b013aedc5a) was put in the wrong place, but wasn't triggered unless shards were configured.

This pull request, if merged, moves the assert to the right place, and updates it to ensure correctness.

Note that normally asserts are only checked in debug builds, so release packages will not be affected.

### Type of Change

## Test Plan

* Run a debug build with shards configured.
* Wait until rippled downloads some shards.
* May need to restart once the shard is downloaded, (I'm not sure, and haven't reproduced yet.)

Before this fix:
* rippled will assert in Shard.cpp line 693.

After this fix:
* rippled will not assert.


drlongle (approved):

drlongle:
LGTM!

PR #4536, opened 2023-05-24
fix node size estimation

dangell7:

This patch fixes a bug in the NODE_SIZE autodetection feature in the Config.cpp file. Specifically, this patch corrects the calculation for the total amount of RAM available, which was previously returned in bytes, but is now being returned in units of the system's memory unit. Additionally, the patch adjusts the node size based on the number of available hardware threads of execution.

### Type of Change

### DIFF
--- a/src/ripple/core/impl/Config.cpp
+++ b/src/ripple/core/impl/Config.cpp
@@ -68,10 +68,8 @@ namespace detail {
 [[nodiscard]] std::uint64_t
 getMemorySize()
 {
-    struct sysinfo si;
-
-    if (sysinfo(&si) == 0)
-        return static_cast<std::uint64_t>(si.totalram);
+    if (struct sysinfo si; sysinfo(&si) == 0)
+        return static_cast<std::uint64_t>(si.totalram) * si.mem_unit;
 
     return 0;
 }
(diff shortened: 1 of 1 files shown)


src/ripple/core/impl/Config.cpp
```diff
@@ -290,22 +289,18 @@ Config::setupControl(bool bQuiet, bool bSilent, bool bStandalone)
...
-        if (auto const hc = std::thread::hardware_concurrency())
-        {
-            if (hc == 1)
-                NODE_SIZE = 0;
-
-            if (hc < 4)
-                NODE_SIZE = std::min<std::size_t>(NODE_SIZE, 1);
-        }
+        if (auto const hc = std::thread::hardware_concurrency(); hc != 0)
+            NODE_SIZE = std::min<std::size_t>(hc / 2, 4);
```
nbougalis, line 303: I think you want to replace the `4` with `NODE_SIZE` here.

nbougalis (changes requested):
Looks good modulo the one minor change I listed. Once that's done, this gets a :+1: from me.

nbougalis (approved):

nbougalis:
This should be included in the next release, as it fixes a bug.

intelliot:
@manojsdoshi to confirm whether this is ok to include in 1.11

Standard procedure is to get 2 approvals, unless the PR is obviously trivial (and in that case, it needs the `Trivial` label applied). Can you recommend another reviewer?

intelliot:
Perhaps @sophiax851 or @ChronusZ can take a look?

nbougalis:
This is a bug fix that impacts servers and, imo, is pretty trivial. Maybe @HowardHinnant can take a quick look?

intelliot:
QA confirmed this is ok to include in 1.11 (if approved by code review)

HowardHinnant (approved):
Looks good to me.

Disclaimer:  Part of this change is on a Linux-only branch which I'm not able to test.  However I know @nbougalis is using Linux and he's approved, so I think we're covered.

intelliot:
Note: This was an older bug (i.e. not a bug introduced in 1.11 nor in an RC - release candidate). It's okay this time, but in the future, QA would like to avoid making non-critical (non-blocker) changes within 2 weeks of a release.

Issue #4303, opened 2022-09-15
“ledger_entry” without default parameters not returning "invalidParams" error.


mounikakun:
## Issue Description
“ledger_entry” with one or more fields are specified incorrectly, or one or more required fields are missing i.e., without index, account_root , directory, offer, ripple_state, check, escrow, payment_channel, deposit_preauth, ticket should result in invalidParams error.

## Steps to Reproduce
{
    "method": "ledger_entry",
    "params": [
        {
            "ledger_index": "validated"
        }
    ]
}

## Expected Result
invalidParams

## Actual Result
unknownOption

arihantkothari:
should `malformedRequest`, `malformedAddress`, `malformedCurrency` also be `invalidParams` ?

arihantkothari:
What is wrong with throwing `unknowOption` error ? 

Accord to [link](https://xrpl.org/ledger_entry.html#possible-errors),
unknownOption - The fields provided in the request did not match any of the expected request formats.
invalidParams - One or more fields are specified incorrectly, or one or more required fields are missing.

PR #4552, opened 2023-05-30
Fix: returning invalidParams instead of unknownOption error code in ledger_entry (#4303)

arihantkothari:

Fix for #4303; Replacing `unknownOption` error with `invalidParams` in `ledger_entry` in API `v2`

### Type of Change

- [ x] Tests (Updated the existing tests)

## Before / After

Test Input Payload 1:
`{
   "method":"ledger_entry",
   "params":[
      {
         "ledger_index":"validated"
      }
   ]
}`

Before:
`{
   "result":{
      "error":"unknownOption",
      "ledger_hash":"CA38067ED8FCDCFDE8CAB00A7F83BD90FE0B75D4906D42CCA13EE02B15EC1D10",
      "ledger_index":29063797,
      "request":{
         "command":"ledger_entry",
         "ledger_index":"validated"
      },
      "status":"error",
      "validated":true
   }
}`

After: remains same as before.

Test Input Payload 2:
`{ "method":"ledger_entry", "params":[ { "api_version": 2, "ledger_index":"validated" } ] }`

Before: Not supported

After:
`{"result":{"error":"invalidParams","ledger_hash":"2532DD58AF4FDC24EAD3F713280C91781C9A0C9CA1DB51EF9BA8946B1D3C1B4E","ledger_index":29343470,"request":{"api_version":2,"command":"ledger_entry","ledger_index":"validated"},"status":"error","validated":true}}`


shawnxie999:
wouldn't this be a small breaking change?

arihantkothari:
How is it that so ? I have updated the existing unit test as well. I was told that we do not want unknownOption at all but the API currently returns it. I might be missing something, do you think I should be updating something else also to make it a non-breaking change if it is a breaking change ?

shawnxie999:
Take it with a grain of salt, but it seems `unknownOption` have existed for quite some while for this API. I think this change would break the error handling for existing applications if the error message is suddenly changed from `unknownOption` to `invalidParams`

shawnxie999:
But I think it would be okay if `invalidParams` is the intended behavior. I would check with @ckniffen if it's affecting anything.

arihantkothari:
Do you think we should be deprecating the `unknownOption` error code ?

arihantkothari:
Thanks @intelliot ! Here are the results of the changes made:


Input:
`{ "method":"ledger_entry", "params":[ { "api_version": 2, "ledger_index":"validated" } ] }`

Output:
`{"result":{"error":"invalidParams","ledger_hash":"2532DD58AF4FDC24EAD3F713280C91781C9A0C9CA1DB51EF9BA8946B1D3C1B4E","ledger_index":29343470,"request":{"api_version":2,"command":"ledger_entry","ledger_index":"validated"},"status":"error","validated":true}}`

Input:
`{ "method":"ledger_entry", "params":[ { "ledger_index":"validated" } ] }`

Output:
`{"result":{"error":"unknownOption","ledger_hash":"BB17A4E96BB0D8411B36AA003AF6E25DC71B3B5E29E16C2FC2F2F88A8C17251B","ledger_index":29343475,"request":{"command":"ledger_entry","ledger_index":"validated"},"status":"error","validated":true}}`

arihantkothari:
Currently, a bunch of `account_info` unit tests fail `(rippled/src/test/rpc/AccountInfo_test.cpp)` because of setting `ApiMaximumSupportedVersion` to `2`. It seems there exists version specific tests in that and by default all tests are run on the max supported version. I am wondering if I should start a new issue regarding this or do the changes here ? I think starting a new issue might make more sense.

intelliot:
Yes, a new issue (or better - a PR) makes sense.

src/test/rpc/LedgerRPC_test.cpp
```diff
@@ -1222,21 +1222,40 @@ class LedgerRPC_test : public beast::unit_test::suite
...
+
+        std::string const ledgerHash{to_string(env.closed()->info().hash)};
+
+        // "features" is not an option supported by ledger_entry.
+        Json::Value jvParams;
+        jvParams[jss::features] = ledgerHash;
+        jvParams[jss::ledger_hash] = ledgerHash;
+        Json::Value const jrr =
+            env.rpc("json", "ledger_entry", to_string(jvParams))[jss::result];
+        checkErrorValue(jrr, "invalidParams", "");
```
thejohnfreeman, line 1256: Related to https://github.com/XRPLF/rippled/pull/4568#discussion_r1227054960, somehow this method tests version 2 even though (a) `RPC::apiVersionIfUnspecified` is 1 and (b) the request object is missing field `api_version`. How?

thejohnfreeman (changes requested):

src/test/rpc/LedgerRPC_test.cpp
arihantkothari: a,b) It only happens in unit-tests (by default it choses max api version). Actual rippled uses v1 by default (if nothing is specified) as you'd expect. Still trying to figure out why this happens unit tests

arihantkothari: Looks like this is what's causing the problem: https://github.com/XRPLF/rippled/blob/beb0904a322bcaef777bedfe17c92c2d2f56e916/src/ripple/net/impl/RPCCall.cpp#L1474

src/ripple/rpc/impl/RPCHelpers.h
```diff
@@ -241,7 +241,7 @@ extern beast::SemanticVersion const lastVersion;
 constexpr unsigned int apiInvalidVersion = 0;
 constexpr unsigned int apiVersionIfUnspecified = 1;
 constexpr unsigned int apiMinimumSupportedVersion = 1;
-constexpr unsigned int apiMaximumSupportedVersion = 1;
+constexpr unsigned int apiMaximumSupportedVersion = 2;
 constexpr unsigned int apiBetaVersion = 2;
```
ximinez, line 245: As I wrote over [on #4571](https://github.com/XRPLF/rippled/pull/4571#discussion_r1228663775), once a version of `rippled` is released with `apiMaximumSupportedVersion` is set to 2, then we are saying that API v2 is **Done**. Once that happens we can not legitimately make any further changes to API version 2. To the best of my knowledge, API version 2 is still in beta, under active development, and still changing. 

Now, if you (and #4571 and #4568) are saying that version 2 is done, let's talk about version 2 being done.

If you want a node to properly handle API v2 requests, the correct way to do that, for now, is with the `[beta_rpc_api]` config option. And if you want to do it in unit tests, set it up in the config similar to how it's done in https://github.com/XRPLF/rippled/blob/develop/src/test/rpc/AccountInfo_test.cpp#L209-L212, or just by setting `cfg.BETA_RPC_API=true;` directly.

ximinez (changes requested):

src/ripple/rpc/impl/RPCHelpers.h
arihantkothari: Thanks for the feedback. I'll fix this.

ximinez: Thanks. Please see also #4573.

arihantkothari: Hi @ximinez , [this](https://github.com/XRPLF/rippled/blob/develop/src/test/rpc/AccountInfo_test.cpp#L209-L222) seems a bit problematic. In the test, `api_version 2` is specified in the payload which means that when let's say `api_version 3` is released (ofc, not soon but whenever it does) - the test will no longer run on `api_version 3` unless this particular behaviour changes and we rewrite the test for it. However ideally, would we not want the new test to run for all versions >= 2 unless the particular behaviour is changed for the next versions ofc ? I may be wrong, but I just wanted to clarify.

arihantkothari: Also now since we are using beta_version and not changing the max supported version, it does not matter for this PR. But I found a weird behaviour when I changed `apiMaximumSupportedVersion ` version to 2. It immediately failed a bunch of tests because the unit-tests started running on `apiMaximumSupportedVersion` by default which is inconsistent with [this](https://github.com/XRPLF/rippled/blob/beb0904a322bcaef777bedfe17c92c2d2f56e916/src/ripple/rpc/impl/RPCHelpers.h#L219-L239). I found [this](https://github.com/XRPLF/rippled/pull/4552/files#r1227259812). It is not important right now probably, but this might be a bug and we might need to fix this when `api_version 2` is done?

thejohnfreeman: This does not give me a good impression of our versioning scheme. Makes me think this is why version 2 has not released.

Do we have guidelines that answer these questions?

- How should we write code that is conditional on version and robust in the face of future versions?
- What do we call the set of versions that is tested? What do we call the set of versions that is enabled during operation?
- How do we configure which versions are in these sets?
- How can we ensure that all to-be-tested versions are actually tested? How can we ensure that code that is not conditional on version maintains its behavior across versions?

intelliot: It's the same as the way APIs are versioned in other API servers, unrelated to rippled.

You write code that is conditional by using conditions. When future versions are created, you have to update the code. Where updates are required depends on the changes being made in the API. The name for the set of versions that is tested is decided the same way as all other names in the code: propose something, and let other people propose alternatives. The same applies for the set of versions that are enabled during operation: you give it a name that makes sense to you, and is the best name you can think of, given limited time and knowledge. Similarly, you propose code that configures which versions are in these sets. And you have to write code that ensures all available versions are actually tested. The purpose of unit tests is to catch breakages when the API version changes, so it is good (and it is a sign of unit tests serving their intended purpose) when tests break in the face of breaking changes in the API. At that point, you update the tests so that they test the previous API version, and the current one.

If you have any other answers to your questions, go ahead and and share them.

thejohnfreeman: @ximinez gave some good answers in our internal Slack:

> 2. You mean like a name? I have no idea. As I said above, the tests should be over minimum to beta, and the set that's enabled during operation (by default) is minimum to maximum.
> 3. They are defined by the variables in `RPCHelpers.h`.
> 4. My thinking is that once [#4573](https://github.com/XRPLF/rippled/pull/4573) is merged, any tests that care about versions should loop from `apiMinimumSupportedVersion` to `apiBetaVersion` and run the tests across all versions. The tests would then have to be written with conditional code to handle the expected results across all versions. This is different than how a lot of the existing tests are written where there's one function for v1 and one for v2, which I agree with you is not sustainable.

He did not answer the first question in my opinion. I think it should be done with zero version _equalities_, only version _inequalities_. In fact, I think we can prescribe a pattern to be strictly-enforced:

```cpp
if (version < FEATURE_X_VERSION) {
  // old behavior
} else {
  // new behavior
}

if (version >= FEATURE_X_VERSION) {
  // new behavior where old behavior was just "do nothing"
}

if (version < FEATURE_X_VERSION) {
  // old behavior where new behavior is "stop doing that"
}
```

ximinez (approved):

src/test/rpc/LedgerRPC_test.cpp
```diff
@@ -1222,7 +1222,7 @@ class LedgerRPC_test : public beast::unit_test::suite
     void
     testLedgerEntryUnknownOption()
```
thejohnfreeman, line 1223: Please:

- Add a parameter for `api_version`.
- Call this method in a loop over versions from `apiMinimumSupportedVersion` to `apiBetaVersion`.
- Remove the other method (`testLedgerEntryInvalidParams`).
- Make the assertion on the error value conditional on the `api_version` parameter, using condition `< 2` for the old behavior and `else` for the new behavior.

src/ripple/rpc/handlers/LedgerEntry.cpp
```diff
@@ -358,7 +358,12 @@ doLedgerEntry(RPC::JsonContext& context)
             }
         }
         else
-            jvResult[jss::error] = "unknownOption";
+        {
+            if (context.apiVersion == 1)
```
thejohnfreeman, line 362: Please change condition to `< 2`.

thejohnfreeman (changes requested):

src/test/rpc/LedgerRPC_test.cpp
arihantkothari: Thanks @thejohnfreeman!

thejohnfreeman (approved):

src/test/rpc/LedgerRPC_test.cpp
```diff
@@ -1220,21 +1221,28 @@ class LedgerRPC_test : public beast::unit_test::suite
...
         // "features" is not an option supported by ledger_entry.
         Json::Value jvParams;
+        jvParams[jss::api_version] = apiVersion;
         jvParams[jss::features] = ledgerHash;
         jvParams[jss::ledger_hash] = ledgerHash;
         Json::Value const jrr =
             env.rpc("json", "ledger_entry", to_string(jvParams))[jss::result];
-        checkErrorValue(jrr, "unknownOption", "");
+
+        if (apiVersion < 2)
```
godexsoft, line 1242: Consider 2u (unsigned int) instead of 2 (signed int)

src/test/rpc/LedgerRPC_test.cpp
```diff
@@ -1220,21 +1221,28 @@ class LedgerRPC_test : public beast::unit_test::suite
     }
 
     void
-    testLedgerEntryUnknownOption()
+    testLedgerEntryInvalidParams(unsigned int apiVersion)
```
godexsoft, line 1224: I'm not sure what the convention is in rippled but i'd use uint32_t or perhaps uint16_t. Fine as it is if others are happy though

src/test/rpc/LedgerRPC_test.cpp
```diff
@@ -1732,11 +1740,18 @@ class LedgerRPC_test : public beast::unit_test::suite
...
         testLedgerEntryRippleState();
         testLedgerEntryTicket();
-        testLedgerEntryUnknownOption();
         testLookupLedger();
         testNoQueue();
         testQueue();
         testLedgerAccountsOption();
+
+        // version specific tests
+        for (auto testVersion = RPC::apiMinimumSupportedVersion;
```
godexsoft, line 1749: It could be beneficial to extract this into some sort of helper function (maybe in the next pr) as i imagine this will be copy-pasted all over the place very soon. Better do it sooner than later. I'd go with a variadic template accepting callables (functions expecting api version as their input).

src/test/rpc/LedgerRPC_test.cpp
```diff
@@ -1732,11 +1740,18 @@ class LedgerRPC_test : public beast::unit_test::suite
...
+             testVersion <= RPC::apiBetaVersion;
```
godexsoft, line 1750: Is there no maximum value, only apiBetaVersion? (i'm on a phone, can't easily check rn).

godexsoft (approved):
Leaving a few suggestions. Overall looks good 👍🚀

src/test/rpc/LedgerRPC_test.cpp
thejohnfreeman: `apiBetaVersion` is the maximum value.

arihantkothari: Yea, from my understanding `apiBetaVersion` >= `apiMaximumSupportedVersion ` is a true statement. However I've seen instances where we've done `std::max(apiBetaVersion, apiMaximumSupportedVersion)` (https://github.com/XRPLF/rippled/blob/11e914fbe97184e0566085948b04942a8301ac36/src/test/rpc/Version_test.cpp#L86)

src/test/rpc/LedgerRPC_test.cpp
arihantkothari: Good catch. thank you!

src/test/rpc/LedgerRPC_test.cpp
arihantkothari: you are right, it will get copied. I'll try this out in the next PR. Thank you for pointing it out.

src/test/rpc/LedgerRPC_test.cpp
```diff
@@ -1724,11 +1732,18 @@ class LedgerRPC_test : public beast::unit_test::suite
...
+             ++testVersion)
+        {
+            testLedgerEntryInvalidParams(testVersion);
+        }
```
shawnxie999, line 1746: nit extra bracket

shawnxie999 (approved):

src/test/rpc/LedgerRPC_test.cpp
arihantkothari: Kept it for readability. It was becoming hard to read after the indentation caused by clang-format if I removed the brackets.

intelliot:
Suggested commit message:
```
APIv2(ledger_entry): return "invalidParams" when fields missing (#4552)

Improve error handling for ledger_entry by returning an "invalidParams"
error when one or more request fields are specified incorrectly, or one
or more required fields are missing.

For example, if none of of the following fields is provided, then the
API should return an invalidParams error:
* index, account_root, directory, offer, ripple_state, check, escrow,
  payment_channel, deposit_preauth, ticket

Prior to this commit, the API returned an "unknownOption" error instead.
Since the error was actually due to invalid parameters, rather than
unknown options, this error was misleading.

Since this is an API breaking change, the "invalidParams" error is only
returned for requests using api_version: 2 and above. To maintain
backward compatibility, the "unknownOption" error is still returned for
api_version: 1.

Related: #4573

Fix #4303
```

intelliot:
@arihantkothari can you confirm the commit message above looks correct to you?

arihantkothari:
> @arihantkothari can you confirm the commit message above looks correct to you?

Hi @intelliot , Sorry I was offline last few days. Yes, this looks good. Do you want me to `git ... --allow-empty` and put this commit message ? Thanks,

src/test/rpc/LedgerRPC_test.cpp
arihantkothari: any idea what would be the most appropriate file to put this helper in ? `RPCHelpers` ? (cc: @ximinez )

ximinez: Since it's only going to be used in tests, it should be under `src/test`. Env.h already has a lot of stuff, including the functions to make RPC calls, so maybe a static member or standalone function in there... Alternately, just create a new RPCUtils.h under src/test/jtx.

Another alternative if you feel like rewriting this test again is to create a member function of `Env`, so you can loop over each test case individually without the overhead of creating a new `Env` every time.

I don't know for sure, though.

PR #4294, opened 2022-09-06
Introduce AMM support (XLS-30d)

gregtatcam:
## High Level Overview of Change

[XLS-30d Specification](https://github.com/XRPLF/XRPL-Standards/discussions/78)

Adds AMM core functionality InstanceCreate, Deposit, Withdraw, Governance, Auctioning
and payment engine integration.

### Context of Change

Adds Transactors for InstanceCreate, Deposit, Withdraw, Governance, and Auctioning.
Adds AMM Root Account, trustline for each IOU AMM token, trustline to track Liquidity Provider Tokens (LPT),
and `ltAMM` object to track fee votes, auction slot bids, AMM tokens pair, total outstanding tokens balance, and
`AMMID` to AMM `RootAccountID` mapping. 
New classes are added to facilitate AMM integration into the payment engine. `BookStep` uses these classes 
to infer if AMM liquidity can be consumed. 
A new RPC method `amm_info` is added to fetch the pool and 
LPT balances.
A new Number class is added to facilitate AMM formula implementation. IOUAmount and STAmount use
Number arithmetic. A `fixUniversalNumber` amendment is added.
`featureAMM`, `fixUniversalNumber`, and `featureFlowCross` amendments are required to support AMM.

### Type of Change

- [ x] New feature (non-breaking change which adds functionality)

## Test Plan

AMM unit tests are added for all core functionality features.

PR #4553, opened 2023-05-31
XLS-39 Clawback:

shawnxie999:

Introduces:
- `AccountRoot` flag: `lsfAllowClawback`
-  New `Clawback` transaction

More info on clawback spec: https://github.com/XRPLF/XRPL-Standards/tree/master/XLS-39d-clawback

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### DIFF
--- a/src/ripple/app/tx/impl/Clawback.cpp
+++ b/src/ripple/app/tx/impl/Clawback.cpp
@@ -21,11 +21,10 @@
 #include <ripple/ledger/Directory.h>
 #include <ripple/ledger/View.h>
 #include <ripple/protocol/Feature.h>
+#include <ripple/protocol/Indexes.h>
 #include <ripple/protocol/Protocol.h>
 #include <ripple/protocol/TxFlags.h>
 #include <ripple/protocol/st.h>
-#include <boost/endian/conversion.hpp>
-#include <array>
 
 namespace ripple {
 
@@ -102,7 +101,7 @@ Clawback::preclaim(PreclaimContext const& ctx)
             issuer,
             fhIGNORE_FREEZE,
             ctx.j) <= beast::zero)
-        return tecNO_LINE;
+        return tecINSUFFICIENT_FUNDS;
 
     return tesSUCCESS;
 }
--- a/src/ripple/app/tx/impl/Clawback.cpp
+++ b/src/ripple/app/tx/impl/Clawback.cpp
@@ -1,0 +1,173 @@
+//------------------------------------------------------------------------------
+/*
+  This file is part of rippled: https://github.com/ripple/rippled
+  Copyright (c) 2023 Ripple Labs Inc.
+
+  Permission to use, copy, modify, and/or distribute this software for any
+  purpose  with  or without fee is hereby granted, provided that the above
+  copyright notice and this permission notice appear in all copies.
+
+  THE  SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+  WITH  REGARD  TO  THIS  SOFTWARE  INCLUDING  ALL  IMPLIED  WARRANTIES  OF
+  MERCHANTABILITY  AND  FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+  ANY  SPECIAL ,  DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+  WHATSOEVER  RESULTING  FROM  LOSS  OF USE, DATA OR PROFITS, WHETHER IN AN
+  ACTION  OF  CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+  OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+//==============================================================================
+
+#include <ripple/app/tx/impl/Clawback.h>
+#include <ripple/ledger/Directory.h>
+#include <ripple/ledger/View.h>
+#include <ripple/protocol/Feature.h>
+#include <ripple/protocol/Protocol.h>
+#include <ripple/protocol/TxFlags.h>
+#include <ripple/protocol/st.h>
+#include <boost/endian/conversion.hpp>
+#include <array>
+
+namespace ripple {
+
+NotTEC
+Clawback::preflight(PreflightContext const& ctx)
+{
+    if (!ctx.rules.enabled(featureClawback))
+        return temDISABLED;
+
+    if (auto const ret = preflight1(ctx); !isTesSuccess(ret))
+        return ret;
+
+    if (ctx.tx.getFlags() & tfClawbackMask)
+        return temINVALID_FLAG;
+
+    AccountID const issuer = ctx.tx.getAccountID(sfAccount);
+    STAmount const clawAmount(ctx.tx.getFieldAmount(sfAmount));
+
+    // The issuer field is used for the token holder instead
+    AccountID const holder = clawAmount.getIssuer();
+
+    if (issuer == holder || isXRP(clawAmount) || clawAmount <= beast::zero)
+        return temBAD_AMOUNT;
+
+    return preflight2(ctx);
+}
+
+TER
+Clawback::preclaim(PreclaimContext const& ctx)
+{
+    AccountID const issuer = ctx.tx.getAccountID(sfAccount);
+    STAmount const clawAmount(ctx.tx.getFieldAmount(sfAmount));
+    AccountID const holder = clawAmount.getIssuer();
+
+    auto const sleIssuer = ctx.view.read(keylet::account(issuer));
+    auto const sleHolder = ctx.view.read(keylet::account(holder));
+    if (!sleIssuer || !sleHolder)
+        return terNO_ACCOUNT;
+
+    std::uint32_t const issuerFlagsIn = sleIssuer->getFieldU32(sfFlags);
+
+    // If AllowClawback is not set or NoFreeze is set, return no permission
+    if (!(issuerFlagsIn & lsfAllowClawback) || (issuerFlagsIn & lsfNoFreeze))
+        return tecNO_PERMISSION;
+
+    auto const sleRippleState =
+        ctx.view.read(keylet::line(holder, issuer, clawAmount.getCurrency()));
+    if (!sleRippleState)
+        return tecNO_LINE;
+
+    STAmount const balance = sleRippleState->getFieldAmount(sfBalance);
+
+    // If balance is positive, issuer must have higher address than holder
+    if (balance > beast::zero && issuer < holder)
+        return tecNO_PERMISSION;
+
+    // If balance is negative, issuer must have lower address than holder
+    if (balance < beast::zero && issuer > holder)
+        return tecNO_PERMISSION;
+
+    // At this point, we know that issuer and holder accounts
+    // are correct and a trustline exists between them.
+    //
+    // Must now explicitly check the balance to make sure
+    // available balance is non-zero.
+    //
+    // We can't directly check the balance of trustline because
+    // the available balance of a trustline is prone to new changes (eg.
+    // XLS-34). So we must use `accountHolds`.
+    if (accountHolds(
+            ctx.view,
+            holder,
+            clawAmount.getCurrency(),
+            issuer,
+            fhIGNORE_FREEZE,
+            ctx.j) <= beast::zero)
+        return tecNO_LINE;
+
+    return tesSUCCESS;
+}
+
+TER
+Clawback::clawback(
+    AccountID const& issuer,
+    AccountID const& holder,
+    STAmount const& amount)
+{
+    // This should never happen
+    if (amount <= beast::zero || issuer != amount.getIssuer())
+        return tecINTERNAL;
+
+    STAmount const holderBalance = accountHolds(
+        view(),
+        holder,
+        amount.getCurrency(),
+        amount.getIssuer(),
+        fhIGNORE_FREEZE,
+        j_);
+
+    // The amount to be clawed back can never exceed the amount that
+    // the token holder current owns
+    if (amount > holderBalance)
+        return tecINTERNAL;
+
+    auto const result = accountSend(view(), holder, issuer, amount, j_);
+
+    if (!isTesSuccess(result))
+        return result;
+
+    if (accountHolds(
+            view(),
+            holder,
+            amount.getCurrency(),
+            amount.getIssuer(),
+            fhIGNORE_FREEZE,
+            j_)
+            .signum() < 0)
+        return tecINTERNAL;
+
+    return tesSUCCESS;
+}
+
+TER
+Clawback::doApply()
+{
+    AccountID const issuer = account_;
+    STAmount clawAmount(ctx_.tx.getFieldAmount(sfAmount));
+    AccountID const holder = clawAmount.getIssuer();
+
+    // Replace the `issuer` field with holder's account
+    clawAmount.setIssuer(issuer);
+
+    // Get the spendable balance. Must use `accountHolds`.
+    STAmount const spendableAmount = accountHolds(
+        view(),
+        holder,
+        clawAmount.getCurrency(),
+        clawAmount.getIssuer(),
+        fhIGNORE_FREEZE,
+        j_);
+
+    return clawback(issuer, holder, std::min(spendableAmount, clawAmount));
+}
+
+}  // namespace ripple
--- a/src/ripple/app/tx/impl/Clawback.cpp
+++ b/src/ripple/app/tx/impl/Clawback.cpp
@@ -39,11 +39,11 @@ Clawback::preflight(PreflightContext const& ctx)
     if (ctx.tx.getFlags() & tfClawbackMask)
         return temINVALID_FLAG;
 
-    AccountID const issuer = ctx.tx.getAccountID(sfAccount);
-    STAmount const clawAmount(ctx.tx.getFieldAmount(sfAmount));
+    AccountID const issuer = ctx.tx[sfAccount];
+    STAmount const clawAmount = ctx.tx[sfAmount];
 
     // The issuer field is used for the token holder instead
-    AccountID const holder = clawAmount.getIssuer();
+    AccountID const& holder = clawAmount.getIssuer();
 
     if (issuer == holder || isXRP(clawAmount) || clawAmount <= beast::zero)
         return temBAD_AMOUNT;
@@ -54,9 +54,9 @@ Clawback::preflight(PreflightContext const& ctx)
 TER
 Clawback::preclaim(PreclaimContext const& ctx)
 {
-    AccountID const issuer = ctx.tx.getAccountID(sfAccount);
-    STAmount const clawAmount(ctx.tx.getFieldAmount(sfAmount));
-    AccountID const holder = clawAmount.getIssuer();
+    AccountID const issuer = ctx.tx[sfAccount];
+    STAmount const clawAmount = ctx.tx[sfAmount];
+    AccountID const& holder = clawAmount.getIssuer();
 
     auto const sleIssuer = ctx.view.read(keylet::account(issuer));
     auto const sleHolder = ctx.view.read(keylet::account(holder));
@@ -74,7 +74,7 @@ Clawback::preclaim(PreclaimContext const& ctx)
     if (!sleRippleState)
         return tecNO_LINE;
 
-    STAmount const balance = sleRippleState->getFieldAmount(sfBalance);
+    STAmount const balance = (*sleRippleState)[sfBalance];
 
     // If balance is positive, issuer must have higher address than holder
     if (balance > beast::zero && issuer < holder)
@@ -108,12 +108,14 @@ Clawback::preclaim(PreclaimContext const& ctx)
 TER
 Clawback::doApply()
 {
-    AccountID const issuer = account_;
-    STAmount clawAmount(ctx_.tx.getFieldAmount(sfAmount));
-    AccountID const holder = clawAmount.getIssuer();
+    AccountID const& issuer = account_;
+    STAmount clawAmount = ctx_.tx[sfAmount];
+    AccountID const holder = clawAmount.getIssuer(); // cannot be reference
 
     // Replace the `issuer` field with issuer's account
     clawAmount.setIssuer(issuer);
+    if (holder == issuer)
+        return tecINTERNAL;
 
     // Get the spendable balance. Must use `accountHolds`.
     STAmount const spendableAmount = accountHolds(
@@ -124,8 +126,13 @@ Clawback::doApply()
         fhIGNORE_FREEZE,
         j_);
 
-    return accountSend(
-        view(), holder, issuer, std::min(spendableAmount, clawAmount), j_);
+    return rippleCredit(
+        view(),
+        holder,
+        issuer,
+        std::min(spendableAmount, clawAmount),
+        true,
+        j_);
 }
 
 }  // namespace ripple
--- a/src/ripple/app/tx/impl/Clawback.cpp
+++ b/src/ripple/app/tx/impl/Clawback.cpp
@@ -110,7 +110,7 @@ Clawback::doApply()
 {
     AccountID const& issuer = account_;
     STAmount clawAmount = ctx_.tx[sfAmount];
-    AccountID const holder = clawAmount.getIssuer(); // cannot be reference
+    AccountID const holder = clawAmount.getIssuer();  // cannot be reference
 
     // Replace the `issuer` field with issuer's account
     clawAmount.setIssuer(issuer);
--- a/src/ripple/app/tx/impl/Clawback.cpp
+++ b/src/ripple/app/tx/impl/Clawback.cpp
@@ -107,19 +107,6 @@ Clawback::preclaim(PreclaimContext const& ctx)
 }
 
 TER
-Clawback::clawback(
-    AccountID const& issuer,
-    AccountID const& holder,
-    STAmount const& amount)
-{
-    // This should never happen
-    if (amount <= beast::zero || issuer != amount.getIssuer())
-        return tecINTERNAL;
-
-    return accountSend(view(), holder, issuer, amount, j_);
-}
-
-TER
 Clawback::doApply()
 {
     AccountID const issuer = account_;
@@ -138,7 +125,7 @@ Clawback::doApply()
         fhIGNORE_FREEZE,
         j_);
 
-    return clawback(issuer, holder, std::min(spendableAmount, clawAmount));
+    return accountSend(view(), holder, issuer, std::min(spendableAmount, clawAmount), j_);
 }
 
 }  // namespace ripple
--- a/src/ripple/app/tx/impl/Clawback.cpp
+++ b/src/ripple/app/tx/impl/Clawback.cpp
@@ -18,7 +18,6 @@
 //==============================================================================
 
 #include <ripple/app/tx/impl/Clawback.h>
-#include <ripple/ledger/Directory.h>
 #include <ripple/ledger/View.h>
 #include <ripple/protocol/Feature.h>
 #include <ripple/protocol/Indexes.h>
--- a/src/ripple/app/tx/impl/Clawback.cpp
+++ b/src/ripple/app/tx/impl/Clawback.cpp
@@ -117,19 +117,6 @@ Clawback::clawback(
     if (amount <= beast::zero || issuer != amount.getIssuer())
         return tecINTERNAL;
 
-    STAmount const holderBalance = accountHolds(
-        view(),
-        holder,
-        amount.getCurrency(),
-        amount.getIssuer(),
-        fhIGNORE_FREEZE,
-        j_);
-
-    // The amount to be clawed back can never exceed the amount that
-    // the token holder current owns
-    if (amount > holderBalance)
-        return tecINTERNAL;
-
     auto const result = accountSend(view(), holder, issuer, amount, j_);
 
     if (!isTesSuccess(result))
@@ -155,7 +142,7 @@ Clawback::doApply()
     STAmount clawAmount(ctx_.tx.getFieldAmount(sfAmount));
     AccountID const holder = clawAmount.getIssuer();
 
-    // Replace the `issuer` field with holder's account
+    // Replace the `issuer` field with issuer's account
     clawAmount.setIssuer(issuer);
 
     // Get the spendable balance. Must use `accountHolds`.
--- a/src/ripple/app/tx/impl/Clawback.cpp
+++ b/src/ripple/app/tx/impl/Clawback.cpp
@@ -116,22 +116,7 @@ Clawback::clawback(
     if (amount <= beast::zero || issuer != amount.getIssuer())
         return tecINTERNAL;
 
-    auto const result = accountSend(view(), holder, issuer, amount, j_);
-
-    if (!isTesSuccess(result))
-        return result;
-
-    if (accountHolds(
-            view(),
-            holder,
-            amount.getCurrency(),
-            amount.getIssuer(),
-            fhIGNORE_FREEZE,
-            j_)
-            .signum() < 0)
-        return tecINTERNAL;
-
-    return tesSUCCESS;
+    return accountSend(view(), holder, issuer, amount, j_);
 }
 
 TER
--- a/src/ripple/app/tx/impl/Clawback.cpp
+++ b/src/ripple/app/tx/impl/Clawback.cpp
@@ -125,7 +125,8 @@ Clawback::doApply()
         fhIGNORE_FREEZE,
         j_);
 
-    return accountSend(view(), holder, issuer, std::min(spendableAmount, clawAmount), j_);
+    return accountSend(
+        view(), holder, issuer, std::min(spendableAmount, clawAmount), j_);
 }
 
 }  // namespace ripple
--- a/src/ripple/app/tx/impl/InvariantCheck.cpp
+++ b/src/ripple/app/tx/impl/InvariantCheck.cpp
@@ -767,7 +767,7 @@ ValidClawback::finalize(
     {
         if (trustlinesChanged != 0)
         {
-            JLOG(j.fatal()) << "Invariant failed: trustline changed.";
+            JLOG(j.fatal()) << "Invariant failed: some trustlines were changed despite failure of the transaction.";
             return false;
         }
     }
--- a/src/ripple/app/tx/impl/InvariantCheck.cpp
+++ b/src/ripple/app/tx/impl/InvariantCheck.cpp
@@ -27,6 +27,7 @@
 #include <ripple/protocol/STArray.h>
 #include <ripple/protocol/SystemParameters.h>
 #include <ripple/protocol/nftPageMask.h>
+#include <ripple/ledger/View.h>
 
 namespace ripple {
 
@@ -711,6 +712,67 @@ NFTokenCountTracking::finalize(
                 << "Invariant failed: burning changed the number of "
                    "minted tokens.";
             return false;
+        }
+    }
+
+    return true;
+}
+
+//------------------------------------------------------------------------------
+
+void
+ValidClawback::visitEntry(
+    bool,
+    std::shared_ptr<SLE const> const& before,
+    std::shared_ptr<SLE const> const& after)
+{
+    if (before && before->getType() == ltRIPPLE_STATE)
+        trustlineChanged++;
+}
+
+bool
+ValidClawback::finalize(
+    STTx const& tx,
+    TER const result,
+    XRPAmount const,
+    ReadView const& view,
+    beast::Journal const& j)
+{
+    if (tx.getTxnType() != ttCLAWBACK)
+        return true;
+
+    if (result == tesSUCCESS)
+    {
+        // a successful clawback transaction will ALWAYS claw back some funds
+        // from ONE trustline
+        if (trustlineChanged != 1){
+            JLOG(j.fatal())
+                << "Invariant failed: wrong number of trustline changed.";
+            return false;                
+        }
+
+        AccountID const issuer = tx.getAccountID(sfAccount);
+        STAmount const amount = tx.getFieldAmount(sfAmount);
+        AccountID const holder = amount.getIssuer();
+        STAmount const holderBalance = accountHolds(
+            view,
+            holder,
+            amount.getCurrency(),
+            issuer,
+            fhIGNORE_FREEZE,
+            j);
+        
+        if (holderBalance.signum() < 0){
+            JLOG(j.fatal())
+                << "Invariant failed: trustline balance is negative";
+            return false;                
+        }
+    }
+    else{
+        if (trustlineChanged != 0){
+            JLOG(j.fatal())
+                << "Invariant failed: trustline changed.";
+            return false;                
         }
     }
 
--- a/src/ripple/app/tx/impl/InvariantCheck.cpp
+++ b/src/ripple/app/tx/impl/InvariantCheck.cpp
@@ -752,7 +752,7 @@ ValidClawback::finalize(
 
         AccountID const issuer = tx.getAccountID(sfAccount);
         STAmount const amount = tx.getFieldAmount(sfAmount);
-        AccountID const holder = amount.getIssuer();
+        AccountID const& holder = amount.getIssuer();
         STAmount const holderBalance = accountHolds(
             view, holder, amount.getCurrency(), issuer, fhIGNORE_FREEZE, j);
 
--- a/src/ripple/app/tx/impl/InvariantCheck.cpp
+++ b/src/ripple/app/tx/impl/InvariantCheck.cpp
@@ -724,7 +724,7 @@ void
 ValidClawback::visitEntry(
     bool,
     std::shared_ptr<SLE const> const& before,
-    std::shared_ptr<SLE const> const& after)
+    std::shared_ptr<SLE const> const&)
 {
     if (before && before->getType() == ltRIPPLE_STATE)
         trustlinesChanged++;
--- a/src/ripple/app/tx/impl/InvariantCheck.cpp
+++ b/src/ripple/app/tx/impl/InvariantCheck.cpp
@@ -743,12 +743,10 @@ ValidClawback::finalize(
 
     if (result == tesSUCCESS)
     {
-        // a successful clawback transaction will ALWAYS claw back some funds
-        // from ONE trustline
-        if (trustlinesChanged != 1)
+        if (trustlinesChanged > 1)
         {
             JLOG(j.fatal())
-                << "Invariant failed: wrong number of trustline changed.";
+                << "Invariant failed: more than one trustline changed.";
             return false;
         }
 
--- a/src/ripple/app/tx/impl/InvariantCheck.cpp
+++ b/src/ripple/app/tx/impl/InvariantCheck.cpp
@@ -767,7 +767,8 @@ ValidClawback::finalize(
     {
         if (trustlinesChanged != 0)
         {
-            JLOG(j.fatal()) << "Invariant failed: some trustlines were changed despite failure of the transaction.";
+            JLOG(j.fatal()) << "Invariant failed: some trustlines were changed "
+                               "despite failure of the transaction.";
             return false;
         }
     }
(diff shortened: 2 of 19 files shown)


src/ripple/app/tx/impl/Clawback.cpp
```diff
@@ -0,0 +1,173 @@
...
+    auto const sleIssuer = ctx.view.read(keylet::account(issuer));
+    auto const sleHolder = ctx.view.read(keylet::account(holder));
+    if (!sleIssuer || !sleHolder)
+        return terNO_ACCOUNT;
+
+    std::uint32_t const issuerFlagsIn = sleIssuer->getFieldU32(sfFlags);
+
+    // If AllowClawback is not set or NoFreeze is set, return no permission
+    if (!(issuerFlagsIn & lsfAllowClawback) || (issuerFlagsIn & lsfNoFreeze))
+        return tecNO_PERMISSION;
```
nbougalis, line 72: Just leaving a comment here, no action needed: although not _strictly_ necessary here (because of how the code involving the two flags is structured) checking *both* flags is a good "belt and suspenders" and imposes no real additional overhead.

nbougalis (changes requested):
Left some comments that need to be addressed before I give this one a pass.

Also, I notice that there no invariant checkers where added. At least one should be possible to add (see one of the comments in this review).

src/ripple/app/tx/impl/Clawback.cpp
```diff
@@ -0,0 +1,173 @@
...
+    if (amount <= beast::zero || issuer != amount.getIssuer())
+        return tecINTERNAL;
+
+    STAmount const holderBalance = accountHolds(
+        view(),
+        holder,
+        amount.getCurrency(),
+        amount.getIssuer(),
+        fhIGNORE_FREEZE,
+        j_);
```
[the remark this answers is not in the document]
shawnxie999: true, it's not needed, will fix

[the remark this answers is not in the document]
shawnxie999: true, it's not needed, will fix

src/ripple/app/tx/impl/SetAccount.cpp
```diff
@@ -562,6 +589,14 @@ SetAccount::doApply()
             uFlagsOut &= ~lsfDisallowIncomingTrustline;
     }
 
+    // Set flag for clawback
+    if (ctx_.view().rules().enabled(featureClawback) &&
+        uSetFlag == asfAllowClawback)
+    {
+        JLOG(j_.trace()) << "set allow clawback";
+        uFlagsOut |= lsfAllowClawback;
+    }
```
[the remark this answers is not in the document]
shawnxie999: this flag is never meant to be cleared. Once set, it cannot be cleared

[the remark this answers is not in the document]
shawnxie999: this flag is never meant to be cleared. Once set, it cannot be cleared

src/ripple/app/tx/impl/Clawback.cpp
```diff
@@ -0,0 +1,173 @@
...
+    std::uint32_t const issuerFlagsIn = sleIssuer->getFieldU32(sfFlags);
+
+    // If AllowClawback is not set or NoFreeze is set, return no permission
+    if (!(issuerFlagsIn & lsfAllowClawback) || (issuerFlagsIn & lsfNoFreeze))
+        return tecNO_PERMISSION;
+
+    auto const sleRippleState =
+        ctx.view.read(keylet::line(holder, issuer, clawAmount.getCurrency()));
+    if (!sleRippleState)
+        return tecNO_LINE;
```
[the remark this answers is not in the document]
shawnxie999: The SLE of the trustline is fetched to check the account must be the issuer in the trustline. Without loading it and purely by using `accountHolds`, I would not be able to know that

```
    // If balance is positive, issuer must have higher address than holder
    if (balance > beast::zero && issuer < holder)
        return tecNO_PERMISSION;

    // If balance is negative, issuer must have lower address than holder
    if (balance < beast::zero && issuer > holder)
        return tecNO_PERMISSION;
```

src/ripple/app/tx/impl/Clawback.cpp
```diff
@@ -0,0 +1,173 @@
...
+
+    STAmount const balance = sleRippleState->getFieldAmount(sfBalance);
+
+    // If balance is positive, issuer must have higher address than holder
+    if (balance > beast::zero && issuer < holder)
+        return tecNO_PERMISSION;
+
+    // If balance is negative, issuer must have lower address than holder
+    if (balance < beast::zero && issuer > holder)
+        return tecNO_PERMISSION;
```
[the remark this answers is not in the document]
shawnxie999: This if-condition is purely used to check that the account of this transaction is the issuer, by comparing high/low account with the polarity of the balance. It's not used to check the balance.

src/ripple/app/tx/impl/SetAccount.cpp
```diff
@@ -361,6 +380,14 @@ SetAccount::doApply()
...
         }
 
+        // Cannot set NoFreeze if clawback is enabled
+        if (ctx_.view().rules().enabled(featureClawback) &&
+            (uFlagsIn & lsfAllowClawback))
+        {
+            JLOG(j_.trace()) << "Can't set NoFreeze if clawback is enabled";
+            return tecNO_PERMISSION;
+        }
+
```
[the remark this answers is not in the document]
shawnxie999: >The current code doesn't support unsetting the lsfAllowClawback flag and it should; once it does, the above scenario becomes possible by having sfSetFlag = asfNoFreeze and sfClearFlag = asfAllowClawback

It is purposely made that `lsfAllowClawback` cannot be cleared. This is reflected in the spec as well.

src/ripple/app/tx/impl/Clawback.cpp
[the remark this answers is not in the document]
shawnxie999: Also, if we want to be able to claw back escrowed amounts, I think it would be better to propose a separate amendment after both PRs are merged. I'm don't think it is a good idea to have two PRs making changes for it in parallel

src/ripple/app/tx/impl/SetAccount.cpp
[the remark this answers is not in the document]
nbougalis: But why? This is the inverse of "no freeze" and it should definitely be a flag you can clear.

src/ripple/app/tx/impl/Clawback.cpp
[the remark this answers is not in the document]
dangell7: I don't think escrow or pay channels should be included in the clawback if AMM is excluded. 

The justification is "it only allows an issuer to claw back the funds that are spendable".

Escrowed or Paychan iou tokens are locked and are not spendable. Therefore escrow and paychannel should be excluded.

src/ripple/app/tx/impl/SetAccount.cpp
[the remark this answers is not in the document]
shawnxie999: @nbougalis The reason for making clawback opt-out by default was purely because, having clawback opt-in was a poor idea and would raise a lot of concerns, just like how having freeze opt-in by default was a bad choice. The verdict is that we would like "opt-out" these kind of regulatory features in the future, and clawback would be the first one to follow that. 

What is really the incentive for making allow clawback a mutable field?  If they really want to mutate it, then I don't see why can't they create a new account.

But I think it would be a good practice to stick with immutable fields for regulatory features like freeze or clawback.

src/ripple/app/tx/impl/Clawback.cpp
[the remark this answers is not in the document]
shawnxie999: i would feel like explicitly checking the high/low account and the balance polarity to verify that the issuer is correct. Even though `accountHolds` could do that too, but that is not the purpose of this function, and behaviors could possibly change too.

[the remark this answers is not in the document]
shawnxie999: > i would feel like explicitly checking the high/low account and the balance polarity to verify that the issuer is correct. Even though checking sign of what `accountHolds` returns could work too, but that is not the purpose of this function, and behaviors could possibly change too.

src/ripple/app/tx/impl/Clawback.cpp
```diff
@@ -0,0 +1,144 @@
...
+    if (ctx.tx.getFlags() & tfClawbackMask)
+        return temINVALID_FLAG;
+
+    AccountID const issuer = ctx.tx.getAccountID(sfAccount);
+    STAmount const clawAmount(ctx.tx.getFieldAmount(sfAmount));
+
+    // The issuer field is used for the token holder instead
+    AccountID const holder = clawAmount.getIssuer();
+
+    if (issuer == holder || isXRP(clawAmount) || clawAmount <= beast::zero)
```
ckeshava, line 49: In the case of `issuer == holder`, I'm curious why this particular error code is being returned.

Wouldn't it be more appropriate to return `tecNoPermission` or any error code pertaining to account ownership?

src/ripple/app/tx/impl/SetAccount.cpp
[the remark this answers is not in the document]
nbougalis: No, you misunderstand: I'm not suggesting that clawback should be opt-out, that would be a bad idea. I was just saying that I would allow it to be unset, but I don't feel strongly enough about it.

src/ripple/app/tx/impl/Clawback.cpp
```diff
@@ -0,0 +1,144 @@
...
+        return ret;
+
+    if (ctx.tx.getFlags() & tfClawbackMask)
+        return temINVALID_FLAG;
+
+    AccountID const issuer = ctx.tx.getAccountID(sfAccount);
+    STAmount const clawAmount(ctx.tx.getFieldAmount(sfAmount));
+
+    // The issuer field is used for the token holder instead
+    AccountID const holder = clawAmount.getIssuer();
```
ckeshava, line 47: can we add a new function to the `STAmount` class which appropriate reflects the true intent of this variable? like `getHolder()` ?

I am not aware of the original intention of `getIssuer()` function name, but it is a counter-party of the intended meaning in the clawback context

nbougalis: Such functionality doesn't belong in the `STAmount` class which is _already_ overburdened with cruft.

src/ripple/app/tx/impl/Clawback.cpp
shawnxie999: I think this error code was chosen loosely. But it makes sense. If the account inside `Amount` parameter is the same as the one submitting this transaction, then something is wrong with this field, and the transaction submitter would need to fix the `Amount` field. So from this perspective,`temBAD_AMOUNT` is specific enough to show where things go wrong. 

Also, `tecNO_PERMISSION` is mainly used in `preclaim` and `doApply`, not in `preflight` I think. `preflight` doesn't access ledger objects, so we wouldn't really be able to know whether someone has the permission to do something

nbougalis: `preflight` can never return a `tec`.

src/ripple/app/tx/impl/Clawback.cpp
```diff
@@ -0,0 +1,173 @@
...
+    // Get the spendable balance. Must use `accountHolds`.
+    STAmount const spendableAmount = accountHolds(
+        view(),
+        holder,
+        clawAmount.getCurrency(),
+        clawAmount.getIssuer(),
+        fhIGNORE_FREEZE,
+        j_);
+
+    return clawback(issuer, holder, std::min(spendableAmount, clawAmount));
```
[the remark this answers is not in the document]
shawnxie999: I removed the `clawback` function

src/ripple/app/tx/impl/Clawback.cpp
shawnxie999: @ckeshava The intent of this is written in the [spec](https://github.com/XRPLF/XRPL-Standards/tree/master/XLS-39d-clawback).

> The sub-field issuer within Amount represents the token holder's address instead of the issuer's.

We are just using the issuer field because it would save us from the need to create one extra parameter. But I agree it's confusing to some people.

src/ripple/protocol/TxFlags.h
```diff
@@ -159,6 +160,9 @@ constexpr std::uint32_t const tfNFTokenCancelOfferMask     = ~(tfUniversal);
 // NFTokenAcceptOffer flags:
 constexpr std::uint32_t const tfNFTokenAcceptOfferMask     = ~tfUniversal;
 
+// Clawback flags:
+constexpr std::uint32_t const tfClawbackMask     = ~tfUniversal;
```
ckeshava, line 164: I did not understand the intent of this variable. I know that this variable is used inside `preflight()`. Source file: `src/ripple/app/tx/impl/Clawback.cpp`

If any of the universal flags are set, then return `temINVALID_FLAG`, is this correct?

shawnxie999: It does the opposite, if any flags other than the universal flag are set, then return `temINVALID_FLAG`.

Here, clawback does not introduce any new flags

In binary:
tfUniversal = `10000000000000000000000000000000` 
tfClawbackMask =  `01111111111111111111111111111111`

If someone enters a flag into the `Clawback` transaction such as `00000000000000000000000000000001`, then
ctx.tx.getFlags() & tfClawbackMask = `00000000000000000000000000000001 & 01111111111111111111111111111111 ` will result in `00000000000000000000000000000001`, which will return `temINVALID_FLAG`

Therefore, we know the flag value of 1 is not valid.

shawnxie999: @scottschurr just curious, is there any use for the `tfFullyCanonicalSig` flag? i think it's deprecated

shawnxie999:
Read for re-review

src/ripple/app/tx/impl/SetAccount.cpp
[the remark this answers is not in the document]
shawnxie999: I'm not quite sure if there is any need for this. But if we find out that unsetting the flag is really wanted somehow, I think it would it would be fine to add it through an amendment later. This would be a rather small change. But I think it would be a good idea to keep it simple for now.

src/ripple/protocol/TxFlags.h
scottschurr: Hi @shawnxie999.  The `tfFullyCanonicalSig` flagis not deprecated, however it has had no effect since July of 2020.  On July 3, 2020 an amendment went live that requires all signatures to be fully canonical, regardless of whether the `tfFullyCanonicalSig ` is set.  So it's okay to set the flag, it just doesn't make things any safer than otherwise.

We certainly don't want the server to reject any transactions with the `tfFullyCanonicalSig ` flag set.  So the flag itself is still valid.

shawnxie999: Thank you!

src/ripple/app/tx/impl/Clawback.cpp
ckeshava: yes, I remember seeing it in the spec. I felt it would be easier to understand if it was the other way around.

src/ripple/app/tx/impl/Clawback.cpp
ckeshava: oh okay, thanks for the explanation 👍

src/ripple/app/tx/impl/Clawback.cpp
```diff
@@ -0,0 +1,132 @@
...
+  MERCHANTABILITY  AND  FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+  ANY  SPECIAL ,  DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+  WHATSOEVER  RESULTING  FROM  LOSS  OF USE, DATA OR PROFITS, WHETHER IN AN
+  ACTION  OF  CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+  OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+//==============================================================================
+
+#include <ripple/app/tx/impl/Clawback.h>
+#include <ripple/ledger/Directory.h>
```
ckeshava, line 21: I think this is an unused import, could be removed

src/ripple/app/tx/impl/Clawback.cpp
```diff
@@ -0,0 +1,132 @@
...
+            ctx.j) <= beast::zero)
+        return tecINSUFFICIENT_FUNDS;
+
+    return tesSUCCESS;
+}
+
+TER
+Clawback::doApply()
+{
+    AccountID const issuer = account_;
```
ckeshava, line 112: Is it helpful to store these three variables as data members of the `Clawback` class instead of recomputing them in almost all the member functions? I notice that these variables are being declared and defined in `preflight`m `preclaim` and `dpApply`

src/ripple/app/tx/impl/InvariantCheck.cpp
```diff
@@ -717,4 +718,61 @@ NFTokenCountTracking::finalize(
...
+            JLOG(j.fatal())
+                << "Invariant failed: trustline balance is negative";
+            return false;
+        }
+    }
+    else
+    {
+        if (trustlinesChanged != 0)
+        {
+            JLOG(j.fatal()) << "Invariant failed: trustline changed.";
```
ckeshava, line 770: I feel a more descriptive error message would be easier to understand here. I'm thinking of something like "Invariant Failed: Some trustlines were changed despite failure of the transaction".

src/ripple/app/tx/impl/InvariantCheck.h
```diff
@@ -365,6 +365,32 @@ class NFTokenCountTracking
...
 
+/**
+ * @brief Invariant: Trust line balance cannot be negative after Clawback.
+ *
+ * We iterate all the trust lines affected by this transaction and ensure
+ * that they have non-negative balance.
+ */
+class ValidClawback
+{
+    std::uint32_t trustlinesChanged = 0;
```
ckeshava, line 376: My understanding of the `ValidClawback::visitEntry` function is that, `trustlinesChanges` can only be `0` or `1` (because of the said function definition in implementation file of `InvariantCheck`). 

Am I missing something? Hypothetically, even if more than one trust line were modified, the `ValidClawback::visitEntry()` would be called only once right?

src/ripple/app/tx/impl/InvariantCheck.h
```diff
@@ -365,6 +365,32 @@ class NFTokenCountTracking
...
+
+public:
+    void
+    visitEntry(
+        bool,
+        std::shared_ptr<SLE const> const&,
+        std::shared_ptr<SLE const> const&);
+
+    bool
+    finalize(
```
ckeshava, line 386: On a related note, can we force all the `InvariantChecks` to adhere to a common interface by inheriting from an abstract class? It appears that all of them have `visitEntry` and `finalize` methods.

src/ripple/app/tx/impl/SetAccount.cpp
```diff
@@ -218,6 +218,37 @@ SetAccount::preclaim(PreclaimContext const& ctx)
...
+    {
+        if (uSetFlag == asfAllowClawback)
+        {
+            if (uFlagsIn & lsfNoFreeze)
+            {
+                JLOG(ctx.j.trace()) << "Can't set Clawback if NoFreeze is set";
+                return tecNO_PERMISSION;
+            }
+
+            if (!dirIsEmpty(ctx.view, keylet::ownerDir(id)))
```
ckeshava, line 234: How is `Clawback` triggering this piece of code which is in `SetAccount` ? I grep'ped for `SetAccount` in `Clawback.cpp` but I couldn't find any instances

src/ripple/app/tx/impl/SetAccount.cpp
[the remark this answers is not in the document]
ckeshava: Going on a tangent here: Since `Clawback` and `NoFreeze` are complimentary to one another, is it a good idea to frame this condition as an invariant? I would eliminate such a check in other parts of the code right?

src/ripple/protocol/LedgerFormats.h
```diff
@@ -243,6 +243,9 @@ enum LedgerSpecificFlags {
         0x10000000,               // True, reject new paychans
     lsfDisallowIncomingTrustline =
         0x20000000,               // True, reject new trustlines (only if no issued assets)
+    lsfAMM [[maybe_unused]] = 0x40000000, // True, AMM account
```
ckeshava, line 246: Why are we adding a flag pertaining to `AMM` in this PR? I couldn't find any usages of this flag in the codebase, what is it used for?

src/ripple/protocol/LedgerFormats.h
```diff
@@ -243,6 +243,9 @@ enum LedgerSpecificFlags {
...
+    lsfAllowClawback =
+        0x80000000,               // True, enable clawback
```
ckeshava, line 248: how did you come up with this particular value `0x80000000` for this flag? I know that it needs to be different from the other flags and complementary to `NoFreeze` flag, but were there any other considerations?

src/ripple/protocol/TxFlags.h
ckeshava: ok, thanks for the detailed explanation @shawnxie999

ckeshava: Why are we not allowing any of the non-universal flags along with a `Clawback` transaction? Where can I learn more about universal flags and their uses?

src/ripple/protocol/impl/TxFormats.cpp
```diff
@@ -328,6 +328,14 @@ TxFormats::TxFormats()
             {sfTicketSequence, soeOPTIONAL},
         },
         commonFields);
+
+    add(jss::Clawback,
+        ttCLAWBACK,
+        {
+            {sfAmount, soeREQUIRED},
+            {sfTicketSequence, soeOPTIONAL},
```
ckeshava, line 336: I'm trying to think of use cases where someone would want to use the tickets feature with `Clawback`, but I'm drawing up a blank.

Since the owner directory needs to be empty for a valid `Clawback` transaction, I suspect there are not a whole lot of transactions which can be done before a `Clawback` right?

src/test/app/Clawback_test.cpp
```diff
@@ -0,0 +1,971 @@
...
+    template <class T>
+    static std::string
+    to_string(T const& t)
+    {
+        return boost::lexical_cast<std::string>(t);
+    }
+
+    // Helper function that returns the owner count of an account root.
+    static std::uint32_t
+    ownerCount(test::jtx::Env const& env, test::jtx::Account const& acct)
```
ckeshava, line 42: I believe these two functions are common to a few other test suites like `NFToken`, `NFTokenBurn` and `Checks`. Given that the function definitions are identical, can we place these functions in a common utility file?

src/ripple/app/tx/impl/Clawback.cpp
shawnxie999: I'm not quite sure if there is other way around this, without adding an additional transaction parameter.

There are two accounts involved in a clawback transaction - the issuer and the holder. The `Account` field is the issuer, who would always be the one submitting the transaction. 

The holder's account is specified in the Amount field like this
```
{
    "currency": "USD",
    "value": "153.75",
    "issuer": "r9cZA1mLK5R5Am25ArfXFmqgNwjZgnfk59"
}
```
Here, and the `issuer` sub-field has the address`r9cZA1mLK5R5Am25ArfXFmqgNwjZgnfk59` which is the holder's address. So after this `Amount` field is serialized into `STAmount` object, calling `getIssuer()` on this object will return the holder's address. This is definitely an exception.

src/ripple/protocol/TxFlags.h
shawnxie999: @ckeshava `Clawback` here does not have introduce any transaction-specific flag, there it does not have any non-universal flags.

The only universal flag that is used by every transaction seems to be  `tfFullyCanonicalSig`, as defined here in the [code](https://github.com/XRPLF/rippled/blob/77c0a62a7472dbc4dc9ed97d3141840a3acb5ce8/src/ripple/protocol/TxFlags.h#L59).

src/ripple/app/tx/impl/Clawback.cpp
shawnxie999: AFAIK, `preflight` and `preclaim` are read only functions, and are not meant to mutate anything.

src/ripple/app/tx/impl/InvariantCheck.h
shawnxie999: From my understanding, I think `ValidClawback::visitEntry` will iterated on every ledger object that was changed as a result of the transaction. And `trustlinesChanged` is incremented each time it encounters a `RIPPLE_STATE`(trustline) object. 

For a `Clawback` transaction, `trustlinesChanged` could only be 0 or 1. This is why in `ValidClawback::finalize`, there is a guard to check for the transaction type:
```
    if (tx.getTxnType() != ttCLAWBACK)
        return true;
```

On the other hand, it's possible that other type of transactions(I think `Payment`?) could potentially change 2 or more trustline objects in the transaction. In that case, `trustlinesChanged` could be some number greater than 1 (since `visitEntry` is iterated on modified objects for _all_ transaction types). 

In your hypothetical scenario, if more than one trustlines are changed, `visitEntry` will keeps count of the total number of modified trustlines as it iterates through all the objects that were changed

```
    if (before && before->getType() == ltRIPPLE_STATE)
        trustlinesChanged++;
```

src/ripple/protocol/LedgerFormats.h
shawnxie999: This flag is added in the AMM PR https://github.com/XRPLF/rippled/pull/4294. I added it as a placeholder to give more clarity for upcoming PRs or resolving conflicts.

src/ripple/protocol/LedgerFormats.h
shawnxie999: I chose it because this is the next available flag. I don't think there is a clean way of deciding the flag value. But what I did is I looked over the [amendment proposals](https://github.com/XRPLF/XRPL-Standards/discussions), as well as their PRs, to find out which flag values are in use. 

Since the AccountRoot `Flags` is a 32-bit bitmask, I chose `0x80000000` because it is simply the last remaining flag value that haven't been used.

src/test/app/Clawback_test.cpp
shawnxie999: I think that makes sense, but not sure exactly where that could be put. I think it makes sense to create a issue later to address/clean this separately.

src/ripple/app/tx/impl/SetAccount.cpp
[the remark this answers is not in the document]
shawnxie999: I don't think this change would eliminate such checks in other parts of the code. Invariant check is executed _after_ a transaction.

And it will still be _mandatory_ to check this complementary property during the `preclaim` (before an transaction) of an `AccountSet`.

src/test/app/Clawback_test.cpp
```diff
@@ -0,0 +1,971 @@
...
+
+            // clawback fails because amendment is disabled
+            env(claw(alice, bob["USD"](5)), ter(temDISABLED));
+            env.close();
+
+            // now enable clawback amendment
+            env.enableFeature(featureClawback);
+            env.close();
+
+            // clawback fails because asfAllowClawback has not been set
```
ckeshava, line 243: regarding adding new features to the rippled codebase, why do we need to enable a new amendment and also set the `asfAllowClawback` flag? Aren't both of them redundant?

I'd like to understand why we have this type of design

src/test/app/Clawback_test.cpp
```diff
@@ -0,0 +1,971 @@
...
+        }
+    }
+
+    void
+    testBidirectionalLine(FeatureBitset features)
+    {
+        testcase("Bidirectional line");
+        using namespace test::jtx;
+
+        // Test when both alice and bob issues USD to each other.
```
ckeshava, line 652: very interesting test case!
Can we include this scenario also in the specification/documentation. I missed this case.

src/test/app/Clawback_test.cpp
shawnxie999: The new amendment is purposed to be voted by the validators. An amendment is required to pass before the feature can be present on the mainnet in the first place.

Now suppose the amendment passes, we have the `asfAllowClawback`  flag that allows someone can set on their _account level_. Without this flag, the account does not have the ability to claw back funds. So, it is individual preference to turn on this flag or not.

src/test/app/Clawback_test.cpp
```diff
@@ -0,0 +1,971 @@
...
+
+        // At this point, both alice and bob are the issuers of USD
+        // and can send USD to each other through one trustline
+
+        // alice fails to clawback. Even though she is also an issuer,
+        // the trustline balance is positive from her perspective
+        env(claw(alice, bob["USD"](200)), ter(tecNO_PERMISSION));
+        env.close();
+
+        // bob is able to successfully clawback from alice because
```
ckeshava, line 709: The party with a negative balance can unilaterally clawback from the counter-party of the trustline right?

Are there real-world analogues to this type of behavior? It feels quite arbitrary (although I'm aware that both the parties need to set the `asfClawback` flag and enable this particular amendment)

src/test/app/Clawback_test.cpp
shawnxie999: It is written in the spec, under `Clawback` transaction

>Indicates the account which is executing this transaction. The account MUST be the issuer of the asset being clawed back. Note that in the XRP Ledger, trustlines are bidirectional and, under some configurations, both sides can be seen as the "issuer" of an asset. In this specification, the term issuer is used to mean the side of the trustline that has an outstanding balance (i.e. 'owes' the issued asset) that it wishes to claw back.

src/test/app/Clawback_test.cpp
```diff
@@ -0,0 +1,971 @@
...
+        // alice issues 1000 USD to bob
+        env.trust(USD(1000), bob);
+        env(pay(alice, bob, USD(1000)));
+        env.close();
+
+        env.require(balance(bob, alice["USD"](1000)));
+        env.require(balance(alice, bob["USD"](-1000)));
+
+        // freeze trustline
+        env(trust(alice, bob["USD"](0), tfSetFreeze));
```
ckeshava, line 819: okay, to summarize:
An issuer `alice` can create a trust line with the `Clawback` feature enabled. She can then proceed to freeze the trust line unilaterally and clawback all of the funds. Is this right?

src/test/app/Clawback_test.cpp
```diff
@@ -0,0 +1,971 @@
...
+    run() override
+    {
+        using namespace test::jtx;
+        FeatureBitset const all{supported_amendments()};
+
+        testWithFeats(all);
+    }
+};
+
+BEAST_DEFINE_TESTSUITE(Clawback, app, ripple);
```
ckeshava, line 970: I like your unit test file. It covers many corner cases, good job!

src/ripple/app/tx/impl/Clawback.cpp
ckeshava: oh, I didn't notice that 👍 
By read-only, you mean that these do not update any global variable right? I know that they can't use the `this` pointer of the class because they are `static`.

src/ripple/app/tx/impl/Clawback.cpp
shawnxie999: Not really, `issuer` is the right naming for `Amount`. `clawback` is merely using this field in a "altered" way.

There is more info in the doc about [Token Amount](https://xrpl.org/currency-formats.html#token-amounts) .
image

src/ripple/protocol/TxFlags.h
ckeshava: ok, thank you!

src/test/app/Clawback_test.cpp
shawnxie999: Yup, `Clawback` should be completely independent of the freeze status of the trustline and should not mutate anything. I put this test in since it won't hurt, even though these are completely unrelated features.

ckeshava:
Okay :+1
Regards,
Keshava

ckeshava:
OK
Regards,
Keshava

ckeshava:
okay
Regards,
Keshava

ckeshava:
okay
Regards,
Keshava

ckeshava:
Regards,
Keshava


On Mon, Jun 12, 2023 at 9:45 AM Shawn Xie ***@***.***> wrote:

> ***@***.**** commented on this pull request.
> ------------------------------
>
> In src/ripple/app/tx/impl/SetAccount.cpp
> <https://urldefense.com/v3/__https://github.com/XRPLF/rippled/pull/4553*discussion_r1226941768__;Iw!!PZTMFYE!5SAwdIBi-qiCGCdzkwthNsiJcUJPW7DVniwwf1FFs1nysDvBBnN3Q2wIUCBLk1ILR8elly8QjHgwFIlmaZjjjeTv0g$>
> :
>
> > +    if (ctx_.view().rules().enabled(featureClawback) &&
> +        uSetFlag == asfAllowClawback)
> +    {
> +        JLOG(j_.trace()) << "set allow clawback";
> +        uFlagsOut |= lsfAllowClawback;
> +    }
>
> @ckeshava
> <https://urldefense.com/v3/__https://github.com/ckeshava__;!!PZTMFYE!5SAwdIBi-qiCGCdzkwthNsiJcUJPW7DVniwwf1FFs1nysDvBBnN3Q2wIUCBLk1ILR8elly8QjHgwFIlmaZjN2lsqbg$>
> I don't think this change would eliminate such checks in other parts of the
> code. Invariant check is executed *after* a transaction.
>
Hmm, I didn't realise that, ok.

ckeshava:
Regards,
Keshava


On Mon, Jun 12, 2023 at 10:45 AM Shawn Xie ***@***.***> wrote:

> ***@***.**** commented on this pull request.
> ------------------------------
>
> In src/test/app/Clawback_test.cpp
> <https://urldefense.com/v3/__https://github.com/XRPLF/rippled/pull/4553*discussion_r1227033168__;Iw!!PZTMFYE!_JVI1d53IQu2xC3PRCDSRVjcxcU3h4Mtu9Tt8n09D6V2cWkR-X65wd8BAv7rtgLo2QcN0TgiuJxhCzLXFgwRSD1Zww$>
> :
>
> > +            env.require(balance(alice, bob["USD"](-100)));
> +            env.require(balance(bob, alice["USD"](100)));
> +
> +            // cindy's balance is reduced
> +            env.require(balance(alice, cindy["USD"](-700)));
> +            env.require(balance(cindy, alice["USD"](700)));
> +        }
> +    }
> +
> +    void
> +    testBidirectionalLine(FeatureBitset features)
> +    {
> +        testcase("Bidirectional line");
> +        using namespace test::jtx;
> +
> +        // Test when both alice and bob issues USD to each other.
>
> It is written in the spec, under Clawback transaction
>
> Indicates the account which is executing this transaction. The account
> MUST be the issuer of the asset being clawed back. Note that in the XRP
> Ledger, trustlines are bidirectional and, under some configurations, both
> sides can be seen as the "issuer" of an asset. In this specification, the
> term issuer is used to mean the side of the trustline that has an
> outstanding balance (i.e. 'owes' the issued asset) that it wishes to claw
> back.
>
>
Okay, I must have missed it, mt bad.

ckeshava:
hmm okay
Regards,
Keshava

src/test/app/Clawback_test.cpp
ckeshava: okay

ckeshava (approved):

src/test/app/Clawback_test.cpp
shawnxie999: Thanks 😀 

I'll address this separately.

src/ripple/app/tx/impl/InvariantCheck.h
scottschurr: No, I don't know this.  I was not in on the design of the invariants.  But inheriting from an abstract class doesn't really _enforce_ a common interface.  The derived class can add whatever methods it wants.  It also is not _required_ to implement inherited methods well.

The thing that enforces the interface of these classes is that they are all consumed by `ApplyContext::checkInvariantHelper()` which expects a regular interface: https://github.com/XRPLF/rippled/blob/develop/src/ripple/app/tx/impl/ApplyContext.cpp#L91-L97

If the invariant does not comply with the interface expected by `checkInvariantsHelper()` then `checkInvariantsHelper()` will fail to compile.

This kind of design is common in template-heavy code, as opposed to code that is inheritance heavy.  Templates (largely) rely on "duck typing".  If it walks like a duck and quacks like a duck, then it is a duck.  All of the InvariantCheckers walk like InvariantCheckers and quack like InvariantCheckers so, as far as `checkInvariantsHelper()` is concerned, they are InvariantCheckers.

I hope that helps.

ckeshava: thanks for the explanation!

src/ripple/app/tx/impl/Clawback.cpp
```diff
@@ -0,0 +1,131 @@
...
+        return temDISABLED;
+
+    if (auto const ret = preflight1(ctx); !isTesSuccess(ret))
+        return ret;
+
+    if (ctx.tx.getFlags() & tfClawbackMask)
+        return temINVALID_FLAG;
+
+    AccountID const issuer = ctx.tx.getAccountID(sfAccount);
+    STAmount const clawAmount(ctx.tx.getFieldAmount(sfAmount));
```
seelabs, line 43: Consider:
```
     AccountID const issuer = ctx.tx[sfAccount];
     STAmount const clawAmount = ctx.tx[sfAmount];
```

src/ripple/app/tx/impl/Clawback.cpp
```diff
@@ -0,0 +1,131 @@
...
+
+    // The issuer field is used for the token holder instead
+    AccountID const holder = clawAmount.getIssuer();
```
seelabs, line 46: Consider:
```
     AccountID const& holder = clawAmount.getIssuer();
```

src/ripple/app/tx/impl/Clawback.cpp
```diff
@@ -0,0 +1,131 @@
...
+
+    return preflight2(ctx);
+}
+
+TER
+Clawback::preclaim(PreclaimContext const& ctx)
+{
+    AccountID const issuer = ctx.tx.getAccountID(sfAccount);
+    STAmount const clawAmount(ctx.tx.getFieldAmount(sfAmount));
+    AccountID const holder = clawAmount.getIssuer();
```
seelabs, line 59: Consider:
```
    AccountID const issuer = ctx.tx[sfAccount];
    STAmount const clawAmount = ctx.tx[sfAmount];
    AccountID const& holder = clawAmount.getIssuer();
```

src/ripple/app/tx/impl/Clawback.cpp
```diff
@@ -0,0 +1,131 @@
...
+
+    auto const sleIssuer = ctx.view.read(keylet::account(issuer));
+    auto const sleHolder = ctx.view.read(keylet::account(holder));
+    if (!sleIssuer || !sleHolder)
```
seelabs, line 63: I don't object to checking for the issuer, but note there will always be an issuer - the account submitting the transactions has to exist.

src/ripple/app/tx/impl/Clawback.cpp
```diff
@@ -0,0 +1,131 @@
...
+    // If AllowClawback is not set or NoFreeze is set, return no permission
+    if (!(issuerFlagsIn & lsfAllowClawback) || (issuerFlagsIn & lsfNoFreeze))
+        return tecNO_PERMISSION;
+
+    auto const sleRippleState =
+        ctx.view.read(keylet::line(holder, issuer, clawAmount.getCurrency()));
+    if (!sleRippleState)
+        return tecNO_LINE;
+
+    STAmount const balance = sleRippleState->getFieldAmount(sfBalance);
```
seelabs, line 77: Consider:
```
    STAmount const balance = (*sleRippleState)[sfBalance];
```

src/ripple/app/tx/impl/Clawback.cpp
```diff
@@ -0,0 +1,131 @@
...
+
+    return tesSUCCESS;
+}
+
+TER
+Clawback::doApply()
+{
+    AccountID const issuer = account_;
+    STAmount clawAmount(ctx_.tx.getFieldAmount(sfAmount));
+    AccountID const holder = clawAmount.getIssuer();
```
seelabs, line 113: Consider:
```
    AccountID const& issuer = account_;
    STAmount clawAmount = ctx_.tx[sfAmount];
    AccountID const& holder = clawAmount.getIssuer();
```

src/ripple/app/tx/impl/Clawback.cpp
```diff
@@ -0,0 +1,131 @@
...
+    STAmount const spendableAmount = accountHolds(
+        view(),
+        holder,
+        clawAmount.getCurrency(),
+        clawAmount.getIssuer(),
+        fhIGNORE_FREEZE,
+        j_);
+
+    return accountSend(
+        view(), holder, issuer, std::min(spendableAmount, clawAmount), j_);
```
seelabs, line 128: Consider:
```
    return rippleCredit(
        view(),
        holder,
        issuer,
        std::min(spendableAmount, clawAmount),
        false,
        j_);
```

src/ripple/app/tx/impl/InvariantCheck.h
```diff
@@ -365,6 +365,32 @@ class NFTokenCountTracking
         beast::Journal const&);
 };
 
+/**
+ * @brief Invariant: Trust line balance cannot be negative after Clawback.
```
seelabs, line 369: Consider re-writing this. You mean something different from "trust line balance" here. `accountHolds` is the balance with respect to one side of the trust line. The balance on the trust line itself _can_ be negative.

src/ripple/app/tx/impl/InvariantCheck.cpp
```diff
@@ -717,4 +718,62 @@ NFTokenCountTracking::finalize(
...
+        if (trustlinesChanged > 1)
+        {
+            JLOG(j.fatal())
+                << "Invariant failed: more than one trustline changed.";
+            return false;
+        }
+
+        AccountID const issuer = tx.getAccountID(sfAccount);
+        STAmount const amount = tx.getFieldAmount(sfAmount);
+        AccountID const holder = amount.getIssuer();
```
seelabs, line 755: Consider:
```
        AccountID const& holder = amount.getIssuer();
```

seelabs (commented):
Initial set of comments.

src/ripple/app/tx/impl/Clawback.cpp
shawnxie999: is there a reason why we preference reference here?

src/ripple/app/tx/impl/Clawback.cpp
shawnxie999: const reference (`AccountID const& holder = clawAmount.getIssuer(); `) actually doesn't work

The following line `clawAmount.setIssuer(issuer);` updates the the issuer field, which changes the value that `holder` stores.

And it makes the assertion   `assert(uSenderID != uReceiverID);` in `rippledCredit` fail .

src/ripple/app/tx/impl/Clawback.cpp
shawnxie999: i think passing `true` for `bCheckIssuer` would work too

src/ripple/app/tx/impl/Clawback.cpp
seelabs: It save a copy. The perf difference is probably unmeasurable, but there's no dowside to not copying, so we may as well not copy.

src/ripple/app/tx/impl/Clawback.cpp
seelabs: I'm fine passing `true` (the parameter only effects debugging anyway).

src/ripple/app/tx/impl/InvariantCheck.h
seelabs: Thanks for the latest patch. I think this is ready to be approved, but I would like to re-write this comment first. Thanks!

seelabs (approved):
:+1:

src/ripple/app/tx/impl/Clawback.cpp
ximinez: Even better:
```
STAmount const balance = sleRippleState->at(sfBalance);
```

shawnxie999:
@nbougalis do you mind revisiting the PR to see it looks acceptable? Thanks

intelliot:
(note: nbougalis intends to review by this weekend)

nbougalis (approved):

PR #4556, opened 2023-06-05
Use Artifactory remote in nix workflow

thejohnfreeman:
Ripple has created a semi-public Artifactory (thanks @shichengripple001 and team!) to hold dependency binaries for our builds. I was able to rewrite our `nix` workflow to use it and cut the time down to a mere [21 minutes](https://github.com/thejohnfreeman/rippled/actions/runs/5178187319). This workflow should continue to work (just more slowly) for forks that do not have access to the Artifactory.

The new bottleneck is the `macos` workflow which took [31 minutes](https://github.com/thejohnfreeman/rippled/actions/runs/5178187323). There is no low-hanging fruit to reduce that time though. AWS sells only one size of macOS instance. The next best improvement we can make is to fix parallel testing on macOS

### DIFF
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -67,11 +67,11 @@ jobs:
           conan profile update env.CC=${{ matrix.profile.cc }} default
           conan profile update env.CXX=${{ matrix.profile.cxx }} default
           conan profile update conf.tools.build:compiler_executables='{"c": "${{ matrix.profile.cc }}", "cpp": "${{ matrix.profile.cxx }}"}' default
-      - name: try to add ripple Conan remote
+          conan remote add ripple ${{ vars.CONAN_URL }} --insert 0
+      - name: try to authenticate to ripple Conan remote
         id: remote
         continue-on-error: true
         run: |
-          conan remote add ripple ${{ secrets.CONAN_URL }} --insert 0
           conan user --remote ripple ${{ secrets.CONAN_USERNAME }} --password ${{ secrets.CONAN_TOKEN }}
       - name: archive profile
         # Create this archive before dependencies are added to the local cache.
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -12,7 +12,8 @@ on: [push, pull_request]
 # and builds and caches binaries for all the dependencies.
 # If an Artifactory remote is configured, they are cached there.
 # If not, they are added to the GitHub artifact.
-# GitHub's "cache" action has a size limit that is too small.
+# GitHub's "cache" action has a size limit (10 GB) that is too small
+# to hold the binaries if they are built locally.
 # We must use the "{upload,download}-artifact" actions instead.
 #
 # The second phase has a job in the matrix for each test configuration.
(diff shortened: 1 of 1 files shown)


intelliot:
@thejohnfreeman can you confirm that this is ready for review?

I think the IP-based URL is fine for now; it can be updated to a human-readable DNS name, when available, in a separate future PR.

legleux (approved):
Worked quick!

.github/workflows/nix.yml
```diff
@@ -1,6 +1,24 @@
...
+# depends on the first ("dependencies").
+#
+# The first phase has a job in the matrix for each combination of
+# variables that affects dependency ABI:
+# platform, compiler, and configuration.
+# It creates a GitHub artifact holding the Conan profile,
+# and builds and caches binaries for all the dependencies.
+# If an Artifactory remote is configured, they are cached there.
+# If not, they are added to the GitHub artifact.
+# GitHub's "cache" action has a size limit that is too small.
```
ximinez, line 15: My understanding is that the cache limit is 10Gb total. These artifacts are < 200Kb. Is that only because the Artifactory remote is available? If so, I suggest rephrasing this as `GitHub's "cache" action has a size limit that is too small to hold the binaries if they are built locally.`

.github/workflows/nix.yml
```diff
@@ -47,19 +69,40 @@ jobs:
...
-        uses: actions/checkout@v3
-      - name: dependencies
+          # Do not quote the URL. An empty string will be accepted (with
+          # a non-fatal warning), but a missing argument will not.
+          conan remote add ripple ${{ env.CONAN_URL }} --insert 0
+      - name: try to authenticate to ripple Conan remote
+        id: remote
+        continue-on-error: true
+        run: |
+          conan user --remote ripple ${{ secrets.CONAN_USERNAME }} --password ${{ secrets.CONAN_TOKEN }}
```
ximinez, line 79: As discussed elsewhere, to avoid the confusing `Process completed with exit code 1` annotation on the summary screen, it may be worth rewriting this to set a status variable and so that this step always succeeds, then check that that variable in the later steps.

e.g.
```
echo outcome=$(conan user --remote ripple ${{ secrets.CONAN_USERNAME }} --password ${{ secrets.CONAN_TOKEN }} && echo "success" || echo "failure") | tee ${GITHUB_OUTPUT}
```
Then the checks would be `steps.remote.outputs.outcome == 'success')`, etc.

.github/workflows/nix.yml
```diff
@@ -47,19 +69,40 @@ jobs:
...
+      - name: archive profile
+        # Create this archive before dependencies are added to the local cache.
+        run: tar -czf conan.tar -C ~/.conan .
+      - name: list missing binaries
+        id: binaries
+        # Print the list of dependencies that would need to be built locally.
+        # A non-zero number means we have "failed" to cache binaries remotely.
```
ximinez, line 86: How about: `A non-zero number of dependencies in the list means we have "failed" to cache binaries remotely.`

ximinez (changes requested):

ximinez (approved):
Nice work!

PR #4558, opened 2023-06-05
Trivial: add comments for NFToken-related invariants

scottschurr:

The docs folks rely on comments describing the invariants.  Two invariants had been added without comments.  This pull request adds the comments so they will be available for docs folks into the future.

### Context of Change

Docs folks were updating the invariants documentation and noted some information was missing.  This adds the information they need.

### Type of Change

- [X] Documentation Updates

### DIFF
--- a/src/ripple/app/tx/impl/InvariantCheck.h
+++ b/src/ripple/app/tx/impl/InvariantCheck.h
@@ -318,6 +318,17 @@ public:
         beast::Journal const&);
 };
 
+/**
+ * @brief Invariant: Validates several invariants for NFToken pages.
+ *
+ * The following checks are made:
+ *  - The page is correctly associated with the owner.
+ *  - The page is correctly ordered between the next and previous links.
+ *  - The page contains a valid number of NFTokens.
+ *  - The NFTokens on this page do not belong on a lower or higher page.
+ *  - The NFTokens are correctly sorted on the page.
+ *  - Each URI, if present, is not empty.
+ */
 class ValidNFTokenPage
 {
     bool badEntry_ = false;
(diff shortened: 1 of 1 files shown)


src/ripple/app/tx/impl/InvariantCheck.h
```diff
@@ -342,6 +353,19 @@ class ValidNFTokenPage
         beast::Journal const&);
 };
 
+/**
+ * @brief Invariant: Validates counts of NFTokens after all transaction types.
+ *
+ * The following checks are made:
+ *  - The number of minted or burned tokens may only be changed by
```
DennisDawson, line 360: For clarity, I avoid the use of _may_, using _can_ or _might_ instead. I would say "...burned tokens can only be changed...."

src/ripple/app/tx/impl/InvariantCheck.h
```diff
@@ -342,6 +353,19 @@ class ValidNFTokenPage
...
+ *    NFTokenMint or NFTokenBurn transactions.
+ *  - A successful NFTokenMint must increase the number of NFTokens.
+ *  - A failed NFTokenMint must not change the number of minted NFTokens.
+ *  - An NFTokenMint transaction may not change the number of burned NFTokens,
```
DennisDawson, line 364: "An NFTokenMint transaction cannot change the number of burned NFTokens..." or "An NFTokenMint transaction must not change the number of burned NFTokens...."

One reason for this is that you might actually mean that the transaction _might_ or _might not_ change the number of burned NFTokens.

Also, the other lines end with a period, rather than a comma.

src/ripple/app/tx/impl/InvariantCheck.h
```diff
@@ -342,6 +353,19 @@ class ValidNFTokenPage
...
+ *  - A successful NFTokenBurn must increase the number of burned NFTokens.
+ *  - A failed NFTokenBurn must not change the number of burned NFTokens.
+ *  - An NFTokenBurn transaction may not change the number of minted NFTokens.
```
DennisDawson, line 367: Again - I would use either _might_ not or _cannot_ change the number.

DennisDawson (approved):
Consider my suggestions, otherwise good to go.

scottschurr:
Thanks for the comments @DennisDawson.  I believe I've addressed them.  Let me know what you think.

src/ripple/app/tx/impl/InvariantCheck.h
```diff
@@ -318,6 +318,17 @@ class ValidNewAccountRoot
         beast::Journal const&);
 };
 
+/**
+ * @brief Invariant: Validates several invariants for NFToken pages.
+ *
+ * The following checks are made:
+ *  - The page is correctly associated with the owner.
+ *  - The page is correctly ordered between the next and previous links.
+ *  - The page contains a valid number of NFTokens.
```
ledhed2222, line 327: might as well put in what a valid number is

ledhed2222 (approved):
LGTM - but one non-blocking comment

scottschurr:
Added the allowed numbers of `NFToken`s per page.

DennisDawson (approved):
Looks great to me - thanks for your patience with me.

intelliot (approved):

PR #4566, opened 2023-06-09
Accept all valid currency codes in API

### LINKED ISSUE #4112: "book_offers" and "ripple_path_find" return no result for 3 letter currencies with a lower case letter (Version: 1.8.5)
## Issue Description
The rippled method to get existing offers ("book_offers") does not return any offers for tokens with a 3 letter currency code where, at least, one letter is lower case.
This does NOT affect currencies where all letters are capital or the currency code is > 3 letters (HEX code)

## Steps to Reproduce
You can simply reproduce this behavior by calling the "book_offers" method with one of the following tokens in the "TakerGets" field:

Example 1:
Issuer: rGbNLwrYNuPim39crh1bjyyauefF1VHPoj
Currency: xSD
Current DEX offers: 97

Example 2:
Issuer: raZ87rZoVjgg8oc5LijSigyN8rbutXGdbw
Currency: Lil
Current DEX offers: 10

Example 3:
Issuer: rJXesdkxGHDtzTEgxxbo7h9C1WyVr8NNgw
Currency: Vtc
Current DEX offers: 3

## Expected Result
Rippled should return the offers for these tokens when calling the "book_offers" method.

## Actual Result
Calling the "book_offers" method results in an empty response:
```
 {
   "result" : {
      "ledger_current_index" : 70037039,
      "offers" : [],
      "status" : "success",
      "validated" : false
   }
}
```

## Environment
I am using Ubuntu 20.04 with rippled 1.8.5

thejohnfreeman:
A few methods, including `book_offers`, take currency codes as parameters. The XRPL doesn't care if the letters in those codes are lowercase or uppercase, as long as they come from an alphabet defined internally. rippled doesn't care either, when they are submitted in a hex representation. When they are submitted in an ASCII string representation, rippled, but not XRPL, is more restrictive, preventing clients from interacting with some currencies already in the XRPL.

This change gets rippled out of the way and lets clients submit currency codes in ASCII using the full alphabet.

Fixes #4112


intelliot:
note: @ximinez is planning to review.

@nixer89 will you review this?

nixer89 (approved):
lgtm

src/ripple/net/impl/RPCCall.cpp
```diff
@@ -121,7 +121,8 @@ class RPCParser
     static Json::Value
     jvParseCurrencyIssuer(std::string const& strCurrencyIssuer)
     {
-        static boost::regex reCurIss("\\`([[:alpha:]]{3})(?:/(.+))?\\'");
+        static boost::regex reCurIss(
+            "\\`([][:alnum:]<>(){}[|?!@#$%^&*]{3})(?:/(.+))?\\'");
```
ckeshava, line 125: How is this regex working correctly? Is the `tilde` character (third character from the beginning) replaceable with the single-quote (last but one character) inside the regex?

can we include a comment referring to [full alphabet](https://github.com/XRPLF/rippled/blob/5d011c7e6b399b988ad193d199e6e3dba92e49ad/src/ripple/protocol/impl/UintTypes.cpp#L33-L39) or reuse the `detail::isoCharSet` variable? [I was thinking: "\\`((detail::isoCharSet){3})(?:/(.+))?\\'", but I'm not sure]


also, I've used regex libraries in other languages, but this one baffles me. what is the purpose of the suffix of this regular expression? [`?:/(.+))?`]

thejohnfreeman: This is a `boost::regex`. It uses Perl syntax, which the library [documents](https://www.boost.org/doc/libs/1_82_0/libs/regex/doc/html/boost_regex/syntax/perl_syntax.html) online. Look at the section "[Buffer boundaries](https://www.boost.org/doc/libs/1_82_0/libs/regex/doc/html/boost_regex/syntax/perl_syntax.html#boost_regex.syntax.perl_syntax.buffer_boundaries)". (The character `` ` `` is called a [backtick](https://en.wikipedia.org/wiki/Backtick); [tilde](https://en.wikipedia.org/wiki/Tilde) is `~`.) `` \` `` matches the start of a buffer and `\'` matches the end of a buffer, so no, they cannot be replaced.

You could try to construct a regex string where `detail::isoCharSet` sits in a character class, but you would have to change the order within `detail::isoCharSet` because `]` cannot appear in the middle of a character class; it can only appear as the first character in the class. I don't think that factoring is worth it, though.

`(?:...)` is a non-capturing group. It groups subatoms into a larger atom without creating a capture group. So `(?:/(.+))?` optionally matches a forward slash and a non-empty sequence of characters, capturing the characters after the slash into a group.

ckeshava: ok, this is helpful. thank you

ckeshava:
```suggestion
// The below regex matches the pattern as specified in UIntTypes.cpp:detail::isoCharSet. It is utilizes boost::regex and adheres to the Perl syntax for regular expressions.
            "\\`([][:alnum:]<>(){}[|?!@#$%^&*]{3})(?:/(.+))?\\'");
```

intelliot: @thejohnfreeman to accept, modify, or decline suggested comment

thejohnfreeman: I've pushed an alternative comment.

ckeshava:
lgtm

intelliot:
@thejohnfreeman at your convenience, please bring this branch up-to-date with `develop` (or give me permission to do so)

PR #4568, opened 2023-06-12
Enable api_version 2

arihantkothari:

Enabling `api_version` 2 by bumping up `apiMaximumSupportedVersion` to 2 and fixing `AccountInfo` tests that were failing as a consequence.

### Context of Change
Changing error code leads to breaking changes for error handling in applications relying on those error codes. Hence, all these changes should be upgraded to the next `api_version`, so that existing users get the existing behaviour, and users who request a higher `api_version`  get the new behaviour.
More context: #4552 

### Type of Change

- [ x] Tests (updated existing tests)


src/test/rpc/AccountInfo_test.cpp
```diff
@@ -89,11 +89,12 @@ class AccountInfo_test : public beast::unit_test::suite
         Account const alice{"alice"};
         env.fund(XRP(1000), alice);
 
-        auto const withoutSigners =
-            std::string("{ ") + "\"account\": \"" + alice.human() + "\"}";
+        auto const withoutSigners = std::string("{ ") +
+            "\"api_version\": 1, \"account\": \"" + alice.human() + "\"}";
```
godexsoft, line 93: I may be wrong here but it seems like `api_version` support is broken somehow. 
I'd expect `v1` to be used when no version is specified (as the comment suggests in `RPCHelpers.h`). In turn, if it would use the right version, the test should not be breaking and you do not need to specify `api_version` here and below.

src/ripple/rpc/impl/RPCHelpers.h
```diff
@@ -241,7 +241,7 @@ extern beast::SemanticVersion const lastVersion;
 constexpr unsigned int apiInvalidVersion = 0;
 constexpr unsigned int apiVersionIfUnspecified = 1;
 constexpr unsigned int apiMinimumSupportedVersion = 1;
-constexpr unsigned int apiMaximumSupportedVersion = 1;
+constexpr unsigned int apiMaximumSupportedVersion = 2;
 constexpr unsigned int apiBetaVersion = 2;
```
ximinez, line 245: As I wrote over [on #4571](https://github.com/XRPLF/rippled/pull/4571#discussion_r1228663775), once a version of `rippled` is released with `apiMaximumSupportedVersion` is set to 2, then we are saying that API v2 is **Done**. Once that happens we can not legitimately make any further changes to API version 2. To the best of my knowledge, API version 2 is still in beta, under active development, and still changing. 

Now, if you (and #4571 and #4552) are saying that version 2 is done, let's talk about version 2 being done. This will be the first I've heard about it.

If you want a node to properly handle API v2 requests, the correct way to do that, for now, is with the `[beta_rpc_api]` config option. And if you want to do it in unit tests, set it up in the config similar to how it's done in https://github.com/XRPLF/rippled/blob/develop/src/test/rpc/AccountInfo_test.cpp#L209-L212, or just by setting `cfg.BETA_RPC_API=true;` directly.

ximinez (changes requested):

src/ripple/rpc/impl/RPCHelpers.h
ximinez: Please see also #4573.

intelliot: Let's say in 1.12 that API v2 is done. Any changes to the API after that point will be for api_version 3+.

arihantkothari: It seems in that case, we might need to work (or get clarity) on the versioning scheme a bit more and get answers to the questions that @thejohnfreeman posted [here](https://github.com/XRPLF/rippled/pull/4552#discussion_r1230276163)

intelliot:
Discussed today and decided to hold this for 1.13 or later.

arihantkothari:
> Discussed today and decided to hold this for 1.13 or later.

Sorry, I missed this comment. Okay! Thanks for the update @intelliot !

arihantkothari:
Leaving this comment here (this is from a while back):

Changing `apiMaximumSupportedVersion` to `2` would result in the failure of the `account_info` version 1 test because the unit tests typically operate based on `apiMaximumSupportedVersion` by default. This behavior is due to the calls made within the code flow, specifically` Env::do_rpc -> rpcClient -> rpcCmdLineToJson -> insert_api_version`, where the `api_version` is set to `RPC::apiMaximumSupportedVersion`. It appears that the command line specific function `rpcCmdLineToJson` is called in order to keep the unit tests aligned with the latest supported API version by default, because the command line exhibits different behavior. The question arises as to whether we should maintain this behavior or not.

src/test/rpc/AccountInfo_test.cpp
arihantkothari: For your second and third question, we actually had a bunch of conversation in this [PR](https://github.com/XRPLF/rippled/pull/4552) and elsewhere. I actually wrote a document summarizing the discussion. I'll send you over.

But I added api_version 1 in this test since it was defaulting to version 2 (maxSupportedVersion) which led it to fail. I have described why this happens in the document as well.

With the current state, all api version 1 tests would have to be explicitly set to api_version 1. Reason why I kept this PR to be a draft, because that is a very bad solution. So if we release api version 3,4,5 those will become the default.

However, it is important to note - this is just happening in tests. In reality, the behaviour is opposite. Default behaviours should be api version 1. This is how it works outside of the tests. For example: we need to specify api_version: 2 in the payload if you want to use v2. But v1 is default.

arihantkothari: But for the first comment, yeah - I stopped working on this after a while due to other stuff. And also because I am not sure what's a good solution to this.

ckeshava: Okay, just to clarify here's @ximinez 's comment on a similar question about default API version for command line: https://github.com/XRPLF/rippled/pull/4759#issuecomment-1771618723

It appears that we are defaulting to `RPC::apiMaximumSupportedVersion` (stealing Ed's [example](https://github.com/XRPLF/rippled/blob/89780c8e4fd4d140fcb912cf2d0c01c1b260539e/src/ripple/net/impl/RPCCall.cpp#L1479C36-L1479C67))

So, irrespective of unit-tests or real-world RPC calls, the API version defaults to `RPC::apiMaximumSupportedVersion`

arihantkothari: No, commandline and unit tests uses max version by default but RPC from my understanding uses the default version which is `apiVersionIfUnspecified` which is currently 1

ckeshava: okay, it appears that the above comment holds true for command line usage and not general RPC calls. It has been detailed in the third paragraph of this [document](https://docs.google.com/document/d/1qEMoDyi7or-7ODEfVtnaFlcFSb0NsE2vnvEw6MIJ3Zo/edit?pli=1)

arihantkothari: that's right

ckeshava: ok, thanks for the clarification 👍

intelliot:
update: @pwang200 will take a look at this and potentially open a new PR

arihantkothari:
Closing this PR since #4785  is now open.

Issue #4545, opened 2023-05-26
[account_tx with ledger_index_max, ledger_index_min and ledger_index do not return error] (Version: [1.11.0-rc2])


mounikakun:
## Steps to Reproduce

`{
    "method": "account_tx",
    "params": [
        {
            "account": "rH9pnDCtgAbFjozbZQuJWvPM8QHweXgZhj",
            "ledger_index_max": 28953065,
            "ledger_index_min": 23333389,
            "ledger_index": 28953075
        }
    ]
}`
## Expected Result

`invalidParams`
## Actual Result

No error

mounikakun:
Issue exists for ledger_index_max, ledger_index_min and ledger_hash also.

Issue #4288, opened 2022-08-30
“account_tx” with ledger_index_max  out of range do not return "lgrIdxMalformed" error.(Version: ["1.9.2"])


mounikakun:
## Issue Description
“account_tx” with the ledger specified by the ledger_index_max does not exist, or if it does exist but the server does not have it should return expected error: lgrIdxMalformed
--

## Sample request
`{
   "method": "account_tx",
   "params": [
       {
           "account": "r4s7P9WUU6AZ9DB315GWKtBaaDji9Zst8h",
           "binary": false,
           "forward": false,
           "ledger_index_max": 31089994,
           "ledger_index_min": -1
       }
   ]
}`

## Expected Result
`"error": "lgrIdxMalformed"`

## Actual Result
`{
    "result": {
        "account": "r4s7P9WUU6AZ9DB315GWKtBaaDji9Zst8h",
        "ledger_index_max": 21320746,
        "ledger_index_min": 21316540,
        "limit": 0,
        "status": "success",
        "transactions": [],
        "validated": true
    }
}`

ManiMounikaKunasani:
“account_tx” with the ledger specified by the ledger_index_min does not exist, or if it does exist but the server does not have it should return expected error: lgrIdxMalformed

Sample request:
`{
   "method": "account_tx",
   "params": [
       {
           "account": "r4s7P9WUU6AZ9DB315GWKtBaaDji9Zst8h",
           "binary": false,
           "forward": false,
           "ledger_index_max": -1,
           "ledger_index_min": 9499032
       }
   ]
}`

Expected Result
`"error": "lgrIdxMalformed"`

Actual Result
`{
    "result": {
        "account": "r4s7P9WUU6AZ9DB315GWKtBaaDji9Zst8h",
        "ledger_index_max": 21320817,
        "ledger_index_min": 21316540,
        "limit": 0,
        "status": "success",
        "transactions": [],
        "validated": true
    }
}`

ManiMounikaKunasani:
“account_tx” with invalid ledger_index returns "invalidParams" instead of "lgrNotFound" error.

PR #4571, opened 2023-06-13
Added error messages on  AccountTx

PeterChen13579:
Fixed error messages for AccountTx method for Github Issue #[4545](https://github.com/XRPLF/rippled/issues/4545) and #[4288](https://github.com/XRPLF/rippled/issues/4288) 

### Context of Change
Certain input for AccountTx method should of error'ed out

### Type of Change
Invalid request by user results in Error message 

Before and After: refer to github link. Using "api_version": 2 calls the updated error messages

Updated existing unit test to check error based off of VPI versioning as adding error messages is considered a breaking c hange.


src/test/rpc/AccountTx_test.cpp
```diff
@@ -166,11 +166,21 @@ class AccountTx_test : public beast::unit_test::suite
 
             p[jss::ledger_index_min] = 0;
             p[jss::ledger_index_max] = 100;
-            BEAST_EXPECT(hasTxs(env.rpc("json", "account_tx", to_string(p))));
+            RPC::apiMaximumSupportedVersion == 1
+                ? BEAST_EXPECT(
+                      hasTxs(env.rpc("json", "account_tx", to_string(p))))
+                : BEAST_EXPECT(isErr(
+                      env.rpc("json", "account_tx", to_string(p)),
+                      rpcLGR_IDX_MALFORMED));
```
thejohnfreeman, line 174: You should copy these tests for an explicit `api_version == 1`, like [this](https://github.com/XRPLF/rippled/pull/4552/files#diff-d49d692a95cf0496bc5f953d37f2535730d925b6ddce2a0eb004f2f6ecef6f05R1232-R1256).

thejohnfreeman (changes requested):

src/ripple/rpc/impl/RPCHelpers.h
```diff
@@ -241,7 +241,7 @@ extern beast::SemanticVersion const lastVersion;
 constexpr unsigned int apiInvalidVersion = 0;
 constexpr unsigned int apiVersionIfUnspecified = 1;
 constexpr unsigned int apiMinimumSupportedVersion = 1;
-constexpr unsigned int apiMaximumSupportedVersion = 1;
+constexpr unsigned int apiMaximumSupportedVersion = 2;
 constexpr unsigned int apiBetaVersion = 2;
```
ximinez, line 245: Don't change the `apiMaximumSupportedVersion` until the updated version is locked in and final and ready for release. We have `apiBetaVersion` variable right here and the `[beta_rpc_api]` config option available to allow testing the changes. (It might not be a bad idea to enable `BETA_RPC_API = true` by default in envconfig.cpp.)

src/test/rpc/AccountTx_test.cpp
ximinez: To follow on @thejohnfreeman's suggestion, loop over both versions for these tests, because we need to keep supporting version 1 indefinitely and ensure it's working correctly.

ximinez (changes requested):

src/ripple/rpc/impl/RPCHelpers.h
PeterChen13579: Just wondering what do you mean by "until the updated version is locked in and final and ready for release."? Do you you mean until PR #[4522](https://github.com/XRPLF/rippled/pull/4552) is merged into develop, so I don't have to change apiMaximumSupportedVersion to 2?

ximinez: Please see also #4573.

thejohnfreeman (approved):

src/test/rpc/AccountTx_test.cpp
```diff
@@ -244,6 +248,83 @@ class AccountTx_test : public beast::unit_test::suite
...
 
+    void
+    testParametersApiV2()
+    {
+        using namespace test::jtx;
+        // test for API version 2
+        Env env{*this, envconfig([](std::unique_ptr<Config> c) {
+                    c->loadFromString("\n[beta_rpc_api]\n1\n");
+                    return c;
+                })};
```
arihantkothari, line 259: Since #4573 is merged, this can probably be changed to `Env env{*this};` ?

arihantkothari (changes requested):

src/test/rpc/AccountTx_test.cpp
ximinez: Agreed.

src/test/rpc/AccountTx_test.cpp
```diff
@@ -244,6 +248,80 @@ class AccountTx_test : public beast::unit_test::suite
         }
     }
 
+    void
+    testParametersApiV2()
```
arihantkothari, line 252: Might wanna:
- Add a parameter for api_version.
- Call this method in a loop over versions from apiMinimumSupportedVersion to apiBetaVersion.
- Remove the other method.
- Make the assertion on the error value conditional on the api_version parameter, using condition < 2 for the old behavior and else for the new behavior.

arihantkothari (changes requested):

src/test/rpc/AccountTx_test.cpp
arihantkothari: https://github.com/XRPLF/rippled/pull/4552#discussion_r1242329180

src/ripple/rpc/handlers/AccountTx.cpp
```diff
@@ -53,9 +53,23 @@ using LedgerSpecifier = RelationalDatabase::LedgerSpecifier;
 
 // parses args into a ledger specifier, or returns a Json object on error
 std::variant<std::optional<LedgerSpecifier>, Json::Value>
-parseLedgerArgs(Json::Value const& params)
+parseLedgerArgs(RPC::Context& context, Json::Value const& params)
 {
     Json::Value response;
+    // if ledger_index_min or max is specified, then ledger_hash or ledger_index
+    // should not be specified. Error out if it is
+    if (context.apiVersion > 1)
```
arihantkothari, line 61: not so important, but I'm wondering if it'll be nice to make constants like API_VERSION_1 and API_VERSION_2 instead of using raw numbers 1 and 2 since we are using it in a lot of places ?

src/test/rpc/AccountTx_test.cpp
```diff
@@ -108,7 +108,7 @@ class AccountTx_test : public beast::unit_test::suite
     };
 
     void
-    testParameters()
+    testParameters(unsigned int apiVersion)
```
arihantkothari, line 111: Not related to this PR, but wondering why the test cases in this file do not have `testcase(testCaseDesctiption)` ?

arihantkothari (approved):

src/test/rpc/AccountTx_test.cpp
thejohnfreeman: My guess is that we don't have any written, enforced guidelines for tests.

src/ripple/rpc/handlers/AccountTx.cpp
thejohnfreeman: I don't think it's worth it when we're just using consecutive integers to name the versions.

src/test/rpc/AccountTx_test.cpp
arihantkothari: I think it might be useful to add `testcase(testCaseDescription + v{apiVersion})` so that users can see if the unit test is actually running on different versions.

src/ripple/rpc/impl/RPCHelpers.h
intelliot: Just to close the loop on this: API v2 will be considered "Done" in the 1.12 release. Any breaking changes in 1.13 will be API v3.

src/test/rpc/AccountTx_test.cpp
thejohnfreeman: I think our test framework and tests are long overdue for a redesign with firm and written guidelines, but I won't worry about the inconsistency in this PR.

PR #4573, opened 2023-06-13
Enable the Beta RPC API (v2) for all unit tests:

ximinez:

I noticed recently a handful of PRs (#4552, #4568, and #4571) that are changing the `apiMaximumSupportedVersion` value from 1 to 2, which would indicate that version two is finalized and out of beta, apparently only to enable API version testing in unit tests.

IMHO, the proper way to enable versioning in unit tests is to enable the `BETA_RPC_API` flag in the config. For simplicity, this PR sets that flag in the default unit test config, so now the beta (v2) is available to all unit tests. The default will still be 1, but any new unit tests should be able to specify an API version and work as expected without any hassle.

### Type of Change

## Test Plan

As this code only effects unit tests, the only testing necessary should be to verify that unit tests pass.

### DIFF
--- a/src/test/jtx/impl/envconfig.cpp
+++ b/src/test/jtx/impl/envconfig.cpp
@@ -49,6 +49,9 @@ setupConfigForUnitTests(Config& cfg)
     cfg.FEES.account_reserve = XRP(200).value().xrp().drops();
     cfg.FEES.owner_reserve = XRP(50).value().xrp().drops();
 
+    // The Beta API (currently v2) is always available to tests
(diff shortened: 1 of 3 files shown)


ximinez:
I nominate the authors of the aforementioned PRs as reviewers for this one: @PeterChen13579 and @arihantkothari.

arihantkothari (approved):

PeterChen13579 (approved):

intelliot (approved):

PR #4575, opened 2023-06-16
Reject negative command line quorum values.

mtrippled:
Check for negative values when setting the quorum from the command line
and reject.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

There's already a test for negative values submitted for them quorum parameter, so it should stop failing.

### DIFF
--- a/src/ripple/app/main/Main.cpp
+++ b/src/ripple/app/main/Main.cpp
@@ -48,6 +48,7 @@
 
 #include <cstdlib>
 #include <iostream>
+#include <sstream>
 #include <stdexcept>
 #include <utility>
 
@@ -375,7 +376,7 @@ run(int argc, char** argv)
         po::value<std::string>(),
         "Specify the node identity for this server.")(
         "quorum",
-        po::value<std::size_t>(),
+        po::value<std::int64_t>(),
         "Override the minimum validation quorum.")(
         "reportingReadOnly", "Run in read-only reporting mode")(
         "silent", "No output to the console after startup.")(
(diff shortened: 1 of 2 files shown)


src/ripple/app/main/Main.cpp
```diff
@@ -704,10 +705,12 @@ run(int argc, char** argv)
     {
         try
         {
-            config->VALIDATION_QUORUM = vm["quorum"].as<std::size_t>();
-            if (config->VALIDATION_QUORUM == std::size_t{})
+            config->VALIDATION_QUORUM = vm["quorum"].as<std::int64_t>();
+            if (*config->VALIDATION_QUORUM < 1)
```
gregtatcam, line 709: Should this be `< 0`? The PR says reject negative and the original check is for negative.

src/ripple/core/Config.h
```diff
@@ -203,7 +203,7 @@ class Config : public BasicConfig
     int PATH_SEARCH_MAX = 3;
 
     // Validation
-    std::optional<std::size_t>
+    std::optional<std::int64_t>
```
nbougalis, line 206: I'd recommend leaving `VALIDATION_QUORUM` as an `std::size_t`, since negative values are not allowed and/or meaningful. It's a way of "self-documenting" the code.

src/ripple/app/main/Main.cpp
nbougalis: I don't think so: a quorum value of `0` doesn't make sense; notice, too, that the original code checked if the value was `0`. 
With that said, I would suggest replacing this code with `lexicalCast`:

```
try
{
    config->VALIDATION_QUORUM = lexicalCast<std::size_t>(vm["quorum"].as<std::string>());

    if (config->VALIDATION_QUORUM == 0)
        throw std::domain_error(s);
}
/* existing catch handlers */
```

Also, this bug is present because `boost::program_options::variable_value::as()` is broken. Per the [documentation](https://www.boost.org/doc/libs/1_56_0/doc/html/boost/program_options/variable_value.html#idp219097696-bb), it should be throwing a `boost::bad_any_cast` exception if a negative value is passed, but it doesn't. So I think this merits a bug report, if anyone feels s inclined.

nbougalis (commented):
Left some comments, at @mtrippled's discretion.

mtrippled:
This is a duplicate, and has already been merged as c5003969dee3fe7fbc1e0a14afab1ccbb8782d82

Issue #4539, opened 2023-05-26
[account_info with invalid signer_lists value do not return an error] (Version: [1.11.0-rc2])


mounikakun:
## Steps to Reproduce
`{
    "method": "account_info",
    "params": [
        {
            "account": "r3kmBwpGRwRoZGM64zkkwoDFB7GNJqxD8Y",
            "strict": true,
            "signer_lists": "ghj"
        }
    ]
}
`
## Expected Result
`{
    "result": {
        "error": "invalidParams",
        "error_code": 31,
        "error_message": "Invalid parameters.",
        "status": "error",
        "type": "response",
        "request": {
            "method": "account_info",
            "params": [
                {
                    "account": "r3kmBwpGRwRoZGM64zkkwoDFB7GNJqxD8Y",
                    "strict": true,
                    "signer_lists": "ghj"
                }
            ]
        }
    },
}`

## Actual Result
`{
    "result": {
        "account_data": {
            "Account": "r3kmBwpGRwRoZGM64zkkwoDFB7GNJqxD8Y",
            "Balance": "39999980",
            "Flags": 0,
            "LedgerEntryType": "AccountRoot",
            "OwnerCount": 1,
            "PreviousTxnID": "F2A7FFCD43B5E90EB3AF0369C14D3AC69E168E148DEDB8739C8DF9FB81F6697E",
            "PreviousTxnLgrSeq": 28951886,
            "Sequence": 28951883,
            "index": "A86A0E675F0CF1771A595A5FB0D9DE90F9E3E4718E23684BE94D92F642D54C48",
            "signer_lists": [
                {
                    "Flags": 65536,
                    "LedgerEntryType": "SignerList",
                    "OwnerNode": "0",
                    "PreviousTxnID": "F2A7FFCD43B5E90EB3AF0369C14D3AC69E168E148DEDB8739C8DF9FB81F6697E",
                    "PreviousTxnLgrSeq": 28951886,
                    "SignerEntries": [
                        {
                            "SignerEntry": {
                                "Account": "rw416oQqMe3haC6ZnagfQaXJJdvMaJo2aG",
                                "SignerWeight": 2
                            }
                        }
                    ],
                    "SignerListID": 0,
                    "SignerQuorum": 2,
                    "index": "E9DE0223C515A49AA2E3BCC71BCF052F34C5B4DD03A196FC3A3EE028385AD291"
                }
            ]
        },
        "account_flags": {
            "defaultRipple": false,
[...]

PR #4585, opened 2023-06-26
account_info with Invalid signer list returns invalid params

PeterChen13579:
Signer_list with invalid input should have error'ed out. Fixed this issue in this PR. (fix for issue [#4539](https://github.com/XRPLF/rippled/issues/4539))

### Context of Change
For API version 2, if signer_list has a value not of type boolean, it will return invalidParam error

### Type of Change

Refer to Github link above.

Added own unit test to test for signer_list in API v2


src/ripple/rpc/handlers/AccountInfo.cpp
```diff
@@ -125,6 +125,15 @@ doAccountInfo(RPC::JsonContext& context)
         }
         result[jss::account_flags] = std::move(acctFlags);
 
+        // The document states that signer_lists is a bool, however
+        // assigning any string value works. Do not allow this
+        // This for api v2 onwards
```
scottschurr, line 130: The tiniest of nits: The comment says "... allow this This for ...".  Consider removing the capitalized "This".

scottschurr (approved):
Looks good.  Thanks for adding the test!  I did leave one nit comment, but I'm good with this pull request either way.  Thanks for the contribution!

src/ripple/rpc/handlers/AccountInfo.cpp
```diff
@@ -125,6 +125,15 @@ doAccountInfo(RPC::JsonContext& context)
...
+        if (!params[jss::signer_lists].isBool() && context.apiVersion > 1)
```
ckeshava, line 131: do we need to check the `context.apiVersion` here? We assume that all the validators are running the latest version of the rippled codebase. Otherwise, the laggards will be incompatible with the rest of the network right?

ckeshava (approved):

src/ripple/rpc/handlers/AccountInfo.cpp
scottschurr: This is a situation where we don't need to worry about validators very much, at least on the main net.  This is an RPC query.  It's not typical for main net validators to be required also handle very many RPC queries.

The `context.api` can be a value passed in by the caller as shown in the test where it says `"\"api_version\": 2`.  So, yes, we should check the `api_version`.  It can be different from one RPC call to the next.

ckeshava: i see okay, thanks for the clarification 👍 
I'm guessing the version of rippled software is preserved in a different variable right?

scottschurr: Yes, the software version (e.g., `1.12.0-b1`) is elsewhere.  The code never needs to check the software version because it is compiled into the binary with whatever software version it has.  The `api_version` is a different kind of beast.

ckeshava: i see, okay

ckeshava (approved):

src/ripple/rpc/handlers/AccountInfo.cpp
```diff
@@ -127,7 +127,7 @@ doAccountInfo(RPC::JsonContext& context)
 
         // The document states that signer_lists is a bool, however
         // assigning any string value works. Do not allow this
-        // This for api v2 onwards
+        // this for api v2 onwards
```
scottschurr, line 130: :smile: Close.  If you read the sentence, it now reads "... allow this this for ...".  The word "this" appears two times in a row.  Now that neither of the instances of "this" are capitalized feel free to remove either one.

PeterChen13579: oops, I meant to write "this is for api v2 onwards". I just updated the comment for more clarity. Hope the updated one is fine 😄

scottschurr (approved):
The revised comment looks great.  Still :+1:

Issue #4588, opened 2023-06-26
`allowClawback` flag is missing from the `account_info.account_flags` response (Version: 1.12.0-beta1)


ckniffen:
## Issue Description
In rippled@1.11.0 a new object called `account_flags` was introduced to the `account_info` response.  An entry needs to be added for `lsfAllowClawback`

This likely was missed because work was being done on both PRs at the same time.

https://github.com/drlongle/rippled/blob/develop/src/ripple/rpc/handlers/AccountInfo.cpp#L80-L100

intelliot:
([Clawback implementation can be found here](https://github.com/XRPLF/rippled/pull/4553))

PR #4590, opened 2023-06-27
Add allowClawback flag for `account_info`

shawnxie999:
Clawback introduced an `allowClawback` flag for `AccountRoot` object. This PR updates the `account_info` API so that this flag is reflected in the API response.

Solves issue https://github.com/XRPLF/rippled/issues/4588

### Context of Change

([Clawback implementation can be found here](https://github.com/XRPLF/rippled/pull/4553))

### Example
```
{
   "result": {
      "account_data": {
         "Account": "rLqYRLKorwEWGzwCJLgtiCdKF416oGt4yx",
         "Balance": "99999999960",
         "Flags": 2147483648,
         "LedgerEntryType": "AccountRoot",
         "OwnerCount": 0,
         "PreviousTxnID": "6696AE4C3F6153E414B114E35EE7A232D03F7B5B8FCDA08A7F10250B49D8BA2C",
         "PreviousTxnLgrSeq": 264,
         "Sequence": 262,
         "index": "87F3974669983DE311498C7F34B2D40E00E9A8FCCCDAC6C026C91EC160934632"
      },
      "account_flags": {
         "allowClawback": true,
         "defaultRipple": false,
         "depositAuth": false,
         "disableMasterKey": false,
         "disallowIncomingXRP": false,
         "globalFreeze": false,
         "noFreeze": false,
         "passwordSpent": false,
         "requireAuthorization": false,
         "requireDestinationTag": false
      },
      "ledger_current_index": 266,
      "status": "success",
      "validated": false
   }
}
```

- [ ] Documentation Updates
- [ ] Release


src/test/rpc/AccountInfo_test.cpp
```diff
@@ -607,6 +632,8 @@ class AccountInfo_test : public beast::unit_test::suite
             ripple::test::jtx::supported_amendments()};
         testAccountFlags(allFeatures);
         testAccountFlags(allFeatures - featureDisallowIncoming);
+        testAccountFlags(
+            allFeatures - featureDisallowIncoming - featureClawback);
```
ckeshava, line 636: why are we removing the `featureClawback` from set of all features? Shouldn't that be enabled for us to run the unit tests?

src/test/rpc/AccountInfo_test.cpp
```diff
@@ -508,13 +508,16 @@ class AccountInfo_test : public beast::unit_test::suite
 
         Env env(*this, features);
         Account const alice{"alice"};
-        env.fund(XRP(1000), alice);
+        Account const bob{"bob"};
+        env.fund(XRP(1000), alice, bob);
```
ckeshava, line 512: is it required to fund `bob`'s account in order to test the account flags? 

Also, shouldn't the `clawback` feature be tested with a custom token (different from XRP) ?

src/test/rpc/AccountInfo_test.cpp
shawnxie999: Amendments are included by default I think. So it has already been tested on lines 633 and 634

src/test/rpc/AccountInfo_test.cpp
shawnxie999: Custom is only needed for trustlines, we don’t do that here. This line is to just fund bob so that it becomes a valid account

ckeshava: okay. But the `clawback` feature is not applicable for XRP token asset right?
Isn't it better to perform this check on an account with custom tokens, instead of XRP tokens?

It appears that we are only checking if certain flags are set/unset, so I guess it doesn't matter.

src/test/rpc/AccountInfo_test.cpp
ckeshava: oh okay, got it.

ckeshava:
it looks good to me. it's strange that only the macOS workflow is failing, but I couldn't access the logs.

ckeshava (approved):

seelabs (approved):
:+1:

intelliot:
> it's strange that only the macOS workflow is failing, but I couldn't access the logs.

Attempt #2 passed. I guess the earlier run failed due to either a glitch in GitHub, or because the macOS runner wasn't available at the time.

shawnxie999:
@intelliot  it failed because of the github downtime last week :-) 

It's ready to merge.

PR #4591, opened 2023-06-27
Fix deb package build

legleux:
Use GCC-11 and update GCC Conan profile

### Context of Change
This is a quick band aid to get this build fixed.
A goal of Q3 is to remove all this package building code.

Also renames debug packages `ddeb` --> `deb` to account for an Artifactory bug where the debug symbol packages with `.ddeb` extensions are not indexed and thus aren't visible in apt clients.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### DIFF
--- a/Builds/containers/packaging/dpkg/debian/rules
+++ b/Builds/containers/packaging/dpkg/debian/rules
@@ -15,12 +15,25 @@ override_dh_systemd_start:
 	dh_systemd_start --no-restart-on-upgrade
 
 override_dh_auto_configure:
+	apt install --yes gcc-11 g++-11
+	update-alternatives --install \
+		/usr/bin/gcc gcc /usr/bin/gcc-11 100 \
+		--slave /usr/bin/g++ g++ /usr/bin/g++-11 \
+		--slave /usr/bin/gcc-ar gcc-ar /usr/bin/gcc-ar-11 \
+		--slave /usr/bin/gcc-nm gcc-nm /usr/bin/gcc-nm-11 \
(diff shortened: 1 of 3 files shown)


Builds/containers/packaging/dpkg/debian/rules
```diff
@@ -38,12 +51,14 @@ override_dh_auto_configure:
...
 
 	conan install . \
+		--profile gcc \
 		--install-folder bld/rippled-reporting \
 		--build missing \
 		--build boost \
 		--build sqlite3 \
 		--build libuv \
 		--settings build_type=Release \
+		--settings compiler.cppstd=17 \
```
thejohnfreeman, line 61: We can't build non-reporting mode _without_ C++20, and we can't build reporting mode _with_ C++20? Is that right?

legleux: It seems to be the case. IIRC, there was an issue building Cassandra.

thejohnfreeman: Yeah, that's the issue I'm having:

> C++20 changes std::memory_order from an `enum` to an `enum class`. Datastax supplies a C++ driver for Cassandra, a dependency of reporting mode. it [defines](https://github.com/datastax/cpp-driver/blob/84389657d4504604c5d05b87769ce3ce983a2296/src/atomic/atomic_std.hpp#L24-L31) an `enum MemoryOrder` whose values are copied from `std::memory_order`, but using values from an `enum class` is not allowed, causing a compile error: https://godbolt.org/z/dG943nvT7

thejohnfreeman (approved):

godexsoft (approved):
LGTM 👍

PR #4592, opened 2023-06-28
[Trivial] Fix build references to deleted ServerHandlerImp

ximinez:

* Fixes the build by changing references to `ServerHandlerImp` to the more correct `ServerHandler`.

### Context of Change

* Commits 0b812cd (#4427) and 11e914f (#4516) conflict. The first added references to `ServerHandlerImp` in files outside of that class's organizational unit (which is technically incorrect). The second removed `ServerHandlerImp`, but was not up to date with develop. This results in the build failing in all circumstances.

### Type of Change

## Before / After

Before: `rippled` doesn't build.
After: It does.

## Future Tasks

Dedicating resources to wiping out all of the levelization issues across the board would help with this kind of thing.

### DIFF
--- a/src/ripple/app/misc/NetworkOPs.cpp
+++ b/src/ripple/app/misc/NetworkOPs.cpp
@@ -64,8 +64,8 @@
 #include <ripple/resource/ResourceManager.h>
 #include <ripple/rpc/BookChanges.h>
 #include <ripple/rpc/DeliveredAmount.h>
+#include <ripple/rpc/ServerHandler.h>
 #include <ripple/rpc/impl/RPCHelpers.h>
-#include <ripple/rpc/impl/ServerHandlerImp.h>
 #include <boost/asio/ip/host_name.hpp>
 #include <boost/asio/steady_timer.hpp>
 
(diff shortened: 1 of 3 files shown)


src/ripple/app/misc/NetworkOPs.cpp
```diff
@@ -64,8 +64,8 @@
 #include <ripple/resource/ResourceManager.h>
 #include <ripple/rpc/BookChanges.h>
 #include <ripple/rpc/DeliveredAmount.h>
+#include <ripple/rpc/ServerHandler.h>
 #include <ripple/rpc/impl/RPCHelpers.h>
```
ckeshava, line 68: If we strictly follow the lexicographic order, shouldn't the import order be reversed?
alphabetically, `S` comes after `i`. While ordering the imports, are we keeping all the `impl` imports together?

ckeshava (approved):

src/ripple/app/misc/NetworkOPs.cpp
scottschurr: It depends on whether the lexicographic order is case insensitive or not.  Lower case "i" comes after capital "S".

ckeshava: okay, didn't think of that :)

scottschurr (commented):
A bit late, but :+1: from me.

PR #4596, opened 2023-06-28
Reenable Windows CI build with Artifactory support

ximinez:

#4556 added Artifactory support to the `nix` builds. This extends that support to the Windows build

The good news is that the build now works. The bad news is that it introduces a new bottleneck, since the build and test takes over an hour.

### Context of Change

Our Windows CI build has been disabled for a while because it was unable to successfully build Conan dependencies. Now that it can download pre-built dependencies from Artificatory, we can put it back in play.

Right now, the job is running on Github's limited hardware. I hope that we can get self-hosted Windows runners available soon. If they become available before this PR is merged, I'll push a trivial commit changing the `runs-on:` line.

## Before / After

Before: No Windows CI jobs enabled.
After: CI will build and test a Windows release build.

## Test Plan

No C++ code was changed. This only affects CI, so there is nothing to test.

## Future Tasks

Making this job run on heavy Windows runners, so it doesn't take so dang long.


thejohnfreeman (approved):

.github/workflows/windows.yml
```diff
@@ -55,45 +53,43 @@ jobs:
...
+        with:
+          generator: '${{ matrix.generator }}'
+          configuration: ${{ matrix.configuration }}
+          # Hard code for now. Move to the matrix if varied options are needed
+          cmake-args: '-Dassert=ON -Dreporting=OFF -Dunity=OFF'
       - name: test
+        shell: bash
         run: |
-          & "$env:build_dir\${{ matrix.configuration }}\rippled.exe" --unittest
+          ${build_dir}/${{ matrix.configuration }}/rippled --unittest --unittest-jobs $(nproc)
```
ckeshava, line 95: the last time I checked, parallel unit tests did not always give the correct results. Some of the tests failed, we were not able to figure out the cause.

.github/actions/dependencies/action.yml
```diff
@@ -14,7 +14,7 @@ runs:
     - name: install dependencies
       shell: bash
       run: |
-        mkdir ${build_dir}
+        mkdir -p ${build_dir}
```
ckeshava, line 17: just so that I understand correctly: the `-p` parameter affects only paths which have a depth of at least level i.e. `a/b/...` where there is at least one parent folder.

.github/workflows/windows.yml
thejohnfreeman: Parallel tests fail on macOS only.

.github/actions/dependencies/action.yml
thejohnfreeman: `-p` will create every missing directory on the given path, not just the last name in the path.

ckeshava: yeah, that's what I meant. I forgot to type "at least level 1" :P 👍

.github/workflows/windows.yml
ckeshava: i didn't know that, okay

ckeshava (approved):

legleux (approved):

.github/workflows/windows.yml
ximinez: To add on: In my testing so far, parallel unit tests have worked fine on every run that ... got that far, at least with release.

.github/actions/dependencies/action.yml
ximinez: Also, it doesn't error out if the directory already exists. Works around two potential problems with one parameter.

ckeshava: okay 👍

.github/workflows/windows.yml
ckeshava: i see, ok

.github/workflows/windows.yml
```diff
@@ -63,7 +63,8 @@ jobs:
           conan profile update settings.compiler.runtime=MT${{ matrix.configuration == 'Debug' && 'd' || '' }} default
           # Do not quote the URL. An empty string will be accepted (with
           # a non-fatal warning), but a missing argument will not.
-          conan remote add ripple ${{ env.CONAN_URL }} --insert 0
+          conan remote add ripple ${{ env.CONAN_URL }} --insert 0 || \
+            conan remote list
```
thejohnfreeman, line 67: What's going on here?

ximinez: The hosted runner seems to be keeping some state between runs, which cause errors when the commands are run again. This was a workaround. Here's an example from my own repo: https://github.com/ximinez/rippled/actions/runs/5490473061/jobs/10005957758#step:8:14

thejohnfreeman:
Judging by the [tests](https://github.com/ximinez/rippled/actions/runs/5554837580/jobs/10145160685), the runner has only 4 hardware threads available. We might be able to cut the build time if we can get that number up to 12 or 16.

.github/workflows/windows.yml
thejohnfreeman: Is it because the runner does not create a fresh environment for each job? I think we should consider this a blocker.

ximinez: That's what it looks like.

ximinez:
We've recently had a few merges to `develop` which had problems with the Windows build, which I think CI would have caught. I'd like to bring attention back to this PR to get it merged.

Unfortunately, the Github Actions Windows build is not working now with self-hosted runner support, so I've rolled back those commits so that we have something that works. (Well, mostly. Some unit tests failed in the last run - I'll fix that.)

I think the trade-off of extra time (a lot of extra time) for more reliability will be worth it. We won't have to wait for every single Windows build to finish - as long as it's checked at some point before merging, I'll be happy. (And we have pretty reliable job cancellation, so they shouldn't build up a huge backlog while changes are being made.)

intelliot:
There have been (only) 2 recent problems on Windows, AFAIK (but please note if there have been others):
- https://github.com/XRPLF/rippled/pull/4702
- https://github.com/XRPLF/rippled/pull/4708

I agree it's still worth having GitHub Actions do a Windows build, as long as it can be done cost-effectively and reliably. And, how long does it take?

ckeshava:
@intelliot in the description of the PR, Ed says that more than an hour is required for the completion of the Windows tests.

On a tangential note, the windows tests seem to use pre-built artifactory dependencies. In a recent discussion, we are planning to move away from internal artifactory builds and instead rely on the conan center for dependencies.

we will need to rethink our discussions in light of this windows CI runner.

thejohnfreeman:
> On a tangential note, the windows tests seem to use pre-built artifactory dependencies. In a recent discussion, we are planning to move away from internal artifactory builds and instead rely on the conan center for dependencies.

I think this was suggested, but no such plan was made. I want to reiterate that this would be a regression. Conan Center is what we _used_ to use for dependencies, and that causes very long build times because Conan Center _cannot_ hold the binaries we need. They do not build the configurations we need, and they will not add them.

.github/actions/dependencies/action.yml
```diff
@@ -17,7 +17,7 @@ runs:
...
```
thejohnfreeman, line 20: Was this actually necessary? I don't think the build directory has multiple steps, so I'm guessing this was added to get around an existing build directory. I suspect we should instead remove and remake the build directory if that was the reason.

.github/workflows/windows.yml
```diff
@@ -55,45 +53,44 @@ jobs:
...
-          pwd
-          ls
-          cmake `
-            -G "${{ matrix.generator }}" `
-            -DCMAKE_TOOLCHAIN_FILE:FILEPATH=build/generators/conan_toolchain.cmake `
-            -Dassert=ON `
-            -Dreporting=OFF `
-            -Dunity=OFF `
-            ..
+          echo missing=$(conan info . --build missing --settings build_type=${{ matrix.configuration }} --json 2>/dev/null  | grep '^\[') | tee ${GITHUB_OUTPUT}
```
thejohnfreeman, line 76: Can we drive-by add the `--settings build_type=${{ matrix.configuration }}` fix in the `nix` workflow too?

.github/workflows/windows.yml
```diff
@@ -55,45 +53,44 @@ jobs:
...
+        with:
+          generator: '${{ matrix.generator }}'
+          configuration: ${{ matrix.configuration }}
+          # Hard code for now. Move to the matrix if varied options are needed
+          cmake-args: '-Dassert=ON -Dreporting=OFF -Dunity=OFF'
       - name: test
+        shell: bash
         run: |
-          & "$env:build_dir\${{ matrix.configuration }}\rippled.exe" --unittest
+          # ${build_dir}/${{ matrix.configuration }}/rippled --unittest --unittest-jobs $(nproc)
```
thejohnfreeman, line 95: Is `nproc` available on Windows? @ximinez please confirm, but I think parallel testing works on Windows. I believe we just have to use `$NUMBER_OF_PROCESSORS`. If neither works, though, we should leave a comment here.

thejohnfreeman (changes requested):

thejohnfreeman:
Latest error message is bizarre:

```
'grpc-proto/cci.20220627' requires 'protobuf/3.21.4' while 'grpc-proto/cci.20220627' requires 'protobuf/3.21.9'.
```

Basically says "A requires B while A requires C". My first instinct is to blow away any Windows binaries cached in our Artifactory, and to ensure that the Conan cache on the Windows runner starts empty (`rm -rf ~/.conan/data`).

ckeshava:
@thejohnfreeman I believe Ed refreshed the binaries on the artifactory a few times (he built the dependency binaries and uploaded them to the internal artifactories)

.github/actions/dependencies/action.yml
ximinez: I _think_ this was to work around an existing directory. IIRC, the workflow has a "build dependencies" step, which calls this script directly *and* a "build" step, which calls this workflow indirectly. It must have been making the second step fail.

I'll go ahead and remove the change, and see how that affects things.

ckeshava: @ximinez are you working with your local windows machine or tinkering with the Windows runner provided by Github Actions?

I don't have a windows machine and I'm wondering how I can try out these build steps.

.github/workflows/windows.yml
ximinez: `nproc` is available. I commented out this line to see if the tests would pass on a single threaded run on the Github runners. [They didn't.](https://github.com/ximinez/rippled/actions/runs/6340590156/job/17222329753#step:14:1372)

Instead what I'm going to do is put the parallel run back, but mark the test step as allowed to fail (`continue-on-error`). At this point, it's more important to me that the build succeeds than that the tests succeed.
1. We've got plenty of test coverage on the other jobs.
2. Test failures are much rarer than build failures because of cross-platform issues.
3. Having a test failure locally doesn't interrupt *my* workflow nearly as much as a build failure.

ximinez:
> @thejohnfreeman I believe Ed refreshed the binaries on the artifactory a few times (he built the dependency binaries and uploaded them to the internal artifactories)

Yeah, [the build worked last time the job ran](https://github.com/ximinez/rippled/actions/runs/6340590156/job/17222329753). I suspect those failures were just something getting caught in an inconsistent state while I was messing with the dependencies.

.github/actions/dependencies/action.yml
ximinez: Yep, that was it. https://github.com/ximinez/rippled/actions/runs/6343277082/job/17230820510#step:13:61

The reason it's not an issue on the `nix` jobs is that those two steps are split across the separate jobs. My problem is a side effect of bringing them back together. I can think of three ways to make this work.
1. Put the `-p` back with a comment explaining that it's for this scenario.
2. `rm -rf ${build_dir}` before creating it to ensure we're starting with a fresh slate.
3. `[ -d ${build_dir} ] || mkdir ${build_dir}` so it doesn't error out if it exists.

I don't really have a strong feeling one way or another. 2 is probably the safest option, even though it'll add a little time to the build because it'll need to recreate the conan toolchain file. What do you think?

ximinez: > @ximinez are you working with your local windows machine or tinkering with the Windows runner provided by Github Actions?

This is with the Windows runner from Github Actions.

> I don't have a windows machine and I'm wondering how I can try out these build steps.

You could copy my branch into a branch in your own repo and play with it from there.

ckeshava: okay, I'll check it out 👍

thejohnfreeman: I think the right fix is to remove the link between the `build` and `dependencies` actions, and just add a step for the `dependencies` action to the `test` job in the `nix` workflow.

Alternatively, but not my preference, you could remove the `dependencies` step in the `windows` workflow and move the upload step after the `build` step, adding an `always() &&` clause to its `if` condition.

ximinez:
@thejohnfreeman @ckeshava I think this PR is as good as it's going to get _given current infrastructure_.

It's an improvement (given that we have no Windows CI at all right now), and doesn't appear to do any harm. It still takes a long time to build, but cancels old jobs, so it won't tie things up indefinitely. The only place where it might cause a problem is if we need to rush something into develop, and need to wait for this build. There are ways around that, so I don't see it as a blocker.

Mind reviewing changes and signing off so we can get this merged? Thanks!

.github/workflows/windows.yml
```diff
@@ -55,45 +53,48 @@ jobs:
...
+          # a non-fatal warning), but a missing argument will not.
+          conan remote add ripple ${{ env.CONAN_URL }} --insert 0
+      - name: try to authenticate to ripple Conan remote
+        shell: bash
+        id: remote
         run: |
-          conan export external/snappy snappy/1.1.9@
-          conan export external/soci soci/4.0.3@
-      - name: install dependencies
+          echo outcome=$(conan user --remote ripple ${{ secrets.CONAN_USERNAME }} \
```
ckeshava, line 69: thanks. this change obviates the need for this [PR](https://github.com/XRPLF/rippled/pull/4716) right?

ckeshava: oh wait, I think this change affects only the windows workflow, never mind

ckeshava (approved):

intelliot:
note: awaiting review from @thejohnfreeman to confirm whether prior requests were addressed

thejohnfreeman (approved):

thejohnfreeman:
Thank you for all the work here @ximinez. I know this was a heavy lift even though the changeset is small.

intelliot:
note: intend to merge this shortly. giving CI a chance to run for a bit

intelliot:
suggested commit message:
```
ci: reenable Windows CI build with Artifactory support (#4596)

Artifactory support was added to the `nix` builds with #4556. This
extends that support to the Windows build. Now, the Windows build works;
CI will build and test a Windows release build. This only affects CI and
does not change any C++ code.

* Copy the remote setup step outcome fix from #4716 discussion
* Allow the Windows job to succeed if tests fail:
  * Currently the tests do not always pass, even on a single threaded
    run on the GitHub runners. So we are using parallel runs and mark
    the test step as allowed to fail (continue-on-error).
  * At this point, it's more important that the build succeeds than that
    the tests succeed, because:
  * We've got plenty of test coverage on the other jobs.
  * Test failures are much rarer than build failures because of
    cross-platform issues.
  * Having a test failure locally doesn't interrupt a workflow nearly as
    much as a build failure.

Note that Conan Center cannot hold the binaries we need. They do not
build the configurations we need, and they will not add them.

## Future Tasks

This introduces a new bottleneck since the build and test takes over an
hour. Speed up the job by:

* Making this job run on heavy Windows runners.
* Increasing the number of hardware threads.
```

PR #4597, opened 2023-06-29
Cancel overridden workflows

thejohnfreeman:
Small quality-of-life improvements to workflows using new [concurrency control features](https://docs.github.com/en/actions/using-jobs/using-concurrency).

### DIFF
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -1,5 +1,8 @@
 name: nix
 on: [push, pull_request]
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
 
 # This workflow has two job matrixes.
 # They can be considered phases because the second matrix ("test")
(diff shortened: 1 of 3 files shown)


intelliot:
LGTM

ximinez (approved):
:+1:

.github/workflows/nix.yml
```diff
@@ -57,7 +60,7 @@ jobs:
           echo ${PATH} | tr ':' '\n'
           conan --version
           cmake --version
-          env
+          env | sort
       - name: configure Conan
         env:
           CONAN_URL: http://18.143.149.228:8081/artifactory/api/conan/conan-non-prod
```
ckeshava, line 66: if the dns mapping changes in the future, won't we be in trouble because we have hard-coded the IP address?

ckeshava (approved):

.github/workflows/nix.yml
thejohnfreeman: There is no domain name for this address. No mapping to change.

The risk is that the public IP address for this machine changes. We have already discussed, internally, adding a domain name for this machine. It will be an improvement, but it is not necessary.

Lastly, the IP address is already hard-coded in the workflows. It is not part of this PR.

ckeshava: I see, ok

PR #4602, opened 2023-06-30
amendment: assign the serial number to the amendment list object

a-noni-mousse:
Tracking the serial number allows for check if amendments have changed since last check without checking of the list one by one

### Type of Change

- [X] amendment
- 
## Before / After
When amendment activates the new field **Sequence** in the amendments object will be increased 1. Before the field does not exist.

### DIFF
--- a/src/ripple/app/tx/impl/Change.cpp
+++ b/src/ripple/app/tx/impl/Change.cpp
@@ -329,6 +329,15 @@ Change::applyAmendment()
                              << " activated: server blocked.";
             ctx_.app.getOPs().setAmendmentBlocked();
         }
+
+        if (view().rules().enabled(featureAmendmentTableSequence) ||
+            amendment == featureAmendmentTableSequence)
+        {
+            // if amendment is enabled then we do an increase on sequence
+            // by 1
+            amendmentObject->setFieldU32(
+                sfSequence, amendmentObject->getFieldU32(sfSequence) + 1);
+        }
     }
 
     if (newMajorities.empty())
--- a/src/ripple/protocol/impl/LedgerFormats.cpp
+++ b/src/ripple/protocol/impl/LedgerFormats.cpp
@@ -141,6 +141,7 @@ LedgerFormats::LedgerFormats()
         {
             {sfAmendments,           soeOPTIONAL},  // Enabled
             {sfMajorities,           soeOPTIONAL},
+            {sfSequence,             soeOPTIONAL},
         },
         commonFields);
 
--- a/src/ripple/ledger/impl/ReadView.cpp
+++ b/src/ripple/ledger/impl/ReadView.cpp
@@ -76,14 +76,26 @@ makeRulesGivenLedger(
     DigestAwareReadView const& ledger,
     std::unordered_set<uint256, beast::uhash<>> const& presets)
 {
-    Keylet const k = keylet::amendments();
-    std::optional digest = ledger.digest(k.key);
-    if (digest)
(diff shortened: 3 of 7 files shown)


src/ripple/app/tx/impl/Change.cpp
```diff
@@ -329,6 +329,15 @@ Change::applyAmendment()
                              << " activated: server blocked.";
             ctx_.app.getOPs().setAmendmentBlocked();
         }
+
+        if (view().rules().enabled(featureAmendmentTableSequence) ||
```
ckniffen, line 333: It might be nice to increment the serial number when amendments gain or loose majority too since those change the amendment ledger entry.

intelliot:
Time and resources are limited, but in general, yes - RippleX engineers make an effort to review contributions from external developers in the community. We have other priorities, so it may take some time for us to get to your PR, but we will.

Can you share the motivations for this PR? How is this change beneficial? How do you weigh that against the downsides (more code/complexity, more memory/compute use, etc)?

ximinez:
> The serial number allows for easy check of if new amendments have activated and with it server does not need to load and process the amendment object that has not been changed.

At least for now, once an amendment is enabled - that is, added to the `Amendments` list - there is no mechanism for it to be removed[^1]. That means that the _size_ of the `Amendments` list will change whenever a new amendment is enabled (strictly increasing currently). Would the size of the list not be a perfectly good substitute or proxy for a serial number?

[^1]: I would be remiss if I did not mention that there have been discussions and efforts to remove amendments from the list after they've been active for several years, and removed / retired from the code. So far, none of those designs have come to fruition. I think it's safe to say that even if such a mechanism were to be introduced, any implementation would remove several old amendments with one new amendment. (It would be absurd to add an entry that only removes one other entry if the goal is to save space.) Thus, if that were to occur, the size of the list would not strictly increase, but it would _change_ any time an amendment was enabled.

src/ripple/protocol/impl/LedgerFormats.cpp
```diff
@@ -141,6 +141,7 @@ LedgerFormats::LedgerFormats()
         {
             {sfAmendments,           soeOPTIONAL},  // Enabled
             {sfMajorities,           soeOPTIONAL},
+            {sfSequence,             soeOPTIONAL},
```
mvadari, line 144: Why have a separate `Sequence` field instead of just using the existing logic for `sfPreviousTxnID`/`sfPreviousTxnLgrSeq`?

ckniffen: I would prefer those fields too.  It would allow walking the activation change txs.

intelliot:
Hi @a-noni-mousse - do you have any thoughts on the comments above?

It should be straightforward to compare the amendment object itself to see if there have been changes; and if the format of the data is a problem, you could compare a hash of the object in order to detect changes.

intelliot:
If there's no reply for a while, this PR will be closed due to inactivity. You'll always be welcome to open a new PR (or reopen this one) if you still feel the change is worth making.

src/ripple/protocol/impl/LedgerFormats.cpp
mvadari: https://github.com/XRPLF/rippled/pull/4751 - IMO this is a better solution since it aligns with the existing ledger structure

intelliot:
Closing due to inactivity. Suggested alternatives are detailed above, e.g. compare the amendment object itself, or a hash of that object. However, if that is insufficient for your use case, please feel free to re-open, or open a new PR.

PR #4603, opened 2023-06-30
Add binary hardening compile and link flags

thejohnfreeman:
Add full RELRO and stack canary when compiling with GCC or Clang.

### DIFF
--- a/Builds/CMake/RippledCompiler.cmake
+++ b/Builds/CMake/RippledCompiler.cmake
@@ -120,7 +120,7 @@ else ()
   target_link_libraries (common
     INTERFACE
       -rdynamic
-      $<$<NOT:$<BOOL:${APPLE}>>:-Wl,-z,relro,-z,now>
+      $<$<BOOL:${is_linux}>:-Wl,-z,relro,-z,now>
       # link to static libc/c++ iff:
       #   * static option set and
       #   * NOT APPLE (AppleClang does not support static libc/c++) and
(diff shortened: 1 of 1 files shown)


nbougalis:
Awesome! Thanks for adding this, @thejohnfreeman! I wanted to add this for a really long time.

nbougalis (approved):

seelabs (approved):

Builds/CMake/RippledCompiler.cmake
```diff
@@ -13,7 +13,6 @@ link_libraries (Ripple::common)
 set_target_properties (common
   PROPERTIES INTERFACE_POSITION_INDEPENDENT_CODE ON)
 set(CMAKE_CXX_EXTENSIONS OFF)
-target_compile_features (common INTERFACE cxx_std_17)
```
intelliot, line 16: Why is this line removed?

intelliot:
Suggested commit message:
```
build: add binary hardening compile and link flags (#4603)

Enhance security during the build process:

* The '-fstack-protector' flag enables stack protection for preventing
  buffer overflow vulnerabilities. If an attempt is made to overflow the
  buffer, the program will terminate, thus protecting the integrity of
  the stack.
* The '-Wl,-z,relro,-z,now' linker flag enables Read-only Relocations
  (RELRO), a feature that helps harden the binary against certain types
  of exploits, particularly those that involve overwriting the Global
  Offset Table (GOT).
  * This flag is only set for Linux builds, due to compatibility issues
    with apple-clang.
  * The `relro` option makes certain sections of memory read-only after
    initialization to prevent them from being overwritten, while `now`
    ensures that all dynamic symbols are resolved immediately on program
    start, reducing the window of opportunity for attacks.
```

Builds/CMake/RippledCompiler.cmake
thejohnfreeman: Our code should be compiled with at least C++20, and the choice of standard is made by the toolchain file, not our CMake project files.

intelliot:
@thejohnfreeman can you bring this up-to-date with develop?

I don't have permission to apply a merge commit to your branch.

Issue #4601, opened 2023-06-30
Commit 1c2ae10dc057d5c7c76d0df2c5a92fc753e4769a causes breakage


nbougalis:
Commit 1c2ae10dc057d5c7c76d0df2c5a92fc753e4769a causes breakage: the way it checks for certain flags breaks, resulting in errors. It should be reverted.

intelliot:
Can you share a demonstration of the breakage (and errors)?

nbougalis:
Consider the function `conditionMet`, specifically lines 121 and 122 and 146-147 of Handler.h: 

https://github.com/XRPLF/rippled/commit/1c2ae10dc057d5c7c76d0df2c5a92fc753e4769a#diff-41d8d23b15cb894b075204dcd88d7bf9ca223fa77f43e28bb22cf38b805cabfcL121-L122

https://github.com/XRPLF/rippled/commit/1c2ae10dc057d5c7c76d0df2c5a92fc753e4769a#diff-41d8d23b15cb894b075204dcd88d7bf9ca223fa77f43e28bb22cf38b805cabfcL121-L122

The change is broken: where it previously, it checked for a _specific_ condition (e.g. `NEEDS_CLOSED_LEDGER`) and only entered the `if` block if it was met, it now executes that code if _any_ condition is set.

A prime example of a RPC request that fails is the `tx` command, which has `NEEDS_NETWORK_CONNECTION` and will now not work unless the server has _both_ a current and a closed ledger.

The problem isn't the powers of two change. It's how the checks where restructured.

intelliot:
Is it possible to write a unit test that detects the error?

ckeshava:
hello Nik, thank you very much for pointing this out. I'm working on this issue.

ckeshava:
I will revert the changes to the condition evaluation.

The existing unit tests do not catch this error. It appears that there are no unit tests for `https://github.com/XRPLF/rippled/blob/develop/src/ripple/rpc/impl/Handler.h`, but this code is indirectly being tested in `https://github.com/XRPLF/rippled/blob/develop/src/test/rpc/RPCCall_test.cpp`. 

Personally, I feel it is better to add a new unit test file for the `Handler.h` (because it would have a smaller granularity) instead of piling on additional tests in `RPCCall_test`. Do you guys think that is okay?

ximinez:
[...]

PR #4604, opened 2023-06-30
Revert "Change enum values to be powers of two (fix #3417)"

### LINKED ISSUE #3417: Software does not distinguish between different Conditions (Version: 1.5)

intelliot:
Consider the function `conditionMet`, specifically lines 121 and 122 and 146-147 of Handler.h: 

https://github.com/XRPLF/rippled/commit/1c2ae10dc057d5c7c76d0df2c5a92fc753e4769a#diff-41d8d23b15cb894b075204dcd88d7bf9ca223fa77f43e28bb22cf38b805cabfcL121-L122

https://github.com/XRPLF/rippled/commit/1c2ae10dc057d5c7c76d0df2c5a92fc753e4769a#diff-41d8d23b15cb894b075204dcd88d7bf9ca223fa77f43e28bb22cf38b805cabfcL146-L147

The change is broken: where it previously, it checked for a _specific_ condition (e.g. `NEEDS_CLOSED_LEDGER`) and only entered the `if` block if it was met, it now executes that code if _any_ condition is set.

A prime example of a RPC request that fails is the `tx` command, which has `NEEDS_NETWORK_CONNECTION` and will now not work unless the server has _both_ a current and a closed ledger.

The problem isn't the powers of two change. It's how the checks where restructured.

_Originally posted by @nbougalis in https://github.com/XRPLF/rippled/issues/4601#issuecomment-1615494006_
            

Reverts XRPLF/rippled#4239

### DIFF
--- a/src/ripple/rpc/impl/Handler.h
+++ b/src/ripple/rpc/impl/Handler.h
@@ -39,8 +39,8 @@ namespace RPC {
 enum Condition {
     NO_CONDITION = 0,
     NEEDS_NETWORK_CONNECTION = 1,
-    NEEDS_CURRENT_LEDGER = 1 << 1,
-    NEEDS_CLOSED_LEDGER = 1 << 2,
+    NEEDS_CURRENT_LEDGER = 2 + NEEDS_NETWORK_CONNECTION,
+    NEEDS_CLOSED_LEDGER = 4 + NEEDS_NETWORK_CONNECTION,
 };
 
 struct Handler
@@ -94,18 +94,20 @@ conditionMet(Condition condition_required, T& context)
     }
 
     if (context.app.getOPs().isAmendmentBlocked() &&
-        (condition_required != NO_CONDITION))
+        (condition_required & NEEDS_CURRENT_LEDGER ||
+         condition_required & NEEDS_CLOSED_LEDGER))
     {
         return rpcAMENDMENT_BLOCKED;
     }
 
     if (context.app.getOPs().isUNLBlocked() &&
-        (condition_required != NO_CONDITION))
+        (condition_required & NEEDS_CURRENT_LEDGER ||
+         condition_required & NEEDS_CLOSED_LEDGER))
     {
         return rpcEXPIRED_VALIDATOR_LIST;
     }
 
-    if ((condition_required != NO_CONDITION) &&
+    if ((condition_required & NEEDS_NETWORK_CONNECTION) &&
         (context.netOps.getOperatingMode() < OperatingMode::SYNCING))
     {
         JLOG(context.j.info()) << "Insufficient network mode for RPC: "
@@ -117,7 +119,7 @@ conditionMet(Condition condition_required, T& context)
     }
 
     if (!context.app.config().standalone() &&
-        condition_required != NO_CONDITION)
+        condition_required & NEEDS_CURRENT_LEDGER)
     {
         if (context.ledgerMaster.getValidatedLedgerAge() >
             Tuning::maxValidatedLedgerAge)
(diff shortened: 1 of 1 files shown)


ckeshava:
Yes, reverting the commit will fix the problem. But I feel we also need to add additional unit tests, so that such blindspots are covered for the future.

ckeshava (approved):

HowardHinnant (changes requested):
This revertible code isn't readable because it checks like a bit-mask even though it is not a bit-mask.

If the power-of-2 fix doesn't have the right behavior, it should be fixed with code that is more readable than this revert results in.  E.g. change the checks to use `==` instead of `&`.

That being said, @ximinez has an analysis that says the behavior between these two versions is the same, in which case I strongly prefer the power-of-2 version.

ximinez:
Veto. Copying my comment from https://github.com/XRPLF/rippled/issues/4601#issuecomment-1622147627:

> A prime example of a RPC request that fails is the `tx` command, which has `NEEDS_NETWORK_CONNECTION` and will now not work unless the server has _both_ a current and a closed ledger.

I don't think there's a behavior change at all. I commented on the original PR to that effect: https://github.com/XRPLF/rippled/pull/4239#discussion_r1224782464

Here's more detail.

In the old code,

```
NO_CONDITION == 0
NEEDS_NETWORK_CONNECTION == 1 == 0b0001
NEEDS_CURRENT_LEDGER     == 3 == 0b0011
NEEDS_CLOSED_LEDGER      == 5 == 0b0101
```

Consider the condition on line 110:

```
condition_required & NEEDS_NETWORK_CONNECTION
```
which is equivalent to 
```
condition_required & 0b0001
```

`condition_required` can be any of those 4 values.

| `condition_required` | `condition_required & NEEDS_NETWORK_CONNECTION` | boolean result |
|-|-|-|
| `NO_CONDITION == 0` | `0` | `false` |
| `NEEDS_NETWORK_CONNECTION == 0b0001` | `0b0001` | `true` |
| `NEEDS_CURRENT_LEDGER == 0b0011` | `0b0001` | `true` |
| `NEEDS_CLOSED_LEDGER == 0b0101` | `0b0001` | `true` |

Basically, for any value except `NO_CONDITION` (0), the expression evaluates to `true`.

The new values are 0, 1, 2, and 4, respectively, but the result is the same. `condition_required != NO_CONDITION` in the new code is functionally equivalent to `condition_required & NEEDS_NETWORK_CONNECTION` in the old code.

It should be clear that the results are the same for the other values.

intelliot:
Closing because @ckeshava and @ximinez agree that this should not be reverted; see https://github.com/XRPLF/rippled/issues/4601#issuecomment-1629688030

Issue #4541, opened 2023-05-26
[account_objects with limit value as zero do not return an error] (Version: [1.11.0-rc2])


mounikakun:
## Steps to Reproduce
`{
    "method": "account_objects",
    "params": [
        {
            "account": "r3kmBwpGRwRoZGM64zkkwoDFB7GNJqxD8Y",
            "ledger_index": "validated",
            "type": "state",
            "deletion_blockers_only": false,
            "limit": 0
        }
    ]
}`

## Expected Result

`invalidParams` error
## Actual Result

No error

mounikakun:
Issue exists in account_tx API also.

Issue #4540, opened 2023-05-26
[Rippled do not follow limit values as per documentation in many API's] (Version: [1.11.0-rc2])


mounikakun:
## Issue Description
As per documentation for [account_channels](https://xrpl.org/account_channels.html#account_channels), Limit cannot be less than 10 or more than 400 but rippled do not follow this boundary values.

## Steps to Reproduce
Provide limit value less than 10 or greater than 400.

## Expected Result
`invalidParams` error

## Actual Result
No error

mounikakun:
As per documentation for [account_objects](https://xrpl.org/account_objects.html#account_objects)limit must be within the inclusive range 10 to 400 but rippled do not follow this boundary values.

mounikakun:
Issue exists for account_offers also.

intelliot:
The API behavior is fine; this is something that should be clarified in the docs. Closing this issue in favor of https://github.com/XRPLF/xrpl-dev-portal/issues/1985

Issue #4315, opened 2022-09-28
For nft_buy_offers and nft_sell_offers with "limit":0 in the request returns empty offers list instead of "invalidParams" error.


mounikakun:
## Steps to Reproduce
`{
  "method": "nft_sell_offers",
  "params": [{
    "nft_id": "000000004C17E5B8E33245D98126F4A6DF4EA57FF71C027B0000099B00000000",
    "ledger_index": "validated",
    "limit": 0
  }]
}`

`{
  "method": "nft_buy_offers",
  "params": [{
    "nft_id": "00080000B99F877D1AA8CD9C4C0110DA670BCB86138D54CF16E5DA9C00000001",
    "ledger_index": "validated",
    "limit": 0
  }]
}`

## Expected Result
`{
    "result": {
        "error": "invalidParams",
        "error_code": 31,
        "error_message": "limitNotPositive",
        "status": "error",
        "type": "response",
        "request": {
            "method": "nft_buy_offers",
            "params": [
                {
                    "nft_id": "00080000B99F877D1AA8CD9C4C0110DA670BCB86138D54CF16E5DA9C00000001",
                    "ledger_index": "validated",
                    "limit": 0
                }
            ]
        }
    },
}`
## Actual Result
`{
    "result": {
        "limit": 0,
        "marker": "122DA77793E227FC22BC42CF4180094A032DB97C1D8CD721D9C47CF40D21BF10",
        "nft_id": "00080000B99F877D1AA8CD9C4C0110DA670BCB86138D54CF16E5DA9C00000001",
        "offers": [],
        "status": "success"
    }
}`

intelliot:
@shawnxie999 @ledhed2222 do you agree that this is an issue? If so, is it worth prioritizing?

ledhed2222:
what environment is this that you're testing this on? I don't see this behavior. What is happening (and what should happen) when you provide a limit of 0 is that it defaults to the default limit of 50. I can't recreate getting an empty array of offers.

PR #4610, opened 2023-07-06
Fix limit

PeterChen13579:
Fix limit error for #4540, #4541 and #4315 

### Context of Change
limit value should be inclusive between 10-400. Any range outside will be considered an error for api version 2 onwards. 
More discussion can be found [here](https://github.com/XRPLF/xrpl-dev-portal/issues/1985): 

### Type of Change

Added unit test that calls limit outside of valid range for api V2

### DIFF
--- a/src/ripple/rpc/impl/RPCHelpers.cpp
+++ b/src/ripple/rpc/impl/RPCHelpers.cpp
@@ -761,6 +761,10 @@ readLimitField(
             return RPC::expected_field_error(jss::limit, "unsigned integer");
 
         limit = jvLimit.asUInt();
+        if ((limit < 10 || limit > 400) && context.apiVersion > 1u)
+        {
+            return rpcError(rpcINVALID_PARAMS);
+        }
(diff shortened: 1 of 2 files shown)


PeterChen13579:
@ximinez I had a discussion with @godexsoft. He mentioned they were following the Rippled Documentation(limit) for Clio, and to change it on their side takes a lot more work than changing it on rippled's side. As rippled should match Clio, we both agreed to change it on rippled. Maybe @godexsoft can give more insight?

ximinez (commented):
This looks fine if we decide this is the correct solution, but I'm going to hold off approving, until a final decision is made.

godexsoft:
Looks good to me too @PeterChen13579 👍 But we did end up implementing the clamping behaviour in Clio instead. Will have to discard this PR.

PeterChen13579:
Discarded PR. Clamping supported on Clio.
Reference: [https://github.com/XRPLF/clio/pull/740](https://github.com/XRPLF/clio/pull/740)

PR #4611, opened 2023-07-09
test: add forAllApiVersions helper function

arihantkothari:

This PR introduces a new variadic template helper function `forAllApiVersions` that accepts callables to execute a set of functions over a range of versions, specifically from RPC::apiMinimumSupportedVersion to RPC::apiBetaVersion. This will be useful to avoid duplication of code. (the for loop)

### Context of Change
#4552

### Type of Change

### DIFF
--- a/src/test/jtx/Env.h
+++ b/src/test/jtx/Env.h
@@ -709,10 +709,10 @@ Env::rpc(std::string const& cmd, Args&&... args)
  * an unsigned integer as its argument and returns void.
  */
 template <class T>
-concept SingleVersionedTestCallable = requires(T callable, unsigned int v)
+concept SingleVersionedTestCallable = requires(T callable, unsigned int version)
 {
     {
-        callable(v)
+        callable(version)
     }
     ->std::same_as<void>;
 };
--- a/src/test/jtx/Env.h
+++ b/src/test/jtx/Env.h
@@ -711,7 +711,10 @@ Env::rpc(std::string const& cmd, Args&&... args)
 template <class T>
 concept SingleVersionedTestCallable = requires(T callable, unsigned int v)
 {
-    {callable(v)} -> std::same_as<void>;
+    {
+        callable(v)
+    }
+    ->std::same_as<void>;
 };
 
 /**
--- a/src/test/jtx/Env.h
+++ b/src/test/jtx/Env.h
@@ -705,21 +705,34 @@ Env::rpc(std::string const& cmd, Args&&... args)
 }
 
 /**
- * Executes a set of provided functions (Fs...) over a range of versions from
- * RPC::apiMinimumSupportedVersion to RPC::apiBetaVersion. Each function in
- * Fs... is expected to take a version number as its argument. This is useful
- * for running a series of tests or operations that need to be performed on
- * multiple versions of an API.
+ * The SingleVersionedTestCallable concept checks for a callable that takes
+ * an unsigned integer as its argument and returns void.
  */
-template <class... Fs>
+template <class T>
+concept SingleVersionedTestCallable = requires(T callable, unsigned int v) {
+    {
+        callable(v)
+    } -> std::same_as<void>;
+};
+
+/**
+ * The VersionedTestCallable concept checks if a set of callables all satisfy
+ * the SingleVersionedTestCallable concept. This allows forAllApiVersions to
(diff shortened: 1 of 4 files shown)


arihantkothari:
The `ordering.txt` file was changed because of the levelization patch.

src/test/jtx/Env.h
```diff
@@ -703,6 +704,25 @@ Env::rpc(std::string const& cmd, Args&&... args)
...
 }
 
+/**
+ * Executes a set of provided functions (Fs...) over a range of versions from
+ * RPC::apiMinimumSupportedVersion to RPC::apiBetaVersion. Each function in
+ * Fs... is expected to take a version number as its argument. This is useful
+ * for running a series of tests or operations that need to be performed on
+ * multiple versions of an API.
+ */
+template <class... Fs>
```
godexsoft, line 714: You don't need the type anywhere inside the body so you could just use `auto`: 
```cpp
void forAllApiVersions(auto... callables)
```

You could also constraint it to require to be callable with a version number using a concept:
```cpp
template<typename T>
concept VersionedTestCallable = requires(T a, uint32_t v) // use same type as real version for `v`
{
    { a(v) } -> std::same_as<void>;
};
// ...
void forAllApiVersions(VersionedTestCallable auto... fs)
```

Without any of this you still have an ok solution in my opinion.

src/test/rpc/AccountTx_test.cpp
```diff
@@ -667,12 +666,8 @@ class AccountTx_test : public beast::unit_test::suite
...
     run() override
     {
-        for (auto testVersion = RPC::apiMinimumSupportedVersion;
-             testVersion <= RPC::apiBetaVersion;
-             ++testVersion)
-        {
-            testParameters(testVersion);
-        }
+        test::jtx::forAllApiVersions(
+            [this](auto version) { testParameters(version); });
```
godexsoft, line 670: You could try to use `std::bind_front` from `<functional>` here as an experiment. See what you prefer in the end. Both work of course 👍

godexsoft (approved):
LGTM. Leaving a few things for you to try out but not required to change them.

PeterChen13579 (approved):

src/test/jtx/Env.h
arihantkothari: Hi @godexsoft , sorry for the late reply - I've never practically used concepts before so I was learning about it.

```
template <typename T>
concept SingleVersionedTestCallable = requires(T a, uint32_t v)
{
    { a(v) } -> std::same_as<void>;
};

template <typename... T>
concept VersionedTestCallable = (... && SingleVersionedTestCallable<T>);
```

It looks like I'll have to do something like this. Does this look good ?

src/test/rpc/AccountTx_test.cpp
arihantkothari: Thanks @godexsoft , I actually love this suggestion - makes the code so much cleaner.

thejohnfreeman (approved):
I think I prefer the loop, perhaps a range-based for-loop over a `boost::counting_range`, but this is fine.

src/test/jtx/Env.h
godexsoft: Yea, something like that. I'd think my version also works but i may be wrong - i did not check it. You can experiment on godbolt (compiler explorer) and see what works.

godexsoft (approved):

shawnxie999 (approved):

intelliot:
Suggested commit message:
```
test: add forAllApiVersions helper function (#4611)

Introduce a new variadic template helper function, `forAllApiVersions`,
that accepts callables to execute a set of functions over a range of
versions - from RPC::apiMinimumSupportedVersion to RPC::apiBetaVersion.
This avoids the duplication of code.

Context: #4552
```

intelliot:
@arihantkothari -
1. Confirm that the commit message suggested above looks good (or offer an alternative)
2. Confirm that this PR is ready to merge (when you confirm, someone will add the `Passed` label)

arihantkothari:
> @arihantkothari -
> 
> 1. Confirm that the commit message suggested above looks good (or offer an alternative)
> 2. Confirm that this PR is ready to merge (when you confirm, someone will add the `Passed` label)

1. This commit looks good to me.
2. I can confirm that it is ready to be merged.
Thanks :)

PR #4617, opened 2023-07-11
Rename `allowClawback` flag to `allowTrustLineClawback`

shawnxie999:
### Context of Change
Reason for this change is here https://github.com/XRPLF/XRPL-Standards/pull/119

We would want to be explicit that this flag is exclusively for trustline. For new token types(eg. CFT), they will not utilize this flag for clawback, instead, they will turn clawback on/off on the token-level, which is more versatile. 

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### DIFF
--- a/src/ripple/protocol/LedgerFormats.h
+++ b/src/ripple/protocol/LedgerFormats.h
@@ -244,7 +244,7 @@ enum LedgerSpecificFlags {
     lsfDisallowIncomingTrustline =
         0x20000000,               // True, reject new trustlines (only if no issued assets)
     lsfAMM [[maybe_unused]] = 0x40000000, // True, AMM account
-    lsfAllowClawback =
+    lsfAllowTrustLineClawback =
         0x80000000,               // True, enable clawback 
 
     // ltOFFER
--- a/src/ripple/protocol/LedgerFormats.h
+++ b/src/ripple/protocol/LedgerFormats.h
@@ -251,7 +251,7 @@ enum LedgerSpecificFlags {
         0x20000000,               // True, reject new trustlines (only if no issued assets)
     lsfAMM [[maybe_unused]] = 0x40000000, // True, AMM account
     lsfAllowTrustLineClawback =
-        0x80000000,               // True, enable clawback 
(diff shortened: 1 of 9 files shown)


src/ripple/protocol/LedgerFormats.h
```diff
@@ -250,8 +250,8 @@ enum LedgerSpecificFlags {
     lsfDisallowIncomingTrustline =
         0x20000000,               // True, reject new trustlines (only if no issued assets)
     lsfAMM [[maybe_unused]] = 0x40000000, // True, AMM account
-    lsfAllowClawback =
-        0x80000000,               // True, enable clawback
+    lsfAllowTrustLineClawback =
+        0x80000000,               // True, enable clawback
```
ledhed2222, line 254: FYI - errant space

ledhed2222 (approved):

kennyzlei (approved):
delete space commit since last review lgtm

intelliot:
@kennyzlei @shawnxie999 this is a breaking API change compared to what was in `develop` prior to merging this PR, right?

If applicable, please add the `API Change` label.

intelliot:
I think the `Breaking change (fix or feature that would cause existing functionality to not work as expected)` checkbox may also need to be checked

shawnxie999:
@intelliot why would that be? clawback hasn't been released yet, this change would affect no one

kennyzlei:
> I think the `Breaking change (fix or feature that would cause existing functionality to not work as expected)` checkbox may also need to be checked

@intelliot I don't believe this is a breaking change either.  The flag was renamed but the value of the flag is still the same (16).  If any previous API usage was using the previous name, it should continue to work.  Currently in devnet, this amendment is still not enabled, so there should not be any usage there

intelliot:
This is a breaking change to version 1.12.0-b1 which included the original Clawback implementation, and needs to be considered for docs and client libraries, which often only look at PRs labeled `API Change` when considering the changes to document and implement. For example, https://github.com/XRPLF/rippled/pull/4612 is only considering PRs with the `API Change` label.

Issue #4546, opened 2023-05-26
[[account_tx] with non existent account do not return an error.] (Version: [1.11.0-rc2])


mounikakun:
## Steps to Reproduce

`Request:
{ "method": "account_tx", "params": [ { "account": "rH9pnDCtgAbFjozbZQuJWvPM8QHweXgZhj" } ] }`
## Expected Result

`actNotFound`
## Actual Result

`{ "result": { "account": "rH9pnDCtgAbFjozbZQuJWvPM8QHweXgZhj", "ledger_index_max": 28952659, "ledger_index_min": 23333389, "limit": 0, "status": "success", "transactions": [], "validated": true } }`

mounikakun:
Issue exists in gateway_balances also.

yinyiqian1:
close. please refer to https://github.com/XRPLF/rippled/pull/4619#issuecomment-1724579487

PR #4619, opened 2023-07-12
Account no exist Error

PeterChen13579:
Fixed issue #4546 where accounts does not exist on ledger does not throw error

### Context of Change
Added error messages when user sends an incorrect request. (Check the above Github issue link for more info.) Everything is under API Version 2 as it is a breaking change.

### Type of Change

Added own unit test for API Version 2;

### DIFF
--- a/src/ripple/rpc/handlers/AccountTx.cpp
+++ b/src/ripple/rpc/handlers/AccountTx.cpp
@@ -29,6 +29,7 @@
 #include <ripple/ledger/ReadView.h>
 #include <ripple/net/RPCErr.h>
 #include <ripple/protocol/ErrorCodes.h>
+#include <ripple/protocol/SystemParameters.h>
 #include <ripple/protocol/UintTypes.h>
 #include <ripple/protocol/jss.h>
 #include <ripple/resource/Fees.h>
@@ -245,12 +246,17 @@ doAccountTxHelp(RPC::Context& context, AccountTxArgs const& args)
     if (context.apiVersion > 1u)
     {
         bool accountFound = false;
-        for (unsigned int idx = result.ledgerRange.min;
-             idx <= result.ledgerRange.max;
-             ++idx)
+        unsigned int const rangeEnd =
+            result.ledgerRange.max + AccountDeleteSeqDelta;
+        for (unsigned int idx = result.ledgerRange.min; idx < rangeEnd;
+             idx += AccountDeleteSeqDelta)
         {
+            if (idx > result.ledgerRange.max)
+                idx = result.ledgerRange.max;
             std::shared_ptr<ReadView const> ledger;
             auto const status = getLedger(ledger, idx, context);
+            if (status.codeString() == "ledgerNotFound")
+                continue;
             if (ledger->exists(keylet::account(args.account)))
             {
                 accountFound = true;
--- a/src/ripple/rpc/handlers/AccountTx.cpp
+++ b/src/ripple/rpc/handlers/AccountTx.cpp
@@ -247,9 +247,9 @@ doAccountTxHelp(RPC::Context& context, AccountTxArgs const& args)
     {
         bool accountFound = false;
         unsigned int const rangeEnd =
-            result.ledgerRange.max + AccountDeleteSeqDelta;
+            result.ledgerRange.max + accountDeleteSeqDelta;
         for (unsigned int idx = result.ledgerRange.min; idx < rangeEnd;
-             idx += AccountDeleteSeqDelta)
+             idx += accountDeleteSeqDelta)
         {
             if (idx > result.ledgerRange.max)
                 idx = result.ledgerRange.max;
--- a/src/ripple/rpc/handlers/AccountTx.cpp
+++ b/src/ripple/rpc/handlers/AccountTx.cpp
@@ -240,6 +240,31 @@ doAccountTxHelp(RPC::Context& context, AccountTxArgs const& args)
 
     result.ledgerRange = std::get<LedgerRange>(lgrRange);
 
+    // if the account does not exist in any of the ledger range,
+    // return an error
+    if (context.apiVersion > 1u)
+    {
+        bool accountFound = false;
+        for (unsigned int idx = result.ledgerRange.min;
+             idx <= result.ledgerRange.max;
+             ++idx)
+        {
+            std::shared_ptr<ReadView const> ledger;
+            auto const status = getLedger(ledger, idx, context);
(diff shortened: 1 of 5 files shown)


intelliot:
@PeterChen13579:
1. Please confirm that this PR is ready for review
2. Please suggest / request 2 code reviewers

PeterChen13579:
@intelliot This PR is ready for review. Anyone is fine

connorjchen (approved):

src/ripple/rpc/handlers/AccountTx.cpp
```diff
@@ -240,6 +240,31 @@ doAccountTxHelp(RPC::Context& context, AccountTxArgs const& args)
...
     result.ledgerRange = std::get<LedgerRange>(lgrRange);
 
+    // if the account does not exist in any of the ledger range,
+    // return an error
+    if (context.apiVersion > 1u)
+    {
+        bool accountFound = false;
+        for (unsigned int idx = result.ledgerRange.min;
+             idx <= result.ledgerRange.max;
+             ++idx)
```
shawnxie999, line 250: Is iterating through the range efficient here? Would there be a more efficient approach?

cc. @ximinez

ximinez: * If the account exists at the ledger range start, it'll be found right away. :heavy_check_mark: 
* If the account is created sometime inside the range, [it can't be deleted for at least 255 ledgers](https://github.com/XRPLF/rippled/blob/c197c4cdd97612b5c3b71b3919add5238fae93a6/src/ripple/app/tx/impl/DeleteAccount.cpp#L213-215). That means that you don't need to check every single ledger. You can check every 255. However, we want to be sure to check the last ledger in case the account is created there. We don't want to check outside the end of the range in case that ledger doesn't even exist yet. :x: 
  * Example: We're looking at ledgers 1-600 for an account. If the account is created in ledger 2, and doesn't send any transactions of its own, then to be deleted `(2 + 255 <= view.seq())` must be true. That happens at ledger 257. So if you iterate every 255 ledgers, you'll start by checking 1, and next check 256, so the account will be there. (I'm choosing to iterate every 255 because I don't think the account will exist in ledger 257 if it's deleted there, and we don't want to miss it).  Now consider if the account is created in ledger 512. You'll check 1, 256, and 511. The next check would be 766, which is out of our range. If the last validated ledger is 605, you won't find the ledger, and that would be bad, so we want to make sure to explicitly check ledger 600 before exiting the loop. The account can't be deleted until `512+255 == 767`, so it'll be there.

Here's what I'm thinking:
1. First refactor AccountDelete.cpp to move `seqDelta` to src/ripple/protocol/SystemParameters.h and call it something like `AccountDeleteSeqDelta`.
2. Change the loop as such:
```
unsigned int const rangeEnd = result.ledgerRange.max + AccountDeleteSeqDelta;
for (unsigned int idx = result.ledgerRange.min;
    idx < rangeEnd;
   idx += AccountDeleteSeqDelta)
{
  if (idx > result.ledgerRange.max)
   idx = result.ledgerRange.max;
 [...]
```
I think I got that right.

src/ripple/rpc/handlers/AccountTx.cpp
```diff
@@ -240,6 +240,31 @@ doAccountTxHelp(RPC::Context& context, AccountTxArgs const& args)
...
+        {
+            std::shared_ptr<ReadView const> ledger;
+            auto const status = getLedger(ledger, idx, context);
```
ximinez, line 253: You're not checking `status` or `ledger`. That's probably not a good idea. What happens if the ledger is not found? I'm not super familiar with the surrounding code, so I don't know if ledger gaps are possible by the time you get here, but if they are, you need to check.

src/ripple/rpc/handlers/AccountTx.cpp
shawnxie999: @ximinez if the user wants to query the max ledger range. Wouldnt it still be inefficient? Iterating at an increment of 255 will still need over 300k iterations for the max ledger range. 

I am curious if the trade off is really worth it to implement an acctNotFound error. IMO we should stick with the current behavior.

ximinez: @shawnxie999 You're talking about full history? That's a good point. Maybe this check should only run if the difference between `result.ledgerRange.min` and `max` is below some reasonable limit (255*10?).

I am not so sure if this trade off of time and code complexity is worth it either.

PeterChen13579: @ximinez @shawnxie999  Thanks for the convo! I never knew the ledger range was over 300k iterations....
If that is the case, I agree that the trade off of time and code complexity is definitely not worth the change, and perhaps the docs should be updated instead.

PeterChen13579:
@ximinez I incorporated the changes you suggested, please have a look. I think the conclusion is to still implement these changes as it is a discrepancy with clio; and the fact that if an account does not exist on the ledger, than imo, it should return accountNotFound error message. Thanks 😄

shawnxie999:
> I think the conclusion is to still implement these changes as it is a discrepancy with clio; 

I think it might be better to have clio reflect rippled's behaviour instead, not the other way around. Introducing this kind of looping might be unnecessarily inefficient in rippled, and we might eventually be looking at millions of iterations as the ledger grows.

PeterChen13579:
@seelabs @thejohnfreeman I was wondering if you both might have any opinions on this matter? Your insights would be greatly appreciated! 😄 
Question: is there an efficient way to loop through the ledger history to check if an account exists on the ledger. The issue #4546 mentioned above on this PR is a discrepancy between clio and rippled. 
Thanks!

thejohnfreeman:
I believe there is no efficient means to search, in rippled's history representation, for an account that may not have ever existed. Indexing history for search is what clio is for.

ximinez:
I agree with @thejohnfreeman . The whole point of clio was to enhance the rippled API, not uncover edge cases that then need to be backported to rippled. If there's truly a bug, that's one thing, but different error handling behavior is perfectly acceptable.

PeterChen13579:
close with fix in clio.  https://github.com/XRPLF/clio/issues/651

intelliot:
- decided not to "fix" this since it potentially makes the API computationally-expensive in `rippled`
- Personally, I think it's fine for Clio to offer "additional" functionality beyond what rippled provides (if there is user demand and it makes sense given Clio's data model), but it should be offered on new/additional APIs - not by changing the behavior of existing ones.

PR #4577, opened 2023-06-20
Added Error messages on Gateway_Balance and Channel_Authorize

PeterChen13579:
## High Level Overview of Change
Fixed/added error messages for Gateway_balance(issue #4290 | #4548 ) and channel_authorize(#4289 )

### Context of Change
Added error messages when user sends an incorrect request. (Check the above Github issue link for more info.) Everything is under API Version 2 as it is a breaking change.

### Type of Change

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [X ] Tests (You added tests for code that already exists, or your new feature included in this PR)

Added own unit test for API Version 2;

PR #4620, opened 2023-07-12
APIv2(account_tx, noripple_check): return error for invalid input (fix #4543)

### LINKED ISSUE #4543: [Rippled do not return error for malformed inputs to optional parameters] (Version: [1.11.0-rc2])
## Issue Description

In most of the APIs, rippled do not validate inputs for optional parameters like binary, forward, strict etc.

## Expected Result

`invalidParams`error
## Actual Result

No error

PeterChen13579:
Fixed issue #4543 resolving malformed input on certain parameters

### Context of Change
Certain inputs for AccountTx and noRippleCheck should of error'ed out

### Type of Change

Added own unit tests for API version 2


arihantkothari (changes requested):
I have similar comments that I had in previous PRs (https://github.com/XRPLF/rippled/pull/4577#discussion_r1252382608). Better to keep it consistent.

PeterChen13579:
@ximinez if you mind reviewing this PR. Thanks 😃

src/ripple/rpc/handlers/AccountInfo.cpp
```diff
@@ -137,7 +137,8 @@ doAccountInfo(RPC::JsonContext& context)
         // The document states that signer_lists is a bool, however
         // assigning any string value works. Do not allow this.
         // This check is for api Version 2 onwards only
-        if (!params[jss::signer_lists].isBool() && context.apiVersion > 1)
+        if (params.isMember(jss::signer_lists) &&
```
ckeshava, line 140: You don't have to explicitly check if `jss::signer_lists` is a member of `params`, because otherwise a `NULL` member is returned as a default. (Source: https://github.com/XRPLF/rippled/blob/aded4a7a925a5822522e6bbbb4adfd3fec39257b/src/ripple/json/json_value.h#L341)

I'm not saying that what you did is wrong, I am making sure that the existing code is not incorrect.

Does it make a difference that you are using `1u` instead of `1`, given that the left-hand-side `context.apiVersion` is an `unsigned int`?

src/ripple/rpc/handlers/AccountTx.cpp
```diff
@@ -389,6 +389,20 @@ doAccountTxJson(RPC::JsonContext& context)
     AccountTxArgs args;
     Json::Value response;
 
+    // The document states that binary and forward params is a boolean value,
```
ckeshava, line 392: what do you mean by "document" here? Is it the XRPL documentation website? Can you include a link for the same?

src/test/rpc/NoRipple_test.cpp
```diff
@@ -222,42 +226,60 @@ class NoRipple_test : public beast::unit_test::suite
...
+                params[jss::account] = bob.human();
+                params[jss::role] = "gateway";
+                params[jss::transactions] = "asdf";
 
-            auto lines = env.rpc("json", "account_lines", to_string(params));
-            auto const& line0 = lines[jss::result][jss::lines][0u];
-            BEAST_EXPECT(line0[jss::no_ripple_peer].asBool() == false);
+                auto lines =
+                    env.rpc("json", "noripple_check", to_string(params));
+                BEAST_EXPECT(lines[jss::result][jss::error] == "invalidParams");
```
ckeshava, line 281: are we getting `invalidParams` due to the `jss::transactions` field? is that the cause of the problem?
I'm trying to understand how this testcase relates to the `noripple_check` property

ckeshava (changes requested):

src/ripple/rpc/handlers/AccountInfo.cpp
```diff
@@ -137,7 +137,8 @@ doAccountInfo(RPC::JsonContext& context)
...
+            !params[jss::signer_lists].isBool() && context.apiVersion > 1u)
```
arihantkothari, line 141: I think you should also check apiVersion before checking anything else to avoid unnecessary checks

src/ripple/rpc/handlers/NoRippleCheck.cpp
```diff
@@ -83,6 +83,15 @@ doNoRippleCheck(RPC::JsonContext& context)
     if (params.isMember(jss::transactions))
         transactions = params["transactions"].asBool();
 
+    // The document states that transactions params is a boolean value,
+    // however, assigning any string value works. Do not allow this.
+    // This check is for api Version 2 onwards only
+    if (params.isMember(jss::transactions) &&
+        !params[jss::transactions].isBool() && context.apiVersion > 1u)
```
arihantkothari, line 90: check apiVersion first to avoid unnecessary checks

src/ripple/rpc/handlers/AccountTx.cpp
```diff
@@ -389,6 +389,20 @@ doAccountTxJson(RPC::JsonContext& context)
...
+    // however, assigning any string value works. Do not allow this.
+    // This check is for api Version 2 onwards only
+    if (params.isMember(jss::binary) && !params[jss::binary].isBool() &&
+        context.apiVersion > 1u)
+    {
+        return rpcError(rpcINVALID_PARAMS);
+    }
+    if (params.isMember(jss::forward) && !params[jss::forward].isBool() &&
+        context.apiVersion > 1u)
```
arihantkothari, line 401: check apiVersion first to avoid unnecessary checks

src/ripple/rpc/handlers/AccountTx.cpp
```diff
@@ -389,6 +389,20 @@ doAccountTxJson(RPC::JsonContext& context)
     AccountTxArgs args;
     Json::Value response;
 
+    // The document states that binary and forward params is a boolean value,
+    // however, assigning any string value works. Do not allow this.
+    // This check is for api Version 2 onwards only
+    if (params.isMember(jss::binary) && !params[jss::binary].isBool() &&
+        context.apiVersion > 1u)
```
arihantkothari, line 396: check apiVersion first to avoid unnecessary checks

arihantkothari (changes requested):

src/ripple/rpc/handlers/AccountInfo.cpp
arihantkothari: Would "1" not do a type conversion from signed to unsigned ? I think "1u" avoids that extra step. No ?

src/test/rpc/NoRipple_test.cpp
```diff
@@ -222,42 +226,60 @@ class NoRipple_test : public beast::unit_test::suite
 
         env(trust(gw, USD(100), alice, 0));
         env(trust(gw, USD(100), bob, 0));
+        Json::Value params;
+        params[jss::api_version] = apiVersion;
 
+        if (params[jss::api_version] < 2)
```
ximinez, line 232: There's a bunch more overhead to look up `params[jss::api_version]` and compare it to `2` than there would be to just compare `apiVersion < 2u`.

But more importantly, this check makes no sense. It makes more sense to run every test, then compare the results based on version. Yes, that will sometimes lead to repetition, but what happens if version 1 is eventually deprecated? Do you want these tests to be deleted? Of course not. You also want to run that version-2-only test under version 1 to demonstrate that the `"asdf"` is accepted as a valid value.

ximinez (changes requested):

ximinez (approved):
Looks good.

ckeshava (approved):

intelliot:
@arihantkothari - could you check whether all of your comments have been addressed?

arihantkothari (approved):
Hey @intelliot , yes all my comments have been addressed. Thanks! 🚀

PR #4626, opened 2023-07-17
Clean up AMM owner dir on AMM account delete, disallow create check to AMM, and fix unconstrained AuthAccount.

gregtatcam:
Proposed commit message, including acknowledgement block, for when this PR is squashed and merged:

```
fix(AMM): prevent orphaned objects, inconsistent ledger state: (#4626)

When an AMM account is deleted, the owner directory entries must be
deleted in order to ensure consistent ledger state.

* When deleting AMM account:
  * Clean up AMM owner dir, linking AMM account and AMM object
  * Delete trust lines to AMM
* Disallow `CheckCreate` to AMM accounts
  * AMM cannot cash a check
* Constrain entries in AuthAccounts array to be accounts
  * AuthAccounts is an array of objects for the AMMBid transaction
* SetTrust (TrustSet): Allow on AMM only for LP tokens
  * If the destination is an AMM account and the trust line doesn't
    exist, then:
    * If the asset is not the AMM LP token, then fail the tx with
      `tecNO_PERMISSION`
    * If the AMM is in empty state, then fail the tx with `tecAMM_EMPTY`
      * This disallows trustlines to AMM in empty state
* Add AMMID to AMM root account
  * Remove lsfAMM flag and use sfAMMID instead
* Remove owner dir entry for ltAMM
* Add `AMMDelete` transaction type to handle amortized deletion
  * Limit number of trust lines to delete on final withdraw + AMMDelete
  * Put AMM in empty state when LPTokens is 0 upon final withdraw
  * Add `tfTwoAssetIfEmpty` deposit option in AMM empty state
  * Fail all AMM transactions in AMM empty state except special deposit
  * Add `tecINCOMPLETE` to indicate that not all AMM trust lines are
    deleted (i.e. partial deletion)
    * This is handled in Transactor similar to deleted offers
  * Fail AMMDelete with `tecINTERNAL` if AMM root account is nullptr
  * Don't validate for invalid asset pair in AMMDelete
* AMMWithdraw deletes AMM trust lines and AMM account/object only if the
  number of trust lines is less than max
  * Current `maxDeletableAMMTrustLines` = 512
  * Check no directory left after AMM trust lines are deleted
  * Enable partial trustline deletion in AMMWithdraw
* Add `tecAMM_NOT_EMPTY` to fail any transaction that expects an AMM in
  empty state
* Clawback considerations
  * Disallow clawback out of AMM account
  * Disallow AMM create if issuer can claw back

This patch applies to the AMM implementation in #4294.

Acknowledgements:
Richard Holland and Nik Bougalis for responsibly disclosing this issue.

We welcome reviews of the project code and urge researchers to

    bugs@xrpl.org
```

Fixes:

- Withdrawing all funds from AMM automatically deletes AMM account. The owner directory entries must be deleted. This fix addresses:
    - Clean up AMM owner dir linking AMM account and AMM object
    - Delete trustlines to AMM
- Disallow check create to AMM
- Fix unconstrained entries in AuthAccount in Bid transaction

### Context of Change

The bugs are introduced in the merged [AMM PR](https://github.com/XRPLF/rippled/pull/4294). They result in resource leak and inconsistent ledger state.

### Type of Change

## Before / After

Fixes the resource leak and unconstrained entries

## Test Plan

Added a unit-test to disallow check create and extended withdraw all unit-test.

### DIFF
--- a/src/ripple/app/misc/impl/AMMUtils.cpp
+++ b/src/ripple/app/misc/impl/AMMUtils.cpp
@@ -251,7 +251,7 @@ deleteAMMAccount(
 
     if (auto const ter =
             deleteAMMTrustLines(sb, ammAccountID, maxDeletableAMMTrustLines, j);
-        ter != tesSUCCESS || ter == tecINCOMPLETE)
+        ter != tesSUCCESS)
         return ter;
 
     auto const ownerDirKeylet = keylet::ownerDir(ammAccountID);
--- a/src/ripple/app/misc/impl/AMMUtils.cpp
+++ b/src/ripple/app/misc/impl/AMMUtils.cpp
@@ -253,6 +253,14 @@ deleteAMMAccount(
     if (!done)
         return tecINCOMPLETE;
 
+    auto const ownerDirKeylet = keylet::ownerDir(ammAccountID);
+    if (sb.exists(ownerDirKeylet) && !sb.emptyDirDelete(ownerDirKeylet))
+    {
+        JLOG(j.error()) << "deleteAMMAccount: cannot delete root dir node of "
+                        << toBase58(ammAccountID);
+        return tecINTERNAL;
+    }
+
     sb.erase(ammSle);
     if (sleAMMRoot)
         sb.erase(sleAMMRoot);
--- a/src/ripple/app/misc/impl/AMMUtils.cpp
+++ b/src/ripple/app/misc/impl/AMMUtils.cpp
@@ -281,7 +281,7 @@ initializeFeeAuctionVote(
     STObject voteEntry{sfVoteEntry};
     if (tfee != 0)
         voteEntry.setFieldU16(sfTradingFee, tfee);
-    voteEntry.setFieldU32(sfVoteWeight, 100000);
+    voteEntry.setFieldU32(sfVoteWeight, VOTE_WEIGHT_SCALE_FACTOR);
     voteEntry.setAccountID(sfAccount, account);
     voteSlots.push_back(voteEntry);
     ammSle->setFieldArray(sfVoteSlots, voteSlots);
--- a/src/ripple/app/misc/impl/AMMUtils.cpp
+++ b/src/ripple/app/misc/impl/AMMUtils.cpp
@@ -188,15 +188,15 @@ ammAccountHolds(
     return STAmount{issue};
 }
 
-static std::pair<TER, AllNodesDeleted>
+static std::pair<TER, bool>
 deleteAMMTrustLines(
-    ApplyView& view,
+    Sandbox& sb,
     AccountID const& ammAccountID,
     std::uint16_t maxTrustlinesToDelete,
     beast::Journal j)
 {
     return cleanupOnAccountDelete(
-        view,
+        sb,
         keylet::ownerDir(ammAccountID),
         [&](LedgerEntryType nodeType,
             uint256 const&,
@@ -219,7 +219,7 @@ deleteAMMTrustLines(
                 return tecINTERNAL;
             }
 
-            return deleteAMMTrustLine(view, sleItem, ammAccountID, j);
+            return deleteAMMTrustLine(sb, sleItem, ammAccountID, j);
         },
         j,
         maxTrustlinesToDelete);
@@ -227,33 +227,37 @@ deleteAMMTrustLines(
 
 TER
 deleteAMMAccount(
-    ApplyView& view,
+    Sandbox& sb,
     Issue const& asset,
     Issue const& asset2,
     beast::Journal j)
 {
-    auto ammSle = view.peek(keylet::amm(asset, asset2));
+    auto ammSle = sb.peek(keylet::amm(asset, asset2));
     if (!ammSle)
         return tecINTERNAL;
 
     auto const ammAccountID = (*ammSle)[sfAccount];
-    auto sleAMMRoot = view.peek(keylet::account(ammAccountID));
+    auto sleAMMRoot = sb.peek(keylet::account(ammAccountID));
     if (!sleAMMRoot)
-        return tecINTERNAL;
-
-    if (auto const res = deleteAMMTrustLines(
-            view, ammAccountID, maxDeletableAMMTrustLines, j);
-        std::get<TER>(res) != tesSUCCESS)
-        return std::get<TER>(res);
-    // All trustlines are deleted, can delete ltAMM and the account
-    else if (res.second == AllNodesDeleted::Yes)
     {
-        view.erase(ammSle);
-        view.erase(sleAMMRoot);
-        return tesSUCCESS;
+        JLOG(j.error()) << "deleteAMMAccount: AMM account does not exist "
+                        << to_string(ammAccountID);
     }
 
-    return tecINCOMPLETE;
+    auto const [ter, done] =
+        deleteAMMTrustLines(sb, ammAccountID, maxDeletableAMMTrustLines, j);
+
+    if (ter != tesSUCCESS)
+        return ter;
+
+    if (!done)
+        return tecINCOMPLETE;
+
+    sb.erase(ammSle);
+    if (sleAMMRoot)
+        sb.erase(sleAMMRoot);
+
+    return tesSUCCESS;
 }
 
 void
--- a/src/ripple/app/misc/impl/AMMUtils.cpp
+++ b/src/ripple/app/misc/impl/AMMUtils.cpp
@@ -188,7 +188,7 @@ ammAccountHolds(
     return STAmount{issue};
 }
 
-static std::pair<TER, bool>
+static TER
 deleteAMMTrustLines(
     Sandbox& sb,
     AccountID const& ammAccountID,
@@ -234,7 +234,11 @@ deleteAMMAccount(
 {
     auto ammSle = sb.peek(keylet::amm(asset, asset2));
     if (!ammSle)
+    {
+        JLOG(j.error()) << "deleteAMMAccount: AMM object does not exist "
+                        << asset << " " << asset2;
         return tecINTERNAL;
+    }
 
     auto const ammAccountID = (*ammSle)[sfAccount];
     auto sleAMMRoot = sb.peek(keylet::account(ammAccountID));
@@ -242,16 +246,13 @@ deleteAMMAccount(
     {
         JLOG(j.error()) << "deleteAMMAccount: AMM account does not exist "
                         << to_string(ammAccountID);
+        return tecINTERNAL;
     }
 
-    auto const [ter, done] =
-        deleteAMMTrustLines(sb, ammAccountID, maxDeletableAMMTrustLines, j);
-
-    if (ter != tesSUCCESS)
+    if (auto const ter =
+            deleteAMMTrustLines(sb, ammAccountID, maxDeletableAMMTrustLines, j);
+        ter != tesSUCCESS || ter == tecINCOMPLETE)
         return ter;
-
-    if (!done)
-        return tecINCOMPLETE;
 
     auto const ownerDirKeylet = keylet::ownerDir(ammAccountID);
     if (sb.exists(ownerDirKeylet) && !sb.emptyDirDelete(ownerDirKeylet))
@@ -262,8 +263,7 @@ deleteAMMAccount(
     }
 
     sb.erase(ammSle);
-    if (sleAMMRoot)
-        sb.erase(sleAMMRoot);
+    sb.erase(sleAMMRoot);
 
     return tesSUCCESS;
 }
--- a/src/ripple/app/misc/impl/AMMUtils.cpp
+++ b/src/ripple/app/misc/impl/AMMUtils.cpp
@@ -188,4 +188,110 @@ ammAccountHolds(
     return STAmount{issue};
 }
 
+static std::pair<TER, AllNodesDeleted>
+deleteAMMTrustLines(
+    ApplyView& view,
+    AccountID const& ammAccountID,
+    std::uint16_t maxTrustlinesToDelete,
+    beast::Journal j)
+{
+    return cleanupOnAccountDelete(
+        view,
+        keylet::ownerDir(ammAccountID),
+        [&](LedgerEntryType nodeType,
+            uint256 const&,
+            std::shared_ptr<SLE>& sleItem) -> TER {
+            // Should only have the trustlines
+            if (nodeType != LedgerEntryType::ltRIPPLE_STATE)
+            {
+                JLOG(j.error())
+                    << "deleteAMMTrustLines: deleting non-trustline "
+                    << nodeType;
+                return tecINTERNAL;
+            }
+
+            // Trustlines must have zero balance
+            if (sleItem->getFieldAmount(sfBalance) != beast::zero)
+            {
+                JLOG(j.error())
+                    << "deleteAMMTrustLines: deleting trustline with "
+                       "non-zero balance.";
+                return tecINTERNAL;
+            }
+
+            return deleteAMMTrustLine(view, sleItem, ammAccountID, j);
+        },
+        j,
+        maxTrustlinesToDelete);
+}
+
+TER
+deleteAMMAccount(
+    ApplyView& view,
+    Issue const& asset,
+    Issue const& asset2,
+    beast::Journal j)
+{
+    auto ammSle = view.peek(keylet::amm(asset, asset2));
+    if (!ammSle)
+        return tecINTERNAL;
+
+    auto const ammAccountID = (*ammSle)[sfAccount];
+    auto sleAMMRoot = view.peek(keylet::account(ammAccountID));
+    if (!sleAMMRoot)
+        return tecINTERNAL;
+
+    if (auto const res = deleteAMMTrustLines(
+            view, ammAccountID, maxDeletableAMMTrustLines, j);
+        std::get<TER>(res) != tesSUCCESS)
+        return std::get<TER>(res);
+    // All trustlines are deleted, can delete ltAMM and the account
+    else if (res.second == AllNodesDeleted::Yes)
+    {
+        view.erase(ammSle);
+        view.erase(sleAMMRoot);
+        return tesSUCCESS;
+    }
+
+    return tecINCOMPLETE;
+}
+
+void
+initializeFeeAuctionVote(
+    ApplyView& view,
+    std::shared_ptr<SLE>& ammSle,
+    AccountID const& account,
+    Issue const& lptIssue,
+    std::uint16_t tfee)
+{
+    // AMM creator gets the voting slot.
+    STArray voteSlots;
+    STObject voteEntry{sfVoteEntry};
+    if (tfee != 0)
+        voteEntry.setFieldU16(sfTradingFee, tfee);
+    voteEntry.setFieldU32(sfVoteWeight, 100000);
+    voteEntry.setAccountID(sfAccount, account);
+    voteSlots.push_back(voteEntry);
+    ammSle->setFieldArray(sfVoteSlots, voteSlots);
+    // AMM creator gets the auction slot for free.
+    auto& auctionSlot = ammSle->peekFieldObject(sfAuctionSlot);
+    auctionSlot.setAccountID(sfAccount, account);
+    // current + sec in 24h
+    auto const expiration = std::chrono::duration_cast<std::chrono::seconds>(
+                                view.info().parentCloseTime.time_since_epoch())
+                                .count() +
+        TOTAL_TIME_SLOT_SECS;
+    auctionSlot.setFieldU32(sfExpiration, expiration);
+    auctionSlot.setFieldAmount(sfPrice, STAmount{lptIssue, 0});
+    // Set the fee
+    if (tfee != 0)
+        ammSle->setFieldU16(sfTradingFee, tfee);
+    else if (ammSle->isFieldPresent(sfTradingFee))
+        ammSle->makeFieldAbsent(sfTradingFee);
+    if (auto const dfee = tfee / AUCTION_SLOT_DISCOUNTED_FEE_FRACTION)
+        auctionSlot.setFieldU16(sfDiscountedFee, dfee);
+    else if (auctionSlot.isFieldPresent(sfDiscountedFee))
+        auctionSlot.makeFieldAbsent(sfDiscountedFee);
+}
+
 }  // namespace ripple
--- a/src/ripple/app/tx/impl/AMMBid.cpp
+++ b/src/ripple/app/tx/impl/AMMBid.cpp
@@ -94,6 +94,10 @@ AMMBid::preclaim(PreclaimContext const& ctx)
         return terNO_AMM;
     }
 
+    auto const lpTokensBalance = (*ammSle)[sfLPTokenBalance];
+    if (lpTokensBalance == beast::zero)
+        return tecAMM_EMPTY;
+
     if (ctx.tx.isFieldPresent(sfAuthAccounts))
     {
         for (auto const& account : ctx.tx.getFieldArray(sfAuthAccounts))
@@ -114,7 +118,6 @@ AMMBid::preclaim(PreclaimContext const& ctx)
         JLOG(ctx.j.debug()) << "AMM Bid: account is not LP.";
         return tecAMM_INVALID_TOKENS;
     }
-    auto const lpTokensBalance = (*ammSle)[sfLPTokenBalance];
 
     auto const bidMin = ctx.tx[~sfBidMin];
 
@@ -204,10 +207,10 @@ applyBid(
                           Number const& burn) -> TER {
         auctionSlot.setAccountID(sfAccount, account_);
         auctionSlot.setFieldU32(sfExpiration, current + TOTAL_TIME_SLOT_SECS);
-        if (fee == 0)
-            auctionSlot.makeFieldAbsent(sfDiscountedFee);
-        else
+        if (fee != 0)
             auctionSlot.setFieldU16(sfDiscountedFee, fee);
+        else if (auctionSlot.isFieldPresent(sfDiscountedFee))
+            auctionSlot.makeFieldAbsent(sfDiscountedFee);
         auctionSlot.setFieldAmount(
             sfPrice, toSTAmount(lpTokens.issue(), minPrice));
         if (ctx_.tx.isFieldPresent(sfAuthAccounts))
--- a/src/ripple/app/tx/impl/AMMDelete.cpp
+++ b/src/ripple/app/tx/impl/AMMDelete.cpp
@@ -1,0 +1,89 @@
+//------------------------------------------------------------------------------
+/*
+    This file is part of rippled: https://github.com/ripple/rippled
+    Copyright (c) 2022 Ripple Labs Inc.
+
+    Permission to use, copy, modify, and/or distribute this software for any
+    purpose  with  or without fee is hereby granted, provided that the above
+    copyright notice and this permission notice appear in all copies.
+
+    THE  SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+    WITH  REGARD  TO  THIS  SOFTWARE  INCLUDING  ALL  IMPLIED  WARRANTIES  OF
+    MERCHANTABILITY  AND  FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+    ANY  SPECIAL ,  DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+    WHATSOEVER  RESULTING  FROM  LOSS  OF USE, DATA OR PROFITS, WHETHER IN AN
+    ACTION  OF  CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+    OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+//==============================================================================
+
+#include <ripple/app/tx/impl/AMMDelete.h>
+
+#include <ripple/app/misc/AMMUtils.h>
+#include <ripple/ledger/Sandbox.h>
+#include <ripple/protocol/AMMCore.h>
+#include <ripple/protocol/Feature.h>
+#include <ripple/protocol/STAccount.h>
+#include <ripple/protocol/TER.h>
+#include <ripple/protocol/TxFlags.h>
+
+namespace ripple {
+
+NotTEC
+AMMDelete::preflight(PreflightContext const& ctx)
+{
+    if (!ammEnabled(ctx.rules))
+        return temDISABLED;
+
+    if (auto const ret = preflight1(ctx); !isTesSuccess(ret))
+        return ret;
+
+    if (ctx.tx.getFlags() & tfUniversalMask)
+    {
+        JLOG(ctx.j.debug()) << "AMM Delete: invalid flags.";
+        return temINVALID_FLAG;
+    }
+
+    if (auto const res = invalidAMMAssetPair(ctx.tx[sfAsset], ctx.tx[sfAsset2]))
+    {
+        JLOG(ctx.j.debug()) << "AMM Delete: Invalid asset pair.";
+        return res;
+    }
+
+    return preflight2(ctx);
+}
+
+TER
+AMMDelete::preclaim(PreclaimContext const& ctx)
+{
+    auto const ammSle =
+        ctx.view.read(keylet::amm(ctx.tx[sfAsset], ctx.tx[sfAsset2]));
+    if (!ammSle)
+    {
+        JLOG(ctx.j.debug()) << "AMM Delete: Invalid asset pair.";
+        return terNO_AMM;
+    }
+
+    auto const lpTokensBalance = (*ammSle)[sfLPTokenBalance];
+    if (lpTokensBalance != beast::zero)
+        return tecAMM_EMPTY;
+
+    return tesSUCCESS;
+}
+
+TER
+AMMDelete::doApply()
+{
+    // This is the ledger view that we work against. Transactions are applied
+    // as we go on processing transactions.
+    Sandbox sb(&ctx_.view());
+
+    auto const ter =
+        deleteAMMAccount(sb, ctx_.tx[sfAsset], ctx_.tx[sfAsset2], j_);
+    if (ter == tesSUCCESS || ter == tecINCOMPLETE)
+        sb.apply(ctx_.rawView());
+
+    return ter;
+}
+
+}  // namespace ripple
--- a/src/ripple/app/tx/impl/AMMDelete.cpp
+++ b/src/ripple/app/tx/impl/AMMDelete.cpp
@@ -44,12 +44,6 @@ AMMDelete::preflight(PreflightContext const& ctx)
         return temINVALID_FLAG;
     }
 
-    if (auto const res = invalidAMMAssetPair(ctx.tx[sfAsset], ctx.tx[sfAsset2]))
-    {
-        JLOG(ctx.j.debug()) << "AMM Delete: Invalid asset pair.";
-        return res;
-    }
-
     return preflight2(ctx);
 }
 
@@ -66,7 +60,7 @@ AMMDelete::preclaim(PreclaimContext const& ctx)
 
     auto const lpTokensBalance = (*ammSle)[sfLPTokenBalance];
     if (lpTokensBalance != beast::zero)
-        return tecAMM_EMPTY;
+        return tecAMM_NOT_EMPTY;
 
     return tesSUCCESS;
 }
--- a/src/ripple/app/tx/impl/AMMWithdraw.cpp
+++ b/src/ripple/app/tx/impl/AMMWithdraw.cpp
@@ -302,6 +302,9 @@ AMMWithdraw::applyGuts(Sandbox& sb)
     if (!ammSle)
         return {tecINTERNAL, false};
     auto const ammAccountID = (*ammSle)[sfAccount];
+    auto const accountSle = sb.read(keylet::account(ammAccountID));
+    if (!accountSle)
+        return {tecINTERNAL, false};
     auto const lpTokens =
         ammLPHolds(ctx_.view(), *ammSle, ctx_.tx[sfAccount], ctx_.journal);
     auto const lpTokensWithdraw =
@@ -377,7 +380,9 @@ AMMWithdraw::applyGuts(Sandbox& sb)
         return {result, false};
 
     bool updateBalance = true;
-    if (newLPTokenBalance == beast::zero)
+    // Delete only if number of trustlines is less or equal than max
+    if (newLPTokenBalance == beast::zero &&
+        accountSle->getFieldU32(sfOwnerCount) <= maxDeletableAMMTrustLines)
     {
         if (auto const ter =
                 deleteAMMAccount(sb, ctx_.tx[sfAsset], ctx_.tx[sfAsset2], j_);
--- a/src/ripple/app/tx/impl/AMMWithdraw.cpp
+++ b/src/ripple/app/tx/impl/AMMWithdraw.cpp
@@ -194,8 +194,10 @@ AMMWithdraw::preclaim(PreclaimContext const& ctx)
     if (!expected)
         return expected.error();
     auto const [amountBalance, amount2Balance, lptAMMBalance] = *expected;
+    if (lptAMMBalance == beast::zero)
+        return tecAMM_EMPTY;
     if (amountBalance <= beast::zero || amount2Balance <= beast::zero ||
-        lptAMMBalance <= beast::zero)
+        lptAMMBalance < beast::zero)
     {
         JLOG(ctx.j.debug())
             << "AMM Withdraw: reserves or tokens balance is zero.";
@@ -371,19 +373,31 @@ AMMWithdraw::applyGuts(Sandbox& sb)
         return std::make_pair(tecINTERNAL, STAmount{});
     }();
 
-    // AMM is deleted if zero tokens balance
-    if (result == tesSUCCESS && newLPTokenBalance != beast::zero)
+    if (result != tesSUCCESS)
+        return {result, false};
+
+    bool updateBalance = true;
+    if (newLPTokenBalance == beast::zero)
+    {
+        if (auto const ter =
+                deleteAMMAccount(sb, ctx_.tx[sfAsset], ctx_.tx[sfAsset2], j_);
+            ter != tesSUCCESS && ter != tecINCOMPLETE)
+            return {ter, false};
+        else
+            updateBalance = (ter == tecINCOMPLETE);
+    }
+
+    if (updateBalance)
     {
         ammSle->setFieldAmount(sfLPTokenBalance, newLPTokenBalance);
         sb.update(ammSle);
-
-        JLOG(ctx_.journal.trace())
-            << "AMM Withdraw: tokens " << to_string(newLPTokenBalance.iou())
-            << " " << to_string(lpTokens.iou()) << " "
-            << to_string(lptAMMBalance.iou());
     }
 
-    return {result, result == tesSUCCESS};
+    JLOG(ctx_.journal.trace())
+        << "AMM Withdraw: tokens " << to_string(newLPTokenBalance.iou()) << " "
+        << to_string(lpTokens.iou()) << " " << to_string(lptAMMBalance.iou());
+
+    return {tesSUCCESS, true};
 }
 
 TER
@@ -398,85 +412,6 @@ AMMWithdraw::doApply()
         sb.apply(ctx_.rawView());
 
     return result.first;
-}
-
-TER
-AMMWithdraw::deleteAccount(Sandbox& sb, AccountID const& ammAccountID)
-{
-    auto sleAMMRoot = sb.peek(keylet::account(ammAccountID));
-    auto ammSle = sb.peek(keylet::amm(ctx_.tx[sfAsset], ctx_.tx[sfAsset2]));
-
-    if (!sleAMMRoot || !ammSle)
-        return tecINTERNAL;
-
-    // Delete the trustlines
-    Keylet const ownerDirKeylet{keylet::ownerDir(ammAccountID)};
-    std::shared_ptr<SLE> sleDirNode{};
-    unsigned int uDirEntry{0};
-    uint256 dirEntry{beast::zero};
-    if (sb.exists(ownerDirKeylet) &&
-        dirFirst(sb, ownerDirKeylet.key, sleDirNode, uDirEntry, dirEntry))
-    {
-        do
-        {
-            auto sleItem = sb.peek(keylet::child(dirEntry));
-            if (!sleItem)
-            {
-                // Directory node has an invalid index
-                JLOG(j_.fatal())
-                    << "AMM Withdraw: Directory node in ledger " << sb.seq()
-                    << " has index to object that is missing: "
-                    << to_string(dirEntry);
-                return tefBAD_LEDGER;
-            }
-
-            // Should only have the trustlines
-            if (auto const nodeType = safe_cast<LedgerEntryType>(
-                    sleItem->getFieldU16(sfLedgerEntryType));
-                nodeType != LedgerEntryType::ltRIPPLE_STATE)
-            {
-                JLOG(j_.error())
-                    << "AMM Withdraw: deleting non-trustline " << nodeType;
-                return tecINTERNAL;
-            }
-
-            // Trustlines must have zero balance
-            if (sleItem->getFieldAmount(sfBalance) != beast::zero)
-            {
-                JLOG(j_.error()) << "AMM Withdraw: deleting trustline with "
-                                    "non-zero balance.";
-                return tecINTERNAL;
-            }
-
-            auto const& [lowAccountID, highAccountID] = std::minmax(
-                sleItem->getFieldAmount(sfHighLimit).getIssuer(), ammAccountID);
-            if (auto const ter = trustDelete(
-                    sb, sleItem, lowAccountID, highAccountID, ctx_.journal);
-                ter != tesSUCCESS)
-            {
-                JLOG(j_.error())
-                    << "AMM Withdraw: failed to delete the trustline.";
-                return ter;
-            }
-
-            // Update the iterator
-            assert(uDirEntry == 1);
-            if (uDirEntry != 1)
-            {
-                JLOG(j_.error())
-                    << "AMM Withdraw: iterator re-validation failed.";
-                return tefBAD_LEDGER;
-            }
-            uDirEntry = 0;
-
-        } while (
-            dirNext(sb, ownerDirKeylet.key, sleDirNode, uDirEntry, dirEntry));
-    }
-
-    sb.erase(ammSle);
-    sb.erase(sleAMMRoot);
-
-    return tesSUCCESS;
 }
 
 std::pair<TER, STAmount>
@@ -621,9 +556,6 @@ AMMWithdraw::withdraw(
             << "AMM Withdraw: failed to withdraw LPTokens";
         return {res, STAmount{}};
     }
-
-    if (lpTokensWithdrawActual == lpTokensAMMBalance)
-        return {deleteAccount(view, ammAccount), STAmount{}};
 
     return {tesSUCCESS, lpTokensAMMBalance - lpTokensWithdrawActual};
 }
--- a/src/ripple/app/tx/impl/AMMWithdraw.cpp
+++ b/src/ripple/app/tx/impl/AMMWithdraw.cpp
@@ -380,9 +380,7 @@ AMMWithdraw::applyGuts(Sandbox& sb)
         return {result, false};
 
     bool updateBalance = true;
-    // Delete only if number of trustlines is less or equal than max
-    if (newLPTokenBalance == beast::zero &&
-        accountSle->getFieldU32(sfOwnerCount) <= maxDeletableAMMTrustLines)
+    if (newLPTokenBalance == beast::zero)
     {
         if (auto const ter =
                 deleteAMMAccount(sb, ctx_.tx[sfAsset], ctx_.tx[sfAsset2], j_);
--- a/src/ripple/app/tx/impl/AMMWithdraw.cpp
+++ b/src/ripple/app/tx/impl/AMMWithdraw.cpp
@@ -25,7 +25,6 @@
 #include <ripple/ledger/Sandbox.h>
 #include <ripple/ledger/View.h>
 #include <ripple/protocol/AMMCore.h>
-#include <ripple/protocol/Feature.h>
 #include <ripple/protocol/STAccount.h>
 #include <ripple/protocol/TxFlags.h>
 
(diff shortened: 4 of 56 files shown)


RichardAH:
If I can ask, why not refactor the DeleteAccount transactor so you can call it from AMMWithdraw? or possibly even better: don't delete the account on the last withdrawal but allow anyone to delete it with the DeleteAccount transactor. I think having two different places in code where all the complexities of account deletion are dealt with is just asking to introduce bugs later. Some new object will be added at some point and one of the delete routines wont be updated for it. We should have a single canonical way to delete an account.

gregtatcam:
> If I can ask, why not refactor the DeleteAccount transactor so you can call it from AMMWithdraw? or possibly even better: don't delete the account on the last withdrawal but allow anyone to delete it with the DeleteAccount transactor. I think having two different places in code where all the complexities of account deletion are dealt with is just asking to introduce bugs later. Some new object will be added at some point and one of the delete routines wont be updated for it. We should have a single canonical way to delete an account.

Thought of reusing AccountDelete crossed my mind. Except there is not much in common between deleting a regular account and deleting the AMM account. In fact, the former explicitly doesn't delete the trustlines. One common part is the loop to iterate over the owner directory objects. I think making this loop, or as much of `AccountDelete::doApply()` as possible, into a function with a callback to handle the specifics of account deletion makes sense. 

I don't quite see how allowing anyone to delete AMM account with AccountDelete is going to work. It doesn't seem to be fair to all liquidity providers (LP). Even if AccountDelete is restricted to LP only it doesn't feel fair either. Perhaps LP voting on deletion might work but this is quite a bit of rework for both the specs and the implementation.

RichardAH:
> I don't quite see how allowing anyone to delete AMM account with AccountDelete is going to work. It doesn't seem to be fair to all liquidity providers (LP). Even if AccountDelete is restricted to LP only it doesn't feel fair either. Perhaps LP voting on deletion might work but this is quite a bit of rework for both the specs and the implementation.

Hi Greg 

This would only be if issued LP tokens = 0. I.e. after final withdraw. Optionally the empty AMM could be left for future users (although you would need to modify AMMDeposit for this I think).

> I think making this loop, or as much of `AccountDelete::doApply()` as possible, into a function with a callback to handle the specifics of account deletion makes sense.

Indeed. Personally would refactor DeleteAccount so the thing apply() calls is a templated function that lets me "dry delete" (try a pretend delete and see if there are issues) and "hard delete" (delete no matter what). For an example of how this can be done see dry/wet runs for IOUEscrow here: https://github.com/XRPLF/rippled/blob/eea7fcb8b1df92fd5e9ba51aa7f10f55f2bb280d/src/ripple/ledger/View.h#L488

There's a few reasons I would do it this way but the most important is destructor consistency. If you consider that ledger objects are created, live on the ledger and then eventually are removed from the ledger, and these objects may have dependent sub objects, then deleting the objects correctly becomes important such that there are no orphaned objects (and no possibility of orphaned objects). This means there should be a canonical way to delete each type of object, not 2 or 3 different deletion routines throughout the code that all worked when they were written but open an exploit when unrelated code is added to the ledger later.

IMHO there should be one canonical place to go to delete an account regardless of why you're deleting it.

nbougalis (approved):
This change looks reasonable and, on first look, addresses the serious issues that we reported last week.

I would recommend one more addition, probably in `TrustSet::preclaim`  (you could do it in `trustCreate`, but that function and its 7,000 arguments ought to die): if one side of the trustline is an AMM, then _ONLY_ create the trustline if the asset is the AMM's token. Since AMMs don't issue _any_ other assets, trustlines for anything else are meaningless and just serve to litter the owner directory of the AMM.

intelliot:
Proposed commit message, including acknowledgement block:

```
fix(AMM): prevent orphaned objects, inconsistent ledger state: (#4626)

When an AMM account is deleted, the owner directory entries must be
deleted in order to ensure consistent ledger state.

* When deleting AMM account:
  * Clean up AMM owner dir, linking AMM account and AMM object
  * Delete trust lines to AMM
* Disallow check create to AMM
* Constrain entries in AuthAccounts array to be accounts
  * AuthAccounts is an array of objects for the AMMBid transaction

This patch applies to the AMM implementation in #4294.

Acknowledgements:
Richard Holland and Nik Bougalis for responsibly disclosing this issue.

We welcome reviews of the project code and urge researchers to

    bugs@xrpl.org
```

src/ripple/app/tx/impl/AMMWithdraw.cpp
```diff
@@ -410,9 +409,78 @@ AMMWithdraw::deleteAccount(Sandbox& sb, AccountID const& ammAccountID)
...
+            if (sleItem->getFieldAmount(sfBalance) != beast::zero)
+            {
+                JLOG(j_.error()) << "AMM Withdraw: deleting trustline with "
+                                    "non-zero balance.";
+                return tecINTERNAL;
+            }
+
+            auto const& [lowAccountID, highAccountID] = std::minmax(
+                sleItem->getFieldAmount(sfHighLimit).getIssuer(), ammAccountID);
+            if (auto const ter = trustDelete(
```
RichardAH, line 461: Suppose I create a trustline to the AMM, increasing my ownercount by 1.
Then the AMM is subsequently deleted, deleting my trustline object.
Where in this code is my ownercount returned to me? I don't believe trustDelete handles this?

RichardAH:
It doesn't appear that lsfAllowTrustLineClawback is checked for? What happens if an issuer of one of the assets that the AMM holds claws it back out of the AMM?

src/ripple/app/tx/impl/AMMWithdraw.cpp
gregtatcam: Thanks, will update.

gregtatcam:
> IMHO there should be one canonical place to go to delete an account regardless of why you're deleting it.

Hi Richard,

This makes sense. It could be addressed by a common function, say `cleanupOnAccountDelete()`, which is called in `AccountDelete::doApply()` to handle a regular account clean up and in `AMMWithdraw::deleteAccount()` as part of the auto-delete on a final withdraw to handle AMM account clean up. This function cleans up the owner directory for either a regular or AMM account. This is more than an iterator with a callback that I suggested before. I think it still makes sense though to keep the auto-delete for the following reasons:

1) This is a big enough change to the way AMMs work that if we made the change we can't just throw it in at the last minute. We're not against making changes that substantially improve things, but unless this does we should keep the current spec.
2) AMMs are different enough from regular accounts that "auto delete" makes a lot of sense, and if we did use "AccountDelete" allowing another account to do the delete is a big change to that transaction.

gregtatcam:
> This makes sense. It could be addressed by a common function, say `cleanupOnAccountDelete()`, which is called in `AccountDelete::doApply()` to handle a regular account clean up and in `AMMWithdraw::deleteAccount()` as part of the auto-delete on a final withdraw to handle AMM account clean up. This function cleans up the owner directory for either a regular or AMM account. This is more than an iterator with a callback that I suggested before. 

Actually I think a callback to get a specific object deleter is still needed. Otherwise would have to bring different ledger entry deletion functions from `AccountDelete` transactor into the module that's going to host this common function (I'm thinking View.h/cpp is a good place for it). So this common function would take `nonObligationDeleter` for a regular account and some other function to delete the trustlines in case of AMM.

intelliot:
@gregtatcam - going forward, please sign your commits, as mentioned in [CONTRIBUTING](https://github.com/XRPLF/rippled/blob/develop/CONTRIBUTING.md). (Past commits are fine to leave as-is)

nbougalis:
Something occurred to me (several hours later) but the new deletion code still has a problem @gregtatcam:  it can be coerced into doing a very large amount of work (since there's no limitation on how many trustlines to the AMM's token can exist). This is a huge problem: not only would the transaction execution take a lot of time, but it could create a very large amount of metadata and that's a separate problem which has been an issue before (see e.g. the offer crossing limits or the remove-found-unfunded-offers-as-they're-found code, which runs even if the offer crossing fails for other reasons.)

Also, @RichardAH that's a nice catch, with the ownercounts. It would have sucked to miss this.

RichardAH:
> Something occurred to me (several hours later) but the new deletion code still has a problem @gregtatcam: it can be coerced into doing a very large amount of work (since there's no limitation on how many trustlines to the AMM's token can exist). This is a huge problem: not only would the transaction execution take a lot of time, but it could create a very large amount of metadata and that's a separate problem which has been an issue before (see e.g. the offer crossing limits or the remove-found-unfunded-offers-as-they're-found code, which runs even if the offer crossing fails for other reasons.)
> 
> Also, @RichardAH that's a nice catch, with the ownercounts. It would have sucked to miss this.

This also is a problem indeed. There aren’t many good solutions to it either. Until the final withdrawal all the LP token trustlines are technically valid because those users might obtain LP tokens.

The only (obvious) safe way to do this is the final withdrawal puts the AMM into a cleanup state. Then another transactor AMMDelete can do an amortised deletion of the trustlines with a maximum per iteration. It would either return tesSUCCESS or tecMORE (meaning it needs to run again.) In cleanup mode it should be impossible to add more trustlines.

It’s not a good solution but it is *a* solution.

gregtatcam:
> This also is a problem indeed. There aren’t many good solutions to it either. Until the final withdrawal all the LP token trustlines are technically valid because those users might obtain LP tokens.

Can we delete a trustline on a withdraw if the account's LP tokens are 0? The auto-delete could only happen when all LP tokens are 0, which means individual LP tokens are also 0. Should this mitigate this issue.

RichardAH:
> > This also is a problem indeed. There aren’t many good solutions to it either. Until the final withdrawal all the LP token trustlines are technically valid because those users might obtain LP tokens.
> 
> Can we delete a trustline on a withdraw if the account's LP tokens are 0? The auto-delete could only happen when all LP tokens are 0, which means individual LP tokens are also 0. Should this mitigate this issue?

That's fine, the problem arises when there are 100,000 trustlines

gregtatcam:
> That's fine, the problem arises when there are 100,000 trustlines

So it doesn't need to be mitigated?

gregtatcam:
> I would recommend one more addition, probably in `TrustSet::preclaim` (you could do it in `trustCreate`, but that function and its 7,000 arguments ought to die): if one side of the trustline is an AMM, then _ONLY_ create the trustline if the asset is the AMM's token. Since AMMs don't issue _any_ other assets, trustlines for anything else are meaningless and just serve to litter the owner directory of the AMM.

@nbougalis 

I can't tell from AMM's root account this AMM's LP Token `Currency`. To figure out I'd need to get AMM object `ltAMM` to find this AMM's `Currency`. This means I'd need to iterate over AMM account owner directory entries, which leads back to your concern of iterating over AMM trustlines.

RichardAH:
> So it doesn't need to be mitigated?

It does. It's entirely possible an AMM has a lot of trustlines at the time it's being deleted. Without a cap on the iteration you will exceed the metadata size limit not to mention taking a long time to close the ledger.

RichardAH:
> I can't tell from AMM's root account this AMM's LP Token `Currency`. To figure out I'd need to get AMM object `ltAMM` to find this AMM's `Currency`. This means I'd need to iterate over AMM account owner directory entries, which leads back to your concern of iterating over AMM trustlines.

This is a design flaw, since these objects are inherently linked you should be able to follow a pointer in either direction.
However a workaround if you don't want to add another optional AccountRoot field is just check the first entry in the first directory page. It's always going to be the ltAMM (unless I'm missing something)

gregtatcam:
>  is just check the first entry in the first directory page. It's always going to be the ltAMM (unless I'm missing something)

Thanks. Because ltAMM is the first object to be inserted.

gregtatcam:
> It does. It's entirely possible an AMM has a lot of trustlines at the time it's being deleted. Without a cap on the iteration you will exceed the metadata size limit not to mention taking a long time to close the ledger.

Would it not make sense to clean up as much as possible on each withdraw if the account's LP tokens balance is 0? My concern is that there is no incentive to call `AMMDelete`, if we take this path, except for being a good XRPL citizen.

gregtatcam:
> It does. It's entirely possible an AMM has a lot of trustlines at the time it's being deleted. Without a cap on the iteration you will exceed the metadata size limit not to mention taking a long time to close the ledger.

The following can be done to address this issue:

- Don't allow the trustline creation to AMM except for LP tokens per @nbougalis  suggestion.
- Delete the account's LP tokens trustline on withdraw if all account's LP tokens are withdrawn. The trustline is auto-created so can as well auto-delete it.
- Auto-delete the trustlines on the final Withdraw; i.e. when the total LP token's balance is 0. Limit the number of the trustlines to delete per @seelabs suggestion. Tentative limit is 1,500 (same as maxOffersToConsider). This should address the clean up for the vast majority of normal AMM usage.
- Add `AMMDelete` for a manual cleanup per @RichardAH suggestion.

RichardAH:
> Would it not make sense to clean up as much as possible on each withdraw if the account's LP tokens balance is 0? My concern is that there is no incentive to call `AMMDelete`, if we take this path, except for being a good XRPL citizen.

You can incentivise it: Instead of "feeing" an owner reserve for creation, store the xrp in the AMM root and release 90% of it to the successful deleter?

RichardAH:
> > > is just check the first entry in the first directory page. It's always going to be the ltAMM (unless I'm missing something)
> 
> Hm, looks like can get a trustline first. Is it reasonable to add an optional AccountRoot field for `sfAMMID`. I think can get rid of the directory entry then.

I think doubly linking two closely connected objects is entirely justified.

nbougalis:
> > I would recommend one more addition, probably in `TrustSet::preclaim` (you could do it in `trustCreate`, but that function and its 7,000 arguments ought to die): if one side of the trustline is an AMM, then _ONLY_ create the trustline if the asset is the AMM's token. Since AMMs don't issue _any_ other assets, trustlines for anything else are meaningless and just serve to litter the owner directory of the AMM.
> 
> @nbougalis
> 
> I can't tell from AMM's root account this AMM's LP Token `Currency`. To figure out I'd need to get AMM object `ltAMM` to find this AMM's `Currency`. This means I'd need to iterate over AMM account owner directory entries, which leads back to your concern of iterating over AMM trustlines.

That seems like a minor design flaw, if I’m being honest, but the good news is that, by design, the AMM will always be in the first page of the owner directory, so the amount of iterations is capped at 32.

Beyond that, there’s another option: a currency identifier is 160-bits, so you could always just make it equal to a well-known 160-bit value that is accessible to you always: the AMM account root ID.

gregtatcam:
> That seems like a minor design flaw, if I’m being honest, but the good news is that, by design, the AMM will always be in the first page of the owner directory, so the amount of iterations is capped at 32.
> 
> Beyond that, there’s another option: a currency identifier is 160-bits, so you could always just make it equal to a well-known 160-bit value that is accessible to you always: the AMM account root ID.

I can just add AMMID to the AMM account root and delete the directory entry linking the root and ltAMM. This saves some space and makes it a lot easier to access ltAMM via the account root.

gregtatcam:
> @gregtatcam - going forward, please sign your commits, as mentioned in [CONTRIBUTING](https://github.com/XRPLF/rippled/blob/develop/CONTRIBUTING.md). (Past commits are fine to leave as-is)

@intelliot I set up the ssh signature. I'm using CLion for dev and github. It doesn't appear CLion took the config I added for signing. So I'm still pushing the unsigned commits. I need to investigate.

gregtatcam:
> @gregtatcam I think I use the instructions here: https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

Thanks. I think I got it. Should work next time.

ximinez:
>     * Delete the account's LP tokens trustline on withdraw if all account's LP tokens are withdrawn. The trustline is auto-created so can as well auto-delete it.

I think the account's LP tokens trustline should be left in place if it's not in the default state on the LP's side. For example, the auto-created trustline will have a limit of 0. Say I want to participate in the ecosystem as someone who takes LP tokens for payment, but not hold them for long. The process might look like this:
1. Set up a trustline to the AMM (via `AMMDeposit` or `TrustSet`).
2. If not done in step 1, change my limit on the trustline via `TrustSet`.
3. Accept a payment.
4. `AMMWithdraw` all of my LP tokens.
5. Repeat steps 3 & 4.

If the trustline is auto-deleted, then I'll have to start over at step 1 every time I withdraw. That's annoying and introduces race conditions - someone could send me a payment before I recreate the trustline - or forces me to keep dust amounts in the trustline. I have no idea how common this use-case might be, but I think it is reasonable to support it.

gregtatcam:
> I think the account's LP tokens trustline should be left in place if it's not in the default state on the LP's side. For example, 

Yeah, makes sense.

RichardAH:
A thought: suppose trustlines can be marked as “issued by AMM”. Firstly if this were the case we could probably do away with placing the LP lines in the AMM ownerdir. These lines can be restricted for use in any pathing or crossing and if the issuer doesn’t exist anymore all the user can do is delete them. They become similar to a frozen line.

gregtatcam:
> A thought: suppose trustlines can be marked as “issued by AMM”. Firstly if this were the case we could probably do away with placing the LP lines in the AMM ownerdir. These lines can be restricted for use in any pathing or crossing and if the issuer doesn’t exist anymore all the user can do is delete them. They become similar to a frozen line.

The two trustlines created by AMM for the token pair have `lsfAMMNode` flag set.

RichardAH:
> > A thought: suppose trustlines can be marked as “issued by AMM”. Firstly if this were the case we could probably do away with placing the LP lines in the AMM ownerdir. These lines can be restricted for use in any pathing or crossing and if the issuer doesn’t exist anymore all the user can do is delete them. They become similar to a frozen line.
> 
> The two trustlines created by AMM for the token pair have `lsfAMMNode` flag set.

Those are easy to clean up. I’m talking about when people open trustlines to an AMM account, those lines should be marked and treated differently. That would allow amortised cleanup rather than having to delete them all in one ledger

gregtatcam:
> Those are easy to clean up. I’m talking about when people open trustlines to an AMM account, those lines should be marked and treated differently. That would allow amortised cleanup rather than having to delete them all in one ledger

How would this cleanup work?

RichardAH:
> How would this cleanup work?
The trustline is not iterated for pathing or crossing and acts as if frozen so each user can delete it when they want to, even after the AMM account is gone.

gregtatcam:
> > How would this cleanup work?
> 
> The trustline is not iterated for pathing or crossing and acts as if frozen so each user can delete it when they want to, even after the AMM account is gone.

I just pushed a commit with AMMDelete per earlier suggestion. The auto-delete on final withdraw has a limit of 1,500 trustlines to delete, and so does AMMDelete. 1,500 should address the vast majority of cases; i.e. the auto-delete on final withdraw will delete all trustlines, amm object and account. AMMDelete return tecINCOMPLETE if it doesn't delete all remaining trustlines so a user can call AMMDelete again. There is a new tfTwoAssetIfEmpty option in deposit to revert AMM from empty to normal state.

Perhaps marking the trustline as AMM can be considered as another enhancement or have it implemented on another AMM iteration.

intelliot:
@gregtatcam Could you summarize all of the APIs introduced for AMM and add to `API-CHANGELOG.md`?

This includes new transactors, flags, transaction result codes, etc.

intelliot:
@RichardAH @nbougalis - please review

src/ripple/app/tx/impl/AMMWithdraw.cpp
```diff
@@ -372,19 +373,31 @@ AMMWithdraw::applyGuts(Sandbox& sb)
...
-            << " " << to_string(lpTokens.iou()) << " "
-            << to_string(lptAMMBalance.iou());
     }
 
-    return {result, result == tesSUCCESS};
+    JLOG(ctx_.journal.trace())
+        << "AMM Withdraw: tokens " << to_string(newLPTokenBalance.iou()) << " "
+        << to_string(lpTokens.iou()) << " " << to_string(lptAMMBalance.iou());
+
+    return {tesSUCCESS, true};
```
RichardAH, line 400: if AMMWithdraw has delete functionality then it should behave the same way as AMMDelete, by returning tecIMCOMPLETE when there is more to delete.

If you are determined to code it in this way then I suggest only running deleteAMMAccount here if you know it will result in complete deletion otherwise don't run it at all here and let someone clean it up with AMMDelete.

src/ripple/app/tx/impl/InvariantCheck.cpp
```diff
@@ -321,12 +321,14 @@ AccountRootsNotDeleted::finalize(
...
     beast::Journal const& j)
 {
-    // AMM account root can be deleted as the result of AMM withdraw
+    // AMM account root can be deleted as the result of AMM withdraw/delete
     // transaction when the total AMM LP Tokens balance goes to 0.
     // Not every AMM withdraw deletes the AMM account, accountsDeleted_
     // is set if it is deleted.
     if ((tx.getTxnType() == ttACCOUNT_DELETE ||
-         (tx.getTxnType() == ttAMM_WITHDRAW && accountsDeleted_ == 1)) &&
+         ((tx.getTxnType() == ttAMM_WITHDRAW ||
```
RichardAH, line 329: This is another reason to only let AMMDelete do deletions. The more txn types that are allowed to result in an account root being removed the less secure this invariant is.

src/ripple/app/tx/impl/Transactor.cpp
```diff
@@ -757,6 +757,25 @@ removeExpiredNFTokenOffers(
...
+    beast::Journal viewJ)
+{
+    std::size_t removed = 0;
+
+    for (auto const& index : trustLines)
+    {
+        if (auto const sleState = view.peek({ltRIPPLE_STATE, index}))
+        {
+            deleteAMMTrustLine(view, sleState, std::nullopt, viewJ);
+            if (++removed == maxDeletableAMMTrustLines)
```
RichardAH, line 773: No reason to use a counter here we know how many entries are in the vector when the function starts. If there are more than maxDeletableAMMTrustLines entries this is an error because the transactor should never have tried to delete more than this?
You should ensure that you tune maxDeletableAMMTrustLines such that metadata will never be oversized.

src/ripple/protocol/Protocol.h
```diff
@@ -95,6 +95,8 @@ using LedgerIndex = std::uint32_t;
 */
 using TxID = uint256;
 
+std::size_t constexpr maxDeletableAMMTrustLines = 1500;
```
RichardAH, line 98: Too high imho. 512?

RichardAH (approved):
No obvious blockers. My main concern is that the final AMMWithdraw might delete 1500 trustlines while still leaving the zombie AMM on the ledger with however many remaining trustlines, while returning tesSUCCESS. I feel this will be utterly confusing if it happens. I think Withdraw should not delete unless it knows it can completely delete.

src/ripple/ledger/impl/View.cpp
```diff
@@ -1544,4 +1544,119 @@ requireAuth(ReadView const& view, Issue const& issue, AccountID const& account)
...
+    if (!sleLow || !sleHigh)
+        return tecINTERNAL;
+    bool const ammLow = sleLow->getFlags() & lsfAMM;
+    bool const ammHigh = sleHigh->getFlags() & lsfAMM;
+
+    // One side must be AMM
+    if (!(ammLow ^ ammHigh) ||
+        (ammAccountID &&
+         ((ammLow && low != ammAccountID) ||
+          (ammHigh && high != ammAccountID))))
```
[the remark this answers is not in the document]
RichardAH: @gregtatcam can you check the std::optional unwrap (in your code) here, should it be `*ammAccountID` ?

gregtatcam:
> No obvious blockers. My main concern is that the final AMMWithdraw might delete 1500 trustlines while still leaving the zombie AMM on the ledger with however many remaining trustlines, while returning tesSUCCESS. I feel this will be utterly confusing if it happens. I think Withdraw should not delete unless it knows it can completely delete.

Failing the final withdraw seems unfair to a user. What can the user do at this point? Reach out to all accounts that have the trustlines and ask them to delete the trustlines? Also, how would then AMM go into the empty state with consequent `AMMDelete` submission. 

How about returning `tecINCOMPLETE` in this case?

gregtatcam:
> > No obvious blockers. My main concern is that the final AMMWithdraw might delete 1500 trustlines while still leaving the zombie AMM on the ledger with however many remaining trustlines, while returning tesSUCCESS. I feel this will be utterly confusing if it happens. I think Withdraw should not delete unless it knows it can completely delete.
> 
> Failing the final withdraw seems unfair to a user. What can the user do at this point? Reach out to all accounts that have the trustlines and ask them to delete the trustlines? Also, how would then AMM go into the empty state with consequent `AMMDelete` submission.
> 
> How about returning `tecINCOMPLETE` in this case?

Ah, sorry. Just saw your other comment regarding returning `tecINCOMPLETE` in withdraw.

src/ripple/app/tx/impl/AMMWithdraw.cpp
gregtatcam: Returning `tecINCOMPLETE` makes sense.

We already went through the trouble of iterating over 1,500 trustlines, deleting them, and adjusting the owner count, so why through this effort away? It also results in fewer calls to AMMDelete.

src/ripple/protocol/Protocol.h
gregtatcam: It takes 14-17 milliseconds to delete 1,500 trustlines and the metadata size is about 700K. So 1,500 seems acceptable.

src/ripple/app/tx/impl/Transactor.cpp
gregtatcam: It should not delete more. What if it does happen? Shouldn't we try to delete as many as we deem safe?

src/ripple/app/tx/impl/InvariantCheck.cpp
gregtatcam: True, but it would make for a bad user experience. I think in vast majority of the cases there will be way fewer than 1,500 trustlines, especially now that we are not allowing trustlines to AMM unless it's LPTokens. Consequently, the final withdraw is going to be able to delete all trustlines and delete AMM account and object in majority of the cases.

src/ripple/app/tx/impl/AMMDelete.cpp
```diff
@@ -0,0 +1,89 @@
...
+    if (auto const ret = preflight1(ctx); !isTesSuccess(ret))
+        return ret;
+
+    if (ctx.tx.getFlags() & tfUniversalMask)
+    {
+        JLOG(ctx.j.debug()) << "AMM Delete: invalid flags.";
+        return temINVALID_FLAG;
+    }
+
+    if (auto const res = invalidAMMAssetPair(ctx.tx[sfAsset], ctx.tx[sfAsset2]))
```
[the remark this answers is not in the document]
RichardAH: It's defensive programming. If _somehow_ the system is in a bad state, is there a graceful way to recover? The "somehow" is not important but for the sake of argument let's suppose a different (future) transactor malfunctioned and deleted the AMM's account root but not the AMM object.

src/ripple/app/tx/impl/AMMWithdraw.cpp
RichardAH: Mainly because the person who did the withdrawal doesn't care about deleting the AMM, they care about withdrawing their funds from the AMM. For every prior withdrawal until this final one, the withdrawer gets tesSUCCESS (i.e. withdrawal was successful.) The last one get tecINCOMPLETE. That's confusing as hell to them. Was the withdrawal incomplete? Overloading the functionality at this edge case requires losing information. Either the result code is consistent with withdraw or it's consistent with delete but it can't be both, unless it's a 100% delete every time.

src/ripple/app/tx/impl/Transactor.cpp
RichardAH: No it's a logic error. This code should not proceed in an unsafe state.

src/ripple/app/tx/impl/AMMWithdraw.cpp
gregtatcam: I think you are suggesting this. Please correct me if I'm wrong.

>        if last withdraw
>            if trustlines less than 1,500
>                delete trustlines
>                delete account
>                return tesSUCCESS
>            set LPTokens balance to 0
>            return tecINCOMPLETE

Is there an efficient way to get the number of account's trustlines?

src/ripple/protocol/Protocol.h
gregtatcam: I changed to 512. AMMDelete unit-tests take too long with 1,500. 512 is good enough to cover majority of cases.

src/ripple/app/tx/impl/AMMWithdraw.cpp
RichardAH: You could overload ownercount to count the number of incoming trustlines (since ownercount isn't really used in AMM?)
I would always return tesSUCCESS from AMMWithdrawal if the *withdrawal* was successful. Whatever ancillary clean-up activity the transactor also does is irrelevant to its return code and purpose. From the perspective of users and end developers they couldn't care less if the GC finished or not, all they want to know about is if the withdrawal was successful or not.

gregtatcam:
@RichardAH I pushed a commit that should address your latest comments.

ximinez (changes requested):
Partial review

src/ripple/app/tx/impl/AMMWithdraw.cpp
ximinez: I propose that in an AMM withdrawal, the withdrawal is the point. If that succeeds, the the transaction succeeds. Deleting trust lines is a helpful side effect. 

I suggest something like

* If withdraw fails
  * return appropriate tec code
* if last withdraw
  * delete up to X trustlines
  * if deletion fails
    * return appropriate tec code
  * if all trust lines deleted
    *  delete account
  * else
    * set LPTokens balance to 0
*   return tesSUCCESS

No need to change the meaning of OwnerCount

src/ripple/app/misc/impl/AMMUtils.cpp
```diff
@@ -188,4 +188,110 @@ ammAccountHolds(
...
+    Issue const& asset2,
+    beast::Journal j)
+{
+    auto ammSle = view.peek(keylet::amm(asset, asset2));
+    if (!ammSle)
+        return tecINTERNAL;
+
+    auto const ammAccountID = (*ammSle)[sfAccount];
+    auto sleAMMRoot = view.peek(keylet::account(ammAccountID));
+    if (!sleAMMRoot)
```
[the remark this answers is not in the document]
RichardAH: > Counterpoint: If the account root doesn't exist, something has gone very, very wrong. Continuing with the deletion process _may_ make things worse. Arguably, the ledger and transaction history should be examined manually to determine what happened and what's going on. The solution may require a fix amendment not just to clean up the ledger, but to prevent the issue from happening again.

This would be an ungraceful failure mode. Half an AMM, if it happened, would probably crash the ledger or block an orderbook. You can't clean it up because of this pedantic check, even after restarting the whole network. Now you need to wait for an amendment and code update. You should seek to build graceful failure modes into production systems.

src/ripple/app/tx/impl/AMMBid.cpp
```diff
@@ -204,10 +207,10 @@ applyBid(
                           Number const& burn) -> TER {
         auctionSlot.setAccountID(sfAccount, account_);
         auctionSlot.setFieldU32(sfExpiration, current + TOTAL_TIME_SLOT_SECS);
-        if (fee == 0)
-            auctionSlot.makeFieldAbsent(sfDiscountedFee);
-        else
+        if (fee != 0)
             auctionSlot.setFieldU16(sfDiscountedFee, fee);
+        else if (auctionSlot.isFieldPresent(sfDiscountedFee))
+            auctionSlot.makeFieldAbsent(sfDiscountedFee);
```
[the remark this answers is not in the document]
RichardAH: I also prefer to check if it exists first. There is an interface inconsistency between STObject and the Ledger. If you try to delete an SLE that doesn't exist, it's a logic error. If you try to delete a field that doesn't exist it ignores you. Just because someone made a mistake programming the APIs doesn't mean programmers should make a mistake using them.

src/ripple/app/misc/impl/AMMUtils.cpp
```diff
@@ -188,4 +188,110 @@ ammAccountHolds(
...
+            view, ammAccountID, maxDeletableAMMTrustLines, j);
+        std::get<TER>(res) != tesSUCCESS)
+        return std::get<TER>(res);
+    // All trustlines are deleted, can delete ltAMM and the account
+    else if (res.second == AllNodesDeleted::Yes)
+    {
+        view.erase(ammSle);
+        view.erase(sleAMMRoot);
+        return tesSUCCESS;
+    }
```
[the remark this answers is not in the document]
gregtatcam: I already pushed a commit yesterday replacing it with bool.

src/ripple/app/tx/impl/AMMWithdraw.cpp
gregtatcam: This is not a good user experience from the withdrawer perspective. The withdrawer just want to withdraw the funds and doesn't care if the delete fails or not.

src/ripple/app/misc/impl/AMMUtils.cpp
```diff
@@ -188,4 +188,110 @@ ammAccountHolds(
...
+    if (!sleAMMRoot)
+        return tecINTERNAL;
+
+    if (auto const res = deleteAMMTrustLines(
+            view, ammAccountID, maxDeletableAMMTrustLines, j);
+        std::get<TER>(res) != tesSUCCESS)
+        return std::get<TER>(res);
+    // All trustlines are deleted, can delete ltAMM and the account
+    else if (res.second == AllNodesDeleted::Yes)
+    {
```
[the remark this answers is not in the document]
nbougalis: Please avoid using `LogicError`. Rationale: that will *kill a server dead*. It should only be used if recovery is actually impossible and almost never in the transaction processing code.

Instead, returning a `tecINTERNAL` and rolling back, allows the transaction to be entombed in the ledger for future review.

src/ripple/app/tx/impl/AMMBid.cpp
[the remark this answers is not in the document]
nbougalis: If you think of the more modern API, which uses `std::optional`, then `makeFieldAbsent` is kind of like assigning `std::nullopt` to a `std::optional`, which is fine whether the `std::optional` is seated or not.

I don’t mind checking, but I don’t see a problem with not checking.

nbougalis (approved):
Left some minor comments, but broadly looks fine. Thanks for quickly tackling the bug report. 🙏

src/ripple/app/tx/impl/AMMDelete.cpp
```diff
@@ -0,0 +1,89 @@
...
+        ctx.view.read(keylet::amm(ctx.tx[sfAsset], ctx.tx[sfAsset2]));
+    if (!ammSle)
+    {
+        JLOG(ctx.j.debug()) << "AMM Delete: Invalid asset pair.";
+        return terNO_AMM;
+    }
+
+    auto const lpTokensBalance = (*ammSle)[sfLPTokenBalance];
+    if (lpTokensBalance != beast::zero)
+        return tecAMM_EMPTY;
```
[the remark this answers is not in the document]
gregtatcam: It's `tecAMM_NOT_EMPTY` in the latest commit.

src/ripple/protocol/impl/TER.cpp
```diff
@@ -45,6 +45,7 @@ transResults()
         MAKE_ERROR(tecAMM_BALANCE,                   "AMM has invalid balance."),
         MAKE_ERROR(tecAMM_INVALID_TOKENS,            "AMM invalid LP tokens."),
         MAKE_ERROR(tecAMM_FAILED,                    "AMM transaction failed."),
+        MAKE_ERROR(tecAMM_EMPTY,                     "AMM is in empty state."),
```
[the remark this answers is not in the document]
gregtatcam: Added `tecAMM_NOT_EMPTY` in the the latest commit.

src/ripple/app/misc/impl/AMMUtils.cpp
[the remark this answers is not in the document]
gregtatcam: Retruning `tecINTERNAL` in the latest commit.

src/ripple/app/tx/impl/AMMBid.cpp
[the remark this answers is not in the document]
gregtatcam: It'll throw `fieldNotFound` if it doesn't exist. It's a default field. I can check whether it has the default value. It's still a check. Try this test:
>     try
>        {
>           auto sle = std::make_shared<SLE>(keylet::amm(xrpIssue(), USD));
>            auto& auctionSlot = sle->peekFieldObject(sfAuctionSlot);
>            auctionSlot.makeFieldAbsent(sfTradingFee);
>        } catch (std::exception const& e)
>        {
>            std::cout << "exception " << e.what() << std::endl;
>        }

src/ripple/app/tx/impl/AMMWithdraw.cpp
gregtatcam: I actually think `OwnerCount` is pretty useful in providing an estimate:
- Clean up scope for `AMMDelete` caller
- Number of Liquidity Providers

gregtatcam:
> It doesn't appear that lsfAllowTrustLineClawback is checked for? What happens if an issuer of one of the assets that the AMM holds claws it back out of the AMM?

@RichardAH I pushed a commit addressing this. Isn't it a bit tricky though? We can't just take the funds out of AMM but then the issues are not going to be able to meet the regulatory requirements. AMM is like a safe heaven from clawback.

src/ripple/app/misc/impl/AMMUtils.cpp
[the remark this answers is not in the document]
ximinez: After thinking about this some more, I think it actually makes sense to add this kind of check as a global invariant: Deleted accounts should have no other objects in the ledger. If that kind of invariant existed before the AMM work was started, the flaw in the original deletion functionality would have been caught as soon as the first unit test was written.

RichardAH:
> AMM is like a safe heaven from clawback.

Perhaps provide issuers an additional account flag: lsfDISALLOW_AMM? Basically signalling that their asset cannot be used in an AMM (because it cannot be clawed back, or for any other reason they or their legal team have.)

gregtatcam:
> Perhaps provide issuers an additional account flag: lsfDISALLOW_AMM? Basically signalling that their asset cannot be used in an AMM (because it cannot be clawed back, or for any other reason they or their legal team have.)

Maybe it's sufficient to check for `lsfAllowTrustLineClawback`. And if set then don't allow AMM creation with the assets from this issuer?

src/ripple/app/misc/impl/AMMUtils.cpp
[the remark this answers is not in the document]
ximinez: If the ledger is corrupt, we're already in a failure mode. Better to stop than to continue in an state that may not be correct.

https://xrpl.org/intro-to-consensus.html#consensus-protocol-properties
> These properties are sometimes summarized as the following principles, in order of priority: Correctness, Agreement, Forward Progress.

If correctness is lost, then we do not want to make forward progress. This is not just me being stubborn, this is baked into the design of the ledger from the beginning.

ximinez (changes requested):

src/ripple/ledger/impl/View.cpp
```diff
@@ -1544,4 +1544,125 @@ requireAuth(ReadView const& view, Issue const& issue, AccountID const& account)
...
+    if (auto const ter = trustDelete(view, sleState, low, high, j);
+        ter != tesSUCCESS)
+    {
+        JLOG(j.error())
+            << "deleteAMMTrustLine: failed to delete the trustline.";
+        return ter;
+    }
+
+    adjustOwnerCount(view, !ammLow ? sleLow : sleHigh, -1, j);
+    adjustOwnerCount(view, ammLow ? sleLow : sleHigh, -1, j);
```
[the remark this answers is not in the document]
RichardAH: Agree this looks wrong indeed

src/ripple/app/tx/impl/SetTrust.cpp
```diff
@@ -128,21 +128,42 @@ SetTrust::preclaim(PreclaimContext const& ctx)
...
+
+    if (!sleDst)
+        return tecNO_DST;
+
+    auto const dstFlags = sleDst->getFlags();
+
     // If the destination has opted to disallow incoming trustlines
     // then honour that flag
     if (ctx.view.rules().enabled(featureDisallowIncoming))
     {
```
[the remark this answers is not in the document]
gregtatcam: Should it return a different error instead of `tecNO_DST` in the second case? If we get there then `featureDisallowIncoming` is not enabled.

src/ripple/app/tx/impl/AMMWithdraw.cpp
```diff
@@ -372,19 +376,33 @@ AMMWithdraw::applyGuts(Sandbox& sb)
...
 
-    // AMM is deleted if zero tokens balance
-    if (result == tesSUCCESS && newLPTokenBalance != beast::zero)
+    if (result != tesSUCCESS)
+        return {result, false};
+
+    bool updateBalance = true;
+    // Delete only if number of trustlines is less or equal than max
+    if (newLPTokenBalance == beast::zero &&
+        accountSle->getFieldU32(sfOwnerCount) <= maxDeletableAMMTrustLines)
```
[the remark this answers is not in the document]
gregtatcam: The `OwnerCount` change is not needed then. I'll roll it back.

src/ripple/app/misc/impl/AMMUtils.cpp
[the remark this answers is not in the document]
gregtatcam: Perhaps this can be addressed by another PR...

src/ripple/protocol/impl/LedgerFormats.cpp
```diff
@@ -56,6 +56,7 @@ LedgerFormats::LedgerFormats()
             {sfMintedNFTokens,       soeDEFAULT},
             {sfBurnedNFTokens,       soeDEFAULT},
             {sfFirstNFTokenSequence, soeOPTIONAL},
+            {sfAMMID,                soeOPTIONAL},
```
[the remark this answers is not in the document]
gregtatcam: I'm not going to hate you for this and it's not a big deal to change. But I don't quite like the account type inference as `AMMID` field side effect. It might also be a problem for the XRPL explorer apps.

src/ripple/app/tx/impl/SetTrust.cpp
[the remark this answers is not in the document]
ximinez: No, I think `tecNO_DST` makes the most sense, since there is no destination account to check for the flag. It's just that we have to check, and it's unfortunate that we have to check twice, but them's the breaks.

I suppose we could structure the check slightly differently, but then we'd be checking the rules twice:
```
// Do not reference sleDst without an amendment. Add the amendment to the if statement below so it can be checked.
auto const sleDst = ctx.view.read(keylet::account(uDstAccountID));
if (ctx.view.rules().enabled(featureDisallowIncoming) || ammEnabled(ctx.view.rules()))
{
    if (!sleDst)
        return tecNO_DST;
}
```

src/ripple/app/misc/impl/AMMUtils.cpp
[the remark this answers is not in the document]
ximinez: > Perhaps this can be addressed by another PR...

Yeah, that would be appropriate.

src/ripple/protocol/impl/LedgerFormats.cpp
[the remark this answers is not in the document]
intelliot: Just a thought (feel free to ignore) - It may be possible to avoid changing the external API by returning a synthetic boolean field based on the presence of AMMID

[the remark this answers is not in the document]
ximinez: > It might also be a problem for the XRPL explorer apps.

Yeah, those are some of the people I expect to hate me. :laughing: 

But kidding aside, it's not like we're repurposing a field to hold the AMMID. This is a new dedicated field, exclusively (at least for now) for AMMs. If it's populated, then it's an AMM. I don't think it's an inference, I think it's a clear flag.

> It may be possible to avoid changing the external API by returning a synthetic boolean field based on the presence of AMMID

If you're talking about the recently-added `account_flags` returned in `account_info`, I notice that it doesn't have support for the `lsfAMM` flag anyway. I think it would be perfectly sensible to add a synthetic "AMM" flag there.

However, anything else that returns the raw `AccountRoot` object would be a little harder, but it'll have the AMMID field, so explorers will have to check for that. The other thing that comes to mind is transaction metadata. I'm pretty sure that'll have the AMMID in the metadata regardless, though. Need to confirm.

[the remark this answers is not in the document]
ximinez: > Need to confirm.

Confirmed:
```
      {
         "ModifiedNode" : {
            "FinalFields" : {
               "AMMID" : "27ED2C339D335D3CA09BB6C85FF19685A65C00A105D68242383EDB3E06942757",
               "Account" : "rnV55X6wNHXQNostoLPQYZZPzBmDnSb14K",
               "Balance" : "11000000000",
               "Flags" : 1099956224,
               "OwnerCount" : 1,
               "Sequence" : 6
            },
            "LedgerEntryType" : "AccountRoot",
            "LedgerIndex" : "DF7F696214208B2C7FB92F28DE6FEE45921D159EF11D7D924294A18CE4A827E7",
            "PreviousFields" : {
               "Balance" : "10000000000"
            },
            "PreviousTxnID" : "7297888514F54699417843AEB220730B00D8EA2B2B06C2EF218A23285922EE7A",
            "PreviousTxnLgrSeq" : 6
         }
      },
```
This would be good to check in a unit test somewhere.

src/ripple/app/tx/impl/Clawback.cpp
```diff
@@ -63,9 +63,6 @@ Clawback::preclaim(PreclaimContext const& ctx)
     if (!sleIssuer || !sleHolder)
         return terNO_ACCOUNT;
 
-    if (sleHolder->getFlags() & lsfAMM)
-        return tecAMM_ACCOUNT;
```
ximinez: I think this check is worth keeping. When it comes to ledger safety, a little redundancy is OK. It should be impossible for the holder to be an AMM, but if it happens, better to stop the transaction from processing.

src/ripple/app/tx/impl/InvariantCheck.cpp
```diff
@@ -326,9 +326,7 @@ AccountRootsNotDeleted::finalize(
     // Not every AMM withdraw deletes the AMM account, accountsDeleted_
     // is set if it is deleted.
     if ((tx.getTxnType() == ttACCOUNT_DELETE ||
```
ximinez, line 328: The comment before this `if` statement needs updating. I suggested:
```
    // A successful AccountDelete or AMMDelete MUST delete exactly
    // one account root
```

src/ripple/ledger/View.h
```diff
@@ -461,11 +461,11 @@ requireAuth(ReadView const& view, Issue const& issue, AccountID const& account);
 /** Cleanup owner directory entries on account delete.
  * Used for a regular and AMM accounts deletion. The caller
  * has to provide the deleter function, which handles details of
- * specific account deletion.
- * @return {tesSUCCESS, false} indicates maxNodesToDelete
+ * specific account-owned object deletion.
+ * @return tesINCOMPLETE indicates maxNodesToDelete
```
ximinez, line 465: Typo: `tesINCOMPLETE` -> `tecINCOMPLETE`

src/ripple/app/tx/impl/SetTrust.cpp
```diff
@@ -541,6 +547,10 @@ SetTrust::doApply()
             uQualityIn,
             uQualityOut,
             viewJ);
+
+        // Maintain number of AMM trustlines in AMM root account
+        if (terResult == tesSUCCESS && (sleDst->getFlags() & lsfAMM))
+            adjustOwnerCount(view(), sleDst, 1, j_);
```
ximinez, line 553: Missed a spot.

ximinez (changes requested):

src/ripple/protocol/LedgerFormats.h
```diff
@@ -249,9 +249,8 @@ enum LedgerSpecificFlags {
         0x10000000,               // True, reject new paychans
     lsfDisallowIncomingTrustline =
         0x20000000,               // True, reject new trustlines (only if no issued assets)
-    lsfAMM = 0x40000000,          // True, AMM account
     lsfAllowTrustLineClawback =
-        0x80000000,               // True, enable clawback
+        0x40000000,               // True, enable clawback
```
ximinez, line 253: We need to make sure this change is well publicized.

gregtatcam: What is the process for this? This feature is pending 12.1.0 release.

ximinez (approved):
I've got a couple of minor comments, but this is ready to sign off regardless.

src/ripple/protocol/LedgerFormats.h
intelliot: Is there any way we can keep the clawback flag the same as it was before?

Client libraries have already merged support for Clawback, [implementing the prior value](https://github.com/XRPLF/xrpl.js/pull/2353/files#diff-ecaed36481e2b2545d4bc7180df7092251f25835586b90d837be4dd11dceed62R208).

gregtatcam: Sure, I'll change it back.

intelliot: The ideal is to change it back.

For future reference: if the flag must be changed, the process is to put the `API Change` label on the PR. (Come to to think of it - there are other API Changes in this PR, so I'll put that label on regardless. They only affect APIs which have not been released yet, but I know people have already been implementing/testing AMM, so we should publicize the changes for their sake.) While it's the responsibility of client library maintainers and API integrators to review release notes and look at the API Changelog, in cases where we know what the relevant/impacted apps/projects are, it would be nice courtesy to inform them directly as well.

ximinez: Woops. Thought I'd posted this earlier:

The big issue is that the flag is specified in the Clawback spec (https://github.com/XRPLF/XRPL-Standards/tree/master/XLS-39d-clawback#32-account-root-modifications). Plus, I'm sure there are downstream projects that have already started implementing Clawback support using this value.

I think it makes the most sense to change this back and leave a comment explicitly saying that `0x40000000` is available.

injaelee:
@RichardAH We would appreciate it if you can allocate some time to do a final review on this PR. We would like to move forward. (cc: @gregtatcam)

src/ripple/app/tx/impl/SetTrust.cpp
```diff
@@ -128,20 +129,46 @@ SetTrust::preclaim(PreclaimContext const& ctx)
...
+    // If destination is AMM and the trustline doesn't exist then only
+    // allow SetTrust if the asset is AMM LP token and AMM is not
+    // in empty state.
+    if (ammEnabled(ctx.view.rules()))
+    {
+        if (!sleDst)
+            return tecNO_DST;
+
+        if (sleDst->isFieldPresent(sfAMMID) &&
+            !ctx.view.read(keylet::line(id, uDstAccountID, currency)))
```
RichardAH, line 155: What happens if the issuer of one of the two currencies that the AMM holds uses SetTrust to alter the lines created by the AMM? It's just allowed to happen? What if they set no-ripple, freeze, etc?

RichardAH:
https://github.com/XRPLF/rippled/blob/2efa0f8d4c3c8b9061efbd03e4e9041aa99aaa63/src/ripple/app/tx/impl/AMMCreate.cpp#L96-L101
What happens if that AMM object exists but is in delete-only mode? I.e. more trustlines left to delete, whilst someone is trying to create it again? Should the create transactor check if it can finish the deletion rather than giving an error here?

RichardAH:
Does the AMM pay transfer fees for the trustlines it uses now? If so can you show me where?

gregtatcam:
> Does the AMM pay transfer fees for the trustlines it uses now? If so can you show me where?
> 
> Because it looks like it doesn't. I reported this previously.
> 
> https://github.com/XRPLF/rippled/blob/2efa0f8d4c3c8b9061efbd03e4e9041aa99aaa63/src/ripple/app/paths/impl/BookStep.cpp#L556-L567

In this case you are referring to the quality function, which is used in one path payment with AMM liquidity to figure out how to adjust the requested `out` amount so that the path produces the requested quality. This calculation does adjust the qualities for the transfer fees. But the fees are not paid at this point. The transfer fees are calculated in [`forEachOffer()`](https://github.com/XRPLF/rippled/blob/develop/src/ripple/app/paths/impl/BookStep.cpp#L717-L730), where they are adjusted by the specific `offer` type. The fees work the same way as for the regular account offer, except on offer crossing, when the offer's owner pays the fees. This is disabled for AMM since we can't just move funds out of AMM.

gregtatcam:
> Does the AMM pay transfer fees for the trustlines it uses now? If so can you show me where?

There are transfer unit-tests for the offer crossing and regular payments: https://github.com/XRPLF/rippled/blob/develop/src/test/app/AMM_test.cpp#L3130-L3209

ximinez (approved):

src/ripple/app/tx/impl/InvariantCheck.cpp
```diff
@@ -321,12 +321,12 @@ AccountRootsNotDeleted::finalize(
...
-    // AMM account root can be deleted as the result of AMM withdraw
+    // AMM account root can be deleted as the result of AMM withdraw/delete
     // transaction when the total AMM LP Tokens balance goes to 0.
-    // Not every AMM withdraw deletes the AMM account, accountsDeleted_
-    // is set if it is deleted.
+    // A successful AccountDelete or AMMDelete MUST delete exactly
+    // one account root.
     if ((tx.getTxnType() == ttACCOUNT_DELETE ||
-         (tx.getTxnType() == ttAMM_WITHDRAW && accountsDeleted_ == 1)) &&
+         tx.getTxnType() == ttAMM_DELETE) &&
```
RichardAH, line 329: This should be tx.getTxnType() == AMM_WITHDRAW || tx.getTxnType() == AMM_DELETE right?

gregtatcam:
> I'm fine with these fixes now

@RichardAH thanks for reviewing.

src/ripple/app/tx/impl/AMMWithdraw.cpp
mDuo13: I'm with Ed on this one, actually. I really don't like the idea that AMMWithdraw returns `tecINCOMPLETE` if it successfully withdrew funds.

IMO if an AMMWithdraw transaction withdraws the requested amount, it should return `tesSUCCESS`, full stop. Whether or not it also succeeded in fully cleaning up the AMM object is a detail that you can figure out from the metadata, but is not usually relevant to the user doing the withdrawal.

The meaning of AMMDelete is different so it makes sense that the semantics for `tecINCOMPLETE` should be different there. The "delete" transaction should return success if and only if the AMM was (fully) deleted. The `tecINCOMPLETE` result code is useful for telling you that it made progress on the thing you asked it to do, but isn't done; that's completely different than the Withdraw case where getting a `tec`-class code should always indicate that the withdraw operation didn't happen.

src/ripple/protocol/impl/TER.cpp
```diff
@@ -45,6 +45,9 @@ transResults()
...
+        MAKE_ERROR(tecAMM_NOT_EMPTY,                 "AMM is not in empty state."),
+        MAKE_ERROR(tecAMM_ACCOUNT,                   "AMM Account."),
```
mDuo13, line 50: This error message is not very informative. How about this?

```suggestion
        MAKE_ERROR(tecAMM_ACCOUNT,                   "This operation is not allowed on an AMM Account."),
```

That said, I also think `tecNO_PERMISSION` could be an acceptable existing code to use in this scenario.

legleux:
LGTM! :rocket:

intelliot:
@gregtatcam - I see that you added a new commit 4 hours ago. Can you confirm if this PR is ready to merge now?

gregtatcam:
> @gregtatcam - I see that you added a new commit 4 hours ago. Can you confirm if this PR is ready to merge now?

I cherry picked commits from #4642 to fix the merge conflict into develop. It's ready to merge now.

intelliot:
Released with 1.12.0 via 58f7aecb0bb1cb0a1f00fef45454c4fb5b407677

PR #4628, opened 2023-07-18
Eliminate built-in SNTP support (fixes #4207):

### LINKED ISSUE #4207: SNTP Clock issues
## Issue Description

The purpose of this issue is to provide a place to discuss a couple of issues surrounding the time keeping used to create ledger close times.

Currently rippled communicates with an NTP server to get the correct time, and this time serves as the basis for determining ledger close times.  Which NTP server is used is configurable in rippled.cfg under the `[sntp-servers]` tag where one or more servers can be listed, and one will be chosen for the job.  It might be possible, however remotely, for a malicious actor to intercept communications between a  node in the XRP network and its NTP server, and shift the network time of the node away from its peers.

1.  Should a node use only one NTP server?  We could, for example, query three or more servers, eliminate any reporting a time sufficiently far from the mean, and average the rest for the current time.
2.   Should we be using NTP servers at all?  We could just query the OS for the current time (e.g. `std::chrono::system_clock::now()`) and leave it up to server operators to keep their machines temporarily synced.  This would allow us to remove the code we currently have that reaches out to NTP servers, and remove the associated config tag.
3.  Is the status quo the best option and thus we should leave well enough alone?

nbougalis:
The code contained a simple SNTP client that could be used to synchronize the clock that the code uses even when the system clock was unsynchronized.

Not only was the SNTP code "crufty" but during a recent audit of the codebase by Bishop-Fox, a potential attack vector that could allow an attacker with the ability to shape or redirect traffic to skew the server's clock was discovered.

This commit removes the SNTP client code, removes a mutex and generally simplifies the `TimeKeeper`.

- Fixes #4207

**Note to server operators:**
It is highly recommended to use time synchronization software to ensure that system clocks are accurate.

### Type of Change

- [X] Documentation Updates
- [ ] Release


intelliot:
How does this compare with https://github.com/HowardHinnant/rippled/commit/6c239bde8d82214286d2f538092eabb1beffafc3 ?

nbougalis:
> How does this compare with [HowardHinnant@6c239bd](https://github.com/HowardHinnant/rippled/commit/6c239bde8d82214286d2f538092eabb1beffafc3) ?

I feel that it is more thorough, restructuring and simplifying the code, instead of simply eliminating the SNTP code.

HowardHinnant (approved):
Looks good to me.

src/ripple/core/TimeKeeper.h
```diff
@@ -22,73 +22,98 @@
...
+                // towards our wall time.
+                if (by > 1s)
+                    return offset + ((by + 3s) / 4);
+
+                if (by < -1s)
+                    return offset + ((by - 3s) / 4);
+
+                return (offset * 3) / 4;
+            }());
+        assert(success);
```
seelabs, line 111: I don't really object, but doesn't asserting here mean you can unconditionally assign? If the assert fails, that means another thread changed the value between the time we read it and changed it. If we're asserting that never fails do we need the compare and exchange strong? Do we need the atomic?

Honestly, I think I'd keep the atomics - it makes the code thread safe, but I might remove the `assert`. What breaks if that `assert` fires?

seelabs (approved):
:+1: LGTM, but see the comment about the assert.

src/ripple/core/TimeKeeper.h
nbougalis: You're right; we could just unconditionally assign and nothing _really_ goes wrong. My intention here was to implement a "weak" form of the precondition that calling this function is _supposed_ to be externally serialized, but it's _extremely_ weak and subject to timing, instead of being a robust check.

I can just change this to be:

```
        closeOffset_ = [by, offset]() {
                // Ignore small offsets and push the close time
                // towards our wall time.
                if (by > 1s)
                    return offset + ((by + 3s) / 4);

                if (by < -1s)
                    return offset + ((by - 3s) / 4);

                return (offset * 3) / 4;
            }();
```

nbougalis:
@intelliot, I've addressed @seelabs's comment and rebased the code against the latest **develop** branch. I will let Scott do a final pass and give a thumbs up, but I think this is ready to go.

intelliot:
> I will let Scott do a final pass and give a thumbs up

Awaiting final pass by @seelabs . Also, there are (new) conflicts that must be resolved.

src/ripple/core/TimeKeeper.h
```diff
@@ -22,73 +22,99 @@
...
-    virtual void
-    run(std::vector<std::string> const& servers) = 0;
-
-    /** Returns the estimate of wall time, in network time.
+    // Adjust system_clock::time_point for NetClock epoch
+    static constexpr time_point
+    adjust(std::chrono::system_clock::time_point when)
+    {
+        return time_point(std::chrono::duration_cast<duration>(
+            when.time_since_epoch() - days(10957)));
```
seelabs, line 40: Just a nit: we usually like to separate the implementation from the interface so it's easier to see interfaces. I'm OK merging this as-is though. It's only a nit and we can clean it up latter (although I did want to flag it; for future changes we should try to stick with separate interfaces/implementations).

seelabs (approved):
:+1: Looks good. Thanks Nik!

intelliot:
```
Eliminate the built-in SNTP support (fixes #4207):
The codebase includes a simple SNTP client that is enabled by
default and used to adjust an application-level clock in case
the operating system's clock is not already synchronized.

A recent audit of the code conducted by Bishop-Fox identified
a potential attack vector, which could allow an attacker with
the capability to manipulate the UDP traffic flow of a server
by redirecting queries or by injecting crafted SNTP responses
to manipulate the correction factor, skewing the clock.

Although the impact of such an attack, if mounted, is minimal
and can be completely eliminated by editing the configuration
file to disable the built-in SNTP client, this commit instead
entirely eliminates the "crufty" 10+ year-old code, removes a
mutex and greatly simplifies the `TimeKeeper`.

**Note to server operators:**

It is highly recommended to use time synchronization software
to ensure that the system clock is accurate. For more details
see https://en.wikipedia.org/wiki/Ntpd#Implementations.
```

Issue #4550, opened 2023-05-30
[ledger_entry with invalid combination of parameters return internal error] (Version: [1.11.0-rc2]])


mounikakun:
## Steps to Reproduce

`{
    "method": "ledger_entry",
    "params": [
        {
            "check": {
                "owner": "rhigTLJJyXXSRUyRCQtqi1NoAZZzZnS4KU",
                "sub_index": 0
            },
            "ledger_index": "validated"
        }
    ]
}
`

## Expected Result

`invalidParams`

## Actual Result

`{
    "result": {
        "error": "internal",
        "error_code": 73,
        "error_message": "Internal error.",
        "request": {
            "check": {
                "owner": "rhigTLJJyXXSRUyRCQtqi1NoAZZzZnS4KU",
                "sub_index": 0
            },
            "command": "ledger_entry",
            "ledger_index": "validated"
        },
        "status": "error"
    }
}`

PR #4630, opened 2023-07-19
APIv2(ledger_entry) : check error

PeterChen13579:
Fix issue #4550  on ledger entry

### Context of Change
"check" field is verified to be a string. (Check the above Github issue link for more info.) Everything is under API Version 2 as it is a breaking change.

### Type of Change

- [X ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

Added own unit test for API Version 2;


src/ripple/rpc/handlers/LedgerEntry.cpp
```diff
@@ -71,7 +71,12 @@ doLedgerEntry(RPC::JsonContext& context)
     {
         expectedType = ltCHECK;
 
-        if (!uNodeIndex.parseHex(context.params[jss::check].asString()))
+        if (!context.params[jss::check].isString() && context.apiVersion > 1u)
```
connorjchen, line 74: Is this the only condition in which params are invalid?

src/test/rpc/LedgerRPC_test.cpp
```diff
@@ -1226,15 +1226,32 @@ class LedgerRPC_test : public beast::unit_test::suite
...
-        jvParams[jss::ledger_hash] = ledgerHash;
-        Json::Value const jrr =
-            env.rpc("json", "ledger_entry", to_string(jvParams))[jss::result];
+        {
+            jvParams[jss::features] = ledgerHash;
+            jvParams[jss::ledger_hash] = ledgerHash;
+            Json::Value const jrr = env.rpc(
+                "json", "ledger_entry", to_string(jvParams))[jss::result];
+
+            if (apiVersion < 2u)
```
connorjchen, line 1235: nit: why is apiVersion checked with < 2u in this file compared > 1u? seems better to keep consistent throughout codebase (if most files have already established some standard)

PeterChen13579: In the test file, it is specifically checking results for apiVersion < 2u and apiVersion other than that (>=2) , while in non-test files, the new error codes only execute if apiVersion is greater than 1. There is a guideline already established [here](https://docs.google.com/document/d/1qEMoDyi7or-7ODEfVtnaFlcFSb0NsE2vnvEw6MIJ3Zo/edit):

src/ripple/rpc/handlers/LedgerEntry.cpp
PeterChen13579: thanks, did a full test and there were other fields that needed to be fixed too

src/test/rpc/LedgerRPC_test.cpp
connorjchen: Don't have access to that doc, but thanks for elaboration. If you don't mind sharing it, my email is connorjchen2784@gmail.com

Is `> 1u` in non-test file equivalent to `>= 2u` in test file? These seem to always be equivalent expressions since it is impossible to have decimal (1.5) here, right?

PeterChen13579: yea > 1u in non-test file is equivalent to >= 2u in test file.

src/ripple/rpc/handlers/LedgerEntry.cpp
```diff
@@ -51,7 +51,15 @@ doLedgerEntry(RPC::JsonContext& context)
 
     if (context.params.isMember(jss::index))
     {
-        if (!uNodeIndex.parseHex(context.params[jss::index].asString()))
+        // Document states "index" field is a string. check for this apiVersion
```
arihantkothari, line 54: What document are you referring to ? xrpl.org ? You can make this clear in the comments and remove ambiguity.

arihantkothari (approved):
LGTM, but comments can be improved.

godexsoft (approved):
LGTM. Feel free to ignore my minor comments or apply the suggestions - up to you 👍

intelliot:
@godexsoft @arihantkothari to review new changes

arihantkothari (approved):
lgtm 👍

ximinez (changes requested):
You made changes to the cases for `jss::index`, `jss::check`, `jss::directory`, and `jss::nft_page`, but you only added a test case for `jss::check`. As noted below, the change for `nft_page` is unnecessary. That aside, you also need to test `index` and `directory`. However, the `directory` case knows how to handle a json object, so you also need to test with something it can't handle: arrays.

ximinez (changes requested):
After posting my previous review, I dug a little further, and ended up going down a rabbit hole.

The original issue that this PR fixes is #4550. The basic problem is that the request is returning an "Internal error" for some bad inputs. The goal is to change that to return "invalidParams".

The question is why is "Internal error" being returned in the first place. Long story short, the `Json::Value::asString()` function `throws` a `Json::error` if the value is actually a Json object or array.

So I went through `doLedgerEntry` and looked for _every_ unchecked call to `asString`. There were actually about a dozen more that were not handled in the original PR. I also expanded the unit tests to check _all_ of those cases, including verifying that the checked calls to `asString` returned something other than "Internal error". That commit is here. https://github.com/ximinez/rippled/commit/4c9bb17133745d1169408fa409f8de02bd985e63 (Feel free to cherry-pick that onto your branch.)

But _then_ I looked over the [whole change set](https://github.com/XRPLF/rippled/compare/develop...ximinez:rippled:4c9bb17133745d1169408fa409f8de02bd985e63) and was really unhappy with all of these piecemeal checks.

There's a much cleaner way to do it, which eliminates the worry of missing any cases and is future-proof (in case later lookups neglect some of those checks). Just catch the thrown `Json::error` once. Here's that commit: https://github.com/XRPLF/rippled/commit/dbe7c5c010bef2366e67a61e1f52dff0b75f15fe (Feel free to cherry-pick that one, too, but you'll have to grab the first one to merge cleanly. Or just merge the branch.)

It's much prettier if you hide whitespace changes.

PeterChen13579:
@ximinez Thank you for all the insight! 😄. I pretty much cherry-picked the try-catch and unittest. I agree it's a LOT cleaner this way and prevents future errors to occur.

ximinez (approved):
Looks good

intelliot:
@godexsoft @arihantkothari - this PR has changed since the time you reviewed it. Could you have a look at the latest?

arihantkothari (approved):

godexsoft:
The diff generated by github for this PR is horrible, very hard to understand what the actual changes are.

My 2c:
- It looks like the change does simplify the code a bit - nice
- However, relying on exceptions to handle logic is not a very good practice (this is verifying **user** input)
- Although this PR does not introduce this behaviour - it was already there in one way or another. Sometimes exceptions are a necessary evil. 

Since @ximinez is happy with the change I have no objections.

godexsoft (approved):

ximinez:
> The diff generated by github for this PR is horrible, very hard to understand what the actual changes are.

Click the gear at the top of the diff, and check "Hide whitespace", then click "Apply and reload". Makes a huge difference on diffs like this.

> * However, relying on exceptions to handle logic is not a very good practice (this is verifying **user** input)

Sometimes it's a _great_ practice, especially when you're dealing with user input, because you never know what kind of wacky input a user is going to send you, and trying to guess everything a user can do wrong can make the code a lot more complex and hard to read. Compare that to a `catch` that handles any unexpected behavior...

Besides, like you said, we already have exception handling - that's what returns the "Internal error" in the current version. This change just puts a prettier error on it.

godexsoft:
@ximinez Thanks for the nifty trick to make the diff more readable!

In my opinion, user input is not a good place for using exceptions but as i said, they are *sometimes* a necessary evil, especially if your codebase is already relying on them. 
A better solution in my opinion would be a wrapper that hides all the verification and conversions away and returns either values or errors thru `std::expected` (or `ripple::Expected` in case of rippled atm.). 
I totally understand that one can't simply change the way things are working so I have no objections in this case, just stating my opinion. 

My point is - the way to simplify code is thru abstractions that can be tested separately and hide irrelevant details. 
Of course i can't disagree with the fact that catching one specific exception is better than "pokemon exception handling" :)

intelliot:
note: updates `rippled` (with api_version 2) to behave as `Clio` already does

PR #4634, opened 2023-07-25
Match unit tests on start of test name, and split some test classes

ximinez:

Changes the matching of the parameters to the `--unittest` command line option to match the patterns against the beginning of the test name.

* For example, without this change, to run the TxQ tests, one must specify `--unittest=TxQ1,TxQ2` on the command line. With this change, one can use `--unittest=TxQ`, and both will be run.
* An exact match will prevent any further partial matching.
* This could have some side effects for different tests with a common start of name. For example, `NFToken`, `NFTokenBurn`, `NFTokenDir`. This might be useful. If not, the shorter-named test(s) can be renamed. For example, `NFToken` to `NFTokenBase`.

Speaking of `NFToken`, the second commit splits it, along with the `NFTokenBurn`, and `Offer` test classes into 4 or 5 sub-classes each, based on the different `FeatureBitset`s that are currently run sequentially. In Windows, this speeds up a Debug build unit test run by about 5x! (Details below.)

There are two commits. The don't necessarily need to be squashed when merging, though it would be fine if they are.

### Context of Change

Long test run times have been a thorn in all of our sides for a while, but are even worse in Windows Debug builds, because there is a ton of overhead. At the recent team summit, someone threw out the idea of adding regex parsing to the matching. This doesn't do full regex matching, but that keeps it pretty simple.

Note that these changes only affect unit tests.

### Type of Change

### Before:

These examples are with the tip of develop, commit aded4a7a92 as of this writing
#### `--unittest=TxQ` (fails)
```
$ rippled --unittest=TxQ
Failed: No tests run
119ms, 0 suites, 0 cases, 0 tests total, 1 failure
```

#### `--unittest=TxQ1, TxQ2` (works)
```
$ rippled --unittest=TxQ1,TxQ2
ripple.app.TxQ1 queue sequence
[...]
ripple.app.TxQ2 acct in queue but empty
[...]
Longest suite times:
   24.4s ripple.app.TxQ2
   21.4s ripple.app.TxQ1
45.8s, 2 suites, 37 cases, 4787 tests total, 0 failures
```

#### Full test run with multiple runners `--unittest --unittest-jobs=16` (2228.9 seconds, bound by the longest running test)
```
$ rippled --unittest --unittest-jobs=16
[...]
Longest suite times:
 2227.4s ripple.tx.NFToken
 1192.5s ripple.tx.NFTokenBurn
  288.9s ripple.tx.Offer
  259.0s ripple.app.TheoreticalQuality
  224.8s ripple.app.ShardArchiveHandler
  224.1s ripple.tx.NFTokenDir
  199.7s ripple.tx.ReducedOffer
  196.5s ripple.app.AMMExtended
  166.9s ripple.app.AMM
  163.4s ripple.app.ReportingETL
2228.9s, 211 suites, 1792 cases, 633774 tests total, 0 failures
```

#### `--unittest=TxQ` (succeeds!)
```
$ rippled --unittest=TxQ
ripple.app.TxQ1 queue sequence
[...]
ripple.app.TxQ2 acct in queue but empty
[...]
Longest suite times:
   25.2s ripple.app.TxQ2
   22.1s ripple.app.TxQ1
47.4s, 2 suites, 37 cases, 4787 tests total, 0 failures
```

#### `--unittest=TxQ1, TxQ2` (results unchanged)
```
$  ./build/cmake/msvc19.ON//Debug/rippled --unittest=TxQ1,TxQ2
ripple.app.TxQ1 queue sequence
[...]
ripple.app.TxQ2 acct in queue but empty
[...]
Longest suite times:
   25.4s ripple.app.TxQ2
   22.1s ripple.app.TxQ1
47.6s, 2 suites, 37 cases, 4787 tests total, 0 failures
```

#### Full test run with multiple runners `--unittest --unittest-jobs=16` (438.7 seconds, *not* bound by the longest running test)
```
$ rippled --unittest --unittest-jobs=16
[...]
Longest suite times:
  314.6s ripple.app.TheoreticalQuality
  293.4s ripple.tx.NFToken0
  281.7s ripple.tx.NFToken4
  279.3s ripple.tx.NFToken1
  273.5s ripple.tx.NFToken2
  271.7s ripple.tx.NFToken3
  261.2s ripple.app.ShardArchiveHandler
  220.9s ripple.tx.NFTokenDir
  216.4s ripple.tx.ReducedOffer
  215.8s ripple.tx.NFTokenBurn4
438.7s, 222 suites, 1806 cases, 633788 tests total, 0 failures
```

The fact that the overall test run takes longer than the longest suite (`TheoreticalQuality` in this case) means that that longest running test suite finished before all the other tests were finished, which allowed the child to be reused to run more tests. I haven't seen that happen in on a full run in any configuration in _years_.

## Test Plan

Because only unit test code has changed, as long as exiting tests pass as they normally do, there's no additional functional testing needed.

## Future Tasks

Full regex parsing?


src/test/app/NFTokenBurn_test.cpp
```diff
@@ -786,22 +786,68 @@ class NFTokenBurn_test : public beast::unit_test::suite
...
-        testWithFeats(
-            all - fixNonFungibleTokensV1_2 - fixNFTDir - fixNFTokenRemint);
-        testWithFeats(all - fixNonFungibleTokensV1_2 - fixNFTokenRemint);
-        testWithFeats(all - fixNFTokenRemint);
-        testWithFeats(all);
+        run(0);
+    }
+};
+
+class NFTokenBurn1_test : public NFTokenBurn0_test
```
shawnxie999, line 818: why is it designed to have subsequent test classes inheriting from `NFTokenBurn0_test`?

Why didn't we create a base class like `NFTokenBurnBase_Test`, which contains all the test cases and the new function `run(std::uint32_t instance, bool last = false)`.  And then have `NFTokenBurn0_test`, `NFTokenBurn1_test`, `NFTokenBurn2_test` and `NFTokenBurn3_test` all inherit from it?

ximinez: It could be done either way. The main reasons I chose to do it this way were
1. That's how I did it in `TxQ_test.cpp`
2. So that the base test class would be useful as a test in and of itself.

If you have a strong opinion, I can change it, but I don't think it really matters.

shawnxie999 (approved):

intelliot:
I guess `ximinez:unittest-prefix` has a branch protection rule requiring linear history? Anyway, GitHub can't automatically update this branch. 

@ximinez , at your convenience, feel free to bring this up-to-date with `develop`. Also, if you'd like the commit message to differ from what you've already used in the commits here, feel free to provide a suggested commit message as well.

intelliot:
@manojsdoshi - are we ok to merge this to `develop` now?

PR #4635, opened 2023-07-27
docs(overlay): add URL of blog post and clarify wording

ckeshava:
There appears to be a mistake in the wording of the Overlay README file. I believe the private key is used to sign the fingerprint and the public key is used to verify the authenticity of the signature. This is a change to that effect.

### Type of Change

- [x] Documentation Updates
- [ ] Release


src/ripple/overlay/README.md
```diff
@@ -344,7 +344,7 @@ messages for the local and remote endpoints, and combine them to generate a uniq
 endpoints.
 
 That fingerprint, which is never shared over the wire (since each endpoint will
-calculate it independently), is then signed by each server using its public
```
intelliot, line 347: I believe this refers to the fact that the server's identity is public. It is not a reference to a public key.

I can see how the wording is confusing, however. Let me see if I can propose a larger rewrite of the paragraph.

ckeshava:
I'd like to ask a question, since we are looking into the `Overlay` corner of the codebase.

Why did we choose the `xor` operation to compare the local and remote data (data received from the peer)? https://github.com/XRPLF/rippled/blob/aded4a7a925a5822522e6bbbb4adfd3fec39257b/src/ripple/overlay/impl/Handshake.cpp#L164

In order to prevent Man-in-the-middle attacks, we are including a "Session-Signature" field in the response. (https://github.com/XRPLF/rippled/blob/aded4a7a925a5822522e6bbbb4adfd3fec39257b/src/ripple/overlay/README.md?plain=1#L346) I don't think we are doing this in a robust way.

Isn't it likely that the local and remote data is identical? (For ex: sharing identical sets of validations, proposals, etc) Isn't it better to concatenate the two sets of data and then sign them with `SecretKey::signDigest()` ? @nbougalis @gregtatcam

intelliot:
Suggestion:

> That fingerprint is calculated by each endpoint independently, so the fingerprint is never transmitted over the network. Each server then utilizes its private key to sign the fingerprint. This is the same keypair that determines the server's public `secp256k1` node identity. The signature is transferred over the secure SSL/TLS encrypted link during the protocol's initial handshake phase.

src/ripple/overlay/README.md
intelliot: Suggestion:

> That fingerprint is calculated by each endpoint independently, so the fingerprint is never transmitted over the network. Each server then utilizes its private key to sign the fingerprint. This is the same keypair that determines the server's public `secp256k1` node identity. The signature is transferred over the secure SSL/TLS encrypted link during the protocol's initial handshake phase.

ckeshava: thanks!

ckeshava: Do you know if I can incorporate your suggestion automatically as a new commit? I can copy the text, but I vaguely remember that there is a way to do it much more easily via the github UI

intelliot: There is a way to make apply-able suggestions in the GitHub UI, but it doesn't work for modifying deleted lines. I went ahead and added my suggestion as a new commit. 👍

src/ripple/overlay/ReduceRelayCommon.h
```diff
@@ -24,6 +24,9 @@
 
 namespace ripple {
 
+// Blog post explaining the rationale behind reduction of flooding gossip
+// protocol: https://xrpl.org/blog/2021/message-routing-optimizations-pt-1-proposal-validation-relaying.html
```
intelliot, line 28: clang-format
```suggestion
// protocol:
// https://xrpl.org/blog/2021/message-routing-optimizations-pt-1-proposal-validation-relaying.html
```

ckeshava:
ok 👍

scottschurr (approved):

gregtatcam (approved):
👍 LGTM

intelliot:
@manojsdoshi I think this could be ok to consider for 1.12.0-rc2 because it is only docs/comments, and does not touch any code

PR #4636, opened 2023-07-31
`DID`: Decentralized identifiers (DIDs) (XLS-40):

mvadari:

Introduces:
* New ledger object: `DID`
* Two new transactions:
  * `DIDSet`, which creates and updates the `DID` object
  * `DIDDelete`, which deletes the `DID` object

### Context of Change

- [XLS-40d specification](https://github.com/XRPLF/XRPL-Standards/tree/master/XLS-0040d-decentralized-identity#readme)

Earlier discussions/versions:
- https://github.com/XRPLF/XRPL-Standards/pull/136
- [Spec discussion](https://github.com/XRPLF/XRPL-Standards/discussions/100)

### Type of Change

- [ ] Documentation Updates
- [ ] Release

## Test Plan
Added tests for the new features.
```
Longest suite times:
   68.2s ripple.app.AMM
   58.4s ripple.app.TheoreticalQuality
   51.6s ripple.app.XChain
   47.9s ripple.tx.NFToken4
   46.2s ripple.tx.NFToken3
   45.9s ripple.tx.NFToken1
   45.7s ripple.tx.NFToken2
   43.5s ripple.tx.NFToken0
   41.8s ripple.tx.NFTokenDir
   39.3s ripple.app.AMMExtended
1119.3s, 227 suites, 1840 cases, 672141 tests total, 0 failures
```
```
 % ./rippled --unittest=ripple.app.DID                               
ripple.app.DID Enabled
ripple.app.DID Account reserve
ripple.app.DID Invalid Set
ripple.app.DID Invalid Delete
ripple.app.DID Modify DID with Set
Longest suite times:
    1.0s ripple.app.DID
1.1s, 1 suite, 5 cases, 83 tests total, 0 failures
```


src/ripple/app/tx/impl/DID.cpp
```diff
@@ -0,0 +1,215 @@
...
+DIDDelete::doApply()
+{
+    AccountID const account = ctx_.tx[sfAccount];
+    auto const didKeylet = keylet::did(account);
+
+    if (auto const ret = deleteSLE(ctx_, didKeylet, account_);
+        !isTesSuccess(ret))
+        return ret;
+
+    return tesSUCCESS;
```
nbougalis, line 212: Same comment as before: why a separate function instead of inlining the code?

src/ripple/app/tx/impl/DID.cpp
```diff
@@ -0,0 +1,215 @@
...
+            if (data->empty())
+            {
+                sleDID->makeFieldAbsent(sfData);
+            }
+            else
+            {
+                (*sleDID)[sfData] = *data;
+            }
+        }
+        if (!sleDID->isFieldPresent(sfURI) && !sleDID->isFieldPresent(sfData))
```
nbougalis, line 129: This bit could be done in `preclaim`; not sure it makes sense to do it, just pointing it out.

src/ripple/protocol/TER.h
```diff
@@ -298,7 +299,8 @@ enum TECcodes : TERUnderlyingType {
     tecUNFUNDED_AMM = 162,
     tecAMM_BALANCE = 163,
     tecAMM_FAILED = 164,
-    tecAMM_INVALID_TOKENS = 165
+    tecAMM_INVALID_TOKENS = 165,
+    tecEMPTY_DID = 166
```
nbougalis, line 303: This conflicts with PR #4626: see [lines 302-305 of TER.h](https://github.com/XRPLF/rippled/pull/4626/files#diff-570760ccc79d416a7e98bf3e851c2e298ea5cf90f7207018e06aa6b46d09a013R302-R305):

```
    tecAMM_EMPTY = 166,
    tecAMM_NOT_EMPTY = 167,
    tecAMM_ACCOUNT = 168,
    tecINCOMPLETE = 169
```

src/ripple/protocol/impl/SField.cpp
```diff
@@ -281,6 +281,7 @@ CONSTRUCT_TYPED_SFIELD(sfCreateCode,            "CreateCode",           VL,
 CONSTRUCT_TYPED_SFIELD(sfMemoType,              "MemoType",             VL,        12);
 CONSTRUCT_TYPED_SFIELD(sfMemoData,              "MemoData",             VL,        13);
 CONSTRUCT_TYPED_SFIELD(sfMemoFormat,            "MemoFormat",           VL,        14);
+CONSTRUCT_TYPED_SFIELD(sfData,                  "Data",                 VL,        15);
```
nbougalis, line 284: You may want to consider using a non-common code here? On the other hand `Data` is generic enough that it may be useful elsewhere, so... maybe it's fine.

nbougalis (changes requested):
Left some comments. Biggest issue, IMO, is the lack of any invariant checks. You should consider what types of checks make sense to code up.

RichardAH:
Is the DID object just a first class NFT that's set and unset by the user themselves on their own account? Am I missing something? Should someone else be able to modify this?

mvadari:
> Is the DID object just a first class NFT that's set and unset by the user themselves on their own account? Am I missing something? Should someone else be able to modify this?

I think you might be referring to an old version of the spec - it was updated (and changed pretty heavily) about a month ago.

src/ripple/app/tx/impl/DID.cpp
mvadari: It seems overly complicated to add that to `preclaim`.

src/ripple/protocol/impl/LedgerFormats.cpp
```diff
@@ -271,13 +271,25 @@ LedgerFormats::LedgerFormats()
...
+
+    add(jss::DID,
+        ltDID,
+        {
+            {sfAccount,              soeREQUIRED},
+            {sfURI,                  soeOPTIONAL},
+            {sfData,                 soeOPTIONAL},
+            {sfOwnerNode,            soeREQUIRED},
+            {sfPreviousTxnID,        soeREQUIRED},
+            {sfPreviousTxnLgrSeq,    soeREQUIRED}
```
cjcobb23, line 292: I feel like `DID` is a weird name for this object. This is just a ledger object that holds two arbitrary bits of data, in the `URI` field and the `Data` field. Why not call this something more generic? I could see this being used for things other than DID.

src/ripple/protocol/impl/TxFormats.cpp
```diff
@@ -394,6 +394,22 @@ TxFormats::TxFormats()
...
+            {sfTicketSequence, soeOPTIONAL},
+        },
+        commonFields);
+
+    add(jss::DIDDelete,
+        ttDID_DELETE,
+        {
+            {sfTicketSequence, soeOPTIONAL},
+        },
+        commonFields);
```
cjcobb23, line 412: Similar to my other comment, why not call these something more generic instead of `DIDSet` and `DIDDelete`?

src/ripple/protocol/impl/LedgerFormats.cpp
mvadari: You can only own one of them, so I don't really know what else you'd use them for.

cjcobb23: This is just arbitrary data storage, I could really use this for anything at all. I could use it for on oracle, a bridge, governance, etc.

src/ripple/app/tx/impl/DID.cpp
```diff
@@ -0,0 +1,215 @@
...
+        if (auto const uri = ctx_.tx[~sfURI])
+        {
+            if (uri->empty())
+            {
+                sleDID->makeFieldAbsent(sfURI);
+            }
+            else
+            {
+                (*sleDID)[sfURI] = *uri;
+            }
```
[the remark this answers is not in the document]
nbougalis: You're right, given what you're trying to do my proposed construct won't work. It's fine as it, but I wouldn't have personally structured this as you have; I'd have included two flags that, if set, indicate the type of value you're setting/clearing. So, if `sfFlags & flagURI` was set, then the `URI` field is replaced with the `URI` field from the transaction.

src/ripple/app/tx/impl/DID.cpp
nbougalis: Agreed. And any performance gains are likely to be overshadowed by the extra work to load/manipulate the objects.

src/ripple/app/tx/impl/DID.cpp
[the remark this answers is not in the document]
mvadari: This was the method that @seelabs and @aanchal4 thought was most user-intuitive.

src/ripple/protocol/impl/LedgerFormats.cpp
seelabs: That's an interesting point @cjcobb23 

We could make this more generic - add a read-only "kind" field that's part of the hash and let people store whatever data they want.

So a DID would be stored in this new generic data object and have a "kind" of "DID", oracles would have a "kind" of "oracle-usd-xrp" or somesuch, and since the "kind" is part of the hash they are distinct objects. Although with this design you can only ever have one object per "kind", but that doesn't seem too bad.

We have talked about general purpose data storage in the past, now that we have a concrete use-case for such an object (DIDs) it seems like it's worth looking at doing the generic solution. We'd have to flesh it out a little more if we want to go in this direction, but it seems like a good idea.

I'd probably rename the fields, maybe add a couple more fields, and a read-only "kind" and everything else works as-is now. It's an easy change from an implementation POV.

One problem with the generic solution is getting everyone to agree on what "kind" means. I might want ids to be "IDS", you might want "DID", and someone else whatever the french word for "ID" is. But that doesn't seem like a killer.

gregtatcam: >    Although with this design you can only ever have one object per "kind", but that doesn't seem too bad.

@seelabs there could be multiple oracles even for the same type. I.e. could have a pricing oracle for XRP/USD by chainlink, DIA, etc. Also could have weather, sports, etc. oracles say from the same Oracle Provider.

seelabs: @gregtatcam If the oracles are from different providers they can be on different accounts (the hash would include the "kind"and the account). Weather, sports, ect oracles from the same providers could be different "kinds".

src/ripple/protocol/impl/TxFormats.cpp
cjcobb23: Something related to whatever we name the object. If we name the object `DataBlob` (not saying that's a good name), we could call these `DataBlobSet` and `DataBlobDelete`

src/ripple/protocol/impl/LedgerFormats.cpp
cjcobb23: Could also add a "description" field as well, that's just some plain text description of the data.

One thing I think worth thinking about is actually storing the schema in the object. The schema could be defined in protobuf. This could enable things like allowing callers to pass in protobuf in the update transaction, safely deserializing the protobuf to other types and returning JSON to the RPC callers by just converting from protobuf to JSON. Obviously this wouldn't work for very large schemas, but for small schemas, say with just a few fields, we could probably store them in the object and get a lot of benefits.

src/ripple/protocol/impl/TxFormats.cpp
dangell7: I like the `Collection/Document` nomenclature. I started in NOSQL and always liked those names. If the account is the collection then these could be "Documents".

I agree that DID is more the type of `Data/Document` that is being stored and we should aim for more agnostic approach with the name of the transactor.

dangell7: I was working on my own amendment to do Collections/Documents and one thing I was doing was adding a log functionally.

So when you updated the document, you create a replica and save that index onto the current object. In my code I use `oldDocumentRef` and `newDocumentRef`. This allows me to traverse the logs. 

Technically something like this might be possible from the transaction meta, but I was going to save directly in the object. It was just an experiment for fun so I'm not saying its the right way. If the meta exists then thats probably how it should be done.

I can share the full function but it does involve saving an extra 2 documents on each update.

Its probably outside the scope of this amendment as well, just sharing some information.

mDuo13: +1 for `Document`

src/ripple/app/tx/impl/DID.cpp
```diff
@@ -0,0 +1,215 @@
...
+
+    if (auto const ret = preflight1(ctx); !isTesSuccess(ret))
+        return ret;
+
+    if (!ctx.tx.isFieldPresent(sfURI) && !ctx.tx.isFieldPresent(sfData))
+        return temEMPTY_DID;
+
+    if (ctx.tx.isFieldPresent(sfURI) && ctx.tx[sfURI].empty() &&
+        ctx.tx.isFieldPresent(sfData) && ctx.tx[sfData].empty())
+        return temEMPTY_DID;
```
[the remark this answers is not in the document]
mvadari: Yes, I missed that - the spec says a max of 256 bytes per field.

src/ripple/app/tx/impl/DID.cpp
```diff
@@ -0,0 +1,215 @@
...
+    (*sleDID)[sfAccount] = account_;
+    if (auto const uri = ctx_.tx[~sfURI]; uri.has_value() && !uri->empty())
+        (*sleDID)[sfURI] = uri.value();
+    if (auto const data = ctx_.tx[~sfData]; data.has_value() && !data->empty())
+        (*sleDID)[sfData] = data.value();
+
+    if (auto const ret = addSLE(ctx_, sleDID, account_); !isTesSuccess(ret))
+        return ret;
+
+    return tesSUCCESS;
```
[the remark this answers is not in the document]
mvadari: The idea behind `addSLE` was that it could eventually become a helper function that all transactors can use - I noticed a lot of the transactors had the same code written, and this would make them (or at least future transactors) a bit easier to read.

src/ripple/app/tx/impl/DID.cpp
mvadari: Ditto here

src/ripple/protocol/TER.h
mvadari: I'll fix that after I rebase when that PR is merged - I don't think it's a super urgent fix.

src/ripple/app/tx/impl/DID.cpp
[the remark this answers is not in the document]
mvadari: Fixed in https://github.com/XRPLF/rippled/pull/4636/commits/99b53beb144c66ed619563b94819f6a6deca8ca0

src/ripple/app/tx/impl/DID.cpp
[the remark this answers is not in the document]
mvadari: Simplified in https://github.com/XRPLF/rippled/pull/4636/commits/99b53beb144c66ed619563b94819f6a6deca8ca0

src/ripple/protocol/impl/LedgerFormats.cpp
ckeshava: @mDuo13 What are the hazards of a large ledger entry object? I'm not familiar with the issues of trust line creation.

The owner reserve fees disincentives spamming an account with frivolous DIDs right?

src/ripple/app/tx/impl/DID.cpp
```diff
@@ -0,0 +1,218 @@
...
+    if (auto const uri = ctx_.tx[~sfURI]; uri.has_value() && !uri->empty())
+        (*sleDID)[sfURI] = uri.value();
+    if (auto const data = ctx_.tx[~sfData]; data.has_value() && !data->empty())
+        (*sleDID)[sfData] = data.value();
+
+    return addSLE(ctx_, sleDID, account_);
+}
+
+NotTEC
+DIDDelete::preflight(PreflightContext const& ctx)
```
ckeshava, line 161: The spec document says:
```
Both URI and Data fields MUST be absent to delete the DID object.
```

Is this condition enforced anywhere implicitly?

src/ripple/protocol/SField.h
```diff
@@ -515,6 +515,7 @@ extern SF_VL const sfCreateCode;
 extern SF_VL const sfMemoType;
 extern SF_VL const sfMemoData;
 extern SF_VL const sfMemoFormat;
+extern SF_VL const sfData;
```
ckeshava, line 518: Will we get any performance benefits if we use a custom `SF_256Byte` field instead of `SF_VL` (variable length) here?

I'm thinking of something like:
```
using SF_256Byte = TypedField<STBitString<256*8>>
```

src/ripple/protocol/impl/TER.cpp
```diff
@@ -92,6 +92,7 @@ transResults()
         MAKE_ERROR(tecINSUFFICIENT_FUNDS,            "Not enough funds available to complete requested transaction."),
         MAKE_ERROR(tecOBJECT_NOT_FOUND,              "A requested object could not be located."),
         MAKE_ERROR(tecINSUFFICIENT_PAYMENT,          "The payment is not sufficient."),
+        MAKE_ERROR(tecEMPTY_DID,                     "The DID object cannot be made to be empty."),
```
ckeshava, line 95: Suggestion: I think it would be useful to say "Atleast URI or Data field must be non-empty in a DID object"

src/test/rpc/LedgerRPC_test.cpp
```diff
@@ -1212,6 +1212,56 @@ class LedgerRPC_test : public beast::unit_test::suite
...
+        env.close();
+
+        std::string const ledgerHash{to_string(env.closed()->info().hash)};
+
+        {
+            // Request the payment channel using its index.
+            Json::Value jvParams;
+            jvParams[jss::did] = alice.human();
+            jvParams[jss::ledger_hash] = ledgerHash;
+            Json::Value const jrr = env.rpc(
```
ckeshava, line 1247: Are we using the DID format `did:xrpl:0:...` to perform any CRUD operation? It appears that we use Account(`alice` in this piece of code) and Type (`DID`).

The implementation does adhere to the spec. Is the DIDID meant to be a universal identifier without any usages inside ledger itself?

src/ripple/protocol/SField.h
mvadari: This is such a simple amendment, I don't think any performance benefits from using `SF_UINT256` would matter.  Also, such a data type would mean that the length of the data type _had_ to be 256 bytes (instead of possibly being less), which sure you could pad with 0s, but that's a little less nice of a DX.

src/test/jtx/DID.h
```diff
@@ -0,0 +1,86 @@
...
+namespace did {
+
+Json::Value
+set(jtx::Account const& account);
+
+Json::Value
+setValid(jtx::Account const& account);
+
+/** Sets the optional URI on a DIDSet. */
+class uri
```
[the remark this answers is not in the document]
mvadari: That's not true in the JTX tests.

src/test/rpc/LedgerRPC_test.cpp
mvadari: No, that format is used in the universal resolver, not in `ledger_entry`. It's unnecessary in `ledger_entry`.

src/ripple/protocol/SField.h
ckeshava: hmm okay, got your point. what is DX?

src/ripple/app/tx/impl/DID.cpp
ckeshava: ok, that makes sense

src/test/jtx/DID.h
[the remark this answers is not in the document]
ckeshava: hmm, i see

src/ripple/protocol/SField.h
mvadari: Developer experience

intelliot:
How should this PR be moved forward? Should it be considered for the next release?

ckeshava (approved):

src/ripple/protocol/impl/LedgerFormats.cpp
mvadari: https://github.com/XRPLF/XRPL-Standards/discussions/132

src/ripple/protocol/impl/TxFormats.cpp
mvadari: https://github.com/XRPLF/XRPL-Standards/discussions/132

mvadari:
> How should this PR be moved forward? Should it be considered for the next release?

@intelliot the spec isn't finalized yet and there isn't much time between the final spec release and the 2.0 release. I'd like it to go into 2.0, but it may get pushed to the next release.

src/ripple/app/tx/impl/DID.cpp
```diff
@@ -0,0 +1,259 @@
...
+            return tecEMPTY_DID;
+        }
+        ctx_.view().update(sleDID);
+        return tesSUCCESS;
+    }
+
+    // Create new ledger object otherwise
+    auto const sleDID = std::make_shared<SLE>(didKeylet);
+    (*sleDID)[sfAccount] = account_;
+    if (auto const uri = ctx_.tx[~sfURI]; uri.has_value() && !uri->empty())
```
gregtatcam, line 189: Consider adding lambda:
>       auto set = [&](auto const& sField) {
>        if (auto const field = ctx_.tx[~sField];
>           field.has_value() && !field->empty())
>            (*sleDID)[sField] = field.value();
>      };
>      set(sfURI);
>      set(sfDIDDocument);
>      set(sfAttestation);

src/ripple/app/tx/impl/DID.cpp
```diff
@@ -0,0 +1,259 @@
...
+    view.update(sleOwner);
+
+    // Remove object from ledger
+    view.erase(sle);
+    return tesSUCCESS;
+}
+
+TER
+DIDDelete::doApply()
+{
```
gregtatcam, line 252: `account_` is equal to `ctx_.tx[sfAccount]`. Consider this:
>     return deleteSLE(ctx_, keylet::did(account_), account_);

src/ripple/app/tx/impl/DID.cpp
```diff
@@ -0,0 +1,259 @@
...
+    // Remove object from owner directory
+    if (!view.dirRemove(
+            keylet::ownerDir(owner), (*sle)[sfOwnerNode], sle->key(), true))
+    {
+        JLOG(j.fatal()) << "Unable to delete DID Token from owner.";
+        return tefBAD_LEDGER;
+    }
+
+    auto const sleOwner = view.peek(keylet::account(owner));
```
gregtatcam, line 241: Should check `sleOwner` and return `tecINTERNAL` if doesn't exist.

src/ripple/app/tx/impl/DID.cpp
```diff
@@ -0,0 +1,259 @@
...
+    return preflight2(ctx);
+}
+
+TER
+DIDSet::preclaim(PreclaimContext const& ctx)
+{
+    AccountID const account = ctx.tx[sfAccount];
+    Keylet const didKeylet = keylet::did(account);
+
+    if (!ctx.view.read(didKeylet))
```
gregtatcam, line 89: Consider simplifying. Don't really need these variables.
>         if (!ctx.view.read(keylet::did(ctx.tx[sfAccount])) &&
>            !ctx.tx.isFieldPresent(sfURI) && !ctx.tx.isFieldPresent(sfDIDDocument))
>            return tecEMPTY_DID;
>
>         return tesSUCCESS;

src/ripple/app/tx/impl/DID.cpp
```diff
@@ -0,0 +1,259 @@
...
+
+    if (!ctx.tx.isFieldPresent(sfURI) &&
+        !ctx.tx.isFieldPresent(sfDIDDocument) &&
+        !ctx.tx.isFieldPresent(sfAttestation))
+        return temEMPTY_DID;
+
+    if (ctx.tx.isFieldPresent(sfURI) && ctx.tx[sfURI].empty() &&
+        ctx.tx.isFieldPresent(sfDIDDocument) && ctx.tx[sfDIDDocument].empty())
+        return temEMPTY_DID;
+
```
gregtatcam, line 61: Consider this:
>     auto malformed = [&](auto const& sField, std::size_t length) -> bool {
>         if (auto field = ctx.tx[~sField])
>             return field->length() > length;
>         return false;
>     };
>     if (malformed(sfURI, maxDIDURILength) ||
>         malformed(sfDIDDocument, maxDIDDocumentLength) ||
>         malformed(sfAttestation, maxDIDAttestationLength))
>         return temMALFORMED;

gregtatcam (changes requested):

src/ripple/protocol/SField.h
```diff
@@ -524,6 +524,8 @@ extern SF_VL const sfCreateCode;
 extern SF_VL const sfMemoType;
 extern SF_VL const sfMemoData;
 extern SF_VL const sfMemoFormat;
+extern SF_VL const sfDIDDocument;
+extern SF_VL const sfAttestation;
```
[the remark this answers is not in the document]
mvadari: It's coming in a spec update that hasn't been added yet (but should happen early next week).

src/test/app/AccountDelete_test.cpp
```diff
@@ -148,13 +148,15 @@ class AccountDelete_test : public beast::unit_test::suite
...
         // Give carol a deposit preauthorization, an offer, a ticket,
-        // and a signer list.  Even with all that she's still deletable.
+        // a signer list, and a DID.  Even with all that she's still deletable.
         env(deposit::auth(carol, becky));
         std::uint32_t const carolOfferSeq{env.seq(carol)};
         env(offer(carol, gw["USD"](51), XRP(51)));
         std::uint32_t const carolTicketSeq{env.seq(carol) + 1};
         env(ticket::create(carol, 1));
         env(signers(carol, 1, {{alice, 1}, {becky, 1}}));
+        env(signers(carol, 1, {{alice, 1}, {becky, 1}}));
```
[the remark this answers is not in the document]
mvadari: Probably a copy-paste error

src/ripple/protocol/TER.h
mvadari: This has been resolved (I don't remember which commit)

src/ripple/app/tx/impl/DID.cpp
mvadari: Done in https://github.com/XRPLF/rippled/pull/4636/commits/298077f443ad9e0a0a8ee3045f2dafd0bd2a590b

src/ripple/app/tx/impl/DID.cpp
mvadari: Done in https://github.com/XRPLF/rippled/pull/4636/commits/298077f443ad9e0a0a8ee3045f2dafd0bd2a590b

src/test/app/AccountDelete_test.cpp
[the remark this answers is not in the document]
mvadari: Fixed in https://github.com/XRPLF/rippled/commit/298077f443ad9e0a0a8ee3045f2dafd0bd2a590b

src/test/rpc/LedgerRPC_test.cpp
```diff
@@ -1530,6 +1530,57 @@ class LedgerRPC_test : public beast::unit_test::suite
...
+            return jv;
+        };
+
+        env(didCreate(alice));
+        env.close();
+
+        std::string const ledgerHash{to_string(env.closed()->info().hash)};
+
+        {
+            // Request the payment channel using its index.
```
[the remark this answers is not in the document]
mvadari: Fixed in https://github.com/XRPLF/rippled/commit/298077f443ad9e0a0a8ee3045f2dafd0bd2a590b

src/test/rpc/LedgerRPC_test.cpp
```diff
@@ -1530,6 +1530,57 @@ class LedgerRPC_test : public beast::unit_test::suite
...
+            Json::Value const jrr = env.rpc(
+                "json", "ledger_entry", to_string(jvParams))[jss::result];
+            BEAST_EXPECT(
+                jrr[jss::node][sfDIDDocument.jsonName] ==
+                strHex(std::string{"data"}));
+            BEAST_EXPECT(
+                jrr[jss::node][sfURI.jsonName] == strHex(std::string{"uri"}));
+        }
+        {
+            // Request an index that is not a payment channel.
```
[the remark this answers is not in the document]
mvadari: Fixed in https://github.com/XRPLF/rippled/commit/298077f443ad9e0a0a8ee3045f2dafd0bd2a590b

src/ripple/app/tx/impl/DID.cpp
```diff
@@ -0,0 +1,259 @@
...
+    return tesSUCCESS;
+}
+
+TER
+DIDSet::doApply()
+{
+    // Edit ledger object if it already exists
+    Keylet const didKeylet = keylet::did(account_);
+    if (auto const sleDID = ctx_.view().peek(didKeylet))
+    {
```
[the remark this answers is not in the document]
mvadari: Done in https://github.com/XRPLF/rippled/commit/298077f443ad9e0a0a8ee3045f2dafd0bd2a590b

src/ripple/app/tx/impl/DID.cpp
mvadari: Done in https://github.com/XRPLF/rippled/pull/4636/commits/04066ae42e9c8d7de59742d984b4202e4d705d54

gregtatcam (approved):
👍 LGTM

intelliot:
@nbougalis - could you check if your requested changes have been satisfactorily addressed?

src/ripple/protocol/SField.h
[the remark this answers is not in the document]
mvadari: @gregtatcam https://github.com/XRPLF/XRPL-Standards/pull/136

src/ripple/app/tx/impl/DID.cpp
```diff
@@ -0,0 +1,234 @@
...
+#include <ripple/protocol/TxFlags.h>
+#include <ripple/protocol/st.h>
+
+namespace ripple {
+
+/*
+    DID
+    ======
+
+    TODO: add docs here
```
seelabs, line 36: Need to add docs before merging.

src/ripple/app/tx/impl/DID.cpp
```diff
@@ -0,0 +1,234 @@
...
+        isTooLong(sfData, maxDIDAttestationLength))
+        return temMALFORMED;
+
+    return preflight2(ctx);
+}
+
+TER
+DIDSet::preclaim(PreclaimContext const& ctx)
+{
+    if (!ctx.view.read(keylet::did(ctx.tx[sfAccount])) &&
```
seelabs, line 78: If you don't use the result of the `read` use `exists` instead so the object isn't deserialized. (FWIW, I do this myself too. It's easy to overlook `exists` and reach for `read`)

src/ripple/app/tx/impl/DID.cpp
```diff
@@ -0,0 +1,234 @@
...
+    }
+
+    // Create new ledger object otherwise
+    auto const sleDID = std::make_shared<SLE>(didKeylet);
+    (*sleDID)[sfAccount] = account_;
+
+    auto set = [&](auto const& sField) {
+        if (auto const field = ctx_.tx[~sField];
+            field.has_value() && !field->empty())
+            (*sleDID)[sField] = field.value();
```
seelabs, line 166: We tend not to use `has_value` and `value` . Consider rewriting this in a similar style to your `update` lambda above. I.e. something like:
```
        if (auto const field = ctx_.tx[~sField]; field && !field->empty())
            (*sleDID)[sField] = *field;
```

seelabs (commented):
Looks good. Left some minor comments.

src/ripple/app/tx/impl/DID.cpp
```diff
@@ -0,0 +1,234 @@
...
+        // Need either the URI or document if the account doesn't already have a
+        // DID
+        return tecEMPTY_DID;
+    }
+
+    return tesSUCCESS;
+}
+
+TER
+addSLE(
```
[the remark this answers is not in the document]
mvadari: I made it a separate function because I'd like to make it a helper function for the future, but I figured that the refactor wasn't worth doing in this PR and could wait for a separate PR.

src/ripple/app/tx/impl/DID.cpp
```diff
@@ -0,0 +1,234 @@
...
+        return temINVALID_FLAG;
+
+    if (auto const ret = preflight1(ctx); !isTesSuccess(ret))
+        return ret;
+
+    return preflight2(ctx);
+}
+
+TER
+DIDDelete::deleteSLE(ApplyContext& ctx, Keylet sleKeylet, AccountID const owner)
```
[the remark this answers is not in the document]
mvadari: I made it a separate function because I'd like to make it a helper function for the future, but I figured that the refactor wasn't worth doing in this PR and could wait for a separate PR. It's also used in `AccountDelete`.

src/ripple/app/tx/impl/DID.cpp
```diff
@@ -0,0 +1,234 @@
...
+}
+
+TER
+DIDSet::doApply()
+{
+    // Edit ledger object if it already exists
+    Keylet const didKeylet = keylet::did(account_);
+    if (auto const sleDID = ctx_.view().peek(didKeylet))
+    {
+        auto update = [&](auto const& sField) {
```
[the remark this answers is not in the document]
mvadari: How do I type a lambda?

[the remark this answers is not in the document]
mvadari: It looks like `auto` is usually recommended for typing a lambda.

src/test/app/DID_test.cpp
```diff
@@ -0,0 +1,415 @@
...
+    s.addRaw(result);
+    return s.getString() == expected;
+}
+
+struct DID_test : public beast::unit_test::suite
+{
+    void
+    testEnabled(FeatureBitset features)
+    {
+        testcase("Enabled");
```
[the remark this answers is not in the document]
mvadari: Fixed in https://github.com/XRPLF/rippled/pull/4636/commits/47059087a2308a0ce1d9032c82419ed2f0992072

src/ripple/app/tx/impl/DID.cpp
mvadari: Fixed in https://github.com/XRPLF/rippled/pull/4636/commits/6bda30b852d448b4b3df2d534b6441ed0af92663

src/ripple/app/tx/impl/DID.cpp
mvadari: Fixed in https://github.com/XRPLF/rippled/pull/4636/commits/6bda30b852d448b4b3df2d534b6441ed0af92663

src/test/app/DID_test.cpp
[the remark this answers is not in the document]
mvadari: Fixed in https://github.com/XRPLF/rippled/pull/4636/commits/f347bae57f9aac9417e082dc83df6f70c5c87af5

src/test/jtx/impl/DID.cpp
```diff
@@ -0,0 +1,67 @@
+//------------------------------------------------------------------------------
```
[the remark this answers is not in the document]
mvadari: Fixed in https://github.com/XRPLF/rippled/pull/4636/commits/47059087a2308a0ce1d9032c82419ed2f0992072

src/ripple/app/tx/impl/DID.cpp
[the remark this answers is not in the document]
gregtatcam: @mvadari Some of those common functions reside in `View.h`. Might make sense to move it there.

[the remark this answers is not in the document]
mvadari: Thanks for the suggestion - moved in https://github.com/XRPLF/rippled/pull/4636/commits/31117f7fccebe37a4faf47d1a840448b8b86e660

src/ripple/app/tx/impl/DID.cpp
mvadari: Added in https://github.com/XRPLF/rippled/pull/4636/commits/4acc54b39e136699c871423871eca4ac405a9052

src/ripple/app/tx/impl/DID.cpp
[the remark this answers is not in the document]
seelabs: I'd be stricter with `auto` on lambdas than on variables. With variables it's always going to be some particular type, with lambdas it's a template, and can be different types. With variables the auto is initialized near where the `auto` occurs, with a lambda you need to look later in the code to see the caller. 

I don't think I agree `auto` is usually recommended for typing a lambda. I think it's easier for the _writer_ of the code to use a lambda - it's shorter and you don't have to think about the concrete type For the _reader_ it's easier if we but the concrete type there. 

We can make arguments for `auto` in lambdas even when it isn't strictly needed - when the type is long, or doesn't matter. Or is obvious (which you can argue in this case). 

I'm OK if you want to keep this as `auto`. I'm weakly in favor of changing it to the concrete type. I would push back on the comment:
```
It looks like auto is usually recommended for typing a lambda.
```
I don't agree with that. And if we are starting to do that we probably need to change it.

src/test/jtx/impl/DID.cpp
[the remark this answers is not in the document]
seelabs: Unfortunately, this doesn't fix things for me. I still get prompted. Honestly, we should probably change this back to `DID.h`. I appreciate the change, but let's change it back (of change it to something that differ by more than just case changes).

src/ripple/app/tx/impl/DID.cpp
```diff
@@ -0,0 +1,165 @@
...
+
+namespace ripple {
+
+/*
+    DID
+    ======
+
+    Decentralized Identifiers (DIDs) are a new type of identifier that enable
+    verifiable, self-sovereign digital identity and are designed to be
+   compatible with any distributed ledger or network. This implementation
```
seelabs, line 38: Nit: formatting

src/ripple/app/tx/impl/DID.cpp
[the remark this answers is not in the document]
mvadari: Reverted in https://github.com/XRPLF/rippled/pull/4636/commits/36716d6417d4e646cf692b6d9fe6fbf2f24e180b

src/ripple/app/tx/impl/DID.cpp
mvadari: Fixed in https://github.com/XRPLF/rippled/pull/4636/commits/c6e6b261001abbc80aaffa8545660893d1242eb9

src/ripple/app/tx/impl/DID.cpp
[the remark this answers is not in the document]
seelabs: That makes sense. I'm fine with it as-is. Thanks for the explanation.

seelabs (approved):
Nice job on this :+1:

gregtatcam (approved):
👍 LGTM

mvadari:
@intelliot this PR is ready to merge.

intelliot:
(note: checking with @nbougalis regarding whether he still has change requests)

sgramkumar:
QA tests passed!

nbougalis (approved):

mvadari:
@intelliot this is good to go

intelliot:
note: @dangell7 will review

src/ripple/protocol/LedgerFormats.h
```diff
@@ -186,6 +186,12 @@ enum LedgerEntryType : std::uint16_t
     */
     ltAMM = 0x0079,
 
+    /** The ledger object which tracks the DID.
+
+       \sa keylet::did
+    */
+    ltDID = 0x0049,
```
dangell7, line 193: nit: This clashes with a LedgerFormat we use in Xahau.

dangell7 (approved):

intelliot:
Suggested commit message:
```
`DID`: Decentralized identifiers (DIDs) (XLS-40): (#4636)

Implement native support for W3C DIDs.

Add a new ledger object: `DID`.

Add two new transactions:
1. `DIDSet`: create or update the `DID` object.
2. `DIDDelete`: delete the `DID` object.

This meets the requirements specified in the DID v1.0 specification
currently recommended by the W3C Credentials Community Group.

The DID format for the XRP Ledger conforms to W3C DID standards. The
objects can be created and owned by any XRPL account holder. The
transactions can be integrated by any service, wallet, or application.
```

intelliot:
note: @sappenin would like the spec to be merged prior to this implementation

- Spec: https://github.com/XRPLF/XRPL-Standards/pull/136

src/ripple/protocol/impl/SField.cpp
mvadari: This was resolved but I forget which commit.

seelabs (approved):
:+1:

gregtatcam (approved):
:+1: LGTM

PR #4637, opened 2023-07-31
refactor: simplify `TxFormats` common fields logic

mvadari:

This PR proposes a minor refactor to `ripple/protocol/TxFormats.cpp`, renaming `commonFields` to `pseudoCommonFields` (since it is the common fields that all pseudo-transactions need) and adding a new static variable, `commonFields`, which represents all the common fields that non-pseudo transactions need (essentially everything that `pseudoCommonFields` contains, plus `sfTicketSequence`).

### Context of Change

This makes it harder to accidentally leave out `sfTicketSequence` in a new transaction.

### Type of Change

- [ ] Documentation Updates
- [ ] Release


src/ripple/protocol/impl/TxFormats.cpp
```diff
@@ -25,6 +25,24 @@ namespace ripple {
 TxFormats::TxFormats()
 {
     // Fields shared by all txFormats:
```
seelabs, line 27: We should probably update this comment and explain what the `common` and `pseudo` vars are doing and how they differ.

src/ripple/protocol/impl/TxFormats.cpp
```diff
@@ -25,6 +25,24 @@ namespace ripple {
...
+        {sfPreviousTxnID, soeOPTIONAL},  // emulate027
+        {sfLastLedgerSequence, soeOPTIONAL},
+        {sfAccountTxnID, soeOPTIONAL},
+        {sfFee, soeREQUIRED},
+        {sfOperationLimit, soeOPTIONAL},
+        {sfMemos, soeOPTIONAL},
+        {sfSigningPubKey, soeREQUIRED},
+        {sfTxnSignature, soeOPTIONAL},
+        {sfSigners, soeOPTIONAL},  // submit_multisigned
+        {sfNetworkID, soeOPTIONAL},
```
seelabs, line 43: Since the only difference between the fields in `pseudo` and `common` is the ticket field, we could consider a macro to factor out the commonality. If you do this make sure you `undef` the macro after it's used.

There are other ways to do this too and some people are allergic to macros (which I understand), but I think it would be OK here. (The other way I had in mind is an immediately invoked lambda which initialized two collections and change `add` to take the collection, but the only reason to do that is to avoid the macro.)

nbougalis:
I’d simply move the ticket field out of the individual transaction formats and into the common fields and then add a manual check in Transactor.cpp to ensure that transactions that aren’t supposed to have this field (i.e. system generated pseudo-transactions) don’t accidentally include it.

HowardHinnant (approved):

intelliot:
> I’d simply move the ticket field out of the individual transaction formats and into the common fields and then add a manual check in Transactor.cpp to ensure that transactions that aren’t supposed to have this field (i.e. system generated pseudo-transactions) don’t accidentally include it.

Given this PR has two approvals, I think we can put this in for now, and consider the above suggestion in a separate/later PR.

intelliot:
@mvadari - at your convenience, you can resolve the conflict so that this PR would be able to merge cleanly on `develop`

intelliot:
(Since this PR was force-pushed, let's get 1 re-review from @HowardHinnant , at his convenience)

mvadari:
> (Since this PR was force-pushed, let's get 1 re-review from @HowardHinnant , at his convenience)

Sorry, was I supposed to merge it? I thought you preferred rebases + force pushes.

HowardHinnant (approved):

intelliot:
FWIW, I prefer merge commits. It's the way git is designed to be used.

PR #4642, opened 2023-08-07
Add unit-tests to verify CLOB/AMM offer and strand selection logic with the transfer fee

gregtatcam:

This PR includes only the unit-tests to verify the payment engine logic to select AMM/CLOB offer and to select the best quality strand when the transfer fee is set.

### Context of Change

The transfer fee changes the offer quality and therefore has to be considered when selecting CLOB/AMM offer or the best quality strand. The `transfer in` rate can be optimized out since it is applied to both CLOB and AMM offer. The `transfer out` rate is set to one in case of the cross currency payment. The quality adjustment for the transfer fee in case of the offer crossing is disabled. Therefore the transfer fee adjustment cancels itself out. 

### Type of Change

- [x] Tests

## Test Plan

The unit-tests consider two cases: 

- If AMM has the same spot price quality as CLOB offer, then AMM can't generate an offer with a better quality than CLOB offer. Therefore, CLOB offer has to be selected over AMM offer and consequently a strand with CLOB offer has to be selected over a strand with AMM offer regardless of the transfer fee.
- If AMM has a slightly better spot price quality than CLOB offer, then AMM can generate an offer with a better quality than CLOB offer. Therefore, AMM offer has to be selected over CLOB offer and consequently a strand with AMM offer has to be selected over a strand with CLOB offer regardless of the transfer fee.

Both cases are considered for the cross-currency payment and the offer crossing.

## Future Tasks

If `ownerPaysFee` amendment is enabled that the transfer fee adjustment has to be reconsidered.

### DIFF
--- a/src/test/app/AMM_test.cpp
+++ b/src/test/app/AMM_test.cpp
@@ -4133,6 +4133,341 @@ private:
     }
 
     void
+    testSelection()
+    {
+        testcase("Offer/Strand Selection");
+        using namespace jtx;
+        Account const ed("ed");
+        Account const gw1("gw1");
+        auto const ETH = gw1["ETH"];
+        auto const CAN = gw1["CAN"];
+
+        auto prep = [&](Env& env, auto gwRate, auto gw1Rate) {
+            fund(env, gw, {alice, carol, bob, ed}, XRP(2'000), {USD(2'000)});
+            env.fund(XRP(2'000), gw1);
+            fund(
+                env,
+                gw1,
+                {alice, carol, bob, ed},
+                {ETH(2'000), CAN(2'000)},
+                Fund::IOUOnly);
+            env(rate(gw, gwRate));
+            env(rate(gw1, gw1Rate));
+            env.close();
+        };
+
+        for (auto const& rates :
+             {std::make_pair(1.5, 1.9), std::make_pair(1.9, 1.5)})
+        {
+            // Offer Selection
+
+            // Cross-currency payment: AMM has the same spot price quality
+            // as CLOB's offer and can't generate a better quality offer.
+            // The transfer fee in this case doesn't change the CLOB quality
+            // because trIn is ignored on adjustment and trOut on payment is
+            // also ignored because ownerPaysTransferFee is false in this case.
+            // Run test for 0) offer, 1) AMM, 2) offer and AMM
+            // to verify that the quality is better in the first case,
+            // and CLOB is selected in the second case.
+            {
+                std::array<Quality, 3> q;
+                for (auto i = 0; i < 3; ++i)
+                {
+                    Env env(*this);
+                    prep(env, rates.first, rates.second);
+                    std::shared_ptr<AMM> amm;
+                    if (i == 0 || i == 2)
+                    {
+                        env(offer(ed, ETH(400), USD(400)), txflags(tfPassive));
+                        env.close();
+                    }
+                    if (i > 0)
+                        amm = std::make_shared<AMM>(
+                            env, ed, USD(1'000), ETH(1'000));
+                    env(pay(carol, bob, USD(100)),
+                        path(~USD),
+                        sendmax(ETH(500)),
+                        txflags(tfPartialPayment));
+                    env.close();
+                    // CLOB and AMM, AMM is not selected
+                    if (i == 2)
+                    {
+                        BEAST_EXPECT(amm->expectBalances(
+                            USD(1'000), ETH(1'000), amm->tokens()));
+                    }
+                    BEAST_EXPECT(expectLine(env, bob, USD(2'100)));
+                    q[i] = Quality(Amounts{
+                        ETH(2'000) - env.balance(carol, ETH),
+                        env.balance(bob, USD) - USD(2'000)});
+                }
+                // CLOB is better quality than AMM
+                BEAST_EXPECT(q[0] > q[1]);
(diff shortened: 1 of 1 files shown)


seelabs (approved):
:+1:

src/test/app/AMM_test.cpp
```diff
@@ -4132,6 +4132,341 @@ struct AMM_test : public jtx::AMMTest
...
+        for (auto const& rates :
+             {std::make_pair(1.5, 1.9), std::make_pair(1.9, 1.5)})
+        {
+            // Offer Selection
+
+            // Cross-currency payment: AMM has the same spot price quality
+            // as CLOB's offer and can't generate a better quality offer.
+            // The transfer fee in this case doesn't change the CLOB quality
+            // because trIn is ignored on adjustment and trOut on payment is
+            // also ignored because ownerPaysTransferFee is false in this case.
```
ximinez, line 4168: The only reason that `ownerPaysTransferFee` is false is because [the OwnerPaysFee feature is unsupported](https://github.com/gregtatcam/rippled/blob/amm-fixes/src/ripple/protocol/impl/Feature.cpp#L413). If that ever changes to `Supported::yes`, some of these tests will fail. If that's intentional, great! If not, you should explicitly create the `Env` objects with a `FeatureBitset` that explicitly removes `featureOwnerPaysFee`.

src/test/app/AMM_test.cpp
```diff
@@ -4132,6 +4132,341 @@ struct AMM_test : public jtx::AMMTest
...
+                    BEAST_EXPECT(expectOffers(env, ed, 0));
+                    BEAST_EXPECT(expectOffers(env, alice, 0));
+                }
+                // Fails to cross because AMM is not selected
+                else
+                {
+                    BEAST_EXPECT(expectOffers(
+                        env, alice, 1, {Amounts{USD(400), ETH(400)}}));
+                    BEAST_EXPECT(expectOffers(env, ed, 0));
+                }
```
ximinez, line 4245: Nit: Both blocks have `BEAST_EXPECT(expectOffers(env, ed, 0));`. It makes sense to move it outside of the `if/else`.

src/test/app/AMM_test.cpp
```diff
@@ -4132,6 +4132,341 @@ struct AMM_test : public jtx::AMMTest
...
+                        txflags(tfPartialPayment));
+                    env.close();
+
+                    BEAST_EXPECT(expectLine(env, bob, USD(2'100)));
+
+                    if (i == 2)
+                    {
+                        // Liquidity is consumed from AMM strand only
+                        if (rates.first == 1.5)
+                        {
```
ximinez, line 4421: This comment only applies to the `rates.first == 1.5`, correct? If so, then it should be inside that block.

src/test/app/AMM_test.cpp
```diff
@@ -4132,6 +4132,341 @@ struct AMM_test : public jtx::AMMTest
...
+            // Run test for 0) offer, 1) AMM, 2) offer and AMM
+            // to verify that the quality is better in the first case,
+            // and CLOB is selected in the second case.
+            {
+                std::array<Quality, 3> q;
+                for (auto i = 0; i < 3; ++i)
+                {
+                    Env env(*this);
+                    prep(env, rates.first, rates.second);
+                    std::shared_ptr<AMM> amm;
```
ximinez, line 4178: Why are you using `shared_ptr` in these tests? Since `amm` is not used anywhere outside this block, it probably makes more sense to use `std::optional`, and `emplace` the object if you want it. I'm not sure if there's a performance benefit either way - it just seems cleaner to me.

src/test/app/AMM_test.cpp
```diff
@@ -4132,6 +4132,341 @@ struct AMM_test : public jtx::AMMTest
...
+                        env(offer(ed, ETH(400), USD(400)), txflags(tfPassive));
+                        env.close();
+                    }
+                    if (i > 0)
+                        amm = std::make_shared<AMM>(
+                            env, ed, USD(1'000), ETH(1'000));
+                    env(pay(carol, bob, USD(100)),
+                        path(~USD),
+                        sendmax(ETH(500)),
+                        txflags(tfPartialPayment));
```
ximinez, line 4190: All of these payments are using the `tfPartialPayment` flag, but delivering the full amount. I'd rather that flag not be used unless you're testing partial payments explicitly, or don't expect the full amount to be delivered.

ximinez (changes requested):

src/test/app/AMM_test.cpp
ximinez: Could you add that to the explanation?

ximinez (approved):
This looks good. I left another comment asking to update the test description, but I figure I don't need to re-review after you add that.

> Could you add that to the explanation? Something like, "These tests are expected to fail if the OwnerPaysFee feature is ever supported. Updates will need to be made to AMM handling in the payment engine, and these tests will need to be updated."

intelliot (approved):

gregtatcam:
> @gregtatcam to confirm whether this is ready to merge. If it's waiting for something, please comment to note that here.

This PR is ready to merge.

intelliot:
This PR was merged as part of #4626.

https://github.com/XRPLF/rippled/pull/4626#issuecomment-1684031952

PR #4650, opened 2023-08-12
Update Ubuntu build image

legleux:
### Context of Change
The DEB build wasn't linking all dependencies statically so failed to run on systems that didn't include the correct version.
The repo that apt pulls GCC from updated the version but apt was not updated so errored trying to retrieve the wrong version.
Moved some commands into scripts and updated the image used to build the packages also.

### Type of Change


thejohnfreeman (approved):
Nothing here sticks out to me as wrong. I know part of it came from our private discussion. If it works, it works.

Aside, I always want to take an opportunity in reviews of this part of the codebase to reiterate that we should split it out into a separate codebase. Hope that's on the roadmap.

legleux:
> Aside, I always want to take an opportunity in reviews of this part of the codebase to reiterate that we should split it out into a separate codebase. Hope that's on the roadmap.

Yes, on the roadmap for this quarter.

intelliot:
What are some examples of systems that didn't include the correct version of gcc?

What error (verbatim) does this PR fix, and who (specifically) benefits from this PR being merged?

legleux:
> What are some examples of systems that didn't include the correct version of gcc?
Ubuntu 18 and 20 , Debian 10 and 11 and CentOS 7

> What error (verbatim) does this PR fix, and who (specifically) benefits from this PR being merged?
A runtime error of a `rippled` built on a system with the non-default GCC and and install time error with packages built with the old commands/image.
```
dpkg: dependency problems prevent configuration of rippled:
 rippled depends on libstdc++6 (>= 12); however:
  Version of libstdc++6:amd64 on system is 10.3.0-1ubuntu1~20.04.
```

intelliot (approved):

intelliot:
FYI @manojsdoshi : @legleux says this will enable us to build packages

manojsdoshi (approved):
LGTM

PR #4653, opened 2023-08-14
Update secp256k1 to 0.3.2

thejohnfreeman:
The new [code](https://github.com/bitcoin-core/secp256k1/commit/acf5c55ae6a94e5ca847e07def40427547876101) was dropped directly in `src/secp256k1` without changes. We can consider changing to a Git submodule, though that will require changes to the build instructions because we are not using submodules anywhere else.


intelliot:
In your PR description, can you link to where you got the secp256k1 source code from, and (if possible) the commit hash of the code used?

CMakeLists.txt
```diff
@@ -71,7 +71,9 @@ find_package(OpenSSL 1.1.1 REQUIRED)
 set_target_properties(OpenSSL::SSL PROPERTIES
   INTERFACE_COMPILE_DEFINITIONS OPENSSL_NO_SSL2
 )
+set(SECP256K1_INSTALL TRUE)
 add_subdirectory(src/secp256k1)
+add_library(secp256k1::secp256k1 ALIAS secp256k1)
```
intelliot, line 76: Why were these two lines added?

intelliot (approved):
I checked out b9b7a71136689525c26a128b5dd942c3cd0a4b0d and confirmed that `src/secp256k1` is identical to https://github.com/bitcoin-core/secp256k1/commit/acf5c55ae6a94e5ca847e07def40427547876101 (v0.3.2)

CMakeLists.txt
thejohnfreeman: The previous `CMakeLists.txt` in `src/secp256k1` was written by us. The new `CMakeLists.txt` was written by the Bitcoin developers. The `SECP256K1_INSTALL` option is necessary to get it to include `install` instructions, and they don't define the `ALIAS` library (a CMake best practice) like we were.

ckeshava:
Hey John,
Do we need to manually update the code for every new version of secp256k1 ? 

If we adopt a git submodule approach and if there is a mistake in the bitcoin-core's implementation of the secp256k1, isn't that a risk for rippled?

thejohnfreeman:
We don't have to update to every new version of any dependency. Our update cadence is up to us. We always manually update dependencies.

A mistake can be found in the implementation whether or not we use submodules. If a mistake is found in a dependency, the most common fix is to switch to a different version.

ckeshava:
@thejohnfreeman  Okay. What are the tradeoffs between manually copying the code versus performing an import using `#include` ? Does it have an impact on the size of the generated binary or the developer productivity?

In an ideal world, if this piece of code were exported as a library by the bitcoin-core developers, we could import it. That is a better solution right?

thejohnfreeman:
@ckeshava I don't understand your questions. Manually copy the code where? We are `#include`-ing it no matter what. There is no choice between the two. Export how? Right now they are publishing it on GitHub. If they were to publish their own Conan recipe, and make it available in Conan Center, then that would be easiest, but using `add_subdirectory` as we do now is not much more difficult.

ckeshava:
ok, I misunderstood that using git submodules has an impact on the size of the rippled binary. thanks for the clarification.

ckeshava (approved):

intelliot:
note: confirmed with @thejohnfreeman that this looks good to go

intelliot:
- could check what bitcoin-core does for testing this
- expect any issues to be caught by existing end to end tests

PR #4657, opened 2023-08-16
Improve `Buffer` to avoid small allocations:

nbougalis:

Many of the uses of the `Buffer` class involve allocations where the size is relatively small and maybe even known at compile time.

This commit introduces code that allows for allocations of up to 112 bytes (sufficient for almost all existing cases) to be serviced without requiring the heap at runtime.

Since instances of `Buffer` are typically allocated on the stack and are short-lived, this change should not have any measurable impact on memory usage.

### Type of Change

- [ ] Documentation Updates
- [ ] Release


src/ripple/basics/Buffer.h
```diff
@@ -113,118 +153,150 @@ class Buffer
...
     }
 
-    /** Return a pointer to beginning of the storage.
+    /** Return a pointer to the storage.
         @note The return type is guaranteed to be a pointer
               to a single byte, to facilitate pointer arithmetic.
     */
     /** @{ */
-    std::uint8_t const*
+    [[nodiscard]] std::uint8_t const*
```
ckeshava, line 187: how does one know the size of the valid data?
the `buf_` has 112 bytes of storage. If I'm using 30 bytes, the remaining will be filled with garbage data. Are the consumers of this function expected to check the size too?

src/ripple/protocol/impl/STVar.h
```diff
@@ -66,12 +73,12 @@ class STVar
 
     STVar(STBase&& t)
     {
-        p_ = t.move(max_size, &d_);
+        p_ = t.move(sizeof(d_), &d_);
```
ckeshava, line 76: Is there any advantage to using the `sizeof` function instead of a `static` variable?
The existing design uses `static constexpr` which caches the size of the buffer. I suspect the compiler performs such an optimization in your code too.

src/ripple/protocol/impl/SecretKey.cpp
```diff
@@ -263,14 +263,15 @@ sign(PublicKey const& pk, SecretKey const& sk, Slice const& m)
                     nullptr) != 1)
                 LogicError("sign: secp256k1_ecdsa_sign failed");
 
-            unsigned char sig[72];
-            size_t len = sizeof(sig);
+            Buffer b(72);
```
ckeshava, line 266: I feel the existing data structure serves the purpose in this case. Is it necessary to use Buffer here?

It's easier to read and understand `unsigned char sig[72]` imho

src/test/basics/Buffer_test.cpp
```diff
@@ -27,15 +29,6 @@ namespace test {
 
 struct Buffer_test : beast::unit_test::suite
 {
-    bool
-    sane(Buffer const& b) const
-    {
-        if (b.size() == 0)
-            return b.data() == nullptr;
```
ckeshava, line 34: I'd like to clarify that the new code does not adhere to the old behavior. When `size_ == 0`, `data_` will point to garbage data in the proposed code, whereas in the existing code, `data_ == nullptr`

src/ripple/basics/Buffer.h
```diff
@@ -21,70 +21,89 @@
...
+
     Meets the requirements of BufferFactory.
 */
 class Buffer
 {
 private:
-    std::unique_ptr<std::uint8_t[]> p_;
-    std::size_t size_ = 0;
+    // Small buffer optimization
+    alignas(std::max_align_t) std::array<std::uint8_t, 112> buff_;
```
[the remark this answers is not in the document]
nbougalis: It _should_ work correctly on existing platforms even without the `alignas`.

src/ripple/basics/Buffer.h
nbougalis: Users of this class have always been expected the know the amount of data they've stored into the buffer. Whether that's a good idea or not, or whether that functionality could be consolidated into this class is another question.

src/ripple/protocol/impl/STVar.h
nbougalis: I believe using `sizeof` is less error prone and communicates the intent better. It may _also_ help reduce false positive warnings by the compiler, though that's a bit of a stretch.

src/test/basics/Buffer_test.cpp
nbougalis: Yes, good eye. That is an implementation detail, which the tests checked for, and that's what's happening here.

Having the internal `data_` pointer always pointing to valid memory (what's known as an [invariant](https://en.wikipedia.org/wiki/Class_invariant)) and never to `nullptr` simplifies other parts of the code (always a good thing), and imposes no additional cost.

src/ripple/protocol/impl/SecretKey.cpp
nbougalis: In _theory_, this should allow the compiler to avoid a copy (see details about [copy elision](https://en.cppreference.com/w/cpp/language/copy_elision#Non-mandatory_copy.2Fmove_.28since_C.2B.2B11.29_elision) and NRVO).

In practice, even if you end up copying, it's just a handful of bytes, and the performance impact (if there is any) will not really be measurable.

_However_, you may be right that this particular change ends up being less clear (and requires the addition of the `resize` member function in `Buffer`). I'll consider reverting this change, thanks!

src/ripple/basics/Buffer.h
ckeshava: okay

src/ripple/protocol/impl/SecretKey.cpp
ckeshava: okay, in some ways, this optimization is similar to the placement-new operation. thanks for resource link!

src/test/basics/Buffer_test.cpp
ckeshava: okay thanks for the clarification.

Is the invariant regarding the non-nullness of the `data_` pointer widely held across the codebase? I'm wondering if stating it as an assert or a comment would help.

ckeshava (approved):

src/test/basics/Buffer_test.cpp
nbougalis: The invariant is internal to the class. External code is unaware of it.

intelliot:
Given this has no measurable impact on memory usage, what's the motivation for the change? Why do it?

nbougalis:
> Given this has no measurable impact on memory usage, what's the motivation for the change? Why do it?

I think it improves and simplifies the codebase, while reducing runtime overhead and minimizing memory fragmentation.

But really, the bottom line is that it's just something I just felt like doing and that's the motivation behind it. :man_shrugging:

PR #4660, opened 2023-08-17
Proposed 1.11.0-devnet

intelliot:

This is 1.11.0, but with the minimal changes necessary to prevent the server from being amendment blocked when `fixReducedOffersV1` and `featureClawback` are enabled.

### Context of Change

The QA team is attempting to reproduce a crash that has been previously observed on Devnet. The crash has been observed with 1.12.0-b1, but no other versions. In order to confirm whether the crash can be triggered in 1.11.0, this proposal can serve as a custom build to run on Devnet for testing purposes.

If a node running this code crashes, then it is possible that 1.11.0 could also crash in the same way.

### Type of Change

- [x] Custom build for Devnet testing only

## Test Plan

Build and run this code on Devnet and try to make it crash. This is only informative if the crash is reliably reproducible with 1.12.0-b1.

## Future Tasks

If this code cannot be made to crash, that is a strong signal that there may be a bug introduced in https://github.com/XRPLF/rippled/compare/1.11.0...beba871 - this is a list of all the changes between 1.11.0 and 1.12.0-b1.

## Reviewers

- @ximinez : to give an opinion on what may be expected to occur if this custom build, with Clawback and fixReducedOffersV1 added to the list of supported features but no other associated code, is deployed on Devnet.
- @kennyzlei : to consider allowing the server(s) to evade [amendment blocking](https://xrpl.org/amendments.html#amendment-blocked-servers) even though they do not have the actual implementation of the amendments in question

## Open Questions

- What's the expected behavior of a server running this code on Devnet?
- What happens when a Clawback tx is submitted?
- What happens when a client asks for a ledger which had a Clawback tx in it?
- What happens when a tx has a different result due to fixReducedOffersV1?
- Should this code be deployed across all Devnet servers (including validators), or only the client handler(s)?

### DIFF
--- a/src/ripple/protocol/Feature.h
+++ b/src/ripple/protocol/Feature.h
@@ -74,7 +74,7 @@ namespace detail {
 // Feature.cpp. Because it's only used to reserve storage, and determine how
 // large to make the FeatureBitset, it MAY be larger. It MUST NOT be less than
 // the actual number of amendments. A LogicError on startup will verify this.
-static constexpr std::size_t numFeatures = 58;
+static constexpr std::size_t numFeatures = 60;
 
 /** Amendments that this server supports and the default voting behavior.
    Whether they are enabled depends on the Rules defined in the validated
@@ -345,6 +345,8 @@ extern uint256 const featureXRPFees;
 extern uint256 const fixUniversalNumber;
 extern uint256 const fixNonFungibleTokensV1_2;
 extern uint256 const fixNFTokenRemint;
+extern uint256 const fixReducedOffersV1;
+extern uint256 const featureClawback;
 
 }  // namespace ripple
 
--- a/src/ripple/protocol/impl/BuildInfo.cpp
+++ b/src/ripple/protocol/impl/BuildInfo.cpp
@@ -33,7 +33,7 @@ namespace BuildInfo {
 //  and follow the format described at http://semver.org/
 //------------------------------------------------------------------------------
 // clang-format off
-char const* const versionString = "1.11.0"
+char const* const versionString = "1.11.0-devnet"
 // clang-format on
 
 #if defined(DEBUG) || defined(SANITIZER)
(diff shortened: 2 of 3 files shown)


intelliot:
also would like a look from @shawnxie999 and @scottschurr who authored the amendments in question

kennyzlei:
thanks for putting this PR together!  given the current crash issues we are observing in devnet with `1.12.*` releases, I think it is valuable to get a stronger indicator of whether this issue existed previously in the `1.11.0` release.  I think the alternatives that include new features implementations beyond `1.11.0` convolute our ability to identify the offending commit and determine whether this is a live issue in mainnet

overall, I believe this warrants allowing the server(s) to evade [amendment blocking](https://xrpl.org/amendments.html#amendment-blocked-servers) even though they do not have the actual implementation of the amendments in question

I will defer to others in considering the open questions around the behavior of this custom build and giving the actual 👍

intelliot:
Thanks @scottschurr. Does the assessment change if we run this code on ALL of the Devnet validators (on UNL)?

Then there should be no different tx outcomes, aside from (possibly) untrusted nodes.

For context: a potential reason to run this on the Devnet validators would be to better mimic the predicted state of Mainnet on 8/21, when [5 amendments](https://xrpl.org/blog/2023/disallowincoming-and-others-expected.html) are expected to be enabled. Of course, desyncs due to mismatched code on validators vs client handlers is not expected there, since those 5 amendments were introduced in 1.10.0.

seelabs (approved):
:+1:

shawnxie999 (approved):

ximinez:
> Thanks @scottschurr. Does the assessment change if we run this code on ALL of the Devnet validators (on UNL)?

I'm not @scottschurr , but I think the assessment changes. If all the validators are running this code, then you shouldn't expect any more desyncs from other nodes (including client handlers) running the same code.

The important condition there is "running the same code". If all nodes on a given network are running the same code, then they will all be doing the same thing. If they all know about the relevant amendments such that they won't get amendment blocked, and all have the same implementation, then they will all do the same thing, and stay happily in sync. This is true whether "do the same thing" means "do nothing".

ximinez (approved):

Issue #4638, opened 2023-08-01
Add an Invariant check that a deleted account has no traceable objects left on the ledger.


ximinez:
## Summary

Add a new class to InvariantCheck.h/cpp that finds any deleted `AccountRoots` (there should only be one, of course) and attempts to read the directory root, and any other objects that can be accessed directly (e.g. the `SignerList`, NFT page... that might be all of them).

## Motivation

The first version of AMM (#4294) support merged to `develop` did not delete the AMM `AccountRoot` correctly. Fortunately, this was discovered and reported, and a fix is in progress (#4626). During the review of the second PR, it occurred to me that this issue would have been avoided if there was an invariant to verify if a deleted account has no "artifacts" left on the ledger. (https://github.com/XRPLF/rippled/pull/4626#discussion_r1279692699)

## Solution

Create a new invariant class (possible name `CompletelyDeletedAccount`). Use the `visitEntry` function to build a vector of deleted `AccountRoots` (There should only be one, but that's checked by the existing `AccountRootsNotDeleted` class). Use the `finalize` function to attempt to `read` any directly accessible objects from the `view`. "Directly accessible" in this case means any object that can be addressed (via a `Keylet`) using only the `sfAccountID`. That includes, but is not necessarily limited to:
* `Keylet ownerDir(AccountID const& id) noexcept;`
* `Keylet signers(AccountID const& account) noexcept;`
* `Keylet nftpage_min(AccountID const& owner);`
* `Keylet nftpage_max(AccountID const& owner);`

*Edit:* And don't forget to check the for an AMM object if `AMMID` is populated.

intelliot:
After the proposed invariant check has been added, what would occur if a transaction attempts to violate the invariant?

ximinez:
> After the proposed invariant check has been added, what would occur if a transaction attempts to violate the invariant?

It would get a `tecINVARIANT_FAILED` result, doing nothing but claim a fee, which would raise alarm bells all over the place. https://github.com/XRPLF/rippled/blob/develop/src/ripple/app/tx/impl/Transactor.cpp#L923-L949

intelliot:
`tecINVARIANT_FAILED` makes sense. Could you be more specific about the alarm bells? I'm not sure where those are implemented, and where exactly they show up. I think we may want to add some alerting, at minimum for us internally, but potentially for the broader community as well.

ximinez:
[...]

PR #4663, opened 2023-08-17
Invariant: prevent a deleted account from leaving (most) artifacts on the ledger.

ximinez:

Adds invariant `AccountRootsDeletedClean`, which checks that a deleted account doesn't leave any directly accessible artifacts behind. Directly accessible artifacts are those for which a `Keylet` can be generated solely from the account ID, or are linked directly in the `AccountRoot`. These are currently:
* `AccountRoot` - redundant, but doesn't hurt to check
* `DirectoryNode` - root node of the account's directory
* `SignerList`
* `NFTokenPage` - min and max, also using the `succ` method used in NFT page handling.
* `AMM` - Uses the `AMMID` in the `AccountRoot`

If any of those items are found, the invariant fails.

Note that this search is not exhaustive. There are plenty of object types that can't be accessed directly, such as `RippleState` and `Escrow`. However, these are all indexed in the directory, so the current assumption is that if the directory is gone, those objects are gone, too. (It is possible to delete the directory without deleting the items, but hopefully something like that would be so obviously wrong, that it would get caught in review. Or it could be checked in a different invariant.)

#### Amendment: `InvariantsV1_1`

Also introduces a new amendment named `InvariantsV1_1`. My intention is to write several separate invariants all gated by this amendment, and have them all included in a single release, sort of like what was done for `fixNonFungibleTokensV1_2`.

The cool thing about this amendment is the checking and logging is done regardless of whether the amendment is enabled. Only if the amendment is enabled will the transaction result be `tecINVARIANT_FAILED`.

This allows most of the benefits of this change to become immediately available, but avoid a fork in the unlikely event that it fails.

Fortunately, the original `EnforceInvariants` amendment has been retired, so there is no need to worry about ordering.

### Context of Change

Explained in #4638. tl;dr: The first version of AMM support did not delete the AMM account object correctly. A fix was merged in #4682. If an invariant like this had existed, the flaw would have been detected during development. To confirm that, run unit tests against https://github.com/ximinez/rippled/tree/invariant-acctdelete-base, which contains the invariant on top of the original AMM code, without the amendment, `AMMID`, or extra unit tests. Many of the AMM tests fail because of this invariant. (Remember, these issues have since been fixed, as indicated by the tests passing on this branch.)

### Type of Change

- [X ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Before / After

Before: If an account is not deleted completely, there will be no way to know it.

After, before the amendment is enabled: If an account is not deleted completely, a `FTL` message will be logged.

After the amendment: If an account is not deleted completely, a `FTL` message will be logged, and the transaction changes will be reverted, and recorded with a `tecINVARIANT_FAILED` result.

## Test Plan

Existing test suites should cover this scenario adequately. If any testing gets an anomalous result, that may indicate an existing, previously undetected problem, or a problem with this change.

## Future Tasks

As mentioned above, I'd like to roll out a handful of other invariants in the same release as this one.


Builds/CMake/RippledCore.cmake
```diff
@@ -1131,5 +1131,6 @@ if (tests)
     # these two seem to produce conflicts in beast teardown template methods
     src/test/rpc/ValidatorRPC_test.cpp
     src/test/rpc/ShardArchiveHandler_test.cpp
+    src/test/ledger/Invariants_test.cpp
```
mvadari, line 1134: Is there a reason this is included both on line 978 and here?

ximinez: This is a list of files that are built individually even on unity builds. I should double-check that it's still necessary to include here. Or as it says on line 1128,
`# any files that don't play well with unity should be added here`

mvadari: I was mainly questioning whether it still needs to be in its original location as well or if it needs to be moved down here.

ximinez: It would need to be both

gregtatcam:
Got a warning on OSX: 
>     InvariantCheck.cpp:398:39: warning: lambda capture 'enforce' is not used [-Wunused-lambda-capture]
>     auto const objectExists = [&view, enforce, &j](auto const& keylet) {

src/ripple/app/tx/impl/InvariantCheck.cpp
```diff
@@ -358,6 +358,101 @@ AccountRootsNotDeleted::finalize(
...
+            uint256 key =
+                view.succ(first.key, last.key.next()).value_or(last.key);
+
+            // current page
+            if (objectExists(Keylet{ltNFTOKEN_PAGE, key}) && enforce)
+                return false;
+        }
+
+        // Keys directly stored in the AccountRoot object
+        if (auto const ammKey = accountSLE->at(~sfAMMID))
```
gregtatcam, line 444: I think this is redundant since you are checking for the `ownerDir`. If AMM's owner directory is deleted then the AMM's object is deleted as well. One can't exist without the other.

src/test/ledger/Invariants_test.cpp
```diff
@@ -44,16 +54,18 @@ class Invariants_test : public beast::unit_test::suite
...
-            tecINVARIANT_FAILED,
-            tefINVARIANT_FAILED})
+        std::initializer_list<TER> ters =
+            {tecINVARIANT_FAILED, tefINVARIANT_FAILED},
+        Preclose const& preclose = {})
     {
         using namespace test::jtx;
         Env env{*this};
 
         Account A1{"A1"};
```
gregtatcam, line 64: Should be `Account const` in both cases.

src/test/ledger/Invariants_test.cpp
```diff
@@ -162,6 +174,167 @@ class Invariants_test : public beast::unit_test::suite
...
+                // Delete the AMM account, cleaning up the directory, but
+                // without deleting the AMM object
+                auto const sle = ac.view().peek(keylet::account(ammAcctID));
+                if (!sle)
+                    return false;
+
+                BEAST_EXPECT(sle->at(~sfAMMID));
+                BEAST_EXPECT(sle->at(~sfAMMID) == ammKey);
+
+                for (auto const trustKeylet :
```
gregtatcam, line 287: Can be `auto const&`.

gregtatcam (approved):
👍 Looks good. I left a few minor comments for your consideration.

src/ripple/app/tx/impl/details/NFTokenUtils.cpp
```diff
@@ -191,6 +191,7 @@ getPageForToken(
         : carr[0].getFieldH256(sfNFTokenID);
 
     auto np = std::make_shared<SLE>(keylet::nftpage(base, tokenIDForNewPage));
+    assert(np->key() > base.key);
```
mvadari, line 194: What's the purpose of this new assert?

src/ripple/app/tx/impl/InvariantCheck.cpp
```diff
@@ -358,6 +358,101 @@ AccountRootsNotDeleted::finalize(
...
+    // This list should include all of the keylet functions that take a single
+    // AccountID parameter, except nftpage_max, because it is exercised below
+    // using the NFT page lookup logic. (nftpage_min is included because it
+    // should _never_ be present.)
+    using fType = std::function<Keylet(AccountID const& id)>;
+    static std::array<fType, 4> const keyletfuncs{
+        &keylet::account,
+        &keylet::ownerDir,
+        &keylet::signers,
+        &keylet::nftpage_min,
```
mvadari, line 388: DID should be added to this list.

src/ripple/app/tx/impl/details/NFTokenUtils.cpp
ximinez: Er, I added that to make sure that I was understanding the key generation process correctly, and left it there because it didn't do any harm. It can be removed if you'd like.

src/ripple/app/tx/impl/InvariantCheck.cpp
```diff
@@ -358,6 +358,101 @@ AccountRootsNotDeleted::finalize(
...
+        //&keylet::nftpage_max
```
mvadari, line 389: Should this line be removed?

src/ripple/app/tx/impl/InvariantCheck.cpp
```diff
@@ -358,6 +358,101 @@ AccountRootsNotDeleted::finalize(
...
+
+                if (item != nullptr)
+                    return item->getName();
+                return std::to_string(sle->getType());
+            }();
+
+            JLOG(j.fatal())
+                << "Invariant failed: account deletion left behind a "
+                << typeName << " object";
+            assert(enforce);
```
mvadari, line 414: Won't this assert mean that it doesn't log if the amendment is enabled?

src/ripple/app/tx/impl/InvariantCheck.cpp
```diff
@@ -358,6 +358,101 @@ AccountRootsNotDeleted::finalize(
...
+    beast::Journal const& j)
+{
+    // This list should include all of the keylet functions that take a single
+    // AccountID parameter, except nftpage_max, because it is exercised below
+    // using the NFT page lookup logic. (nftpage_min is included because it
+    // should _never_ be present.)
+    using fType = std::function<Keylet(AccountID const& id)>;
+    static std::array<fType, 4> const keyletfuncs{
+        &keylet::account,
+        &keylet::ownerDir,
```
mvadari, line 386: I'm not familiar with how the owner directory structure works. Does not having the first owner directory page mean that there aren't subsequent pages still in existence?

src/ripple/app/tx/impl/InvariantCheck.cpp
ximinez: I centralized the list of functions into `Indexes.h`, included all the functions, and rewrote the invariant slightly to not duplicate effort.

src/ripple/app/tx/impl/InvariantCheck.cpp
ximinez: It should log - the logging is single-threaded last I looked, so it'll finish before the assert fires.

src/ripple/app/tx/impl/InvariantCheck.cpp
ximinez: It should. More to the point, though, the directory root is the only page I can look up with only the `AccountID`. I want to add more extensive directory checking down the line.

Background: The directory structure is basically a doubly-linked list. The root has index 0, and contains optional fields for `IndexNext` and `IndexPrevious`. If there is more than one page, those fields can be used in `page()` to find those pages. The page pointed to by `IndexNext` will have an `IndexPrevious` field of 0, and vice versa.

src/ripple/app/tx/impl/details/NFTokenUtils.cpp
ximinez: Done

src/ripple/app/tx/impl/InvariantCheck.cpp
mvadari: I was mostly confused by the `assert` statement before the `return` statement.

src/test/ledger/Invariants_test.cpp
```diff
@@ -162,6 +174,157 @@ class Invariants_test : public beast::unit_test::suite
...
+                ac.view().erase(sle);
+
+                return true;
+            },
+            XRPAmount{},
+            STTx{ttAMM_WITHDRAW, [](STObject& tx) {}},
+            {tecINVARIANT_FAILED, tefINVARIANT_FAILED},
+            [&](Account const& A1, Account const& A2, Env& env) {
+                // Delete the AMM account, cleaning up the directory, but
+                // without deleting the AMM object
```
mvadari, line 319: I think this comment is incorrect (I was a little confused by it)

src/test/ledger/Invariants_test.cpp
```diff
@@ -162,6 +174,157 @@ class Invariants_test : public beast::unit_test::suite
...
+            {tecINVARIANT_FAILED, tefINVARIANT_FAILED},
+            [&](Account const& A1, Account const& A2, Env& env) {
+                // Preclose callback to create the AMM which will be deleted in
+                // the Precheck callback above.
+                AMM const amm(env, A1, XRP(100), A1["USD"](50));
+                ammAcctID = amm.ammAccount();
+                ammKey = amm.ammID();
+                ammIssue = amm.lptIssue();
+                return true;
+            });
```
mvadari, line 264: Should this test be for a DirectoryNode instead of an AMM?

src/test/ledger/Invariants_test.cpp
ximinez: Fixed. This comment was duplicated in both the callback functions. I clarified one, and fixed the other.

src/test/ledger/Invariants_test.cpp
ximinez: This test is the closest reproduction to the orphaned object bug that inspired this PR in the first AMM PR that I could get. The alert flags as "DirectoryNode", just because that's checked before the other object types that are left.

intelliot:
> Also introduces a new amendment named `InvariantsV1_1`. My intention is to write several separate invariants all gated by this amendment, and have them all included in a single release, sort of like what was done for `fixNonFungibleTokensV1_2`.

Will those other invariants be in separate PRs, or will they be added to this same PR #4663?

My personal preference would be to break them out into separate PRs. You can still introduce the amendment in #4663 but mark it as `Supported::no`, and then we could merge this PR without waiting further. As you noted, "the checking and logging is done regardless of whether the amendment is enabled", so that would be active right away. After the amendment is complete (and has all the invariants you expect to complete in the short term), the last PR touching the amendment can update it to `Supported::yes`. That would work well, right?

ximinez:
> Will those other invariants be in separate PRs, or will they be added to this same PR #4663?

Separate.

> My personal preference would be to break them out into separate PRs. You can still introduce the amendment in #4663 but mark it as `Supported::no`, and then we could merge this PR without waiting further. As you noted, "the checking and logging is done regardless of whether the amendment is enabled", so that would be active right away. After the amendment is complete (and has all the invariants you expect to complete in the short term), the last PR touching the amendment can update it to `Supported::yes`. That would work well, right?

Yes! Thank you for suggesting that.

src/ripple/app/tx/impl/InvariantCheck.cpp
mvadari: Does this invariant check not crash if the assert fails/if the amendment isn't enabled? Or is that what you meant by the logging happens before the assert fires?

ximinez: > Does this invariant check not crash if the assert fails/if the amendment isn't enabled? Or is that what you meant by the logging happens before the assert fires?

If the amendment isn't enabled, and a problem is detected
1. The message will be written to the log.
2. In a debug build (or a release build with asserts enabled), the assert will fire, which will crash rippled.
3. In a release build:
    1. The `objectExists` lambda will return `true`, indicating the object exists.
    2. The callers all have `&& enabled` after the call to `objectExists`, so they will not return a failure to the transaction engine. It'll look like the invariant was fine.

The important thing here is that production servers typically run release builds. Anybody running a debug build will typically know what they're doing, and want to hit `assert`s if something goes wrong.

I wanted to implement it this way so that developers (who are more likely to run debug builds) will run into that `assert` if they make a change that breaks the invariant (hopefully while running unit tests!), regardless of whether the amendment has been enabled on the ledger.

intelliot:
@mvadari - would you be able to review this, or should we request another reviewer?

src/test/ledger/Invariants_test.cpp
```diff
@@ -162,6 +176,164 @@ class Invariants_test : public beast::unit_test::suite
...
+    testAccountRootsDeletedClean()
+    {
+        using namespace test::jtx;
+        testcase << "account root deletion left artifact";
+
+        for (auto const& keyletInfo : directAccountKeylets)
+        {
+            // TODO: Use structured binding once LLVM issue
+            // https://github.com/llvm/llvm-project/issues/48582
+            // is fixed.
```
mvadari, line 189: Looks like this issue has been fixed: https://github.com/llvm/llvm-project/issues/48582

src/ripple/app/tx/impl/InvariantCheck.cpp
mvadari: Ah, the part I was missing was that `assert`s only run on a debug build. Thanks!

src/ripple/app/tx/impl/InvariantCheck.cpp
```diff
@@ -358,6 +358,88 @@ AccountRootsNotDeleted::finalize(
...
+        {
+            Keylet const first = keylet::nftpage_min(accountID);
+            Keylet const last = keylet::nftpage_max(accountID);
+
+            std::optional<uint256> key = view.succ(first.key, last.key.next());
+
+            // current page
+            if (key && objectExists(Keylet{ltNFTOKEN_PAGE, *key}) && enforce)
+                return false;
+        }
```
mvadari, line 428: Just confirming my understanding: the `nftpage_min` and `nftpage_max` keylets are checked in the above `directAccountKeylets` for loop, and this section checks all `NFTokenPage`s in between?

src/test/ledger/Invariants_test.cpp
```diff
@@ -44,16 +54,20 @@ class Invariants_test : public beast::unit_test::suite
         Precheck const& precheck,
```
mvadari, line 54: Minor note: some sort of comment/docstring on this function would be super helpful

mvadari (approved):

src/test/ledger/Invariants_test.cpp
ximinez: Nice. Looks like it'll be in clang 16. Think I should update the comment?

src/ripple/app/tx/impl/InvariantCheck.cpp
ximinez: Yes, that is the intent.

src/test/ledger/Invariants_test.cpp
ximinez: Updated in 527a964

src/ripple/app/tx/impl/InvariantCheck.cpp
ximinez: Added a comment to explain in 527a964

src/test/ledger/Invariants_test.cpp
ximinez: Added in 527a964

ximinez:
Had an accidental force push there, but cleaned it up to push 51aa79c updating some comments.

ximinez:
Internal tracker: https://ripplelabs.atlassian.net/browse/RPFC-111

intelliot:
Perf signed off by Antonio Sun on May 7, 2024

ximinez:
Perf testing has passed while I was out. I've merged in `develop`, including the physical redesign. The only issue is the spurious MacOS test failure, but I think we're allowed to ignore that, so this is ready to merge.

PR #4671, opened 2023-08-23
Improve Slab Allocator & `SHAMapItem` slabber:

nbougalis:
The existing slab allocator has significant performance advantages over normal dynamic memory allocation codepaths (e.g. via `new` or `malloc`) but sacrifices the ability to release memory back to the system.

As a result, otherwise transient spikes in memory usage might result in increased memory usage for the lifetime of the process.

This commit retains the lock-free management of individual slabs, while also making it possible to reclamation memory from slabs that are empty and improving the slab selection strategy to favor fuller slabs.

The commit also adjusts the sizes of slabs used to back `SHAMapItem` to better match the characteristics of the XRP Ledger "mainnet."

### DIFF
--- a/src/ripple/basics/SlabAllocator.h
+++ b/src/ripple/basics/SlabAllocator.h
@@ -30,32 +30,109 @@
 #include <atomic>
 #include <cassert>
 #include <cstdint>
-#include <cstring>
 #include <mutex>
-
-#if BOOST_OS_LINUX
-#include <sys/mman.h>
-#endif
 
 namespace ripple {
 
 template <typename Type>
 class SlabAllocator
 {
+    // The code keeps track of the free list using a singly-linked list, that
+    // is allocated in and maintained within the unused memory of a slab. The
+    // assert ensures that we can store the 'next' pointer.
     static_assert(
         sizeof(Type) >= sizeof(std::uint8_t*),
         "SlabAllocator: the requested object must be larger than a pointer.");
 
     static_assert(alignof(Type) == 8 || alignof(Type) == 4);
 
+    // A simple linked list node to keep track of the free list.
+    //
+    // The nodes for this linked list are constructed inside unused memory
+    // in the slab's allocated block.
+    //
+    // By way of example, on a system with 1 byte long pointers, where the
+    // slab allocator allocates 4 byte long items from a 16 byte long slab
+    // the memory layout is as follows:
+    //
+    //     slot: 0      1      2      3
+    //           [ .... | .... | .... | ....]
+    //
+    // We don't need anything too complicated: the only operations that we
+    // perform are insertion and removal, but only at the head of the list
+    // which means that all we need is very basic CAS.
+    //
+    // Normally, to keep a linked list we need a "node" structure, similar
+    // to this. And, if we're going to have a node, we need some memory in
+    // which to "store" it.
+    //
+    // But remember that we only need a "node" when an item is going to be
+    // on the free list, and something is on the free list because there's
+    // some memory that is otherwise unused.
+    //
+    // So if the size of the unused memory is greater than the size of our
+    // node (which is the size of a pointer) then we can use that directly
+    // and overlay a singly linked list of free objects onto the memory of
+    // those free objects.
+    //
+    // Here's what the memory may look like if slots 1 and 3 are allocated
+    // and slots 0 and 2 are free, with the free list overlaid on top:
+    //
+    //     list_ --------------+
+    //                         |
+    //                         v
+    //     slot: 0      1      2      3
+    //           [ N... | .... | N... | .... ]
+    //           ^ |             |
+    //           | '---> NULL    |
+    //           |               |
+    //           '---------------'
+    //
+    // Notice that although the first empty slot in memory order is slot 0
+    // the free list points to slot 2 as the first available slot and slot
+    // 2 points to slot 0, which points to NULL, as it is it the last free
+    // slot.
+    //
+    // Now imagine that the object in slot 1 is released. The free list is
+    // updated to point to slot 1 as the first available slot. The  memory
+    // now looks like this:
+    //
+    //  list_ ----------+
+    //                  |
+    //                  v
+    //     slot: 0      1      2      3
+    //           [ N... | N... | N... | .... ]
+    //           ^ |      |    ^ |
+    //           | |      |    | |
+    //           | |      '----' |
+    //           | '---> NULL    |
+    //           |               |
+    //           '---------------'
+    //
+    // Note that the order of slots within the free list at any given time
+    // is random, as it depends on the pattern and timing of memory access
+    // of the application.
+    struct alignas(Type) FreeItem
+    {
+        FreeItem* next;
+
+        FreeItem(FreeItem* n = nullptr) noexcept : next(n)
+        {
+        }
+
+        ~FreeItem() noexcept = default;
+    };
+
+    static_assert(sizeof(Type) >= sizeof(FreeItem));
+
     /** A block of memory that is owned by a slab allocator */
     struct SlabBlock
     {
-        // A mutex to protect the freelist for this block:
+        // A mutex to protect this block:
         std::mutex m_;
 
         // A linked list of appropriately sized free buffers:
-        std::uint8_t* list_ = nullptr;
+        FreeItem* list_ = nullptr;
 
         // The next block:
         SlabBlock* next_;
@@ -94,41 +171,35 @@ class SlabAllocator
 
             std::lock_guard _(m_);
 
-            // It's possible that this block has an underlying buffer. If it
-            // does, and it has space available, use it.
+            assert(
+                (avail_ == 0 && list_ == nullptr) ||
+                (avail_ != 0 && list_ != nullptr));
+
+            // It's possible that this block has an underlying buffer. If
+            // it does, and it has space available, we use it.
             if (data_ != nullptr)
             {
                 auto ret = list_;
 
                 if (ret != nullptr)
                 {
-                    // We use memcpy to avoid UB for unaligned accesses
-                    std::memcpy(&list_, ret, sizeof(std::uint8_t*));
+                    list_ = ret->next;
                     --avail_;
                 }
 
-                return ret;
+                return reinterpret_cast<std::uint8_t*>(ret);
             }
 
-            assert(data_ == nullptr);
+            // We are going to take the passed-in memory block and use it
+            // to service allocations. The first item will be returned to
+            // the caller, and we chain the rest into a linked list.
+            for (std::size_t i = 1; i != count; ++i)
+                list_ = new (data + (item * i)) FreeItem(list_);
 
             data_ = data;
-            list_ = nullptr;
-
-            // We're needing an allocation, so grab it now.
-            auto ret = data;
-
-            for (std::size_t i = 0; i != count - 1; ++i)
-            {
-                data += item;
-                // We use memcpy to avoid UB for unaligned accesses
-                std::memcpy(data, &list_, sizeof(std::uint8_t*));
-                list_ = data;
-            }
-
             avail_ = count - 1;
 
-            return ret;
+            return data;
         }
 
         SlabBlock(SlabBlock const& other) = delete;
@@ -159,26 +230,20 @@ class SlabAllocator
     // The size of each individual slab:
     std::size_t const slabSize_;
 
-    std::uint8_t*
+    [[nodiscard]] std::uint8_t*
     allocate(SlabBlock* slab) noexcept
     {
         std::lock_guard l(slab->m_);
 
-        std::uint8_t* ret = slab->list_;
+        auto ret = slab->list_;
 
-        if (ret)
+        if (ret != nullptr)
         {
-            assert(
-                (slab->avail_ != 0) && (slab->avail_ <= itemCount_) &&
-                (ret >= slab->data_) &&
-                (ret + itemSize_ <= slab->data_ + slabSize_));
-
-            // We use memcpy to avoid UB for unaligned accesses
-            std::memcpy(&slab->list_, ret, sizeof(std::uint8_t*));
+            slab->list_ = ret->next;
             --slab->avail_;
         }
 
-        return ret;
+        return reinterpret_cast<std::uint8_t*>(ret);
     }
 
     /** Return an item to this allocator's freelist.
@@ -204,9 +269,7 @@ class SlabAllocator
             // doesn't extend past the end of our buffer:
             assert(ptr + itemSize_ <= slab->data_ + slabSize_);
 
-            // We use memcpy to avoid UB for unaligned accesses
-            std::memcpy(ptr, &slab->list_, sizeof(std::uint8_t*));
-            slab->list_ = ptr;
+            slab->list_ = new (ptr) FreeItem(slab->list_);
 
             if (++slab->avail_ != itemCount_)
                 return true;
@@ -271,7 +334,7 @@ public:
     [[nodiscard]] std::uint8_t*
     allocate() noexcept
     {
-        auto active = active_.load();
+        auto active = active_.load(std::memory_order::relaxed);
 
         if (active)
         {
@@ -284,7 +347,8 @@ public:
             if (auto ret = allocate(slab))
             {
                 // This may fail, but that's OK.
-                active_.compare_exchange_strong(active, slab);
+                active_.compare_exchange_weak(
+                    active, slab, std::memory_order::relaxed);
                 return ret;
             }
         }
@@ -298,11 +362,6 @@ public:
 
         if (!buf) [[unlikely]]
             return nullptr;
-
-#if BOOST_OS_LINUX
-        if (slabSize_ >= megabytes(std::size_t(4)))
-            madvise(buf, slabSize_, MADV_HUGEPAGE);
-#endif
 
         // Check whether there's an existing slab with no associated buffer
         // that we can give our newly allocated buffer to:
@@ -320,10 +379,7 @@ public:
 
                 // This may fail, but that's OK.
                 active_.compare_exchange_weak(
-                    active,
-                    slab,
-                    std::memory_order_relaxed,
-                    std::memory_order_relaxed);
+                    active, slab, std::memory_order::relaxed);
 
                 return ret;
             }
@@ -344,15 +400,12 @@ public:
 
         // Finally, link the new slab
         while (!slabs_.compare_exchange_weak(
-            slab->next_,
-            slab,
-            std::memory_order_release,
-            std::memory_order_relaxed))
+            slab->next_, slab, std::memory_order::release))
         {
             // Nothing to do
         }
 
-        active_.store(slab, std::memory_order_relaxed);
+        active_.store(slab, std::memory_order::relaxed);
         return ret;
     }
 
@@ -367,7 +420,7 @@ public:
     {
         assert(ptr);
 
-        for (auto slab = slabs_.load(std::memory_order_relaxed);
+        for (auto slab = slabs_.load(std::memory_order::relaxed);
              slab != nullptr;
              slab = slab->next_)
         {
@@ -399,6 +452,11 @@ public:
         std::size_t align;
 
     public:
+        /**
+         * @param extra_ The number of additional bytes to allocate per item.
+         * @param alloc_ The number of bytes to allocate for the slab.
+         * @param align_ The alignment of returned pointers.
+         */
         constexpr SlabConfig(
             std::size_t extra_,
             std::size_t alloc_ = 0,
@@ -410,6 +468,13 @@ public:
 
     constexpr SlabAllocatorSet(std::vector<SlabConfig> cfg)
     {
+        if (cfg.size() > allocators_.capacity())
+        {
+            throw std::runtime_error(
+                "SlabAllocatorSet<" + beast::type_name<Type>() +
+                ">: too many slab config options");
+        }
+
         // Ensure that the specified allocators are sorted from smallest to
         // largest by size:
         std::sort(
--- a/src/ripple/basics/SlabAllocator.h
+++ b/src/ripple/basics/SlabAllocator.h
@@ -55,42 +55,80 @@ class SlabAllocator
         std::mutex m_;
 
         // A linked list of appropriately sized free buffers:
-        std::uint8_t* l_ = nullptr;
+        std::uint8_t* list_ = nullptr;
 
-        // The next memory block
+        // The next block:
         SlabBlock* next_;
 
         // The underlying memory block:
-        std::uint8_t const* const p_ = nullptr;
+        std::uint8_t* data_ = nullptr;
 
-        // The extent of the underlying memory block:
-        std::size_t const size_;
+        // The number of items available in this block
+        std::size_t avail_{0};
 
-        SlabBlock(
-            SlabBlock* next,
-            std::uint8_t* data,
-            std::size_t size,
-            std::size_t item)
-            : next_(next), p_(data), size_(size)
+        explicit SlabBlock(SlabBlock* next) : next_(next)
         {
-            // We don't need to grab the mutex here, since we're the only
-            // ones with access at this moment.
-
-            while (data + item <= p_ + size_)
-            {
-                // Use memcpy to avoid unaligned UB
-                // (will optimize to equivalent code)
-                std::memcpy(data, &l_, sizeof(std::uint8_t*));
-                l_ = data;
-                data += item;
-            }
         }
 
-        ~SlabBlock()
+        // We don't release the allocated memory here: slabs are only
+        // destroyed when the process is shutting down, and we cannot
+        // guarantee that this isn't called before the destruction of
+        // all the objects that are allocated in the buffer we have.
+        ~SlabBlock() = default;
+
+        std::uint8_t*
+        assign_and_allocate(
+            std::uint8_t* data,
+            std::size_t size,
+            std::size_t item,
+            std::size_t count)
         {
-            // Calling this destructor will release the allocated memory but
-            // will not properly destroy any objects that are constructed in
-            // the block itself.
+            assert(
+                reinterpret_cast<std::uintptr_t>(data) %
+                    alignof(decltype(data_)) ==
+                0);
+
+            assert(count == (size / item));
+
+            assert(data != nullptr);
+
+            std::lock_guard _(m_);
+
+            // It's possible that this block has an underlying buffer. If it
+            // does, and it has space available, use it.
+            if (data_ != nullptr)
+            {
+                auto ret = list_;
+
+                if (ret != nullptr)
+                {
+                    // We use memcpy to avoid UB for unaligned accesses
+                    std::memcpy(&list_, ret, sizeof(std::uint8_t*));
+                    --avail_;
+                }
+
+                return ret;
+            }
+
+            assert(data_ == nullptr);
+
+            data_ = data;
+            list_ = nullptr;
+
+            // We're needing an allocation, so grab it now.
+            auto ret = data;
+
+            for (std::size_t i = 0; i != count - 1; ++i)
+            {
+                data += item;
+                // We use memcpy to avoid UB for unaligned accesses
+                std::memcpy(data, &list_, sizeof(std::uint8_t*));
+                list_ = data;
+            }
+
+            avail_ = count - 1;
+
+            return ret;
         }
 
         SlabBlock(SlabBlock const& other) = delete;
@@ -100,91 +138,109 @@ class SlabAllocator
         SlabBlock(SlabBlock&& other) = delete;
         SlabBlock&
         operator=(SlabBlock&& other) = delete;
-
-        /** Determines whether the given pointer belongs to this allocator */
-        bool
-        own(std::uint8_t const* p) const noexcept
-        {
-            return (p >= p_) && (p < p_ + size_);
-        }
-
-        std::uint8_t*
-        allocate() noexcept
-        {
-            std::uint8_t* ret;
-
-            {
-                std::lock_guard l(m_);
-
-                ret = l_;
-
-                if (ret)
-                {
-                    // Use memcpy to avoid unaligned UB
-                    // (will optimize to equivalent code)
-                    std::memcpy(&l_, ret, sizeof(std::uint8_t*));
-                }
-            }
-
-            return ret;
-        }
-
-        /** Return an item to this allocator's freelist.
-
-            @param ptr The pointer to the chunk of memory being deallocated.
-
-            @note This is a dangerous, private interface; the item being
-                  returned should belong to this allocator. Debug builds
-                  will check and assert if this is not the case. Release
-                  builds will not.
-         */
-        void
-        deallocate(std::uint8_t* ptr) noexcept
-        {
-            assert(own(ptr));
-
-            std::lock_guard l(m_);
-
-            // Use memcpy to avoid unaligned UB
-            // (will optimize to equivalent code)
-            std::memcpy(ptr, &l_, sizeof(std::uint8_t*));
-            l_ = ptr;
-        }
     };
 
-private:
+    // The first slab, which is always present.
+    SlabBlock slab_;
+
     // A linked list of slabs
     std::atomic<SlabBlock*> slabs_ = nullptr;
 
-    // The alignment requirements of the item we're allocating:
-    std::size_t const itemAlignment_;
+    // A pointer to the slab we're attempting to allocate from
+    std::atomic<SlabBlock*> active_ = nullptr;
 
     // The size of an item, including the extra bytes requested and
-    // any padding needed for alignment purposes:
+    // any padding needed for alignment purposes between items:
     std::size_t const itemSize_;
+
+    // The maximum number of items that this slab can allocate
+    std::size_t const itemCount_;
 
     // The size of each individual slab:
     std::size_t const slabSize_;
 
+    std::uint8_t*
+    allocate(SlabBlock* slab) noexcept
+    {
+        std::lock_guard l(slab->m_);
+
+        std::uint8_t* ret = slab->list_;
+
+        if (ret)
+        {
+            assert(
+                (slab->avail_ != 0) && (slab->avail_ <= itemCount_) &&
+                (ret >= slab->data_) &&
+                (ret + itemSize_ <= slab->data_ + slabSize_));
+
+            // We use memcpy to avoid UB for unaligned accesses
+            std::memcpy(&slab->list_, ret, sizeof(std::uint8_t*));
+            --slab->avail_;
+        }
+
+        return ret;
+    }
+
+    /** Return an item to this allocator's freelist.
+
+        @param slab The slab in which to attempt to return the memory
+        @param ptr The pointer to the chunk of memory being deallocated.
+
+        @return true if the item belongs to the given slab and was freed, false
+                     otherwise.
+    */
+    [[nodiscard]] bool
+    deallocate(SlabBlock* slab, std::uint8_t* ptr) noexcept
+    {
+        assert(ptr != nullptr);
+
+        {
+            std::lock_guard l(slab->m_);
+
+            if (auto const d = slab->data_; (ptr < d) || (ptr >= d + slabSize_))
+                return false;
+
+            // It looks like this objects belongs to us. Ensure that it
+            // doesn't extend past the end of our buffer:
+            assert(ptr + itemSize_ <= slab->data_ + slabSize_);
+
+            // We use memcpy to avoid UB for unaligned accesses
+            std::memcpy(ptr, &slab->list_, sizeof(std::uint8_t*));
+            slab->list_ = ptr;
+
+            if (++slab->avail_ != itemCount_)
+                return true;
+
+            // The slab is now fully unused and doesn't need its data block:
+            ptr = std::exchange(slab->data_, nullptr);
+            slab->list_ = nullptr;
+            slab->avail_ = 0;
+        }
+
+        // Release the memory of this fully unused block back to the system
+        // without holding a lock.
+        boost::alignment::aligned_free(ptr);
+        return true;
+    }
+
 public:
     /** Constructs a slab allocator able to allocate objects of a fixed size
 
-        @param count the number of items the slab allocator can allocate; note
-                     that a count of 0 is valid and means that the allocator
-                     is, effectively, disabled. This can be very useful in some
-                     contexts (e.g. when mimimal memory usage is needed) and
-                     allows for graceful failure.
+        @param extra The number of extra bytes per item, on top of sizeof(Type)
+        @param alloc The number of bytes to allocate for this slab
+        @param align The alignment of returned pointers, normally alignof(Type)
      */
     constexpr explicit SlabAllocator(
         std::size_t extra,
         std::size_t alloc = 0,
-        std::size_t align = 0)
-        : itemAlignment_(align ? align : alignof(Type))
-        , itemSize_(
-              boost::alignment::align_up(sizeof(Type) + extra, itemAlignment_))
+        std::size_t align = alignof(Type))
+        : slab_(nullptr)
+        , slabs_(&slab_)
+        , itemSize_(boost::alignment::align_up(sizeof(Type) + extra, align))
+        , itemCount_(alloc / itemSize_)
         , slabSize_(alloc)
     {
-        assert((itemAlignment_ & (itemAlignment_ - 1)) == 0);
+        assert(std::has_single_bit(align));
     }
 
     SlabAllocator(SlabAllocator const& other) = delete;
@@ -195,15 +251,13 @@ public:
     SlabAllocator&
     operator=(SlabAllocator&& other) = delete;
 
-    ~SlabAllocator()
-    {
-        // FIXME: We can't destroy the memory blocks we've allocated, because
-        //        we can't be sure that they are not being used. Cleaning the
-        //        shutdown process up could make this possible.
-    }
+    // FIXME: We can't destroy the memory blocks we've allocated, because
+    //        we can't be sure that they are not being used. Cleaning the
+    //        shutdown process up could make this possible.
+    ~SlabAllocator() = default;
 
     /** Returns the size of the memory block this allocator returns. */
-    constexpr std::size_t
+    [[nodiscard]] constexpr std::size_t
     size() const noexcept
     {
         return itemSize_;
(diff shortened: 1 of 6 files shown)


seelabs:
I'm seeing a bunch of unit test failures locally (I see that CI is passing, this is just on my local machine). I don't know what's up yet, but thought I'd report it:
```
447.6s, 212 suites, 1797 cases, 638417 tests total, 42 failures
```

src/ripple/basics/SlabAllocator.h
```diff
@@ -57,40 +57,66 @@ class SlabAllocator
...
+        ~SlabBlock()
+        {
+            // We don't release the allocated memory here: slabs are only
+            // destroyed when the process is shutting down, and we cannot
+            // guarantee that this isn't called before the destruction of
+            // all the objects that are allocated in the buffer we have.
+        }
+
+        std::uint8_t*
+        assign(
```
seelabs, line 82: This function both assigns and allocates, maybe rename it?

src/ripple/basics/SlabAllocator.h
```diff
@@ -57,40 +57,66 @@ class SlabAllocator
...
         SlabBlock* next_;
 
         // The underlying memory block:
-        std::uint8_t const* const p_ = nullptr;
+        std::atomic<std::uint8_t*> data_ = nullptr;
 
-        // The extent of the underlying memory block:
-        std::size_t const size_;
+        // The number of items allocated from this block
+        std::atomic<std::size_t> avail_{0};
```
seelabs, line 67: `avail_` is always accessed within a lock so I don't think this needs to be `atomic`

src/ripple/basics/SlabAllocator.h
```diff
@@ -57,40 +57,66 @@ class SlabAllocator
         // A linked list of appropriately sized free buffers:
         std::uint8_t* l_ = nullptr;
 
-        // The next memory block
+        // The next block:
         SlabBlock* next_;
 
         // The underlying memory block:
-        std::uint8_t const* const p_ = nullptr;
+        std::atomic<std::uint8_t*> data_ = nullptr;
```
seelabs, line 64: With one exception, `data_` is always access within a lock. I _think_ we can get rid of that exception (see comment below); if we do I don't think this needs to be `atomic`

src/ripple/basics/SlabAllocator.h
```diff
@@ -57,40 +57,66 @@ class SlabAllocator
...
             std::uint8_t* data,
             std::size_t size,
-            std::size_t item)
-            : next_(next), p_(data), size_(size)
+            std::size_t item,
+            std::size_t count)
         {
-            // We don't need to grab the mutex here, since we're the only
-            // ones with access at this moment.
+            assert(
```
seelabs, line 88: I don't really object, but our code style is to put function implementations outside of class definitions. I understand this file already did it this, so I don't mind if you keep it, but thought I'd point it out.

src/ripple/basics/SlabAllocator.h
```diff
@@ -100,91 +126,109 @@ class SlabAllocator
...
-        @param count the number of items the slab allocator can allocate; note
-                     that a count of 0 is valid and means that the allocator
-                     is, effectively, disabled. This can be very useful in some
-                     contexts (e.g. when mimimal memory usage is needed) and
-                     allows for graceful failure.
+        @param extra The number of extra bytes per item, on top of sizeof(Type)
+        @param alloc The number of bytes to allocate for this slab
+        @param align The alignment of returned pointers, normally alignof(Type)
      */
     constexpr explicit SlabAllocator(
```
seelabs, line 222: I don't think this slab allocator will ever be used in a `constexpr` environment. I might remove the `constexpr` tags. They don't do any harm, but they also don't add value.

src/ripple/basics/SlabAllocator.h
```diff
@@ -100,91 +126,109 @@ class SlabAllocator
...
+    std::size_t const slabSize_;
 
-            return ret;
+private:
+    std::uint8_t*
+    allocate(SlabBlock* slab) noexcept
+    {
+        std::uint8_t* ret = nullptr;
+
+        if (slab->data_ != nullptr)
```
seelabs, line 154: I believe this is the only place where slab `data_` is accessed outside of a lock, and it's to check for what will be a rare occurrence and then immediately takes out a lock. What do you think about putting this inside a lock and removing all the atomic variables/atomic swaps and making the mutex protect `data_`?

seelabs (commented):
Did a first pass and left some comments to consider.

src/ripple/basics/SlabAllocator.h
```diff
@@ -57,40 +57,66 @@ class SlabAllocator
...
+            std::size_t item,
+            std::size_t count)
         {
-            // We don't need to grab the mutex here, since we're the only
-            // ones with access at this moment.
+            assert(
+                reinterpret_cast<std::uintptr_t>(data) %
+                    alignof(decltype(data_)) ==
+                0);
+            assert(
```
HowardHinnant, line 92: This should be `static_assert` as it is testing compile-time information.

src/ripple/basics/SlabAllocator.h
HowardHinnant: Once we get fully conforming C++20 compilers, I _believe_ (can't test that I'm right) that the `slabber` variable in SHAMapitem.h, line 117 could be decorated with `constinit`, as long as this constructor remains `constexpr`, which would ensure that this constructor runs at compile-time.

HowardHinnant (approved):

src/ripple/basics/SlabAllocator.h
nbougalis: This function doesn't allocate: the memory block is passed _into_ this function.

src/ripple/basics/SlabAllocator.h
nbougalis: I get it, and I could definitely move the functions below the classes if you feel strongly enough about it, but I think the better option here (given that this is somewhat esoteric, unlikely to be casually edited, etc) is to simply inline the function implementations directly into the class.

nbougalis:
Review comments addressed and code rebased to latest `develop`.

src/ripple/basics/SlabAllocator.h
```diff
@@ -100,91 +138,109 @@ class SlabAllocator
...
+    std::uint8_t*
+    allocate(SlabBlock* slab) noexcept
+    {
+        std::lock_guard l(slab->m_);
+
+        std::uint8_t* ret = slab->list_;
+
+        if (ret)
+        {
+            assert(
```
ckeshava, line 171: are asserts checked at runtime in non-debug builds? I'm wondering if there are situations where this assert is not checked.

src/ripple/basics/SlabAllocator.h
seelabs: I'm fine with it as-is. Just thought I'd point it out.

seelabs:
@nbougalis It's easier for me if you don't squash things down when you address review comments. It isn't _that_ big a deal, but it makes it easy to see exactly what changed from the last time I looked at it.

src/ripple/basics/SlabAllocator.h
```diff
@@ -214,95 +268,111 @@ class SlabAllocator
...
 #endif
 
-        // We need to carve out a bit of memory for the slab header
-        // and then align the rest appropriately:
-        auto slabData = reinterpret_cast<void*>(
-            reinterpret_cast<std::uint8_t*>(buf) + sizeof(SlabBlock));
-        auto slabSize = size - sizeof(SlabBlock);
+        // Check whether there's an existing slab with no associated buffer
+        // that we can give our newly allocated buffer to:
+        for (auto slab = slabs_.load(std::memory_order_relaxed);
```
seelabs, line 309: This loop does the load using `memory_order_relaxed` and the loop on line 282 uses `memory_order_seq_cst` (the default). Is this intentional? I _think_ they could both use the same memory order, and I _think_ they could both be relaxed (although I don't use memory orders that often, so I'm not confident here). If there's a reason for the difference we should probably add a comment. If there's no reason for the difference we should be consistent.

seelabs (commented):
I think this looks good, but I'm going to wait to give a thumbs up. I want to give a closer look to all the `memory_order_relaxed` loads and stores before I approve this. Also left a couple of minor comments to consider, but nothing that should hold this up.

src/ripple/basics/SlabAllocator.h
```diff
@@ -55,42 +55,80 @@ class SlabAllocator
...
+        std::size_t avail_{0};
 
-        SlabBlock(
-            SlabBlock* next,
+        explicit SlabBlock(SlabBlock* next) : next_(next)
+        {
+        }
+
+        // We don't release the allocated memory here: slabs are only
+        // destroyed when the process is shutting down, and we cannot
```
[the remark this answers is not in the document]
nbougalis: No, not quite: what this comment says is that the *destructor* of a `SlabBlock` that points to an allocated chunk of memory does not deallocate that chunk. Of course, this destructor will _only_ ever be called when the process is shutting down, at which point any and all allocated memory will be released back to the O/S.

Note that when a `SlabBlock` becomes empty, the buffer it points to is dynamically deallocated. But as the code is written, during shutdown it's possible for the destructor of a SlabBlock to be called while items are still allocated out of the slab block.

It's possible to fix this but I don't know if the computational overhead is worth it for a hot path.

src/ripple/basics/SlabAllocator.h
```diff
@@ -55,42 +55,80 @@ class SlabAllocator
...
+            // It's possible that this block has an underlying buffer. If it
+            // does, and it has space available, use it.
+            if (data_ != nullptr)
+            {
+                auto ret = list_;
+
+                if (ret != nullptr)
+                {
+                    // We use memcpy to avoid UB for unaligned accesses
+                    std::memcpy(&list_, ret, sizeof(std::uint8_t*));
```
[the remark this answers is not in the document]
nbougalis: I agree, it's tricky, idiomatic code. We could define a simple `struct` and use it to keep the free list. The code then reads better (and avoids the need for `std::memcpy`, which simplifies things further), and I suspect it'll optimize just as well.

```
struct FreeList
{
    FreeList *next;
};
```

I'll take a look.

src/ripple/basics/SlabAllocator.h
```diff
@@ -55,42 +55,80 @@ class SlabAllocator
...
+                    --avail_;
+                }
+
+                return ret;
+            }
+
+            assert(data_ == nullptr);
```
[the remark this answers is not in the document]
nbougalis: Not required; I'll remove it.

src/ripple/basics/SlabAllocator.h
```diff
@@ -100,91 +138,109 @@ class SlabAllocator
...
+            slab->list_ = nullptr;
+            slab->avail_ = 0;
+        }
 
-    // The size of each individual slab:
-    std::size_t const slabSize_;
+        // Release the memory of this fully unused block back to the system
+        // without holding a lock.
+        boost::alignment::aligned_free(ptr);
+        return true;
```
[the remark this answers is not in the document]
nbougalis: There's no object stored at `ptr`.

src/ripple/basics/SlabAllocator.h
```diff
@@ -100,91 +138,109 @@ class SlabAllocator
...
-    };
+            // It looks like this objects belongs to us. Ensure that it
+            // doesn't extend past the end of our buffer:
+            assert(ptr + itemSize_ <= slab->data_ + slabSize_);
 
-private:
-    // A linked list of slabs
-    std::atomic<SlabBlock*> slabs_ = nullptr;
+            // We use memcpy to avoid UB for unaligned accesses
+            std::memcpy(ptr, &slab->list_, sizeof(std::uint8_t*));
```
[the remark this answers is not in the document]
nbougalis: As @seelabs explained in a previous post, this basically creates a linked list, and uses the memory of the block as the storage for the linked list itself.

Let's look at an example: imagine a system where pointers are 4 bytes long, and the slab allocator is configured to allocate 60 bytes at a time, and divvy it up into 3 20 byte blocks.

Here's what this looks like:

```
 +------  slot 0  ------+ +------  slot 1  ------+ +------  slot 2  ------+
[.... .... .... .... ....|.... .... .... .... ....|.... .... .... .... ....|
```

We need to keep track of a "free list" of items. We don't need anything too complicated (we only work with the "head" of the list, nothing too complicated) so a simple singly-linked list will work.

Normally, to keep a linked list we need some kind of "node" structure. Something like:

```
struct Node
{
    Node *next;
};
```

And, if we're going to have a node, we need some memory in which to "store" it.

But remember that we only need a "node" when an item is going to be on the free list, and something is on the free list because there's some memory that is otherwise unused.

So if the size of the unused memory is greater than the size of our node (which, is the size of a pointer!) then we can just use that memory! In other words, we overlay a singly-linked list of free objects onto the memory of those free objects.

Let's imagine that the 60-byte block of memory we have contains one item (slot 1) that's allocated, and two items that are free (slots 0 and 2) . Here's what that looks link:


```
           +------   free   ------+ +------  object  ------+ +------   free   ------+
          [NEXT .... .... .... ....|.... .... .... .... ....|NEXT .... .... .... ....|
          ^   |                                             ^   |
          |   |                                             |   `------> NULL
          |   `---------------------------------------------'
          |
list_ = --+
```

Now, let's imagine that the object that's in use is released. Here's how things change (sorry, the ASCII graphics are getting more complex here):

```
           +------   free   ------+ +------   free   ------+ +------   free   ------+
          [NEXT .... .... .... ....|NEXT .... .... .... ....|NEXT .... .... .... ....|
          ^   |                    ^   |                    ^   |
          |   |                    |   |                    |   `------> NULL
          `----------------------------'                    |
              |                    |                        |
              `--------------------|------------------------'
                                   |
list_ = ---------------------------+
```

Notice how the head linked list now points to slot 1, slot 1 points to slot 0 and slot 0 points to `nullptr` (i.e. the end of the linked list).

Notice too how the "nodes" of the linked list are occupying the same memory that objects do when the slots are allocated.

I hope this helps.

nbougalis:
> @nbougalis It's easier for me if you don't squash things down when you address review comments. It isn't _that_ big a deal, but it makes it easy to see exactly what changed from the last time I looked at it.

Apologies! I rebased against the latest develop, and must have squashed out of habit.

src/ripple/basics/SlabAllocator.h
seelabs: @nbougalis  I'm leaving a comment here so we can have a threaded conversation, even though it doesn't relate to this particular line of code (github top comments can't be threaded).

I'm trying to think though the memory orders. What they do is define dependencies between the loads and stores of different atomic variables in the code. What I'm thinking about is the dependencies between the atomic variables `slabs_` and `active_`. For example, would the code blow up if a write to `active_` was seen before a write to `slabs_` and the active slab was not part of the slabs collection? What are the constraints here?

As far as I can tell, there is no dependency between `slabs_` and `active_` and the code works fine if `active_` is not part of the `slabs_` collection. I can't think of other constraints. If there aren't any other constraints, can't *all* the memory orders be `relaxed`? If there are constraints that I'm missing, can we document them so I can look at the code and make sure the memory orders will satisfy the constraints.

As for right now, I believe ALL the memory orders can be `relaxed`, but I may be missing important constraints.

src/ripple/basics/SlabAllocator.h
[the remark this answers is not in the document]
ckeshava: Okay, the dynamic deallocation of unused buffers occurs through `boost::alignment::aligned_free(ptr);` calls, am I right?

Why is the fix complicated? If we can ensure that the destructor of `SlabBlock` is made after all the other objects have been destroyed, we will not face this issue. 

I'm guessing the shutdown process is concurrent with multiple threads, hence causes this stated overhead?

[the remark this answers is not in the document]
nbougalis: It’s complicated because the order of destruction of objects isn’t, necessarily, trivial or “fixed” and depends of a lot of variables that aren’t always directly controllable and, to the degree that they are, to exercise control can be onerous.

It’s unlikely that a program of even moderate complexity could successfully manage the ordering without relying on compiler extensions, and do so reliably in the face of changes in the codebase and the minutiae of compiling.

As I explained, when the process shuts down, the O/S reclaims all the allocated memory, whether it was freed or not. IMO it seems counterproductive to invest too much time and effort and to complicate the code in order to ensure we release the memory block a few milliseconds before the process shuts down.

src/ripple/basics/SlabAllocator.h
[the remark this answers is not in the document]
ckeshava: Thanks a lot for the detailed explanation, this helps a lot!!

1. Are we using linked lists instead of contiguous memory allocation in the `SlabAllocator` in order to reduce the fragmentation/wastage of memory? I imagine dealing with a contiguous memory blocks would be simpler than with linked lists.

2. In the above example, `itemSize_ == 20 bytes`, `itemCount_ == 3` and `slabSize_ == 60 bytes`. is that correct?

src/ripple/basics/SlabAllocator.h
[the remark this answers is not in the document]
ckeshava: alright, that's fair. thanks for the explanation!

intelliot:
Can you quantify the impact this is expected to have on memory use (or any other measurable metrics)?

src/ripple/basics/SlabAllocator.h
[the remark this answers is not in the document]
nbougalis: Yes, in the example we have `itemSize_ == 20`, `itemCount_ == 3` and `slabSize_ == 60`.

To your first question, we _are_ using a contiguous block of memory in the `SlabAllocator` (in the example above, a 60 byte contiguous block). That's the whole point and it's to reduce fragmentation and the (necessary) overhead that's imposed by general purpose allocators like `malloc` that must be able to hand out chunks of any size.

The linked list is there to allow us to keep track of which "pieces" of the contiguous chunk of memory are in use, and to be able to return them. Remember that deallocations can happen in a different order than allocations.

In the above example, assume the chunk of memory is at address 100. Let's say you try to use the `SlabAllocator` to allocate 3 items.

The memory for the first item will be at address 100 (range: 100-119). The memory for the second item will be at address 120 (range: 120-139). The memory for the third item will be at address 140 (range: 140-159).

Now, assume you are done with the item at address 120 and you return it to the `SlabAllocator`. It needs to know that this item is now available for reallocation, so it keeps track of that in the linked list. It looks like this: `list_` points to address 120. The *value* at address 120 (interpreted as a pointer) points to NULL.

Now, assume you are done with the item at address 140 and you return it to the `SlabAllocator`. It, again, adds the new item to the linked list. It looks like this: `list_` points to address 140. The *value* at address 140 (interpreted as a pointer) points to 120, and the *value* at address 120 points to NULL.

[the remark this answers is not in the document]
ckeshava: thank you, I understand the design 👍 
This does not necessarily pertain to this PR, but I'm thinking about an alternative design choice. 

You could maintain the set of free slots in a `std::unordered_set<int> freeSlabs;` (I'm glossing over the mapping details from integer to the memory address) to achieve the same goals. Allocation is faster because you don't have to traverse through the entire linked list to find the free slab, deallocation is just as fast.

src/ripple/basics/SlabAllocator.h
nbougalis: @scottschurr I agree! Using anything other than `std::memory_order::seq_cst`  is an optimization (often times it's a micro- or pico-optimization) and can be tricky and error-prone, even for experts, and I certainly don't consider myself an expert at this.

Given that we share that belief, it is mildly ironic that I am trying to use anything other than `std::memory_order::seq_cst` (and I think I'm the only one that is trying to do this in the codebase).

With that said, I do believe that it's worth trying to optimize code, especially low-level and performance critical code, and I don't think we should shy away from tools (even "dangerous" power-tools) that allow us to do this, just as long as we are careful and disciplined in their use.

I am fairly confident the semantics here are correct, but I will carefully re-think them (which, I think, is what @seelabs was trying to do as well). I will also run some benchmarks to determine the performance impact of using `std::memory_order::seq_cst` as opposed to `std::memory_order::relaxed`; if the performance impact is not measurable, then the increased "safety" and "understandability" of `std::memory_order::seq_cst` might make sense even if `std::memory_order::relaxed` semantics are fine.

nbougalis:
> Can you quantify the impact this is expected to have on memory use (or any other measurable metrics)?

I guess I could run against the existing code and snapshot memory usage after 30, 60 and 90 minutes and compare then against this, but I don't know that I need to do this. The reason is simple: I contend that the net result of this change will result in a net decrease in memory usage. Proof by vigorous handwaving:

The current code allocates a `SlabBlock` together with the memory that will service that slab in a single operation; once allocated, a `SlabBlock` is never deleted. Several operations might result in a server allocating a large number of items, which might, in turn, result in the allocation of a number of `SlabBlock` instances. Once the server settles in a steady state, the number of items it needs drops, but the memory taken up the `SlabBlock` cannot be released.

The version of the code in this PR _decouples_ the allocation of the `SlabBlock` instance from the memory used to service the slab. Like in the current code a `SlabBlock` can never be deleted, but it is a relatively small object (under 96 bytes) so it doesn't matter. _Unlike_ the current code, however, the memory that will service that slab (which can be tens or hundreds of MB, depending on the configuration of the slab) _is_ allocated separately and _it_ **can be deallocated** once a slab becomes empty. Once the server settles in a steady state and the number of items drops, this code will relinquish unused memory back to the operating system, which is something the existing code doesn't do.

In other words, this change does nothing to increase memory usage but implements functionality that could result in a (measurable!) decrease. ∎

src/ripple/basics/SlabAllocator.h
scottschurr: @nbougalis, thanks for offering to run those benchmarks.  Yes, if the performance impact of the riskier `memory_order` is measurable, then the benefits may justify the risks.

seelabs: Just BTW, here's a quote from the book "Rust Atomics and Locks", a very good book on this topic, even for C++ people like us:
```
While it might seem like the easiest memory ordering to reason about, SeqCst ordering is almost never necessary in practice. In nearly all cases, regular acquire and release ordering suffice.
```

So yeah, `relaxed` is difficult to reason about. `acquire/release` is much less difficult. I'd be OK with `acquire/release` over `seq_cst` (especially since as far as I can tell `relaxed` would work too). But of course, all of this is moot if there's no perf gain.

I'm also not sure how `relaxed` memory ordering work on intel. I vaguely remember that on intel there's no difference between `relaxed` and `acquire/release`, although I may be misremembering. I'll search for this when I get a chance...

src/ripple/shamap/SHAMapItem.h
```diff
@@ -149,36 +149,42 @@ intrusive_ptr_release(SHAMapItem const* x)
...
-    // If we can't grab memory from the slab allocators, we fall back to
-    // the standard library and try to grab a precisely-sized memory block:
+    // If we can't grab memory from the slab allocators, we use new to
+    // get a precisely-sized and properly aligned memory block:
     if (raw == nullptr)
-        raw = new std::uint8_t[sizeof(SHAMapItem) + data.size()];
+        raw = new (std::align_val_t(alignof(SHAMapItem)))
+            std::uint8_t[sizeof(SHAMapItem) + data.size()];
+
+    assert(reinterpret_cast<std::uintptr_t>(raw) % alignof(SHAMapItem) == 0);
```
ckeshava, line 172: Why do you need `raw` to be compatibly aligned with `std::uintptr_t` ? I don't see `raw` being type-casted and used as `std::uintptr_t`

src/ripple/shamap/SHAMapItem.h
```diff
@@ -149,36 +149,42 @@ intrusive_ptr_release(SHAMapItem const* x)
...
     // that the refcount can never be zero before incrementing as an
     // invariant.
-    return {new (raw) SHAMapItem{tag, data}, false};
+    return {
+        new (std::assume_aligned<alignof(SHAMapItem)>(raw))
+            SHAMapItem{tag, data},
+        false};
 }
 
 static_assert(alignof(SHAMapItem) != 40);
```
ckeshava, line 184: this appears arbitrary, is there any reason why the alignment of `SHAMapItem` cannot be `40`?
Moreover, the line below obviates the need for the current line.

src/ripple/shamap/SHAMapItem.h
```diff
@@ -149,36 +149,42 @@ intrusive_ptr_release(SHAMapItem const* x)
         if constexpr (!std::is_trivially_destructible_v<SHAMapItem>)
             std::destroy_at(x);
 
-        // If the slabber doens't claim this pointer, it was allocated
+        // If the slabber doesn't claim this pointer, it was allocated
         // manually, so we free it manually.
         if (!detail::slabber.deallocate(const_cast<std::uint8_t*>(p)))
```
ckeshava, line 154: I'll take this opportunity to ask you about the choice of `boost::intrusive_ptr`, although its unrelated to this PR.
Why didn't you use a `std::shared_ptr` instead? I know that `boost::intrusive_ptr` has a smaller memory footprint.

The `intrusive_ptr_release` appears to be non-trivial. Can it not be implemented with a `std::shared_ptr` ?

src/ripple/basics/SlabAllocator.h
nbougalis: > @nbougalis, thanks for offering to run those benchmarks. Yes, if the performance impact of the riskier `memory_order` is measurable, then the benefits may justify the risks.

So, I ran some benchmarks on _synthetic code_ that implements the same "linked list" behavior that we have here, comparing the performance of `std::memory_order::relaxed` against `std::memory_order::seq_cst`.

Depending on the amount of concurrency (i.e. how many threads compete to modify the linked list), it appears that `std::memory_order::seq_cst` is between 25% and 100% slower that `std::memory_order::relaxed`. Here's a sample run (compiled with `-O3`):

```
------------------------------------------------------
Benchmark            Time             CPU   Iterations
------------------------------------------------------
relaxedTest   91517753 ns       513274 ns          100
cstTest      192449311 ns       994084 ns          100
```

Disclaimer: I _can't_ be sure if the test fully isolates the relevant code and eliminates all other influences, but the results generally seem to  suggest that we are accurately capturing the overhead of `std::memory_order::seq_cst`.

I am confident that the code does what is intended, whether we use `std::memory_order::relaxed` or `std::memory_order::seq_cst`, and so believe that it's hard to justify (the admittedly small) additional overhead of the latter. But if the consensus is that using `std::memory_order::seq_cst` is the desired option, I'm fine with that.

src/ripple/shamap/SHAMapItem.h
nbougalis: The question "Can it not be implemented with a `std::shared_ptr` is not the right question to be asking. The right question is "is `std::shared_ptr` appropriate?

And in many cases (most, even!) it will be appropriate. But in this case it isn't.

Using `boost::intrusive_ptr` allows us to squeeze a lot of savings in terms of size. And because there are _so many_ instances of this class, those small savings in terms of size add up.

Even if that was not a concern, `std::shared_ptr` would impose other limitations as well:

- First note that the fact that the memory allocation performed by `std::make_shared` is out of our control (that is, it cannot use the slabbed allocator).
- Second, it's important to understand what `std::make_shared<SHAMapItem>(...)` does: it allocates a block of memory that is large enough for the `std::shared_ptr` control block _and_ at least `sizeof(SHAMapItem)` extra bytes. But that's not enough: we need more data for the `SHAMapItem` buffer and we need that data to be _contiguous_ to the `SHAMapItem` itself, and there's no mechanism to allow `std::make_shared` to do that either.

It's true that `intrusive_ptr_release` is not trivial, but it's also not overly complex either. It basically does what `std::shared_ptr` does, only that is abstracted away and hidden from your view. And that does matter in many cases, but in _this_ case, it doesn't matter enough, at least IMO.

src/ripple/shamap/SHAMapItem.h
nbougalis: Yeah, that was left over from some testing I was conducting. I'll remove it.

src/ripple/shamap/SHAMapItem.h
nbougalis: That's not what's happening here: I am checking whether the pointer to the newly allocated memory is properly aligned for a `SHAMapItem`. It should always be (the code is carefully implemented to ensure that) but I want to be _extra_ careful, because of the placement `new` operation below, which uses [`std::assume_aligned`](https://en.cppreference.com/w/cpp/memory/assume_aligned).

src/ripple/basics/SlabAllocator.h
seelabs: Another option is `acquire/release`; I suspect you won't see any perf difference between `relaxed` and `acquire/release` (at least on intel platforms). If that's true, that would probably be my vote. As per the rust book referenced above `seq_cst` is almost never needed, and we don't need it in this case. If I'm wrong and there _is_ a perf difference, then we can look at `relaxed`.

src/ripple/shamap/SHAMapItem.h
ckeshava: Okay, I get that. But why do you choose to type-cast `raw` into `std::uintptr_t` ? 

Would it make a difference if you used `std::intptr_t` (or another pointer type) instead of `std::uintptr_t` ?

src/ripple/shamap/SHAMapItem.h
ckeshava: I see, okay.
Appropos the second bullet point: 
```
... But that's not enough: we need more data for the SHAMapItem buffer and we need that data to be contiguous to the SHAMapItem itself ...
```

Why do we need more contiguous buffer for the SHAMapItem? Is it required to store the next pointer of linked list of items? (as you explained elsewhere in the comments) ?

nbougalis: No, it's required to store the "contents" of the `SHAMapItem`. As I explained, previously a `SHAMapItem` contained a `std::vector` or a `Buffer` as a member (don't remember which off-hand) and that object required a separate allocation.

The new code allocates the space for the `SHAMapItem` and its dynamically-sized contents in one go, so that the contents are located immediately after the `SHAMapItem` in memory. For more details see commit c3acbce82dc5db8765cfc9dfeae36954cec0312f and checkout the changes in `SHAMapItem.h` and `SHAMapItem.cpp`.

src/ripple/basics/SlabAllocator.h
```diff
@@ -30,67 +30,176 @@
...
+            // It's possible that this block has an underlying buffer. If
+            // it does, and it has space available, we use it.
+            if (data_ != nullptr)
             {
-                // Use memcpy to avoid unaligned UB
-                // (will optimize to equivalent code)
-                std::memcpy(data, &l_, sizeof(std::uint8_t*));
-                l_ = data;
-                data += item;
+                auto ret = list_;
```
seelabs, line 182: Technically I think we need to run a destructor on `*list_` or this is UB. I don't think it actually matters in practice, but we may as well be technically correct here, and it's an easy change.

src/ripple/basics/SlabAllocator.h
```diff
@@ -30,67 +30,176 @@
...
+
+                if (ret != nullptr)
+                {
+                    list_ = ret->next;
+                    --avail_;
+                }
+
+                return reinterpret_cast<std::uint8_t*>(ret);
```
seelabs, line 190: We _may_ need `std::launder` here. I'll admit I don't actually know for sure, I'm trying to read the `std::launder` docs to know for sure.

seelabs (commented):
A couple comments on some possible UB behavior (that I don't think we'd ever see in practice; but I still tagged them).

src/ripple/basics/SlabAllocator.h
nbougalis: The destructor is trivial (see [`std::is_trivially_destructible`](https://en.cppreference.com/w/cpp/types/is_destructible)) which, if I understand things correctly, means that the destructor does nothing and so not to not explicitly invoke it  is fine.

With that said, I'm perfectly fine with adding a call to `std::destroy_at(ret)` here and it might make sense to eliminate any confusion, even if it's technically correct, which is the best kind of correct.

Note that there is a somewhat analogous situation in `SHAMapItem` (which I introduced) that it may make sense to also address by removing the `if constexpr` check and calling `std::destroy_at` unconditionally:

https://github.com/XRPLF/rippled/blob/c3acbce82dc5db8765cfc9dfeae36954cec0312f/src/ripple/shamap/SHAMapItem.h#L146-L150

seelabs: I think it's UB if we don't run the destructor even if it's trivailly destructable. Yes, I agree it won't matter in practice, but I think we should do it (unless someone can correct me and show it's not UB).

src/ripple/basics/SlabAllocator.h
nbougalis: I don't think we do, but would welcome more conversation on the topic.

Given:

```
template< class T >
constexpr T* launder( T* p ) noexcept;
```

The documentation states:

> Formally, given
>
>* the pointer `p` represents the address `A` of a byte in memory
>* an object `x` is located at the address `A`
>* `x` is within its [lifetime](https://en.cppreference.com/w/cpp/language/lifetime)
>* the type of `x` is the same as `T`, ignoring cv-qualifiers at every level
>* every byte that would be reachable through the result is reachable through `p` (bytes are reachable through a pointer that points to an object `y` if those bytes are within the storage of an object `z` that is [pointer-interconvertible](https://en.cppreference.com/w/cpp/language/static_cast#pointer-interconvertible) with `y`, or within the immediately enclosing array of which `z` is an element). 
>
>Then `std::launder(p)` returns a value of type `T*` that points to the object `x`.

I'll admit that I find all this _really_ confusing, and wish someone would do the ELI5 thing.

But, my read is that in this case `p` is a `FreeItem`, and so `std::launder` can only give us _that_ back. Also, a `FreeItem` is likely to be smaller in size than whatever the new memory is used for by the caller, so I'm not sure that the reachability analysis is satisfied either.

Given this description of a use case from the docs: "[o]btaining a pointer to an object created by placement new from a pointer to an object providing storage for that object" I do think that there's an argument to be made that whoever invokes placement new to construct an object in the returned buffer should pass the result of `std::launder` but at the same time I'm confused about how to do that without violating one of the above requirements and, therefore, invoking UB.

Frankly, I think that, as is, `std::launder` is nigh unusable.

src/ripple/basics/SlabAllocator.h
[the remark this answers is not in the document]
ckeshava: okay, that's true. thanks for the explanation 👍

src/ripple/basics/SlabAllocator.h
HowardHinnant: I unfortunately agree that we have to run the destructor even if trivial.  The optimizers have gotten too scary when it comes to UB.

nbougalis: Interesting. I'll update the code to run the destructor here and update the code in `SHAMapItem`. Thanks @HowardHinnant!

nbougalis:
I have pushed an update that (a) ensures that the constructor for `FreeItem` is always called, even if trivial; and (b) switches from `std::memory_order::relaxed` to `std::memory_order::acquire` and `std::memory_order::release` semantics for `slabs_` which benchmarking suggests has no real overhead, at least on my x86-64 setup; ymmv.

src/ripple/basics/SlabAllocator.h
```diff
@@ -30,67 +30,176 @@
...
-                // Use memcpy to avoid unaligned UB
-                // (will optimize to equivalent code)
-                std::memcpy(data, &l_, sizeof(std::uint8_t*));
-                l_ = data;
-                data += item;
+                auto ret = list_;
+
+                if (ret != nullptr)
+                {
+                    list_ = ret->next;
```
seelabs, line 186: I think we need a destructor for `ret` here.

src/ripple/basics/SlabAllocator.h
```diff
@@ -214,95 +345,102 @@ class SlabAllocator
         @return a pointer to a block of memory from the allocator, or
                 nullptr if the allocator can't satisfy this request.
      */
-    std::uint8_t*
+    [[nodiscard]] std::uint8_t*
     allocate() noexcept
     {
-        auto slab = slabs_.load();
+        auto active = active_.load(std::memory_order::relaxed);
```
seelabs, line 351: May as well use acquire/release semantics for `active_` too. To be clear, I'll approve with `relaxed`, but on intel there won't be a perf difference so we may as well use the stronger semantics.

src/ripple/shamap/SHAMapItem.h
ckeshava: okay, this makes sense 👍

src/ripple/basics/SlabAllocator.h
```diff
@@ -100,91 +209,115 @@ class SlabAllocator
...
-        }
-    };
+            // It looks like this objects belongs to us. Ensure that it
+            // doesn't extend past the end of our buffer:
+            assert(ptr + itemSize_ <= slab->data_ + slabSize_);
 
-private:
-    // A linked list of slabs
-    std::atomic<SlabBlock*> slabs_ = nullptr;
+            slab->list_ = new (std::assume_aligned<alignof(FreeItem)>(ptr))
```
ckeshava, line 278: isn't it necessary to assert that `ptr` is aligned correctly? I believe `ptr` can only be `SHAMapItem*` as invoked from here: https://github.com/nbougalis/rippled/blob/a4e27672a29d39e26fedd1e074b68b71ef6b2ebd/src/ripple/shamap/SHAMapItem.h#L158

Since `Type` is aligned at 4 or 8 byte boundaries, I feel this should be fine, just wanted to draw your attention for confirmation.

src/ripple/basics/SlabAllocator.h
```diff
@@ -214,95 +345,102 @@ class SlabAllocator
...
+             slab != nullptr;
+             slab = slab->next_)
+        {
+            if (auto ret = slab->assign_and_allocate(
+                    buf, slabSize_, itemSize_, itemCount_))
+            {
+                // This slab had available space and we used it. The block
+                // of memory we allocated is no longer needed.
+                if (ret != buf)
+                    boost::alignment::aligned_free(buf);
```
ckeshava, line 393: I'm uncomfortable with this extraneous allocation and free'ing up of the `buf` 's memory.

We could split up `assign_and_allocate` into two functions -- split up at https://github.com/nbougalis/rippled/blob/a4e27672a29d39e26fedd1e074b68b71ef6b2ebd/src/ripple/basics/SlabAllocator.h#L192

This would differentiate between reusing existing buffers versus using the new buffer.

We could do a first pass over `slabs_`, if there is an existing free buffer, that can be used. Otherwise, we can proceed with the allocation for `buf`.

Do you think the benefits is worth the refactor?

src/ripple/basics/SlabAllocator.h
nbougalis: Yes, it will be correctly aligned, which is why `std::assume_aligned` is used here. Per the documentation:

>[`std::assume_aligned`] informs the implementation that the object `ptr` points to is aligned to at least `N`. The implementation may use this information to generate more efficient code, but it might only make this assumption if the object is accessed via the return value of `assume_aligned`.

So, we are _communicating_ information to the compiler here that allows it to potentially make better decisions since, by using placement new, it doesn't have control of the alignment of the returned block and so it can't really be sure it's aligned.

That said, it's almost certain that using `std::assume_aligned` doesn't do much and could be removed, but I think it's better to have it, especially considering the pain we go to properly align things.

src/ripple/basics/SlabAllocator.h
nbougalis: > We could split up `assign_and_allocate` into two functions -- split up at https://github.com/nbougalis/rippled/blob/a4e27672a29d39e26fedd1e074b68b71ef6b2ebd/src/ripple/basics/SlabAllocator.h#L192
> 
> This would differentiate between reusing existing buffers versus using the new buffer.
> 
> We could do a first pass over `slabs_`, if there is an existing free buffer, that can be used. Otherwise, we can proceed with the allocation for `buf`.

The iteration over `slabs_` is lock-free: that is, multiple threads can be iterating the list at the same time and, potentially, multiple threads might select the _same_ slab and try to give it a new buffer.

"Ah, don't do that!" you might say. But in order to **not** do that we'd either need to make the iteration require a lock, which will have a significant and *unnecessary* performance impact.

And locking a single slab isn't much better. First, if we lock some slab X and then allocate the memory with the slab lock held, in order to assign it, every other allocation and deallocation attempt might stall, waiting for that slab lock to be released. And _sure_, we could fix this by doing a `try_lock` and skipping a slab that's locked, but that's also additional complexity and it would only work for allocations, not deallocations.

Remember too that the system may take its sweet time to actually allocate memory. You're doing a system call, after all, and it may require a lot of work before it can be serviced.

The advantage of the current setup is that the work happens _without any locks_ and so, other memory allocation/deallocation attempts that use that slab will proceed independently at full speed.


> Do you think the benefits is worth the refactor?

I don't think that the benefits are worth the refactor; in fact, I think the net result will be a pessimization with performance implications.

Which brings us back to the opening sentence in your post:

> I'm uncomfortable with this extraneous allocation and free'ing up of the `buf` 's memory.

Why? What's making you uncomfortable?

src/ripple/basics/SlabAllocator.h
ckeshava: In the Notes section of the cpp-reference: 

> It is up to the program to ensure that the alignment assumption actually holds. A call to assume_aligned does not cause the compiler to verify or enforce this.

`ptr` is obtained as an input to this function. If `ptr` is not aligned to `alignof(FreeItem)`, we could be breaking things. Can we be a 100% sure that `ptr` adheres to this alignment requirements? 

Would an assert like this make the code more robust?
`assert( reinterpret_cast<std::uintptr_t>(ptr) %
                    alignof(FreeItem) ==
                0);`

src/ripple/basics/SlabAllocator.h
ckeshava: Thanks for the explanation about the lock-free code and its challenges, that was helpful.




> The advantage of the current setup is that the work happens _without any locks_ and so, other memory allocation/deallocation attempts that use that slab will proceed independently at full speed.
> 
To clarify, `assign_annd_allocate` function can be entered by one-thread only (https://github.com/nbougalis/rippled/blob/17f88fd151f9810693f1a5e3c41723bbd6005fbd/src/ripple/basics/SlabAllocator.h#L172)

> Why? What's making you uncomfortable?

> Remember too that the system may take its sweet time to actually allocate memory. You're doing a system call, after all, and it may require a lot of work before it can be serviced.

We are performing a memory allocation at https://github.com/nbougalis/rippled/blob/17f88fd151f9810693f1a5e3c41723bbd6005fbd/src/ripple/basics/SlabAllocator.h#L376

In one of the code paths (if a slab is found with an empty slot), that memory is not given back to the OS without any usage here: https://github.com/nbougalis/rippled/blob/17f88fd151f9810693f1a5e3c41723bbd6005fbd/src/ripple/basics/SlabAllocator.h#L394

I'm wondering if we can eliminate the unnecessary memory allocation and free system calls. Can we allocate memory only when we are absolutely sure that we will use it?

src/ripple/basics/SlabAllocator.h
```diff
@@ -30,67 +30,177 @@
...
+    //           | |      |    | |
+    //           | |      '----' |
+    //           | '---> NULL    |
+    //           |               |
+    //           '---------------'
+    //
+    // Note that the order of slots within the free list at any given time
+    // is random, as it depends on the pattern and timing of memory access
+    // of the application.
+    struct alignas(Type) FreeItem
```
seelabs, line 115: I'm seeing a bunch of warnings of the form: 
```
src/ripple/basics/SlabAllocator.h:115:12: error: requested alignment is less than minimum alignment of 8 for type 'ripple::SlabAllocator<ripple::SHAMapItem>::FreeItem'
```

src/ripple/basics/SlabAllocator.h
```diff
@@ -30,67 +30,177 @@
...
+        FreeItem* next;
+
+        FreeItem(FreeItem* n = nullptr) noexcept : next(n)
+        {
+        }
+
+        ~FreeItem() noexcept = default;
+    };
+
+    static_assert(sizeof(Type) >= sizeof(FreeItem));
```
seelabs, line 126: Should we also assert that alignof(Type) >= alignof(FreeItem)? If I do that it's failing on clang.

src/ripple/shamap/SHAMapItem.h
```diff
@@ -139,46 +143,49 @@ intrusive_ptr_add_ref(SHAMapItem const* x)
...
     // invariant.
-    return {new (raw) SHAMapItem{tag, data}, false};
+    return {
+        new (std::assume_aligned<alignof(SHAMapItem)>(raw))
+            SHAMapItem{tag, data},
+        false};
 }
 
-static_assert(alignof(SHAMapItem) != 40);
-static_assert(alignof(SHAMapItem) == 8 || alignof(SHAMapItem) == 4);
```
ckeshava, line 179: @seelabs  There was an assert pertaining to alignment in a previous commit, but I don't think it helps solve the clang warning.

src/ripple/basics/SlabAllocator.h
seelabs: You know, between the uglyness with the destructors and the alignment issues, I wonder if we're better off without the `FreeItem` class after all.

ckeshava: How do you propose to store the `next` pointers without `FreeItem` class?

seelabs: `memcpy` or some function that wraps `memcpy` to make it clearer we're dealing with a linked list. A previous patch did that.

src/ripple/basics/SlabAllocator.h
nbougalis: >In one of the code paths (if a slab is found with an empty slot), that memory is not given back to the OS without any usage here:

Huh? I'm sorry but you seem confused. Let's walk through this:

We allocate some memory, and we begin iterating the slabs.

If a slab is found with a free slot, we use that slot and *release* the memory we just allocated (since it's not needed).

If a slab is found without a free slot _and_ without a memory buffer, we give it the memory we just allocated, and it gives us back a slot (it's guaranteed to have one now!)

If we run out of slabs, we create a _new_ slab and give it the memory we just allocated.

In all cases, either the memory is used, or it's properly free. There are no leaks.

>I'm wondering if we can eliminate the unnecessary memory allocation and free system calls. Can we allocate memory only when we are absolutely sure that we will use it?

No, for the reasons I explained previously. If you want to perform the allocation only after you're sure you need to, you must hold the `std::mutex` of the slab you intend to use, which is what we're trying to avoid.

src/ripple/basics/SlabAllocator.h
nbougalis: Yeah, this was silly of me: on LP64 systems, pointers will end up being aligned on 8 byte boundaries, but `SHAMapItem` is perfectly content with 4 byte alignment.

I _think_ a reasonable solution is to just use an appropriate `alignas()` for `SHAMapItem` (which we can easily check in the slab allocator with a `static_assert`) because the code is a _LOT_ clearer using `FreeItem` instead of the weird/confusing `std::memcpy` syntax.

That said, I'm not opposed to that either.

src/ripple/shamap/SHAMapItem.h
nbougalis: It won't. The problem is that on an LP64 system, pointers are aligned on 8 byte boundaries, and `FreeItem` contains a pointer so its natural alignment is 8 bytes, but `SHAMapItem` doesn't contain any pointers, and its natural alignment is 4 bytes.

src/ripple/basics/SlabAllocator.h
HowardHinnant: I just repro'd these on macOS clang 15 as well.

PR #4674, opened 2023-08-24
AMMBid: use tecINTERNAL for 'impossible' errors

mDuo13:
Modifies two error cases in AMMBid to return `tecINTERNAL` rather than another error code, to more clearly indicate that these types of errors should not be possible unless operating in some sort of unforeseen circumstances. Upgrade the related log message to warning level, since it likely indicates a bug.

**Caution:** I have not actually built this branch myself. Reviewers, please confirm that the code compiles and passes tests.

Modifies two specific transaction error cases in the processing for the AMMBid transactor:

- `tecAMM_BALANCE` - In this case, this error (total LP Tokens outstanding is lower than the amount to be burned for the bid) is a subset of the case where the user doesn't have enough LP Tokens to pay for the bid. When this case is reached, the bidder's LP Tokens balance has already been checked first. The user's LP Tokens should always be a subset of total LP Tokens issued, so this should be impossible.
- `tecINSUFFICIENT_PAYMENT` - In this case, the amount to be refunded as a result of the bid is greater than the price paid for the auction slot. This should never occur unless something is wrong with the math for calculating the refund amount.

### Context of Change

Both error cases in question are "defense in depth" measures meant to protect against making things worse if the code has already reached a state that is supposed to be impossible, likely due to a bug elsewhere. (For example, some sort of race condition where multiple threads are accessing the same ledger entries in parallel without proper locks, simple memory corruption, or some sort of math failure elsewhere—rounding errors, missing range checks, and so on.) 

If these cases have "normal" error codes, that suggests that QA should be able to add test cases that intentionally trigger them, and if they occur then the code is functioning as intended. But, in reality, it should be impossible for an integration test to trigger this case, and if you can trigger these reliably it's indicative of a bug that should be fixed elsewhere. (It's possible that this sort of thing could occur randomly due to, say, cosmic rays flipping bits at random, in which case reproducing the error is likely to be, well, very difficult.)

Any "shouldn't ever occur" checks should use an error code that categorically indicates a larger problem. This is similar to how `tecINVARIANT_FAILED` is a warning sign that something went wrong and likely could've been worse, but since there isn't a separate-level Invariant Check applying here, `tecINTERNAL` is the appropriate error code. The [official description for `tecINTERNAL` in the docs](https://xrpl.org/tec-codes.html) is:

> This error code should not normally be returned. If you can reproduce this error, please [report an issue](https://github.com/ripple/rippled/issues).

### Type of Change

This is "debatably" a transaction processing change since it could hypothetically change how transactions are processed if there's a bug we don't know about, but as far as I know it doesn't actually change anything that can actually happen. Regardless, as long as the changes are incorporated into the AMM amendment _before_ said amendment is part of any stable release, no additional amendment is necessary.


mDuo13:
It seems I don't know how to run clang-format. I tried, but it didn't work. ¯\_(ツ)_/¯

gregtatcam (approved):
👍 LGTM. Builds/passes the unit-tests. Approved, assuming the clang format errors are going to be fixed.
This works for me on OSX. Run from `rippled` folder: clang-format -style=format -i src/ripple/app/tx/impl/AMMBid.cpp. clang-format version is 15.0.7.

intelliot:
I don't have permission to push to your branch, but I applied the clang-format patch and pushed the commit here: https://github.com/intelliot/rippled/commit/925552f080cb352524cf54c0fda6859634891cfa

Feel free to merge or cherry-pick it

ximinez:
Just a quick comment before I dig into this, I think it would be more accurate to log those messages as "error".

ximinez:
Just a quick comment before I dig into this, I think it would be more accurate to log those messages as "error".

intelliot:
notes:
- not sure why this PR doesn't show for me on https://github.com/XRPLF/rippled/milestone/7 - a bug in github, I guess.
- this impacts the [AMM amendment](https://xrpl.org/known-amendments.html#amm) but the change is small and benign

intelliot:
note: this is arguably an API change but it only affects the result code from `submit` / transaction processing.

intelliot:
note: no impact on Clio

ximinez (changes requested):
Other than the thing about the log level, this looks fine.

intelliot:
notes:
- this should log at `fatal` level
- should also update other `tecINTERNAL` logs to be at the same `fatal` level (as long as they are expected to be impossible)

mDuo13:
As for changing to log levels to fatal in other cases where we return `tecINTERNAL` because the error should be impossible—I'm leaving that to someone else in another PR. Maybe @ximinez?

intelliot:
Yes, that should be a separate PR. No urgency, so I'm content to leave it as an open issue for now. Maybe a `Good First Issue` for a new contributor, even.

intelliot:
@ximinez - please review this as well

intelliot:
This is a "tiny" PR and review by @seelabs would be welcome, but not required.

ximinez (approved):
Looks good. Just running through my Windows-based checks. I can't imagine any reason they'd fail.

intelliot:
> Just running through my Windows-based checks.

If those do turn up anything, feel free to open a new issue/PR. This one will be merged shortly...

PR #4676, opened 2023-08-27
Fix asan stack use after scope (fix #4675)

### LINKED ISSUE #4675: ASAN stack-use-after-scope in soci::use with rvalues (Version: [1.12.0-rc1])
## Issue Description
We have stack-use-after-scope bugs since we're using ravlues with `soci::use`

## Steps to Reproduce
Build rippled with -Dsan=address, start a fresh Ledger server and trigger e.g. peer_reservations_add from a second instance. You may need to use a rather recent compiler for asan to catch most stack-use-after-scope errors.
```
mkdir build
cd build
conan install .. --output-folder . --build missing --settings build_type=Debug -o reporting=True
cmake -DCMAKE_TOOLCHAIN_FILE:FILEPATH=build/generators/conan_toolchain.cmake -DCMAKE_BUILD_TYPE=Debug -Dsan=address -Dreporting=ON -DCMAKE_C_COMPILER=clang-15 -DCMAKE_CXX_COMPILER=clang++-15 -DCMAKE_EXPORT_COMPILE_COMMANDS=ON ..
cmake --build . -j 12
./rippled-reporting --start -a
./rippled-reporting peer_reservations_add n9Mxf6qD4J55XeLSCEpqaePW4GjoCR5U1ZeGZGJUCNe3bQa4yQbG 
```

## Actual Result
ASAN detects stack use after scope which at least for `insertPeerReservation `and `deletePeerReservation `means we are accessing a std::string which already has been deleted. In `updateLedgerDBs` this may be more of an academic issue. But for the peer reservations we could actually get the nodeId overwritten by a different thread while we're reading it since the actual string content is on the heap.

## Environment
My Dockerfile:
```
FROM ubuntu:22.04
RUN apt-get update
RUN apt-get -y upgrade
RUN apt-get -y install git scons ninja-build clang
RUN DEBIAN_FRONTEND=noninteractive apt-get -yq install pkg-config
RUN apt-get -y install protobuf-compiler libprotobuf-dev libssl-dev libboost-all-dev python-is-python3 cmake clang-15 clangd-15 g++ pip dotnet-runtime-6.0 curl clang-format-11
RUN rm /usr/bin/clang && rm /usr/bin/clang++ && ln -s /usr/bin/clangd-15 /usr/bin/clangd && ln -s /usr/bin/clang-15 /usr/bin/clang && ln -s /usr/bin/clang++-15 /usr/bin/clang++ && ln -s /usr/bin/clang-format-11 /usr/bin/clang-format
ENV CC=clang  
ENV CXX=clang++
RUN pip install conan==1.60.2 && conan profile new default --detect && conan profile update settings.compiler.cppstd=20 default && conan profile update settings.compiler.libcxx=libstdc++11 default
```
```
git rev-parse HEAD
300b7e078a4bc511f30b74509d416e5081ec3650
```
## Supporting Files
See also https://github.com/SOCI/soci/blob/master/docs/binding.md#object-lifetime-and-immutability and https://github.com/SOCI/soci/issues/1070

[server log with ASAN error](https://gist.github.com/StefanVK/85abcc665cd29cd02db41071d7405485)
My rippled.cfg equals rippled-example.cfg and validators.txt equals validators-example.txt

StefanVK:
Fix asan stack-use-after-scope: https://github.com/XRPLF/rippled/issues/4675

Replace ravlues with lvalues for calls to `soci::use` to have the scope go beyond the end of the expression.

### Context of Change

For `insertPeerReservation` and `deletePeerReservation` we could be reading nodeIds that get overwritten while we're reading them.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

I built rippled with asan with and without the fix. Without the fix asan finds stack-use-after-scope with those commands
```
./rippled-reporting --start -a
./rippled-reporting peer_reservations_add n9Mxf6qD4J55XeLSCEpqaePW4GjoCR5U1ZeGZGJUCNe3bQa4yQbG 
```
After the fix, asan no longer reports an issue.


ckeshava:
Hello @StefanVK  thanks for pointing this out. On which platform/environment did you observe this issue?
I don't understand how the rvalues could suffer from being overwritten in the existing code. My understanding it that the rvalue parameters are being copied into the `soci::use()` calls.

StefanVK:
Hi @ckeshava it was on x86_64 with clang-15 with asan. More details are in the issue https://github.com/XRPLF/rippled/issues/4675

Basically what soci ends up doing is taking a reference to the ravlue in a [soci::use_container](https://github.com/SOCI/soci/blob/master/include/soci/use.h) and then the address in a [soci_use_type](https://github.com/SOCI/soci/blob/master/include/soci/use-type.h). It does not copy the value itself and does not extend its lifetime.

The evaluation of the statement is triggered by the destructor of a soci::details::once_temp_type which is the result of operator<< on the [soci::session](https://github.com/SOCI/soci/blob/master/include/soci/session.h). Other temporaries from the same expression can already be beyond its lifetime at this point.

Now for trivial types such as int it is undefined behavior to access it after the lifetime has ended but realistically you won't see the memory being overwritten within the same statement. For a std::string however when the lifetime ends, the non trivial destructor is invoked and the backing heap memory (for strings too large for small string optimization) gets freed and could be reused by a different thread (depending on the allocator) or another non trivial destructor in the same thread for another object whose lifetime ends in the same statement (not the case for rippled).

I did not observe the memory being overwritten, I only observed asan pointing out that memory is being accessed beyond its lifetime. But I'm convinced that for the std::string case it would be possible for the memory to be overwritten, or for a segmentation fault to occur. We had the same bug in our codebase and I started looking for the same issue in different codebases on github and found rippled.

ckeshava:
thanks for the detailed explanation @StefanVK It sounds like this is something that needs to be fixed in the soci codebase, but thank you for the fix.

if I want to observe the overwriting of a piece of memory, I'll need to put breakpoints (before and after) and observe certain addresses in the heap right? is there a more efficient way to check this?

ckeshava (approved):

HowardHinnant (approved):

intelliot:
@StefanVK: Can you confirm that, from your perspective, this PR still looks good to merge?

StefanVK:
@intelliot Yes, as far as I'm concerned this is still a correct fix.

intelliot (approved):

PR #4677, opened 2023-08-27
[TRIVIAL] fix clang unused-but-set-variable warning

StefanVK:
Fix clang warning about unused but set variable: removed variable.

### Context of Change
(clang 15)
/workspaces/rippled/src/ripple/app/rdb/backend/impl/PostgresDatabase.cpp:178:14: warning: variable 'expNumResults' set but not used [-Wunused-but-set-variable]
    uint32_t expNumResults = 1;

### Type of Change

- [ ] Documentation Updates
- [ ] Release


scottschurr (approved):
:+1:  LGTM.  Thanks for the contribution!

src/ripple/app/rdb/backend/impl/PostgresDatabase.cpp
```diff
@@ -189,8 +187,6 @@ loadLedgerInfos(
         auto minAndMax =
             std::get_if<std::pair<uint32_t, uint32_t>>(&whichLedger))
     {
-        expNumResults = minAndMax->second - minAndMax->first;
```
ckeshava, line 192: Is it useful to assert the validity of the output of the Postgres database?
`expNumResults` could be set to an initial absurd value (say `-1`), later we can do:
```
if(expNumResults != -1)
    assert(expNumResults == res.ntuples());
```

Similar to the validation at https://github.com/XRPLF/rippled/blob/3c321f54cc03575f8f85fc1fa73737134f0ada09/src/ripple/app/rdb/backend/impl/PostgresDatabase.cpp#L223

Is this useful?

intelliot:
Hi @StefanVK - could you address the comment above?

StefanVK:
@ckeshava and @intelliot 
I'm not sure I'm the right person to answer this since I don't know the invariants of the database nor the intended preconditions of the function. I only suggested an obviously behavior preserving fix to a compiler warning here.

I myself would not feel comfortable adding that assert since it's not obvious to me that it's a failure to call this function with a range that exceeds the ledger range in the database. Returning whichever subset of LedgerInfos we have in the database seems like reasonable behavior to me. It's not clear to me that calling the function with a range that is not completely in our database is bug in the calling code and we can and do handle the case gracefully in loadLedgerInfos.

If you can argue that it is indeed always a bug to call loadLedgeInfos with a ledgerRange or seq that is not or not fully contained in our database, go ahead and add an assert. I'm not saying it's wrong to add the assert, just that I don't know enough about the function to feel comfortable adding it.

ckeshava:
Hello,
I have a couple of thoughts. 
1. [loadLedgerHelper](https://github.com/XRPLF/rippled/blob/397268394b80574419fddafe77aa75ea11ad9cfe/src/ripple/app/rdb/backend/impl/PostgresDatabase.cpp#L304) does not accurately reflect the input-possibilities for the `whichLedger` parameter. Although [loadLedgerInfo]() supports a pair of minLedgerSeq and maxLedgerSeq as inputs, that is not reflected in the `loadLedgerHelper` function signature. Hence, validation at `loadLedgerHelper` is not exhaustive.

2. [PostgresDatabaseImp::getHashesByIndex](https://github.com/XRPLF/rippled/blob/397268394b80574419fddafe77aa75ea11ad9cfe/src/ripple/app/rdb/backend/impl/PostgresDatabase.cpp#L656) make use of the above-mentioned pair of indices as input. But it does not perform sanity checks on the output. Not all the downstream callers of this function perform the sanity checks [here](https://github.com/XRPLF/rippled/blob/397268394b80574419fddafe77aa75ea11ad9cfe/src/ripple/nodestore/impl/DatabaseShardImp.cpp#L927) and [here](https://github.com/XRPLF/rippled/blob/397268394b80574419fddafe77aa75ea11ad9cfe/src/ripple/app/ledger/impl/LedgerMaster.cpp#L750)

3. Suppose the user asks for ledgers between sequence numbers 100 to 200 -- and `rippled` returns only 87 such ledgers, due to lack of relevant data. Is this situation possible? Does `rippled` have any fail-safe in this scenario? If not, I'd recommend adding this assert. 

I have written the suggested changes at the tip of this [branch](https://github.com/ckeshava/rippled/tree/pgAssert)

@scottschurr I'm not familiar with the databases used it `rippled`. What do you think?

scottschurr:
@ckeshava, I get the feeling that your questions are expanding far beyond the boundaries of the initial change request.  All of the suggested changes are inside of a
```
#ifdef RIPPLED_REPORTING
```
conditional macro block.  Any questions regarding Postgres should be addressed to @mtrippled, who developed reporting mode.

However, fundamentally, this pull request is about a C++ variable that ...

1. Is initialized, and
2. Is possibly assigned to later, but
3. Is never read from.

It is possible that the lack of reads of the variable points out a bug in the code.  Only @mtrippled is qualified to answer that question.  But, lacking that bug, removing the local variable `expNumResults` is a sensible call.

@mtrippled, could you look at the code in question and decide whether you want `expNumResults` removed, or if you think that value should be used later down in the code?  Thanks.

mtrippled (approved):
Looks good. :+1:

mtrippled:
@ckeshava What's the answer to your 3rd question?

ckeshava:
@mtrippled I don't know the answer myself, I wanted to verify if that situation is possible. If we don't have all the relevant data in the local database, does `rippled` attempt to fetch the appropriate ledgers from its peers? I vaguely remember FLR and synchronization deal with this problem, but I don't know

mtrippled:
@ckeshava I don't have the answer either. But as Scott alluded, before killing the process, make sure of a couple of things first:
1) it always represents an error, and especially
2) killing the process is the best way to handle the error

Generally, criteria 2 is only met if the alternative is incorrect data being persisted or returned to the user.

ckeshava:
okay, I'll dig into it

ckeshava:
As it stands, I agree with the changes and it is okay to merge it. 
I believe I can improve this portion of the codebase, I'll look into it after I'm done with the current PRs and open a separate refactor PR. 👍

ckeshava (approved):

PR #4682, opened 2023-08-30
Update amm_info to fetch AMM by amm account and add owner directory for AMM object

gregtatcam:

Add back the owner directory, which was deleted by #4626. 
Update amm_info to fetch AMM object by amm account id instead of the assets pair.

### Context of Change

#4626 removed the AMM object directory entry and added AMMID. This was done for efficiency consideration because it could take up to 32 iterations to get AMM object. This change broke `account_objects` for `amm` type. This PR adds back the owner directory and also updates `amm_info` to fetch AMM object via amm account id instead of the asset pair.

### Type of Change

## Before / After

`account_objects` works with `amm` type. `amm_info` can be used with either the asset pair or amm account id to retrieve AMM object.

## Test Plan

AccountObjects is extended for `amm` type tests.
AMMInfo is extended to test for `amm_info` with amm account id.


intelliot:
I think this should _also_ be considered "Breaking change (fix or feature that would cause existing functionality to not work as expected)" because it changes the behavior of the AMM amendment, even if only in a small way.

src/ripple/app/misc/impl/AMMUtils.cpp
```diff
@@ -201,6 +201,9 @@ deleteAMMTrustLines(
         [&](LedgerEntryType nodeType,
             uint256 const&,
             std::shared_ptr<SLE>& sleItem) -> TER {
+            // Skip AMM
+            if (nodeType == LedgerEntryType::ltAMM)
+                return tecOBJECT_NOT_FOUND;
```
ximinez, line 206: I'm not a huge fan of using `tecOBJECT_NOT_FOUND`. I think it would make more sense to find (or even create) a `ter` code that serves this purpose. Even though it'll never be exposed externally, even internally, I'd rather see something that tells me this object will be retried later. Maybe `terNO_AMM`? Or even just the base `terRETRY` (it has the advantage of not being used anywhere else).

ximinez (changes requested):
Incremental "first impressions" review.

src/ripple/protocol/jss.h
```diff
@@ -160,6 +160,7 @@ JSS(alternatives);           // out: PathRequest, RipplePathFind
 JSS(amendment_blocked);      // out: NetworkOPs
 JSS(amendments);             // in: AccountObjects, out: NetworkOPs
 JSS(amm);                    // out: amm_info
+JSS(amm_account);            // in: amm_info
```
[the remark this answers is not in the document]
gregtatcam: I thought that was pretty clear. But I'm not good with names. @mDuo13 what do you think about the name change?

src/ripple/app/misc/impl/AMMUtils.cpp
gregtatcam: I actually had `terRETRY` at first. Thanks, will change.

seelabs: You're abusing ter codes to mean things they don't mean. `tecOBJECT_NOT_FOUND` is no good (the object is found), and `terRETRY` doesn't mean "skip this entry". I don't think we can use a ter code to mean "skip this entry". Consider changing the callback function to return both the ter and an enum that says to skip or not. There are other interface changes we can consider too, but I abusing ter codes doesn't seem like the right approach here.

src/ripple/app/tx/impl/AMMCreate.cpp
```diff
@@ -279,6 +279,19 @@ applyCreate(
...
+            keylet::ownerDir(*ammAccount),
+            ammSle->key(),
+            describeOwnerDir(*ammAccount));
+        !page)
+    {
+        JLOG(j_.debug()) << "AMM Instance: failed to insert owner dir";
+        return {tecDIR_FULL, false};
+    }
+    else
+        ammSle->setFieldU64(sfOwnerNode, *page);
```
seelabs, line 294: Consider:
```
    if (auto const page = sb.dirInsert(
            keylet::ownerDir(*ammAccount),
            ammSle->key(),
            describeOwnerDir(*ammAccount)))
    {
        ammSle->setFieldU64(sfOwnerNode, *page);
    }
    else
    {
        JLOG(j_.debug()) << "AMM Instance: failed to insert owner dir";
        return {tecDIR_FULL, false};
    }
```

src/ripple/ledger/impl/View.cpp
```diff
@@ -1567,8 +1567,8 @@ cleanupOnAccountDelete(
 
             // Deleter handles the details of specific account-owned object
             // deletion
-            if (auto const ter = deleter(nodeType, dirEntry, sleItem);
-                ter != tesSUCCESS)
+            auto const ter = deleter(nodeType, dirEntry, sleItem);
+            if (ter != tesSUCCESS && ter != tecOBJECT_NOT_FOUND)
```
seelabs, line 1571: See previous comment. I think we're abusing ter codes here.

src/ripple/ledger/impl/View.cpp
```diff
@@ -1580,21 +1580,22 @@ cleanupOnAccountDelete(
...
+            //     back to 'it' to "un-invalidate" the iterator.
+            assert(uDirEntry >= 1);
+            if (uDirEntry == 0)
             {
                 JLOG(j.error())
                     << "DeleteAccount iterator re-validation failed.";
                 return tefBAD_LEDGER;
             }
-            uDirEntry = 0;
+            if (ter == tesSUCCESS)
```
seelabs, line 1597: We really mean "don't skip" not "success"

src/ripple/rpc/handlers/AMMInfo.cpp
```diff
@@ -78,32 +78,54 @@ doAMMInfo(RPC::JsonContext& context)
 
     Issue issue1;
```
seelabs, line 79: `Issue` probably shouldn't have a default constructor at all. At any rate, consider making this a `std::optional` until it has a valid value.

src/ripple/rpc/handlers/AMMInfo.cpp
```diff
@@ -78,32 +78,54 @@ doAMMInfo(RPC::JsonContext& context)
...
     Issue issue2;
+    uint256 ammID;
+    bool const isAssets =
+        params.isMember(jss::asset) && params.isMember(jss::asset2);
+    bool const isAMMAccount = params.isMember(jss::amm_account);
```
seelabs, line 84: If the variables are made optional we can remove the `isAssets` and `isAMMAccount` and just check if the optional is set.

src/ripple/rpc/handlers/AMMInfo.cpp
```diff
@@ -78,32 +78,54 @@ doAMMInfo(RPC::JsonContext& context)
...
 
-    if (!params.isMember(jss::asset) || !params.isMember(jss::asset2))
+    if ((isAssets && isAMMAccount) || (!isAssets && !isAMMAccount))
```
seelabs, line 86: You could just check for equality.

src/ripple/rpc/handlers/AMMInfo.cpp
```diff
@@ -78,32 +78,54 @@ doAMMInfo(RPC::JsonContext& context)
...
+
+    if (isAssets)
+    {
+        if (auto const i = getIssue(params[jss::asset], context.j); !i)
+        {
+            RPC::inject_error(i.error(), result);
+            return result;
+        }
+        else
+            issue1 = *i;
```
seelabs, line 105: I see this pattern in this code a lot - where the error condition doesn't use the "if local" variable, and we handle the error condition first. I think this is simpler to read by handling the non-error case in the main branch and removing the `; !i` section. Here and other places as well.

src/ripple/rpc/handlers/AMMInfo.cpp
```diff
@@ -78,32 +78,54 @@ doAMMInfo(RPC::JsonContext& context)
...
+            RPC::inject_error(rpcACT_MALFORMED, result);
+            return result;
+        }
+        else if (auto const sle = ledger->read(keylet::account(*id)); !sle)
+        {
+            RPC::inject_error(rpcACT_MALFORMED, result);
+            return result;
+        }
+        else
+            ammID = sle->getFieldH256(sfAMMID);
```
seelabs, line 127: I think this code is clearer without the cascading local variables declared in the if statements. That pattern is OK if it prevents the declared variables from polluting the parent scope. Here the parent scope can be small.

src/ripple/rpc/handlers/AMMInfo.cpp
```diff
@@ -115,10 +137,19 @@ doAMMInfo(RPC::JsonContext& context)
...
+        return keylet::amm(ammID);
+    }();
     auto const amm = ledger->read(ammKeylet);
     if (!amm)
         return rpcError(rpcACT_NOT_FOUND);
+    if (isAMMAccount)
+    {
+        issue1 = (*amm)[sfAsset];
+        issue2 = (*amm)[sfAsset2];
+    }
```
seelabs, line 152: This whole section is sued to get the `accountID`, `issue1` `issue2` and `amm`. Consider re-writing this with a lambda. Something like this (this incorporates the comments above as well):
```
Json::Value
doAMMInfo(RPC::JsonContext& context)
{
    auto const& params(context.params);
    Json::Value result;

    std::shared_ptr<ReadView const> ledger;
    result = RPC::lookupLedger(ledger, context);
    if (!ledger)
        return result;

    struct ValuesFromContextParams
    {
        std::optional<AccountID> accountID;
        Issue issue1;
        Issue issue2;
        std::shared_ptr<SLE const> amm;
    };

    auto getValuesFromContextParams =
        [&]() -> Expected<ValuesFromContextParams, error_code_i> {
        std::optional<AccountID> accountID;
        std::optional<Issue> issue1;
        std::optional<Issue> issue2;
        std::optional<uint256> ammID;

        if (params.isMember(jss::asset) && params.isMember(jss::asset2))
        {
            if (auto const i = getIssue(params[jss::asset], context.j))
                issue1 = *i;
            else
                return Unexpected(i.error());

            if (auto const i = getIssue(params[jss::asset2], context.j))
                issue2 = *i;
            else
                return Unexpected(i.error());
        }

        if (params.isMember(jss::amm_account))
        {
            auto const id = getAccount(params[jss::amm_account], result);
            if (!id)
                return Unexpected(rpcACT_MALFORMED);
            auto const sle = ledger->read(keylet::account(*id));
            if (!sle)
                return Unexpected(rpcACT_MALFORMED);
            ammID = sle->getFieldH256(sfAMMID);
        }

        if (issue1.has_value() == ammID.has_value())
            return Unexpected(rpcINVALID_PARAMS);

        if (params.isMember(jss::account))
        {
            accountID = getAccount(params[jss::account], result);
            if (!accountID || !ledger->read(keylet::account(*accountID)))
                return Unexpected(rpcACT_MALFORMED);
        }

        auto const ammKeylet = [&]() {
            if (issue1 && issue2)
                return keylet::amm(*issue1, *issue2);
            assert(ammID);
            return keylet::amm(*ammID);
        }();
        auto const amm = ledger->read(ammKeylet);
        if (!amm)
            return Unexpected(rpcACT_NOT_FOUND);
        if (!issue1 && !issue2)
        {
            issue1 = (*amm)[sfAsset];
            issue2 = (*amm)[sfAsset2];
        }

        return ValuesFromContextParams{
            accountID, *issue1, *issue2, std::move(amm)};
    };

    auto r = getValuesFromContextParams();
    if (!r)
    {
        RPC::inject_error(r.error(), result);
        return result;
    }

    auto& [accountID, issue1, issue2, amm] = *r;

```

seelabs (commented):
Left some suggestions about code cleanup. Functionality looks good.

src/ripple/ledger/View.h
```diff
@@ -469,8 +478,7 @@ requireAuth(ReadView const& view, Issue const& issue, AccountID const& account);
 cleanupOnAccountDelete(
     ApplyView& view,
     Keylet const& ownerDirKeylet,
-    std::function<TER(LedgerEntryType, uint256 const&, std::shared_ptr<SLE>&)>
-        deleter,
+    EntryDeleter deleter,
```
seelabs, line 481: This shouldn't be passed by value. Pass by ref instead.

seelabs (commented):
Found one minor nit; other than that LGTM. I'll look at Ed's comment's too, of course.

gregtatcam:
> Found one minor nit; other than that LGTM. I'll look at Ed's comment's too, of course.

Thanks, will fix. I'll wait with pushing another commit until you review Ed's comments.

ximinez (commented):
Other than the ref issue pointed out by @seelabs, and the parameter name question, I think this is good.

src/ripple/protocol/jss.h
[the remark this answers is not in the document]
mDuo13: Both seem fine to me

src/ripple/rpc/handlers/AMMInfo.cpp
```diff
@@ -74,83 +74,89 @@ doAMMInfo(RPC::JsonContext& context)
...
+            auto const sle = ledger->read(keylet::account(*id));
+            if (!sle)
+                return Unexpected(rpcACT_MALFORMED);
+            ammID = sle->getFieldH256(sfAMMID);
         }
-        else if (auto const sle = ledger->read(keylet::account(*id)); !sle)
+
+        if (issue1.has_value() == ammID.has_value())
+            return Unexpected(rpcINVALID_PARAMS);
+
```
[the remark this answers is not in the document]
mDuo13: Yes, that should be an `invalidParams` error.

khancode:
@gregtatcam "Update amm_info to fetch AMM object by amm account id instead of the assets pair."

If I'm understanding this correctly, is amm account id replacing asset pair params in `amm_info` request?

So will this be no longer valid?
```JSON
{
    "command": "amm_info",
    "asset": {
      "currency": "XRP"
    },
    "asset2": {
      "currency": "TST",
      "issuer": "rP9jPyP5kyvFRb6ZiRghAGw5u8SGAmU4bd"
    }
}
```

gregtatcam:
> @gregtatcam "Update amm_info to fetch AMM object by amm account id instead of the assets pair."
> 
> If I'm understanding this correctly, is amm account id replacing asset pair params in `amm_info` request?
> 
> So will this be no longer valid?
> 
> ```json
> {
>     "command": "amm_info",
>     "asset": {
>       "currency": "XRP"
>     },
>     "asset2": {
>       "currency": "TST",
>       "issuer": "rP9jPyP5kyvFRb6ZiRghAGw5u8SGAmU4bd"
>     }
> }
> ```

It doesn't replace the asset pair. Either one but not both can be provided. You can either fetch amm info by `amm_account` or by `asset` and `asset2`.

src/ripple/rpc/handlers/AMMInfo.cpp
```diff
@@ -119,7 +122,8 @@ doAMMInfo(RPC::JsonContext& context)
             ammID = sle->getFieldH256(sfAMMID);
         }
 
-        if (issue1.has_value() == ammID.has_value())
+        if ((issue1.has_value() ^ issue2.has_value()) ||
+            (issue1.has_value() == ammID.has_value()))
```
seelabs, line 126: Since this isn't a bitwise operation, I'd prefer `!=` instead of exclusive or (yes, I understand they act the same here). Also, if issue1 has a value and issue2 does not, that's a logic error, not a malformed error. I don't object to the check, but I'd assert and return `rpcINTERNAL` on that error.

src/ripple/ledger/View.h
```diff
@@ -478,7 +478,7 @@ using EntryDeleter = std::function<std::pair<TER, SkipEntry>(
 cleanupOnAccountDelete(
     ApplyView& view,
     Keylet const& ownerDirKeylet,
-    EntryDeleter deleter,
+    EntryDeleter const& deleter,
```
seelabs, line 481: I probably wouldn't have made this `const`. If the callback wants to change itself (increment some counter say), the interface shouldn't stop that.

src/ripple/rpc/handlers/AMMInfo.cpp
seelabs: Now that I look, we should probably detect when `params.isMember(jss::asset) != params.isMember(jss::asset2)` and return malformed on that too.

src/ripple/ledger/View.h
gregtatcam: It would not compile otherwise.

src/ripple/rpc/handlers/AMMInfo.cpp
gregtatcam: Isn't it invalid parameters if one issue is set but not the other? It means that either `asset` or `asset2` is included but not both. `params.isMember(jss::asset) != params.isMember(jss::asset2)` is handled by `issue1.has_value() != issue2.has_value()`.

seelabs (approved):
:+1:

ximinez (changes requested):

src/ripple/protocol/jss.h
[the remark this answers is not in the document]
gregtatcam: Yes, this is correct.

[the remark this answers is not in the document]
intelliot: Expanding on this for future readers and tech writers / documenters:

- `amm_account` is AMM account ID. It allows users to fetch AMM Object with AMM account ID instead of using the asset pair.
- Either `amm_account` or `asset` and `asset2` have to be included.
- `account` is a liquidity provider (LP) account. If `account` is included, then `lp_token` is set to that LP's token balance.
  - `account` is an `r...` address and is kind of like the "perspective" account.
  - It is like the `peer` in `account_lines`, or the `taker` in `book_offers`.

ximinez (approved):

PR #4685, opened 2023-09-01
Proposed 1.12.0-rc4

intelliot:

Compared to rc3, this version includes the following:
- #4674
- #4682

### Context of Change

In the process of documenting AMM, two "impossible" error cases were identified. These now return `tecINTERNAL` because they should never occur; if they occur, please [open an issue](https://github.com/XRPLF/rippled/issues) to ensure the mechanism of action is investigated and understood.

With the experience of using the AMM APIs, it was determined that it may be useful to retrieve AMM with `account_objects`, so the AMM object owner directory entry was added back in. Additionally, `amm_info` was enhanced to also allow users to retrieve an AMM object by amm account id.

### Type of Change

- [x] Release


injaelee (approved):
✅

PR #4688, opened 2023-09-05
Fix bug with `account_lines` RPC, improve RPC performance & memory behavior JSON

nbougalis:
This commit is based off of PR #4671 (comments to the first commit, which is part of that PR, will be addressed there, and this changeset rebased on top of it, as needed).

It is composed of three commits (which should stay separate);

### Type of Change

- [ ] Documentation Updates
- [ ] Release

#### Fix broken "default ripple" reporting in `account_lines`:

The `TrustLineBase::getDefaultRipple` function, introduced in commit 8116b569c7235fc83d61b68f5b3c697b8042df35 was broken.

The bug is caused by the fact that the `sfFlags` fields of an object are _type specific_; that is, the same flag value may mean different things depending on the _type_ of object.

Here the code would examine the `sfFlags` field of trustline objects and check for the account-specific `lsfDefaultRipple` flag, but that flag actually means `lsfHighFreeze` for trustline objects.

The desired semantics of the original commit are unclear, but likely meant to check the flags of the "view" account. This commit does not attempt to implement that logic, but it should be fixed, so that the rippling flag of the trustline can be correctly interpreted.

#### Improve trustline handling during RPC and pathfinding:

he code used to handle trustlines during pathfinding and RPC handling was suboptimal. This commit improves the code and minimizes its memory footprint (by as much as 50%, from 256 bytes to 128).

Several RPC handlers that previously allocated large amounts of memory for working space and did their work in two steps, first by collecting the trustlines and then generating iterating over them to generate the JSON response, now complete the same work in one step.

These changes result in measurable performance improvements, primarily when traversing owner directories with a large number of entries (e.g. when enumerating the trustlines of the issuer of an asset).

#### Improve `Json::Value` memory allocation for strings:

The memory allocation patterns of Json::Value benefit greatly from the slabbed allocator. This commit adds a global slabbed allocator dedicated to `Json::Value`.

Real-world data indicates that only 2% of allocation requests are over 72 bytes long. The remaining 98% of allocations fall into the following 3 buckets, calculated across 9,500,000,000 allocation calls:

    [ 1, 32]: 17% of all allocations
    [33, 48]: 27% of all allocations
    [49, 72]: 57% of all allocations

This commit should result in improved performance for servers that have JSON-heavy workloads, typically those servicing RPC and WebSocket workloads, and less memory fragmentation.


src/ripple/rpc/handlers/AccountLines.cpp
```diff
@@ -54,10 +54,10 @@ addLine(Json::Value& jsonLines, RPCTrustLine const& line)
...
     if (line.getAuthPeer())
         jPeer[jss::peer_authorized] = true;
-    if (line.getNoRipple() || !line.getDefaultRipple())
-        jPeer[jss::no_ripple] = line.getNoRipple();
-    if (line.getNoRipplePeer() || !line.getDefaultRipple())
-        jPeer[jss::no_ripple_peer] = line.getNoRipplePeer();
+    if (line.getNoRipple())
+        jPeer[jss::no_ripple] = true;
+    if (line.getNoRipplePeer())
+        jPeer[jss::no_ripple_peer] = true;
```
ximinez: This should probably delete the `TrustLineBase::getDefaultRipple()` function, too.

Issue #4684, opened 2023-08-31
OfferCreate with tfFillOrKill fails with tecKILLED if offer is better than, but doesn't exactly match, open offer rate


nhartner:
## Issue Description

`CreateOffer` with `tfFillOrKill` flag will always results in a `tecKILLED` result if the exchange rate on the offer is better than, but doesn't exactly match, the order book rate. 

## Steps to Reproduce

1. Create 3 accounts and fund them with faucet:
- ISSUER (ex. https://test.bithomp.com/explorer/r1MfR17bEHV1VhFgfZey6TizQKueEJwLV)
- MAKER (ex. https://test.bithomp.com/explorer/rML3sUiiYUnMj7WPc9ECNdMXnYe5Z5Xbdb)
- TAKER (ex. https://test.bithomp.com/explorer/rSeyhj86z7GCQ96EV3X8naQA6V57VQoN9)

2. Enable _Default Ripple_ flag on ISSUER (ex. https://test.bithomp.com/explorer/424CF9443FBBE23734DF9E49BA2297D2A2ECE58C11F01B4EE8FD251F63E85A6C)

3. Create trustlines:
- From MAKER to ISSUER for currency ONE (ex. https://test.bithomp.com/explorer/9F03819996B0289137B0BC9297DEDA53C9DEDD4DA99531989719BAD5DC2CC776)
- From TAKER to ISSUER for currency ONE (ex. https://test.bithomp.com/explorer/72EB305366A3897BF70BB37F1ECF5065A5D4DC48B3D244569B68786242D02F33)

4. Send a payment of 100 units of currency ONE from ISSUER to MAKER (ex. https://test.bithomp.com/explorer/BBFBC0CB6B6ECE68050E8C57E7F7F33F0437D9EF03E638F74D70DFF5DBF1B06B)

5. Create an offer on MAKER account to sell 100 units of currency ONE for 100 XRP (ex. https://test.bithomp.com/explorer/DDFF0C28494332C491147A9A85AC8F5C1A988A848B4366271E72A67266AF5556)

6. Create an offer on TAKER account to buy 100 units of currency ONE for 101 XRP with `tfFillOrKill` flag (ex. https://test.bithomp.com/explorer/7251BC0895848886B9419FC1E209DABAAA2BE3BD66CB690BABFD92D660EA4743)

## Expected Result

[...]

PR #4694, opened 2023-09-08
Fix OfferCreate with tfFillOrKill if offer is better than open offer rate

gregtatcam:

This PR addresses #4684 issue. This is the payment engine breaking change and requires an amendment.

### Context of Change

flowCross amendment introduces an issue where the offer crossing with tfFillOrKill flag set and tfSell not set fails to cross an offer with a better rate. For instance, given two offers:
(takerPays/takerGets)
XRP(100)/USD(100)
USD(100)/XRP(101) has tfFillOrKill set
The second offer fails to cross but it should succeed because the taker can buy USD(100) for XRP(100) instead of XRP(101). XRPL reference for the [offer create with tfFillOrKill](https://xrpl.org/offercreate.html) states: 
>    ...the owner must receive the full TakerPays amount; if the tfSell flag is enabled, the owner must be able to spend the entire TakerGets amount instead.

### Type of Change

- [ x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x ] Tests (You added tests for code that already exists, or your new feature included in this PR)

## Before / After

The amendment code added in StrandFlow.cpp:flow() handles tfFillOrKill and tfFillOrKill+tfSell as two separate cases, whereas the original code handled both as one case.

## Test Plan

Added testFillOrKill() to Offer_test.


src/test/app/Offer_test.cpp
```diff
@@ -5081,6 +5081,157 @@ class Offer_test : public beast::unit_test::suite
         pass();
     }
 
+    void
+    testFillOrKill(FeatureBitset features)
```
scottschurr, line 5085: Thanks very much for adding this test.  I especially like the way it demonstrates identical behavior before flowCross and after flowCross but with your fix.  Nicely done!

scottschurr (approved):
I really like both the fix and the unit test that demonstrates the problem.  I left a few comments on the unit test, but they are all at your discretion.  If you would like to pick up all of the unit test suggestions then you can, if you would like, cherry-pick them from the top-most commit here: https://github.com/scottschurr/rippled/commits/gregt-fix-fillorkill

Very nicely done.  Thanks for fixing my bug!  :+1:

src/ripple/app/paths/impl/StrandFlow.h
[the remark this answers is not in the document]
gregtatcam: Thanks for reviewing and the suggestions. I added/pushed your commit.

intelliot:
@seelabs : will you review this?

intelliot:
note: seelabs plans to review (eventually :)

src/ripple/app/paths/impl/StrandFlow.h
```diff
@@ -817,6 +817,35 @@ flow(
...
+     *   are reversed, the owner must send the entire TakerGets.
+     */
+    bool const fixFillOrKill =
+        baseView.rules().enabled(fixFillOrKillOnFlowCross);
+    // tfSell is handled by setting the deliver amount to max
+    auto const sell = [&]() -> bool {
+        if constexpr (std::is_same_v<TOutAmt, XRPAmount>)
+        {
+            static auto max = XRPAmount{STAmount::cMaxNative};
+            return outReq == max;
```
seelabs, line 838: We can talk about this, but I don't love this solution. It's using a special value of output request to change behavior. Any payment that uses this value will get this special behavior, even payments that aren't offers and don't have the flag set. It also means if the implementation changes so we no long set these max amounts this code breaks.

seelabs (changes requested):

src/ripple/protocol/Feature.h
```diff
@@ -348,6 +348,7 @@ extern uint256 const fixNonFungibleTokensV1_2;
 extern uint256 const fixNFTokenRemint;
 extern uint256 const fixReducedOffersV1;
 extern uint256 const featureClawback;
+extern uint256 const fixFillOrKillOnFlowCross;
```
seelabs, line 351: I probably wouldn't mention "FlowCross". Just `fixOfferFillOrKill`. There's only one offer crossing implementation, there's no need to mention it by name.

src/ripple/app/paths/impl/StrandFlow.h
scottschurr: I can't argue that this is a great solution, but that's a problem with the FlowCross implementation, not with this fix.  @gregtatcam is using the only indication that FlowCross sends down through the payment engine that the `tfSell` flag is set:

https://github.com/XRPLF/rippled/blob/develop/src/ripple/app/tx/impl/CreateOffer.cpp#L739-L755

But, yes, there is also an `offerCrossing` flag that is passed as well.  We could add that to the `sell` detection.  Maybe like this...
```
    // FlowCross handles tfSell by setting the deliver amount to max
    auto const sell = [&]() -> bool {
        if (offerCrossing)
        {
            if constexpr (std::is_same_v<TOutAmt, XRPAmount>)
            {
                static const auto max = XRPAmount{STAmount::cMaxNative};
                return outReq == max;
            }
            else if constexpr (std::is_same_v<TOutAmt, IOUAmount>)
            {
                static const auto max =
                    IOUAmount{STAmount::cMaxValue / 2, STAmount::cMaxOffset};
                return outReq >= max;
            }
        }
        return false;
    }();
```
That's still not great, but it makes sure that the `sell` flag is only set when offer crossing.

gregtatcam: Thanks @scottschurr, it is better but as you said not great. I wanted to minimize the payment engine impact and not change the arguments to `flow()`. If it's ok to change the args then the `offerCrossing` can be changed from `bool` to `optional<uint32>`, which if seated contains the offer crossing flags and if not seated then it's not offer crossing. The semantics of the `offerCrossing` flag remains the same but it does provide more information. The only thing I don't like is that `partialPayment` flag carries a duplicate info (it's set to `!(txFlags & tfFillOrKill)`) but it's probably ok. @seelabs does this work?

intelliot:
Given there are requested changes awaiting review by @gregtatcam , this PR is currently targeting a 2024 release. If there is more urgency to it, please comment here to make the case. Thanks!

scottschurr (changes requested):
First, your solution of changing the `bool` to an `enum` is terrific!  It clarifies a bunch of code by changing a nondescript `bool` into something that has useful descriptive names.  Great choice!

I did leave two comments.  You need to fix the name of a local in `Offer_test::run()`.  You can also consider fixing the names of your new enumerators so they match our code base conventions.

But other than those two things I really like this approach.  Thanks!

src/test/app/Offer_test.cpp
```diff
@@ -5153,12 +5344,17 @@ class Offer_test : public beast::unit_test::suite
         FeatureBitset const takerDryOffer{fixTakerDryOfferRemoval};
         FeatureBitset const rmSmallIncreasedQOffers{fixRmSmallIncreasedQOffers};
         FeatureBitset const immediateOfferKilled{featureImmediateOfferKilled};
+        FeatureBitset const fixFillOrKill{fixFillOrKill};
```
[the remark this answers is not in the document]
gregtatcam: Thanks, it's fixed in this commit [3859760](https://github.com/XRPLF/rippled/pull/4694/commits/385976048582b13a7df761dc4fcc4ec4b2edf5f5).

scottschurr (approved):
:+1:

src/ripple/app/paths/impl/StrandFlow.h
seelabs: Looks good. Thanks for making that change, I appreciate it.

seelabs (approved):
:+1:

intelliot:
@gregtatcam This branch has conflicts that must be resolved

The conflicts look simple but I'll let you take care of them, preferably by adding a merge commit. (As we've been doing, this branch will be squashed to a single commit when it lands in `develop`)

intelliot:
@gregtatcam - I am not sure you merged with latest `develop` (if this repo is `upstream` then maybe do something like `git fetch upstream` and `git merge upstream/develop`) - there are still a couple small conflicts (i.e. for featureDID)

at your convenience, please resolve

gregtatcam:
> @gregtatcam - I am not sure you merged with latest `develop` (if this repo is `upstream` then maybe do something like `git fetch upstream` and `git merge upstream/develop`) - there are still a couple small conflicts (i.e. for featureDID)
> 
> at your convenience, please resolve

Updated.

intelliot:
@seelabs or @scottschurr - could you have a quick re-review to confirm that the merge looks OK?

scottschurr:
When I attempt to build the latest version of this branch the build fails.  But that failure stems from the APIv2: remove tx_history and ledger header commit (https://github.com/XRPLF/rippled/commit/1eac4d2c07eaac04178ff53c7288bc46c6ce7da0)

The failure is in src/test/jtx/TestHelpers.h:38...
```
make_vector(auto const& input) requires std::ranges::range<decltype(input)>
```
The compiler says
```
error: no member named 'range' in namespace 'std::ranges'
```
I suspect that the library for the version of clang I'm running does not yet have support for `std::ranges::range`.

My inability to compile just happened with the introduction of this commit.  There are newer versions of Xcode I could upgrade to.  I'm running 13.4.1, release June 22, 2022 -- about a year and a half old. There is a version 15.0.1 available, released Oct 18, 2023.  Is it time to upgrade?

gregtatcam:
> My inability to compile just happened with the introduction of this commit. There are newer versions of Xcode I could upgrade to. I'm running 13.4.1, release June 22, 2022 -- about a year and a half old. There is a version 15.0.1 available, released Oct 18, 2023. Is it time to upgrade?

Builds for me. I have Apple clang version 15.0.0 (clang-1500.0.40.1).

scottschurr:
According to this website, ranges support requires Xcode 14.3 or later: https://developer.apple.com/xcode/cpp/

scottschurr:
Xcode 14.3 was released March 30, 2023 according to Wikipedia.

scottschurr:
In the mean time, we'd better have @seelabs review the merge commit, since the code is not building for me at the moment.

intelliot:
note: https://github.com/XRPLF/rippled/pull/4788 is expected to fix the error scottschurr mentioned above

intelliot:
Now that #4788 has been merged, this PR can be merged once CI passes. Meanwhile, @gregtatcam - can you confirm this looks good to merge?

(If so, we can put the `Passed` label on it)

gregtatcam:
> Now that #4788 has been merged, this PR can be merged once CI passes. Meanwhile, @gregtatcam - can you confirm this looks good to merge?
> 
> (If so, we can put the `Passed` label on it)

It's good to merge.

PR #4701, opened 2023-09-12
Update README.md

wojake:
Update old links & descriptions.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 The [XRP Ledger](https://xrpl.org/) is a decentralized cryptographic ledger powered by a network of peer-to-peer nodes. The XRP Ledger uses a novel Byzantine Fault Tolerant consensus algorithm to settle and record transactions in a secure distributed database without a central operator.
 
 ## XRP
-[XRP](https://xrpl.org/xrp.html) is a public, counterparty-free asset native to the XRP Ledger, and is designed to bridge the many different currencies in use worldwide. XRP is traded on the open-market and is available for anyone to access. The XRP Ledger was created in 2012 with a finite supply of 100 billion units of XRP.
+[XRP](https://xrpl.org/xrp.html) is a public, counterparty-free crypto-asset native to the XRP Ledger, and is designed as a gas token for network services and to bridge different currencies. XRP is traded on the open-market and is available for anyone to access. The XRP Ledger was created in 2012 with a finite supply of 100 billion units of XRP.
 
 ## rippled
 The server software that powers the XRP Ledger is called `rippled` and is available in this repository under the permissive [ISC open-source license](LICENSE.md). The `rippled` server software is written primarily in C++ and runs on a variety of platforms. The `rippled` server software can run in several modes depending on its [configuration](https://xrpl.org/rippled-server-modes.html).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -3,7 +3,7 @@
 The [XRP Ledger](https://xrpl.org/) is a decentralized cryptographic ledger powered by a network of peer-to-peer nodes. The XRP Ledger uses a novel Byzantine Fault Tolerant consensus algorithm to settle and record transactions in a secure distributed database without a central operator.
 
 ## XRP
-[XRP](https://xrpl.org/xrp.html) is a public, counterparty-free asset native to the XRP Ledger, and is designed to bridge the many different currencies in use worldwide. XRP is traded on the open-market and is available for anyone to access. The XRP Ledger was created in 2012 with a finite supply of 100 billion units of XRP.
+[XRP](https://xrpl.org/xrp.html) is a public, counterparty-free crypto-asset native to the XRP Ledger, and is designed as a gas token for network services and to bridge different currencies. XRP is traded on the open-market and is available for anyone to access. The XRP Ledger was created in 2012 with a finite supply of 100 billion units of XRP.
```
ckeshava, line 6: Nit: I'm not sure if the gas fees needs to be included in the definition of XRP. Although it is used for that purpose, this would not be the defining use of XRP.

ckeshava: Regarding the definition of XRP as a "crypto" asset, I'd defer that decision to @intelliot or someone with expertise in law. 

Other changes look good to me.

wojake: It is the core, fundamental use of XRP. In the past, XRP was created, used and marketed as that, a gas token.

While it's true that XRP has a second widely known feature to bridge currencies, the core fundamental use of XRP is a gas token.

Additionally, this was David's reply to a comment regarding XRP as a gas token (deleted question): https://twitter.com/JoelKatz/status/1682238728906088449. There was a number of materials detailing XRP's attribute as a gas token but I don't hold any of them, I can certainly try to find them.

> Regarding the definition of XRP as a "crypto" asset, I'd defer that decision to @intelliot or someone with expertise in law.

"crypto-asset" could be changed to "crypto", yes. I just referred to MiCA.

MarkusTeufelberger: > In the past, XRP was created, used and marketed as that, a gas token.

No, it was created, used and marketed as DoS and spam prevention mechanism.

wojake: Yes, the same use is applied to Bitcoin, Ether, and many more L1 tokens.

There's a lot of definition mix-up in crypto, some call X as X2 and some call X as Y due to differentiation in protocol design and evolution in status quo. Transaction fees, no matter the differentiation in consensus protocol, incentive model and the re-direction of said fees, is being referred to as **gas**.

David and a few other key developers have stated that XRP's **core** network use is it being a gas token, this is in recent times btw.

The term "gas" is mainly being referred to as transaction fees on Ethereum and a few other chains. There is a growing use of the term "gas" in crypto, from what I'm seeing and even Ripple adopted it in recent years:

image

wojake: > @wojake who is MiCA?

Moms into Crypto-Assets: https://eur-lex.europa.eu/legal-content/EN/TXT/PDF/?uri=CELEX:32023R1114

wojake: The core, fundamental, network use (consumption) of XRP is to process transaction fees. That is a gas token.

With the addition of external, network features like fee escalation and auto-bridging, XRP has additional attributes such as the role as a "world bridge currency" and as a mechanism to prevent spam & DoS attack.

The reason that XRP is being used to bridge currencies & as a mechanism to prevent traffic-related attacks is because, XRP is a gas token; To be consumed to process & validate transactions. Fee Escalation is the component in charged to ensure that the economics of all of this is, valid & rationale according to the code.

This is all flexible in terms of definition due to the maintainers having absolute authority to introduce new features and call it as another use case for XRP. As opposed to the feature being marketed as a network feature, and XRP as a gas token to use such feature on the network.

The XRPL is the use case, XRP is the XRPL's gas token.

wojake: > "The available supply of XRP decreases slowly over time as small amounts are destroyed to pay transaction costs."
~ https://github.com/XRPLF/rippled/blob/develop/README.md?plain=1#L22

"destroyed" as in, consumed. Mathematically, it is destroyed. Technically, it is consumed to cover transaction cost.

README.md
```diff
@@ -21,19 +21,19 @@ If you are interested in running an **API Server** (including a **Full History S
...
 - **[Secure, Adaptable Cryptography][]:** The XRP Ledger relies on industry standard digital signature systems like ECDSA (the same scheme used by Bitcoin) but also supports modern, efficient algorithms like Ed25519. The extensible nature of the XRP Ledger's software makes it possible to add and disable algorithms as the state of the art in cryptography advances.
-- **[Modern Features for Smart Contracts][]:** Features like Escrow, Checks, and Payment Channels support cutting-edge financial applications including the [Interledger Protocol](https://interledger.org/). This toolbox of advanced features comes with safety features like a process for amending the network and separate checks against invariant constraints.
+- **[Modern Features][]:** Features like Escrow, Checks, and Payment Channels support financial applications atop of the XRP Ledger. This toolbox of advanced features comes with safety features like a process for amending the network and separate checks against invariant constraints.
```
mDuo13, line 27: I feel like this would be a great opportunity to update this with another area of significant XRPL development recently, namely tools for on-chain regulation compliance through tools like decentralized IDs, credentials, clawback, and similar, making it a perfect fit for the modern industry where sensible crypto regulations are making it possible to integrate traditional finance with decentralized systems.

mDuo13 (approved):
There's way more that you could do to improve this README, but these changes on their own are an improvement and I would be fine with merging them.

ckeshava:
> @ckeshava Can you please take another look at this one, or let us know if @wojake needs to make changes?

@bthomee  this looks correct to me 👍

README.md
ckeshava: I second this thought ^

ckeshava (approved):

PR #1972, opened 2017-01-18
BEAST_DEFINE_TESTSUITE_MANUAL(ProtocolJson,protocol,ripple);

sublimator:
Just leaving this quick hack here as a starting point for you C++ gurus to work off/consider. This is a manually run test that dumps the protocol definitions in json, which I used to update ripple-lib-java for omnigate.com. 

It's too error-prone to do manually so I hacked this in to spit out json, and then loaded the json inside a java unit test and made sure all the Java definitions (using native enums/classes etc) were in sync.

It would be quite a useful command/feature to have, to keep the various ripple-lib-* up to date. It might be more nicely exposed as an RPC command.

PR #4703, opened 2023-09-13
Add RPC for `server_definitions` to rippled

### LINKED ISSUE #3657: Consider automatically generate enum definitions during compilation
The binary enum definitions should be generated from the rippled code automatically whenever there is a new rippled release.

We should consider how this (interpreted broadly) can be handled so that other packages that rely on the binary representation can be automatically updated.

Options include having a descriptor file in some DSL and a script that parses it to generate the necessary definitions. The script could be packaged and be used as part of the build, and we could have some kind of “hook” on GitHub that runs, generates the files, and updates repos we have push access to (e.g. the JS binary codec).

Exported from RIPD-1305

mvadari:

Add a new RPC / WS call: `server_definitions`, which returns an SDK-compatible `definitions.json` generated from the `rippled` instance currently running. This will alleviate definitions juggling, especially with side chains which may have exotic features and serialized fields. Software will be able to query `server_definitions` against a node from the network they want to connect to and immediately know how to speak that node's language, even if new features are added to it in the future (as long as there are no new serialized types that the software doesn't know how to serialize/deserialize).

Example:

```js
> {"command": "server_definitions"}
< {
    "result": {
        "FIELDS": [
            [
                "Generic",
                {
                    "isSerialized": false,
                    "isSigningField": false,
                    "isVLEncoded": false,
                    "nth": 0,
                    "type": "Unknown"
                }
            ],
            [
                "Invalid",
                {
                    "isSerialized": false,
                    "isSigningField": false,
                    "isVLEncoded": false,
                    "nth": -1,
                    "type": "Unknown"
                }
            ],
            [
                "ObjectEndMarker",
                {
                    "isSerialized": false,
                    "isSigningField": true,
                    "isVLEncoded": false,
                    "nth": 1,
                    "type": "STObject"
                }
            ],
        ...
```

### Context of Change

I took #4402 and modified it to avoid using `magic_enum`. Closes #3657.

### Type of Change

## Test Plan
I haven't written tests yet - I wanted to get this out first, because I'm sure there will be plenty of other issues.


src/ripple/rpc/handlers/ServerInfo.cpp
```diff
@@ -22,13 +22,313 @@
...
+            a[1U] = v;
+            ret[jss::FIELDS][i++] = a;
+        }
+
+        {
+            Json::Value a = Json::arrayValue;
+            a[0U] = "taker_pays_funded";
+            Json::Value v = Json::objectValue;
+            v[jss::nth] = 259;
+            v[jss::isVLEncoded] = false;
```
ckeshava, line 225: what is the rationale for setting these fields to `true/false`? How are you parsing the source code?

mvadari: I don't understand your question. 
Here's an example of the output of this RPC in a library: https://github.com/XRPLF/xrpl.js/blob/main/packages/ripple-binary-codec/src/enums/definitions.json
All this data is needed so that libraries know how to encode and decode transactions/ledger objects/other binary data that's in the custom rippled serialization format. 
The scripts that are currently used to generate these files are here: https://github.com/RichardAH/xrpl-codec-gen

src/ripple/rpc/handlers/ServerInfo.cpp
```diff
@@ -22,13 +22,313 @@
...
+
+    uint256 hash;
+    if (params.isMember(jss::hash))
+    {
+        if (!params[jss::hash].isString() ||
+            !hash.parseHex(params[jss::hash].asString()))
+            return RPC::invalid_field_error(jss::hash);
+    }
+
+    static const Definitions defs{};
```
nbougalis, line 322: Minor nit, but I don't know that you actually need an instance of a class for this: you may as well just have something like `inline const Json::Value definitions = []() { ... }();`

nbougalis (dismissed):

src/ripple/protocol/KnownFormats.h
```diff
@@ -145,6 +145,12 @@ class KnownFormats
         return formats_.end();
     }
 
+    std::map<char const*, KeyType> const
```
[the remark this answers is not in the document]
mvadari: The copy was on purpose, actually, so you can't accidentally mess with the internals. Perhaps I should switch to a static copy in this function instead and return the reference instead...

src/ripple/rpc/handlers/ServerInfo.cpp
```diff
@@ -22,13 +22,313 @@
...
+    if (defs.hashMatches(hash))
+    {
+        Json::Value jv = Json::objectValue;
+        jv[jss::hash] = to_string(hash);
+        return jv;
```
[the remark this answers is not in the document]
mvadari: Most of the code changes in this file were written by @RichardAH in #4402 so I haven't looked over it in detail.

src/ripple/protocol/KnownFormats.h
```diff
@@ -185,6 +191,8 @@ class KnownFormats
         names_[name] = &item;
         types_[type] = &item;
 
+        namesAndTypes_.insert({name, type});
```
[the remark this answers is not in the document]
mvadari: Happy to switch, I copied formatting from somewhere else (I forget where) that used a `char const*` so I thought that was preferred.

src/ripple/protocol/SField.h
```diff
@@ -50,43 +50,56 @@ class STBitString;
...
+    STYPE(STI_UNKNOWN, -2)        \
+    STYPE(STI_NOTPRESENT, 0)      \
+    STYPE(STI_UINT16, 1)          \
+    STYPE(STI_UINT32, 2)          \
+    STYPE(STI_UINT64, 3)          \
+    STYPE(STI_UINT128, 4)         \
+    STYPE(STI_UINT256, 5)         \
+    STYPE(STI_AMOUNT, 6)          \
+    STYPE(STI_VL, 7)              \
+    STYPE(STI_ACCOUNT, 8)         \
```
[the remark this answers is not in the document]
mvadari: I planned on adding them back in, I just wasn't sure how to add comments into multi-line macros (the standard `//` didn't seem to work).

src/ripple/protocol/KnownFormats.h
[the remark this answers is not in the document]
nbougalis: Just change the return `std::map<char const*, KeyType> const&` and you're safe (although, as I mention, this `std::map` may be unnecesary).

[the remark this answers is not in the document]
mvadari: Unneeded in https://github.com/XRPLF/rippled/pull/4703/commits/ff4f55de326dafbdf37a982c0016c12a3ad3aec0

src/ripple/protocol/KnownFormats.h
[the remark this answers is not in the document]
mvadari: Fixed in https://github.com/XRPLF/rippled/pull/4703/commits/ff4f55de326dafbdf37a982c0016c12a3ad3aec0

src/ripple/protocol/KnownFormats.h
```diff
@@ -198,6 +206,8 @@ class KnownFormats
 
     boost::container::flat_map<std::string, Item const*> names_;
     boost::container::flat_map<KeyType, Item const*> types_;
+
+    std::map<char const*, KeyType> namesAndTypes_;
```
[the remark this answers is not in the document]
mvadari: Fixed in https://github.com/XRPLF/rippled/pull/4703/commits/ff4f55de326dafbdf37a982c0016c12a3ad3aec0

src/ripple/protocol/SField.h
```diff
@@ -50,43 +50,56 @@ class STBitString;
...
+    STYPE(STI_VECTOR256, 19)      \
+    STYPE(STI_UINT96, 20)         \
+    STYPE(STI_UINT192, 21)        \
+    STYPE(STI_UINT384, 22)        \
+    STYPE(STI_UINT512, 23)        \
+    STYPE(STI_ISSUE, 24)          \
+    STYPE(STI_TRANSACTION, 10001) \
+    STYPE(STI_LEDGERENTRY, 10002) \
+    STYPE(STI_VALIDATION, 10003)  \
+    STYPE(STI_METADATA, 10004)
```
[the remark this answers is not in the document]
mvadari: Fixed in https://github.com/XRPLF/rippled/pull/4703/commits/166ab29be45393d7bc370ed2959b082fc095ead1

src/ripple/protocol/SField.h
```diff
@@ -50,43 +50,56 @@ class STBitString;
...
+    STYPE(STI_UINT32, 2)          \
+    STYPE(STI_UINT64, 3)          \
+    STYPE(STI_UINT128, 4)         \
+    STYPE(STI_UINT256, 5)         \
+    STYPE(STI_AMOUNT, 6)          \
+    STYPE(STI_VL, 7)              \
+    STYPE(STI_ACCOUNT, 8)         \
+    STYPE(STI_OBJECT, 14)         \
+    STYPE(STI_ARRAY, 15)          \
+    STYPE(STI_UINT8, 16)          \
```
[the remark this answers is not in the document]
mvadari: Fixed in https://github.com/XRPLF/rippled/pull/4703/commits/166ab29be45393d7bc370ed2959b082fc095ead1

src/ripple/protocol/SField.h
```diff
@@ -50,43 +50,56 @@ class STBitString;
...
-    STI_VALIDATION = 10003,
-    STI_METADATA = 10004,
-};
+class Definitions;
+
+#pragma push_macro("XMACRO")
+#undef XMACRO
+
+#define XMACRO(STYPE)             \
+    STYPE(STI_UNKNOWN, -2)        \
```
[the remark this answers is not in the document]
mvadari: Fixed in https://github.com/XRPLF/rippled/pull/4703/commits/166ab29be45393d7bc370ed2959b082fc095ead1

src/ripple/protocol/SField.h
```diff
@@ -50,43 +50,56 @@ class STBitString;
...
+    STYPE(STI_NOTPRESENT, 0)      \
+    STYPE(STI_UINT16, 1)          \
```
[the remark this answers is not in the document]
mvadari: Fixed in https://github.com/XRPLF/rippled/pull/4703/commits/166ab29be45393d7bc370ed2959b082fc095ead1

src/ripple/rpc/handlers/ServerInfo.cpp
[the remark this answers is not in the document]
mvadari: The output does seem to have a `"status": "success"` element when I run the command locally.

src/ripple/protocol/SField.h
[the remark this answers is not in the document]
mvadari: Fixed in https://github.com/XRPLF/rippled/commit/166ab29be45393d7bc370ed2959b082fc095ead1

src/ripple/rpc/handlers/ServerInfo.cpp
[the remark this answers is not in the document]
RichardAH: When you query the API, if you have a pre-existing set of definitions then you supply the hash of those definitions. If the server has the same definitions it just sends back the hash again to confirm, rather than the whole definitions. If the server has different definitions then it sends those plus the new hash.

sublimator:
Nice. Been wanting this for a while:
https://github.com/XRPLF/rippled/pull/1972

src/ripple/app/main/Main.cpp
```diff
@@ -169,6 +169,7 @@ printHelp(const po::options_description& desc)
            "     peer_reservations_list\n"
            "     ripple ...\n"
            "     ripple_path_find <json> [<ledger>]\n"
+           "     server_definitions\n"
            "     server_info [counters]\n"
```
ckeshava, line 173: This is not related to this PR. What does `counters` signify in `server_info [counters]` ?
`server_info` request does not take in any parameters: https://xrpl.org/server_info.html

src/ripple/rpc/impl/Handler.cpp
```diff
@@ -146,6 +146,10 @@ Handler const handlerArray[]{
      byRef(&doSubmitMultiSigned),
      Role::USER,
      NEEDS_CURRENT_LEDGER},
+    {"server_definitions",
+     byRef(&doServerDefinitions),
+     Role::USER,
```
ckeshava, line 151: Is `Role::User` the lowest privileged role in the system? I assumed `Role::GUEST` would be the lowest privilege amongst all of the `Roles`.

If that is the case, shouldn't `server_definitions` be invocable by `Role::GUEST` ?

src/ripple/app/main/Main.cpp
nbougalis: Try it and see.

ckeshava: I collected the outputs of `./rippled server_info > a.txt` and `./rippled server_info 2 > b.txt` and compared the diff between the files.

```
7c7
<             "hash" : "BEA5D3D76787DE85172890B884F951FB3C4020138D53D6D380046A2F75DCC780",
---
>             "hash" : "053DB62FB722A8F9F11CD612350C760F4149B98705FADBAB734475B2769A1888",
10c10
<             "seq" : 1885
---
>             "seq" : 1801
13c13
<          "fetch_pack" : 2554,
---
>          "fetch_pack" : 4677,
25c25
<                   "per_second" : 7
---
>                   "per_second" : 3
33c33
<                   "per_second" : 22
---
>                   "per_second" : 20
```

There are some more lines differing in the time statistics, but I'm not able to reason the difference. I couldn't find documentation for `fetch_pack`, but all the other diff'ed fields appear to be consistent with the slightly different start times

src/ripple/rpc/handlers/ServerInfo.cpp
```diff
@@ -22,13 +22,308 @@
...
+    uint256 hash;
+    if (params.isMember(jss::hash))
+    {
+        if (!params[jss::hash].isString() ||
+            !hash.parseHex(params[jss::hash].asString()))
+            return RPC::invalid_field_error(jss::hash);
+    }
+
+    static const Definitions defs{};
+    if (defs.hashMatches(hash))
```
ckeshava, line 318: The value of `hash` is default-constructed on line 309. How does it represent a user's input?

src/ripple/rpc/handlers/ServerInfo.cpp
```diff
@@ -22,13 +22,308 @@
...
+    Definitions() : defs(generate()){};
+
+    bool
+    hashMatches(uint256 hash) const
+    {
+        return defsHash && *defsHash == hash;
+    }
+
+    Json::Value const&
+    operator()(void) const
```
ckeshava, line 298: why do we need the `void` keyword here? doesn't `operator()() const` suffice?

src/ripple/rpc/handlers/ServerInfo.cpp
ckeshava: okay thanks

src/ripple/protocol/SField.h
```diff
@@ -51,44 +51,67 @@ template <class>
...
-    STI_LEDGERENTRY = 10002,
-    STI_VALIDATION = 10003,
-    STI_METADATA = 10004,
-};
+class Definitions;
+
+#pragma push_macro("XMACRO")
+#undef XMACRO
+
+#define XMACRO(STYPE)                             \
```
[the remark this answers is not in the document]
mvadari: The advantage is that all of the contents of `SerializedTypeID` are in a map. There's no way to iterate through an enum in C++. The macro ensures that you only need to have one line of code for each type ID.

src/ripple/protocol/jss.h
```diff
@@ -317,6 +317,7 @@ JSS(fee_level);             // out: AccountInfo
 JSS(fee_mult_max);          // in: TransactionSign
 JSS(fee_ref);               // out: NetworkOPs
 JSS(fetch_pack);            // out: NetworkOPs
+JSS(FIELDS);                // out: RPC server_definitions
```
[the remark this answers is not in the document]
mvadari: This is the format that is being copied: https://github.com/XRPLF/xrpl.js/blob/main/packages/ripple-binary-codec/src/enums/definitions.json

src/ripple/rpc/handlers/ServerInfo.cpp
mvadari: https://github.com/XRPLF/rippled/pull/4703#discussion_r1330641082

src/ripple/protocol/SField.h
```diff
@@ -50,43 +50,56 @@ class STBitString;
...
-    STI_LEDGERENTRY = 10002,
-    STI_VALIDATION = 10003,
-    STI_METADATA = 10004,
-};
+class Definitions;
+
+#pragma push_macro("XMACRO")
+#undef XMACRO
+
+#define XMACRO(STYPE)             \
```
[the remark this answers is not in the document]
mvadari: I can't figure out how to properly use `Boost.Describe` - I keep getting a ton of errors.

src/ripple/protocol/SField.h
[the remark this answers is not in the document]
ckeshava: i see okay. 

Do we need both `SerializedTypeID` and `sTypeMap`? 
Could we get rid of `enum SerializedTypeID` and live with just the map? (because their contents are identical) I feel that would obviate the macros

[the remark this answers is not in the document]
mvadari: No, the enum is used all over the repo, and it would likely slow down a lot of rippled to force it to call a map.
The contents are supposed to be identical - it needs to be both in an enum _and_ in a map. The enum is for ease of use all over the codebase, and the map is specifically for `server_definitions`, so you can iterate through the map instead of needing to somehow iterate through the enum.

src/ripple/protocol/jss.h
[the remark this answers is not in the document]
ckeshava: okay, I'll go through that file.
What is the rationale for the `CAPITALIZATION_EXCEPTIONS` dictionary? Are those terms being used in another context? https://github.com/RichardAH/xrpl-codec-gen/blob/9157e62e7d585d9dd1bb17eb3f74f2a209fddcd0/gen.py#L4

src/ripple/rpc/handlers/ServerInfo.cpp
ckeshava: Okay. I still don't understand how `hash` captures the user input? `defs` refer to the definitions contained by the server. Shouldn't we read the user input from `params.context` ?

src/ripple/protocol/SField.h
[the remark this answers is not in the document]
ckeshava: I see, alright. 

is this because of performance difference between accessing an `enum` versus `std::unordered_map` or is it more of developer-ease?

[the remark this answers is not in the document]
mvadari: Yes.

src/ripple/protocol/jss.h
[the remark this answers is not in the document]
mvadari: That's because it's parsing the enums - e.g. `XCHAIN_BRIDGE` - which needs to be turned into `XChainBridge`. If you naively capitalize when converting from snake_case to CamelCase you'll get `XchainBridge`. Hence the exceptions.

src/ripple/rpc/handlers/ServerInfo.cpp
mvadari: That's exactly what's happening here - on line 313.

src/ripple/rpc/impl/Handler.cpp
mvadari: I don't see `Role::GUEST` used anywhere. `server_info` uses `Role::USER`, so I think that's the correct option here.

src/ripple/rpc/handlers/ServerInfo.cpp
```diff
@@ -22,13 +22,313 @@
...
+            const std::string out = Json::FastWriter().write(ret);
+            defsHash =
+                ripple::sha512Half(ripple::Slice{out.data(), out.size()});
+            ret[jss::hash] = to_string(*defsHash);
+        }
+
+        return ret;
+    }
+
+    std::optional<uint256> defsHash;
```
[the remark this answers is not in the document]
mvadari: Fixed in https://github.com/XRPLF/rippled/pull/4703/commits/5fc2ee5bda04496327a577d6425650f3ac3c9ee1

src/ripple/rpc/handlers/ServerInfo.cpp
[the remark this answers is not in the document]
mvadari: It looks like the `"status": "success"` element is added automatically to RPCs and doesn't need to be done in the handlers themselves: https://github.com/XRPLF/rippled/blob/develop/src/ripple/rpc/impl/ServerHandler.cpp#L531

src/ripple/rpc/handlers/ServerInfo.cpp
mvadari: Fixed in https://github.com/XRPLF/rippled/pull/4703/commits/e67f72c6ddb71b2fe343fd13e6c5c2d5e3f0075c

src/ripple/rpc/handlers/ServerInfo.cpp
mvadari: IMO things like the `hashMatches` method are a bit neater with a class as opposed to something inline. If you think I should refactor to remove the class though, I can.

src/ripple/rpc/handlers/ServerInfo.cpp
ckeshava: i completely missed that, understood 👍

intelliot:
@mvadari - just confirming: Is this PR ready for review, with all comments up to this point addressed/considered?

mvadari:
> @mvadari - just confirming: Is this PR ready for review, with all comments up to this point addressed/considered?

@intelliot I have 1-2 more of Nik's comments to respond to (I've been going through them slowly) but I've gone through all of Keshava's.

src/ripple/protocol/SField.h
```diff
@@ -52,43 +52,65 @@ class STInteger;
...
+#undef TO_ENUM
+#pragma push_macro("TO_MAP")
+#undef TO_MAP
+
+#define TO_ENUM(name, value) name = value,
+#define TO_MAP(name, value) {#name, value},
+
+enum SerializedTypeID { XMACRO(TO_ENUM) };
+
+static std::map<char const*, int> const sTypeMap = {XMACRO(TO_MAP)};
```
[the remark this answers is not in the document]
mvadari: No specific reason really. I can switch it.

src/ripple/protocol/SField.h
```diff
@@ -52,43 +52,65 @@ class STInteger;
...
-    // cannot be serialized inside other types
-    STI_TRANSACTION = 10001,
-    STI_LEDGERENTRY = 10002,
-    STI_VALIDATION = 10003,
-    STI_METADATA = 10004,
-};
+#pragma push_macro("XMACRO")
+#undef XMACRO
+
+#define XMACRO(STYPE)                             \
```
seelabs, line 58: I just wanted to note that this is a problem that the C++ language itself should solve, and there really isn't a nice, clean solution to reflection and enums. I prefer to keep our solution reasonably "magic free", and it's easy to understand what's going on with this macro solution. When we're reaching for macros the first thing we're going to want to see is if there's a better way. I know there are other solutions to this, but for this limited use-case I think macros are OK.

src/ripple/protocol/SField.h
```diff
@@ -266,9 +288,10 @@ class SField
     static int
     compare(const SField& f1, const SField& f2);
 
+    static std::map<int, SField const*> knownCodeToField;
```
seelabs, line 291: Instead of making this public consider a function that provides `const&` access.

src/ripple/protocol/TER.h
```diff
@@ -641,6 +642,15 @@ isTecClaim(TER x)
     return ((x) >= tecCLAIM);
 }
 
+namespace detail {
+
```
seelabs, line 646: Since other modules are accessing this, it shouldn't be in a `detail` namespace.

src/ripple/protocol/jss.h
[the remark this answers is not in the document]
seelabs: Please add a short comment to every field that does not follow the naming convention and explain that it was done intentionally. (this, and things like `isSerialized`, `isSigningField`, `isVLEncoded`, `LEDGER_ENTRY_TYPES`

src/ripple/rpc/handlers/ServerInfo.cpp
```diff
@@ -22,13 +22,298 @@
...
 #include <ripple/net/RPCErr.h>
+#include <ripple/protocol/LedgerFormats.h>
+#include <ripple/protocol/SField.h>
+#include <ripple/protocol/TER.h>
+#include <ripple/protocol/TxFormats.h>
+#include <ripple/protocol/digest.h>
 #include <ripple/protocol/jss.h>
 #include <ripple/rpc/Context.h>
 #include <ripple/rpc/Role.h>
-#include <ripple/rpc/impl/TransactionSign.h>
```
seelabs, line 28: Add blank lines between the include groups. I.e. no lines between the ripple includes, then a blank line then the boot includes, then a blank line, then the standard includes.

src/ripple/rpc/handlers/ServerInfo.cpp
```diff
@@ -22,13 +22,298 @@
...
+#include <boost/algorithm/string.hpp>
+#include <sstream>
 
 namespace ripple {
 
+class Definitions
```
seelabs, line 38: `Definitions` is too generic a name for the `ripple` namespace. Consider putting it into a sub-namespace or renaming it to something less generix.

src/ripple/rpc/handlers/ServerInfo.cpp
```diff
@@ -22,13 +22,298 @@
...
+{
+private:
+    std::string
+    translate(std::string inp)
+    {
+        auto replace = [&](const char* f, const char* r) -> std::string {
```
seelabs, line 44: f and r seem like strange param names. Maybe give them more descriptive names?

src/ripple/rpc/handlers/ServerInfo.cpp
```diff
@@ -22,13 +22,298 @@
...
+#include <boost/algorithm/string.hpp>
+#include <sstream>
 
 namespace ripple {
 
+class Definitions
+{
+private:
+    std::string
+    translate(std::string inp)
```
seelabs, line 42: Pass the param by `const&`, unless I'm missing the reason it's passed by value.

src/ripple/rpc/handlers/ServerInfo.cpp
```diff
@@ -22,13 +22,298 @@
...
+    {
```
seelabs, line 43: Separate the declarations from the implementations. We do that so it's easier to see class interfaces. (I honestly don't care that much, but it is our code style, and other people do strongly prefer it).

src/ripple/rpc/handlers/ServerInfo.cpp
```diff
@@ -22,13 +22,298 @@
...
+        auto replace = [&](const char* f, const char* r) -> std::string {
+            std::string out = inp;
+            boost::replace_all(out, f, r);
+            return out;
+        };
+
+        auto find = [&](const char* s) -> bool {
```
seelabs, line 50: `find` is already a member function on `string` and it acts differently from this lambda. Consider renaming this. C++ 23 has a `contains` member function that is similar to this. Maybe rename this `contains`?

src/ripple/rpc/handlers/ServerInfo.cpp
```diff
@@ -22,13 +22,298 @@
...
+
+        if (find("UINT"))
+        {
+            if (find("256") || find("160") || find("128"))
+                return replace("UINT", "Hash");
+            else
+                return replace("UINT", "UInt");
+        }
+
+        if (inp == "OBJECT")
```
seelabs, line 62: Consider using a hash table here. Also fine as-is (we don't really care about efficiency here. This code is only ever run once).

src/ripple/rpc/handlers/ServerInfo.cpp
```diff
@@ -22,13 +22,298 @@
...
+    };
+
+    Json::Value
+    generate()
+    {
+        Json::Value ret{Json::objectValue};
+        ret[jss::TYPES] = Json::objectValue;
+
+        ret[jss::TYPES]["Done"] = -1;
+        std::map<int32_t, std::string> type_map{{-1, "Done"}};
```
seelabs, line 115: Don't use `snake_case`. Here and other places.

src/ripple/rpc/handlers/ServerInfo.cpp
```diff
@@ -22,13 +22,298 @@
...
+        for (const auto& pair : sTypeMap)
```
seelabs, line 116: Consider using a structured binding here. Also, the `const` goes after the `auto`. I.e.

src/ripple/rpc/handlers/ServerInfo.cpp
```diff
@@ -22,13 +22,298 @@
...
+            Json::Value innerArray = Json::arrayValue;
+            innerArray[0U] = f->fieldName;
+            innerArray[1U] = innerObj;
+
+            ret[jss::FIELDS][i++] = innerArray;
+        }
+
+        ret[jss::TRANSACTION_RESULTS] = Json::objectValue;
+
+        for (const auto& pair : detail::transResults())
```
seelabs, line 250: We shouldn't use the `detail` namespace from another module. (remove the function from the detail namespace).

seelabs, line 250: The `const` goes after the `auto`. Also, consider using a structured binding here.

src/ripple/rpc/handlers/ServerInfo.cpp
```diff
@@ -22,13 +22,298 @@
...
+        {
+            ret[jss::TRANSACTION_RESULTS][pair.second.first] = pair.first;
+        }
+
+        ret[jss::TRANSACTION_TYPES] = Json::objectValue;
+        ret[jss::TRANSACTION_TYPES][jss::Invalid] = -1;
+        for (auto it = TxFormats::getInstance().begin();
```
seelabs, line 257: range-based for look would work well here. Something like:
`for(auto const& f: TxFormats::getInstance())`

src/ripple/rpc/handlers/ServerInfo.cpp
```diff
@@ -22,13 +22,298 @@
...
+    Definitions() : defs_(generate()){};
+
+    bool
+    hashMatches(uint256 hash) const
+    {
+        return defsHash_ == hash;
+    }
+
+    Json::Value const&
+    operator()() const
```
seelabs, line 288: This doesn't really need to be an operation. I'd probably make this a regular function. But also fine as-is if there's a reason I'm missing or you strongly prefer the operator.

seelabs (commented):
Looks good. Left some comments.

src/ripple/rpc/handlers/ServerInfo.cpp
```diff
@@ -22,13 +22,298 @@
...
+                translate(std::string(pair.first).substr(4) /* remove STI_ */);
+            int32_t type_value = pair.second;
+            ret[jss::TYPES][type_name] = type_value;
+            type_map[type_value] = type_name;
+        }
+
+        ret[jss::LEDGER_ENTRY_TYPES] = Json::objectValue;
+        ret[jss::LEDGER_ENTRY_TYPES][jss::Invalid] = -1;
+
+        for (auto it = LedgerFormats::getInstance().begin();
```
ckeshava, line 128: I feel this portion of the code can be refactored into a common utility function inside the `KnownFormats` class. I came up with [this](a38e4e78e5500ffd939402ff006ed0ed5c857705)

src/ripple/rpc/handlers/ServerInfo.cpp
```diff
@@ -22,13 +22,298 @@
...
+                out += token;
+            if (pos == inp.size())
+                break;
+            inp = inp.substr(pos + 1);
+        }
+        return out;
+    };
+
+    Json::Value
+    generate()
```
ckeshava, line 109: Why do you have this `generate` function? Why not place this functionality inside the `Definitions` constructor? (`generate` is not being invoked anywhere else)

Since `Definitions` need to be generated at most once in the life time of the program, is `Singleton` pattern appropriate for this requirement? Does it suffice to have one `static` instance of the `Definitions` object?

src/ripple/rpc/handlers/ServerInfo.cpp
mvadari: I get a lot of these warnings when I do that:
```
warning: returning reference to local temporary object [-Wreturn-stack-address]
            return "STObject";
```

ckeshava: I was able to build this commit without any errors/warnings: 3e5617064ca5c0bac37b3aa8f5316d659fdd69af

src/ripple/rpc/handlers/ServerInfo.cpp
mvadari: Fixed in https://github.com/XRPLF/rippled/pull/4703/commits/d558b665cd2658b8fd6d2176662758f9efd8ae3f

src/ripple/rpc/handlers/ServerInfo.cpp
mvadari: Fixed in https://github.com/XRPLF/rippled/pull/4703/commits/d558b665cd2658b8fd6d2176662758f9efd8ae3f and learned what a structured binding is :)

mvadari: Fixed in https://github.com/XRPLF/rippled/pull/4703/commits/d558b665cd2658b8fd6d2176662758f9efd8ae3f

src/ripple/rpc/handlers/ServerInfo.cpp
mvadari: Fixed in https://github.com/XRPLF/rippled/pull/4703/commits/d558b665cd2658b8fd6d2176662758f9efd8ae3f

src/ripple/rpc/handlers/ServerInfo.cpp
mvadari: Fixed in https://github.com/XRPLF/rippled/pull/4703/commits/d558b665cd2658b8fd6d2176662758f9efd8ae3f

src/ripple/rpc/handlers/ServerInfo.cpp
mvadari: Fixed in https://github.com/XRPLF/rippled/pull/4703/commits/d558b665cd2658b8fd6d2176662758f9efd8ae3f

src/ripple/protocol/TER.h
mvadari: Fixed in https://github.com/XRPLF/rippled/pull/4703/commits/d558b665cd2658b8fd6d2176662758f9efd8ae3f

src/ripple/rpc/handlers/ServerInfo.cpp
mvadari: Fixed in https://github.com/XRPLF/rippled/pull/4703/commits/c34bcd0a628f0c866b0b5bcef0aa5af1e22d43aa

src/ripple/rpc/handlers/ServerInfo.cpp
mvadari: Oops, I was changing the return type instead of the param type. Fixed in https://github.com/XRPLF/rippled/pull/4703/commits/c34bcd0a628f0c866b0b5bcef0aa5af1e22d43aa

src/ripple/rpc/handlers/ServerInfo.cpp
mvadari: Fixed in https://github.com/XRPLF/rippled/commit/d558b665cd2658b8fd6d2176662758f9efd8ae3f

src/ripple/protocol/SField.h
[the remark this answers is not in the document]
mvadari: Fixed in https://github.com/XRPLF/rippled/pull/4703/commits/d558b665cd2658b8fd6d2176662758f9efd8ae3f

src/ripple/rpc/handlers/ServerInfo.cpp
mvadari: That link doesn't work for me. Does this fix satisfy you? https://github.com/XRPLF/rippled/pull/4703#discussion_r1350810440

ckeshava: No, I have different idea, here's the PR: https://github.com/mvadari/rippled/pull/3

mvadari: IMO that's unnecessary.

src/ripple/rpc/handlers/ServerInfo.cpp
mvadari: I thought it was a bit more readable as a hash table, so I changed it in https://github.com/XRPLF/rippled/pull/4703/commits/ebe4258ac26d67cb544adfd7c7d693a3d72ce08e

src/ripple/rpc/handlers/ServerInfo.cpp
mvadari: Fixed in https://github.com/XRPLF/rippled/pull/4703/commits/ebe4258ac26d67cb544adfd7c7d693a3d72ce08e

src/ripple/protocol/SField.h
mvadari: Fixed in https://github.com/XRPLF/rippled/pull/4703/commits/ebe4258ac26d67cb544adfd7c7d693a3d72ce08e

src/ripple/rpc/handlers/ServerInfo.cpp
```diff
@@ -22,13 +22,297 @@
...
+#include <unordered_map>
 
 namespace ripple {
 
+namespace detail {
+
+class ServerDefinitions
+{
+private:
+    std::string const
```
seelabs, line 45: Remove the `const` from the return type. I suspect this is just a typo from when you changed the param to `const&`

src/ripple/rpc/handlers/ServerInfo.cpp
```diff
@@ -22,13 +22,297 @@
...
+    operator()() const
+    {
+        return defs_;
+    }
+};
+
+std::string const
+ServerDefinitions::translate(std::string const& inp)
+{
+    auto replace = [&](const char* oldStr, const char* newStr) -> std::string {
```
seelabs, line 73: Nit: `const` is in the wrong place (should be `char const*`)

seelabs (approved):
:+1: Nice job on this. I'm approving this because the only issues left are minor nits. I do think we should take care of the nits (and I will review if you do), but I'm OK approving this now.

intelliot:
note: needs an entry in `API-CHANGELOG.md`

src/ripple/rpc/handlers/ServerInfo.cpp
mvadari: Fixed in https://github.com/XRPLF/rippled/pull/4703/commits/b628b4203dd54345ce73597005be6a3e1727be7f

src/ripple/rpc/handlers/ServerInfo.cpp
mvadari: Fixed in https://github.com/XRPLF/rippled/pull/4703/commits/b628b4203dd54345ce73597005be6a3e1727be7f

ckeshava:
@mvadari Are you planning to add unit tests for server-definitions?

src/ripple/rpc/handlers/ServerInfo.cpp
```diff
@@ -22,13 +22,313 @@
...
+    std::optional<uint256> defsHash;
+    Json::Value defs;
+
+public:
+    Definitions() : defs(generate()){};
+
+    bool
+    hashMatches(uint256 hash) const
+    {
+        return defsHash && *defsHash == hash;
```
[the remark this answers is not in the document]
mvadari: I removed the `std::optional` in https://github.com/XRPLF/rippled/commit/5fc2ee5bda04496327a577d6425650f3ac3c9ee1, IMO it's a bit clearer and slightly more performant to have the separate variable.

src/ripple/protocol/jss.h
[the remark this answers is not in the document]
mvadari: Done in https://github.com/XRPLF/rippled/pull/4703/commits/ebe4258ac26d67cb544adfd7c7d693a3d72ce08e

src/ripple/rpc/handlers/ServerInfo.cpp
mvadari: Fixed in https://github.com/XRPLF/rippled/pull/4703/commits/7f17a784dccd4a5ac232470275f2dc52cb7f6c45

src/ripple/protocol/jss.h
```diff
@@ -361,6 +363,12 @@ JSS(invalid_API_version);  // out: Many, when a request has an invalid
 JSS(io_latency_ms);        // out: NetworkOPs
 JSS(ip);                   // in: Connect, out: OverlayImpl
 JSS(is_burned);            // out: nft_info (clio)
+JSS(isSerialized);         // out: RPC server_definitions
+                           // matches server_definitions.json format
+JSS(isSigningField);       // out: RPC server_definitions
+                           // matches server_definitions.json format
+JSS(isVLEncoded);          // out: RPC server_definitions
+                           // matches server_definitions.json format
```
ckeshava, line 371: this comment is not very informative. it would help if you could describe this field, where and why its being used

src/ripple/rpc/handlers/ServerInfo.cpp
```diff
@@ -22,13 +22,297 @@
...
+
+    if (auto const& it = replacements.find(inp); it != replacements.end())
+    {
+        return it->second;
+    }
+
+    std::string out;
+    size_t pos = 0;
+    std::string inpToProcess = inp;
+    for (;;)
```
ckeshava, line 110:
```suggestion
    // this loop converts "snake_case" types into "camelCase"
    for (;;)
```

src/ripple/rpc/handlers/ServerInfo.cpp
```diff
@@ -22,13 +22,297 @@
...
+    translate(std::string const& inp);
+
+    Json::Value
+    generate();
+
+    uint256 defsHash_;
+    Json::Value defs_;
+
+public:
+    ServerDefinitions() : defs_(generate()){};
```
ckeshava, line 55: Why do we need a separate function `generate()` ? The constructor appears to be the only place, where this piece of code is executed.

Wouldn't the singleton pattern be more appropriate here? Do you intend to create one single instance of this class? I feel that would better communicate your intent

intelliot:
Would this PR resolve https://github.com/XRPLF/rippled/issues/3657?

src/ripple/protocol/jss.h
mvadari: What more info would you like? I think all that info is already there, to the same degree that explanations exist elsewhere in the file.

mvadari:
> Would this PR resolve #3657?

I believe it would. Added to the PR description.

src/ripple/rpc/handlers/ServerInfo.cpp
mvadari: There is only one `static` instance of the `Definitions` object - in the `doServerDefinitions` method.

src/ripple/rpc/handlers/ServerInfo.cpp
mvadari: Done in https://github.com/XRPLF/rippled/pull/4703/commits/205c0f5bc37b99d22becd640c4f587114f961edc

src/ripple/rpc/handlers/ServerInfo.cpp
mvadari: Done in https://github.com/XRPLF/rippled/pull/4703/commits/205c0f5bc37b99d22becd640c4f587114f961edc

src/ripple/protocol/jss.h
ckeshava: this is not a binding requirement: I feel explaining the motivation for the use of the variable would be useful, especially because `server_definitions.json` is constructed from the python script. It would save future developers from having to read the python script.

src/ripple/rpc/handlers/ServerInfo.cpp
ckeshava: I understand -- but it might not be readily apparent to the readers of the code. I felt there are alternatives which are better poised to convey that there's only one instance

ckeshava (approved):

ckeshava:
The PR looks nice. I've still left some suggestions, but they are my subjective opinions and shouldn't stop the PR from going into production

src/ripple/rpc/handlers/ServerInfo.cpp
mvadari: I don't think that's necessary, since this is only a helper class and isn't used anywhere else.

src/ripple/protocol/jss.h
mvadari: I added some more explanations where these values are actually used - is that enough?

src/ripple/rpc/handlers/ServerInfo.cpp
mvadari: It's pretty easy to see the only place it's used is a static variable IMO - and it'd be a comparable search to finding the `getInstance` method. If you think it'd be easier I can move the `ServerDefinitions` implementation below the `doServerDefinitions` method, so that code is front-and-center?

src/ripple/protocol/jss.h
ckeshava: yes, the comments are helpful, thanks

src/ripple/rpc/handlers/ServerInfo.cpp
ckeshava: no, I think that is not required 👍

mvadari:
@intelliot this is ready to merge 👍

mvadari:
I don't have the powers to retry actions but it looks like something failed with the container setup in the dependency test.

intelliot:
1. I clicked the button to retry the CI job
2. I've asked @nbougalis to confirm that he doesn't have any remaining blockers (i.e. comments/concerns that still must be addressed)

intelliot (approved):
@ckeshava has approved the latest code in this PR (aside from the merge commit)

intelliot:
Suggested commit message:
```
feat(rpc): add `server_definitions` method (#4703)

Add a new RPC / WS call for `server_definitions`, which returns an
SDK-compatible `definitions.json` (binary enum definitions) generated by
the server. This enables clients/libraries to dynamically work with new
fields and features, such as ones that may become available on side
chains. Clients query `server_definitions` on a node from the network
they want to work with, and immediately know how to speak that node's
binary "language", even if new features are added to it in the future
(as long as there are no new serialized types that the software doesn't
know how to serialize/deserialize).

Example:

```js
> {"command": "server_definitions"}
< {
    "result": {
        "FIELDS": [
            [
                "Generic",
                {
                    "isSerialized": false,
                    "isSigningField": false,
                    "isVLEncoded": false,
                    "nth": 0,
                    "type": "Unknown"
                }
            ],
            [
                "Invalid",
                {
                    "isSerialized": false,
                    "isSigningField": false,
                    "isVLEncoded": false,
                    "nth": -1,
                    "type": "Unknown"
                }
            ],
            [
                "ObjectEndMarker",
                {
                    "isSerialized": false,
                    "isSigningField": true,
                    "isVLEncoded": false,
                    "nth": 1,
                    "type": "STObject"
                }
            ],
        ...
```

Built on #4402 and modified to avoid using `magic_enum`.

Close RIPD-1305

Close #3657

---------

Co-authored-by: Richard Holland <richard.holland@starstone.co.nz>
```

kennyzlei:
For Clio, we are considering adding this support in 2.2 release as it needs some time to investigate support for this API.  This means for s1/s2, rippled 2.0 release would mean we would need to wait for Clio 2.2 to have this RPC accessible

https://github.com/XRPLF/clio/issues/942

mvadari:
> For Clio, we are considering adding this support in 2.2 release as it needs some time to investigate support for this API. This means for s1/s2, rippled 2.0 release would mean we would need to wait for Clio 2.2 to have this RPC accessible
> 
> [XRPLF/clio#942](https://github.com/XRPLF/clio/issues/942)

@kennyzlei is it possible to just forward to rippled in 2.1? It should be pretty performant as-is and shouldn't need to be called too often.

SaxenaKaustubh:
Worked on this with @mvadari on this change. After a few iterations, we have this working the way we want. This concludes the manual testing.

PR #4707, opened 2023-09-14
Revert CMake changes

legleux:
Early return from `RippledRelease.cmake` preventing targets being created during packaging.


intelliot:
(@seelabs confirmed this should be ok)

intelliot (approved):

PR #4708, opened 2023-09-14
Remove two unused functions

seelabs:


ximinez (approved):
Looks like a good, safe change. And it builds.

scottschurr (approved):
:+1: Looks great.  Good to get rid of unused/unimplemented interfaces.  Everything builds / passes tests / syncs on MacOS.

PR #4710, opened 2023-09-15
Reduce boilerplate in applySteps:

seelabs:

When a new transactor is added, there are several places in applySteps that need to be modified. This patch refactors the code so only one function needs to be modified.

### Type of Change

### DIFF
--- a/src/ripple/app/tx/impl/applySteps.cpp
+++ b/src/ripple/app/tx/impl/applySteps.cpp
@@ -50,7 +50,115 @@
 #include <ripple/app/tx/impl/XChainBridge.h>
 #include <ripple/protocol/TxFormats.h>
 
+#include <stdexcept>
+
 namespace ripple {
+
+namespace {
+
+struct UnknownTxnType : std::exception
+{
+    TxType txnType;
+    UnknownTxnType(TxType t) : txnType{t}
+    {
+    }
+};
+
+// Call a lambda with the concrete transaction type as a template parameter
+// throw an "UnknownTxnType" exception on error
+template <class F>
+auto
+with_txn_type(TxType txnType, F&& f)
+{
+    switch (txnType)
+    {
+        case ttACCOUNT_DELETE:
+            return f.template operator()<DeleteAccount>();
+        case ttACCOUNT_SET:
+            return f.template operator()<SetAccount>();
+        case ttCHECK_CANCEL:
+            return f.template operator()<CancelCheck>();
+        case ttCHECK_CASH:
+            return f.template operator()<CashCheck>();
+        case ttCHECK_CREATE:
+            return f.template operator()<CreateCheck>();
+        case ttDEPOSIT_PREAUTH:
+            return f.template operator()<DepositPreauth>();
+        case ttOFFER_CANCEL:
+            return f.template operator()<CancelOffer>();
+        case ttOFFER_CREATE:
+            return f.template operator()<CreateOffer>();
+        case ttESCROW_CREATE:
+            return f.template operator()<EscrowCreate>();
+        case ttESCROW_FINISH:
+            return f.template operator()<EscrowFinish>();
+        case ttESCROW_CANCEL:
+            return f.template operator()<EscrowCancel>();
+        case ttPAYCHAN_CLAIM:
+            return f.template operator()<PayChanClaim>();
+        case ttPAYCHAN_CREATE:
+            return f.template operator()<PayChanCreate>();
+        case ttPAYCHAN_FUND:
+            return f.template operator()<PayChanFund>();
+        case ttPAYMENT:
+            return f.template operator()<Payment>();
+        case ttREGULAR_KEY_SET:
+            return f.template operator()<SetRegularKey>();
+        case ttSIGNER_LIST_SET:
+            return f.template operator()<SetSignerList>();
+        case ttTICKET_CREATE:
+            return f.template operator()<CreateTicket>();
+        case ttTRUST_SET:
+            return f.template operator()<SetTrust>();
+        case ttAMENDMENT:
+        case ttFEE:
+        case ttUNL_MODIFY:
+            return f.template operator()<Change>();
+        case ttNFTOKEN_MINT:
+            return f.template operator()<NFTokenMint>();
+        case ttNFTOKEN_BURN:
+            return f.template operator()<NFTokenBurn>();
+        case ttNFTOKEN_CREATE_OFFER:
+            return f.template operator()<NFTokenCreateOffer>();
+        case ttNFTOKEN_CANCEL_OFFER:
+            return f.template operator()<NFTokenCancelOffer>();
+        case ttNFTOKEN_ACCEPT_OFFER:
+            return f.template operator()<NFTokenAcceptOffer>();
+        case ttCLAWBACK:
+            return f.template operator()<Clawback>();
+        case ttAMM_CREATE:
+            return f.template operator()<AMMCreate>();
+        case ttAMM_DEPOSIT:
+            return f.template operator()<AMMDeposit>();
+        case ttAMM_WITHDRAW:
+            return f.template operator()<AMMWithdraw>();
+        case ttAMM_VOTE:
+            return f.template operator()<AMMVote>();
+        case ttAMM_BID:
+            return f.template operator()<AMMBid>();
+        case ttAMM_DELETE:
+            return f.template operator()<AMMDelete>();
+        case ttXCHAIN_CREATE_BRIDGE:
+            return f.template operator()<XChainCreateBridge>();
+        case ttXCHAIN_MODIFY_BRIDGE:
+            return f.template operator()<BridgeModify>();
+        case ttXCHAIN_CREATE_CLAIM_ID:
+            return f.template operator()<XChainCreateClaimID>();
+        case ttXCHAIN_COMMIT:
+            return f.template operator()<XChainCommit>();
+        case ttXCHAIN_CLAIM:
+            return f.template operator()<XChainClaim>();
+        case ttXCHAIN_ADD_CLAIM_ATTESTATION:
+            return f.template operator()<XChainAddClaimAttestation>();
+        case ttXCHAIN_ADD_ACCOUNT_CREATE_ATTESTATION:
+            return f.template operator()<XChainAddAccountCreateAttestation>();
+        case ttXCHAIN_ACCOUNT_CREATE_COMMIT:
+            return f.template operator()<XChainCreateAccountCommit>();
+        default:
+            throw UnknownTxnType(txnType);
+    }
+}
+}  // namespace
 
 // Templates so preflight does the right thing with T::ConsequencesFactory.
 //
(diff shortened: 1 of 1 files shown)


src/ripple/app/tx/impl/applySteps.cpp
```diff
@@ -50,371 +50,226 @@
...
+            return f.template operator()<XChainCreateBridge>();
         case ttXCHAIN_MODIFY_BRIDGE:
-            return invoke_preflight_helper<BridgeModify>(ctx);
+            return f.template operator()<BridgeModify>();
         case ttXCHAIN_CREATE_CLAIM_ID:
-            return invoke_preflight_helper<XChainCreateClaimID>(ctx);
+            return f.template operator()<XChainCreateClaimID>();
         case ttXCHAIN_COMMIT:
-            return invoke_preflight_helper<XChainCommit>(ctx);
+            return f.template operator()<XChainCommit>();
```
ckeshava, line 144: I had not come across the usage of the `template` keyword, I learnt a new construct :) thanks.

src/ripple/app/tx/impl/applySteps.cpp
```diff
@@ -50,371 +50,226 @@
...
+            auto const tec = T::preflight(ctx);
+            return std::make_pair(
+                tec,
+                isTesSuccess(tec) ? consequences_helper<T>(ctx)
+                                  : TxConsequences{tec});
+        });
+    }
+    catch (UnknownTxnType const& e)
+    {
+        assert(false);
```
ckeshava, line 211: Why is this `assert` statement necessary? The exception will be caught and throw without the assert too.

src/ripple/app/tx/impl/applySteps.cpp
```diff
@@ -460,173 +315,17 @@ TxConsequences::TxConsequences(STTx const& tx, std::uint32_t sequencesConsumed)
...
-            Payment p(ctx);
-            return p();
-        }
-        case ttREGULAR_KEY_SET: {
-            SetRegularKey p(ctx);
-            return p();
-        }
-        case ttSIGNER_LIST_SET: {
-            SetSignerList p(ctx);
+        return with_txn_type(ctx.tx.getTxnType(), [&]<typename T>() {
```
ckeshava, line 320: Does the developer need to specify `invoke_apply.template operator<Some_Txn_Type>()` the type explicitly like this? 

So you have effectively collected the all the common "functionality" from `invoke_apply`, `invoke_preflight_helper`, `invoke_preclaim` and `invoke_calculateBaseFee` and placed them under `with_txn_type`. Am I understanding this correctly?

src/ripple/app/tx/impl/applySteps.cpp
```diff
@@ -50,371 +50,226 @@
...
 };
 
-template <class T>
-std::pair<NotTEC, TxConsequences>
-invoke_preflight_helper(PreflightContext const& ctx)
+// Call a lambda with the concrete transaction type as a template parameter
+// throw an "UnknownTxnType" exception on error
+template <class F>
+auto
+with_txn_type(TxType txnType, F const& f)
```
nbougalis, line 67: Why not `F&&`?

nbougalis (commented):
Nice! I like this a lot. It's certainly advanced C++ and some people may go "wha... ?!?" when they read it, but it does make it less error-prone to add new transaction types.

Good job!

src/ripple/app/tx/impl/applySteps.cpp
nbougalis: The intention is to have servers running in debug mode to "break" into the debugger if they encounter an unknown transaction. The wisdom of this is... debatable, but this bit was already there before.

ckeshava: ok, thanks for the clarification

src/ripple/app/tx/impl/applySteps.cpp
seelabs: There's no reason it has to be `const`. Fixed in `e253734ede [fold] Use universal ref for callback`, and yeah, I tend to prefer universal refs for callbacks myself. I don't know why I wrote it as a `const&` (although it doesn't really cause harm either).

src/ripple/app/tx/impl/applySteps.cpp
seelabs: I may be misunderstanding your question, but `invoke_apply` is called like this: `invoke_apply(ctx);`. It's a regular function and doesn't have an `operator()`. 

`with_txn_type` factors out the `switch` statement that used to live in those four places, and handles the variability between those functions by using a callback. So now when a new transaction is added we need to change one place rather than four.

src/ripple/app/tx/impl/applySteps.cpp
```diff
@@ -50,371 +50,236 @@
...
-            return XChainAddClaimAttestation::calculateBaseFee(view, tx);
-        case ttXCHAIN_ADD_ACCOUNT_CREATE_ATTESTATION:
-            return XChainAddAccountCreateAttestation::calculateBaseFee(
-                view, tx);
-        case ttXCHAIN_ACCOUNT_CREATE_COMMIT:
-            return XChainCreateAccountCommit::calculateBaseFee(view, tx);
-        default:
-            assert(false);
-            return XRPAmount{0};
+        return with_txn_type(tx.getTxnType(), [&]<typename T>() {
```
ckeshava, line 275: @seelabs  How does this lambda deduce the type T? T needs to be explicitly specified by the caller of `invoke_apply` or any of the other three functions right?

src/ripple/app/tx/impl/applySteps.cpp
```diff
@@ -50,371 +50,236 @@
...
+    {
+        // Should never happen
+        JLOG(ctx.j.fatal())
+            << "Unknown transaction type in preflight: " << e.txnType;
+        assert(false);
+        return {temUNKNOWN, TxConsequences{temUNKNOWN}};
     }
 }
 
-/* invoke_preclaim<T> uses name hiding to accomplish
```
mvadari, line 196: Shouldn't these comments stay?

src/ripple/app/tx/impl/applySteps.cpp
mvadari: C++20 adds support for templated lambdas: https://ggulgulia.medium.com/c-20-lambda-expression-and-template-syntax-7f6ae2f9e7cd
`with_txn_type` will provide the type to the lambda based on the value of `tx.getTxnType`.

mvadari, line 275: Why do the `view` and `tx` parameters not need to be captured in the lambda?

src/ripple/app/tx/impl/applySteps.cpp
mvadari: `invoke_apply` is a normal function, it's just doing complex things under the hood.

src/ripple/app/tx/impl/applySteps.cpp
seelabs: I don't find the comment that helpful - I think the code is clear without it. But I also don't mind leaving it in. Fixed in `eb4547ad04 [fold] Re add comment`

src/ripple/app/tx/impl/applySteps.cpp
seelabs: The `[&]` will default capture all the needed variables by ref. So `view` and `tx` _are_ both captured. This could have been written as `[&view, &tx]` but there's no reason to call attention to what variables are captured.

ckeshava: On a related note, will the const-ness of these refs be preserved through the lambda? I remember watching a lecture about this a long time ago, but I'm not able to discern it now.

ckeshava: thanks! this makes a lot more sense now 👍

ckeshava: i found some solutions for the const question here: https://stackoverflow.com/questions/3772867/lambda-capture-as-const-reference

mvadari (commented):
I'm not entirely sure this refactor will play nicely with plugins, but it doesn't really hurt anything, because the plugin code can just continue to do what it's doing, with separate code in each of the methods.

mvadari (approved):

intelliot:
@seelabs - do you you have an opinion on whether this should go in 2.0 (with XChainBridge and lots of other changes), or wait for 2024?

I hope this is a safe refactor, but you never know ;)

ckeshava (approved):

ckeshava:
@mvadari What do you mean by plug-in code? Is it a different PR?

mvadari:
> @mvadari What do you mean by plug-in code? Is it a different PR?

It will be in a future PR - it's still WIP.

seelabs:
@intelliot I don't think we need to delay this. I think we can merge this into the next release.

gregtatcam (approved):
👍 Nice update! One minor suggestion to consider - change `std::enable_if_t` in `consequences_helper()` to `requires`. It just makes it a bit more readable.

seelabs:
Force pushed to resync with my local branch. I also disabled maintainers from pushing to this branch. It's easier for me to track things if I'm the only one who pushes.

intelliot:
note: do not merge until after PR author applies `Passed` label

seelabs:
@gregtatcam Nice suggestion! I agree the "requires" is nicer (it's not part of vocabulary yet so reached for `enable_if` just out of habbit). Fixed in `c887c0bfd0 [fold] Use requires clause in place of enable_if`. Can you do me a favor and give a thumbs up after you take a quick look so I know it's OK with you.

gregtatcam:
> @gregtatcam Nice suggestion! I agree the "requires" is nicer (it's not part of vocabulary yet so reached for `enable_if` just out of habbit). Fixed in `c887c0bfd0 [fold] Use requires clause in place of enable_if`. Can you do me a favor and give a thumbs up after you take a quick look so I know it's OK with you.

Sure, will do.

gregtatcam (approved):
👍 LGTM

ximinez (approved):
I freakin' love this. Approved as-is.

...But I am going to throw you a curve ball. What do you think of getting rid of the exception type, and using an `UnknownTxn` placholder transactor instead? To me, the big advantage is that it puts all the error conditions into one place (the class) instead of spread around the `catch` clauses.

Here's the change: https://github.com/seelabs/rippled/compare/refactor-apply-steps...ximinez:rippled:sd-4710-apply-steps

seelabs:
@ximinez (I really wish I could respond to top level comment in github). That's an interesting design! It is nice that it gets rid of the try/catch. My biggest objection is it's a non-standard way to handle errors. It's relatively easy to look at the current code and understand what's going on - exceptions mean errors, there's a try/catch block...yeah, yeah, yeah we've seen it a million times. Your design is codes the error as concrete class that returns the correct error code. It's really clever! I kinda like it! But it's also a way of handling errors in a way we don't usually do in rippled. I'm a little hesitant to introduce another error handling mechanism.

Having said all that, I do like what you did. What do you think about getting this in now and maybe do your patch as a later PR? (I'll admit another minor concern is another round of reviews, and I'm trying to get this merged; But that shouldn't be a strong criteria).

seelabs:
Squashed and rebased onto the latest develop

PR #4720, opened 2023-09-21
Disallow the same bridge on one chain:

seelabs:
Before this patch, two door accounts on the same chain could could own the same bridge spec (of course, one would have to be the issuer and one would have to be the locker). While this is silly, it does not violate any bridge invariants. However, on further review if we allow this then the `claim` transactions would need to change. Since it's hard to see a use case for two doors to own the same bridge, this patch disallows it. (The transaction will return tecDUPLICATE).

### DIFF
--- a/src/ripple/app/tx/impl/XChainBridge.cpp
+++ b/src/ripple/app/tx/impl/XChainBridge.cpp
@@ -1452,13 +1452,19 @@ XChainCreateBridge::preclaim(PreclaimContext const& ctx)
 {
     auto const account = ctx.tx[sfAccount];
     auto const bridgeSpec = ctx.tx[sfXChainBridge];
-
     STXChainBridge::ChainType const chainType =
         STXChainBridge::srcChain(account == bridgeSpec.lockingChainDoor());
 
-    if (ctx.view.read(keylet::bridge(bridgeSpec, chainType)))
     {
-        return tecDUPLICATE;
+        auto hasBridge = [&](STXChainBridge::ChainType ct) -> bool {
+            return ctx.view.exists(keylet::bridge(bridgeSpec, ct));
+        };
+
+        if (hasBridge(STXChainBridge::ChainType::issuing) ||
+            hasBridge(STXChainBridge::ChainType::locking))
+        {
+            return tecDUPLICATE;
+        }
     }
 
     if (!isXRP(bridgeSpec.issue(chainType)))
--- a/src/test/app/XChain_test.cpp
+++ b/src/test/app/XChain_test.cpp
@@ -472,15 +472,16 @@ struct XChain_test : public beast::unit_test::suite,
                 mcBob, bridge(mcBob, mcGw["USD"], mcBob, mcGw["USD"])),
             ter(temXCHAIN_EQUAL_DOOR_ACCOUNTS));
 
-        // Both door accounts are on the same chain is allowed (they likely
-        // belong to different chains. If they do belong to the same chain, that
-        // is silly, but doesn't violate any invariants).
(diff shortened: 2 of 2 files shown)


src/test/app/XChain_test.cpp
```diff
@@ -631,10 +632,12 @@ struct XChain_test : public beast::unit_test::suite,
         env.tx(create_bridge(a1, bridge(a1, GUSD, B, BUSD))).close();
         env.tx(create_bridge(a2, bridge(a2, GUSD, B, BUSD))).close();
 
-        // Add the exact same bridge to two different accoutns (one must locking
-        // account and one must be issuing)
+        // Add the exact same bridge to two different accoutns (one locking
```
mvadari, line 635: While you're here:
```suggestion
        // Add the exact same bridge to two different accounts (one locking
```

src/test/app/XChain_test.cpp
```diff
@@ -631,10 +632,12 @@ struct XChain_test : public beast::unit_test::suite,
...
+        // account and one issuing)
         env.tx(create_bridge(a3, bridge(a3, GUSD, a4, a4["USD"]))).close();
-        env.tx(create_bridge(a4, bridge(a3, GUSD, a4, a4["USD"]))).close();
+        env.tx(create_bridge(a4, bridge(a3, GUSD, a4, a4["USD"])),
+               ter(tecDUPLICATE))
+            .close();
```
mvadari, line 640: Perhaps this should check the reverse situation compared to the above test change - issuing chain then locking chain.

src/test/app/XChain_test.cpp
seelabs: Fixed in `211ccc6fd9 [fold] Add additional test`

src/test/app/XChain_test.cpp
seelabs: Fixed in `211ccc6fd9 [fold] Add additional test`

mvadari (approved):

src/ripple/app/tx/impl/XChainBridge.cpp
```diff
@@ -1452,13 +1452,19 @@ XChainCreateBridge::preclaim(PreclaimContext const& ctx)
...
     auto const bridgeSpec = ctx.tx[sfXChainBridge];
-
     STXChainBridge::ChainType const chainType =
         STXChainBridge::srcChain(account == bridgeSpec.lockingChainDoor());
 
-    if (ctx.view.read(keylet::bridge(bridgeSpec, chainType)))
     {
-        return tecDUPLICATE;
+        auto hasBridge = [&](STXChainBridge::ChainType ct) -> bool {
+            return !!ctx.view.read(keylet::bridge(bridgeSpec, ct));
```
scottschurr, line 1460: As a small optimization, consider calling `ctx.view.exists()` instead, like this:
```
return ctx.view.exists(keylet::bridge(bridgeSpec, ct));
```
This will avoid deserializing the bridge to check the `LedgerEntryType`.  Just a thought.

scottschurr (approved):
LGTM :+1:.  I've offered a small optimization to consider, but it's at your discretion.

src/ripple/app/tx/impl/XChainBridge.cpp
seelabs: Love it. Fixed in `8bbf7b9e21 [fold] Use exists instead of read to check if ledger object exists`

scottschurr (approved):
:+1: Looks great.

intelliot:
note: have tests for this (Ram)

PR #4721, opened 2023-09-21
fixDisallowIncomingV1

dangell7:
This commit introduces a new fix amendment (fixDisallowIncomingV1) which fixes an issue that occurs with authorized trustlines and the `lsfDisallowIncomingTrustline` flag.

Currently if a user creates an authorized trustline then sets the `lsfDisallowIncomingTrustline` flag on their account, the gateway/issuer cannot then approve the trustline.

To recreate the issue:

1. Issuer sets `asfRequireAuth` on their account.
2. User sets `asfDisallowIncomingTrustline` on their account.
3. User submits SetTrust tx to Issuer.
4. Issuer cannot  then authorize the trustline.

Once the amendment is activated, the issuer can authorize the trustline after the User sets the `asfDisallowIncomingTrustline` flag because the trustline already exists.


src/ripple/app/tx/impl/SetTrust.cpp
```diff
@@ -133,14 +133,22 @@ SetTrust::preclaim(PreclaimContext const& ctx)
...
         if (sleDst->getFlags() & lsfDisallowIncomingTrustline)
-            return tecNO_PERMISSION;
+        {
+            if (ctx.view.rules().enabled(fixDisallowIncomingV1) &&
+                ctx.view.exists(keylet::line(id, uDstAccountID, currency)))
+            {
+                // pass
+            }
+            else
+                return tecNO_PERMISSION;
```
mvadari, line 150: Why not just invert the `if` and return inside it, instead of having an empty `if` and returning in the `else`?

dangell7: I actually have it like that on my local branch but with `clang-format` it makes it very "compact". I'm open to doing it like that if its preferred. It was just done like that for readability.

```
if (ctx.view.rules().enabled(fixDisallowIncomingV1) &&
        !ctx.view.exists(keylet::line(id, uDstAccountID, currency)))
        return tecNO_PERMISSION;
```

RichardAH: In general positive conditions should be preferred over negative conditions. They are quick and easy to scan through, it's very obvious to the reader what's happening. The more complicated the condition the more likely the next programmer will misunderstand it and insert a mistake.

scottschurr:
I just looked over this pull request.  The proposed approach looks like it would work and solve the identified problem.

However, I'm not yet convinced that the identified problem justifies an amendment.  Let me walk through why.

If `asfDisallowIncomingTrustline` were a sticky flag (that couldn't be cleared), then I'd say you have a great justification.  But the flag is not sticky.  `asfDisallowIncomingTrustline` can be set and cleared as often as an account wants.  So if an account gets into the situation solved by this `fix`, they can also solve the problem by

1. Clearing their `asfDisallowIncomingTrustline` flag briefly,
2. Letting the `tfSetfAuth` get set on the trust line, and then
3. Re-enabling their `asfDisallowIncomingTrustline` flag.

This unit test fragment demonstrates the approach I'm suggesting.
```
        {
            Env env{*this, (features | disallowIncoming) - fixDisallowIncomingV1};
            auto const dist = Account("dist");
            auto const gw = Account("gw");
            auto const USD = gw["USD"];
            auto const distUSD = dist["USD"];

            env.fund(XRP(1000), gw, dist);
            env.close();

            env(fset(gw, asfRequireAuth));
            env.close();

            env(fset(dist, asfDisallowIncomingTrustline));
            env.close();

            env(trust(dist, USD(10000)));
            env.close();

            // Can't set trust line because dist has set the
            // asfDisallowIncomingTrustline flag.
            env(trust(gw, distUSD(10000)),
                txflags(tfSetfAuth),
                ter(tecNO_PERMISSION));
            env.close();

            // But dist can clear the asfDisallowIncomingTrustline flag.  Then
            // the trust line can have `tfSetfAuth` set.
            env(fclear(dist, asfDisallowIncomingTrustline));
            env.close();

            env(trust(gw, distUSD(10000)), txflags(tfSetfAuth));
            env.close();

            // Now dist can re-enable asfDisallowIncomingTrustline.
            env(fclear(dist, asfDisallowIncomingTrustline));
            env.close();

            env(pay(gw, dist, USD(1000)));
            env.close();
        }
```
This approach has the advantage that it works today (without an amendment) and it's pretty easy to explain.

Is it elegant?  No.  But it solves the problem for what I suspect is a pretty small user population.

I'm also open to being wrong on this.  Convince me that this proposed amendment has a good justification.

Thanks for reading.

dangell7:
> I just looked over this pull request. The proposed approach looks like it would work and solve the identified problem.
> 
> However, I'm not yet convinced that the identified problem justifies an amendment. Let me walk through why.
> 
> If `asfDisallowIncomingTrustline` were a sticky flag (that couldn't be cleared), then I'd say you have a great justification. But the flag is not sticky. `asfDisallowIncomingTrustline` can be set and cleared as often as an account wants. So if an account gets into the situation solved by this `fix`, they can also solve the problem by
> 
> 1. Clearing their `asfDisallowIncomingTrustline` flag briefly,
> 2. Letting the `tfSetfAuth` get set on the trust line, and then
> 3. Re-enabling their `asfDisallowIncomingTrustline` flag.
> 
> This unit test fragment demonstrates the approach I'm suggesting.
> 
> ```
>         {
>             Env env{*this, (features | disallowIncoming) - fixDisallowIncomingV1};
>             auto const dist = Account("dist");
>             auto const gw = Account("gw");
>             auto const USD = gw["USD"];
>             auto const distUSD = dist["USD"];
> 
>             env.fund(XRP(1000), gw, dist);
>             env.close();
> 
>             env(fset(gw, asfRequireAuth));
>             env.close();
> 
>             env(fset(dist, asfDisallowIncomingTrustline));
>             env.close();
> 
>             env(trust(dist, USD(10000)));
>             env.close();
> 
>             // Can't set trust line because dist has set the
>             // asfDisallowIncomingTrustline flag.
>             env(trust(gw, distUSD(10000)),
>                 txflags(tfSetfAuth),
>                 ter(tecNO_PERMISSION));
>             env.close();
> 
>             // But dist can clear the asfDisallowIncomingTrustline flag.  Then
>             // the trust line can have `tfSetfAuth` set.
>             env(fclear(dist, asfDisallowIncomingTrustline));
>             env.close();
> 
>             env(trust(gw, distUSD(10000)), txflags(tfSetfAuth));
>             env.close();
> 
>             // Now dist can re-enable asfDisallowIncomingTrustline.
>             env(fclear(dist, asfDisallowIncomingTrustline));
>             env.close();
> 
>             env(pay(gw, dist, USD(1000)));
>             env.close();
>         }
> ```
> 
> This approach has the advantage that it works today (without an amendment) and it's pretty easy to explain.
> 
> Is it elegant? No. But it solves the problem for what I suspect is a pretty small user population.
> 
> I'm also open to being wrong on this. Convince me that this proposed amendment has a good justification.
> 
> Thanks for reading.

I would argue a few things. 

1. This means that the issuer needs to communicate to me, for me to clear my flag then reset it.
2. They/I would need to do this every time, for every issuer that has an authorized trustline.
3. I already gave permission/trust to the issuer through adding their trustline therefore this process is redundant.

mvadari:
> @dangell7 wrote:
> 
> > I would argue a few things.
> > 
> > 1. This means that the issuer needs to communicate to me, for me to clear my flag then reset it.
> > 2. They/I would need to do this every time, for every issuer that has an authorized trustline.
> > 3. I already gave permission/trust to the issuer through adding their trustline therefore this process is redundant.
> 
> Thanks for bringing those up. I think all of those are legitimate points. However I'm not completely convinced they are important enough to merit an amendment.
> 
> On the other hand I recognize that I'm not a very good representative of ledger usability concerns. So my hope is that someone with more usability focus will weigh in. Possibly @mvadari or @ximinez?

While I agree that the flow @scottschurr described would be an easier solution to the problem (easier in that it doesn't require enabling an amendment), I think this amendment fixes the behavior to be what it _should_ be. IMO this amendment should still be approved, but in the meanwhile we can use other mechanisms to deal with the problem (and the explanations will be a bit simpler when the fix is enabled). I'm of the opinion that if a fix amendment isn't going to change the user API (which would be a pain to deal with in tooling), then there's no reason why it shouldn't be in the codebase.

src/ripple/app/tx/impl/SetTrust.cpp
```diff
@@ -133,14 +133,22 @@ SetTrust::preclaim(PreclaimContext const& ctx)
...
+    // then honour that flag ONLY if the trustline doesn't exist
     if (ctx.view.rules().enabled(featureDisallowIncoming))
     {
         if (!sleDst)
             return tecNO_DST;
 
         if (sleDst->getFlags() & lsfDisallowIncomingTrustline)
-            return tecNO_PERMISSION;
+        {
+            if (ctx.view.rules().enabled(fixDisallowIncomingV1) &&
```
scottschurr, line 144: If I were reading this code in the future, I think I'd find the changed comment on line 136 misplaced.  My inclination is to remove the change from line 136 and, instead, add something like this comment just in front of this `if`:
```
            // The original implementation of featureDisallowIncoming was
            // too restrictive.  If
            //   o fixDisallowIncomingV1 is enabled and
            //   o The trust line already exists
            // Then allow the TrustSet.
```
I think this is an improvement.  But it's only a suggestion.  Take it, modify it, or leave it as you will.

scottschurr (approved):
:+1: Nice job with this pull request.  The implementation is straightforward and easy to read.  The unit test demonstrates the intention and utility of the change.

I left one note about a comment that might be changed.  But I leave that your discretion.

mvadari (approved):

intelliot:
This PR has sufficient approvals. @dangell7 , please:
1. Bring this branch up-to-date with `develop`
2. Confirm that, from your perspective, this PR is ready to merge.

Thanks!

dangell7:
LGTM

Issue #4015, opened 2021-12-07
Rename the port_grpc section to avoid confusion


cjcobb23:
The gRPC related config info needs to be in a section called `port_grpc`. However, this section cannot be added to the `server` section, or else rippled will fail to start. This is because the code that handles the other ports (and reads from the `server` section) does not handle gRPC. I think the cleanest thing here would be to change the name from `port_grpc` to just `gRPC`, though there are a variety of other solutions here. Regardless, this trips a lot of people up and it's not clear to them what's wrong.

Issue #4557, opened 2023-06-05
grpc port config included server block bug (Version: 1.10.1)


interc0der:
## Issue Description
When building the rippled config, the grpc port cannot be included in the server block stanza. 
If included, rippled will not boot. This seems odd, and I am suggesting that all ports/server stanzas should be included in the server block.

## Steps to Reproduce

A handful of different configurations were tested to verify the behavior.
1. All ports removed and isolated down to grpc_port - node stop
2. Isolated down to grpc_port and add protocol - node stop
3. grpc_port removed from server block - successful boot

___

1. All ports removed and isolated down to grpc_port - node stop

Reason : 
```
2023-May-25 06:23:58.886615258 UTC Application:ERR Missing 'protocol' in [port_grpc]
```

```
[server]
port_grpc
#ssl_key = /etc/letsencrypt/live/
#ssl_cert = /etc/letsencrypt/live/

[port_grpc]
port = 50051
ip = 0.0.0.0
secure_gateway = 127.0.0.1
```
___

2. Isolated down to grpc_port and add protocol - node stop

Reason: 
```
2023-May-25 06:25:31.239412338 UTC gRPC Server:NFO Starting gRPC server at 0.0.0.0:50051
E0525 06:25:31.239850513       1 chttp2_server.cc:1057]      UNKNOWN:No address added out of total 1 resolved for '0.0.0.0:50051' {created_time:"2023-05-25T06:25:31.239801722+00:00", children:[UNKNOWN:Failed to add any wildcard listeners {created_time:"2023-05-25T06:25:31.239796042+00:00", children:[UNKNOWN:Address family not supported by protocol {created_time:"2023-05-25T06:25:31.239732694+00:00", errno:97, os_error:"Address family not supported by protocol", syscall:"socket", target_address:"[::]:50051"}, UNKNOWN:Unable to configure socket {created_time:"2023-05-25T06:25:31.239763291+00:00", fd:32, children:[UNKNOWN:Address already in use {syscall:"bind", os_error:"Address already in use", errno:98, created_time:"2023-05-25T06:25:31.239761217+00:00"}]}]}]}
```

```
[server]
port_grpc
#ssl_key = /etc/letsencrypt/live/
#ssl_cert = /etc/letsencrypt/live/

[port_grpc]
port = 50051
ip = 0.0.0.0
secure_gateway = 127.0.0.1
protocol = http
```

___
3. grpc_port removed from server block - successful boot

```
[server]
[...]

PR #4728, opened 2023-09-25
Validate grpc port info in GRPCServer ctor

ckeshava:
Fix https://github.com/XRPLF/rippled/issues/4015, https://github.com/XRPLF/rippled/issues/4557: GRPCServer and ServerHandler are created during the construction of Application instance. Do not validate GRPC IP/Port/Protocol information in ServerHandler. Instead, pass the responsibility of grpc port information validation to GRPCServer constructor. This ensures that configuration file can specify the port_grpc section.

### Context of Change
Based on suggestions from @ximinez , I implemented this code change. @cjcobb23 suggests an alternative solution to the same issue https://github.com/XRPLF/rippled/issues/4015

I welcome any other suggestions too. I felt this is a clean way to handle this problem without breaking backwards compatibility.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

This pertains to renaming of variables inside the codebase. It should suffice to test it through the existing unit tests.


src/ripple/app/main/GRPCServer.cpp
```diff
@@ -427,9 +428,9 @@ GRPCServerImpl::GRPCServerImpl(Application& app)
     : app_(app), journal_(app_.journal("gRPC Server"))
 {
     // if present, get endpoint from config
-    if (app_.config().exists("port_grpc"))
+    if (app_.config().exists(SECTION_PORT_GRPC))
     {
-        const auto& section = app_.config().section("port_grpc");
+        Section section = app_.config().section(SECTION_PORT_GRPC);
```
ximinez, line 433: Don't make a copy of the `Section`. 
```
Section const& section = ...
```

ximinez (changes requested):
This is only a partial review after taking a quick glance

src/ripple/app/main/GRPCServer.cpp
ckeshava: fixed in dc4a1beb066afbce6ed123923b1d134aea68e372

ximinez (changes requested):
One minor suggestion. Also, please merge in the latest develop branch.

src/ripple/core/ConfigSections.h
```diff
@@ -101,6 +101,8 @@ struct ConfigSection
 #define SECTION_SWEEP_INTERVAL "sweep_interval"
 #define SECTION_NETWORK_ID "network_id"
 
+#define SECTION_PORT_GRPC "port_grpc"
```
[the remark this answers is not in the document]
ckeshava: The last 6 `#defines` are not in alphabetical order, I've fixed them too

ximinez (approved):

legleux (approved):

intelliot:
@ckeshava confirmed this is ready from his perspective; adding `Passed` label.

intelliot:
@ckeshava @legleux - did you confirm that this PR, as-is, fully fixes the issue reported in https://github.com/XRPLF/rippled/issues/4557#issue-1742480269?

Specifically:
> 1. All ports removed and isolated down to grpc_port - node stop
> 2. Isolated down to grpc_port and add protocol - node stop
> 3. grpc_port removed from server block - successful boot


In all of those cases, with this PR, the node will successfully boot with a correctly configured grpc_port, right?

intelliot:
Suggested commit message:
```
feat: allow `[port_grpc]` to be specified in `[server]` stanza (#4728)

Instead of validating gRPC IP/Port/Protocol information in
ServerHandler, validate grpc port info in GRPCServer constructor. This
should not break backwards compatibility.

gRPC-related config info must be in a section (stanza) called
[port_gprc]. Prior to this commit, this section could not be added to
the [server] stanza.

* Close #4015 - That was an alternate solution. It was decided that with
  relaxed validation, it is not necessary to rename port_grpc.
* Fix #4557
```

legleux:
:+1: 
With this PR, `rippled` no longer stops with 
```
Application:ERR Missing 'protocol' in [port_grpc]
```
when `port_grpc` appears in the `[server]` block which was the cause of confusion.

intelliot:
Thanks. I think that's ok - in fact it's probably what we want, as the change should be 100% backward-compatible with any existing working cfg.

intelliot:
Suggested commit message (updated):
```
feat: allow `port_grpc` to be specified in `[server]` stanza (#4728)

Prior to this commit, `port_grpc` could not be added to the [server]
stanza. Instead of validating gRPC IP/Port/Protocol information in
ServerHandler, validate grpc port info in GRPCServer constructor. This
should not break backwards compatibility.

gRPC-related config info must be in a section (stanza) called
[port_gprc].

* Close #4015 - That was an alternate solution. It was decided that with
  relaxed validation, it is not necessary to rename port_grpc.
* Fix #4557
```

ckeshava:
@legleux  Thanks for verifying the output of this PR! I was not able to look into this one for the past few days

PR #4618, opened 2023-07-12
APIv2(gateway_balances, channel_authorize): update errors

PeterChen13579:
## High Level Overview of Change
Fixed/added error messages for Gateway_balance(issue https://github.com/XRPLF/rippled/issues/4290 | https://github.com/XRPLF/rippled/issues/4548 ) and channel_authorize(https://github.com/XRPLF/rippled/issues/4289 )

### Context of Change
Added error messages when user sends an incorrect request. (Check the above Github issue link for more info.) Everything is under API Version 2 as it is a breaking change.

### Type of Change

- [ X] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ X] Tests (You added tests for code that already exists, or your new feature included in this PR)

Added own unit test for API Version 2;

PR #4731, opened 2023-09-27
[TRIVIAL] Fix build error: `uint` is not universally defined

ximinez:

Fixes the Windows build

### Context of Change

Introduced by #4618

### Type of Change

### DIFF
--- a/src/ripple/rpc/impl/RPCHelpers.h
+++ b/src/ripple/rpc/impl/RPCHelpers.h
@@ -296,7 +296,7 @@ std::pair<PublicKey, SecretKey>
 keypairForSignature(
     Json::Value const& params,
     Json::Value& error,
-    uint apiVersion = apiVersionIfUnspecified);
+    unsigned int apiVersion = apiVersionIfUnspecified);
 /** Helper to parse submit_mode parameter to RPC submit.
  *
  * @param params RPC parameters


thejohnfreeman (approved):

ckeshava (approved):

ckeshava:
Is there a reason why `uint` isn't defined for windows platform? I feel this would increase the compatibility between linux, mac and windows, from a developer experience perspective.

ximinez:
> Is there a reason why `uint` isn't defined for windows platform? I feel this would increase the compatibility between linux, mac and windows, from a developer experience perspective.

I really have no idea. I am pretty sure that it's not defined in the standard as a separate type (https://en.cppreference.com/w/cpp/types/integer). My best guess is that it's either a compiler extension, or some slight difference in the dependency/header tree from one of the included libraries. We shouldn't rely on either when `unsigned int` *is* part of the standard and works perfectly well here. 

Also notice that I only changed the `.h` file. [It's already defined as `unsigned int` in the `.cpp` file.](https://github.com/ximinez/rippled/blob/uint-fix/src/ripple/rpc/impl/RPCHelpers.cpp#L849-L853)

intelliot:
note: confirmed with ximinez this is ready to merge

Issue #4736, opened 2023-10-02
Amendments not enabled on networks with network_id greater than 1024 (Version: 2.0.0-b1)


sgramkumar:
## Issue Description
Enabling amendments on network with network_id upto 1024 works fine, but on network with id greater than 1024, amendments are not enabled. Probably this is because the transaction doing the voting is not including the `NetworkID` [which is required if network_id > 1024](https://xrpl.org/transaction-common-fields.html#networkid-field) and so those transactions are not closed.

**On networks with network_id <= 1024**
- network_id: 258 => 2:00pm-2:56pm - amendments enabled
- network_id: 1020 => 3:04pm-4:50pm - amendments enabled

**On networks with network_id > 1024**
- network_id: 1030 => 3:04pm-4:50pm - amendments not enabled
- network_id: 1234 => waited more than 12 hrs wait - amendments not enabled

## Steps to Reproduce
1. Update rippled.cfg with `network_id` to a value greater than 1024
```
[network_id]
1030
```
2. Set `amendment_majority_time` to 30 mins to accelerate amendment voting 
```
[amendment_majority_time]
30 minutes
```
3. Make sure to remove `[features]` stanza
4. Bring up a private network (2 validators are enough to reproduce this issue)
5. Once the network is up, vote to enable an amendment (for example: Checks)
`/opt/ripple/bin/rippled --conf /opt/ripple/etc/rippled.cfg feature Checks accept`
6. Wait for 30-45 mins (for the next flag ledger to cross)
7. Verify with command: `rippled feature | grep -A5 -B3 Checks`

## Expected Result
Amendment should be enabled for `Checks` with this response

```
         "157D2D480E006395B76F948E3E07A45A05FE10230D88A7993C71F97AE4B1F2D1" : {
            "enabled" : true,
            "name" : "Checks",
            "supported" : true
         },
```

## Actual Result
But on network with network_id > 1024, amendment is not enabled after 45 mins too

Command: `rippled feature | grep -A5 -B3 Checks`

```
         "157D2D480E006395B76F948E3E07A45A05FE10230D88A7993C71F97AE4B1F2D1" : {
            "count" : 5,
            "enabled" : false,
            "name" : "Checks",
            "supported" : true,
            "threshold" : 3,
            "validations" : 5,
            "vetoed" : false
         },

```

## Environment
Any local network with network_id > 1024

intelliot:
- [Context on NetworkID](https://github.com/XRPLF/rippled/pull/4370)

PR #4737, opened 2023-10-02
Network ID logic should not be applied to pseudo-transactions

dangell7:
This PR fixes the issue with enabling new amendments on a network with an ID > 1024. 

See an example here: https://github.com/XRPLF/rippled/issues/4736

- [Context on pseudo-transactions](https://xrpl.org/pseudo-transaction-types.html)


intelliot:
Just curious - is there a reason this solution is better than adding the NetworkID field to pseudo-transactions?

mvadari:
> Just curious - is there a reason this solution is better than adding the NetworkID field to pseudo-transactions?

I don't fully understand how pseudo-transactions work, but is there a chance that a signed pseudo-transaction could be submitted on another network to cause unwanted changes?

ximinez:
Pseudo-transactions don't get signed, nor are they submitted. I think it's possible to request one, but I'm not 100% sure of that. Point is, that's not a risk.

That said, adding the NetworkID to pseudo-transactions _should_ be an equally viable solution. Perhaps better, because someone seeing one of these pseudo-txs would know exactly where it came from.

RichardAH:
> That said, adding the NetworkID to pseudo-transactions _should_ be an equally viable solution. Perhaps better, because someone seeing one of these pseudo-txs would know exactly where it came from.
> 
> P.S. Either way, unit tests should be added to ensure that pseudo-transactions are generated and processed correctly with a network ID > 1024.

It's not obvious to me why pseudos would ever need a network ID? If you want to differentiate networks just add NetworkID to the fees (or another singleton) object?

intelliot:
It would never be _needed_. When looking some pseudo-transaction JSON, you would be able to see which network it came from, however.

src/ripple/app/tx/impl/Transactor.cpp
```diff
@@ -40,25 +40,32 @@ namespace ripple {
 NotTEC
 preflight0(PreflightContext const& ctx)
 {
-    uint32_t const nodeNID = ctx.app.config().NETWORK_ID;
-    std::optional<uint32_t> const txNID = ctx.tx[~sfNetworkID];
-
-    if (nodeNID <= 1024)
+    if (isPseudoTx(ctx.tx))
```
gregtatcam, line 43: It's better not to have the empty block and check `if (!isPseudoTx(ctx.tx))`

gregtatcam: I'm not having much luck with testing this fix. I run a local network of 5 nodes. I added this to config:
>      [network_id]
>      1026
>      [amendment_majority_time]
>      16 minutes
I ran this to every node:
>              {
>                  "method": "feature",
>                  "params": [
>                    {
>                      "feature": "AMM",
>                      "vetoed": false
>                    }
>                  ]
>                }
This is the response:
>       {'count': 1,
>         'enabled': False,
>         'name': 'AMM',
>         'supported': True,
>         'threshold': 3,
>         'validations': 5,
>         'vetoed': False}

gregtatcam: I'll re-test with the suggested wait time to double check.

gregtatcam: I waited 45 minutes - the same result.

intelliot:
Notes
- Concern: consistency, and having all transactions follow the same rules
- May be preferable to avoid making exceptions which may be forgotten in the future
- Should try to minimize exceptions that apply only to pseudo-transactions

intelliot:
Just to close the loop on my earlier question: I'm personally fine with this solution, given these are pseudo-transactions, which are already different from normal transactions anyway. It makes sense to review (and potentially merge) this change as-is.

src/ripple/app/tx/impl/Transactor.cpp
```diff
@@ -40,25 +40,33 @@ namespace ripple {
...
     {
-        // legacy networks have IDs 1024 and below. These networks cannot
-        // specify NetworkID in txn
-        if (txNID)
-            return telNETWORK_ID_MAKES_TX_NON_CANONICAL;
+        // all emitted and pseudo transactions are free to pass, do not need
+        // network id
```
seelabs, line 46: What should we do if a pseudo transaction has a network id? Right now it's ignored, which is probably fine. But some other options would be:
1) If present, it needs to follow the same rules as other transactions (i.e. match the configured network id), fail if the network is less than 1024.
2) Fail if present in any form.
3) Assert if present in any form, but allow to continue processing in release mode.

I'd probably prefer (3). If someone does add a network id to a pseudo transaction, and they did it intentionally, the assert will let them know this code needs to change. But you're call if you want to add it or not. I'm fine either way.

As a side note, I also prefer not to have the empty `if` block, but that's just a preference. I'm also fine if you want to keep it as-is.

ximinez: I'd vote for option 1. It has the advantage of being simple to add that check. Simply change the check to
```
if (isPseudoTx(ctx.tx) && !ctx.tx.isFieldPresent(sfNetworkID))
```

dangell7: I like option 1 because its simple.

seelabs: @dangell7 I don't think that patch is going to work If the network ID is required, but isn't present the new patch let's the transaction succeed.

dangell7: fixed in [d8b2ba4](https://github.com/XRPLF/rippled/pull/4737/commits/d8b2ba42a6b2ba1a327ecb7f8021bdff662f3e75)

seelabs (approved):
:+1:

gregtatcam (approved):
👍

intelliot:
note: needs to be brought up-to-date with `develop`

sgramkumar:
Verified on networks with network_id 12 & 1212. Amendment is enabled in both cases.

PR #4750, opened 2023-10-05
Synergizee `server_info` & `server_state`

dangell7:

Currently `server_info` and `server_state` return 2 different types of values in the rpc. 

`server_info` returns the fee in xrp while `server_state` returns the fee in drops. 

The difference is [here](https://github.com/XRPLF/rippled/blob/4e84ad6cd7ecf7d51e8e8bdded9af195443daa8b/src/ripple/app/misc/NetworkOPs.cpp#L2605) and is only called once with `true` from [here](https://github.com/XRPLF/rippled/blob/4e84ad6cd7ecf7d51e8e8bdded9af195443daa8b/src/ripple/rpc/handlers/ServerInfo.cpp#L38)

Thank you.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

### DIFF
--- a/src/ripple/overlay/impl/OverlayImpl.cpp
+++ b/src/ripple/overlay/impl/OverlayImpl.cpp
@@ -867,8 +867,8 @@ OverlayImpl::getServerInfo()
         Json::Value& validated_ledger = server_info[jss::validated_ledger];
 
         validated_ledger.removeMember(jss::base_fee);
-        validated_ledger.removeMember(jss::reserve_base_xrp);
-        validated_ledger.removeMember(jss::reserve_inc_xrp);
+        validated_ledger.removeMember(jss::reserve_base);
+        validated_ledger.removeMember(jss::reserve_inc);
     }
 
     return server_info;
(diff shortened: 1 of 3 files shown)


ckniffen:
This looks like a good candidate for the v2 api.

intelliot:
@dangell7 - since this is potentially a breaking change for existing `server_info` consumers, could you put this on api_version 2?

mvadari:
I thought this difference was on purpose, since the values in `server_info` are human-readable and the values in `server_state` are machine-readable.

src/ripple/overlay/impl/OverlayImpl.cpp
```diff
@@ -867,8 +867,8 @@ OverlayImpl::getServerInfo()
         Json::Value& validated_ledger = server_info[jss::validated_ledger];
 
         validated_ledger.removeMember(jss::base_fee);
-        validated_ledger.removeMember(jss::reserve_base_xrp);
-        validated_ledger.removeMember(jss::reserve_inc_xrp);
+        validated_ledger.removeMember(jss::reserve_base);
+        validated_ledger.removeMember(jss::reserve_inc);
     }
```
dangell7, line 872: @mvadari thank you for that piece of information. Can we confirm this code here is correct? This was just a weird piece in noticed during this PR.

Issue #4727, opened 2023-09-25
Unify JSON serialization format of transactions (Version: 2.0.0-b2)


intelliot:
There are many API methods that can return transactions. Currently, the return formats are inconsistent. The following table shows where fields are located as of rippled v1.12.0 (unchanged since v1.0.1).

**Note:** In the following table, the "top" level is with respect to the individual object representing the transaction; for methods that return arrays of transactions, the "top" level is per element of the array. "Transaction Instructions" refer to the canonical fields of the transaction as defined in the [transaction format](https://xrpl.org/transaction-formats.html), such as the `Account` and `Flags` fields.

| Method | Transaction Instructions | Transaction Metadata | `hash` | `ledger_index` | `inLedger` | `validated` | `ledger_hash` | Binary blob | `date` |
|----|----|---|---|---|---|---|---|---|---|
| `ledger` (with transactions expanded) | Top level | `metaData` field (JSON) <br> _or_ in `meta` field¹ (binary) | Top level | Above transaction level | None | Above transaction level | None |  `tx_blob` field¹ | None |
| `tx` | Top level | `meta` field | Top level | Top level | Top level | Top level | None | `tx` field¹ | Top level |
| `tx_history` (deprecated) | Top level | None | Top level | Top level | Top level | None | None | Not supported | None |
| `account_tx` | `tx` field | `meta` field | `tx` field | `tx` field | `tx` field | Top level | None | `tx_blob` field¹ | None |
| `transaction_entry` ([would like to deprecate](https://github.com/ripple/rippled/issues/2598)) | `tx_json` field | `metadata` field | `tx_json` field | Top level | None | Top level | Top level | Not supported | None |
| `sign`, `sign_for`, `submit`, and `submit_multisigned` | `tx_json` field | (N/A) | `tx_json` field | (N/A) | (N/A) | (N/A) | (N/A) | `tx_blob` field | None |
| Streams from `subscribe` | `transaction` field | `meta` field² | Top level | Top level² | None | Top level² | Top level² | Not supported | `transaction` field |
| [Data API v2](https://developers.ripple.com/data-api.html#transaction-objects) methods | `tx` field | `meta` field | Top level | Top level | None | None (Data API _only_ reports validated transaction data) | None | `tx`¹ | Top level (as ISO8601 timestamp, not Ripple time) |

¹ Only if the request asked for binary data
² The `transactions_proposed` stream omits these fields because the transactions' outcomes are not yet final.

## Recommendation

[...]

Issue #3638, opened 2020-10-15
Deprecate tx_history RPC call


carlhua:
The tx_history method of the WebSocket and JSON-RPC APIs is problematic in several ways:

- It lacks the options to specify an account or ledger version to narrow down the results to something specific
- The startIndex field is woefully inadequate for pagination (no control over how many transactions to return) and inconsistent with how we do pagination on other APIs (limit, marker)
The only use that it could be for (getting a selection of recent transactions) is better served by subscribe.

We should consider deprecating this call and use `ledger ` with the correct parameters instead.

Exported from RIPD-346

Issue #3539, opened 2020-07-15
Revise or deprecate ledger_header command


carlhua:
The ledger_header admin command provides the ledger header information in JSON and binary formats.

The ledger public command, meanwhile, provides the ledger header information in JSON format only.

Rather than having two commands that overlap so closely in functionality, we should consolidate them. Since it doesn't make sense to retrieve the same information in two different formats at once, the most effective solution would be to:

1. [x] Add a "binary" option to the ledger command (default:false). If true, then return the binary form of the ledger header instead of the JSON form. If false or omitted, return the same information as currently provided.
2. [ ] Deprecate the now completely redundant ledger_header admin command.

Exported from RIPD-889

mDuo13:
I thought this was done already... Now I'm unsure.

cjcobb23:
I confirmed that 1 is already done.

mDuo13:
It's [not documented](https://xrpl.org/ledger-methods.html) either, so in that sense it's kind of like it's deprecated?

mDuo13:
I guess the only remaining step would be to remove `ledger_header` entirely from APIv2?

mDuo13:
The `ledger_header` API method is not currently documented and is not implemented in Clio but is currently still available in rippled under API v1. IMO we should remove it (maybe from both, definitely from API v2)

Bronek:
I'm on it

PR #4759, opened 2023-10-11
Remove tx_history and ledger_header in API version 2

Bronek:

Remove `tx_history` and `ledger_header` methods from API version 2. Also update `RPC::Handler` to allow for methods (or methods implementation) to be version specific. This partially resolves https://github.com/XRPLF/rippled/issues/4727 and fully resolves https://github.com/XRPLF/rippled/issues/3638 and https://github.com/XRPLF/rippled/issues/3539 

### Context of Change

Note design change in `RPC::Handler` - we can now store multiple handlers with the same name, as long as they belong to different (non-overlapping) API versions. This necessarily impacts the handler lookup algorithm and its complexity, however there is no performance loss on x86_64 architecture, and only minimal performance loss on arm64 (around 10ns). This design change gives us extra flexibility evolving API in the future, including other parts of https://github.com/XRPLF/rippled/issues/4727 

### Type of Change

- [X] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation Updates
- [ ] Release

### Before

In any API version `tx_history` and `ledger_header` are recognised methods

### After
In API version 1 `tx_history` and `ledger_header` are a valid API method
In API version 2 `tx_history` and `ledger_header` are not recognised method; `rippled` will return error `unknownCmd` if they are called.


mDuo13:
Fixes #3638

mDuo13:
P.S. while we're at it, maybe do #3539 also?

API-CHANGELOG.md
```diff
@@ -125,6 +125,13 @@ Currently (prior to the release of 2.0), it is available as a "beta" version, me
 
 Since `api_version` 2 is in "beta", breaking changes to V2 can currently be made at any time.
 
+#### Retired methods
+
+In API version 2, the following methods are no longer available:
+
+- `tx_history` - please use other appropriate method, e.g. `tx` or `account_tx` or `ledger` with `transactions` field set to `true`
+- `ledger_header` - please use `ledger` with `binary` field set to `true`
```
mDuo13, line 133: The point of `tx_history` is to get a list of recently applied transactions. The `account_tx` and `ledger` methods are the next best thing for that effect. (I did find one narrow case where tx_history is actually more useful, which is on a chain with very little activity, it can give you recent transactions with fewer steps/foreknowledge than the other two. That need is very niche, mostly served by the ledger command, and less useful because of tx_history's non-standard format, so it's still fine to deprecate IMO.) The `tx` method just looks up an individual transaction if you already know its hash, which is different enough that I think we should not suggest it here.

The `binary` note on `ledger_header` is a red herring; both `ledger` and `ledger_header` return the same (non-binary) info by default. Binary support was the last thing on the `ledger_header` side of the venn diagram, but now the `ledger` command is a superset of the functionality.

```suggestion
- `tx_history` - Instead, use other methods such as `account_tx` or `ledger` with the `transactions` field set to `true`.
- `ledger_header` - Instead, use the `ledger` method.
```

P.S. The word "please" can convey a lot of different tones, not all of them good, especially if we only include it in some places and not others. For the changelog, I think it's best to omit the word "please" from these types of notes.

mDuo13 (commented):
Excited to finally get rid of some of this old cruft!

Bronek:
Apologies to reviewers about all these `forced push`, I was trying to make the micro-benchmark `Handler_test.cpp` little less stupid. Which happens to be first commit in the series because I use it to establish baseline ...

API-CHANGELOG.md
```diff
@@ -125,6 +125,13 @@ Currently (prior to the release of 2.0), it is available as a "beta" version, me
 
 Since `api_version` 2 is in "beta", breaking changes to V2 can currently be made at any time.
 
+#### Retired methods
```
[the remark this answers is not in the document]
Bronek: Thanks !

API-CHANGELOG.md
Bronek: Thanks !

intelliot:
@Bronek - please comment here to confirm when this PR is ready for review

Bronek:
This is ready for review, thanks !

src/ripple/rpc/impl/Handler.cpp
```diff
@@ -174,14 +189,39 @@ Handler const handlerArray[]{
...
     template <std::size_t N>
     explicit HandlerTable(const Handler (&entries)[N])
     {
-        for (std::size_t i = 0; i < N; ++i)
+        for (auto const& entry : entries)
         {
-            auto const& entry = entries[i];
-            assert(table_.find(entry.name_) == table_.end());
-            table_[entry.name_] = entry;
+            assert(!overlappingApiVersion(
```
ckeshava, line 219: Instead of the `assert`, can we throw a `LogicError` instead? I'm not sure if asserts are enabled in all types of builds. If there are duplicate/overlapping handlers, it sounds like we need to throw a halting error message.

src/ripple/rpc/impl/Handler.cpp
```diff
@@ -205,36 +245,47 @@ class HandlerTable
...
     std::vector<char const*>
     getHandlerNames() const
     {
         std::vector<char const*> ret;
-        ret.reserve(table_.size());
         for (auto const& i : table_)
-            ret.push_back(i.second.name_);
+        {
+            // Note, table_ is always ordered, allowing such a simple check
+            if (ret.empty() || std::strcmp(ret.back(), i.second.name_) != 0)
```
ckeshava, line 266: Can we return a `std::set<char const*>` instead of manually checking for duplicates in the vector?

src/ripple/rpc/impl/Handler.cpp
```diff
@@ -205,36 +245,47 @@ class HandlerTable
...
-        assert(table_.find(HandlerImpl::name()) == table_.end());
+        static_assert(HandlerImpl::minApiVer <= HandlerImpl::maxApiVer);
+        static_assert(HandlerImpl::maxApiVer <= RPC::apiMaximumValidVersion);
 
-        Handler h;
-        h.name_ = HandlerImpl::name();
-        h.valueMethod_ = &handle<Json::Value, HandlerImpl>;
-        h.role_ = HandlerImpl::role();
-        h.condition_ = HandlerImpl::condition();
+        assert(!overlappingApiVersion(
```
ckeshava, line 283: At first blush, this check appears to be redundant. We are performing this check  in the `HandlerTable` constructor, do we need to repeat ourselves?

`addHandler` is not being invoked from any other place in the codebase.

src/test/rpc/Handler_test.cpp
```diff
@@ -0,0 +1,114 @@
...
+        std::size_t dummy = 0;
+        auto const [mean, stdev, n] = time(
+            1'000'000,
+            [&](std::size_t i) {
+                auto const d = getHandler(1, false, names[i]);
+                dummy = dummy + i + (int)d->role_;
+            },
+            [&]() -> std::size_t { return distr(prng); });
+
+        std::cout << "mean=" << mean << " stdev=" << stdev << " N=" << n
```
ckeshava, line 98: thanks for computing the metrics. can you also include the unit of time in the print statements? I believe it's micro-seconds for this line.

ckeshava (changes requested):
thanks Bronek, I've put down some comments

src/ripple/rpc/impl/Handler.cpp
ckeshava: I believe we can remove the check in the `HandlerTable` constructor. I verified that the existing unit tests work correctly without this check.

But this is a suggestion, please don't feel its binding to follow it.

src/ripple/rpc/impl/Handler.cpp
ckeshava: I've made the necessary changes for using a `std::set` in at the tip of this [branch](https://github.com/ckeshava/rippled/tree/handlerSet)

feel free use it. this is the [commit hash](b705fc4a3c0ce73e6e93480134e1027ead34ab17)

src/test/rpc/TransactionHistory_test.cpp
```diff
@@ -54,6 +54,21 @@ class TransactionHistory_test : public beast::unit_test::suite
...
+        testcase("Command retired from API v2");
+        using namespace test::jtx;
+        Env env{*this, envconfig(no_admin)};
+
+        Json::Value params{Json::objectValue};
+        params[jss::api_version] = 2;
+        auto const result =
+            env.client().invoke("tx_history", params)[jss::result];
+        BEAST_EXPECT(result[jss::error] == "unknownCmd");
+        BEAST_EXPECT(result[jss::status] == "error");
```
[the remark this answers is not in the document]
Bronek: it's in newly added  `src/test/rpc/LedgerHeader_test.cpp`

src/ripple/rpc/impl/Handler.cpp
Bronek: sure, can do. Note that debug binary is tested in CI (meaning, an assert would be caught), and the population of handlers is exactly the same in both debug and release.

src/ripple/rpc/impl/Handler.cpp
Bronek: thank you, will grab your commit - agree that's an improvement.

src/test/rpc/Handler_test.cpp
Bronek: will do, thanks !

src/test/rpc/Handler_test.cpp
```diff
@@ -0,0 +1,133 @@
...
+#include <limits>
+#include <numeric>
+#include <random>
+
+namespace ripple::RPC {
+
+// NOTE: there should be no need for this function;
+// `std::cout << some_duration` should just work if built with a compliant
+// C++20 compiler. Sadly, we are not using one, as of today
+// TODO: remove this operator<< overload when we bump compiler version
```
Bronek, line 35: @HowardHinnant would you like to comment on this ?

intelliot:
(note: these API methods weren't in Clio, so no need for Clio to change anything here)

src/test/rpc/Handler_test.cpp
```diff
@@ -0,0 +1,133 @@
...
+            handlerNames.begin(), handlerNames.end());
+
+        std::uniform_int_distribution<std::size_t> distr{0, names.size() - 1};
+
+        std::size_t dummy = 0;
+        auto const [mean, stdev, n] = time(
+            1'000'000,
+            [&](std::size_t i) {
+                auto const d = getHandler(1, false, names[i]);
+                dummy = dummy + i + (int)d->role_;
```
ckeshava, line 113: Why do we need to insert this `dummy` computation? Won't it add extra time into the results?

src/test/rpc/Handler_test.cpp
```diff
@@ -0,0 +1,133 @@
...
+    void
+    run() override
+    {
+        reportLookupPerformance();
+    }
+};
+
+BEAST_DEFINE_TESTSUITE(Handler, rpc, ripple);
+
+}  // namespace ripple::RPC
```
ckeshava, line 133: Can you add the extra line at the end of this file? I think it's part of coding convention, it's followed in all the other files.

src/ripple/rpc/impl/Handler.cpp
```diff
@@ -17,11 +17,16 @@
 */
 //==============================================================================
 
+#include <ripple/basics/contract.h>
 #include <ripple/rpc/handlers/Handlers.h>
 #include <ripple/rpc/handlers/Version.h>
 #include <ripple/rpc/impl/Handler.h>
 #include <ripple/rpc/impl/RPCHelpers.h>
 
+#include <chrono>
```
ckeshava, line 26: `chrono` and `type-traits` includes are not required

src/ripple/rpc/impl/Handler.cpp
```diff
@@ -57,6 +62,19 @@ handle(JsonContext& context, Object& object)
     return status;
 };
```
ckeshava, line 63: unrelated to your PR, but there is a stray `;` ^^

src/ripple/rpc/impl/Handler.cpp
```diff
@@ -205,36 +251,46 @@ class HandlerTable
...
+            range.first, range.second, [version](auto const& entry) {
+                return version >= entry.second.minApiVer_ &&
+                    version <= entry.second.maxApiVer_;
+            });
+
+        return i == range.second ? nullptr : &i->second;
     }
 
-    std::vector<char const*>
+    std::set<char const*>
```
ckeshava, line 265: I forgot to add a `[[nodiscard]]` tag, can you please add that?

ckeshava (commented):
I have a few clarifying questions for this PR.

Although the unit tests are sufficient, I'd like to test the API changes via the command line. How can I enable the API-version 2 in my local `rippled`? I set `beta_rpc_api` to `1` in the config file, but that didn't seem to help.

src/ripple/rpc/handlers/Version.h
```diff
@@ -46,23 +46,15 @@ class VersionHandler
...
-        return Role::USER;
-    }
+    static constexpr unsigned minApiVer = 1;
 
-    static Condition
-    condition()
-    {
-        return NO_CONDITION;
-    }
+    static constexpr unsigned maxApiVer = RPC::apiMaximumValidVersion;
```
ckeshava, line 53: Should we add an assert/check in the [constructor](https://github.com/XRPLF/rippled/blob/1fde585003c286841e3884012bef000d1ff5f4ee/src/ripple/rpc/handlers/Version.h#L33) of `VersionHandler` for the compatibility of `apiVersion_`? 

I'm trying to understand the expected behavior for wrong `apiVersion` inputs. At the moment, the unit tests fail (the `result` is a `null` JSON Value). I feel it would be nice to return a descriptive error message for the end-user (Ex: unknown version ?)


We need to ensure that `VersionHandler` and `LedgerHandler` behave in a similar way.

ckeshava: To reproduce the unit test: I set the `apiVersion = 4` [here](https://github.com/Bronek/rippled/blob/6de4b6c175d80880901f7f95c9f05421152adfb7/src/test/rpc/TransactionHistory_test.cpp#L65)

src/test/rpc/Handler_test.cpp
HowardHinnant: MSVC (recent versions) have C++20 chrono.  gcc-14 has it, but hasn't shipped yet.  And LLVM is not close.

src/ripple/rpc/handlers/Version.h
Bronek: I will add assert in `template < . . . class HandlerImpl> handle(JsonContext& context, Object& object)` which is shared by both `LedgerHandler` and `VersionHandler` ; as for the actual behaviour, we expect the user to get `unknownCmd` as presented in unit  tests, no plan for more diagnostic. We get that for the simplicity of handler lookup , in `getHandler`. We can certainly document that this error could be triggered by invalid selection of api version, but that probably belongs to https://xrpl.org  ( @mDuo13 - what's your opinion ?)

src/test/rpc/Handler_test.cpp
ckeshava: i see, that's an interesting point 👍

src/ripple/rpc/impl/Handler.cpp
```diff
@@ -47,6 +50,9 @@ template <class Object, class HandlerImpl>
 Status
 handle(JsonContext& context, Object& object)
 {
+    assert(
+        context.apiVersion >= HandlerImpl.minApiVer &&
```
ckeshava, line 54: My compiler (apple-clang v15) is throwing this error on these two lines:
```
/Users/ckeshavabs/personal-code/rippled/src/ripple/rpc/impl/Handler.cpp:54:31: error: 'HandlerImpl' does not refer to a value
        context.apiVersion >= HandlerImpl.minApiVer &&
                              ^
/Users/ckeshavabs/personal-code/rippled/src/ripple/rpc/impl/Handler.cpp:49:31: note: declared here
template <class Object, class HandlerImpl>
                              ^
/Users/ckeshavabs/personal-code/rippled/src/ripple/rpc/impl/Handler.cpp:55:31: error: 'HandlerImpl' does not refer to a value
        context.apiVersion <= HandlerImpl.maxApiVer);
                              ^
/Users/ckeshavabs/personal-code/rippled/src/ripple/rpc/impl/Handler.cpp:49:31: note: declared here
template <class Object, class HandlerImpl>
```

Did you mean to use the scope-resolution operator? `HandlerImpl::minApiVer` -> this builds for me without a hitch

src/ripple/rpc/handlers/Version.h
ckeshava: Thank you, I verified that this works correctly.

I believe it is better to not construct incorrect `VersionHandler` and `LedgerHandler` objects. Although the `assert` in the `handle` function catches this error, it'd probably be better to catch this issue in the constructors itself, as early as possible. (or possibly in `RPCHelpers.cpp`, even before the ctors)

ckeshava:
This PR looks good. I have left a few more comments, but they are non-binding suggestions.

I want to draw attention to this earlier question (https://github.com/XRPLF/rippled/pull/4759#pullrequestreview-1683130236)
I compile and run this branch of `rippled` on my local machine. The command `rippled tx_history 0` displays old output, I do not get the expected `unknownCmd` error. Isn't this wrong?
The version is set to `2.0.0-b3`([here](https://github.com/XRPLF/rippled/blob/b421945e717a29e81b029adea576b24cd2068051/src/ripple/protocol/impl/BuildInfo.cpp#L36)) Do I need to do anything else to switch to version 2 for my local rippled server?

The unit tests are satisfactory, but I'm not able to verify this change through a local version of `rippled`.

src/ripple/rpc/impl/Handler.cpp
Bronek: you are right, thanks !

Bronek:
> I have a few clarifying questions for this PR.
> 
> Although the unit tests are sufficient, I'd like to test the API changes via the command line. How can I enable the API-version 2 in my local `rippled`? I set `beta_rpc_api` to `1` in the config file, but that didn't seem to help.

I do not think it's possible to select API version in command line; @mDuo13 ?

intelliot:
Yes, I think we can always default to the most recent API version for command line. As [stated here](https://github.com/xrplf/rippled/blob/develop/API-CHANGELOG.md#api-version-1), the command line is intended for ad-hoc usage by humans, not programs or automated scripts. The command line is not meant for use in production code. 

I don't think anyone is using tx_history, not even from the command line. If they are, they are welcome to later open an issue, where we can direct them to use alternative method to get the data they're looking for.

ckeshava (approved):

ximinez:
@ckeshava Change `apiMaximumSupportedVersion` to 2 and rebuild. The command line uses the maximum _non-beta_ API version.
https://github.com/XRPLF/rippled/blob/1.12.0/src/ripple/net/impl/RPCCall.cpp#L1485

src/test/rpc/Handler_test.cpp
```diff
@@ -0,0 +1,133 @@
...
+// `std::cout << some_duration` should just work if built with a compliant
+// C++20 compiler. Sadly, we are not using one, as of today
+// TODO: remove this operator<< overload when we bump compiler version
+std::ostream&
+operator<<(std::ostream& os, std::chrono::nanoseconds ns)
+{
+    return (os << ns.count() << "ns");
+}
+
+// NOTE This is a rather naiive effort at a microbenchmark. Ideally we want
```
ximinez, line 42: Spelling nit: "naive"

src/test/rpc/Handler_test.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+public:
+    void
+    run() override
+    {
+        reportLookupPerformance();
+    }
+};
+
+BEAST_DEFINE_TESTSUITE(Handler, rpc, ripple);
```
ximinez, line 131: Since this is testing performance timing and not really correctness, I think it could be changed to a manual test.
```
BEAST_DEFINE_TESTSUITE_MANUAL(Handler, rpc, ripple);
```

intelliot:
note: we intend to [promote](https://github.com/XRPLF/rippled/pull/4568) API version 2 to non-beta after #4733 is merged, and maybe after #4775; and almost certainly before the first 2.0.0-rc

src/test/basics/PerfLog_test.cpp
```diff
@@ -309,7 +309,9 @@ class PerfLog_test : public beast::unit_test::suite
 
         // Get the all the labels we can use for RPC interfaces without
         // causing an assert.
-        std::vector<char const*> labels{ripple::RPC::getHandlerNames()};
+        std::set<const char*> handlerNames = ripple::RPC::getHandlerNames();
+        std::vector<char const*> labels(
+            handlerNames.begin(), handlerNames.end());
```
ximinez, line 314: You're taking this `set` and converting it to a `vector` a couple of times in unit tests. The only other usage of `getHandlerNames()` is in `PerfLog`, where the collection type doesn't matter as long as it's iterable.

I was going to suggest changing the return type back to `vector`, but the overhead to copy the set is unnecessary in the non-test code. What do you think about adding a function to `TestHelpers.h/cpp`, that returns a `vector`? It could even have the same name, since it would be in the `ripple::test` namespace.

src/test/rpc/Handler_test.cpp
```diff
@@ -0,0 +1,133 @@
...
+    return (os << ns.count() << "ns");
+}
+
+// NOTE This is a rather naiive effort at a microbenchmark. Ideally we want
+// Google Benchmark, or something similar. Also, this actually does not belong
+// to unit tests, as it makes little sense to run it in conditions very
+// dissimilar to how rippled will normally work.
+// TODO as https://github.com/XRPLF/rippled/issues/4765
+
+class Handler_test : public beast::unit_test::suite
```
ximinez, line 48: I don't think the test class shouldn't be in the `ripple::RPC` namespace. `ripple::test` makes more sense.

ximinez (changes requested):
A few minor comments. Overall looks great.

src/ripple/rpc/impl/Handler.cpp
```diff
@@ -174,14 +193,42 @@ Handler const handlerArray[]{
...
+        assert(minVer <= maxVer);
+        assert(maxVer <= RPC::apiMaximumValidVersion);
+
+        for (; range.first != range.second; range.first++)
+        {
+            if (range.first->second.minApiVer_ <= maxVer &&
+                range.first->second.maxApiVer_ >= minVer)
+                return true;
+        }
+        return false;
```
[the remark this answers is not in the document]
Bronek: Yes, exactly, thank you !

src/test/rpc/Handler_test.cpp
Bronek: Oh great, wan't aware of that macro

intelliot:
@Bronek please comment here to let us know when you think all suggestions have been addressed

src/ripple/rpc/handlers/LedgerHandler.h
```diff
@@ -58,23 +59,15 @@ class LedgerHandler
...
-        return "ledger";
-    }
+    static constexpr char name[] = "ledger";
 
-    static Role
-    role()
-    {
-        return Role::USER;
-    }
+    static constexpr unsigned minApiVer = 1;
```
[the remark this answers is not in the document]
Bronek: Thanks, done

src/ripple/rpc/handlers/Version.h
```diff
@@ -46,23 +46,15 @@ class VersionHandler
...
-        return "version";
-    }
+    static constexpr char const* name = "version";
 
-    static Role
-    role()
-    {
-        return Role::USER;
-    }
+    static constexpr unsigned minApiVer = 1;
```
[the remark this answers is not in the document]
Bronek: Thanks, done.

src/ripple/rpc/impl/Handler.cpp
```diff
@@ -197,44 +244,54 @@ class HandlerTable
...
                                    : RPC::apiMaximumSupportedVersion))
             return nullptr;
-        auto i = table_.find(name);
-        return i == table_.end() ? nullptr : &i->second;
+
+        auto const range = table_.equal_range(name);
+        auto const i = std::find_if(
+            range.first, range.second, [version](auto const& entry) {
+                return version >= entry.second.minApiVer_ &&
+                    version <= entry.second.maxApiVer_;
```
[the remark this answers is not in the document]
Bronek: Yes, that looks better - thanks !

src/test/rpc/Handler_test.cpp
Bronek: Thanks, fixed

src/test/basics/PerfLog_test.cpp
Bronek: Thanks, added a helper function in `TestHelpers.h`

Bronek:
> @Bronek please comment here to let us know when you think all suggestions have been addressed

Yes, all addressed

src/test/rpc/Handler_test.cpp
Bronek: Thanks, moved to `test`

Bronek:
This is ready to merge

intelliot:
awaiting updated review from ximinez (verifying the changes that were made)

Bronek:
> awaiting updated review from ximinez (verifying the changes that were made)

Well in that case I will have one more minor improvement ...

ximinez (approved):
Looks good!

intelliot:
Suggested commit message:
```
APIv2: remove tx_history and ledger_header (#4759)

Remove `tx_history` and `ledger_header` methods from API version 2.

Update `RPC::Handler` to allow for methods (or method implementations)
to be API version specific. This partially resolves #4727. We can now
store multiple handlers with the same name, as long as they belong to
different (non-overlapping) API versions. This necessarily impacts the
handler lookup algorithm and its complexity; however, there is no
performance loss on x86_64 architecture, and only minimal performance
loss on arm64 (around 10ns). This design change gives us extra
flexibility evolving the API in the future, including other parts of
#4727.

In API version 2, `tx_history` and `ledger_header` are no longer
recognised; if they are called, `rippled` will return error
`unknownCmd`

Resolve #3638

Resolve #3539
```

PR #4760, opened 2023-10-12
Optimize calculation of close time to avoid impasse and minimize gratuitous proposal changes

mtrippled:
This fixes a bug introduced in https://github.com/XRPLF/rippled/pull/4505 that causes gratuitous position changes, resulting in validator desyncs. 

### Context of Change
This change refactors the way that proposal close times are determined for purposes of a node adopting a new close time. The change in PR4505 which introduced the bug also fixed a bug that resulted in impasses in testing. However, it also increased the likelihood of a peer changing its close time, and, therefore, its position. This fix avoids the impasse but minimizes changing positions. Namely, it will only avoid an impasse if consensus taking a long time already.

### Type of Change

- [ ] Documentation Updates
- [ ] Release

Run throughput test, observe network stability after transaction processing stops.


intelliot:
I applied the clang-format.patch and merged 'develop'. @mtrippled , if that inconveniences you in any way, please feel free to force push over my commits.

src/ripple/consensus/Consensus.h
```diff
@@ -1787,47 +1802,91 @@ Consensus<Adaptor>::updateOurPositions(bool const share)
...
+        // As consensus time increases, the threshVote threshold also increases.
+        // Once threshVote exceeds 60%, no members of either set of validators
+        // will change their close times.
+        //
+        // Avoiding the impasse means that validators should not necessarily
+        // count their own close time towards the total until they know
+        // the most popular among their peers, and then change their vote
+        // to that of the most popular. In this case, the validators in set
+        // t2 from the example would have switched to close time t1 and ended
+        // the network impasse.
```
thejohnfreeman, line 1832: If the validators in t1 do not count their own close times, then they perceive two equal sets of 2 validators each for different close times. Are we sure they won't switch, even if t2 is greater? Perhaps the code handles this correctly (I have not yet analyzed it), but it seems ambiguous in at least this comment.

I'm a little worried that we're adding more and more special cases to the code instead of carefully designing and implementing a general algorithm. I'm disturbed by how long this function is even before this change. Is it possible to state the current consensus algorithm in simple terms? How far has it strayed from the description in the whitepapers? I don't think I'm concerned enough right now to try to get in the way of this change, but I'd like to hear opinions from others. How much confidence do we have in the convergence of consensus at this point?

mtrippled: I observed the impasse behavior in testing, and commented on how specifically it occurred:
// Here's how the impasse could occur:
        // Imagine 5 validators. 3 have close time t1, and 2 have t2.
        // As consensus time increases, the threshVote threshold also increases.
        // Once threshVote exceeds 60%, no members of either set of validators
        // will change their close times.
       
This is a point fix to a problem observed in testing that also potentially affects the live network. It is not a holistic refactor of consensus.

mtrippled: As duration increases, threshVote gets higher. threshVote is the criteria for whether or not to change one's close time. So if it's >60% and only 60% of peers have the other close time, then none will ever change their close time. That's the impasse which was observed in testing. The original fix led to gratuitous position changes, which were again observed in testing. This fixes both issues, and no problems are observed in testing.

thejohnfreeman:
```
        // Here's how the impasse could occur:
        // Imagine 5 validators. 3 have close time t1, and 2 have t2.
        // As consensus time increases, the threshVote threshold also increases.
        // Once threshVote exceeds 60%, no members of either set of validators
        // will change their close times.
```
If t2 > t1, then:

- the validators in t1 will see two times tied with two votes each, and switch to the greater of the two, which is t2
- the validators in t2 will see t1 with 3 votes and t2 with 1 vote, and switch to the one with more votes, which is t1

After, t2 will have 3 votes and t1 will have 2 votes, and there will be one more flip as everyone converges on the time that is both greater and has more votes, which is t2. Correct? Can that be done in one step instead of two? Are there chances for more discord with 35 validators and more than two close times? Does this feel hacky to anyone else?

thejohnfreeman: What is the logic behind the `threshVote` restriction anyway? Why is switching halted after all times fall below `threshVote`? Why not just remove that restriction? Could the fix be as simple as that?

mtrippled: > ```
>         // Here's how the impasse could occur:
>         // Imagine 5 validators. 3 have close time t1, and 2 have t2.
>         // As consensus time increases, the threshVote threshold also increases.
>         // Once threshVote exceeds 60%, no members of either set of validators
>         // will change their close times.
> ```
> 
> If t2 > t1, then:
> 
> * the validators in t1 will see two times tied with two votes each, and switch to the greater of the two, which is t2
> * the validators in t2 will see t1 with 3 votes and t2 with 1 vote, and switch to the one with more votes, which is t1
> 
> After, t2 will have 3 votes and t1 will have 2 votes, and there will be one more flip as everyone converges on the time that is both greater and has more votes, which is t2. Correct? Can that be done in one step instead of two? Are there chances for more discord with 35 validators and more than two close times? Does this feel hacky to anyone else?

No, the first criteria is having more votes. The greater time is only evaluated if there's a tie in the number of votes. That's what this comment means:



No, the first criteria is having more votes. The greater time is only evaluated if there's a tie in the number of votes. That's what this comment means:

// vote. Instead, they add their own close time vote in with
// those of their peers only after they discover the most popular,
// and the greatest of those if there is a tie. This solution avoids

mtrippled: > What is the logic behind the `threshVote` restriction anyway? Why is switching halted after all times fall below `threshVote`? Why not just remove that restriction? Could the fix be as simple as that?

I assume that threshVote is there to reduce the amount of position changes as time increases--that's the only effect that it seems to have. With too many position changes, consensus could be delayed because the ground is moving under everybody's feet. I think the assumption is that, as time passes, more peers will tend to have submitted proposals, so the likelihood is good that there is a favorite close time. But early on, few proposals have been submitted so changing one's position to match the early favorite probably will result in achieving faster agreement.

thejohnfreeman: > No, the first criteria is having more votes. The greater time is only evaluated if there's a tie in the number of votes.

Yes, but from the perspective of three validators (the ones initially proposing close time t1), when they exclude their own votes (as you are making them do), then they perceive a tie vote: two votes for t1 and two votes for t2.

thejohnfreeman: > With too many position changes, consensus could be delayed because the ground is moving under everybody's feet.

Almost no validator will be changing their position if there is an early favorite. Most validators will be voting for an early favorite (that's what makes it an early favorite), and they will have no incentive to change their vote. If there's an early favorite, we don't have to worry about "too many position changes".

If there is not an early favorite, and you want to reduce position changes while early votes arrive, then validators should freeze their position while the threshold is _low_ (when consensus is in an early phase), and then start changing position when the threshold becomes _high_ (when consensus is in a late phase). That's the opposite of the current behavior as you describe it.

The fact that there is no definitive explanation for the current behavior is a problem. The fact that the suggested explanation for the current behavior defies reason is even worse. Who understands this part of the code?

mtrippled: > > No, the first criteria is having more votes. The greater time is only evaluated if there's a tie in the number of votes.
> 
> Yes, but from the perspective of three validators (the ones initially proposing close time t1), when they exclude their own votes (as you are making them do), then they perceive a tie vote: two votes for t1 and two votes for t2.

In the case of consensus taking a long time, t1 & t2 are tied, because the validator is only counting the votes of the others. Since they are then tied, the latter time is chosen. The validator switches its vote. So do the other members of t1 (assuming proposals are getting around). They all settle on t2.

mtrippled: > > With too many position changes, consensus could be delayed because the ground is moving under everybody's feet.
> 
> Almost no validator will be changing their position if there is an early favorite. Most validators will be voting for an early favorite (that's what makes it an early favorite), and they will have no incentive to change their vote. If there's an early favorite, we don't have to worry about "too many position changes".
> 
> If there is not an early favorite, and you want to reduce position changes while early votes arrive, then validators should freeze their position while the threshold is _low_ (when consensus is in an early phase), and then start changing position when the threshold becomes _high_ (when consensus is in a late phase). That's the opposite of the current behavior as you describe it.
> 
> The fact that there is no definitive explanation for the current behavior is a problem. The fact that the suggested explanation for the current behavior defies reason is even worse. Who understands this part of the code?

By the ground moving under everybody's feet, I mean gratuitous position changes. This causes a problem because of timing. At one brief window of time, one of the validators sees positions and close times sufficient to reach consensus, and then goes into the accept phase, builds a ledger, and so on. However, the other validators didn't quite have the same view because they received proposals a few ms later. Several of them decide that their own close times are not ideal, so they decide to switch them, and end up with consensus on another proposal because of the close time change. This leaves the 1st validator described out of sync. I have observed this behavior in testing, and the bug that this PR is fixing causes gratuitous position changes that tend to cause longer consensus duration and, sometimes, desyncs. This is what I mean by "the ground moving under everybody's feet." This PR fixes that issue by doing a position change only if consensus duration is already taking a long time, and the problem described no longer appears in testing.

There may be a better way to have implemented this several years ago by the original author. But this PR is responding to a particular circumstance--rather 2 of them--found in testing recently One is the impasse as described, and the other is to avoid gratuitous position changes. I feel it is the most minimally invasive to what already exists while fixing those issues and those issues only.

src/ripple/consensus/Consensus.h
```diff
@@ -1787,47 +1802,89 @@ Consensus<Adaptor>::updateOurPositions(bool const share)
...
-        // to break the impasse. It's also OK for those agreeing
-        // with the 3 to pretend to vote for the one with 2, because
-        // that will never exceed the threshold of 75%, even with as
-        // few as 3 validators. The most it can achieve is 2/3.
+        // Choose a close time and decide whether there is consensus for it.
+        //
+        // Close time is chosen based on the threshVote threshold
+        // calculated above. If a close time has votes equal to or greater than
+        // that threshold, then that is the best close time. If two
+        // close times are equal, then choose the greatest of them. Ensure
```
HowardHinnant, line 1810: Is this a correct and improved modification of this comment?

// If <del>two</del> close times <del>are</del> <ins>have an</ins> equal <ins>number of votes</ins>, then choose the greatest of them.

src/ripple/consensus/Consensus.h
```diff
@@ -1787,47 +1802,89 @@ Consensus<Adaptor>::updateOurPositions(bool const share)
...
+        // Determining whether there is close time consensus is very simple
+        // in comparison: if votes for the best close time meet or exceed
+        // threshConsensus, then we have close time consensus. Otherwise, not.
+
+        // First, find the best close time with which to agree.
+        std::multimap<int, NetClock::time_point> ctVotesByCount;
         for (auto& [t, v] : closeTimeVotes)
+            ctVotesByCount.insert({v, t});
+        int maxVote = ctVotesByCount.rbegin()->first;
+        NetClock::time_point const& bestCtime = ctVotesByCount.rbegin()->second;
```
HowardHinnant, line 1860: I believe the following will be a faster way to compute `maxVote` and `bestCtime` because it avoids allocations:

```c++
    int maxVote = 0;
    NetClock::time_point bestCtime{};
    for (auto const& [t, v] : closeTimeVotes)
    {
        if (maxVote <= v)
        {
            maxVote = v;
            if (bestCtime < t)
                bestCtime = t;
        }
    }
```

HowardHinnant (approved):
Left a couple of optional changes.

src/ripple/consensus/Consensus.h
mtrippled: fixed

src/ripple/consensus/Consensus.h
mtrippled: Fixed. I used your implementation. Initially, I liked what I did because the new container placed the best choice at the very end for easy access. But your implementation makes explicit the criteria-- (1) that which has the max votes, and if there's a tie (2) the latest.

HowardHinnant:
LGTM.

src/ripple/consensus/Consensus.h
thejohnfreeman: > In the case of consensus taking a long time, t1 & t2 are tied, because the validator is only counting the votes of the others. Since they are then tied, the latter time is chosen. The validator switches its vote. So do the other members of t1 (assuming proposals are getting around). They all settle on t2.

They do **NOT** settle on t2 with the way you have described it. From the perspective of every validator on t2, they perceive they are in the _minority_, that t1 has the _majority_, and so _they_ switch to t1 while the validators on t1 switch to t2, and the process repeats itself indefinitely.

sophiax851: @mtrippled In reality we have a lot more validators than 5 and each will be receiving votes from each other at different time points due to network topologies and latencies, if we change close_time based on what's proposed here, how can we eliminate the "indefinite" flapping possibility?

mtrippled: > > In the case of consensus taking a long time, t1 & t2 are tied, because the validator is only counting the votes of the others. Since they are then tied, the latter time is chosen. The validator switches its vote. So do the other members of t1 (assuming proposals are getting around). They all settle on t2.
> 
> They do **NOT** settle on t2 with the way you have described it. From the perspective of every validator on t2, they perceive they are in the _minority_, that t1 has the _majority_, and so _they_ switch to t1 while the validators on t1 switch to t2, and the process repeats itself indefinitely.

@thejohnfreeman 

You are correct and have discovered another impasse. The original goal of the PR that this fixes is to avoid impasse. Impasse is still possible.

OK, t1 has 2 voters and t2 has 2. They are at an impasse as described, from about 6-12s. One voter is AWOL. 

At 12s, each member revokes its own vote: they only see the votes of others. Each of who voted for t1 now see only 1 vote for t1 and 2 for t2. They vote for t2. Each of who voted for t2 now see only 1 vote for t2 and 2 for t1. They vote for t1. They each flip.

This cycle repeats every second that Consensus runs.

Another impasse.

I think a fix for this is to bias towards t2, meaning switch votes on tie only if not the latter time.

mtrippled: > @mtrippled In reality we have a lot more validators than 5 and each will be receiving votes from each other at different time points due to network topologies and latencies, if we change close_time based on what's proposed here, how can we eliminate the "indefinite" flapping possibility?

The fix I plan to make based on @thejohnfreeman's observation should address this. Namely, coming to consensus is referred to as "avalanching" -- I think @JoelKatz came up with the term. Maybe not. Anyway, each validator takes an initial position when they close their ledger. That initial position includes the close time. This is a field in the ledger header, so it's part of consensus. It has nothing to do with transactions, but is an approximation of the current time. This fix will wait until about 12s since the last consensus round, and will then assume that there is a possible impasse in close times. To resolve the impasse, each validator needs to choose the close time with the most votes. If there is a tie, then each validator needs to choose one of the 2 close times. Either earlier or latter. It doesn't matter as long as each uses the same criteria. The problem that John found caused them to flip their votes to the opposite in this case, so that simply continued the impasse. The fix is relatively simple: in a tie, if a validator is currently voting with the latter close time, then do nothing. But if a validator is voting with the earlier close time, change to the latter. This will result in resolution of the impasse.

By waiting 12s, that should cause most validators to have their proposals distribute through the network. There is an existing threshold that tends to avoid changing positions early. However, that threshold also makes an impasse possible. So, to answer your question, once John's issue is addressed, there will be no infinite flipping.

mtrippled: @sophiax851 Another thing that keeps infinite flipping from happening is that, during the establish phase, validators only make up a close time once from scratch based on the previous close time. If they change their position, it would only be to another close time proposed by another validator. There is no possibility of validators coming up with new close times outside of the set that is already there. Of course, with network and other latency, other validators will make proposals with possibly new close times. But in this case, they will still end up "avalanching" towards the majority close time, with the tie-breaker being the latter close time.

Take 100 validators. Imagine 90 of them send proposals within 8s. 45 say t1 and 45 say t2. The other 10 validators send their proposals with entirely different close times at 11s, t3-t12, each with 1 vote. Even if they don't all see things identically, the latter 10 will each choose either t1 or t2 to vote for (unless they lose like 90% of network traffic). Their close times will disappear. Each second this determination is made again by each validator. Eventually, they will coalesce onto a single close time if the rule is implemented correctly.

sophiax851: @mtrippled Both of https://github.com/XRPLF/rippled/pull/4505 and this PR is to address the possible impasses when there are a LOT of transactions(> 40k??) made into the open ledger (when we disable transactions queue, and disable fee escalation etc.. ) and to allow each phase to be as long as it takes. Let's put aside if this scenario will ever be possible, but I feel strongly that we can not make protocol level changes, especially time-based algorithm changes without proper testing - testing which involves more validators and various network latency variables.

mtrippled: @sophiax851 This is a bug fix discovered in testing. The original impasse is possible at any transaction volume, but it was noticed under testing at a high volume. It is better to fix bugs found in testing than to wait for them to maybe appear. This PR is a bugfix for another PR trying to fix the bug. Then John discovered that this PR also may result in an impasse. So I'm fixing that.

One of the things that perf testing does is allow us to be proactive about things that may later appear in production--this has happened to us in the past where a fix was ready to go for something I saw in testing before it hit production.

sophiax851: @mtrippled How is the impasse for which https://github.com/XRPLF/rippled/pull/4505 addresses possible for any transaction volume? The protocol has been working well with the 1950ms being the min_duration for the establish phase and how the countdown starts, again, if we want to change how the close_time should be established, we need to test it under a more realistic env. and test case.  I agree totally with the proactive testing, but the testing has to be valid in terms of how it can appear in production.

mtrippled: The original impasse description is commented: https://github.com/XRPLF/rippled/pull/4760/files#diff-742221137528fb505ea44914e395d7576a77eb3d97e32ced8565b7a6f1c9ff7bR1822

mtrippled: And simply making a test environment more like production doesn't necessarily cause the bug  to occur, at least immediately. As in the description, the impasse potential is latent at any volume, having to do with decisions about when to change a close time.

src/ripple/consensus/Consensus.h
```diff
@@ -1787,57 +1799,149 @@ Consensus<Adaptor>::updateOurPositions(bool const share)
...
+        // Highest vote getter is last. Track each close time that is tied
+        // with the highest.
+        for (auto rit = votesByCloseTime.crbegin();
+            rit != votesByCloseTime.crend(); ++rit)
+        {
+            int const& voteCount = rit->first;
+            if (!maxVote.has_value())
+                maxVote = voteCount;
+            else if (voteCount < *maxVote)
+                break;
```
thejohnfreeman, line 1889:
```suggestion
                continue;
```

thejohnfreeman (changes requested):

src/ripple/consensus/Consensus.h
thejohnfreeman: The code is fine, but this comment is wrong. Clearing the impasse takes _one more round_.

```
        // Avoiding the impasse means that validators should not necessarily
        // count their own close time towards the total until they know
        // the most popular among their peers, and then change their vote
        // to that of the most popular. In this case, the validators in set
        // t2 from the example would have switched to close time t1 and ended
        // the network impasse.
```

Assume all these validators are "stuck", i.e. they all perceive a long consensus time. Our starting condition is this:

| validator | voting for |
|---|---|
| A | t1
| B | t1
| C | t1
| D | t2
| E | t2

Every validator looks at the count, but excludes their own vote. Here is how they see the votes:

| validator | votes for t1 | votes for t2
|---|---|---
| A | 2 | 2
| B | 2 | 2
| C | 2 | 2
| D | 3 | 1
| E | 3 | 1

The t1 validators A, B, C decide to switch to t2 because they perceive a tie between t1 and t2 and t2 is the later value. The t2 validators D, E decide to switch because they perceive a sole winning time of t1. All validators flip. Here is the next state:

| validator | voting for |
|---|---|
| A | t2
| B | t2
| C | t2
| D | t1
| E | t1

The impasse remains. Now they count votes again, excluding their own vote. Here is what they see:

| validator | votes for t1 | votes for t2
|---|---|---
| A | 2 | 2
| B | 2 | 2
| C | 2 | 2
| D | 1 | 3
| E | 1 | 3

The validators A, B, C, which are now voting for t2, _still_ perceive a tie between t1 and t2, but _they do not switch their vote_, because t2 is the later value. The validators D, E, which are now voting for t1, _do_ switch their vote because they perceive a sole winning time of t2. _Now_ the impasse is cleared. Here is the next state:

| validator | voting for |
|---|---|
| A | t2
| B | t2
| C | t2
| D | t2
| E | t2

src/ripple/consensus/Consensus.h
```diff
@@ -1787,57 +1799,149 @@ Consensus<Adaptor>::updateOurPositions(bool const share)
...
+        // time this current consensus round has taken. This is reflected
+        // in the "stuck" boolean.
+        //
+        // The best close time is the close time with the highest number of
+        // votes. If multiple close times are tied, choose the latest one.
+        // Whether earlier or later does not matter. What matters is
+        // coming to agreement with the same criteria.
+        //
+        // For validators that are stuck and do not have the best close time,
+        // they change their close time position add their vote to the tally
```
thejohnfreeman, line 1850:
```suggestion
        // they change their close time position and add their vote to the tally
```

src/ripple/consensus/Consensus.h
```diff
@@ -1787,57 +1799,149 @@ Consensus<Adaptor>::updateOurPositions(bool const share)
...
-        // with the 3 to pretend to vote for the one with 2, because
-        // that will never exceed the threshold of 75%, even with as
-        // few as 3 validators. The most it can achieve is 2/3.
-        for (auto& [t, v] : closeTimeVotes)
+        // Choose a close time and decide whether there is consensus for it.
+        //
+        // Close time is chosen based on the threshVote threshold
+        // calculated above. If a close time has votes equal to or greater than
+        // that threshold, then that is the best close time. If two
+        // close times have an equal number of votes, then choose the greatest
```
thejohnfreeman, line 1807:
```suggestion
        // close times have an equal number of votes, then choose the greater
```

thejohnfreeman (changes requested):
I'll leave it up to Mark to decide how to fix the portion of the comment that talks through the example.

src/ripple/consensus/Consensus.h
mtrippled: fixed

src/ripple/consensus/Consensus.h
mtrippled: fixed

thejohnfreeman (approved):

src/ripple/consensus/Consensus.h
mtrippled: I modified the comment to reflect the current behavior.

Regarding the table you made--the new behavior is different than described based on the problem you pointed out. No longer do validators revoke their vote in the first step after being "stuck". Instead, they all vote as they always have and tally for each. Now, however, they determine which is the best close time based on the criteria: most votes, then if tie, latest close time. In this case, t1 will have 3 and t2 start with 2 votes. Only the t2 set will change their position and vote to t1. The t1 set know they are the best, and will not change their votes. Impasse resolved next iteration.

In the case of the problem with t1 having 2 votes and t2 having 2 votes--that was possible with before my previous commit. What happens now is that each validator first counts their own close time towards the tallies of each possible close time. They they see that there is a tie between t1 and t2. However, t2 is greater. Therefore, the set in t1 will switch their close time to t2, and t2 will keep the same. Resolved in the next iteration.

scottschurr (approved):
This all looks good as far as I can see.  I left a few comments about things that might be a little more efficient.  But nothing that I think should stop this pull request.  It's also possible I misunderstand something, so be cautious.  Don't let my "efficiencies" introduce bugs.

I do have one concern, however.  There are a few places that simply assume that STL containers have at least one entry.  I'm _guessing_ that's a legit assumption.  If it's not, then this code could lead to occasional crashes when some of these containers are empty.

Personally, I wish the code were handling the empty container cases a bit more defensively.  But it may not be necessary.

mtrippled:
> This all looks good as far as I can see. I left a few comments about things that might be a little more efficient. But nothing that I think should stop this pull request. It's also possible I misunderstand something, so be cautious. Don't let my "efficiencies" introduce bugs.
> 
> I do have one concern, however. There are a few places that simply assume that STL containers have at least one entry. I'm _guessing_ that's a legit assumption. If it's not, then this code could lead to occasional crashes when some of these containers are empty.
> 
> Personally, I wish the code were handling the empty container cases a bit more defensively. But it may not be necessary.

I commented above those optionals, and also put an assert() there. It's safe

src/ripple/consensus/Consensus.h
```diff
@@ -1759,15 +1759,27 @@ Consensus<Adaptor>::updateOurPositions(bool const share)
...
+        }
         else if (convergePercent_ < parms.avSTUCK_CONSENSUS_TIME)
+        {
             neededWeight = parms.avLATE_CONSENSUS_PCT;
+        }
         else
+        {
             neededWeight = parms.avSTUCK_CONSENSUS_PCT;
+            stuck = true;
+        }
```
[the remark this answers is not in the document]
mtrippled: I'm keeping it as is.

src/ripple/consensus/Consensus.h
```diff
@@ -1787,57 +1799,144 @@ Consensus<Adaptor>::updateOurPositions(bool const share)
...
+        // with the highest.
+        for (auto rit = votesByCloseTime.crbegin();
+            rit != votesByCloseTime.crend(); ++rit)
+        {
+            int const& voteCount = rit->first;
+            if (!maxVote.has_value())
+                maxVote = voteCount;
+            else if (voteCount < *maxVote)
+                break;
+            maxCloseTimes.insert(rit->second);
```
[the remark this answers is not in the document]
mtrippled: I'm keeping the sledgehammer for the same reason.

src/ripple/consensus/Consensus.h
```diff
@@ -1787,57 +1799,144 @@ Consensus<Adaptor>::updateOurPositions(bool const share)
...
+        // The 2nd check has been nested within the first historically.
+        // It's possible that this can be optimized by doing the
+        // 2nd check independently--this may make consensus happen faster in
+        // some cases. Then again, the trade-offs have not been modelled.
+        if (*maxVote >= threshVote)
+        {
+            consensusCloseTime = bestCloseTime;
+            ss << "Close time " << bestCloseTime.time_since_epoch().count() << " has " <<
+                *maxVote << " votes, which is >= the threshold (" << threshVote
+                << " to make that our positions if it isn't already.";
```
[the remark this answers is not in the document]
mtrippled: fixed

src/ripple/consensus/Consensus.h
```diff
@@ -1787,57 +1799,144 @@ Consensus<Adaptor>::updateOurPositions(bool const share)
...
+            ss << closeTime.time_since_epoch().count() << ':' << voteCount;
+        }
+        std::optional<int> maxVote;
+        std::set<NetClock::time_point> maxCloseTimes;
+        // Highest vote getter is last. Track each close time that is tied
+        // with the highest.
+        for (auto rit = votesByCloseTime.crbegin();
+            rit != votesByCloseTime.crend(); ++rit)
+        {
+            int const& voteCount = rit->first;
```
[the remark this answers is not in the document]
mtrippled: fixed
I had no idea about const& being less efficient than int copy.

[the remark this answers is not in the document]
scottschurr: FWIW, the reason that it's less efficient is that an `&` is really just a pointer in disguise.  So any time there's an `&` there's the likelihood of an indirection through a pointer.  If you make a local copy of an `int`, that can just stay in a register, which is much faster than accessing memory through pointer indirection.

If the thing you're accessing is too big to fit in a register then `const&` is a good choice.  If the thing is small enough to fit in a register then it often makes sense to make a local copy.

PR #4762, opened 2023-10-12
Don't suspend writes based on batch size. Instead, log if it seems to be too big

mtrippled:
Removes a limit that blocks nodestore writes. This delays consensus because each new ledger is persisted during consensus processing.

### Context of Change
This removes the check that limits write batch size and blocks when the limit is reached. It also adds logging to indicate when the limit is exceeded. The trade-off is that the pending writes consume memory until they are persisted, so memory can increase without the limit. But since blocking writes result in consensus instability, the existing behavior is a proactive pessimization. Ultimately, pending writes either increase due to increased input (transaction volume), or because of I/O slowdown. If the IO and memory subsystem can't handle the throughput then the process will be unstable regardless.

### Type of Change

- [ ] Documentation Updates
- [ ] Performance
- [ ] Release

### DIFF
--- a/src/ripple/nodestore/impl/BatchWriter.cpp
+++ b/src/ripple/nodestore/impl/BatchWriter.cpp
@@ -37,14 +37,11 @@ BatchWriter::~BatchWriter()
 }
 
 void
-BatchWriter::store(std::shared_ptr<NodeObject> const& object)
+BatchWriter::store(
+    std::shared_ptr<NodeObject> const& object,
+    beast::Journal const& j)
 {
     std::unique_lock<decltype(mWriteMutex)> sl(mWriteMutex);
-
-    // If the batch has reached its limit, we wait
-    // until the batch writer is finished
-    while (mWriteSet.size() >= batchWriteLimitSize)
-        mWriteCondition.wait(sl);
 
     mWriteSet.push_back(object);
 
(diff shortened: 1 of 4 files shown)


HowardHinnant:
This looks fine to me, but I have a question:  It looks like the code just adds logging.  I don't see code that does the other part of the title: "Don't suspend writes based on batch size."  What am I missing?

src/ripple/nodestore/impl/BatchWriter.cpp
```diff
@@ -37,21 +37,29 @@ BatchWriter::~BatchWriter()
...
 void
-BatchWriter::store(std::shared_ptr<NodeObject> const& object)
+BatchWriter::store(std::shared_ptr<NodeObject> const& object,
+    beast::Journal const& j)
 {
     std::unique_lock<decltype(mWriteMutex)> sl(mWriteMutex);
 
-    // If the batch has reached its limit, we wait
-    // until the batch writer is finished
-    while (mWriteSet.size() >= batchWriteLimitSize)
```
mtrippled, line 46: @HowardHinnant Here's the logic change.

mtrippled:
> This looks fine to me, but I have a question: It looks like the code just adds logging. I don't see code that does the other part of the title: "Don't suspend writes based on batch size." What am I missing?

fixed

HowardHinnant (approved):

seelabs (approved):
:+1: This patch fixes a potential deadlock as well. Needs to be reformatted before merging though. Nice job Mark!

ximinez (approved):

intelliot:
@sophiax851 this may require perf signoff

intelliot:
Internal tracker: RPFC-80

intelliot:
May be replaced by #4882

intelliot:
- Could we just slow down the writes since there is no rush?

intelliot:
https://github.com/XRPLF/rippled/pull/4503 writes asynchronously to nudb with a background thread, but it has an upper limit of records that can be in the queue to be written. When reached, writes are blocked. Online deletion copies a ledger and ensures that it writes the entire ledger to nudb. This amount of data being written likely causes further writes to be blocked until the data is written. This includes persisting new records for each new ledger produced, which is synchronous for consensus. This causes desyncs. This PR (#4762) is expected to fix this.

If #4882 (which just reverts #4503) is merged, then #4503+#4762 (combined) can still be considered for testing+merging in the future.

intelliot:
note via @mtrippled : there is no reason to merge this PR. We can consider https://github.com/XRPLF/rippled/pull/4503#issuecomment-1902323779 - tweaking NuDB's write buffer is probably better design. We can design micro-benchmarks to show how it affects NuDB itself. Write prioritization is too much complexity, but if NuDB could write faster, then that was the whole point of #4503.

PR #4763, opened 2023-10-12
Don't reach consensus as fast if no other proposals seen.

mtrippled:
This fixes a case where a peer can desync under a certain timing circumstance--if it reaches a certain point in consensus before it receives proposals.

### Context of Change
This was noticed under high transaction volumes. Namely, when we arrive at the point of deciding whether consensus is reached after minimum establish phase duration but before having received any proposals. This could be caused by finishing the previous round slightly faster and/or having some delay in receiving proposals. Existing behavior arrives at consensus immediately after the minimum establish duration with no proposals. This causes us to desync because we then close a non-validated ledger. The change in this PR causes us to wait for a configured threshold before making the decision to arrive at consensus with no proposals. This allows validators to catch up and for brief delays in receiving proposals to be absorbed. There should be no drawback since, with no proposals coming in, we needn't be in a huge rush to jump ahead.

### Type of Change

- [ ] Documentation Updates
- [ ] Performance
- [ ] Release


src/ripple/consensus/Consensus.cpp
```diff
@@ -127,7 +140,12 @@ checkConsensus(
                     << prevProposers << " agree=" << currentAgree
                     << " validated=" << currentFinished
                     << " time=" << currentAgreeTime.count() << "/"
-                    << previousAgreeTime.count();
+                    << previousAgreeTime.count()
+                    << " proposing? " << proposing
+                    << " minimum duration to reach consensus: "
+                    << parms.ledgerMIN_CONSENSUS.count() << "ms"
+                    << " max consensus time " << parms.ledgerMAX_CONSENSUS.count() << 's'
```
Bronek, line 147: `ledgerMAX_CONSENSUS` type is milliseconds, not seconds.

src/ripple/consensus/Consensus.cpp
```diff
@@ -94,11 +94,24 @@ checkConsensusReached(
...
-    std::size_t minConsensusPct)
+    std::size_t minConsensusPct,
+    bool reachedMax)
 {
-    // If we are alone, we have a consensus
+    // If we are alone for too long, we have a consensus
+    // Delaying consensus like this avoids a circumstance where a peer
+    // gets ahead of proposers insofar as it has not received any proposals.
+    // This could happen if there's a slowdown in receiving proposals. Reaching
+    // consensus prematurely in this way means that the peer with desync.
```
Bronek, line 104: "that the peer with desync." - is this an unfinished sentence ?

src/ripple/consensus/Consensus.cpp
mtrippled: fixed

src/ripple/consensus/Consensus.cpp
mtrippled: fixed

Bronek (approved):

seelabs (commented):
:+1: Changes to consensus always make me nervous, but this is a straightforward change to a very specific corner case. LGTM.

seelabs:
@mtrippled We need to fix the formatting before we merge this.

seelabs (approved):
:+1: Although this needs to be reformatted before it is merged.

intelliot:
@mtrippled : does this depend on other PRs, or could it be tested on its own?

mtrippled:
@intelliot this could be tested on its own I think.

intelliot:
(Ignore removing the label, that was a misclick.)

Internal tracker: RPFC-104

intelliot:
@seelabs could we get a quick re-review now that this has been reformatted? 

Also, someone please confirm whether the "1 lines in your changes are missing coverage" is worth addressing prior to merging this.

seelabs (approved):

PR #4764, opened 2023-10-12
Optimize when to acquire ledgers from the network.

mtrippled:
Particulary avoid acquiring ledgers likely to be
produced locally very soon.

Peer ledger acquisition is expensive. This reduces peer ledger aquisition.

### Context of Change
Existing code attempts to acquire ledgers if a validation for a new ledger is received, even if it comes moments before we have created the same ledger through normal consensus processing. A peer only needs to be slightly behind the first validator for this to be triggered. This essentially causes all nodes to have expensive background ledgerData acquisition jobs to be running. The fix is to not try to acquire ledgers that are very likely to be created locally through normal processing.

### Type of Change

- [ ] Documentation Updates
- [X] Performance
- [ ] Release


HowardHinnant (approved):

ximinez (approved):

intelliot:
@mtrippled to confirm that this should be put on hold for now

PR #4775, opened 2023-10-18
Unify JSON serialization format of transactions

Bronek:

Unify JSON format of transaction as per https://github.com/XRPLF/rippled/issues/4727 , intended for API version 2 release

### Type of Change

- [ ] Documentation update
- [ ] Release

### API Impact


intelliot:
@ximinez @ckeshava this is ready for review

Bronek:
> @ximinez @ckeshava this is ready for review

Please note I will appreciate if you verify that I have not missed (or indeed, missed) anything in https://github.com/XRPLF/rippled/issues/4727

src/ripple/app/ledger/impl/LedgerToJson.cpp
```diff
@@ -118,24 +122,69 @@ fillJsonTx(
...
+                fill.context->ledgerMaster, fill.ledger, fill.context->app);
+            txJson[jss::validated] = validated;
+            if (validated)
+            {
+                txJson[jss::ledger_index] = to_string(fill.ledger.seq());
+                txJson[jss::ledger_hash] = to_string(fill.ledger.info().hash);
+                if (auto close_time =
+                        fill.context->ledgerMaster.getCloseTimeBySeq(
+                            fill.ledger.seq()))
+                    txJson[jss::close_time_iso] = to_string_iso(*close_time);
```
ximinez, line 169: I'm not sure how expensive it is to compute these fields, especially `close_time`, but would it make sense to initialize them once in the ctor for `LedgerFill` and reuse them on every iteration?

ximinez (changes requested):
Partial review. I have really barely started, so I don't have the whole picture yet. If any of these suggestions don't make sense, just say so. 

I'm putting some of the ideas into a branch while I go along just so I can see if they make sense. https://github.com/Bronek/rippled/compare/feature/unify_transaction_json...ximinez:rippled:bk-4775-unify-tx-json

src/ripple/app/ledger/impl/LedgerToJson.cpp
```diff
@@ -17,14 +17,17 @@
...
 #include <ripple/app/ledger/LedgerToJson.h>
 #include <ripple/app/main/Application.h>
 #include <ripple/app/misc/DeliverMax.h>
 #include <ripple/app/misc/TxQ.h>
 #include <ripple/basics/base_uint.h>
 #include <ripple/core/Pg.h>
+#include <ripple/protocol/jss.h>
 #include <ripple/rpc/Context.h>
 #include <ripple/rpc/DeliveredAmount.h>
+#include <ripple/rpc/impl/RPCHelpers.h>
```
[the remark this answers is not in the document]
Bronek: Thanks, I will try to fix it as you suggest.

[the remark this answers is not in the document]
Bronek: Done

src/ripple/app/ledger/impl/LedgerMaster.cpp
```diff
@@ -603,6 +603,55 @@ LedgerMaster::clearLedger(std::uint32_t seq)
     mCompleteLedgers.erase(seq);
 }
 
+bool
+LedgerMaster::isValidated(ReadView const& ledger)
+{
+    if (app_.config().reporting())
+        return haveLedger(ledger.seq());  // TODO Is this correct ?
```
Bronek, line 610: @ximinez @seelabs I am not sure if this is correct, will appreciate feedback. I got the idea from `getValidatedRange` but perhaps a better check would be `return ledger.seq() <= getValidLedgerIndex()`.

FWIW The old `isValidated` in `RPCHelper` would just `return true;` but that did not seem right to me.

src/ripple/app/ledger/impl/LedgerToJson.cpp
Bronek: It definitely does - done, thanks !

src/ripple/app/misc/impl/Transaction.cpp
```diff
@@ -165,13 +165,19 @@ Transaction::load(
...
+Transaction::getJson(
+    JsonOptions options,
+    bool binary,
+    bool showInLedger,
+    std::optional<std::reference_wrapper<std::string>> hash) const
 {
-    Json::Value ret(mTransaction->getJson(JsonOptions::none, binary));
+    Json::Value ret(mTransaction->getJson(JsonOptions::none, binary, hash));
 
     if (mInLedger)
```
ckeshava, line 176: Since `inLedger` field is being deprecated, it would be nice to rename the `mInLedger` data member into `mLedgerIndex` or something like that. It would help with the readability of the code.

src/ripple/protocol/impl/STTx.cpp
```diff
@@ -234,17 +234,33 @@ Json::Value STTx::getJson(JsonOptions) const
...
     {
-        Json::Value ret;
         Serializer s = STObject::getSerializer();
-        ret[jss::tx] = strHex(s.peekData());
-        ret[jss::hash] = to_string(getTransactionID());
+        Json::Value ret = strHex(s.peekData());
         return ret;
     }
-    return getJson(options);
+    return STObject::getJson(JsonOptions::none);  // Yes, want `none`
```
ckeshava, line 263: Can we you please explain why we are disregarding the input parameter `options` here? (instead setting it to `JsonOptions::none`) ?

src/ripple/protocol/STTx.h
```diff
@@ -108,8 +108,16 @@ class STTx final : public STObject, public CountedObject<STTx>
 
     Json::Value
     getJson(JsonOptions options) const override;
+
+    /// If `hash` is set, will store hash inside the provided string. Otherwise
```
ckeshava, line 112: Is it more accurate to say -- "If `hash` is set, will store transaction ID inside the provided string" ?

src/ripple/app/misc/NetworkOPs.cpp
```diff
@@ -3101,7 +3103,12 @@ NetworkOPsImp::transJson(
     transResultInfo(result, sToken, sHuman);
 
     jvObj[jss::type] = "transaction";
-    jvObj[jss::transaction] = transaction->getJson(JsonOptions::none);
+    // NOTE jvObj which is not a finished object for either API version. After
+    // it's populated, we need to finish it for a specific API version. This is
+    // done in a loop, near the end of this function.
+    std::string hash = {};
```
ckeshava, line 3109: I feel it would be more clear if we separate the serialization logic and data initialization in the `getJson()` function. `hash` can be directly initialized here as `hash = transaction->getTransactionID();` (instead of passing an optional ref-wrapper)

This is not a strong opinion, an alternative design suggestion, that's all

src/ripple/app/ledger/impl/LedgerToJson.cpp
```diff
@@ -118,24 +121,68 @@ fillJsonTx(
...
+                        {txn->getTransactionID(), fill.ledger.seq(), *stMeta});
+            }
+
+            const bool validated =
+                fill.context->ledgerMaster.isValidated(fill.ledger);
+            txJson[jss::validated] = validated;
+            if (validated)
+            {
+                txJson[jss::ledger_index] = to_string(fill.ledger.seq());
+                txJson[jss::ledger_hash] = to_string(fill.ledger.info().hash);
```
ckeshava, line 164: As per [Recommendations](https://github.com/XRPLF/rippled/issues/4727),  `ledger_hash` is expected to be included in closed ledgers. The code includes this field in validated ledgers, not [closed ledgers](https://xrpl.org/open-closed-validated-ledgers.html).

Should we update the recommendations to reflect the code?
image

src/ripple/app/ledger/impl/LedgerMaster.cpp
```diff
@@ -603,6 +603,55 @@ LedgerMaster::clearLedger(std::uint32_t seq)
...
+                    // SQL database doesn't match ledger chain
+                    clearLedger(seq);
+                }
+            }
+            return false;
+        }
+    }
+    catch (SHAMapMissingNode const& mn)
+    {
+        auto stream = app_.journal("IsValidated").warn();  // TODO Better name ?
```
ckeshava, line 645: I feel `app_.journal("LedgerMaster::IsValidated")` is a descriptive name

src/ripple/protocol/impl/STTx.cpp
```diff
@@ -234,17 +234,33 @@ Json::Value STTx::getJson(JsonOptions) const
...
+    bool binary,
+    std::optional<std::reference_wrapper<std::string>> hash) const
 {
+    if (!hash)  // Old behaviour - default because `hash = {}` in declaration
+    {
+        if (binary)
+        {
+            Json::Value ret;
+            Serializer s = STObject::getSerializer();
+            ret[jss::tx] = strHex(s.peekData());
```
ckeshava, line 248: what is the logic for inclusion/exclusion of the `jss::tx` field?

src/test/jtx/TestHelpers.h
```diff
@@ -27,17 +27,25 @@
...
+concept iterable = requires(T& v)
+{
+    std::begin(v);
+    std::end(v);
+};
+
+template <typename Input>
 auto
-make_vector(auto const& input) requires std::ranges::range<decltype(input)>
+make_vector(Input const& input) requires iterable<Input>
```
ckeshava, line 46: I'm not able to find usages of this helper function in this PR, have you used it in another PR?

src/test/rpc/JSONRPC_test.cpp
```diff
@@ -2586,6 +2588,7 @@ class JSONRPC_test : public beast::unit_test::suite
                         assert(get<1>(testFunc) == nullptr);
                         result = signFn(
                             req,
+                            1,
```
ckeshava, line 2591: I want to understand why we're not looping through all the API versions here. Is it because this PR only modified the response formats? 
I don't think the request format has been modified by this PR.

src/ripple/rpc/handlers/Tx.cpp
```diff
@@ -311,17 +325,50 @@ populateJsonResponse(
...
+                    context.apiVersion);
+            }
+            response[jss::hash] = hash;
+            if (result.validated)
+            {
+                response[jss::ledger_index] = result.txn->getLedger();
+                if (result.ledgerHash)
+                    response[jss::ledger_hash] = to_string(*result.ledgerHash);
+
+                if (result.closeTime)
```
ckeshava, line 351: I think `response[jss::validated] = true;` can be added here. This is with reference to [Recommendations](https://github.com/XRPLF/rippled/issues/4727) point-7 

I notice that you are adding this field in `TransactionEntry.cpp` and `fillJsonTx`,  but I'm not sure if that's covering all possibilities.

src/test/rpc/Subscribe_test.cpp
```diff
@@ -307,6 +308,80 @@ class Subscribe_test : public beast::unit_test::suite
...
+            env.close();
+
+            // Check stream update for payment transaction
+            BEAST_EXPECT(wsc->findMsg(5s, [&](auto const& jv) {
+                return jv[jss::meta]["AffectedNodes"][1u]["CreatedNode"]
+                         ["NewFields"][jss::Account]  //
+                    == Account("alice").human() &&
+                    jv[jss::close_time_iso]  //
+                    == "2000-01-01T00:00:10Z" &&
+                    jv[jss::validated] == true &&           //
```
ckeshava, line 350: We can test more fields at this juncture -- like `ledger_hash`, `ledger_index` and non-existence of `inLedger`. 
https://github.com/ckeshava/rippled/tree/moreTests

src/ripple/rpc/handlers/AMMInfo.cpp
```diff
@@ -66,7 +66,7 @@ to_iso8601(NetClock::time_point tp)
     return date::format(
         "%Y-%Om-%dT%H:%M:%OS%z",
```
HowardHinnant, line 68: The `%Om` should just be `%m`.  Adding the `O` gives the locale's alternative representation.  But this is just the numbers 1-12, and I'd be surprised if we really intend on alternative representation of decimal digits.  Ditto for `%OS`.  It should just be `%S`.

Bronek: I replaced this with `%FT%TZ` and changed type conversion instead, to preserve full-second period of `NetClock`. See last commit  https://github.com/XRPLF/rippled/pull/4775/commits/9f1d544dcbae542d9d1d11ac43d36c5e168eba9a

HowardHinnant: That looks perfect! :-)

src/ripple/rpc/handlers/Tx.cpp
Bronek: No, we set this field some 40 lines below
```
response[jss::validated] = result.validated;
```

src/test/jtx/TestHelpers.h
Bronek: it is used in `PerfLog_test.cpp` and `Handler_test.cpp`

src/ripple/app/misc/NetworkOPs.cpp
Bronek: Maybe. Trouble is that we do not have flexible design choice here, since `Transaction::getJson` is used in quite a few locations.

src/ripple/protocol/STTx.h
Bronek: thanks , will do that !

src/ripple/protocol/impl/STTx.cpp
Bronek: Thanks, added comment to explain

src/ripple/app/misc/impl/Transaction.cpp
Bronek: Thanks, will do that. In general I do not like mixing refactoring with functional changes, but here it is a clear readability improvement with very small added churn.

src/ripple/app/ledger/impl/LedgerMaster.cpp
Bronek: I do not know how these names are rendered, but it appears we almost never use them with `::` in the middle ...

ckeshava: hmm, i see

Bronek:
I made changes from review discussion. Can I have more review please ?

src/test/jtx/TestHelpers.h
ckeshava: okay, I didn't find these results a few days ago, thanks 👍

src/ripple/app/ledger/impl/LedgerMaster.cpp
Bronek: I switched to `return true;` with a comment explainig the reason.

API-CHANGELOG.md
```diff
@@ -141,6 +141,31 @@ In API version 2, the following methods are no longer available:
...
+- `validated` - Bool element set to `true` if the transaction is in a validated ledger, otherwise `false`
+
+This change affects the following methods:
+
+- `tx` - Transaction data moved into element `tx_json` (was inline inside `result`) or, if binary output was requested, moved from `tx` to `tx_blob`. Renamed binary transaction metadata element (if it was requsted) from `meta` to `meta_blob`. Changed location of `hash` and added new elements
+- `account_tx` - Renamed transaction element from `tx` to `tx_json`. Renamed binary transaction metadata element (if it was requsted) from `meta` to `meta_blob`. Changed location of `hash` and added new elements
+- `transaction_entry` - Renamed transaction metadata element from `metadata` to `meta`. Changed location of `hash` and added new elements
+- `subscribe` - Renamed transaction element from `transaction` to `tx_json`. Changed location of `hash` and added new elements
+- `sign`, `sign_for`, `submit` and `submit_multisigned` - Changed location of `hash` element.
+
```
Bronek, line 168: @mDuo13 do you mind checking this file ?

src/ripple/app/ledger/impl/LedgerMaster.cpp
ckeshava: are you still referring to line 610? github shows me the old code, perhaps your new code has not been pushed into github

API-CHANGELOG.md
```diff
@@ -141,6 +141,31 @@ In API version 2, the following methods are no longer available:
...
+- `hash` - Transaction ID. This data was stored inside transaction output in API version 1, but in API version 2 is a sibling element.
+- `ledger_index` - Ledger index (only set on validated ledgers)
+- `ledger_hash` - Ledger hash (only set on closed or validated ledgers)
+- `close_time_iso` - Ledger close time expressed in ISO 8601 time format (only set on validated ledgers)
+- `validated` - Bool element set to `true` if the transaction is in a validated ledger, otherwise `false`
+
+This change affects the following methods:
+
+- `tx` - Transaction data moved into element `tx_json` (was inline inside `result`) or, if binary output was requested, moved from `tx` to `tx_blob`. Renamed binary transaction metadata element (if it was requsted) from `meta` to `meta_blob`. Changed location of `hash` and added new elements
```
ckeshava, line 163: typo: "requested"

API-CHANGELOG.md
```diff
@@ -141,6 +141,31 @@ In API version 2, the following methods are no longer available:
...
+- `account_tx` - Renamed transaction element from `tx` to `tx_json`. Renamed binary transaction metadata element (if it was requsted) from `meta` to `meta_blob`. Changed location of `hash` and added new elements
```
ckeshava, line 164: typo: "requested"

src/test/rpc/AccountTx_test.cpp
```diff
@@ -120,22 +120,37 @@ class AccountTx_test : public beast::unit_test::suite
...
+                        (j[jss::result][jss::transactions].size() == 2) &&
+                        (j[jss::result][jss::transactions][0u][jss::tx_json]
+                          [jss::TransactionType] == jss::AccountSet) &&
+                        (j[jss::result][jss::transactions][1u][jss::tx_json]
+                          [jss::TransactionType] == jss::Payment) &&
+                        (j[jss::result][jss::transactions][1u][jss::tx_json]
+                          [jss::DeliverMax] == "10000000010") &&
+                        (!j[jss::result][jss::transactions][1u][jss::tx_json]
+                              .isMember(jss::Amount));
+                default:
```
arihantkothari, line 151: I may be wrong here, but what happens if v3 comes in and we don't do any specific change to this command or test. Should it not still follow the logic of case 2 ?

src/ripple/rpc/handlers/AccountTx.cpp
```diff
@@ -338,6 +356,11 @@ populateJsonResponse(
                         insertDeliveredAmount(
                             jvObj[jss::meta], context, txn, *txnMeta);
                         insertNFTSyntheticInJson(jvObj, sttx, *txnMeta);
+                        if (auto closeTime =
```
ckeshava, line 359: Why is the presence of `closeTime` predicated on the existence of `txnMeta` ? Can't we have this block of code at line 341?
```
if (auto closeTime = context.ledgerMaster.getCloseTimeBySeq(txn->getLedger()))
     jvObj[jss::close_time_iso] =
     to_string_iso(*closeTime);
```

src/test/rpc/AccountTx_test.cpp
ckeshava: I think we can coalesce multiple cases:
```
case 2:
case 3:
// existing tests
```

src/test/rpc/AccountTx_test.cpp
```diff
@@ -120,22 +120,37 @@ class AccountTx_test : public beast::unit_test::suite
...
+                        (j[jss::result][jss::status] == "success") &&
+                        (j[jss::result][jss::transactions].size() == 2) &&
+                        (j[jss::result][jss::transactions][0u][jss::tx_json]
+                          [jss::TransactionType] == jss::AccountSet) &&
+                        (j[jss::result][jss::transactions][1u][jss::tx_json]
+                          [jss::TransactionType] == jss::Payment) &&
+                        (j[jss::result][jss::transactions][1u][jss::tx_json]
+                          [jss::DeliverMax] == "10000000010") &&
+                        (!j[jss::result][jss::transactions][1u][jss::tx_json]
+                              .isMember(jss::Amount));
```
ckeshava, line 150: To be thorough, we can add the checks for other fields like `jss::hash`, `jss::close_time_iso` etc

ckeshava (changes requested):
I have left a few comments, some of them are non-binding suggestions. feel free to pick/choose from the tip of this branch: https://github.com/ckeshava/rippled/tree/jsonChanges

src/test/rpc/AccountTx_test.cpp
arihantkothari: Would it not be better to follow this pattern ? (again, i'm not sure if switch case statements makes more sense...)
```
if (version < FEATURE_X_VERSION) {
  // old behavior
} else {
  // new behavior
}

if (version >= FEATURE_X_VERSION) {
  // new behavior where old behavior was just "do nothing"
}

if (version < FEATURE_X_VERSION) {
  // old behavior where new behavior is "stop doing that"
}
```

In this particular case just:

```
if (version < 2u) {
  // old behavior
} else {
  // new behavior
}
```

arihantkothari: not the biggest deal in any case 😆

Bronek: compilers rather good at catching bugs which arise from missing a `case` in a `switch` statement, and very bad at catching bugs from bad logic in `if ... else if .. else` series of statements. This is why I prefer using `switch`. As soon as we add version 3 and start testing it, this will fail (due to `default: return false;`) which is exactly as we want it, so we know to make adjustment. If these adjustments become a problem, we can change style, but I hope by then we will have a better way of testing different API versions (I aim to work on it, and `forAllApiVersions` is a good first effort).

src/ripple/app/ledger/impl/LedgerToJson.cpp
Bronek: Thanks, I implemented your suggestion.

src/test/rpc/AccountTx_test.cpp
arihantkothari: Thanks for the reply, that is good to know.

src/ripple/rpc/handlers/AccountTx.cpp
Bronek: that's because we only have meta in a validated ledger, and we only show close time in a validated ledger as well

src/test/rpc/AccountTx_test.cpp
Bronek: Great idea, and actually helped me uncover a bug (see last commit)

src/ripple/rpc/handlers/AccountTx.cpp
ckeshava: I see. It would be more clear if all the fields `jss::ledger_index`, `jss::ledger_hash` and `jss::close_time_iso` are grouped under one conditional check `if (ledger is validated) then {<fill these fields>}`

src/ripple/rpc/handlers/AccountTx.cpp
```diff
@@ -326,10 +325,29 @@ populateJsonResponse(
...
+                    {
+                        std::string hash;
+                        jvObj[json_tx] = txn->getJson(
+                            JsonOptions::include_date,
+                            false,
+                            false,
+                            {std::ref(hash)});
+                        jvObj[jss::hash] = hash;
+                        jvObj[jss::ledger_index] = txn->getLedger();
+                        jvObj[jss::ledger_hash] =
```
ckeshava, line 340: @Bronek I felt this would be a good place to add the `jss::close_time_iso` field because other fields are also populated here.

But I get the rationale about fencing it behind `txMeta` 👍

src/test/rpc/AccountTx_test.cpp
ckeshava: hmm, I tested it myself but I didn't run into any bugs. anyway, I'm glad you found it 👍

ckeshava: oh, you had called `txnMeta->getIndex` instead of `txn->getLedger` :)
Honestly, I find it hard to tell the differences between `TxMeta::mIndex`, `TxMeta::mTransactionID`, `TxMeta::mLedger` --> they have similar types and appear to fulfill similar functionality.

src/ripple/app/ledger/impl/LedgerToJson.cpp
ckeshava: thank you for looking into it 👍

ckeshava (changes requested):
Hello @Bronek 
I have added a few more unit tests at the tip of this [branch](https://github.com/ckeshava/rippled/tree/lgrIdx), if you'd like feel free to use it.

I'd like to draw your attention to the `jss::tx_json` field. In `apiVersion > 1`, is it supposed to contain the `jss::ledger_index` and `jss::date` fields? The [Recommendations](https://github.com/XRPLF/rippled/issues/4727) suggest that these fields need to be removed.

I have fixed that and added unit tests to that effect in the above branch

src/test/rpc/TransactionEntry_test.cpp
```diff
@@ -143,29 +144,43 @@ class TransactionEntry_test : public beast::unit_test::suite
...
                 return;
 
-            BEAST_EXPECT(resIndex[jss::tx_json][jss::hash] == txhash);
+            if (apiVersion > 1)
+            {
+                BEAST_EXPECT(resIndex[jss::hash] == txhash);
+                BEAST_EXPECT(resIndex[jss::validated] == true);
+                BEAST_EXPECT(!resIndex[jss::tx_json].isMember(jss::Amount));
+
+                if (!close_time_iso.empty())
```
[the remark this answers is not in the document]
ckeshava: The fix that I proposed causes failures in `AccountTxPaging_test`, so I'm not certain if that's the right way to go about it. LMK if you have other ideas.

API-CHANGELOG.md
Bronek: Thanks, fixed

API-CHANGELOG.md
Bronek: thanks, fixed

src/ripple/rpc/handlers/AccountTx.cpp
Bronek: thanks for a very good comment ! So I've spent some more time reading `AccountTx.cpp` and it turns out, this function only ever returns validated ledger. Which is actually rather visible near the top of `populateJsonResponse` function, where we do `response[jss::validated] = true;` . This also means that the line I added `response[jss::validated] = false;` would be a regression. I have removed it now, and will move other lines accordingly.

src/test/rpc/AccountTx_test.cpp
Bronek: Yeah, you are absolutely right :cry: Luckily there's a simple strategy to avoid bugs this causes i.e. strong typing - but that's refactoring work, definitely not this PR

src/ripple/rpc/handlers/AccountTx.cpp
Bronek: Done that now, thanks !

src/ripple/app/misc/NetworkOPs.cpp
Bronek: Actually, having seen your (good, but unfortunately breaking API version 1) idea to remove `ledger_index` from inside transaction JSON output, I will try to do that properly both for `ledger_index` and for transaction `hash`. Hopefully in this PR, but unsure as this might come with larger churn that I am want here.

Bronek: ok done! It wasn't so bad - see the latest commit https://github.com/XRPLF/rippled/pull/4775/commits/95b6055ac487910b3eb2ba5d6194b94c751042b0

src/ripple/rpc/handlers/Tx.cpp
```diff
@@ -311,17 +327,51 @@ populateJsonResponse(
...
+            {
+                response[jss::tx_json] = result.txn->getJson(
+                    JsonOptions::include_date, false, false, {std::ref(hash)});
+                RPC::insertDeliverMax(
+                    response[jss::tx_json],
+                    sttx->getTxnType(),
+                    context.apiVersion);
+            }
+
+            if (result.ledgerHash)
```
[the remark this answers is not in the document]
Bronek: I will add a comment here, thanks

src/test/rpc/TransactionEntry_test.cpp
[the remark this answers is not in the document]
Bronek: I think that's because  your proposed change has removed `ledger_index` from inside transaction in API version 1. I avoided this mistake in my attempt to improve `getJson`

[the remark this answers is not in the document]
Bronek: also, re. your suggestions branch, what's the story behind this suggestion in `LedgerMaster::isValidated` ?

```
         // SQL database doesn't match ledger chain
         clearLedger(seq);
+         app_.getInboundLedgers().acquire(
+                       *hash, seq, InboundLedger::Reason::GENERIC);
```

Bronek:
@ximinez I think I implemented all your remarks. Do you mind reviewing again ? Thank you !

src/ripple/rpc/handlers/AccountTx.cpp
Bronek: @mDuo13 you might want to update documentation of `account_tx` - this function only ever returns transactions from validated ledger. It's not a new API version 2 behaviour.

src/test/rpc/TransactionEntry_test.cpp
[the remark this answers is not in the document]
ckeshava: I know that you faithfully copied the previous (API version 1) behavior over to the new place, to ensure backwards compatibility.

Since we have detected that there is a problem with a ledger, we have an opportunity to correct it. After erasing the wrong ledger data, we can fetch the correct ledger through the `acquire` call.

Does any other process in `rippled` rectify the broken ledgers? (I believe some components of FLR can solve this problem)

src/test/rpc/AccountTx_test.cpp
ckeshava: yeah! Strong typing is a great idea

src/ripple/rpc/handlers/AccountTx.cpp
ckeshava: On a similar note, I feel line 360 (in this file -- setting of the `jss::validated` field) can be pulled out and placed here (along with other fields like `ledger_index`, `ledger_hash`, etc)

ckeshava (changes requested):
hey bronek, I really like the latest set of changes on `getJson`, it really simplifies things. I've left a few minor comments.

I think the presence of `jss::ledger_index` inside `tx_json` still needs to be addressed.

src/ripple/app/misc/impl/Transaction.cpp
```diff
@@ -167,16 +167,25 @@ Transaction::load(
...
-        ret[jss::ledger_index] = mInLedger;
+        if (!(options & JsonOptions::disable_API_prior_V2))
+        {
+            // Behaviour before API version 2
+            ret[jss::inLedger] = mLedgerIndex;
+        }
+
+        // TODO: disable_API_prior_V3 to disable output of both `date` and
+        // `ledger_index` elements (taking precedence over include_date)
+        ret[jss::ledger_index] = mLedgerIndex;
```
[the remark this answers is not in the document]
Bronek: No, this must be enabled in API version 1, but can be _also_ present in version 2. Since in API version 1 there's no alternative, I'd rather keep in version 2 as well, to make transition to version 2 easier. It's enough that folks need to deal with `hash` field changing location in this release.

src/ripple/app/misc/Transaction.h
```diff
@@ -24,10 +24,11 @@
 #include <ripple/beast/utility/Journal.h>
 #include <ripple/protocol/ErrorCodes.h>
 #include <ripple/protocol/Protocol.h>
+#include <ripple/protocol/STBase.h>
```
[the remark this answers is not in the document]
Bronek: Because of the definition of `JsonOptions`

src/ripple/rpc/handlers/AccountTx.cpp
```diff
@@ -326,10 +325,34 @@ populateJsonResponse(
...
+                    auto const json_tx =
+                        (context.apiVersion > 1 ? jss::tx_json : jss::tx);
+                    if (context.apiVersion > 1)
+                    {
+                        jvObj[json_tx] = txn->getJson(
+                            JsonOptions::include_date |
+                                JsonOptions::disable_API_prior_V2,
+                            false);
+                        jvObj[jss::hash] = to_string(
+                            txn->getSTransaction()->getTransactionID());
```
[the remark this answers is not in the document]
Bronek: thanks, I will take it :) was looking for function like this in `Transaction` and missed :rofl:

src/ripple/protocol/STTx.h
```diff
@@ -108,6 +110,13 @@ class STTx final : public STObject, public CountedObject<STTx>
 
     Json::Value
     getJson(JsonOptions options) const override;
+
+    /**
+    If `hash` is set, will store transaction id inside the provided string.
```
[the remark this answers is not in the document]
Bronek: No, that's not strictly correct. Even though that's true when we do not have API version 3, it would have to be rewritten when we _do_ have it, to mean that we disable features of API versions prior to version 2, and then document similarly another enum `JsonOptions::disable_API_prior_V3` - all of which is extra work we do not really need to do _at all_, because I took care to give these enums really descriptive names. 

So, I just removed the comment instead. Thanks for spotting it!

src/ripple/rpc/handlers/AccountTx.cpp
Bronek: actually it needs to be moved - thanks for spotting

Bronek:
> I'd like to draw your attention to the `jss::tx_json` field. In `apiVersion > 1`, is it supposed to contain the `jss::ledger_index` and `jss::date` fields? The [Recommendations](https://github.com/XRPLF/rippled/issues/4727) suggest that these fields need to be removed.

no, want to keep it where it is (in addition to the new location) in API version  2; then remove in API version 3

src/ripple/app/misc/impl/Transaction.cpp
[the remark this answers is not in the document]
ckeshava: I see, I thought it needs to be removed from the `tx_json` field.

src/test/rpc/AccountTx_test.cpp
```diff
@@ -120,22 +120,56 @@ class AccountTx_test : public beast::unit_test::suite
...
+                            (payment[jss::tx_json][jss::DeliverMax] ==
+                             "10000000010") &&
+                            (!payment[jss::tx_json].isMember(jss::Amount)) &&
+                            (!payment[jss::tx_json].isMember(jss::hash)) &&
+                            (payment[jss::hash] ==
+                             "9F3085D85F472D1CC29627F260DF68EDE59D42D1D0C33E345"
+                             "ECF0D4CE981D0A8") &&
+                            (payment[jss::validated] == true) &&
+                            (payment[jss::ledger_index] == 3) &&
+                            (payment[jss::ledger_hash] ==
```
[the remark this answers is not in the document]
ckeshava: these tests are more stringent - they check that the `tx_json` field does not contain the `ledger_index`. You might want to skip these because it hasn't been implemented in the code base

src/ripple/protocol/STTx.h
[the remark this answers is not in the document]
ckeshava: ok makes sense

ckeshava (approved):
thanks bronek, this PR looks good.

src/ripple/rpc/handlers/AccountTx.cpp
ckeshava: did you move this field?  I can't spot it for this if-block

src/ripple/app/misc/Transaction.h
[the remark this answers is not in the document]
ckeshava: Okay. I'm able to compile the code without these imports, but this is not a strong opinion.

src/ripple/rpc/handlers/AccountTx.cpp
Bronek: it's above the `if` block now, as not to break API version 1 users of this method.

ckeshava: thanks, github hadn't updated the code 😅  👍

src/ripple/rpc/handlers/AccountTx.cpp
intelliot: https://github.com/XRPLF/xrpl-dev-portal/pull/2250

ximinez (changes requested):
Partial review. Some of these comments may be outdated due to changes since I originally wrote them.

src/ripple/protocol/STBase.h
```diff
@@ -31,7 +31,58 @@
...
+struct JsonOptions
+{
+    using underlying_t = unsigned int;
+    underlying_t value;
+
+    enum values : underlying_t {
+        none = 0u,
+        include_date = 1u,
+        disable_API_prior_V2 = 2u,
+        _all = 3u  // NOTE see `operator~` and bump as needed when adding enums
```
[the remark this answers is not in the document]
Bronek: Oh, good idea - will do that

src/ripple/app/misc/impl/Transaction.cpp
```diff
@@ -167,16 +167,25 @@ Transaction::load(
...
     {
-        ret[jss::inLedger] = mInLedger;  // Deprecated.
-        ret[jss::ledger_index] = mInLedger;
+        if (!(options & JsonOptions::disable_API_prior_V2))
+        {
+            // Behaviour before API version 2
+            ret[jss::inLedger] = mLedgerIndex;
+        }
+
+        // TODO: disable_API_prior_V3 to disable output of both `date` and
```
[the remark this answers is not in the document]
Bronek: yes. See my other comment on naming

src/ripple/protocol/STTx.h
```diff
@@ -108,6 +110,13 @@ class STTx final : public STObject, public CountedObject<STTx>
...
+    Otherwise it will be stored as nested jss::hash element inside the returned
+    JSON. Additionally, if `hash` is set and `binary` is true, will not create
+    nested jss::tx for binary hex; instead will return it as JSON string
+    */
```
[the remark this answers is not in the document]
Bronek: Thanks, removed now

src/ripple/app/ledger/impl/LedgerMaster.cpp
```diff
@@ -603,6 +603,55 @@ LedgerMaster::clearLedger(std::uint32_t seq)
...
+    if (ledger.info().validated)
+        return true;
+
+    auto const seq = ledger.info().seq;
+    try
+    {
+        // Use the skip list in the last validated ledger to see if ledger
+        // comes before the last validated ledger (and thus has been
+        // validated).
+        auto hash = walkHashBySeq(seq, InboundLedger::Reason::GENERIC);
```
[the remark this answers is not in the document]
Bronek: Thanks, will do

src/ripple/protocol/STBase.h
```diff
@@ -31,7 +31,58 @@
...
+/// Note, should be treated as flags that can be | and &
+struct JsonOptions
+{
+    using underlying_t = unsigned int;
+    underlying_t value;
+
+    enum values : underlying_t {
+        none = 0u,
+        include_date = 1u,
+        disable_API_prior_V2 = 2u,
```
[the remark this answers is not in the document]
Bronek: There is a reason why I want to define this in an exclusive, rather than inclusive manner. The reason is that `prior_V2` is by definition _bounded_ i.e. its meaning does not change as we add new versions. So, if we add version 3 but keep our code say `disable_API_prior_V2` that simply implies that there are no changes between V2 and V3 (which should be the default for most of API, unless we want our users to hate us). On the other hand, `post_V1` is _unbounded_ - as we add new version, it becomes _implicitly_ included in the _meaning_ of `post_V1` and that is not what we want, unless we wanted to either rewrite all the call sites or tolerate potentially confusing name. Similar reason why I am not using `select_API_V2` which, although bounded, implies incompatibility with future versions (hence, again, would need to rewrite call sites as we add new versions).

Yes I know this goes against `if (apiVersion > 1)` programming pattern on which I seem to have settled, but such code can be easily (re)structured to make sense as we add new versions - e.g. you can nest `if (apiVersion > 2)` inside, or change to `switch` statement or use something else, because it's all imperative style (could be changed to OO or functional, but the same applies - your code is still saying "here's _how_ I want to achieve my goal"). Enums on the other hand are declarative (your code says "here's _what_ I want to achieve"), which makes them more constrained and powerful - and also why it is important to give them _bounded_ semantics.

src/ripple/protocol/STTx.h
```diff
@@ -24,11 +24,13 @@
 #include <ripple/protocol/Feature.h>
 #include <ripple/protocol/PublicKey.h>
 #include <ripple/protocol/Rules.h>
+#include <ripple/protocol/STBase.h>
```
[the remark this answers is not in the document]
Bronek: Thanks, indeed.

src/ripple/protocol/impl/STTx.cpp
```diff
@@ -236,15 +237,31 @@ Json::Value STTx::getJson(JsonOptions) const
...
-        ret[jss::tx] = strHex(s.peekData());
-        ret[jss::hash] = to_string(getTransactionID());
+        Json::Value ret = strHex(s.peekData());
         return ret;
     }
-    return getJson(options);
+    // We want virtualy the same output as `getJson(JsonOptions)` overload
+    // above, but without the hash element. Since `getJson(JsonOptions)`
+    // is using STObject::getJson(JsonOptions::none), we use it here as well
+    return STObject::getJson(JsonOptions::none);
```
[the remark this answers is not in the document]
Bronek: The proposed code is incorrect - in API version 2 we are not storing binary inside `tx` element, instead we return it directly (just like we return non-binary). As for returning transaction id - the way I wrote it, the decision to use transaction id (and/or any other field we need, and indeed we do need ledger index) belongs to the caller. No need to expand the return value; just use the accessors which are right there. And yes I do not like accessors either, but fixing _that_ belongs to a refactoring PR rather than this one.

However, I do see what you mean by _almost_ duplicating the two options, that indeed can be confusing to the reader.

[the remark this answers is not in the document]
Bronek: ok rewrote it to reduce code duplication, the only duplicate part now is setting `ret[jss::hash] = to_string(getTransactionID());` to support API version 1. Hopefully one day we can rid of it.

src/ripple/protocol/STBase.h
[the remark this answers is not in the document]
Bronek: I decided to use binary literals with digit separator and manual padding - I think it's readable now

https://github.com/XRPLF/rippled/pull/4775/files#diff-f2cc18725ab9c3dcd52ab8f6b93a0672d2527b49462ea89198c8d1380005b5ccR40-R49

src/test/rpc/Subscribe_test.cpp
```diff
@@ -22,6 +22,7 @@
 #include <ripple/core/ConfigSections.h>
 #include <ripple/protocol/Feature.h>
 #include <ripple/protocol/jss.h>
+#include "ripple/json/json_value.h"
```
ximinez, line 25: `""` instead of `<>`?

src/test/rpc/Subscribe_test.cpp
```diff
@@ -307,6 +308,85 @@ class Subscribe_test : public beast::unit_test::suite
         BEAST_EXPECT(jv[jss::status] == "success");
     }
 
+    void
+    testTransactions_APIv2()
```
ximinez, line 312: Would it be good to rename the original `testTransactions` to `testTransactions_APIv1` so there's no confusion?

src/test/rpc/TransactionEntry_test.cpp
```diff
@@ -143,29 +144,43 @@ class TransactionEntry_test : public beast::unit_test::suite
...
+                BEAST_EXPECT(resIndex[jss::hash] == txhash);
+                BEAST_EXPECT(resIndex[jss::validated] == true);
+                BEAST_EXPECT(!resIndex[jss::tx_json].isMember(jss::Amount));
+
+                if (BEAST_EXPECT(!close_time_iso.empty()))
+                    BEAST_EXPECT(
+                        resIndex[jss::close_time_iso] == close_time_iso);
+            }
+            else
+                BEAST_EXPECT(resIndex[jss::tx_json][jss::hash] == txhash);
```
ximinez, line 182: Could you add checks for the same fields, or their expected absence, as in the `apiVersion > 1` block?

src/test/rpc/TransactionEntry_test.cpp
```diff
@@ -244,10 +265,13 @@ class TransactionEntry_test : public beast::unit_test::suite
     "SigningPubKey" : "0324CAAFA2212D2AEAB9D42D481535614AED486293E1FB1380FF070C3DD7FB4264",
     "TransactionType" : "AccountSet",
     "TxnSignature" : "3044022007B35E3B99460534FF6BC3A66FBBA03591C355CC38E38588968E87CCD01BE229022071A443026DE45041B55ABB1CC76812A87EA701E475BBB7E165513B4B242D3474",
-    "hash" : "F4E9DF90D829A9E8B423FF68C34413E240D8D8BB0EFD080DF08114ED398E2506"
```
ximinez, line 247: I hate to see the hash check just be thrown away. Add a `hash` parameter to `check_tx`, and then check for it in the appropriate place depending on API version.

src/test/rpc/TransactionEntry_test.cpp
```diff
@@ -333,17 +360,18 @@ class TransactionEntry_test : public beast::unit_test::suite
...
     }
 
 public:
     void
     run() override
     {
         testBadInput();
-        testRequest();
+        testRequest(1);
+        testRequest(2);
```
ximinez, line 374: This should use the `forAllApiVersions` function to call `testRequest`.

src/test/rpc/Transaction_test.cpp
```diff
@@ -763,7 +790,8 @@ class Transaction_test : public beast::unit_test::suite
         testRangeCTIDRequest(features);
         testCTIDValidation(features);
         testCTIDRPC(features);
-        testRequest(features);
+        testRequest(features, 1);
+        testRequest(features, 2);
```
ximinez, line 794: This should also use the `forAllApiVersions` helper function.

ximinez (changes requested):
I have a handful of comments here, but they're all pretty minor.

src/ripple/protocol/STBase.h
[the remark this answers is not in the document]
ximinez: You're right - I flubbed the examples I gave. Obviously you figured out what I meant.

src/ripple/protocol/STBase.h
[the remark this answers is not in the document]
ximinez: Yes, all good points. Agreed.

src/ripple/rpc/handlers/AccountTx.cpp
```diff
@@ -325,16 +324,39 @@ populateJsonResponse(
...
-                    jvObj[jss::tx] = txn->getJson(JsonOptions::include_date);
                     auto const& sttx = txn->getSTransaction();
                     RPC::insertDeliverMax(
-                        jvObj[jss::tx], sttx->getTxnType(), context.apiVersion);
+                        jvObj[json_tx], sttx->getTxnType(), context.apiVersion);
                     if (txnMeta)
                     {
                         jvObj[jss::meta] =
                             txnMeta->getJson(JsonOptions::include_date);
-                        jvObj[jss::validated] = true;
```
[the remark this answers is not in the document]
Bronek: I was thinking about it and even wrote this block
```c++
}
else
{
    JLOG(context.j.error())
        << __func__
        << " Missing transaction medatata on (supposedly) "
            "validated ledger, index="
        << txn->getLedger();
    // Override response and we are done.
    response = Json::Value(Json::objectValue);
    break;
}
```
... but then, seeing we set `jvObj[jss::validated] = true;` in two more locations in the current release, decided that this PR is not the right place to clean this up. So, assert it is.

src/test/rpc/TransactionEntry_test.cpp
Bronek: There already is `txhash` parameter in `check_tx` and it is exactly what you are looking for i.e. `env.tx()->getTransactionID()`. However, I do agree that throwing these hash values is bad, so I added them back to verify the exact transaction hashes they were originally referring to, but with a code structure that's more natural for this particular check. Also, in the same spirit I added ledger hashes.

src/test/rpc/TransactionEntry_test.cpp
Bronek: Thanks, done.

Bronek: Also, thanks to this feedback from you, I found and fixed (harmless, but still) bug in `STTx::getJson` - we were populating `hash` inside transaction output from API version 2, if overload without `bool binary` was used.

ximinez (approved):

src/test/rpc/Subscribe_test.cpp
```diff
@@ -163,7 +164,7 @@ class Subscribe_test : public beast::unit_test::suite
     }
 
     void
-    testTransactions()
+    testTransactions_APIv1()
```
arihantkothari, line 167: This PR is great and I'm approving it. But I think we should also add `testcase("transactions API Version 1")`. Small change :) I know that a bunch of tests don't have the testcase, and I'm not sure why... but if we have the facility, we should use it 😆

src/test/rpc/Transaction_test.cpp
```diff
@@ -694,15 +694,13 @@ class Transaction_test : public beast::unit_test::suite
     }
 
     void
-    testRequest(FeatureBitset features)
+    testRequest(FeatureBitset features, unsigned apiVersion)
```
arihantkothari, line 697: Also a nit: I know unsigned = unsigned int but I think it might be better to have the same everywhere aka` unsigned int`. But this is personal preference and I'm fine with both :)

arihantkothari (approved):

src/test/rpc/Transaction_test.cpp
arihantkothari: or better if we have an alias for `unsigned int` for apiVersions

Bronek: That kind of a thing is probably best left for `.clang-format` rules . As for strong typing - I am for it, but that's (similarly to formatting rules) is outside of the scope of this PR

arihantkothari: okay 👍

PR #4781, opened 2023-10-23
feat: add user version of `feature` RPC

mvadari:

This PR modifies the `feature` RPC to have a non-admin mode that allows conversion from the hex encodings of amendments to human-readable strings.

Example:
```
> {
    "command": "feature",
    "feature": "2E2FB9CF8A44EB80F4694D38AADAE9B8B7ADAFD2F092E10068E61C98C4F092B0"
  }
< {
    "result": {
      "2E2FB9CF8A44EB80F4694D38AADAE9B8B7ADAFD2F092E10068E61C98C4F092B0": {
        "enabled": false,
        "name": "fixUniversalNumber",
        "supported": true
      }
    },
    "type": "response"
  }
```

### Context of Change

This has been requested by many explorer operators, so that the on-ledger hex amendment names can be decoded.

### Type of Change

### API Impact

- [x] Public API: New feature (new methods and/or new fields)

## Test Plan
Added tests to handle this case.

### DIFF
--- a/src/ripple/app/misc/impl/AmendmentTable.cpp
+++ b/src/ripple/app/misc/impl/AmendmentTable.cpp
@@ -388,6 +388,7 @@ private:
         Json::Value& v,
         uint256 const& amendment,
         AmendmentState const& state,
+        bool isAdmin,
         std::lock_guard<std::mutex> const& lock) const;
 
     void
@@ -428,9 +429,9 @@ public:
     firstUnsupportedExpected() const override;
 
     Json::Value
-    getJson() const override;
+    getJson(bool isAdmin) const override;
     Json::Value
-    getJson(uint256 const&) const override;
+    getJson(uint256 const&, bool isAdmin) const override;
 
     bool
     needValidatedLedger(LedgerIndex seq) const override;
@@ -906,13 +907,14 @@ AmendmentTableImpl::injectJson(
     Json::Value& v,
     const uint256& id,
     const AmendmentState& fs,
+    bool isAdmin,
     std::lock_guard<std::mutex> const&) const
 {
     if (!fs.name.empty())
         v[jss::name] = fs.name;
 
     v[jss::supported] = fs.supported;
-    if (!fs.enabled)
+    if (!fs.enabled && isAdmin)
     {
         if (fs.vote == AmendmentVote::obsolete)
             v[jss::vetoed] = "Obsolete";
@@ -921,7 +923,7 @@ AmendmentTableImpl::injectJson(
     }
     v[jss::enabled] = fs.enabled;
 
-    if (!fs.enabled && lastVote_)
+    if (!fs.enabled && lastVote_ && isAdmin)
     {
         auto const votesTotal = lastVote_->trustedValidations();
         auto const votesNeeded = lastVote_->threshold();
@@ -936,7 +938,7 @@ AmendmentTableImpl::injectJson(
 }
 
 Json::Value
-AmendmentTableImpl::getJson() const
+AmendmentTableImpl::getJson(bool isAdmin) const
 {
     Json::Value ret(Json::objectValue);
     {
@@ -947,6 +949,7 @@ AmendmentTableImpl::getJson() const
                 ret[to_string(e.first)] = Json::objectValue,
                 e.first,
                 e.second,
+                isAdmin,
                 lock);
         }
     }
@@ -954,16 +957,19 @@ AmendmentTableImpl::getJson() const
 }
 
 Json::Value
-AmendmentTableImpl::getJson(uint256 const& amendmentID) const
+AmendmentTableImpl::getJson(uint256 const& amendmentID, bool isAdmin) const
 {
     Json::Value ret = Json::objectValue;
-    Json::Value& jAmendment = (ret[to_string(amendmentID)] = Json::objectValue);
 
     {
         std::lock_guard lock(mutex_);
         AmendmentState const* a = get(amendmentID, lock);
         if (a)
-            injectJson(jAmendment, amendmentID, *a, lock);
+        {
+            Json::Value& jAmendment =
+                (ret[to_string(amendmentID)] = Json::objectValue);
+            injectJson(jAmendment, amendmentID, *a, isAdmin, lock);
+        }
     }
 
     return ret;
--- a/src/ripple/rpc/handlers/Feature1.cpp
+++ b/src/ripple/rpc/handlers/Feature1.cpp
@@ -22,6 +22,7 @@
 #include <ripple/app/misc/AmendmentTable.h>
 #include <ripple/net/RPCErr.h>
 #include <ripple/protocol/ErrorCodes.h>
+#include <ripple/protocol/Feature.h>
 #include <ripple/protocol/jss.h>
 #include <ripple/rpc/Context.h>
 
@@ -37,6 +38,7 @@ doFeature(RPC::JsonContext& context)
     if (context.app.config().reporting())
         return rpcError(rpcREPORTING_UNSUPPORTED);
 
+    bool const isAdmin = context.role == Role::ADMIN;
     // Get majority amendment status
     majorityAmendments_t majorities;
 
@@ -47,7 +49,7 @@ doFeature(RPC::JsonContext& context)
 
     if (!context.params.isMember(jss::feature))
     {
-        auto features = table.getJson();
+        auto features = table.getJson(isAdmin);
 
         for (auto const& [h, t] : majorities)
         {
@@ -69,13 +71,18 @@ doFeature(RPC::JsonContext& context)
 
     if (context.params.isMember(jss::vetoed))
     {
+        if (!isAdmin)
+            return rpcError(rpcNO_PERMISSION);
+
         if (context.params[jss::vetoed].asBool())
             table.veto(feature);
         else
             table.unVeto(feature);
     }
 
-    Json::Value jvReply = table.getJson(feature);
+    Json::Value jvReply = table.getJson(feature, isAdmin);
+    if (!jvReply)
+        return rpcError(rpcBAD_FEATURE);
 
     auto m = majorities.find(feature);
     if (m != majorities.end())
--- a/API-CHANGELOG.md
+++ b/API-CHANGELOG.md
@@ -28,6 +28,12 @@ The `network_id` field was added in the `server_info` response in version 1.5.0 
 
 ## XRP Ledger server version 2.0.0
 
+### Additions in 2.2
+
+Additions are intended to be non-breaking (because they are purely additive).
+
+- `feature`: A non-admin mode that uses the same formatting as admin RPC, but hides potentially-sensitive data.
+
 ### Additions in 2.0
 
 Additions are intended to be non-breaking (because they are purely additive).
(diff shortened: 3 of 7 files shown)


JST5000:
As an additional note, this is a feature I tried to find when I had when trying to set up our docker container for CI since the `Amendments` stanza to force-enable amendments is really hard to read without the human-readable versions of amendments.

dangell7:
This seems overkill.

```python
import hashlib

def hash_string(k: str) -> str:
    return hashlib.sha512(k.encode('utf-8')).hexdigest().upper()[:64]
```

```typescript
import * as crypto from 'crypto';

function sha512Hash(k: string): string {
  return crypto.createHash('sha512').update(k, 'utf8').digest('hex').toUpperCase().substring(0, 64);
}
```

mvadari:
> This seems overkill.

This works for going from the name to the hex, but not the other way around, since you can't reverse a hash.

pkcs8:
Would the non-admin version of this command have similar response shape as the admin version? I.e., if I skip the `feature` property, and just send `{"command":"feature"}`, it should return a map of all amendments known to this rippled sans voting data.

mvadari:
> Would the non-admin version of this command have similar response shape as the admin version? I.e., if I skip the `feature` property, and just send `{"command":"feature"}`, it should return a map of all amendments known to this rippled sans voting data.

I didn't add that here because it's a bit harder to do. But if that's a need, I can add it.

pkcs8:
I read rippled websocket url from the environment, and often switch and use admin and non-admin ports interchangeably. I'm not sure  now complicated it would be, but would be great if response shape remains same so I don't have to write conditions. I like how `account_objects` command works with its `type` filter.

lathanbritz:
> I didn't add that here because it's a bit 

I'm actually a +1 for this I need a way to figure the list of amendments on the network. Currently using external API for mainnet or hardcoded. Would help a lot.

mvadari:
@pkcs8 @lathanbritz I have switched to the existing admin feature format (removing the admin fields of course) and added the ability to get a list of all features in https://github.com/XRPLF/rippled/pull/4781/commits/43a1ce7b1516b93d3abbf034050afa27c641c956.

mvadari:
@intelliot can I add @ximinez and @thejohnfreeman as reviewers (if they're okay with that)?

intelliot:
@mvadari I requested ximinez and thejohnfreeman but the PRs listed here are higher priority: https://github.com/XRPLF/rippled/milestone/8 <- these are 2.0 release blockers

This PR (#4781) is currently planned for 2.1.0

dangell7:
We don't like this on `feature` endpoint. We believe opening up feature like this is dangerous. We have opted to move this to the ServerDefinitions.cpp file.

What are your thoughts on this?

I know this goes against what XRPScan asked for but in our opinion its a "safer" solution.

mvadari:
> We don't like this on `feature` endpoint. We believe opening up feature like this is dangerous. We have opted to move this to the ServerDefinitions.cpp file.
> 
> What are your thoughts on this?
> 
> I know this goes against what XRPScan asked for but in our opinion its a "safer" solution.

This code change is pretty small and therefore easy to audit and test from a security perspective. There is also prior art for functions that have separate user versions and admin versions, like `server_info`, and there have been no issues with those. I'm personally not concerned, but there's also a reason we go through code review 🙂

src/ripple/rpc/handlers/Feature1.cpp
```diff
@@ -47,12 +49,15 @@ doFeature(RPC::JsonContext& context)
...
 
-        for (auto const& [h, t] : majorities)
+        if (isAdmin)
         {
-            features[to_string(h)][jss::majority] =
-                t.time_since_epoch().count();
+            for (auto const& [h, t] : majorities)
+            {
+                features[to_string(h)][jss::majority] =
+                    t.time_since_epoch().count();
```
ximinez, line 59: I wouldn't think `majority` time is sensitive information either.

src/ripple/rpc/handlers/Feature1.cpp
```diff
@@ -67,6 +72,10 @@ doFeature(RPC::JsonContext& context)
     if (!feature && !feature.parseHex(context.params[jss::feature].asString()))
         return rpcError(rpcBAD_FEATURE);
 
+    if (!isAdmin && context.params.isMember(jss::vetoed))
+    {
+        return rpcError(rpcNO_PERMISSION);
+    }
     if (context.params.isMember(jss::vetoed))
     {
         if (context.params[jss::vetoed].asBool())
```
ximinez, line 81: You could make this a lot safer and easier to reason about by simply moving the `isAdmin` check inside the voting block.
```
    if (context.params.isMember(jss::vetoed))
    {
        if (!isAdmin)
            return rpcError(rpcNO_PERMISSION);

        if (context.params[jss::vetoed].asBool())
            table.veto(feature);
        else
            table.unVeto(feature);
    }
```

So now the "voting" code is all in one block, and the logic is like
1. Am I trying to vote?
   1. Am I *allowed* to vote?
      1. No? Fail!
      2. Yes? Vote!

src/ripple/rpc/handlers/Feature1.cpp
```diff
@@ -75,7 +84,13 @@ doFeature(RPC::JsonContext& context)
...
     }
 
-    Json::Value jvReply = table.getJson(feature);
+    Json::Value jvReply = table.getJson(feature, isAdmin);
+    if (!jvReply)
+        return rpcError(rpcBAD_FEATURE);
+    if (!isAdmin)
+    {
+        return jvReply;
+    }
```
ximinez, line 94: If you changing the `majority` check above, change it here, too.

ximinez (changes requested):

src/ripple/app/misc/impl/AmendmentTable.cpp
```diff
@@ -921,7 +923,7 @@ AmendmentTableImpl::injectJson(
     }
     v[jss::enabled] = fs.enabled;
 
-    if (!fs.enabled && lastVote_)
+    if (!fs.enabled && lastVote_ && isAdmin)
```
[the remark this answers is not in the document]
mvadari: That's what I was concerned about.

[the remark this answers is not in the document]
ximinez: In that case, better safe than sorry.

ximinez (approved):
Looks good and useful

thejohnfreeman (approved):

intelliot:
`Perf Test Desired (Optional)`: This is a new feature in the API, but the perf impact should be negligible.

Internal tracker: RPFC-109

intelliot:
I think this can be merged once the (trivial) merge conflicts are resolved

mvadari:
@intelliot merge conflicts have been resolved and I believe this is ready to be merged.

mvadari:
The coverage check failed to upload, but I don't have permissions to re-run it.

intelliot:
@ximinez or @Bronek or @mvadari - could you have a quick look to see if the "2 lines in your changes are missing coverage" would be easy to address?

I'm ok to merge as-is if that's what y'all want, but let's confirm that's what we want to do (rather than adding some coverage first)

Bronek:
> @ximinez or @Bronek or @mvadari - could you have a quick look to see if the "2 lines in your changes are missing coverage" would be easy to address?
> 
> I'm ok to merge as-is if that's what y'all want, but let's confirm that's what we want to do (rather than adding some coverage first)

This is referring lines 917 and 926 of `AmendmentTable.cpp`. There's a newly added `isAdmin` parameter which is used in branching instructions in these lines. Looking at `AmendmentTable_test.cpp` this parameter seem to be only set to `true`, but I do not know about other tests. I think it should be easy to add unit tests to cover `isAdmin=false`, too.

Bronek:
> > @ximinez or @Bronek or @mvadari - could you have a quick look to see if the "2 lines in your changes are missing coverage" would be easy to address?
> > I'm ok to merge as-is if that's what y'all want, but let's confirm that's what we want to do (rather than adding some coverage first)
> 
> This is referring lines 917 and 926 of `AmendmentTable.cpp`. There's a newly added `isAdmin` parameter which is used in branching instructions in these lines. Looking at `AmendmentTable_test.cpp` this parameter seem to be only set to `true`, but I do not know about other tests. I think it should be easy to add unit tests to cover `isAdmin=false`, too.

I looked at test coverage report of the `develop` branch and see similar partial branch coverage remarks 
image

at [codecov.io](https://app.codecov.io/gh/XRPLF/rippled/blob/develop/src%2Fripple%2Fapp%2Fmisc%2Fimpl%2FAmendmentTable.cpp)

This makes me thing that the coverage warning is not new actually, and also that it refers to `fs.enabled` and not `isAdmin`.

intelliot:
- @mvadari to confirm whether she thinks, given the above, this is still ready to be merged now

mvadari:
> @mvadari to confirm whether she thinks, given the above, this is still ready to be merged now

I checked the tests and I'm pretty sure both branches of `isAdmin` are already checked.

intelliot:
> (cc @Bronek to double check since he's already taken a look at some of the code)

src/ripple/app/misc/impl/AmendmentTable.cpp
Bronek, line 926: You are missing case case for `lastVote_=true` and `isAdmin=false`

Bronek (changes requested):
I recommend adding something as simple as 

```
--- a/src/test/app/AmendmentTable_test.cpp
+++ b/src/test/app/AmendmentTable_test.cpp
@@ -638,6 +638,20 @@ public:
         BEAST_EXPECT(enabled.empty());
         BEAST_EXPECT(majority.empty());

+        uint256 const unsupportedID = amendmentId(unsupported_[0]);
+        {
+            Json::Value const unsupp =
+                table->getJson(unsupportedID, false)[to_string(unsupportedID)];
+            BEAST_EXPECT(unsupp.size() == 0);
+        }
+
+        table->veto(unsupportedID);
+        {
+            Json::Value const unsupp =
+                table->getJson(unsupportedID, false)[to_string(unsupportedID)];
+            BEAST_EXPECT(!unsupp[jss::vetoed].asBool());
+        }
+
         votes.emplace_back(testAmendment, validators.size());

         doRound(
```

This will cover the one missing test case for `isAdmin=false` _after_ we have voted; the other missing branches are around `fs.enabled` and are unrelated to this PR, and so should be ignored here.

Bronek:
> > (cc @Bronek to double check since he's already taken a look at some of the code)

Thank you @intelliot ; I looked some more and there is one thing that can be improved in unit tests; added it in my feedback.

I use `cmake -Dcoverage-format=json-detail -Dcoverage=ON . . .` to prepare `coverage.json` report and see from it that the proposed addition will increase branch coverage of line 929 in `AmendmentTable.cpp` from 6/8 to 7/8 .

Bronek (approved):

ximinez:
@mvadari Is this ready to merge?

mvadari:
> @mvadari Is this ready to merge?

@ximinez with Bronek's ✅ , yep - I think it was just waiting on the 2.1 release.

ximinez:
Proposed commit message:

```
feat: add user version of `feature` RPC (#4781)

* hides potentially sensitive data
* uses same formatting as admin RPC
```

mvadari:
> Proposed commit message:
> 
> ```
> feat: add user version of `feature` RPC (#4781)
> 
> * hides potentially sensitive data
> * uses same formatting as admin RPC
> ```

nit: I'd switch the order of the bullets
```
feat: add user version of `feature` RPC (#4781)

* uses same formatting as admin RPC
* hides potentially sensitive data
```

intelliot:
From my perspective, @ximinez or @seelabs are welcome to merge this. (no big rush though)

PR #4785, opened 2023-10-25
Fix unit test api_version to enable api_version 2

pwang200:
This PR attempts to address the unit test api_version issues discussed in a previous PR #4568

- the command line API still uses `apiMaximumSupportedVersion`.
- the unit test RPCs will uses `apiMinimumSupportedVersion` if unspecified.


src/test/rpc/RPCCall_test.cpp
```diff
@@ -6403,7 +6403,7 @@ std::string
 updateAPIVersionString(const char* const req)
 {
     static std::string version_str =
-        std::to_string(RPC::apiMaximumSupportedVersion);
+        std::to_string(RPC::apiMinimumSupportedVersion);
     static auto place_holder = "%MAX_API_VER%";
```
arihantkothari, line 6407: If we are putting minimum version as `version_str`. Should this line instead be `static auto place_holder = "%MIN_API_VER%";` ?

src/test/rpc/RPCCall_test.cpp
```diff
@@ -6403,7 +6403,7 @@ std::string
 updateAPIVersionString(const char* const req)
 {
     static std::string version_str =
-        std::to_string(RPC::apiMaximumSupportedVersion);
+        std::to_string(RPC::apiMinimumSupportedVersion);
```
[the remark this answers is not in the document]
arihantkothari: I know most of the times (maybe all the times), `apiVersionIfUnspecified = apiMinimumSupportedVersion` always.

Bronek (changes requested):

ximinez:
IMHO, this solution is backwards. When a new API version becomes officially supported, we _want_ incompatible unit tests to break. That is a signal that they are version dependent and need to be updated to test all API versions (preferably), or a least updated to the latest API version.

Bronek:
> IMHO, this solution is backwards. When a new API version becomes officially supported, we _want_ incompatible unit tests to break. That is a signal that they are version dependent and need to be updated to test all API versions (preferably), or a least updated to the latest API version.

Agree. Exposing `apiVersion` parameter, where appropriate, to unit tests, would make that easier.

ximinez:
> Agree. Exposing `apiVersion` parameter, where appropriate, to unit tests, would make that easier.

There's a helper function, `forAllApiVersions` for just this purpose.

src/test/rpc/RPCCall_test.cpp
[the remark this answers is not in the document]
pwang200: I cannot say no. I.e., probably make sense to use `apiVersionIfUnspecified` as default.

src/test/rpc/RPCCall_test.cpp
pwang200: I changed it back to `RPC::apiMaximumSupportedVersion` and we will see what we eventually decide to use as default. Could be `apiVersionIfUnspecified` as you said earlier.

src/ripple/net/RPCCall.h
```diff
@@ -65,18 +65,20 @@ fromNetwork(
     std::unordered_map<std::string, std::string> headers = {});
 }  // namespace RPCCall
 
-/** Given a rippled command line, return the corresponding JSON.
+/** Given a rippled unit test rpc command, return the corresponding JSON.
  */
 Json::Value
-cmdLineToJSONRPC(std::vector<std::string> const& args, beast::Journal j);
+unitTestCmdToJSONRPC(std::vector<std::string> const& args, beast::Journal j);
```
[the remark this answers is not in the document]
pwang200: Thanks, it is moved.

src/ripple/net/impl/RPCCall.cpp
```diff
@@ -1561,7 +1563,8 @@ rpcClient(
     try
     {
         Json::Value jvRpc = Json::Value(Json::objectValue);
-        jvRequest = rpcCmdLineToJson(args, jvRpc, logs.journal("RPCParser"));
+        auto apiVersion = fromCmdLine? RPC::apiMaximumSupportedVersion : RPC::apiMinimumSupportedVersion;
```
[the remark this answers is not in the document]
pwang200: Thanks, changed.

src/ripple/net/impl/RPCCall.cpp
```diff
@@ -1698,7 +1701,7 @@ fromCommandLine(
     const std::vector<std::string>& vCmd,
     Logs& logs)
 {
-    auto const result = rpcClient(vCmd, config, logs);
+    auto const result = rpcClient(vCmd, config, logs, true);
```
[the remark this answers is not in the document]
pwang200: done.

pwang200:
I kind of agree with what @ximinez said about "this solution is backwards". That could be the reason we had `RPC::apiMaximumSupportedVersion` as default. I changed the default back to `RPC::apiMaximumSupportedVersion`. (shows my (lack of) understanding of the issue.) I addressed @Bronek comments and the apiVersion is unit tests' choice, instead of having to use what the command line use. Maybe we should reopen #4568?

intelliot:
@pwang200 - from your perspective, is this PR ready for review + merge?

src/ripple/net/RPCCall.h
```diff
@@ -65,18 +65,23 @@ fromNetwork(
     std::unordered_map<std::string, std::string> headers = {});
 }  // namespace RPCCall
 
-/** Given a rippled command line, return the corresponding JSON.
- */
+
```
intelliot, line 68:
```suggestion
```

Bronek (approved):
Nice, that's an overall improvement.

src/ripple/net/impl/RPCCall.cpp
```diff
@@ -1698,7 +1672,8 @@ fromCommandLine(
     const std::vector<std::string>& vCmd,
     Logs& logs)
 {
-    auto const result = rpcClient(vCmd, config, logs);
+    auto const result =
+        rpcClient(vCmd, config, logs, RPC::apiMaximumSupportedVersion);
```
Bronek, line 1676: Note, this means that as soon as we bump `apiMaximumSupportedVersion`, the `tx_history` and `ledger_header` commands will stop working (as they are no longer recognized in API version 2). Ideally at this point we should remove them from `parseCommand` in this file (and also remove function `RPCParser::parseTxHistory`). This probably needs its own issue.

pwang200:
> @pwang200 - from your perspective, is this PR ready for review + merge?

yes

src/ripple/net/RPCCall.h
```diff
@@ -65,18 +65,22 @@ fromNetwork(
...
 }  // namespace RPCCall
 
-/** Given a rippled command line, return the corresponding JSON.
- */
 Json::Value
-cmdLineToJSONRPC(std::vector<std::string> const& args, beast::Journal j);
+rpcCmdToJson(
+    std::vector<std::string> const& args,
+    Json::Value& retParams,
+    beast::Journal j,
```
seelabs, line 72: Usually we make `journal` the last parameter.

src/ripple/net/RPCCall.h
```diff
@@ -65,18 +65,22 @@ fromNetwork(
...
+    unsigned int apiVersion);
```
seelabs, line 73: Maybe not for this PR, but we should consider making apiVersion either an enum type with one enum per API version or maybe a strongly-typed int (int with a tag, like we do with base_uint).

seelabs (approved):
:+1: This looks good. Approving as-is, but left two nits to consider. (I'm fine if the nits are not addressed).

arihantkothari (approved):
lgtm 👍

intelliot:
re-requested review since a new commit was pushed after the last review.

Also, should we try to merge #4775 prior to this one?

Bronek (approved):

Bronek:
> Also, should we try to merge #4775 prior to this one?

There's no need to delay this PR

Issue #4787, opened 2023-10-25
Remove dependency on <ranges> 


Bronek:
In https://github.com/XRPLF/rippled/pull/4759 I inadvertently managed to sneak in dependency on `<ranges>` . Given that at this time we want this project to compile on relatively old compilers (gcc 11, clang 13) which may not have this header, this dependency is potentially breaking. Also, the code in question (below) can trivially switch to something else.

## Problematic code excerpt

```
#include <ranges>

auto make_vector(auto const& input) requires std::ranges::range<decltype(input)>
{
    return std::vector(std::ranges::begin(input), std::ranges::end(input));
}
```

PR #4788, opened 2023-10-25
Remove include <ranges>

Bronek:

Remove dependency on `<ranges>` header, since it is not implemented by all compilers which we want to support. Resolves https://github.com/XRPLF/rippled/issues/4787 

### Context of Change

We only use ranges for `std::ranges::begin` and `std::ranges::end`, and for a matching concept. This can be trivially replaced with `std::begin` , `std::end` and an equivalent, trivial to write, concept.

The code change in question only affects unit tests.

### Type of Change

- [ ] Documentation update
- [ ] Release

### API Impact

### DIFF
--- a/src/test/jtx/TestHelpers.h
+++ b/src/test/jtx/TestHelpers.h
@@ -35,15 +35,15 @@ namespace jtx {
 
 // Helper to make vector from iterable
 template <typename T>
-concept iterable = requires(T& v) {
-                       std::begin(v);
-                       std::end(v);
-                   };
+concept iterable = requires(T& v)
+{
+    std::begin(v);
+    std::end(v);
+};
 
 template <typename Input>
 auto
-make_vector(Input const& input)
-    requires iterable<Input>
+make_vector(Input const& input) requires iterable<Input>
 {
     return std::vector(std::begin(input), std::end(input));
 }
--- a/src/test/jtx/TestHelpers.h
+++ b/src/test/jtx/TestHelpers.h
@@ -27,17 +27,25 @@
 #include <ripple/protocol/jss.h>
 #include <test/jtx/Env.h>
 
-#include <ranges>
+#include <vector>
 
 namespace ripple {
 namespace test {
 namespace jtx {
 
 // Helper to make vector from iterable
+template <typename T>
+concept iterable = requires(T& v) {
+                       std::begin(v);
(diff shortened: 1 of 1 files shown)


scottschurr (approved):
Looks great.  Works for me using Xcode 13.4.1.

scottschurr:
Having slept on this, the current solution works for me.  However, I'm having some pause about introducing a custom `concept` that is only used in one place.  We could remove the `concept` and use a `requires` expression instead, like this:
```
template <typename Input>
auto
make_vector(Input const& input)
    requires requires(Input& v) {
                 std::begin(v);
                 std::end(v);
             }
{
    return std::vector(std::begin(input), std::end(input));
}
```
That would remove the need for a custom `concept`.  Just a thought.

Bronek:
> Having slept on this, the current solution works for me. However, I'm having some pause about introducing a custom `concept` that is only used in one place. We could remove the `concept` and use a `requires` expression instead, like this:
> 
> ```
> template <typename Input>
> auto
> make_vector(Input const& input)
>     requires requires(Input& v) {
>                  std::begin(v);
>                  std::end(v);
>              }
> {
>     return std::vector(std::begin(input), std::end(input));
> }
> ```
> 
> That would remove the need for a custom `concept`. Just a thought.

It's generic concept, and so is this function. I consider `requires requires` a little ugly - they are good to describe contract that a given function needs, but if something is generic enough I prefer to give it a name. Of course that raises a question of location of such a utility - the location is not generic at all here. But IMO it's better to have a named concept and plan moving it elsewhere in the future (how about XRPLF/toolset repo ?) rather than something that pretends to be very specific when actually it is not.

seelabs (approved):

seelabs:
I'm approving this, but in my opinion the concept doesn't add much. I probably prefer the code without the concept.

Bronek:
> I'm approving this, but in my opinion the concept doesn't add much. I probably prefer the code without the concept. The concept isn't used to help with an overload, and the expected interface of `Input` is trivially clear from code.

OK, I think appropriate compromise is to put a comment why this code is the way it is and why we need it temporarily i.e. because we do not yet have `<ranges>` on all supported compilers

scottschurr (approved):
:+1:

Bronek:
@intelliot this is ready to merge

PR #4794, opened 2023-11-01
Refactor

dangell7:

Change all test suite names to match parent directory name.

### Context of Change

This is better because when reviewing the test failures for the first time its hard to understand where the test file is located in the source code. Sometimes the test ripple.app... means its in the test/app directory, sometimes ripple.app means its in the test/rpc directory. 

This makes it clear.

### Type of Change

- [ ] Documentation update
- [ ] Release

### API Impact

### DIFF
--- a/src/test/app/Check_test.cpp
+++ b/src/test/app/Check_test.cpp
@@ -2725,6 +2725,6 @@ public:
     }
 };
 
-BEAST_DEFINE_TESTSUITE(Check, tx, ripple);
+BEAST_DEFINE_TESTSUITE(Check, app, ripple);
 
 }  // namespace ripple
--- a/src/test/app/CrossingLimits_test.cpp
+++ b/src/test/app/CrossingLimits_test.cpp
@@ -562,7 +562,7 @@ public:
     }
 };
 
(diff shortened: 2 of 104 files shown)


intelliot:
Thanks for this PR. Could you update the PR title to be bit more specific/descriptive of the changes being made? (`Refactor` is a bit vague :)

scottschurr (approved):
:+1: LGTM.  It makes sense for these to be consistent with _something_ and the subdirectory makes a lot of sense for that.  I appreciate you making this change.

I did a `grep -r BEAST_DEFINE_ src/test`, and it looks like you missed just a few.  You could do the `grep` yourself, or you could cherry-pick the top-most commit here if you want to fix up those last few remnants: https://github.com/scottschurr/rippled/commits/dangell7-refactor/

mvadari:
Does this PR need to be updated after everything got moved around?

bthomee:
@dangell7 We're going to moving off the existing test framework onto ctest/doctest, to allow us to more easily run specific test suites/cases and see better test failure outputs. The new framework also allows us to move the tests out of the binary, with very limited exceptions for security-related tests that will remain embedded. The initial change should be submitted soon, and the existing tests will be moved over slowly over time. As such, I'll close this PR & hope the test failures will be easier to understand in the future with the new framework 🤞

PR #4799, opened 2023-11-03
Revert "fix: stabilize voting threshold for amendment majority"

scottschurr:

This reverts commit 2bb8de030fc887787f311e428e2e91466e41a240.  This pull request consists exclusively of a git-generated revert commit.

Manual testing showed some unexpected behavior in the amendment flapping changes.  We're removing the commit since the release date is close and fixing the problem has some risk associated with it.  A tested fixed commit will be put into a later release.

#### Type of Change

## Future Tasks

Fix the bug, test, and re-submit the corrected pull request.

### DIFF
--- a/src/ripple/app/main/Application.cpp
+++ b/src/ripple/app/main/Application.cpp
@@ -1333,9 +1333,6 @@ ApplicationImp::setup(boost::program_options::variables_map const& cmdline)
                 << "Invalid entry in [" << SECTION_VALIDATOR_LIST_SITES << "]";
             return false;
         }
-
-        // Tell the AmendmentTable who the trusted validators are.
-        m_amendmentTable->trustChanged(validators_->getQuorumKeys().second);
     }
     //----------------------------------------------------------------------
     //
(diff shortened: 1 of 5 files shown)


intelliot:
I believe/assume this is a clean revert, so this should be a relatively quick/safe code review.

Bronek (approved):
I verified that this corresponds to clean revert and that there were no unrelated changes in the affected files that might affect the semantics of this revert

scottschurr:
For what it's worth, testing is ongoing.  At this point it's unclear whether there's an actual bug in this code or if the testing is running afoul of other amendments (specifically `fixAmendmentMajorityCalc`) not being enabled.

In short, let's not pull the trigger on this revert immediately.  There's more news coming from QA.  But it's good to have this available if we need it.

sgramkumar:
Executed tests to verify the original commit https://github.com/XRPLF/rippled/commit/2bb8de030fc887787f311e428e2e91466e41a240 and they work as expected. 👍

scottschurr:
We have determined that expectations for manual testing were confounded due to interactions with `fixAmendmentMajorityCalc`.  Now that we understand that interaction, manual testing of the original code shows us the expected (desirable) results.  So I'm closing this revert pull request.

I really appreciate all the work that @sgramkumar put into testing and understanding the nature of the failures.  :trophy:

Issue #4790, opened 2023-10-29
OrderBook update logic query


sublimator:
```C++
void
OrderBookDB::setup(std::shared_ptr<ReadView const> const& ledger)
{
    if (!app_.config().standalone() && app_.getOPs().isNeedNetworkLedger())
    {
        JLOG(j_.warn()) << "Eliding full order book update: no ledger";
        return;
    }

    auto seq = seq_.load();

    if (seq != 0)
    {
        if ((seq > ledger->seq()) && ((ledger->seq() - seq) < 25600))
            return;

        if ((ledger->seq() <= seq) && ((seq - ledger->seq()) < 16))
            return;
    }
```

This part seems a bit confusing to me:

```c++
        if ((seq > ledger->seq()) && ((ledger->seq() - seq) < 25600))
```

If bigger > smaller && (smaller - bigger) < 25600 

The old logic:

```C++
        auto seq = ledger->info().seq;

        // Do a full update every 256 ledgers
        if (mSeq != 0)
        {
            if (seq == mSeq)
                return;
            if ((seq > mSeq) && ((seq - mSeq) < 256))
                return;
            if ((seq < mSeq) && ((mSeq - seq) < 16))
                return;
        }
```

sublimator:
Maybe it was supposed to be:
```c++
if ((ledger->seq() > seq) && ((ledger->seq() - seq) < 25600))
```

I see before the variable `seq` was used for the passed in ledger' sequence, but now `seq` is the last sequence (which before was mSeq) so maybe there was some copy/paste issues?

seelabs:
Thanks @sublimator. After a quick look this does look like a bug to me. Let me take a closer look now.

seelabs:
@sublimator There's a PR for this here: https://github.com/XRPLF/rippled/pull/4890

PR #4804, opened 2023-11-04
feat: tweak sparse array settings

sublimator:


sublimator:
I had (have?) a plan for loading binary dumps that involved creating a tree efficiently with no reallocations and while beating my head against the SM code, in the process noticed some opportunities to make the creation of inner nodes a little more efficient.

sublimator:
From ledger 83,592,201:

image

This is drawn using the prior boundary points of 2,4,6,16  
There does seem to be a fair bit of overallocation in the middle of the tree

The TaggedPointer's stated reason to live:

```
    The motivation for this class is saving RAM. A large percentage of inner
    nodes only store a small number of children. Memory can be saved by
    storing the inner node's children in sparse arrays. Measurements show
    that on average a typical SHAMap's inner nodes can be stored using only
    25% of the original space.
```

The "Exact Allocation" referenced there is based on calculations involving spending 8 bytes extra "somewhere" to encode the exact number of children. 

I'm still foggy on the lifecycle of nodes in rippled, and when they can be treated as immutable, but it seems like in principle for hashed nodes, it ought to be able to have the exact amount of children allocated "somehow".

src/ripple/shamap/impl/SHAMapInnerNode.cpp
```diff
@@ -132,22 +132,67 @@ SHAMapInnerNode::makeFullInner(
...
+    std::uint16_t isBranch = 0;
+    for (int i = 0; i < branchFactor; ++i)
+    {
+        // If the serialized inners included a branch header we wouldn't have to
+        // do this walk to check for non-zero branches. Note that the node store
+        // actually defines a custom compression specifically for inner nodes,
+        // and decompresses it into this form before handing off to other
+        // components. It may be worth investigating whether the compressed form
+        // is actually more useful here.
+        if (si.getBitString<256>().isNonZero())
```
sublimator, line 150: Could probably just scan the Slice bytes directly?

src/ripple/basics/Slice.h
```diff
@@ -187,6 +187,14 @@ class Slice
 
         return {data_ + pos, std::min(count, size() - pos)};
     }
+
+    /** Checks if any byte in the slice is non-zero */
+    [[nodiscard]] bool
+    isNonZero() const noexcept
```
sublimator, line 193: This probably better as a free "somewhere" function eh

src/ripple/shamap/impl/SHAMapInnerNode.cpp
```diff
@@ -132,35 +132,86 @@ SHAMapInnerNode::makeFullInner(
...
-    auto ret = std::make_shared<SHAMapInnerNode>(0, branchFactor);
-
     SerialIter si(data);
 
+    // Determine the non-empty branches so we only allocate once what we need
+    // with no reallocation and moving around data.
+    // When allocating 16 branches, it only has to turn around and reallocate
+    // the right size and compact, which is less than ideal. In principle,
+    // with less allocations we should see less fragmentation in the arena.
+    std::uint16_t isBranch = 0;
```
sublimator, line 142: This probably isn't really much of a win, but if TaggedPointer goes to all its trouble, in context, it seems justified?

src/ripple/shamap/impl/SHAMapInnerNode.cpp
```diff
@@ -169,24 +220,28 @@ SHAMapInnerNode::makeCompressedInner(Slice data)
...
+        if (pos >= branchFactor || prevPos > pos || !ix.has_value())
+            Throw<std::runtime_error>("invalid CI node");
 
-        if (hashes[pos].isNonZero())
-            ret->isBranch_ |= (1 << pos);
+        hashes[ix.value()].as_uint256() = hash;
+        prevPos = pos;
     }
 
+    // Should effectively be be a noop at this point
```
sublimator, line 244: so .... just remove it??

src/ripple/shamap/impl/TaggedPointer.ipp
```diff
@@ -33,10 +33,34 @@ namespace {
...
+//  |        1 |   578,512 |
+//  |        2 | 2,840,810 | <-------
+//  |        3 |   624,837 |
+//  |        4 |   120,458 | <-------
+//  |        5 |    45,802 |
+//  |        6 |    73,588 |
+//  |        7 |   129,694 |
+//  |        8 |   184,122 |
+//  |        9 |   208,545 |
+//  |       10 |   187,800 | <-------
```
sublimator, line 52: @mtrippled @seelabs

The sampling data from ledger 83,592,201 suggests a potential optimization opportunity by adjusting boundaries to cater to common child counts in inner nodes. This could lead to memory allocation improvements.

Thoughts?

seelabs:
Thanks @sublimator I will assign myself to review this.

seelabs:
@sublimator Just FYI: I _think_ this repository has some the analysis I did to choose the boundaries: https://github.com/seelabs/shamap_compress/ In particular, I should review this python notebook: https://github.com/seelabs/shamap_compress/blob/master/shamap_compress.ipynb I haven't had time to review this PR yet, but I'll look forward to refreshing my memory on this stuff.

sublimator:
Thanks for taking a look Scott :)

I am trying to stay off the computer for a few days - but I looked  (on
phone, so technically ...) quickly at the child count histogram in that
repo and saw pretty different numbers to what I got.

When I am back on, I will double-check the child counts I posted. I did
note some deltas in some of the stats in historical vs latest ledgers I
looked at recently.

sublimator:
At the risk of tricking myself back onto the computer -perhaps we could
expand the scope here to some kind of stats command?

src/ripple/shamap/impl/TaggedPointer.ipp
thejohnfreeman: I can confirm that given these counts, these are the optimal boundaries, when limited to exactly four, that minimize the allocated size. I brute-forced the computation in a [spreadsheet](https://docs.google.com/spreadsheets/d/13PUUJrpNc-gvbQwTNoxR-keMVGRajpEu6EsghvLGPy0/edit?usp=sharing). A closed form solution wasn't coming to me, but I'm interested to know if there is one.

sublimator: @thejohnfreeman 

Thanks for looking. 

p.s. feel free to push to this branch

src/ripple/shamap/impl/SHAMapInnerNode.cpp
sublimator:
```C++
void
SHAMapInnerNode::resizeChildArrays(std::uint8_t toAllocate)
{
    hashesAndChildren_ =
        TaggedPointer(std::move(hashesAndChildren_), isBranch_, toAllocate);
}
```

```C++
inline TaggedPointer::TaggedPointer(
    TaggedPointer&& other,
    std::uint16_t isBranch,
    std::uint8_t toAllocate)
    : TaggedPointer(std::move(other))
{
    auto const oldNumAllocated = capacity();
    toAllocate = numAllocatedChildren(toAllocate);
    if (toAllocate == oldNumAllocated)
        return;
```

```C++
inline TaggedPointer::TaggedPointer(TaggedPointer&& other) : tp_{other.tp_}
{
    other.tp_ = 0;
}
```


It seems it would still be a new pointer, but a redundant one.

I removed the call to `resizeChildArrays` in `SHAMapInnerNode::makeFullInner` but was inconsistent, forgetting to apply it `SHAMapInnerNode::makeCompressedInner`.

I just pushed a commit to consistently remove it.

src/ripple/rpc/handlers/SHAMapInfo.cpp
```diff
@@ -0,0 +1,72 @@
...
+#include <ripple/shamap/SHAMapLeafNode.h>
+#include <ripple/shamap/SHAMapTreeNode.h>
+
+namespace ripple {
+
+Json::Value
+doSHAMapInfo(RPC::JsonContext& context)
+{
+    Json::Value jvResult;
+    auto& composition = jvResult["composition"] = Json::objectValue;
```
sublimator, line 36: If there's interest, I can expand this to include some more interesting stats, but kept scoped for now. 

Out of sight, out of mind?

src/ripple/rpc/handlers/SHAMapInfo.cpp
```diff
@@ -0,0 +1,72 @@
...
+    });
+
+    composition["inners_count"] = inners;
+    composition["leaves_count"] = leaves;
+    auto& branches = composition["inners_child_count"] = Json::objectValue;
+
+    for (int i = 0; i < SHAMapInnerNode::branchFactor; ++i)
+    {
+        auto count = to_string(i + 1);
+        branches[(count.size() == 1 ? "0" + count : count)] = children[i];
```
sublimator, line 66: Maybe just return an array to be formatted by consumer scripts?

sublimator:
@seelabs 

?

sublimator:
@seelabs @thejohnfreeman 

bumpy bump :)

sublimator:
Bump :)

sublimator:
Ping :)

intelliot:
Much of the team is currently on vacation. We will certainly have a look at this in January. Thanks for the ping!

sublimator:
Ok, I will disable the ping bot until NYD! haha

sublimator:
Ping

sublimator:
Ping :)

thejohnfreeman (approved):

thejohnfreeman:
One request for the ping bot: remove old ping comments when adding a new one. @sublimator

sublimator:
Hahaha :)

sublimator:
image

seelabs:
Sorry for the delay. I'll try to get to this later this week.

sublimator:
Nice one! Thanks Scott

intelliot:
@sublimator as these changes are expected to impact performance, please provide these details for the performance team to understand and test:

1. Is this a new feature, bug fix, or improvement to existing functionality?
2. What behavior/functionality does the change impact?
3. In what processing can the impact be measured? Be as specific as possible - e.g. RPC client call, payment transaction that involves LOB, AMM, caching, DB operations, etc.
4. Does this change affect concurrent processing - e.g. does it involve acquiring locks, multi-threaded processing, or async processing?

Thanks!

sublimator:
@intelliot 

Will wait on Scott's input first (@seelabs) as I'm much busier now on other things than when I first opened this PR.

That said, I can quickly say:
1. I expect the performance changes to be rather minor
2. This PR represents initial baby steps and a renewed focus on this part of the code after many years (the child counts seem to have changed considerably)
3. Includes the beginnings of a command to get shamap stats which could prove useful for subsequent changes that could improve performance

seelabs:
I'll plan on running this patch overnight and compare memory footprints a server with and without this patch.

sublimator:
Thanks!

seelabs:
mem_usage
mem_diff
mem_percent_change

This is what I got from last night's run, but I'll need to do it again. I realized that one servers had different path finding settings, and that makes a difference. I'll re-run the test again tonight. (I thought I'd show this result anyway, but we'll have a better apples to apples comparison tomorrow).

sublimator:
@seelabs 

Btw, somewhat related to path finding, I stumbled across this a while back:
https://github.com/XRPLF/rippled/issues/4790

I think the order book is only used for path finding right?

seelabs:
Here are the charts from last night. I made sure both servers had the same configs (other than db paths since I ran them on the same machine:
mem_usage
mem_diff
mem_percent_change

Given those charts it looks like this change uses a bit more memory than the current version - low hundreds of mb. Honestly, that's surprising, I did expect a memory savings. Let me double check the scripts I used to run this to make sure I didn't make a dumb error like swapping "base" and "new" (after a quick check the script looks right to me though).

seelabs:
@sublimator Thanks for the bug report on path finding. I'll comment there.

sublimator:
>>> Given those charts it looks like this change uses a bit more memory
than the current version

Interesting!
I will have to have a look at it again.
Been a while.

sublimator:
Do you have any theory as to why it might use more ram? Assuming it is and consistently so?
I expected it to be marginal (even bordering on theoretical) at best, but I still thought there would be a measurable improvement. I spent more time eyeballing the [experiment] branch to be fair, but figured this would be less controversial.

sublimator:
How much time do you have to dig into this?
What can we salvage here?
Does reverting the boundaries "fix" the memory increase?
If so, potentially worth just merging that until have time for next round of investigations?
But I'm not sure ripple "works like that"

thejohnfreeman:
New is greater than old in the first chart, but the delta is negative in the subsequent charts, which suggests to me there is some mistaken switching going on. Is the first chart mislabeled, or are the other charts oddly reporting the change _from_ new _to_ old instead of the other way?

seelabs:
@thejohnfreeman The delta is negative because it's measuring "memory savings". So positive is a memory savings since I use these scripts to find out "how much memory savings do I get with some new code". Yeah, that's a little weird, but it is working as indended.

seelabs:
@sublimator Let me revert the boundary changes and make sure there's no memory difference and if they're aren't I'll take a close look at the code. I doing the correct allocation upfront is a good change (I think, I still need to look closely...) so yeah, I do think we want some of these changes even if we don't want the boundary changes yet.

sublimator:
I don't really have bandwidth atm to dive back into this much. While you're
in the area it may be worth looking at the experiment branch (in so much as
it seemed to reduce memory )

I think the shamap_info command could grow into something useful

Cheers

sublimator:
@seelabs @thejohnfreeman 

Did you ever look at this more?
What's the plan?
What should be done with this?

Thanks!

sublimator:
Ok, I will close this

sublimator:
@seelabs 

Do you think it's possible that one of the servers in your test was using an existing/primed database?

PR #4805, opened 2023-11-04
[experiment] feat: compact after hashing 

sublimator:
See: https://github.com/XRPLF/rippled/pull/4804#issuecomment-1793341910

Reduce re-allocations
1. Allocate storage for children precisely when creating from compressed wire and "full inner" node store. 
2. Hack the TaggedPointer to store MetaData in first 8 bytes, to store the child count, to allow compacting to exact size
3. Allocate 16 branches for new nodes and only compact just before hashing

Seems to shave a few hundred megs off, even when 'spending' the 8 bytes for the child count. 
*Seems* to work, so seems worth investigating further along this line.


src/ripple/shamap/impl/TaggedPointer.h
```diff
@@ -65,10 +65,12 @@ class TaggedPointer
         A moved-from object will have a tp_ of zero.
     */
     std::uintptr_t tp_ = 0;
-    /** bit-and with this mask to get the tag bits (lowest two bits) */
-    static constexpr std::uintptr_t tagMask = 3;
-    /** bit-and with this mask to get the pointer bits (mask out the tag) */
-    static constexpr std::uintptr_t ptrMask = ~tagMask;
+    std::uint32_t children_ = 0;
```
sublimator, line 68: ofc this ruins atomicity of pointer switcheroo - but does show the potential ram savings

src/ripple/shamap/impl/SHAMapInnerNode.cpp
```diff
@@ -74,50 +74,34 @@ SHAMapInnerNode::getChildIndex(int i) const
 std::shared_ptr<SHAMapTreeNode>
 SHAMapInnerNode::clone(std::uint32_t cowid) const
 {
-    auto const branchCount = getBranchCount();
-    auto const thisIsSparse = !hashesAndChildren_.isDense();
-    auto p = std::make_shared<SHAMapInnerNode>(cowid, branchCount);
+    // We are cloning it for writing, and who knows what it'll be used for,
```
sublimator, line 77: You could probably use some depth based heuristics here to try not to overallocate - but given what I /assume/ is the lifecycle of a node - not sure how much of a win that would be.

src/ripple/shamap/impl/SHAMapInnerNode.cpp
```diff
@@ -132,35 +116,86 @@ SHAMapInnerNode::makeFullInner(
...
 
-    auto ret = std::make_shared<SHAMapInnerNode>(0, branchFactor);
-
     SerialIter si(data);
 
+    // Determine the non-empty branches so we only allocate once what we need
+    // with no reallocation and moving around data.
+    // When allocating 16 branches, it only has to turn around and reallocate
+    // the right size and compact, which is less than ideal. In principle,
+    // with less allocations we should see less fragmentation in the arena.
```
sublimator, line 125: Seems like there's `branchFactor` allocation zones, so the "in principle" fragmentation probably not an issue, so much as redundant copying?

sublimator:
No interest, closing

PR #4807, opened 2023-11-06
Support for the mold linker

seelabs:
This adds support for the mold linker on linux. On my system this significantly improves link time (from about 18 seconds to about 1 second).


sublimator:
Siiiick, but awwww, linux only?

sublimator:
https://github.com/bluewhalesystems/sold

Payware, but would be totally worth it

thejohnfreeman (approved):

Bronek:
I tested this under Ubuntu 22.04 container - in this environment `gold` is always installed because it is included in `binutils` , and `mold` is an extra package, so we can have both, but we cannot have `mold` alone without `gold`. With the last commit "Use first non-default linker" it all works and the link time with `mold` is so fast, I can barely spot it (not kidding!)

Bronek (approved):
Thanks, it works !

Issue #4811, opened 2023-11-13
Regression: crash when rendering JSON binary transaction with API v.2 enabled


Bronek:
## Issue Description

In PR #4775 the API version 2 output of `tx` method has significantly changed, and in the process has also the JSON type of output from `STTx::getJson` . If we enable both API v. 2 and binary output, that JSON is no longer an object but a string. It follows that it cannot have any additional JSON fields added, and yet we try to add fields inside `Transaction::getJson` which causes a crash.

## Steps to Reproduce

Run `tx` method requesting binary output with `api_version` 2

## Expected Result

Output transaction data with `tx_blob` field (among others) 

## Actual Result

Crash.

PR #4812, opened 2023-11-13
Fix 2.0 regression in tx method with binary output

Bronek:

Fixes regression caused by https://github.com/XRPLF/rippled/pull/4775 
This PR resolves https://github.com/XRPLF/rippled/issues/4811 

### Context of Change

The `STTx::getJson` was changed in API version 2 to allow for returning binary output as a single hex-encoded JSON string (not a JSON object). This caused regression inside `Transaction::getJson` which will attempt to append fields to that not-an-object, resulting in failed JSON operation. This change is fixing this regression and adds a relevant test.

### Type of Change

- [ ] Documentation update
- [ ] Release


seelabs (approved):
:+1:

src/test/rpc/Transaction_test.cpp
```diff
@@ -792,6 +862,8 @@ class Transaction_test : public beast::unit_test::suite
         testCTIDRPC(features);
         test::jtx::forAllApiVersions(
             std::bind_front(&Transaction_test::testRequest, this, features));
+        test::jtx::forAllApiVersions(
+            std::bind_front(&Transaction_test::testBinaryRequest, this));
```
arihantkothari, line 866: Why not just:
```
test::jtx::forAllApiVersions(
           std::bind_front(&Transaction_test::testRequest, this, features)
           std::bind_front(&Transaction_test::testBinaryRequest, this));
```
instead of calling `test::jtx::forAllApiVersions` twice ?

Bronek: good point, will do

arihantkothari (approved):
Lgtm 👍 Thanks for the changes.

src/test/rpc/Transaction_test.cpp
```diff
@@ -774,6 +776,74 @@ class Transaction_test : public beast::unit_test::suite
...
+        auto const USD{gw["USD"]};
+
+        env.fund(XRP(1000000), alice, gw);
+        std::shared_ptr<STTx const> txn = env.tx();
+        BEAST_EXPECT(
+            to_string(txn->getTransactionID()) ==
+            "3F8BDE5A5F82C4F4708E5E9255B713E303E6E1A371FD5C7A704AFD1387C23981");
+        env.close();
+        std::shared_ptr<STObject const> meta =
+            env.closed()->txRead(env.tx()->getTransactionID()).second;
```
ximinez, line 800: Should this be reading `txn->getTransactionID()`?

src/test/rpc/Transaction_test.cpp
```diff
@@ -791,7 +861,8 @@ class Transaction_test : public beast::unit_test::suite
         testCTIDValidation(features);
         testCTIDRPC(features);
         test::jtx::forAllApiVersions(
-            std::bind_front(&Transaction_test::testRequest, this, features));
+            std::bind_front(&Transaction_test::testRequest, this, features),
+            std::bind_front(&Transaction_test::testBinaryRequest, this));
```
ximinez, line 865: Since `testBinaryRequest` doesn't take a `features` parameter, it probably makes sense to call it from `run()`.

src/test/rpc/Transaction_test.cpp
```diff
@@ -774,6 +776,74 @@ class Transaction_test : public beast::unit_test::suite
...
+        using namespace test::jtx;
+        using std::to_string;
+
+        Env env{*this};
+        Account const alice{"alice"};
+        Account const gw{"gw"};
+        auto const USD{gw["USD"]};
+
+        env.fund(XRP(1000000), alice, gw);
+        std::shared_ptr<STTx const> txn = env.tx();
```
ximinez, line 794: Could be `std::shared_ptr<STTX const> const txn`

ximinez (changes requested):

src/test/rpc/Transaction_test.cpp
Bronek: Thanks, done

src/test/rpc/Transaction_test.cpp
Bronek: Thanks, done

src/test/rpc/Transaction_test.cpp
Bronek: Thanks, done

ximinez (approved):
This is good to go as-is, but I left a suggestion to improve readability.

PR #4817, opened 2023-11-16
Workarounds for gcc-13 compatibility

Bronek:

Workaround for compilation errors with gcc-13 and other compilers relying on `libstdc++` version 13

This is temporary until actual fix https://github.com/boostorg/beast/pull/2682/ is available for us to use.

### Context of Change

Some boost.beast files (which we do use) rely on an old gcc-12 behaviour where `#include <cstdint>` was not needed even though types from this header were used. This was broken by change in libstdc++ version 13 https://gcc.gnu.org/gcc-13/porting_to.html#header-dep-changes . The necessary fix was implemented in boost.beast in https://github.com/boostorg/beast/pull/2682/ however it is not yet available. Until it is available we can use this workaround to enable compilation of `rippled` with gcc-13, clang-16 etc.

### Type of Change

- [ ] Documentation update
- [ ] Release


intelliot:
Merged due to request from Bronek and because the change is trivial.

PR #4818, opened 2023-11-17
Proposed 2.0.0-rc2

manojsdoshi:

PR #4819, opened 2023-11-17
Merging release back to develop

manojsdoshi:


ximinez (approved):
As this is the same code I just approved in #4818 , I'll approve it here, too.

scottschurr (approved):
:+1: Builds, passes unit tests, and syncs for me on macOS.

PR #4822, opened 2023-11-20
Improve lifetime management of ledger objects (`SLE`s) to prevent runaway memory usage. AKA "Is it caching? It's always caching."

### LINKED ISSUE #4224: Rippled locks up when client use pathing (all versions)
When clients connect to the node using pathing the node soon locks up.

## Steps to Reproduce
Run node and connect some clients via websocket and path_find (note they need to be requesting the path_find from this node).  https://xrpl-pathfinding.netlify.app is a simple way to do this.

## Expected Result
Node does not lock up.

## Actual Result
Node fails with, 
Screen Shot 2022-07-07 at 18 38 19
Now the node fails to restart and rippled server_info reports error.
Only way I have found to fix this is to remove the DB, then restart the node

## Environment
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 20.04.4 LTS
Release:	20.04
Codename:	focal

rippled 1.9.1
installed from apt.

ximinez:

This PR, if merged, changes `CachedView` to only cache the key needed to find each ledger object (`SLE`) in the global `CachedSLEs`, instead of a strong reference to the `SLE` itself. Further, it releases a hard reference to the last `Ledger` used for path finding work as soon as there is no more work to be done. Additionally, a bunch of logging is added to monitor cache sizes before and after sweeping for expired objects. The logging is not strictly necessary for the solution, but it doesn't appear to hurt, and can be useful if there are similar issues in the future. Finally, several more objects, and some statistics for the `CachedSLEs`, are tracked for the `get_counts` admin command.

Resolves #4224

### Context of Change

Currently, the `CachedView` object holds and keeps strong references (`std::shared_ptr`) to objects being loaded from the ledger. This prevents those objects from being completely released from the global cache (`CachedSLEs`, which is a `TaggedCache`) and releasing their memory until the owning `Ledger` is released.

This is not a significant problem during typical server operations, but can lead to memory exhaustion in some circumstances, including path finding. This looks like a memory leak, but isn't really, because the memory would be recovered _eventually_ if rippled was able to continue.

It needs to be noted that this is not a guaranteed complete solution for all path finding memory issues. It's still possible that sufficiently large `path_search` configuration values, or a sufficiently large set of steps, accounts, and trust lines could still consume enough memory to cause problems.

Problems with memory usage during path finding has been an ongoing issue. See also:
* 417cfc2 (#4061)
* #4080 
* #4099
* #4111 
* #4409 

I hope that this is the last word on this issue, but real world performance remains to be seen.

### Type of Change

### API Impact

- [X ] Public API: New feature (new methods and/or new fields)
  * Potentially adds several new key-value pairs to the output of [`get_counts`](https://xrpl.org/get_counts.html).

### Before

Certain combinations of path finding requests could cause rippled to exhaust available memory and crash, or be killed by the operating system.

### After

The same scenarios will not cause rippled to crash, or at worst will allow rippled to service many more requests and run a lot longer before crashing.

## Test Plan

This is an optimization that shouldn't affect anything other than this particular edge case. Thus existing test cases should cover it.

Beta testers are encouraged to build this branch, put it on public facing servers with path finding configured, and report back on their results.


intelliot:
- Requires additional testing since this makes changes that aren't necessarily specific to pathfinding.
- There's a smaller / more targeted change that could be made, which cleans out the pointers: just the changes in `src/ripple/app/ledger/impl/LedgerMaster.cpp`.
- Would like node operators (supporting pathfinding) to run this change to confirm that it resolves the problem.

lathanbritz:
Trying to get this built on my node box and seem to keep getting stuck here

```
[57/661] Building CXX object CMakeFiles/xrpl_core.dir/src/ripple/protocol/impl/STIssue.cpp.o
FAILED: CMakeFiles/xrpl_core.dir/src/ripple/protocol/impl/STIssue.cpp.o 
/usr/bin/g++ -DBOOST_ASIO_DISABLE_HANDLER_TYPE_REQUIREMENTS -DBOOST_ASIO_USE_TS_EXECUTOR_AS_DEFAULT -DBOOST_BEAST_ALLOW_DEPRECATED -DBOOST_CONTAINER_FWD_BAD_DEQUE -DBOOST_COROUTINES_NO_DEPRECATION_WARNING -DBOOST_FILESYSTEM_DEPRECATED -DCARES_STATICLIB -DDATE_HEADER_ONLY -DDEBUG -DGIT_COMMIT_HASH=\"7bfdc8f2bb514b43d7483f4301b453d1b4e4e1dc\" -DHAS_UNCAUGHT_EXCEPTIONS=1 -DOPENSSL_NO_SSL2 -D_DEBUG -I/home/xrpl/rippled/src -I/home/xrpl/rippled/src/secp256k1/include -I/home/xrpl/rippled/src/ed25519-donna -isystem /home/xrpl/rippled/build/proto_gen -isystem /home/xrpl/rippled/build/proto_gen/src -isystem /home/xrpl/rippled/build/proto_gen_grpc -isystem /home/xrpl/rippled/build/proto_gen_grpc/ripple/proto -isystem /root/.conan/data/openssl/1.1.1u/_/_/package/23b828d52c0630e6b0b96d2945419feb7843c4f8/include -isystem /root/.conan/data/boost/1.82.0/_/_/package/3a52104b31a9c302344831090b42e066cefe22fe/include -isystem /root/.conan/data/date/3.0.1/_/_/package/5ab84d6acfe1f23c4fae0ab88f26e3a396351ac9/include -isystem /root/.conan/data/protobuf/3.21.9/_/_/package/e7b3aae6b77ab32a1300f43e340d3856a5c70e32/include -isystem /root/.conan/data/zlib/1.2.13/_/_/package/23b828d52c0630e6b0b96d2945419feb7843c4f8/include -isystem /root/.conan/data/grpc/1.50.1/_/_/package/e4b22279357772621ebd0f91c7be7bc55142e551/include -isystem /root/.conan/data/abseil/20230125.3/_/_/package/a0d04ca949f5f2ad68af2af705b4e0f9a95967a0/include -isystem /root/.conan/data/c-ares/1.19.1/_/_/package/5a258fabd5aeaa11cbc1951f7243b664f46595f6/include -isystem /root/.conan/data/re2/20230301/_/_/package/a0d04ca949f5f2ad68af2af705b4e0f9a95967a0/include -DBOOST_ASIO_HAS_STD_INVOKE_RESULT -Wall -Wdeprecated -Wextra -Wno-unused-parameter -g -fPIC -Wno-maybe-uninitialized -Wno-deprecated-declarations -frtti -Wnon-virtual-dtor -Wno-sign-compare -Wno-char-subscripts -Wno-format -Wno-unused-local-typedefs -fstack-protector -Wno-unused-but-set-variable -Wno-deprecated -O0 -pthread -Wsuggest-override -std=c++2a -MD -MT CMakeFiles/xrpl_core.dir/src/ripple/protocol/impl/STIssue.cpp.o -MF CMakeFiles/xrpl_core.dir/src/ripple/protocol/impl/STIssue.cpp.o.d -o CMakeFiles/xrpl_core.dir/src/ripple/protocol/impl/STIssue.cpp.o -c /home/xrpl/rippled/src/ripple/protocol/impl/STIssue.cpp
In file included from /home/xrpl/rippled/src/ripple/basics/base_uint.h:28,
                 from /home/xrpl/rippled/src/ripple/protocol/UintTypes.h:24,
                 from /home/xrpl/rippled/src/ripple/protocol/Issue.h:24,
                 from /home/xrpl/rippled/src/ripple/protocol/STIssue.h:24,
                 from /home/xrpl/rippled/src/ripple/protocol/impl/STIssue.cpp:20:
/home/xrpl/rippled/src/ripple/basics/Expected.h:139:5: error: ‘requires’ does not name a type
  139 |     requires std::convertible_to<U, T> constexpr Expected(U && r)
      |     ^~~~~~~~
/home/xrpl/rippled/src/ripple/basics/Expected.h:145:9: error: ‘requires’ does not name a type
  145 |         requires std::convertible_to<U, E> &&
      |         ^~~~~~~~
/home/xrpl/rippled/src/ripple/basics/Expected.h:221:9: error: ‘requires’ does not name a type
  221 |         requires std::convertible_to<U, E> &&
      |         ^~~~~~~~
In file included from /home/xrpl/rippled/src/ripple/protocol/UintTypes.h:24,
                 from /home/xrpl/rippled/src/ripple/protocol/Issue.h:24,
                 from /home/xrpl/rippled/src/ripple/protocol/STIssue.h:24,
                 from /home/xrpl/rippled/src/ripple/protocol/impl/STIssue.cpp:20:
/home/xrpl/rippled/src/ripple/basics/base_uint.h:553:37: error: ‘strong_ordering’ in namespace ‘std’ does not name a type
  553 | [[nodiscard]] inline constexpr std::strong_ordering
      |                                     ^~~~~~~~~~~~~~~
/home/xrpl/rippled/src/ripple/basics/base_uint.h: In function ‘constexpr bool ripple::operator==(const ripple::base_uint<Bits, Tag>&, const ripple::base_uint<Bits, Tag>&)’:
/home/xrpl/rippled/src/ripple/basics/base_uint.h:578:19: error: expected primary-expression before ‘>’ token
  578 |     return (lhs <=> rhs) == 0;
      |                   ^
In file included from /home/xrpl/rippled/src/ripple/protocol/STIssue.h:24,
                 from /home/xrpl/rippled/src/ripple/protocol/impl/STIssue.cpp:20:
/home/xrpl/rippled/src/ripple/protocol/Issue.h: At global scope:
/home/xrpl/rippled/src/ripple/protocol/Issue.h:88:37: error: ‘weak_ordering’ in namespace ‘std’ does not name a type
   88 | [[nodiscard]] inline constexpr std::weak_ordering
      |                                     ^~~~~~~~~~~~~
In file included from /home/xrpl/rippled/src/ripple/protocol/STIssue.h:26,
                 from /home/xrpl/rippled/src/ripple/protocol/impl/STIssue.cpp:20:
/home/xrpl/rippled/src/ripple/protocol/STBase.h:64:67: error: ‘constexpr bool ripple::operator==(ripple::JsonOptions, ripple::JsonOptions)’ cannot be defaulted
   64 |     operator==(JsonOptions lh, JsonOptions rh) noexcept -> bool = default;
      |                                                                   ^~~~~~~
/home/xrpl/rippled/src/ripple/protocol/STBase.h:66:67: error: ‘constexpr bool ripple::operator!=(ripple::JsonOptions, ripple::JsonOptions)’ cannot be defaulted
   66 |     operator!=(JsonOptions lh, JsonOptions rh) noexcept -> bool = default;
      |                                                                   ^~~~~~~
In file included from /home/xrpl/rippled/src/ripple/protocol/impl/STIssue.cpp:20:
/home/xrpl/rippled/src/ripple/protocol/STIssue.h: In function ‘bool ripple::operator<(const ripple::STIssue&, const ripple::STIssue&)’:
/home/xrpl/rippled/src/ripple/protocol/STIssue.h:125:24: error: no match for ‘operator<’ (operand types are ‘const ripple::Issue’ and ‘const ripple::Issue’)
  125 |     return lhs.issue() < rhs.issue();
      |            ~~~~~~~~~~~ ^ ~~~~~~~~~~~
      |                     |             |
      |                     |             const ripple::Issue
      |                     const ripple::Issue
In file included from /home/xrpl/rippled/src/ripple/basics/base_uint.h:29,
                 from /home/xrpl/rippled/src/ripple/protocol/UintTypes.h:24,
                 from /home/xrpl/rippled/src/ripple/protocol/Issue.h:24,
                 from /home/xrpl/rippled/src/ripple/protocol/STIssue.h:24,
                 from /home/xrpl/rippled/src/ripple/protocol/impl/STIssue.cpp:20:
/home/xrpl/rippled/src/ripple/basics/Slice.h:220:1: note: candidate: ‘bool ripple::operator<(const ripple::Slice&, const ripple::Slice&)’
  220 | operator<(Slice const& lhs, Slice const& rhs) noexcept
      | ^~~~~~~~
/home/xrpl/rippled/src/ripple/basics/Slice.h:220:24: note:   no known conversion for argument 1 from ‘const ripple::Issue’ to ‘const ripple::Slice&’
  220 | operator<(Slice const& lhs, Slice const& rhs) noexcept
      |           ~~~~~~~~~~~~~^~~
In file included from /home/xrpl/rippled/src/ripple/protocol/impl/STIssue.cpp:20:
/home/xrpl/rippled/src/ripple/protocol/STIssue.h:123:1: note: candidate: ‘bool ripple::operator<(const ripple::STIssue&, const ripple::STIssue&)’
  123 | operator<(STIssue const& lhs, STIssue const& rhs)
      | ^~~~~~~~
/home/xrpl/rippled/src/ripple/protocol/STIssue.h:123:26: note:   no known conversion for argument 1 from ‘const ripple::Issue’ to ‘const ripple::STIssue&’
  123 | operator<(STIssue const& lhs, STIssue const& rhs)
      |           ~~~~~~~~~~~~~~~^~~
/home/xrpl/rippled/src/ripple/protocol/STIssue.h: In function ‘bool ripple::operator<(const ripple::STIssue&, const ripple::Issue&)’:
/home/xrpl/rippled/src/ripple/protocol/STIssue.h:137:24: error: no match for ‘operator<’ (operand types are ‘const ripple::Issue’ and ‘const ripple::Issue’)
  137 |     return lhs.issue() < rhs;
      |            ~~~~~~~~~~~ ^ ~~~
      |                     |    |
      |                     |    const ripple::Issue
      |                     const ripple::Issue
In file included from /home/xrpl/rippled/src/ripple/basics/base_uint.h:29,
                 from /home/xrpl/rippled/src/ripple/protocol/UintTypes.h:24,
                 from /home/xrpl/rippled/src/ripple/protocol/Issue.h:24,
                 from /home/xrpl/rippled/src/ripple/protocol/STIssue.h:24,
                 from /home/xrpl/rippled/src/ripple/protocol/impl/STIssue.cpp:20:
/home/xrpl/rippled/src/ripple/basics/Slice.h:220:1: note: candidate: ‘bool ripple::operator<(const ripple::Slice&, const ripple::Slice&)’
  220 | operator<(Slice const& lhs, Slice const& rhs) noexcept
      | ^~~~~~~~
/home/xrpl/rippled/src/ripple/basics/Slice.h:220:24: note:   no known conversion for argument 1 from ‘const ripple::Issue’ to ‘const ripple::Slice&’
  220 | operator<(Slice const& lhs, Slice const& rhs) noexcept
      |           ~~~~~~~~~~~~~^~~
In file included from /home/xrpl/rippled/src/ripple/protocol/impl/STIssue.cpp:20:
/home/xrpl/rippled/src/ripple/protocol/STIssue.h:123:1: note: candidate: ‘bool ripple::operator<(const ripple::STIssue&, const ripple::STIssue&)’
  123 | operator<(STIssue const& lhs, STIssue const& rhs)
      | ^~~~~~~~
/home/xrpl/rippled/src/ripple/protocol/STIssue.h:123:26: note:   no known conversion for argument 1 from ‘const ripple::Issue’ to ‘const ripple::STIssue&’
  123 | operator<(STIssue const& lhs, STIssue const& rhs)
      |           ~~~~~~~~~~~~~~~^~~
/home/xrpl/rippled/src/ripple/protocol/STIssue.h:135:1: note: candidate: ‘bool ripple::operator<(const ripple::STIssue&, const ripple::Issue&)’
  135 | operator<(STIssue const& lhs, Issue const& rhs)
      | ^~~~~~~~
/home/xrpl/rippled/src/ripple/protocol/STIssue.h:135:26: note:   no known conversion for argument 1 from ‘const ripple::Issue’ to ‘const ripple::STIssue&’
  135 | operator<(STIssue const& lhs, Issue const& rhs)
      |           ~~~~~~~~~~~~~~~^~~
ninja: build stopped: subcommand failed.
```

lathanbritz:
Feed back so far.

1. follow up responses to the initial path_find request seem slower than what we have today.. the production build seems to return results much more frequently.
2. memory seems to behave for the 2-3 hrs I let node run on "normal" requests and some light pathing requests. 
Screenshot 2023-12-04 at 18 47 28
3. asked few a few people to hit a front end that adds a bunch of load on the box and managed to bring it down. Here is error log...
```
2023-Dec-04 21:38:43.756954454 UTC NetworkOPs:WRN Missing offer
2023-Dec-04 21:38:43.757092579 UTC NetworkOPs:WRN Missing offer
2023-Dec-04 21:38:43.758713948 UTC NetworkOPs:WRN Missing offer
2023-Dec-04 21:38:43.759163464 UTC NetworkOPs:WRN Missing offer
2023-Dec-04 21:38:43.829825677 UTC Resource:WRN Consumer entry 127.0.0.1 dropped with balance 15010 at or above drop threshold 15000
2023-Dec-04 21:38:43.830112890 UTC Resource:WRN Consumer entry 127.0.0.1 dropped with balance 15104 at or above drop threshold 15000
rippled: /root/.conan/data/boost/1.82.0/_/_/package/3a52104b31a9c302344831090b42e066cefe22fe/include/boost/beast/websocket/detail/soft_mutex.hpp:89: bool boost::beast::websocket::detail::soft_mutex::try_lock(const T*) [with T = boost::beast::websocket::stream<boost::beast::basic_stream<boost::asio::ip::tcp, boost::asio::executor, boost::beast::unlimited_rate_policy> >::close_op<boost::asio::executor_binder<ripple::BaseWSPeer<ripple::ServerHandler, ripple::PlainWSPeer<ripple::ServerHandler> >::close(const boost::beast::websocket::close_reason&)::<lambda(const error_code&)>, boost::asio::strand<boost::asio::executor> > >]: Assertion `id_ != T::id' failed.
Aborted
```

ximinez:
> 3. asked few a few people to hit a front end that adds a bunch of load on the box and managed to bring it down. Here is error log...

That looks like a problem at the websocket layer, which path finding shouldn't affect. Can you tell if the same issue occurs under heavy load without this change?

lathanbritz:
>  Can you tell if the same issue occurs under heavy load without this change?

yeah it was fine with just me hitting the box for 2-3 hours, it did not fall over with that error.

lathanbritz:
right did some more testing off the develop branch this morning @ximinez 
 
Screenshot 2023-12-05 at 11 31 03


can validate that the issue is present outside of this change as well.

ximinez:
> I run that interface off the 1.12.0 prod build and it does not fail with this error under decent load.

When you say "prod build", do you mean the published packages? Those have asserts disabled, so even if the issue exists there, you wouldn't see it. Could you build https://github.com/XRPLF/rippled/tree/1.12.0 the same way you built `develop`, and see if the issue exists there, too?

intelliot:
> right did some more testing off the develop branch this morning @ximinez
>
> can validate that the issue is present outside of this change as well.

Given the issue appears on `develop` (without #4822 - and therefore doesn't seem to be related to this PR), can you open a new issue so we can track/discuss your findings there? @lathanbritz

src/ripple/ledger/impl/CachedView.cpp
```diff
@@ -33,25 +33,38 @@ CachedViewImpl::exists(Keylet const& k) const
 std::shared_ptr<SLE const>
 CachedViewImpl::read(Keylet const& k) const
 {
-    {
+    static CountedObjects::Counter hits{"CachedView::cache_hit"};
+    static CountedObjects::Counter hitsexpired{"CachedView::cache_hit_expired"};
```
Bronek, line 37: this name irks me, unsure why. Can we give it capitalised "E", like so: `hitsExpired` ?

src/ripple/ledger/impl/CachedView.cpp
```diff
@@ -33,25 +33,38 @@ CachedViewImpl::exists(Keylet const& k) const
...
         if (iter != map_.end())
         {
-            if (!iter->second || !k.check(*iter->second))
-                return nullptr;
+            cacheHit = true;
             return iter->second;
         }
-    }
-    auto const digest = base_.digest(k.key);
+        return base_.digest(k.key);
```
Bronek, line 50: If I read this right, the `digest` operation is expensive since it reads the whole of the object, and is the reason why we use `map_`. However before this change, `digest` call was outside of the scope of the critical section and now it is within this scope, making it unnecessarily longer than it needs to be. I would add an extra scope right before `lock_guard` and close it just above this line (just like it was done before this change).

Alternatively we could just extend this critical section to the whole of the function, as it will have some benefits (just one opportunity to block on mutex, instead of two), but it is very much a speculation whether or not that would be an improvement.

Bronek (changes requested):

src/ripple/ledger/impl/CachedView.cpp
ximinez: Sure. We can call it whatever makes sense. I chose the name based on the other names that appear at the top of a `get_counts` call:
```
      "AL_hit_rate" : 38.9652214050293,                                                        
      "AL_size" : 50,                          
      "SLE_hit_rate" : 0.9263432896252034,
      "dbKBLedger" : 6469,         
      "dbKBTotal" : 142664,                                                                    
      "dbKBTransaction" : 134785,         
      "fullbelow_size" : 1066268,     
      "historical_perminute" : 21,     
      "ledger_hit_rate" : 4.897829055786133,                                                   
      "node_read_bytes" : "3305869523",                                                        
      "node_reads_duration_us" : "72651617585", 
      "node_reads_hit" : "198704709",
      "node_reads_total" : "267571236",                                                        
      "node_writes" : "56873374",                                                              
      "node_written_bytes" : "28242167146",
[...]
```
But I don't have a problem using `hitsExpired`, etc.

src/ripple/ledger/impl/CachedView.cpp
ximinez: Good catch! Yes, I'd rather keep the critical section as narrow as possible, if only to prevent potential problems down the road.

Bronek (approved):

thejohnfreeman (approved):

ximinez:
@Bronek @thejohnfreeman I pushed a commit to change the `"miss"` label to be lower-case. I don't think it requires re-review, but I wanted to give you a heads up.

intelliot:
@sophiax851 this may require perf signoff

mtrippled (approved):
LGTM :+1:

intelliot:
Internal tracker: RPFC-79

intelliot:
`Perf SignedOff`: Examined the tagged cache sweeping during the test during lab testing and using a Mainnet node, didn’t observe abnormal behavior. No performance regression observed either.

Released in 2.0.1.

sophiax851:
It's worth mentioning that this "Perf SignedOff" testing was conducted to ensure that there is no performance regression in the normal transaction processing with the changes made in this PR which was intended to address the root cause of the issue that I identified for [#4224 ](https://github.com/XRPLF/rippled/issues/4224). There was separate testing conducted in verifying the #4224.

PR #4820, opened 2023-11-17
Consistently show ledger_index as integer on JSON output

Bronek:
## High Level Overview of Change

Several methods in API version 1 return `ledger_index` as a quoted number, i.e. string type.  This value should be consistently returned as a number. It was reported in https://github.com/XRPLF/clio/issues/722#issuecomment-1614973809 and I missed it in earlier changes meant for API version 2

### Context of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Refactor (non-breaking change that only restructures code)
- [ ] Tests (you added tests for code that already exists, or your new feature included in this PR)
- [ ] Documentation update
- [ ] Chore (no impact to binary, e.g. `.gitignore`, formatting, dropping support for older tooling)
- [ ] Release

### API Impact

- [ ] Public API: New feature (new methods and/or new fields)
- [x] Public API: Breaking change (in general, breaking changes should only impact the next api_version)
- [ ] `libxrpl` change (any change that may affect `libxrpl` or dependents of `libxrpl`)
- [ ] Peer protocol change (must be backward compatible or bump the peer protocol version)

#### Before:
JSON output from `ledger` would contain ledger as string e.g. `"ledger_index": "54300932"` 

#### After
JSON output from `ledger` would contain ledger as number e.g. `"ledger_index": 54300932`

PR #4833, opened 2023-11-22
Write proper `forAllApiVersions` used in NetworkOPs.cpp

Bronek:

We have `forAllApiVersions` in tests, but since https://github.com/XRPLF/rippled/pull/4820 we also need something similar in `NetworkOPs.cpp`. Write it properly so we can use it both in prod code and in tests

#### What's `MultiApiJson` and why do we need it (not this PR)

Class `MultiApiJson` [was added](https://github.com/XRPLF/rippled/pull/4733) in version 2.0, to enable us to publish JSON data from `NetworkOPs.cpp` (where we handle data subscriptions) to clients supporting different API version - that is, slightly different JSON objects, because of differences between API versions. A lazy (and very wasteful) solution would have been to build new JSON object for each subscription on each update, however that would cause a serious performance degradation. So instead, when sending publications to subscribed clients, I chose to expand on the existing (faster) solution, which is to build JSON once and then publish it to all clients. The trouble is, we need to send subtly different JSON depending on the API version supported by the client - the good news is that there are not that many supported API versions, and hopefully that number will stay low. This is where `MultiApiJson` comes in (technically, it is a specialization of `MultivarJson`). It is a very small (size equals to the number of valid API versions, 3 at this moment) array of distinct JSON objects, which are indexed by API version. All these JSON objects are created as a copy of a single JSON received in the class constructor - which should be the object that's closest to what we want to publish. After construction, each individual copy can be altered to accommodate variations between versions (this is best seen in the body of `NetworkOPsImp::transJson`), or all can be altered at the same time (there's a `set` method for this, which is not affected by this PR). The ownership and functions allowing for inspection and alteration of the individual JSON objects is the responsibility of `MultiApiJson`. This PR changes how we can inspect and alter these individual JSON objects.

#### Back to this PR

The purpose of this change is to enable the user of the `MultiApiJson` class to "visit" (that is, execute a piece of code, typically a lambda) a JSON object selected by API version. If the user wants to "visit" _all_ API versions, the `MultiApiJson` in this PR is smart enough to know what specific versions that means. The best example is in the `NetworkOPsImp::transJson` function, where the following:

```c++
    MultiApiJson multiObj{jvObj};
    visit<RPC::apiMinimumSupportedVersion, RPC::apiMaximumValidVersion>(
        multiObj,  //
        [&](Json::Value& jvTx, unsigned int apiVersion) {
            RPC::insertDeliverMax(
                jvTx[jss::transaction], transaction->getTxnType(), apiVersion);

            if (apiVersion > 1)
              // . . .
        });

```

has changed to:

```c++
    MultiApiJson multiObj{jvObj};
    forAllApiVersions(
        multiObj.visit(),  //
        [&](Json::Value& jvTx, auto v) {
            constexpr unsigned version = decltype(v)::value;
            RPC::insertDeliverMax(
                jvTx[jss::transaction], transaction->getTxnType(), version);

            if constexpr (version > 1)
              // . . .
        });

```

This came about because a very similar code was also needed in https://github.com/XRPLF/rippled/pull/4820 , inside function `NetworkOPsImp::pubValidation`, and of course, many instances of `forAllApiVersions` inside tests. Such uses are likely to increase as we add new API versions. Hence it is important to establish good practice which minimises potential for bugs.

The approach adopted in this PR is to:

- factor out API versions from `ripple/rpc/impl/RPCHelpers.h` to a new file `ripple/protocol/ApiVersion.h` 
- factor out a "policy" template parameter, used by `MultivarJson` for conversion from version to index in the array of JSON objects
- factor out `MultiApiJson` from `ripple/json/MultivarJson.h` to `ApiVersion.h`; this is where "policy" `ApiVersionHelper` (used to specialize `MultivarJson`) is defined
- redefine API version numbers as compile-time constants, using `std::integral_constant<unsigned, ...>` inside `ApiVersion.h`
- define new functions `forAllApiVersions` (removing equivalent from tests) and `forApiVersions` (similar to the previously defined `visit` template) inside `ApiVersion.h`
- replace `MultivarJson::select` with a more powerful `MultivarJson::visit` and a helper nested class `visitor_t`

The type change of API version from unsigned to `std::integral_constant<unsigned, ...>` has several consequences:
- it is now possible to validate during compilation the correctness of a hardcoded version number. Here's an example from `NetworkOPsImp::pubValidation`:
```c++
        multiObj.visit(
            RPC::apiVersion<1>,  // Invalid API version here will trigger compilation error
            [](Json::Value& jvTx) {
                // . . .
            });
```
- it is possible to pass a strongly typed API version to the user function, which in turn enables it to perform dispatch via overloading or `if constexpr` (instead of branching instruction), as seen in `NetworkOPsImp::transJson`
- few existing uses of the API version require an explicit cast to `unsigned` (in `ServerHandler.cpp` and `RPCHelpers.h`)
- since the API client connection handling does not create different types for requests for different API versions, we still need to support all uses of API version as `unsigned` (this could in time change, but the cost in terms of code churn is not worth it, at this time).
- `MultivarJson::visitor_t` has to support both variants of API version selection:
    - with a `std::integral_constant<unsigned, ...>` (which enforces version correctness via `static_assert`) 
    - with an `unsigned` (which enforces version correctness via normal `assert`) 

Additionally, the switch from `select` to `visit` reverses the calling convention used to publish JSON inside `NetworkOPs.cpp`, from:

```c++
	p->send(
	    multiObj.select(apiVersionSelector(p->getApiVersion())),
	    true);
```

to:

```c++
        multiObj.visit(
            p->getApiVersion(),
            [&](Json::Value const& jv) { p->send(jv, true); });
```

As for the complexity of change, the implementation of `MultivarJson::visitor_t` is probably the most challenging, mostly due to the number of defined overloads. The class itself is a stateless [niebloid](https://brevzin.github.io/c++/2020/12/19/cpo-niebloid/) which should make the reasoning about it slightly easier. The number of overloads inside this class also explains the size of unit tests, both `MultivarJson_test.cpp` and `ApiVersion_test.cpp`.

Additionally, the newly added `visit` function inside `MultivarJson` has two slightly different meanings depending on the parameters:
- the overload taking no parameters just returns the `visitor` object for the lazy execution (wrapped in a tiny lambda to provide the `*this` object); this object is next called by `forAllApiVersions` etc.
- the overload taking one or more parameters just calls the `visitor` directly, forwarding both `*this` and all the provided parameters, for the eager execution

I realise that this approach might be foreign to some readers, but I feel it is consistent with the paradigms of functional programming, where function evaluation is lazy when possible and eager when needed.

There is also a small drive-by fix in `LedgerToJson.cpp` where I removed the superfluous `(fill.context->apiVersion > 1)` conditional inside the scope of `else if (fill.context->apiVersion > 1)` section (starting at the line 140).

### Type of Change

- [ ] Documentation update
- [ ] Release

`Perf impact not expected`: This does not change any algorithms. The binary will be little different for `NetworkOPs` compared to before the change, but not in a way that would change anything perceptible. It gives the compiler slightly more space to apply optimisations when building JSON and when publishing but this is very minor.


thejohnfreeman:
Before I dive into the code, can you please share in an English comment (a) what problem this is meant to solve and (b) the general design / intuition of this solution?

intelliot:
@Bronek - John's request would be a good use for the `Context of Change` section of the PR template :)
```
### Context of Change

<!--
Please include the context of a change.
If a bug fix, when was the bug introduced? What was the behavior?
If a new feature, why was this architecture chosen? What were the alternatives?
If a refactor, how is this better than the previous implementation?

If there is a spec or design document for this feature, please link it here.
-->
```

Bronek:
> Before I dive into the code, can you please share in an English comment (a) what problem this is meant to solve and (b) the general design / intuition of this solution?

Thanks, done.

HowardHinnant (approved):

src/ripple/protocol/ApiVersion.h
```diff
@@ -0,0 +1,126 @@
...
+static_assert(apiMaximumValidVersion >= apiMaximumSupportedVersion);
+
+}  // namespace RPC
+
+template <unsigned minVer, unsigned maxVer, typename Fn, typename... Args>
+    requires(maxVer >= minVer) &&                   //
+    (minVer >= RPC::apiMinimumSupportedVersion) &&  //
+    (RPC::apiMaximumValidVersion >= maxVer) void forApiVersions(
+        Fn const& fn,
+        Args&&... args) requires requires()
```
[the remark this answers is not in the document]
Bronek: It's just an empty parameter declaration clause. Will remove it, I see how it can be confusing. Thanks !

src/ripple/json/MultivarJson.h
```diff
@@ -75,75 +88,149 @@ struct MultivarJson
...
-         ...);
+        return visitor(*this, std::forward<decltype(args)>(args)...);
     }
-    (std::make_index_sequence<size>{});
-}
+
+    void
+    visit(auto...) && = delete;
+    void
+    visit(auto...) const&& = delete;
```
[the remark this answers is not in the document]
Bronek: note, this is outdated. Now the `visit` functions are forwarding to `visitor` (also see the overview of change at the top of this PR)

> Additionally, the newly added visit function inside MultivarJson has two slightly different meanings depending on the parameters:

the overload taking no parameters just returns the visitor object for the lazy execution (wrapped in a tiny lambda to provide the *this object); this object is next called by forAllApiVersions etc.
the overload taking one or more parameters just calls the visitor directly, forwarding both *this and all the provided parameters, for the eager execution

src/ripple/protocol/ApiVersion.h
```diff
@@ -0,0 +1,132 @@
...
+             std::integral_constant<unsigned int, minVer + offset>{},
+             std::forward<Args>(args)...)),
+         ...);
+    }
+    (std::make_index_sequence<size>{});
+}
+
+template <typename Fn, typename... Args>
+void
+forAllApiVersions(Fn const& fn, Args&&... args) requires requires()
```
thejohnfreeman, line 97: If the requires expression parameter list is empty, what do you think of just omitting it? You do in some places but not here.

src/ripple/protocol/ApiVersion.h
```diff
@@ -0,0 +1,132 @@
...
+{
+    forApiVersions<
+        RPC::apiMinimumSupportedVersion,
+        RPC::apiMaximumValidVersion>(fn, std::forward<Args>(args)...);
+}
```
thejohnfreeman, line 102: What do these requires clauses buy us when they look exactly like the body?

src/ripple/protocol/ApiVersion.h
```diff
@@ -0,0 +1,132 @@
...
+    requires(maxVer >= minVer) &&                   //
+    (minVer >= RPC::apiMinimumSupportedVersion) &&  //
+    (RPC::apiMaximumValidVersion >=
+     maxVer) void forApiVersions(Fn const& fn, Args&&... args) requires requires
+{
+    fn(std::integral_constant<unsigned int, minVer>{},
+       std::forward<Args>(args)...);
+    fn(std::integral_constant<unsigned int, maxVer>{},
+       std::forward<Args>(args)...);
+}
```
thejohnfreeman, line 82: Style-wise, why split the requires clause both before and after the function declarator?

src/ripple/json/MultivarJson.h
```diff
@@ -75,75 +107,222 @@ struct MultivarJson
...
+                Args&&...> requires requires()
+        {
+            fn(json.val[index(Version)], version, std::forward<Args>(args)...);
+        }
+        {
+            static_assert(
+                valid(Version) && index(Version) >= 0 && index(Version) < size);
+            return fn(
+                json.val[index(Version)], version, std::forward<Args>(args)...);
+        }
```
thejohnfreeman, line 174: Can we collapse these definitions by moving the type of the first function parameter to the template parameter list?

src/ripple/json/MultivarJson.h
```diff
@@ -75,75 +107,222 @@ struct MultivarJson
...
+            Fn fn) const
+            -> std::invoke_result_t<Fn, Json::Value const&> requires requires()
+        {
+            fn(json.val[index(Version)]);
+        }
+        {
+            static_assert(
+                valid(Version) && index(Version) >= 0 && index(Version) < size);
+            return fn(json.val[index(Version)]);
+        }
```
thejohnfreeman, line 191: Can we collapse this definition with the above two by defining a custom invoker that does the right thing depending on the signature of `fn`?

thejohnfreeman (changes requested):

src/ripple/protocol/ApiVersion.h
```diff
@@ -0,0 +1,132 @@
...
+constexpr static auto apiInvalidVersion = apiVersion<0>;
+constexpr static auto apiVersionIfUnspecified = apiVersion<1>;
+constexpr static auto apiMinimumSupportedVersion = apiVersion<1>;
+constexpr static auto apiMaximumSupportedVersion = apiVersion<2>;
+constexpr static auto apiCommandLineVersion =
+    apiVersion<1>;  // TODO Bump to 2 later
+constexpr static auto apiBetaVersion = apiVersion<3>;
+constexpr static auto apiMaximumValidVersion = apiBetaVersion;
+
+static_assert(apiMinimumSupportedVersion >= apiVersionIfUnspecified);
```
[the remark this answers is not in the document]
Bronek: Then we change the assertion; at this moment, the precondition being enforced here is that `apiVersionUnspecified` must not overlap with any valid version. Unspecified is not the same as default (version 0 is not valid) and it requires special handling upon receipt of a customer request.

src/ripple/protocol/ApiVersion.h
```diff
@@ -0,0 +1,132 @@
...
+*/
+//==============================================================================
+
+#ifndef RIPPLE_PROTOCOL_APIVERSION_H_INCLUDED
+#define RIPPLE_PROTOCOL_APIVERSION_H_INCLUDED
+
+#include <ripple/json/MultivarJson.h>
+
+#include <functional>
+
```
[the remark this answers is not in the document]
Bronek: thank you, will do.

src/ripple/protocol/ApiVersion.h
```diff
@@ -0,0 +1,132 @@
...
+        return static_cast<std::size_t>(v <= 1 ? 0 : (v <= 2 ? 1 : 2));
+    }
+
+    constexpr static auto
+    valid(unsigned int v) noexcept -> bool
+    {
+        return v >= RPC::apiMinimumSupportedVersion &&
+            v <= RPC::apiMaximumValidVersion;
+    }
+};
```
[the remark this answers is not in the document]
Bronek: We can, but how do you propose we manage our internal dependencies (levelization) versions ? Do you suggest we should move the content of `MultivarJson.h` inside `ApiVersion.h` ? It certainly can be done.

src/ripple/protocol/ApiVersion.h
Bronek: they make the code SFINAE-friendly - i.e. it excludes the function from the overload set, rather than give you an outright compilation error. Admittedly I am using it perhaps too generously (a habit from the times when I was bitten hard by not using it enough).

src/ripple/json/MultivarJson.h
Bronek: Good idea, will do that. Will also remove `requires requires` , since the computation of the return type with `std::invoke_result_t` makes it redundant.

src/ripple/app/misc/NetworkOPs.cpp
```diff
@@ -3166,13 +3167,14 @@ NetworkOPsImp::transJson(
...
+    forAllApiVersions(
+        multiObj.visit(),  //
+        [&](Json::Value& jvTx, auto v) {
+            constexpr unsigned version = decltype(v)::value;
             RPC::insertDeliverMax(
-                jvTx[jss::transaction], transaction->getTxnType(), apiVersion);
+                jvTx[jss::transaction], transaction->getTxnType(), version);
 
-            if (apiVersion > 1)
+            if constexpr (version > 1)
```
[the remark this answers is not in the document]
Bronek: Ugh that would make things pretty complex elsewhere (because we can not longer rely on overloading alone), only to support this ugly syntax `template operator()<version>`

What if I simply replaced the lambda signature at line 3172 with this ?

```cpp
        [&]<unsigned Version>(Json::Value& jvTx, std::integral_constant<unsigned, Version>) {
```

src/ripple/json/MultivarJson.h
Bronek: I am not sure that's a good idea, complexity wise. It would save us 2 overloads (so we get to 2 total, one for integral_constant and another for other types) but at the cost of adding 2 extra overloads elsewhere (or an `if constexpr`) for the invoker and extra glue to make it work.

src/ripple/protocol/ApiVersion.h
Bronek: good point, will pull them all after the declarator (need `args`)

src/ripple/protocol/ApiVersion.h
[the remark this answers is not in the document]
thejohnfreeman: I'm missing something. Can't it overlap when `apiMinimumSupportedVersion == apiVersionIfUnspecified`, and isn't that the current state of affairs?

src/ripple/protocol/ApiVersion.h
thejohnfreeman: There are more. I just singled out this one.

src/ripple/protocol/ApiVersion.h
thejohnfreeman: There may be more. Please check.

src/ripple/protocol/ApiVersion.h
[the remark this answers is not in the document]
Bronek: I apologize - you are right and I did not read carefully. This check is wrong way round, I will fix it. Also I need a similar check for `apiCommandLineVersion`, and probably also for `apiInvalidVersion` (which is what I **thought** this check is against, when I wrote my reply above).

src/ripple/protocol/ApiVersion.h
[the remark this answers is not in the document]
Bronek: `ApiVersion.h` is in `protocol` and not in `rpc`, because pulling dependencies from `rpc` into `protocol` would violate levelization ; we need API versions for doing anything version specific inside `protocol` (as we do inside `NetworkOPs.cpp`, baring some serious refactoring).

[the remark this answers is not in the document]
Bronek: There was this dependency here https://github.com/XRPLF/rippled/pull/4833/files#diff-86ea7ba43fce0b564af71cb3f2070ce71cb27d9ec9b9fd93ab5298960a3b36dbL71 which I glad to have removed, would rather not bring anything like it back.

thejohnfreeman (approved):

intelliot:
@Bronek - any preference on whether this goes in 2.1 or 2.2? Given it is a refactor, I'm not sure if there is some urgency to getting it merged/released.

Bronek:
@intelliot After 2.1 release is fine.

Bronek:
from the coverage report:

> [src/ripple/app/misc/NetworkOPs.cpp](https://app.codecov.io/gh/XRPLF/rippled/pull/4833?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=XRPLF#diff-c3JjL3JpcHBsZS9hcHAvbWlzYy9OZXR3b3JrT1BzLmNwcA==) 
is referring to old tests of that same part, which does not have (and did not have) full coverage

> [src/ripple/protocol/MultiApiJson.h](https://app.codecov.io/gh/XRPLF/rippled/pull/4833?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=XRPLF#diff-c3JjL3JpcHBsZS9wcm90b2NvbC9NdWx0aUFwaUpzb24uaA==)
is referring to two `asserts` in new code which are impossible to test

> [src/ripple/app/ledger/impl/LedgerToJson.cpp](https://app.codecov.io/gh/XRPLF/rippled/pull/4833?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=XRPLF#diff-c3JjL3JpcHBsZS9hcHAvbGVkZ2VyL2ltcGwvTGVkZ2VyVG9Kc29uLmNwcA==)
is referring to old tests of that same part, which does not have (and did not have) full coverage

so basically these gaps are either unavoidable, or old.

PR #4838, opened 2023-11-27
XLS-46: Dynamic Non Fungible Tokens (dNFTs)

xVet:

This Amendment adds functionality to update the `URI` of [XLS-20 NFTs](https://github.com/XRPLF/XRPL-Standards/discussions/46) by adding:

- NFTokenModify: Transactor to initiate the altering of the `URI`
- lsfMutable: Flag to be set in a `NFTokenMint` transaction to allow `NFTokenModify` to execute successfully on given NFT.

### Context of Change

Changes and spec: [XLS-46](https://github.com/XRPLF/XRPL-Standards/discussions/130)

### Type of Change

- [ ] Documentation update
- [ ] Release

### API Impact

- [X] Public API: New feature (new methods and/or new fields)

## Future Tasks

Test case for:
- `NFTokenModify`
-  `lsfMutable` (new flag for `NFTokenMint`)


src/ripple/protocol/TxFlags.h
```diff
@@ -129,6 +129,7 @@ constexpr std::uint32_t const tfBurnable                   = 0x00000001;
 constexpr std::uint32_t const tfOnlyXRP                    = 0x00000002;
 constexpr std::uint32_t const tfTrustLine                  = 0x00000004;
 constexpr std::uint32_t const tfTransferable               = 0x00000008;
+constexpr std::uint32_t const tfMutable               = 0x00000016;
```
mvadari, line 132:
```suggestion
constexpr std::uint32_t const tfTransferable               = 0x00000008;
constexpr std::uint32_t const tfMutable                    = 0x00000016;
```

src/ripple/app/tx/impl/NFTokenModify.cpp
```diff
@@ -0,0 +1,96 @@
...
+#include <ripple/protocol/TxFlags.h>
+#include <ripple/protocol/st.h>
+
+namespace ripple {
+
+NotTEC
+NFTokenModify::preflight(PreflightContext const& ctx)
+{
+    if (!ctx.rules.enabled(featureNonFungibleTokensV1))
+        return temDISABLED;
```
mvadari, line 16: This needs to be guarded by a new amendment, not the existing NFT amendment.

src/ripple/protocol/TxFlags.h
```diff
@@ -147,7 +148,7 @@ constexpr std::uint32_t const tfNFTokenMintOldMask =
     ~(tfUniversal | tfBurnable | tfOnlyXRP | tfTrustLine | tfTransferable);
 
 constexpr std::uint32_t const tfNFTokenMintMask =
-    ~(tfUniversal | tfBurnable | tfOnlyXRP | tfTransferable);
+    ~(tfUniversal | tfBurnable | tfOnlyXRP | tfTransferable | tfMutable);
```
mvadari, line 151: The use of this flag in `NFTokenMint` needs to be guarded by an amendment

src/ripple/protocol/nft.h
```diff
@@ -54,6 +54,7 @@ constexpr std::uint16_t const flagBurnable = 0x0001;
 constexpr std::uint16_t const flagOnlyXRP = 0x0002;
 constexpr std::uint16_t const flagCreateTrustLines = 0x0004;
 constexpr std::uint16_t const flagTransferable = 0x0008;
+constexpr std::uint16_t const flagMutable = 0x0016;
```
mvadari, line 57: The number is in hex
```suggestion
constexpr std::uint16_t const flagMutable = 0x0010;
```

xVet:
close/reopen (comment will be edited with link)

xVet:
Tests are for the next commits. Tests, guards and code related to tfTrustLine/fixRemoveNFTokenAutoTrustLine are removed in my draft as well

src/ripple/protocol/nft.h
LetsEffingGo: It seems the flag numbers increment by a power of 2 (in hex). Is this the case, and if so, what's the reasoning for it?

src/ripple/protocol/TxFlags.h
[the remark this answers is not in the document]
mvadari: This is unnecessary - all the other transactions that don't have any flags just use the `tfUniversalMask` variable instead of creating a new one.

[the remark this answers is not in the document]
shawnxie999: True, unless new flags are looking to be added in the future 🤷

src/ripple/protocol/Feature.h
```diff
@@ -352,6 +352,8 @@ extern uint256 const featureXChainBridge;
 extern uint256 const fixDisallowIncomingV1;
 extern uint256 const featureDID;
 extern uint256 const fixFillOrKill;
+extern uint256 const featureDNFT;
```
mvadari, line 355: nit: I would call this `featureDynamicNFT` to be a bit more descriptive

src/ripple/protocol/TxFlags.h
```diff
@@ -127,27 +127,14 @@ constexpr std::uint32_t tfPayChanClaimMask = ~(tfUniversal | tfRenew | tfClose);
 // NFTokenMint flags:
 constexpr std::uint32_t const tfBurnable                   = 0x00000001;
 constexpr std::uint32_t const tfOnlyXRP                    = 0x00000002;
-constexpr std::uint32_t const tfTrustLine                  = 0x00000004;
 constexpr std::uint32_t const tfTransferable               = 0x00000008;
+constexpr std::uint32_t const tfMutable                    = 0x00000016;
```
mvadari, line 131: Flags are bitmasks, they should only be one bit and the number is already in hex.
```suggestion
constexpr std::uint32_t const tfMutable                    = 0x00000010;
```

src/ripple/app/tx/impl/NFTokenModify.cpp
```diff
@@ -0,0 +1,96 @@
...
+    return tesSUCCESS;
+}
+
+TER
+NFTokenModify::doApply()
+{
+    auto const nftokenID = ctx_.tx[sfNFTokenID];
+    auto const account = ctx_.tx[sfAccount];
+
+
```
mvadari, line 78: nit:
```suggestion

```

src/ripple/app/tx/impl/NFTokenModify.cpp
```diff
@@ -0,0 +1,96 @@
...
+        tokenAndPage->token[sfURI] = newURI; 
+    }
+
+    // Apply the changes to the token
+    if (auto const ret = nft::updateToken(view(), account, std::move(tokenAndPage->token), std::move(tokenAndPage->page)); !isTesSuccess(ret))
+        return ret;
+
+    return tesSUCCESS;    
+}
+}
```
mvadari, line 96: nit: files should end with a newline

src/ripple/app/tx/impl/NFTokenModify.h
```diff
@@ -0,0 +1,31 @@
+
```
mvadari, line 1: This needs the standard license/disclaimer text

src/ripple/app/tx/impl/NFTokenModify.cpp
```diff
@@ -0,0 +1,96 @@
...
+        return temDISABLED;
+
+    if (auto const ret = preflight1(ctx); !isTesSuccess(ret))
+        return ret;
+
+    auto const nftID = ctx.tx[sfNFTokenID];
+    auto const account = ctx.tx[sfAccount];
+
+    if (!account || !nftID)
+       return temMALFORMED;
```
mvadari, line 25: This check is unneeded - it'll be checked by `TxFormats`

src/ripple/app/tx/impl/NFTokenModify.cpp
```diff
@@ -0,0 +1,96 @@
+
+#include <ripple/app/tx/impl/NFTokenModify.h>
```
mvadari, line 2: This new transaction type needs to be added to `ripple/protocol/TxFormats.h` and `ripple/protocol/impl/TxFormats.cpp` and `ripple/app/tx/impl/applySteps.cpp` - otherwise rippled doesn't know it exists

shawnxie999:
I would suggest add a validation in `NFTokenModify`, where it returns failure if the NFT still has outstanding buy and sell offers.

Otherwise, I see two problems:

1.  If the owner changes the URI and someone else accepts a current offer around the same time, the buyer may not realize that the URI has been updated, resulting in buying a NFT that they may not have want. I think we would like to avoid this confusion and potential bad act at the protocol layer.
2. From a technical perspective, the Clio ETL does not handle a edge case very well - if a `NFTokenModify` and `NFTokenAcceptOffer` occur at the the _same ledger_, Clio will result in data loss. This a design with the clio that hasn't considered the possibility where the URI change and a transfer of ownership can happen at the same ledger. Trying to solve this problem is rather messy to do in the Clio end, so I would prefer not go that route. 

So I think adding a validation to check for offers, even though it will be some extra work as the owner would need to cancel all the offers, but that's reasonable.

Any thoughts?

shawnxie999:
> Great point! I agree it makes totally sense to first check if offers are available and if so, to cancel them all ? Would you say, if offers available then let `NFTokenModify` fail or rather cancel all offers and then execute? Basically guaranteeing a successful execution either way.
> 
> Thanks for the noticing this edge scenario Shawn 🤗

Both are possible solutions, I don't really have a strong opinion.

To summarize the two options:

1. `NFTokenModify` fails if the NFT still has offers, and does not cancel them. This should be pretty straightforward, all we need is checking if the sell and buy offer directories are empty or not.

2. `NFTokenModify` implicitly cancels all the existing offers. However, we need to make sure the the number of existing offers is within a threshold(500), so that the metadata of the transaction won't be too huge. The code to implement this is very similar to `NFTokenBurn` before the activation of the amendment `fixNonFungibleTokensV1_2`. In preclaim of NFTokenBurn.cpp, there is a [line of code ](https://github.com/XRPLF/rippled/blob/46f3d3ef610414b48321ded1d4900b56b2bce116/src/ripple/app/tx/impl/NFTokenBurn.cpp#L84):
```
    if (!ctx.view.rules().enabled(fixNonFungibleTokensV1_2))
    {
        // If there are too many offers, then burning the token would produce
        // too much metadata.  Disallow burning a token with too many offers.
        return nft::notTooManyOffers(ctx.view, ctx.tx[sfNFTokenID]);
    }
```
This code will return error if the number of offers exceeds 500, and therefore would require the owner to manually cancel them.

src/ripple/app/tx/impl/NFTokenModify.cpp
```diff
@@ -0,0 +1,96 @@
...
+    // Find the token and its page
+    auto tokenAndPage = nft::findTokenAndPage(view(), account, nftokenID);
+    if (!tokenAndPage)
+        return tecINTERNAL;
+
+    // Replace the URI if present in the transaction
+    if (ctx_.tx.isFieldPresent(sfURI)) {
+        auto newURI = ctx_.tx[sfURI];
+        tokenAndPage->token[sfURI] = newURI; 
+    }
```
tequdev, line 88: Since the URI is a required field in the Standard, this condition is always true if it is followed.

If the URI can be deleted, the URI field should be an optional field.

I think it would be appropriate to change the URI field in Standard to an optional field.
And overwrite the URI if it exists in the transaction and delete the URI if it does not.

dangell7:
> > But there is still a concern, what if, in the same ledger, the owner submits a `NFTokenModify` at the same time someone else creates a buy offer? If that can happen, then the buyer submits a buy offer without knowing the URI changed, and the offer wasn't cancelled because `NFTokenModify` and `NFTokenCreateOffer` occured in the same ledger (if `NFTokenModify` is executed before `NFTokenCreateOffer` in that ledger)
> > I'm not sure of this either and I think we may want to double check this edge case in unit test if it's something we would care about.
> 
> To resolve this concern. I would propose an update to the spec, such that the owner can only submit a `NFTokenModify` transaction if the the NFT owner account has set `lsfDisallowIncomingNFTokenOffer`, to block any incoming offers until the URI has changed. But another concern of this approach is that it would block incoming offers for all NFTs, which may not be desirable.

I think all of that is irrelevant as the issuer can just update the NFToken after the sale. Wether the change happens before the sale or after is irrelevant imo. 

Personally I wouldn't do anything with the offers. If clio cant handle this then thats a design flaw on clio and no one should be limiting themselves because clio cant do something.

shawnxie999:
> I think all of that is irrelevant as the issuer can just update the NFToken after the sale. Wether the change happens before the sale or after is irrelevant imo.
> 
> Personally I wouldn't do anything with the offers. If clio cant handle this then thats a design flaw on clio and no one should be limiting themselves because clio cant do something.

This actually has nothing to do with clio, but yes it's only a minor detail. If everyone feels good about it, then I don't have any concern.

mvadari:
Closing this in favor of #5048

PR #4840, opened 2023-11-27
Proposed 2.0.0-rc3

manojsdoshi:


intelliot (approved):
We merge version number bumps (like this one) with a fast-forward merge (do not use the GitHub UI) 👍

PR #4841, opened 2023-11-29
Proposed 2.0.0-rc4

intelliot:
Includes the two API fixes that we've been discussing.

Due to the lack of "2 weeks" of test time, @manojsdoshi has the right to reject this release candidate in favor of 2.0.0-rc3. However, these changes only impact the RPC API and are considered to be well-reviewed and low-risk bug fixes.

### DIFF
--- a/src/ripple/app/ledger/impl/LedgerToJson.cpp
+++ b/src/ripple/app/ledger/impl/LedgerToJson.cpp
@@ -27,6 +27,7 @@
 #include <ripple/protocol/jss.h>
 #include <ripple/rpc/Context.h>
 #include <ripple/rpc/DeliveredAmount.h>
+#include <ripple/rpc/impl/RPCHelpers.h>
 
 namespace ripple {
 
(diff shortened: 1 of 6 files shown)


Bronek (approved):

src/ripple/app/misc/NetworkOPs.cpp
```diff
@@ -2220,12 +2222,28 @@ NetworkOPsImp::pubValidation(std::shared_ptr<STValidation> const& val)
             reserveIncXRP && reserveIncXRP->native())
             jvObj[jss::reserve_inc] = reserveIncXRP->xrp().jsonClipped();
 
+        // NOTE Use MultiApiJson to publish two slightly different JSON objects
+        // for consumers supporting different API versions
+        MultiApiJson multiObj{jvObj};
+        visit<RPC::apiMinimumSupportedVersion, RPC::apiMaximumValidVersion>(
```
Bronek, line 2228: This PR is meant to make the change trivial switching between API version 3 and version 2, which is why it looks like this. If this is indeed released for API version 2 then I have another PR lined up to clean this up, so we simply modify one version of `Json::Value` rather than `visit` here - i.e. simplify accordingly

src/ripple/app/ledger/impl/LedgerToJson.cpp
```diff
@@ -159,7 +167,10 @@ fillJsonTx(
         txJson[jss::validated] = validated;
         if (validated)
         {
-            txJson[jss::ledger_index] = to_string(fill.ledger.seq());
+            auto const seq = fill.ledger.seq();
+            txJson[jss::ledger_index] = (fill.context->apiVersion > 1)
```
Bronek, line 171: This PR is meant to make the change trivial switching between API version 3 and version 2, which is why it looks like this. If this is indeed released for API version 2 then I have another PR lined up to remove redundant conditional expression here and instead unconditionally set `ledger_index` to a number (since this is inside `else if (fill.context->apiVersion > 1)` block starting at line 140)

ximinez (commented):
:shipit:

PR #4845, opened 2023-12-05
XLS-52d: NFTokenMintOffer

tequdev:

Fixed to allow creating NFTokenOffer in a NFTokenMint transaction.

Added 3 optional fields to NFTokenMint transaction:
- `Amount`
- `Destination`
- `Expiration`

### Context of Change

- https://github.com/XRPLF/XRPL-Standards/discussions/147

### Type of Change


shawnxie999:
Is this ready to review?

src/ripple/app/tx/impl/NFTokenMint.cpp
```diff
@@ -80,6 +86,40 @@ NFTokenMint::preflight(PreflightContext const& ctx)
...
+        {
+            if (ctx.tx.isFlag(tfOnlyXRP))
+                return temBAD_AMOUNT;
+
+            if (!*amount)
+                return temBAD_AMOUNT;
+        }
+    }
+
+    if (auto exp = ctx.tx[~sfExpiration]; exp == 0)
```
shawnxie999, line 113: `exp` should be const

src/ripple/app/tx/impl/NFTokenMint.cpp
```diff
@@ -80,6 +86,40 @@ NFTokenMint::preflight(PreflightContext const& ctx)
...
+        return temBAD_EXPIRATION;
+
+    if (auto dest = ctx.tx[~sfDestination])
```
shawnxie999, line 116: `dest` should be const

src/ripple/app/tx/impl/NFTokenMint.cpp
```diff
@@ -146,6 +186,54 @@ NFTokenMint::preclaim(PreclaimContext const& ctx)
             return tecNO_PERMISSION;
     }
 
+    if (auto const amount = ctx.tx[~sfAmount])
+    {
+        if (hasExpired(ctx.view, ctx.tx[~sfExpiration]))
+            return tecEXPIRED;
+
+        if (!(ctx.tx.getFlags() & nft::flagCreateTrustLines) &&
```
shawnxie999, line 194: I don't think this check is necessary because the preflight already checked `NFTokenMintMask` to exclude the old `flagCreateTrustLines` flag

shawnxie999 (commented):
initial set of minor comments...

src/ripple/app/tx/impl/NFTokenMint.cpp
```diff
@@ -260,6 +346,62 @@ NFTokenMint::doApply()
...
+                [&nftokenID](std::shared_ptr<SLE> const& sle) {
+                    (*sle)[sfFlags] = lsfNFTokenSellOffers;
+                    (*sle)[sfNFTokenID] = nftokenID;
+                });
+
+            if (!offerNode)
+                return tecDIR_FULL;
+
+            // Only sell offer are supported at NFTokenMint.
+            std::uint32_t sleFlags = lsfSellNFToken;
```
shawnxie999, line 379: const

src/test/app/NFToken_test.cpp
```diff
@@ -6566,6 +6586,236 @@ class NFTokenBaseUtil_test : public beast::unit_test::suite
...
+                ter(temDISABLED));
+            env.close();
+
+            env(token::mint(alice),
+                token::expiration(lastClose(env) + 25),
+                ter(temDISABLED));
+            env.close();
+        }
+
+        if (features[featureNFTokenMintOffer])
```
shawnxie999, line 6621: We may also want to introduce a small test that checks the reserve requirements if the issuer has enough reserve to own the new NFT and the offer. You can refer to `testFixNFTokenBuyerReserve` test (but doesn't need to be as detailed as this one though)

shawnxie999 (approved):
Looks good.

src/ripple/app/tx/impl/NFTokenMint.cpp
```diff
@@ -146,6 +186,54 @@ NFTokenMint::preclaim(PreclaimContext const& ctx)
...
+        if (hasExpired(ctx.view, ctx.tx[~sfExpiration]))
+            return tecEXPIRED;
+
+        if (!(ctx.tx.getFlags() & nft::flagCreateTrustLines) &&
+            !amount->native() && ctx.tx[~sfTransferFee])
+        {
+            auto const nftIssuer =
+                ctx.tx[~sfIssuer].value_or(ctx.tx[sfAccount]);
+
+            if (!ctx.view.exists(keylet::line(nftIssuer, amount->issue())))
```
shawnxie999, line 200: Due to a bug would cause failure if the NFT issuer is also the issuer of the IOU (if transferFee is present), this amendment would also need to this change

```
            if (issuer != amount.getIssuer() &&
                !ctx.view.read(keylet::line(issuer, amount.issue())))
                return tecNO_LINE;
```

This is fixed in another amendment [here](https://github.com/XRPLF/rippled/pull/4946/files). We would need to add test case for this as well.

shawnxie999 (changes requested):
There is an outstanding bug in `NFTokenCreateOffer` that we would need to fix in this amendment as well

src/ripple/app/tx/impl/NFTokenMint.cpp
tequdev: This Amendment has not yet been merged, so the code in `ctx.view.rules().enabled(fixNFTokenTrustlineSurprise)` cannot be used.

How can I do this?
Should this be the default process without `ctx.view.rules().enabled(fixNFTokenTrustlineSurprise)`?

shawnxie999: I don't think there is the need to check for the fixNFTokenTrustlineSurprise amendment at all.

shawnxie999 (approved):
Thanks! Looks good!

scottschurr:
For the most part, this is a well done pull request.  The tests seem to have good coverage of the new code.

I have one serious concern with the implementation.  A bunch of the code from NFTokenCreateOffer.cpp has been copied into NFTokenMint.  That means when one part of the code changes for maintenance the copy must change as well.  Inevitably the two copies will fall out of sync.  That's not an immediate problem, but is absolutely a long term problem.

One way in which this problem is blindingly apparent is when @shawnxie999 asked that you include code for an amendment that is in review.  Optimally that fix should not need to be coded in two different places.

One way to address this concern is to share the code, rather than copy it.  If the common code is shared, for example by calling a shared routine, then future maintenance will still only need to happen in one place: in the shared code.

I wanted to make sure this was possible before I suggested this change.  So I have an example of how this code sharing might be done.  The top-most commit here shows one way it might be implementated: https://github.com/scottschurr/rippled/commits/tequ-featureNFTokeMintOffer/  You are welcome to cherry-pick that commit if you want.  But also feel free to take a different path to a similar solution.

If you do cherry-pick that commit then I should probably recuse myself from the remainder of this code review, since I would be blind to mistakes that I might have introduced.

I hope that helps.

tequdev:
Thanks, @scottschurr!
I will check your commit and cherry pick!

scottschurr:
@shawnxie999, the most recent commit introduced significant changes.  It would be good if you could re-review this pull request.  Thanks.

src/ripple/app/tx/impl/NFTokenMint.cpp
```diff
@@ -31,16 +31,23 @@
 
 namespace ripple {
 
+static std::uint16_t
+extractNFTokenFlagsFromTxFlags(std::uint32_t txFlags)
```
shawnxie999, line 35: const?

src/test/app/NFToken_test.cpp
```diff
@@ -7417,6 +7416,164 @@ class NFTokenBaseUtil_test : public beast::unit_test::suite
         }
     }
 
+    void
+    testNFTIssuerIsIOUIssuer(FeatureBitset features)
```
shawnxie999, line 7420: @scottschurr  isn't this test redundant with the tests you introduced in fixEnforceNFTokenTrustline?

scottschurr: Yes, they are.  We'll have to see which amendment gets into develop first.  If this one goes in first (which seems likely) then I'll need to remove the nearly identical tests from the `fixEnforceNFTokenTrustline` pull request.

Good spotting.

src/ripple/app/tx/impl/NFTokenMint.cpp
scottschurr: It could be `const` with absolutely no harm.  Typically literals like this are passed as non-`const` since no harm will come to the calling function if they get modified in the called function.  Since it is a literal, not a reference, this is a copy.  In C++ it is usually cheaper compute-wise to pass small objects, like a `uint32_t`, by value.

So, I'm happy to add the `const` if you would like.  But I don't think it increases safety in this particular case.

shawnxie999: cool! i agree we we don't need change here 👍

shawnxie999 (approved):

scottschurr:
I think this pull request looks good, but my view is colored by my own changes to the pull request.  So I'm removing myself as a reviewer.  @HowardHinnant has agreed to review this pull request in my place, although he's tied up at the moment.  He can hopefully get to the review in the next week or two.  Sorry for the additional delay.

HowardHinnant:
Needs rebasing against current develop to resolve conflicts.  Doing so will also help me see how this PR would change develop.

src/ripple/app/tx/impl/NFTokenMint.cpp
```diff
@@ -197,7 +197,10 @@ NFTokenMint::preclaim(PreclaimContext const& ctx)
             auto const nftIssuer =
                 ctx.tx[~sfIssuer].value_or(ctx.tx[sfAccount]);
 
-            if (!ctx.view.exists(keylet::line(nftIssuer, amount->issue())))
+            // If the IOU issuer and the NFToken issuer are the same,
```
HowardHinnant: The commit message should be prefixed with `[FOLD]` to indicate that this commit should be merged (likely with the previous commit).

HowardHinnant (approved):

PR #4848, opened 2023-12-07
Websocket should only call async_close once

ximinez:

Prevents Websocket connections from trying to close twice.

### Context of Change

While testing #4822, @lathanbritz discovered that `rippled` would assert under heavy websocket client load. Later testing determined that this issue was unrelated to that PR, and in fact, exists at least as far back as 1.12.0. In fact, the relevant code appears to be 7 years old. It's entirely possible that the code was fine as originally written, and changes in the underlying Boost library uncovered it.

The issue only occurs in debug builds (assertions are disabled in release builds, including published packages), and when the websocket connections are unprivileged.

More specifically, the issue seems to occur when a client drives up the resource balance enough to be forcibly disconnected while there are still messages pending to be sent. Relevant log messages will look like:
```
2023-Dec-07 00:02:21.624011245 UTC Resource:WRN Consumer entry 127.0.0.1 dropped with balance 16368 at or above drop threshold 15000
```

### Type of Change

## Test Plan

Running the following python script several times in parallel reliably recreates the issue, usually within a few seconds on a debug build from a non-admin host.
```
from datetime import datetime
from websocket import create_connection
import sys
# ws = create_connection('ws://badger-w:51233')
ws = create_connection('ws://localhost:51233')
acct="<insert any r-address here>"

while True:
  ws.send('{ "command" : "path_find", "destination_account" : "' + acct + '", "destination_amount" : "10000000", "id" : "example", "source_account" : "' + acct + '", "subcommand" : "create" }')

  ws.send('{ "command" : "ledger", "transactions" : true, "expand" : true }')

  # resp=ws.recv()
  # print(resp)

  ws.send('{ "command" : "ping" }')

  resp=ws.recv()
  print(resp)

  resp=ws.recv()
  print(resp)

  ws.send('{ "command" : "path_find", "subcommand" : "close" }')

  resp=ws.recv()
  print(resp)

  resp=ws.recv()
  print(resp)

  print(datetime.now())

```


src/ripple/server/impl/BaseWSPeer.h
```diff
@@ -348,6 +356,8 @@ BaseWSPeer<Handler, Impl>::on_write_fin(error_code const& ec)
         return fail(ec, "write_fin");
     wq_.pop_front();
     if (do_close_)
+    {
```
gregtatcam, line 359: Should `do_close_` be reset to `false` here? What if `on_write_fin` is called again?

gregtatcam (approved):
👍 LGTM

scottschurr (approved):
:+1:

src/ripple/server/impl/BaseWSPeer.h
ximinez: Thanks for the feedback. Coincidentally, I had the realization about not needing `closing_` while having dinner, and came in to fix it before I even saw your comments. :grinning:

intelliot:
Suggested commit message:
```
WebSocket should only call async_close once (#4848)

Prevent WebSocket connections from trying to close twice.

The issue only occurs in debug builds (assertions are disabled in
release builds, including published packages), and when the WebSocket
connections are unprivileged. The assert (and WRN log) occurs when a
client drives up the resource balance enough to be forcibly disconnected
while there are still messages pending to be sent.

Thanks to @lathanbritz for discovering this issue in #4822.
```

intelliot:
note: 694fc8a is mistaken and would need review

ximinez:
Branch is reverted back to the version that was approved by reviewers, and merged with latest `develop`.

SaxenaKaustubh:
LGTM! QA Approved.

PR #4849, opened 2023-12-08
Improve coverage reporting 

Bronek:

Currently coverage reports take a very long time to generate. This can be improved by using multiple cores, with tools which know how to use them. Also add codecov integration.

Newly added `CodeCoverage.cmake` has been copied from [cmake-modules](https://github.com/bilke/cmake-modules/blob/master/CodeCoverage.cmake), then updated.

**NOTE** This PR should only be merged by someone with administrative access to the repo, since it makes use of `secrets` in `.github` workflow. A commit (merge) made by someone without access to repo secrets would result in this change not having this access either, which will lead to [spurious failures of codecov integration](https://community.codecov.com/t/upload-issues-unable-to-locate-build-via-github-actions-api/3954)

### Type of Change

- [ ] Documentation update
- [ ] Release


Bronek:
I plan to rewrite all the commits, so they are all signed. The overall diff won't change. (_this comment will be deleted when I'm done_).

.github/actions/build/action.yml
```diff
@@ -26,4 +30,9 @@ runs:
         cmake \
           --build ${build_dir} \
           --config ${{ inputs.configuration }} \
-          --parallel ${NUM_PROCESSORS:-$(nproc)}
+          --parallel ${NUM_PROCESSORS:-$(nproc)} \
+          --target ${{ inputs.cmake-target }}
+        [[ -n "${{ inputs.artifact }}" ]] \
+          && [[ -e "${build_dir}/${{ inputs.artifact }}" ]] \
+          && mv "${build_dir}/${{ inputs.artifact }}" ./ \
```
thejohnfreeman, line 37: This condition has three terms. Why would the second term ever be false when the first term true?

BUILD.md
```diff
@@ -262,17 +262,75 @@ It patches their CMake to correctly import its dependencies.
...
 | --- | ---| ---|
 | `assert` | OFF | Enable assertions.
 | `reporting` | OFF | Build the reporting mode feature. |
+| `coverage` | OFF | Prepare the coverage report. |
 | `tests` | ON | Build tests. |
 | `unity` | ON | Configure a unity build. |
 | `san` | N/A | Enable a sanitizer with Clang. Choices are `thread` and `address`. |
 
-[Unity builds][5] may be faster for the first build
+[Unity builds][7] may be faster for the first build
```
thejohnfreeman, line 333: Existing links should not be relabeled when a new one is added. Not all numbers need to be used. Just add the new link at an unused number. Otherwise the diff gets these frivolous changes.

Builds/CMake/RippledCov.cmake
```diff
@@ -2,97 +2,39 @@
...
-    set (extract_pattern "")
-    if (coverage_core_only)
-      set (extract_pattern "${CMAKE_CURRENT_SOURCE_DIR}/src/ripple/")
-    endif ()
+if (coverage)
+  if (DEFINED CODE_COVERAGE_TEST_PARALLELISM)
+    set(TEST_PARALLELISM ${CODE_COVERAGE_TEST_PARALLELISM})
+  else()
+    set(TEST_PARALLELISM ${PROCESSOR_COUNT})
+  endif()
```
thejohnfreeman, line 12: What is the difference between `TEST_PARALLELISM` and `CODE_COVERAGE_TEST_PARALLELISM`? If you're just setting a default value here, the best practice is to use `option()`.

CMakeLists.txt
```diff
@@ -28,6 +28,12 @@ if(Git_FOUND)
     endif()
 endif() #git
 
+include (ProcessorCount)
+if (NOT CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
+  include (CodeCoverage)
+  include (RippledCov)
```
thejohnfreeman, line 34: Why aren't these includes conditional on the `coverage` option?

.github/actions/build/action.yml
Bronek: in case of action being misconfigured. You make a good point, though - if the action is misconfigured, it will be preferable to fail, rather than silently skip moving the artifact. I will change.

BUILD.md
Bronek: Ok thanks, will change that. I assumed the reader would prefer the links to be in order

Builds/CMake/RippledCov.cmake
Bronek: I need an actual default value, will use `set (... CACHE ...)`

CMakeLists.txt
Bronek: thanks, I forgot to clean up this part.

.github/actions/build/action.yml
thejohnfreeman: I'm wondering: why not just add a separate step, in the jobs that use this option, to move the artifact? It seems like the coverage report job doesn't even need to move the report to upload it. Can we skip moving it altogether and remove that step from this action?

Builds/CMake/RippledCov.cmake
```diff
@@ -2,26 +2,10 @@
    coverage report target
 #]===================================================================]
 
-ProcessorCount(PROCESSOR_COUNT)
-
 if (coverage)
```
thejohnfreeman, line 5: We don't even include this file unless `coverage=ON` right? Can we just assert that `coverage=ON` here and de-dent the rest of the file?

CMakeLists.txt
```diff
@@ -131,6 +127,11 @@ if(reporting)
   )
 endif()
 
+if (coverage)
+  include (CodeCoverage)
+  include (RippledCov)
+endif()
```
thejohnfreeman, line 133: No spaces between function name and parentheses, please. This style is not yet consistent throughout our CMake, but it is the mandatory style for all future changes (matching CMake's documentation). Please fix the other changes in this PR too.

Builds/CMake/RippledCov.cmake
```diff
@@ -2,26 +2,10 @@
...
-
-  if (DEFINED CODE_COVERAGE_REPORT_FORMAT)
-    set(CODE_COVERAGE_FORMAT ${CODE_COVERAGE_REPORT_FORMAT})
-  else()
-    set(CODE_COVERAGE_FORMAT html-details)
-  endif()
-
-  if (DEFINED CODE_COVERAGE_EXTRA_ARGS)
-    set(GCOVR_ADDITIONAL_ARGS ${CODE_COVERAGE_EXTRA_ARGS})
+  set(GCOVR_ADDITIONAL_ARGS ${coverage_extra_args})
```
thejohnfreeman, line 6: Can we add a comment here like the following, assuming it is correct?

```
# The instructions for these commands come from the `CodeCoverage` module,
# which was copied from https://github.com/bilke/cmake-modules,
# commit <sha1>.
```

I think this file should hold the command for `include(CodeCoverage)` too.

BUILD.md
```diff
@@ -262,12 +262,72 @@ It patches their CMake to correctly import its dependencies.
    generator. Pass `--help` to see the rest of the command line options.
 
 
+## Coverage report
+
+Coverage report is intended for the developers using compilers GCC
```
thejohnfreeman, line 267:
```suggestion
The coverage report is intended for the developers using compilers GCC
```

BUILD.md
```diff
@@ -262,12 +262,72 @@ It patches their CMake to correctly import its dependencies.
...
+or Clang (including Apple Clang). It is generated by the build target `coverage_report`,
```
thejohnfreeman, line 268: Can we just rename the target to `coverage`?

BUILD.md
thejohnfreeman: The are many other instances in this section where the definite article is omitted. I think it'd be good to have a technical writer edit the whole section for grammar and style.

CMakeLists.txt
```diff
@@ -28,6 +28,8 @@ if(Git_FOUND)
     endif()
 endif() #git
 
+include (ProcessorCount)
```
thejohnfreeman, line 31: This `include` should be moved to the file that calls `ProcessorCount`.

thejohnfreeman (changes requested):

Builds/CMake/RippledCov.cmake
Bronek: will do, thanks

Builds/CMake/RippledCov.cmake
Bronek: there's a significant number of local changes, I will point this out.

CMakeLists.txt
Bronek: there are actually two such files - `RippledSettings.cmake` and `RippledSanity.cmake`

CMakeLists.txt
Bronek: Will fix that, thanks

BUILD.md
Bronek: Will do that, thanks

.github/actions/build/action.yml
Bronek: will do that, simpler is better

CMakeLists.txt
thejohnfreeman: They should both include it. CMake modules have include guards when necessary just like C++ headers, and the best practice is the same: include what you use.

thejohnfreeman (approved):

intelliot:
note: awaiting review by @ximinez

ximinez (approved):
Obviously, this doesn't work in Windows / Visual Studio, because the tools are missing. The good news is that it doesn't blow up or otherwise fail.

Do we want it to fail?

intelliot:
I think that's a question for @Bronek ? 

(I don't personally have an opinion on that yet.)

Bronek:
> Obviously, this doesn't work in Windows / Visual Studio, because the tools are missing. The good news is that it doesn't blow up or otherwise fail.

Thanks for pointing it out. We should not enable coverage target on Windows, and warn if the `coverage` flag was enabled. Added a commit to this effect now.

CMakeLists.txt
```diff
@@ -126,12 +125,20 @@ if(reporting)
   )
 endif()
 
+if(coverage)
+  if(CMAKE_CXX_COMPILER_ID MATCHES "MSVC")
+    message(WARNING "Code coverage on Windows is not supported, ignoring 'coverage' flag")
+  else()
+    include(CodeCoverage)
```
thejohnfreeman, line 132: I think it would be good to move this include into `RippledCov.cmake`.

Bronek: Thank you, done :+1:

intelliot:
Let's get at least 1 quick re-review of the latest diff, and confirmation from @Bronek that this is ready to merge (from his perspective).

ximinez (approved):

thejohnfreeman (approved):
Two small nits that aren't blockers. I want to echo my request for no force-pushes in PRs. They completely break the feature to "review changes since your last review".

Builds/CMake/RippledSettings.cmake
```diff
@@ -17,15 +21,17 @@ endif ()
 if (is_gcc OR is_clang)
   option (coverage "Generates coverage info." OFF)
   option (profile "Add profiling flags" OFF)
+  set (coverage_test_parallelism "${PROCESSOR_COUNT}" CACHE STRING
+    "Unit tests parallelism for the purpose of coverage report.")
+  set (coverage_format "html-details" CACHE STRING
+    "Output format of the coverage report.")
+  set (coverage_extra_args "" CACHE STRING
```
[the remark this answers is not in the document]
Bronek: thanks, I added commit to reformat this file so it is internally consistent.

CMakeLists.txt
```diff
@@ -126,12 +125,16 @@ if(reporting)
   )
 endif()
 
+if(coverage)
+  include(CodeCoverage)
```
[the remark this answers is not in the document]
Bronek: thanks, I moved this include inside `RippledCov.cmake`

Builds/CMake/RippledCov.cmake
```diff
@@ -9,6 +9,8 @@ endif()
 if(CMAKE_CXX_COMPILER_ID MATCHES "MSVC")
   message(WARNING "Code coverage on Windows is not supported, ignoring 'coverage' flag")
 else()
```
thejohnfreeman, line 11: Just end this branch with `return()` and de-dent the remainder of the file.

thejohnfreeman (changes requested):

thejohnfreeman (approved):

Bronek:
this is good to merge (pls. squash when merging)

PR #4856, opened 2023-12-21
[Trivial] Remove unused constants from resource/Fees.h

scottschurr:

While reviewing other code I stumbled across a number of constants defined in resource/Fees.h that are unused anywhere in the code base.  They did no harm, but it seems untidy to leave them lying around.  This commit removes the unused constants.

I also tidied up some messiness left by clang-format.

### Type of Change


scottschurr:
This pull request had become particularly stale.  Since there have been no active reviews yet, I rebased the pull request to the latest develop.

ximinez (approved):
I was surprised to see that none of the removed variables are defined in Fees.cpp. It's not just that they aren't used in code, they're not even defined, so if anyone did try to use one, they'd get a linker error.

Since this is so old, I think it would be ok to rebase it on current `develop`, instead of needing to do the work of a merge that includes the reorginization. Do you want to do it, or should I?

scottschurr:
I'll do the rebase.  I should be able to get it done today.

ximinez:
Per offline discussion, the merge was actually easy, so I pushed it.

ximinez:
Presumably, this will be ready for merge once CI passes (unless it needs to be caught up to `develop` again).

PR #4860, opened 2024-01-02
[TRIVIAL] Add unit test for redundant payment

Bronek:

The preflight check to enforce that payee and payer must not be the same account have been in rippled at least 9 years, however we do not have unit test for this case. Add it.

### Context of Change

Apparently other crypto ledgers are attacked by creating large number of transactions with the same payee and payer. We are immune to this type of attack, but it is not obvious. Let's make it clear and also improve unit test coverage.

### Type of Change

- [ ] Documentation update
- [ ] Release


Bronek:
@intelliot this is ready to merge

PR #4863, opened 2024-01-03
Ignore python error about modifying system python

legleux:
Initially an alpine docker image is used at the beginning of the packaging pipeline. 
This ignores the error from adding a package to a non-virtualenv python install.

[Relevant PEP](https://peps.python.org/pep-0668/)

Error from pipeline:
```
error: externally-managed-environment

× This environment is externally managed
╰─> 
    The system-wide python installation should be maintained using the system
    package manager (apk) only.
    
    If the package in question is not packaged already (and hence installable via
    "apk add py3-somepackage"), please consider installing it inside a virtual
    environment, e.g.:
    
    python3 -m venv /path/to/venv
    . /path/to/venv/bin/activate
    pip install mypackage
    
    To exit the virtual environment, run:
    
    deactivate
    
    The virtual environment is not deleted, and can be re-entered by re-sourcing
    the activate file.
    
    To automatically manage virtual environments, consider using pipx (from the
    pipx package).

note: If you believe this is a mistake, please contact your Python installation or OS distribution provider. You can override this, at the risk of breaking your Python installation or OS, by passing --break-system-packages.
hint: See PEP 668 for the detailed specification.
```

- [ ] Documentation update
- [ ] Release

### DIFF
--- a/Builds/containers/gitlab-ci/docker_alpine_setup.sh
+++ b/Builds/containers/gitlab-ci/docker_alpine_setup.sh
@@ -10,6 +10,6 @@ apk add \
     bash util-linux coreutils binutils grep \
     make ninja cmake build-base gcc g++ abuild git \
     python3 python3-dev
-pip3 install awscli
+pip3 install awscli --break-system-packages
 # list curdir contents to build log:
 ls -la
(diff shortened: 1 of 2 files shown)


manojsdoshi (approved):
LGTM

ximinez (approved):
Looks fine, so long as the job works as expected.

thejohnfreeman (approved):

PR #4867, opened 2024-01-05
Change default send_queue_limit to 500

### LINKED ISSUE #4866: [Rippled disconnects Websockets under current network load] (Version: [1.12.0])
## Issue Description
rippled disconnects websockets due to high transaction volume / load on the network

## Steps to Reproduce
1. Run a `rippled` instance with high transaction volume on the ledger (such as Mainnet right now)
2. subscribe via websocket to the `transactions` stream

## Expected Result
Websocket stays connected and transactions are received

## Actual Result
`rippled` disconnects websocket

## Environment
- using `rippled` 1.12.0
- using any websocket client such as xrpl.py, xrpl.js, xrpl-client

## Additional Information
It turns out that the default value of the parameter `send_queue_limit` is not sufficiently set for the current transaction load on the XRPL Mainnet with around 200-400 transactions per ledger.

The default value is `100`. When a websocket client connects and simply subscribes to the `transactions` stream (no other load on the server or other connections), then the websocket client receives a `disconnected` error/message from rippled with code `1008`.

I set the `send_queue_limit` parameter to a higher value now and I don't have any websocket disconnects anymore.

So I suggest to change the `default` value of the `send_queue_limit` to a more appropriate value since the default config does not run stable on the current main net anymore.

thejohnfreeman:
Recent load on mainnet has averaged 250 txn/ledger. Clients subscribed to `transactions` over web socket are being disconnected because the traffic exceeds the default `send_queue_limit` of 100.

This changeset only touches the default configuration, not the default in code.

Fixes #4866

### DIFF
--- a/cfg/rippled-example.cfg
+++ b/cfg/rippled-example.cfg
@@ -1641,6 +1641,7 @@ port = 6006
 ip = 127.0.0.1
 admin = 127.0.0.1
 protocol = ws
+send_queue_limit = 500
 
 [port_grpc]
 port = 50051
@@ -1651,6 +1652,7 @@ secure_gateway = 127.0.0.1
 #port = 6005
 #ip = 127.0.0.1
 #protocol = wss
+#send_queue_limit = 500
 
 #-------------------------------------------------------------------------------
 


ximinez:
Since you're changing the config file, I suggest changing the title and commit message from "default" to "recommended".

nixer89 (commented):
lgtm!

nixer89 (approved):

dangell7 (approved):

intelliot:
ok for 2.0.1: only touches the example configuration - no code change.

intelliot:
> Since you're changing the config file, I suggest changing the title and commit message from "default" to "recommended".

Sorry I missed this comment 😵‍💫. In the future, you can always do a code review and "Request changes" in order to request changes.

ximinez:
> Sorry I missed this comment 😵‍💫. In the future, you can always do a code review and "Request changes" in order to request changes.

My bad. I thought I had.

legleux:
:+1: QA confirmed increasing the queue via config prevents `Client too slow` disconnections.

PR #4869, opened 2024-01-07
append the network id to the validation stream

### LINKED ISSUE #4783: Validation stream Network ID (Version: 1.12.0)
## Summary
Network id is impossible to establish from the validation stream, the stream contains validators of all networks.

## Motivation
Explorers, hosting tools, etc. Or anything try to validate data out of this stream will need to figure out which network the node is on.

## Solution
Add a network_id field to the validation data returned in the stream of data https://xrpl.org/subscribe.html#validations-stream

## Paths Not Taken
Unable to 100% day if the id is present in another hash or id it could possibly be derived from something already present.

lathanbritz:
Help identify which network the node should be part of this help explorers and the like. 

This fixes #4783

### DIFF
--- a/src/ripple/app/misc/NetworkOPs.cpp
+++ b/src/ripple/app/misc/NetworkOPs.cpp
@@ -2153,6 +2153,7 @@ NetworkOPsImp::pubValidation(std::shared_ptr<STValidation> const& val)
         jvObj[jss::flags] = val->getFlags();
         jvObj[jss::signing_time] = *(*val)[~sfSigningTime];
         jvObj[jss::data] = strHex(val->getSerializer().slice());
+        jvObj[jss::network_id] = app_.config().NETWORK_ID;
 
         if (auto version = (*val)[~sfServerVersion])
             jvObj[jss::server_version] = std::to_string(*version);
--- a/src/ripple/app/misc/NetworkOPs.cpp
+++ b/src/ripple/app/misc/NetworkOPs.cpp
@@ -2153,7 +2153,9 @@ NetworkOPsImp::pubValidation(std::shared_ptr<STValidation> const& val)
         jvObj[jss::flags] = val->getFlags();
         jvObj[jss::signing_time] = *(*val)[~sfSigningTime];
         jvObj[jss::data] = strHex(val->getSerializer().slice());
-        jvObj[jss::network_id] = app_.config().NETWORK_ID;
+        if (auto const netid = app_.overlay().networkID())
+            jvObj[jss::network_id] = static_cast<Json::UInt>(*netid);
+
 
         if (auto version = (*val)[~sfServerVersion])
             jvObj[jss::server_version] = std::to_string(*version);
--- a/src/ripple/app/misc/NetworkOPs.cpp
+++ b/src/ripple/app/misc/NetworkOPs.cpp
@@ -2156,7 +2156,6 @@ NetworkOPsImp::pubValidation(std::shared_ptr<STValidation> const& val)
         if (auto const netid = app_.overlay().networkID())
             jvObj[jss::network_id] = static_cast<Json::UInt>(*netid);
 
-
         if (auto version = (*val)[~sfServerVersion])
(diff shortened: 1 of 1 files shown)


src/ripple/app/misc/NetworkOPs.cpp
```diff
@@ -2153,6 +2153,7 @@ NetworkOPsImp::pubValidation(std::shared_ptr<STValidation> const& val)
         jvObj[jss::flags] = val->getFlags();
         jvObj[jss::signing_time] = *(*val)[~sfSigningTime];
         jvObj[jss::data] = strHex(val->getSerializer().slice());
+        jvObj[jss::network_id] = app_.config().NETWORK_ID;
```
mvadari, line 2156: Does this data need to be added to the `val` object?

ckeshava: `network_id` is returned in the `server_info` command output. It is computed in a slightly different way at https://github.com/XRPLF/rippled/blob/350d213ee8a709e0ed417da0f081291ed411480a/src/ripple/app/misc/NetworkOPs.cpp#L2523

I understand that this information is very convinient in the Validation stream, but can't we grab it from the `server_info` command too?

ckniffen: Making a call to server info is not always possible because not all validators expose their IPs.

mvadari: Most don't (especially UNL validators) for security reasons.

ckeshava: okay, I didn't realise that, thanks!

ckeshava (approved):

src/ripple/app/misc/NetworkOPs.cpp
```diff
@@ -2153,6 +2153,8 @@ NetworkOPsImp::pubValidation(std::shared_ptr<STValidation> const& val)
         jvObj[jss::flags] = val->getFlags();
         jvObj[jss::signing_time] = *(*val)[~sfSigningTime];
         jvObj[jss::data] = strHex(val->getSerializer().slice());
+        if (auto const netid = app_.overlay().networkID())
```
mvadari, line 2156: Is there a situation where there is no network ID?

ckeshava: Yeah, that occurs here: https://github.com/XRPLF/rippled/blob/fad9d639bf52e727aa9cc1bedbdcdb49aa4d36de/src/ripple/overlay/impl/OverlayImpl.cpp#L1617

But if no network_id is specified, how will a validator know which peers it must connect with?

intelliot: I think the ideal would be for nodes for nodes on different networks to not connect to each other in the first place. And if they are connected, they should disconnect from each other. That should automatically solve the issue since you wouldn't get validations from other networks to begin with.

lathanbritz: My understanding that would be done when choosing which peers to use and add? @mvadari @ckeshava is that correct. 

It also has a dependency on the id being present meaning all nodes need to update to a rippled version publishing this information with its peers information. Only then would it be possible to then exclude peers not on your network.

That fine if that is the chosen route to achieve that but it would need to be split across a few releases.

ckeshava: I don't know much about this topic. 
We are validating the `network_id` inside the transaction engine, but we are not validating it inside the classes `Peer`, `Overlay` or whilst setting up the UNL. Fixing the gap in these classes should fix the chatter in the gossip protocol.

I don't foresee any breaking of functionality if this idea were to be implemented immediately. The cross-communication of validations across different networks occurs due to mis-configurations. Yes, I believe it presents more work for the validator operator.

lathanbritz: im going to leave that off to another request as I also don't fully understand how best to achieve that.

PR #4871, opened 2024-01-09
Windows CI unit tests need to pass

ximinez:

Disable the Windows CI unit tests "allowed to fail" workaround

### Context of Change

The runner hardware was upgraded a little while ago, and the unit tests have been passing.

### Type of Change

- [X ] Chore (no impact to binary, e.g. `.gitignore`, formatting, dropping support for older tooling)

## Before / After

Before: In Windows CI, the unit tests were allowed to silently fail without causing the whole job to fail. This was because it was more important to ensure no inconsistencies were introduced into the build.

After: In Windows CI, if the unit tests fail, the job will fail.


ckeshava (approved):

thejohnfreeman (approved):

ximinez:
Suggested commit message:
```
Check for success/failure of Windows CI unit tests (#4871)

* Disable the Windows CI unit tests "allowed to fail" workaround
* The runner hardware was upgraded, and the unit tests have been passing since then.
```
Note: This is slated for 2.1.0, so do not merge until after 2.0.1 is frozen / released.

intelliot:
Same commit message as above but hard wrapped at 72 characters
```
Check for success/failure of Windows CI unit tests (#4871)

* Disable the Windows CI unit tests "allowed to fail" workaround.
* The runner hardware was upgraded, and the unit tests have been passing
  since then.
```

PR #4874, opened 2024-01-09
Update xxHasher library to v0.8.2

ckeshava:
Update the xxHasher library to use v0.8.2. The README page of the project lists several improvements in the performance of the algorithms in the newer version of the xxHasher library.

Kudos to @Bronek for brainstorming with me and also completing a lion's share of this pull request.

### Context of Change

We briefly considered using conan build system to pull the latest version of xxHash library. But NuDB has a dependency on an older version of xxHash (v0.6.2), this would make the conan approach a little long-winded.

### Type of Change

- [x] Performance
- [ ] Documentation update
- [ ] Release

### API Impact

No changes to API

Internal tracker: RPFC-85


ckeshava:
I'm pretty sure we haven't modified anything pertaining to the cryptographic algorithms implementation, but it would be great if this PR can be reviewed by an expert @JoelKatz  (cc Murat Cenk)

Bronek:
The compilation errors on Windows and MacOS are because the implementation file `xxhash.h` which we pulled from xxhash repo is littered with various `#include` directives, which are pulling symbols inside the `namespace beast::detail` rather than global namespace, as it was originally designed to. I can see two ways to solve this:
1. move all these includes near the top of the file, so they appear before `namespace beast::detail`. This will require untangling all the conditional pre-processor directives there are conditional to.
2. switch from using (edited) `xxhash.h` to straight conan dependency. This will require few relatively simple changes inside `xxhasher.h` so that the `XXH64_state_t` can be replaced with `XXH64_state_t*` 

Additionally I think we should switch to `XXH3` API because that's where the SIMD instructions are used, improving performance twofold (roughly, depending on platform and input size)

ckeshava:
Yes you are right, upon cursory inspection, `XXH3` appears to be superior than the previous algorithms

Bronek:
> @Bronek at present, the code compiles and unit tests pass correctly in `arm64-apple-darwin23.2.0` environment. So I don't know if we need as drastic a process as Step-1. I suspect there's something going wrong with the x86 `#include` conditionals in the code.

Well, yes. This error  `error: '__m128d' was not declared in this scope; did you mean 'beast::detail::__m128d'?` literally tells us that the names which should be global, got defined inside `namespace beast::details`. That happens if you do:

```c++
namespace beast::detail {
#include <someheader.h>
}
```
... which is what our edited `xxhash.h` header does. So, that's wrong and we need to do something about it.

> If we use the conan dependency solution, will it not cause a pollution of the `#define` macros? Some dependencies like NuDB use a different version of the library (xxHash v0.6.2)

Yes, there is a simple way to do that. Remove `#include "xxhash.h"` from `xxhasher.h` and move all the function definitions to `xxhasher.cpp`, replacing the XXH state with state pointer.

ckeshava:
@Bronek  sorry about the force-push, I wanted to simplify the commit history.
I have not picked up your commit (625a17d) yet, I'll do that once these changes get reviewed.

I pulled some of the `#include`'s out of the `beast::detail` namespace. There are one or two more `#include's` inside the `beast::detail` namespace, but those aren't affecting the compilation or the unit tests.

Bronek:
@ckeshava I was unable to push to your branch , so I have created a new PR https://github.com/XRPLF/rippled/pull/4893 to replace this one. It entirely removes `xxhash.h` and `xxhash.cpp` and replaces these with conan dependency, while also upgrading to `XXH3_...` i.e. SIMD version of hash operations.

intelliot:
Replaced by #4893

PR #4882, opened 2024-01-16
Revert "Asynchronously write batches to NuDB. (#4503)"

mtrippled:
This reverts commit 1d9db1bfdd84e980e44e2f521990eaca42633d5e.

Revert code that introduces problems associated with online deletion.

### Context of Change
The bug was introduced in PR #4503, which causes writes to block during online deletion. In turn, this delays persisting records to disk during consensus, which causes node desyncs.

### Type of Change

- [ ] Documentation update
- [ ] Release

### API Impact

Run online deletion as normal. It should stay stable, and revert to behavior from 1.12 and previous.

## Future Tasks

If we proceed with this revert, there will there be no reason to consider #4762.

Closes #4762

### DIFF
--- a/src/ripple/nodestore/Types.h
+++ b/src/ripple/nodestore/Types.h
@@ -36,9 +36,8 @@ enum {
     // This sets a limit on the maximum number of writes
     // in a batch. Actual usage can be twice this since
     // we have a new batch growing as we write the old.
-    // The main goal is to avoid delays while persisting the ledger.
     //
-    batchWriteLimitSize = 262144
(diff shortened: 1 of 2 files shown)


intelliot:
QA will check to see how this can be tested

ximinez (approved):
Clean revert.

intelliot:
Internal tracker: RPFC-80

thejohnfreeman (approved):

intelliot:
`Perf SignedOff`: "I think 4882 is good to go"

PR #4885, opened 2024-01-17
Headers needed in Conan package to build xbridge witness server

legleux:
Headers required for the xrpl Conan package to build [xbwd](https://github.com/ripple/xbridge_witness)

- [x] `libxrpl` change (any change that may affect `libxrpl` or dependents of `libxrpl`)

### DIFF
--- a/Builds/CMake/RippledCore.cmake
+++ b/Builds/CMake/RippledCore.cmake
@@ -230,6 +230,11 @@ install (
   FILES
     src/ripple/json/impl/json_assert.h
   DESTINATION include/ripple/json/impl)
+
+install (
+    FILES
+      src/ripple/net/RPCErr.h
+    DESTINATION include/ripple/net)
 install (
   FILES
     src/ripple/protocol/AccountID.h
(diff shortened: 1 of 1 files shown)


thejohnfreeman (changes requested):
The general pattern for `libxrpl` headers is to export entire "modules" (directories under `src/ripple`) rather than, within the same module, export some files in `libxrpl` and keep the rest in `rippled`. I don't recall if there are any existing exceptions, but we should try to avoid adding new ones.

For `server`, this changeset excludes `impl/JSONRPCUtil.h` and its source file. They appear to have no dependencies outside of `libxrpl`, so I think we should include them too, even if the witness server doesn't use them.

The `net` and `resource` modules are different. Most of their files are excluded here. I think that `net/RPCErr.h` can fit into the `protocol` module that is already in `libxrpl`. Either way, we must not forget to add `net/impl/RPCErr.cpp` to the sources.

`resource` appears to have no `rippled` dependencies outside of the `beast`, `json`, and `protocol` modules which are already in `libxrpl`. Perhaps we can export the entire module in `libxrpl`.

thejohnfreeman (approved):

intelliot:
note: legleux currently on PTO; to follow up after returning

legleux:
This is ready.
I did notice a few more things could be alphabetized in the `CMakeLists.txt`.

oleks-rip (approved):
Looks good

intelliot:
@legleux at your convenience, confirm that you still think this PR is ready to merge.

(I'm not sure who has permission to apply the `Passed` label, but if you can, feel free to do that when this is ready.)

intelliot:
I assume the force-push was just to rebase and resolve merge conflicts?

PR #4887, opened 2024-01-18
test: env unit test RPC errors return a unique result:

ximinez:

Changes the error used in unit tests `submit` returns an RPC error (including failed connections, etc.) from `temINVALID` to `telLOCAL_ERROR`. Also adds a few log messages that should be helpful to diagnose some test failure cases, and changes the description of `telLOCAL_ERROR`.

### Context of Change

`temINVALID` is currently returned when `submit` gets an RPC failure, but it is also used in several places in the transaction engine. This can make diagnosing test failures, especially spurious test failures, confusing and annoying to debug.
`telLOCAL_ERROR` is not new, but it is not used anywhere in the transaction engine. This will make those types of errors distinct and easier to diagnose.
I did not create a new code, because the available numeric range for codes is large, but not unlimited, and it's wasteful to create a new one when there's one which fits the purpose already available.

### Type of Change

- [X ] Tests (you added tests for code that already exists, or your new feature included in this PR)

### API Impact

None.

## Before / After

No externally visible effects when things are working correctly.

## Test Plan

Since this is limited to unit tests, there is no need for additional testing if those tests pass.

### DIFF
--- a/src/test/app/MultiSign_test.cpp
+++ b/src/test/app/MultiSign_test.cpp
@@ -247,7 +247,10 @@ public:
 
         // Duplicate signers should fail.
         aliceSeq = env.seq(alice);
-        env(noop(alice), msig(demon, demon), fee(3 * baseFee), ter(temINVALID));
+        env(noop(alice),
+            msig(demon, demon),
+            fee(3 * baseFee),
+            ter(telENV_RPC_FAILED));
         env.close();
         BEAST_EXPECT(env.seq(alice) == aliceSeq);
 
@@ -358,7 +361,7 @@ public:
         msig phantoms{bogie, demon};
         std::reverse(phantoms.signers.begin(), phantoms.signers.end());
         std::uint32_t const aliceSeq = env.seq(alice);
-        env(noop(alice), phantoms, ter(temINVALID));
+        env(noop(alice), phantoms, ter(telENV_RPC_FAILED));
         env.close();
         BEAST_EXPECT(env.seq(alice) == aliceSeq);
     }
@@ -1634,7 +1637,10 @@ public:
 
         // Duplicate signers should fail.
         aliceSeq = env.seq(alice);
-        env(noop(alice), msig(demon, demon), fee(3 * baseFee), ter(temINVALID));
+        env(noop(alice),
+            msig(demon, demon),
+            fee(3 * baseFee),
+            ter(telENV_RPC_FAILED));
         env.close();
         BEAST_EXPECT(env.seq(alice) == aliceSeq);
 
--- a/src/test/jtx/impl/Env.cpp
+++ b/src/test/jtx/impl/Env.cpp
@@ -280,7 +280,9 @@ Env::parseResult(Json::Value const& jr)
         jr[jss::result].isMember(jss::engine_result_code))
         ter = TER::fromInt(jr[jss::result][jss::engine_result_code].asInt());
     else
-        ter = temINVALID;
+        // Use an error code that is not used anywhere in the transaction engine
+        // to distinguish this case.
+        ter = telENV_RPC_FAILED;
     return std::make_pair(ter, isTesSuccess(ter) || isTecClaim(ter));
 }
 
@@ -288,23 +290,29 @@ void
 Env::submit(JTx const& jt)
 {
     bool didApply;
-    if (jt.stx)
-    {
-        txid_ = jt.stx->getTransactionID();
-        Serializer s;
-        jt.stx->add(s);
-        auto const jr = rpc("submit", strHex(s.slice()));
+    auto const jr = [&]() {
+        if (jt.stx)
+        {
+            txid_ = jt.stx->getTransactionID();
+            Serializer s;
+            jt.stx->add(s);
+            auto const jr = rpc("submit", strHex(s.slice()));
 
-        std::tie(ter_, didApply) = parseResult(jr);
-    }
-    else
-    {
-        // Parsing failed or the JTx is
-        // otherwise missing the stx field.
-        ter_ = temMALFORMED;
-        didApply = false;
-    }
-    return postconditions(jt, ter_, didApply);
+            std::tie(ter_, didApply) = parseResult(jr);
+
+            return jr;
+        }
+        else
+        {
+            // Parsing failed or the JTx is
+            // otherwise missing the stx field.
+            ter_ = temMALFORMED;
+            didApply = false;
+
+            return Json::Value();
+        }
+    }();
+    return postconditions(jt, ter_, didApply, jr);
 }
 
 void
@@ -342,11 +350,15 @@ Env::sign_and_submit(JTx const& jt, Json::Value params)
 
     std::tie(ter_, didApply) = parseResult(jr);
 
-    return postconditions(jt, ter_, didApply);
+    return postconditions(jt, ter_, didApply, jr);
 }
 
 void
-Env::postconditions(JTx const& jt, TER ter, bool didApply)
+Env::postconditions(
+    JTx const& jt,
+    TER ter,
+    bool didApply,
+    Json::Value const& jr)
 {
     if (jt.ter &&
         !test.expect(
@@ -356,6 +368,8 @@ Env::postconditions(JTx const& jt, TER ter, bool didApply)
                 transHuman(*jt.ter) + ")"))
     {
         test.log << pretty(jt.jv) << std::endl;
+        if (jr)
+            test.log << pretty(jr) << std::endl;
         // Don't check postconditions if
         // we didn't get the expected result.
         return;
--- a/src/test/protocol/Memo_test.cpp
+++ b/src/test/protocol/Memo_test.cpp
@@ -56,7 +56,7 @@ public:
             JTx memoSize = makeJtxWithMemo();
             memoSize.jv[sfMemos.jsonName][0u][sfMemo.jsonName]
                        [sfMemoData.jsonName] = std::string(2020, '0');
-            env(memoSize, ter(temINVALID));
+            env(memoSize, ter(telENV_RPC_FAILED));
 
             // This memo is just barely small enough.
             memoSize.jv[sfMemos.jsonName][0u][sfMemo.jsonName]
@@ -72,7 +72,7 @@ public:
             auto& m = mi[sfCreatedNode.jsonName];  // CreatedNode in Memos
             m[sfMemoData.jsonName] = "3030303030";
 
-            env(memoNonMemo, ter(temINVALID));
+            env(memoNonMemo, ter(telENV_RPC_FAILED));
         }
         {
             // Put an invalid field in a Memo object.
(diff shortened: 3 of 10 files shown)


ximinez:
@scottschurr @ckeshava The discussion in https://github.com/XRPLF/rippled/pull/4846#discussion_r1454285073 finally prompted me to clean up the branch enough to create a PR.

ckeshava:
I observed two other unit tests that are using `temINVALID`. I think they should use `telLOCAL_ERROR` instead. here is a branch with the change: https://github.com/ckeshava/rippled/tree/ed-localerr

I'd prefer to use `telUNIT_TEST_ERR` instead of `telLOCAL_ERR`, because it's more readable.

ckeshava:
Sorry, my bad. Ignore the above modifications to the `Ticket` and `PseudoTx` test classes. The changes are unwarranted and are correct as is

ckeshava (approved):

scottschurr (approved):
These changes look good.  Thanks for doing this.

That said, I think it could be better.  I think that @ckeshava is right, it would be better to use a new `TER` code with a proper name.

We have not historically burned through through `tel` codes very quickly.  Because we rely so much on our unit test framework, I think that sacrificing a `tel` code in order to be really clear would be very worthwhile.  The top-most branch here has made that change: https://github.com/scottschurr/rippled/commits/ed-env-failures/

For one thing, I think it makes it much more obvious what's happening if someone is casually reading a unit test.

But I won't insist on it.  This is a useful change either way.

ximinez:
@scottschurr Ok, I'm convinced. I grabbed your commit.

scottschurr (approved):
:+1: Looks great.  Thanks!

intelliot:
- awaiting `Passed` label by ximinez

ximinez:
Suggested squash commit message:
```
test: Env unit test RPC errors return a unique result: (#4877)

* telENV_RPC_FAILED is a new code, reserved exclusively
  for unit tests when RPC fails unexpectedly. This will
  make those types of errors distinct and easier to test
  for when expect and/or diagnose when not.
```

scottschurr (approved):
:+1:  I suspect that logging the returned JSON will help in some tricky debugging situations.  Thanks for adding that.

src/test/app/MultiSign_test.cpp
```diff
@@ -1634,7 +1637,10 @@ class MultiSign_test : public beast::unit_test::suite
 
         // Duplicate signers should fail.
         aliceSeq = env.seq(alice);
-        env(noop(alice), msig(demon, demon), fee(3 * baseFee), ter(temINVALID));
+        env(noop(alice),
+            msig(demon, demon),
+            fee(3 * baseFee),
+            ter(telENV_RPC_FAILED));
```
ckeshava, line 1643: @ximinez Only the error code in `Env.cpp`'s `Env::parseResult` should be returning `telENV_RPC_FAILED`. Why are all the other unit tests using this new error code?

For instance, if I use duplicate signers in an RPC call to a mainnet validator (non-unit test scenario), I should get `temINVALID` error code, not a `telRNV_RPC_FAILED`

src/test/jtx/impl/Env.cpp
```diff
@@ -280,31 +280,39 @@ Env::parseResult(Json::Value const& jr)
...
+        {
+            // Parsing failed or the JTx is
+            // otherwise missing the stx field.
+            ter_ = temMALFORMED;
+            didApply = false;
+
+            return Json::Value();
+        }
+    }();
+    return postconditions(jt, ter_, didApply, jr);
```
ckeshava, line 315: this is a useful change, thanks! it saves developers from printing out the suspicious outputs in debug sessions

src/test/protocol/Memo_test.cpp
```diff
@@ -72,23 +72,23 @@ class Memo_test : public beast::unit_test::suite
...
+            env(memoNonMemo, ter(telENV_RPC_FAILED));
         }
         {
             // Put an invalid field in a Memo object.
             JTx memoExtra = makeJtxWithMemo();
             memoExtra
                 .jv[sfMemos.jsonName][0u][sfMemo.jsonName][sfFlags.jsonName] =
                 13;
-            env(memoExtra, ter(temINVALID));
+            env(memoExtra, ter(telENV_RPC_FAILED));
```
ckeshava, line 83: To emphasize my point: this test case underscores invalid fields in a memo object. It is more appropriate to return a `temINVALID` code, instead of the current change. That would help localize `telENV_RPC_FAILED` to a few places in the code base.

ckeshava (commented):
I have a few clarifying questions about the intended usage of `telENV_RPC_FAILURES`. Otherwise, the PR looks fine

src/test/app/MultiSign_test.cpp
ximinez: > Why are all the other unit tests using this new error code?

Excellent question. The reason is that the RPC layer does some validation before ever invoking the transaction engine. While some of these checks are redundant, catching them early prevents a lot more work from being done later, and so saves time overall. In this example, I think the duplicate signer check is done before any signature verification, which is way more expensive.

Either way, this request is failing on those RPC checks, and so the result has no `TER` code. You can verify this yourself by changing the result to something else (e.g. `tesSUCCESS`), and noting that the new log message will show an RPC-level error, and not a transactor-level error.

src/test/protocol/Memo_test.cpp
ximinez: My explanation above covers this test case as well.

src/test/app/MultiSign_test.cpp
ximinez: You know.... It might not be a bad idea to introduce a new "requirement" parameter, similar to `ter()` for RPC result code values.... Then the test could explicitly specify _which_ kind of RPC failure it's expecting.

scottschurr: Potentially something not called `ter()` but `rpcErr()`?  So if a particular RPC error was expected we could check for that.  It might have the side effect of preventing the check for `tesSUCCESS`?

How hard would it be to code something like that up?

ximinez: I don't think it needs to change. I just hadn't looked at it in a while, and didn't remember the reasoning. In fact, `rpcErr()` could take advantage of the fact that it's `optional`.

ckeshava: what do you mean? set `rpcErr` to be an optional value in case certain tests do not perform an rpc calls?

ximinez: It would almost be easier to code up a proof of concept than to explain, but I'll try. The general idea is to add a `std::optional<error_code_i> rpc = std::nullopt;` field to `JTx` alongside `ter`. The `rpcErr` class would be like the `ter` class, but the `operator()` would set both `jt.ter = std::nullopt; // (or telENV_RPC_FAILED)` and `jt.rpc = <whatever>;`. Thus, you could use it instead of `ter()`.

ckeshava: I'm doing something similar, but I haven't used an optional: https://github.com/ckeshava/rippled/blob/38ecfe19227deefeb86db5624d740533c9d8d31f/src/test/jtx/ter.h#L58

this is still a WIP

ckeshava: Hello,
I have a question pertaining to this idea.

How can I access the `rpcError` of a particular rpc call? `jr[jss::result][jss::error_code]` could be a useful candidate, but rippled does not explicitly return an error code in many places.

Here are two examples: https://xrpl.org/docs/references/http-websocket-apis/api-conventions/error-formatting/ (although `rpcINVALID_PARAMS` is the rpcError in this instance, it is not bubbled up to the output)

Three instances in this file: https://github.com/XRPLF/rippled/blob/c88166e055e4378f78fb0236e0bf347d6544e517/src/ripple/rpc/handlers/Submit.cpp#L114 (Search for `jvResult[jss::error]` in this file)

Issues like Duplicate Signer lists bubble up through the Submit command, but they are not assigned a specific `jss::error_code` field in the result `Json::Value`.

A draft of my work is at the tip of this branch: https://github.com/ckeshava/rippled/tree/envRpcErr

ximinez: Here's a proof of concept I knocked out last week: https://github.com/XRPLF/rippled/compare/develop...ximinez:rippled:ci/jtx_rpc

I'm not sure it's 100% correct or done, but when I saw you were working on it, I tabled it. Here's the change where I extract the error information. https://github.com/XRPLF/rippled/compare/develop...ximinez:rippled:ci/jtx_rpc#diff-c3c592e646bf717a28060dbb4f27c750962afc4a7ad779fee4eadd1f11e6b500R278-R303

ckeshava: thanks Ed, I'll take a look

ckeshava: Hello Ed,
You have done a great job! your branch passes all the tests, I don't think I can do better 😅 

I have a minor nit about the placement of a comment, you can find it here: https://github.com/ckeshava/rippled/pull/new/minorNit

Having said that, I wanted to discuss three things:
1. I feel we can avoid using `exception` as a variable name (to avoid confusing with `std::exception`) here - https://github.com/XRPLF/rippled/compare/develop...ximinez:rippled:ci/jtx_rpc#diff-19c35fddcebc11f5dd0b6b647315df1340bdddc6f9a79c3a6de4b6a98a43a5a5R49

2. Regarding the presentation of RPC errors, you have structured it as one of two alternatives: `<rpcCode, epcError> or <rpcError, epcExceptionMessage>`. This does capture all the existing unit tests.
But I came across three instances where `<rpcCode, epcError, rpcExceptionMessage>` are all returned by rippled. Here are some examples:
- https://github.com/XRPLF/rippled/blob/develop/src/ripple/rpc/impl/TransactionSign.cpp#L537
- https://github.com/XRPLF/rippled/blob/develop/src/ripple/rpc/impl/ServerHandler.cpp#L906
- https://github.com/XRPLF/rippled/blob/develop/src/ripple/protocol/ErrorCodes.h#L221

As part of future work, it would be nice if we can have consistency across all such instances. Would it be too hard to specify an `rpcErrorCode` in all such instances? I don't know.

3. I realised that we don't unit-test `rpc<error_code>` nearly as much as the `tem/tec` error codes. A handful of unit tests expect `rpc`-related errors. I know that Bronek is working to add more unit tests. I feel such unit tests are useful to detect breaking of Public APIs. What do you think?

ximinez: Awesome. Thanks! I'll make those changes and open a PR!

ckeshava: I have tried to do a PoC for point-2 in my previous comment. Please check the top two commits: https://github.com/ximinez/rippled/compare/ci/jtx_rpc...ckeshava:minorNit

I think it would be nice if all the error cases would return a `jss::error_code`, instead of only returning a `jss::error` or `jss::error_message`. What do you think?

I think this will be a non-trivial amount of work.

The Developer Growth team is working on an OpenAPI specification. I believe it would be useful if we can have a consistent set of error codes -- it would be easier to model the RPC error cases. (This is my read of the situation, I haven't discussed this with the team yet)

ximinez: This looks like it would be a valuable addition, but I think adding new RPC codes is beyond the scope of this change, partly because it could change the RPC API.

ckeshava: yeah, that's true

PR #4890, opened 2024-01-18
fix: order book update variable swap:

seelabs:
This is likely the result of a typo when the code was simplified.

### Context of Change

This was reported by @sublimator in https://github.com/XRPLF/rippled/issues/4790

### Type of Change

### DIFF
--- a/src/ripple/app/ledger/OrderBookDB.cpp
+++ b/src/ripple/app/ledger/OrderBookDB.cpp
@@ -47,7 +47,7 @@ OrderBookDB::setup(std::shared_ptr<ReadView const> const& ledger)
 
     if (seq != 0)
     {
-        if ((seq > ledger->seq()) && ((ledger->seq() - seq) < 25600))
+        if ((ledger->seq() > seq) && ((ledger->seq() - seq) < 25600))
             return;
 
         if ((ledger->seq() <= seq) && ((seq - ledger->seq()) < 16))


Bronek (approved):
This is good. I would also recommend reading `ledger->seq()` once into a stack const variable once and also adding const to `auto seq`, but since the ledger is immutable it does not change anything except making it more explicit.

Bronek:
It just occurred to me that, since Apr 2022 when the buggy code was deployed, we were more-or-less performing full book update every time (instead of the intended every 25600 ledgers). When this PR is released we will bump it to every 25600 ledgers. This will probably have performance implications.

sublimator:
> we were more-or-less performing full book update every time

This is just from (foggy) memory at this point, but I think the take away I got was that it was simply NOT showing new order book pairs (which doesn't really happen all that often) until a restart, and was doing LESS work if anything. But I could have been wrong and/or misremembering. But yeah, if it does start doing the work it was supposed to, I guess there will be perf implications.

intelliot:
notes
- This change should only affect path finding. No amendment needed.
- Possible perf impact, but only back to old/previous behavior.

intelliot:
`Testable`:
- Notice that the bug snuck in and unit tests pass either way - that's not good.
- It would be good to add a test case. Let's try to write the test.

ximinez (approved):

seelabs:
Rebased onto the tip of develop. Will merge after CI passes

PR #4893, opened 2024-01-19
Upgrade to xxhash 0.8.2 as conan dependency, enable SIMD hashing

Bronek:

Upgrade from xxhash from 0.6.2 to 0.8.2 , switch to conan dependency , enable SIMD

### Context of Change

We are currently using old version 0.6.2 of [xxhash](https://github.com/Cyan4973/xxHash), as a verbatim copy & paste from its header file `xxhash.h`. Switch to the more recent version 0.8.2 . Since this version is supported by conan (and properly protects its ABI by keeping the state object incomplete), add it as a conan dependency. Also, since this new version supports SIMD instructions (in the new `XXH3` family of functions), switch to the new functions to make use of these more performant CPU instructions.

### Type of Change

- [x] Performance (increase or change in throughput and/or latency)
- [ ] Documentation update
- [ ] Release

This change requires performance testing.

Internal tracker: RPFC-85


Bronek:
This has cleared performance testing in https://ripplelabs.atlassian.net/browse/RPFC-85

thejohnfreeman (approved):

intelliot:
- awaiting review by @ximinez (or suggestion of alternate reviewer)

scottschurr (approved):
:+1: Looks good to me.  Nice job removing the two beast/hash/impl/xxhash.* files!

Bronek:
> `XXH64_update` instead of `XXH3_64bits_update`

Yes, this the upgrade to SIMD instructions part. In both cases (old and new code) we are operating on 64-bit data type.

Bronek:
> Everything looks good. I'm not very familiar with this library, so I took a quick look through the [documentation example](https://github.com/Cyan4973/xxHash/tree/v0.8.2?tab=readme-ov-file#example). I noticed that it does a lot more error checking than this version. Is that something we should be concerned about? 

Huh, maybe I need to add handling running out of memory in the constructors, otherwise other operations will silently fail. How do we normally handle it in this project ?

```c++
XXH_PUBLIC_API XXH3_state_t* XXH3_createState(void)
{
    XXH3_state_t* const state = (XXH3_state_t*)XXH_alignedMalloc(sizeof(XXH3_state_t), 64);
    if (state==NULL) return NULL;
```

scottschurr:
No, you don't need to do anything special about running out of memory.  If rippled runs out of memory it bursts into flame.  There's no special handling for low memory situations.

Bronek:
> No, you don't need to do anything special about running out of memory. If rippled runs out of memory it bursts into flame. There's no special handling for low memory situations.

In this case throwing `std::bad_alloc` is better than failing silently. I will add it.

Bronek:
@ximinez do you mind checking again ? I added `throw std::bad_alloc()` where appropriate. There are no other error paths that I can see in in xxhash sources.

ximinez (approved):

PR #4906, opened 2024-01-29
Add STObject constructor to explicitly set inner object template

gregtatcam:

Add an `STObject` constructor overload that includes an additional bool argument to set the inner object template. Set the argument to true if the constructed object is an inner object and if the `fixInnerObjTemplate` amendment is enabled.

### Context of Change

Currently, the inner object template is not set upon object creation. If the object contains an `soeDEFAULT` field, and this field is initially set to the default value, accessing the field results in `tefEXCEPTION` error. Conversely, if the top object's `soeDEFAULT` field, initially set to the default value, is accessed, it returns the default value since top objects have the template set during construction. This fix aims to make the inner object behavior consistent with the top object behavior. The inner object must be explicitly constructed, as described in the summary, for this fix to work.

Currently, two `AMM` fields, `sfTradingFee` and `sfDiscountedFee`, included in the inner objects `sfVoteEntry` and `sfAuctionSlot` respectively, are declared as `soeDEFAULT` fields. This fix employs the new `STObject` constructor when constructing inner objects that include these fields.

### Type of Change

- [ ] Documentation update
- [ ] Release

## Test Plan

`AMM` unit-test is extended with a test, which demonstrates the behavior before and after this fix.


seelabs:
Our current setup for amendments works well when modifying and add transactors. It works less well with bug fixes - especially bug fixes that effect lower level classes. We've run into this several times with `STAmount` and number classes, and now we're running into it with `STObject`. I dislike passing in a `bool` into the constructor like this patch does, although I understand why you did it.

src/ripple/app/misc/impl/AMMUtils.cpp
```diff
@@ -287,17 +287,18 @@ initializeFeeAuctionVote(
...
+    STObject voteEntry{sfVoteEntry, fixInnerObj};
     if (tfee != 0)
         voteEntry.setFieldU16(sfTradingFee, tfee);
     voteEntry.setFieldU32(sfVoteWeight, VOTE_WEIGHT_SCALE_FACTOR);
     voteEntry.setAccountID(sfAccount, account);
     voteSlots.push_back(voteEntry);
     ammSle->setFieldArray(sfVoteSlots, voteSlots);
     // AMM creator gets the auction slot for free.
-    auto& auctionSlot = ammSle->peekFieldObject(sfAuctionSlot);
+    STObject auctionSlot{sfAuctionSlot, fixInnerObj};
```
seelabs, line 301: Our current setup for amendments works well when modifying and add transactors. It works less well with bug fixes - especially bug fixes that effect lower level classes. We've run into this several times with STAmount and number classes, and now we're running into it with STObject. I dislike passing in a bool into the constructor like this patch does, although I understand why you did it.

What do you think about a `thread_local` variable that holds the current `Rules` (defaulting to a null ptr if there is no transaction active)? The variable would be setup on entry to a transaction and torn down on exit. The problems I see with this are:

1. It won't work with coroutines, since these can be resumed on different threads. I don't believe this is a problem now, but it could let bugs sneak in later if we use coroutines.
2. It wouldn't work with recursive transactions (which we don't have, but "batch" transactions may have a problem depending how they are implemented). This is fixable with a stack but could be another source of errors.
3. It may break leveling. This `thread_local` needs to sit at a lowest level of the hierarchy of a class that would use it. For now it could sit at the `protocol` level, which is fine. But later this could be an issue.

All in all I think I favor going in this directions, but thought I'd open it up for discussion. I'm leaving it as a comment on code because top-level comments cannot have threaded discussions.

gregtatcam: I like this idea. This could be made into a more general solution then. The overloaded `STObject` constructor with `bool` argument is not needed and existing `STObject(SField const&)` constructor can be used instead. This variable can set/torn down in `applyTransaction()` and some added AMM code can be removed. This is somewhat similar to `thread_local Number::rounding_mode Number::mode_`. Does this sound right?

If `applyTransaction()` is the right place then wouldn't it address 2?

ximinez: @seelabs Because, as far as I know, inner objects are always and only created in transactor classes and related helper functions, what do you think about undoing the `STObject` ctor changes, and adding a `makeInnerObject` function that has a `Rules` parameter?
```
STObject makeInnerObject(SField const& name, Rules const& rules)
{
    STObject obj{name};
    if (rules.enabled(fixInnerObjTemplate))
    {
        SOTemplate const* elements =
            InnerObjectFormats::getInstance().findSOTemplateBySField(name);
        if (elements)
            set(*elements);
    }
}
```
Both `STObject` and `Rules` live in `src/ripple/protocol`, so there are no levelization implications.

If a new amendment is added that uses inner objects, the `if` could check for all of them.

gregtatcam: Then each inner object has to be explicitly constructed this way rather then implicitly handled via `STObject` constructor. Why is this a better approach? To avoid 1-3 issues that @seelabs mentioned?

ximinez: Yeah, it avoids those issues, but it may not be a better solution. Maybe you could just pass `Rules` (or more specifically `std::optional<Rules>` into the current ctor and not add the helper).

gregtatcam: Actually, I would rather use the helper. The constructor is obscure.

seelabs: I like @ximinez idea of passing `Rules` - good thinking Ed.

src/ripple/app/misc/impl/AMMUtils.cpp
```diff
@@ -315,6 +316,7 @@ initializeFeeAuctionVote(
         auctionSlot.setFieldU16(sfDiscountedFee, dfee);
     else if (auctionSlot.isFieldPresent(sfDiscountedFee))
         auctionSlot.makeFieldAbsent(sfDiscountedFee);
+    ammSle->set(&auctionSlot);
```
ximinez, line 319: What exactly does this line do that wasn't being done before? It probably needs to also be gated by the amendment.

src/ripple/protocol/impl/STObject.cpp
```diff
@@ -57,6 +59,20 @@ STObject::STObject(SerialIter& sit, SField const& name, int depth) noexcept(
...
+STObject
+STObject::makeInnerObject(SField const& name, Rules const& rules)
+{
+    STObject obj{name};
+    if (rules.enabled(fixInnerObjTemplate))
+    {
+        SOTemplate const* elements =
+            InnerObjectFormats::getInstance().findSOTemplateBySField(name);
+        if (elements)
+            obj.set(*elements);
```
ximinez, line 71: Minor tweak:
```
        if (SOTemplate const* elements =
                InnerObjectFormats::getInstance().findSOTemplateBySField(name))
            obj.set(*elements);
```

src/ripple/protocol/impl/STObject.cpp
```diff
@@ -629,6 +645,12 @@ STObject::getFieldArray(SField const& field) const
 
 void
 STObject::set(std::unique_ptr<STBase> v)
+{
+    set(v.get());
+}
+
+void
+STObject::set(STBase* v)
 {
```
ximinez, line 654: Suggestion:
```
assert(v);
```

src/ripple/protocol/SOTemplate.h
```diff
@@ -35,6 +35,8 @@ enum SOEStyle {
     soeREQUIRED = 0,  // required
     soeOPTIONAL = 1,  // optional, may be present with default value
     soeDEFAULT = 2,   // optional, if present, must not have default value
+                      // inner object with the default fields has to be
+                      // constructed with STObject::makeInnerObject()
```
ximinez, line 39: I think this comment should copied to InnerObjectFormats.cpp, probably before line 138 (the closing `}`) or after line 27 (the opening `{`). That's where people are likely going to be making additions, and so will be more likely to see and heed the message.

ximinez (changes requested):

src/ripple/app/misc/impl/AMMUtils.cpp
gregtatcam: `auctionSlot` was originally fetched via `peekFieldObject()`, which was returning a reference to `auctionSlot`. The latest implementation constructs `auctionSlot` as the local variable, sets its fields, and then sets `auctionSlot` in `AMM` object. `set` originally had declaration only and was not implemented. I don't think it has to be gated.

seelabs: The original code got an existing object and modified it. The new code creates a new object and sets it. So if there are any fields that aren't explicitly set that could have been present when it was retrieved in the previous version with `peekFieldObject` then those fields will be different between the old and new version and that's transaction breaking.

It looks like `sfAuthAccounts` is such a field. This looks transaction breaking without being protected with an amendment to me. If for some reason there's no way `sfAuthAccount` could have ever been set then we'd at the very minimum need a comment, but we'd probably want to rewrite this code to protect that case anyway.

As a side note, is it OK that `sfAuthAccounts` got overwritten with the new code or do we want to preserve that?

gregtatcam: `sfAuthAccounts` is an optional field and is not going to be present in either the old version or the new version. The final result in terms of which fields are present and which are not is the same in both versions. What changes is how `sfDiscountedFee` is handled when it's fetched from `sfAuctionSlot` because `sfDiscountedFee` is `soeDEFAULT`. The old version fails if the field is not present. The new version fails if the amendment is disabled and succeeds if the amendment is enabled.

seelabs: Just so I understand, `sfAuctionSlot` is guaranteed not exist in `ammSle` on entry to this function? Or it might exist and `sfAuthAccounts` is guaranteed not to exist?

ximinez: Ok. I missed the `peekFieldObject()` change because I was focused on the `set`. I think this is fine, but another option to consider is that you could restore the `peekFieldObject()`, then set up the template here. The only real advantages I can think of offhand are that it would remove the need to create the temporary local, and it avoids a move assignment. The disadvantage is it adds yet another copy of that code block from the ctor. May not be worth it, but I offer for your consideration:
```
    // AMM creator gets the auction slot for free.
    auto& auctionSlot = ammSle->peekFieldObject(sfAuctionSlot);
    if (rules.enabled(fixInnerObjTemplate))
    {
        if (SOTemplate const* elements =
                InnerObjectFormats::getInstance().findSOTemplateBySField(sfAuctionSlot))
            auctionSlot.set(*elements);
    }
    auctionSlot.setAccountID(sfAccount, account);
[etc]
```
But enough bike-shedding.

ximinez (approved):
Looks good. One optional comment to consider.

src/ripple/app/misc/impl/AMMUtils.cpp
gregtatcam: Actually, `sfAuctionSlot` does not exist when called from `AMMCreate` and exists when called from `AMMDeposit`. The latter can be called to restore `AMM` from an empty state. So `sfAuthAccounts` have to be preserved in case it's called from `AMMDeposit` if the amendment is not enabled and reset if the amendment is enabled.

gregtatcam: How about something like this:
>        STObject auctionSlot = STObject::makeInnerObject(sfAuctionSlot, rules);
>        if (!rules.enabled(fixInnerObjTemplate) && ammSle->isFieldPresent(sfAuctionSlot))
>        {
>            auto& as = ammSle->peekFieldObject(sfAuctionSlot);
>            if (as.isFieldPresent(sfAuthAccounts))
>                auctionSlot.setFieldArray(sfAuthAccounts, as.getFieldArray(sfAuthAccounts));
>        }

danielwwf:
After talking with some people, we would like to see a 2.0.2 for this fix that gets pushed ASAP. Waiting for 2.1.0 seems too long imho.

gregtatcam:
> Looks good. One optional comment to consider.

There is another update. Please re-review. Thanks.

src/ripple/app/misc/impl/AMMUtils.cpp
```diff
@@ -138,6 +138,10 @@ std::uint16_t
 getTradingFee(ReadView const& view, SLE const& ammSle, AccountID const& account)
 {
     using namespace std::chrono;
+    // should not happen
```
seelabs, line 141: Asserts by definition "should not happen". The comment doesn't add value. I'd remove it.

src/ripple/app/misc/impl/AMMUtils.cpp
```diff
@@ -287,17 +291,25 @@ initializeFeeAuctionVote(
...
         voteEntry.setFieldU16(sfTradingFee, tfee);
     voteEntry.setFieldU32(sfVoteWeight, VOTE_WEIGHT_SCALE_FACTOR);
     voteEntry.setAccountID(sfAccount, account);
     voteSlots.push_back(voteEntry);
     ammSle->setFieldArray(sfVoteSlots, voteSlots);
     // AMM creator gets the auction slot for free.
-    auto& auctionSlot = ammSle->peekFieldObject(sfAuctionSlot);
+    // AuctionSlot is created on AMMCreate and updated on AMMDeposit
+    // when AMM is in an empty state
+    if (!ammSle->isFieldPresent(sfAuctionSlot))
```
seelabs, line 307: This needs to be protected by an an amendment check. In general, when adding transaction breaking code, the code should follow the pattern:
```
if (amendment_on){
// do the new thing
}
else{
// do the old thing
}
```
It makes the code much easier to audit and makes it much less likely that transaction breaking changes will sneak in.

src/ripple/app/tx/impl/AMMBid.cpp
```diff
@@ -173,8 +173,16 @@ applyBid(
         return {tecINTERNAL, false};
     STAmount const lptAMMBalance = (*ammSle)[sfLPTokenBalance];
     auto const lpTokens = ammLPHolds(sb, *ammSle, account_, ctx_.journal);
+    // should not happen
```
seelabs, line 176: Rm comment.

src/ripple/app/tx/impl/AMMBid.cpp
```diff
@@ -173,8 +173,16 @@ applyBid(
...
+    assert(
+        !ctx_.view().rules().enabled(fixInnerObjTemplate) ||
+        ammSle->isFieldPresent(sfAuctionSlot));
     if (!ammSle->isFieldPresent(sfAuctionSlot))
```
seelabs, line 180: This needs to be protected by an an amendment check. In general, when adding transaction breaking code, the code should follow the pattern:
```
if (amendment_on){
// do the new thing
}
else{
// do the old thing
}
```
It makes the code much easier to audit and makes it much less likely that transaction breaking changes will sneak in.

src/ripple/app/tx/impl/AMMVote.cpp
```diff
@@ -199,6 +200,11 @@ applyVote(
         }
     }
 
+    // should not happen
```
seelabs, line 203: Rm comment.

src/ripple/protocol/impl/STObject.cpp
```diff
@@ -630,6 +645,13 @@ STObject::getFieldArray(SField const& field) const
 void
 STObject::set(std::unique_ptr<STBase> v)
 {
+    set(v.get());
+}
+
+void
+STObject::set(STBase* v)
```
seelabs, line 652: Why does this take a pointer as an argument? It seems like `STBase&&` would be clearer.

src/ripple/rpc/handlers/AMMInfo.cpp
```diff
@@ -200,6 +200,10 @@ doAMMInfo(RPC::JsonContext& context)
     }
     if (voteSlots.size() > 0)
         ammResult[jss::vote_slots] = std::move(voteSlots);
+    // should not happen
```
seelabs, line 203: Rm comment.

src/ripple/protocol/impl/STObject.cpp
gregtatcam: This function was already declared in `STObject` but it was not defined. I just added the implementation.

seelabs: I'm still not understanding why this interface uses a pointer and not an rvalue ref. rvalue ref seems better given how this is implemented. Am I missing something?

gregtatcam: Changed to rvalue ref.

seelabs (approved):
:+1:

dangell7:
Why don't we just make these fields `soeOPTIONAL`? Whats the purpose of these being `soeDEFAULT`?

Also you stated `If the object contains an soeDEFAULT field, and this field is initially set to the default value, accessing the field results in tefEXCEPTION error.`

Isn't this exactly what default behavior is supposed to be? `soeDEFAULT // optional, if present, must not have default value`

Thanks

gregtatcam:
> Why don't we just make these fields `soeOPTIONAL`? Whats the purpose of these being `soeDEFAULT`?
> 
> Also you stated `If the object contains an soeDEFAULT field, and this field is initially set to the default value, accessing the field results in tefEXCEPTION error.`
> 
> Isn't this exactly what default behavior is supposed to be? `soeDEFAULT // optional, if present, must not have default value`
> 
> Thanks

Changing the field's style is risky and it's a lot more code changes.

`soeDEFAULT` field saves space. If it has the default value then it should not be present. But if it is accessed then it should return the default value if it's not present.

dangell7:
> > Why don't we just make these fields `soeOPTIONAL`? Whats the purpose of these being `soeDEFAULT`?
> > Also you stated `If the object contains an soeDEFAULT field, and this field is initially set to the default value, accessing the field results in tefEXCEPTION error.`
> > Isn't this exactly what default behavior is supposed to be? `soeDEFAULT // optional, if present, must not have default value`
> > Thanks
> 
> Changing the field's style is risky and it's a lot more code changes.
> 
> `soeDEFAULT` field saves space. If it has the default value then it should not be present. But if it is accessed then it should return the default value if it's not present.

Ah I see thank you

src/test/jtx/impl/AMM.cpp
```diff
@@ -494,6 +494,22 @@ AMM::deposit(
...
+        arg.asset1In,
+        arg.asset2In,
+        arg.maxEP,
+        arg.flags,
+        arg.assets,
+        arg.seq,
+        arg.tfee,
+        arg.err);
+}
+
```
ximinez, line 512: Maybe not for this PR, but it might be worth refactoring this at some point to get rid of the overloads with multiple arguments, and just use these handy structures.

src/test/jtx/impl/AMM.cpp
```diff
@@ -722,20 +776,28 @@ bool
...
-            auto const slotPrice = auctionSlot[sfPrice].iou();
-            auto const authAccounts = auctionSlot.getFieldArray(sfAuthAccounts);
-            return cb(slotFee, slotInterval, slotPrice, authAccounts);
+            auto const& auctionSlot =
+                static_cast<STObject const&>(amm->peekAtField(sfAuctionSlot));
+            if (auctionSlot.isFieldPresent(sfAccount))
+            {
+                // this could fail in pre-fixInnerObjTemplate test if one
+                // the fail-cases. access as optional to avoid the failure
+                auto const slotFee = auctionSlot[~sfDiscountedFee].value_or(0);
```
ximinez, line 792: I think this comment is trying to say "This could fail in pre-fixInnerObjTemplate tests if the submitted transactions recreate one of the failure scenarios. Access as optional to avoid the failure." Could you update to be clearer? (Feel free to use what I wrote, or rephrase it another way.)

ximinez (changes requested):

src/test/jtx/impl/AMM.cpp
gregtatcam: I opened a JIRA ticket for this.

seelabs (approved):

ximinez (approved):

intelliot:
- @legleux reviewing for QA

PR #4907, opened 2024-01-31
Periodically pause copying ledger nodes during online_delete

ximinez:

Mitigates disk write contention while old ledgers are being deleted, and specifically while the full ledger is being copied to the "new" node store.

### Context of Change

PR #4503 (reverted by #4882) attempted to improve `rippled` performance by writing batches to NuDB asynchronously. However, it had an unintended side effect that when `online_delete` writes the entire ledger to disk, it tends to cause the buffer to fill up, which results in blocking new ledgers from being persisted.

### Type of Change

- [X] Performance (increase or change in throughput and/or latency)

## Before / After

Cost: This change will cause `online_delete` to take significantly longer to copy the full ledger from the old node store to the new one, which is the last significant step in the process.
Benefit: `rippled` should be much less likely to desync, and put less load on the disk during `online_delete`.

`online_delete` is still a demanding process, so this won't be a panacea, but it should be a significant improvment. (Significant improvement to be measured.)

### Performance details

1. This is an improvement to existing functionality, which could be considered a bug fix.
2. The change impacts node store writes. Specifically, it should reduce contention between `online_delete` and writing new ledgers.
3. The impact should be measured in a couple of different ways.
    1. `rippled` should struggle less during `online_delete` to stay synced, and other functions.
    2. `rippled` should put less strain/demand on the disk during `online_delete`.
5. This change affects concurrent processing, in the sense that multiple threads are writing to the node store, especially during `online_delete`.

Note that `back_off_milliseconds` is configurable, defaulting to 100. Node operators can de-prioritize `online_delete` operations more by increasing this value to whatever they are comfortable with.


ximinez:
Internal tracker: https://ripplelabs.atlassian.net/browse/RPFC-107

mtrippled (approved):
:+1: LGTM

scottschurr (approved):
:+1: LGTM!

intelliot:
Note: As of May 28, 2024 - perf testing is still in progress.

ximinez:
Even though this has two approvals, I'm changing it back to draft because early perf tests did not show any change, and there may be a better / different solution.

ximinez:
Closed in favor of #5276

PR #4921, opened 2024-02-19
Network id validation stream and forwarding

shortthefomo:

- append the network id to the ledger and validation stream.
- featureNetworkIDValidation which also stops forwarding validations to peers to stop cross network talk.

### Context of Change

Rework of #4869 and add the feature request by @intelliot  in #4783

### Type of Change

### API Impact

- [x] Peer protocol change (must be backward compatible or bump the peer protocol version)


lathanbritz:
@intelliot this one is ready for review

src/ripple/app/misc/NetworkOPs.cpp
```diff
@@ -2305,6 +2308,16 @@ NetworkOPsImp::recvValidation(
     std::shared_ptr<STValidation> const& val,
     std::string const& source)
 {
+    // Don't relay validations from another network.
+    if (auto networkID = app_.overlay().networkID())
+    {
+        auto const view = m_ledgerMaster.getCurrentLedger();
+
+        if (view->rules().enabled(featureNetworkIDValidation) &&
+            val->getNetworkID() != static_cast<Json::UInt>(*networkID))
```
exunda, line 2317: This code is probably wrong. The cast is unnecessary and confusing but (if my understanding is correct) the call `val->getNetworkID()` which calls `getFieldU32` which calls `STObject::getFieldByValue` will `throw` an exception if the `sfNetworkID` field is not present in the received validation.

src/ripple/overlay/impl/OverlayImpl.h
```diff
@@ -389,7 +389,9 @@ class OverlayImpl : public Overlay, public reduce_relay::SquelchHandler
     std::optional<std::uint32_t>
     networkID() const override
     {
-        return setup_.networkID;
+        if (auto const netid = setup_.networkID)
+            return netid;
+        return 0;
```
exunda, line 394: Previously this code could return `std::nullopt` but now it will return an `std::optional` with a value of `0` which (if I understand the code correctly) is the ID of the mainnet. That makes no sense and seems very wrong. Why even have the optional return?

src/ripple/protocol/impl/STValidation.cpp
```diff
@@ -90,6 +92,12 @@ STValidation::getConsensusHash() const
     return getFieldH256(sfConsensusHash);
 }
 
+std::uint32_t
+STValidation::getNetworkID() const
+{
+    return getFieldU32(sfNetworkID);
```
exunda, line 98: I think you should return optional here. Something like:

```
std::optional<std::uint32_t>
STValidation::getNetworkID() const
{
    if (!isFieldPresent(sfNetworkID))
        return std::nullopt;
    return getFieldU32(sfNetworkID);
}
```

There may be a way to do this more efficiently using the weird `~sfNetworkID` syntax but I'm still learning code.

exunda (commented):
Hi 👋  I've been looking at this code for a few weeks but still learning and this is my first code review on it. Hopefully you find the comments useful!

src/ripple/app/misc/NetworkOPs.cpp
shortthefomo: This is my first attempted commit to rippled and still trying to pick up C, so also learning lots as I go here. That being said will, test and validate this and possibly also add a test case to prove that we get the desired result here.

src/ripple/overlay/impl/OverlayImpl.h
exunda: I think that this will result in a change of the behavior and I am not sure it is correct. The code previously could return `std::nullopt` and now won't and maybe some parts of the code relied on that? This change should be signed off by someone with better understanding of the codebase than me before you put the change in and admins of servers shoudl be informed.

Bronek: I agree with @exunda , this looks incorrect

PR #4923, opened 2024-02-20
XLS-60d: DefaultAutoBridge

tequdev:

Changes the internal default path to include the path of the XRP bridge if sfPaths is unspecified in the Payment transaction.

### Context of Change

https://github.com/XRPLF/XRPL-Standards/discussions/175

### Type of Change

## Before / After
If relevant, use this section for an English description of the change at a technical level.
If no path is specified, the impact of the transaction on the node is increased, but if a path is set to mediate XRP, the impact on the node is unaffected.


ximinez:
Since the [XLS is still in discussion](https://github.com/XRPLF/XRPL-Standards/discussions/175), and may or may not move forward, I'm going to close this PR for the time being. Feel free to reopen or recreate this PR when the spec is ready.

Also note, that there are conflicts with `develop`, and the changes may need to be updated to account for the retirement of `FlowCross`.

PR #4924, opened 2024-02-20
Change order of checks in amm_info

Bronek:

Checking if account is well-formed currently happens before the check if `asset` , `asset2` and `amm_account` are supplied, which results in an unexpected error messages if any of the fields is malformed. Swap the order of checks, so validity of fields is verified first. Since this is a breaking change, retain the existing behaviour `if (apiVersion < 3)` and add identical  check, but **after** checks of assets and accounts, `if (apiVersion >= 3)`. Also add unit tests to demonstrate the old and new behaviour.

### Type of Change

- [ ] Documentation update
- [ ] Release

### API Impact

Before: sending an invalid account or asset to `amm_info` RPC method while other parameters are not set as expected would result in `rpcINVALID_PARAMS` error. This behaviour is **retained** in API version 2

After:  sending an invalid account or asset to `amm_info` RPC method while other parameters are not set as expected will result in `rpcISSUE_MALFORMED` or `rpcACT_MALFORMED` error. This behaviour is **new** in API version 3

This resolves https://github.com/XRPLF/clio/issues/1157


godexsoft (approved):
LGTM 👍

src/test/rpc/AMMInfo_test.cpp
```diff
@@ -75,6 +75,52 @@ class AMMInfo_test : public jtx::AMMTestBase
             }
         });
 
+        // Invalid parameters *and* invalid LP account, default API version
+        testAMM([&](AMM& ammAlice, Env&) {
+            Account bogie("bogie");
```
gregtatcam, line 80: `bogie` should be const. 
Consider declaring it once outside of `testAMM` and then using in each test.

src/test/rpc/AMMInfo_test.cpp
```diff
@@ -75,6 +75,52 @@ class AMMInfo_test : public jtx::AMMTestBase
...
+            std::vector<std::tuple<
+                std::optional<Issue>,
+                std::optional<Issue>,
+                std::optional<AccountID>,
+                bool>>
+                vals = {
```
gregtatcam, line 86: `vals` is the same in a few tests. Consider declaring it once outside of `testAMM` and then using it in each test.

gregtatcam (approved):
👍 LGTM

Bronek:
this is good to merge - note it preserves the current behaviour in API version 2 and adds a test for it.

seelabs:
@godexsoft Yes, it should be merged today.

godexsoft:
@Bronek this still requires `[beta_rpc_api]` to be set to `1` in config. Don't we want this to be enabled by default as the maximum allowed version yet? What's the plan for the switch if any?

ximinez:
> this still requires `[beta_rpc_api]` to be set to `1` in config. Don't we want this to be enabled by default as the maximum allowed version yet? What's the plan for the switch if any?

We never want `[beta_rpc_api]` to be 1 by default. If we want to declare a particular version of the API as done, then we make that version supported by increasing `apiMaximumSupportedVersion` and then we open up a "new" beta version by incrementing `apiBetaVersion`. `apiBetaVersion` should probably always be one more than `apiMaximumSupportedVersion`.

godexsoft:
> > this still requires `[beta_rpc_api]` to be set to `1` in config. Don't we want this to be enabled by default as the maximum allowed version yet? What's the plan for the switch if any?
> 
> We never want `[beta_rpc_api]` to be 1 by default. If we want to declare a particular version of the API as done, then we make that version supported by increasing `apiMaximumSupportedVersion` and then we open up a "new" beta version by incrementing `apiBetaVersion`. `apiBetaVersion` should probably always be one more than `apiMaximumSupportedVersion`.

I did not mean to have the beta enabled by default, that would be silly. I mean to have v3 as the new maximum version. I guess my wording was a little off 🔢  
In Clio we will start treating v3 as the maximum (but not default) version in one of the next releases just so we can forward to `rippled` with v3. But turns out v3 is still in beta so this fix here can't be used unless the operator enables beta or `rippled` moves on to v3 as the new allowed max version to request.

PR #4934, opened 2024-02-28
remove the type filter from "ledger" RPC command

ckeshava:

This issue was reported on the Javascript client library: https://github.com/XRPLF/xrpl.js/issues/2611

The `type` filter (Note: as of the latest version of rippled, `type` parameter is deprecated) does not work as expected. This PR removes the `type` filter from the `ledger` command.

## Context of the bug
I have not investigated the full extent of this issue. I have reproduced the issue on this branch: https://github.com/XRPLF/rippled/compare/develop...ckeshava:rippled:failingTestCasesTypeFilter?expand=1

Note: The above branch has annotations about which portions of the codebase are responsible for the wrong behavior.

Both `ledger_data` and `ledger` commands must behave identically with respect to the `type` filter, but that is not the case. While the above branch demonstrates failing tests, I'd like to highlight two tests that work correctly: https://github.com/XRPLF/rippled/blob/8a2f6bec33c6a250a6f5793a45c87e320e446214/src/test/rpc/LedgerRPC_test.cpp#L2241

This behavior occurs because of a specific combination of `expand`, `accounts` and `full` parameters in the `ledger` command. 

### Type of Change

- [ ] Documentation update
- [ ] Release
- [ ] Removal of feature, removal of associated unit tests too

### API Impact

`type` parameter can no longer be used with the `ledger` RPC command

No impact on performance of rippled


ckeshava:
@ximinez  @HowardHinnant  please review this PR at your convenience

mvadari (approved):

bthomee:
@ckeshava can you update your branch and resolve the conflicts?
@godexsoft adding you for an RPC sanity check.

godexsoft (approved):
If i understand correctly, this is the `type` that has been deprecated in Clio about a year ago. If that's correct then there should be no issue with Clio compatibility from this change.
EDIT: I should probably say that Clio will report a "deprecated" error when `type` is passed while `rippled` probably will not. But since it's now removed, this may be expected. If we want to have no discrepancy at all in the output IF `type` is used then perhaps it should be adjusted in this PR. @mounikakun do we check for this in integration?

mounikakun:
@godexsoft We do have integration tests for `type` in ledger but all of them forwards to rippled as `type` is deprecated in Clio. 

Also Clio do not return `deprecated` error if there's `type` in `ledger` request but returns default ledger response with a warning message. If rippled returns the same there won't be any discrepancy.

Clio request:
```json
{
    "method": "ledger",
    "params": [
      {
        "type": "ticket"
      }
    ]
  }
```

Response:
```json
 {
    "result": {
      "ledger_hash": "B93129538E2C211567FAC16702575CC278F288B9BEAB2FA8819E8142935256B4",
      "ledger_index": 3007200,
      "validated": True,
      "ledger": {
        "account_hash": "BD5DBDC8FAF4A66F8DD00FCC3311D6B7273690C58E219322E9E7482C25EA8AD1",
        "close_flags": 0,
        "close_time": 801266581,
        "close_time_human": "2025-May-22 22:03:01.000000000 UTC",
        "close_time_resolution": 10,
        "close_time_iso": "2025-05-22T22:03:01Z",
        "ledger_hash": "B93129538E2C211567FAC16702575CC278F288B9BEAB2FA8819E8142935256B4",
        "parent_close_time": 801266580,
        "parent_hash": "FE727AF6B6890F0D6BD62E95BB5786EE4BD6B90B641BC66B25A117C4E60F316D",
        "total_coins": "99999926515219548",
        "transaction_hash": "698B0C5B6CDB666D5B0FA2E480303AA8F53EC129D0D1963DA12EEB940D12CE6F",
        "ledger_index": "3007200",
        "closed": True
      },
      "status": "success"
    },
    "warnings": [
      {
        "id": 2004,
        "message": "Some fields from your request are deprecated. Please check the documentation at https://xrpl.org/docs/references/http-websocket-apis/ and update your request. Field "type" is deprecated."
      },
      {
        "id": 2001,
        "message": "This is a clio server. clio only serves validated data. If you want to talk to rippled, include "ledger_index":"current" in your request"
      },
      {
        "id": 2002,
        "message": "This server may be out of date"
      }
    ]
  }

```

Bronek:
Hi @ckeshava , it would be great if you could:

1. update this branch from the current `develop` and

2. act on this feedback from @mounikakun 
> Also Clio do not return deprecated error if there's type in ledger request but returns default ledger response with a warning message. If rippled returns similar response there won't be any discrepancy.

Also , a warning is preferred over error for reasons of backwards compatibility with older `rippled` versions.

ckeshava:
Hello, I'm working on rebasing my branch. This PR is quite old, so I'll need some time to wrangle out the old commits.

The `type` filter has been deprecated for many years now (as per the documentation on xrpl.org). Is there an official policy on when we can remove the deprecated features?

Furthermore, I do not agree with the current behavior of the Clio server. Since Clio ignores the `type` filter command, users might mis-interpret the response. I feel throwing an error is a better behavior.

However, I'm happy to make the necessary changes to provide a warning in this PR.

ckeshava:
Re-opening the PR after rebasing with the latest. The existing tests have been modified to validate the warnings thrown in the response.

ckeshava:
Hello! Gentle reminder: Is this PR acceptable? @Bronek 
Let me know if you have feedback

Bronek (approved):

src/test/rpc/LedgerRPC_test.cpp
```diff
@@ -711,6 +711,7 @@ class LedgerRPC_test : public beast::unit_test::suite
         env.close();
 
         std::string index;
+        int hashes_ledger_entry_index = -1;
```
mvadari, line 714: nit: should use camelCase

ckeshava: 20bc8bb900

PR #4944, opened 2024-03-08
Enforce no duplicate slots from incoming connections

Bronek:

We do not currently enforce that incoming peer connection does not have `remote_endpoint` which is already used (either by incoming or outgoing connection), hence already stored in `slots_`. If we happen to receive a connection from such a duplicate `remote_endpoint`, it will eventually result in a crash (when disconnecting) or weird behaviour (when updating slot state), as a result of an apparently matching `remote_endpoint` in `slots_` being used by a different connection

### Context of Change

This came from troubleshooting a crash of our node in livenet, as reported in [RIPD-1848](https://ripplelabs.atlassian.net/browse/RIPD-1848)

### Type of Change

- [ ] Documentation update
- [ ] Release

## Future Tasks

A different approach is also possible, where we do allow duplicate `remote_endpoint` and change the type of `slots_` accordingly. This would be a more intrusive change


Bronek:
## TODO (will remove this comment when done)

Add unit tests in `PeerFinder_test.cpp`

Bronek:
If the fix is reverted, then the newly added test will result in the same crash as reported (in the Release build)

```
% ./rippled -u PeerFinder
ripple.PeerFinder.PeerFinder backoff 1
Logic error: PeerFinder::Logic::remove(): remote_endpoint missing from slots_
[1]    950283 segmentation fault  ./rippled -u PeerFinder
```

scottschurr (approved):
:+1: LGTM!  Thanks for adding the unit test.

thejohnfreeman (changes requested):

src/test/peerfinder/PeerFinder_test.cpp
```diff
@@ -24,6 +24,7 @@
 #include <ripple/peerfinder/impl/Logic.h>
 #include <ripple/protocol/PublicKey.h>
 #include <ripple/protocol/SecretKey.h>
+#include "ripple/beast/unit_test/suite.hpp"
```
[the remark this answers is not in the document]
scottschurr: Good spotting! Sorry I missed that.

src/test/peerfinder/PeerFinder_test.cpp
```diff
@@ -157,6 +158,108 @@ class PeerFinder_test : public beast::unit_test::suite
...
+            "test", beast::IP::Endpoint::from_string("65.0.0.1:5"));
+        {
+            Config c;
+            c.autoConnect = false;
+            c.listeningPort = 1024;
+            c.ipLimit = 2;
+            logic.config(c);
+        }
+
+        PublicKey const pk(randomKeyPair(KeyType::secp256k1).first);
```
gregtatcam, line 180: This is not used.

src/test/peerfinder/PeerFinder_test.cpp
```diff
@@ -157,6 +158,108 @@ class PeerFinder_test : public beast::unit_test::suite
...
```
gregtatcam, line 231: This is not used.

src/test/peerfinder/PeerFinder_test.cpp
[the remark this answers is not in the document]
thejohnfreeman: Clangd is constantly trying to sneak these in on me. I checked and sadly there is no clang-format setting that can catch this for us.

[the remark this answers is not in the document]
Bronek: Thanks @thejohnfreeman , fixed !

I think that `clangd` is correct, see how the source directory is added in generated build files as a `-I` parameter, rather than `-isystem`. But that's a discussion for some other time and place.

src/test/peerfinder/PeerFinder_test.cpp
```diff
@@ -157,6 +158,108 @@ class PeerFinder_test : public beast::unit_test::suite
...
+        std::size_t n = 0;
+
+        for (std::size_t i = 0; i < seconds; ++i)
```
gregtatcam, line 181: What scenarios are tested by iterating 10,000 times? Slots are added by `new_*_slot()` functions and removed by `remove()` in `on_closed()`. Those functions are only executed if the list is not empty. Basically the test is just repeated a few times.

thejohnfreeman (approved):

src/test/peerfinder/PeerFinder_test.cpp
Bronek: I do  not know. I just reuse the same logic as other tests in this suite.

gregtatcam: Not sure about other tests but it doesn't look like there is any benefit in repeating this test N times in this test. I think we just want to test `new_inbound_slot()`, `new_outbound_slot()`, and `on_closed()`.

src/test/peerfinder/PeerFinder_test.cpp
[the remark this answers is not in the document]
thejohnfreeman: I think the opinions around this, held by people and tools, are conflicting:

- There is a convention that includes of system headers should use angles and non-system headers should use quotes.
- Some IDEs treat system headers differently.
- Some compilers turn off warnings for system headers.

The first convention turns into a problem for dependent packages. Those packages want to treat libxrpl headers as system headers. If our headers include each other using quotes, then they are violating the convention. There is an alternative convention, which we follow, to include all public headers with angles, and who is to say which convention is right? It's just preference.

clangd should just make this configurable, but [progress](https://github.com/clangd/clangd/issues/1247) is slow.

src/test/peerfinder/PeerFinder_test.cpp
gregtatcam: Up to you if you want to change it. I'm going to approve.

gregtatcam (approved):
LGTM

gregtatcam (approved):

Bronek:
This is ready to merge (but might want to delay merging after https://github.com/XRPLF/rippled/pull/4947 , so we have working Windows build as well)

scottschurr (approved):
The `test_duplicateOutIn()`test without the loop still causes the `LogicError` if the fix is removed.  The `test_duplicateInOut()` without the loop does not fire the `LogicError`, but it also did not fire the `LogicError` with the loop either.

Changes look good.  :+1:

Bronek:
> The `test_duplicateOutIn()`test without the loop still causes the `LogicError` if the fix is removed. The `test_duplicateInOut()` without the loop does not fire the `LogicError`, but it also did not fire the `LogicError` with the loop either.

Yes, that's correct - `new_outbound_slot` already did have the required check, so this test also increases test coverage outside of its immediate focus.

PR #4945, opened 2024-03-09
Address rare corruption of NFTokenPage linked list

scottschurr:

It was discovered that there are a couple of NFT directories on the MainNet that are missing links in the middle of the chain.  This pull request does several things to address the corruption:

1. The source of the corruption was identified and fixed, which should prevent the problem from happening in the future.
2. A new transaction, `LedgerStateFix`, is added which allows any account on ledger to pay the fee to repair the links of any single account.
3. More invariants are added which should prevent similar problems from happening in the future.
4. And unit tests are added to exercise all of the above changes.

A new amendment, `fixNFTokenPageLinks`, includes all of these changes.

Details follow.

### The Source of the Corruption

There are two different transactions that introduced the corruption into the two directories.  In both cases the following conditions were met:

1. There were at least two NFToken pages in the directory.
2. The next-to-last page was completely full, holding 32 NFTokens.
3. The very last page of the directory contained only one NFToken, and the transaction removed that last remaining token from the last page.

When these conditions were met, the last NFToken page was removed and the next-to-last page was left as the final page in the directory.

That would be fine except the NFToken directory has a an expectation that the last page has a specific index.  The page with that index was just deleted.  So when an NFToken is added to the directory, and that token has a high enough value that it does not belong on the new last page, then a new last page is created that has no links to the previous page.

Now there's a linked list with a hole in the middle of the list.

The `fixNFTokenPageLinks` amendment modifies the NFToken page coalescing code to notice when the very last page of the directory would be removed.  In that case it simply moves all of the contents of the next lower page into the last page and deletes the next-to-last page (and fixes up the links).

New invariant checks also validate aspects of the links on pages, so a similar corruption should cause a `tecINVARIANT_FAILED` transaction result.  That should prevent similar corruptions in the future.

### The New Transaction: `LedgerStateFix`

Once the amendment goes live then that prevents any similar NFToken directory corruption in the future.  But we are still left with two damaged NFToken directories.  And potentially more that could be created between now and when the amendment goes live.

We could leave the corruptions.  I don't like that idea for two reasons:

1. These corruptions seriously inconvenience those accounts that suffered the corruption, through no fault of their own.
2. When you drop food on the kitchen floor you clean it up.  You don't put up a sign that says, please don't step here.  You clean it up.

We could put in code that cleans up the two known corrupted directories when the amendment goes live.  That's not a good approach for two reasons:

1. As mentioned before, additional directories could face similar corruption between now and when the amendment goes live.  We would like to be able to clean up all damaged NFToken directories, regardless of when they are created.
2. The MainNet is not the only place running the XRP Ledger code.  Other networks may have similar corruptions that need to be fixed.  The approach to the fix should be applicable to other networks as well.

We need a transaction to clean up the directories.  There are no transactions that just naturally walk an NFToken directory.  So we can't leverage an existing transaction.  There are RPC commands that walk the directory, but RPC commands can't modify ledger state.

But we would also rather not introduce a new single-use transaction for this very tiny corner case.

So this approach creates a new transaction aimed at cleaning up messes in the ledger: `LedgerStateFix`.  We'll use it for this situation.  If there are other repairable corruptions in the future we can re-use this transaction to repair those future corruptions.

#### `LedgerStateFix`

`LedgerStateFix` fields (in addition to common fields) are:
|    Field Name       |  Data Type  | Required/Optional | Description                        |
|---------------------|-------------|-------------------|------------------------------------|
| `TransactionType`   | `uint16`    | Required          | `LedgerStateFix`                   |
| `Account`           | `STAccount` | Required          | Account that signs and pays fee    |
| `Fee`               | `STAmount`  | Required          | Fee paid for transaction           |
| `Flags`             | `uint32`    | Optional          | Available for future fixes         |
| **`LedgerFixType`** | `uint16`    | Required          | Only `1` is currently supported    |
| **`Owner`**         | `STAccount` | Optional          | Required if `LedgerFixType` == `1` |

**`TransactionType`** — identifies this as a `LedgerStateFix` transaction.

**`Account`** — identifies the account signing and submitting the transaction as well as paying the `Fee`.

**`Fee`** — this transaction is rare and potentially compute intensive.  We want to discourage folks from spamming the ledger with these.  So the minimum fee is the same as the fee for an `AccountDelete` transaction.  And, yes, if the transaction fails with a `tec` code the fee is still charged.

**`Flags`** — not needed for `LedgerFixType == 1`.  May be used for some unknown future ledger fix.

**`LedgerFixType`** — for now the only valid value is `1`, which fixes the NFToken directory for a single account.

**`Owner`** — the account that owns the NFToken directory that needs fixing.  Need not have any relationship to `Account`.

#### Fixing NFTokenPage Links

Once the amendment goes live, any corrupted NFTokenPage directory can be fixed by submitting a transaction like this:
```
{
   "Account" : "<Signer and fee payer>",
   "Fee" : "2000000",
   "LedgerFixType" : 1,
   "Owner" : "<Account with corrupted NFTokenPage directory",
   "Sequence" : <n>,
   "SigningPubKey" : "<Account's public key>",
   "TransactionType" : "LedgerStateFix",
   "TxnSignature" : "<Signature>",
}
```
If a repair is actually completed, then `tesSUCCESS` is returned.  If there is an error or if the directory did not need repair then a `tec` code is returned.

#### Future Fixes

Unfortunately this is not the first time ledger state has needed tidying up.  Amendment [`fixTrustLinesToSelf`](https://xrpl.org/resources/known-amendments/#fixtrustlinestoself) removed invalid state from the ledger.  Since programmers are humans we can expect similar kinds of problems to show up in the future.

If there are future ledger state problems, they can be attached to the `LedgerStateFix` transaction by picking a `LedgerFixType` value other than `1`.  Any additional parameters can be added as optional values.

### Context of Change

While I was research how efficient NFTokenPages actually are in use, I stumbled across two NFToken directories that were missing links in the middle of the chain.  This is a form of ledger corruption.  But its consequences do not seem too dire, even for the owners of those directories.  Known consequences are:

1. The `account_objects` RPC command returns only those objects at the beginning of the NFToken directory up to the point where a link is missing.  So, for those accounts with missing links, the NFTokens returned by `account_objects` is incomplete.
2. The `account_nfts` RPC command has a problem similar to `account_objects`.  `account_nfts` returns NFTokens only up to the point where a link is missing.

There may be other consequences, but I've looked for and not found any.  Particularly, I've not seen any evidence of _lost_ NFTokens.  NFTokens still get added to the correct owner's page.  They simply are not reported by the RPC commands.

### Type of Change

- [x] Tests (there are tests for the new code)
- [x] Documentation will need to be added to cover the new transaction.

### API Impact

- [x] Public API: New feature (a new transaction type is added)

### Consideration

This NFT-related amendment arrives at about the same time as a different NFT-related amendment: https://github.com/XRPLF/rippled/pull/4946

Should these two pull requests be merged into a single amendment?


src/ripple/app/tx/impl/InvariantCheck.cpp
```diff
@@ -633,6 +663,22 @@ ValidNFTokenPage::finalize(
         return false;
     }
 
+    // Only enforce these invariants if fixNFTokenPageLinks is enabled.
```
seelabs, line 666: I prefer the code without the comment. The if statement below is not confusing.

src/ripple/app/tx/impl/LedgerStateFix.cpp
```diff
@@ -0,0 +1,99 @@
...
+    if (ctx.tx.getFlags() & tfUniversalMask)
+        return temINVALID_FLAG;
+
+    if (auto const ret = preflight1(ctx); !isTesSuccess(ret))
+        return ret;
+
+    switch (ctx.tx[sfLedgerFixType])
+    {
+        case FixType::nfTokenPageLink:
+            // FixType::nfTokenPageLink requires an Owner field.
```
seelabs, line 46: I prefer the code without the comment. We can see from the case statement immediately above that this is "FixType::nfTokenPageLink" and the code immediately below that it requires an owner field.

src/ripple/app/tx/impl/details/NFTokenUtils.cpp
```diff
@@ -428,10 +428,48 @@ removeToken(
         return tesSUCCESS;
     }
 
-    // The page is empty, so we can just unlink it and then remove it.
     if (prev)
     {
-        // Make our previous page point to our next page:
+        // With fixNFTokenPageLinks...
+        // The page is empty and there is a prev.  If the last page of the
```
seelabs, line 434: This comment is double spaced (I noticed the others were too). I don't think double spacing is very common, but I don't object too strongly if you prefer it.

src/ripple/app/tx/impl/details/NFTokenUtils.cpp
```diff
@@ -636,5 +674,139 @@ deleteTokenOffer(ApplyView& view, std::shared_ptr<SLE> const& offer)
     return true;
 }
 
+bool
+repairNFTokenDirectoryLinks(ApplyView& view, AccountID const& owner)
+{
+    // Assume that no repair is required.
```
seelabs, line 680: I'm seeing lots of comments of like this - a comment and then a line of code immediately below it that doesn't need the comment. I'll tag the ones I'd consider removing, but I'll stop saying "because it mirrors the line of code below it and the code is better without things that don't pull their weight".

src/ripple/app/tx/impl/details/NFTokenUtils.cpp
```diff
@@ -636,5 +674,139 @@ deleteTokenOffer(ApplyView& view, std::shared_ptr<SLE> const& offer)
...
+    bool didRepair = false;
+
+    // Walk from the first NFTokenPage owned by owner to the last
```
seelabs, line 683: Consider removing.

src/ripple/app/tx/impl/details/NFTokenUtils.cpp
```diff
@@ -636,5 +674,139 @@ deleteTokenOffer(ApplyView& view, std::shared_ptr<SLE> const& offer)
...
+    // find along the way.
+    auto const last = keylet::nftpage_max(owner);
+
+    std::shared_ptr<SLE> page = view.peek(Keylet(
+        ltNFTOKEN_PAGE,
+        view.succ(keylet::nftpage_min(owner).key, last.key.next())
+            .value_or(last.key)));
+
+    // If no page was found then owner has no NFToken pages that might need
+    // repair.
```
seelabs, line 694: Consider removing.

src/ripple/app/tx/impl/details/NFTokenUtils.cpp
```diff
@@ -636,5 +674,139 @@ deleteTokenOffer(ApplyView& view, std::shared_ptr<SLE> const& offer)
...
+
+    if (page->key() == last.key)
+    {
+        // There is one and only one page in this entire directory.  There
+        // should be no links on that page.
+        bool const nextPresent = page->isFieldPresent(sfNextPageMin);
+        bool const prevPresent = page->isFieldPresent(sfPreviousPageMin);
+        if (nextPresent || prevPresent)
+        {
+            // This page should have no links but does.  Remove the links.
```
seelabs, line 706: Consider removing.

src/ripple/app/tx/impl/details/NFTokenUtils.cpp
```diff
@@ -636,5 +674,139 @@ deleteTokenOffer(ApplyView& view, std::shared_ptr<SLE> const& offer)
...
+
+    // If no page was found then owner has no NFToken pages that might need
+    // repair.
+    if (!page)
+        return didRepair;
+
+    if (page->key() == last.key)
+    {
+        // There is one and only one page in this entire directory.  There
+        // should be no links on that page.
```
seelabs, line 701: Consider rewriting "The only page in this directory"

src/ripple/app/tx/impl/details/NFTokenUtils.cpp
```diff
@@ -636,5 +674,139 @@ deleteTokenOffer(ApplyView& view, std::shared_ptr<SLE> const& offer)
...
+                page->makeFieldAbsent(sfPreviousPageMin);
+            if (nextPresent)
+                page->makeFieldAbsent(sfNextPageMin);
+            view.update(page);
+        }
+        return didRepair;
+    }
+
+    // There's a first page that is not the same as last page.  The first page
+    // should not contain a previous link.
```
seelabs, line 718: Consider rewriting: "first page that is not the same as the last page"

src/ripple/app/tx/impl/details/NFTokenUtils.cpp
```diff
@@ -636,5 +674,139 @@ deleteTokenOffer(ApplyView& view, std::shared_ptr<SLE> const& offer)
...
+    if (page->isFieldPresent(sfPreviousPageMin))
+    {
+        // Remove the unneeded previous link.
```
seelabs, line 721: Consider removing.

src/ripple/app/tx/impl/details/NFTokenUtils.cpp
```diff
@@ -636,5 +674,139 @@ deleteTokenOffer(ApplyView& view, std::shared_ptr<SLE> const& offer)
...
+        didRepair = true;
+        page->makeFieldAbsent(sfPreviousPageMin);
+        view.update(page);
+    }
+
+    // For every page between the first and last...
```
seelabs, line 727: Consider removing.

src/ripple/app/tx/impl/details/NFTokenUtils.cpp
```diff
@@ -636,5 +674,139 @@ deleteTokenOffer(ApplyView& view, std::shared_ptr<SLE> const& offer)
...
+    std::shared_ptr<SLE> nextPage;
+    while (
+        (nextPage = view.peek(Keylet(
+             ltNFTOKEN_PAGE,
+             view.succ(page->key().next(), last.key.next())
+                 .value_or(last.key)))))
+    {
+        // page should have a link to nextPage.
```
seelabs, line 735: Consider removing.

src/ripple/app/tx/impl/details/NFTokenUtils.cpp
```diff
@@ -636,5 +674,139 @@ deleteTokenOffer(ApplyView& view, std::shared_ptr<SLE> const& offer)
...
+        if (!page->isFieldPresent(sfNextPageMin) ||
+            page->getFieldH256(sfNextPageMin) != nextPage->key())
+        {
+            // Repair the link to nextPage.
```
seelabs, line 739: Consider removing.

src/ripple/app/tx/impl/details/NFTokenUtils.cpp
```diff
@@ -636,5 +674,139 @@ deleteTokenOffer(ApplyView& view, std::shared_ptr<SLE> const& offer)
...
+            didRepair = true;
+            page->setFieldH256(sfNextPageMin, nextPage->key());
+            view.update(page);
+        }
+
+        // nextPage should have a link to page;
```
seelabs, line 745: Consider removing.

src/ripple/app/tx/impl/details/NFTokenUtils.cpp
```diff
@@ -636,5 +674,139 @@ deleteTokenOffer(ApplyView& view, std::shared_ptr<SLE> const& offer)
...
+        if (!nextPage->isFieldPresent(sfPreviousPageMin) ||
+            nextPage->getFieldH256(sfPreviousPageMin) != page->key())
+        {
+            // Repair the link to page.
```
seelabs, line 749: Consider removing.

src/ripple/app/tx/impl/details/NFTokenUtils.cpp
```diff
@@ -636,5 +674,139 @@ deleteTokenOffer(ApplyView& view, std::shared_ptr<SLE> const& offer)
...
+            didRepair = true;
+            nextPage->setFieldH256(sfPreviousPageMin, page->key());
+            view.update(nextPage);
+        }
+
+        if (nextPage->key() == last.key)
+            // We need special handling for the last page.
+            break;
+
+        // Continue to walk the links.
```
seelabs, line 759: Consider removing.

src/ripple/app/tx/impl/details/NFTokenUtils.cpp
```diff
@@ -636,5 +674,139 @@ deleteTokenOffer(ApplyView& view, std::shared_ptr<SLE> const& offer)
...
+            view.update(newPrev);
+        }
+        view.erase(page);
+        view.insert(nextPage);
+        return didRepair;
+    }
+
+    // The lastPage is present in the ledger and has the right index.  There's
+    // also a previous page.  Make sure that the last page does not have a
+    // next link.
```
seelabs, line 801: Consider removing.

src/ripple/protocol/Feature.h
```diff
@@ -74,7 +74,7 @@ namespace detail {
 // Feature.cpp. Because it's only used to reserve storage, and determine how
 // large to make the FeatureBitset, it MAY be larger. It MUST NOT be less than
 // the actual number of amendments. A LogicError on startup will verify this.
-static constexpr std::size_t numFeatures = 68;
+static constexpr std::size_t numFeatures = 69;
```
seelabs, line 77: Just a reminder that this needs to be incremented when the other amendment is merged. We won't get a merge conflict, since they change to the same number.

seelabs (approved):
:+1: Nice job on this!

You're probably going to be unhappy with the number of comments that I tagged as "consider removing", but in almost all cases they are comments where the comment mirrors a single line of code and doesn't add any benefit (and we don't want things in the code that don't pull their weight).

But :+1: no matter how you decide to resolve the comments, it's your call.

src/ripple/app/tx/impl/details/NFTokenUtils.cpp
```diff
@@ -636,5 +674,122 @@ deleteTokenOffer(ApplyView& view, std::shared_ptr<SLE> const& offer)
...
+        return didRepair;
+    }
+
+    // First page is not the same as last page.  The first page should not
+    // contain a previous link.
+    if (page->isFieldPresent(sfPreviousPageMin))
+    {
+        didRepair = true;
+        page->makeFieldAbsent(sfPreviousPageMin);
+        view.update(page);
```
shawnxie999, line 716: I remember Nik mentioned that the nodes at the end of the doubly linked list should point to the other end. Would we want to make `sfPreviousPageMin` of the first page point to the `last` page? 

Nik's comment:
https://github.com/XRPLF/rippled/pull/4346#discussion_r1026600781

Though I don't think this is actually enforced in the current NFTPage code...

src/ripple/app/tx/impl/details/NFTokenUtils.cpp
```diff
@@ -636,5 +674,122 @@ deleteTokenOffer(ApplyView& view, std::shared_ptr<SLE> const& offer)
...
+             ltNFTOKEN_PAGE,
+             view.succ(page->key().next(), last.key.next())
+                 .value_or(last.key)))))
+    {
+        if (!page->isFieldPresent(sfNextPageMin) ||
+            page->getFieldH256(sfNextPageMin) != nextPage->key())
+        {
+            didRepair = true;
+            page->setFieldH256(sfNextPageMin, nextPage->key());
+            view.update(page);
```
shawnxie999, line 731: From my understanding, this code checks if `sfNextPageMin` of the current page links correctly to the next page.

But is this related to the address corruption bug? It seems to me this has nothing to do with the bug...

src/ripple/app/tx/impl/details/NFTokenUtils.cpp
```diff
@@ -636,5 +674,122 @@ deleteTokenOffer(ApplyView& view, std::shared_ptr<SLE> const& offer)
...
+    // contain a previous link.
+    if (page->isFieldPresent(sfPreviousPageMin))
+    {
+        didRepair = true;
+        page->makeFieldAbsent(sfPreviousPageMin);
+        view.update(page);
+    }
+
+    std::shared_ptr<SLE> nextPage;
+    while (
```
shawnxie999, line 720: would this while-loop require performance testing?

src/ripple/app/tx/impl/details/NFTokenUtils.cpp
shawnxie999: Im kind of curious if this code is here for safety, or is it because some page is actually mislinked?

src/ripple/app/tx/impl/details/NFTokenUtils.cpp
scottschurr: Hi @shawnxie999, great question.  There have definitely been times that I've wished that the first and last links in the list linked to each other.  But they don't and, as far as I know, never have.

The spec (https://github.com/XRPLF/XRPL-Standards/discussions/46) does not give any specifics about how `sfPreviousPageMin` and `NextPageMin` are handled.  So I think this is a case where "the code is the spec".  The existing code does not link the back of the list to the front of the list.  And the existing unit tests do not expect such links to exist.  This amendment does not, and should not, change that behavior.

src/ripple/app/tx/impl/details/NFTokenUtils.cpp
scottschurr: I put in the code that looks for missing links because that's what we actually see in the ledger.  I added the code that verifies the link is valid because I was here and the check was cheap to do.

I've seen no instances of pages that were linked with incorrect values in the wild.  I've only seen links that are missing.

src/ripple/app/tx/impl/details/NFTokenUtils.cpp
shawnxie999: When the owner has a `last` page, is that sufficient to deduce that this particular owner is not a "victim" of the broken link?  Is there still the need to execute this loop? I'm just wondering if that's a viable approach instead of iterating through the entire doubly linked `NFTokenPage` structure every time.

src/ripple/app/tx/impl/LedgerStateFix.cpp
```diff
@@ -0,0 +1,98 @@
...
+            return tefINVALID_LEDGER_FIX_TYPE;
+    }
+
+    return preflight2(ctx);
+}
+
+XRPAmount
+LedgerStateFix::calculateBaseFee(ReadView const& view, STTx const& tx)
+{
+    // The fee required for AccountDelete is one owner reserve.
```
shawnxie999, line 60: why does it say AccountDelete?

src/ripple/app/tx/impl/details/NFTokenUtils.cpp
scottschurr: Good question.  But no, the presence of a valid `last` page is not sufficient to know that the directory tree is undamaged.  In the two current on-ledger cases both of the directories have valid `last` pages and there is a single break between two pages in the middle of the directory.

The way this happens is that

1. The last page is removed.  This leaves the next page down as the new `last` page.  But that new `last` page does not have the official `last` index for this directory.
2. The directory owner adds an NFToken with an ID greater the index of the current `last` page.  This causes a new `last` page to be created with the correct `last` index.  But the new `last` page does not have a link to the previous page.  So the link is missing in the middle of the chain.

It's possible to have the case where the last page is simply missing.  And I also have a unit test where I repair that situation.  But what I've seen is that, typically, the link is missing in the middle of the chain.

It would be possible for there to be more than one broken link in the chain.  But I've not seen that in the wild.

Does that make sense?  Questions?

src/ripple/app/tx/impl/LedgerStateFix.cpp
scottschurr: Good eyes!  That's a copy-and-paste-o and needs to be fixed.

src/ripple/app/tx/impl/details/NFTokenUtils.cpp
shawnxie999: Thanks for the detailed explanation! This makes a lot of sense!

shawnxie999 (approved):
looks good other than a minor typo 😃

exunda:
Why do you use the custom transaction method instead of doing this automatically which is what happens in `activateTrustLinesToSelfFix` which is called when the amendment is activated? Are you worried that more corrupted pages will create before the amendment activates?

scottschurr:
@exunda asked:
> Why do you use the custom transaction method instead of doing this automatically which is what happens in activateTrustLinesToSelfFix which is called when the amendment is activated?

That's a fair question.  It was answered in the (admittedly long) description of the pull request.  Here's the quote:
> We could put in code that cleans up the two known corrupted directories when the amendment goes live. That's not a good approach for two reasons:
>
> 1. As mentioned before, additional directories could face similar corruption between now and when the amendment goes live. We would like to be able to clean up all damaged NFToken directories, regardless of when they are created.
> 2. The MainNet is not the only place running the XRP Ledger code. Other networks may have similar corruptions that need to be fixed. The approach to the fix should be applicable to other networks as well.
>
> We need a transaction to clean up the directories. There are no transactions that just naturally walk an NFToken directory. So we can't leverage an existing transaction. There are RPC commands that walk the directory, but RPC commands can't modify ledger state.

Does that answer your question?

intelliot:
Due to changes in `develop`, there are now merge conflicts. @scottschurr would it make sense for you to bring this branch up-to-date?

scottschurr:
@intelliot, I'll work on that merge.

ximinez:
Once the merge is done, will this PR be ready to merge?

scottschurr:
@ximinez, the merge is almost done -- I need to fix some clang formatting.  Gahh, merges across the restructure are painful.

Once clang-format is fixed (soon) I think it _could_ be merged to develop.  It has two approvals.  But, given that it's an amendment, it would be nice if there were one more review from a senior developer.  I had not marked the branch as passed because I was waiting for that additional review.

src/xrpld/app/tx/detail/LedgerStateFix.cpp
```diff
@@ -0,0 +1,99 @@
...
+        return temINVALID_FLAG;
+
+    if (auto const ret = preflight1(ctx); !isTesSuccess(ret))
+        return ret;
+
+    switch (ctx.tx[sfLedgerFixType])
+    {
+        case FixType::nfTokenPageLink:
+            if (!ctx.tx.isFieldPresent(sfOwner))
+                return temINVALID;
```
ximinez, line 47: This line isn't covered. Can you add a test that removes the `sfOwner` from a `LedgerStateFix` tx?

src/xrpld/app/tx/detail/NFTokenUtils.cpp
```diff
@@ -637,6 +675,123 @@ deleteTokenOffer(ApplyView& view, std::shared_ptr<SLE> const& offer)
...
+                    " cannot be repaired.  Unexpected link problem.");
+            newPrev->at(sfNextPageMin) = nextPage->key();
+            view.update(newPrev);
+        }
+        view.erase(page);
+        view.insert(nextPage);
+        return didRepair;
+    }
+
+    if (nextPage->isFieldPresent(sfNextPageMin))
```
ximinez, line 786:
```suggestion
    assert(nextPage);
    if (nextPage->isFieldPresent(sfNextPageMin))
```

ximinez (changes requested):

scottschurr (commented):
Thanks for the review, @ximinez!  They were all good suggestions.  I got to most of them, but I ran out of time on a couple that I noted.  Maybe someone else can do those?  I think it's also not the end of the world if those changes to the tests don't happen.

Thanks again!

src/test/app/FixNFTokenPageLinks_test.cpp
```diff
@@ -0,0 +1,631 @@
...
+        {
+            auto bobLastNFTokenPage = env.le(keylet::nftpage_max(bob));
+            if (!BEAST_EXPECT(bobLastNFTokenPage))
+                return;
+
+            BEAST_EXPECT(bobLastNFTokenPage->isFieldPresent(sfPreviousPageMin));
+            BEAST_EXPECT(
+                bobLastNFTokenPage->at(sfPreviousPageMin) !=
+                bobMiddleNFTokenPageIndex);
+            BEAST_EXPECT(!bobLastNFTokenPage->isFieldPresent(sfNextPageMin));
```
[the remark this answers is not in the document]
ximinez: I went ahead and pushed this change. Someone else will need to reverse it.

src/test/app/FixNFTokenPageLinks_test.cpp
```diff
@@ -0,0 +1,631 @@
...
+                carolMiddleNFTokenPage->isFieldPresent(sfPreviousPageMin));
+            BEAST_EXPECT(carolMiddleNFTokenPage->isFieldPresent(sfNextPageMin));
+        }
+        // carol has a "last" page that includes a PreviousPageMin field.
+        {
+            auto carolLastNFTokenPage = env.le(keylet::nftpage_max(carol));
+
+            BEAST_EXPECT(
+                carolLastNFTokenPage->isFieldPresent(sfPreviousPageMin));
+            BEAST_EXPECT(!carolLastNFTokenPage->isFieldPresent(sfNextPageMin));
```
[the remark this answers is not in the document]
ximinez: I went ahead and pushed this change. Someone else will need to reverse it.

shawnxie999 (approved):
still looks good

ximinez (approved):
Since @scottschurr is unavailable to give final sign off, I'll do it: This is ready to merge.

PR #4946, opened 2024-03-09
Add the fixEnforceNFTokenTrustline amendment:

### LINKED ISSUE #4925: NFT issuer is inadvertently creating a trustline due to third-party transactions(Version: 2.1.0)
## Issue Description
Since the NFTokenAcceptOffer does not check the trust line that the issuer receives as a transfer fee in the NFTokenAcceptOffer, if the issuer deletes the trust line after NFTokenCreateOffer, the trust line is created for the issuer by the NFTokenAcceptOffer.

## Steps to Reproduce

1. Issuer has Trustline (ie: USD)
2. Mint NFToken with transfer fee
3. Transfer NFToken to Alice.
4. Alice create NFTokenOffer to bob with 10 USD
5. Issuer delete USD Trustline
6. Accept NFTokenOffer

## Expected Result

NFTokenAcceptOffer transaction fails with `terNO_LINE`

## Actual Result

NFTokenAcceptOffer transaction succeeds and a trustline is created for the issuer.

## Environment
rippled 2.1.0

## Supporting Files

https://gist.github.com/tequdev/f5cfd3f3cfbfcf4a22205052962e6cfe

### LINKED ISSUE #4941: If same account issues an IOU asset and an NFToken, that account's IOU cannot be used to pay the fee for that account's NFTs
## Issue Description

If an NFToken has a transfer fee then, when an offer is accepted, a portion of the sale price goes to the issuer.

It is possible for an issuer to issue both an IOU (for remittances) and NFTokens.  If the issuer's IOU is used to pay for the transfer of one of the issuer's NFTokens, that transfer will fail.

The problem occurs because the NFT code looks for a trust line.  However the issuer of an IOU does not need a trust line to accept their own issuance and, in fact, is not allowed to have a trust line to themselves.

## Steps to Reproduce

1. Issuer issues an IOU (e.g, USD).
2. Becky and Cheri get trust lines for Issuer's USD and acquire some.
3. Issuer mints NFToken with transfer fee.
4. Becky acquires the NFToken, paying with XRP.
5. Becky attempts to create an offer to sell the NFToken for Issuer's USD.  The attempt fails with `tecNO_LINE`.

## Expected Result

Becky should be able to create the offer and Cheri should be able to accept the offer, paying with Issuer's USD.

## Actual Result

Becky's `NFTokenOfferCreate` transaction fails with `tecNO_LINE`.

## Environment
rippled 2.1.0

scottschurr:

Introduces the `fixEnforceNFTokenTrustline ` amendment which fixes bugs in interactions between NFTokenOffers and trust lines.

Resolves #4925.
Resolves #4941.

#### #4925

This issue describes a wide timing window that causes a rare bug.  To understand the process you need some background in how NFTokens work.

When an NFToken is minted, it may have a transfer fee associated with it.  If that transfer fee is non-zero, then a portion of the funds paid for the NFToken's transfer are syphoned off for the NFT's issuer.  That transfer fee is always paid in the same funds as were used for the transfer.

The `NFTokenCreateOffer` transaction verifies that,

- If the NFToken has a transfer fee and
- The funds for the transfer are an IOU (not XRP) then
- The NFToken issuer must have a trust line for the funds.  

If the issuer does not have the necessary trust line then the `NFTokenCreateOffer` fails.

But here's the timing window.  When the `NFTokenCreateOffer` happens no funds are transferred.  The funds are transferred when the corresponding `NFTokenAcceptOffer` transaction completes.  But `NFTokenAcceptOffer` does not verify the presence of the trust line.

With that background, here are the steps for the bug:

1. Alice creates a trust line for USD.
2. Alice mints an NFToken with a non-zero transfer fee.
3. Becky acquires the NFToken from Alice.
4. Becky creates an offer to sell the NFToken for USD(100).
5. Alice removes her trust line for USD.
6. Carol accepts Becky's offer for the NFToken and pays in USD.
7.  Alice's trust line is recreated and receives the transfer fee in USD.

The bug appears in step 7, where the trust line that Alice deleted in step 5 is recreated.

The fix is simply to put a check in the `NFTokenAcceptOffer` transaction that is similar to the check in `NFTokenCreateOffer`.  But the check is transaction changing, so it must be guarded by an amendment.

#### #4941

This issue describes an even rarer, but related bug.

The checks described in the previous issue involve looking for a trust line connected to the NFToken issuer.  That _almost_ always works.  But there's one situation where it doesn't.  An IOU issuer can always accept any IOU that they issue.  And there is no trust line to represent that capability.  In fact, having a trust line from an account to itself is forbidden.

Here are the steps for the bug:

1. Alice issues USD
2. Alice mints an NFToken with a non-zero transfer fee.
3. Becky acquires the NFToken from Alice.
4. Becky attempts to creates an offer to sell the NFToken for Alice's USD(100).  This attempt fails with `tecNO_LINE`

The bug appears in step 4, where Becky cannot create the offer.

The fix is simply to check for the presence of the trust line _or_ if the IOU's issuer is the same as the NFToken issuer.  But the check is transaction changing, so it must be guarded by an amendment.

### Type of Change

- [x] Tests (unit tests for these changes were added)

### Consideration

This NFT-related amendment arrives at about the same time as a different NFT-related amendment: https://github.com/XRPLF/rippled/pull/4945

Should these two pull requests be merged into a single amendment?

## Proposed Commit Message
```
Add the fixEnforceNFTokenTrustline amendment:

Fix bugs in interactions between NFTokenOffers and trust lines.

Since the NFTokenAcceptOffer does not check the trust line that
the issuer receives as a transfer fee in the NFTokenAcceptOffer,
if the issuer deletes the trust line after NFTokenCreateOffer,
the trust line is created for the issuer by the
NFTokenAcceptOffer.  That's fixed.

Also if Alice issues an IOU and also mints NFTokens with a
transfer fee, Alice's IOU cannot be used to pay for transfers
of those NFTokens.  That's fixed.

Resolves #4925.
Resolves #4941.
```

### DIFF
--- a/src/ripple/app/tx/impl/NFTokenAcceptOffer.cpp
+++ b/src/ripple/app/tx/impl/NFTokenAcceptOffer.cpp
@@ -270,6 +270,28 @@ NFTokenAcceptOffer::preclaim(PreclaimContext const& ctx)
         }
     }
 
+    // Fix a bug where the transfer of an NFToken with a transfer fee could
+    // give the NFToken issuer an undesired trust line.
+    if (ctx.view.rules().enabled(fixEnforceNFTokenTrustline))
+    {
+        std::shared_ptr<SLE const> const& offer = bo ? bo : so;
+        if (!offer)
+            // Should be caught in preflight.
+            return tecINTERNAL;
+
+        uint256 const& tokenID = offer->at(sfNFTokenID);
+        STAmount const& amount = offer->at(sfAmount);
+        if (nft::getTransferFee(tokenID) != 0 &&
+            (nft::getFlags(tokenID) & nft::flagCreateTrustLines) == 0 &&
+            !amount.native())
+        {
+            auto const issuer = nft::getIssuer(tokenID);
+            // Issuer doesn't need a trust line to accept their own currency.
+            if (issuer != amount.getIssuer() &&
+                !ctx.view.read(keylet::line(issuer, amount.issue())))
+                return tecNO_LINE;
+        }
+    }
     return tesSUCCESS;
 }
 
--- a/src/ripple/protocol/Feature.h
+++ b/src/ripple/protocol/Feature.h
@@ -80,7 +80,7 @@ namespace detail {
 // Feature.cpp. Because it's only used to reserve storage, and determine how
 // large to make the FeatureBitset, it MAY be larger. It MUST NOT be less than
 // the actual number of amendments. A LogicError on startup will verify this.
-static constexpr std::size_t numFeatures = 75;
+static constexpr std::size_t numFeatures = 76;
 
 /** Amendments that this server supports and the default voting behavior.
    Whether they are enabled depends on the Rules defined in the validated
@@ -368,6 +368,7 @@ extern uint256 const fixPreviousTxnID;
 extern uint256 const fixAMMv1_1;
 extern uint256 const featureNFTokenMintOffer;
 extern uint256 const fixReducedOffersV2;
+extern uint256 const fixEnforceNFTokenTrustline;
 
 }  // namespace ripple
 
--- a/src/ripple/protocol/impl/Feature.cpp
+++ b/src/ripple/protocol/impl/Feature.cpp
@@ -495,6 +495,7 @@ REGISTER_FIX    (fixPreviousTxnID,              Supported::yes, VoteBehavior::De
 REGISTER_FIX    (fixAMMv1_1,                    Supported::yes, VoteBehavior::DefaultNo);
 REGISTER_FEATURE(NFTokenMintOffer,              Supported::yes, VoteBehavior::DefaultNo);
 REGISTER_FIX    (fixReducedOffersV2,            Supported::yes, VoteBehavior::DefaultNo);
+REGISTER_FIX    (fixEnforceNFTokenTrustline,    Supported::yes, VoteBehavior::DefaultNo);
 
 // The following amendments are obsolete, but must remain supported
 // because they could potentially get enabled.
--- a/src/test/app/NFToken_test.cpp
+++ b/src/test/app/NFToken_test.cpp
@@ -7417,6 +7417,176 @@ class NFTokenBaseUtil_test : public beast::unit_test::suite
     }
 
     void
+    testUnaskedForAutoTrustline(FeatureBitset features)
+    {
+        testcase("Test fix unasked for auto-trustline.");
+
+        using namespace test::jtx;
+
+        Account const issuer{"issuer"};
+        Account const becky{"becky"};
+        Account const cheri{"cheri"};
+        Account const gw("gw");
+        IOU const gwAUD(gw["AUD"]);
+
+        // This test case covers issue...
+        // https://github.com/XRPLF/rippled/issues/4925
+        //
+        // For an NFToken with a transfer fee, the issuer must be able to
+        // accept the transfer fee or else a transfer should fail.  If the
+        // NFToken is transferred for a non-XRP asset, then the issuer must
+        // have a trustline to that asset to receive the fee.
+        //
+        // This test looks at a situation where issuer would get a trustline
+        // for the fee without the issuer's consent.  Here are the steps:
+        //  1. Issuer has a trustline (i.e., USD)
+        //  2. Issuer mints NFToken with transfer fee.
+        //  3. Becky acquires the NFToken, paying with XRP.
+        //  4. Becky creates offer to sell NFToken for USD(100).
+        //  5. Issuer deletes trustline for USD.
+        //  6. Carol buys NFToken from Becky for USD(100).
+        //  7. The transfer fee from Carol's purchase re-establishes issuer's
+        //     USD trustline.
+        //
+        // The fixEnforceNFTokenTrustline amendment addresses this oversight.
+        //
+        // We run this test case both with and without
+        // fixEnforceNFTokenTrustline enabled so we can see the change
+        // in behavior.
+        //
+        // In both cases we remove the fixRemoveNFTokenAutoTrustLine amendment.
+        // Otherwise we can't create NFTokens with tfTrustLine enabled.
+        FeatureBitset const localFeatures =
+            features - fixRemoveNFTokenAutoTrustLine;
+        for (FeatureBitset feats :
+             {localFeatures - fixEnforceNFTokenTrustline,
+              localFeatures | fixEnforceNFTokenTrustline})
+        {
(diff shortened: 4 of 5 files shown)


src/ripple/app/tx/impl/NFTokenAcceptOffer.cpp
```diff
@@ -270,6 +270,34 @@ NFTokenAcceptOffer::preclaim(PreclaimContext const& ctx)
...
+        // - If the NFToken has a transfer fee, and
+        // - If the NFToken doesn't have the flagCreateTrustLines flag set, and
+        // - If the Amount is not XRP, and
+        // - If the NFToken issuer is not the Amount issuer, and
+        // - If the NFToken issuer does not have a trust line for the Amount
+        // - Then reject the token accept.
+        uint256 const& tokenID = offer->at(sfNFTokenID);
+        STAmount const& amount = offer->at(sfAmount);
+        if (nft::getTransferFee(tokenID) != 0 &&
+            (nft::getFlags(tokenID) & nft::flagCreateTrustLines) == 0 &&
```
shawnxie999, line 291: Since the flagCreateTrustLines flag has been deprecated, would there be any consequences if we don't add this check? 

I think checking for the flag makes sense, I would just like to learn more about what kind of scenario this check would guard against. I'm thinking that some nodes may decide to enable `NonFungibleTokensV1_1` and `fixNFTokenTrustlineSurprise` amendments, but not enabling
 `fixRemoveNFTokenAutoTrustLine` for some reason? Just guessing here.

dangell7:
I thought the recreation of the trustline was a feature and that was the reason for `disallowIncomingTrustline`?

src/ripple/app/tx/impl/NFTokenAcceptOffer.cpp
scottschurr: Hi @shawnxie999, great question.  The `fixRemoveNFTokenAutoTrustLine` amendment did not deprecate `nft::flagCreateTrustLines`.  But it did deprecate the transaction flag `tfTrustLine`.  So after the amendment went live on October 27, 2022, new NFTokens could not be minted with `nft::flagCreateTrustLines` set.  That code lives here: https://github.com/XRPLF/rippled/blob/develop/src/ripple/app/tx/impl/NFTokenMint.cpp#L56-L58

But there may be any number of NFTokens minted before October 27, 2022 that _do_ have the `nft::flagCreateTrustLines` flag set.  The code needs to continue to respect the flags on those older NFTokens.  That's why we need to continue to honor `nft::flagCreateTrustLines`.

shawnxie999: Thanks!

scottschurr:
Hi @dangell7, in this particular case the recreation of the trust line is a bug.  The NFToken in the bug report does not have the `nft::flagCreateTrustLines` flag set.  So the code should not be spontaneously adding trust lines to the issuer of the NFToken.

The test code that I added may be confusing.  Two NFTokens are created:

- One with `tfTrustLine` and
- One without.  

The NFToken created _with_ `tfTrustLine` is there to verify that my changes don't break the workflow when that trust line is _supposed_ to be created.  The one _without_ `tfTrustLine` tests the reported bug.

NFTokens have a lot of (too many?) options.  So it's easy to get tangled up.

Does that make sense?  Feel free to ask further questions.

shawnxie999 (approved):
Thanks for helping out with this! Looks good!

dangell7:
> Hi @dangell7, in this particular case the recreation of the trust line is a bug. The NFToken in the bug report does not have the `nft::flagCreateTrustLines` flag set. So the code should not be spontaneously adding trust lines to the issuer of the NFToken.
> 
> The test code that I added may be confusing. Two NFTokens are created:
> 
> * One with `tfTrustLine` and
> * One without.
> 
> The NFToken created _with_ `tfTrustLine` is there to verify that my changes don't break the workflow when that trust line is _supposed_ to be created. The one _without_ `tfTrustLine` tests the reported bug.
> 
> NFTokens have a lot of (too many?) options. So it's easy to get tangled up.
> 
> Does that make sense? Feel free to ask further questions.

Yepp I totally forgot about the flag on the NFToken.

src/ripple/app/tx/impl/NFTokenAcceptOffer.cpp
```diff
@@ -270,6 +270,34 @@ NFTokenAcceptOffer::preclaim(PreclaimContext const& ctx)
...
+        if (!offer)
+            // Should be caught in preflight.
+            return tecINTERNAL;
+
+        // - If the NFToken has a transfer fee, and
+        // - If the NFToken doesn't have the flagCreateTrustLines flag set, and
+        // - If the Amount is not XRP, and
+        // - If the NFToken issuer is not the Amount issuer, and
+        // - If the NFToken issuer does not have a trust line for the Amount
+        // - Then reject the token accept.
```
seelabs, line 287: Nit: The above comment doesn't clarify the code, it just repeats it. I prefer the code with out comment.

src/ripple/protocol/Feature.h
```diff
@@ -355,6 +355,7 @@ extern uint256 const fixFillOrKill;
 extern uint256 const fixNFTokenReserve;
 extern uint256 const fixInnerObjTemplate;
 extern uint256 const featurePriceOracle;
+extern uint256 const fixNFTokenTrustlineSurprise;
```
seelabs, line 358: That's a cute name, but it seems a little to jocular. Maybe `fixEnforceNFTokenTrustline`?

seelabs (approved):
:+1: Two nits, fix them (or not) at your discretion.

scottschurr:
The XLS-25d: NFTokenMintOffer commit ( https://github.com/XRPLF/rippled/commit/9f7c619e4fbbb0e110dc3fa6e9d13dad74b5830f) interacted a great deal with this pull request.

1. That commit absorbed the fix for issue #4941 that was previously in this pull request.
2. Similarly, that commit absorbed the unit tests for issue #4941 that were previously in this commit.

So I rebased this commit to the tip of develop and resolved the differences.  I also squashed the commits and changed the commit message to ...

1. Reflect the changed amendment name and
2. That this commit only addresses issue #4925 now.

I'm not marking this as passed yet, since I want to allow my reviewers an opportunity to check my work with the squash and rebase. But as far as I'm concerned I think this branch is ready.

seelabs (approved):
Still :+1: after latest changes

shawnxie999 (approved):
looks good 👍

PR #4947, opened 2024-03-09
Embed patched recipe for RocksDB 6.29.5

thejohnfreeman:
We depend on RocksDB 6.29.5. I don't know how we ever built a working package for it with MSVC, but when we removed it from the Ripple cache recently, CI could not build its replacement. This PR copies the recipe from Conan Center and adds a small patch to remove the statement `include(thirdparty.inc)` from its `CMakeLists.txt`, which fixes the build.


BUILD.md
```diff
@@ -147,6 +147,13 @@ which allows you to statically link it with GCC, if you want.
    conan export external/snappy snappy/1.1.10@
    ```
 
+Export our [Conan recipe for RocksDB](./external/rocksdb).
```
godexsoft, line 150: Are we forced to use a custom recipe again? If @Bronek had a branch that updates to a newer version without a custom recipe then that might be preferable, wdyt?

thejohnfreeman: I'd rather fix 6.29.5 before updating to 8+. It's easier to backport this recipe than the set of code changes that might require.

godexsoft (approved):

Bronek (approved):
I like it. Perhaps we need "release notes" for developers to run `conan remove rocksdb`  so they can properly replace it in conan cache, but I do not feel strongly about it. I like how you moved stuff from workflows to action.

BUILD.md
Bronek: I agree with @thejohnfreeman - that branch with upgrade to 8.8.1 is not yet ready, and when I think it is ready it might still cause fair amount of hassle (e.g. performance tests) and uncertainty, meaning it would be some time before we can use it.

PR #4948, opened 2024-03-13
Fix workflows

thejohnfreeman:
The problem was `CONAN_USERNAME` environment variable, which Conan 1.x uses as the default user in package references.

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -22,28 +22,27 @@ jobs:
       NUM_PROCESSORS: 12
     steps:
       - name: checkout
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
       - name: install Ninja
         if: matrix.generator == 'Ninja'
         run: brew install ninja
       - name: check environment
         run: |
+          env | sort
           echo ${PATH} | tr ':' '\n'
           python --version
           conan --version
           cmake --version
-          env | sort
       - name: configure Conan
         run : |
           conan profile get env.CXXFLAGS default || true
           conan profile update 'conf.tools.build:cxxflags+=["-DBOOST_ASIO_DISABLE_CONCEPTS"]' default
-          conan remote remove ripple
       - name: build dependencies
         uses: ./.github/actions/dependencies
         env:
           CONAN_URL: http://18.143.149.228:8081/artifactory/api/conan/conan-non-prod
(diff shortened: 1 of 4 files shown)


Bronek:
What's the purpose of `CONAN_PASSWORD_RIPPLE` and `CONAN_LOGIN_USERNAME_RIPPLE` ? I see that they are set when invoking `dependencies` action, the PR does explain why **not** use the name `CONAN_USERNAME`, but I do not see these new environment variables actually used. It looks as if they were used implicitly (in which case please add a comment inside `dependencies/action.yml`) or not used at all ?

Bronek (commented):
What's the purpose of `CONAN_PASSWORD_RIPPLE` and `CONAN_LOGIN_USERNAME_RIPPLE` ? I see that they are set when invoking `dependencies` action, the PR **does** explain why not use the name `CONAN_USERNAME`, but I do **not** see these new environment variables actually used. It looks as if they were used implicitly (in which case please add a comment inside `dependencies/action.yml`) or not used at all ?

Bronek:
> What's the purpose of `CONAN_PASSWORD_RIPPLE` and `CONAN_LOGIN_USERNAME_RIPPLE` ?

BTW I do understand that it is perfectly normal for an action to implicitly use environment variables. It's the `_RIPPLE` suffix which makes it confusing, because one possible use of such a suffix is to *prevent* such an implicit use of an environment variable.

Bronek (approved):
Thank you !

.github/workflows/macos.yml
```diff
@@ -22,28 +22,27 @@ jobs:
       NUM_PROCESSORS: 12
     steps:
       - name: checkout
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
```
mvadari, line 25: Can this update be made in the other workflows as well? v3 is deprecated.

ximinez (approved):
Thanks for chasing these issues down.

.github/workflows/macos.yml
thejohnfreeman: Yes, but because they are low priority, I will get to them in another PR so that this can be merged now.

thejohnfreeman: https://github.com/XRPLF/rippled/pull/4949

PR #4950, opened 2024-03-13
Add the`fixEmptyDID` amendment

mvadari:

This PR introduces an amendment that fixes a problem where DIDs can be empty, which shouldn't be able to happen. An empty DID object is equivalent to not having a DID object at all, so there is no need to take up the extra object reserve.

### Context of Change

If a`DIDSet` transaction is sent with one or two of the fields (`DIDDocument`/`Data`/`URI`) as an empty string, when there is no pre-existing DID, the `DID` object is created with empty fields.

Recreation: https://devnet.xrpl.org/transactions/A73FEA27EF6A24438A7E9A3B8192B1C0B430E503A14212933CAFF54C66C51FCD

### Type of Change

- [ ] Documentation update
- [ ] Release

## Test Plan

A unit test was added that reproduces the issue. The test ensures that behavior remains the same pre-amendment, and is the desired behavior post-amendment.

### DIFF
--- a/src/ripple/app/tx/impl/DID.cpp
+++ b/src/ripple/app/tx/impl/DID.cpp
@@ -161,6 +161,13 @@ DIDSet::doApply()
     set(sfURI);
     set(sfDIDDocument);
     set(sfData);
+    if (ctx_.view().rules().enabled(fixEmptyDID) &&
+        !sleDID->isFieldPresent(sfURI) &&
+        !sleDID->isFieldPresent(sfDIDDocument) &&
+        !sleDID->isFieldPresent(sfData))
+    {
+        return tecEMPTY_DID;
+    }
 
     return addSLE(ctx_, sleDID, account_);
 }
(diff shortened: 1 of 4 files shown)


seelabs:
This needs a better commit message. When opening a PR I'd like it to have the proposed commit message and that commit message should be part of the review (plus a good commit message helps the reviewers as well).

Many people do this (I'm sure I do this myself), but I'd like us to get out of the habit of doing.

seelabs (approved):
:+1: LGTM

gregtatcam (approved):
👍 LGTM

mvadari:
Proposed commit message:
```
`fixEmptyDID`: ensure that empty DIDs cannot be created (#4950)

This amendment fixes an edge case where an empty DID object can be
created. It adds an additional check to ensure that DIDs are
non-empty when created, and throws a `tecEMPTY_DID` error if the DID
would be empty.
```

mvadari:
@seelabs @ximinez this is ready to merge.

PR #4959, opened 2024-03-20
Elaborate Error Message for rpcInternal

ckeshava:

Validator operators are often flummoxed by the below error:

73-error

It would be helpful if the error message can instruct the users to run the rippled server in another process (`./rippled`) before executing any of the command-line client commands. 

### Type of Change

- [x] Documentation update
- [ ] Release

This is an elaboration of an error message.

### API Impact

No change to API. No impact on performance either.

I manually verified (on MacOS) that the new error message is displayed, if rippled server isn't running in the background.


thejohnfreeman (approved):
This feels like a consequence of lumping client and server into the same executable. :/

seelabs:
Needs to be reformatted.

intelliot (approved):

PR #4972, opened 2024-03-31
Fixed relative paths in BUILD.md

tequdev:

Fixed relative paths in BUILD.md

### Type of Change

- [x] Documentation update


thejohnfreeman:
Oh, that's not a relative path. That's an ellipsis that refers to the rest of the `conan install` command, the exact arguments to which depend on the builder's environment and goals. We'll need to either close this PR or find another way to convey the meaning of the ellipsis.

PR #4979, opened 2024-04-04
Proposed 2.2.0-b2

ximinez:

This is a beta for the 2.2.0 release.

Highlights:
- #4327
- #4781
- #4950
- #4763
- #4933

The base branch is `release`. [All releases (including betas)](https://github.com/XRPLF/rippled/blob/develop/CONTRIBUTING.md#before-you-start) go in `release`. This PR will be merged with `--ff-only` (not squashed or rebased, and not using the GitHub UI) to both `release` and `develop`.

### Context of Change

This introduces
- the fast base58 codec, which speeds up base58 encoding and decoding significantly
- a user version of the `feature` RPC command, allowing lookups without changing voting settings
- the `fixEmptyDID` amendment to prevent creation of empty DID objects
- the `fixXChainRewardRounding` amendment to improve the calculation of reward shares
- consensus improvements to reduce de-syncs

It also fixes up some technical debt and build issues, and fixes the 'install' target so that downstream projects like `clio` can build properly.

### Type of Change

- [x] Release

### API Impact

No API impact.


Bronek:
Important caveat - because of https://github.com/XRPLF/rippled/pull/4979 , we have added a conan dependency on `xxHash::xxhash`, which is **transitive** for users of `libxrpl` . We could hide this dependency if we think it's a problem (i.e. move implementation of `class xxhasher` from header to `.cpp` so the dependency `#include <xxhash.h>` does not show up in header file)

ximinez:
@Bronek What is the worst case for users of `libxrpl` once they bump their dependency to this version or later? I assume they'd have to re-run `conan install`. Maybe wipe out their build directory?

Bronek:
> @Bronek What is the worst case for users of `libxrpl` once they bump their dependency to this version or later? I assume they'd have to re-run `conan install`. Maybe wipe out their build directory?

re-running `conan install` will be definitely needed; luckily it does not rebuild anything that's already in conan cache, so it should be quick, no need to wipe anytihing. The bigger problem is for projects which depend in `libxrpl`, as they will also (transitively) gain this dependency, without necessarily knowing about it. So they will also have to re-run `conan install`

scottschurr (approved):

PR #4981, opened 2024-04-04
AMM Unit test: Create AMM Pools with tiny pool balances

ckeshava:
This PR introduces new unit tests to test small AMM Pool creation and tiny Payment transactions against AMMs. AMM pool sizes are in order of `1e-6` and payment transactions are in order of `1e-7`. Please let me know if I should test smaller ranges.

The behavior of the code matches our understanding for both XRP/IOU and IOU/IOU AMM pools. Many thanks to @gregtatcam  for helping with these tests.

### Type of Change

- [ ] Documentation update
- [ ] Release

### API Impact

No impact on performance. No changes to Public APIs either.

### DIFF
--- a/src/test/app/AMM_test.cpp
+++ b/src/test/app/AMM_test.cpp
@@ -2294,6 +2294,58 @@ private:
         using namespace jtx;
         using namespace std::chrono;
 
+        // burn all the LPTokens through a AMMBid transaction
+        {
+            Env env(*this);
+            fund(env, gw, {alice}, XRP(2'000), {USD(2'000)});
+            AMM amm(env, gw, XRP(1'000), USD(1'000), false, 1'000);
+
+            // auction slot is owned by the creator of the AMM i.e. gw
+            BEAST_EXPECT(amm.expectAuctionSlot(100, 0, IOUAmount{0}));
+
+            // gw attempts to burn all her LPTokens through a bid transaction
+            // this transaction fails because AMMBid transaction can not burn
+            // all the outstanding LPTokens
+            amm.bid(BidArg{
+                .account = gw,
+                .bidMin = 1'000'000,
+                .err = ter(tecAMM_INVALID_TOKENS)});
+        }
+
+        // burn all the LPTokens through a AMMBid transaction
+        {
+            Env env(*this);
+            fund(env, gw, {alice}, XRP(2'000), {USD(2'000)});
+            AMM amm(env, gw, XRP(1'000), USD(1'000), false, 1'000);
+
+            // auction slot is owned by the creator of the AMM i.e. gw
+            BEAST_EXPECT(amm.expectAuctionSlot(100, 0, IOUAmount{0}));
+
+            // gw burns all but one of her LPTokens through a bid transaction
+            // this transaction suceeds because the bid price is less than
+            // the total outstanding LPToken balance
+            amm.bid(BidArg{
+                .account = gw,
+                .bidMin = STAmount{amm.lptIssue(), UINT64_C(999'999)},
+                .err = ter(tesSUCCESS)});
+
+            // gw must posses the auction slot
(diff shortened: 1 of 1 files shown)


ckeshava:
@seelabs  can you please review this PR? I'm not able to add reviewers through the Github UI

seelabs:
I think I'd separate these tests out as a separate test case - I'm not sure they belong in "basic" tests. Maybe call it "extreme values test" or somesuch. I'd also tests every combination of the following amounts. The amounts should be used for both AMM pool balances and for transaction amounts (deposit/withdrawal/payments)

1. max allowed IOU 
2. 10^-9 IOU
3. min allowed IOU
4. 1 drop
5. 2 drops
6. 1 XRP
7. 10^7 XRP
8. 1 IOU
9. 10^7 IOU

I'd expect that you'd want to create these combinations programmatically.

ckeshava:
@seelabs  what would be the appropriate values for minimum and maximum allowed IOUs? `STAmount` represents numbers with a `mantissa` and an `exponent`. I believe these numbers can have 16 significant digits. So the max and min values would be `1e16` and `1e-16` ?

src/test/app/AMM_test.cpp
```diff
@@ -2986,6 +2985,208 @@ struct AMM_test : public jtx::AMMTest
...
+                    },
+                    ammAlice.balances(XRP, USD)));
+            });
+        }
+
+        // test AMM with tiny pool sizes and tiny payment amounts, XRP/IOU
+        // Initial AMM Pool composition (1 drop XRP, 1e-6 USD), 0 trading fees
+        // Swap transaction: Amount = 1e-7 USD, sendmax = 1 drop XRP
+        // Final expected pool composition: 2 drop XRP, 9e-7 USD
+        testAMM(
```
ckeshava, line 3126: I believe the two below unit tests are still useful. They validate small payment transactions against a small AMM pool. But if you feel otherwise (or if there are similar tests), I can remove them.

src/test/app/AMM_test.cpp
```diff
@@ -2986,6 +2985,208 @@ struct AMM_test : public jtx::AMMTest
...
+                    },
+                    ammAlice.balances(XRP, USD)));
+
+                // payment transaction: sendmax is the parameterised value
+                env(pay(bob, carol, STAmount{XRP, 1, 7}),
+                    sendmax(money),
+                    txflags(tfPartialPayment));
+                env.close();
+
+                BEAST_EXPECT(std::apply(
```
ckeshava, line 3098: If anyone has any ideas on how I can remove this ugly repetitive invariant checks, please let me know.

I'd like to perform the AMMInvariant check after every transaction. I couldn't get `BEAST_EXPECT(std::apply(AMMInvariantCheck, ammAlice.balances(XRP, USD)));` this to work.
The compiler complains about the need to use a non-static callable inside `std::apply` ://

src/test/app/AMM_test.cpp
```diff
@@ -2986,6 +2985,208 @@ struct AMM_test : public jtx::AMMTest
...
+    void
+    testExtremeValues() {
+        testcase("Extreme values test for AMM Pool Composition; Payments, "
+            "Deposits and Withdrawls against the AMM");
+
+        using namespace jtx;
+
+        // test the creation of AMM pools with a combination of the following
+        // initial XRP/USD balances
+        for (STAmount const& usd :
```
ckeshava, line 3002: @seelabs  is this what you meant by creating AMM pools with differing pool sizes? I'm not validating any other property of the AMMs except for their successful creation.

ckeshava:
FYI: When using IOUs with AMMDeposit, Payment and AMMWithdraw transactions, it is not feasible to use less than `1e-3` values. The individual transactions have differing limits, but this is the common denominator. `1e-3` appears to be the minimum value of IOUs

PR #4982, opened 2024-04-05
AMM Unit tests: rounding down of equal asset deposit LPToken calculation

ckeshava:
This PR exercises a specific portion of the AMMHelpers.cpp file. It triggers a downward rounding of LPTokens in `adjustLPTokens` function in the `AMMHelpers.cpp` file.

### Type of Change

- [ ] Documentation update
- [ ] Release

### API Impact

No impact on performance. No change to APIs


src/test/app/AMM_test.cpp
```diff
@@ -4934,30 +4934,30 @@ struct AMM_test : public jtx::AMMTest
...
+//        testFlags();
+//        testRippling();
+//        testAMMAndCLOB();
+//        testTradingFee();
+//        testAdjustedTokens();
+//        testAutoDelete();
+//        testClawback();
+//        testAMMID();
+//        testSelection();
+//        testFixDefaultInnerObj();
```
HowardHinnant, line 4960: Why comment out all of the tests?

ckeshava: my bad, fixed it

HowardHinnant:
This looks good to me.  But I was hoping it would hit these lines: https://app.codecov.io/gh/XRPLF/rippled/pull/4982/blob/src/ripple/app/misc/impl/AMMHelpers.cpp#L171. And I don't have a good explanation as to why it didn't.  Nevertheless, this looks like a good addition to me.

ckeshava:
@HowardHinnant  I have updated the numbers so that the execution hits the necessary lines.

I expected `10'000.1` to be rounded down to `10'000` and consequently hit the said line. I think my debugger is malfunctioning.

HowardHinnant (approved):

src/test/app/AMM_test.cpp
```diff
@@ -1128,6 +1128,46 @@ struct AMM_test : public jtx::AMMTest
...
+                                                 USD(10'000) + depositUSD,
+                                                 IOUAmount{1, 7} + newLPTokens));
+
+            // 30,000 less deposited depositUSD
+            BEAST_EXPECT(expectLine(env, carol, USD(30'000) - depositUSD));
+            // 30,000 less deposited depositXRP and 10 drops tx fee
+            BEAST_EXPECT(expectLedgerEntryRoot(env, carol, XRP(30'000) -
+                                                   depositXRP - txfee(env, 1)));
+        });
+
```
seelabs, line 1170: Maybe my brain isn't working today, but I'm not following how this tests rounding down. Where would the calculation differ if we rounded to nearest, say? Can you add some clarifying comments to help me follow this test?

src/ripple/app/misc/impl/AMMHelpers.cpp
```diff
@@ -153,6 +153,11 @@ adjustAmountsByLPTokens(
     std::uint16_t tfee,
     bool isDeposit)
 {
+    // If lpTokens contains a fractional part and at least 16 significant
+    // digits, it is possible that lpTokensActual is rounded down. This is to
+    // ensure that we don't create/burn more LPTokens (depending on
+    // whether it's AMMDeposit or AMMWithdraw transaction) than what the trader
+    // asked for.
```
ckeshava, line 160: @seelabs  This is one of the points in the codebase where LPTokens are rounded down. Check the function above `adjustLPTokens`.

Consequently, `lpTokensActual` < `lpTokens` by a very small margin.

src/test/app/AMM_test.cpp
ckeshava: The values computed by `AMMHelpers.cpp:adjustLPTokens` would be different, if we used the nearest rounding mode. I have tagged that line above ^

seelabs: I'm still not sure how the test is showing this difference. I think what I'm looking for is a comment in the test that explains how the test proves that the rounding is working as we intended. The test may be doing that, but I'm not sure how. I think I need a comment to explain how the test works.

HowardHinnant: Fwiw, I checked the test coverage of AMMHelpers.cpp before and after this PR and this PR improves the coverage in this file, specifically in `adjustAmountsByLPTokens`.

ckeshava: thanks howard.
@seelabs I have added a test case and asserts on line 1155 to show the rounding down behavior

src/test/app/AMM_test.cpp
```diff
@@ -1128,6 +1128,62 @@ struct AMM_test : public jtx::AMMTest
...
+                    STAmount{USD, deltaUSD.mantissa(), deltaUSD.exponent()};
+
+                const STAmount depositXRP =
+                    STAmount{XRP, deltaXRP.mantissa(), deltaXRP.exponent()};
+
+                // initial LPTokens (1e7) + newLPTokens
+                BEAST_EXPECT(ammAlice.expectBalances(
+                    XRP(10'000) + depositXRP,
+                    USD(10'000) + depositUSD,
+                    IOUAmount{1, 7} + newLPTokens));
```
ckeshava, line 1177: I don't understand how lines 1156 and 1177 can be true at the same. I'm guessing it's due to the loss of precision in the last few digits.

`newLPTokens` == `deltaLPTokens` (numerically)
`initLPTokens` + `newLPTokens` = `finalLPTokens`, but `finalLPToken - initLPToken < deltaLPTokens`

seelabs (approved):
:+1: nice job on this!

PR #4983, opened 2024-04-05
Add the fixAMMOfferRounding amendment

gregtatcam:

Introduces `fixAMMOfferRounding` amendment to improve quality of the generated AMM offer for a single path payment when one side of the offer is XRP and there is a Limit Order Book (LOB) offer for the same asset pair on the DEX.

### Context of Change

A single-path AMM offer with account offer on DEX, is always generated
starting with the takerPays first, which is rounded up, and then
the takerGets, which is rounded down. This rounding ensures that the pool's
product invariant is maintained. However, when one of the offer's side
is XRP, this rounding can result in the AMM offer having a lower
quality, potentially causing offer generation to fail if the quality
is lower than the account's offer quality.
    
To address this issue, the proposed fix adjusts the offer generation process
to start with the XRP side first and always rounds it down. This results
in a smaller offer size, improving the offer's quality.
This change still ensures the product invariant, as the other generated
side is the exact result of the swap-in or swap-out equations.

### Type of Change

- [ ] Documentation update
- [ ] Release

### API Impact

## Test Plan

A unit-test `testFixAMMOfferRounding()` is added to AMM_test to verify this amendment

### DIFF
--- a/src/ripple/app/misc/AMMHelpers.h
+++ b/src/ripple/app/misc/AMMHelpers.h
@@ -21,7 +21,9 @@
 #define RIPPLE_APP_MISC_AMMHELPERS_H_INCLUDED
 
 #include <ripple/basics/IOUAmount.h>
+#include <ripple/basics/Log.h>
 #include <ripple/basics/Number.h>
+#include <ripple/beast/utility/Journal.h>
 #include <ripple/protocol/AMMCore.h>
 #include <ripple/protocol/AmountConversions.h>
 #include <ripple/protocol/Feature.h>
@@ -147,12 +149,223 @@ withinRelativeDistance(Amt const& calc, Amt const& req, Number const& dist)
 }
 // clang-format on
 
-/** Finds takerPays (i) and takerGets (o) such that given pool composition
- * poolGets(I) and poolPays(O): (O - o) / (I + i) = quality.
- * Where takerGets is calculated as the swapAssetIn (see below).
- * The above equation produces the quadratic equation:
- * i^2*(1-fee) + i*I*(2-fee) + I^2 - I*O/quality,
- * which is solved for i, and o is found with swapAssetIn().
+/** Solve quadratic equation to find takerGets or takerPays. Round
+ * to minimize the amount in order to maximize the quality.
+ */
+std::optional<Number>
+solveQuadraticEqSmallest(Number const& a, Number const& b, Number const& c);
+
+/** Generate AMM offer starting with takerGets when AMM pool
+ * from the payment perspective is IOU(in)/XRP(out)
+ * Equations:
+ * Spot Price Quality after the offer is consumed:
+ *     Qsp = (O - o) / (I + i) (1)
+ *  where O is poolPays, I is poolGets, o is takerGets, i is takerPays
+ * Swap out:
+ *     i = (I * o) / (O - o) * f (2)
+ *  where f is (1 - tfee/100000), tfee is in basis points
+ * Effective price targetQuality:
+ *     Qep = o / i (3)
+ * There are two scenarios to consider
+ * A) Qsp = Qep. Substitute i in (1) with (2) and solve for o
+ *    and Qsp = targetQuality(Qt):
+ *     o**2 + o * (I * Qt * (1 - 1 / f) - 2 * O) + O**2 - Qt * I * O = 0
+ * B) Qep = Qsp. Substitute i in (3) with (2) and solve for o
+ *    and Qep = targetQuality(Qt):
+ *     o = O - I * Qt / f
+ * Since the scenario is not know a priori, both A and B are solved and
+ * the lowest value of o is takerGets. takerPays is calculated with
+ * swap out eq (2). If o is less or equal to 0 then the offer can't
+ * be generated.
+ */
+template <typename TIn, typename TOut>
+std::optional<TAmounts<TIn, TOut>>
+getAMMOfferStartWithTakerGets(
+    TAmounts<TIn, TOut> const& pool,
+    Quality const& targetQuality,
+    std::uint16_t const& tfee)
+{
+    assert(targetQuality.rate() != beast::zero);
+    // calculate a, b, c to minimize quadratic equation solution
+    // b is negative
+    // calculate:
+    // (-b - root2(b * b - 4 * a * c)) / (2 * a)
+    // minimize:
+    // (-b - root2(b * b  - 4 * a * c)) / (2 * a)
+    // b
+    // maximize:
+    // root2(b * b - 4 * a * c)
+    // b * b
+    // minimize:
+    // 4 * a * c
+    // a
+    // c
+    // maximize:
+    // 2 * a
+    // a
+    // a must be maximized and minimized. a is 1, no rounding
+    // b must be maximized and minimized. choose minimize since the effect of
+    // the rounding outside of root2 is greater than the inside of root2.
+    // fee is always maximized
+    // feeMult is always minimized
+    auto const fee = upward()(getFee(tfee));
+    auto const feeMult = downward()(1 - fee);
+    auto const a = 1;
+    // minimize b
+    // b = pool.in * (1 - 1 / f) / targetQuality.rate() - 2 * pool.out
+    // minimize:
+    // pool.in * (1 - 1 / f) / targetQuality.rate()
+    // pool.in * (1 - 1 / f)
+    // maximize:
+    // 1 / f
+    // 2 * pool.out
+    auto const b = downward()(
+        pool.in * (1 - upward()(1 / feeMult)) / targetQuality.rate() -
+        upward()(2 * pool.out));
+    // minimize c
+    // c = pool.out * pool.out - pool.in * pool.out / targetQuality.rate()
+    // minimize:
+    // pool.out * pool.out - pool.in * pool.out / targetQuality.rate()
+    // pool.out * pool.out
+    // maximize:
+    // pool.in * pool.out / targetQuality.rate()
+    // pool.in * pool.out
+    auto const c = downward()(
+        pool.out * pool.out -
+        upward()((pool.in * pool.out) / targetQuality.rate()));
+
+    auto nTakerGets = solveQuadraticEqSmallest(a, b, c);
+    if (!nTakerGets || *nTakerGets <= 0)
+        return std::nullopt;
+
+    // minimize constraint
+    // constraint = pool.out - pool.in / (targetQuality.rate() * f)
+    // minimize:
+    // pool.out - pool.in / (targetQuality.rate() * f)
+    // maximize:
+    // pool.in / (targetQuality.rate() * f)
+    // minimize:
+    // (targetQuality.rate() * f)
+    auto const nTakerGetsConstraint = downward()(
+        pool.out -
+        upward()(pool.in / downward()(targetQuality.rate() * feeMult)));
+    if (nTakerGetsConstraint <= 0)
+        return std::nullopt;
+
+    // Pick the smallest to make the quality better
+    if (nTakerGetsConstraint < nTakerGets)
+        nTakerGets = nTakerGetsConstraint;
+
+    auto const takerGets = toAmount<TOut>(
+        getIssue(pool.out), *nTakerGets, Number::rounding_mode::downward);
+
+    return TAmounts<TIn, TOut>{swapAssetOut(pool, takerGets, tfee), takerGets};
+}
+
+/** Generate AMM offer starting with takerPays when AMM pool
+ * from the payment perspective is XRP(in)/IOU(out) or IOU(in)/IOU(out).
+ * Equations:
+ * Spot Price Quality after the offer is consumed:
+ *     Qsp = (O - o) / (I + i) (1)
+ *  where O is poolPays, I is poolGets, o is takerGets, i is takerPays
+ * Swap in:
+ *     o = (O * i * f) / (I + i * f) (2)
+ *  where f is (1 - tfee/100000), tfee is in basis points
+ * Effective price quality:
+ *     Qep = o / i (3)
+ * There are two scenarios to consider
+ * A) Qsp = Qep. Substitute o in (1) with (2) and solve for i
+ *    and Qsp = targetQuality(Qt):
+ *     i**2 * f + i * I * (1 + f) + I**2 - I * O / Qt = 0
+ * B) Qep = Qsp. Substitute i in (3) with (2) and solve for i
+ *    and Qep = targetQuality(Qt):
+ *     i = O / Qt - I / f
+ * Since the scenario is not know a priori, both A and B are solved and
+ * the lowest value of i is takerPays. takerGets is calculated with
+ * swap in eq (2). If i is less or equal to 0 then the offer can't
+ * be generated.
+ */
+template <typename TIn, typename TOut>
+std::optional<TAmounts<TIn, TOut>>
+getAMMOfferStartWithTakerPays(
+    TAmounts<TIn, TOut> const& pool,
+    Quality const& targetQuality,
+    std::uint16_t tfee)
+{
+    // calculate a, b, c to minimize quadratic equation solution
+    // b is positive
+    // calculate:
+    // (-b + root2(b * b - 4 * a * c)) / (2 * a)
+    // minimize:
+    // (-b + root2(b * b - 4 * a * c))
+    // maximize
+    // b
+    // minimize:
+    // root2(b * b - 4 * a * c)
+    // b * b
+    // maximize:
+    // 4 * a * c
+    // a
+    // c
+    // maximize:
+    // 2 * a
+    // a
+    // b must be maximized and minimized. choose maximize since the effect of
+    // the rounding outside of root2 is greater than the inside of root2.
+    // fee is always maximized
+    // feeMult is always minimized
+    auto const fee = upward()(getFee(tfee));
+    auto const f = downward()(1 - fee);
+    auto const& a = f;
+    // maximize b
+    auto const b = upward()(pool.in * (1 + f));
+    // maximize c
+    // c = pool.in * pool.in - pool.in * pool.out * targetQuality.rate()
+    // maximize:
+    // pool.in * pool.in
+    // minimize:
+    // pool.in * pool.out * targetQuality.rate()
+    auto const c = upward()(
+        pool.in * pool.in -
+        downward()(pool.in * pool.out * targetQuality.rate()));
+
+    auto nTakerPays = solveQuadraticEqSmallest(a, b, c);
+    if (!nTakerPays || nTakerPays <= 0)
+        return std::nullopt;
+
+    // minimize constraint
+    // constraint = pool.out * targetQuality.rate() - pool.in / f
+    // minimize:
+    // pool.out * targetQuality.rate() - pool.in / f
+    // pool.out * targetQuality.rate()
+    // maximize:
+    // pool.in / f
+    auto const nTakerPaysConstraint =
+        downward()(pool.out * targetQuality.rate() - upward()(pool.in / f));
+    if (nTakerPaysConstraint <= 0)
+        return std::nullopt;
+
+    // Pick the smallest to make the quality better
+    if (nTakerPaysConstraint < nTakerPays)
+        nTakerPays = nTakerPaysConstraint;
+
+    auto const takerPays = toAmount<TIn>(
+        getIssue(pool.in), *nTakerPays, Number::rounding_mode::downward);
+
+    return TAmounts<TIn, TOut>{takerPays, swapAssetIn(pool, takerPays, tfee)};
+}
+
+/**   Generate AMM offer so that either updated Spot Price Quality (SPQ)
+ * is equal to LOB quality (in this case AMM offer quality is
+ * better than LOB quality) or AMM offer is equal to LOB quality
+ * (in this case SPQ is better than LOB quality).
+ * Pre-amendment code calculates takerPays first. If takerGets is XRP,
+ * it is rounded down, which results in worse offer quality than
+ * LOB quality, and the offer might fail to generate.
+ * Post-amendment code calculates the XRP offer side first. The result
+ * is rounded down, which makes the offer quality better.
+ *   It might not be possible to match either SPQ or AMM offer to LOB
+ * quality. This generally happens at higher fees.
  * @param pool AMM pool balances
  * @param quality requested quality
  * @param tfee trading fee in basis points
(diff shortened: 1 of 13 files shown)


src/ripple/app/misc/AMMHelpers.h
```diff
@@ -145,12 +149,184 @@ withinRelativeDistance(Amt const& calc, Amt const& req, Number const& dist)
...
+ * Assumes
+ * b > 0 & 4*a*c < 0 || b < 0 & 4*a*c > 0
+ * x = (-b + sqrt(b**2 - 4*a*c))/(2*a), if b > 0
+ * x = (-b - sqrt(b**2 - 4*a*c))/(2*a), if b < 0
+ */
+inline std::optional<Number>
+solveQuadraticEqSmallest(Number const& a, Number const& b, Number const& c)
+{
+    if (auto const r = b * b - 4 * a * c; r < 0)
+        return std::nullopt;
```
Bronek, line 164: This `return std::nullopt` is not covered by the unit tests.

src/ripple/app/misc/AMMHelpers.h
```diff
@@ -145,12 +149,184 @@ withinRelativeDistance(Amt const& calc, Amt const& req, Number const& dist)
...
+    assert(targetQuality.rate() != beast::zero);
+    auto const f = feeMult(tfee);
+    auto const a = 1;
+    auto const b = pool.in * (1 - 1 / f) / targetQuality.rate() - 2 * pool.out;
+    auto const c =
+        pool.out * pool.out - pool.in * pool.out / targetQuality.rate();
+
+    auto nTakerGets = solveQuadraticEqSmallest(a, b, c);
+    if (!nTakerGets || *nTakerGets <= 0)
+        return std::nullopt;
```
Bronek, line 227: Needs two more test cases, one for `nTakerGets` being `nullopt` (i.e. when the solution is less than zero, as commended in `solveQuadraticEqSmallest`) and other when `nTakerGets` is exactly 0

Bronek (changes requested):
Unit tests need more coverage for the error cases where you fail to calculate  AMM offer,  pointed out two extra cases.

src/ripple/app/paths/impl/AMMLiquidity.cpp
```diff
@@ -239,7 +239,12 @@ AMMLiquidity<TIn, TOut>::getOffer(
             return offer;
         }
 
-        JLOG(j_.error()) << "AMMLiquidity::getOffer, failed";
+        JLOG(j_.error()) << "AMMLiquidity::getOffer, failed "
+                         << ammContext_.multiPath() << " "
+                         << ammContext_.curIters() << " "
+                         << (clobQuality ? clobQuality->rate() : STAmount{})
+                         << " " << to_string(balances.in) << " "
+                         << to_string(balances.out);
```
Bronek, line 247: Do not worry about missing coverage for the log lines. Fixing this problem globally belongs elsewhere.

src/ripple/basics/Number.h
```diff
@@ -387,6 +387,71 @@ class saveNumberRoundMode
...
+    operator=(downward const&) = delete;
+    downward&
+    operator=(downward const&&) = delete;
+    Number const&
+    operator()(Number const& n)
+    {
+        Number::setround(mode_);
+        return n;
+    }
+};
```
HowardHinnant, line 453: These are clever.  And I like the way they clean up the code.  But @scottschurr pointed out to me a potential to cause a run-time error:  Subexpressions in C++ are not guaranteed to be executed in any given order.  See: https://en.cppreference.com/w/cpp/language/eval_order#:~:text=Order%20of%20evaluation%20of%20any,same%20expression%20is%20evaluated%20again.

For example in:
```c++
z = downward()(a * b) / upward()(c - d);
```
I am concerned that the order could be:
1. construct downward
2. construct upward
3. execute a * b
4. execute c - d
5. call the downward() with a*b
6. call the upward with c-d
7. do the division

When I tried to simulate such an order with clang, I failed to do so.  But even if all of our compilers don't create a run-time error now, if they started to in the future, the error would be silent, and extremely subtle, hard to detect and hard to find and fix.

gregtatcam: Thanks for investigating. So have to break expressions into subexpressions. The functor would still work to make it less verbose, right? I.e.

    num = downard()(a * b);
    den = upward()(c - d);
    res = downard() (num / den);

HowardHinnant: Yes, that will work.  We will have to watch out for maintenance in the future to make sure it doesn't combine such expressions.

For other's, here's what the above looks like without these helpers:
```c++
saveNumberRoundMode _{Number::setround(Number::downward)};
num = a * b;
Number::setround(Number::upward);
den = c - d;
Number::setround(Number::downward);
res = num / den;
```

The basic rule for both of these styles is that we need a `;` between calls to `Number::setround`.

gregtatcam: I'm wondering if parenthesis are going to force the order, for instance. Also note additional `downward()` to set the division rounding.

    z = downward()( (downward()(a * b) ) / ( upward()(c - d) ) )

scottschurr: An approach that might allow combining expressions would be named arithmetic functions that take three parameters: two `Number's and a rounding mode.  So something vaguely like:
```
Number z = div(mul(a, b, downward), sub(c, d, upward), downward);
```
That's pretty ugly to read.  I haven't tried it, but it _might_ be a livable compromise?

gregtatcam: It is really hard to read. I think I'd rather break into subexpression if there is no better way. 
The standard does say this `Any expression in parentheses is also classified as a primary expression: this guarantees that the parentheses have higher precedence than any operator. Parentheses preserve value, type, and value category.` [standard](https://en.cppreference.com/w/cpp/language/expressions#:~:text=Any%20expression%20in%20parentheses%20is%20also%20classified%20as%20a%20primary%20expression) so everything in parenthesis should be evaluated first. So it should work as long as the constructor is called first.

scottschurr: Regarding the constructor, here's the best I could find.  In the C++23 Standard it's section 6.7 Temporary objects [class.temporary] paragraph 2:

> The materialization of a temporary object is generally delayed as long as possible in order to avoid creating unnecessary temporary objects.

Note that it says "generally", which I take to mean it is not guaranteed.

gregtatcam: So the functor can't be used the way it's currently used. This works (I abbreviated dwn-downward, up-upward) and they are functions, no functors.

    Number z = dwn(dwn(a) + b)) / dwn(up(c) - d)

Can also add unary prefix functions to Number, which work like identity functions and set the rounding, for instance: `~` is downward and `&` is upward. Then:

    Number z = ~(~a + b) / ~(&c - d)

Or could use a variation of Scott's suggestion:
muld(a, b) -> mul(a, b, downward) and so on. Then:

    Number z = divd(muld(a, b), subu(c, d))

seelabs: It's a shame the functor syntax won't work, I liked that a lot.

Doesn't everything except:
```
Number z = div(mul(a, b, downward), sub(c, d, upward), downward);
```
Run into the order of operations problem? (Even if it doesn't I don't love the `dwn(dwn(a) + b)) / dwn(up(c) - d)` syntax). 

One thought: if we used expression templates we could use: 
```
z = downward(downward(a * b) / upward(c - d))
```
And get around the order of operations issue. However, expression templates aren't a lot of fun to write or debug. There are serious downsides to going that way. I'm probably opposed to doing that, but maybe it will spark an idea from someone else. (ref: expression templates: https://en.wikipedia.org/wiki/Expression_templates)

gregtatcam: Interesting. I glanced it over and need to take another pass but it gave me an idea. How about `upward` and `downward` classes with constructors to take Number and overloaded operators +,-,*,/ returning Number. Then it could be something like this. 

     z = downward(downward(a) * downward(b)) / downward((upward(c) - upward(d)))

If too verbose then can add an overload that takes Number; for inst. operator + (Number const&), then:

    z = downward(downward(a) * b)) / (upward(c) - d)

HowardHinnant: I am seriously questioning if the cure is not worse than the disease.

There is also going to be some non-zero cost of setting the thread_local.  I'm not sure what that cost is.  It may be negligible.  But whatever it is, it is being doubled compared to doing this with calls to `Number::setround`.  If `Number::setround` is too verbose, perhaps we should just move `setround` outside of `Number`.

src/ripple/app/misc/AMMHelpers.h
```diff
@@ -145,12 +149,184 @@ withinRelativeDistance(Amt const& calc, Amt const& req, Number const& dist)
 }
 // clang-format on
 
-/** Finds takerPays (i) and takerGets (o) such that given pool composition
- * poolGets(I) and poolPays(O): (O - o) / (I + i) = quality.
- * Where takerGets is calculated as the swapAssetIn (see below).
- * The above equation produces the quadratic equation:
- * i^2*(1-fee) + i*I*(2-fee) + I^2 - I*O/quality,
- * which is solved for i, and o is found with swapAssetIn().
+namespace {
```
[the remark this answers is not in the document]
scottschurr: Another choice that's been used elsewhere in the code base would be to use `namespace detail`, rather than the anonymous `namespace`.  I don't feel strongly about it, just presenting an option.

src/ripple/app/misc/impl/AMMHelpers.cpp
```diff
@@ -203,4 +203,52 @@ solveQuadraticEq(Number const& a, Number const& b, Number const& c)
...
+    // maximize:
+    // 4 * a * c
+    // (2 * a)
+    if (b > 0)
+    {
+        auto const d = downward()((downward()(b * b) - upward()(4 * a * c)));
+        if (d < 0)
+            return std::nullopt;
+        return downward()(
+            downward()(-b + downward()(root2(d))) / upward()(2 * a));
```
seelabs, line 228: There are more numerically stable forms of the quadratic formula. For example, look for the "Citardauq Formula"

src/ripple/basics/Number.h
gregtatcam: It's not the problem with `setround`, it's breaking a complex arithmetic expression into subexpressions to get the rounding right. `Number` class made those expressions super easy to implement and read and the rounding makes it difficult.

seelabs: My suggestion is not to solve the "find a nice syntax for equations with rounding" in this PR. So something like we did in this patch: https://github.com/XRPLF/rippled/pull/5002/commits/b90f58d3a8ced5d7cbef12cd8f3215546f87ca25 and we can find a nice syntax later if we find the code in that style too difficult to read and maintain.

src/ripple/app/misc/AMMHelpers.h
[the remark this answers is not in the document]
Bronek: I feel pretty strongly against anonymous namespaces in header files, because of internal linkage. Fine if everything inside such a namespace is explicitly `static` but otherwise can lead to ODR.

src/ripple/app/misc/AMMHelpers.h
```diff
@@ -163,43 +376,114 @@ std::optional<TAmounts<TIn, TOut>>
...
+                        << " " << to_string(pool.out) << " "
+                        << Number{1} / quality.rate() << " " << tfee
+                        << std::endl;
+        return std::nullopt;
     }
-    return std::nullopt;
+
+    // Might fail due to finite precision. Should the relative difference be
+    // allowed?
+    if (Quality{*amounts} < quality)
```
seelabs, line 471: Since we're defining rounding to favor the AMM, won't this trigger almost all the time? I.e. I'd _expect_ the computed quality to be less than the requested quality (by tiny amounts, of course). I might be wrong on that, but if I am can you leave a comment in the code explaining why we don't expect to hit this very often?

src/ripple/app/misc/AMMHelpers.h
```diff
@@ -147,12 +149,223 @@ withinRelativeDistance(Amt const& calc, Amt const& req, Number const& dist)
...
+ * be generated.
+ */
+template <typename TIn, typename TOut>
+std::optional<TAmounts<TIn, TOut>>
+getAMMOfferStartWithTakerGets(
+    TAmounts<TIn, TOut> const& pool,
+    Quality const& targetQuality,
+    std::uint16_t const& tfee)
+{
+    assert(targetQuality.rate() != beast::zero);
```
seelabs, line 188: Instead of asserting, consider returning a nullopt as well, so release builds don't crash.

src/ripple/basics/Number.h
scottschurr: Thanks for that comment, @Bronek.  I also think that toward zero and away from zero feel a lot more natural.  However I know that the values being operated on are almost always positive, so maybe it's not a big deal.

src/ripple/app/misc/AMMHelpers.h
```diff
@@ -147,12 +149,223 @@ withinRelativeDistance(Amt const& calc, Amt const& req, Number const& dist)
...
+    // minimize:
+    // 4 * a * c
+    // a
+    // c
+    // maximize:
+    // 2 * a
+    // a
+    // a must be maximized and minimized. a is 1, no rounding
+    // b must be maximized and minimized. choose minimize since the effect of
+    // the rounding outside of root2 is greater than the inside of root2.
```
seelabs, line 208: Except b is squared inside the root2. What you're saying may still be true, but it's not immediately clear to me that it is.

src/ripple/basics/Number.h
HowardHinnant: We have `Number::towards_zero`, but not `Number::away_zero`.  If we decide we need `Number::away_zero` I think I could add it fairly quickly.

src/ripple/app/paths/impl/BookStep.cpp
```diff
@@ -758,8 +781,17 @@ BookStep<TIn, TOut, TDerived>::forEachOffer(
...
 
     // At any payment engine iteration, AMM offer can only be consumed once.
-    auto tryAMM = [&](std::optional<Quality> const& quality) -> bool {
-        auto ammOffer = getAMMOffer(sb, quality);
+    auto tryAMM = [&](std::optional<Quality> const& lobQuality) -> bool {
+        // If offer crossing then use the best quality out of LOB and
+        // limit quality to prevent AMM being blocked by a lower quality
+        // LOB.
+        auto const bestQuality = [&]() -> std::optional<Quality> {
+            if (sb.rules().enabled(fixAMMRounding) && lobQuality)
```
seelabs, line 789: This change doesn't relate to "AMMRounding". We might want to rename the feature.

src/ripple/app/misc/AMMHelpers.h
gregtatcam: There are two scenario. One is - AMM offer is generated so that the new Spot Price Quality is equal to the requested quality. In this case the offer's quality is better than updated SPQ. The second is - if there is a trading fee then matching the new SPQ to the target quality will push AMM offer quality bellow the target quality. In this case the AMM offer is generated to match the target quality and SPQ quality is going to be better than the target quality. In both cases we start the offer generation with the XRP side (if applicable) and we round it downward. This minimizes the offer size and maximizes the quality. If the offer is IOU/IOU, we still round downward the side of the offer that we start generating first. This again minimizes the offer and maximizes the quality. So the scenario when we succeed in the offer generation but it's quality is less than the target quality should be pretty rare. It's still possible though and perhaps we should add some error allowance.

src/ripple/app/misc/AMMHelpers.h
```diff
@@ -145,12 +149,184 @@ withinRelativeDistance(Amt const& calc, Amt const& req, Number const& dist)
...
+namespace {
+
+/** Smallest solution for quadratic equation.
+ * Assumes
+ * b > 0 & 4*a*c < 0 || b < 0 & 4*a*c > 0
+ * x = (-b + sqrt(b**2 - 4*a*c))/(2*a), if b > 0
+ * x = (-b - sqrt(b**2 - 4*a*c))/(2*a), if b < 0
+ */
+inline std::optional<Number>
+solveQuadraticEqSmallest(Number const& a, Number const& b, Number const& c)
```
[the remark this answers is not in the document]
gregtatcam: These two functions use different solutions and the rounding strategy could be also different.

src/ripple/app/paths/impl/BookStep.cpp
scottschurr: For NFTs there was a rollup of three different fixes named `fixNonFungibleTokensV1_2`.  Perhaps something like `fixAMMv1_1` would work?

src/ripple/app/misc/AMMHelpers.h
gregtatcam: To put it differently, we maximize the left side of the expression in several ways: `b * b` is maximized, `4 * a * c` is minimized, `b * b - 4 * a * c` is maximized and then the root is maximized as well. So I think the overall minimization effect is going to be greater if `b` is rounded downward. If this the right way, then `b` has to be actually rounded upward because `b` is negative in this case; i.e. we need to make the absolute `b` to be smaller.

src/ripple/app/paths/impl/BookStep.cpp
seelabs: We could think about rolling up the other AMM rounding amendment that we added to 2.2 into this one as well.

gregtatcam: We only have `fixAMMRounding` in 2.2, no?

src/ripple/app/misc/AMMHelpers.h
```diff
@@ -147,12 +149,223 @@ withinRelativeDistance(Amt const& calc, Amt const& req, Number const& dist)
...
+    // maximize:
+    // 2 * a
+    // a
+    // a must be maximized and minimized. a is 1, no rounding
+    // b must be maximized and minimized. choose minimize since the effect of
+    // the rounding outside of root2 is greater than the inside of root2.
+    // fee is always maximized
+    // feeMult is always minimized
+    auto const fee = upward()(getFee(tfee));
+    auto const feeMult = downward()(1 - fee);
```
scottschurr, line 212: Just something to consider that might improve readability.  There's a `feeMult()` function called in this file (defined in AMMCore.h).  There are also just a few locals in this file named `feeMult`, like this one.  Also, it seems to me that, throughout most of this file, locals filling the role of `feeMult` are named `f`.

Consider renaming those few locals in this file named `feeMult` to `f`.  This may increase naming consistency in the file.  It might also reduce cognitive dissonance caused by a local name and a function name being the same.

src/ripple/app/misc/AMMHelpers.h
```diff
@@ -147,12 +149,223 @@ withinRelativeDistance(Amt const& calc, Amt const& req, Number const& dist)
...
+ * from the payment perspective is IOU(in)/XRP(out)
+ * Equations:
+ * Spot Price Quality after the offer is consumed:
+ *     Qsp = (O - o) / (I + i) (1)
+ *  where O is poolPays, I is poolGets, o is takerGets, i is takerPays
+ * Swap out:
+ *     i = (I * o) / (O - o) * f (2)
+ *  where f is (1 - tfee/100000), tfee is in basis points
+ * Effective price targetQuality:
+ *     Qep = o / i (3)
```
scottschurr, line 168: Walking into this cold, it took me a while to understand that, for example, `(1)` designates equation 1 (not multiply by 1).  To avoid that confusion, consider documenting the equations like this:
```
 *     Qsp = (O - o) / (I + i) -- Equation (1)
 *     i = (I * o) / (O - o) * f -- Equation (2)
 *     Qep = o / I -- Equation (3)
```

src/ripple/basics/Number.h
gregtatcam: > Regarding the constructor, here's the best I could find. In the C++23 Standard it's section 6.7 Temporary objects [class.temporary] paragraph 2:
> 
> > The materialization of a temporary object is generally delayed as long as possible in order to avoid creating unnecessary temporary objects.
> 
> Note that it says "generally", which I take to mean it is not guaranteed.

@scottschurr doesn't this [sequenced before](https://en.cppreference.com/w/cpp/language/eval_order#:~:text=In%20a%20function%2Dcall%20expression%2C%20the%20expression%20that%20names%20the%20function%20is%20sequenced) rule imply that in `downward()(a + b)`, `downward()` is going to be evaluated first?

src/ripple/app/paths/impl/BookStep.cpp
seelabs: You're right. My bad.

src/ripple/basics/Number.h
scottschurr: Yes, I'm pretty sure that in your example `downward()` will be evaluated first.

However I'm most concerned about @HowardHinnant's example at the top of this discussion.  If both `upward()` and `downward()` are constructed before either of the functors are called then one of the computations will round the wrong way.  The only way I personally know of to guarantee order of construction is with a sequence point.

The order of evaluation discussion says very little about when _constructors_ will be called.  My experience with undefined behavior has made me leery of surprises that optimizers can bring.  But I haven't found wording that _proves_ that the compiler will do the wrong thing.

gregtatcam: I give up (for now)!

src/ripple/app/misc/AMMHelpers.h
seelabs: Here's what I did in sage:

```
sage: def q(a,b,c): return -b+sqrt(b*b-4*a*c)
sage: small_delta = 10^-20
sage: large_delta = 9*10^-20
sage: R=RealField(400)
sage: R(q(1,10+large_delta,1)) > R(q(1,10-small_delta,1))
True
```

I _think_ that shows that if we round up we can end up with larger values than if we round down. In this case, it's simulating rounding 10.00...001 up to 10.00...01 or down to 10.00...000). When I round `b` down I get a smaller value. It's possible I messed up this calculation (please dummy check me), but I'm still not convinced we'll find a single rounding for b that minimizes this. We may need one rounding inside the root2 and one outside the root2.

gregtatcam: A few things:
- The result is negative given `a,b,c` values.
- There is a relation between `b` and `c`. I think a real test should take it into consideration.
- How does this test factor in other rounding that I mentioned?
- Doesn't having different `b` kind of breaks the basic algebra? `b` must be the same in quadratic equation solution . x = (-b1 +- sqrt(b2*b2 - 4 * a *c)) / (2 * a) is not a valid quadratic equation solution even if `b1` and `b2` differ by a small delta. Am I missing something?

gregtatcam: I'm actually thinking how much of the rounding pays off in this function. The offer generation definitely must start with XRP and rounded downward. This makes a big difference, especially for small offers. But in case of IOU/IOU offer the rounding might not do much. To your earlier point, swap functions round in favor of the pool. IOU/IOU offer generation starts with takerPays and rounds it downward but takerGets is also rounded downward by swapAssetOut. As the result the AMM offer's quality might be less than the target quality. Which means we might have to allow for some error anyways and the rounding doesn't add much in this case except for more complexity.

HowardHinnant: Just a general comment, not necessarily related to this bit of code:  I'm beginning to question the wisdom of directed rounding on every single operation.  The one other place I've seen this done is when one is trying to put lower and upper limits on the exact answer.  One does the computation twice, with consistent rounding one way, and then the other.

That's not what we're trying to do here.  What we want is to compute as accurately as possible, but make sure that round off error doesn't violate an invariant, such as creating XRP.  Perhaps we should stick to round_to_nearest, but then clip the final result to meet our invariants.  Perhaps clipping the final result, as opposed to rounding individual operations is the better tool here.

gregtatcam: This makes sense to me.

seelabs: I'm really poking at this statement:
```
b must be maximized and minimized. choose minimize since the effect of
the rounding outside of root2 is greater than the inside of root2
```
 It's not obvious to me that that's true, and I _think_ the test I wrote shows that it's not always true. If we really do want to choose the rounding that minimizes the whole expression sometimes we'll have to round up, and sometimes we'll have to round down (I _think_). There may be other constraints that I'm not considering (although I don't _think_ a relationship between b and c would matter, but maybe it does.)

So my real concern is that we always honor the invariant. If we can show that the invariant is honored in all cases, then we're good. An easy way to do that is to show we always set roundings to favor the AMM. If we can't do that then we have a much harder analysis job to do. (or we can give up trying to honor the invariant, but I'd rather not do that). Anyway, let's talk some more tomorrow about this.

gregtatcam: The invariant is honored in all cases. The reason for the rounding in this function is to minimize the generated offer so that the quality matches the provided target quality. This is accomplished by rounding downward the side of the offer that we start generating first. It doesn't matter which side it is - the smaller the amount the better the quality. The other side of the offer is calculated with either `swapAssetIn` or `swapAssetOut`. Those functions ensure the invariant is honored and round the result downward (to minimize takerGets) or upward (to maximize takerPays) respectively.

seelabs: As long as the invariant is honored I'm good. And I agree we want the most accurate result while honoring the invariant.

gregtatcam: I'm going to change the rounding per @HowardHinnant suggestion and address all current comments.

src/ripple/basics/Number.h
scottschurr: It seems like the direction we're going is that `upward()` and `downward()` will be a lot less important.  That's probably for the best.  However I cobbled up an approach that is not as pretty as the original `upward()` and `downward()` but includes the necessary sequence points.  It leans heavily on lambdas.  The code looks like this:
```
template<Number::rounding_mode Rm, typename Functor>
requires std::is_invocable_v<Functor>
auto
numberRounder(Functor&& f)
{
    saveNumberRoundMode guard(Number::getround());
    Number::setround(Rm);
    return std::forward<Functor>(f)();
}

template<typename Functor>
requires std::is_invocable_v<Functor>
auto
roundUp(Functor&& f)
{
    return numberRounder<Number::upward>(std::forward<Functor>(f));
}

template<typename Functor>
requires std::is_invocable_v<Functor>
auto
roundDn(Functor&& f)
{
    return numberRounder<Number::downward>(std::forward<Functor>(f));
}
```
Using it looks like this (including clang-formatting):
```
    auto const nTakerPaysConstraint = roundDn([&] {
        return pool.out * targetQuality.rate() -
            roundUp([&] { return pool.in / f; });
    });
```
Just for your consideration.

Also, I've found the two-phase use of `saveNumberRoundMode` to be a bit more awkward than necessary.  So a while back I cobbled up something I find easier to use.  The implementation is in STAmount.cpp, since that's the only place it's currently used:
```
class NumberRoundModeGuard
{
    saveNumberRoundMode saved_;

public:
    explicit NumberRoundModeGuard(Number::rounding_mode mode) noexcept
        : saved_{Number::setround(mode)}
    {
    }

    NumberRoundModeGuard(NumberRoundModeGuard const&) = delete;

    NumberRoundModeGuard&
    operator=(NumberRoundModeGuard const&) = delete;
};
```
I personally find `NumberRoundModeGuard` to be easier to think about and use than `saveNumberRoundMode`.  If that's true for you as well, then I encourage you to move `NumberRoundModeGuard` into the Number.h header and use it freely.

gregtatcam: Thanks for both suggestions! I was considering lambdas but thought it would not be easy to read. But now that I look at it, I think it's a good alternative.

seelabs: @scottschurr Nice!

gregtatcam:
I pushed [commit]( https://github.com/XRPLF/rippled/pull/4983/commits/33ccc5e4ad3ff4fc79ac3ee1012d752f908e7882), which should address all feedback. The commit message details all the updates, but the main change is that the rounding is changed to `nearest` when the AMM offer matching target quality is generated.

scottschurr (commented):
I'm going to be AFK tomorrow, so I thought I'd get the comments I have so far to you today.  I'll continue the review on Monday.  These are mostly nits.  But there's one I'm concerned about.

src/ripple/app/paths/impl/BookStep.cpp
```diff
@@ -297,6 +297,13 @@ class BookPaymentStep : public BookStep<TIn, TOut, BookPaymentStep<TIn, TOut>>
         return true;
     }
 
+    // A payment doesn't use max quality threshold (limitQuality)
```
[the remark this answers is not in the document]
gregtatcam: You are right, AMM supports cross-currency Payment. The comment is not correct. I meant to say that the strand's limitQuality doesn't relate to the step's quality in case of Payment.

src/ripple/app/misc/AMMHelpers.h
```diff
@@ -147,12 +155,163 @@ withinRelativeDistance(Amt const& calc, Amt const& req, Number const& dist)
...
+    auto const nTakerPaysConstraint =
+        pool.out * targetQuality.rate() - pool.in / f;
+    if (nTakerPaysConstraint <= 0)
+        return std::nullopt;
+
+    // Select the smallest to maximize the quality
+    if (nTakerPaysConstraint < *nTakerPays)
+        nTakerPays = nTakerPaysConstraint;
+
+    auto getAmounts = [&](Number const& nTakerPays) {
```
seelabs, line 286: This param shadows the local variable. Consider changing the name.

src/ripple/app/misc/AMMHelpers.h
```diff
@@ -147,12 +155,163 @@ withinRelativeDistance(Amt const& calc, Amt const& req, Number const& dist)
...
+    auto const nTakerGetsConstraint =
+        pool.out - pool.in / (targetQuality.rate() * f);
+    if (nTakerGetsConstraint <= 0)
+        return std::nullopt;
+
+    // Select the smallest to maximize the quality
+    if (nTakerGetsConstraint < *nTakerGets)
+        nTakerGets = nTakerGetsConstraint;
+
+    auto getAmounts = [&](Number const& nTakerGets) {
```
seelabs, line 218: This param shadows the local variable, consider changing the name.

src/ripple/app/misc/AMMHelpers.h
```diff
@@ -147,12 +163,165 @@ withinRelativeDistance(Amt const& calc, Amt const& req, Number const& dist)
...
+    auto const nTakerGetsConstraint =
+        pool.out - pool.in / (targetQuality.rate() * f);
+    if (nTakerGetsConstraint <= 0)
+        return std::nullopt;
+
+    // Select the smallest to maximize the quality
+    if (nTakerGetsConstraint < *nTakerGets)
+        nTakerGets = nTakerGetsConstraint;
+
+    auto getAmounts = [&pool, &tfee](Number const& nTakerGets_) {
```
seelabs, line 225: I think we need a different fix here. Trailing underscores are for member variables.

seelabs (approved):
:+1:. Thanks Greg, and nice job on this!

Bronek (approved):

seelabs (approved):
Still :+1: after latest patch

scottschurr (approved):
:+1: Looks good to me.  I found a few nits or suggestions for changes.  But I didn't find anything that compromises functionality.  All these comments are at your discretion.

seelabs (approved):
Still :+1:

scottschurr (approved):
:+1: Thanks for considering those nits.  Looks great.

PR #4996, opened 2024-04-19
[TRIVIAL] fix: Remove redundant STAmount conversion in test

ximinez:
Does what it says on the tin.


thejohnfreeman (approved):

scottschurr (approved):
Works for me on macOS with Apple clang version 13.1.6 (clang-1316.0.21.2.5 and warnings-as-errors.

PR #5002, opened 2024-04-25
fix: AMM swap rounding

seelabs:

fix amendment: AMM swap should honor invariants:
    
The AMM has an invariant for swaps where:
```
new_balance_1*new_balance_2 >= old_balance_1*old_balance_2
```
Due to rounding, this invariant could sometimes be violated (although by very small amounts).
    
This patch introduces an amendment `fixAMMRounding` that changes the rounding to always favor the AMM. Doing this should maintain the invariant.

### Type of Change

- [ x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ x] Tests (you added tests for code that already exists, or your new feature included in this PR)
- [ ] Documentation update
- [ ] Release

### DIFF
--- a/src/ripple/app/misc/AMMHelpers.h
+++ b/src/ripple/app/misc/AMMHelpers.h
@@ -24,8 +24,10 @@
 #include <ripple/basics/Number.h>
 #include <ripple/protocol/AMMCore.h>
 #include <ripple/protocol/AmountConversions.h>
+#include <ripple/protocol/Feature.h>
 #include <ripple/protocol/Issue.h>
 #include <ripple/protocol/Quality.h>
+#include <ripple/protocol/Rules.h>
 #include <ripple/protocol/STAccount.h>
 #include <ripple/protocol/STAmount.h>
 
@@ -228,10 +230,62 @@ swapAssetIn(
     TIn const& assetIn,
     std::uint16_t tfee)
 {
-    return toAmount<TOut>(
-        getIssue(pool.out),
-        pool.out - (pool.in * pool.out) / (pool.in + assetIn * feeMult(tfee)),
-        Number::rounding_mode::downward);
+    if (auto const& rules = getCurrentTransactionRules();
+        rules && rules->enabled(fixAMMRounding))
+    {
+        // set rounding to always favor the amm. Clip to zero.
+        // calculate:
+        // pool.out -
+        // (pool.in * pool.out) / (pool.in + assetIn * feeMult(tfee)),
+        // and explicitly set the rounding modes
+        // Favoring the amm means we should:
+        // minimize:
+        // pool.out -
+        // (pool.in * pool.out) / (pool.in + assetIn * feeMult(tfee)),
+        // maximize:
+        // (pool.in * pool.out) / (pool.in + assetIn * feeMult(tfee)),
+        // (pool.in * pool.out)
+        // minimize:
+        // (pool.in + assetIn * feeMult(tfee)),
+        // minimize:
+        // assetIn * feeMult(tfee)
+        // feeMult is: (1-fee), fee is tfee/100000
+        // minimize:
+        // 1-fee
+        // maximize:
+        // fee
+        saveNumberRoundMode _{Number::getround()};
+
+        Number::setround(Number::upward);
+        auto const numerator = pool.in * pool.out;
+        auto const fee = getFee(tfee);
+
+        Number::setround(Number::downward);
+        auto const denom = pool.in + assetIn * (1 - fee);
+
+        if (denom.signum() <= 0)
+            return toAmount<TOut>(getIssue(pool.out), 0);
+
+        Number::setround(Number::upward);
+        auto const ratio = numerator / denom;
+
+        Number::setround(Number::downward);
+        auto const swapOut = pool.out - ratio;
+
+        if (swapOut.signum() < 0)
+            return toAmount<TOut>(getIssue(pool.out), 0);
+
+        return toAmount<TOut>(
+            getIssue(pool.out), swapOut, Number::rounding_mode::downward);
+    }
+    else
+    {
+        return toAmount<TOut>(
+            getIssue(pool.out),
+            pool.out -
+                (pool.in * pool.out) / (pool.in + assetIn * feeMult(tfee)),
+            Number::rounding_mode::downward);
+    }
 }
 
 /** Swap assetOut out of the pool and swap in a proportional amount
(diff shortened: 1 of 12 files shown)


src/test/app/AMM_test.cpp
```diff
@@ -3493,10 +3621,24 @@ struct AMM_test : public jtx::AMMTest
...
+                    BEAST_EXPECT(ammAlice.expectBalances(
+                        XRP(10'100), USD(10'000), ammAlice.tokens()));
+                    // Initial 30,000 + 200
+                    BEAST_EXPECT(expectLine(env, carol, USD(30'200)));
+                }
+                else
+                {
+                    BEAST_EXPECT(ammAlice.expectBalances(
+                        XRP(10'100),
+                        STAmount(USD, 10'000'00000000001, -11),
```
gregtatcam, line 3635: Doesn't build on apple clang. Missed `UINT64_C`.

src/test/app/AMM_test.cpp
```diff
@@ -4339,15 +4624,28 @@ struct AMM_test : public jtx::AMMTest
...
+            if (!features[fixAMMRounding])
+            {
+                BEAST_EXPECT(expectLine(env, bob, EUR(1'990)));
+                BEAST_EXPECT(ammAlice.expectBalances(
+                    USD(1'000), EUR(1'005), ammAlice.tokens()));
+            }
+            else
+            {
+                BEAST_EXPECT(expectLine(
+                    env, bob, STAmount(EUR, 1989'999999999999, -12)));
```
gregtatcam, line 4636: Doesn't build on apple clang. Missed `UINT64_C`.

src/test/app/AMM_test.cpp
```diff
@@ -4339,15 +4624,28 @@ struct AMM_test : public jtx::AMMTest
...
+                BEAST_EXPECT(ammAlice.expectBalances(
+                    USD(1'000),
+                    STAmount(EUR, 1005'000000000001, -12),
```
gregtatcam, line 4639: Doesn't build on apple clang. Missed `UINT64_C`.

gregtatcam (changes requested):

src/test/app/AMM_test.cpp
Bronek: fixed

src/test/app/AMM_test.cpp
Bronek: fixed

src/test/app/AMM_test.cpp
Bronek: fixed

src/ripple/protocol/AmountConversions.h
```diff
@@ -130,16 +132,48 @@ toAmount(
...
             return STAmount(issue, static_cast<std::int64_t>(n));
         return STAmount(issue, n.mantissa(), n.exponent());
     }
+    else
+    {
+        constexpr bool alwaysFalse = !std::is_same_v<T, T>;
+        static_assert(alwaysFalse, "Unsupported type for toAmount");
+    }
+}
+
```
gregtatcam, line 152: How about using `concept` instead of the `static_assert` here and other places?

    template <typename T>
    concept ValidAmount =
        std::is_same_v<T, IOUAmount> ||
        std::is_same_v<T, XRPAmount> ||
        std::is_same_v<T, STAmount>;

    template <ValidAmount T>
    T
    toMaxAmount(Issue const& issue)
    {
     ...
    }

I don't think max `Number` is applicable, at least not in the context that `toMaxAmount()` is used.

seelabs: I thought I got a compile error without the supporting `Number`.

I don't object to the `concept`, but I weakly prefer the code without it. I don't love that the `is_same` is repeated in both the body of the function and the concept, and the concept isn't important for distinguishing between overloads. I'll change it if you or others feel strongly, but I think I prefer the non-concept version.

gregtatcam: It builds for me without `Number` on OSX, Linux, and Windows.

I prefer `concept` since it's clearly defines the type constraint. I don't feel strongly about it though.

src/ripple/app/misc/AMMHelpers.h
```diff
@@ -228,10 +230,62 @@ swapAssetIn(
...
+        // feeMult is: (1-fee), fee is tfee/100000
+        // minimize:
+        // 1-fee
+        // maximize:
+        // fee
+        saveNumberRoundMode _{Number::getround()};
+
+        Number::setround(Number::upward);
+        auto const numerator = pool.in * pool.out;
+        auto const fee = getFee(tfee);
```
gregtatcam, line 261: Does it make sense to always round upwards `getFee()` and downward `feeMult()`. The fee is added to the pool so we want always to maximize it's effect. Can have the default just in case it needs to be changed in some instances:

    inline Number
    getFee(std::uint16_t tfee, Number::rounding_mode rounding = Number::rounding_mode::upward)
    {
        if (auto const& rules = getCurrentTransactionRules();
            rules && rules->enabled(fixAMMRounding))
        {
            saveNumberRoundMode rm(Number::setround(rounding));
            return Number{tfee} / AUCTION_SLOT_FEE_SCALE_FACTOR;
        }
        return Number{tfee} / AUCTION_SLOT_FEE_SCALE_FACTOR;
    }

    inline Number
    feeMult(std::uint16_t tfee, Number::rounding_mode rounding = Number::rounding_mode::downward)
    {
        if (auto const& rules = getCurrentTransactionRules();
            rules && rules->enabled(fixAMMRounding))
        {
            saveNumberRoundMode rm(Number::setround(rounding));
            return 1 - getFee(tfee);
        }
        return 1 - getFee(tfee);
    }

src/ripple/protocol/AmountConversions.h
seelabs: I hear you. I don't think this is worth spending too much time on. I'm going to keep it as is.

gregtatcam (approved):
👍 LGTM

src/ripple/protocol/Rules.h
```diff
@@ -90,5 +96,31 @@ class Rules
...
+
+void
+setCurrentTransactionRules(std::optional<Rules> r);
+
+/** RAII class to set and restore the current transaction rules
+ */
+class CurrentTransactionRulesGuard
+{
+public:
+    explicit CurrentTransactionRulesGuard(Rules r)
```
HowardHinnant, line 110: This class should not be CopyConstructible nor CopyAssignable.

HowardHinnant (approved):
Approved, but I would like to make `CurrentTransactionRulesGuard` not copyable either now or in the near future.  All it needs is deleted copy members.  We already do not copy it, so this won't break anything.

src/ripple/protocol/Rules.h
seelabs: Yep, agreed, thanks! Fixed in `94bbc9c8a9 [fold] CurrentTransactionRulesGuard should not be copyable or assignable`

scottschurr (approved):
:+1: LGTM.  I left one note that might be a small improvement for future maintainability.  But, in general, it looks like it should do the job.

PR #5006, opened 2024-04-29
Add external directory for Conan recipe's exports

legleux:

Adds a the `external` directory to the Conan exports required for generating the `libxrpl` Conan package.

### Context of Change

Without exporting the contents of `external/`, package generation errors out with:

```
CMake Error at CMakeLists.txt:73 (add_subdirectory):
  add_subdirectory given source "external/secp256k1" which is not an existing
  directory.
CMake Error at CMakeLists.txt:74 (add_library):
  add_library cannot create ALIAS target "secp256k1::secp256k1" because
  target "secp256k1" does not already exist.
CMake Error at CMakeLists.txt:75 (add_subdirectory):
  add_subdirectory given source "external/ed25519-donna" which is not an
  existing directory.
```

### Type of Change

### API Impact

- [x] `libxrpl` change (any change that may affect `libxrpl` or dependents of `libxrpl`)

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -116,7 +116,7 @@ class Xrpl(ConanFile):
             self.requires('rocksdb/6.29.5')
 
     exports_sources = (
-        'CMakeLists.txt', 'Builds/*', 'bin/getRippledInfo', 'src/*', 'cfg/*'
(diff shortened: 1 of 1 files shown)


thejohnfreeman (approved):

thejohnfreeman:
There will be yet more new exports after the physical redesign is complete, but this reminds me that I need to add them to the final PR (#4997).

intelliot (approved):

PR #5009, opened 2024-05-01
Remove flow assert:

seelabs:
Rounding in the payment engine is causing an assert to sometimes fire with "dust" amounts. This is causing issues when running debug builds of rippled. This issue will be addressed, but the assert is no longer serving its purpose.

### Type of Change

### DIFF
--- a/src/ripple/app/paths/impl/StrandFlow.h
+++ b/src/ripple/app/paths/impl/StrandFlow.h
@@ -834,7 +834,13 @@ flow(
     {
         if (actualOut > outReq)
         {
-            assert(0);
+            // Rounding in the payment engine is causing this assert to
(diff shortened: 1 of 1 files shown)


Bronek (approved):

scottschurr (approved):
I know it was hard to choose to do this, but I think this is a good choice.  Having this `assert` fire was training me to be dismissive of `assert`s.

Bronek:
> I know it was hard to choose to do this, but I think this is a good choice. Having this `assert` fire was training me to be dismissive of `assert`s.

I think as long as we:

* know the reason for the `assert` fail and
* are confident the severity does not warrant it and
* reject the transaction anyway

... it's fine not  to have it

Issue #4965, opened 2024-03-24
AMMCreate transaction missing Asset field in affected nodes (Version: 2.1.0)


Silkjaer:
## Issue Description
When a new AMM is created, the created node in affected nodes miss the Asset field if it is XRP.

## Steps to Reproduce
Example transaction: https://xrpscan.com/tx/BDCC35ED3EEA1F6E4AC76C21395F98239BACF9D59708CF3A4560830FF734AEF7

shawnxie999:
This is expected behavior. The metadata would omit fields that have default values. In this case, `Asset` field's default value is XRP, and therefore it would be omitted. There is really no need to look into the metadata for this info - one could simply look at the transaction payload and find out the values both `Amount` and `Amount2`

Silkjaer:
I disagree, as other objects do not omit default values in the metadata when they are created.

joynancy:
The `amount` and `amount2` values are omitted, they are included in preload transactions. The other explorers are displaying the assets and amounts without any issue. e.g. https://bithomp.com/explorer/BDCC35ED3EEA1F6E4AC76C21395F98239BACF9D59708CF3A4560830FF734AEF7; https://livenet.xrpl.org/transactions/BDCC35ED3EEA1F6E4AC76C21395F98239BACF9D59708CF3A4560830FF734AEF7

I am reaching out to XRPScan team to see if they agree to update.

shawnxie999:
@Silkjaer Other objects do have default values that are omitted (like empty array), but they might be less obvious. I do agree with you that it would be more clear if XRP is not omitted. However, I don't think it would cause any kind information loss, as the user could retrieve the AMM data in other ways. In the case of `AMM` metadata, they could look at what the submitter entered for the transaction payload for the asset amounts, which is more efficient than looking at the metadata. I'm open to oppositions.

ximinez:
> In the case of `AMM` metadata, they could look at what the submitter entered for the transaction payload for the asset amounts, which is more efficient than looking at the metadata. I'm open to oppositions.

@shawnxie999 As a general rule, we want to discourage looking at the transaction payload as a source of truth. That's how we got things like the [partial payment exploit](https://xrpl.org/docs/concepts/payment-types/partial-payments#partial-payments-exploit). The metadata is the canonical truth about what happened during a transaction, so it should be accurate and complete enough that there is no **need** to look at the transaction payload.

[...]

PR #5012, opened 2024-05-02
Fix AMM missing XRP Asset Metadata

shawnxie999:

In the `NewFields` of a AMM object after a `AMMCreate` transaction, `Asset` would be omitted if it is XRP, which is incorrect. A new amendment fixAMMMetadata fixes this problem 

### Type of Change

- [ ] Documentation update
- [ ] Release

### API Impact

### DIFF
--- a/src/ripple/protocol/Feature.h
+++ b/src/ripple/protocol/Feature.h
@@ -74,7 +74,7 @@ namespace detail {
 // Feature.cpp. Because it's only used to reserve storage, and determine how
 // large to make the FeatureBitset, it MAY be larger. It MUST NOT be less than
 // the actual number of amendments. A LogicError on startup will verify this.
-static constexpr std::size_t numFeatures = 73;
+static constexpr std::size_t numFeatures = 74;
 
 /** Amendments that this server supports and the default voting behavior.
    Whether they are enabled depends on the Rules defined in the validated
@@ -360,6 +360,7 @@ extern uint256 const fixEmptyDID;
 extern uint256 const fixXChainRewardRounding;
 extern uint256 const fixPreviousTxnID;
 extern uint256 const fixAMMRounding;
+extern uint256 const fixAMMMetadata;
 
 }  // namespace ripple
 
--- a/src/ripple/protocol/impl/STCurrency.cpp
+++ b/src/ripple/protocol/impl/STCurrency.cpp
@@ -17,6 +17,8 @@
 */
 //==============================================================================
 
+#include <ripple/protocol/Feature.h>
+#include <ripple/protocol/Rules.h>
 #include <ripple/protocol/STCurrency.h>
 #include <ripple/protocol/jss.h>
 
@@ -71,6 +73,12 @@ STCurrency::isEquivalent(const STBase& t) const
 bool
 STCurrency::isDefault() const
 {
+    if (auto const rules = getCurrentTransactionRules())
+    {
+        if (rules->enabled(fixAMMMetadata))
+            return false;
+    }
+
     return isXRP(currency_);
 }
 
--- a/src/test/app/Oracle_test.cpp
+++ b/src/test/app/Oracle_test.cpp
@@ -666,6 +666,44 @@ private:
         }
     }
 
+    void
+    testXRPAssetMeta(FeatureBitset features)
+    {
+        testcase("XRP Asset Meta");
+
+        Env env(*this, features);
+        Account const owner("owner");
+        env.fund(XRP(1'000), owner);
+        Oracle oracle(
+            env, {.owner = owner, .series = {{"XRP", "USD", 740, 1}}});
+
+        std::string const txHash{
+            env.tx()->getJson(JsonOptions::none)[jss::hash].asString()};
+
+        env.close();
+        Json::Value const meta = env.rpc("tx", txHash)[jss::result][jss::meta];
+
+        for (Json::Value const& metaNode : meta[sfAffectedNodes.jsonName])
+        {
+            // Ensure XRP asset is shown in PriceData regardless of the
+            // fixAMMMetadata amendment
+            if (metaNode.isMember(sfCreatedNode.jsonName) &&
+                metaNode[sfCreatedNode.jsonName][sfLedgerEntryType.jsonName]
+                        .asString() == jss::Oracle)
+            {
+                auto const newFields =
+                    metaNode[sfCreatedNode.jsonName][sfNewFields.jsonName];
+                BEAST_EXPECT(newFields.isMember(sfPriceDataSeries.jsonName));
+
+                auto const priceData = newFields[sfPriceDataSeries.jsonName][0u]
+                                                [sfPriceData.jsonName];
+                BEAST_EXPECT(priceData.isMember(sfBaseAsset.jsonName));
+                BEAST_EXPECT(
+                    priceData[sfBaseAsset.jsonName].asString() == "XRP");
+            }
+        }
+    }
+
 public:
     void
     run() override
@@ -684,6 +722,8 @@ public:
               all - featureExpandedSignerList})
             testMultisig(features);
         testLedgerEntry();
+        testXRPAssetMeta(all);
+        testXRPAssetMeta(all - fixAMMMetadata);
     }
 };
 
--- a/src/ripple/protocol/impl/Feature.cpp
+++ b/src/ripple/protocol/impl/Feature.cpp
@@ -467,6 +467,7 @@ REGISTER_FIX    (fixEmptyDID,                   Supported::yes, VoteBehavior::De
 REGISTER_FIX    (fixXChainRewardRounding,       Supported::yes, VoteBehavior::DefaultNo);
 REGISTER_FIX    (fixPreviousTxnID,              Supported::yes, VoteBehavior::DefaultNo);
 REGISTER_FIX    (fixAMMRounding,                Supported::yes, VoteBehavior::DefaultNo);
+REGISTER_FIX    (fixAMMMetadata,                Supported::yes, VoteBehavior::DefaultNo);
 
 // The following amendments are obsolete, but must remain supported
 // because they could potentially get enabled.
--- a/src/ripple/protocol/impl/STIssue.cpp
+++ b/src/ripple/protocol/impl/STIssue.cpp
@@ -21,6 +21,8 @@
 
 #include <ripple/basics/Log.h>
 #include <ripple/basics/contract.h>
+#include <ripple/protocol/Feature.h>
+#include <ripple/protocol/Rules.h>
 #include <ripple/protocol/SystemParameters.h>
 #include <ripple/protocol/jss.h>
 
@@ -88,6 +90,12 @@ STIssue::isEquivalent(const STBase& t) const
 bool
 STIssue::isDefault() const
 {
+    if (auto const rules = getCurrentTransactionRules())
+    {
+        if (rules->enabled(fixAMMMetadata))
+            return false;
+    }
+
     return issue_ == xrpIssue();
 }
 
--- a/src/test/app/AMM_test.cpp
+++ b/src/test/app/AMM_test.cpp
@@ -5981,6 +5981,49 @@ private:
     }
 
     void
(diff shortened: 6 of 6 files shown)


src/ripple/protocol/impl/STCurrency.cpp
```diff
@@ -71,6 +73,12 @@ STCurrency::isEquivalent(const STBase& t) const
 bool
 STCurrency::isDefault() const
 {
+    if (auto const rules = getCurrentTransactionRules())
```
gregtatcam, line 76: This changes semantics of this call. Post-amendment `STCurrency` and `STIssue` will not have the default value. Is this really the right behavior?

shawnxie999: https://github.com/XRPLF/rippled/issues/4965#issuecomment-2079688630

here is what Ed said about default value

gregtatcam: Is `isDefault()` used for the metadata insertion only? If not then is returning always `false` consistent with other uses?

src/test/app/Oracle_test.cpp
```diff
@@ -666,6 +666,44 @@ struct Oracle_test : public beast::unit_test::suite
...
+        std::string const txHash{
+            env.tx()->getJson(JsonOptions::none)[jss::hash].asString()};
+
+        env.close();
+        Json::Value const meta = env.rpc("tx", txHash)[jss::result][jss::meta];
+
+        for (Json::Value const& metaNode : meta[sfAffectedNodes.jsonName])
+        {
+            // Ensure XRP asset is shown in PriceData regardless of the
+            // fixAMMMetadata amendment
```
gregtatcam, line 689: Why is this not amendment guarded?

gregtatcam:
Does it make sense to add another virtual method to `STBase`, say `canInsertMetadata()` with the default implementation returning `isDefault()` and `STIssue`/`STCurrency` returning false? `canInsertMetadata()` is only called in the metadata handling. This ensures that `isDefault()` semantic is not broken.

src/test/app/Oracle_test.cpp
shawnxie999: This is not a problem with Oracle I believe, since the asset is an inner object

src/ripple/protocol/impl/STCurrency.cpp
shawnxie999: Yes i believe isDefault() is only used for metadata

src/test/app/Oracle_test.cpp
gregtatcam: Why it should not be amendment guarded if it's an inner object?

gregtatcam:
> > Does it make sense to add another virtual method to `STBase`, say `canInsertMetadata()` with the default implementation returning `isDefault()` and `STIssue`/`STCurrency` returning false? `canInsertMetadata()` is only called in the metadata handling. This ensures that `isDefault()` semantic is not broken.
> 
> I'm not sure if that would be needed. `isDefault()` itself is only used for metadata purposes(though would need to double check this), so I don't think we would need to introduce another method for this. And by Ed's definition, `isDefault()` is only supposed to return true if the value represents "nothing", so omitting XRP would clearly be wrong behavior.

Can you confirm it's only used for the metadata? 

Does `badCurrency()` represent "nothing"? By the way, `STAmount` has `isDefault()` as `return (mValue == 0) && mIsNative;`; i.e. XRP(0), which is clearly not "nothing".

src/test/app/Oracle_test.cpp
shawnxie999: Inner object default values will not omitted. Metadata would only check the outer [object](https://github.com/XRPLF/rippled/blob/f6509495731b396ce33d96778e8df0366a8200bf/src/ripple/ledger/impl/ApplyStateTable.cpp#L236)

exunda:
> > > Does it make sense to add another virtual method to `STBase`, say `canInsertMetadata()` with the default implementation returning `isDefault()` and `STIssue`/`STCurrency` returning false? `canInsertMetadata()` is only called in the metadata handling. This ensures that `isDefault()` semantic is not broken.
> > 
> > 
> > I'm not sure if that would be needed. `isDefault()` itself is only used for metadata purposes(though would need to double check this), so I don't think we would need to introduce another method for this. And by Ed's definition, `isDefault()` is only supposed to return true if the value represents "nothing", so omitting XRP would clearly be wrong behavior.
> 
> Can you confirm it's only used for the metadata?
> 
> Does `badCurrency()` represent "nothing"? By the way, `STAmount` has `isDefault()` as `return (mValue == 0) && mIsNative;`; i.e. XRP(0), which is clearly not "nothing".

That is really a scary question for a Rippel employee to ask! 😨 The `badCurrency` is the encoding of the string  XRP as 160-bit asset code and used to prevent trust lines with that asset code from being created, probably as a safety to prevent people from being tricked. But attackers can still do this trick by using other character case or even a space. Very sloppy. I am surprsied nobody else has discussed this!

Bronek:
FWIW, I am uncomfortable with this change. Both `STCurrency` and `STIssue` are basic protocol building blocks, and because of this we cannot assume that `isDefault()` functions will be only ever used for metadata generation (that is, regardless of the answer to the question posed by @gregtatcam , things may change in the future). IMO we should either add a new function with the required semantics and change other bits (i.e. where metadata is generated) to use it, or do something else.

dangell7:
> FWIW, I am uncomfortable with this change. Both `STCurrency` and `STIssue` are basic protocol building blocks, and because of this we cannot assume that `isDefault()` functions will be only ever used for metadata generation (that is, regardless of the answer to the question posed by @gregtatcam , things may change in the future). IMO we should either add a new function with the required semantics and change other bits (i.e. where metadata is generated) to use it, or do something else.

I agree with this.

Here is what we did:

```
bool const recordDefaultAmounts = to.rules().enabled(fixXahauV1);
...
bool const shouldRecord =
                    (obj.getSType() == STI_AMOUNT && recordDefaultAmounts) ||
                    !obj.isDefault();
```

Bronek:
> > FWIW, I am uncomfortable with this change. Both `STCurrency` and `STIssue` are basic protocol building blocks, and because of this we cannot assume that `isDefault()` functions will be only ever used for metadata generation (that is, regardless of the answer to the question posed by @gregtatcam , things may change in the future). IMO we should either add a new function with the required semantics and change other bits (i.e. where metadata is generated) to use it, or do something else.
> 
> I agree with this.
> 
> Here is what we did:
> 
> ```
> bool const recordDefaultAmounts = to.rules().enabled(fixXahauV1);
> ...
> bool const shouldRecord =
>                     (obj.getSType() == STI_AMOUNT && recordDefaultAmounts) ||
>                     !obj.isDefault();
> ```

Having read https://github.com/XRPLF/rippled/issues/4965#issuecomment-2079688630 I think I was mistaken. The semantics of `isDefault()` is defined in `STBase` ("does the object hold any useful data ?"); neither `STIssue` nor `STCurrency` implement it correctly, as of today. Having said that, I also think that the proposed amendment is badly named, as this is not an AMM amendment. It is a currency/IOU metadata amendment.

dangell7:
> > > FWIW, I am uncomfortable with this change. Both `STCurrency` and `STIssue` are basic protocol building blocks, and because of this we cannot assume that `isDefault()` functions will be only ever used for metadata generation (that is, regardless of the answer to the question posed by @gregtatcam , things may change in the future). IMO we should either add a new function with the required semantics and change other bits (i.e. where metadata is generated) to use it, or do something else.
> > 
> > 
> > I agree with this.
> > Here is what we did:
> > ```
> > bool const recordDefaultAmounts = to.rules().enabled(fixXahauV1);
> > ...
> > bool const shouldRecord =
> >                     (obj.getSType() == STI_AMOUNT && recordDefaultAmounts) ||
> >                     !obj.isDefault();
> > ```
> 
> Having read [#4965 (comment)](https://github.com/XRPLF/rippled/issues/4965#issuecomment-2079688630) I think I was mistaken. The semantics of `isDefault()` is defined in `STBase` ("does the object hold any useful data ?"); neither `STIssue` nor `STCurrency` implement it correctly, as of today. Having said that, I also think that the proposed amendment is badly named, as this is not an AMM amendment. It is a currency/IOU metadata amendment.

I disagree as [STAmount](https://github.com/XRPLF/rippled/blob/40b4adc9cc296a7e3c6e8c94b5a977a54c835613/src/ripple/protocol/impl/STAmount.cpp#L701) clearly defines the `isDefault()` as `return (mValue == 0) && mIsNative;` so if that is incorrect then yeah it should just be false AND STAmount should be fixed as well, however if that is the correct default response then STIssue and STCurrency are correct in their response and this fix needs to be handled outside of the `isDefault()` behavior. 

But alas I really don't have any firm position on how its handled in ripple(d).

Bronek:
> `return (mValue == 0) && mIsNative;`

I think the condition that you are citing here just happens to match the definition of "no useful data here", in the context of `STAmount`.

src/ripple/protocol/Feature.h
```diff
@@ -360,6 +360,7 @@ extern uint256 const fixEmptyDID;
 extern uint256 const fixXChainRewardRounding;
 extern uint256 const fixPreviousTxnID;
 extern uint256 const fixAMMRounding;
+extern uint256 const fixAMMMetadata;
```
Bronek, line 363: Having read https://github.com/XRPLF/rippled/issues/4965#issuecomment-2079688630 I think this is a good fix, but I would like the amendment to use a different name. The use of `AMM` in name is misleading, this is a currency metadata fix (i.e. a lower abstraction level), which just happens to show up in AMM (but also in Oracle, as the unit tests demonstrate).

Bronek (changes requested):

src/ripple/protocol/Feature.h
Bronek: Could be also named as asset metadata, rather than currency metadata.

bthomee:
@shawnxie999 can you please rebase, resolve the conflicts, and address the open items in the comments, if any?

shawnxie999:
@bthomee I won't be able to get to this PR for a while.

vlntb:
@shawnxie999 Do you have time to return back to this PR?

vlntb:
@gregtatcam Is this change still relevant?

shawnxie999:
@vlntb Unfortunately, I won't be able to prioritize this PR for quite a while

gregtatcam:
> @gregtatcam Is this change still relevant?

Yes, it's still relevant.

godexsoft:
We recently merged a refactor to `develop` that enables clang-tidy's `readability-identifier-naming`. Your branch now has heavy conflicts that are largely mechanical. Below is a workflow that aligns your branch's naming with develop _before_ merging, which should minimize the merge conflicts.

## One-time setup

If you don't already have clang-tidy working in your env, on macOS:

```bash
brew install llvm@21
# Follow brew's hint to put $(brew --prefix llvm@21)/bin on PATH so run-clang-tidy is found.
```

## Workflow on your branch (before merging develop)

**1. Grab the new `.clang-tidy` from develop without pulling anything else.** Sync your fork on GitHub first, then:

```bash
git remote -v   # should show 'upstream' among others; if not:
# git remote set-url upstream git@github.com:XRPLF/rippled.git
git fetch upstream
git checkout upstream/develop -- .clang-tidy
```

**2. Reconfigure conan/cmake so `compile_commands.json` is fresh.**

**3. Apply renames for the files modified in your PR:**

```bash
git diff --name-only $(git merge-base HEAD upstream/develop) HEAD \
  | grep -E '\.(cpp|h|hpp|ipp)$' \
  | xargs run-clang-tidy -p build -fix -allow-no-checks
# or -p .build, or whatever your build dir is called
```

**4. Build + test, then commit as a single dedicated commit:**

```bash
cmake --build build -j8
git commit -am "refactor: Align identifier naming with develop"
```

**5. Now merge develop:**

```bash
git merge upstream/develop
```

## Extra

Run clang-tidy once more after the merge to catch any stragglers introduced from develop's side:

```bash
run-clang-tidy -p build -fix -allow-no-checks src tests
# or -p .build, or whatever your build dir is called
```

PR #5013, opened 2024-05-03
Price Oracle: validate input parameters and extend test coverage:

gregtatcam:

Validate trim, time_threshold, document_id are valid Int, UInt, or string convertible to UInt. Validate base_asset and quote_asset are valid currency. Update error codes. Extend Oracle and GetAggregatePrice unit-tests.
Denote unreachable coverage code.

### Type of Change

- [ ] Documentation update
- [ ] Release

## Before / After

Fixes invalid input parameters conversion to uint. For instance, 1.2 was converted to 1 if passed to trim, time_threshold, or document_id. This only allows valid UInt values. Similarly, base_asset and quote_asset were not validated to be a valid currency code. This fix only allows valid currency codes in base_asset and quote_asset.

## Test Plan

Extended Oracle and GetAggregatePrices tests.

### DIFF
--- a/src/ripple/rpc/handlers/GetAggregatePrice.cpp
+++ b/src/ripple/rpc/handlers/GetAggregatePrice.cpp
@@ -85,10 +85,15 @@ iteratePriceData(
 
         auto const ledger = context.ledgerMaster.getLedgerBySeq(prevSeq);
         if (!ledger)
+        {
+            // LCOV_EXCL_START
             return;
+            // LCOV_EXCL_STOP
+        }
 
         meta = ledger->txRead(prevTx).second;
 
+        prevChain = chain;
         for (STObject const& node : meta->getFieldArray(sfAffectedNodes))
         {
             if (node.getFieldU16(sfLedgerEntryType) != ltORACLE)
@@ -96,7 +101,6 @@ iteratePriceData(
                 continue;
             }
 
-            prevChain = chain;
             chain = &node;
             isNew = node.isFieldPresent(sfNewFields);
             // if a meta is for the new and this is the first
@@ -170,19 +174,47 @@ doGetAggregatePrice(RPC::JsonContext& context)
     if (!params.isMember(jss::quote_asset))
         return RPC::missing_field_error(jss::quote_asset);
 
+    // Lambda to validate uint type
+    // support positive int, uint, and a number represented as a string
+    auto validUInt = [](Json::Value const& params,
+                        Json::StaticString const& field) {
+        auto const& jv = params[field];
+        std::uint32_t v;
+        return jv.isUInt() || (jv.isInt() && jv.asInt() >= 0) ||
+            (jv.isString() && beast::lexicalCastChecked(v, jv.asString()));
+    };
+
     // Lambda to get `trim` and `time_threshold` fields. If the field
     // is not included in the input then a default value is returned.
-    auto getField = [&params](
+    auto getField = [&params, &validUInt](
                         Json::StaticString const& field,
                         unsigned int def =
                             0) -> std::variant<std::uint32_t, error_code_i> {
         if (params.isMember(field))
         {
-            if (!params[field].isConvertibleTo(Json::ValueType::uintValue))
-                return rpcORACLE_MALFORMED;
+            if (!validUInt(params, field))
+                return rpcINVALID_PARAMS;
             return params[field].asUInt();
         }
         return def;
+    };
+
+    // Lambda to get `base_asset` and `quote_asset`. The values have
+    // to conform to the Currency type.
+    auto getCurrency =
+        [&params](SField const& sField, Json::StaticString const& field)
+        -> std::variant<Json::Value, error_code_i> {
+        try
+        {
+            if (params[field].asString().empty())
+                return rpcINVALID_PARAMS;
+            currencyFromJson(sField, params[field]);
+            return params[field];
+        }
+        catch (...)
+        {
+            return rpcINVALID_PARAMS;
+        }
     };
 
     auto const trim = getField(jss::trim);
@@ -206,8 +238,18 @@ doGetAggregatePrice(RPC::JsonContext& context)
         return result;
     }
 
-    auto const& baseAsset = params[jss::base_asset];
-    auto const& quoteAsset = params[jss::quote_asset];
+    auto const baseAsset = getCurrency(sfBaseAsset, jss::base_asset);
+    if (std::holds_alternative<error_code_i>(baseAsset))
+    {
+        RPC::inject_error(std::get<error_code_i>(baseAsset), result);
+        return result;
+    }
+    auto const quoteAsset = getCurrency(sfQuoteAsset, jss::quote_asset);
+    if (std::holds_alternative<error_code_i>(quoteAsset))
+    {
+        RPC::inject_error(std::get<error_code_i>(quoteAsset), result);
+        return result;
+    }
 
     // Collect the dataset into bimap keyed by lastUpdateTime and
     // STAmount (Number is int64 and price is uint64)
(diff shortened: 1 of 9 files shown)


thejohnfreeman (approved):

src/ripple/rpc/handlers/GetAggregatePrice.cpp
```diff
@@ -170,21 +170,49 @@ doGetAggregatePrice(RPC::JsonContext& context)
...
-    auto getField = [&params](
+    auto getField = [&params, &validUInt](
                         Json::StaticString const& field,
                         unsigned int def =
                             0) -> std::variant<std::uint32_t, error_code_i> {
         if (params.isMember(field))
         {
-            if (!params[field].isConvertibleTo(Json::ValueType::uintValue))
-                return rpcORACLE_MALFORMED;
+            if (!validUInt(params, field))
```
mvadari, line 191: What is the difference between these two lines of code?

mvadari: If I'm reading the code correctly, it's that the old version accepted booleans and floats and would just convert them?

gregtatcam: `isConvertible` is 
>      return (other == nullValue && value_.uint_ == 0) ||
>                (other == intValue && value_.uint_ <= (unsigned)maxInt) ||
>               other == uintValue || other == realValue ||
>               other == stringValue || other == booleanValue;
which means it can convert a wider range of other type. For instance, it'll convert 1.2 to 1, which is invalid in this case.

gregtatcam: > If I'm reading the code correctly, it's that the old version accepted booleans and floats and would just convert them?

Correct.

src/ripple/rpc/handlers/GetAggregatePrice.cpp
```diff
@@ -85,18 +85,18 @@ iteratePriceData(
 
         auto const ledger = context.ledgerMaster.getLedgerBySeq(prevSeq);
         if (!ledger)
-            return;
+            return;  // LCOV_EXCL_LINE
 
         meta = ledger->txRead(prevTx).second;
 
+        prevChain = chain;
```
mvadari, line 92: What does moving this line do? I don't really understand what `prevChain` is doing, and there don't appear to be any tests checking this behavior.

src/ripple/rpc/handlers/GetAggregatePrice.cpp
```diff
@@ -170,21 +170,49 @@ doGetAggregatePrice(RPC::JsonContext& context)
...
 
+    // Lambda to get `base_asset` and `quote_asset`. The values have
+    // to conform to the Currency type.
+    auto getCurrency =
+        [&params](SField const& sField, Json::StaticString const& field)
+        -> std::variant<Json::Value, error_code_i> {
+        try
+        {
+            if (params[field].asString().empty())
+                return rpcINVALID_PARAMS;
```
mvadari, line 206: Isn't this check already included in `currencyFromJson`?

src/test/app/Oracle_test.cpp
```diff
@@ -224,27 +231,38 @@ struct Oracle_test : public beast::unit_test::suite
...
             oracle.set(UpdateArg{
                 .series = {{"XRP", "USD", 740, 1}},
-                .lastUpdateTime = testStartTime.count() + 400 + 301,
+                .lastUpdateTime = static_cast<std::uint32_t>(closeTime() + 311),
                 .err = ter(tecINVALID_UPDATE_TIME)});
             oracle.set(UpdateArg{.series = {{"XRP", "USD", 740, 1}}});
-            BEAST_EXPECT(
-                oracle.expectLastUpdateTime(testStartTime.count() + 450));
+            BEAST_EXPECT(oracle.expectLastUpdateTime(
+                static_cast<std::uint32_t>(testStartTime.count() + 450)));
```
mvadari, line 256: Is there a reason the other lines were switched to using `closeTime() but this one still uses `testStartTime`?

src/ripple/rpc/handlers/GetAggregatePrice.cpp
gregtatcam: This handles an unlikely case when `CreatedNode/ModifiedNode` not found in the inner loop for `AffectedNodes`. If this happens then `prevChain` is going to be equal to `chain` and the iteration stops. I don't know if it's possible to reproduce it unless the ledger state is changed in a way, which is not possible to do with the transactions.

mvadari: Does moving this line of code make any difference other than the slight efficiency improvement from running fewer times?

src/ripple/rpc/handlers/GetAggregatePrice.cpp
gregtatcam: Yes, it could be removed. It's just a bit faster.

src/ripple/rpc/handlers/GetAggregatePrice.cpp
gregtatcam: `prevChain` has to be assigned outside of the inner loop. Initially `prevChain` is `nullptr` and `chain` points to `oracle`.  If there is no `AffectedNodes` then both `prevChain` and `chain` are unchanged and the check `(prevChain == chain)` is not going to fail.

src/test/jtx/Oracle.h
```diff
@@ -39,16 +55,16 @@ using DataSeries = std::vector<std::tuple<
 struct CreateArg
 {
     std::optional<AccountID> owner = std::nullopt;
-    std::optional<std::uint32_t> documentID = 1;
+    std::optional<AnyValue> documentID = 1;
```
mvadari, line 58: Why are these types becoming less specific? Isn't that checked at the validation level?

src/test/app/Oracle_test.cpp
```diff
@@ -622,6 +678,39 @@ struct Oracle_test : public beast::unit_test::suite
         }
     }
 
+    void
+    testInvalidLedgerEntry()
```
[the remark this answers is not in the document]
gregtatcam: Thanks. It should be moved.

src/test/jtx/Oracle.h
gregtatcam: I wanted to be able to test a wider range of valid/invalid values similar to how QA does it. Unfortunately the test framework does its own validation and will not call rpc if the parameters don't match their defined SField type. The validation is done because the transaction is signed. I left this change so that eventually can extend the testing to support unsigned transactions.

src/test/app/Oracle_test.cpp
gregtatcam: The two lines that used `testStartTime` should have not actually used it because `testStartTime` is already added in `Oracle::set` to the passed in `LastUpdateTime` value.

src/test/jtx/impl/Oracle.cpp
```diff
@@ -156,26 +158,34 @@ Oracle::aggregatePrice(
...
+        toJson(jv[jss::base_asset], *baseAsset);
     if (quoteAsset)
-        jv[jss::quote_asset] = *quoteAsset;
+        toJson(jv[jss::quote_asset], *quoteAsset);
     if (timeThreshold)
-        jv[jss::time_threshold] = *timeThreshold;
-
-    auto jr = env.rpc("json", "get_aggregate_price", to_string(jv));
+        toJson(jv[jss::time_threshold], *timeThreshold);
+    // Convert "%None%" to None
```
mvadari, line 177: Should`"%None%" be a global test constant, since it's a special case?

mvadari (approved):

PR #5016, opened 2024-05-13
Fix offer crossing with transfer fee over AMM and other fixes

gregtatcam:

This PR addresses three issues:
- Offer crossing with the transfer fee via AMM 
- LPToken adjustment to factor in the LPTokenBalance rounding
- Possible conflict between Liquidity Provider trustline LPToken balance
   and LPTokenBalance when last Liquidity Provider withdraws all tokens

### Type of Change

- [ ] Documentation update
- [ ] Release

## Before / After

1. `qualityUpperBound()` and `QualityFunction` calculation doesn't factor in the transfer fee on offer crossing.
    This is correct behavior when crossing an account offer. Single-path with AMM limits the `out` value based 
    on the limitQuality. If the transfer fee is not factored in then the out value is going to be such that
    the generated AMM offer quality matches the limitQuality.  But the effective quality is smaller because 
    `takerPays` amount factors in the transfer fee when the strand's quality is calculated. The fix factors in
    the transfer fee when the quality is adjusted for the fees on offer crossing.
2. `adjustAmountsByLPTokens()` function is not returning adjusted tokens and amounts in some cases.
    The fix returns the adjusted LPToken and adjusted amounts.
3.  Liquidity Provider's (LP) LPToken trustline balances might not add up to the total LPTokenBalance due to the rounding. 
     This results in a dust amount left in the balances on the last LP withdrawal or the withdrawal transaction failing. The fix
     sets the LPTokenBalance to the trustline balance if LP is the only remaining LP on withdrawal transaction.

## Test Plan

Updated AMM_test and AMMExtended unit-tests.

### DIFF
--- a/src/ripple/app/misc/impl/AMMHelpers.cpp
+++ b/src/ripple/app/misc/impl/AMMHelpers.cpp
@@ -167,7 +167,7 @@ adjustAmountsByLPTokens(
     {
         bool const ammRoundingEnabled = [&]() {
             if (auto const& rules = getCurrentTransactionRules();
-                rules && rules->enabled(fixAMMRounding))
+                rules && rules->enabled(fixAMMv1_1))
                 return true;
             return false;
         }();
--- a/src/ripple/app/misc/impl/AMMHelpers.cpp
+++ b/src/ripple/app/misc/impl/AMMHelpers.cpp
@@ -165,6 +165,13 @@ adjustAmountsByLPTokens(
 
     if (lpTokensActual < lpTokens)
     {
+        bool const ammRoundingEnabled = [&]() {
+            if (auto const& rules = getCurrentTransactionRules();
+                rules && rules->enabled(fixAMMRounding))
+                return true;
+            return false;
+        }();
+
         // Equal trade
         if (amount2)
         {
@@ -172,10 +179,14 @@ adjustAmountsByLPTokens(
             auto const amountActual = toSTAmount(amount.issue(), fr * amount);
             auto const amount2Actual =
                 toSTAmount(amount2->issue(), fr * *amount2);
-            return std::make_tuple(
-                amountActual < amount ? amountActual : amount,
-                amount2Actual < amount2 ? amount2Actual : amount2,
-                lpTokensActual);
+            if (!ammRoundingEnabled)
+                return std::make_tuple(
+                    amountActual < amount ? amountActual : amount,
+                    amount2Actual < amount2 ? amount2Actual : amount2,
+                    lpTokensActual);
+            else
+                return std::make_tuple(
+                    amountActual, amount2Actual, lpTokensActual);
         }
 
         // Single trade
@@ -183,13 +194,19 @@ adjustAmountsByLPTokens(
             if (isDeposit)
                 return ammAssetIn(
                     amountBalance, lptAMMBalance, lpTokensActual, tfee);
-            else
+            else if (!ammRoundingEnabled)
                 return withdrawByTokens(
                     amountBalance, lptAMMBalance, lpTokens, tfee);
+            else
+                return withdrawByTokens(
+                    amountBalance, lptAMMBalance, lpTokensActual, tfee);
         }();
-        return amountActual < amount
-            ? std::make_tuple(amountActual, std::nullopt, lpTokensActual)
-            : std::make_tuple(amount, std::nullopt, lpTokensActual);
+        if (!ammRoundingEnabled)
+            return amountActual < amount
+                ? std::make_tuple(amountActual, std::nullopt, lpTokensActual)
+                : std::make_tuple(amount, std::nullopt, lpTokensActual);
+        else
+            return std::make_tuple(amountActual, std::nullopt, lpTokensActual);
     }
 
     assert(lpTokensActual == lpTokens);
--- a/src/ripple/app/misc/impl/AMMUtils.cpp
+++ b/src/ripple/app/misc/impl/AMMUtils.cpp
@@ -348,4 +348,86 @@ initializeFeeAuctionVote(
         auctionSlot.makeFieldAbsent(sfDiscountedFee);  // LCOV_EXCL_LINE
 }
 
+Expected<bool, TER>
+isOnlyLiquidityProvider(
+    ReadView const& view,
+    Issue const& ammIssue,
+    AccountID const& lpAccount)
+{
+    // Liquidity Provider (LP) must have one LPToken trustline
+    std::uint8_t nLPTokenTrustLines = 0;
+    // There are at most two IOU trustlines. One or both could be to the LP
+    // if LP is the issuer, or a different account if LP is not an issuer.
+    // For instance, if AMM has two tokens USD and EUR and LP is not the issuer
+    // of the tokens then the trustlines are between AMM account and the issuer.
+    std::uint8_t nIOUTrustLines = 0;
+    // There is only one AMM object
+    bool hasAMM = false;
+    // AMM LP has at most three trustlines and only one AMM object must exist.
+    // If there are more than five objects then it's either an error or
+    // there are more than one LP. Ten pages should be sufficient to include
+    // five objects.
+    std::uint8_t limit = 10;
+    auto const root = keylet::ownerDir(ammIssue.account);
+    auto currentIndex = root;
+
+    // Iterate over AMM owner directory objects.
+    while (limit-- >= 1)
+    {
+        auto const ownerDir = view.read(currentIndex);
+        if (!ownerDir)
+            return Unexpected<TER>(tecINTERNAL);  // LCOV_EXCL_LINE
+        for (auto const& key : ownerDir->getFieldV256(sfIndexes))
+        {
+            auto const sle = view.read(keylet::child(key));
+            if (!sle)
+                return Unexpected<TER>(tecINTERNAL);  // LCOV_EXCL_LINE
+            // Only one AMM object
+            if (sle->getFieldU16(sfLedgerEntryType) == ltAMM)
+            {
+                if (hasAMM)
+                    return Unexpected<TER>(tecINTERNAL);  // LCOV_EXCL_LINE
+                hasAMM = true;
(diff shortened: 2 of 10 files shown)


src/ripple/app/misc/impl/AMMUtils.cpp
```diff
@@ -348,4 +348,56 @@ initializeFeeAuctionVote(
...
+            auto const highLimit = sle->getFieldAmount(sfHighLimit);
+            if ((lowLimit.getIssuer() != lpAccount &&
+                 highLimit.getIssuer() != lpAccount))
+                return false;
+            ++nTrustLines;
+        }
+        auto const uNodeNext = ownerDir->getFieldU64(sfIndexNext);
+        if (uNodeNext == 0)
+            return nTrustLines == 2 || nTrustLines == 3;
+        currentIndex = keylet::page(root, uNodeNext);
```
scottschurr, line 398: This linear search is kinda terrifying.  I looked in a recent ledger and the largest number of owner directory pages owned by a single account was 284,820.  But there were quite a few accounts with over 10,000 owner directory pages.  This linear search signs us up for a potentially huge amount of computing.  Calling this function from a transactor looks really scary to me.

Is there anything we know about AMM Liquidity Providers that will limit the number of directory pages that they can own?  Is there any way that could be leveraged to limit this linear search to just a few directory nodes?

gregtatcam: There are two things in this function that prevent this possible huge number of iterations. One is that the limit is set to at most 10 iterations. Two is that if there is more than one Liquidity Provider (LP) then we'll break out of the loop a lot sooner because any LP can have at most three trustlines to AMM: one for LPToken and at most two for IOU. So we should break out after at most four iterations in the worst case scenario if there are more than one LP.
   
    if ((lowLimit.getIssuer() != lpAccount &&
                 highLimit.getIssuer() != lpAccount))
                return false;

scottschurr: Sorry.  I missed what `limit` was up to when I first looked at the `while` loop.  But I still don't understand why 10 directory pages is a legit upper bound.

Owner directories can hold a lot of things other than trust lines.  The number of `Offer`s (both IOU and NFT), `Escrow`s, `PayChan`s, and `DepositPreauth`s they can hold has a very large upper bound.  Additionally, an account can own up to 250 tickets.

Now, maybe it's impossible for an AMM Liquidity Provider account to own any of these kinds of objects.  In which case 10 iterations is (probably?) plenty.  But I am not familiar enough with the AMM to know whether that's the case.  If it _is_ the case I think there should be a comment explaining why 10 iterations will always be sufficient.

There should probably also be some kind of error indication if the number of expected directory pages is ever surpassed.  It may not be possible for an AMM Liquidity Provider to ever exceed that number of pages today.  But people are excited about AMMs, which means rules will be changing in the future.  So either the function must throw or the return value needs to be more complicated than a `bool`.  Maybe an `Expected`?

Does that make any sense?

gregtatcam: We are iterating over the trustlines for AMM account. AMM account holds one AMM object and the rest are LP trustlines. Each LP has one trustline for LPToken. If an issuer created AMM then the issuer could have two more trustlines for IOU. Actually the function should return `true` if trustlines >= 1 && truslines <= 3, not 2 or 3.

It makes sense to return an error in some cases: there are 0 trustlines, non-issuer LP has more than 1 trustline, an issuer LP has more than 3 trustlines, there are more than 1 AMM object. `Expected` with `tecINTERNAL` works.

gregtatcam: I think it is sufficient to just look for LPToken trustlines and return an error if it is 0 or greater than 1.

scottschurr: I just did a little more reading about the AMM implementation.  I'm sorry that a) I'm not really up on how the AMM works and b) that I did not look more closely at the loop before I commented.

Given the constraints as I (finally) understand them, we can have high confidence that the number of loops will be very bounded by adding one constraint to the loop.  If `nTrustLines > 3` return `false`.  I think if that constraint is added then (assuming there's only one AMM object) the number of directory pages examined cannot exceed 6.

src/test/jtx/impl/AMMTest.cpp
```diff
@@ -103,44 +103,45 @@ AMMTestBase::testAMM(
     std::optional<std::pair<STAmount, STAmount>> const& pool,
     std::uint16_t tfee,
     std::optional<jtx::ter> const& ter,
-    std::optional<FeatureBitset> const& features)
+    std::vector<FeatureBitset> const& vfeatures)
```
seelabs, line 106: I don't love this interface - running the same test with a different set of features seems like you'd want another function call (with possibly another `ter`). I'm not that picky about test code, so no change needed if you prefer this.

src/ripple/app/misc/impl/AMMUtils.cpp
seelabs: Yeah, I don't love this. I don't see a better solution though.

seelabs (commented):
This looks OK to me, but I want to take another look tomorrow before I approve it.

src/ripple/app/misc/impl/AMMHelpers.cpp
```diff
@@ -165,31 +165,48 @@ adjustAmountsByLPTokens(
 
     if (lpTokensActual < lpTokens)
     {
+        bool const ammRoundingEnabled = [&]() {
+            if (auto const& rules = getCurrentTransactionRules();
+                rules && rules->enabled(fixAMMRounding))
```
scottschurr, line 170: When this is merged with develop will `fixAMMRounding` be an additional amendment?  Or will you be leveraging the recently added `fixAMMv1_1`?  I'm fine either way, but it's worth being aware of the merge consequences.

gregtatcam: I'll leverage `fixAMMv1_1`.

src/ripple/app/tx/impl/AMMWithdraw.cpp
```diff
@@ -310,6 +310,26 @@ AMMWithdraw::applyGuts(Sandbox& sb)
...
+    // might not match the LP's trustline balance
+    if (sb.rules().enabled(fixAMMRounding))
+    {
+        if (auto const res =
+                isOnlyLiquidityProvider(sb, lpTokens.issue(), account_);
+            !res)
+            return {res.error(), false};
+        else if (
+            res.value() &&
+            withinRelativeDistance(
```
seelabs, line 323: I'm not understanding the reasoning behind the `withinRelativeDistance` check. If it's the only trust line why do we check if there's a relative distance? I'm guessing this is as a sanity check. But I would probably make the sanity check looser if this is the case.

gregtatcam: Right, it's a sanity check and I made it more restrictive. Do you feel e-7 is sufficient?

src/ripple/app/misc/impl/AMMHelpers.cpp
gregtatcam: I need to merge with the other AMM PR #4983.

src/ripple/app/tx/impl/AMMWithdraw.cpp
seelabs: I don't think there's a value where we can safely say "the sanity check must always be below this value". This sanity check is really protecting against a bug in `isOnlyLiquidityProvider`. As long as that's solid we don't need the sanity check. Thinking...

If this is the last remaining trust line and the sanity check fails, what should happen? If we return a `tec` we may be saving ourselves against an exploit/bad bug at the cost of not allowing the last person to get their funds. If we don't have a check, we open ourselves up to bugs in `isOnlyLiqueidityProvider`. If we keep it as-is, then what? Isn't that still a bug if the sanity check fails and we keep the code as-is?

gregtatcam: The only reason to have the sanity check is in case `isOnlyLiquidityProvider` is not solid. I think it looks right but then essentially all tokens are assigned to the user. I don't see a way around it.

If the sanity check fails then it's still a bug, only more obscure. The transaction might succeed or might fail. It might make sense to return `tec`. The user can still withdraw with a fewer tokens.

seelabs: I vote that we loosen the constraint considerably - maybe 10^-1? 10^-2? - and also change this to return a tec if the sanity check fails.

gregtatcam: 10^-2 is 1%. This is pretty big difference, no?

👍 `tec`.

seelabs: Yes, I think I'm voting for a very large difference. I'm fine if you disagree and want a tighter bound, but given we can't but a bound on the difference I'd rather have it large. Let's get @scottschurr 's opinion on this too.

gregtatcam: I was thinking e-4 at least to prevent/minimize a possible exploit if there is one. I'm fine with e-2 if everyone votes for it.

seelabs: It's your call. I'm also find with e-4.

gregtatcam: Let's wait for @scottschurr vote.

scottschurr (approved):
:+1: Looks good to me.  I found one typo in a comment, but nothing that critically needs to be changed as far as I was able to see.

src/ripple/app/tx/impl/AMMWithdraw.cpp
gregtatcam: Let's do e-3.

PR #5021, opened 2024-05-20
Fix Oracle's token pair deterministic order

gregtatcam:

Price Oracle data-series logic uses `unordered_map` to update the Oracle object. This results in different servers disagreeing on the order of that hash table. Consequently, the generated ledgers will have different hashes. The fix uses `map` instead to guarantee the order of the token pairs in the data-series.

### Type of Change

- [ ] Documentation update
- [ ] Release


scottschurr (approved):
Fix looks good to me.  The fix builds, passes unit tests, syncs, and stops for me on macOS.  :+1:

seelabs (approved):

PR #5025, opened 2024-05-21
Fix gcc-14 build

oleks-rip:

Fix fstream usage to be compatible with gcc-14

### Context of Change

gcc-14  require explicit definition of the fstream classes

### Type of Change

### API Impact

none


Bronek:
Note, this broke both MacOS and Windows build, because of no matching constructor for `std::ifstream` (and `std::ostream`) from `boost::filesystem::path` . Try adding `.native()` calls.

oleks-rip:
Thank you. I've updated to  2.2.0-rc3 and everything seems ok, without this fix. I'll close it.

PR #5026, opened 2024-05-22
Expose all amendments known by libxrpl

godexsoft:

There is no way for consumers like Clio to get all known amendments/features thru `libxrpl`.
This PR adds a public function to do just that as well as some tests for it.

Please note that this adds only what is needed for Clio and nothing more. If there is desire to expose the raw `VoteBehavior` it can be added. Let me know.

### Context of Change

Clio needs the ability to get all known amendments in order to implement the `feature` API (non-admin part).

### Type of Change

### API Impact

None.


src/ripple/protocol/Feature.h
```diff
@@ -67,6 +68,11 @@
 namespace ripple {
 
 enum class VoteBehavior : int { Obsolete = -1, DefaultNo = 0, DefaultYes };
+enum class AmendmentSupport : int { Retired = -1, Supported = 0, Unsupported };
+
+/** All amendments libxrpl knows about. */
+std::unordered_map<std::string, AmendmentSupport> const&
```
godexsoft, line 74: Note that i used `unordered_map` here because for Clio's usecase the order is not important. 
But i'm now thinking that perhaps there will be a usecase in the future where it could come in handy if the names came out sorted. What do you think?

intelliot: From a cost/benefit perspective, it's hard for me to imagine that the perf degradation of using a sorted map would be significant for amendments. It's O(log n) versus O(1) and n is small, maybe < 100, and certainly < 1000. So yes, I think it would be nice for the names to come out sorted.

godexsoft: Agreed. After sleeping on it I also think it's better to keep the API similar to what it was, so definitely changing this back to match.

intelliot (approved):
Untested; LGTM

seelabs (approved):

PR #5027, opened 2024-05-23
Add new command line option to make replaying transactions easier

Bronek:

Add new command line option `trap_tx_hash` to make replaying transactions easier

This new option can be used only if `replay` is also enabled. It takes a transaction hash from the ledger loaded for replay, and will cause a specific line to be hit in `Transactor::operator()`, right before `apply()`.

Emit an error if:
- the new option is used without `replay`
- the supplied parameter cannot be parsed as transaction hash (i.e. `uint256` in hex)
- the supplied transaction hash is not found on the ledger being replayed

### Context of Change

While replaying specific ledger is relatively easy, things get more difficult if we want to debug a specific transaction within the ledger. There are several options, but all of them are a hassle. Why not just a new option to `rippled` to hit a very specific line of code, right before a specific transaction on the replayed ledger is being applied.

This change adds new virtual function `std::optional<uint256> trapTxID() const` to `ripple::Application` class. This new function is called inside `Transactor::operator()` to check if:
- trap transaction is set
- the currently processed transaction is a match

If both conditions are met, `rippled` will simply log `Transaction trapped` followed by a hash. The user can put a breakpoint at this line to break into the debugger.

### Type of Change

- [ ] Documentation update
- [ ] Release


Bronek:
I will try to add add unit test for this change, since apparently `Application` class is usable in unit tests.

src/ripple/app/main/Application.cpp
```diff
@@ -1254,6 +1255,12 @@ class ApplicationImp : public Application, public BasicApp
         return maxDisallowedLedger_;
     }
 
+    virtual std::optional<uint256>
```
exunda, line 1258: I would return a `const std::optional<uint256>&` instead of a copy

src/ripple/app/main/Application.cpp
```diff
@@ -2233,6 +2246,11 @@ ApplicationImp::loadOldLedger(
             {
                 (void)_;
                 auto txID = tx->getTransactionID();
+                if (trapTxID && *trapTxID == txID)
```
exunda, line 2249: You can just write `if(trapTxID == txID)` here

src/ripple/app/tx/impl/Transactor.cpp
```diff
@@ -854,6 +854,12 @@ Transactor::operator()()
     }
 #endif
 
+    if (auto trap = ctx_.app.trapTxID();
+        trap && ctx_.tx.getTransactionID() == *trap)
+    {
+        JLOG(j_.debug()) << "Transaction trapped: " << *trap;
+    }
```
exunda, line 861: After you change the function to retun a `const std::optional<uint256>&` you can just wrtite

    if(ctx_.app.trapTxID() == ctx.tx.getTransactionID())
        JLOG(j_.debug()) << "Transaction trapped: " << ctx.tx.getTransactionID();

src/ripple/core/Config.h
```diff
@@ -162,6 +162,8 @@ class Config : public BasicConfig
 
     std::string START_LEDGER;
 
+    std::optional<uint256> TRAP_TX_HASH = {};
```
exunda, line 165: You do not need squigly brackets here

src/ripple/app/main/Application.cpp
Bronek: Good idea.

src/ripple/core/Config.h
Bronek: I like initializers to be explicit, it makes reading easier.

src/ripple/app/tx/impl/Transactor.cpp
Bronek: On second thought, I think it's not such a good idea, as it will mean unconditional copying of `uint256` from `ctx.tx` for the purpose of equality comparison. I've changed it back again to enable bool shortcircuit evaluation, where in normal production use we only check if the trap is set and that's all. Of course that means more branching, but in normal production use we have one branch anyway.

exunda: I  looked at this more carefully because no copying should happen given overload 21 on https://en.cppreference.com/w/cpp/utility/optional/operator_cmp but then I looked closer and noticed that `STTx::getTransactionID` returns a copy 🤦  

I agree that because of that it will copy and I suggest you fix that code to return a `const uint256&` instead!

Bronek: Thank you for the suggestion, however such a change does not belong to this PR.

thejohnfreeman (approved):

src/ripple/app/main/Application.cpp
```diff
@@ -2247,6 +2265,14 @@ ApplicationImp::loadOldLedger(
             }
 
             m_ledgerMaster->takeReplay(std::move(replayData));
+
+            if (trapTxID && !trapTxID_)
+            {
+                JLOG(m_journal.fatal())
+                    << "Ledger " << replayLedger->info().seq
+                    << " does not contain the transaction hash " << *trapTxID;
+                return false;
```
gregtatcam, line 2274: Should we return `false` in this case or just log? The semantics of this call is to return `false` if the ledger fails to load.

Bronek: Yes, that's intentional - you only use this functionality when replaying a transaction, and failing a little earlier (still need several minutes to load that ledger) will save your time.

src/test/app/LedgerLoad_test.cpp
```diff
@@ -199,6 +233,99 @@ class LedgerLoad_test : public beast::unit_test::suite
...
+                Config::REPLAY,
+                std::nullopt),
+            nullptr,
+            beast::severities::kDisabled);
+        auto const jrb = env.rpc("ledger", "current", "full")[jss::result];
+        BEAST_EXPECT(jrb[jss::ledger][jss::accountState].size() == 97);
+
+        env.close();
+        auto const closed = env.rpc("ledger", "current", "full")[jss::result];
+        BEAST_EXPECT(closed[jss::ledger][jss::accountState].size() == 98);
```
gregtatcam, line 260: What is the meaning of the tests for 97, 98, "<=", etc? Can you add some comments to clarify. Just adding the comments in one code block would be useful.

src/ripple/core/Config.h
seelabs: Most of the rest of rippled doesn't do this. It's a nit, but I prefer the version without the explicit initialization (for example, the string above isn't initialized like this). Let's not bike shed this too much either - it's not a huge deal one way or the other, and I understand your preference, but let's stick to relying on default constructors for now.

seelabs (commented):
I'm a little torn on this patch. On the one hand, this makes it easier to replay transactions without needing to set conditional breakpoints. On the other hand, knowing how to set conditional breakpoints is a skill I'd expect anyone wanting to replay transactions to have. On the other hand, this is pretty simple code, so adding it isn't too large of a cost.

If other people find this useful I think it's OK to add.

seelabs (approved):
:+1:

src/test/app/LedgerLoad_test.cpp
Bronek: Added comment to explain - in `replay` mode, the ledger being replayed needs to be explicitly accepted e.g. with `env.close()` . Before this call, we are getting an earlier ledger as `current`, with one fewer state objects.

PR #4968, opened 2024-03-26
Proposed 2.1.1

seelabs:
## High Level Overview of Change

This is a hotfix release that fixes a critical bug when handling of a large synthetic AMM offers in the payment engine.

### Type of Change

- [ x] Bug fix (non-breaking change which fixes an issue)
- [ x] Release

PR #4870, opened 2024-01-08
Proposed 2.0.0

intelliot:
This is the proposed version 2.0.0 of `rippled`, the reference server implementation of the XRP Ledger protocol. This release adds new features and bug fixes, and introduces these amendments:

- `DID`
- `XChainBridge`
- `fixDisallowIncomingV1`
- `fixFillOrKill`

[Sign Up for Future Release Announcements](https://groups.google.com/g/ripple-server)

PR #4615, opened 2023-07-10
Adapt to change in Conan recipe for NuDB

thejohnfreeman:

PR #5028, opened 2024-05-24
Upload libXRPL flow

godexsoft:

Adding a workflow that automatically publishes `libxrpl` to a custom channel (`clio/pr_NUMBER`) Artifactory and triggers a workflow in Clio's CI using `repository_dispatch`. 

This flow will only run if the `BuildInfo.cpp` file or the workflow itself was updated in a PR. Typically this means a new version is proposed. 

Publishing the library as soon as possible (i.e. not after merging it or tagging the version) gives us the ability to notify Clio's CI and trigger a full check against the new version of `libxrpl`. In the future we could automate a notification back from Clio to report whether the checks were successful or not.

This flow will start together with others (if conditions are met) but will not proceed with the upload until essential other checks are done and are either successful or skipped (not cancelled or failed). Currently the _essential_ checks are all `test` running on `linux`. Mac and Windows are not on the list. Main reason for that is Mac being flaky and we don't want to fail this step (at least not often).

Two secrets that need to be present in the repo (are already present):
- CONAN_USERNAME
- CONAN_TOKEN

And a new one that *already was added* in order to trigger a workflow on Clio's CI:
- CLIO_NOTIFY_TOKEN

Here is the flow in Clio CI that does the rest: https://github.com/XRPLF/clio/pull/1433
Currently Clio's side is broken because Conan's channel syntax contains a forward slash - a symbol that is not allowed for artifact names. This will be fixed on Clio side asap (https://github.com/XRPLF/clio/pull/1583).

### Context of Change

Recently there was an `rc` of `libxrpl` that did not allow consumers to link correctly. We should try and avoid this in the future.

### Type of Change

### API Impact

None.


.github/workflows/libxrpl.yml
```diff
@@ -0,0 +1,86 @@
...
+env:
+  CONAN_URL: http://18.143.149.228:8081/artifactory/api/conan/conan-non-prod
+  CONAN_LOGIN_USERNAME_RIPPLE: ${{ secrets.CONAN_USERNAME }}
+  CONAN_PASSWORD_RIPPLE: ${{ secrets.CONAN_TOKEN }}
+on:
+  pull_request:
+    paths:
+      - 'src/ripple/protocol/impl/BuildInfo.cpp'
+      - '.github/workflows/libxrpl.yml'
+    branches: [release]
```
thejohnfreeman, line 11: - [Here](https://github.com/XRPLF/rippled/pull/4968) is a patch release PR that targeted `master`.
- [Here](https://github.com/XRPLF/rippled/pull/4870) is a major release PR that targeted `develop`.

Do we channel all of our releases through a single branch? I don't know. Even if we do, I'm not sure we can count on it always happening through a PR. I understand you want to publish a package before the commit is pushed to a protected branch or tagged. Can we get an expert on the release process to chime in? @seelabs?

Once we determine the right condition to check, we can think about how to deduplicate some of the steps here. Maybe we can move this workflow into a single conditional step at the end of the test workflow.

thejohnfreeman (changes requested):

.github/workflows/libxrpl.yml
godexsoft: Would be indeed useful to know if `release` is supposed to always be used for this purpose @seelabs 👍 

When implementing this i deliberately sacrificed duplication for non-intrusiveness. I agree we can deduplicate quite a bit but i thought we would be better off creating a separate issue/PR for that. Let me know what you think @thejohnfreeman @seelabs .

thejohnfreeman: I've added a workflow in a [branch](https://github.com/thejohnfreeman/rippled/actions/runs/9484998818/job/26139475923) that builds the Conan package locally and tests that an example downstream package can import it successfully. We could add a job to upload the recipe (and binary?) conditional on that test (and whatever conditions we decide determine a release).

godexsoft: Not sure how useful it is to build an "example" package. Unless you have a package that uses everything libxrpl provides of course, in which case that's great. 
I only really care that Clio (the only real/important consumer afaik) still works so I'd certainly prefer to leave the notification step that triggers Clio's CI. If you want to add an example package build in between - i have no problem with that. 
I personally prefer to get simple things like CI out of the way and functional quicker and then improve if needed later.

thejohnfreeman: It is useful to test that, at least, `libxrpl` can be found and linked by a dependent package. It is possible to build a Conan package that doesn't actually export anything that can be linked.

thejohnfreeman: Clio could be our example package, instead, if we want to make it so. But the test that Clio can build against the `xrpl` Conan package must be moved to this project, so that it is a blocker for our commits. The example package I wrote is much simpler, but it is in this project. That is the key attribute.

godexsoft: Well, i personally don't like the idea of building Clio from this repo - that's as intrusive as it gets. But i do see your points. Please do what you want with this PR or let's discuss next steps next week.

godexsoft: @seelabs can we get your input on the release process please?

seelabs: @thejohnfreeman @godexsoft: 
 
You can rely on releases and release candidates always being pushed to `release`. In most cases, you can rely on betas being pushed to `release` as well, but that may not happen if there are `develop` has diverged from `release` and we aren't merging betas into release until we transition the release candidate to release.

Here's the process that happens when `develop` and `master` have not diverged (usually the branches have not diverged, but this can happen for hot fixes or when a rc is active and we merge commits to develop that shouldn't be in the rc):

* betas and release candidates are merged to `develop` and `release`
* releases are merged to `develop`, `release` and `master`
 
I understand that happened with the old process when `develop` and `master` diverged, but it's less clear to me what happens in the new process. Here's my understanding for the new process if release and develop have diverged:

* betas to to `develop`, but not master
* release candidates to to `release` but not `develop`
* once a release candidate transitions to a release, we merge the `release` branch into `develop` with a merge commit.
* once the reverse happens, new betas will be merged to `master` again until the branches diverge

We usually open beta PRs and rc PRs against the `release` branch, but sometimes the PR is accidentally opened against the `develop` branch and if the branches haven't diverged there's no harm.

thejohnfreeman: It sounds like the two important conditions are:

1. All releases will pass through the `release` branch.
2. For Clio to be notified before a release actually lands, the push to `release` must first pass through a PR.

In that case, the conditions of the workflow in this PR are already correct.

I think the workflows are ripe for a re-factoring after this, but it doesn't need to be a blocker.

thejohnfreeman (approved):

.github/workflows/libxrpl.yml
godexsoft: If everyone are happy we can merge this but we will need to setup an account (and a token from that) with access to execute the remote repo workflow.

thejohnfreeman: Are you sure a new account is required? I think when you create the PAT, you can just set the "resource owner" field to the XRPLF organization and limit its scope to the clio respository.

godexsoft: No, it's not required. But if i create a PAT from say my account it will probably use my account when creating PRs/issues from that workflow.
It's the easiest option.. but perhaps having some "XRPLD[bot]" account would be preferred?

godexsoft: or would setting "resource owner" change the PR/issue author, do you know?

thejohnfreeman: I don't know. It can be tested though. Is there someone with organization permissions who we can delegate this to?

godexsoft: I have evidence that it will still use the `github-actions` account to create the PR even if i use a PAT created for my account. I can create one and ask @ximinez to set it up in this repo.

ximinez: Not all releases are guaranteed to pass through `release`. Specifically, a hotfix (e.g. [2.2.1](https://github.com/XRPLF/rippled/pull/5078)) can go directly to `master` without touching `develop` or `release` until it is backported. Also the `paths` and `branches` conditions [are essentially `AND`ed together](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#running-your-pull_request-workflow-based-on-the-head-or-base-branch-of-a-pull-request). No changes should ever get to `release` (or `master` for that matter) without the version number changing in `BuildInfo.cpp`, which would make that check redundant. Conversely, `BuildInfo.cpp` is almost never changed except to change the version number.

This contradicts what @seelabs said above. Take the example of 2.2.1, the reason is simple: The issue wasn't discovered in 2.2.0 until well after the first 2.3.0 beta had been merged to `release`. At that point, the only options were either roll back `release` (very bad), or let `master` and `release` diverge (not so bad).

Therefore, I suggest either:
* Remove `paths`, and change `branches` to use both `release` and `master`. 
* Remove `branches`, and leave `paths` as-is.

Both risk adding extra builds, but drastically reduce the chances of missing a build, which seems like it would be worse.

ximinez (changes requested):

.github/workflows/libxrpl.yml
```diff
@@ -0,0 +1,85 @@
+name: Upload libXRPL to Conan
```
ximinez, line 1: What do you think of changing the name to something like "Check libXRPL compatibility with Clio"?

.github/workflows/libxrpl.yml
```diff
@@ -0,0 +1,85 @@
...
+      version: ${{ steps.version.outputs.version }}
+    runs-on: [self-hosted, heavy]
+    container: rippleci/rippled-build-ubuntu:aaf5e3e
+    steps:
+      - name: Wait for essential checks to succeed
+        uses: lewagon/wait-on-check-action@v1.3.4
+        with:
+          ref: ${{ github.event.pull_request.head.sha || github.sha }}
+          running-workflow-name: wait-for-check-regexp
+          check-regexp: '(dependencies|test).*linux.*' # Ignore windows and mac tests but make sure linux passes
```
ximinez, line 29: You could probably simplify this to just check `test`, since `test` depends on `dependencies`.

.github/workflows/libxrpl.yml
```diff
@@ -0,0 +1,85 @@
...
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
+          wait-interval: 10
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Unlock Conan
+        shell: bash
+        run: conan remove --locks
```
ximinez, line 36: The environment should always be reset after each job, so `conan` shouldn't be locked. Has this been an issue?

ximinez (changes requested):

.github/workflows/libxrpl.yml
```diff
@@ -0,0 +1,85 @@
+name: Upload libXRPL to Conan
+env:
+  CONAN_URL: http://18.143.149.228:8081/artifactory/api/conan/conan-non-prod
+  CONAN_LOGIN_USERNAME_RIPPLE: ${{ secrets.CONAN_USERNAME }}
+  CONAN_PASSWORD_RIPPLE: ${{ secrets.CONAN_TOKEN }}
+on:
+  pull_request:
+    paths:
+      - 'src/ripple/protocol/impl/BuildInfo.cpp'
```
ximinez, line 9: This file is is moved to `src/libxrpl/protocol/BuildInfo.cpp` in `develop`.

.github/workflows/libxrpl.yml
```diff
@@ -0,0 +1,85 @@
...
+        run: |
+          conan remote list
+          conan remote remove ripple || true
+          # Do not quote the URL. An empty string will be accepted (with a non-fatal warning), but a missing argument will not.
+          conan remote add ripple ${{ env.CONAN_URL }} --insert 0
+      - name: Parse new version
+        id: version
+        shell: bash
+        run: |
+          echo version=$(cat src/ripple/protocol/impl/BuildInfo.cpp \
```
ximinez, line 52: This file is is moved to `src/libxrpl/protocol/BuildInfo.cpp` in `develop`.

ximinez (changes requested):

.github/workflows/libxrpl.yml
godexsoft: I don't know tbh., this is copy-pasted from the `dependencies/action.yml`. I assume it was needed but perhaps it is only needed when used through the action.

godexsoft: I verified - it works fine without. Removed now 👍

.github/workflows/libxrpl.yml
godexsoft: Tried but it fails. I already forgot why i had to do it in the first place: this workflow launches together with all others. Dependencies launch before and test targets so the check-regexp above need to contain `dependencies`, otherwise it will immediately exit stating that `test` was never ran against this ref.

.github/workflows/libxrpl.yml
thejohnfreeman: Don't have to change it right now, but please let's all get in the habit of writing the library name in all lowercase: libxrpl. Unique capitalization is a frequent source of frustration when working with C++ libraries. CMake package names, and the variable and target names derived from them, are case-sensitive. It is easier for everyone if we all agree to use only lowercase. Who can remember "[NuDB](https://github.com/XRPLF/rippled/pull/4615/files)"?

.github/workflows/libxrpl.yml
thejohnfreeman: It was needed on macOS. Maybe still is. Like I said, after this goes in, the workflows are ripe for refactoring.

.github/workflows/libxrpl.yml
ximinez: Oooooh. Interesting. Thanks for the explanation.

godexsoft:
@ximinez should we wrap this one up? Anything in the PR needs changing?

ximinez (approved):
Yep. Looks good.

Issue #4937, opened 2024-02-29
fixReducedOffersV1 still not covering all cases and book blocking (Version: 2.1.0)


shortthefomo:
## Issue Description
Amendment fixReducedOffersV1 https://xrpl.org/resources/known-amendments#fixreducedoffersv1 is still not fixing all case where the order book is blocked 

two cases have now been identified on mainnet.

## Supporting Files
![GHhUFv5XYAAVWtV](https://github.com/XRPLF/rippled/assets/508629/8c1daa43-1b20-4f8e-a498-9e5f0afb27d9)
![GHYCDJ_X0AAEvjm](https://github.com/XRPLF/rippled/assets/508629/2f6df5fa-b0f1-4086-870d-60bfc2f3c8eb)

scottschurr:
@lathanbritz, I'm disappointed but not surprised that I didn't fix all the ways that order books can be blocked.  I have something else I'm working on right now, but this issue is next on my list after that's done.

It would help a lot if you can identify a specific historic transaction that failed because of this blockage.  Of course you can only point at transactions that fail with a `tec` code.  So there may not be any such transactions on the network.  But being able to replay such a transaction might help me see the reason behind the blockage.  If you know of a transaction like that please mention it here.  Thanks.

lathanbritz:
@scottschurr rLtdxMMXPs21PYnnSiwm85EabdHDH4oVhi has blocked the USD book (its the one with the partially filled order), it does not have too  many tx 

<img width="620" alt="Screenshot 2024-03-01 at 17 33 02" src="https://github.com/XRPLF/rippled/assets/508629/7dc0a8f1-dfe8-4964-8cde-74e34f853748">

scottschurr:
@lathanbritz that helps, thanks.  That's current?  Like that's the state of the book for, say, ledger 86325875?  Is there a way you can get the ledger index of the blocking offer (from rLtdxMMXPs21PYnnSiwm85EabdHDH4oVhi).  That would be a big help.  Then I can look back at that offer's history.

lathanbritz:
@scottschurr we have another instance today


ledger index 86,858,489 its present
account https://xrpscan.com/account/rJTyFN9PYy9uWJ3ixsSBfvhZGgvvjNt8MV

scottschurr:
@lathanbritz, thanks for the information on that recent blockage.

[...]

PR #5032, opened 2024-05-31
Amendment fixReducedOffersV2 fixes issue #4937

scottschurr:

In issue https://github.com/XRPLF/rippled/issues/4937 @shortthefomo reported additional instances of order book blocking since the `fixReducedOffersV1` amendment went live.  Thanks to that report I was able to identify an additional case where blocking offers could be generated.  This pull request fixes the newly identified case.

### Context of Change

Having order books occasionally blocked by offers that were modified so that their quality was worse than the originally placed quality has been a problem for a while.  Three different ways that this could occur were identified and addressed by amendment `fixReducedOffersV1`.  All three of those case were addressed by carefully looking at and fixing the numeric rounding applied during calculations.

This new amendment, `fixReducedOffersV2`, identifies and fixes an additional rounding case that was missed during the research for `fixReducedOffersV1 `.

Beyond the fix itself, there are two kinds of unit test additions in this pull request.

1. There's a new test that directly exercises the failing case both with and without the amendment.  It demonstrates that the old rounding behavior could lead to blocked order books.  It also shows that with the new rounding behavior the order books are no longer blocked for this range of cases.

2. A few pre-existing test cases failed due to the changed rounding.  Those test cases were examined and the tests patched to handle the new rounding behavior.

### Type of Change

### API Impact

None.


src/test/app/Flow_test.cpp
```diff
@@ -1375,9 +1386,12 @@ struct Flow_test : public beast::unit_test::suite
     {
         using namespace jtx;
         FeatureBitset const ownerPaysFee{featureOwnerPaysFee};
+        FeatureBitset const reducedOffersV2(fixReducedOffersV2);
 
         testLineQuality(features);
         testFalseDry(features);
+        testDirectStep(features - reducedOffersV2);
```
gregtatcam, line 1393: Why need to run this test with `reducedOffersV2` enabled? There is no change to `testDirectStep` with/without the amendment.

src/test/app/Flow_test.cpp
```diff
@@ -518,14 +517,26 @@ struct Flow_test : public beast::unit_test::suite
...
+                ter(expectedTER));
 
-            env.require(balance(carol, EUR(1)));
-            env.require(balance(bob, USD(0.4)));
-            env.require(balance(bob, EUR(999)));
+            if (!reducedOffersV2)
+            {
+                env.require(balance(carol, EUR(1)));
+                env.require(balance(bob, USD(0.4)));
+                env.require(balance(bob, EUR(999)));
```
gregtatcam, line 538: Does it make sense to extend this test to show that the order book is actually blocked without the amendment?

src/ripple/app/paths/impl/BookStep.cpp
```diff
@@ -661,14 +661,18 @@ limitStepIn(
...
+    Rules const& rules)
 {
     if (limit < stpAmt.in)
     {
         stpAmt.in = limit;
         auto const inLmt =
             mulRatio(stpAmt.in, QUALITY_ONE, transferRateIn, /*roundUp*/ false);
-        ofrAmt = offer.limitIn(ofrAmt, inLmt);
+        // It turns out we can prevent order book blocking by (strictly)
+        // rounding down the ceil_in() result.  This adjustment changes
```
gregtatcam, line 673: Would it be beneficial to add a comment here and in `limitStepOut` explaining why the choice of the rounding in each case prevents order book blocking?

src/test/app/ReducedOffer_test.cpp
```diff
@@ -606,13 +609,193 @@ class ReducedOffer_test : public beast::unit_test::suite
...
+            //
+            // Also verify that if fixReducedOffersV2 is not enabled then
+            // some of the test cases produced a potentially blocking rate.
+            if (features[fixReducedOffersV2])
+            {
+                BEAST_EXPECT(blockedCount == 0);
+            }
+            else
+            {
+                BEAST_EXPECT(blockedCount > 80);
```
gregtatcam, line 786: Shouldn't we check for the exact count of the blocking rates since the same tests are executed on every run? If this count changes in the next release then doesn't it indicate some change that needs to be investigated?

gregtatcam (approved):
👍 Looks good. I left a few minor comments for you to consider.

scottschurr (commented):
Thanks for the review @gregtatcam.  I've addressed a couple of your comments.  There were also two that I didn't do anything with, but I left notes about why I they are the way they are.  I'm open to continuing the conversation on those if you'd like.

src/test/app/Flow_test.cpp
gregtatcam: I think it should be removed if it doesn't accomplish anything. It just increases the tests running time without any benefit.

src/test/app/ReducedOffer_test.cpp
gregtatcam: Yeah, it's fine.

gregtatcam (approved):
👍

src/ripple/app/paths/AMMOffer.h
```diff
@@ -103,15 +104,19 @@ class AMMOffer
     limitOut(
         TAmounts<TIn, TOut> const& offrAmt,
         TOut const& limit,
-        bool fixReducedOffers,
+        Rules const& rules,
```
seelabs, line 107: Consider using the globally available rules rather than passing this as a parameter. (also fine as-is, but I think I'd like to get away from passing rules around as much as we do and just using the new global).

src/ripple/app/paths/AMMOffer.h
```diff
@@ -103,15 +104,19 @@ class AMMOffer
...
     /** Limit in of the provided offer. If one-path then swapIn
      * using current balances. If multi-path then ceil_in using
      * current quality.
      */
     TAmounts<TIn, TOut>
-    limitIn(TAmounts<TIn, TOut> const& offrAmt, TIn const& limit) const;
+    limitIn(
+        TAmounts<TIn, TOut> const& offrAmt,
+        TIn const& limit,
+        Rules const& rules,
```
seelabs, line 118: See comment above about passing rules as a param.

src/ripple/protocol/Quality.h
```diff
@@ -200,6 +200,29 @@ class Quality
             toAmount<In>(stRes.in), toAmount<Out>(stRes.out));
     }
 
+    Amounts
+    ceil_in_strict(Amounts const& amount, STAmount const& limit, bool roundUp)
```
seelabs, line 204: Could use a comment for how "strict" differs from the non-strict function.

src/ripple/protocol/Quality.h
```diff
@@ -200,6 +200,29 @@ class Quality
...
+        const;
+
+    template <class In, class Out>
+    TAmounts<In, Out>
+    ceil_in_strict(
```
seelabs, line 209: We usually don't define functions inline in the class. I know this is a member template and that gets a little tedious outside of the class. I'd prefer it was defined out of the class, but it's a minor nit.

src/ripple/protocol/Quality.h
```diff
@@ -200,6 +200,29 @@ class Quality
...
+        TAmounts<In, Out> const& amount,
+        In const& limit,
+        bool roundUp) const
+    {
```
seelabs, line 213: This looks like a copy of `ceil_in` with one function called changed. 

If `ceil_in` is being deprecated and will be replaced with `ceil_in_strict` then I actually prefer the code duplication. And I think this is the case. If this is not the case, then let's look at merging the two.

On the same topic, if the `ceil_in` is being deprecated, consider renaming `ceil_in` to `ceil_in_deprecated` or somesuch and renaming `ceil_in_strict` to `ceil_in`.

src/ripple/protocol/impl/Quality.cpp
```diff
@@ -64,13 +64,20 @@ Quality::operator--(int)
     return prev;
 }
 
-Amounts
-Quality::ceil_in(Amounts const& amount, STAmount const& limit) const
+template <STAmount (
+    *DivRoundFunc)(STAmount const&, STAmount const&, Issue const&, bool)>
```
seelabs, line 68: I usually like passing the callback as a parameter rather than as a template. But I'm also fine with this as-is.

seelabs (commented):
I think this is ready to go, but I left a couple of nits for you to consider. I'll approve it after you take a look. I'm fine with however you decide to resolve the nits (including keeping them as they are).

src/ripple/protocol/Quality.h
scottschurr: I'm not sure about `ceil_in` being deprecated.  But it certainly has an odd behavior regarding rounding.  So we _might_ eventually want to replace all uses of it with `ceil_in_strict`.  But I'd like to hold off on renaming either of them until we know that we're actually going to make that change (which would have to go under an amendment).  There are a few things in this code base that have been "deprecated" for quite a few years without ever being removed.  I'd like to not add to that confusion.

seelabs (approved):
:+1:

scottschurr:
Proposed commit message:
```
fixReducedOffersV2: prevent offers from blocking order books:

Fixes issue #4937.

The fxiReducedOffersV1 amendment fixed certain forms of offer
modification that could lead to blocked order books.  Reduced
offers can block order books if the effective quality of the
reduced offer is worse than the quality of the original offer
(from the perspective of the taker). It turns out that, for
small values, the quality of the reduced offer can be
significantly affected by the rounding mode used during
scaling computations.

Issue #4937 identified an additional code path that modified
offers in a way that could lead to blocked order books.  This
commit changes the rounding in that newly located code path so
the quality of the modified offer is never worse than the
quality of the offer as it was originally placed.

It is possible that additional ways of producing blocking
offers will come to light.  Therefore there be a future need
for a V3 amendment.
```

PR #5042, opened 2024-06-10
Add Antithesis intrumentation

Bronek:

Add Antithesis instrumentation.

### Context of Change

This change adds the support for [continuous testing (fuzzing)](https://youtu.be/KoxXb2mBVuY) of `rippled` on [Antithesis platform](https://antithesis.com/docs/introduction/how_antithesis_works.html). A copy of [Antithesis C++ SDK](https://github.com/antithesishq/antithesis-sdk-cpp/) is added to `external/antithesis-sdk` (with `clang-format` disabled, to make it easier to keep up).

The instrumentation macros `ASSERT` and `UNREACHABLE` are defined  inside `include/xrpl/beast/utility/instrumentation.h`. In a normal build, these macros are just wrappers (with added name) for regular `assert` macro. In such build, there is no dependency on `external/antithesis-sdk`. The documentation of macros is also in `include/xrpl/beast/utility/instrumentation.h`, and the naming convention for instrumentation macros has been added to `CONTRIBUTING.md`

If `voidstar` CMake option is enabled (which is only supported with clang compiler, version 16 or later, running on Linux, and requires a Debug build), these macros will register and forward instrumentation calls to `external/antithesis-sdk`; this in turn informs the Antithesis platform of all the contracts. In case if a contract is violated during fuzzing, a report will be created for `rippled` developers for troubleshooting. The purpose of name in instrumentation macros is to provide instrumentation calls with stable identity, which does not rely on line numbers.

The `voidstar` CMake option is not exposed in `conanfile.py` because it is not useful for programs with a dependency on this project; it is only useful when building `rippled` binary with the intent of running it on the Antithesis platform. This means that, for ordinary users, macros `ASSERT` or `UNREACHABLE` are just like named `assert`.

Both `ASSERT` or `UNREACHABLE` are so-called "always assertions" - that is, they are [test properties](https://antithesis.com/docs/introduction/how_antithesis_works.html#test-properties), which must be at all times true. Antithesis also supports a ["sometimes assertions"](https://antithesis.com/docs/best_practices/sometimes_assertions.html) which this PR does **not** define; we will add them later, on a case-by-case basis.

Also, this PR does not use ["rich assertions"](https://antithesis.com/blog/zelda/) which are added in [Antithesis SDK 0.4](https://github.com/antithesishq/antithesis-sdk-cpp/releases/tag/v0.4.0); this will be also used in a separate PR, in order to keep this PR as close as possible to regular `assert` model (even though this makes fuzzing slower to discover bugs, than it would have been otherwise).

#### Summary of changes

1. Define and document instrumentation macros in `include/xrpl/beast/utility/instrumentation.h`
2. Add external dependency `antithesis-sdk-cpp` in `external/antithesis-sdk`
3. Add support for `antithesis-sdk-cpp` in `CMakeLists.txt` and `cmake/RippledInstall.cmake`
4. Add new CMake option `voidstar` to `cmake/RippledCompiler.cmake`, `cmake/RippledCore.cmake` and `cmake/RippledSettings.cmake`
5. Set compilation option `--build-id` unconditionally in `cmake/RippledCompiler.cmake`
6. Update `CONTRIBUTING.md` to document contracts and instrumentation
7. Add `UNREACHABLE` in `LogicError` in `src/libxrpl/basics/contract.cpp`
8. All the remaining code changes (_several hundred lines_) are replacing `assert` with either `ASSERT` or `UNREACHABLE` (where suitable), with added name
9. The naming guidance is explained inside `CONTRIBUTING.md`

I strived to make the naming of contracts consistent everywhere, however that was difficult given the size of change.

#### Macro `ASSERT` is `ALWAYS_OR_UNREACHABLE` rather than `ALWAYS`

Instrumentation macros are registered in the Antithesis platform and they can be _also_ used for the purpose of coverage reporting (that is, whether or not a given contract has been reached is also a test property on the Antithesis platform). The semantics of `ALWAYS` macro (as defined in [Antithesis C++ SDK](https://antithesis.com/docs/using_antithesis/sdk/cpp/assert.html#always-assertions)) is not only to check that a condition is always true, it is **also** that the given line of code is _always_ executed during fuzzing. Since we do not expect all `assert`s to be always executed by a program during testing, we need a different macro, with the semantics that the condition is true **if the given line is executed**, otherwise the test platform should ignore the fact that a given line was **not executed**. This is the semantics of `ALWAYS_OR_UNREACHABLE` macro.

#### Macro `ASSERT` is not the same as `assert`

There are also locations where we do not use `ASSERT` or `UNREACHABLE` because the old style `assert` or `assert(false)` make more sense:
  * `constexpr` functions (`ASSERT` would fail compilation)
  * unit tests i.e. files under `src/test` (no need for instrumentation)
  * unit tests-related modules (files under `beast/test` and `beast/unit_test`; no need for instrumentation)

Also, as explained in `include/xrpl/beast/utility/instrumentation.h`, there are minor differences between `ASSERT` and `assert`:
 * `ASSERT` must have an unique name (naming convention in CONTRIBUTING.md)
 * the condition (which comes first) must be *implicitly* convertible to bool
 * during fuzzing, the program will continue execution past a failed `ASSERT` (like Release build) however the parameters will be fully evaluated (which is why CMake option `voidstar` requires Debug build) 

#### No change to the github workflow, yet

We need to upgrade clang compiler from version 14 to version no older than 16 in order to enable `voidstar` option, which would be needed to validate that the instrumentation macros do work with the the headers in `antithesis-sdk-cpp` (in particular that the `ASSERT` conditions are **implicitly** convertible to `bool`). I raised an issue https://github.com/XRPLF/rippled/issues/5153 for this upgrade. Once it is done, I intend to submit a separate PR which will define a new job to build with the `voidstar` option.

### Type of Change

- [ ] Documentation update
- [ ] Release


Builds/CMake/RippledCompiler.cmake
```diff
@@ -120,7 +120,7 @@ else ()
   target_link_libraries (common
     INTERFACE
       -rdynamic
-      $<$<BOOL:${is_linux}>:-Wl,-z,relro,-z,now>
+      $<$<BOOL:${is_linux}>:-Wl,-z,relro,-z,now,--build-id>
```
Bronek, line 123: Th unconditional addition of `--build-id` is intentional, but could be made conditional (on `voidstar`), like other options are. I just think it is a useful option to have even in a regular build.

Bronek:
The choice of cmake option name `voidstar` comes from the reference to `libvoidstar.so` in the implementation of the Antithesis instrumentation - which is inside `external/antithesis-sdk/antithesis_instrumentation.h` file. I am open to replacing it with something else, e.g. `instrumentation`.

Bronek:
TODO: Alter the consensus threshold to 80% (only when `voidstar` enabled),  so we can test the system with 6 validators only.

intelliot:
6 validators is a good start and should yield useful results.

But just curious, how difficult would it be to later scale up the test to 35 (or more) validators?

(This would make the testing even more realistic and predict what could happen in the future if more validators are added)

include/xrpl/basics/instrumentation.h
```diff
@@ -0,0 +1,47 @@
...
+#ifndef RIPPLE_BASICS_INSTRUMENTATION_H_INCLUDED
+#define RIPPLE_BASICS_INSTRUMENTATION_H_INCLUDED
+
+#include <cassert>
+
+#ifdef ENABLE_VOIDSTAR
+#ifndef NDEBUG
+#include <antithesis_sdk.h>
+#else
+#pragma GCC error "Antithesis instrumentation requires Debug build"
```
thejohnfreeman, line 29: Can this just be `#error`? Maybe even just:

```cpp
#ifdef NDEBUG
#error "Antithesis instrumentation requires Debug build"
#endif
#include <antithesis_sdk.h>
```

include/xrpl/basics/instrumentation.h
```diff
@@ -0,0 +1,47 @@
...
+#endif
+#else
+#define ALWAYS(cond, message, ...)
+#define ALWAYS_OR_UNREACHABLE(cond, message, ...) assert(message&& cond)
```
thejohnfreeman, line 33: Wrap `cond` in parentheses?

Builds/levelization/results/loops.txt
```diff
@@ -4,6 +4,9 @@ Loop: test.jtx test.toplevel
 Loop: test.jtx test.unit_test
   test.unit_test == test.jtx
 
+Loop: xrpl.basics xrpl.beast
+  xrpl.basics > xrpl.beast
+
```
thejohnfreeman, line 9: This is very disappointing to me. Wondering what you think of some alternatives.

a. We cannot just directly use `ALWAYS_OR_UNREACHABLE` in `xrpl/beast/` because it is not defined on non-Linux platforms.
b. We could write our own `<cassert>` and put it on the include path before the standard library.
c. We could not rewrite assertions within `xrpl/beast/`.
d. We could move the instrumentation header into `xrpl/beast/`.
e. We could add a new module, at a level beneath the `beast` module, that includes the instrumentation header.

thejohnfreeman (changes requested):
Am I correct in guessing that all of the edits under `include/` and `src/` are just changing assertions to use the new macros?

Bronek:
> 6 validators is a good start and should yield useful results.
> 
> But just curious, how difficult would it be to later scale up the test to 35 (or more) validators?
> 
> (This would make the testing even more realistic and predict what could happen in the future if more validators are added)

It's just a question of CPU utilisation on the Antithesis platform, that is balancing price/speed

include/xrpl/basics/instrumentation.h
Bronek: Yes, thanks will do that.

include/xrpl/basics/instrumentation.h
Bronek: yes, good point.

Builds/levelization/results/loops.txt
Bronek: I like d. Would that be ok for you as well ?

Bronek:
> Am I correct in guessing that all of the edits under `include/` and `src/` are just changing assertions to use the new macros?

Yes.

thejohnfreeman (approved):

Bronek:
Note, recent set of changes are related to the addition of names to instrumentation - which helps to ensure that Antithesis testing reports do not rapidly become obsolete.

Bronek:
> > Am I correct in guessing that all of the edits under `include/` and `src/` are just changing assertions to use the new macros?
> 
> Yes.

@thejohnfreeman not anymore. With the commit  https://github.com/XRPLF/rippled/pull/5042/commits/8d7cd9c558aeea3eabe490342a926985b10c552b all the asserts have now gained names, which I tried to deduce by reading the surrounding code. This was needed to ensure that Antithesis reports do not lose the history of passing/failing asserts, as the location of asserts in code changes due to code churn.

include/xrpl/beast/utility/instrumentation.h
```diff
@@ -0,0 +1,58 @@
...
+#define UNREACHABLE(name, ...) assert((name) && false)
+#endif
+
+#define ASSERT ALWAYS_OR_UNREACHABLE
+
+// How to use the above macros:
+//
+// ALWAYS if it must be reached during fuzzing
+// ASSERT if you expect that it might, or might not, be reached during fuzzing
+// REACHABLE if it must be reached during fuzzing
```
bthomee, line 44: The `ALWAYS` and `REACHABLE` have the same usage description, but must have different definitions. Can you clarify their differences here?

include/xrpl/beast/utility/instrumentation.h
```diff
@@ -0,0 +1,58 @@
...
+#include <antithesis_sdk.h>
+#else
+#define ALWAYS(cond, name, ...) assert((name) && (cond))
+#define ALWAYS_OR_UNREACHABLE(cond, name, ...) assert((name) && (cond))
+#define SOMETIMES(cond, name, ...)
+#define REACHABLE(name, ...)
+#define UNREACHABLE(name, ...) assert((name) && false)
+#endif
+
+#define ASSERT ALWAYS_OR_UNREACHABLE
```
bthomee, line 38: You can avoid guarding the `#include <antithesis_sdk.h>` as it already contains these define statements if their SDK is not enabled.

You could do something like:

```
#ifndef ENABLE_VOIDSTAR
#define NO_ANTITHESIS_SDK
#endif
#include <antithesis_sdk.h>
```

I left off the `NDEBUG` check above, because the cmake config already takes care of when the voidstar option can and cannot be turned on. And the Antithesis header then has the following check to determine whether or not to enable the SDK: `#if defined(NO_ANTITHESIS_SDK) || __cplusplus < 202000L || (defined(__clang__) && __clang_major__ < 16)`.

That being said, the Antithesis header does have a few system library includes plus the namespace definition containing a struct (top of file) and template function (bottom of file). If it is indeed a concern that these would be included in all builds, including releases, then your current code would be fine. That being said, if you're open to modifying the `antithesis_sdk.h` file then you could possibly move that `#if defined(NO_ANTITHESIS_SDK) ...` statement around the namespace or at least the struct/function to get the best of both worlds.

include/xrpl/beast/utility/instrumentation.h
Bronek: the difference is that one takes `cond` parameter and the other does not - I think this is visible enough ?

include/xrpl/beast/utility/instrumentation.h
Bronek: The idea was to entirely exclude the dependency on `external/antithesis-sdk` if the build is not using the `voidstar` option. The reason is that, at least for the time being, we want `rippled` to build successfully on a wider selection of compilers than supported by `external/antithesis-sdk` - which includes clang 15, which would trip on this:

```
#if defined(__clang__) && __clang_major__ < 16
    #error "The Antithesis C++ API requires clang version 16 or higher"
#endif
```

Having said that I like the idea of not having to repeat definitions which are in this section of `#include <antithesis_sdk.h>` ; maybe I can talk Antithesis into doing something to make this header easier to swallow for clang 15 (when `NO_ANTITHESIS_SDK` is defined)

Bronek: No wait, I would still need to guard `#include <antithesis_sdk.h>`, because (if `voidstar` is not being used) I need to define all the polyfills for instrumentation macros. The polyfills from `<antithesis_sdk.h>` are not good, as they do not `assert`.

bthomee: Ah yes, you are correct - they just define the names but don't attach any actions to them.

include/xrpl/beast/utility/instrumentation.h
bthomee: It is visible, but requires the reader to look at both the define statements and the comments to figure that out. I figured the reachable description could be something like `REACHABLE if it must be reached during fuzzing provided the specified condition holds`.

Bronek: It's not "provided" - this would imply that, if the condition does not hold, the location may not be reached. I will figure out better explanation.

src/libxrpl/basics/contract.cpp
```diff
@@ -48,6 +49,7 @@ LogicError(std::string const& s) noexcept
 {
     JLOG(debugLog().fatal()) << s;
     std::cerr << "Logic error: " << s << std::endl;
+    UNREACHABLE("LogicError", {{"message", s}});
```
bthomee, line 52:
```suggestion
    UNREACHABLE("ripple::detail::LogicError", {{"message", s}});
```

src/libxrpl/json/Object.cpp
```diff
@@ -168,7 +168,7 @@ Array::append(Json::Value const& v)
             return;
         }
     }
-    assert(false);  // Can't get here.
+    UNREACHABLE("Json::Array::append : invalid type");  // Can't get here.
```
bthomee, line 171:
```suggestion
    UNREACHABLE("Json::Array::append : invalid type"); 
```

The `UNREACHABLE` statement now makes it clear no path can get here.

src/libxrpl/json/Object.cpp
```diff
@@ -203,7 +203,7 @@ Object::set(std::string const& k, Json::Value const& v)
             return;
         }
     }
-    assert(false);  // Can't get here.
+    UNREACHABLE("Json::Object::set : invalid type");  // Can't get here.
```
bthomee, line 206:
```suggestion
    UNREACHABLE("Json::Object::set : invalid type");
```

src/libxrpl/json/json_reader.cpp
```diff
@@ -953,7 +953,7 @@ operator>>(std::istream& sin, Value& root)
     Json::Reader reader;
     bool ok = reader.parse(sin, root);
 
-    // JSON_ASSERT( ok );
+    // ASSERT( ok );
```
bthomee, line 956:
```suggestion
```

Commented-out code can be removed. Moreover, if you want to keep it then the `ASSERT` macro must be fully implemented, including giving it a name.

src/test/csf/Histogram.h
```diff
@@ -19,8 +19,8 @@
 #ifndef RIPPLE_TEST_CSF_HISTOGRAM_H_INCLUDED
 #define RIPPLE_TEST_CSF_HISTOGRAM_H_INCLUDED
 
+#include <xrpl/beast/utility/instrumentation.h>
```
bthomee, line 22:
```suggestion
```

Include does not seem to be used, so can be removed.

I see this occurring in several files. I gave a few additional suggestions for removal below, but please check all files to see if the include isn't unnecessarily present if none of the macros are called

src/test/jtx/impl/amount.cpp
```diff
@@ -20,7 +20,7 @@
 #include <test/jtx/Account.h>
 #include <test/jtx/amount.h>
 #include <xrpl/basics/safe_cast.h>
-#include <cassert>
+#include <xrpl/beast/utility/instrumentation.h>
```
bthomee, line 23:
```suggestion
```

src/test/overlay/short_read_test.cpp
```diff
@@ -21,12 +21,12 @@
 #include <xrpl/basics/make_SSLContext.h>
 #include <xrpl/beast/core/CurrentThreadName.h>
 #include <xrpl/beast/unit_test.h>
+#include <xrpl/beast/utility/instrumentation.h>
```
bthomee, line 24:
```suggestion
```

src/test/unit_test/multi_runner.cpp
```diff
@@ -20,6 +20,7 @@
 #include <test/unit_test/multi_runner.h>
 
 #include <xrpl/beast/unit_test/amount.h>
+#include <xrpl/beast/utility/instrumentation.h>
```
bthomee, line 23:
```suggestion
```

src/xrpld/app/consensus/RCLConsensus.cpp
```diff
@@ -92,7 +92,9 @@ RCLConsensus::Adaptor::Adaptor(
               std::numeric_limits<std::uint64_t>::max() - 1))
     , nUnlVote_(validatorKeys_.nodeID, j_)
 {
-    assert(valCookie_ != 0);
+    ASSERT(
```
bthomee, line 95: Shouldn't this file have a `#include <xrpl/beast/utility/instrumentation.h>` statement? I presume it compiles due to transitive includes, but it would be better for the file to fully mention all dependencies as imports.

I see this across many other files too, so please double-check all.

src/xrpld/app/tx/detail/InvariantCheck.cpp
```diff
@@ -380,7 +380,8 @@ AccountRootsDeletedClean::finalize(
     // transaction processing results, however unlikely, only fail if the
     // feature is enabled. Enabled, or not, though, a fatal-level message will
     // be logged
-    bool const enforce = view.rules().enabled(featureInvariantsV1_1);
+    [[maybe_unused]] bool const enforce =
```
bthomee, line 383: Would this be a good candidate for `SOMETIMES`? If the fuzzer never can get the enforce variable to be true, perhaps it indeed is unused.

src/xrpld/app/tx/detail/OfferStream.cpp
```diff
@@ -339,7 +340,9 @@ TOfferStreamBase<TIn, TOut>::step()
                                 std::is_same_v<TOut, XRPAmount>))
                     return shouldRmSmallIncreasedQOffer<IOUAmount, IOUAmount>();
             }
-            assert(0);  // xrp/xrp offer!?! should never happen
+            UNREACHABLE(
+                "rippls::TOfferStreamBase::step::rmSmallIncreasedQOffer : XRP "
+                "vs XRP offer");  // xrp/xrp offer!?! should never happen
```
bthomee, line 345:
```suggestion
                "vs XRP offer");
```

The extra comment is now superfluous.

src/xrpld/app/tx/detail/SetSignerList.cpp
```diff
@@ -128,7 +128,10 @@ SetSignerList::doApply()
         default:
             break;
     }
-    assert(false);  // Should not be possible to get here.
+    UNREACHABLE(
+        "ripple::SetSignerList::doApply : invalid operation");  // Should not be
+                                                                // possible to
+                                                                // get here.
```
bthomee, line 134:
```suggestion
        "ripple::SetSignerList::doApply : invalid operation");
```

src/xrpld/core/JobTypes.h
```diff
@@ -53,17 +53,22 @@ class JobTypes
...
+            [[maybe_unused]] auto const inserted =
+                m_map
+                    .emplace(
+                        std::piecewise_construct,
+                        std::forward_as_tuple(jt),
+                        std::forward_as_tuple(
+                            jt, name, limit, avgLatency, peakLatency))
+                    .second;
+
+            ASSERT(
```
bthomee, line 69: As before, would this be a candidate for using `SOMETIMES`, if you don't know if it's actually used?

external/antithesis-sdk/antithesis_instrumentation.h
```diff
@@ -0,0 +1,113 @@
...
+        trace_pc_guard_init(start, stop);
+    }
+    return;
+}
+
+extern 
+#ifdef __cplusplus
+    "C"
+#endif
+void __sanitizer_cov_trace_pc_guard( uint32_t *guard ) {
```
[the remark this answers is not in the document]
Bronek: Nope, I'd rather take `external/antithesis-sdk`  exactly as-is, so we do not have think hard when upgrading to a newer version

external/antithesis-sdk/antithesis_instrumentation.h
```diff
@@ -0,0 +1,113 @@
...
+    if (has_libvoidstar) {
+        uint64_t edge = (uint64_t)(__builtin_return_address(0));
+        trace_pc_guard(guard, edge);
+    } else {
+        if (guard) {
+          *guard = 0;
+        }
+    }
```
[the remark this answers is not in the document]
Bronek: I'd rather use `external/antithesis-sdk` exactly as provided

src/xrpld/core/JobTypes.h
Bronek: No, `SOMETIMES` is only a hint for the fuzzer, it does not check anything.

src/xrpld/app/tx/detail/InvariantCheck.cpp
Bronek: The `[[maybe_unused]]` attribute means that the compiler should not warn when the variable is never used. This variable `enforce` is only ever used inside an `ASSERT` (down below in this function), meaning the compiler will see it being used in any Debug build (including `voidstar` one) but will _not_ see it being used in a Release build. That's how we use `[[maybe_unused]]` in several other location in this project. On the other hand `SOMETIMES` disables check altogether, and that's not what we want - since before this PR we had actual `assert` in this location.

src/xrpld/app/consensus/RCLConsensus.cpp
Bronek: great point.

src/xrpld/app/paths/detail/StrandFlow.h
```diff
@@ -839,7 +847,7 @@ flow(
             // running debug builds of rippled. While this issue still needs to
             // be resolved, the assert is causing more harm than good at this
             // point.
-            // assert(0);
+            // UNREACHABLE();
```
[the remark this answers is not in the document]
Bronek: there was commented-out `assert(0)` before; replacing it with an actual `UNREACHABLE()` would change the program behaviour from "we put a comment here" to "we crash", and that's not a good PR for this kind of change.

CONTRIBUTING.md
```diff
@@ -343,6 +343,62 @@ pip3 install pre-commit
...
+To learn more about Antithesis, see
+[How Antithesis Works](https://antithesis.com/docs/introduction/how_antithesis_works.html)
+and [C++ SDK](https://antithesis.com/docs/using_antithesis/sdk/cpp/overview.html#)
+
+We continue to use the old style `assert` or `assert(false)` in certain
+locations, where the reporting of contract violations on the Antithesis
+platform is either not possible or not useful.
+
+For this reason:
+* The locations where `assert` or `assert(false)` contracts can be used:
```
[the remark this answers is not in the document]
Bronek: We do not seem to care that much about 80 chars per line in `.md` files, so :man_shrugging:

external/antithesis-sdk/antithesis_instrumentation.h
```diff
@@ -0,0 +1,113 @@
...
+    "C"
+#endif
+void __sanitizer_cov_trace_pc_guard_init(uint32_t *start, uint32_t *stop) {
+    debug_message_out("SDK forwarding to libvoidstar for __sanitizer_cov_trace_pc_guard_init()");
+    if (!did_check_libvoidstar) {
+        antithesis_load_libvoidstar();
+    }
+    if (has_libvoidstar) {
+        trace_pc_guard_init(start, stop);
+    }
```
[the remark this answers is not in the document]
Bronek: I leave that to the authors of Antithesis SDK ; they decided that `antithesis_load_libvoidstar` should have no parameters and that seems OK to me. If it were to call `trace_pc_guard_init(start, stop);` then it would have to also receive `start` and `stop` parameters.

src/libxrpl/basics/contract.cpp
Bronek: I'd rather keep this one; it is special because there are different ways `LogicError` may happen, which is also why the extra parameter.

src/xrpld/app/paths/detail/StrandFlow.h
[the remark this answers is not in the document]
bthomee: Let me rephrase: It was a commented-out `assert`, and you've replaced it by a commented-out `UNREACHABLE`. My assumption here is that you want to leave this here commented out, so it can be re-enabled in the future when needed. If so, the old `assert(0)` statement would compile fine but the new `UNREACHABLE()` statement would fail to compile as that macro expects at least a name to be provided.

bthomee (approved):

bthomee (approved):

Bronek:
Blocked on #5143 and #5103

intelliot (approved):
note: this will be merged _after_ rippled 2.3.0

ximinez:
@Bronek Before I merge this, can you write a commit message with a consise summary?

Bronek:
> @Bronek Before I merge this, can you write a commit message with a consise summary?

Add Antithesis instrumentation

* Copy Antithesis SDK version 0.4.0 to directory external/
* Add build option `voidstar` to enable instrumentation with Antithesis SDK
* Define instrumentation macros ASSERT and UNREACHABLE in terms of regular C assert
* Replace all asserts with named ASSERT or UNREACHABLE
* Add UNREACHABLE to LogicError
* Document instrumentation macros in CONTRIBUTING.md

ximinez:
Merged manually as d7e949193fe3a42c67f36c021daba9fd0d52780c.
```
$ git diff --exit-code Bronek/feature/antithesis-instrumentation && echo Identical
Identical
```

PR #5046, opened 2024-06-13
fix account_objects with invalid marker do not return an error

### LINKED ISSUE #4542: [account_objects with invalid marker do not return an error] (Version: [1.11.0-rc2])
## Steps to Reproduce

For account_objects API, use a marker generated from another API.
Sample request:
`{
    "method": "account_objects",
    "params": [
        {
            "account": "r3kmBwpGRwRoZGM64zkkwoDFB7GNJqxD8Y",
            "marker": "044C3E9F80DDA8381673FB0B149C642045952FC2DF82243AF80F5F2997988B15,0"
        }
    ]
}`
## Expected Result

`invalidParams`
## Actual Result

No error

yinyiqian1:
fix #4542 

### Type of Change

- [ ] Documentation update
- [ ] Release

### API Impact


src/xrpld/rpc/detail/RPCHelpers.cpp
```diff
@@ -250,7 +250,10 @@ getAccountObjects(
     {
         // it's possible the user had nftoken pages but no
         // directory entries
-        return mlimit < limit;
+        if (mlimit >= limit)
+            jvResult[jss::account_objects] = Json::arrayValue;
+
```
gregtatcam, line 255: Please add comments why `true` is always returned in this case.

yinyiqian1: updated. also added for  `if (mlimit >= limit)`. please check

gregtatcam (approved):
👍 LGTM

src/test/rpc/AccountObjects_test.cpp
```diff
@@ -1220,6 +1220,104 @@ class AccountObjects_test : public beast::unit_test::suite
...
+        {
+            Json::Value params;
+            params[jss::account] = bob.human();
+            params[jss::limit] = limit;
+            params[jss::marker] = marker;
+            params[jss::ledger_index] = "validated";
+            auto resp = env.rpc("json", "account_objects", to_string(params));
+            auto& accountObjects = resp[jss::result][jss::account_objects];
+            BEAST_EXPECT(!resp[jss::result].isMember(jss::error));
+            BEAST_EXPECT(
```
Bronek, line 1280: please also add here check that this is the last page:
```c++
BEAST_EXPECT(!resp[jss::result].isMember(jss::marker));
```

src/xrpld/rpc/handlers/AccountObjects.cpp
```diff
@@ -283,6 +283,10 @@ doAccountObjects(RPC::JsonContext& context)
             return RPC::invalid_field_error(jss::marker);
     }
 
+    // check if dirIndex is valid
+    if (!dirIndex.isZero() && !ledger->read({ltDIR_NODE, dirIndex}))
+        return RPC::invalid_field_error(jss::marker);
```
Bronek, line 288: since `RPC::getAccountObjects` returning `false` would have the same result, I would move this check into the body of `RPC::getAccountObjects`; this will keep the checks (other than the `marker` format, in the code segment above) in that one function only, which is more robust.

src/xrpld/rpc/detail/RPCHelpers.cpp
```diff
@@ -249,8 +249,18 @@ getAccountObjects(
...
         // it's possible the user had nftoken pages but no
-        // directory entries
-        return mlimit < limit;
+        // directory entries. If there's no nftoken page, we will
+        // give empty array for account_objects.
+        if (mlimit >= limit)
+            jvResult[jss::account_objects] = Json::arrayValue;
+
+        // non-zero dirIndex validity was checked in the caller function;
+        // by this point, it should be zero.
```
Bronek, line 258: as commented elsewhere, I would prefer if this function did not rely on the validation of `dirIndex` happening in the caller, but rather move that check into the body of `getAccountObjects`

Bronek (commented):
Thanks for fixing this bug. I have some comments which would hopefully make this fix a little better.

Bronek:
While you are at it, I see that `account_objects` has pretty good unit test coverage but it could be still improved. See the red line 274 in https://app.codecov.io/gh/XRPLF/rippled/blob/develop/src%2Fxrpld%2Frpc%2Fhandlers%2FAccountObjects.cpp and red line 183 in https://app.codecov.io/gh/XRPLF/rippled/blob/develop/src%2Fxrpld%2Frpc%2Fdetail%2FRPCHelpers.cpp ; perhaps you could add tests for those two lines as well ?

yinyiqian1:
> While you are at it, I see that `account_objects` has pretty good unit test coverage but it could be still improved. See the red line 274 in https://app.codecov.io/gh/XRPLF/rippled/blob/develop/src%2Fxrpld%2Frpc%2Fhandlers%2FAccountObjects.cpp and red line 183 in https://app.codecov.io/gh/XRPLF/rippled/blob/develop/src%2Fxrpld%2Frpc%2Fdetail%2FRPCHelpers.cpp ; perhaps you could add tests for those two lines as well ?

Sure. I just added. But the following line is unreachable because if string s does not have ','   , std::getline will just read to the string end. It wont return error. 
```
if (!std::getline(ss, s, ','))
            return RPC::invalid_field_error(jss::marker);
```

So I also changed it to 
```
if (s.find(',') == std::string::npos)
            return RPC::invalid_field_error(jss::marker);
```

yinyiqian1:
> While you are at it, I see that `account_objects` has pretty good unit test coverage but it could be still improved. See the red line 274 in https://app.codecov.io/gh/XRPLF/rippled/blob/develop/src%2Fxrpld%2Frpc%2Fhandlers%2FAccountObjects.cpp and red line 183 in https://app.codecov.io/gh/XRPLF/rippled/blob/develop/src%2Fxrpld%2Frpc%2Fdetail%2FRPCHelpers.cpp ; perhaps you could add tests for those two lines as well ?

Yes. I just added. 

There might be a potential bug here, if we pass in some marker like `xxxxxx,yyyyyy,,,,,,,,1234`, suppose dirIndex and entryIndex part are valid, it will just ignore all the characters after the second `,`. Do you think we should fix this as well?
Also just double check, the marker will always be two parts seperated by `,`, right?
```
std::stringstream ss(marker.asString());
        std::string s;
        if (!std::getline(ss, s, ','))
            return RPC::invalid_field_error(jss::marker);

        if (!dirIndex.parseHex(s))
            return RPC::invalid_field_error(jss::marker);

        if (!std::getline(ss, s, ','))
            return RPC::invalid_field_error(jss::marker);

        if (!entryIndex.parseHex(s))
            return RPC::invalid_field_error(jss::marker);
```

src/test/rpc/AccountObjects_test.cpp
yinyiqian1: added

src/xrpld/rpc/handlers/AccountObjects.cpp
yinyiqian1: updated

src/xrpld/rpc/detail/RPCHelpers.cpp
yinyiqian1: updated

Bronek:
> > While you are at it, I see that `account_objects` has pretty good unit test coverage but it could be still improved. See the red line 274 in https://app.codecov.io/gh/XRPLF/rippled/blob/develop/src%2Fxrpld%2Frpc%2Fhandlers%2FAccountObjects.cpp and red line 183 in https://app.codecov.io/gh/XRPLF/rippled/blob/develop/src%2Fxrpld%2Frpc%2Fdetail%2FRPCHelpers.cpp ; perhaps you could add tests for those two lines as well ?
> 
> Yes. I just added.
> 
> There might be a potential bug here, if we pass in some marker like `xxxxxx,yyyyyy,,,,,,,,1234`, suppose dirIndex and entryIndex part are valid, it will just ignore all the characters after the second `,`. Do you think we should fix this as well? Also just double check, the marker will always be two parts seperated by `,`, right?
> 
> ```
> std::stringstream ss(marker.asString());
>         std::string s;
>         if (!std::getline(ss, s, ','))
>             return RPC::invalid_field_error(jss::marker);
> 
>         if (!dirIndex.parseHex(s))
>             return RPC::invalid_field_error(jss::marker);
> 
>         if (!std::getline(ss, s, ','))
>             return RPC::invalid_field_error(jss::marker);
> 
>         if (!entryIndex.parseHex(s))
>             return RPC::invalid_field_error(jss::marker);
> ```

It would be great to fix it, yes - @mDuo13 any comments ?

Bronek (approved):
nice !

yinyiqian1:
> It would be great to fix it, yes - @mDuo13 any comments ?

@Bronek  I also fix it in the recent commit. Could you review as well?

src/xrpld/rpc/handlers/AccountObjects.cpp
```diff
@@ -268,18 +268,15 @@ doAccountObjects(RPC::JsonContext& context)
...
+        if (!dirIndex.parseHex(markerStr.substr(0, idx)))
             return RPC::invalid_field_error(jss::marker);
 
-        if (!std::getline(ss, s, ','))
-            return RPC::invalid_field_error(jss::marker);
-
-        if (!entryIndex.parseHex(s))
+        if (!entryIndex.parseHex(markerStr.substr(idx + 1)))
             return RPC::invalid_field_error(jss::marker);
     }
```
Bronek, line 281: really nice improvement in parsing here ! :+1:

src/test/rpc/AccountObjects_test.cpp
```diff
@@ -1220,6 +1220,142 @@ class AccountObjects_test : public beast::unit_test::suite
...
+        }
+
+        // test account_objects with an invalid marker containing invalid
+        // entryIndex
+        {
+            // construct an invalid entry
+            std::stringstream ss(marker.asString());
+            std::string s;
+            std::getline(ss, s, ',');
+            s = s + ",0";
```
Bronek, line 1328: can you pls also test when this gets corrupted not by `,0` but by a `,` alone ? i.e. valid `dirIndex` followed by a coma, then end of string. Also reverse, i.e. string starts with a coma, followed by valid `entryIndex`, then end of string. Reading up the implementation these both should return the same kind of error, but you need to have unit tests so that the code (which I think is good at this moment) does not degrade due to bad changes in the future.

src/test/rpc/AccountObjects_test.cpp
```diff
@@ -1220,6 +1220,142 @@ class AccountObjects_test : public beast::unit_test::suite
...
+            auto& accountObjects = resp[jss::result][jss::account_objects];
+            marker = resp[jss::result][jss::marker];
+            BEAST_EXPECT(!resp[jss::result].isMember(jss::error));
+            BEAST_EXPECT(accountObjects.size() == limit);
+        }
+
+        // test invalid marker by adding invalid string after the maker.
+        {
+            std::string s = marker.asString();
+            s.append(",1234");
```
Bronek, line 1264: I suggest replacing this test with a different one, where `entryIndex` is valid, see my other comment.

Bronek (approved):
Good change, just some comments to improve unit tests.

src/test/rpc/AccountObjects_test.cpp
yinyiqian1: This is used to test adding some extra characters after the entryIndex. The marker format is like `xxxx,yyyy,1234`.

Bronek: Nice, I missed it.

Bronek:
@yinyiqian1 I allowed myself to merge the current `develop` to your branch, since it was presenting a confusing diff in `API-CHANGELOG.md` . Hope you do not mind.

src/test/rpc/AccountObjects_test.cpp
```diff
@@ -1220,6 +1220,164 @@ class AccountObjects_test : public beast::unit_test::suite
...
+        unsigned const accountObjectSize = 30;
+        for (unsigned i = 0; i < accountObjectSize; i++)
+            env(check::create(alice, bob, XRP(10)));
+
+        for (unsigned i = 0; i < 10; i++)
+            env(token::mint(carol, 0));
+
+        env.close();
+
+        unsigned const limit = 11;
```
Bronek, line 1245: if you make it `static constexpr` ...

src/test/rpc/AccountObjects_test.cpp
```diff
@@ -1220,6 +1220,164 @@ class AccountObjects_test : public beast::unit_test::suite
...
+            marker = resp[jss::result][jss::marker];
+            BEAST_EXPECT(!resp[jss::result].isMember(jss::error));
+            BEAST_EXPECT(accountObjects.size() == limit);
+        }
+
+        // this lambda function is used to check invalid marker response.
+        auto testInvalidMarker = [&env, &bob](std::string& marker) {
+            Json::Value params;
+            params[jss::account] = bob.human();
+            params[jss::limit] = 11;
```
Bronek, line 1279: ... then you will be able to use `limit` here, without explicit capture.

src/test/rpc/AccountObjects_test.cpp
```diff
@@ -1220,6 +1220,164 @@ class AccountObjects_test : public beast::unit_test::suite
...
+            Json::Value const resp =
+                env.rpc("json", "account_nfts", to_string(params));
+            return resp[jss::result][jss::error_message] ==
+                "Invalid field \'marker\'.";
+        };
+
+        auto const markerStr = marker.asString();
+        auto const& idx = markerStr.find(',');
+        auto const dirIndex = markerStr.substr(0, idx);
+        auto const entryIndex = markerStr.substr(idx + 1);
```
Bronek, line 1291: nice !

src/test/rpc/AccountObjects_test.cpp
```diff
@@ -1220,6 +1220,164 @@ class AccountObjects_test : public beast::unit_test::suite
...
+        // entryIndex with marker: "dirIndex,"
+        {
+            std::string s = dirIndex + ',';
+            BEAST_EXPECT(testInvalidMarker(s));
+        }
+
+        // test account_objects with an invalid marker containing invalid
+        // entryIndex with marker: "dirIndex,0"
+        {
+            std::string s = dirIndex + ',';
```
Bronek, line 1338: you missed trailing `0` here

src/test/rpc/AccountObjects_test.cpp
```diff
@@ -1220,6 +1220,164 @@ class AccountObjects_test : public beast::unit_test::suite
...
+        }
+
+        // test account_objects with an invalid marker containing invalid
+        // entryIndex by replacing some characters from the entryIndex.
+        {
+            std::string s = entryIndex;
+            s.replace(0, 7, "FFFFFFF");
+            s = dirIndex + ',' + s;
+            BEAST_EXPECT(testInvalidMarker(s));
+        }
```
Bronek, line 1349: nice test, but can you please move it to follow immediately after the one where you are replacing part of `dirIndex` with `FFFFFFF` ? This way it will be easier to understand how these two are different.

Bronek (approved):
Good change, just minor remarks on unit tests.

src/test/rpc/AccountObjects_test.cpp
Bronek: never mind, you can capture `limit` also without this change. Although `constexpr` would make it more obvious.

yinyiqian1: can I just update the lambda function capture to `[&]`, you can review in my recent commit.

src/test/rpc/AccountObjects_test.cpp
yinyiqian1: Updated in the recent commit. Please review

src/test/rpc/AccountObjects_test.cpp
yinyiqian1: Updated in the recent commit. Please review

Bronek (approved):

Bronek:
nice ! @yinyiqian1 do you agree this is ready to merge ?

yinyiqian1:
> nice ! @yinyiqian1 do you agree this is ready to merge ?

Yes. Please merge.

yinyiqian1:
> nice ! @yinyiqian1 do you agree this is ready to merge ?

Please do not merge as I found some issue. will let you know once fixed.

Bronek:
@mDuo13 I think this PR does not require change in `API-CHANGELOG.md` because the erroneous behaviour being fixed here is not documented nor makes practical sense to use. You opinion ?

mDuo13:
Yes, this sounds like a bug fix that does not require an API version change.

PR #5047, opened 2024-06-14
fixInnerObjTemplate2 amendment

scottschurr:

Apply inner object templates to all remaining (non-AMM) inner objects.

### Context of Change

It was discovered that the inner objects added to the code base for the AMM were not properly protected by their templates.  That oversight was corrected by the `fixInnerObjTemplate` amendment.

An audit showed that there were additional inner objects that did not have their templates applied.  This pull request applies templates to those additional inner objects.

Adds a unit test for applying the template to sfMajorities. Other remaining inner objects showed no problems having templates applied.

### Type of Change

- [x] Tests (unit tests were added for the templates added to Amendments)
- [x] Amendment.


src/ripple/protocol/impl/STObject.cpp
```diff
@@ -61,10 +61,19 @@ STObject::STObject(SerialIter& sit, SField const& name, int depth) noexcept(
...
 STObject
-STObject::makeInnerObject(SField const& name, Rules const& rules)
+STObject::makeInnerObject(SField const& name)
 {
     STObject obj{name};
-    if (rules.enabled(fixInnerObjTemplate))
+
+    // The if is complicated because inner object templates were added in
+    // two phases:
+    //  1. If there are no available Rules, then always apply the template.
```
gregtatcam, line 70: Why should the template be applied if there is no `Rules`?

scottschurr: @gregtatcam, great question.

The way `getCurrentTransactionRules()` works is:

1. If we're currently inside a transaction then the `Rules` are set and available.
2. If this code is ever called when we're not inside a transaction, then the optional is unseated.

This means that if the optional is unseated then we're not inside a transaction, and any changes we make will not be transaction breaking.  We have two choices we could make at this point:

1. If we're not inside a transaction, stick with the old behavior.
2. If we're not inside a transaction apply the new behavior.

In my opinion the new behavior (apply the template) is strictly safer than the old behavior (don't apply the template).  So I'm choosing to bias the change to honor safety.

Thoughts?

gregtatcam: Thanks for the clarification. Applying the new behavior sounds right to me. One argument to use the new behavior is when a unit-test doesn't call a transaction but does call a function with this amendment, for instance some utility. We want to make sure that the unit-test fails in this case if there is a difference in before/after.

gregtatcam (approved):
👍 LGTM

src/ripple/protocol/impl/STObject.cpp
```diff
@@ -61,10 +61,19 @@ STObject::STObject(SerialIter& sit, SField const& name, int depth) noexcept(
...
+    //  2. fixInnerObjTemplate added templates to two AMM inner objects.
+    //  3. fixInnerObjTemplate2 added templates to all remaining inner objects.
+    std::optional<Rules> const& rules = getCurrentTransactionRules();
+    bool const isAMMObj = name == sfAuctionSlot || name == sfVoteEntry;
+    if (!rules || (rules->enabled(fixInnerObjTemplate) && isAMMObj) ||
+        (rules->enabled(fixInnerObjTemplate2) && !isAMMObj))
```
seelabs, line 76: This code was my only concern. Usually, when we introduce an amendment I want code where it's trivially clear that the code with the amendment switched off is not transaction breaking. That's not true here. I understand why it's not true, but that means we need to audit the old code to make sure all the old existing calls to `makeInnerObject` are for auction slots or vote entries. 

I took a look at the old code and that does appear to be true. No action needed on this comment, I just wanted to add a note that I did that audit in case anyone wants to dummy check me.

seelabs (approved):
:+1:

src/ripple/protocol/impl/STObject.cpp
gregtatcam: This block of code also got my attention and I did review the old existing calls to `makeInnerObject` to make sure it looked right.

scottschurr:
Thanks for the reviews!  This PR needs to go through the physical restructure re-mapping process before it can be merged.  So I'll do the remapping soon.  Once that is done I'll mark the PR as passed.

scottschurr:
The physical restructure is done.  Unit tests, start, sync, and graceful stop all work on my Mac for both debug and release builds.  The CI problem with coverage looks like a host thing.  I'm marking this pull request passed.

PR #5048, opened 2024-06-18
XLS-46: DynamicNFT

tequdev:
Replace from https://github.com/XRPLF/rippled/pull/4838
Co-Author: @xVet

Spec: [XLS-46d: Dynamic Non Fungible Tokens (dNFTs)](https://github.com/XRPLF/XRPL-Standards/discussions/130)

This Amendment adds functionality to update the URI of [NFToken objects](https://github.com/XRPLF/XRPL-Standards/discussions/46) by adding:

`NFTokenModify:` Transactor to initiate the altering of the URI
`lsfMutable`: Flag to be set in a NFTokenMint transaction to allow NFTokenModify to execute successfully on given NFT.

### Context of Change

https://github.com/XRPLF/XRPL-Standards/discussions/130

### Type of Change

- [x ] Tests (you added tests for code that already exists, or your new feature included in this PR)

### API Impact


shawnxie999:
In other PR, I suggested adding a [check](https://github.com/XRPLF/rippled/pull/4838#issuecomment-1838855814) so that NFTokenModify would fail if there is any outstanding offers. This is to prevent the issuer from changing the URI and then accept an existing offer.

tequdev:
> In other PR, I suggested adding a [check](https://github.com/XRPLF/rippled/pull/4838#issuecomment-1838855814) so that NFTokenModify would fail if there is any outstanding offers. This is to prevent the issuer from changing the URI and then accept an existing offer.

As mentioned in this comment, wouldn't that make sense since the issuer can change the URI after the user has bought it?
Users need to be aware of the risk to mutable NFTs, just as they need to be aware of the possibility of being burned by the issuer after purchasing a Burnable NFT.

https://github.com/XRPLF/rippled/pull/4838#issuecomment-1994709096

xVet:
Correct, if you own a dynamic NFT then you agree that it can be changed by the issuer. I wouldn't personally put a check on offers, because this might cut into some use cases people come up with.

shawnxie999:
BTW, the "Context of change" section linked to the XLS20 proposal, I think you meant to link to this one https://github.com/XRPLF/XRPL-Standards/discussions/130

tequdev:
> BTW, the "Context of change" section linked to the XLS20 proposal, I think you meant to link to this one [XRPLF/XRPL-Standards#130](https://github.com/XRPLF/XRPL-Standards/discussions/130)

good catch, fixed!

scottschurr (changes requested):
Nice job on this pull request!  The unit test coverage is pretty good.  I appreciate that you took the time to do that.

I left quite a number of comments, but don't be dismayed by that ...

1. Most of the comments are personal preferences of mine (although I justify them).
2. I can be a very picky reviewer.

That said, I think that my comment regarding the `nft::updateToken()` implementation is pretty important.  If you choose not to implement that change I'll want to understand why.

For your convenience I made commits that implement all of the changes I've suggested.  If you want, you're welcome to cherry pick them, or you can just use them as examples of the changes I am suggesting.  The commits are the top-most three on branch https://github.com/scottschurr/rippled/commits/tequ-featureDynamicNFT/

- Suggested approach to NFTokenMintMask: 8aec32c0b8d814ce4504c6353816aab8764a8003
- Suggested change for `testNFTokenModify()`: 813c302ff8348c0f60bcb730806cc4a6fc1a6bb0
- Suggested changes to NFTokenModify.cpp and NFTokenUtils.*: 1393268ecd41902b9fd321971ffe5994f589e757

I hope you find those useful.

Thanks for the submission!

src/libxrpl/protocol/TxFormats.cpp
```diff
@@ -374,6 +374,15 @@ TxFormats::TxFormats()
         },
         commonFields);
 
+    add(jss::NFTokenModify,
+        ttNFTOKEN_MODIFY,
+        {
+            {sfNFTokenID, soeREQUIRED},
+            {sfOwner, soeOPTIONAL},
+            {sfURI, soeOPTIONAL},
```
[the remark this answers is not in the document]
tequdev: Since you pointed it out, as of Jun 29 the `sfURI` field has been updated to an optional field.

tequdev:
@scottschurr 
Thanks for the review. I fully agree with your suggestions and have cherry-picked all three commits.

scottschurr (approved):
Looks great to me!  Thanks for your patience with my review.

src/xrpld/app/tx/detail/NFTokenUtils.cpp
```diff
@@ -238,6 +238,39 @@ compareTokens(uint256 const& a, uint256 const& b)
     return a < b;
 }
 
+TER
+changeTokenURI(
+    ApplyView& view,
+    AccountID const& owner,
+    uint256 const& nftokenID,
+    std::optional<ripple::Slice> const& uri)
```
shawnxie999, line 246: @scottschurr may I ask why is `Slice` type used here, and not `Buffer`?

Also another related question, in `STBlob`, it uses`Buffer` to store `value_`, but why `value_type` have type `Slice` and not `Buffer`? In `STBlob`:
```
class STBlob : public STBase, public CountedObject<STBlob>
{
    Buffer value_;

public:
    using value_type = Slice;
```

Thanks!

src/libxrpl/protocol/TxFormats.cpp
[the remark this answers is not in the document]
shawnxie999: Is there any use case for removing the URI from an existing NFT? Is there any difference from burning the NFT after it becomes an NFT without an URI?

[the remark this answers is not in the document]
tequdev: Such a use case may not be common, but it is not impossible to say that there is no use case.

Since the NFTokenMint transaction can be executed with the URI field unspecified, it is reasonable that the NFTokenModify transaction can change the URI field to unspecified.

shawnxie999 (approved):
Looks good!

src/libxrpl/protocol/TxFormats.cpp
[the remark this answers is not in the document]
mvadari: Should there be a flag for removing the URI from the NFT, like with `NFTokenMinter`, instead of just using its absence?

src/xrpld/app/tx/detail/NFTokenUtils.cpp
scottschurr: > @scottschurr may I ask why is Slice type used here, and not Buffer?

Sure.  The simple answer is that I was looking for the most obvious type to pass in.  The only call in the code base (currently) to `nft::changeTokenURI` looks like this:
```
    return nft::changeTokenURI(view(), owner, nftokenID, ctx_.tx[~sfURI]);
```
The awkward question is "What type does `ctx_.tx[~sfURI]` return?"  I could dig through a bunch of templates in `STObject` and try to figure that out.  But it's really easy to get lost in there.  So in situations like this I usually rely on my compiler to tell me.

I picked a type that I'm pretty sure is not constructible or assignable from whatever `ctx_.tx[~sfURI]` returns.  I don't remember the specific type I picked this time, but I often pick `int*`.  Then I tell the compiler that I want to assign `ctx_.tx[~sfURI]` to an `int*`.  It might look like this:
```
    [[maybe_unused]] int* junk = ctx_.tx[~sfURI];  // !!!! DEBUG !!!!
```
Then I compile hoping to get a useful error message out.  I use clang, so I this case I got:
```
/Users/scott/Development/rippled/src/xrpld/app/tx/detail/NFTokenModify.cpp:87:27: error: no viable conversion from 'std::optional<std::decay_t<typename STBlob::value_type>>' (aka 'optional<ripple::Slice>') to 'int *'
    [[maybe_unused]] int* junk = ctx_.tx[~sfURI];
```
Way off at the end of the error message it tells me "no viable conversion from ...  (aka `optional<ripple::Slice>`) to `int *`.

So the error message told me that, in order to avoid conversions, I needed to pass an `optional<ripple::Slice>` to `changeTokenURI()`.  Once I had that information I removed the test code from the code base.

A `Slice` is really cheap to copy, so it often makes sense to pass a `Slice` by value.  But I'm all the way up to an `optional<ripple::Slice>` which is not quite so obvious about pass-by-value or pass-by-const-ref.  I was conservative and picked pass-by-const-ref.

> Also another related question, in `STBlob`, it uses `Buffer` to store `value_`, but why `value_type` have type `Slice` and not `Buffer`?

That's a great question, and it's all about efficiency.  If we look inside `Buffer` it holds onto the buffer contents like this:
```
    std::unique_ptr<std::uint8_t[]> p_;
```
It's not a `shared_ptr`, it's a `unique_ptr`.  So the `Buffer` holds a `unique_ptr` to the only copy of the data.  If you're not acquainted with `unique_ptr`, it's worth getting to know.  Here's what CppReference has to say: https://en.cppreference.com/w/cpp/memory/unique_ptr

As CppReference notes, a `unique_ptr` is neither copy constructible or copy assignable.  So we _can't_ copy the `Buffer`.  The way it holds its data does not allow us to do so.  We have to get information about the contents of the `Buffer` some other way.

That way is called a `Slice`.  If you look inside a `Slice`, you'll see that it only holds two pieces of data:
```
private:
    std::uint8_t const* data_ = nullptr;
    std::size_t size_ = 0;
```
We have a pointer to the `const` data (so we can't mess with the data inside the `Buffer`) and a length.  If this code had been written with C++20, we'd probably be using the `ranges` library.  A `Slice` looks very much like a `ranges::contiguous_range`.  But this was done using C++11.

So `Slice` gives us a way to pass around information about the contents of the `Buffer` without actually copying `Buffer` or moving its contents around.

Why go to all this effort instead of putting the data inside of a `shared_ptr` and passing around `shared_ptr`s?  We gain both efficiency and safety.

**Efficiency**

A `shared_ptr` requires atomic operations, which are compute expensive.  If you copy a `shared_ptr` you have to increment its `atomic` count.  And every time a `shared_ptr` is destroyed its `atomic` count is decremented.  The `unique_ptr` has none of that overhead.

**Safety**
A `shared_ptr` is risky in two ways.

1. First is that if you have multiple `shared_ptr`s to non-`const` data, then anybody with one of those `shared_ptr` can modify the contents of the `Buffer`.  That leads to debugging mayhem if anyone doesn't follow rules closely.  That's why a `Slice` holds a `std::uint8_t const* data_`.  That `const` keeps anyone holding a`Slice` from messing with the contents of the `Buffer`.

2. The other problem with `shared_ptr` is that the contents of the `shared_ptr` hang around until the last copy of that `shared_ptr` is destroyed.  It is often difficult to know who has responsibility for destroying the contents of a `shared_ptr`.  But a `unique_ptr` has none of that ambiguity.

With `unique_ptr` the lifetime is obvious.  Whoever holds the `unique_ptr` has responsibility for the lifetime.  If the lifetime needs to exceed the lifetime of the holding object, then the `unique_ptr` must be _moved_, so the originally holding object no longer has the contents of the `unique_ptr`.

That's a very long winded answer.  I hope it made some sense.

shawnxie999: Really appreciate the detailed response. Thanks Scott! ❤️

yinyiqian1 (approved):
looks good

src/test/app/NFToken_test.cpp
```diff
@@ -7744,6 +7744,273 @@ class NFTokenBaseUtil_test : public beast::unit_test::suite
         }
     }
 
+    void
+    testNFTokenModify(FeatureBitset features)
+    {
+        testcase("Test modify");
```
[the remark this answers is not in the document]
tequdev: fixed!

shawnxie999:
I suggest holding off on merging this PR until the [Clio counterpart](https://github.com/XRPLF/clio/pull/1525) is also approved. While this isn't standard practice, the NFT functionality is in a unique position where its Clio implementation is necessary for it to work correctly. IMO ideally we should make sure the feature works on both ends before merging both into develop.

tequdev:
> I suggest holding off on merging this PR until the [Clio counterpart](https://github.com/XRPLF/clio/pull/1525) is also approved. While this isn't standard practice, the NFT functionality is in a unique position where its Clio implementation is necessary for it to work correctly. IMO ideally we should make sure the feature works on both ends before merging both into develop.

The PR has been inactive for about a month...
If it is not included in the next release (2.3.0), we will have to wait a few more months.

cindyyan317:
> > I suggest holding off on merging this PR until the [Clio counterpart](https://github.com/XRPLF/clio/pull/1525) is also approved. While this isn't standard practice, the NFT functionality is in a unique position where its Clio implementation is necessary for it to work correctly. IMO ideally we should make sure the feature works on both ends before merging both into develop.
> 
> The PR has been inactive for about a month... If it is not included in the next release (2.3.0), we will have to wait a few more months.

Hi @tequdev ,
The  [Clio counterpart](https://github.com/XRPLF/clio/pull/1525)  has been finished. After this PR is merged to rippled, Clio will update the XRPL library and do the final test.

src/xrpld/app/tx/detail/NFTokenModify.cpp
```diff
@@ -0,0 +1,90 @@
...
+    if (!(nft::getFlags(ctx.tx[sfNFTokenID]) & nft::flagMutable))
+        return tecNO_PERMISSION;
+
+    // Verify permissions for the issuer
+    if (AccountID const issuer = nft::getIssuer(ctx.tx[sfNFTokenID]);
+        issuer != account)
+    {
+        if (auto const sle = ctx.view.read(keylet::account(issuer)); !sle)
+            return tecNO_PERMISSION;  // Should not be possible
+        else if (auto const minter = (*sle)[~sfNFTokenMinter];
```
mvadari, line 72:
```suggestion
        if (auto const minter = (*sle)[~sfNFTokenMinter];
```

src/xrpld/app/tx/detail/NFTokenModify.cpp
```diff
@@ -0,0 +1,90 @@
...
+    // Check if the NFT is mutable
+    if (!(nft::getFlags(ctx.tx[sfNFTokenID]) & nft::flagMutable))
+        return tecNO_PERMISSION;
+
+    // Verify permissions for the issuer
+    if (AccountID const issuer = nft::getIssuer(ctx.tx[sfNFTokenID]);
+        issuer != account)
+    {
+        if (auto const sle = ctx.view.read(keylet::account(issuer)); !sle)
+            return tecNO_PERMISSION;  // Should not be possible
```
mvadari, line 71: Shouldn't this have a different error (like `tecINTERNAL` or `tefEXCEPTION`) if this shouldn't be possible?

src/xrpld/app/tx/detail/NFTokenUtils.cpp
```diff
@@ -241,6 +241,39 @@ compareTokens(uint256 const& a, uint256 const& b)
...
+{
+    std::shared_ptr<SLE> const page = locatePage(view, owner, nftokenID);
+
+    // If the page couldn't be found, the given NFT isn't owned by this account
+    if (!page)
+        return tecNO_ENTRY;
```
mvadari, line 255: Same here

src/xrpld/app/tx/detail/NFTokenUtils.cpp
```diff
@@ -241,6 +241,39 @@ compareTokens(uint256 const& a, uint256 const& b)
...
+    // Locate the NFT in the page
+    STArray& arr = page->peekFieldArray(sfNFTokens);
+
+    auto const nftIter =
+        std::find_if(arr.begin(), arr.end(), [&nftokenID](STObject const& obj) {
+            return (obj[sfNFTokenID] == nftokenID);
+        });
+
+    if (nftIter == arr.end())
+        return tecNO_ENTRY;
```
mvadari, line 266: Same here

src/xrpld/app/tx/detail/NFTokenModify.cpp
tequdev: fixed

src/xrpld/app/tx/detail/NFTokenUtils.cpp
tequdev: fixed

src/xrpld/app/tx/detail/NFTokenUtils.cpp
tequdev: fixed

mvadari:
Build error - 
```
src/xrpld/app/tx/detail/NFTokenModify.cpp:72:35: error: use of undeclared identifier 'sle'
   72 |         if (auto const minter = (*sle)[~sfNFTokenMinter]; minter != account)
```

include/xrpl/protocol/detail/transactions.macro
```diff
@@ -447,6 +447,13 @@ TRANSACTION(ttCREDENTIAL_DELETE, 60, CredentialDelete, ({
     {sfCredentialType, soeREQUIRED},
 }))
 
+/** This transaction type modify a NFToken */
+TRANSACTION(ttNFTokenModify, 61, NFTokenModify, ({
```
cindyyan317, line 451: 🤔  This name seemed changed. Why `ttNFTokenModify` is using different naming style?

tequdev: Good catch! fixed!

mvadari:
Good to go from the automation tests. @tequdev please update the branch and ensure that this is ready to merge from your side.

tequdev:
ready to merge👍

mvadari:
@bthomee this is ready to merge whenever you get a chance

bthomee (approved):

mvadari:
@tequdev it looks like this code doesn't check for invalid flags - can that be added in a separate PR?

tequdev:
> @tequdev it looks like this code doesn't check for invalid flags - can that be added in a separate PR?

Nice catch!

You can check the changes in this PR.
#5246

PR #5052, opened 2024-06-21
Add xrpld build option and Conan package test

thejohnfreeman:
Related to #5028. The Clio trigger from that branch needs to be merged with this one.

### DIFF
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -232,3 +232,52 @@ jobs:
             token: ${{ secrets.CODECOV_TOKEN }}
           attempt_limit: 5
           attempt_delay: 210000 # in milliseconds
+
+  conan:
+    needs: dependencies
+    runs-on: [self-hosted, heavy]
+    container: rippleci/rippled-build-ubuntu:aaf5e3e
+    env:
+      build_dir: .build
+      configuration: Release
+    steps:
+      - name: download cache
+        uses: actions/download-artifact@v3
+        with:
+          name: linux-gcc-${{ env.configuration }}
+      - name: extract cache
+        run: |
+          mkdir -p ~/.conan
+          tar -xzf conan.tar -C ~/.conan
+      - name: check environment
+        run: |
+          env | sort
+          echo ${PATH} | tr ':' '\n'
+          conan --version
+          cmake --version
+      - name: checkout
+        uses: actions/checkout@v4
+      - name: dependencies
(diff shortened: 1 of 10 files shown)


examples/example/src/example.cpp
```diff
@@ -0,0 +1,8 @@
+#include <cstdio>
+
+#include <xrpl/protocol/BuildInfo.h>
+
+int main(int argc, const char** argv) {
```
godexsoft, line 5: nit: I thought rippled codebase preferred east-const, not that it matters ☝️

godexsoft (approved):
I like the `with_xrpld` change. We should start using it in Clio asap too 👍
Imo worth merging. We can add the Clio notification on top of it right here or in the other PR once this is merged.

godexsoft:
@Bronek this is still awaiting your review 🙇

.github/workflows/nix.yml
```diff
@@ -232,3 +232,52 @@ jobs:
...
+        run: |
+          cd examples/example
+          mkdir ${build_dir}
+          cd ${build_dir}
+          conan install .. --output-folder . \
+            --require-override ${reference} --build missing
+          cmake .. \
+            -DCMAKE_TOOLCHAIN_FILE:FILEPATH=./build/${configuration}/generators/conan_toolchain.cmake \
+            -DCMAKE_BUILD_TYPE=${configuration}
+          cmake --build .
```
Bronek, line 283: I would also add a step to run the example and ` | grep -F "${version}"` from its output, but that's just me.

Also, I love the idea behind this newly added workflow.

examples/example/conanfile.py
```diff
@@ -0,0 +1,59 @@
...
+    default_options = {
+        'shared': False,
+        'fPIC': True,
+        'xrpl:with_xrpld': False,
+    }
+
+    requires = ['xrpl/2.2.0-rc1@jfreeman/nodestore']
+    generators = ['CMakeDeps', 'CMakeToolchain']
+
+    exports_sources = [
```
Bronek, line 28: Is this needed in this `example` project ?

Bronek (approved):
I like it, just some minor comments

examples/example/conanfile.py
thejohnfreeman: No, it's just part of the template I use.

thejohnfreeman:
I renamed the option to be more consistent with the other option names (especially `tests`).

PR #5054, opened 2024-06-24
chore: replaced friend for unit-test with mock

vlntb:
### Context of Change

Replacing friend relationship between tested class (STLedgerEntry) and unit-test (Invariants_test) in favour of using mock. The change has no functional impact.

### Type of Change

- [x] Removed Invariants_test as a friend from STLedgerEntry.
- [x] Declared a mock class that avoids checking LedgerFormats on construction.
- [x] Made a dummy Nickname ledger entry to trigger an invalid ledger entry error.

### API Impact

- None

### DIFF
--- a/src/test/ledger/Invariants_test.cpp
+++ b/src/test/ledger/Invariants_test.cpp
@@ -23,7 +23,7 @@
 #include <xrpld/app/tx/detail/ApplyContext.h>
 #include <xrpld/app/tx/detail/Transactor.h>
 #include <xrpl/beast/utility/Journal.h>
-#include <xrpl/protocol/STLedgerEntry.h>
+#include <xrpl/protocol/digest.h>
 #include <boost/algorithm/string/predicate.hpp>
 
 namespace ripple {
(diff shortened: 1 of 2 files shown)


src/test/ledger/Invariants_test.cpp
```diff
@@ -188,15 +188,19 @@ class Invariants_test : public beast::unit_test::suite
...
+                {
+                    public:
+                    STLedgerEntryMock(Keylet const& k): STObject(sfLedgerEntry)
+                    {
+                    }
+                };
+
+                // Make a dummy Nickname ledger entry to trigger an invalid ledger entry error.
+                Keylet nickname = {static_cast<LedgerEntryType>(0x006e), sha512Half('n', A1, (*sle)[sfSequence] + 2)};
+                auto sleNew = std::reinterpret_pointer_cast<STLedgerEntry>(std::make_shared<STLedgerEntryMock>(nickname));
```
seelabs, line 202: Nice idea! I'm honestly not sure if this is UB or not (I suspect it is, but I'd also need to look at what the rules are for type punning).

Bronek: I think it is an UB, because of `reinterpret_cast` between two unrelated types. Also, in `STLedgerEntry::key()` (called by `ac.view().insert(sleNew)`), where does the key value of the `sleNew` comes from ?

I tried something similar with `std::bit_cast` which is considered to be "a better way" to do type punning in C++, but despite getting code that both compiles and passes tests, I think it _might_ be also UB. In any case:

```c++
class STLedgerEntryMock : public STObject, public CountedObject<STLedgerEntry> 
{
    uint256 key_;
    LedgerEntryType type_;

public:
    STLedgerEntryMock(Keylet const& k):
        STObject(sfLedgerEntry), key_(k.key), type_(k.type)
    {}
};
static_assert(sizeof(STLedgerEntryMock) == sizeof(STLedgerEntry));

// Make a dummy Nickname ledger entry to trigger an invalid ledger entry error.
Keylet nickname = {static_cast<LedgerEntryType>(0x006e), sha512Half('n', A1, (*sle)[sfSequence] + 2)};                 std::shared_ptr<STLedgerEntryMock> sleMockPtr{new STLedgerEntryMock{nickname}};
auto sleNew = std::shared_ptr<STLedgerEntry>(sleMockPtr, std::bit_cast<STLedgerEntry*>(sleMockPtr.get()));
```

Bronek: FWIW If we did `std::bit_cast` on the actual object (i.e. no a pointer) that won't compile, because `std::bit_cast` is not enabled on classes which are not trivially copyable. If we used `std::memcpy` instead of `std::bit_cast` on the actual object, it would compile but also certainly an UB, because of missing side-effects of the `CountedObject` destructor. So I cannot really think of a standard-compliant way of doing it.

Bronek: Right, so type punning is subject to this rule in C++ 

https://eel.is/c++draft/basic.lval#11

... which makes all of the above an UB. Also, there is very little semantic difference between `reinterpret_cast<T*>` and `std::bit_cast<T*>` , which in this case does not matter.

Bronek (changes requested):
Unfortunately, this is an UB. Type punning rules in C++ https://eel.is/c++draft/basic.lval#11 do not allow for this kind of cast as attempted in the updated unit test.

vlntb:
Thanks for the feedback. I agree that comparing leaking test information into the code with UB from type punning, the latter is the most harmful crime. If we want to support white-box testing in the unit tests, using friends might not be that bad. I am closing this PR.

PR #5056, opened 2024-06-24
fix account_objects to disallow filtering by types that an account ca…

### LINKED ISSUE #4681: account_objects: Disallow filtering by types that an account can't own.
Inspired by #4678, it would reduce confusion if the `account_objects` API method threw an error in cases where the user attempted to filter by a ledger entry type that accounts can't own.

(Note: Clio already does this. The documentation has always omitted mention of the types that accounts can't own.)

## Summary

The `account_objects` method uses shared code to allow the user to specify a `type` of ledger entry and filter results to only the given type.

Some types, like the `Amendments` table, can't be owned by any account, so specifying that type in `account_objects` always returns an empty list. It might even be a paginated empty list, since the API implementation currently—as best I can tell—needs to walk the entire owner directory to confirm that no entry of the given type is present; it doesn't have a concept of which entries can or can't be present.

[Source file: `AccountObjects.cpp`](https://github.com/XRPLF/rippled/blob/develop/src/ripple/rpc/handlers/AccountObjects.cpp)

[Example request:](https://xrpl.org/websocket-api-tool.html?server=wss%3A%2F%2Fxrplcluster.com%2F&req=%7B%22id%22%3A1%2C%22command%22%3A%22account_objects%22%2C%22account%22%3A%22rf1BiGeXwwQoi8Z2ueFYTEXSwuJYfV2Jpn%22%2C%22ledger_index%22%3A%22validated%22%2C%22type%22%3A%22fee%22%2C%22limit%22%3A10%7D)

```json
{
  "id": 1,
  "command": "account_objects",
  "account": "rf1BiGeXwwQoi8Z2ueFYTEXSwuJYfV2Jpn",
  "ledger_index": "validated",
  "type": "fee",
  "limit": 10
}
```

## Motivation
Users may be confused why they get no results if they expect something to appear that isn't there (this came up with AMM). An error more clearly explains the situation. Also, if they do, the server does extra work looking for something that doesn't exist.

## Solution
The following types cannot ever appear in an owner directory, and should return an `invalidParams` error if the user specifies their short name in the `type` field of an `account_objects` request (short name in parens is what the user would specify for the command—see also #4393):

- Amendments (`amendments`)
- DirectoryNode (`directory`)
- FeeSettings (`fee`)
- LedgerHashes (`hashes`)
- NegativeUNL (no short name—can't currently be filtered, but should be disallowed if a filter is added in the future)

Currently it's TBD whether or not an AMM entry gets added to the associated AMM-account's owner directory, but if not, it should be excluded also.

Filtering in the `ledger` and `ledger_data` methods should be unaffected. Those methods can return _full_ ledger data, which does include these entry types, so it makes sense to allow filtering on them there.

## Notes on API Compatibility

Because (a) the Clio API already works this way, and (b) the documentation only lists the types that can appear and would still be considered valid, I consider this a non-breaking change and think it does not need to be put on an API version switch.

yinyiqian1:
fix #4681 

### Type of Change

- [ ] Documentation update
- [ ] Release

### API Impact

### DIFF
--- a/src/test/rpc/AccountObjects_test.cpp
+++ b/src/test/rpc/AccountObjects_test.cpp
@@ -556,11 +556,11 @@ public:
         Env env(*this, features);
 
         // Make a lambda we can use to get "account_objects" easily.
-        auto acct_objs = [&env](
-                             AccountID const& acct,
-                             std::optional<Json::StaticString> const& type,
-                             std::optional<std::uint16_t> limit = std::nullopt,
-                             std::optional<std::string> marker = std::nullopt) {
+        auto acctObjs = [&env](
+                            AccountID const& acct,
+                            std::optional<Json::StaticString> const& type,
+                            std::optional<std::uint16_t> limit = std::nullopt,
+                            std::optional<std::string> marker = std::nullopt) {
             Json::Value params;
             params[jss::account] = to_string(acct);
             if (type)
@@ -574,7 +574,7 @@ public:
         };
 
         // Make a lambda that easily identifies the size of account objects.
-        auto acct_objs_is_size = [](Json::Value const& resp, unsigned size) {
+        auto acctObjsIsSize = [](Json::Value const& resp, unsigned size) {
             return resp[jss::result][jss::account_objects].isArray() &&
                 (resp[jss::result][jss::account_objects].size() == size);
         };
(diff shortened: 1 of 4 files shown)


src/test/rpc/AccountObjects_test.cpp
```diff
@@ -579,19 +579,34 @@ class AccountObjects_test : public beast::unit_test::suite
...
+            return resp[jss::result].isMember(jss::error) &&
+                resp[jss::result][jss::error_message] ==
+                "Invalid field \'type\'.";
+        };
+
+        // Make a lambda that checks if different filtering type is invalid
+        auto check_invalid_filter_types = [&](AccountID const& acct) {
+            // we expect invalid field type reported for the following types
+            if (!acct_objs_type_is_invalid(acct_objs(acct, jss::amendments)) ||
+                !acct_objs_type_is_invalid(acct_objs(acct, jss::directory)) ||
```
mvadari, line 593: Can't an account own directory objects?

src/xrpld/rpc/detail/RPCHelpers.cpp
```diff
@@ -986,6 +986,20 @@ chooseLedgerEntryType(Json::Value const& params)
     return result;
 }
 
+bool
+isAccountObjectsValidType(LedgerEntryType const& type)
+{
+    switch (type)
+    {
+        case LedgerEntryType::ltAMENDMENTS:
+        case LedgerEntryType::ltDIR_NODE:
```
mvadari, line 995: Can't an account own directory objects?

src/xrpld/rpc/handlers/AccountObjects.cpp
```diff
@@ -220,6 +220,9 @@ doAccountObjects(RPC::JsonContext& context)
     {
         auto [rpcStatus, type] = RPC::chooseLedgerEntryType(params);
 
+        if (!RPC::isAccountObjectsValidType(type))
```
cindyyan317, line 223: Looks like this isAccountObjectsValidType function is used in this handler only. We can just add a function/lambda here ?

src/test/rpc/AccountObjects_test.cpp
```diff
@@ -579,19 +579,35 @@ class AccountObjects_test : public beast::unit_test::suite
                 (resp[jss::result][jss::account_objects].size() == size);
         };
 
+        // Make a lambda that checks if the response has error for invalid type
+        auto acct_objs_type_is_invalid = [](Json::Value const& resp) {
```
scottschurr, line 583: I respect and honor that you're following file-local naming conventions by naming this lambda in snake case.  You're following the model of the `acct_objs` and `acct_objs_is_size` lambdas just above.

Unfortunately the above lambda, `acct_objs_is_size`, does not follow the general guidance in this code base that functions are named in `camelCase` with an initial lower case letter.  So I encourage you to rename your new lambda to `acctObjsTypeIsInvalid`.  If you're feeling really helpful, you could also consider renaming...

- `acct_objs`,
- `acct_objs_is_size`, and
- `scenv_acct_objs`

to use `camelCase` as well.  That would help to stem the spread of the rot.  But I also understand that you may not want to extend the scope of the pull request.

scottschurr (commented):
The fix looks good as far as I'm concerned.  However,

- I raised a concern about how test failures will be handled, and suggested how I would resolve it.  But I'm open to other thoughts.
- I also raised a concern about lambda naming (in general) in the test file.

I pushed commits with my suggested fixes to this branch: https://github.com/scottschurr/rippled/commits/yinyiqian1-br_fix_invalid_type/

- My suggestions for improving test failure messages are in commit ad508ffcc31fdb10217ded16cfe06fd7765d57c4
- My suggestions for renaming the lambdas are in this commit: b419bd0e5fa0e9bdcf31523ac422c369597fd951

You are welcome to look at, cherry-pick, or ignore those commits as you wish.  But at the very least I hope for some discussion.

Thanks for the contribution!

scottschurr (approved):
Looks great to me.  Thanks for being willing to take on the drive-by naming fix!  In my opinion things like that can help reduce bit rot in the code base.

cindyyan317 (approved):
LGTM ! 👍

shawnxie999 (approved):
👍

PR #5061, opened 2024-07-10
[TRIVIAL] Fix NuDB build error

ximinez:

Fix a build error that seems to be limited to non-unity builds on Windows (Visual Studio) due to a missing `#include` in NuDB.

### Context of Change

I think the reordering of `#includes` in the physical redesign PR (#4997) removed a silent dependency from one of the headers in NuDB. The dependency version of NuDB didn't change, as far as I can tell. CI didn't catch the issue because the Windows build uses unity to save time.

The error is:
```
C:\Users\passp\.conan\data\nudb\2.0.8\_\_\package\5ab84d6acfe1f23c4fae0ab88f26e3a396351ac9\include\nudb/detail/stream.hpp(64,20): error C2039: 'logic_error': is not a member of 'std' [C:\Dev\rippled\upstream\build\cmake\msvc19.OFF\rippled.vcxproj]
```
`std::logic_error` requires `<stdexcept>`, but `nudb/detail/stream.hpp` does not include it.

My assumption is some other header that had previously been ordered before `stream.hpp` and included `<stdexcept>` is now ordered after `stream.hpp`. I don't think it's worth figuring out which file it is, because it's obviously a problem with NuDB.

### Type of Change

- [X] Build

## Future Tasks

See also: https://github.com/cppalliance/NuDB/pull/100

### DIFF
--- a/BUILD.md
+++ b/BUILD.md
@@ -98,6 +98,12 @@ Update the compiler settings:
    conan profile update settings.compiler.cppstd=20 default
    ```
 
+Configure Conan to use recipe revisions:
+
+   ```
+   conan config set general.revisions_enabled=1
+   ```
+
 **Linux** developers will commonly have a default Conan [profile][] that compiles
 with GCC and links with libstdc++.
 If you are linking with libstdc++ (see profile setting `compiler.libcxx`),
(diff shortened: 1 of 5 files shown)


ximinez:
Rewrote this to patch the source via Conan.

thejohnfreeman (commented):
Do you want to add an instruction to `BUILD.md` explaining how to export the right version of `nudb` from `external/nudb`?

BUILD.md
```diff
@@ -98,6 +98,12 @@ Update the compiler settings:
    conan profile update settings.compiler.cppstd=20 default
    ```
 
+Configure Conan to use recipe revisions:
```
thejohnfreeman, line 101:
```suggestion
Configure Conan (1.x only) to use recipe revisions:
```

thejohnfreeman (changes requested):

BUILD.md
ximinez: Done

thejohnfreeman (approved):

PR #5063, opened 2024-07-11
fix: change error for invalid `feature` param in `feature` RPC

mvadari:

This PR changes the error message for invalid values of `feature` in the `feature` RPC. Now, if a non-string is passed into the function, it will return `invalidParams`, instead of `badFeature` as it did before.

### Context of Change

In rippled 2.2.0, a non-admin version of `feature` was added. This makes the need for proper error handling more important.

### Type of Change

### API Impact

While this is technically a breaking change, I consider it more of a bugfix. Only the error message is changing.

## Test Plan

Additional unit tests were added.


godexsoft (approved):
Looks correct to me 👍

intelliot (approved):
OK because this is a relatively new RPC, and it is not involved in how integrations do tx processing (where any potential API breakage would be more painful)

mvadari:
This is ready to merge.

PR #5064, opened 2024-07-12
docs: clean up API-CHANGELOG.md

intelliot:

Reorganize the contents of API-CHANGELOG so that the newest (and thus most relevant) information is at the top (i.e., reverse chronological order within each of the two sections, API Versions and XRP Ledger server versions)

### Context of Change

A reasonable follow-up would be documentation of how/when the `full`/`accounts` parameters will be deprecated and eventually removed. Ref: https://github.com/XRPLF/clio/issues/1537

### Type of Change

- [x] Documentation update


intelliot:
Trivial: documentation only, no code change

PR #5065, opened 2024-07-12
fix: better error handling in account RPCs

mvadari:

This PR adds a check for whether the `account` and `ident` params in several account-based RPCs are strings. If not, it returns `invalidParams`. Some error messages are also improved to be more specific.

Note: This PR looks a lot larger than it is; it is really just the same change(s) applied over and over again across several RPCs.

### Context of Change

rippled currently returns `Internal error` in this scenario. 

The error in the logs is `Caught throw: Type is not convertible to string`

### Type of Change

### API Impact

This is a bugfix, not a breaking change, but there isn't an option for that.

## Test Plan
Added unit tests that fail before this patch and now pass.

## Future Tasks
Audit other RPCs for this sort of issue.

### DIFF
--- a/src/ripple/rpc/handlers/AccountCurrenciesHandler.cpp
+++ b/src/ripple/rpc/handlers/AccountCurrenciesHandler.cpp
@@ -40,13 +40,13 @@ doAccountCurrencies(RPC::JsonContext& context)
     if (params.isMember(jss::account))
     {
         if (!params[jss::account].isString())
-            return rpcError(rpcINVALID_PARAMS);
+            return RPC::invalid_field_error(jss::account);
         strIdent = params[jss::account].asString();
     }
     else if (params.isMember(jss::ident))
     {
         if (!params[jss::ident].isString())
-            return rpcError(rpcINVALID_PARAMS);
+            return RPC::invalid_field_error(jss::ident);
         strIdent = params[jss::ident].asString();
     }
 
--- a/src/ripple/rpc/handlers/AccountCurrenciesHandler.cpp
+++ b/src/ripple/rpc/handlers/AccountCurrenciesHandler.cpp
@@ -33,18 +33,28 @@ doAccountCurrencies(RPC::JsonContext& context)
 {
     auto& params = context.params;
 
+    if (!(params.isMember(jss::account) || params.isMember(jss::ident)))
+        return RPC::missing_field_error(jss::account);
+
+    std::string strIdent;
+    if (params.isMember(jss::account))
+    {
+        if (!params[jss::account].isString())
+            return rpcError(rpcINVALID_PARAMS);
+        strIdent = params[jss::account].asString();
+    }
+    else if (params.isMember(jss::ident))
+    {
+        if (!params[jss::ident].isString())
+            return rpcError(rpcINVALID_PARAMS);
+        strIdent = params[jss::ident].asString();
+    }
+
     // Get the current ledger
     std::shared_ptr<ReadView const> ledger;
     auto result = RPC::lookupLedger(ledger, context);
     if (!ledger)
         return result;
-
-    if (!(params.isMember(jss::account) || params.isMember(jss::ident)))
-        return RPC::missing_field_error(jss::account);
-
-    std::string const strIdent(
-        params.isMember(jss::account) ? params[jss::account].asString()
-                                      : params[jss::ident].asString());
 
     // Get info on account.
     auto id = parseBase58<AccountID>(strIdent);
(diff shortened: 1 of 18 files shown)


mvadari:
This change could also be done by editing `parseBase58` to return an empty `std::optional` instead of crashing (which would result in an `rpcACT_MALFORMED` instead of `rpcINVALID_PARAMS`, which I believe differs from what Clio returns in this case).

src/ripple/rpc/handlers/AccountCurrenciesHandler.cpp
```diff
@@ -33,19 +33,29 @@ doAccountCurrencies(RPC::JsonContext& context)
 {
     auto& params = context.params;
 
+    if (!(params.isMember(jss::account) || params.isMember(jss::ident)))
+        return RPC::missing_field_error(jss::account);
+
+    std::string strIdent;
```
godexsoft, line 39: nit: can be avoided via immediately invoked lambda

src/ripple/rpc/handlers/AccountInfo.cpp
```diff
@@ -23,6 +23,7 @@
 #include <ripple/ledger/ReadView.h>
 #include <ripple/protocol/ErrorCodes.h>
 #include <ripple/protocol/Indexes.h>
+#include <ripple/protocol/RPCErr.h>
```
godexsoft, line 26: Wasn't all of the `ripple` stuff renamed to `xrpl`? or is it not merged yet? I know there is an alias - i'd refactor the entire codebase to use the new `xrpl` (in a separate PR). surprised it's not already done tbh.

src/ripple/rpc/handlers/AccountInfo.cpp
```diff
@@ -53,9 +54,17 @@ doAccountInfo(RPC::JsonContext& context)
 
     std::string strIdent;
```
godexsoft, line 55: nit: Same improvement possible

godexsoft (approved):
Looks fine to me. Leaving a few nits and questions 👍

src/ripple/rpc/handlers/AccountCurrenciesHandler.cpp
mvadari: Given that the lambda would be handling errors, it'll be more readable etc. if it's left as-is, and probably wouldn't have any performance improvements to switch.

src/test/app/AccountTxPaging_test.cpp
```diff
@@ -271,6 +271,6 @@ class AccountTxPaging_test : public beast::unit_test::suite
     }
 };
 
-BEAST_DEFINE_TESTSUITE(AccountTxPaging, app, ripple);
+BEAST_DEFINE_TESTSUITE(AccountTxPaging, rpc, ripple);
```
ximinez, line 274: Is this change intentional?

src/test/app/PayChan_test.cpp
```diff
@@ -873,6 +873,25 @@ struct PayChan_test : public beast::unit_test::suite
         auto const chan1Str = to_string(channel(alice, bob, env.seq(alice)));
         env(create(alice, bob, channelFunds, settleDelay, pk));
         env.close();
+        {
+            // test account non-string
+            auto testInvalidAccountParam = [&](auto const& param) {
+                Json::Value params;
+                params[jss::account] = param;
+                auto jrr = env.rpc(
+                    "json", "account_tx", to_string(params))[jss::result];
```
ximinez, line 882:
```suggestion
                    "json", "account_channels", to_string(params))[jss::result];
```
?

ximinez (changes requested):
Because this branch is based on `master`, `RippledCode.cmake` needs to be updated to reflect the renaming of `AccountLinesRPC_test.cpp` to `AccountLines_test.cpp`.

src/test/app/AccountTxPaging_test.cpp
mvadari: Yes, because it made more sense to me to have it under `rpc` and I figured I'd update it while I was here. If it's a breaking change or anything or just ill advised, I'm happy to change it back.

src/test/app/PayChan_test.cpp
mvadari: Typo 🤦

src/test/app/AccountTxPaging_test.cpp
ximinez: Only problem is that we've already got a pending PR (#4794) that updates a bunch of unit tests to make them match their directory names. This would conflict with that, at least in principal, by moving away from that type of naming convention. If you think this makes more sense to be named `rpc`, it might make sense to move the files to `src/test/rpc`.

mvadari: Upon further review, I think I confused this test file with the `account_tx` RPC and this test file seems to be testing something else. I rolled back the change.

ximinez (approved):

mvadari:
Closing in favor of #5078

PR #5066, opened 2024-07-13
Remove shards

thejohnfreeman:
This change removes all text related to shards, including code, tests, documentation, and configuration settings.

99% of the change is removed lines and all remaining tests pass. Only 3 tests are removed:

- `nodestore/DatabaseShard_test.cpp`
- `rpc/NodeToShardRPC_test.cpp`
- `rpc/ShardArchiveHandler_test.cpp`

Only 2 tests are modified:

- `nodestore/Database_test.cpp`: Remove tests of `Database::ledgersPerShard()`.
- `rpc/RPCCall_test.cpp`: Remove tests of `download_shard` and `node_to_shard`.


MarkusTeufelberger:
So what's the replacement method to reliably and trustlessly share a full history node database?

thejohnfreeman:
@MarkusTeufelberger tarball, rsync, all the most popular methods for copying data. The sharding feature was never completed. What is there to replace?

src/test/rpc/ShardArchiveHandler_test.cpp
Bronek, line 1: I can see this file listed in `src/test/rpc/ShardArchiveHandler_test.cpp` line 145

src/xrpld/rpc/ShardArchiveHandler.h
Bronek, line 1: I can see this file mentioned in `src/xrpld/net/uml/interrupt_sequence.pu`

Bronek:
There's a comment mentioning shards in `src/xrpld/nodestore/backend/NuDBFactory.cpp` (lines 152-158), it probably needs to be updated

Bronek:
I also see mention of `HTTPDownloader::do_session` in `src/xrpld/net/detail/HTTPDownloader.cpp` (line 127)

Bronek:
> ## [Codecov](https://app.codecov.io/gh/XRPLF/rippled/pull/5066?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=XRPLF) Report
> Attention: Patch coverage is `36.23188%` with `44 lines` in your changes missing coverage. Please review.
> 
> > Project coverage is 74.8%. Comparing base [(`a39720e`)](https://app.codecov.io/gh/XRPLF/rippled/commit/a39720e94a3fb5be07d0b526c840f2e607db42cb?dropdown=coverage&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=XRPLF) to head [(`dd6b21a`)](https://app.codecov.io/gh/XRPLF/rippled/commit/dd6b21abbdb056151ac928f0828e4e2921552726?dropdown=coverage&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=XRPLF).
> > Report is 1 commits behind head on develop.
> 
> Additional details and impacted files

We have to ignore the "failing" check from codecov, the one that is showing "36% of diff hit (target 71%)". The alternative is to add more code in unit tests, which this PR should not do.

Bronek (approved):
Good change.

src/xrpld/rpc/handlers/Tx.cpp
```diff
@@ -100,7 +100,6 @@ doTxPostgres(RPC::Context& context, TxArgs const& args)
     if (locator.isFound())
     {
         auto start = std::chrono::system_clock::now();
-        // The second argument of fetch is ignored when not using shards
         if (auto obj = context.app.getNodeFamily().db().fetchNodeObject(
                 locator.getNodestoreHash(), locator.getLedgerSequence()))
```
ximinez, line 104: > The second argument of fetch is ignored when not using shards

So does it make sense to remove the second argument to `fetchNodeObject` everywhere?

ximinez (changes requested):
A good change, which will simplify a lot of the remaining code. I left a few suggestions and questions.

However, I have to ask, sort of as a follow-up to  @MarkusTeufelberger's question: Why not finish the feature? It wasn't _bad_ code, and it had potential for making it a lot easier to manage space, especially as pertains to `online_delete`. I don't know what else was planned for it, either.

scottschurr:
@ximinez asked...
> Why not finish the feature?

In part because the code is particularly stale.  Mickey gave me a core dump on the state of shards in November of 2021.  No one has done any development (at all) on shards since then.  That's a good 2.5 years worth of stale.  Mickey was hoping that I would pick up development of shards.  But I didn't.  Ripple had other plans for what I needed to work on.

Also, Ripple ended up throwing it weight into clio, which was described as an alternative to shards as a solution for the full history problem.  If you believe that clio solves the full history problem, then shards are a lot less useful.

Another reason to not finish Shards is because Mickey was just discovering some additional difficult parts of shards.  The code is not bad code, but the feature is not actually done.  There's no one with a vision for how the feature should be completed.

I hope that history helps.

src/xrpld/app/rdb/README.md
```diff
@@ -2,17 +2,16 @@
...
-* For future use, the shard store will be used if the node store is absent.
 
 ## Overview
 
-Firstly, the interface `RelationalDatabase` is inherited by the classes `SQLiteDatabase` and `PostgresDatabase` which are used to operate the software's main data store (for storing transactions, accounts, ledgers, etc.). Secondly, the files under the `detail` directory provide supplementary functions that are used by these derived classes to access the underlying databases. Lastly, the remaining files in the interface (located at the top level of the module) are used by varied parts of the software to access any secondary relational databases.
+Firstly, the interface `RelationalDatabase` is inherited by the class `SQLiteDatabase` which is used to operate the software's main data store (for storing transactions, accounts, ledgers, etc.). Secondly, the files under the `detail` directory provide supplementary functions that are used by these derived classes to access the underlying databases. Lastly, the remaining files in the interface (located at the top level of the module) are used by varied parts of the software to access any secondary relational databases.
```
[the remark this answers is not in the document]
thejohnfreeman: Oops! This change was meant to go in my next PR which removes reporting mode. I'll move it over. Good catch!

src/xrpld/rpc/handlers/Tx.cpp
thejohnfreeman: It does. I started going down that route, but it is all over the place, and I started having second thoughts. We might want to keep this information available for the future. It seems to be a lesson learned the hard way that we often want _both_ the ledger digest _and_ the ledger sequence together. In fact, I added a type in my FLR2 branch just for representing the pair, which I call `LedgerIdentifier`.

src/xrpld/overlay/detail/ProtocolMessage.h
```diff
@@ -484,14 +468,6 @@ invokeProtocolMessage(
...
-        case protocol::mtGET_PEER_SHARD_INFO_V2:
-            success = detail::invoke<protocol::TMGetPeerShardInfoV2>(
-                *header, buffers, handler);
-            break;
-        case protocol::mtPEER_SHARD_INFO_V2:
-            success = detail::invoke<protocol::TMPeerShardInfoV2>(
-                *header, buffers, handler);
-            break;
         default:
             handler.onMessageUnknown(header->message_type);
```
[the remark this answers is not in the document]
thejohnfreeman: I have no strong feelings about it. I'm fine incrementing the protocol version if someone can teach me how.

thejohnfreeman:
Based on my understanding of the old design, I have reservations.

I imagine that activity among ledger objects follows a power law distribution. Most objects are rarely touched, and a few objects represent most of the touches in any given block. If we pack new objects into a shard until it reaches a size limit, and then seal that shard and distribute it, then most objects in the shard are in the small set of frequently touched objects, but a few objects in the shard will be from the infrequently touched set.

How does this affect the performance of serving particular queries I might have?

- If I want one particular object as it existed in one particular ledger, then I need a proof path from the header of that ledger to that object. If the shards are storing inner nodes, then the search might involve a path through multiple shards, held by different servers, and shards will fill up quickly because they are storing so much data from inner nodes. If the shards are not storing inner nodes, then they will store less data, but the search will involve _even more_ shards, that are _even more_ likely to be held by different servers, because we have to collect the leaves of the entire subtree to recompute the inner nodes.

- If I want a complete copy of one ledger, then as time goes on, each object becomes more likely to be the only object of interest in its shard. The number of bytes I'm downloading explodes because most of it is for other ledgers.

I think what we really want is consistent hashing of ledger objects, not bundles of ledger objects that we call shards. But we need consistent hashing in the context of a trustless, decentralized peer-to-peer network with high turnover. There may be good solutions in this problem space. I haven't seen them, but I haven't done any research.

Then, there are my engineering sentiments. I don't like the proliferation of branches that shards introduced. I'd rather just see a clean replacement of the nodestore abstraction.

src/xrpld/overlay/detail/ProtocolMessage.h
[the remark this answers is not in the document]
Bronek: FWIW @ximinez that exactly "the peer won't be punished or booted, so not harm no foul" was my thinking when I reviewed this.

src/xrpld/rpc/handlers/Tx.cpp
ximinez: That ledger sequence has a default value of 0. Do you know if there are any places that are using the default, or passing 0 in explicitly? If there aren't any, then I'm fine with leaving this as-is.

ximinez (approved):
Looks good. 

Thanks for the detailed explanations and historical info @scottschurr and @thejohnfreeman. It sets my mind at ease, and hopefully will be helpful for anyone else asking the same question.

src/xrpld/overlay/detail/ProtocolMessage.h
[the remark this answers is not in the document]
ximinez: Here's an example of how one would add a protocol version. https://github.com/ximinez/rippled/commit/f97844700229c7c0712eb79883cea2ef01b7af0f, but it's not actually useful, because older clients won't speak this protocol version, and newer clients won't support Shards anyway, so there's no need to test for it.

I updated the `PeerImp::supportsFeature` function, which for a "real" new protocol version would be called to check the feature before doing the new function or sending the new message or whatever.

src/xrpld/rpc/handlers/Tx.cpp
thejohnfreeman: I know there are places that pass 0 explicitly. Here is [one](https://github.com/XRPLF/rippled/blob/8b181ed81883ce9f0a1c54f8bad9347ac7066627/src/xrpld/app/ledger/detail/SkipListAcquire.cpp#L80) I found (in FLR1) in a quick search. I believe there are more. 0 is not a valid sequence though. The genesis ledger [starts](https://github.com/XRPLF/rippled/blob/8b181ed81883ce9f0a1c54f8bad9347ac7066627/src/xrpld/app/ledger/Ledger.cpp#L190) at 1. 0 is used to mean "unknown".

ximinez: Yeah, that's what I figured.

PR #5068, opened 2024-07-18
Enable publishing subsequent ledgers even if a prior one has not been acquired yet.

mtrippled:
Removes logjam of ledgers to update bookkeeping for if one of them has not been acquired.

### Context of Change
The bug was first noticed in testing of 2.x builds in an internal to Ripple test environment.

### Type of Change

### API Impact

This change has run internally for over a week by now, with no reproduction of the original
problem, and no new problems introduced. Prior to this, the problem was happening as 
frequently as each day.

Investigate problems acquiring recent ledgers that have yet to be published which triggered
the logjam fixed by this PR. There's possibly peer resource contention in that environment, but TBD.

### DIFF
--- a/src/xrpld/app/ledger/detail/LedgerMaster.cpp
+++ b/src/xrpld/app/ledger/detail/LedgerMaster.cpp
@@ -1445,11 +1445,21 @@ LedgerMaster::findNewLedgersToPublish(
             }
 
             // Did we acquire the next ledger we need to publish?
-            if (ledger && (ledger->info().seq == pubSeq))
+            if (ledger && (ledger->info().seq >= pubSeq))
             {
                 ledger->setValidated();
                 ret.push_back(ledger);
-                ++pubSeq;
+                if (ledger->info().seq > pubSeq)
+                {
+                    JLOG(m_journal.info())
+                        << "ready to publish " << ledger->info().seq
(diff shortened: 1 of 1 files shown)


src/xrpld/app/ledger/detail/LedgerMaster.cpp
```diff
@@ -1445,11 +1445,21 @@ LedgerMaster::findNewLedgersToPublish(
...
                 ledger->setValidated();
                 ret.push_back(ledger);
-                ++pubSeq;
+                if (ledger->info().seq > pubSeq)
+                {
+                    JLOG(m_journal.info())
+                        << "ready to publish " << ledger->info().seq
+                        << ", but skipping " << pubSeq
+                        << " because it has not been acquired.";
+                }
```
Bronek, line 1458: Is it possible to write a unit test to exercise this new execution branch ?

mtrippled: There's no coverage of this function or its caller. It's caller's caller is tested twice, but for entirely different reasons. This function's caller, LedgerMaster::doAdvance(), does a lot of things, none of which are currently tested, let alone this particular tweak.

Bronek (approved):

PR #5071, opened 2024-07-24
Fix crash inside `OverlayImpl` loops over `ids_`

Bronek:

Functions `OverlayImpl::getActivePeers` and `OverlayImpl::findPeerByPublicKey` may cause a crash when iterating over `ids_` 

### Context of Change

Functions `OverlayImpl::getActivePeers` and `OverlayImpl::findPeerByPublicKey` use the following _buggy_ pattern to iterate over `ids_` :

```c++
for (auto& [id, w] : ids_)
{
    if (auto p = w.lock())
    {
        // ...
    } // p destroyed here
}
```

The problem is that, given specific thread scheduling, when `p` is destroyed at a marked location, it might also cause the destruction of the `PeerImp` that it owned since the start of `if (auto p = w.lock())` block. This destruction in turn will call `OverlayImpl::onPeerDeactivate` which will remove the current element from `ids_` collection, hence invalidating the iterator to the removed element. This is all happening while `p` is being destroyed, that is before the `for` at the start of the loop tries to increment the iterator (that just got invalidated) to reach to the next element. Incrementing the invalidated operator is an invalid operation, in practice it will yield bad data, causing a crash.

### Type of Change

- [ ] Documentation update
- [ ] Release

### DIFF
--- a/src/test/overlay/tx_reduce_relay_test.cpp
+++ b/src/test/overlay/tx_reduce_relay_test.cpp
@@ -189,7 +189,10 @@ private:
             consumer,
             std::move(stream_ptr),
             overlay);
+        BEAST_EXPECT(
+            overlay.findPeerByPublicKey(key) == std::shared_ptr<PeerImp>{});
         overlay.add_active(peer);
+        BEAST_EXPECT(overlay.findPeerByPublicKey(key) == peer);
         peers.emplace_back(peer);  // overlay stores week ptr to PeerImp
         lid_ += 2;
         rid_ += 2;
--- a/src/xrpld/overlay/detail/OverlayImpl.cpp
+++ b/src/xrpld/overlay/detail/OverlayImpl.cpp
@@ -1207,14 +1207,14 @@ std::shared_ptr<Peer>
 OverlayImpl::findPeerByPublicKey(PublicKey const& pubKey)
 {
     std::lock_guard lock(mutex_);
-    // NOTE The purpose of p is to delay the destruction of PeerImp
-    std::shared_ptr<PeerImp> p;
-    for (auto [_, w] : ids_)
+    // NOTE The purpose of peer is to delay the destruction of PeerImp
+    std::shared_ptr<PeerImp> peer;
+    for (auto const& e : ids_)
     {
-        if (p = w.lock(); p != nullptr)
+        if (peer = e.second.lock(); peer != nullptr)
         {
-            if (p->getNodePublic() == pubKey)
-                return p;
+            if (peer->getNodePublic() == pubKey)
+                return peer;
         }
     }
     return {};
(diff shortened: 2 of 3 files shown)


Bronek:
Here's a smaller reproduction of the bug https://godbolt.org/z/9rGsx9eEP and of the proposed fix https://godbolt.org/z/qPbc9oYM9

thejohnfreeman (approved):

ximinez (approved):
This looks good.

I did a little bit of a search, and came across a few more places that _at a glance_ could have a non-zero chance of having the same problem, and could probably be quickly and easily fixed with this same pattern. Would they be worth changing in the PR, or another one?

* https://github.com/ximinez/rippled/blob/1e3ab9f76615cf9bb74b7bfeb6d2bd1cd257fe33/include/xrpl/server/detail/io_list.h#L246
* https://github.com/ximinez/rippled/blob/1e3ab9f76615cf9bb74b7bfeb6d2bd1cd257fe33/src/xrpld/app/ledger/detail/InboundLedger.cpp#L1311
* https://github.com/ximinez/rippled/blob/1e3ab9f76615cf9bb74b7bfeb6d2bd1cd257fe33/src/xrpld/app/ledger/detail/LedgerReplayer.cpp#L274 (because the lambda is used in a call to `std::for_each`)

thejohnfreeman:
@ximinez those first two are fine because the collection being iterated is held in a local variable, and the destructor of the objects in the collection cannot remove that object from the collection. I don't think the third is in danger, either, but I haven't looked very deeply.

Bronek:
> @ximinez those first two are fine because the collection being iterated is held in a local variable, and the destructor of the objects in the collection cannot remove that object from the collection. I don't think the third is in danger, either, but I haven't looked very deeply.

Yes, the first two operate on a stack-based collection within the scope of the function, so these are not a problem. The last one is not a problem either, because both collections `deltas_` and `skipLists_` operate on objects which do not remove themselves from a collection inside their respective destructors :
https://github.com/ximinez/rippled/blob/1e3ab9f76615cf9bb74b7bfeb6d2bd1cd257fe33/src/xrpld/app/ledger/detail/SkipListAcquire.cpp#L48
https://github.com/ximinez/rippled/blob/1e3ab9f76615cf9bb74b7bfeb6d2bd1cd257fe33/src/xrpld/app/ledger/detail/LedgerDeltaAcquire.cpp#L52

ximinez:
> The last one is not a problem either, because both collections `deltas_` and `skipLists_` operate on objects which do not remove themselves from a collection inside their respective destructors

I am thinking a little bit about future-proofing. This seems like a good pattern to use in general _just in case_ the behavior of these objects changes in the future. However, that is definitely beyond the scope of this PR.

PR #5092, opened 2024-08-03
Remove reporting mode

thejohnfreeman:
Builds and passes all tests, but I have not yet tested compatibility with Clio. Asking for help from the Clio team with that.


include/xrpl/protocol/ErrorCodes.h
```diff
@@ -136,9 +136,8 @@ enum error_code_i {
     rpcINVALID_LGR_RANGE = 79,
     rpcEXPIRED_VALIDATOR_LIST = 80,
 
-    // Reporting
-    rpcFAILED_TO_FORWARD = 90,
-    rpcREPORTING_UNSUPPORTED = 91,
```
godexsoft, line 141: This is used by Clio. 
We can either leave this code be or replace it in Clio with a Clio-native code instead.
The latter would slightly change the output of the 'subscribe' RPC which uses this error code.

src/xrpld/app/ledger/detail/LedgerMaster.cpp
```diff
@@ -34,10 +34,9 @@
 #include <xrpld/app/misc/TxQ.h>
 #include <xrpld/app/misc/ValidatorList.h>
 #include <xrpld/app/paths/PathRequests.h>
-#include <xrpld/app/rdb/backend/PostgresDatabase.h>
+#include <xrpld/app/rdb/RelationalDatabase.h>
```
godexsoft, line 37: Is this include still needed?

src/xrpld/app/main/Application.cpp
```diff
@@ -45,9 +45,8 @@
 #include <xrpld/app/misc/ValidatorKeys.h>
 #include <xrpld/app/misc/ValidatorSite.h>
 #include <xrpld/app/paths/PathRequests.h>
+#include <xrpld/app/rdb/RelationalDatabase.h>
```
godexsoft, line 48: Is it still needed?

src/xrpld/app/main/GRPCServer.cpp
```diff
@@ -381,8 +278,7 @@ GRPCServerImpl::CallData<Request, Response>::clientIsUnlimited()
     if (!getUser())
         return false;
     auto clientIp = getClientIpAddress();
-    auto proxiedIp = getProxiedClientIpAddress();
```
godexsoft, line 384: Are we sure this is only used when in reporting mode? This wont break Clio's unlimited use of gRPC, right?

godexsoft (approved):
Great effort 🙇 I'm leaving just a few questions to absolutely make sure we don't break Clio 👍

src/xrpld/app/ledger/detail/LedgerMaster.cpp
thejohnfreeman: Yes, for type `LedgerHashPair`, which is a struct containing two `uint256`.

src/xrpld/app/main/Application.cpp
thejohnfreeman: Yes, for `std::unique_ptr<RelationalDatabase> mRelationalDatabase;` on line 218.

src/xrpld/app/main/GRPCServer.cpp
thejohnfreeman: I can't prove it, but I think it is. I don't think there is any other gRPC client in practice. This will return `true` for all clients it did before, plus some more. In fact, this would have returned `false` for a reporting mode client acting on behalf of another client. If Clio is unlimited, that must be because it is not configured as a proxy, right?

godexsoft: Clio needs its IP in the `secure_gateway` of the `[port_grpc]`. I'm not sure if that makes it a proxy or what but without it we get `tooBusy` sometimes. Keep in mind that Clio uses the `X-User` header to get unlimited role. I'm not sure how any of that relates to this but from your explanation it sounds like some separate thing.

include/xrpl/protocol/ErrorCodes.h
godexsoft: Looks like we are going to use our own error code on Clio's side. 👍

Bronek: I suggest to just keep these two in `ErrorCodes.h`, at least for the time being - since this is libxrpl include, there's a reasonable expectation that it defines the rippled API.

Also, these names belong to API version 2, even if unused. So removing them is an API breaking change. I'm happy to compromise to rename `rpcREPORTING_UNSUPPORTED` to a more generic name (keeping the error code 91).

godexsoft: Renaming the `rpcREPORTING_UNSUPPORTED` is an option but i suppose we want to keep the associated error strings as they are to avoid breaking the API.. in that case the output of Clio will remain with 'reportingMode' in the output. We can live with that. Hopefully can be properly addressed with OpenAPI later.

cfg/rippled-example.cfg
```diff
@@ -883,133 +883,14 @@
...
-#   the gRPC section, in order to bypass the server's rate limiting.
-#   This section needs to be added to the config of the ETL source, not
-#   the config of the reporting node. In the example below, the
-#   reporting server is running at 127.0.0.1. Multiple IPs can be
-#   specified in secure_gateway via a comma separated list.
-#
-#   [port_grpc]
-#   ip = 0.0.0.0
-#   port = 50051
-#   secure_gateway = 127.0.0.1
```
thejohnfreeman, line 994: @godexsoft should this stanza be left in, but changed to refer to Clio instead of "reporting node"?

src/xrpld/app/main/GRPCServer.cpp
thejohnfreeman: Clio is the client.

With reporting mode, the configuration looks like this:

non-reporting rippled <--> reporting rippled <--> app

In this scenario, when the app sends a request to the reporting server, and it doesn't have that information, it forwards the request to the non-reporting server. In that forwarded request, the reporting server is the client, and the app is the proxied client.

If Clio is not attaching the correct field (`client_ip`) to indicate it is forwarding a request on behalf of a proxied client, then it will never be throttled as long as its address (the client address) is in the list of secure gateways.

(Yes, this is confusing. Clio is the "client" in this code block, but the message's `client_ip` field names the "proxied client".)

cfg/rippled-example.cfg
godexsoft: I think it would be helpful for operators setting up their ETL nodes, yes 👍

src/xrpld/app/main/GRPCServer.cpp
godexsoft: For forwarding to `rippled` what we do with the real client's IP is 
```cpp
if (forwardToRippledClientIp) {
    connectionBuilder.addHeader(
        {boost::beast::http::field::forwarded, fmt::format("for={}", *forwardToRippledClientIp)}
    );
}
```
and we also set the `X-User` to either `clio_admin` or `clio_user` depending whether the connection is an `admin` one (in Clio terms).
However this has nothing to do with gRPC.

For the gRPC connection we don't set the `client_ip` field but we do set the `user` to `ETL` (probably does nothing).
What happens if you don't add Clio's IP to the list of secure gateways, Clio itself gets throttled. Makes sense.
Note that Clio is not doing any gRPC forwarding for Clio's clients, gRPC is only used by Clio to get the data for Clio's use.

include/xrpl/protocol/ErrorCodes.h
thejohnfreeman: Ok. I'll restore these two values and add a comment.

thejohnfreeman: On further research, it appears that Clio uses only `rpcREPORTING_UNSUPPORTED`, so I'll restore only that one.

seelabs (approved):
:+1:

Issue #5036, opened 2024-06-03
book_changes takes a long time querying full-history servers


mDuo13:
## Issue Description
When calling the `book_changes` method using WebSocket or JSON-RPC, asking for data from older ledgers leads to a long delay or even a request timeout.

## Steps to Reproduce
1. Connect to a full-history rippled server via WebSocket
2. Send the following request:
3. 
```json
{
  "command": "book_changes",
  "ledger_index": 72381103
}
```

## Expected Result
Server should respond quickly with a value like this:

```json
{
  "result": {
    "type": "bookChanges",
    "ledger_hash": "33C53CF838DEE3C1F74E0BA8BEDEE28D2527E11B1C53846A22E9DBB33EC048C2",
    "ledger_index": 72381103,
    "ledger_time": 708708711,
    "validated": true,
    "changes": [
      {
        "currency_a": "XRP_drops",
        "currency_b": "rLqUC2eCPohYvJCEBJ77eCCqVL2uEiczjA/5852646F67650000000000000000000000000000",
        "volume_a": "520687381",
        "volume_b": "8568683",
        "high": "60.76767128172012",
        "low": "60.76159531559149",
        "open": "60.76159531559149",
        "close": "60.76767128172012"
      },
      {
        "currency_a": "XRP_drops",
        "currency_b": "rhLr8bGvHvBgYXAHNPyXrQAcKGrQ2X5nU4/5854726976694100000000000000000000000000",
        "volume_a": "3852",
        "volume_b": "4950495.0495049",
        "high": "0.000778104000000008",
        "low": "0.000778104000000008",
        "open": "0.000778104000000008",
        "close": "0.000778104000000008"
      },
      {
        "currency_a": "rKWH2NPtXn636uodWJ3dLLM7wHmyG2qpz6/KWH",
        "currency_b": "rTGoNeK6vpu28U2oE5tiqhH3x8z2jBhxv/TGO",
        "volume_a": "0.8",
        "volume_b": "8",
        "high": "0.1",
        "low": "0.1",
        "open": "0.1",
        "close": "0.1"
      }
    ]
  },
  "status": "success",
  "type": "response"
}
```

## Actual Result
Server takes a long time (>30s) to respond if it responds at all. On xrplcluster.com this results in the gateway returning a 504 error.

For more recent ledgers, the server responds quickly with a valid response, so this has something to do with fetching older data specifically. It's unclear how old a ledger you have to ask for before it becomes a problem.

## Environment
[...]

PR #5096, opened 2024-08-10
updates to book_changes RPC command

### LINKED ISSUE #5033: Invalid "success" response from book_changes (possible race condition)
## Issue Description
When querying for a recent ledger by index, sometimes the result doesn't match the expected format.

## Steps to Reproduce
Make WebSocket API requests such as the following, replacing the ledger index with the most recent one:

```json
{
  "id": "example_book_changes",
  "command": "book_changes",
  "ledger_index": 88376663
}
```

The problem is intermittent, so it may take some tries to reproduce. Be sure to keep updating the ledger index with the most recent one as new ledgers are closed.

## Expected Result
The API should either return a response in the expected format or an error message.

**Example valid response (with non-empty changes):**

```json
{
  "id": "example_book_changes",
  "result": {
    "changes": [
      {
        "close": "1303424095097822e-4",
        "currency_a": "XRP_drops",
        "currency_b": "rchGBxcD1A1C2tdxF6papQYZ8kjRKMYcL/BTC",
        "high": "1304070002477734e-4",
        "low": "1302405543037992e-4",
        "open": "1303084400776639e-4",
        "volume_a": "60000000",
        "volume_b": "0.000460371"
      }
    ],
    "ledger_hash": "45765B07A378A94FCB6CEDD3D498E0CC15C73468F5F15AAA779079E67A234EB9",
    "ledger_index": 88376645,
    "ledger_time": 770512422,
    "type": "bookChanges"
  },
  "status": "success",
  "type": "response"
}
```

**Example expected error response if the ledger index is too high:**

```json
{
  "error": "invalidParams",
  "error_code": 31,
  "error_message": "Ledger index too large",
  "id": "example_book_changes",
  "request": {
    "command": "book_changes",
    "id": "example_book_changes",
    "ledger_index": 88376663
  },
  "status": "error",
  "type": "response"
}
```

## Actual Result

In some cases the response claims to be a success but it contains the wrong data, apparently some sort of status update on the status of acquiring the requested ledger. Looks like a [Ledger Request Object](https://xrpl.org/docs/references/http-websocket-apis/admin-api-methods/logging-and-data-management-methods/ledger_request#ledger-request-object).

```json
{
  "id": "example_book_changes",
  "result": {
    "hash": "63C34CA1EE6CD80B89A035E077A0266BF2AF8F4A5CC729D1B595940D304BC05A",
    "have_header": true,
    "have_state": false,
    "have_transactions": false,
    "needed_state_hashes": [],
    "needed_transaction_hashes": [],
    "peers": 8,
    "timeouts": 1
  },
  "status": "success",
  "type": "response"
}
```

## Environment

Tested using the xrplcluster.com servers running 2.2.0-rc1, but I think this API hasn't been changed much if at all since #4212 two years ago so the problems probably exist in older versions.

The s1.ripple.com cluster mostly goes to Clio servers; I haven't been able to reproduce on there.

ckeshava:

Updates to the `book_changes` RPC command.
Pending clarification, I believe this PR also fixes #5033  and #5036 (improves the latency of the command)

### Context of Change
1. accepts shortcut strings (`current, closed, validated`) as inputs to `ledger_index` parameter in the RPC command.
2. returns `validated: true/false` in the RPC output

### Type of Change

- [ ] Documentation update
- [ ] Release

### API Impact


src/xrpld/rpc/BookChanges.h
Bronek, line 1: Can you pls do a drive-by fix and include the missing headers ?

```
#include <xrpl/json/json_value.h>
#include <xrpl/protocol/LedgerFormats.h>
#include <xrpl/protocol/STAmount.h>
#include <xrpl/protocol/STObject.h>
#include <xrpl/protocol/TxFormats.h>
#include <xrpl/protocol/jss.h>

#include <memory>
```

ckeshava: I have specified these include's in a8a21bfe4

Did you obtain these `#include`'s based on any compiler's warning? Is this based on visual observation or any compiler special flag/setting?

ckeshava: How do you use a unit test file "on its own"? I usually do `cmake --build .` inside the build directory. It automatically re-compiles the modified-source-code files.

Bronek: A unit test for a header file on its own will have one `#include` , only for the header file being tested. Aside from that it can only `#include` files related to the unit test framework being used, but it must not include any other project files. If you have a test like that, and it fails to compile, then that's an indication that the file being tested is missing includes.

There are tools which do the equivalent testing for you, without actually creating any additional files in the project, for example `clangd` - that's how I find such errors.

ckeshava: thanks for the explanation, that seems reasonable to me. I've removed some of the extraneous include's from BookChanges_test in a recent commit

Bronek:
@godexsoft you might want to test this branch with clio.

Bronek (approved):
I love that it gets rid of `getLedgerByContext` call in a USER context, also the newly added unit tests seem to bump test coverage overall. Nice work !

ckeshava:
@Bronek  thanks for the quick review!

What do you mean by USER context? I feel the `getLedgerByContext` should only be used by the consensus mechanism (and Full History Nodes) to collect+store all ledgers. Its in-efficient for a `tiny` or `medium` resources validator to engage in acquisition of ledgers.

The `getLedgerByContext` method can also be removed in the `ledger_request` RPC call. I expect to see similar issues in that RPC call. Check my comment here: https://github.com/XRPLF/rippled/issues/5036#issuecomment-2279269503

src/xrpld/rpc/handlers/BookOffers.cpp
```diff
@@ -204,13 +204,13 @@ doBookOffers(RPC::JsonContext& context)
 Json::Value
 doBookChanges(RPC::JsonContext& context)
 {
-    auto res = RPC::getLedgerByContext(context);
+    std::shared_ptr<ReadView const> ledger;
 
-    if (std::holds_alternative<Json::Value>(res))
-        return std::get<Json::Value>(res);
+    Json::Value result = RPC::lookupLedger(ledger, context);
```
godexsoft, line 209: 1. Does this change still yield the same error from checking ledger_index etc.?
2. Can we do the same in `LedgerRequest.cpp` and eliminate getLedgerByContext() entirely?

godexsoft (commented):
I don't see anything that could impact Clio. 
My only concern is whether the new code still does the same error handling as the old did.

src/xrpld/rpc/handlers/BookOffers.cpp
Bronek: 1. @godexsoft the `ledger_index` handling (and errors) will be the same as in almost every other command, since the vast majority of commands use `lookupLedger`
2. No; the ADMIN method `ledger_request` which, with this PR, will be the only method using `getLedgerByContext()`, and it  needs to rely on this function in order to download ledgers from the other nodes.

godexsoft (approved):

Bronek (changes requested):
@ckeshava please also update API-CHANGELOG.md for release 2.3 . You need to add section for release 2.3 and list the changes in behaviour of `book_changes` (basically making it like all other methods, but also will now return `ledgerNotFound` if the ledger is very old). I suggest asking @mDuo13 for help.

API-CHANGELOG.md
```diff
@@ -28,6 +28,11 @@ The `network_id` field was added in the `server_info` response in version 1.5.0
 
 ## XRP Ledger server version 2.0.0
 
+### Updates in 2.3.0
+
+(Breaking Changes)
+- Changes to the `book_changes` RPC: This RPC now returns a `validated` field in its response, which was missing in prior versions. Additionally, `book_changes` RPC exhibits similar behavior as the other non-Admin RPC commands, insofar as ledger-retrieval is concerned. Specifically, if the requested ledger version is old, then a `ledgerNotFound` error is returned. This is a breaking change, because prior versions would attempt to acquire such an old ledger.
```
ckeshava, line 34: @mDuo13  @Bronek  If need be, please suggest modifications to this message.

Bronek (approved):

API-CHANGELOG.md
```diff
@@ -28,6 +28,11 @@ The `network_id` field was added in the `server_info` response in version 1.5.0
...
+
```
intelliot, line 35:
```suggestion
### Breaking change in 2.3

- `book_changes`: If the requested ledger version is not available on this node, a `ledgerNotFound` error is returned and the node does not attempt to acquire the ledger from the p2p network (as with other non-admin RPCs).

Admins can still attempt to retrieve old ledgers with the `ledger_request` RPC.

### Addition in 2.3

- `book_changes`: Returns a `validated` field in its response, which was missing in prior versions.

```

intelliot (approved):
Suggested a rewording to the changelog for consistency. I also think it's ok to merge as-is and improve the docs later on (which is easy to do).

PR #5106, opened 2024-08-26
Pr/2.2.2 rc2

ximinez:
### Type of Change

- [ ] Documentation update
- [ ] Release

### API Impact

### DIFF
--- a/src/ripple/app/ledger/InboundLedgers.h
+++ b/src/ripple/app/ledger/InboundLedgers.h
@@ -42,6 +42,12 @@ public:
     virtual std::shared_ptr<Ledger const>
     acquire(uint256 const& hash, std::uint32_t seq, InboundLedger::Reason) = 0;
 
+    virtual void
+    acquireAsync(
+        uint256 const& hash,
+        std::uint32_t seq,
+        InboundLedger::Reason reason) = 0;
+
     virtual std::shared_ptr<InboundLedger>
     find(LedgerHash const& hash) = 0;
 
--- a/src/ripple/app/ledger/InboundLedgers.h
+++ b/src/ripple/app/ledger/InboundLedgers.h
@@ -23,6 +23,7 @@
 #include <ripple/app/ledger/InboundLedger.h>
 #include <ripple/protocol/RippleLedgerHash.h>
 #include <memory>
+#include <set>
 
 namespace ripple {
 
@@ -37,11 +38,14 @@ public:
 
     virtual ~InboundLedgers() = default;
 
-    // VFALCO TODO Should this be called findOrAdd ?
-    //
+    // Callers should use this if they possibly need an authoritative
+    // response immediately.
     virtual std::shared_ptr<Ledger const>
     acquire(uint256 const& hash, std::uint32_t seq, InboundLedger::Reason) = 0;
 
+    // Callers should use this if they are known to be executing on the Job
+    // Queue. TODO review whether all callers of acquire() can use this
+    // instead. Inbound ledger acquisition is asynchronous anyway.
     virtual void
     acquireAsync(
         uint256 const& hash,
--- a/src/ripple/app/ledger/impl/InboundLedgers.cpp
+++ b/src/ripple/app/ledger/impl/InboundLedgers.cpp
@@ -29,6 +29,7 @@
 #include <ripple/core/JobQueue.h>
 #include <ripple/nodestore/DatabaseShard.h>
 #include <ripple/protocol/jss.h>
+#include <exception>
 #include <memory>
 #include <mutex>
 #include <vector>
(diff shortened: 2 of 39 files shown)


src/ripple/app/ledger/impl/InboundLedgers.cpp
```diff
@@ -69,76 +70,99 @@ class InboundLedgersImp : public InboundLedgers
...
+    }
+
+    void
+    acquireAsync(
+        uint256 const& hash,
+        std::uint32_t seq,
+        InboundLedger::Reason reason) override
+    {
+        std::unique_lock lock(acquiresMutex_);
+        if (pendingAcquires_.contains(hash))
```
vlntb, line 159: In the context of `acquireLedger` we use ledger ID and ledger hash interchangeably. Is it possible that in the case of using ID, we might be skipping on ledger update that is different from the one that already applied if we use ID as a differentiator?
i.e. `adaptor_.acquireLedger(prevLedgerID) in Consensus.h`.

src/ripple/app/ledger/impl/InboundLedgers.cpp
```diff
@@ -69,76 +70,99 @@ class InboundLedgersImp : public InboundLedgers
...
+            return;
```
vlntb, line 160: Can we add a debug trace here to check what is being skipped?

ximinez: There are already log messages at the caller sites. Is that sufficient?

src/ripple/app/ledger/impl/InboundLedgers.cpp
mtrippled: ledger ID and ledger hash are the same thing.

src/ripple/app/misc/NetworkOPs.cpp
```diff
@@ -2307,7 +2310,15 @@ NetworkOPsImp::recvValidation(
     JLOG(m_journal.trace())
         << "recvValidation " << val->getLedgerHash() << " from " << source;
 
+    std::unique_lock lock(validationsMutex_);
+    if (pendingValidations_.contains(val->getLedgerHash()))
+        return false;
+    pendingValidations_.insert(val->getLedgerHash());
+    lock.unlock();
     handleNewValidation(app_, val, source);
```
bachase, line 2320: @mtrippled, my understanding is this is meant to prevent later on the calls to [`checkAccept`](
https://github.com/ximinez/rippled/blob/098546f0b1ba259d144e3e60fc9b3274f85e5d05/src/ripple/app/consensus/RCLValidations.cpp#L189) in`handleNewValidation` that may end up calling starting of a chain of `InboundLedger` requests here https://github.com/ximinez/rippled/blob/098546f0b1ba259d144e3e60fc9b3274f85e5d05/src/ripple/app/ledger/impl/LedgerMaster.cpp#L1070-L1071. 

Asking to understand if you instead considered change the call at this `LedgerMaster` callsite to use the new `acquireAsync` method you added, and if so, why you went this way? I'm guessing you are trying to filter out redundant validations at the highest level?

src/ripple/app/ledger/impl/InboundLedgers.cpp
```diff
@@ -69,76 +70,99 @@ class InboundLedgersImp : public InboundLedgers
...
+        pendingAcquires_.insert(hash);
+        lock.unlock();
```
bachase, line 162: Nit: for this and the similar callsite for validations --- should we use RAII/scoped locks vs. this explicit lock and unlock? I get it doesn't matter now, but if code drifts over time, that might be a stronger sign to the future code writer to not break this.

src/ripple/app/ledger/InboundLedgers.h
```diff
@@ -42,6 +42,12 @@ class InboundLedgers
     virtual std::shared_ptr<Ledger const>
     acquire(uint256 const& hash, std::uint32_t seq, InboundLedger::Reason) = 0;
 
+    virtual void
+    acquireAsync(
```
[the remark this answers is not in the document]
mtrippled: @mtrippled Yes, I can add comments, though I'm not sure if this is in a "freeze" state yet, but can add comments. I made the new function for things to be submitted via the job queue.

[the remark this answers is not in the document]
bachase: Its also ok if that goes when we fold this into the mainline versus hotfix.

src/ripple/app/misc/NetworkOPs.cpp
```diff
@@ -2311,13 +2311,14 @@ NetworkOPsImp::recvValidation(
         << "recvValidation " << val->getLedgerHash() << " from " << source;
 
     std::unique_lock lock(validationsMutex_);
-    if (pendingValidations_.contains(val->getLedgerHash()))
+    if (pendingValidations_.contains(
```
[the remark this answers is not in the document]
mtrippled: @bachase I don't know. Also, I don't know how many validations saturating the job queue are duplicates from the same validator vs all the validations coming from distinct validators. In the latter case, none would be suppressed. I don't think the sequence number is useful if the hashes are distinct--also, I think some place else checks that the sequence number matches the ledger we're trying to validate. I don't know anything about the validation cookie either.

src/ripple/app/ledger/impl/InboundLedgers.cpp
vlntb: We discussed this offline with Mark and I don't have any reservations about ledger identity.

ximinez:
I'm closing this PR in favor of #5115, which is ready to merge.

PR #5112, opened 2024-08-28
Implement threshold for the intersection of lists of validators

Bronek:

Improve UNL security by allowing validators to set a minimum number of UNL publishers to agree on validators

### Context of Change

In the current system, every node builds an internal list of validators to use, by performing **union** of UNLs received from all configured UNL publishers (validator lists), where the authenticity of each single UNL is verified by a manifest with a known master key.

This **union** is a weak security spot of the UNL publication system, as it allows a single UNL publisher to add an arbitrary number of validators, which will be subsequently used by all nodes trusting this UNL publisher. Currently the risk is managed by recommending only the use of two UNL publishers: vl.xrplf.org and vl.ripple.com . With this union approach, a hypothetical addition of a third (or more) UNL publisher would only increase the attack surface, in this way (counterintuitively) degrading the UNL system security.

This PR is replacing union with a configurable **intersection threshold** between UNL publishers, which by default is calculated as:

```
threshold =  size(validator_list_keys) < 3
  ? 1
  : floor(size(validator_list_keys) / 2) + 1
```

This means that for a default 2 UNL publishers, the default threshold will be 1. Since "intersection of 1 sets" is not an intersection (it's an union) this makes the default behaviour (for one or two UNL publishers) identical to the current union system.

If a node configuration contains a third UNL publisher, the default threshold will be calculated as `floor(3 / 2) + 1`, that is `1 + 1` that is 2. This means that, for 3 validator lists, the node will **only** use validators which are present on 2 or more lists (and will silently ignore validators which are on one list only).

If the configured validator list contains 4 or 5 publishers, the default threshold will be 3. If the configured validator list contains 6 or 7 publishers, the default threshold will be 4 etc.

This PR adds an optional configuration option `[validator_list_threshold]` to `validators.txt` which can  be explicitly set to the minimum number of the lists on which a validator must be listed in order to be used (this number must not be greater than the size of `[validator_list_keys]`). If it is not set, or set to 0, the value will be calculated at startup from the size of `[validator_list_keys]` as explained above.

The actually used threshold value will be also displayed in the output of [validators](https://xrpl.org/docs/references/http-websocket-apis/admin-api-methods/status-and-debugging-methods/validators) RPC method (alongside with the used list of validators, and the content of each configured validator list, as is the current output).

This change is **not** altering the list of the default UNL publishers. The intent of this PR is to make it safe to extend this list in the future.

---

The security problems of the **union** UNL publication system were originally described by @mDuo13 and the idea to use intersection of validator lists as a solution came from @ximinez 

### Type of Change

- [ ] Documentation update
- [ ] Release

### API Impact

Add new value `validator_list_threshold` to the output from [validators](https://xrpl.org/docs/references/http-websocket-apis/admin-api-methods/status-and-debugging-methods/validators) RPC method

Add new (optional) section `[validator_list_threshold]` to validators file.

- [x] Public API: New feature (new methods and/or new fields)


src/xrpld/app/misc/detail/ValidatorList.cpp
```diff
@@ -190,6 +192,33 @@ ValidatorList::load(
         ++count;
     }
 
+    std::size_t const validPublishers = std::count_if(
+        publisherLists_.begin(),
+        publisherLists_.end(),
+        [](std::pair<PublicKey, PublisherListCollection> const& p) -> bool {
+            return p.second.status != PublisherStatus::revoked;
```
Bronek, line 199: I guess should also check for `expired`

src/test/app/ValidatorList_test.cpp
```diff
@@ -1593,6 +1673,164 @@ class ValidatorList_test : public beast::unit_test::suite
...
+                activeValidators,
+                env.timeKeeper().now(),
+                env.app().getOPs(),
+                env.app().overlay(),
+                env.app().getHashRouter());
+
+            BEAST_EXPECT(
+                trustedKeys->quorum() ==
+                std::numeric_limits<std::size_t>::max());
+            // TODO This is safer; above is not good
```
Bronek, line 1807: This is referring to check in `ValidatorList::calculateQuorum`

src/xrpld/app/misc/detail/ValidatorList.cpp
```diff
@@ -190,6 +192,34 @@ ValidatorList::load(
...
+    {
+        listThreshold_ = *listThreshold;
+        // This should be enforced by Config class
+        assert(listThreshold_ > 0 && listThreshold_ <= publisherLists_.size());
+        if (listThreshold_ > validPublishers)
+        {
+            JLOG(j_.fatal()) << "Validator list threshold set to "
+                             << listThreshold_ << " however there are only "
+                             << validPublishers << " valid publishers";
+            return false;
```
Bronek, line 212: Consider downgrading this to "just log an error and carry on".

src/xrpld/app/misc/detail/ValidatorList.cpp
```diff
@@ -190,6 +192,34 @@ ValidatorList::load(
...
+        }
+    }
+    else
+    {
+        // Want truncated result when dividing an odd integer
+        listThreshold_ = (validPublishers >= 3) ? (validPublishers / 2 + 1) : 1;
```
Bronek, line 218: This is wrong

src/xrpld/app/misc/detail/ValidatorList.cpp
```diff
@@ -1749,11 +1779,38 @@ ValidatorList::calculateQuorum(
...
+        // There are two, subtly different, sides to list threshold
+        //
+        // 1. The minimum required intersection between lists listThreshold_
+        //    for a validator to be included in trustedMasterKeys_.
+        //    If this many (or more) publishers are unavailable, we are likely
+        //    to NOT include a validator which otherwise would have been used.
+        // 2. The minimum number of lists which, when UNAVAILABLE, will prevent
+        //    us from hitting the above threshold on ANY validator. This is
+        //    calculated as:
+        //      N - M + 1
```
bachase, line 1801: To aid in my review, can you provide a bit more detail on this view @Bronek ? I'm not following the distinction between case 1 and case 2 here. I think case #2 is saying that until we have at least the `listThreshold_` number of distinct* published lists available, then nothing can hit the threshold because there is simply not enough lists.

I'm not clear on #2. Is there an example you could work through in a GH comment?

* - distinct means having a different publisher key versus a distinct URL/domain.

bachase (commented):
Mostly style/nits so far. Will do another pass after I better understand how the unavailable lists is being considered.

src/xrpld/app/misc/ValidatorList.h
```diff
@@ -343,7 +346,8 @@ class ValidatorList
     load(
         std::optional<PublicKey> const& localSigningKey,
         std::vector<std::string> const& configKeys,
-        std::vector<std::string> const& publisherKeys);
+        std::vector<std::string> const& publisherKeys,
+        std::optional<std::size_t> listThreshold = {});
```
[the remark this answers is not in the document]
Bronek: Good question, and admittedly I am not 100% certain that this is the right way, so I will present both arguments for and against making a change here (to remove the default)

1. for - in `rippled` proper (as opposed to unit tests), we call this function in one place only `ApplicationImp::setup` and it is important that, in this one location, we definitely **do** provide this parameter, as parsed from the configuration.

2. against - we do recommend using default in configuration, for a good reason. It makes sense to follow this advice, in the most concise way possible (that is, not providing the parameter at all) also in unit tests etc. where we call this function outside of `rippled` proper

3. against - making it a required parameter would cause relatively large churn in unit tests

```
[feature/validator_list_threshold] % git diff --stat
 src/test/app/Manifest_test.cpp          |  2 +-
 src/test/app/ValidatorList_test.cpp     | 66 +++++++++++++++++++++++++++++++++++++-----------------------------
 src/test/app/ValidatorSite_test.cpp     |  2 +-
 src/test/consensus/NegativeUNL_test.cpp |  2 +-
 src/xrpld/app/misc/ValidatorList.h      |  2 +-
 5 files changed, 41 insertions(+), 33 deletions(-)
```

Again, I am not 100% convinced that points 2. and 3. do outweigh point 1. , and (since it's not much work) if you think that 1. (or some other argument) dictates to remove the default value, I'd be happy to do it.

src/test/core/Config_test.cpp
```diff
@@ -711,6 +863,38 @@ trustthesevalidators.gov
...
             BEAST_EXPECT(
                 c.section(SECTION_VALIDATOR_LIST_KEYS).values().size() == 3);
+            BEAST_EXPECT(
+                c.section(SECTION_VALIDATOR_LIST_THRESHOLD).values().size() ==
+                1);
+            BEAST_EXPECT(c.VALIDATOR_LIST_THRESHOLD == 2);
+        }
+        {
+            // load should throw if [validator_list_threshold] is present both
+            // in rippled cfg and validators file
```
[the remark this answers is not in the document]
Bronek: Yes. The test is loading validators file (which contains [validator_list_threshold], see line 228) and also loads [validator_list_threshold] section from the config file itself (right below this comment), which results in two values being loaded into `Config` (which is standard processing of validators file, I think in this case consistency is a virtue). This is why we expect an exception thrown, which will set the error.

src/xrpld/app/misc/detail/ValidatorList.cpp
Bronek: I will welcome suggestions how to formulate this better !

1. refers to the situation when, out of N configured lists, we are missing M or more. In this situation we will not be using validators which these lists might have provided us with. This retains the original intent of this section, when 1 was logically a hardcoded "errorThreshold" (an "intersection of 1" is just an union)

2. refers to situation - as you correctly recognized - when we are missing so many lists that it is impossible for any validator to meed the required threshold M (any, except _self_). This prevents an occurrence when we would calculate a quorum of 1, making this _self_ validator its own 100% quorum. In the original logic it is not possible, and I wanted to avoid introducing such behaviour here. Also see unit test change in commit https://github.com/XRPLF/rippled/pull/5112/commits/57b451cab070f696ef72ba1a1d9a1e2604f991b3

Bronek:
> Good catch, however:
> 
> The use of `Union` of X sets is still valid, however as stated by OP, an attack surface opens up with a malicious publisher. There may be cases where an operator would want to use the Union of X sets instead of the intersect. I recommend allowing the use of both `Union` and `Intersect`, where `Intersect` is the default as to the clear benefits of it.

The use of union is definitely allowed and this PR won't change it, it's just not the default. The user only needs to configure the "intersection of 1", which is functionally equivalent to union.

```
[validator_list_threshold]
1
```

> > The intent of this PR is to make it safe to extend this list in the future.
> 
> Agreed, this would welcome an influx of (reputable) UNL publishers to greater network relevance. I must argue that standards are still to be applied with the addition of new(er) publishers to the default lists ([validators-example.txt](https://github.com/XRPLF/rippled/blob/develop/cfg/validators-example.txt)). Maintainers and trustees of rippled should not add any publishers or author to this list unless:
> 
> * a proven track record **on the network** is present and backed.
> * author is recognized, present, relevant and of sound mind within the "community" of network users and network operators.
> 
> This so called standard should not be ambiguous or overreaching in its limited scope. The consequence of such will result in extreme ends.

Agree, but this is outside of the scope of this PR.

src/xrpld/app/misc/ValidatorList.h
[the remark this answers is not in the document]
bachase: I don't have a super strong point of view. Trust your judgement (or defer to any other reviewers)

src/xrpld/app/misc/detail/ValidatorList.cpp
bachase: Got it. I think this formulation works. The only alternative I can see is framing it in terms of available rather than unavailable number of lists, but thats not so different. 

On second read, I believe the "Ideally we want both to be equal ..." bit is what threw me off the most; maybe just drop that? But that's just me.

bachase (approved):
👍 Great job with the unit tests and the careful thought into how to incrementally improve UNL management.

src/xrpld/app/misc/detail/ValidatorList.cpp
Bronek: I updated this comment, hopefully it is clearer now.

bachase: That change does help. 👍

src/test/app/ValidatorList_test.cpp
```diff
@@ -452,15 +479,57 @@ class ValidatorList_test : public beast::unit_test::suite
...
+            //  -- thus should not be loaded
+            pubManifests.applyManifest(*deserializeManifest(makeManifestString(
+                pubRevokedPublic,
+                pubRevokedSecret,
+                pubRevokedSigning.first,
+                pubRevokedSigning.second,
+                std::numeric_limits<std::uint32_t>::max())));
+
+            // these two are not revoked (and not in the manifest cache at all.)
             auto legitKey = randomMasterKey();
```
ximinez, line 522:
```suggestion
            // this one is not revoked (and not in the manifest cache at all.)
            auto legitKey = randomMasterKey();
```

src/test/app/ValidatorList_test.cpp
```diff
@@ -1550,16 +1630,18 @@ class ValidatorList_test : public beast::unit_test::suite
                     {strHex(publisherPublic)});
                 std::vector<std::string> emptyCfgKeys;
 
-                BEAST_EXPECT(
-                    trustedKeys->load({}, emptyCfgKeys, cfgPublishers));
+                BEAST_EXPECT(trustedKeys->load(
+                    {}, emptyCfgKeys, cfgPublishers, std::size_t(1)));
```
ximinez, line 1634:
```suggestion
                // Threshold of 1 will result in a union of all the lists
                BEAST_EXPECT(trustedKeys->load(
                    {}, emptyCfgKeys, cfgPublishers, std::size_t(1)));
```

src/test/app/ValidatorList_test.cpp
```diff
@@ -2390,6 +2658,1151 @@ class ValidatorList_test : public beast::unit_test::suite
...
+
+            hash_set<NodeID> removed;
+            BEAST_EXPECT(trustedKeys->trusted(self.masterPublic));
+            for (auto const& val : valKeys)
+            {
+                BEAST_EXPECT(trustedKeys->listed(val.masterPublic));
+                BEAST_EXPECT(!trustedKeys->trusted(val.masterPublic));
+                removed.insert(calcNodeID(val.masterPublic));
+            }
+            BEAST_EXPECT(trustedKeys->trusted(self.masterPublic));
```
ximinez, line 2836: You're checking that `self.masterPublic` is trusted twice. Is that intentional?

src/xrpld/app/misc/detail/ValidatorList.cpp
```diff
@@ -190,14 +192,42 @@ ValidatorList::load(
         ++count;
     }
 
+    if (listThreshold)
+    {
+        listThreshold_ = *listThreshold;
+        // This should be enforced by Config class
+        assert(listThreshold_ > 0 && listThreshold_ <= publisherLists_.size());
```
ximinez, line 199: Should this be an `XRPL_ASSERT`?

src/xrpld/app/misc/detail/ValidatorList.cpp
```diff
@@ -1794,11 +1826,40 @@ ValidatorList::calculateQuorum(
...
+        //      M: minimum required intersection i.e. listThreshold_
+        //    If this happens, we still have this local validator and we do not
+        //    want it to form a quorum of 1, so we disable quorum as well.
+        //
+        // We disable quorum if the number of unavailable publishers exceeds
+        // either of the above thresholds
+        auto const errorThreshold = std::min(
+            listThreshold_,  //
+            publisherLists_.size() - listThreshold_ + 1);
+        assert(errorThreshold > 0);
```
ximinez, line 1861: Should this be an `XRPL_ASSERT`?

ximinez (changes requested):
Looks good overall. I just left a few small suggestions, questions, and comments.

src/test/app/ValidatorList_test.cpp
Bronek: no, just too much copy&paste

src/xrpld/app/misc/detail/ValidatorList.cpp
```diff
@@ -1943,20 +2004,25 @@ ValidatorList::updateTrusted(
...
+        if (kit == keyListings_.end() ||     //
+            kit->second < listThreshold_ ||  //
+            validatorManifests_.revoked(*it))
         {
             trustChanges.removed.insert(calcNodeID(*it));
             it = trustedMasterKeys_.erase(it);
         }
         else
         {
+            assert(kit->second > 0);
```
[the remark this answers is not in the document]
Bronek: Yes, thanks. Missed it after merge from `develop`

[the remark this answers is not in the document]
Bronek: As for the actual assert condition, that's a good idea.

src/xrpld/app/misc/detail/ValidatorList.cpp
Bronek: Yes, thanks.

ximinez (approved):
Looks good!

Bronek:
Proposed merge message

```
Add [validator_list_threshold] to validators.txt to improve UNL security
```

src/test/core/Config_test.cpp
```diff
@@ -222,6 +223,9 @@ moreripplevalidators.net
 [validator_list_keys]
 03E74EE14CB525AFBB9F1B7D86CD58ECC4B91452294B42AB4E78F260BD905C091D
 030775A669685BD6ABCEBD80385921C7851783D991A8055FD21D2F3966C96F1B56
+
+[validator_list_threshold]
+2
```
bthomee, line 228: I'm missing two tests:
1. A validator list threshold larger than 3, to test `publisherLists_.size() / 2 + 1`.
2. A validator list threshold greater than the number of provided lists.
Would you be able to add those?

Bronek: 1. the calculation is inside `ValidatorList.cpp` so the test is in `ValidatorList_test.cpp` lines 494 (for two publishers) and 3022, 3090, 3158 (for five publishers). Do you want me to add test for three publishers as well ?
2. it is tested at `Config_test.cpp` line 647

Issue #5102, opened 2024-08-23
SQLite3 transaction.db out of Space on Full History Servers (Version: all)


jscottbranson:
## Issue Description
rippled relies on SQLite3 for the `transaction.db`.

SQLite3 databases have limitations (`max_page_count` and `page_size`) to prevent databases from growing excessively.

`page_size` defaults to 4096 bytes across all recent versions of SQLite3, and [rippled passes a "PRAGMA" statement to SQLite3](https://github.com/XRPLF/rippled/blob/d9bd75e68326861fb38fd5b27d47da1054a7fc3b/src/xrpld/app/main/DBInit.h#L79) when the connection is opened reinforcing this page size. The page_size is stored in the database itself, and [changing the page_size](https://www.sqlite.org/pragma.html#pragma_page_size) requires the database to be recreated. Recreating the database takes significant time as well as storage space, and I have not successfully created a transaction.db with larger page sizes, despite multiple attempts. There are also potential performance implications associated with changing page_size, and SQLite documentation suggests most applications will benefit from using the default 4096 byte setting.

In contrast, `max_page_count` seems to vary significantly across SQLite3 versions, and it is not stored in the database itself. Rather, max_page_count is [set via a PRAGMA statement](https://www.sqlite.org/pragma.html#pragma_max_page_count) when the SQLite3 connection is opened, or the default value for a given SQLite3 version is used. [rippled passes a PRAGMA statement](https://github.com/XRPLF/rippled/blob/d9bd75e68326861fb38fd5b27d47da1054a7fc3b/src/xrpld/app/main/DBInit.h#L80) to SQLite3 upon opening the database, defining `max_page_count = 2147483646` pages. With `page_size of 4096 bytes * max_page_count of 2147483646`, the maximum size for the transaction.db file is 8.796093 TB. Currently, my full history servers' databases are 8.3 TB. Thus, it is possible that all full history servers are approaching the point where the transaction.db file will be full. **_It seems urgent to remedy this, or risk all full history servers crashing over the next few months_**.

The most obvious solution seems to be increasing the value passed by rippled for max_page_count to the [current (2024) default](https://www.sqlite.org/limits.html) of `4294967294` pages (the SQLite [change log](https://www.sqlite.org/changes.html) indicates a prior default of `1073741824` pages). However, defining max_page_count via a PRAGMA statement is not always successful.

[...]

PR #5114, opened 2024-08-31
Update SQLite3 max_page_count to match current defaults

jscottbranson:
When rippled initiates a connection to SQLite3, rippled sends a "PRAGMA" statement defining the maximum number of pages allowed in the database. This commit updates the max_page_count so it is consistent with the default for newer versions of SQLite3. Increasing max_page_count is critical for keeping full history servers online.

### Context of Change
Fixes XRPLF/rippled#5102

### Type of Change

- [x] Performance (increase or change in throughput and/or latency)

## Future Tasks
Older versions of SQLite3 should be tested to ensure they can successfully use the larger max_page_count

### DIFF
--- a/src/ripple/app/main/DBInit.h
+++ b/src/ripple/app/main/DBInit.h
@@ -77,7 +77,7 @@ inline constexpr auto TxDBName{"transaction.db"};
 inline constexpr std::array<char const*, 4> TxDBPragma
 {
     "PRAGMA page_size=4096;", "PRAGMA journal_size_limit=1582080;",
-        "PRAGMA max_page_count=2147483646;",
+        "PRAGMA max_page_count=4294967294;",
 
 #if (ULONG_MAX > UINT_MAX) && !defined(NO_SQLITE_MMAP)
         "PRAGMA mmap_size=17179869184;"


dangell7:
These settings imo should be configurable imo. Also, updating the max_page_count might fix the issue today but we will not be able to update this again. We should be increasing the page_size for FH nodes.

vvysokikh1:
> These settings imo should be configurable imo. Also, updating the max_page_count might fix the issue today but we will not be able to update this again. We should be increasing the page_size for FH nodes.

We had a discussion like this a couple of days ago. We thought there's no point to make max_page_count configurable. But regarding page_size - yes, there's a plan to increase a default value and make it configurable.

jscottbranson:
@dangell7 this PR sets max_page_count to the highest value currently supported in SQLite3. I don't know if there would be a reason for someone to desire a lower value.

We could also increase page_size or make that configurable. AFAIK, vacuuming isn't necessary if page count is also increased, since the larger page size will be used moving forward. Vacuuming is still ideal to transition historic DB pages to the larger page size.

Sadly, I don't seem to have the resources to try vacuuming the DB.

interc0der:
Sqlite vacuuming requires an equal amount of unallocated space in order to cleanup a database file. For example, if an 8TB transaction.db file needs to be vacuumed, at least 8TB (probably more) of free space would be required to process the file.

For this reason, if there comes a time when developers need to vacuum any sqlite database files, it is advantageous to have a handful of smaller files compared to one large file.

Did we consider a solution that breaks up the transaction.db into smaller chucks? This would alleviate the max_page_count threshold and improve `auto_vacuum` performance.

dangell7 (approved):

intelliot (approved):

intelliot (approved):
All CI checks passed on 65ded6bc1d56989713fc3d78732d1ce9a58a1519. Merging to `master` for 2.2.3 hotfix.

PR #5115, opened 2024-08-31
Proposed version 2.2.2

ximinez:

This is a hotfix release that fixes a ongoing issue on Mainnet where a majority of UNL validators will stall during consensus processing due to lock contention, preventing ledgers from being validated for up to two minutes.

The base branch is `master`. This PR will be merged with `--ff-only` (not squashed or rebased, and not using the GitHub UI) to `master`, and reverse-merged to `develop`. **Do not use the Github UI to merge this PR**

---

This PR / branch is identical to #5106, which is currently being tested, with two exceptions:
* The version number is updated
* The commits are reorganized and "cleaned up" for release.
```
$ git fetch origin && git diff origin/pr/2.2.2-rc1 origin/master-next 
diff --git a/src/ripple/protocol/impl/BuildInfo.cpp b/src/ripple/protocol/impl/BuildInfo.cpp
index 44aef03533..0086824b7e 100644
--- a/src/ripple/protocol/impl/BuildInfo.cpp
+++ b/src/ripple/protocol/impl/BuildInfo.cpp
@@ -33,7 +33,7 @@ namespace BuildInfo {
 //  and follow the format described at http://semver.org/
 // clang-format off
-char const* const versionString = "2.2.2-rc2"
+char const* const versionString = "2.2.2"
 // clang-format on
 
 #if defined(DEBUG) || defined(SANITIZER)
```

### Type of Change

- [x] Release

### DIFF
--- a/src/ripple/core/DatabaseCon.h
+++ b/src/ripple/core/DatabaseCon.h
@@ -21,6 +21,7 @@
 #define RIPPLE_APP_DATA_DATABASECON_H_INCLUDED
 
 #include <ripple/app/main/DBInit.h>
+#include <ripple/basics/PerfLog.h>
 #include <ripple/core/Config.h>
 #include <ripple/core/SociDB.h>
 #include <boost/filesystem/path.hpp>
@@ -115,7 +116,8 @@ public:
         Setup const& setup,
         std::string const& dbName,
         std::array<char const*, N> const& pragma,
-        std::array<char const*, M> const& initSQL)
+        std::array<char const*, M> const& initSQL,
+        beast::Journal journal)
         // Use temporary files or regular DB files?
         : DatabaseCon(
               setup.standAlone && !setup.reporting &&
@@ -126,7 +128,8 @@ public:
                   : (setup.dataDir / dbName),
               setup.commonPragma(),
               pragma,
-              initSQL)
+              initSQL,
+              journal)
     {
     }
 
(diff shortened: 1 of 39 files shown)


src/ripple/core/DatabaseCon.h
```diff
@@ -177,7 +183,14 @@ class DatabaseCon
     LockedSociSession
     checkoutDb()
```
Bronek, line 184: I think this function should take a `beast::Journal j` parameter rather than adding data member `beast::Journal const j_` . Yes that means adding of this parameter everywhere, but it's better than giving `Config` an accessor to `Journal` and then storing that inside `DatabaseCon`. Also, that means the journal will be tied to where the `checkoutDb` is called rather than where `DatabaseCon` is initialised, which is good news for `doVacuumDB` (and perhaps other places as well)

Bronek: NVM, I am starting to think this is not viable. Will have to revisit this at some point later, with proper performance logging/monitoring (no journal).

Bronek (approved):
LGTM

mDuo13:
This PR should edit `RELEASENOTES.md` to add the new release.

The changes can be relatively small. The opening sentence of this PR is good for the overview.

ximinez:
> This PR should edit `RELEASENOTES.md` to add the new release.
> 
> The changes can be relatively small. The opening sentence of this PR is good for the overview.

I saw this after merging the PR. Do you think we'd be ok adding release notes when we reverse merge into `develop`?

WietseWind:
FYI @ximinez - this seems to make a difference on FH nodes @ xrplcluster - less transitions, less hiccups in ledger steram over websocket.

ximinez:
This is great news, @WietseWind ! Thanks for the update!

PR #5118, opened 2024-09-06
test: enable TxQ unit tests work with variable reference fee

vvysokikh1:

Fix TxQ unit tests to be able to process reference fee value other than 10.

### Context of Change

In preparation for potential reference fee change we would like to verify that fee change works as expected. The first step is to fix all unit tests to be able to work with different reference fee values.

### Type of Change

- [] Tests (you added tests for code that already exists, or your new feature included in this PR)

### API Impact

None

### Test Plan

Tested following reference fee values: `10, 20, 50, 100, 200, 500, 1000` by changing https://github.com/vvysokikh1/rippled/blob/2f432e812cb773048530ebfaf2e0e6def51e3cc2/src/test/jtx/impl/envconfig.cpp#L47


src/test/app/TxQ_test.cpp
```diff
@@ -250,147 +284,154 @@ class TxQPosNegFlows_test : public beast::unit_test::suite
...
+        env(noop(fred), fee(baseFee * 1.7), queued);
+        env(noop(gwen), fee(baseFee * 1.8), queued);
+        env(noop(hank), fee(baseFee * 1.9), queued);
+        env(noop(alice), fee(baseFee * 2.0), queued);
 
         // Queue is full now.
-        checkMetrics(__LINE__, env, 6, 6, 4, 3, 385);
-
+        // clang-format off
+        checkMetrics(__LINE__, env, 6, 6, 4, 3, txFeeLevelByAccount(env, daria) + 1);
```
bthomee, line 408: I'm trying to work out how `txFeeLevelByAccount(env, daria) + 1` ends up being 385, based on looking at the other references to `daria` above. How does this work?

src/test/app/TxQ_test.cpp
```diff
@@ -490,23 +529,24 @@ class TxQPosNegFlows_test : public beast::unit_test::suite
...
-        env(noop(alice), ticket::use(tkt1 + 9), fee(19), queued);
-        env(noop(alice), ticket::use(tkt1 + 10), fee(20), queued);
-        env(noop(alice), ticket::use(tkt1 + 11), fee(21), queued);
-        env(noop(alice), ticket::use(tkt1 + 12), fee(22), queued);
+        env(noop(alice), ticket::use(tkt1 + 1), fee(baseFee * 1.1));
+        env(noop(alice), ticket::use(tkt1 + 2), fee(baseFee * 1.2));
+        env(noop(alice), ticket::use(tkt1 + 3), fee(baseFee * 1.3));
+        env(noop(alice), ticket::use(tkt1 + 4), fee(baseFee * 1.4));
+        env(noop(alice), ticket::use(tkt1 + 5), fee(baseFee * 1.5), queued);
+        auto const expectedMinFeeLevel = txFeeLevelByAccount(env, alice) + 1;
```
bthomee, line 537: To confirm, the expected min fee level is the value of the fee once the queue starts filling up?

src/test/app/TxQ_test.cpp
```diff
@@ -1647,22 +1709,40 @@ class TxQPosNegFlows_test : public beast::unit_test::suite
...
+                // We know there's going to be 11 transactions in ledger
+                // Need to fetch 5th transaction cost here to be able to
+                // calculate median fee level.
+                if (i == 4)
+                {
+                    double const feeMultiplier =
+                        static_cast<double>(cost.drops()) / baseFee;
+                    medFeeLevel = FeeLevel64{static_cast<uint64_t>(
+                        feeMultiplier * baseFeeLevel.fee())};
+                }
```
bthomee, line 1734: Wouldn't it be better to modify `calcMedFeeLevel` to take in a list of transactions and determine the median fee level there?

src/test/app/TxQ_test.cpp
```diff
@@ -151,7 +154,38 @@ class TxQPosNegFlows_test : public beast::unit_test::suite
...
+        auto const& txq = env.app().getTxQ();
+        auto const& txs = txq.getAccountTxs(account.id());
+
+        return txs.begin()->feeLevel.fee();
+    }
+
+    // Calculating expected median fee level based on known fee levels of median
+    // transaction levels.
+    auto
+    calcMedFeeLevel(FeeLevel64 const feeLevel1, FeeLevel64 const feeLevel2)
```
bthomee, line 177: Isn't there an existing function you can call? Since the fee level is determined by the transactions seen and used in the non-test code, surely there must be a function that already does this?

vvysokikh1: There's the code that does that: https://github.com/XRPLF/rippled/blob/5bf3a308d5d550c60b30514af77313aaf91b37b5/src/xrpld/app/misc/detail/TxQ.cpp#L172

Unfortunately this is inside of a struct that is declared private inside of transaction queue. Without changes to the actual code it's not possible to use

src/test/app/TxQ_test.cpp
vvysokikh1: This is possible, but it's the only occasion where such functionality is required. Not worth generalizing it I think

vvysokikh1: plus it's really annoying to take out the transactions that get directly into the ledger...

src/test/app/TxQ_test.cpp
bthomee: I understood that transactions using a higher fee could be prioritized over those using a lower fee - but here the ones using lower fee go into the ledger while those using a higher fee end up being queued.

Note that this is just asking about functionality - not related to this PR as your changes here are clear.

bthomee (approved):

src/test/app/TxQ_test.cpp
vvysokikh1: here the size pf the ledger is 4. So first 4 transactions just get into the ledger, essentially they by-pass the queue completely. That's why their fee don't matter at all. The queue comes into an effect only when the ledger is full.

PR #5108, opened 2024-08-27
Introduce MPT V1 support (XLS-33d)

gregtatcam:
## High Level Overview of Change

New Transactions:
- MPTokenIssuanceCreate
- MPTokenIssuanceDestory
- MPTokenIssuanceSet
- MPTokenAuthorize

Modified Transactions:
- Payment
- Clawback

New Objects:
- MPTokenIssuance
- MPToken

API updates:
- ledger_entry
- account_objects
- ledger_data

Refactor:
- Add new types STMPTAmount, MPTIssue, MPTAmount to define MPT amount.
- Add STEitherAmount as a variant of STAmount and STMPTAmount. STEitherAmount
  replaces STAmount as a serializable type for amount fields.
- Add TypedVariantField and OptionaledVariantField to represent generic serializable variant type.
- Add SOElement constructor overload to describe the amount fields, which may have MPT amount (currently only sfAmount). The constructors are used in TxFormats to define fields supporting MPT.
- Add STObject::operator[] overload for *VariantField to return alternate type, held by a variant field.

### Context of Change

- [XLS-33d specification](https://github.com/XRPLF/XRPL-Standards/tree/master/XLS-0033d-multi-purpose-tokens)

### Type of Change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Refactor (non-breaking change that only restructures code)
- [ ] Performance (increase or change in throughput and/or latency)
- [x] Tests (you added tests for code that already exists, or your new feature included in this PR)
- [ ] Documentation update
- [ ] Chore (no impact to binary, e.g. `.gitignore`, formatting, dropping support for older tooling)
- [ ] Release

### API Impact

- [x] Public API: New feature (new methods and/or new fields)
- [ ] Public API: Breaking change (in general, breaking changes should only impact the next api_version)
- [ ] `libxrpl` change (any change that may affect `libxrpl` or dependents of `libxrpl`)
- [ ] Peer protocol change (must be backward compatible or bump the peer protocol version)

## Test Plan

Added test for new feature:
- MPT

## Future Tasks

Integrate MPT into XRPL DEX.

PR #5121, opened 2024-09-06
Add a new serialized type: STNumber

thejohnfreeman:
This type should let objects and transactions contain multiple fields for quantities of XRP, IOU, or [MPT](https://github.com/XRPLF/rippled/pull/5108) without duplicating information about the "issue" (represented by `STIssue`). It is a straightforward serialization of our internal `Number` type that uniformly represents those quantities.

### DIFF
--- a/include/xrpl/protocol/STNumber.h
+++ b/include/xrpl/protocol/STNumber.h
@@ -1,0 +1,54 @@
+//------------------------------------------------------------------------------
+/*
+    This file is part of rippled: https://github.com/ripple/rippled
+    Copyright (c) 2024 Ripple Labs Inc.
+
+    Permission to use, copy, modify, and/or distribute this software for any
+    purpose  with  or without fee is hereby granted, provided that the above
+    copyright notice and this permission notice appear in all copies.
+
+    THE  SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+    WITH  REGARD  TO  THIS  SOFTWARE  INCLUDING  ALL  IMPLIED  WARRANTIES  OF
+    MERCHANTABILITY  AND  FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+    ANY  SPECIAL ,  DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+    WHATSOEVER  RESULTING  FROM  LOSS  OF USE, DATA OR PROFITS, WHETHER IN AN
+    ACTION  OF  CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+    OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+//==============================================================================
+
+#ifndef XRPL_PROTOCOL_STNUMBER_H_INCLUDED
+#define XRPL_PROTOCOL_STNUMBER_H_INCLUDED
+
+#include <xrpl/basics/Number.h>
+#include <xrpl/protocol/STBase.h>
+
+namespace ripple {
+
+class STNumber final : public STBase
+{
+private:
+    Number value_;
+
+public:
+    STNumber(SField const& n, Number const& v = Number());
+    STNumber(SerialIter& sit, SField const& name);
+
+    SerializedTypeID getSType() const override;
+    std::string getText() const override;
+    Json::Value getJson(JsonOptions) const override;
+    void add(Serializer& s) const override;
+
+    Number value() const;
+    void setValue(Number const& v);
+
+    STBase* copy(std::size_t n, void* buf) const override;
+    STBase* move(std::size_t n, void* buf) override;
+
+    bool isEquivalent(const STBase& t) const override;
+    bool isDefault() const override;
+};
+
+} // namespace ripple
+
+#endif
--- a/src/libxrpl/protocol/STNumber.cpp
+++ b/src/libxrpl/protocol/STNumber.cpp
@@ -1,0 +1,107 @@
+//------------------------------------------------------------------------------
+/*
+    This file is part of rippled: https://github.com/ripple/rippled
+    Copyright (c) 2023 Ripple Labs Inc.
+
+    Permission to use, copy, modify, and/or distribute this software for any
+    purpose  with  or without fee is hereby granted, provided that the above
+    copyright notice and this permission notice appear in all copies.
+
+    THE  SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+    WITH  REGARD  TO  THIS  SOFTWARE  INCLUDING  ALL  IMPLIED  WARRANTIES  OF
+    MERCHANTABILITY  AND  FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+    ANY  SPECIAL ,  DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+    WHATSOEVER  RESULTING  FROM  LOSS  OF USE, DATA OR PROFITS, WHETHER IN AN
+    ACTION  OF  CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+    OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+//==============================================================================
+
+#include <xrpl/protocol/STNumber.h>
+
+#include <xrpl/protocol/SField.h>
+
+namespace ripple {
+
+STNumber::STNumber(SField const& n, Number const& v)
+    : STBase(n), value_(v)
+{
+}
+
+STNumber::STNumber(SerialIter& sit, SField const& name)
+    : STBase(name)
+{
+    // We must call these methods in separate statements
+    // to guarantee their order of execution.
+    auto mantissa = sit.geti64();
+    auto exponent = sit.geti32();
+    value_ = Number{mantissa, exponent};
+}
+
+SerializedTypeID
+STNumber::getSType() const
+{
+    return STI_NUMBER;
+}
+
+std::string
+STNumber::getText() const
+{
+    return to_string(value_);
+}
+
+Json::Value
+STNumber::getJson(JsonOptions) const
+{
+    // TODO: come up with a string representation.
+    return to_string(value_);
+}
+
+void
+STNumber::add(Serializer& s) const
+{
+    assert(getFName().isBinary());
+    assert(getFName().fieldType == getSType());
+    s.addi64(value_.mantissa());
+    s.addi32(value_.exponent());
+}
+
+Number
+STNumber::value() const
+{
+    return value_;
+}
+
+void
+STNumber::setValue(Number const& v)
+{
+    value_ = v;
+}
+
+STBase*
+STNumber::copy(std::size_t n, void* buf) const
+{
+    return emplace(n, buf, *this);
+}
+
+STBase*
+STNumber::move(std::size_t n, void* buf)
+{
+    return emplace(n, buf, std::move(*this));
+}
+
+bool
+STNumber::isEquivalent(const STBase& t) const
+{
+    assert(t.getSType() == this->getSType());
+    const STNumber* v = dynamic_cast<const STNumber*>(&t);
+    return v && (value_ == v->value_);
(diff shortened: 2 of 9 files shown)


dangell7:
I like this. I had all sorts of fun when doing `sfQuantity` on the options proposal. I think I just ended up using an i32. Some things I would be curious about. 

1. Will there be a `getFieldNumber(sfQuantity)`? (Assuming yes)
2. Also how can one multiply / divide / add / subtract an STNumber with an STAmount. That might be good to show in tests if its not straight forward. For example I have 100 USD and I want to multiply by 100. 

I ended up with `STAmount const totalValue = STAmount(strikePrice.issue(), (strikePrice.mantissa() * quantity));` but maybe I took a wrong turn. 

Thank you

src/libxrpl/protocol/Serializer.cpp
```diff
@@ -71,6 +72,32 @@ Serializer::add64(std::uint64_t i)
...
+int
+Serializer::addi32(std::int32_t i)
+{
+    int ret = mData.size();
+    static constexpr std::size_t SIZE =
+        sizeof(std::int32_t) / sizeof(unsigned char);
+    static_assert(SIZE == 4);
+    mData.resize(ret + SIZE);
+    unsigned char* p = &mData.back() + 1 - SIZE;
+    boost::endian::store_big_s32(p, i);
```
gregtatcam, line 84: Why not use the same serialization approach as in other methods? I.e. `push_back/shift`?

include/xrpl/protocol/STNumber.h
```diff
@@ -0,0 +1,63 @@
...
+
+#ifndef XRPL_PROTOCOL_STNUMBER_H_INCLUDED
+#define XRPL_PROTOCOL_STNUMBER_H_INCLUDED
+
+#include <xrpl/basics/Number.h>
+#include <xrpl/protocol/STBase.h>
+
+namespace ripple {
+
+class STNumber final : public STBase
```
gregtatcam, line 28: Do we need to inherit from `CountedObject`? Most `ST*` types do.

gregtatcam, line 28: Why contain and not inherit from `Number`? All arithmetics then just works.

src/test/protocol/STNumber_test.cpp
```diff
@@ -0,0 +1,61 @@
...
+#include <limits>
+
+namespace ripple {
+
+struct STNumber_test : public beast::unit_test::suite
+{
+    void
+    run() override
+    {
+        {
```
gregtatcam, line 32: Does it make sense to have this block of code in a separate function, for instance `testNumber`. And have `run` call `testNumber`? We'll probably be adding more tests for `STNumber`.

gregtatcam: I guess we don't want to have the arithmetics work between not compatible types.

thejohnfreeman: I split tests out into separate functions at the point it becomes meaningful.

thejohnfreeman:
@dangell7 I added `getFieldNumber`. I generally add methods as-needed, and the `getField*` methods seem incomplete (there is no `getFieldIssue`, for example). Let me know if there are others you want.

I added a test to demonstrate multiplication. The general pattern is to convert everything to `Number`, compute your arithmetic, and then reconstruct final types like `STAmount`. We can add arithmetic helpers for those final types where it makes sense. I'm a little disappointed that conversion from `STAmount` to `Number` must first pass through constructing either an `XRPAmount` or an `IOUAmount`, though I guess it is inlined.

include/xrpl/protocol/STNumber.h
```diff
@@ -0,0 +1,66 @@
...
+                       public CountedObject<STNumber>,
+                       public Number
+{
+public:
+    STNumber() = default;
+    explicit STNumber(SField const& n, Number const& v = Number());
+    STNumber(SerialIter& sit, SField const& name);
+
+    using Number::operator=;
+
```
gregtatcam, line 39: I think should add `using value_type = Number;`. It's used by `STObject::operator[]`, etc.

include/xrpl/protocol/STNumber.h
```diff
@@ -0,0 +1,66 @@
...
+    std::string
+    getText() const override;
+    Json::Value getJson(JsonOptions) const override;
+    void
+    add(Serializer& s) const override;
+
+    Number const&
+    value() const;
+    void
+    setValue(Number const& v);
```
gregtatcam, line 51: Is this needed? Why not use assignment operator?

src/test/protocol/Serializer_test.cpp
```diff
@@ -0,0 +1,51 @@
...
+#include <limits>
+
+namespace ripple {
+
+struct Serializer_test : public beast::unit_test::suite
+{
+    void
+    run() override
+    {
+        std::initializer_list<std::int64_t> const values = {
```
gregtatcam, line 32: Does it also make sense to add `int32` tests?

src/test/protocol/STNumber_test.cpp
```diff
@@ -0,0 +1,72 @@
...
+            STNumber const stnum{sfQuantity};
+            BEAST_EXPECT(stnum.getSType() == STI_NUMBER);
+            BEAST_EXPECT(stnum.getText() == "0");
+            BEAST_EXPECT(stnum.isDefault() == true);
+            BEAST_EXPECT(stnum.value() == Number{0});
+        }
+
+        std::initializer_list<std::int64_t> const values = {
```
gregtatcam, line 42: Should add `int32` tests?

src/test/protocol/STNumber_test.cpp
```diff
@@ -0,0 +1,72 @@
...
+        }
+
+        {
+            STAmount const strikePrice{noIssue(), 100};
+            STNumber const factor{sfQuantity, 100};
+            auto const iouValue = strikePrice.iou();
+            IOUAmount totalValue{iouValue * factor};
+            STAmount const totalAmount{totalValue, strikePrice.issue()};
+            BEAST_EXPECT(totalAmount == Number{10'000});
+        }
```
gregtatcam, line 66: Should we test arithmetics with some large/small fractional values to make sure `STAmount` and `STNumber` produce the same results?

src/libxrpl/protocol/STNumber.cpp
```diff
@@ -0,0 +1,104 @@
...
+
+std::string
+STNumber::getText() const
+{
+    return to_string(*this);
+}
+
+Json::Value STNumber::getJson(JsonOptions) const
+{
+    // TODO: come up with a string representation.
```
gregtatcam, line 53: How about 

     elem[jss::value] = getText();

include/xrpl/protocol/STNumber.h
Speedtraders48: Yes and also make it perfect

Speedtraders48: From the decimal point but it is perfect

src/libxrpl/protocol/Serializer.cpp
Speedtraders48: Because the hour and time ratio are not the same.

src/libxrpl/protocol/Serializer.cpp
```diff
@@ -71,6 +72,32 @@ Serializer::add64(std::uint64_t i)
     return ret;
 }
 
+int
+Serializer::addi32(std::int32_t i)
```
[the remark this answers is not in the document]
thejohnfreeman: My thinking was grouping signed vs unsigned before bit length, but I'm fine either way.

include/xrpl/protocol/STNumber.h
```diff
@@ -0,0 +1,66 @@
...
+#ifndef XRPL_PROTOCOL_STNUMBER_H_INCLUDED
+#define XRPL_PROTOCOL_STNUMBER_H_INCLUDED
+
+#include <xrpl/basics/CountedObject.h>
+#include <xrpl/basics/Number.h>
+#include <xrpl/protocol/STBase.h>
+
+namespace ripple {
+
+class STNumber final : public STBase,
```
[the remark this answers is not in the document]
thejohnfreeman: I didn't put much thought into it. Most of the `ST` classes are final. We can always remove it when it's time to inherit, but I'll remove it now anyways.

include/xrpl/protocol/STNumber.h
```diff
@@ -0,0 +1,66 @@
...
+
+    STBase*
+    copy(std::size_t n, void* buf) const override;
+    STBase*
+    move(std::size_t n, void* buf) override;
+
+    bool
+    isEquivalent(STBase const& t) const override;
+    bool
+    isDefault() const override;
```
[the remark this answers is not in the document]
thejohnfreeman: Won't it just use the existing overload for `Number`?

include/xrpl/protocol/STNumber.h
thejohnfreeman: I think it's for `STObject::setFieldUsingSetValue`.

src/libxrpl/protocol/STNumber.cpp
thejohnfreeman: That will give us things like `{ "sfFieldName": { "value": "1234567" }}` instead of just `{ "sfFieldName": "1234567" }`. I think `to_string` might just be the correct representation. I'll remove the comment.

include/xrpl/protocol/STNumber.h
```diff
@@ -0,0 +1,66 @@
...
+namespace ripple {
+
+class STNumber final : public STBase,
+                       public CountedObject<STNumber>,
+                       public Number
+{
+public:
+    STNumber() = default;
+    explicit STNumber(SField const& n, Number const& v = Number());
+    STNumber(SerialIter& sit, SField const& name);
```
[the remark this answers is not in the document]
thejohnfreeman: I'm hesitant to add this before we have any callers that need it.

include/xrpl/protocol/STNumber.h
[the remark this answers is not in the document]
gregtatcam: It's ambiguous because both `Number` and `STBase` have it overloaded.

src/test/protocol/STNumber_test.cpp
thejohnfreeman: I welcome all tests written by others, but I don't want to spend my time designing those tests right now.

gregtatcam (approved):
👍 LGTM
I left a couple of minor comments to consider.

HowardHinnant:
One thing that bothers me is the ability to destruct a `STNumber` via a reference or pointer to a `Number`.  This would call `Number`'s non-virtual destructor and not call `STNumber`'s destructor.

That error can be prevented at compile-time with this refactor: https://github.com/HowardHinnant/rippled/commit/05ecfc6ef279475d695b065a35ae82476ba65f13

Only very lightly tested.  So if accepted, this does need further scrutiny.

gregtatcam:
> One thing that bothers me is the ability to destruct a `STNumber` via a reference or pointer to a `Number`. This would call `Number`'s non-virtual destructor and not call `STNumber`'s destructor.
> 
> That error can be prevented at compile-time with this refactor: [HowardHinnant@05ecfc6](https://github.com/HowardHinnant/rippled/commit/05ecfc6ef279475d695b065a35ae82476ba65f13)
> 
> Only very lightly tested. So if accepted, this does need further scrutiny.

@HowardHinnant Why not adding a virtual destructor? If `STNumber` is a subclass of `Number` then the math just works.

HowardHinnant:
> @HowardHinnant Why not adding a virtual destructor? If `STNumber` is a subclass of `Number` then the math just works.

The math also just works with the way I refactored it.  Note that I did not have to change any use cases of `STNumber`, in code or in tests.

Adding a virtual destructor to `Number` adds an expense to every client of `Number` whether they use it or not.  My refactor adds no expense to `Number` at all.  It does not touch the declaration or implementation of `Number`. That makes it a cleaner separation of duties.  And the resulting API that clients see is identical, minus the ability to implicitly convert `STNumber*` to `Number*` (and `STNumber&` to `Number&`), which is both unneeded and undesirable.

HowardHinnant (approved):

ximinez:
Waiting on 2.3.0 release

thejohnfreeman:
```
Add a new serialized type: STNumber (#5121)

`STNumber` lets objects and transactions contain multiple fields for
quantities of XRP, IOU, or MPT without duplicating information about the
"issue" (represented by `STIssue`). It is a straightforward serialization of
the `Number` type that uniformly represents those quantities.
```

PR #5122, opened 2024-09-06
Consolidate definitions of fields, objects, transactions, and features

thejohnfreeman:
Leverages a preprocessor technique to put the entire definitions of fields, objects, transactions, and features into one file for each type. When you want to add a field, object, transaction, or feature, you touch just one file. When you want to inspect a definition, you look up just one file.


thejohnfreeman:
Yes, that would be great, but the preprocessor cannot parse JSON or CSV or any other format. It would require the introduction of an additional build step to generate these headers from that definition file. I would welcome that contribution from someone else, but it's out of scope for my current work. This was a drive-by change I made while working on single asset vault.

include/xrpl/protocol/LedgerFormats.h
```diff
@@ -52,150 +52,13 @@ namespace ripple {
...
-    */
-    ltDID = 0x0049,
-
-    /** A ledger object which tracks Oracle
-        \sa keylet::oracle
-     */
-    ltORACLE = 0x0080,
+#pragma push_macro("OBJECT")
+#undef OBJECT
+#define OBJECT(tag, value, name, fields) tag = value,
```
Bronek, line 58: For consistency, please follow the same style as in `Feature.h` (vertical spacing)

include/xrpl/protocol/SField.h
```diff
@@ -355,329 +355,29 @@ using SF_XCHAIN_BRIDGE = TypedField<STXChainBridge>;
...
+
+#define UNTYPED_SFIELD(name, stiSuffix, fieldValue, ...) \
+    extern SField const sf##name;
+
+#pragma push_macro("TYPED_SFIELD")
+#undef TYPED_SFIELD
+
+#define TYPED_SFIELD(name, stiSuffix, fieldValue, ...) \
+    extern SF_##stiSuffix const sf##name;
+
```
Bronek, line 370: For consistency, please follow the same style as in `Feature.h` (order of operations where `#push_macro` and `#undef` go interleaved, then all new `#define`, then `#undef` and `#pop_macro` go interleaved)

include/xrpl/protocol/TxFormats.h
```diff
@@ -55,168 +55,25 @@ namespace ripple {
...
-    ttESCROW_CANCEL = 4,
-
-    /** This transaction type sets or clears an account's "regular key". */
-    ttREGULAR_KEY_SET = 5,
+#pragma push_macro("TRANSACTION")
+#undef TRANSACTION
+#define TRANSACTION(tag, value, name, fields) tag = value,
+#include <xrpl/protocol/transactions.h>
+#undef TRANSACTION
+#pragma pop_macro("TRANSACTION")
```
Bronek, line 64: For consistency, please follow the same style as in Feature.h (vertical spacing)

include/xrpl/protocol/features.h
Bronek, line 1: I would put it in `include/xrpl/protocol/detail` . Also, how about name which clearly refers to a header it is used in, e.g. `Feature_list.h` or even better `Feature.inl` ?

include/xrpl/protocol/objects.h
Bronek, line 1: I would put it in `include/xrpl/protocol/detail`. Also, how about name which clearly refers to a header it is used in, e.g. `LedgerFormats_list.h` or even better `LedgerFormats.inl` ?

include/xrpl/protocol/sfields.h
Bronek, line 1: I would put it in `include/xrpl/protocol/detail`. Also, how about name which clearly refers to a header it is used in, e.g. `SField_list.h` or `SField.inl` ?

include/xrpl/protocol/transactions.h
Bronek, line 1: I would put it in `include/xrpl/protocol/detail. `Also, how about name which clearly refers to a header it is used in, e.g. `TxFormats_list.h` or `TxFormats.inl` ?

src/libxrpl/protocol/LedgerFormats.cpp
```diff
@@ -25,347 +25,25 @@ namespace ripple {
...
+#define UNWRAP(...) __VA_ARGS__
+#pragma push_macro("OBJECT")
+#undef OBJECT
+#define OBJECT(tag, value, name, fields) \
+    add(jss::name, tag, UNWRAP fields, commonFields);
+#include <xrpl/protocol/objects.h>
+#undef OBJECT
+#pragma pop_macro("OBJECT")
+#undef UNWRAP
+#pragma pop_macro("UNWRAP")
```
Bronek, line 46: For consistency, please follow the same style as in Feature.cpp (vertical spacing)

src/libxrpl/protocol/SField.cpp
```diff
@@ -49,386 +49,32 @@ TypedField<T>::TypedField(private_access_tag_t pat, Args&&... args)
...
-    SF_##stiSuffix const sfName(                                            \
-        access, STI_##stiSuffix, fieldValue, txtName, ##__VA_ARGS__);       \
-    static_assert(                                                          \
-        std::string_view(#sfName) == "sf" txtName,                          \
-        "Declaration of SField does not match its text name")
-
-// clang-format off
+#define TYPED_SFIELD(name, stiSuffix, fieldValue, ...) \
+    SF_##stiSuffix const sf##name(                     \
+        access, STI_##stiSuffix, fieldValue, #name, ##__VA_ARGS__);
```
Bronek, line 64: For consistency, please follow the same style as in Feature.cpp (order of operations where `#push_macro` and `#undef` go interleaved, then all new `#define`, then `#undef` and `#pop_macro` go interleaved)

src/libxrpl/protocol/TxFormats.cpp
```diff
@@ -47,472 +47,18 @@ TxFormats::TxFormats()
...
+#define UNWRAP(...) __VA_ARGS__
+#pragma push_macro("TRANSACTION")
+#undef TRANSACTION
+#define TRANSACTION(tag, value, name, fields) \
+    add(jss::name, tag, UNWRAP fields, commonFields);
+#include <xrpl/protocol/transactions.h>
+#undef TRANSACTION
+#pragma pop_macro("TRANSACTION")
+#undef UNWRAP
+#pragma pop_macro("UNWRAP")
```
Bronek, line 61: For consistency, please follow the same style as in Feature.cpp (vertical spacing)

include/xrpl/protocol/features.h
```diff
@@ -0,0 +1,107 @@
...
+    THE  SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+    WITH  REGARD  TO  THIS  SOFTWARE  INCLUDING  ALL  IMPLIED  WARRANTIES  OF
+    MERCHANTABILITY  AND  FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+    ANY  SPECIAL ,  DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+    WHATSOEVER  RESULTING  FROM  LOSS  OF USE, DATA OR PROFITS, WHETHER IN AN
+    ACTION  OF  CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+    OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+//==============================================================================
+
```
Bronek, line 19: I would put here something like:

```c++
#if !defined(XRPL_FEATURE) || !defined(XRPL_FIX)
#error "This header cannot be used on its own"
#endif
```

include/xrpl/protocol/objects.h
```diff
@@ -0,0 +1,386 @@
...
+    THE  SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+    WITH  REGARD  TO  THIS  SOFTWARE  INCLUDING  ALL  IMPLIED  WARRANTIES  OF
+    MERCHANTABILITY  AND  FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+    ANY  SPECIAL ,  DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+    WHATSOEVER  RESULTING  FROM  LOSS  OF USE, DATA OR PROFITS, WHETHER IN AN
+    ACTION  OF  CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+    OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+//==============================================================================
+
```
Bronek, line 19: I would put here something like:

```c++
#if !defined(OBJECT)
#error "This header cannot be used on its own"
#endif
```

include/xrpl/protocol/sfields.h
```diff
@@ -0,0 +1,354 @@
...
+    THE  SOFTWARE IS PROVIDED AS IS AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+    WITH  REGARD  TO  THIS  SOFTWARE  INCLUDING  ALL  IMPLIED  WARRANTIES  OF
+    MERCHANTABILITY  AND  FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+    ANY  SPECIAL ,  DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+    WHATSOEVER  RESULTING  FROM  LOSS  OF USE, DATA OR PROFITS, WHETHER IN AN
+    ACTION  OF  CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+    OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+//==============================================================================
+
```
Bronek, line 19: I would put here something like:

```c++
#if !defined(UNTYPED_SFIELD) || !defined(TYPED_SFIELD)
#error "This header cannot be used on its own"
#endif
```

include/xrpl/protocol/transactions.h
```diff
@@ -0,0 +1,414 @@
...
+    THE  SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+    WITH  REGARD  TO  THIS  SOFTWARE  INCLUDING  ALL  IMPLIED  WARRANTIES  OF
+    MERCHANTABILITY  AND  FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+    ANY  SPECIAL ,  DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+    WHATSOEVER  RESULTING  FROM  LOSS  OF USE, DATA OR PROFITS, WHETHER IN AN
+    ACTION  OF  CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+    OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+//==============================================================================
+
```
Bronek, line 19: I would put here something like:

```c++
#if !defined(TRANSACTION)
#error "This header cannot be used on its own"
#endif
```

Bronek (commented):
Few initial comments, also this: I wonder if this pattern is perhaps too permissive

```c++
#pragma push_macro("FOO)
#undef FOO

#define FOO . . .;

#include <xrpl/protocol/. . .>

#undef FOO
#pragma pop_macro("FOO")
```

Could this be better (i.e. more robust) ?

```c++
#ifdef FOO
#error "Macro FOO was found to be unexpectedly defined"
#endif

#define FOO . . .;

#include <xrpl/protocol/. . .>

#undef FOO
```

mvadari:
> This is a useful change and would make the manual process of updating third party libraries with new field types, transactions and ledger entries a little easier. However, an even more useful change would be to abstract the definition files into easily machine parsable formats, maybe either JSON or CSV, so that any third party API libraries can choose to auto-generate their data structures and stream encoders and decoders from those files, thereby keeping up to date much more easily than using git blame to find recent changes to the large definition files.

@donovanhide does the `server_definitions` RPC solve your problem? https://xrpl.org/docs/references/http-websocket-apis/public-api-methods/server-info-methods/server_definitions

include/xrpl/protocol/features.h
thejohnfreeman: I much prefer the simple `features.h` naming.

mvadari:
> Just to be clear of my use case, it is updating the data structures here:
> 
> https://github.com/rubblelabs/ripple/blob/master/data/ledgerentry.go https://github.com/rubblelabs/ripple/blob/master/data/transaction.go https://github.com/rubblelabs/ripple/blob/master/data/format.go https://github.com/rubblelabs/ripple/blob/master/data/flags.go
> 
> And the canonical source of truth for these data structures, encodings and flags are the source code files being changed in this PR. It would be much more efficient if my library and others could be derived from an easily machine readable file rather than a hard to diff C++ file. Seeing as this is the essence of the Ripple protocol, it doesn't seem like an unreasonable request :-)

@donovanhide that's a fair statement, but I think it should be in an RPC rather than in the source code (perhaps added to `server_definitions`). Let's move this discussion to [this issue](https://github.com/XRPLF/rippled/issues/5151) I just opened on the topic.

thejohnfreeman:
Whenever it comes time to merge this branch with the changes in #5143, regardless of whichever lands in `develop` first, the conflict resolutions can be cherry picked from my `vault` [branch](https://github.com/thejohnfreeman/rippled/tree/vault).

include/xrpl/protocol/features.h
Bronek: I totally do not mind `.ipp`, or even stick with `.h` if you insist, but then I would very much prefer if it was moved to `detail` subdir as not to make clutter. These files are not standalone, like all well-written header files should be, so having a different extension is a hint to the user that IDE errors are expected when viewing them

Bronek: never mind, I see they have been moved to `detail`, thanks !

Bronek (approved):

include/xrpl/protocol/features.h
thejohnfreeman: I moved them to `detail/` subdirectory.

gregtatcam (commented):
I like this idea of changing just one file a lot but IMHO there are two inconvenient issues:

- Because of the macros, some IDE (I have CLion) are not going to be able to resolve the references. For instance, clicking on a field name in transactions.m or ledger_entries.m doesn't navigate to the fields declaration. And searching for the field with the `sf` prefix is not going to be helpful either - see the next issue.

- I'd prefer that the fields keep `sf` prefix. For instance, when I search for `sfAmount` I can see the field type and the occurrences. But try to search for `Amount` and you'll get huge number of hits. Of course I know where to find the field declaration but I have to navigate to this file separately.

include/xrpl/protocol/detail/features.m
```diff
@@ -0,0 +1,112 @@
+//------------------------------------------------------------------------------
```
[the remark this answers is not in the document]
Bronek: I asked to use a different extension that cannot be confused with a header, because these files are not headers in normal sense of the word (because of macros they rely on) - they cannot be included as-is, they won't parse in an IDE, they cannot be formatted with clang-format. I suggested `.inl` as that's typically used for "inline files" but @thejohnfreeman chose different.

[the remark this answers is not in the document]
gregtatcam: It's confusing IMO, but I'm fine with this.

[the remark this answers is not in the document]
gregtatcam: This : https://gcc.gnu.org/onlinedocs/cpp/Header-Files.html suggests headers can be used for macros.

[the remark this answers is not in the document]
Bronek: Yes, headers are often used for macro definitions. Also the language attaches no meaning to file extension. However things get a little tricky for IDEs and other tools when, what appears to be a header file from extension, does not work like one. In this case the apparent header files rely on a macro definitions brought from the outside _without_ the relevant `#include` inside the file, which is not normal practice of how header files work.

If that helps to get the PR approved, I am happy for @thejohnfreeman to rename these files to any extension, including `.h`. I guess it does not matter much.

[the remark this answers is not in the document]
gregtatcam: `.m` feels odd but not a showstopper I guess. I'm more concerned about two other issues that I mentioned.

include/xrpl/protocol/detail/ledger_entries.m
```diff
@@ -0,0 +1,394 @@
...
+    MERCHANTABILITY  AND  FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+    ANY  SPECIAL ,  DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+    WHATSOEVER  RESULTING  FROM  LOSS  OF USE, DATA OR PROFITS, WHETHER IN AN
+    ACTION  OF  CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
+    OR IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+*/
+//==============================================================================
+
+#if !defined(LEDGER_ENTRY)
+#error "undefined macro: LEDGER_ENTRY"
```
Bronek, line 21: I just occurred to me that in order to make this file (and others) similar to a regular header, we could do this instead of `#error` :

```
#define LEDGER_ENTRY(...)
#pragma message("using dummy LEDGER_ENTRY")
```

Note the dummy macro is empty. This in turn would make the file well-behaved in IDEs and other tools, meaning it could be renamed back to `.h` if you so desire.

Bronek: @gregtatcam any comments on this ?

gregtatcam: Sounds good to me.

gregtatcam: It still doesn't solve the issue of resolving the fields reference by IDE, does it?

Bronek: You are right - it does not. An actual macro doing something could do that, maybe ?

include/xrpl/protocol/detail/features.m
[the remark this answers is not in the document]
thejohnfreeman: Would everyone like `.macro` more than both `.m` and `.h`?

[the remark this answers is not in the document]
gregtatcam: If for some reason `.h` is not an option then `.macro` is better. 
But as I mentioned earlier, using macros in general prevents IDE (at least CLion) resolving the fields reference. I find this really inconvenient.

gregtatcam (approved):
👍 LGTM

PR #5123, opened 2024-09-09
Enforce non-standard currency rule of not starting with 0x00

gregtatcam:

This amendment enforces non-standard currency rule per [XRPL documentation](https://xrpl.org/docs/references/protocol/data-types/currency-formats#nonstandard-currency-codes).

* Add amendment fixNonStandardCurrency
* Fail AMMCreate, CheckCreate, OfferCreate, TrustSet, and Payment transactions if non-standard currency starts with 0x00

### Type of Change

- [x] Documentation update
- [ ] Release

## Test Plan
Updated unit-tests:
- AMM
- Check
- Flow
- Offer
- SetTrust

### DIFF
--- a/src/libxrpl/protocol/UintTypes.cpp
+++ b/src/libxrpl/protocol/UintTypes.cpp
@@ -133,8 +133,31 @@ badCurrency()
 }
 
 bool
-validCurrency(Currency const& currency)
+validCurrency(Currency const& currency, PaymentTx paymentTx)
 {
+    // Allow payments for invalid non-standard currencies
+    // created pre fixNonStandardCurrency amendment
+    static std::set<Currency> whiteList = {
+        Currency{"0000000000000000000000000000000078415059"},
+        Currency{"00000000004150756E6B30310000000000000000"},
+        Currency{"0000000000D9A928EFBCBEE297A1EFBCBE29DBB6"},
+        Currency{"0000000000414C6F676F30330000000000000000"},
+        Currency{"0000000000000000000000005852500000000000"},
+        Currency{"000028E0B2A05FE0B2A029E2948CE288A9E29490"},
+        Currency{"00000028E2989EEFBE9FE28880EFBE9F29E2989E"},
+        Currency{"00000028E381A3E29795E280BFE2979529E381A3"},
+        Currency{"000000000000005C6D2F5F283E5F3C295F5C6D2F"},
+        Currency{"00000028E295AFC2B0E296A1C2B0EFBC89E295AF"},
+        Currency{"0000000000000000000000005852527570656500"},
+        Currency{"000000000000000000000000302E310000000000"},
+        Currency{"0000000000E1839A28E0B2A05FE0B2A0E1839A29"},
+        Currency{"0000000048617070794E6577596561725852504C"},
+        Currency{"0000E29D9AE29688E29590E29590E29688E29D9A"},
+        Currency{"000028E297A35FE297A229E2948CE288A9E29490"},
+        Currency{"00000000CA95E0B2A0E0B2BFE1B4A5E0B2A0CA94"},
+        Currency{"000000282D5F282D5F282D5F2D295F2D295F2D29"},
+        Currency{"0000000000000000000000000000000078415049"},
+        Currency{"00000000000028E295ADE0B2B05FE280A2CC8129"}};
     static constexpr Currency sIsoBits(
         "FFFFFFFFFFFFFFFFFFFFFFFF000000FFFFFFFFFF");
 
@@ -146,7 +169,7 @@ validCurrency(Currency const& currency)
     if (*currency.cbegin() != 0x00)
         return true;
 
-    return false;
+    return paymentTx == PaymentTx::Yes && whiteList.contains(currency);
 }
 
 }  // namespace ripple
--- a/src/libxrpl/protocol/UintTypes.cpp
+++ b/src/libxrpl/protocol/UintTypes.cpp
@@ -132,4 +132,21 @@ badCurrency()
     return currency;
 }
 
+bool
+validCurrency(Currency const& currency)
+{
+    static constexpr Currency sIsoBits(
+        "FFFFFFFFFFFFFFFFFFFFFFFF000000FFFFFFFFFF");
+
+    // XRP or standard currency
+    if (currency == xrpCurrency() || ((currency & sIsoBits).isZero()))
+        return true;
+
+    // Non-standard currency must not start with 0x00
+    if (*currency.cbegin() != 0x00)
+        return true;
+
+    return false;
+}
+
 }  // namespace ripple
--- a/include/xrpl/protocol/Feature.h
+++ b/include/xrpl/protocol/Feature.h
@@ -80,7 +80,7 @@ namespace detail {
 // Feature.cpp. Because it's only used to reserve storage, and determine how
 // large to make the FeatureBitset, it MAY be larger. It MUST NOT be less than
 // the actual number of amendments. A LogicError on startup will verify this.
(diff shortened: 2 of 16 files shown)


thejohnfreeman (changes requested):
I think we need to add checks to `NFTokenCreateOffer` (and `NFTokenMint` when `featureNFTokenMintOffer` is enabled).

mvadari:
How will this work for existing tokens that already start with `0x00`? Will those effectively be banned? Can the reserves in those objects be reclaimed?

src/libxrpl/protocol/UintTypes.cpp
```diff
@@ -132,4 +132,44 @@ badCurrency()
...
+        Currency{"0000000000000000000000005852527570656500"},
+        Currency{"000000000000000000000000302E310000000000"},
+        Currency{"0000000000E1839A28E0B2A05FE0B2A0E1839A29"},
+        Currency{"0000000048617070794E6577596561725852504C"},
+        Currency{"0000E29D9AE29688E29590E29590E29688E29D9A"},
+        Currency{"000028E297A35FE297A229E2948CE288A9E29490"},
+        Currency{"00000000CA95E0B2A0E0B2BFE1B4A5E0B2A0CA94"},
+        Currency{"000000282D5F282D5F282D5F2D295F2D295F2D29"},
+        Currency{"0000000000000000000000000000000078415049"},
+        Currency{"00000000000028E295ADE0B2B05FE280A2CC8129"}};
```
Bronek, line 160: An interesting thing might happen if someone creates a new invalid currency during the voting period of this amendment. Even worse if it happens during the two-week period after voting has completed. I do not think there's anything we can do about it, other than allow errors to happen on transactions with such an invalid currency in the future - I think there should be **no** future PR to extend this list.

gregtatcam: Yes, I agree. There should be no future PR to extend the list. But it might make sense to check if there are more invalid currencies created before this PR is merged. This list is good as of couple of weeks ago.

MarkusTeufelberger:
This list should be in client applications, not in the protocol.

src/libxrpl/protocol/UintTypes.cpp
MarkusTeufelberger: It would be trivial to create thousands of additional entries for this list until the PR is merged. The way currency codes are being interpreted is up to client applications and should not be part of the protocol.

See also the documentation: "To prevent this from **being treated as** a "standard" currency code" - there is nothing stating that "standard currency codes" are a rule.

If you want to enforce standards, you could also start blocking non-ISO-4217 currency codes. This would include BTC, ETH and XRP by the way.

gregtatcam:
> This list should be in client applications, not in the protocol.

If the white-listed currencies are in the client application and the protocol enforces the rule for the non-standard currencies then the transactions with the white-listed currencies are going to fail. The intention is to allow the non-standard currencies created pre-amendment to work.

MarkusTeufelberger:
The protocol should not interpret currency codes at all. Otherwise you should also start checking for valid ASCII characters and even only allow ISO-4217 codes for "standard currencies" per the documentation.

src/libxrpl/protocol/UintTypes.cpp
gregtatcam: There are about 140,000 currencies codes out of which ~4,700 are standard and ~135,300 are non-standard. The 22 non-standard currencies that don't follow the rule were clearly created unintentionally. Creating thousands more  doesn't serve any purpose and is a violation of the rule.

MarkusTeufelberger: This is a guideline for interpreting a 160 bit field when using some APIs, not a rule. I also don't see any reason or benefit of limiting this in the protocol (especially with such a clunky allowlist approach, considering that there are more than one ledger out there using this software).

gregtatcam:
The motivation for enforcing the non-standard currency rule is efficient serialization of Multi Purpose Token (MPT) in the lending protocol and AMM. AMM is currently identified by a token pair where each token is currency (160 bits)+account (160 bits) and is serialized as 320 bits. MPT is identified by 192 bits MPTID: 32 bits sequence + 160 bits account. Without the rule enforced, MPT is going to be serialized as 160+160+32 = 352 bits. With the rule enforced, MPT is serialized as 16+192 bits, where 16 bits is a mask starting with zero byte. If the first two bytes of the serialized data matches the mask then the next 192 bits is MPTID, otherwise it's a currency code followed by account. The saving is 144 bits/18 bytes. Based on this PR feedback, there are clearly downsides/challenges to this approach, but sharing this added context to see if any other suggestions for eliminating those downsides while enabling more efficient serialization.

gregtatcam:
> Are you referring to this? https://github.com/XRPLF/XRPL-Standards/tree/master/XLS-0033d-multi-purpose-tokens#a114-why-isnt-mptokenissuanceid-constructed-from-the-hash-of-an-issuer-address-and-currency-code
> 
> I can't find any reference to the 16 bit mask issue you have just raised in your previous comment. Could you please elucidate?

Correct. This spec doesn't include MPT serialization for AMM because MPT is not integrated into DEX in V1 release.

We can serialize MPT as 0x00ff + 32 bit sequence + 160 bit account. The mask is 0x00ff. When we deserialize, we look at the first 16 bits. Standard currency starts with 0x00 but is followed by another 88 zero bit. So it can't be standard currency. If non-standard currency rule is enforced then it can't be non-standard currency since it can't start with 0x00. We infer this to be MPTID and read next 192 bits. If the first two bytes are not 0x00ff then the data must be 160 bit currency + 160 bit account.

gregtatcam:
> Why not check if all the bytes are zero except the three standard currency bytes rather than just checking the first two bytes? Then you have completely validated the currency code? If so, come up with a short encoding where the actual three bytes of the standard currency code are encoded rather than the full 20 bytes. Maybe you need a prefix byte which says which of the two options you are using giving you a 16 byte saving in the standard currency code case. TBH, this all seems a bit too clever, much like using the inf bit of the float to encode both fixed and floating point numbers into an Amount.

I'm not following. I don't need to validate anything. I need to distinguish MPT from currency+account. Current encoding for AMM token pair (STIssue type) is asset: 160 currency + 160 bit account; asset2: 160 currency + 160 bit account. MPT is 192 bit data. We need to serialize MPT so that it is backward compatible with the current STIssue serialization.

src/libxrpl/protocol/UintTypes.cpp
ximinez: What if, instead of a hard-coded whitelist, you put the whitelist on the ledger in a new singleton object that just holds an array of Currencies, much like the Amendments singleton?

It could use the `LedgerFix` transaction introduced in #4945 with a new `LedgerFixType`. Add fields to `LedgerFix` to specify the currency code, and one account that has an existing trust line (thus ensuring that this is a currency that already exists on the ledger, and nobody can whitelist a currency preemptively). Use the default base fee, since this isn't doing a lot of work. You've already got an amendment gating all of this, so you wouldn't need another one. It might be worth reusing the ideas from `Rules` so that the object doesn't have to keep being loaded, but this doesn't have to do the extra work that `Rules` does to handle presets, so the existing object caches may be enough - it isn't likely to expire from the cache since so many transactions will read it.

This would
1. Get rid of the hard-coded list.
2. Allow the whitelist to work across all networks.
3. Allow any new invalid currencies created before the amendment is enabled to also be whitelisted.
4. Keeping it cheap, should encourage issuers to pay the few drops to fix their currencies, but if they're not, anyone can do it. Whitelisting all the current ones would only cost 200 drops total.

PR #5126, opened 2024-09-11
Reduce duplicate peer traffic for ledger data

ximinez:

Several changes to help reduce message traffic and improve logging and visibility.

* Suppress duplicate `TMGetLedger` and `TMLedgerData` messages, reducing the overhead of processing those messages.
* Reduce the number of those messages sent to peers.
* Improve logging related to those messages, as well as a few other areas.
* Introduces a new protocol version to gate a new feature on the `TMLedgerData` message, which allows multiple identical requests to be replied to with one message.

These changes are organized into several commits which are organized logically separating each functional operation. They can be merged as-is, or squashed.

### Context of Change

Analysis of the issue that led to #5115 identified heavy `TMGetLedger` request and `TMLedgerData` response traffic between nodes leading up to the syncing incidents. It was later determined that those messages were more a symptom of the problem, and not the root cause. However, leading up to identification of the root cause, these changes were being implemented to cut down on those messages, detect duplicates, etc. That reduction in unnecessary traffic is still valuable, so it's being included here.

### Type of Change

- [X] Performance (increase or change in throughput and/or latency)

### API Impact

None.

## Test Plan

1. Verify that nodes are at least as reliable at staying in sync as they are now.
2. Verify that the indicated messages are sent less frequently by upgraded nodes.
3. Verify that peers that both have this change send fewer ledger messages to each other.

## Future Tasks

I am still working on a follow-up to #4764 that makes use of these changes and other improvements to reduce the number of requests initiated by a given node in the first place.


include/xrpl/basics/CanProcess.h
```diff
@@ -0,0 +1,108 @@
...
+    {
+        return canProcess_;
+    }
+
+private:
+    bool
+    insert()
+    {
+        std::unique_lock<Mutex> lock_(mtx_);
+        bool exists = collection_.contains(item_);
```
Bronek, line 96: This would avoid extra lookup

```c++
auto [_, inserted] = collection_.insert(item_)
return inserted;
```

ximinez: Good catch. Fixed.

include/xrpl/basics/base_uint.h
```diff
@@ -623,6 +623,13 @@ to_string(base_uint<Bits, Tag> const& a)
     return strHex(a.cbegin(), a.cend());
 }
 
+template <std::size_t Bits, class Tag>
+inline std::string
+to_short_string(base_uint<Bits, Tag> const& a)
```
vlntb, line 628: [nit]: Adding checks for the to_short_string in the base_unit_test next to existing to_string cases would be good.

src/xrpld/overlay/detail/PeerImp.cpp
```diff
@@ -3008,7 +3243,41 @@ PeerImp::getLedger(std::shared_ptr<protocol::TMGetLedger> const& m)
...
+                        m->set_requestcookie(id());
+                        peer->send(std::make_shared<Message>(
+                            *m, protocol::mtGET_LEDGER));
+                        JLOG(p_journal_.debug())
+                            << "getLedger: Request relayed to peer ["
+                            << peer->id() << "]: " << mHash;
+                        return ledger;
+                    }
+
+                    JLOG(p_journal_.trace())
```
vlntb, line 3277: [nit]: Should this be a `warn` instead of `trace`? Not having a peer to relay the request may indicate some configuration or environment issues.

src/xrpld/overlay/detail/PeerImp.cpp
```diff
@@ -2936,7 +3078,9 @@ getPeerWithLedger(
 void
 PeerImp::sendLedgerBase(
     std::shared_ptr<Ledger const> const& ledger,
-    protocol::TMLedgerData& ledgerData)
+    protocol::TMLedgerData& ledgerData,
+    std::map<std::shared_ptr<Peer>, std::set<std::optional<uint64_t>>> const&
```
vlntb, line 3082: `std::map<std::shared_ptr<Peer>, std::set<std::optional<uint64_t>>>` is mentioned in four places. It would be more readable to define an alias:
`using PeerCookieMap = std::map<std::shared_ptr<Peer>, std::set<std::optional<uint64_t>>>;`

vlntb (commented):
Only minor comments. Happy to approve once addressed.

include/xrpl/basics/base_uint.h
ximinez: I wouldn't call that a nit. Missing test coverage is pretty significant. Thanks for catching it. Fixed.

src/xrpld/overlay/detail/PeerImp.cpp
ximinez: Not necessarily, though. It could mean that the node has already sent the request to all it's peers. Also not that the original message is `trace`.

But there's something odd here. It looks like this code block was somehow duplicated! I must have messed up resolving a conflict when I rebased from `master` to `develop`. I've removed the duplicate.

src/xrpld/overlay/detail/PeerImp.cpp
ximinez: Good suggestion! Fixed.

vlntb (approved):

include/xrpl/basics/CanProcess.h
```diff
@@ -0,0 +1,106 @@
...
+        : mtx_(mtx), collection_(collection), item_(item), canProcess_(insert())
+    {
+    }
+
+    ~CanProcess()
+    {
+        if (canProcess_)
+        {
+            std::unique_lock<Mutex> lock_(mtx_);
+            collection_.erase(item_);
```
Bronek, line 76: How about replacing the (likely) `O(log N)` element lookup with `O(1)` erase of an iterator ? The iterator which is returned from `insert` to be specific, and currently ignored. In this case the `Item` wouldn't have to be stored inside `CanProcesa` object, so that's also one less template parameter and probably also smaller object size (hashes are larger than iterators I guess).

Could even go one step further and replace all data members with `std::function<void()> cleanup_` which would capture all that it needs if `insert` succeeded, or is empty if `insert` failed. In this case no template parameters would be needed at all.

src/xrpld/app/ledger/detail/InboundLedgers.cpp
```diff
@@ -75,18 +75,98 @@ class InboundLedgersImp : public InboundLedgers
...
+            bool const shouldAcquire = [&]() {
+                if (!app_.getOPs().isNeedNetworkLedger())
+                    return true;
+                if (reason == InboundLedger::Reason::GENERIC)
+                    return true;
+                if (reason == InboundLedger::Reason::CONSENSUS)
+                    return true;
+                return false;
+            }();
+            assert(
```
Bronek, line 87: This looks like unnecessary repetition of the condition used to initialise `shouldAcquire` but there is also a subtle bug here. We are reading `getOPs().isNeedNetworkLedger()` more than once, but this function returns an atomic (I mean its implementation in `NetworkOPs.cpp`) so in principle each time we call it, we might get a different response. I am not certain this condition can actually happen, but at the very least it is brittle code.

Also please note 3rd call to `getOPs().isNeedNetworkLedger()` inside logging below, the same applies as it could generate some confusing logs

src/xrpld/app/ledger/detail/InboundLedgers.cpp
```diff
@@ -75,18 +75,98 @@ class InboundLedgersImp : public InboundLedgers
...
+                // the validated ledger. If the requested ledger is between
+                // 1 and 19 inclusive ledgers ahead of the valid ledger this
+                // node has not built it yet, but it's possible/likely it
+                // has the tx's necessary to build it and get caught up.
+                // Plus it might not become validated. On the other hand, if
+                // it's more than 20 in the future, this node should request
+                // it so that it can jump ahead and get caught up.
+                LedgerIndex const validSeq =
+                    app_.getLedgerMaster().getValidLedgerIndex();
+                constexpr std::size_t lagLeeway = 20;
```
Bronek, line 123: is this the best location for this constant ? I am asking, not judging.

src/xrpld/app/ledger/detail/InboundLedgers.cpp
```diff
@@ -135,28 +221,25 @@ class InboundLedgersImp : public InboundLedgers
...
-            acquire(hash, seq, reason);
-        }
-        catch (std::exception const& e)
-        {
-            JLOG(j_.warn())
-                << "Exception thrown for acquiring new inbound ledger " << hash
-                << ": " << e.what();
-        }
-        catch (...)
+        if (CanProcess check{acquiresMutex_, pendingAcquires_, hash})
```
Bronek, line 224: Looking good 👍

Bronek (commented):
Several notes for now, also I noticed there's a bunch of `Added line not covered by tests` annotations from codecov, perhaps worth checking these.

include/xrpl/basics/CanProcess.h
```diff
@@ -0,0 +1,106 @@
...
+        }
+    }
+
+    bool
+    canProcess() const
+    {
+        return canProcess_;
+    }
+
+    operator bool() const
```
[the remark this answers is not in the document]
ximinez: > Please use `explicit` here; it would be consistent with `std::optional` and most other `operator bool` inside the project.

Fixed

include/xrpl/basics/CanProcess.h
```diff
@@ -0,0 +1,106 @@
...
+    {
+        if (canProcess_)
+        {
+            std::unique_lock<Mutex> lock_(mtx_);
+            collection_.erase(item_);
+        }
+    }
+
+    bool
+    canProcess() const
```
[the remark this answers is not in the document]
ximinez: > given that it is not covered by tests, perhaps we do not need this function ?

It's used in the call path for network traffic, most of which is not covered by tests. But I changed that call to use the boolean conversion, and everything seems fine.

include/xrpl/basics/CanProcess.h
ximinez: > How about replacing the (likely) `O(log N)` element lookup with `O(1)` erase of an iterator ? The iterator which is returned from `insert` to be specific, and currently ignored. In this case the `Item` wouldn't have to be stored inside `CanProcess` object, so that's also one less template parameter and probably also smaller object size (hashes are larger than iterators I guess).

I didn't really consider using an iterator because I didn't want to consider the risks of it getting invalidated for any valid `Collection` type. For example, it looks like `unordered_map` would fit the template, but can invalidate iterators. One option is to force the collection to be `std::set`, which the current callers use.

> Could even go one step further and replace all data members with `std::function<void()> cleanup_` which would capture all that it needs if `insert` succeeded, or is empty if `insert` failed. In this case no template parameters would be needed at all.

So only the ctor and `insert` would have template parameters? That could work.

ximinez: I was incorrect in that last comment. `unordered_map` doesn't fit the template, but `unordered_set` does, and it also [says it can potentially invalidate operators](https://en.cppreference.com/w/cpp/container/unordered_set/insert). So, I wrote a generic `insert()` that doesn't use iterators, and a specialized one that does, and wrote tests for both.  It may be overkill....

src/xrpld/app/ledger/detail/InboundLedgers.cpp
ximinez: Ugh. I should have removed that `assert` before opening the PR, but forgot. I was just double-checking that I had rewritten old code into new code. I've done that, and stored the result of `isNeedNetworkLedger` into a local variable, so that all the uses are guaranteed consistent.

src/xrpld/app/ledger/detail/InboundLedgers.cpp
ximinez: > is this the best location for this constant ? I am asking, not judging.

I think it's fine for now. It's specific to this function, mostly just to give it a name instead of using an unnamed magic number statement, and I can't imagine using it anywhere else. (I stole this code from @mtrippled, and didn't see any reason to change it.)

ximinez: @Bronek I'll push the changes I described tomorrow.

ximinez: @Bronek Changes pushed. I'm not 100% sure they're correct because Boost is giving me a new compiler error about `string_view` conversions in Windows.

src/xrpld/app/ledger/detail/InboundLedger.cpp
```diff
@@ -392,7 +392,14 @@ InboundLedger::onTimer(bool wasProgress, ScopedLockType&)
 
     if (!wasProgress)
     {
-        checkLocal();
+        if (checkLocal())
+        {
+            // Done. Something else (probably consensus) built the ledger
+            // locally while waiting for data (or possibly before requesting)
+            assert(isDone());
```
Bronek, line 399: please replace with `XRPL_ASSERT` and a short description why; see `CONTRIBUTING.md` for contracts and instrumentation naming guidelines

src/xrpld/app/misc/HashRouter.cpp
```diff
@@ -128,4 +142,20 @@ HashRouter::shouldRelay(uint256 const& key)
...
+    uint256 const& key,
+    std::function<bool(PeerShortID)> callback)
+{
+    std::lock_guard lock(mutex_);
+
+    auto& s = emplace(key).first;
+
+    for (auto const peerID : s.peekPeerSet())
+    {
+        if (!callback(peerID))
```
Bronek, line 156: How fast are these callbacks - we are running them over a loop inside critical section. Might need to think of an alternative solution, to avoid hogging the `mutex_`. Not a major issue, because as indicated in first sentence - I do not know if these callbacks are fast or slow (or we don't know ?)

src/xrpld/overlay/detail/PeerImp.cpp
Bronek, line 1: There are 10 `assert` added here, can you please turn them to `XRPL_ASSERT`, following naming guideline in `CONTRIBUTING.md` ?

Bronek (approved):
Only minor remarks related to use of `assert`, once fixed feel free to merge

Note to self: add unit tests for `PeerImp.cpp` in a future PR

Bronek (commented):
Actually I realized I added a comment re. use of a loop and callback inside critical section. A sufficient explanation why this is safe might be sufficient to address this comment, but for now this is not approved.

Bronek:
> @Bronek Any outstanding issues for @ximinez to address?

Yes, just added some. The loop and callback inside critical section can be probably resolved with an explanation e.g. a comment in code. The `assert` ones are trivial to fix.

ximinez:
Commit message:
```
Reduce duplicate peer traffic for ledger data (#5126)

- Drop duplicate outgoing TMGetLedger messages per peer
  - Allow a retry after 30s in case of peer or network congestion.
  - Addresses RIPD-1870
  - (Changes levelization. That is not desirable, and will need to be fixed.)
- Drop duplicate incoming TMGetLedger messages per peer
  - Allow a retry after 15s in case of peer or network congestion.
  - The requestCookie is ignored when computing the hash, thus increasing
    the chances of detecting duplicate messages.
  - With duplicate messages, keep track of the different requestCookies
    (or lack of cookie). When work is finally done for a given request,
    send the response to all the peers that are waiting on the request,
    sending one message per peer, including all the cookies and
    a "directResponse" flag indicating the data is intended for the
    sender, too.
  - Addresses RIPD-1871
- Drop duplicate incoming TMLedgerData messages
  - Addresses RIPD-1869
- Improve logging related to ledger acquisition
- Class "CanProcess" to keep track of processing of distinct items

---------

Co-authored-by: Valentin Balaschenko <13349202+vlntb@users.noreply.github.com>
```

src/xrpld/app/ledger/InboundLedger.h
```diff
@@ -196,6 +196,24 @@ class InboundLedger final : public TimeoutCounter,
...
+    switch (reason)
+    {
+        case HISTORY:
+            return "HISTORY";
+        case GENERIC:
+            return "GENERIC";
+        case CONSENSUS:
+            return "CONSENSUS";
+        default:
+            assert(false);
```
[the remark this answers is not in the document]
ximinez: > please replace with `UNREACHABLE` and a nice & short description why (see `CONTRIBUTING.md` , section " Contracts and instrumentation" for naming guideline)

Done.

src/xrpld/app/ledger/detail/InboundLedger.cpp
ximinez: > please replace with `XRPL_ASSERT` and a short description why; see `CONTRIBUTING.md` for contracts and instrumentation naming guidelines

Done

src/xrpld/app/misc/HashRouter.cpp
ximinez: The callbacks are very fast, but now I don't remember why I wanted to do this under lock.
PeerImp.cpp line 3528
```
        std::set<HashRouter::PeerShortID> peers;
        app_.getHashRouter().forEachPeer(
            mHash, [&](HashRouter::PeerShortID const& peerID) {
                // callback is called under lock, so finish as fast as
                // possible.
                peers.insert(peerID);
                return true;
            });
```
Maybe I wanted avoid making a copy of the `std::set`? But this is effectively making a copy of the set anyway, so that's dumb. I'd guess I originally wrote it with the loop contents inside the callback, but decided that was too slow, and didn't think before just copying.

Bronek (commented):
Thank you @ximinez , this is just to confirm that you have covered all the feedback I had for this PR, and I particularly like how you changed `forEachPeer` into `getPeers` and also paid attention to contract names. Nice !

PR #5127, opened 2024-09-11
buffer Underflow vulnarability assertion check fix for RFC1751.cpp

nathanogaga118:
an assertion check fix has been  updated with the code 

https://github.com/nathanogaga118/rippled/blob/develop/src/libxrpl/crypto/RFC1751.cpp

Key Changes and fixes:

Assertions Added:

In btoe: assert(strData.size() >= 8 && "strData must be at least 8 bytes long");

In getEnglishFromKey: assert(strKey.size() >= 16 && "strKey must be at least 16 bytes long");

These assertions ensure that the preconditions for the btoe function are met, preventing potential buffer underflow issues.

### DIFF
--- a/src/libxrpl/crypto/RFC1751.cpp
+++ b/src/libxrpl/crypto/RFC1751.cpp
@@ -1,6 +1,6 @@
 //------------------------------------------------------------------------------
 /*
-    This file is part of rippled: https://github.com/ripple/rippled
+ This file is part of rippled: https://github.com/ripple/rippled
     Copyright (c) 2012, 2013 Ripple Labs Inc.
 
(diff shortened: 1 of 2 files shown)


Bronek:
Please kindly re-format this PR, as per to https://github.com/XRPLF/rippled/blob/develop/CONTRIBUTING.md#formatting  Note, we are using old clang-format-10, but it's trivial to setup with the help of containerised ubuntu:20.04 (https://packages.ubuntu.com/focal/clang-format-10 )

Bronek:
Also, this PR seems to be removing existing comments from the codebase in a rather arbitrary manner, please do not do that without a good reason.

Bronek:
Closed due to author's inactivity

PR #5135, opened 2024-09-15
SQLite: Allow configurable sqlite database pragma variables.

dangell7:
### Type of Change

- [ ] Documentation update
- [ ] Release

### API Impact

### DIFF
--- a/cfg/rippled-example.cfg
+++ b/cfg/rippled-example.cfg
@@ -1107,7 +1107,7 @@
 #                           "safety_level" setting.
 #
 #       page_size           Valid values: integer (typically 512, 1024, 2048, 4096)
-#                           The default is 1024 bytes. This setting determines
+#                           The default is 4096 bytes. This setting determines
 #                           the size of a page in the database file.
 #
 #       journal_size_limit  Valid values: integer
@@ -1116,7 +1116,7 @@
 #                           reached, older entries will be deleted.
 #
 #       max_page_count      Valid values: integer
-#                           The default is 1073741823. This setting limits
+#                           The default is 4294967294. This setting limits
 #                           the total number of pages in the database. If the
 #                           limit is reached, the database will not allow
 #                           additional pages to be allocated.
--- a/cfg/rippled-example.cfg
+++ b/cfg/rippled-example.cfg
@@ -1106,7 +1106,7 @@
 #                           This setting may not be combined with the
 #                           "safety_level" setting.
 #
-#       page_size           Valid values: integer (typically 512, 1024, 2048, 4096)
+#       page_size           Valid values: integer (MUST be power of 2 between 512 and 65536)
 #                           The default is 4096 bytes. This setting determines
 #                           the size of a page in the database file.
 #
--- a/cfg/rippled-example.cfg
+++ b/cfg/rippled-example.cfg
@@ -1115,11 +1115,6 @@
 #                           the size of the journal file. When the limit is
 #                           reached, older entries will be deleted.
 #
-#       max_page_count      Valid values: integer
-#                           The default is 4294967294. This setting limits
-#                           the total number of pages in the database. If the
-#                           limit is reached, the database will not allow
-#                           additional pages to be allocated.
 #
 #-------------------------------------------------------------------------------
 #
--- a/cfg/rippled-example.cfg
+++ b/cfg/rippled-example.cfg
@@ -1108,13 +1108,13 @@
 #
 #       page_size           Valid values: integer (MUST be power of 2 between 512 and 65536)
 #                           The default is 4096 bytes. This setting determines
-#                           the size of a page in the database file.
+#                           the size of a page in the transaction.db file.
 #                           See https://www.sqlite.org/pragma.html#pragma_page_size 
 #                           for more details about the available options.
 #
 #       journal_size_limit  Valid values: integer
 #                           The default is 1582080. This setting limits
-#                           the size of the journal file. When the limit is
+#                           the size of the journal for transaction.db file. When the limit is
 #                           reached, older entries will be deleted.
 #                           See https://www.sqlite.org/pragma.html#pragma_journal_size_limit
 #                           for more details about the available options.
--- a/cfg/rippled-example.cfg
+++ b/cfg/rippled-example.cfg
@@ -1109,11 +1109,15 @@
 #       page_size           Valid values: integer (MUST be power of 2 between 512 and 65536)
 #                           The default is 4096 bytes. This setting determines
 #                           the size of a page in the database file.
+#                           See https://www.sqlite.org/pragma.html#pragma_page_size 
+#                           for more details about the available options.
 #
 #       journal_size_limit  Valid values: integer
 #                           The default is 1582080. This setting limits
 #                           the size of the journal file. When the limit is
 #                           reached, older entries will be deleted.
+#                           See https://www.sqlite.org/pragma.html#pragma_journal_size_limit
+#                           for more details about the available options.
 #
 #
 #-------------------------------------------------------------------------------
--- a/cfg/rippled-example.cfg
+++ b/cfg/rippled-example.cfg
@@ -1106,6 +1106,21 @@
 #                           This setting may not be combined with the
 #                           "safety_level" setting.
 #
+#       page_size           Valid values: integer (typically 512, 1024, 2048, 4096)
+#                           The default is 1024 bytes. This setting determines
+#                           the size of a page in the database file.
+#
+#       journal_size_limit  Valid values: integer
+#                           The default is 1582080. This setting limits
+#                           the size of the journal file. When the limit is
+#                           reached, older entries will be deleted.
+#
+#       max_page_count      Valid values: integer
+#                           The default is 1073741823. This setting limits
+#                           the total number of pages in the database. If the
+#                           limit is reached, the database will not allow
+#                           additional pages to be allocated.
+#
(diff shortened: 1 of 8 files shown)


vvysokikh1:
hey, I've been testing this PR and I can't make page_size work. It always sets it to 4096 with brand new transaction.db file. Does it work for you?

vvysokikh1 (changes requested):
As mentioned in previous comment, this change doesn't actually work. I've debugged the main problem and here're some nitpicks mostly

src/xrpld/core/DatabaseCon.h
```diff
@@ -114,7 +116,7 @@ class DatabaseCon
     DatabaseCon(
         Setup const& setup,
         std::string const& dbName,
-        std::array<char const*, N> const& pragma,
+        std::array<std::string, N> const& pragma,
```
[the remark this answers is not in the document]
dangell7: It looks like setup is only used to specify the dataDir. If we want to change that then thats fine too.

cfg/rippled-example.cfg
```diff
@@ -1106,6 +1106,21 @@
...
+#       page_size           Valid values: integer (typically 512, 1024, 2048, 4096)
+#                           The default is 4096 bytes. This setting determines
+#                           the size of a page in the database file.
+#
+#       journal_size_limit  Valid values: integer
+#                           The default is 1582080. This setting limits
+#                           the size of the journal file. When the limit is
+#                           reached, older entries will be deleted.
+#
+#       max_page_count      Valid values: integer
```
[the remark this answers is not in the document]
dangell7: Thats fine

dangell7:
> hey, I've been testing this PR and I can't make page_size work. It always sets it to 4096 with brand new transaction.db file. Does it work for you?
> 
> UPD: I think I've found the problem after debugging it for some time. Looks like the following code has an issue:
> 
> ```
>     template <std::size_t N, std::size_t M>
>     DatabaseCon(
>         boost::filesystem::path const& pPath,
>         std::vector<std::string> const* commonPragma,
>         std::array<std::string, N> const& pragma,
>         std::array<char const*, M> const& initSQL,
>         beast::Journal journal)
>         : session_(std::make_shared<soci::session>()), j_(journal)
>     {
>         open(*session_, "sqlite", pPath.string());
> 
>         if (commonPragma)
>         {
>             for (auto const& p : *commonPragma) <<< some of the statement here creates a page in DB file, lockoing it to default 4096
>             {
>                 soci::statement st = session_->prepare << p;
>                 st.execute(true);
>             }
>         }
>         for (auto const& p : pragma) <<< This is the place where we set page_size
>         {
>             soci::statement st = session_->prepare << p;
>             st.execute(true);
>         }
>         for (auto const& sql : initSQL)
>         {
>             soci::statement st = session_->prepare << sql;
>             st.execute(true);
>         }
>     }
> ```
> 
> Changing the ordering of commonPragma and pragma helps to set the correct page_size. I'm not entirely sure this is correct solution though
> 
> UPD2: I understand this issue is not directly linked to your PR so I can take over the fix in this/other branch if you don't want to bother

Thank you for looking into it. You're welcome to take over.

src/xrpld/core/detail/DatabaseCon.cpp
```diff
@@ -237,6 +237,40 @@ setup_DatabaseCon(Config const& c, std::optional<beast::Journal> j)
...
+    int64_t page_size = 4096;
+    int64_t journal_size_limit = 1582080;
+    int64_t max_page_count = 4294967294;
+
+    if (c.exists("sqlite"))
+    {
+        auto& s = c.section("sqlite");
+        set(page_size, "page_size", s);
+        set(journal_size_limit, "journal_size_limit", s);
+    }
```
Bronek, line 258: needs to enforce valid `page_size` values, as per https://www.sqlite.org/pragma.html#pragma_page_size

src/xrpld/core/detail/DatabaseCon.cpp
```diff
@@ -253,8 +261,16 @@ setup_DatabaseCon(Config const& c, std::optional<beast::Journal> j)
     if (c.exists("sqlite"))
     {
         auto& s = c.section("sqlite");
-        set(page_size, "page_size", s);
         set(journal_size_limit, "journal_size_limit", s);
+        if (!isPageSizeValid(page_size))
+            Throw<std::runtime_error>(
+                "Invalid page_size. Must be between 512 and 65536.");
+
+        if (!isPageSizePowerOfTwo(page_size))
```
vvysokikh1, line 269: I assume these checks have to be after `set(page_size, "page_size", s);`?

vvysokikh1 (approved):
I've tested the last change and looks good to me. Ideally I would prefer the tests to check those <512 and >65536 and not power of 2 values but it's good to go anyways

dangell7:
> I've tested the last change and looks good to me. Ideally I would prefer the tests to check those <512 and >65536 and not power of 2 values but it's good to go anyways

Yeah tests are a must before this is merged. I'll write them today.

src/xrpld/core/detail/DatabaseCon.cpp
```diff
@@ -237,6 +237,55 @@ setup_DatabaseCon(Config const& c, std::optional<beast::Journal> j)
...
+    }
+
+    setPragma(setup.txPragma[0], "page_size", page_size);
+    setPragma(setup.txPragma[1], "journal_size_limit", journal_size_limit);
+    setPragma(setup.txPragma[2], "max_page_count", max_page_count);
+
+#if (ULONG_MAX > UINT_MAX) && !defined(NO_SQLITE_MMAP)
+    setup.txPragma[3] = "PRAGMA mmap_size=17179869184;";
+#else
+    // Provide an explicit `no-op` SQL statement
```
Bronek, line 282: I'd prefer is we switched the type of `setup.txPragmas` to `std::vector<std::string>` rather than use a dummy statement.

Bronek (changes requested):

src/xrpld/core/detail/DatabaseCon.cpp
Bronek: I think the reason why we used `std::array` was for `constexpr` initialisation inside `DBInit.h`; this is out of the window, now that we are building the set of init pragmas in runtime, so there is no point sticking to the limitations of `std::array` either.

src/xrpld/core/detail/DatabaseCon.cpp
```diff
@@ -237,6 +237,55 @@ setup_DatabaseCon(Config const& c, std::optional<beast::Journal> j)
     }
     setup.useGlobalPragma = true;
 
+    auto setPragma =
+        [](std::string& pragma, std::string const& key, int64_t value) {
+            pragma = "PRAGMA " + key + "=" + std::to_string(value) + ";";
+        };
+
+    auto isPageSizeValid = [](int64_t page_size) {
```
Bronek, line 245: this lambda is not needed, the condition should just be an inline `if` statement

src/xrpld/core/detail/DatabaseCon.cpp
```diff
@@ -237,6 +237,55 @@ setup_DatabaseCon(Config const& c, std::optional<beast::Journal> j)
...
+        return (page_size >= 512 && page_size <= 65536);
+    };
+
+    auto isPageSizePowerOfTwo = [](int64_t page_size) {
```
Bronek, line 249: same here - this lambda is not needed, the condition should be just an inline `if` statement

src/xrpld/core/detail/DatabaseCon.cpp
Bronek: Actually, we can keep `std::array<std::string>` ; the whole conditional is pointless and we can remove the `#ifdef #else #endif`.

We can safely assume that the `#if` condition is true because we do not support 32bit builds; the `NO_SQLITE_MMAP` is never passed by the cmake files and this would only matter if someone was trying to use version of sqlite3 older than [version 3.7.17, released in 2013](https://www.sqlite.org/releaselog/3_7_17.html).

dangell7: Thank you for taking the time to look that up and review.

src/xrpld/core/detail/DatabaseCon.cpp
```diff
@@ -276,16 +268,6 @@ setup_DatabaseCon(Config const& c, std::optional<beast::Journal> j)
     setPragma(setup.txPragma[1], "journal_size_limit", journal_size_limit);
     setPragma(setup.txPragma[2], "max_page_count", max_page_count);
 
-#if (ULONG_MAX > UINT_MAX) && !defined(NO_SQLITE_MMAP)
-    setup.txPragma[3] = "PRAGMA mmap_size=17179869184;";
```
Bronek: sorry if I was unclear; I meant in https://github.com/XRPLF/rippled/pull/5135#discussion_r1764989451 to change this as-if the `#if` condition was always true, which means to keep this line and remove the directive `#if` above and the block '#else .... #endif` below. This means that txPragma will always have size 4, not 3

Bronek (changes requested):

Bronek (approved):
Just two more minor changes in the configuration documentation section, this aside this is good.

Bronek:
@dangell7 can you please merge into your branch from `develop` branch ? Please remember to use `-S`. Thanks !

dangell7:
@Bronek yes thank you! LGTM

PR #5140, opened 2024-09-17
Fix order of SQLite PRAGMAS to allow configurable page_size

vvysokikh1:

Fixed the order of application of pragmas when opening SQLite connection. Due to specific pragmas set with `commonPragmas`, it looks like a page gets allocated in the DB file BEFORE `page_size` is set, thus locking the page_size to default value 4096.

### Context of Change
This issue has no effect before #5135 is merged.

### Type of Change

## Test Plan
1. Run rippled creating brand new transaction.db file with #5135, setting rippled.cfg to have the following:
```
...
[sqlite]
page_size = 8192
...
```
2. After a couple of seconds you should be able to run sqlite command to check the page_size of new transaction.db file:
```
sqlite3 <path>/transaction.db 'PRAGMA page_size;'
```

### DIFF
--- a/src/xrpld/core/DatabaseCon.h
+++ b/src/xrpld/core/DatabaseCon.h
@@ -206,6 +206,12 @@ private:
     {
         open(*session_, "sqlite", pPath.string());
 
+        for (auto const& p : pragma)
+        {
+            soci::statement st = session_->prepare << p;
+            st.execute(true);
+        }
+
         if (commonPragma)
         {
             for (auto const& p : *commonPragma)
@@ -214,11 +220,7 @@ private:
                 st.execute(true);
             }
         }
-        for (auto const& p : pragma)
-        {
-            soci::statement st = session_->prepare << p;
-            st.execute(true);
-        }
+
         for (auto const& sql : initSQL)
         {
             soci::statement st = session_->prepare << sql;


Bronek (approved):
LGTM

Bronek:
For readers of this PR:

* this is about swapping the order of execution of SQLite `PRAGMA` statements in `commonPragma` parameter and `pragma` parameter 
* `commonPragma` is either `nullptr` or `globalPragma` initialised in [xrpld/core/detail/DatabaseCon.cpp](https://github.com/vvysokikh1/rippled/blob/sqlite_pragma_ordering/src/xrpld/core/detail/DatabaseCon.cpp#L120) (look for `result->emplace_back`)
* the actual `PRAGMA` statements are constructed from the three strings `CommonDBPragma...` in [xrpld/app/main/DBInit.h](https://github.com/vvysokikh1/rippled/blob/sqlite_pragma_ordering/src/xrpld/app/main/DBInit.h#L32)
* the problem this PR solves is caused by `PRAGMA journal_mode` which creates journal file - hence committing `page_size` to the database, whatever it is at that point
* so if we want a non-default `page_size`, it needs to be set _before_ `PRAGMA journal_mode`
* we set `page_size` in `pragma` parameter , hence it needs to be processed before `commonPragma` parameter
* we have not noticed the problem until today because the hardcoded `page_size` [happens to be 4KB](https://github.com/vvysokikh1/rippled/blob/sqlite_pragma_ordering/src/xrpld/app/main/DBInit.h#L79), which is identical as the [SQLite default](https://www.sqlite.org/pragma.html#pragma_page_size)
* however there's a PR in flight #5135 to make `page_size` configurable and that will, obviously, do not work

FWIW `globalPragma` is [`static`](https://github.com/vvysokikh1/rippled/blob/sqlite_pragma_ordering/src/xrpld/core/DatabaseCon.h#L104) which is a bad code smell, but changing it belongs to a different PR (refactoring only)

intelliot (approved):
Important to move these pragmas first so that `PRAGMA page_size` is executed before `PRAGMA journal_mode` (which accesses the database and commits the page_size)

PR #5145, opened 2024-09-20
test: enable unit tests to work with variable reference fee

vvysokikh1:
High Level Overview of Change
This PR is a continuation of work done in: #5118 
Fix remaining unit tests to be able to process reference fee value other than 10.

Context of Change
In preparation for potential reference fee change we would like to verify that fee change works as expected. The first step is to fix all unit tests to be able to work with different reference fee values.

Type of Change
[x ] Tests (you added tests for code that already exists, or your new feature included in this PR)
API Impact
None

Test Plan
Tested following reference fee values: 10, 20, 50, 100, 200, 500, 1000 by changing https://github.com/vvysokikh1/rippled/blob/2f432e812cb773048530ebfaf2e0e6def51e3cc2/src/test/jtx/impl/envconfig.cpp#L47


Bronek:
> Test Plan Tested following reference fee values: 10, 20, 50, 100, 200, 500, 1000 by changing https://github.com/vvysokikh1/rippled/blob/2f432e812cb773048530ebfaf2e0e6def51e3cc2/src/test/jtx/impl/envconfig.cpp#L47

Can you pls add a command line option to set this value ? It will probably go somewhere in app/main/Main.cpp (look for sections guarded by `#ifdef ENABLE_TESTS` )

vvysokikh1:
> > Test Plan Tested following reference fee values: 10, 20, 50, 100, 200, 500, 1000 by changing https://github.com/vvysokikh1/rippled/blob/2f432e812cb773048530ebfaf2e0e6def51e3cc2/src/test/jtx/impl/envconfig.cpp#L47
> 
> Can you pls add a command line option to set this value ? It will probably go somewhere in app/main/Main.cpp (look for sections guarded by `#ifdef ENABLE_TESTS` )

This is actually quite tricky to do in runtime. Almost all tests allocate the `Env' and `Config' independently and it would require significant effort to propagate the fee value from startup params.

Feasible option is to make it compile time setting in CMake

Bronek:
> > > Test Plan Tested following reference fee values: 10, 20, 50, 100, 200, 500, 1000 by changing https://github.com/vvysokikh1/rippled/blob/2f432e812cb773048530ebfaf2e0e6def51e3cc2/src/test/jtx/impl/envconfig.cpp#L47
> > 
> > 
> > Can you pls add a command line option to set this value ? It will probably go somewhere in app/main/Main.cpp (look for sections guarded by `#ifdef ENABLE_TESTS` )
> 
> This is actually quite tricky to do in runtime. Almost all tests allocate the `Env` and `Config` independently and it would require significant effort to propagate the fee value from startup params.
> 
> Feasible option is to make it compile time setting in CMake

Yeah in that case just pls add an option to CMake (files `RippledSettings.cmake` and `RippledCore.cmake`); it is important that we can run unit tests with an arbitrary fee, so the new features do not revert to the hardcoded default 10.

src/test/app/Oracle_test.cpp
```diff
@@ -70,38 +80,47 @@ struct Oracle_test : public beast::unit_test::suite
...
                         {"XRP", "AUD", 740, 1},
                     },
+                .fee = static_cast<int>(env.current()->fees().base.drops()),
                 .err = ter(tecINSUFFICIENT_RESERVE)});
         }
 
         {
             Env env(*this);
+            auto const baseFee =
+                static_cast<int>(env.current()->fees().base.drops());
```
bthomee, line 90: If you move this to the top of the test, then all `Oracle` instantiations can use `.fee = baseFee`. Right now the code uses an inconsistent mix-and-match.

src/test/app/Oracle_test.cpp
```diff
@@ -70,38 +80,47 @@ struct Oracle_test : public beast::unit_test::suite
                         {"XRP", "CAD", 740, 1},
                         {"XRP", "AUD", 740, 1},
                     },
+                .fee = static_cast<int>(env.current()->fees().base.drops()),
```
bthomee, line 83: Looking at `src/test/jtx/Oracle.h` and `src/test/jtx/impl/Oracle.cpp`, when setting Oracle args they won't be updated if they are not provided.

So, here and below, you don't need to additionally include `.fee = static_cast<int>(env.current()->fees().base.drops())` in these `oracle.set` calls, unless the fee is different from what the oracle was initially created with.

This applies to both `set` using `CreateArgs` or `UpdateArgs` as input, since they end up calling the same function. It may also apply to the `remove` calls too, where including the same base fee as was already in use is unnecessary.

src/test/rpc/Transaction_test.cpp
```diff
@@ -827,7 +830,10 @@ class Transaction_test : public beast::unit_test::suite
         using namespace test::jtx;
         using std::to_string;
 
-        Env env{*this};
+        Env env{*this, envconfig([](std::unique_ptr<Config> cfg) {
+                    cfg->FEES.reference_fee = 10;
+                    return cfg;
+                })};
```
bthomee, line 836: I've seen so many copies of this function in so many different test files. Can this be moved into a common place, such as `src/test/jtx/Env.h`, so there's only a single place where a default `Env` is instantiated?

src/test/app/AMMExtended_test.cpp
```diff
@@ -3498,51 +3524,23 @@ struct AMMExtended_test : public jtx::AMMTest
...
-                JsonOptions::none)[sfAffectedNodes.fieldName];
-            if (!BEAST_EXPECT(checkArraySize(affected, 2u)))
-                return;
-            auto ff =
-                affected[1u][sfModifiedNode.fieldName][sfFinalFields.fieldName];
-            BEAST_EXPECT(
-                ff[sfLowLimit.fieldName] ==
-                G1["USD"](0).value().getJson(JsonOptions::none));
-            BEAST_EXPECT(ff[jss::Flags].asUInt() & lsfLowFreeze);
-            BEAST_EXPECT(!(ff[jss::Flags].asUInt() & lsfHighFreeze));
```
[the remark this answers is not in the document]
vvysokikh1: when fee changes, the layout of `affected` json changes as well. It's not easy to programmatically figure it out (we need to know correct index), but also this check if completely irrelevant, as such tests would make sense in freeze_test, not here.

src/test/app/MPToken_test.cpp
```diff
@@ -1863,7 +1863,9 @@ class MPToken_test : public beast::unit_test::suite
 
         Account const alice{"alice"};
 
-        Env env{*this, features};
+        auto cfg = envconfig();
+        cfg->FEES.reference_fee = 10;
```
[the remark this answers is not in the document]
vvysokikh1: because once fee changes (the one in envconfig != 10), this is the way of returning it back to 10 for test cases that need it.

src/test/app/LedgerMaster_test.cpp
```diff
@@ -33,6 +33,8 @@ class LedgerMaster_test : public beast::unit_test::suite
         using namespace jtx;
         return envconfig([&](std::unique_ptr<Config> cfg) {
```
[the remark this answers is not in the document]
vvysokikh1: I'm not sure what you mean. I don't want each test to have a different fee. They all should be passing the same fee except few that calculate hashes, rely on ordering too much. Those tests have fee 10 set in the way of `cfg->FEES.reference_fee = 10;`

src/test/app/Oracle_test.cpp
vvysokikh1: baseFee is a part of Env. Env is create one line above. I don't have any other way to get fee earlier

src/test/app/Oracle_test.cpp
vvysokikh1: If you look at `CreateArgs` struct, it initializes fee with 10 (hardcoded). And there's no way of initializing it in a different way as it's a part of Env. 

This is the only way to construct `CreateArgs` (and Oracle) with correct fee.

src/test/app/Offer_test.cpp
```diff
@@ -2343,6 +2378,7 @@ class OfferBaseUtil_test : public beast::unit_test::suite
         Env env{*this, features};
 
         env.fund(XRP(10000000), gw);
+        env.close();
 
         // The fee that's charged for transactions
         auto const f = env.current()->fees().base;
```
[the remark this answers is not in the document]
vvysokikh1: this is existing code and I prefer not to mess it. Plus there's tons of places where this is the case so it would introduce line 100+ lines of additional change. I can do that no problem but was hesitant due to having no clear benefits.

src/test/rpc/Transaction_test.cpp
vvysokikh1: Because this is the test with hardcoded fee 10 that I want to always stay 10, independently of what fee is set by configuration

src/test/app/PayChan_test.cpp
```diff
@@ -2017,6 +2016,7 @@ struct PayChan_test : public beast::unit_test::suite
             rmAccount(env, bob, carol);
             BEAST_EXPECT(!env.closed()->exists(keylet::account(bob.id())));
 
+            auto const feeDrops = env.current()->fees().base;
```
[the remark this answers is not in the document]
vvysokikh1: this is (kind of) consistent with this particular file

src/test/app/Oracle_test.cpp
```diff
@@ -42,15 +42,22 @@ struct Oracle_test : public beast::unit_test::suite
             Account const bad("bad");
             env.memoize(bad);
             Oracle oracle(
-                env, {.owner = bad, .seq = seq(1), .err = ter(terNO_ACCOUNT)});
+                env,
+                {.owner = bad,
+                 .seq = seq(1),
+                 .fee = static_cast<int>(env.current()->fees().base.drops()),
```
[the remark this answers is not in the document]
vvysokikh1: What is the motivation? I've seen you doing that in other PR, but I struggle to understand what is the purpose of this.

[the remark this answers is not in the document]
bthomee: Per @ximinez on a different PR:
> we have two sets of overrides: [safe_cast and unsafe_cast](https://github.com/XRPLF/rippled/blob/develop/include/xrpl/basics/safe_cast.h) that specify at compile time whether or not all the values of the source type can fit into the destination type.

Now, if there's little value gained, I'd rather see us rip it out completely everywhere, rather than having it only partially being used. I personally find them not very useful, especially because data types have different sizes on different platforms, so in the end what is a `safe_cast` on one platform ends up being an `unsafe_cast` on another, which then doesn't compile on one of them, and the end result is just using `static_cast` instead. It took me tons of whack-a-mole commits to figure out which of these functions worked on all three platforms.

@ximinez: Any strong opinions on maintaining use of (un)safe cast whenever possible and preferring its use if it compiles on all platforms, or can we just get rid of it (in a separate PR) as the benefits do not outweigh the pains of getting code to compile everywhere when using them?

src/test/app/Oracle_test.cpp
vvysokikh1: I usually create a variable if this value is needed more than once. Here these test cases only use it once so it kind of doesn't make sense for it to be a variable.

src/test/app/Oracle_test.cpp
vvysokikh1: That's a pretty big change unfortunately and I'm not keen on doing such refactoring.

src/test/app/Offer_test.cpp
```diff
@@ -2428,13 +2464,13 @@ class OfferBaseUtil_test : public beast::unit_test::suite
...
                 env(trust(gw, acct["USD"](1)));
+            env.close();
 
             // Optionally pre-establish a trustline between acct and gw.
             // Note this is not really part of the test, so we expect there
             // to be enough XRP reserve for acct to create the trust line.
             if (t.preTrust == acctPreTrust)
                 env(trust(acct, USD(1)));
-
             env.close();
```
bthomee, line 2474: These three `env.close()` calls here and above (GitHub only allowed me to select a limited number of lines), should they be inside the `if` statement, since there's no point in calling it if the `env(trust(...))` call isn't made?

bthomee (commented):
Almost good to go. I'm wondering, in a possible follow-up PR, whether we can make it a lot harder to forget to close the test ledger by making it explicit. For instance, the various env function calls, such as `fund`, `offer`, and `trust` could take an extra parameter, `bool close = true`, which would then call `env.close()` unless the argument is explicitly set to false.

src/test/app/Offer_test.cpp
vvysokikh1: yeah they could be inside of the if-block, but at the same time there's no harm if empty ledger closes, so I'm not worried about this

src/test/app/Offer_test.cpp
```diff
@@ -756,8 +760,10 @@ class OfferBaseUtil_test : public beast::unit_test::suite
             auto const xrpOffer2 = XRP(500);
 
             env.fund(r + f + xrpOffer, bob);
+
             env(offer(bob, usdOffer2, xrpOffer2), ter(tesSUCCESS));
             env.fund(r + f, alice);
+
```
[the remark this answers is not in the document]
vvysokikh1: ok, I now have a better context to answer.

In this whole test, there's no .close() at all. So, whichever transactions happen, they happen in the order the code is written. So basically what you see is what you get. No problems with this.

The problem would happen on the first .close(). All transactions would be reapplied, and the ordering of this reapplication would be dependent on fee and other factors.

vvysokikh1:
> Almost good to go. I'm wondering, in a possible follow-up PR, whether we can make it a lot harder to forget to close the test ledger by making it explicit. For instance, the various env function calls, such as `fund`, `offer`, and `trust` could take an extra parameter, `bool close = true`, which would then call `env.close()` unless the argument is explicitly set to false.

From the moment we enable different reference fee testing, it will be not easy to push the tests that break this since they would fail on CI. We could technically add some additional precautions to make ledger close after fund, but I'm not capable to answer how much efforts we would need to make it work. There's plenty of tests that don't close ledger at all so modifying all those would be grind job

bthomee (approved):

PR #5157, opened 2024-10-15
Reformat code with clang-format-18

thejohnfreeman:
This is a second attempt at reformatting our first-party sources with a newer version of clang-format. The first was #4938. That PR was created shortly before Ubuntu LTS 24.04 was released, and internally we decided to hold off on a flag commit like this until the new LTS was available in GitHub Actions, letting us easily update to an even newer version of clang-format (18 instead of 15). That day has come.


Bronek (approved):

vvysokikh1 (approved):

mvadari:
Confirmed that the pre-commit hook still works.

PR #5159, opened 2024-10-16
test: enable compile time param to change reference fee value

vvysokikh1:
### Type of Change

- [ ] Documentation update
- [ ] Release

### API Impact


.github/workflows/nix.yml
```diff
@@ -163,6 +163,60 @@ jobs:
...
+        platform:
+          - linux
+        compiler:
+          - gcc
+        configuration:
+          - Debug
+        cmake-args:
+          - "-DUNIT_TEST_REFERENCE_FEE=200"
+          - "-DUNIT_TEST_REFERENCE_FEE=1000"
+    needs: dependencies
```
bthomee, line 179: Would it make sense to have this run after the other tests succeed? If there's an issue with the code and tests fail, we already have a lot of jobs running in parallel for nothing, and this would add two more.

.github/workflows/nix.yml
```diff
@@ -163,6 +163,60 @@ jobs:
...
+  reference-fee-test:
+    strategy:
+      fail-fast: false
+      matrix:
+        platform:
+          - linux
+        compiler:
+          - gcc
+        configuration:
+          - Debug
```
bthomee, line 175: I'd recommend setting this to Release to have these new jobs run faster - unless you think there's vital information we'd get from a debug build.

I suspect that a test failure only when using a non-standard fee will be relatively rare going forward, and either way our current testing framework doesn't offer any additional insights in debug builds compared with release builds.

vvysokikh1: it was release until today, Ed suggested to switch to debug.

.github/workflows/nix.yml
vvysokikh1: That would make the total run longer by > 10 minutes.

.github/workflows/nix.yml
ximinez: > @ximinez can you please elaborate?

Short answer is so that asserts are included. I'd want to know if any non-test code fails with an assert due to a different base fee value. In the "nix" tests, the Debug builds run just as fast or faster than Release builds. I think this is because debug doesn't add any significant runtime delays to tests, but release has overhead optimizing the build.

bthomee (approved):

PR #5164, opened 2024-10-25
Support AccountPermission XLS-74d and XLS-75d

yinyiqian1:
spec:
https://github.com/XRPLF/XRPL-Standards/discussions/217
https://github.com/XRPLF/XRPL-Standards/discussions/218

This PR includes the following changes:
1. Added AccountPermissionSet transaction.
    to enable an account(delegating account) delegate some permissions to another account(delegated account), so that the delegated account can send transactions on behalf of the delegating account.

2. Added Account_Permission ledger object
    The AccountPermissionSet transaction will create AccountPermission ledger object, with keylet(delegating account, delegated account)

3. Permission has two types:
     a. transaction level permission
     b. granular permission which will be part of a transaction

3. Added `sfOnBehalfOf` common field, which is optional
    this field is set when an account is sending transaction on behalf of another account

4. Added `sfDelegateSequence` which are optional
    to avoid double insertion, when a delegating transaction happens, we have to increment both of the delegating account's sequence and the delegated account's sequence. So sfDelegateSequence is used to track the delegating account's sequence.

5. Added `sfDelegateTicketSequence` which are optional
     we allow a delegated account to create ticket on behalf of the delegating account of he has the permission. In this case, the delegated account should be able to use the ticket he created on behalf of the delegating account. So sfDelegateTicketSequence is introduced.

6. Also added a wrapper class STTxDelegated, so that STTxDelegated[sfAccount] will return the delegated account if it's a delegating transaction, and  STTxDelegated[sfAccount] will return the account if not delegated as before.

7. a set of Granular Permissions `permissions` is added to ApplyContext. It contains the granular permissions enabled for that transaction. But if the transaction is fully delegated, then even there are granular permissions related to that transaction, the permissions will be cleared up. To be more specific:
   a. isDelegated=true && permissions not empty: only granular permissions delegated to the transaction
   b. isDelegated=true && permissions is empty:  the transaction is fully delegated

### Type of Change

- [ ] Documentation update
- [ ] Release

### API Impact


include/xrpl/protocol/Permissions.h
```diff
@@ -0,0 +1,82 @@
...
+    gpPaymentMint = 65545,
+
+    gpPaymentBurn = 65546,
+
+    gpMPTokenIssuanceLock = 65547,
+
+    gpMPTokenIssuanceUnlock = 65548,
+};
+
+class Permission
```
shawnxie999, line 61: should probably delete copy and copy assignment constructors

src/xrpld/app/tx/detail/AccountPermissionSet.cpp
```diff
@@ -0,0 +1,161 @@
...
+        STAmount const reserve{ctx_.view().fees().accountReserve(
+            sleOwner->getFieldU32(sfOwnerCount) + 1)};
+
+        if (mPriorBalance < reserve)
+            return tecINSUFFICIENT_RESERVE;
+
+        sle = std::make_shared<SLE>(accountPermissionKey);
+        auto const& permissions = ctx_.tx.getFieldArray(sfPermissions);
+        sle->setFieldArray(sfPermissions, permissions);
+        auto page = ctx_.view().dirInsert(
```
shawnxie999, line 109:
```suggestion
        auto const page = ctx_.view().dirInsert(
```

src/xrpld/app/tx/detail/AccountPermissionSet.cpp
```diff
@@ -0,0 +1,161 @@
...
+{
+    auto const sleOwner = ctx_.view().peek(keylet::account(account_));
+    if (!sleOwner)
+        return {tefINTERNAL};  // LCOV_EXCL_LINE
+
+    auto const accountPermissionKey =
+        keylet::accountPermission(account_, ctx_.tx[sfAuthorize]);
+
+    auto sle = ctx_.view().peek(accountPermissionKey);
+    if (!sle)
```
shawnxie999, line 98: can be refactored like
   ```
   auto sle = ctx_.view().peek(accountPermissionKey);
   if (sle)
   {
        auto const& permissions = ctx_.tx.getFieldArray(sfPermissions);
        sle->setFieldArray(sfPermissions, permissions);
        ctx_.view().update(sle); 
        return tesSUCCESS;
   }
   STAmount const reserve{ctx_.view().fees().accountReserve(
            sleOwner->getFieldU32(sfOwnerCount) + 1)};
    ...
```

File: src/xrpld/app/tx/detail/AccountPermissionSet.cpp [hunk 4]
```diff
@@ -0,0 +1,161 @@
...
+
+    return tesSUCCESS;
+}
+
+TER
+AccountPermissionSet::doApply()
+{
+    auto const sleOwner = ctx_.view().peek(keylet::account(account_));
+    if (!sleOwner)
+        return {tefINTERNAL};  // LCOV_EXCL_LINE
```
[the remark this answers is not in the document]
yinyiqian1: updated

src/xrpld/app/tx/detail/AccountPermissionSet.cpp
yinyiqian1: updated

src/xrpld/app/tx/detail/AccountPermissionSet.cpp
yinyiqian1: updated

include/xrpl/protocol/Permissions.h
shawnxie999: does this still needs to be done?

src/xrpld/app/tx/detail/applySteps.cpp
```diff
@@ -294,15 +304,22 @@ preflight(
...
-    PreflightContext const pfctx(app, tx, rules, flags, j);
+    bool const isDelegated =
+        rules.enabled(featureAccountPermission) && tx[~sfOnBehalfOf];
+    AccountID const account = isDelegated ? *tx[~sfOnBehalfOf] : tx[sfAccount];
+    PreflightContext const pfctx(app, tx, rules, flags, account, j);
     try
     {
-        return {pfctx, invoke_preflight(pfctx)};
+        // if AccountPermission is not enabled, do not use OnBehalfOf field.
+        if (!rules.enabled(featureAccountPermission) && tx[~sfOnBehalfOf])
```
shawnxie999, line 314: so what's the behavior when someone submits a tx with sfOnBehalfOf but not enabled the amendment? what error code does it return. To me, i feel we should not be throwing a runtime error, because it's a completely acceptable input. pls consider performing this check in the `Transactor::preflight1` instead and return an code like `temDISABLED`

gregtatcam:
PR description should be added with a link to technical specifications if applicable. This is essential for reviewers not familiar with the subject.

yinyiqian1:
> PR description should be added with a link to technical specifications if applicable.

just added

yinyiqian1:
FYI: unit test failures are transient, which have passed locally. "ERR:Env Env::close() failed: no response from server"

src/xrpld/app/tx/detail/AMMDeposit.cpp
```diff
@@ -168,7 +168,7 @@ AMMDeposit::preflight(PreflightContext const& ctx)
 TER
 AMMDeposit::preclaim(PreclaimContext const& ctx)
 {
-    auto const accountID = ctx.tx[sfAccount];
+    auto const accountID = ctx.account;
```
gregtatcam, line 171: Should `auto const&`

src/xrpld/app/tx/detail/AMMWithdraw.cpp
```diff
@@ -170,7 +170,7 @@ tokensWithdraw(
 TER
 AMMWithdraw::preclaim(PreclaimContext const& ctx)
 {
-    auto const accountID = ctx.tx[sfAccount];
+    auto const accountID = ctx.account;
```
gregtatcam, line 173: Should be `auto const&`.

src/xrpld/app/tx/detail/CancelCheck.cpp
```diff
@@ -73,7 +73,7 @@ CancelCheck::preclaim(PreclaimContext const& ctx)
     {
         // If the check is not yet expired, then only the creator or the
         // destination may cancel the check.
-        AccountID const acctId{ctx.tx[sfAccount]};
+        AccountID const acctId{ctx.account};
```
gregtatcam, line 76: Can be a ref now:
   AccountID const& acctId = ctx.account;

src/xrpld/app/tx/detail/CancelOffer.cpp
```diff
@@ -53,7 +53,7 @@ CancelOffer::preflight(PreflightContext const& ctx)
 TER
 CancelOffer::preclaim(PreclaimContext const& ctx)
 {
-    auto const id = ctx.tx[sfAccount];
+    auto const id = ctx.account;
```
gregtatcam, line 56: Should be a const&

src/xrpld/app/tx/detail/Change.cpp
```diff
@@ -38,7 +38,7 @@ Change::preflight(PreflightContext const& ctx)
     if (!isTesSuccess(ret))
         return ret;
 
-    auto account = ctx.tx.getAccountID(sfAccount);
+    auto account = ctx.account;
```
gregtatcam, line 41: Don't need this temp. Can be simplified:

    if (ctx.account != beast::zero)

src/xrpld/app/tx/detail/Clawback.cpp
```diff
@@ -39,7 +39,7 @@ preflightHelper<Issue>(PreflightContext const& ctx)
     if (ctx.tx.isFieldPresent(sfHolder))
         return temMALFORMED;
 
-    AccountID const issuer = ctx.tx[sfAccount];
+    AccountID const issuer = ctx.account;
```
gregtatcam, line 42: Should be const&.

src/xrpld/app/tx/detail/Clawback.cpp
```diff
@@ -197,7 +197,7 @@ preclaimHelper<MPTIssue>(
 TER
 Clawback::preclaim(PreclaimContext const& ctx)
 {
-    AccountID const issuer = ctx.tx[sfAccount];
+    AccountID const issuer = ctx.account;
```
gregtatcam, line 200: Should be const&. Please check everywhere where an alias variable is used for `ctx.account`. It should be `AccountID const&`.

src/xrpld/app/tx/detail/MPTokenIssuanceSet.cpp
```diff
@@ -102,9 +102,21 @@ MPTokenIssuanceSet::doApply()
     std::uint32_t const flagsIn = sle->getFieldU32(sfFlags);
     std::uint32_t flagsOut = flagsIn;
 
-    if (txFlags & tfMPTLock)
+    bool bLock = txFlags & tfMPTLock;
+    bool bUnlock = txFlags & tfMPTUnlock;
+    if (ctx_.isDelegated && !ctx_.gpSet.empty())
+    {
+        // if gpSet is not empty, granular delegation is happening.
+        if (bLock && ctx_.gpSet.find(gpMPTokenIssuanceLock) == ctx_.gpSet.end())
```
gregtatcam, line 110: Can simplify here and below.

    if (bLock && !ctx.gpSet.contains(gpMPTokenIssuanceLock))

src/xrpld/app/tx/detail/Payment.cpp
```diff
@@ -341,6 +341,28 @@ Payment::doApply()
...
+    {
+        // If gpSet is not empty, granular delegation is happening.
+        // Currently we only support PaymentMint and PaymentBurn granular
+        // permission. PaymentMint means the sender is the issuer. PaymentBurn
+        // means the destination is the issuer.
+        bool authorized = false;
+        auto const amountIssue = dstAmount.issue();
+        if (isXRP(amountIssue))
+            return tecNO_AUTH;
+        if (amountIssue.account == account_ &&
```
gregtatcam, line 355: Can simplify here and below:

    if (amountIssue.account == account_ &&
            ctx_.gpSet.contains(gpPaymentMint))

src/xrpld/app/tx/detail/SetTrust.cpp
```diff
@@ -243,6 +243,26 @@ SetTrust::doApply()
...
+    if (ctx_.isDelegated && !ctx_.gpSet.empty())
+    {
+        granularDelegated = true;
+        // If granular permission is delegated under the TrustSet transaction.
+        // Currently we only support TrustlineAuthorize, TrustlineFreeze and
+        // TrustlineUnfreeze granular permission.
+        if (bSetNoRipple || bClearNoRipple || bQualityIn || bQualityOut)
+            return terNO_AUTH;
+        if (bSetAuth &&
+            ctx_.gpSet.find(gpTrustlineAuthorize) == ctx_.gpSet.end())
```
gregtatcam, line 256: Can simplify here and below:

    if (bSetAuth &&
            !ctx_.gpSet.contains(gpTrustlineAuthorize))

src/xrpld/app/tx/detail/InvariantCheck.cpp
```diff
@@ -458,6 +458,7 @@ LedgerEntryTypesMatch::visitEntry(
         switch (after->getType())
         {
             case ltACCOUNT_ROOT:
+            case ltACCOUNT_PERMISSION:
```
gregtatcam, line 461: I can't leave a comment on unchanged lines so commenting here for ValidClawback::finalize:924, which is this line:

     AccountID const issuer = tx.getAccountID(sfAccount);

The `issuer` in this case could be either `sfAccount` or `sfOnBehalfOf`, right?

gregtatcam (commented):
I took a pass and left a few comments.
My big concern is how `sfAccount` is obfuscated by `sfOnBehalfOf`. `sfAccount` pretty much should never be used in the transactors with some exceptions. It's easy to miss these kind of errors where `sfAccount` is used instead of the `account` included in the context.

src/xrpld/app/tx/detail/Clawback.cpp
yinyiqian1: since it only used once, I'll remove this line and simplify to
```
    if (ctx.account  == holder || isXRP(clawAmount) || clawAmount <= beast::zero)
        return temBAD_AMOUNT;
```

src/xrpld/app/tx/detail/InvariantCheck.cpp
yinyiqian1: This is missed out. I'll also check all the other similar places which is missed out.

yinyiqian1: @gregtatcam 
I also need to make change to AcceptedLedgerTx.cpp, LedgerToJson.cpp, and NetworkOPs.cpp, because they also check the account id when the transaction type is ttOFFER_CREATE. I'm working on this as well

include/xrpl/protocol/Permissions.h
```diff
@@ -0,0 +1,88 @@
...
+namespace ripple {
+
+/**
+ * We have transaction type permissions and granular type
+ * permissions. Since we will reuse the TransactionFormats to parse the
+ * Transaction Permissions, we only define the GranularPermissionType here.
+ */
+
+enum GranularPermissionType : std::uint32_t {
+    gpTrustlineAuthorize = 65537,
```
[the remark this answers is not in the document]
yinyiqian1: `gp` is to indicate it is a granular permission set's member. I'm ok with either way.

include/xrpl/protocol/detail/ledger_entries.macro
```diff
@@ -436,3 +436,16 @@ LEDGER_ENTRY(ltCREDENTIAL, 0x0081, Credential, ({
...
 }))
+
+/** A ledger object representing permissions an account has delegated to another account.
+
+    \sa keylet::accountPermission
+ */
+LEDGER_ENTRY(ltACCOUNT_PERMISSION, 0x0082, AccountPermission, ({
+    {sfAccount,              soeREQUIRED},
+    {sfAuthorize,            soeREQUIRED},
+    {sfPermissions,          soeREQUIRED},
```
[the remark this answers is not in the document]
yinyiqian1: Since being empty means deleting all the permissions, can we keep it as required? So that the user won't deleting the permissions by accident when they mistakenly leave out the Permissions field. Giving an empty field is more clear that they want to delete all the permissions.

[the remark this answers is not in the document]
gregtatcam: This is an on-ledger object, not the transaction. It's empty if there is no permissions so it could be optional. Can save a bit of space this way, right?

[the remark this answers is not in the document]
yinyiqian1: yes. make sense.

include/xrpl/protocol/Permissions.h
```diff
@@ -0,0 +1,88 @@
...
+private:
+    Permission();
+
+    std::unordered_map<std::string, GranularPermissionType>
+        granularPermissionMap;
+
+    std::unordered_map<GranularPermissionType, TxType> granularTxTypeMap;
+
+public:
+    static Permission const&
```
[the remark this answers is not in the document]
yinyiqian1: need to access private members in the functions, for example, `granularPermissionMap` and `granularTxTypeMap`

include/xrpl/protocol/Permissions.h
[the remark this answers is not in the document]
yinyiqian1: just deleted `gp`

include/xrpl/protocol/Permissions.h
yinyiqian1: just added

include/xrpl/protocol/Permissions.h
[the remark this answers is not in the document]
gregtatcam: They should be also static. There should be only one instance of those maps.

include/xrpl/protocol/STTxWr.h
```diff
@@ -0,0 +1,210 @@
...
+        : tx_(tx), isDelegated_(isDelegated)
+    {
+    }
+
+    STTx
+    getTx() const
+    {
+        return tx_;
+    }
+
```
gregtatcam, line 52: I think it also makes sense to add implicit conversion operator to `STTx const&`. Then don't need to call tx.getTx() in many cases. Can just pass tx.

gregtatcam: And perhaps change getTx() to something else, maybe getSTTx? tx.getTx() doesn't look right.

src/xrpld/app/tx/detail/Transactor.cpp
```diff
@@ -126,6 +130,10 @@ preflight1(PreflightContext const& ctx)
         ctx.tx.isFieldPresent(sfAccountTxnID))
         return temINVALID;
 
+    if (!ctx.rules.enabled(featureAccountPermission) &&
```
gregtatcam, line 133: This is already checked at the beginning of the function.

src/xrpld/app/tx/detail/Transactor.cpp
```diff
@@ -180,6 +188,51 @@ Transactor::calculateBaseFee(ReadView const& view, STTx const& tx)
     return baseFee + (signerCount * baseFee);
 }
 
+TER
+Transactor::checkPermissions(
+    ReadView const& view,
+    STTx const& tx,
+    std::unordered_set<GranularPermissionType>& permissions)
+{
+    if (!view.read(keylet::account(tx[sfOnBehalfOf])))
```
gregtatcam, line 197: Should be `exists` instead.

Should we check if `sfOnBehalfOf` is included in `tx`?  `checkPermissions` is so far only called in delegated but still might make sense to check.

src/xrpld/app/tx/detail/Transactor.cpp
```diff
@@ -180,6 +188,51 @@ Transactor::calculateBaseFee(ReadView const& view, STTx const& tx)
...
+        return terNO_ACCOUNT;
+
+    auto const accountPermissionKey =
+        keylet::accountPermission(tx[sfOnBehalfOf], tx[sfAccount]);
+    auto const sle = view.read(accountPermissionKey);
+    if (!sle)
+        return temMALFORMED;  // todo: change error code
```
gregtatcam, line 204: Change the error code?

src/xrpld/app/tx/detail/Transactor.cpp
```diff
@@ -247,6 +300,21 @@ TER
 Transactor::payFee()
 {
     auto const feePaid = ctx_.tx[sfFee].xrp();
+    if (ctx_.tx.isDelegated())
```
gregtatcam, line 303: Why is this needed? We  always need the account sender to update the balance. I.e. it's always `sfAccount`. And if it's not delegated then `account_` is `sfAccount`.

src/xrpld/rpc/handlers/LedgerEntry.cpp
```diff
@@ -834,6 +869,7 @@ doLedgerEntry(RPC::JsonContext& context)
         {jss::index, parseIndex, ltANY},
         {jss::account_root, parseAccountRoot, ltACCOUNT_ROOT},
         // TODO: add amendments
+        {jss::amm, parseAccountPermission, ltACCOUNT_PERMISSION},
```
gregtatcam, line 872: This is not `amm`, is it?

src/xrpld/app/ledger/detail/LedgerToJson.cpp
```diff
@@ -206,7 +206,9 @@ fillJsonTx(
     if ((fill.options & LedgerFill::ownerFunds) &&
         txn->getTxnType() == ttOFFER_CREATE)
     {
-        auto const account = txn->getAccountID(sfAccount);
+        auto const& account = txn->isFieldPresent(sfOnBehalfOf)
```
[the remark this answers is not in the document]
yinyiqian1: can we give it a more precise name like `getDelegatingOrOriginalAccountID()` or `getEffectiveAccountID()`?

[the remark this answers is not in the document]
gregtatcam: `getEffectiveAccountID()` sounds good.

src/xrpld/app/tx/detail/Transactor.cpp
yinyiqian1: how about `tecNO_PERMISSION` because the permission does not exist?

src/xrpld/rpc/handlers/LedgerEntry.cpp
yinyiqian1: forgot to change when copying :(

src/xrpld/app/tx/detail/Transactor.cpp
yinyiqian1: I removed the check here. But I still need to add
`if (!ctx_.tx.isDelegated()) mSourceBalance -= feePaid;`
to keep mSourceBalance value precise.

gregtatcam: Shouldn't this always be done for the sender account? Why is it needed to differentiate between delegated/non-delegated transaction.

yinyiqian1: Yes, I combined them. My updated code is like:
```
Transactor::payFee()
{
    auto const feePaid = ctx_.tx[sfFee].xrp();
        // whether the transaction is being delegated to another account or not,
        // the sender account will pay the fee.
        auto const sender = ctx_.tx.getSTTx().getAccountID(sfAccount);
        auto const sleSender = view().peek(keylet::account(sender));
        if (!sleSender)
            return tefINTERNAL;

        auto senderBalance = STAmount{(*sleSender)[sfBalance]}.xrp();
        senderBalance -= feePaid;
        sleSender->setFieldAmount(sfBalance, senderBalance);
        view().update(sleSender);

        if (!ctx_.tx.isDelegated())
            mSourceBalance -= feePaid;

        return tesSUCCESS;
}
```

yinyiqian1: Yes. updated code is like below. `mSouceBalance` is tracking the effective account's balance after fee, so if it's delegated, the effective account is not paying the fee because the sender will pay for it, then mSouceBalance should not be updated. That's the reason for adding this line: `
        if (!ctx_.tx.isDelegated())
            mSourceBalance -= feePaid;`
```
TER
Transactor::payFee()
{
    auto const feePaid = ctx_.tx[sfFee].xrp();
        // whether the transaction is being delegated to another account or not,
        // the sender account will pay the fee.
        auto const sender = ctx_.tx.getSTTx().getAccountID(sfAccount);
        auto const sleSender = view().peek(keylet::account(sender));
        if (!sleSender)
            return tefINTERNAL;

        auto senderBalance = STAmount{(*sleSender)[sfBalance]}.xrp();
        senderBalance -= feePaid;
        sleSender->setFieldAmount(sfBalance, senderBalance);
        view().update(sleSender);

        if (!ctx_.tx.isDelegated())
            mSourceBalance -= feePaid;

        return tesSUCCESS;
}
```

include/xrpl/protocol/detail/ledger_entries.macro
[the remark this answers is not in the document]
yinyiqian1: @gregtatcam  when there's no permissions, can we erase the SLE? so that we can still keep `sfPermissions` as required?

src/xrpld/app/tx/detail/Transactor.cpp
gregtatcam: Shouldn't `mSourceBalance` always be updated? `mSourceBalance` is owned by the sender and is used in some transactors, albeit minimally. So I don't see why it shouldn't be updated if this a delegated transaction.

include/xrpl/protocol/detail/ledger_entries.macro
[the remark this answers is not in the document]
gregtatcam: Do you mean delete SLE? I don't think we should delete it if there is no permissions. Where is a transaction for this. Do you see a problem with making it optional?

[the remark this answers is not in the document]
yinyiqian1: I think it's easier to just delete the ledger entry if there's no permission between two accounts. Is there any reason why should we keep it and just leave out the permissions field?

include/xrpl/protocol/detail/transactions.macro
```diff
@@ -447,6 +447,12 @@ TRANSACTION(ttCREDENTIAL_DELETE, 60, CredentialDelete, ({
     {sfCredentialType, soeREQUIRED},
 }))
 
+/** This transaction type delegates authorized account specified permissions */
+TRANSACTION(ttACCOUNT_PERMISSION_SET, 61, AccountPermissionSet, ({
+    {sfAuthorize, soeREQUIRED},
+    {sfPermissions, soeREQUIRED},
```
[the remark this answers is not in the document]
yinyiqian1: Being empty means deleting all the permissions, so that the user won't delete the permissions by accident when they mistakenly leave out the Permissions field. So I think it's better to keep it as soeREQUIRED here in the transaction.

include/xrpl/protocol/detail/ledger_entries.macro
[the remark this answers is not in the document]
gregtatcam: I don't think there is a precedent for this, is there? A transaction is used to delete a ledger entry.

include/xrpl/protocol/detail/transactions.macro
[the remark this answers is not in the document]
gregtatcam: But it's a required field in the transaction, isn't it? It could be an empty array of permissions when it's set. So if it's empty then we delete it from the ledger object. The optional field has to be handled appropriately but it's used in a very few places. So if a user can make the permissions array empty then why is it a problem deleting it from the ledger object?

include/xrpl/protocol/detail/ledger_entries.macro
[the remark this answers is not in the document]
shawnxie999: I don't think there is a hard requirement that a specific txn is needed to delete an object. Ex. unfunded offers can be cancelled implicitly, associated NFT offers are automatically deleted when the NFT itself is deleted (see https://github.com/XRPLF/rippled/pull/4346)

include/xrpl/protocol/detail/transactions.macro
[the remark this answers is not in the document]
gregtatcam: It's kept required in the transaction but optional in the on-ledger object.

include/xrpl/protocol/detail/ledger_entries.macro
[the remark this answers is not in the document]
gregtatcam: Can delete it then.

src/xrpld/app/tx/detail/Transactor.cpp
yinyiqian1: yes. correct. I'll update it anyway. By the way, I might need to take care of mPriorBalance in the later usage, for example, in `Payment.cpp`, `mPriorBalance` is meant to be used as a effective account balance, it's currently tracking the sender's balance, when it's delegated, we need to get the delegating account's balance.

yinyiqian1: `account_` is the delegating account if it's a delegated transaction. And `mSourceBalance` is owned by the effective account, which means if it is a delegated transaction, `mSourceBalance` is tracking the delegating account.

src/xrpld/app/misc/NetworkOPs.cpp
```diff
@@ -3016,7 +3016,9 @@ NetworkOPsImp::transJson(
 
     if (transaction->getTxnType() == ttOFFER_CREATE)
     {
-        auto const account = transaction->getAccountID(sfAccount);
+        auto const& account = transaction->isFieldPresent(sfOnBehalfOf)
```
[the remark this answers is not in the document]
ckeshava: All transactions could be delegated to other accounts, is that correct?

src/xrpld/app/tx/detail/AMMWithdraw.cpp
```diff
@@ -250,8 +250,7 @@ AMMWithdraw::preclaim(PreclaimContext const& ctx)
     if (auto const ter = checkAmount(amount2, amount2Balance))
         return ter;
 
-    auto const lpTokens =
-        ammLPHolds(ctx.view, *ammSle, ctx.tx[sfAccount], ctx.j);
+    auto const lpTokens = ammLPHolds(ctx.view, *ammSle, accountID, ctx.j);
```
ckeshava, line 253: Shouldn't the initialization of accountID variable be updated? (at the beginning of the preclaim override)

```
auto const accountID = ctx.tx[sfAccount]; // shouldn't this be updated to getEffectiveAccountID() ?
```

yinyiqian1: if this is a delegated transaction, `ctx.tx[sfAccount]`  will return the account who delegates the sender. For example, if A sends the transaction on behalf of B, `ctx.tx[sfAccount]` is actually B. 
`getEffectiveAccountID` is for STTx,  ctx.tx is of `STTxDelegated` type, and `ctx.tx[sfAccount]` is delegated to return the actual account.
STTxDelegated is created to avoid changing the code everywhere to get the account id.

src/xrpld/app/misc/NetworkOPs.cpp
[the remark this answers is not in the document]
yinyiqian1: Yes. but here the original code is only checking the `ttOFFER_CREATE`. 
it is not getting account for other transactions here.

src/test/app/AccountPermission_test.cpp
```diff
@@ -0,0 +1,4463 @@
...
+        testDID(all);
+        testEscrow(all);
+        testMPTokenIssuanceSetGranular(all);
+        testMPToken(all);
+        testNFToken(all);
+        testOracle(all);
+        testPaymentChannel(all);
+        testPayment(all);
+        testPaymentGranular(all);
+        testPayment(all);
```
ckeshava, line 4451: `testPayment(all);` has been invoked twice

src/xrpld/app/misc/NetworkOPs.cpp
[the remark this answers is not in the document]
ckeshava: ah, I missed that, thanks.

The possibility of under-funded offers eludes me sometimes :)

src/xrpld/app/tx/detail/AMMWithdraw.cpp
ckeshava: that makes sense, thanks for the explanation.

include/xrpl/protocol/STTxDelegated.h
```diff
@@ -0,0 +1,256 @@
...
+    {
+        return tx_.getAccountID(sfAccount);
+    }
+
+    std::uint32_t
+    getEffectiveSeq() const
+    {
+        return isDelegated_ ? tx_.getDelegateSeqProxy().value()
+                            : tx_.getSeqProxy().value();
+    }
```
mvadari, line 70: Why not just add these helper functions to `STTx` or `Transactor` directly? Would be much simpler and means that the diff would be smaller.

yinyiqian1: The main purpose of STTxDelegated is that we do not need to change ctx.tx[sfAccout] everywhere to something like `ctx.tx.getEffectiveAccount()` which could easily make mistakes.
Another option is that add directly in STTx but change it everywhere for ctx.tx[sfAccout], I was doing this for my previous implementation but Greg suggested adding the wrapper STTxDelegated. So I changed it into this way.
What do you think? @gregtatcam

src/xrpld/app/tx/detail/MPTokenIssuanceSet.cpp
```diff
@@ -102,9 +102,21 @@ MPTokenIssuanceSet::doApply()
...
+    bool unlock = txFlags & tfMPTUnlock;
+
+    if (ctx_.tx.isDelegated() && !ctx_.permissions.empty())
+    {
+        // if permissions is not empty, granular delegation is happening.
+        if (lock && !ctx_.permissions.contains(MPTokenIssuanceLock))
+            return tecNO_PERMISSION;
+        if (unlock && !ctx_.permissions.contains(MPTokenIssuanceUnlock))
+            return tecNO_PERMISSION;
+    }
```
shawnxie999, line 115: why isn't this checked in preclaim? (same question for other transactors as well)

src/xrpld/app/tx/detail/Payment.cpp
```diff
@@ -341,6 +342,30 @@ Payment::doApply()
 
     AccountID const dstAccountID(ctx_.tx.getAccountID(sfDestination));
     STAmount const dstAmount(ctx_.tx.getFieldAmount(sfAmount));
+
+    // if the transaction is transaction-level delegated, ctx_.permissions
+    // was already cleaned up.
+    if (ctx_.tx.isDelegated() && !ctx_.permissions.empty())
```
shawnxie999, line 348: should prob create a helper function for the logic `ctx_.tx.isDelegated() && !ctx_.permissions.empty()` so that the same code isn't in every transactor

src/xrpld/app/tx/detail/Payment.cpp
```diff
@@ -341,6 +342,30 @@ Payment::doApply()
...
+        // permission. PaymentMint means the sender is the issuer. PaymentBurn
+        // means the destination is the issuer.
+        bool authorized = false;
+        auto const amountIssue = dstAmount.issue();
+        if (isXRP(amountIssue))
+            return tecNO_PERMISSION;
+        if (amountIssue.account == account_ &&
+            ctx_.permissions.contains(PaymentMint))
+            authorized = true;
+        if (amountIssue.account == dstAccountID &&
```
shawnxie999, line 361: what happens when it contains both PaymentBurn and PaymentMint? does such tx make sense? if not, should we make an assertion or return an err code?

src/xrpld/app/tx/detail/Transactor.cpp
```diff
@@ -208,6 +220,52 @@ Transactor::calculateBaseFee(ReadView const& view, STTx const& tx)
     return baseFee + (signerCount * baseFee);
 }
 
+TER
+Transactor::checkPermissions(
+    ReadView const& view,
+    STTx const& tx,
+    std::unordered_set<GranularPermissionType>& permissions)
+{
+    if (!tx.isFieldPresent(sfOnBehalfOf) ||
```
shawnxie999, line 229: do we also check that sfOnBehalfOf is different from sfAccount somewhere?

src/xrpld/app/tx/detail/Transactor.cpp
```diff
@@ -208,6 +220,52 @@ Transactor::calculateBaseFee(ReadView const& view, STTx const& tx)
...
+    if (!sle)
+        return tecNO_PERMISSION;
+
+    auto const permissionArray = sle->getFieldArray(sfPermissions);
+    auto const transactionType = tx.getTxnType();
+
+    for (auto const& permission : permissionArray)
+    {
+        auto const permissionValue = permission[sfPermissionValue];
+        if (permissionValue == transactionType + 1)
```
shawnxie999, line 245: probably shouldn't hardcode the logic `transactionType + 1` here. we should prob have a function named `TransactionPermissionValue(TxType)` that does this and use it across all the places (I believe STParsedJSON is  hardcoded too)

src/xrpld/rpc/detail/TransactionSign.cpp
```diff
@@ -447,6 +447,47 @@ transactionPreProcessImpl(
...
+                keylet::account(delegatingAccount));
+
+        if (verify && !delegatingAccountSle)
+        {
+            // If not offline and did not find on behalf of account, error.
+            JLOG(j.debug())
+                << "transactionSign: Failed to find onBehalfOf account "
+                << "in current ledger: " << toBase58(delegatingAccount);
+
+            return rpcError(rpcSRC_ACT_NOT_FOUND);
```
shawnxie999, line 480: can be refactored

```
if (verify){
    delegatingAccountSle = app.openLedger().current()->read(
        keylet::account(delegatingAccount));

    if (!delegatingAccountSle){
       // If not offline and did not find on behalf of account, error.
       JLOG(j.debug())
           << "transactionSign: Failed to find onBehalfOf account "
           << "in current ledger: " << toBase58(delegatingAccount);

       return rpcError(rpcSRC_ACT_NOT_FOUND);
   }
}

```

src/xrpld/rpc/detail/TransactionSign.cpp
```diff
@@ -447,6 +447,47 @@ transactionPreProcessImpl(
         return rpcError(rpcSRC_ACT_NOT_FOUND);
     }
 
+    AccountID delegatingAccount;
+    std::shared_ptr<SLE const> delegatingAccountSle;
+    if (tx_json.isMember(jss::OnBehalfOf))
+    {
+        auto const onBehalfOf =
```
shawnxie999, line 454: we should prob check if the transaction type is supported as well

src/xrpld/rpc/detail/TransactionSign.cpp
```diff
@@ -465,6 +506,16 @@ transactionPreProcessImpl(
                 hasTicketSeq ? 0 : app.getTxQ().nextQueuableSeq(sle).value();
         }
 
+        if (tx_json.isMember(jss::OnBehalfOf) &&
+            !tx_json.isMember(jss::DelegateSequence))
+        {
+            bool const hasDelegateTicketSeq =
+                tx_json.isMember(sfDelegateTicketSequence.jsonName);
+            tx_json[jss::DelegateSequence] = hasDelegateTicketSeq
+                ? 0
```
shawnxie999, line 515: why is the value of DelegateSequence 0? shouldn't the DelegateSequence  field be left out entirely?

src/xrpld/rpc/detail/TransactionSign.cpp
```diff
@@ -994,6 +1045,19 @@ checkMultiSignFields(Json::Value const& jvRequest)
     if (!tx_json.isMember(jss::Sequence))
         return RPC::missing_field_error("tx_json.Sequence");
 
+    if (tx_json.isMember(jss::OnBehalfOf))
+    {
+        if (!tx_json.isMember(jss::DelegateSequence))
```
shawnxie999, line 1050: if DelegateSequence isn't present, then is DelegateTicketSequence guarenteed to be present?

src/xrpld/rpc/handlers/LedgerEntry.cpp
```diff
@@ -67,6 +67,43 @@ parseAuthorizeCredentials(Json::Value const& jv)
...
+        jvResult[jss::error] = "malformedAccount";
+        return std::nullopt;
+    }
+    auto const authorize =
+        parseBase58<AccountID>(params[jss::authorize].asString());
+    if (!authorize)
+    {
+        jvResult[jss::error] = "malformedAuthorize";
+        return std::nullopt;
+    }
```
shawnxie999, line 103: should double check with clio team for these error msgs. sometimes they may have trouble supporting new error msgs so we should avoid adding new ones if not necessary. the safe approach is to just return  `malformedAddress`

src/libxrpl/protocol/STParsedJSON.cpp
```diff
@@ -360,10 +361,45 @@ parseLeaf(
...
+                            ret = detail::make_stvar<STUInt32>(
+                                field, *granularPermission);
+                    }
+                    else
+                    {
+                        ret = detail::make_stvar<STUInt32>(
+                            field,
+                            beast::lexicalCastThrow<std::uint32_t>(
+                                value.asString()));
+                    }
```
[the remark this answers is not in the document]
shawnxie999: it would create additional burden for SDKs to do field-specific serialization, especially it's only trying to 1 in the case of transaction permission. do we really need to add 1 to the transaction type? what's the reasoning for this?

src/xrpld/app/tx/detail/SetAccount.cpp
shawnxie999, line 1: can the sender account set asfAuthorizedNFTokenMinter on behalf of the delegated account? there is no such check (or do we have unittests that checks it?).

I'm worried that certain features could be slipped by where the sender does not have the perm but still able to modify something. should we consider creating some kind of white list of tx field map that limits what the delegating account can modify? For ex:

For SetAccount, it would map to only valid fields (like sfSetFlag, sfMessageKey, sfEmailHash, etc), and functions/lambdas that checks the validity of the permissions. It can look like

`permRules = {SetAccount, {{CommonFields, {}}, {sfSetFlag, checkFlagFunc}, {sfMessageKey, checkMessageKeyFunc}, {sfEmailHash, checkEmailHashFunc}}`

Then we can easily iterate the tx and compare if specific fields are allowed.

I think it would be cleaner and safer to have the checks in one place rather than scattered in different places which could be risky. what do you think?

src/xrpld/app/tx/detail/Transactor.cpp
```diff
@@ -889,13 +1038,28 @@ Transactor::reset(XRPAmount fee)
...
     XRPL_ASSERT(
         isTesSuccess(ter), "ripple::Transactor::reset : result is tesSUCCESS");
 
     if (isTesSuccess(ter))
         view().update(txnAcct);
 
+    if (ctx_.tx.isDelegated())
+    {
+        auto const accountDelegating = ctx_.tx.getAccountID(sfAccount);
+        auto const sleDelegating =
```
shawnxie999, line 1051:
```
  if (!sleDelegating)
        return {tefINTERNAL, beast::zero};
```

should check that sle exists

src/xrpld/app/tx/applySteps.h
```diff
@@ -163,7 +165,7 @@ struct PreflightResult
 {
 public:
     /// From the input - the transaction
-    STTx const& tx;
+    STTxDelegated const tx;
```
shawnxie999, line 168:
```suggestion
    STTxDelegated const& tx;
```

src/xrpld/app/tx/applySteps.h
```diff
@@ -208,26 +209,33 @@ struct PreclaimResult
     /// From the input - the ledger view
     ReadView const& view;
     /// From the input - the transaction
-    STTx const& tx;
+    STTxDelegated const tx;
```
shawnxie999, line 212:
```suggestion
    STTxDelegated const& tx;
```

src/xrpld/app/tx/applySteps.h
```diff
@@ -208,26 +209,33 @@ struct PreclaimResult
...
     /// From the input - the flags
     ApplyFlags const flags;
     /// From the input - the journal
     beast::Journal const j;
 
     /// Intermediate transaction result
     TER const ter;
+    /// granular permissions enabled for the transaction. If empty and
+    /// isDelegated=true, then the entire transaction is authorized.
+    std::unordered_set<GranularPermissionType> permissions;
```
shawnxie999, line 222:
```suggestion
    std::unordered_set<GranularPermissionType> const permissions;
```

shawnxie999:
Just a heads up I believe this PR would have some conflicts with the Batch feature, specifically with the account sequences.

include/xrpl/protocol/STTxDelegated.h
```diff
@@ -0,0 +1,256 @@
...
+class STTxDelegated
+{
+private:
+    const STTx& tx_;    // Wrap an instance of STTx
+    bool isDelegated_;  // if the transaction is delegated by another account
+
+public:
+    explicit STTxDelegated(STTx const& tx, bool isDelegated)
+        : tx_(tx), isDelegated_(isDelegated)
+    {
```
shawnxie999, line 44: should assert that isDelegated can only be true if onBehalfOf is present

include/xrpl/protocol/STTxDelegated.h
```diff
@@ -0,0 +1,256 @@
...
+// amendment. It wraps STTx, and delegates to STTx methods. The key change is
+// getAccountID and operator[]. We need to first check if the transaction is
+// delegated by another account by checking if the sfOnBehalfOf field is
+// present. If it is present, we need to return the sfOnBehalfOf field as the
+// account when calling getAccountID(sfAccount) and tx[sfAccount].
+class STTxDelegated
+{
+private:
+    const STTx& tx_;    // Wrap an instance of STTx
+    bool isDelegated_;  // if the transaction is delegated by another account
```
shawnxie999, line 39:
```suggestion
    bool const isDelegated_;  // if the transaction is delegated by another account
```

include/xrpl/protocol/jss.h
```diff
@@ -55,10 +55,13 @@ JSS(AssetClass);           // in: Oracle
 JSS(AssetPrice);           // in: Oracle
 JSS(AuthAccount);          // in: AMM Auction Slot
 JSS(AuthAccounts);         // in: AMM Auction Slot
+JSS(Authorize);            // in: account_permission
 JSS(BaseAsset);            // in: Oracle
 JSS(BidMax);               // in: AMM Bid
 JSS(BidMin);               // in: AMM Bid
 JSS(ClearFlag);            // field.
+JSS(DelegateTicketSequence); // in: Delegate transactions
+JSS(DelegateSequence);     // in: Delegate transactions
```
shawnxie999, line 64: we dont need these here if already defined in `sfield.macro`. the JSON field can be called using `sfDelegateTicketSequence.jsonName` and `sfDelegateSequence.jsonName`

include/xrpl/protocol/jss.h
```diff
@@ -76,9 +79,13 @@ JSS(LPTokenOut);           // in: AMM Liquidity Provider deposit tokens
 JSS(LPTokenIn);            // in: AMM Liquidity Provider withdraw tokens
 JSS(LPToken);              // out: AMM Liquidity Provider tokens info
 JSS(OfferSequence);        // field.
+JSS(OnBehalfOf);           // in: AccountPermission
```
shawnxie999, line 82: ditto. not needed, can be called using sfOnBehalfOf.jsonName

include/xrpl/protocol/jss.h
```diff
@@ -76,9 +79,13 @@ JSS(LPTokenOut);           // in: AMM Liquidity Provider deposit tokens
...
 JSS(OracleDocumentID);     // field
 JSS(Owner);                // field
 JSS(Paths);                // in/out: TransactionSign
+JSS(Permission);           // in: AccountPermission
+JSS(Permissions);          // in: AccountPermission
+JSS(PermissionValue);      // in: AccountPermission
```
shawnxie999, line 88: ditto. dont need to define these i think

src/xrpld/app/tx/detail/MPTokenIssuanceSet.cpp
```diff
@@ -102,9 +102,21 @@ MPTokenIssuanceSet::doApply()
     std::uint32_t const flagsIn = sle->getFieldU32(sfFlags);
     std::uint32_t flagsOut = flagsIn;
 
-    if (txFlags & tfMPTLock)
+    bool lock = txFlags & tfMPTLock;
```
gregtatcam, line 105: should be `bool const` in both cases.

src/xrpld/app/tx/detail/Transactor.cpp
```diff
@@ -312,6 +375,8 @@ Transactor::checkSeqProxy(
     SeqProxy const t_seqProx = tx.getSeqProxy();
     SeqProxy const a_seq = SeqProxy::sequence((*sle)[sfSequence]);
 
+    // check delegate seq
```
gregtatcam, line 378: Is this comment relevant?

src/libxrpl/protocol/Permissions.cpp
```diff
@@ -0,0 +1,147 @@
...
+        {PaymentBurn, ttPAYMENT},
+        {MPTokenIssuanceLock, ttMPTOKEN_ISSUANCE_SET},
+        {MPTokenIssuanceUnlock, ttMPTOKEN_ISSUANCE_SET}};
+
+    // We do not allow delegating AccountSet, SetRegularKey, SignerListSet,
+    // AccountPermissionSet and AccountDelete for security reason. To support
+    // more newly added transactions, we need to add it here and have it fully
+    // tested. Newly addded transactions are not supported for delegation
+    // automatically for safety.
+    supportedTransactions = {
```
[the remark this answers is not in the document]
shawnxie999: according to spec https://github.com/XRPLF/XRPL-Standards/discussions/217

> Adding a new transaction type to the XRPL will automatically be supported by any feature that uses these permissions.

I think it's saying new transactions get automatically supported by permissions? if my understanding is right, then it's probably easier to maintain an "exception" list as opposed to an "allow" list, as it's going to be pretty easy to forget to add the new transaction to the list.

@mvadari can you confirm if new future transactions should be supported automatically?

src/libxrpl/protocol/STParsedJSON.cpp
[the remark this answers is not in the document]
shawnxie999: resolving this as it's not longer relevant. current implementation is good.

src/libxrpl/protocol/Permissions.cpp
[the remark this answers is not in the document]
yinyiqian1: It's risky that new transactions are automatically supported. It's possible that other developers are unaware of account permission and call the incorrect function to get account and sequence number. And it's safer to enable newly added transactions for account permission in new amendments so that we can have the code fully tested.

src/xrpld/app/tx/detail/CashCheck.cpp
```diff
@@ -138,7 +138,7 @@ CashCheck::preclaim(PreclaimContext const& ctx)
         STAmount const value{[](STTx const& tx) {
             auto const optAmount = tx[~sfAmount];
             return optAmount ? *optAmount : tx[sfDeliverMin];
-        }(ctx.tx)};
+        }(ctx.tx.getSTTx())};
```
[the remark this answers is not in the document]
yinyiqian1: there are two options here, use `ctx.tx.getSTTx()` or pass `STTxDelegated` as the function's parameter.

src/xrpld/app/tx/detail/AMMDeposit.cpp
yinyiqian1: it's refactored to STTxDelegated so this line is not changed now, do you want to add `&` here anyway in this PR?

src/xrpld/app/tx/detail/AMMWithdraw.cpp
yinyiqian1: it's refactored to STTxDelegated so this line is not changed now, do you want to add `&` here anyway in this PR?

src/xrpld/app/tx/detail/CancelCheck.cpp
yinyiqian1: it's refactored to STTxDelegated so this line is not changed now, do you want to add `&` here anyway in this PR?

src/xrpld/app/tx/detail/CancelOffer.cpp
yinyiqian1: it's refactored to STTxDelegated so this line is not changed now, do you want to add `&` here anyway in this PR?

src/xrpld/app/tx/detail/Clawback.cpp
yinyiqian1: it's refactored to STTxDelegated so this line is not changed now, do you want to add `&` here anyway in this PR?

src/xrpld/app/tx/detail/InvariantCheck.cpp
yinyiqian1: It's updated to use STTxDelegated, so I'll resolve this for now.

src/libxrpl/protocol/STTx.cpp
```diff
@@ -197,6 +197,23 @@ STTx::getSeqProxy() const
     return SeqProxy{SeqProxy::ticket, *ticketSeq};
 }
 
+SeqProxy
+STTx::getDelegateSeqProxy() const
+{
+    std::uint32_t const seq{getFieldU32(sfDelegateSequence)};
```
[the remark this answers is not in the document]
yinyiqian1: since this optional field must exist if the code reaches here, so I added an assert for sfDelegateSequence existence.
sfDelegateSequence can exist along with sfDelegateTicketSequence only when sfDelegateSequence is zero. (Same logic as sfSequence and sfTicketSequence). The fields are validated in transactor::checkDelegateSeqProxy.
updated to this:
```
std::optional<std::uint32_t> const delegateSeq{operator[](~sfDelegateSequence)};
XRPL_ASSERT(
        delegateSeq,
        "ripple::STTx::getDelegateSeqProxy : delegateSeq must be present");
```

src/xrpld/rpc/detail/TransactionSign.cpp
yinyiqian1: this is following the same logic as Sequence. DelegateSequence is set to zero when the user did not provide DelegateSequence but provided DelegateTicketSequence. If user wants to use ticket, they can choose to set sequence to 0 or leave out sequence. If sequence is left out, then it's auto-filled to 0. ( refer to `Sequence` field description https://xrpl.org/docs/references/protocol/transactions/common-fields )

src/xrpld/rpc/detail/TransactionSign.cpp
yinyiqian1: This is for multi sign. Sequence must be present. And if it's a delegating transaction, then DelegateSequence must also be present. Although `SignerListSet ` is disabled for AccountPermission, it is possible that a delegating transaction can be multi signed. For example, A can send payment on behalf of B; C can D can multi sign for A. Then the payment on behalf of B can be signed by C and D.

src/xrpld/rpc/handlers/Simulate.cpp
```diff
@@ -144,6 +182,25 @@ autofillTx(Json::Value& tx_json, RPC::JsonContext& context)
...
 
+    if (tx_json.isMember(jss::OnBehalfOf))
+    {
+        if (!tx_json.isMember(jss::DelegateSequence))
+        {
+            auto const delegateSeq =
+                getAutofillDelegateSequence(tx_json, context);
+            if (!delegateSeq)
+                return delegateSeq.error();
+            tx_json[sfDelegateSequence.jsonName] = *delegateSeq;
```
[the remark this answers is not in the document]
yinyiqian1: This is following Sequence's logic, so that using Sequence and DelegateSequence follow same the rule.

dangell7:
```
Added sfDelegateSequence which are optional
to avoid double insertion, when a delegating transaction happens, we have to increment both of the delegating account's sequence and the delegated account's sequence. So sfDelegateSequence is used to track the delegating account's sequence.

Added sfDelegateTicketSequence which are optional
we allow a delegated account to create ticket on behalf of the delegating account of he has the permission. In this case, the delegated account should be able to use the ticket he created on behalf of the delegating account. So sfDelegateTicketSequence is introduced.
```

Why is this PR not just an extension of "Regular Key Type" functionality? Why not just verify the signature of the delegated account and then process everything as per usual? Or if you want to verify both signatures add `sfDelegateSigners` and verify the signatures.

src/xrpld/app/tx/detail/CreateOffer.cpp
```diff
@@ -968,7 +969,7 @@ CreateOffer::applyGuts(Sandbox& sb, Sandbox& sbCancel)
 
     // Note that we we use the value from the sequence or ticket as the
     // offer sequence.  For more explanation see comments in SeqProxy.h.
-    auto const offerSequence = ctx_.tx.getSeqProxy().value();
+    auto const offerSequence = ctx_.tx.getEffectiveSeq();
```
dangell7, line 972: Does this mean that the delegate is the owner of the offer? Why? Or is the owner the submitting account but the sequence was created with the delegate? Why?

include/xrpl/protocol/STTxDelegated.h
yinyiqian1: @ximinez Do you have any preference?

src/xrpld/app/tx/detail/CreateOffer.cpp
yinyiqian1: I'm going to change to another implementation, something like:
```
{
  TransactionType: Payment,
  Account: "rDELEGATING..."
  Sequence: [rDELEGATING's sequence]
  DelegatedAccount: "rDELEGATED..."
  SigningPubKey: [rDELEGATED's public key]
  Signature: [rDELEGATED's signature]
}
```
Another PR will be open and I'll close this PR. Also the spec XLS-75d will be updated.

yinyiqian1: Hi @dangell7  ,  both the owner and the sequence will be the actual account instead of the submitting account. (But we will adopt a totally different spec and implementation, so I'll create a new PR on this)

Here's the explanation for the current implementation:
Background: A submit a transaction of `CreateOffer` of behalf of B. In the request, `account` is A(delegated account), `onBehalfOf` is B (delegating account, and also the actual account).
According to the spec XLS-75d, https://github.com/XRPLF/XRPL-Standards/discussions/218
The transaction is like:
```
{
    Transaction: "CreateOffer",
    Account: "DelegatedAccountID",    // account A
    OnBehalfOf: "DelegatingAccountID". // account B
     ...
}
```
Here the `Account` field is A, which is the sender of the transaction. B will be the owner of the offer. So I need to use B's account and B's sequence to create the offer.

I wrapped `STTx` in a wrapper, so now `ctx_.tx` is of type `STTxDelegated`. As a result, `ctx.tx[sfAccount]` will return B instead of A. Although A initiated the transaction on behalf of B, B is the actual owner. Therefore, we need to use B's sequence to create the object. So here the `getEffectiveSeq()` is returning B's sequence, not the sender's sequence.

Reason for this: According to the spec spec XLS-75d, I need to keep the `Account` field to be the sender, which is the account being delegated. The `onBehalfOf` is the actual account. That's why I need to find a way to return the correct account.
So the wrapper class `STTxDelegated` is introduced and `[]` is overridden. Basically for normal transactions, it returns the `Account`. But for delegating transactions which contains `OnbehalfOf`, it returns `OnBehalfOf` instead.
```
    template <class T>
        requires(!std::is_same<TypedField<T>, SF_ACCOUNT>::value)
    typename T::value_type
    operator[](TypedField<T> const& f) const
    {
        return tx_[f];
    }

    // When Type is SF_ACCOUNT and also field name is sfAccount, we need to
    // check if the transaction is delegated by another account. If it is,
    // return sfOnBehalfOf field instead.
    template <class T>
        requires std::is_same<TypedField<T>, SF_ACCOUNT>::value
    AccountID
    operator[](TypedField<T> const& f) const
    {
        if (f == sfAccount)
            return tx_.isFieldPresent(sfOnBehalfOf) ? *tx_[~sfOnBehalfOf]
                                                    : tx_[sfAccount];
        return tx_[f];
    }
```

The complexity is because the Account field refers to the delegated account, rather than the delegating account, which is the actual owner of the offer.

To avoid the complexity,
We plan to make significant change to make the `Account` field being the delegating account, in this new design, `Account` will be B instead of A. And A is given in another field `DelegatedAccount`. And I'll verify that `DelegatedAccount` (instead of `Account`) matches the id generated from SigningPubKey. So you will expect huge change in both implementation and spec definition.
```
{
  TransactionType: Payment,
  Account: "rDELEGATING..."
  Sequence: [rDELEGATING's sequence]
  DelegatedAccount: "rDELEGATED..."
  SigningPubKey: [rDELEGATED's public key]
  Signature: [rDELEGATED's signature]
}
```

dangell7: Glad to hear you are taking a different approach, if you would like help please feel free to reach out to me.

yinyiqian1: I think it is more clear to add `DelegatedAccount` to indicate a delegating transaction. And I'll check the permissions before the delegated account signing the transaction. So basically the request only provides:
```
{
TransactionType: Payment,
Account: "rDELEGATING..."
Sequence: [rDELEGATING's sequence]
DelegatedAccount: "rDELEGATED..."
}
```
It only adds `DelegatedAccount` field, and then signed by the delegated account.

src/xrpld/app/tx/detail/Payment.cpp
yinyiqian1: it will be the issuer sends token to the issuer itself, 
it's checked in payment::preflight
```
if (account == dstAccountID && equalTokens(srcAsset, dstAsset) && !hasPaths)
    {
        // You're signing yourself a payment.
        // If hasPaths is true, you might be trying some arbitrage.
        JLOG(j.trace()) << "Malformed transaction: "
                        << "Redundant payment from " << to_string(account)
                        << " to self without path for " << to_string(dstAsset);
        return temREDUNDANT;
    }
```

src/xrpld/app/tx/detail/SetAccount.cpp
yinyiqian1: setting and clearing any flag are prohibited. There's some unit test covering setting and clearing flags returns `tecNO_PERMISSION`.
I put the logic in the `checkPermission` function instead of `doApply()` in the new PR https://github.com/XRPLF/rippled/pull/5354
But this `checkPermission` function still needs to be taken care of when new logic are added to `doApply`.

PR #5165, opened 2024-10-26
Add vl.xrpl.vision to validators-example.txt

xrplvision:
Add the list provider vl.xrpl.vision (xrpl.vision) to validators-example.txt and include the keys.


Bronek:
I recommend waiting until https://github.com/XRPLF/rippled/pull/5112 is merged **and released** ; without this change, any additional UNL publisher will reduce the security of the UNL system

PR #5166, opened 2024-10-29
Add hubs.xrpkuwait.com to bootstrap

alloynetworks:
@Xrpkuwait is a committed contributor to XRPL infrastructure. His hubs will provide a much needed extra set of bootstrap servers. They have shown good uptime and capacity.

### DIFF
--- a/src/xrpld/overlay/detail/OverlayImpl.cpp
+++ b/src/xrpld/overlay/detail/OverlayImpl.cpp
@@ -492,7 +492,7 @@ OverlayImpl::start()
         // Pool of servers operated by ISRDC - https://isrdc.in
         bootstrapIps.push_back("sahyadri.isrdc.in 51235");
 
-        //Pool of servers operated by @Xrpkuwait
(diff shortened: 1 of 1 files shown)


Xrpkuwait:
Thank you.

I confirm my commitment to operate the **hubs.xrpkuwait.com** servers for a minimum of 2 years.
**XRPKuwait.com** is an XRPL infrastructure provider based in Kuwait, operating 7 nodes, including public hubs and a validator.

ximinez (changes requested):
Please also update the default list of servers in the example config https://github.com/XRPLF/rippled/blob/develop/cfg/rippled-example.cfg#L417-L419

ximinez:
> > Please also update the default list of servers in the example config https://github.com/XRPLF/rippled/blob/develop/cfg/rippled-example.cfg#L417-L419
> 
> Added via #5167

Sorry, my intention was that both changes would be included in this PR. It's one change that should be merged together, specifically using the "Squash and merge" button to put the changes into a single commit.

alloynetworks:
Combined with 5167 in 5169

PR #5167, opened 2024-10-29
Add hubs.xrpkuwait.com to example rippled.cfg

alloynetworks:
Comment added to indicate that hubs.xrpkuwait.com is also in the bootstrap list


ximinez:
See https://github.com/XRPLF/rippled/pull/5166#issuecomment-2445019170

alloynetworks:
Combined with 5166 in 5169

PR #5120, opened 2024-09-06
Retry RPC commands to try to fix MacOS CI jobs

ximinez:
## High Level Overview of Change

Retries failed RPC connection timeouts up to 3 times, unless the test is expecting a timeout. Geared toward fixing MacOS test run failures. 

Also remove an orphaned free function: `getAccounts`.

### Context of Change

MacOS CI jobs have been failing spuriously for a while due to timeouts. This seems to resolve the issue.

### Type of Change

- [X] Tests (you added tests for code that already exists, or your new feature included in this PR)

### API Impact

None

PR #5171, opened 2024-10-30
DRAFT: test: Retry all RPC commands to fix MacOS CI jobs

ximinez:

Retry all failed RPC connections / commands in unit tests.

### Context of Change

Follow up to #5120 (23991c9), which added a retry for submit commands. It improved MacOS test reliability, but other tests are failing now.

### Type of Change

- [X] Tests (you added tests for code that already exists, or your new feature included in this PR)


src/test/jtx/impl/Env.cpp
```diff
@@ -562,12 +563,37 @@ Env::ust(JTx const& jt)
...
-        .second;
+    // We shouldn't need to retry, but it fixes the test on macOS for
+    // the moment.
+    if (!cb)
+    {
+        static auto const defaultCallback = [](Json::Value const& jr) {
+            auto const parsedResult = parseResult(jr, false);
+            return parsedResult.rpcCode != rpcINTERNAL;
+        };
+        cb = defaultCallback;
```
thejohnfreeman, line 579: Can we move this out of the function, into the definition of `useDefaultCallback` (either inline or extern), and rename `useDefaultCallback` to something like `rejectInternalError`? If you prefer to frame the predicate names as `accept{Condition}` instead of `reject{Condition}`, then I'd be fine with that too.

src/test/jtx/Env.h
```diff
@@ -292,6 +292,17 @@ class Env
         The command is examined and used to build
         the correct JSON as per the arguments.
     */
+    using rpcCallback = std::function<bool(Json::Value const&)>;
+    const rpcCallback useDefaultCallback{};
+    const rpcCallback expectInternalRPCError = [](Json::Value const&) {
```
thejohnfreeman, line 297: Can we rename this to `rejectNever` and add a comment explaining that it is meant to be used for tests that expect an internal error and do not want to trigger a retry?

src/test/jtx/Env.h
```diff
@@ -292,6 +292,17 @@ class Env
...
+        return true;
+    };
+    rpcCallback
+    getInternalFailureCallback(bool shouldFail)
+    {
+        return shouldFail ? expectInternalRPCError : useDefaultCallback;
+    }
```
thejohnfreeman, line 304: Can we rename this to `rejectInternalErrorIf`?

src/test/jtx/Env.h
```diff
@@ -514,7 +536,7 @@ class Env
     /** Gets the TER result and `didApply` flag from a RPC Json result object.
      */
     static ParsedResult
-    parseResult(Json::Value const& jr);
+    parseResult(Json::Value const& jr, bool checkTER);
```
thejohnfreeman, line 539: I don't quite understand what this parameter is supposed to be doing. Can we add an explanation in the docstring?

thejohnfreeman (changes requested):

src/test/jtx/Env.h
```diff
@@ -292,6 +292,17 @@ class Env
         The command is examined and used to build
         the correct JSON as per the arguments.
     */
+    using rpcCallback = std::function<bool(Json::Value const&)>;
```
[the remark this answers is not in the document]
ximinez: Yes. Fixed.

src/test/jtx/impl/Env.cpp
ximinez: Yes. Fixed.

src/test/jtx/Env.h
ximinez: Yes. Fixed.

src/test/jtx/Env.h
ximinez: A few more checks, and I was able to get rid of `checkTER` entirely. The intention was to flag whether the request was related to submitting a transaction, and thus a `TER` would be expected in the result. There is no `TER` in the result if there isn't a transaction.

ximinez:
@thejohnfreeman @vlntb I want to bring your attention to draft PR #5180. It builds on top of this PR, but adds three commits:
1. The MacOS CI jobs have 4 configurations: The combinations of release vs debug, and unity vs non-unity.
2. The `Env` tests cycle through a range of ports so that if the OS is slow to close a port when one test ends, the next test doesn't immediately try to open it. A general `retry` operation is added so that lower-level network issues can also be retried, not just failed RPC commands.
3. The MacOS CI jobs use parallel runners! (It only uses half the available CPU threads, because using all was still not reliable, but it's better than nothing!)

I would like to roll those changes into this PR so we can get the MacOS jobs fixed all in one shot. Do you think that's a good idea, or should that be a separate PR to be merged separately sometime after this one?

thejohnfreeman:
I don't see any approvals in here yet, so I say roll them into this PR.

thejohnfreeman (approved):

ximinez:
Closed in favor of #5196

PR #5172, opened 2024-10-31
Fix tokens comparison in Payment for temREDUNDANT error

gregtatcam:

Fix semantics of tokens comparison in the payment transactor when checking for temREDUNDANT.
This check prevents an account sending tokens to self.
Current code checks for equal tokens (currency or MPTID) AND the issuers. It should only check
if currency or MPTID are equal.

### Type of Change

- [ ] Documentation update
- [ ] Release

### DIFF
--- a/include/xrpl/protocol/Asset.h
+++ b/include/xrpl/protocol/Asset.h
@@ -199,6 +199,9 @@ validJSONAsset(Json::Value const& jv);
 Asset
 assetFromJson(Json::Value const& jv);
 
+Json::Value
+to_json(Asset const& asset);
+
 }  // namespace ripple
 
 #endif  // RIPPLE_PROTOCOL_ASSET_H_INCLUDED
--- a/include/xrpl/protocol/Asset.h
+++ b/include/xrpl/protocol/Asset.h
@@ -97,6 +97,13 @@ public:
 
     friend constexpr bool
     operator==(Currency const& lhs, Asset const& rhs);
+
+    /** Return true if both asset's issue (Issue or MPTIssue)
+     * are the same and hold the same token (currency or MPTID).
+     * Otherwise return false.
+     */
+    friend constexpr bool
+    equalTokens(Asset const& lhs, Asset const& rhs);
 };
 
 template <ValidIssueType TIss>
@@ -155,6 +162,26 @@ constexpr bool
 operator==(Currency const& lhs, Asset const& rhs)
 {
     return rhs.holds<Issue>() && rhs.get<Issue>().currency == lhs;
+}
+
+constexpr bool
+equalTokens(Asset const& lhs, Asset const& rhs)
+{
+    return std::visit(
+        [&]<typename TLhs, typename TRhs>(
+            TLhs const& issLhs, TRhs const& issRhs) {
+            if constexpr (
+                std::is_same_v<TLhs, Issue> && std::is_same_v<TRhs, Issue>)
+                return issLhs.currency == issRhs.currency;
+            else if constexpr (
+                std::is_same_v<TLhs, MPTIssue> &&
+                std::is_same_v<TRhs, MPTIssue>)
+                return issLhs.getMptID() == issRhs.getMptID();
+            else
+                return false;
+        },
+        lhs.issue_,
+        rhs.issue_);
 }
 
 inline bool
--- a/include/xrpl/protocol/Asset.h
+++ b/include/xrpl/protocol/Asset.h
@@ -98,9 +98,8 @@ public:
     friend constexpr bool
     operator==(Currency const& lhs, Asset const& rhs);
 
-    /** Return true if both asset's issue (Issue or MPTIssue)
-     * are the same and hold the same token (currency or MPTID).
-     * Otherwise return false.
+    /** Return true if both assets refer to the same currency (regardless of
+     * issuer) or MPT issuance. Otherwise return false.
      */
     friend constexpr bool
     equalTokens(Asset const& lhs, Asset const& rhs);
--- a/src/test/app/MPToken_test.cpp
+++ b/src/test/app/MPToken_test.cpp
@@ -1938,6 +1938,49 @@ class MPToken_test : public beast::unit_test::suite
         }
     }
 
+    void
+    testTokensEquality()
+    {
+        using namespace test::jtx;
+        testcase("Tokens Equality");
+        Currency const cur1{to_currency("CU1")};
+        Currency const cur2{to_currency("CU2")};
+        Account const gw1{"gw1"};
+        Account const gw2{"gw2"};
+        MPTID const mpt1 = makeMptID(1, gw1);
+        MPTID const mpt1a = makeMptID(1, gw1);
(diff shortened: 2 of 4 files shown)


sgramkumar:
This patch fixes the test failure 👍

src/test/app/MPToken_test.cpp
```diff
@@ -1938,6 +1938,124 @@ class MPToken_test : public beast::unit_test::suite
...
+            BEAST_EXPECT((mptAmt1 -= mptAmt2) == mptAmt1);
+            BEAST_EXPECT(mptAmt1 == mptAmt2);
+            BEAST_EXPECT(mptAmt1 == 100);
+            BEAST_EXPECT(MPTAmount::minPositiveAmount() == MPTAmount{1});
+        }
+
+        {
+            testcase("Test MPTIssue from/to Json");
+            MPTIssue const issue1{asset1.get<MPTIssue>()};
+            Json::Value const jv = to_json(issue1);
```
kennyzlei, line 2046: nit: can we explicitly check what the contents of `jv` are here?  Otherwise we just have confidence that `to_json` and `mptIssueFromJson` are inverse functions without knowing if the json `jv` was properly created

src/test/app/MPToken_test.cpp
```diff
@@ -1938,6 +1938,124 @@ class MPToken_test : public beast::unit_test::suite
...
+            MPTIssue const issue2 = mptIssueFromJson(jv);
+            BEAST_EXPECT(issue1 == issue2);
+        }
+
+        {
+            testcase("Test Asset from/to Json");
+            Json::Value jv;
```
kennyzlei, line 2053: nit: I think it would be more thorough to test more than just an empty json value here

kennyzlei (approved):
Unit tests lgtm

include/xrpl/protocol/Asset.h
```diff
@@ -97,6 +97,13 @@ class Asset
 
     friend constexpr bool
     operator==(Currency const& lhs, Asset const& rhs);
+
+    /** Return true if both asset's issue (Issue or MPTIssue)
+     * are the same and hold the same token (currency or MPTID).
+     * Otherwise return false.
```
thejohnfreeman, line 103: I'm suggesting a change to the whole string, but wanted to note the original string was one of the rare occasions you'd need the plural possessive "assets'" (apostrophe _after_ the s). You'd need to pluralize "issues" too.

```suggestion
    /** Return true if both assets refer to the same currency (regardless of issuer) or MPT issuance.
     * Otherwise return false.
```

src/test/app/MPToken_test.cpp
```diff
@@ -1938,6 +1938,126 @@ class MPToken_test : public beast::unit_test::suite
...
+            STAmount mptOverflow{asset2, 3037000500ull};
+            try
+            {
+                res = multiply(mptOverflow, mptOverflow, asset3);
+                fail("should throw runtime exception 1");
+            }
+            catch (std::runtime_error const&)
+            {
+                pass();
+            }
```
ximinez, line 2017: Capture the error here and check that the `what()` is what you're expecting.
```suggestion
            catch (std::runtime_error const& e)
            {
                BEAST_EXPECTS(e.what() == "MPT value overflow", e.what());
            }
```

src/test/app/MPToken_test.cpp
```diff
@@ -1938,6 +1938,126 @@ class MPToken_test : public beast::unit_test::suite
...
+            try
+            {
+                res = multiply(
+                    STAmount{asset1, (2ull << 32) + 2}, mptOverflow, asset3);
+                fail("should throw runtime exception 2");
+            }
+            catch (std::runtime_error const&)
+            {
+                pass();
+            }
```
ximinez, line 2029: Same here.
```suggestion
            catch (std::runtime_error const& e)
            {
                BEAST_EXPECTS(e.what() == "MPT value overflow", e.what());
            }
```

src/test/app/MPToken_test.cpp
```diff
@@ -1938,6 +1938,126 @@ class MPToken_test : public beast::unit_test::suite
...
+        }
+
+        {
+            testcase("Test MPTAmount arithmetics");
+            MPTAmount mptAmt1{100};
+            MPTAmount const mptAmt2{100};
+            BEAST_EXPECT((mptAmt1 += mptAmt2) == MPTAmount{200});
+            BEAST_EXPECT((mptAmt1 -= mptAmt2) == mptAmt1);
```
ximinez, line 2037: If the `+=` and `-=` operators are badly messed up, and return the summed value without actually modifying the amount, these tests will still pass. Even though that's far fetched, it doesn't hurt to check the value result explicitly.
```suggestion
            BEAST_EXPECT((mptAmt1 += mptAmt2) == MPTAmount{200});
            BEAST_EXPECT(mptAmt1 == 200);
            BEAST_EXPECT((mptAmt1 -= mptAmt2) == mptAmt1);
```

src/test/app/MPToken_test.cpp
```diff
@@ -1938,6 +1938,126 @@ class MPToken_test : public beast::unit_test::suite
...
+                jv[jss::mpt_issuance_id] == to_string(asset1.get<MPTIssue>()));
+            BEAST_EXPECT(issue1 == mptIssueFromJson(jv));
+        }
+
+        {
+            testcase("Test Asset from/to Json");
+            Json::Value const jv = to_json(asset1);
+            BEAST_EXPECT(
+                jv[jss::mpt_issuance_id] == to_string(asset1.get<MPTIssue>()));
+            BEAST_EXPECT(asset1 == assetFromJson(jv));
```
ximinez, line 2057: Could we also explicitly check the contents of `jv` here?

ximinez (changes requested):

ximinez:
Also, I notice the linux CI builds are failing.

include/xrpl/protocol/Asset.h
```diff
@@ -97,6 +97,12 @@ class Asset
 
     friend constexpr bool
     operator==(Currency const& lhs, Asset const& rhs);
+
+    /** Return true if both assets refer to the same currency (regardless of
+     * issuer) or MPT issuance. Otherwise return false.
+     */
+    friend constexpr bool
+    equalTokens(Asset const& lhs, Asset const& rhs);
```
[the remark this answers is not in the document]
gregtatcam: Well, MPT is not a currency. I thought that `token` better reflects both Currency and MPT.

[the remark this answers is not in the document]
ximinez: Fair enough.

src/test/app/MPToken_test.cpp
gregtatcam: I'm not following. We checked above for `jv[jss::mpt_issuance_id]`.

gregtatcam:
> Also, I notice the linux CI builds are failing.

Thanks, I'm investigating.

ximinez (approved):

thejohnfreeman (approved):

PR #5176, opened 2024-11-01
Add fixAMMv1_2 amendment

gregtatcam:

Add fixAMMv1_2 amendment:

- It adds reserve check on AMMWithdraw
- It generates AMM max offer if changeSpotPriceQuality() fails

### Type of Change

- [ ] Documentation update
- [ ] Release

## Test Plan

Added unit test to check reserve requirements on AMMWithdraw.

### DIFF
--- a/src/xrpld/app/tx/detail/AMMWithdraw.cpp
+++ b/src/xrpld/app/tx/detail/AMMWithdraw.cpp
@@ -472,6 +472,7 @@ AMMWithdraw::withdraw(
         tfee,
         FreezeHandling::fhZERO_IF_FROZEN,
         isWithdrawAll(ctx_.tx),
+        mPriorBalance,
         j_);
     return {ter, newLPTokenBalance};
 }
@@ -490,6 +491,7 @@ AMMWithdraw::withdraw(
     std::uint16_t tfee,
     FreezeHandling freezeHandling,
     WithdrawAll withdrawAll,
+    XRPAmount const& priorBalance,
     beast::Journal const& journal)
 {
     auto const lpTokens = ammLPHolds(view, ammSle, account, journal);
@@ -594,9 +596,9 @@ AMMWithdraw::withdraw(
     auto sufficientReserve = [&](Issue const& issue) -> TER {
         if (!enabledFixAMMv1_2 || isXRP(issue))
             return tesSUCCESS;
-        if (!view.exists(keylet::line(account_, issue)))
+        if (!view.exists(keylet::line(account, issue)))
         {
-            auto const sleAccount = view.read(keylet::account(account_));
+            auto const sleAccount = view.read(keylet::account(account));
             if (!sleAccount)
                 return tecINTERNAL;  // LCOV_EXCL_LINE
             auto const balance = (*sleAccount)[sfBalance].xrp();
@@ -607,14 +609,14 @@ AMMWithdraw::withdraw(
                 (ownerCount < 2) ? XRPAmount(beast::zero)
                                  : view.fees().accountReserve(ownerCount + 1));
 
-            if (std::max(mPriorBalance, balance) < reserve)
+            if (std::max(priorBalance, balance) < reserve)
                 return tecINSUFFICIENT_RESERVE;
         }
         return tesSUCCESS;
     };
 
     if (auto const err = sufficientReserve(amountWithdrawActual.issue()))
-        return {err, STAmount{}};
+        return {err, STAmount{}, STAmount{}, STAmount{}};
 
     // Withdraw amountWithdraw
     auto res = accountSend(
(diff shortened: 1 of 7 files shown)


thejohnfreeman (approved):

ximinez (approved):
LGTM

thejohnfreeman:
Normally these checks are supposed to use the "prior balance", i.e. the balance of the account before the transaction started executing. I just want to note that we're making an exception here. In fact, it would be easier if we could ignore the "prior balance" everywhere. The context for most helper functions is just the `ApplyView`, which does not track the "prior balance" of any account. (The transaction function in this PR is not a helper, but one day it might be.) Helpers that require a prior balance demand that it be drilled through the caller as a parameter.

src/xrpld/app/tx/detail/AMMWithdraw.cpp
```diff
@@ -552,6 +552,29 @@ AMMWithdraw::withdraw(
...
+        {
+            auto const sleAccount = view.read(keylet::account(account_));
+            if (!sleAccount)
+                return tecINTERNAL;  // LCOV_EXCL_LINE
+            auto const balance = (*sleAccount)[sfBalance].xrp();
+            auto const reserve =
+                view.fees().accountReserve((*sleAccount)[sfOwnerCount] + 1);
+
+            if (balance < reserve)
+                return tecINSUFFICIENT_RESERVE;
```
ximinez, line 570: I hate to do this after approving, but [John's comment](https://github.com/XRPLF/rippled/pull/5176#issuecomment-2455233999) raises an important point.

The reason most reserve checks use `mPriorBalance` is because that is the account's balance _before_ fees are deducted. Unless the transaction causes the account to receive XRP, that value _will_ be higher (or equal in one edge case) than the current `sfBalance` value. This is important because an account is allowed to dip into its reserves to pay fees, so we need to examine the account's balance before those fees are paid. AFAIK, this is the only transaction that can cause the transaction-sending account to receive XRP _and_ create a trustline.

This transaction is different from all(?) other transactions, though, in that it is possible for an account to receive *both* XRP *and* an IOU, which could create a trustline.

Consider a couple of scenarios (assuming account reserve is 20XRP, and owner reserve is 5XRP):
1. An account has exactly 30XRP, and a USD trustline. It submits an `AMMWithdraw` with 10 drop fee to get equal amounts of USD and EUR. Using the current rule, that transaction will fail, because the balance will be just under 30XRP. But if the account had submitted a `TrustSet` for EUR, that would have succeeded, and then the account could submit the same `AMMWIthdraw`. If we're going to allow `AMMWithdraw` to create trustlines, they should follow the same rules.
2. An account has exactly 25XRP, and no trustlines. It submits an `AMMWithdraw` with 10 drop fee to get 1000 XRP and 500 USD. Using the current rule, that transaction will succeed, because the balance will be 1025 XRP-10 drops which means we can't _just_ check the `mPriorBalance`.

Fortunately, I think the fix is simple:
```suggestion
            if (max(mPriorBalance, balance) < reserve)
                return tecINSUFFICIENT_RESERVE;
```

That also makes me realize this change neglects one other rule: The first two trust lines shouldn't count against the account's reserve, so in my example above, the transaction should succeed, even if the account only had 20XRP. I'll address that in a separate comment.

src/xrpld/app/tx/detail/AMMWithdraw.cpp
```diff
@@ -552,6 +552,29 @@ AMMWithdraw::withdraw(
...
+        if (!enabledFixAMMv1_2 || isXRP(issue))
+            return tesSUCCESS;
+        if (!view.exists(keylet::line(account_, issue)))
+        {
+            auto const sleAccount = view.read(keylet::account(account_));
+            if (!sleAccount)
+                return tecINTERNAL;  // LCOV_EXCL_LINE
+            auto const balance = (*sleAccount)[sfBalance].xrp();
+            auto const reserve =
+                view.fees().accountReserve((*sleAccount)[sfOwnerCount] + 1);
```
ximinez, line 567: This should use the [same reserve logic as `SetTrust`](https://github.com/ximinez/rippled/blob/8d0b19cd6ed98e2e96d37e46254c111546761877/src/xrpld/app/tx/detail/SetTrust.cpp#L207-L229).
```suggestion
            std::uint32_t const ownerCount = sle->at(sfOwnerCount);

            // See also SetTrust::doApply()
            XRPAmount const reserve(
                (ownerCount < 2) ? XRPAmount(beast::zero)
                          : view().fees().accountReserve(ownerCount + 1));
```
It might be worth refactoring this into a helper function because the same logic is used in MPT, too.

ximinez (changes requested):

ximinez (approved):

